PR #328, opened 2018-11-08
Squid crashes when ICAPS and a sslcrtvalidator used together

chtsanti:
Squid hits an assertions when tries to remove an comm_close handler from
the already closed connection object.

### DIFF
--- a/src/adaptation/icap/Xaction.cc
+++ b/src/adaptation/icap/Xaction.cc
@@ -741,7 +741,7 @@ Adaptation::Icap::Xaction::handleSecuredPeer(Security::EncryptorAnswer &answer)
     securer = NULL;
 
     if (closer != NULL) {
-        if (answer.conn != NULL && answer.conn->isOpen())
+        if (Comm::IsConnOpen(answer.conn))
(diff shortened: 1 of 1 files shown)


src/adaptation/icap/Xaction.cc
```diff
@@ -741,7 +741,7 @@ Adaptation::Icap::Xaction::handleSecuredPeer(Security::EncryptorAnswer &answer)
     securer = NULL;
 
     if (closer != NULL) {
-        if (answer.conn != NULL)
+        if (answer.conn != NULL && answer.conn->isOpen())
```
yadij: Please do not use NULL in new or changed lines.

```suggestion
        if (answer.conn && answer.conn->isOpen())
```

yadij: or better use the Comm::IsConnOpen test:

```suggestion
        if (Comm::IsConnOpen(answer.conn))
```

chtsanti: I used the second form.

yadij:
Sorry this one got overlooked for so long. Once Anubis has done its thing I will add it to the backport queue for 4.5 release.

PR #330, opened 2018-11-10
Fix tls-min-version= being ignored

yadij:
PeerOptions::parse() called parseOptions() only when 'options='
altered sslOptions instead of delaying the parse to context
creation. This missed the fact that for GnuTLS the tlsMinVersion
was also updating the sslOptions string rather than the
parsedOptions variable later in the configuration process.

Call parseOptions() to reset the parsedOptions value whenever
sslOptions string is altered.


yadij:
This previous comment does not compute. There is no pretense at parsing - the string *is* parsed into a binary representation stored in parsedOptions.

The OpenSSL representation being a bitmap is just much easier to make arbitrary ad-hoc changes to. The GnuTLS one is an opaque/private data structure so we cannot easily emulate that ad-hoc changing without an update function. I am still looking for such a function in the GnuTLS API to resolve that XXX, for now we are forced to re-parse into a fresh gnutls_priority_st data structure each time the sslOptions string is altered.

An alternative design would be to keep separate binary representations of the tls-options and the tls-min-version settings then combine them when updating context or session state. That would be more complex than what we have now and still require some way to be found to merge the gnutls_priority_t representations - so the XXX would just move around the code.

They way we have it now we are at least hiding these nasty issues behind the libsecurity API where it can be eventually fixed without refactoring the caller code.

yadij:
Oh, I see. You want me to revert the parsedOptions back to the way I had it operating in the very first attempts to merge GnuTLS code?
 * PeerOptions::parse just storing the settings in string format, and 
 * the call to parseOptions() being made from updateContextOptions() when the ports are initialized.

rousskov:
> You want me to revert the parsedOptions back to the way I had it operating in the very first attempts to merge GnuTLS code?

Naturally, I do not know or remember what your first merging attempts were. If you want me to review a specific code or design outside of this PR, please point me to it. However, if that design matches what I have suggested, and you are not aware of any problems with it, then implementing it in this PR may be more efficient than discussing some old and probably out-of-sync code.

> PeerOptions::parse just storing the settings in string format, and
> the call to parseOptions() being made from updateContextOptions() when the ports are initialized.

If possible, I would not delay interpreting/compiling parsed options until updateContextOptions() or some other future events. I would interpret them immediately after we are done parsing them. In other words:

* PeerOptions::parse() parses configured settings and, once all the settings are parsed, compiles them.

rousskov:
> can you point at where this "all the settings are parsed" point occurs in the code?

My quick check points to parse_securePeerOptions() and parse_peer() as two of those places. There is probably at least one more such place (for *_port parsing). However, this answer sounds too obvious to be the one you are looking for -- you probably would not have asked the above question if that was the correct answer. I do not know why this answer is wrong though. Perhaps I do not understand the question?

> Because that looks like configDoConfigure() to me - exactly where the contexts are initialized.

configDoConfigure() feels too high-level -- it orchestrates parsing of all directives. The peer/security options in question ought to be specific/local to their individual directives.

yadij:
> > Because that looks like configDoConfigure() to me - exactly where the contexts are initialized.
> 
> configDoConfigure() feels too high-level -- it orchestrates parsing of all directives. The peer/security options in question ought to be specific/local to their individual directives.

a06e3a0 does what you ask for early parsing. This will result in repeated re-parsing of sslOptions string for configurations where the sslproxy_* directives or multiple tls_outgoing_options are used.

yadij:
Marking this for merge provided the re-tests pass. This is a needed security feature that has been unresolved too long already. The current code is in use by several testers, so the issue seems resolved even if the design is not perfect. It can be updated in a later PR if anyone wishes to polish any further.

yadij:
Fedora tests fail due to general build failures in unrelated code on their new compiler versions. Skipping that all builds pass.

PR #332, opened 2018-11-14
Remove 8K limit for single access.log line

redmeadowman:
The function logfilePrintf() currently limits a single line
to no more than 8K characters.  It allocates an 8K buffer
on the stack, which is not safe. This PR eliminates
the 8K limit, and moves the buffer to the heap.


kinkie:
ok to test

src/log/File.cc
```diff
@@ -112,21 +112,35 @@ void
...
+        // Try again with a bigger buffer
+        delete buf;
+        buf = NULL;
+        bufsize *= 2;
+    }
 
-        if (fmt[strlen(fmt) - 1] == '\n')
-            buf[8191] = '\n';
+    if (!buf) {
+        logfileWrite(lf, (char*)"-", 1);
```
rousskov: Please do not cast const away. Fix logfileWrite() profile instead. AFAICT, that fix is simple and will not lead to many cascading changes.

src/log/File.cc
```diff
@@ -112,21 +112,35 @@ void
...
+        auto s = vsnprintf(buf, bufsize, fmt, args);
+        if (s < bufsize) {
+            logfileWrite(lf, buf, (size_t) s);
+            delete buf;
+            break;
+        }
+
+        // Try again with a bigger buffer
+        delete buf;
+        buf = NULL;
```
rousskov: Please use `nullptr` on new or modified lines.

src/log/File.cc
```diff
@@ -112,21 +112,35 @@ void
...
-    int s;
-
     va_start(args, fmt);
 
-    s = vsnprintf(buf, 8192, fmt, args);
-
-    if (s > 8192) {
-        s = 8192;
+    // bufsize will be large enough for most log lines
+    int bufsize = 16 * 1024;
```
rousskov: Please use `size_t` for sizes.

src/log/File.cc
```diff
@@ -112,21 +112,35 @@ void
...
+    while (true) {
+        buf = new char[bufsize];
+        if (!buf) {
+            break;
+        }
+
+        auto s = vsnprintf(buf, bufsize, fmt, args);
+        if (s < bufsize) {
+            logfileWrite(lf, buf, (size_t) s);
+            delete buf;
```
rousskov: Please do not leave dangling pointers, especially if code checks them for being nil, even if the (current) code does not actually dereference them.

```suggestion
            delete buf;
            buf = nullptr;
```

You will need to refactor other code to accomodate this change, but I hope that this and other changes will become unnecessary after the higher level problems are addressed.

src/log/File.cc
```diff
@@ -112,21 +112,35 @@ void
...
-    s = vsnprintf(buf, 8192, fmt, args);
-
-    if (s > 8192) {
-        s = 8192;
+    // bufsize will be large enough for most log lines
+    int bufsize = 16 * 1024;
+    char* buf = NULL;
+
+    while (true) {
+        buf = new char[bufsize];
```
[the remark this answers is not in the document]
redmeadowman: Buffers allocated on stacks are the #1 cause of vulnerabilities in software.  It is much safer to allocate memory in the heap.  Also, 16K is an excessive amount of memory to allocate on the stack (the former amount was 8K, which is also excessive for the stack).  Allocating a static buffer would be fine, but then the function is no longer thread-safe, unless the memory can be allocated with the C++11 thread_local directive.

src/log/File.cc
```diff
@@ -112,21 +112,35 @@ void
...
+        if (!buf) {
```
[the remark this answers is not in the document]
redmeadowman: Good catch, thanks.

src/log/File.cc
```diff
@@ -112,21 +112,35 @@ void
...
+        }
+
+        auto s = vsnprintf(buf, bufsize, fmt, args);
+        if (s < bufsize) {
+            logfileWrite(lf, buf, (size_t) s);
+            delete buf;
+            break;
+        }
+
+        // Try again with a bigger buffer
```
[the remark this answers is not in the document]
redmeadowman: Agreed.

src/log/File.cc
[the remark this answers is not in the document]
rousskov: AFAIK, this function does not need to be thread safe. In fact, virtually no Squid function needs to be thread safe except a very small set of specialized functions dedicated to AUFS code (src/DiskIO/DiskThreads/). Outside of threaded AUFS code, virtually any "I will do X because of thread safety" argument is invalid. This logging code is outside of threaded AUFS code.

And if you convert this code to use SBuf, then we do not need to argue whether allocating X KB on stack is "safe" -- SBuf does not use stack for its storage.

rousskov:
> On the architectural level please be aware that this 8KB buffer is not the only limit imposed on log lines.

AFAICT, it _is_ the only 8KB limit in some environments, and even if it is never the only limit, the PR intent to increase that limit is still valid. You have not said that it is invalid, but I want to explicitly clarify that nothing in your comment should be misinterpreted as PR intent invalidation.

BTW, your summary is very useful for many purposes! Please consider copying it to some wiki page.

> IMO this magic 8196 number should just be replaced with LOGFILE_BUFSZ which can be made replaceable at build time with -DLOGFILE_BUFSZ when it is moved to log/File.h

I do not think a `LOGFILE_BUFSZ` macro is a good solution here because 8KB is too small for most production deployments these days while significantly larger defaults become increasingly inappropriate for stack allocation. Having unlimited buffer is probably the best solution here, and that eliminates the need for defining `LOGFILE_BUFSZ`.

rousskov:
> The Log:: objects are overdue for a full refactoring to use C++ instead of C virtual functions - which seems out of scope here.

Agreed on both counts. That refactoring should be done, but in a separate independent PR.

redmeadowman:
Completed requested modifications and ran tests.

redmeadowman:
LOGFILE_BUFSZ is currently defined as 32K, which should be more than adequate for most lines.

rousskov:
> LOGFILE_BUFSZ is currently defined as 32K, which should be more than adequate for most lines.

Agreed. In case you are implying the opposite, increasing initial buffer size does _not_ address my change request regarding code duplication. I am not sure why you are ignoring it (and a few other change requests), but I am happy to answer any questions about them.

Please go back to my [review](https://github.com/squid-cache/squid/pull/332#pullrequestreview-175130857) and make sure that change requests not already marked as "resolved" are addressed, starting with the code duplication request (which starts with `Please find a way to avoid duplicating SBuf::vappendf()`). Please note that change requests marked as "outdated" by GitHub may still need to be addressed -- the essence of the problem may remain even if the code the change request was attached to has changed.

redmeadowman:
Silly question, why did Squid create SBuf instead of using std::string?

rousskov:
> Silly question, why did Squid create SBuf instead of using std::string?

In short, std::string is a bad choice for frequently shared, parsed, and peeked at raw buffers that Squid needs for shoveling I/O results around and similar performance-sensitive tasks. The general-purpose class was not designed for that kind of performance work. For example, modern std::string cannot be safely shared without copying. Also, std::string has some scary and slow locale-dependent code (at least it was present in older std::string implementations).

src/log/File.cc
```diff
@@ -112,21 +113,11 @@ void
...
-        s = 8192;
-
-        if (fmt[strlen(fmt) - 1] == '\n')
-            buf[8191] = '\n';
-    }
-
-    logfileWrite(lf, buf, (size_t) s);
+    static SBuf sbuf;
+    sbuf.clear();
+    sbuf.vappendf(fmt, args);
```
rousskov: Other bugs notwithstanding, the maximum line size is now limited to 256 MB (SBuf::maxSize or 268435455 bytes to be precise). That is more than enough, of course, but I wonder whether we should add code to handle overflows. Without such code, an unlikely huge access.log line will either kill Squid or not appear in the access.log at all. Thus, we are back to the six overflow handling options discussed earlier, except that the simplest option 6 is off the table. Since we no longer deal with raw formatting code here, I suggest working around this problem instead of solving it directly:

1. Do not handle overflows in logfilePrintf(). Just check that if SBuf throws, there will some ERROR message about that in cache.log. To check, temporary hack logfilePrintf() code to add `Must(false)` and see what happens -- do that before moving on to step2 below.
2. Change Log::Format::SquidCustom() to print the already formatted(!!) line using logfileWrite() instead of abusing logfilePrintf(). You may need to refactor a little so that Format::assemble() appends a new line when needed.

The above approach is based on the assumption that the most likely (if not the only source) of logfilePrintf() overflows is _custom_ log formats. And if we want to improve overflow handling there, we have to improve Format::assemble(), where overflow handling approaches like the nice option 5 posted earlier are much easier to implement. We do not have to do those improvements in this PR though -- the above two steps should be enough for now as far as overflow handling is concerned.

rousskov: AFAICT, SBuf::vappendf() is completely broken right now. It does not increase the size in a loop and some of its return value tests do not make sense to me. I would expect it to fail on many logformat specifications, especially a trivial one that logs just the URL (i.e., short format specs but long parameter string).

If I am right, please fix that method so that we can actually reuse it here without breaking everything. You already have already written most of the code that you can resurrect in SBuf::vappendf() to fix that method. Just keep my earlier change requests in mind when doing so.

src/log/File.cc
```diff
@@ -11,6 +11,7 @@
 #include "squid.h"
 #include "fatal.h"
 #include "fde.h"
+#include "sbuf/SBuf.h"
```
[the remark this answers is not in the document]
yadij: FYI: You can run scripts/sourcemaintenance.sh to do this change automatically. Just ignore or revert anything it does outside of the lines you are changing in this PR.

src/log/File.cc
```diff
@@ -112,21 +113,11 @@ void
...
-
-        if (fmt[strlen(fmt) - 1] == '\n')
-            buf[8191] = '\n';
-    }
-
-    logfileWrite(lf, buf, (size_t) s);
+    static SBuf sbuf;
+    sbuf.clear();
+    sbuf.vappendf(fmt, args);
+    logfileWrite(lf, sbuf.c_str(), sbuf.length());
```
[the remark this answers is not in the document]
yadij: At least logfile_mod_syslog_writeline requires it to be a 0-terminated C-string. That is fixable if you go with this change.

Other UDP receivers may be assuming 0-teriminated strings in the packets they receive based on that being what Squid currently sends them today. We cannot really be sure in advance whether this string termination change is going to be a problem for anyone. That said, you can always do it anyway - this line size change is already a UI alteration restricting this PR to v5 series and needing release notes about the behaviour differences.

src/log/File.cc
rousskov: > Alex: the most likely source of logfilePrintf() overflows is custom log formats

> Amos: that assumption is wrong. [...] all log formats built into Squid are possible overflows

Your statement does not contradict my assumption AFAICT, but, more importantly, I believe that the already implemented PR changes invalidate your statement: Some logfilePrintf() callers cannot overflow at all (even before the PR), and only custom log format has any chance of producing log lines that exceed 268435455 characters (the post-PR limit).

> care needs to be taken not to break any of those by forcing implicit line-wrapping into Format::assemble() output

Yes, that is why I said "appends a new line _when needed_". One way to handle this without adding parameters/options to assemle() could be to append a newline to the configured logformat string.

src/log/File.cc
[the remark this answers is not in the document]
rousskov: > At least logfile_mod_syslog_writeline requires it to be a 0-terminated C-string.

FWIW, that looks like a logfile_mod_syslog_writeline() bug: The logfileWrite() API implies unterminated buffer, and the official logfilePrintf() implementation does not always terminate the buffer it gives to logfileWrite().

Given that some of the underlying code is broken, and the full extent of the problem is unclear, I suggest leaving `c_str()` in here with a comment:

```suggestion
    // XXX: Fix logfileWrite() implementations that assume 0-termination
    logfileWrite(lf, sbuf.c_str(), sbuf.length());
```

> Other UDP receivers may be assuming 0-teriminated strings in the packets they receive based on that being what Squid currently sends them today

I hope Squid does not send string termination characters in UDP packets, but this uncertainty is one of the reasons to leave `c_str()` in here until somebody can get to the bottom of it.

src/log/File.cc
redmeadowman: Reading back through the comments above it is not clear to me what changes have been requested for this pull request.

rousskov: This change request is still unresolved. The two-step plan I have suggested still applies AFAICT.

src/log/File.cc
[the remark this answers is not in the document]
rousskov: This change request is still unresolved. IMO, you can refuse to resolve it because it is a minor style change that should be automated instead, but you need to do so explicitly. I recommend just moving the line where it belongs instead (as I originally suggested).

rousskov:
> Reading back through the comments above it is not clear to me what changes have been requested for this pull request.

I went through all my change requests and marked the addressed ones (and the ones that do not need to be addressed) as resolved. AFAICT, there are two [change](https://github.com/squid-cache/squid/pull/332#discussion_r236476980) [requests](https://github.com/squid-cache/squid/pull/332#discussion_r236475542) remaining. I added a "not addressed" comment to each of them. [One](https://github.com/squid-cache/squid/pull/332#discussion_r236476980) is optional/trivial. The [other](https://github.com/squid-cache/squid/pull/332#discussion_r236475542) one requires a little bit of work.

Rant: GitHub does not make it very easy to find what needs to be done, but if you use the web interface and start from the actual review [post](https://github.com/squid-cache/squid/pull/332#pullrequestreview-175559588) (rather than notification emails or some random change requests), you can usually find all the change requests that may need your attention. This can be time consuming, especially when there are outdated blobs and when reviewers do not mark addressed change requests promptly, but that is the only way I know that works. You can always ask about a particular change request status if you are not sure, of course.

Please avoid using individual change requests for posting general PR questions or comments (like this one). Each change request should be focused on that specific change.

src/log/File.cc
redmeadowman: I am out of time on this project.  I will not implement the suggested change.

rousskov: > I will not implement the suggested change.

Thank you for sharing your decision. I believe the requested change should be implemented before this PR is approved because it is not difficult to implement, will address a potentially important in-scope problem, and even provide a significant performance improvement. Hopefully, somebody else will take over and finish this work.

However, if other @squid-cache/commiters disagree, they may be able to approve this PR (after a thorough from-scratch review) without this change because they may decide that the PR is a step forward, while that known problem is unlikely to surface in real deployments.

rousskov:
@yadij, do you want to link the new bug [5026](https://bugs.squid-cache.org/show_bug.cgi?id=5026) from this PR title before it is merged?

yadij:
If it actually fixes that bug, yes

yadij:
It looks like something is going wrong with the pr-auto triggers. The status says "build triggered for merge commit", but Jenkins has no queued build for this PR.

rousskov:
> It looks like something is going wrong with the pr-auto triggers. The status says "build triggered for merge commit", but Jenkins has no queued build for this PR.

I was worried about the lack of Jenkins results as well. @eduard-bagdasaryan, could you please take a look? I do not know much about Jenkins setup...

rousskov:
> If it actually fixes that bug, yes.

Based on the bug and PR descriptions, I assume it does, but I do not know for sure. The reporter has not updated the bug report yet(?) with this information (but that could also be because this PR does not apply cleanly to v4).

> AFAIK once Anubis starts triggering its build commit it is too late, the commit already being tested with the existing title will either be merged or fail.

It should not be too late until the tested commit is merged into master. Anubis should notice the title change and regenerate the merge commit, which should restart testing (IIRC).

AFAICT, as of now, Anubis has not even created the merge commit (on the auto branch) because Anubis waits for Jenkins test results (of the GitHub-generated commit).

eduard-bagdasaryan:
> It looks like something is going wrong with the pr-auto triggers.

It looks like Jenkins was down at the time of fc2501d, thus missing the notification. The easiest way to re-trigger it is probably to re-send the corresponding webhook push event (the list of notifications should be available on Webhooks GitHub page). I cannot do it myself due to permission restrictions.

rousskov:
> The easiest way to re-trigger it is probably to re-send the corresponding webhook push event

Hmm.. I do not see a push event that corresponds to fc2501dc43280cbb5c0455ec9834346844f3cb93 dated 2020-02-12. The latest events are 2020-02-08 and  2020-02-18. Neither event is about this PR AFAICT. I am going to commit a non-change in hope to trigger a new Jenkins test.

rousskov:
@yadij, please consider backporting this fix (i.e. commit a03343c) to v5 which suffers from the same logline truncation problem. I have heard that a similar fix works for v5. Please let me know if you want Factory to confirm and post a backporting pull request to save you time.

rousskov:
> @yadij, please consider backporting this fix (i.e. commit [a03343c](https://github.com/squid-cache/squid/commit/a03343c2acbc382a59a176713d5e33387b82f6a9)) to v5 which suffers from the same logline truncation problem. I have heard that a similar fix works for v5. Please let me know if you want Factory to confirm and post a backporting pull request to save you time.

@yadij, please let me know whether you would backport commit a03343c into v5 (or accept a backport). A year of silence may imply "no", but it could also imply that you just did not see the question or forgot about it... Let's close this question, one way or another.

PR #334, opened 2018-11-18
Maintenance script updates

yadij:
Fix/workaround some bugs and incorrect errors in the
source-maintenance script.

### DIFF
--- a/scripts/source-maintenance.sh
+++ b/scripts/source-maintenance.sh
@@ -141,7 +141,7 @@ for FILENAME in `git ls-files`; do
 	# sprintf() - not allowed anywhere.
 	#
 	STRDUP=`grep -e "[^x]strdup(" ${FILENAME}`;
-	if test "x${STRDUP}" != "x" -a "${FILENAME}" != "xstring.h"; then
+	if test "x${STRDUP}" != "x" -a "${FILENAME}" != "compat/xstring.h"; then
 		echo "ERROR: ${FILENAME} contains unprotected use of strdup()"
 	fi
 	SPRINTF=`grep -e "[^v]sprintf(" ${FILENAME}`;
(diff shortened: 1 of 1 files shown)


scripts/source-maintenance.sh
```diff
@@ -118,12 +118,12 @@ for FILENAME in `git ls-files`; do
...
 	#
-	if test "${FILENAME}" != "compat/openssl.h" && \
-	   (FI=`grep "#include.*openssl/" "${FILENAME}" 2>/dev/null`; \
-	    test "x${FI}" != "x") && \
-	   (FI=`grep "#include \"compat/openssl\.h\"" "${FILENAME}" 2>/dev/null`; \
-	    test "x${FI}" == "x"); then
-		echo "ERROR: ${FILENAME} includes openssl headers without including \"compat/openssl.h\""
+	if test "${FILENAME}" != "compat/openssl.h"; then
+		FA=`grep "#include.*openssl/" "${FILENAME}" 2>/dev/null | head -1`;
+		FB=`grep "#include.*compat/openssl" "${FILENAME}" 2>/dev/null | head -1`;
```
rousskov:
```suggestion
		FB=`grep '#include.*compat/openssl[.]h' "${FILENAME}" 2>/dev/null | head -1`;
```

... to better match the blob stated goal, the ERROR text and the intent. If we do want to relax this rule to allow for other compat/openssl headers, then we should adjust the blob stated goal and the ERROR text.

scripts/source-maintenance.sh
```diff
@@ -118,12 +118,12 @@ for FILENAME in `git ls-files`; do
...
+		if test "x${FA}" != "x" && test "x${FB}" = "x"; then
```
rousskov:
```suggestion
		if test "x${FA}" != "x" -a "x${FB}" = "x"; then
```

... for consistency with other similar "a and b" tests in this script.

scripts/source-maintenance.sh
yadij: This may not be portable grep syntax, but okay lets give it a try.

PR #143, opened 2018-02-02
Cleanup: storeAppendPrintf removal script

yadij:
Automatically convert deprecated uses of storeAppendPrintf()
to use the Packable API method.

The store API definitions and some difficult to auto-remove
uses remain for future manual removal once all developer
branches have been updated by this script.

PR #336, opened 2018-11-27
Cleanup and simplify unit tests

yadij:
Polish unit tests to reduce their dependency lists inline with
the updated test documentation requirements.

Some further cleanup of unit tests documentation based on
experience pruning existing unit tests.

 * Only the test logic files actually need to be distributed by
   a test. All files being tested are supposed to be distributed
   from elsewhere and the test should rely on that to prevent
   future issues like the base/RefCount.h bug mentioned below.

 * Starting to deprecate TESTSOURCES which is causing more
   trouble than benefit by pulling in too many needless
   dependencies. eg the globals.cc objects and symbols. Tools it
   used to provide are largely superseded by the stub mechanisms.

Add missing stub files necessary for pruning dependencies. Also
fix some bugs in existing stub files.

Fix base/RefCount.h distribution. This file was not included in
the base/libbase.la dependency list but indirectly being
distributed by the unit-test existence. Which in some builds
could cause it not to exist in generated minimal tarballs.


yadij:
Removing the WIP, this can be applied without completing all test pruning. Most of the remaining pruning is blocked by storeAppend*() API dependency - for which PR #143 was the optimal fix.

yadij:
@rousskov Do you wish to review any of this? or shall I clear Anubis to proceed?

rousskov:
> Do you wish to review any of this?

No, I do not. I would not have [added](https://github.com/squid-cache/squid/pull/336#event-2056509351) an `S-waiting-for-committer` label if I wanted to review this PR.

PR #339, opened 2018-12-04
Fixed on_unsupported_protocol tunnel action

chtsanti:
Instead of tunneling traffic, a matching on_unsupported_protocol
"tunnel" action resulted in a Squid error response sent to the client
(or, where an error response was not possible, in a connection closure).
The following three cases were fixed:

    * port: http_port (real CONNECT)
    * ssl_bump action: client-first or step1 bump
    * handling phase: parsing TLS client handshake
    * expected data: TLS Client Hello

    * port: http_port (real CONNECT)
    * ssl_bump action: client-first or step1 bump
    * handling phase: parsing the first bumped HTTP request
    * expected data: HTTP request header

    * port: https_port (fake CONNECT)
    * ssl_bump action: any action except terminate
    * handling phase: parsing TLS client handshake
    * expected data: TLS Client Hello

Also, when on_unsupported_protocol was configured, Squid wasted RAM and
CPU cycles to buffer client HTTP requests beyond the point of no return
(i.e., roughly, beyond the first HTTP request on a connection or in a
tunnel), when on_unsupported_protocol settings no longer apply.

Client handshake accumulation is now driven by preservingClientData_. We
set that data member when the connection is accepted (because we may
decide to start preserving bytes right away) and reset it whenever that
decision may change, including when switching to a new protocol inside
CONNECT tunnel and confirming the expected/supported protocol by
successfully parsing its handshake.

Squid does not stop handshake preservation when on_unsupported_protocol
gets disabled during reconfiguration, but Squid will not tunnel
preserved bytes if that happens (and will not tunnel a partial handshake
if on_unsupported_protocol configuration keeps changing).

Also changed how IPv6-based certificates are generated. Their CN field
value is no longer surrounded by [square brackets]. This change was done
to improve Squid code that had to be modified to fix
on_unsupported_protocol. It affects certificate cache key so old
IPv6-based certificates will never be found (and will eventually be
purged) while new ones will be generated and cached instead. We believe
these IPv6-based certificates are rare and untrusted by browsers so the
change in their CN should not have a significant affect on users.

### DIFF
--- a/src/anyp/Uri.cc
+++ b/src/anyp/Uri.cc
@@ -63,8 +63,10 @@ SBuf
 AnyP::Uri::hostOrIp() const
 {
     static char ip[MAX_IPSTRLEN];
-    SBuf hostOrIp_(hostIsNumeric() ? hostIP().toStr(ip, sizeof(ip)) : host());
-    return hostOrIp_;
+    if (hostIsNumeric())
+        return  SBuf(hostIP().toStr(ip, sizeof(ip)));
+    else
+        return SBuf(host());
 }
 
 const SBuf &
--- a/src/anyp/Uri.cc
+++ b/src/anyp/Uri.cc
@@ -59,6 +59,14 @@ AnyP::Uri::host(const char *src)
     touch();
 }
 
+SBuf
+AnyP::Uri::hostOrIp() const
+{
+    static char ip[MAX_IPSTRLEN];
+    SBuf hostOrIp_(hostIsNumeric() ? hostIP().toStr(ip, sizeof(ip)) : host());
+    return hostOrIp_;
+}
+
 const SBuf &
 AnyP::Uri::path() const
 {
--- a/src/anyp/Uri.cc
+++ b/src/anyp/Uri.cc
@@ -64,7 +64,7 @@ AnyP::Uri::hostOrIp() const
 {
     static char ip[MAX_IPSTRLEN];
     if (hostIsNumeric())
-        return  SBuf(hostIP().toStr(ip, sizeof(ip)));
+        return SBuf(hostIP().toStr(ip, sizeof(ip)));
     else
         return SBuf(host());
 }
--- a/src/anyp/Uri.h
+++ b/src/anyp/Uri.h
@@ -80,7 +80,9 @@ public:
     int hostIsNumeric(void) const {return hostIsNumeric_;}
     Ip::Address const & hostIP(void) const {return hostAddr_;}
 
-    /// \return the hostname or the unformatted ip address
+    /// \returns the host subcomponent of the authority component
+    /// If the host is an IPv6 address, returns that IP address without
+    /// [brackets]! See RFC 3986 Section 3.2.2.
     SBuf hostOrIp() const;
 
     void port(unsigned short p) {port_=p; touch();}
--- a/src/anyp/Uri.h
+++ b/src/anyp/Uri.h
@@ -80,6 +80,9 @@ public:
     int hostIsNumeric(void) const {return hostIsNumeric_;}
     Ip::Address const & hostIP(void) const {return hostAddr_;}
 
+    /// \return the hostname or the unformatted ip address
+    SBuf hostOrIp() const;
+
     void port(unsigned short p) {port_=p; touch();}
     unsigned short port() const {return port_;}
     /// reset the port to the default port number for the current scheme
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1929,6 +1929,11 @@ ConnStateData::clientParseRequests()
 
             context->registerWithConn();
 
+#if USE_OPENSSL
+            if (switchedToHttps())
+                parsedBumpedRequestCount++;
+#endif
+
             processParsedRequest(context);
 
             parsed_req = true; // XXX: do we really need to parse everything right NOW ?
@@ -2181,6 +2186,7 @@ ConnStateData::ConnStateData(const MasterXaction::Pointer &xact) :
 #if USE_OPENSSL
     switchedToHttps_(false),
     parsingTlsHandshake(false),
+    parsedBumpedRequestCount(0),
     tlsConnectPort(0),
     sslServerBump(NULL),
     signAlgorithm(Ssl::algSignTrusted),
@@ -3986,20 +3992,17 @@ ConnStateData::mayTunnelUnsupportedProto()
 {
     if (!Config.accessList.on_unsupported_protocol)
         return false;
- #if USE_OPENSSL
-    // subject to SslBump processing and currently parsing the TLS
-    // client handshake
-    if (sslBumpMode != Ssl::bumpEnd && !switchedToHttps_)
-        return true;
 
-    // The first request inside a bumped connection. pipeline.nrequests is 2
-    // because it counts the (fake or real) CONNECT request as well.
-    if (sslBumpMode != Ssl::bumpEnd && pipeline.nrequests == 2)
+#if USE_OPENSSL
+    // SslBump processing:
+    //  1) parsing the TLS client handshake
+    //  2) reading and processing the first bumped HTTP request
+    if (sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1)
         return true;
- #endif
+#endif
 
     // the first request in a connection to a plain intercepting port
-    if (transparent() && pipeline.nrequests == 1)
+    if (transparent() && pipeline.nrequests <= 1)
         return true;
 
     return false;
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -2931,14 +2931,17 @@ ConnStateData::switchToHttps(ClientHttpRequest *http, Ssl::BumpMode bumpServerMo
     // Depending on receivedFirstByte_, we are at the start of either an
     // established CONNECT tunnel with the client or an intercepted TCP (and
     // presumably TLS) connection from the client. Expect TLS Client Hello.
-    debugs(33, 5, (receivedFirstByte_ ? "post-CONNECT " : "raw TLS ") << clientConnection);
+    const auto insideConnectTunnel = receivedFirstByte_;
+    debugs(33, 5, (insideConnectTunnel ? "post-CONNECT " : "raw TLS ") << clientConnection);
 
     tlsConnectHostOrIp =  request->url.hostOrIp();
     tlsConnectPort = request->url.port();
     resetSslCommonName(request->url.host());
 
-    // Able to tunnel after a timeout without receiving any data
-    preservingClientData_ = true;
+    // If the protocol has changed, then reset preservingClientData_.
+    // Otherwise, its value initially set in start() is still valid/fresh.
+    if (insideConnectTunnel)
+        preservingClientData_ = preserveHttpBytesForTunnellingUnsupportedProto();
 
     // We are going to read new request
     flags.readMore = true;
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -4015,15 +4015,18 @@ ConnStateData::preserveHttpBytesForTunnellingUnsupportedProto() const
 
 #if USE_OPENSSL
     // The first HTTP request on a bumped connection.
-    const auto firstBumpedRequest = sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1;
-#else
-    const auto firstBumpedRequest = false;
+    // XXX: sslBumpMode != bumpEnd includes non-bumped bumpSplice and bumpNone.
+    // XXX: parsedBumpedRequestCount == 1 means the _second_ HTTP request
+    if (sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1)
+        return true;
 #endif
 
     // the first request in a connection to a plain intercepting port
-    const auto firstTransparentPlainRequest = !port->secure.encryptTransport && transparent() && pipeline.nrequests <= 1;
+    // XXX: pipeline.nrequests == 1 may mean the _second_ request
+    if (!port->secure.encryptTransport && transparent() && pipeline.nrequests <= 1)
+        return true;
 
-    return firstBumpedRequest || firstTransparentPlainRequest;
+    return false;
 }
 
 bool
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -2675,18 +2675,18 @@ ConnStateData::sslCrtdHandleReply(const Helper::Reply &reply)
     }
 
     if (reply.result == Helper::BrokenHelper) {
-        debugs(33, 5, HERE << "Certificate for " << tlsConnectHostOrIp << " cannot be generated. ssl_crtd response: " << reply);
+        debugs(33, 5, "Certificate for " << tlsConnectHostOrIp << " cannot be generated. ssl_crtd response: " << reply);
     } else if (!reply.other().hasContent()) {
         debugs(1, DBG_IMPORTANT, HERE << "\"ssl_crtd\" helper returned <NULL> reply.");
     } else {
         Ssl::CrtdMessage reply_message(Ssl::CrtdMessage::REPLY);
         if (reply_message.parse(reply.other().content(), reply.other().contentSize()) != Ssl::CrtdMessage::OK) {
-            debugs(33, 5, HERE << "Reply from ssl_crtd for " << tlsConnectHostOrIp << " is incorrect");
+            debugs(33, 5, "Reply from ssl_crtd for " << tlsConnectHostOrIp << " is incorrect");
         } else {
             if (reply.result != Helper::Okay) {
-                debugs(33, 5, HERE << "Certificate for " << tlsConnectHostOrIp << " cannot be generated. ssl_crtd response: " << reply_message.getBody());
+                debugs(33, 5, "Certificate for " << tlsConnectHostOrIp << " cannot be generated. ssl_crtd response: " << reply_message.getBody());
             } else {
-                debugs(33, 5, HERE << "Certificate for " << tlsConnectHostOrIp << " was successfully recieved from ssl_crtd");
+                debugs(33, 5, "Certificate for " << tlsConnectHostOrIp << " was successfully recieved from ssl_crtd");
                 if (sslServerBump && (sslServerBump->act.step1 == Ssl::bumpPeek || sslServerBump->act.step1 == Ssl::bumpStare)) {
                     doPeekAndSpliceStep();
                     auto ssl = fd_table[clientConnection->fd].ssl.get();
@@ -3188,9 +3188,9 @@ ConnStateData::httpsPeeked(PinnedIdleContext pic)
 
     if (Comm::IsConnOpen(pic.connection)) {
         notePinnedConnectionBecameIdle(pic);
-        debugs(33, 5, HERE << "bumped HTTPS server: " << tlsConnectHostOrIp);
+        debugs(33, 5, "bumped HTTPS server: " << tlsConnectHostOrIp);
     } else
-        debugs(33, 5, HERE << "Error while bumping: " << tlsConnectHostOrIp);
+        debugs(33, 5, "Error while bumping: " << tlsConnectHostOrIp);
 
     getSslContextStart();
 }
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -2954,11 +2954,6 @@ ConnStateData::switchToHttps(ClientHttpRequest *http, Ssl::BumpMode bumpServerMo
     tlsConnectPort = request->url.port();
     resetSslCommonName(request->url.host());
 
-    // If the protocol has changed, then reset preservingClientData_.
-    // Otherwise, its value initially set in start() is still valid/fresh.
-    if (insideConnectTunnel)
-        preservingClientData_ = shouldPreserveClientData();
-
     // We are going to read new request
     flags.readMore = true;
 
@@ -2988,6 +2983,13 @@ ConnStateData::switchToHttps(ClientHttpRequest *http, Ssl::BumpMode bumpServerMo
     receivedFirstByte_ = false;
     // Get more data to peek at Tls
     parsingTlsHandshake = true;
+
+    // If the protocol has changed, then reset preservingClientData_.
+    // Otherwise, its value initially set in start() is still valid/fresh.
+    // Requires parsingTlsHandshake which is initialized above.
+    if (insideConnectTunnel)
+        preservingClientData_ = shouldPreserveClientData();
+
     readSomeData();
 }
 
@@ -4025,9 +4027,12 @@ ConnStateData::shouldPreserveClientData() const
         return false;
 
 #if USE_OPENSSL
+    // We are parsing client hello request
+    if (parsingTlsHandshake)
+        return true;
+
     // the 1st HTTP request on a bumped connection
-    // XXX: sslBumpMode != bumpEnd includes non-bumped bumpSplice and bumpNone.
-    if (!parsedBumpedRequestCount && sslBumpMode != Ssl::bumpEnd)
+    if (!parsedBumpedRequestCount && switchedToHttps())
         return true;
 #endif
 
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1570,7 +1570,7 @@ ConnStateData::tunnelOnError(const HttpRequestMethod &method, const err_type req
     checklist.my_addr = clientConnection->local;
     checklist.conn(this);
     const char *log_uri = http ? http->log_uri : nullptr;
-    checklist.syncAle(request.getRaw(), log_uri);
+    checklist.syncAle(request, log_uri);
     auto answer = checklist.fastCheck();
     if (answer.allowed() && answer.kind == 1) {
         debugs(33, 3, "Request will be tunneled to server");
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -2987,7 +2987,7 @@ ConnStateData::switchToHttps(ClientHttpRequest *http, Ssl::BumpMode bumpServerMo
 
     // If the protocol has changed, then reset preservingClientData_.
     // Otherwise, its value initially set in start() is still valid/fresh.
-    // Requires parsingTlsHandshake which is initialized above.
+    // shouldPreserveClientData() uses parsingTlsHandshake which is reset above.
     if (insideConnectTunnel)
         preservingClientData_ = shouldPreserveClientData();
 
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1300,7 +1300,10 @@ parseHttpRequest(ConnStateData *csd, const Http1::RequestParserPointer &hp)
             return NULL;
         }
 
-        if (csd->mayTunnelUnsupportedProto()) {
+        if (csd->ableToTunnelUnsupportedProto()) {
+            // The preserveDataForTunnelingUnsupportedProto is already checked
+            // inside Http::One::Server::parseOneRequest and the hp parser
+            // must preserved parsed data.
             csd->preservedClientData = hp->parsed();
             csd->preservedClientData.append(csd->inBuf);
         }
@@ -1556,7 +1559,7 @@ bool ConnStateData::serveDelayedError(Http::Stream *context)
 bool
 clientTunnelOnError(ConnStateData *conn, Http::StreamPointer &context, HttpRequest::Pointer &request, const HttpRequestMethod& method, err_type requestError)
 {
-    if (conn->mayTunnelUnsupportedProto()) {
+    if (conn->ableToTunnelUnsupportedProto()) {
         ACLFilledChecklist checklist(Config.accessList.on_unsupported_protocol, request.getRaw(), nullptr);
         checklist.al = (context && context->http) ? context->http->al : nullptr;
         checklist.requestErrorType = requestError;
@@ -2145,7 +2148,7 @@ ConnStateData::requestTimeout(const CommTimeoutCbParams &io)
     if (!Comm::IsConnOpen(io.conn))
         return;
 
-    if (mayTunnelUnsupportedProto() && !receivedFirstByte_) {
+    if (!receivedFirstByte_ && preserveDataForTunnellingUnsupportedProto()) {
         Http::StreamPointer context = pipeline.front();
         Must(context && context->http);
         HttpRequest::Pointer request = context->http->request;
@@ -2990,6 +2993,7 @@ ConnStateData::parseTlsHandshake()
     // client data may be needed for splicing and for
     // tunneling unsupportedProtocol after an error
     preservedClientData = inBuf;
+    ableToTunnelUnsupportedProto_ = true;
 
     // Even if the parser failed, each TLS detail should either be set
     // correctly or still be "unknown"; copying unknown detail is a no-op.
@@ -3985,23 +3989,25 @@ ConnStateData::checkLogging()
 }
 
 bool
-ConnStateData::mayTunnelUnsupportedProto()
+ConnStateData::preserveDataForTunnellingUnsupportedProto()
 {
-    if (!Config.accessList.on_unsupported_protocol)
-        return false;
-
 #if USE_OPENSSL
     // Tunneling may be possible when parsing the TLS client handshake
     // and when parsing the first HTTP request on a bumped connection.
-    if (sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1)
-        return true;
+    bool firstBumpedRequest = sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1;
 #endif
 
     // the first request in a connection to a plain intercepting port
-    if (transparent() && pipeline.nrequests <= 1)
-        return true;
+    bool firstTransparentPlainRequest = !port->secure.encryptTransport && transparent() && pipeline.nrequests <= 1;
+    ableToTunnelUnsupportedProto_ = Config.accessList.on_unsupported_protocol &&
+                                    (
+#if USE_OPENSSL
+                                     firstBumpedRequest ||
+#endif
+                                     firstTransparentPlainRequest
+                                    );
 
-    return false;
+    return ableToTunnelUnsupportedProto_;
 }
 
 NotePairs::Pointer
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -2265,7 +2265,7 @@ ConnStateData::start()
         whenClientIpKnown();
 
     // requires needProxyProtocolHeader_ which is initialized above
-    preservingClientData_ = preserveHttpBytesForTunnellingUnsupportedProto();
+    preservingClientData_ = shouldPreserveClientData();
 }
 
 void
@@ -2957,7 +2957,7 @@ ConnStateData::switchToHttps(ClientHttpRequest *http, Ssl::BumpMode bumpServerMo
     // If the protocol has changed, then reset preservingClientData_.
     // Otherwise, its value initially set in start() is still valid/fresh.
     if (insideConnectTunnel)
-        preservingClientData_ = preserveHttpBytesForTunnellingUnsupportedProto();
+        preservingClientData_ = shouldPreserveClientData();
 
     // We are going to read new request
     flags.readMore = true;
@@ -4013,7 +4013,7 @@ ConnStateData::checkLogging()
 }
 
 bool
-ConnStateData::preserveHttpBytesForTunnellingUnsupportedProto() const
+ConnStateData::shouldPreserveClientData() const
 {
     // do not tunnel on errors while expecting PROXY protocol bytes
     if (needProxyProtocolHeader_)
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -4026,6 +4026,10 @@ ConnStateData::shouldPreserveClientData() const
     if (!Config.accessList.on_unsupported_protocol)
         return false;
 
+    // Exclude FTP ports. They do not support on_unsupported_protocol.
+    if (port->transport.protocol == AnyP::PROTO_FTP)
+        return false;
+
 #if USE_OPENSSL
     // We are parsing client hello request
     if (parsingTlsHandshake)
@@ -4036,8 +4040,7 @@ ConnStateData::shouldPreserveClientData() const
         return true;
 #endif
 
-    // the 1st HTTP or FTP request on a connection to a plain intercepting port
-    // XXX: Exclude FTP ports. They do not support on_unsupported_protocol.
+    // the 1st HTTP request on a connection to a plain intercepting port
     if (!pipeline.nrequests && !port->secure.encryptTransport && transparent())
         return true;
 
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1563,25 +1563,21 @@ ConnStateData::tunnelOnError(const HttpRequestMethod &method, const err_type req
     const auto http = context ? context->http : nullptr;
     const auto request = http ? http->request : nullptr;
 
-    // TODO: Remove this change-reducing hack.
-    const auto conn = this;
-    {
-        ACLFilledChecklist checklist(Config.accessList.on_unsupported_protocol, request, nullptr);
-        checklist.al = http ? http->al : nullptr;
-        checklist.requestErrorType = requestError;
-        checklist.src_addr = conn->clientConnection->remote;
-        checklist.my_addr = conn->clientConnection->local;
-        checklist.conn(conn);
-        const char *log_uri = http ? http->log_uri : nullptr;
-        checklist.syncAle(request.getRaw(), log_uri);
-        auto answer = checklist.fastCheck();
-        if (answer.allowed() && answer.kind == 1) {
-            debugs(33, 3, "Request will be tunneled to server");
-            if (context)
-                context->finished(); // Will remove from pipeline queue
-            Comm::SetSelect(conn->clientConnection->fd, COMM_SELECT_READ, NULL, NULL, 0);
-            return conn->initiateTunneledRequest(request, Http::METHOD_NONE, "unknown-protocol", conn->preservedClientData);
-        }
+    ACLFilledChecklist checklist(Config.accessList.on_unsupported_protocol, request, nullptr);
+    checklist.al = http ? http->al : nullptr;
+    checklist.requestErrorType = requestError;
+    checklist.src_addr = clientConnection->remote;
+    checklist.my_addr = clientConnection->local;
+    checklist.conn(this);
+    const char *log_uri = http ? http->log_uri : nullptr;
+    checklist.syncAle(request.getRaw(), log_uri);
+    auto answer = checklist.fastCheck();
+    if (answer.allowed() && answer.kind == 1) {
+        debugs(33, 3, "Request will be tunneled to server");
+        if (context)
+            context->finished(); // Will remove from pipeline queue
+        Comm::SetSelect(clientConnection->fd, COMM_SELECT_READ, NULL, NULL, 0);
+        return initiateTunneledRequest(request, Http::METHOD_NONE, "unknown-protocol", preservedClientData);
     }
     debugs(33, 3, "denied; send error: " << requestError);
     return false;
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1547,37 +1547,54 @@ bool ConnStateData::serveDelayedError(Http::Stream *context)
 }
 #endif // USE_OPENSSL
 
-/**
- * Check on_unsupported_protocol checklist and return true if tunnel mode selected
- * or false otherwise
- */
+/// ConnStateData::tunnelOnError() wrapper. Reduces code changes. TODO: Remove.
 bool
 clientTunnelOnError(ConnStateData *conn, Http::StreamPointer &context, HttpRequest::Pointer &request, const HttpRequestMethod& method, err_type requestError)
 {
-    if (Config.accessList.on_unsupported_protocol && conn->ableToTunnelUnsupportedProto()) {
-        ACLFilledChecklist checklist(Config.accessList.on_unsupported_protocol, request.getRaw(), nullptr);
-        checklist.al = (context && context->http) ? context->http->al : nullptr;
+    assert(conn);
+    assert(conn->pipeline.front() == context);
+    return conn->tunnelOnError(method, requestError);
+}
+
+/// initiate tunneling if possible or return false otherwise
+bool
+ConnStateData::tunnelOnError(const HttpRequestMethod &method, const err_type requestError)
+{
+    if (!Config.accessList.on_unsupported_protocol) {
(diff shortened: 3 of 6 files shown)


src/client_side.cc
```diff
@@ -4103,13 +4116,25 @@ ConnStateData::checkLogging()
...
-    return Config.accessList.on_unsupported_protocol
-#if USE_OPENSSL
-           &&
-           ((port->flags.isIntercepted() && port->flags.tunnelSslBumping)
-            || (serverBump() && pinning.serverConnection))
-#endif
-           ;
+    if (!Config.accessList.on_unsupported_protocol)
+        return false;
+ #if USE_OPENSSL
```
yadij: missing empty line here.
```suggestion

 #if USE_OPENSSL
```

rousskov:
What is the current mayTunnelUnsupportedProto() semantics? In other words, what question does the method answer? Here are a few options (and there are more):

1. Assuming unsupported protocol data(**) has been or will be received on this connection, should it be (correctly) tunneled? The answer may change after that data was received, but, misconfigurations aside, all before- answers ought to be the same, and all after- answers ought to be the same.
1. Assuming unsupported protocol data(**) has been received on this connection, should it be (correctly) tunneled? Misconfigurations aside, the answer is the same no matter when the question is asked, but the question should only be asked _after_ the unsupported protocol data has been received.
1. Assuming unsupported protocol data(**) will be received on this connection, should it be (correctly) tunneled? Misconfigurations aside, the answer is the same no matter when the question is asked.
1. Assuming the caller can start tunneling this connection right now, should it? The answer may change depending on when the question is asked, including changes before- and after unsupported protocol data(**) is received.
 
I suspect it is option 1: AFAICT, right now, the method may be called both before and after the unsupported protocol data is received (or an equivalent event like a timeout has occurred). Each method caller should interpret the answer in its own context. Some callers may need to account for pending responses to previously pipelined requests; other callers may not care about pipelining at all.

When Squid initiates tunneling, it must preserve responses for still-pending non-CONNECT requests, of course. If Squid does not do that today, it is an out-of-scope Squid bug (like Amos has said). I agree that it would be nice to mark that bug now with an XXX if we know which caller(s) are broken. However, we may need to clarify (and possibly fix) the method semantics first. For example, adding such an XXX to mayTunnelUnsupportedProto() using option-1 semantics would be wrong/awkward because that particular method semantics should not depend on the call timing while the presence of pending pipelined requests changes with time as requests get served.

Should Squid disable CONNECT pipelining for now? In other words, if the N+1st request is a CONNECT request, should Squid wait for the first N requests to be fully served until Squid starts handing that CONNECT request? I would not be surprised if the correct answer to that question is unrelated to on_unsupported_protocol. If answering that question is difficult, we may have to place it out of this PR scope.

----

(**) Or a similar event like a client read timeout.

rousskov:
> Instead of tunnelling traffic, a matching on_unsupported_protocol "tunnel" action after a client-first or step1 bumped connection resulted to a connection abort.

> Instead of tunnelling traffic a matching on_unsupported_protocol "tunnel" action after received the first bytes of an unknown protocol tunnelled through a client-first or step1 bumped connection resulted to a connection abort.

I do not understand what the second paragraph is about. The difference between the two PR description paragraphs above is in "after received the first bytes of an unknown protocol" and in "tunnelled through" parts. I would expect that the first paragraph assumes the same "after received the first bytes of an unknown protocol" part because those bytes trigger on_unsupported_protocol processing. I do not understand the meaning/significance of the second different part at all.

Please detail the two cases to clarify the difference between them. I can polish the wording later.

rousskov:
The PR description focuses on the mishandled "tunnel" action. Was the matching "respond" action processed correctly in all those cases?

Other than removing RAM/CPU waste, this PR does not affect configurations that were only using the "respond" action, right?

chtsanti:
> 
> Please detail the two cases to clarify the difference between them. I can polish the wording later.

1st case) The CONNECT request is received and squid decides to bump on step1 or with a client-first. Squid is waiting the TLS hello message but unrecognised bytes are received. Squid needs to examine on_unsupported_protocol for the correct action, but it doesn't do it and just respond with an error.

2nd case)  The CONNECT request is received and squid decides to bump on step1 or with a client-first. Squid is waiting the TLS hello messages and then it is establishing correctly the TLS connection with the client.  Squid is waiting for the first request through bumped connection, but what it receives it is not an HTTP protocol. Squid needs to examine on_unsupported_protocol but it doesn't do it. 

Note that in the 2nd case when the first request is coming through the client-to-squid TLS(bumped) connection the  squid-to-server TLS connection may not be established yet.

The above CONNECT requests may be fake in the case of transparent https bumping.

> The PR description focuses on the mishandled "tunnel" action. Was the matching "respond" action processed correctly in all those cases?

Looks that it handled correctly. The issues I found was that squid does not examine on_unsupported_protocol in all cases it should do it, and tunnel if required. If the on_unsupported_protocol is not examined then the behaviour is "respond". Always "if possible" because it is not always possible to respond and in some cases we just closing the connection, for example after receiving unknown protocol instead of TLS data.

> Other than removing RAM/CPU waste, this PR does not affect configurations that were only using the "respond" action, right?

Except I am missing some cases, yes, only respond configurations should not affected. I checked for this also.

chtsanti:
First of all my new implementation of the mayTunnelUnsupportedProto() is again buggy. My fault (again)

> Have you tested what the new mayTunnelUnsupportedProto() causes to happen when there are one or more non-CONNECT requests still in the pipeline ahead of the CONNECT being processed? That case should also tunnel, but only after completing the required responses.

I am expecting that currently squid tunnels after completing the required responses yes. If not then probably we have problems in other cases too.
Also disabling CONNECT pipelining should not be required, but if it is required then I believe that this is because of bugs unrelated to unsupported protocol.
I did not check with pipelined requests however I can run some checks if you believe that it is required.

All the confusion come from my WRONG idea to use 'pipeline.nrequests' to get first request inside bumped connection.  It is wrong because the pipeline.nrequests counts the requests seen in the pipeline (not currently pipelined) and we may have many GET requests before got a CONNECT to bump. 

The pipeline.nrequests can be used only for transparent proxying (but also here I am using it wrongly, but it can be fixed, anyway).

So I still have the problem to count the first Http Request inside a bumped connection.
I am seeing two options for these:
1) Add a new counter to count HttpRequests inside bumped connection, or a related flag (eg ConnStateData::firstBumpedHttpRequestParsed).
2) Buffer received data for all HttpRequests inside bumped connections (wasting CPU and RAM  if bump and on_unsupported_protocol are enabled).

Opinions?

rousskov:
> Add a new counter to count HttpRequests inside bumped connection

I suggest doing that (instead of a flag or excessive buffering). The new counter may also be useful for triage. Suggested name: parsedBumpedRequests or parsedBumpedRequestCount.

rousskov:
@chtsanti, your clarifications of fixed cases helped a lot, thank you! Please review the updated PR title and description.

src/client_side.cc
chtsanti: addressed at 27122e

chtsanti:
> The new counter may also be useful for triage. Suggested name: parsedBumpedRequests or parsedBumpedRequestCount.

I used the parsedBumpedRequestCount

> I don't quite understand why a new counter is needed exactly. The Pipeline interface has a count() method for retrieving the current count of outstanding transactions if that is what you intended. 

The Pipeline::count() return the requests currently pipelined. No it is not what we need.
The new counter help us to match the state after the SSL bump is decided and before the first bumped HTTP request is received.  This is looks good solution.

> The state data should never have been mixed together like this.

Probably you are right. The ConnStateData become a huge and complex class. But will be huge project to split it.

chtsanti:
> The new counter may also be useful for triage. Suggested name: parsedBumpedRequests or parsedBumpedRequestCount.

I used the parsedBumpedRequestCount

> I don't quite understand why a new counter is needed exactly. The Pipeline interface has a count() method for retrieving the current count of outstanding transactions if that is what you intended. 

The Pipeline::count() return the requests currently pipelined. No it is not what we need.
The new counter help us to match the state after the SSL bump is decided and before the first bumped HTTP request is received.  This is looks good solution.

> The state data should never have been mixed together like this.

Probably you are right. The ConnStateData become a huge and complex class. But will be huge project to split it.

chtsanti:
> The new counter may also be useful for triage. Suggested name: parsedBumpedRequests or parsedBumpedRequestCount.

I used the parsedBumpedRequestCount

> I don't quite understand why a new counter is needed exactly. The Pipeline interface has a count() method for retrieving the current count of outstanding transactions if that is what you intended. 

The Pipeline::count() return the requests currently pipelined. No it is not what we need.
The new counter help us to match the state after the SSL bump is decided and before the first bumped HTTP request is received.  This is looks good solution.

> The state data should never have been mixed together like this.

Probably you are right. The ConnStateData become a huge and complex class. But will be huge project to split it.

chtsanti:
There is a new commit 27122e please re-review.

rousskov:
> Pre-bumping HTTP, TLS handshake protocol, and after-bumping HTTPS are supposed to be mutually exclusive uses of a socket

As your own questions about pipelining illustrate, pipelining is incompatible with the idea of exclusive socket access. And HTTP/2 requirements make shared control absolutely necessary. Ideally, a single socket can be used by plain HTTP, TLS, and bumping HTTP at the same time (and that list is not exhaustive).

ConnStateData should be aware of (and coordinating) those individual pieces/layers (all tied to a to-Squid TCP connection). We should extract layer-specific work/state into layer-specific classes, but I discourage you from taking up that challenge, especially now. If you do decide to work on those proposals, please start by re-merging Server and ConnStateData back together.

src/client_side.h
```diff
@@ -394,9 +394,11 @@ class ConnStateData : public Server, public HttpControlMsgSink, private Independ
 #if USE_OPENSSL
     bool switchedToHttps_;
     bool parsingTlsHandshake; ///< whether we are getting/parsing TLS Hello bytes
+    uint32_t parsedBumpedRequestCount; ///< The number of bumped HTTP requests
```
rousskov: Please use either 

* `unsigned int` (i.e. the simplest type that still works OK) if you do not care about overflows or
*  `uint64_t` (i.e., the type that works best) if you do care about overflows.

I recommend the latter.

rousskov: AFAICT, this new member is more accurately described as `the number of successfully parsed HTTP request headers on a bumped client connection`. An more accurate description here may help answer the "how is this counter differs from the ones we already have?" questions.

src/client_side.cc
```diff
@@ -4103,13 +4122,22 @@ ConnStateData::checkLogging()
...
 {
-    return Config.accessList.on_unsupported_protocol
+    if (!Config.accessList.on_unsupported_protocol)
+        return false;
+
 #if USE_OPENSSL
-           &&
-           ((port->flags.isIntercepted() && port->flags.tunnelSslBumping)
-            || (serverBump() && pinning.serverConnection))
+    // SslBump processing:
```
rousskov: The `1)` and `2)` items below can be misinterpreted as two "steps" rather than different "states" (handled by the same `if` statement).

```suggestion
    // Tunneling may be possible when parsing the TLS client handshake
    // and when parsing the first HTTP request on a bumped connection.
```

src/client_side.cc
```diff
@@ -4103,13 +4122,22 @@ ConnStateData::checkLogging()
...
+    //  1) parsing the TLS client handshake
+    //  2) reading and processing the first bumped HTTP request
+    if (sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1)
+        return true;
 #endif
-           ;
+
+    // the first request in a connection to a plain intercepting port
+    if (transparent() && pipeline.nrequests <= 1)
```
rousskov: AFAICT, transparent() does not imply "plain" mentioned in your comment. If you want to exclude TLS connections, then you need more checks.

src/client_side.cc
```diff
@@ -3050,7 +3057,10 @@ ConnStateData::switchToHttps(HttpRequest *request, Ssl::BumpMode bumpServerMode)
 {
     assert(!switchedToHttps_);
 
-    sslConnectHostOrIp = request->url.host();
+    static char ip[MAX_IPSTRLEN];
+    sslConnectHostOrIp =  request->url.hostIsNumeric() ?
+        request->url.hostIP().toStr(ip, sizeof(ip)) :request->url.host();
```
rousskov: Please avoid duplicating this code -- add something like a `url.hostOrIp()` method if needed. BTW, why do you need to change sslConnectHostOrIp value? Does that change fix a bug?

src/client_side.h
rousskov: The `successfully` part of my suggestion may be wrong -- please double check whether we are counting successful parsing events or just parsing attempts (the last of which may have failed).

src/client_side.cc
chtsanti: OK I added and use the Anyp::Uri::hostOrIp method.

Also I changed the way the ClientHttpRequest::uri computed inside buildFakeRequest was buggy.

> why do you need to change sslConnectHostOrIp value? Does that change fix a bug?

Nope. just to use it with initiateTunneledRequest.

But I just realised that the sslConnectHostOrIp is used to generate certificates.  This fix affects ssl_crtd daemon results in the case we are connecting to an ipv6 TLS server and no SNI exist nor a hostname in related CONNECT request.

Current squid code in this case will generate '[ipv6_address]' as sslConnectHostOrIp (inside '[]' brackets) and this patch will changes this behaviour and will generate a simple 'ipv6_address'. This is will cause squid to ignore currently cached certificates and generate new ones.

My opinion is that this is not a huge problem and it will happen only for a small part of cached certificates.

Also looks that we are not generating correctly such certificates (with ip as CN which is not checked by browsers) but we have to build also a subjectAltName with an ip entry:
    https://tools.ietf.org/html/rfc5280#section-4.2.1.6

Opinions?

src/client_side.h
yadij: FYI: HTTP/1.x has no defined limit on pipeline lengths (it is only a tradition to use 5-10) and HTTP/2 and HTTP/3 have a 2^31 requirement which our Pipeline was being prepared for. Remember these are counters for accounting the entire potential week/month/year long lifetime of the connection.

chtsanti:
> I cannot fully review the code changes yet because you have not clarified the intended mayTunnelUnsupportedProto() scope/definition/semantics as discussed earlier.

I am a little confused with your questions. However I will try to describe for what the mayTunnelUnsupportedProto() is currently used:
1) to initialise related structures to preserve received data in the future. This is done before any data re received
2) To preserve received data
3) After we found unsupported protocol data call this function to decide if we should start acl check and finally splice if required or not.

 Actually we have to answer on two questions:
 1) Are we expecting unsupported data now or on next steps? 
 2) Are we configured to tunnel an unsupported protocol?

The answer to (1) is  (a) while we are trying to Bump a TLS connection, (b)when receives the first data after bumped connection is established, (c) while trying to do transparent proxing.

chtsanti:
@rousskov  The commit 4dd1ac8 should address your requests.
Please re-review

yadij:
If I am understanding you right then this method name is slightly off - what it is actually stating is the answer to the question "is (future) tunnel still possible?". So it is not specific to unsupported protocols, but can be shared with splice or any protocol we *do* support which require tunneling.

It is the specific callers which have the knowledge about "unsupported" protocols and/or whether it is a tunnel-now situation or defer-delivery situation, not the test itself. The test only needs to know if something already happened that prevents tunnel.

So perhapse name the the function clearer, and do its internal checks by protocol. Along these lines:

bool canTunnel() const {
   if [TLS thing(s) that block tunneling]
    return false;
   if [HTTP thing(s) that block tunneling]
    return false;
   if [protocol X thing(s) that block tunneling]
    return false;
  return true;
}

chtsanti:
> If I am understanding you right then this method name is slightly off - what it is actually stating is the answer to the question "is (future) tunnel still possible?". So it is not specific to unsupported protocols, but can be shared with splice or any protocol we _do_ support which require tunneling.

Maybe but not sure.
A future decision for tunnelling maybe is possible after the Nth  request is received for example while we are upgrading to a non HTTP protocol. But if we are talking about an unsupported protocol this is should detected before or while the first bytes are received.

I do not believe we should consider these cases as similar. For the "upgrading to an other protocol" case, may not required at all a mayTunnel() method. And for sure it will have completely different restrictions.

> 
> It is the specific callers which have the knowledge about "unsupported" protocols and/or whether it is a tunnel-now situation or defer-delivery situation, not the test itself. The test only needs to know if something already happened that prevents tunnel activities by the caller.

I am not sure that this is the case.

> 
> So perhapse name the function clearer, and do its internal checks by protocol. Along these lines:
> 
> bool canTunnel() const {
> if [TLS thing(s) that block tunneling]
> return false;
> if [HTTP thing(s) that block tunneling]
> return false;
> if [protocol X thing(s) that block tunneling]
> return false;
> return true;
> }

I think there are not thinks which are blocking tunnelling but instead there are cases where we are allowing tunnelling.
 
Please look again in latest "mayTunnelUnsupportedProto" to see in which cases we are allowing tunnelling.

yadij:
> > If I am understanding you right then this method name is slightly off - what it is actually stating is the answer to the question "is (future) tunnel still possible?". So it is not specific to unsupported protocols, but can be shared with splice or any protocol we _do_ support which require tunneling.
> 
> Maybe but not sure.
> A future decision for tunnelling maybe is possible after the Nth request is received for example while we are upgrading to a non HTTP protocol. But if we are talking about an unsupported protocol this is should detected before or while the first bytes are received.
> 

That was part of my thinking - there are a lot of cases and conditions under which tunnel MIGHT happen later, and a finite set of conditions which absolutely prevent it. So we can check for those smaller set of limitations and default to tunneling unless a reason is known not to.

> I do not believe we should consider these cases as similar. For the "upgrading to an other protocol" case, may not required at all a mayTunnel() method. And for sure it will have completely different restrictions.
> 

Such as?  TLS can be nested in CONNECT, TLS inside TLS, and port 80 can receive unknown protocol - so IMO the cases do have overlaps, we just do not have support for them all yet. This is a good time to prepare for handling those cases by making the test method/function not depend on the exact traffic type being handled.

> > It is the specific callers which have the knowledge about "unsupported" protocols and/or whether it is a tunnel-now situation or defer-delivery situation, not the test itself. The test only needs to know if something already happened that prevents tunnel activities by the caller.
> 
> I am not sure that this is the case.
> 

For the cases we have already and a few I expect we will be adding in the mid-term future I am pretty sure that is true. Long-term we may want something else entirely, but that is going to need a huge amount more design and time to find out what the future actually needs.

> > So perhapse name the function clearer, and do its internal checks by protocol. Along these lines:
...
> 
> I think there are not thinks which are blocking tunnelling but instead there are cases where we are allowing tunnelling.
> 

I am basing this design around the issue that some of the callers need to do preparation for future tunneling.

If we start with the assumption/default that tunnel is not possible until allowed, those prep will be calling this test function during a time before tunnel is allowed - thus they would be told "no" and skip the required prep. Inverting the test to something like if(!forbidTunnel()) it would produce "!false == true" and do the prep for the cases when it was absolutely not possible to do.

On the flip side, once one of the things happen that prevent tunneling (usually bytes sent) it tends to stay impossible to tunnel. So we can let the callers do prep (ie store client bytes for later use or delivery) and cancel that later (ie send those saved bytes in a tunnel).


> Please look again in latest "mayTunnelUnsupportedProto" to see in which cases we are allowing tunnelling.

I do not have a strong preference for "can" vs "may", but it definitely should avoid tying it by name to only ever be used regarding "UnsupportedProto".

src/client_side.cc
rousskov: > Also looks that we are not generating correctly such certificates (with ip as CN which is not checked by browsers) but we have to build also a subjectAltName with an ip entry:
https://tools.ietf.org/html/rfc5280#section-4.2.1.6

> Opinions?

If I understand the question correctly, you are asking whether Squid should change how IP-based certificates are generated. I think that question should not be asked during this PR. We have enough problems here without that question! I do not know the answer to the question though -- the RFC text you linked to appears to talk about how to add an IP address as an Subject Alternative Name, not which fields/names should be used when generating an IP-based certificate. If you agree that this problem is out of this PR scope, then let's not try to answer that question here. Please file a bug report instead.

rousskov: I am resolving this change request by adding the following paragraph to the PR description (based on my interpretation of your answer). As always, please edit the PR description as needed so that it reflects your understanding of the current PR state.

> Also changed how IPv6-based certificates are generated. Their CN field value is no longer surrounded by [square brackets]. This change was done to streamline Squid code. It affects certificate cache key so old IPv6-based certificates will never be found (and will eventually be purged) while new ones will be generated and cached instead. We believe these IPv6-based certificates are rare and untrusted by browsers so the change in their CN should not have a significant affect on users.

src/client_side.cc
rousskov: I am resolving this change request although I am not sure that the "when A or when B" comment actually matches the "X && Y" code. I will ask you to clarify why there is a match in the next review.

src/client_side.h
rousskov: > The "successfully" is not accurate.

Agreed, but _not_ because of:

> Later we are doing more checks and the request may dropped.

The "successfully" part is wrong because the counter is incremented when parseOneRequest() returns a non-nil pointer, indicating either that the request was successfully parsed or that parsing failed. Thus, the new counter counts failure cases as well as successful ones. The adjusted description reflects that.

I am resolving this change request.

src/client_side.cc
rousskov: > The TLS connections in these cases should handled (and always return true) in previous checks.

I see no evidence of that. AFAICT, the only "previous check" that returns in this method uses the following condition:

```
USE_OPENSSL && sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1
```

How could that possibly handle all transparent() TLS connections?! The trivial counter-example is `parsedBumpedRequestCount == 100`.

I will resolve this change request here but will reopen it in the new review because this problem has not been addressed.

src/client_side.cc
```diff
@@ -4103,13 +4119,21 @@ ConnStateData::checkLogging()
...
-           ((port->flags.isIntercepted() && port->flags.tunnelSslBumping)
-            || (serverBump() && pinning.serverConnection))
+    // Tunneling may be possible when parsing the TLS client handshake
+    // and when parsing the first HTTP request on a bumped connection.
+    if (sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1)
+        return true;
 #endif
-           ;
+
+    // the first request in a connection to a plain intercepting port
```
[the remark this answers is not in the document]
chtsanti: Well, in this case this method will not called at all.
We can add a check for port->secure.encryptTransport too.

I did not resolve it here please check "split mayTunnelUnsupportedProto() into two functions" related comments about.

src/client_side.cc
```diff
@@ -4103,13 +4119,21 @@ ConnStateData::checkLogging()
...
+    if (!Config.accessList.on_unsupported_protocol)
+        return false;
+
 #if USE_OPENSSL
-           &&
-           ((port->flags.isIntercepted() && port->flags.tunnelSslBumping)
-            || (serverBump() && pinning.serverConnection))
+    // Tunneling may be possible when parsing the TLS client handshake
+    // and when parsing the first HTTP request on a bumped connection.
+    if (sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1)
```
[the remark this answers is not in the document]
chtsanti: There is not a safe check for "while parsing the TLS handshake".  While we are parsing tls handshake we are always preserve data to be able to splice if required by configuration.
In all cases the mayTunnelUnsupportedProto is called after the sslBumpMode is set and while parsedBumpedRequestCount is less or equal to 1.

I did not resolve it here please check "split mayTunnelUnsupportedProto() into two functions" related comments about.

src/anyp/Uri.cc
```diff
@@ -59,6 +59,14 @@ AnyP::Uri::host(const char *src)
     touch();
 }
 
+SBuf
+AnyP::Uri::hostOrIp() const
+{
+    static char ip[MAX_IPSTRLEN];
+    SBuf hostOrIp_(hostIsNumeric() ? hostIP().toStr(ip, sizeof(ip)) : host());
```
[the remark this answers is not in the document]
chtsanti: resolved at commit 5a693724

src/anyp/Uri.h
```diff
@@ -80,6 +80,9 @@ class Uri
     int hostIsNumeric(void) const {return hostIsNumeric_;}
     Ip::Address const & hostIP(void) const {return hostAddr_;}
 
+    /// \return the hostname or the unformatted ip address
```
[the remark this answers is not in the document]
chtsanti: resolved at commit 5a693724

src/client_side.cc
```diff
@@ -3050,7 +3057,8 @@ ConnStateData::switchToHttps(HttpRequest *request, Ssl::BumpMode bumpServerMode)
 {
     assert(!switchedToHttps_);
 
-    sslConnectHostOrIp = request->url.host();
+    sslConnectHostOrIp =  request->url.hostOrIp();
```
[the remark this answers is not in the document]
chtsanti: This is resulted to other 4-5  more unrelated changes. I did not used a new class, just renamed the sslConnectHostOrIp to tlsConnectHostOrIp.

Resolved at commit e01be963

chtsanti:
@rousskov @yadij 
please check my latest commits.

> Please split mayTunnelUnsupportedProto() into two functions:
please check my commit 

>     1. preserveDataForTunnellingUnsupportedProto() -- used before Squid starts parsing (and potentially consuming) a request. In some cases, such parsing may lead to an unsupported protocol detection, of course.
>     2. ableToTunnelUnsupportedProto() -- used after Squid detects an unsupported protocol.

The patch does something like that.

> 
> 
> If _needed_, it is OK for preserveDataForTunnellingUnsupportedProto() to save some state/flag that ableToTunnelUnsupportedProto() can check as one of the preconditions for returning true.

It uses the flag ableToTunnelUnsupportedProto_.
Noted that this flag is also directly set by ConnStateData::parseTlsHandshake.

> We can polish the function names later as their implementation and use cases become clear.

The names are not correct. However I do not like the result too. Personally I prefer to use the previous form (mayTunnelUnsupportedProtocol), even if it was not correct too.

The problem I am seeing is that even if the preserveDataForTunnelingUnsupportedProto decided at the begging of transaction that we can tunnel on error, this is may not be true after a while (for example after two requests executed). 

I think that it is better than this form  to directly set/unset  the ableToTunnelUnsupportedProto_ in various places inside client_side.cc.  But this is will be complex too.

src/client_side.cc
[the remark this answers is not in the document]
rousskov: Please feel free to undo https://github.com/squid-cache/squid/commit/e01be9639713d226c93ac6ee77b0ca5c407fb3c9 if you believe that renaming causes too many otherwise unnecessary changes -- I did start my change request with an `If...` precondition that can be false.

If you keep the renamed member, then please fix `HERE` and also `s/SSL/TLS/` in the renamed data member description, to improve symmetry with the sister data member.

src/client_side.cc
```diff
@@ -2275,7 +2283,7 @@ ConnStateData::requestTimeout(const CommTimeoutCbParams &io)
     if (!Comm::IsConnOpen(io.conn))
         return;
 
-    if (mayTunnelUnsupportedProto() && !receivedFirstByte_) {
+    if (!receivedFirstByte_ && preserveDataForTunnellingUnsupportedProto()) {
```
rousskov: I would expect to see an ableToTunnelUnsupportedProto() check here. preserveDataForTunnellingUnsupportedProto() decides whether to preserve data. We should have decided that earlier, _before_ the timeout. Here, after a timeout, we can only check whether the preconditions for tunneling are satisfied (probably _including_ the !receivedFirstByte_ precondition, but I am not sure about that):

```suggestion
    if (ableToTunnelUnsupportedProto()) {
```

src/client_side.cc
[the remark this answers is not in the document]
rousskov: > Well, in this case this method will not called at all.

That is not what the code implies. Please note that I am not saying that this method will or will not be called for a GnuTLS https_port. I am saying that the current method code (including comments) are either confusing (at best) or lying (at worst). The code (possibly just comments -- not sure) needs to be adjusted to make sense. For _example_, if this code is specific to OpenSSL builds, then the entire method should be surrounded by USE_OPENSSL. If it is not specific to OpenSSL builds, then perhaps there should be an `#else` clause for other builds. I do _not_ know what the exact solution is, but we should not say `the first request in a connection to a plain intercepting port` when the code clearly implies that other cases may also be possible.

src/client_side.cc
```diff
@@ -4117,23 +4121,25 @@ ConnStateData::checkLogging()
...
 {
-    if (!Config.accessList.on_unsupported_protocol)
-        return false;
-
 #if USE_OPENSSL
     // Tunneling may be possible when parsing the TLS client handshake
     // and when parsing the first HTTP request on a bumped connection.
-    if (sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1)
-        return true;
+    bool firstBumpedRequest = sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1;
```
rousskov: The comment says `Tunneling may be possible when A and when B`. The boolean variable is called `whenB`. What happened to A?! I do not know what the exact solution is, but the code (including comments and variable names) does not make sense right now. Something needs to be changed!

> There is not a safe check for "while parsing the TLS handshake".

I do not understand what that means or implies in this specific context. The comment is lying about us checking that condition? The code is unsafe? The comment is irrelevant to the code it is covering? The existing check covers both cases and cannot (or should not) be split into two (but should be renamed)? The correct solution to address this problem may depend on the correct answers to those questions. I do not know those answers, but you probably do.

> While we are parsing tls handshake we are always preserve data to be able to splice if required by configuration.

OK.

>  In all cases the mayTunnelUnsupportedProto is called after the sslBumpMode is set and while parsedBumpedRequestCount is less or equal to 1.

I do not know what you mean by "all cases". Since we are checking that condition, that condition is probably not true in some cases. The variable name should reflect what the condition means.

src/client_side.cc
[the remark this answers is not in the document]
rousskov: The problem remains, but I am marking this change request as resolved and moving the discussion into the new review where better/fresh context is available.

src/client_side.cc
```diff
@@ -1253,7 +1253,10 @@ parseHttpRequest(ConnStateData *csd, const Http1::RequestParserPointer &hp)
             return NULL;
         }
 
-        if (csd->mayTunnelUnsupportedProto()) {
+        if (csd->ableToTunnelUnsupportedProto()) {
+            // The preserveDataForTunnelingUnsupportedProto is already checked
+            // inside Http::One::Server::parseOneRequest and the hp parser
+            // must preserved parsed data.
```
rousskov: This overly detailed comment is very likely to become stale (and hence confusing) soon IMHO. There is already a Must() inside the parsed() method that will trigger if this overall logic fails. And the reader can check what makes ableToTunnelUnsupportedProto() true. I think that is enough and suggest removing this comment.

If you insist on keeping the comment:
```suggestion
            // was told to preserve its parsed data.
```

I am also uncomfortable with the fact that this code is _not_ inside Http::One::Server -- it is not obvious that it does not apply to FTP requests.

src/client_side.cc
```diff
@@ -4101,15 +4121,25 @@ ConnStateData::checkLogging()
...
 bool
-ConnStateData::mayTunnelUnsupportedProto()
+ConnStateData::preserveDataForTunnellingUnsupportedProto()
 {
-    return Config.accessList.on_unsupported_protocol
 #if USE_OPENSSL
-           &&
-           ((port->flags.isIntercepted() && port->flags.tunnelSslBumping)
-            || (serverBump() && pinning.serverConnection))
+    // Tunneling may be possible when parsing the TLS client handshake
```
rousskov: If this method is not called when we are parsing the TLS client handshake, please remove that condition. I also suggest renaming this method to preserveHttpBytesForTunnellingUnsupportedProto() or similar, to underscore that it only deals with plain HTTP requests and decrypted HTTPS requests.

src/client_side.cc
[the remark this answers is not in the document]
chtsanti: With the "I did not resolve it here" I was meaning that probably whis was should discussed with the "split mayTunnelUnsupportedProto() into two methods" related comments. Sorry.

This code changed in commit fd887c2be ("split mayTunnelUnsupportedProto() into two methods") to be:
>`// the first request in a connection to a plain intercepting port`
> `bool firstTransparentPlainRequest = !port->secure.encryptTransport && transparent() && pipeline.nrequests <= 1;` 

I think this is now handles correctly the "first conenction in plain intercepting port".

src/client_side.cc
chtsanti: Well. You are right, however not easy to find all earlier stages where the decision should taken. Moreover not easy to make a function preserveDataForTunnellingUnsupportedProto() which should take the decision in various cases.

In the latest commit, I am setting directly the ConnStateData::ableToTunnelUnsupportedProto_  inside ConnStateData::switchToHttps (after this is called we are expecting TLS hello data). This member also set directly inside ConnStateData::parseTlsHandshake(), and it is reset after the first byte received (inside ConnStateData::receivedFirstByte).

I renamed the preserveDataForTunnellingUnsupportedProto()  to be a preserveHttpForTunnellingUnsupportedProto()  and it is now used only while we are expecting plain HTTP requests. It is used:
  * Http::One::Server::start() to detect timeouts before any HTTP request is received inside bumped connections or in intercepted plain HTTP connections
  * and in Http::One::Server::parseOneRequest to recognise the 1st HTTP request on bumped connections or in intercepted plain HTTP connections.

Please check my commit e757c76

src/client_side.cc
chtsanti: > The comment says Tunneling may be possible when A and when B. The boolean variable is called whenB. What happened to A?

The A (was/)is handled setting directly inside the ConnStateData::parseTlsHandshake() method.

> Or that this method should be called preserveHttpBytesForTunnellingUnsupportedProto because it only deals with plain HTTP and decrypted HTTPS bytes?

Yes. I used this name in my commit e757c76

src/client_side.cc
```diff
@@ -4117,23 +4121,25 @@ ConnStateData::checkLogging()
...
     // and when parsing the first HTTP request on a bumped connection.
-    if (sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1)
-        return true;
+    bool firstBumpedRequest = sslBumpMode != Ssl::bumpEnd && parsedBumpedRequestCount <= 1;
 #endif
 
     // the first request in a connection to a plain intercepting port
-    if (transparent() && pipeline.nrequests <= 1)
-        return true;
+    bool firstTransparentPlainRequest = !port->secure.encryptTransport && transparent() && pipeline.nrequests <= 1;
```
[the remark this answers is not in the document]
chtsanti: resolved at commit d90977f

src/client_side.cc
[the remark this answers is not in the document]
chtsanti: resolved at commit d90977f

src/servers/Http1Server.cc
```diff
@@ -78,7 +78,7 @@ Http::One::Server::parseOneRequest()
     // a) do not have one already
     // b) have completed the previous request parsing already
     if (!parser_ || !parser_->needsMoreData())
-        parser_ = new Http1::RequestParser(mayTunnelUnsupportedProto());
+        parser_ = new Http1::RequestParser(preserveDataForTunnellingUnsupportedProto());
```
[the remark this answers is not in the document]
chtsanti: > method and changing this code to
```
ableToTunnelUnsupportedProto_ = preserveDataForTunnellingUnsupportedProto();
parser_ = new Http1::RequestParser(mayTunnelUnsupportedProto());
```
I used this scheme.

 > if preserveDataForTunnellingUnsupportedProto() becomes a decision-only method we can rename it to something like shouldPreserveDataForTunnellingUnsupportedProto().

I am using the name  preserveHttpBytesForTunnellingUnsupportedProto() but we can change it.

please check commit e757c76

src/client_side.cc
chtsanti: Ok, this condition is removed. The preserveHttpBytesForTunnellingUnsupportedProto() name is used.
Resolved at commit e757c76

src/client_side.cc
chtsanti: I removed the comment in commit e757c76

> I am also uncomfortable with the fact that this code is not inside Http::One::Server -- it is not obvious that it does not apply to FTP requests.

Not easy to move it here.
I do not know if we can support ftp protocol. Are we supporting transparent FTP proxying?

chtsanti:
@rousskov, @yadij 

I applied two new commits.
I am not sure that the patch is final (probably we should rename again members and methods) but I tried to handle the problems or at least to show the problems I  found trying to solve various problems.

Please re-review

src/client_side.cc
rousskov: > Are we supporting transparent FTP proxying?

Yes, AFAICT: `ftp_port` is explicitly documented to accept `intercept` and `tproxy` modes. We do not support on_unsupported_protocol for FTP though (yet?).

However, perhaps this is not important: I said that it is not obvious that this code does not apply to FTP requests (and, hence, should be moved to Http::One::Server), but it can be argued that it is obvious enough -- there are many HTTP-specific words/names in this context. Thus, I am OK with the code staying where it is if you think its current location works well.

Thank you for addressing my primary concern in this change request. I am resolving this change request.

src/client_side.cc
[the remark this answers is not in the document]
chtsanti: addressed at 7c79ddfb

chtsanti:
My sense is that the smaphoreci and Jenkins build are stuck.
I rebased to latest master and jenkings reports that it can not merge with master. And the PR waits for the semaphorci status for many hours.

rousskov:
Sempahore seems confused by your force-pushing -- it keeps rebuilding the old (no longer associated with your branch) commit https://github.com/squid-cache/squid/commit/d90977f31224c6ba9605fa1b54f94bbf7a9bc7e4 instead of trying to build the new branch tip (https://github.com/squid-cache/squid/pull/339/commits/7c79ddfb104c081bc30b7399b567b622210158a9).

Jenkins [claims](https://build.squid-cache.org/job/5-pr-test/1114/console) that commit 7c79ddfb104c081bc30b7399b567b622210158a9 has merge conflicts (presumably with master). Is Jenkins lying?

N.B. I have committed a few changes but have not pushed them yet. Still working on them, but hope to finish next week. Hopefully, Sempahore will realize what is going on when I push new commits (without force-pushing).

chtsanti:
> Jenkins claims that commit 7c79ddf has merge conflicts (presumably with master). Is Jenkins lying?

How jenkins works? The commit 7c79ddf can not be applied directly over master. The branch has about 10 commits. Is it possible that jenkins miss 1 or 2 of the commits?
The branch now is based on master a75bf5d. It is only one (completely unrelated) commit behind of latest master , should not have any problem.

chtsanti:
> @chtsanti, thank you for getting this PR to the point where I could polish it further. I pushed several untested commits in hope to address most of the remaining concerns. Some of my changes were substantial, but I tried to leave detailed commit messages. Please review the changes (including the updated PR description) and thoroughly test them after addressing the following XXXs:
> 
>     * [x]  XXX: sslBumpMode != bumpEnd includes non-bumped bumpSplice and bumpNone.

Please review my 3bab87d commit.  It uses two new if/checks instead of the old one check. 
Also moves the code which sets the preserviingClientData_ flag at the end of  switchToHttps method after setting the  parsingTlsHandshake flag.
I hope it is ok.

> 
>     * [x]  XXX: Exclude FTP ports. They do not support on_unsupported_protocol.

 commit ef9db3f5

chtsanti:
We are mixed up reviewers and authros into this PR, I am not able to make a github review, so I will post as simple comment my findings.
Inside ConnStateData::requestTimeout: 
```
if (tunnelOnError(HttpRequestMethod(), ERR_REQUEST_START_TIMEOUT))
        return;
```
The requestTimeout method will be called in the following cases:
 - After a request_start_timeout
 - After a request_timeout

We need to tunnel inside this method to handle the case the client waits a server message imediatelly after connect, eg an SSH or IMAP protocol. If some bytes are received we are able to check if we have an HTTP protocol or a foreign protocol.

However the requestTimeout method also tunnels after a HTTP request timeout (eg the first line of HTTP request is received, but the next headers, or the end of headers  did not received in a request_timeout period).  The additional check for "receivedFirstByte_" is required before call tunnelOnError().

In general the final code works well. Still the ConnStateData::preservingClientData may lay in some cases, eg we received POST request headers, we are reading request body, we are not actually still preserving data in this case).  But these cases does not affect squid operation,  we are not asking to tunnel on those cases.

rousskov:
> the requestTimeout method also tunnels after a HTTP request timeout (eg the first line of HTTP request is received, but the next headers, or the end of headers did not received in a request_timeout period).

The requestTimeout() method checks on_unsupported_protocol. Whether Squid will tunnel after a request reading/parsing timeout, should depend on the admin decision. No hard-coded logic will cover all useful use cases!

I think we should give the admin enough information to make the right decision for their environment. If there is no ACL that can match !receivedFirstByte_ state, then let's split ERR_REQUEST_START_TIMEOUT into 

* ERR_REQUEST_START_TIMEOUT (has not receivedFirstByte_ yet) and
* ERR_REQUEST_PARSE_TIMEOUT (has receivedFirstByte_ already).

Note that the proposed split can be considered backward-compatible:

* Old configurations that use ERR_REQUEST_START_TIMEOUT explicitly will continue to work as they did (i.e. closing the connection instead of tunneling the ERR_REQUEST_PARSE_TIMEOUT cases).
* Old configurations that use "all" ACL to tunnel on all errors will also preserve their "tunnel everything" behavior if we interpret "all errors" or "everything" as "all problems recognized/supported by this version of Squid", which is a reasonable interpretation.

> The additional check for "receivedFirstByte_" is required before call tunnelOnError().

While that receivedFirstByte_ check would indeed address the specific use case you are describing, I do not think we should hard-code the "no tunneling" decision for non-HTTP requests that, for example, start with a single command word and then wait for the server to respond. If we can easily give the admin the control over what to tunnel (and, in some sense, over "unsupported protocol" detection), then we should.

What do you think about splitting ERR_REQUEST_START_TIMEOUT into to distinct errors, reflecting the difference in the receivedFirstByte_ state?

rousskov:
> ConnStateData::preservingClientData [lies after] we received POST request headers, we are reading request body, we are not actually still preserving data in this case

Should we fix this by clearing preservingClientData_ after successfully parsing request headers? At that time, we know that we are not dealing with an unsupported protocol anymore.

chtsanti:
> What do you think about splitting ERR_REQUEST_START_TIMEOUT into to distinct errors, reflecting the difference in the receivedFirstByte_ state?

Does not looks bad idea to give this functionality/choice to admins. 
By the way I just realised that we are not documenting at all the squid_error acl : )

> Should we fix this by clearing preservingClientData_ after successfully parsing request headers? At that time, we know that we are not dealing with an unsupported protocol anymore.

Mixed feelings about. Because we are not using this flag after successfully parsing request headers we do not really need it. Also we may forget cases where such reset is needed (eg after bumping). Trying to reset this flag may make related code more complex.
We can just comment that it is only used and set while parsing headers and client tls hello message.

chtsanti:
> What do you think about splitting ERR_REQUEST_START_TIMEOUT into to distinct errors, reflecting the difference in the receivedFirstByte_ state?

implemented at 707831c45

chtsanti:
@rousskov 
I (hope I) handle all of your requests: 
  - https://github.com/squid-cache/squid/pull/339#pullrequestreview-229860135
    commits 3bab87d and ef9db3f
  - https://github.com/squid-cache/squid/pull/339#issuecomment-487648956
   commit 707831c

Please re-review.

After the review I need to handle the following TODOs:
   - TODO: Convert parseHttpRequest() into a method and inline this call.
  - TODO: Remove this change-reducing hack.

chtsanti:
> I pushed a few source code formatting and comment changes. No functionality changes. Please check that I did not misrepresent something.

Looks ok.

> 
> I added one TODO comment that should be addressed in this PR: We should briefly explain why FTP port has to be excluded from on_unsupported_protocol support. Intercepted ftp_port knows where the traffic is going, so I would expect us to be able to tunnel that traffic at TCP level. Is this about DATA transfers? Active, passive, or both? Something else? In other words, what do we expect to break if we remove the following condition (now marked with a TODO):
> 
> ```c++
>     // TODO: Explain why we do not tunnel intercepted ftp_port bytes.
>     if (port->transport.protocol == AnyP::PROTO_FTP)
>         return false;
> ```

When we build this feature we have the requirement for HTTP port.  For those ports we just used the pre-existing Http Tunnel  code.  We did not check for Ftp port at all.

The reasons does not have to do with active or passive DATA requests. On ftp port the incoming requests are for the control socket. But supporting correctly on ftp port probably require changes on FtpServer.cc file. We never had check about.

I think a comment like "TODO: support ftp_port" is enough.
Maybe to implement such support requires to just remove the check for ftp ports.. Maybe require more changes. But this is a different project.

> 
> Since there were no change requests from @yadij, the following old polishing/change-minimization TODOs can also be completed now:
> 
>     * [ ]  TODO: Convert parseHttpRequest() into a method and inline this call.

done in commit 0946a02d.
I started move the parseHttpRequest in Http::One::Server but required to move a lot of code and I stopped. I think it is a huge change to be included with this issue.

> 
>     * [ ]  TODO: Remove this change-reducing hack.

The commits d61bf2 and ae6a86

> As discussed earlier, the following new TODOs are outside this PR scope IMO (but can and perhaps should be done separately shortly after):
> 
>     * ConnStateData::tunnelOnError() wrapper. Reduces code changes. TODO: Remove.
> 
>     * TODO: move to the protected section when removing clientTunnelOnError()

Yep.

rousskov:
> I think a comment like "TODO: support ftp_port" is enough. Maybe to implement such support requires to just remove the check for ftp ports.. Maybe require more changes. But this is a different project.

Aha, it is now clear to me that we do not know what it takes to support FTP tunneling. That is OK! I thought we knew. I agree that adding the corresponding comment is enough. I suggest:

```C++
// TODO: Figure out whether/how we can support FTP tunneling.
```

rousskov:
> Aha, it is now clear to me that we do not know what it takes to support FTP tunneling. That is OK! I thought we knew. I agree that adding the corresponding comment is enough. I suggest:
> 
> ```c++
> // TODO: Figure out whether/how we can support FTP tunneling.
> ```

I added that comment.

rousskov:
If there are no last-minute objections, I will clear this PR for commit.

yadij:
No objections from me.

rousskov:
@yadij, will this fix be eventually included in v4? We can help with porting if needed.

rousskov:
> @yadij, will this fix be eventually included in v4? We can help with porting if needed.

That v4 port is #461. Please use it if you need it or just close that PR if you do not.

PR #344, opened 2018-12-29
Bug 4856: Exit when GoIntoBackground() fork() call fails

marcosfrm:
Ignoring the (implicit/default) command-line instruction to background
Squid is wrong in general and confuses at least some daemon managers.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1860,13 +1860,12 @@ GoIntoBackground()
     pid_t pid;
     if ((pid = fork()) < 0) {
         int xerrno = errno;
-        syslog(LOG_ALERT, "fork failed: %s", xstrerr(xerrno));
-        // continue anyway, mimicking --foreground mode (XXX?)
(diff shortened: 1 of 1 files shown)


src/main.cc
```diff
@@ -1860,13 +1860,12 @@ GoIntoBackground()
     pid_t pid;
     if ((pid = fork()) < 0) {
         int xerrno = errno;
-        syslog(LOG_ALERT, "fork failed: %s", xstrerr(xerrno));
-        // continue anyway, mimicking --foreground mode (XXX?)
+        fatalf("fork failed: %s", xstrerr(xerrno));
```
rousskov: If possible, please throw an exception instead. Modern Squids use C++ exceptions for error reporting/handling (where those exceptions work and are appropriate). This is probably one of those works-and-appropriate cases.
```suggestion
        throw TexcHere(ToSBuf("failed to fork(2) the master process: ", xstrerr(xerrno)));
```

rousskov:
OK to test

PR #347, opened 2019-01-01
basic_ldap_auth: Return BH on internal errors; polished messages

amishmm:
Basic LDAP auth helper now returns BH instead of ERR in case of errors
other than LDAP_SECURITY_ERROR, per helper guidelines.

Motivation: I have a wrapper around Basic LDAP auth helper. If an LDAP
server is down, then the helper returns BH, and the wrapper uses
a fallback authentication source.

Also converted printf() to SEND_*() macros and reduced message
verbosity.

### DIFF
--- a/src/auth/basic/LDAP/basic_ldap_auth.cc
+++ b/src/auth/basic/LDAP/basic_ldap_auth.cc
@@ -607,9 +607,9 @@ recover:
                 goto recover;
             }
             if (LDAP_SECURITY_ERROR(e))
-                SEND_ERR(HLP_MSG(ldap_err2string(e)));
+                SEND_ERR(ldap_err2string(e));
             else
-                SEND_BH(HLP_MSG(ldap_err2string(e)));
+                SEND_BH(ldap_err2string(e));
         } else {
             SEND_OK("");
         }
--- a/src/auth/basic/LDAP/basic_ldap_auth.cc
+++ b/src/auth/basic/LDAP/basic_ldap_auth.cc
@@ -23,7 +23,7 @@
  *
  * Usage: squid_ldap_auth -b basedn [-s searchscope]
  *                        [-f searchfilter] [-D binddn -w bindpasswd]
- *                        [-u attr] [-h host] [-p port] [-P] [-R] [ldap_server_name[:port]] ...
+ *                        [-u attr] [-h host] [-p port] [-B] [-P] [-R] [ldap_server_name[:port]] ...
  *
  * Dependencies: You need to get the OpenLDAP libraries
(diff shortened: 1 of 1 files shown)


src/auth/basic/LDAP/basic_ldap_auth.cc
```diff
@@ -23,7 +23,7 @@
  *
  * Usage: squid_ldap_auth -b basedn [-s searchscope]
  *                        [-f searchfilter] [-D binddn -w bindpasswd]
- *                        [-u attr] [-h host] [-p port] [-P] [-R] [ldap_server_name[:port]] ...
+ *                        [-u attr] [-h host] [-p port] [-B] [-P] [-R] [ldap_server_name[:port]] ...
```
yadij: Since Squid-3.x are no longer supported this option can be removed from this merge PR. (People back-porting may want it, but not the official merge).

src/auth/basic/LDAP/basic_ldap_auth.cc
```diff
@@ -602,7 +610,7 @@ main(int argc, char **argv)
                 ld = NULL;
                 goto recover;
             }
-            printf("ERR %s\n", ldap_err2string(squid_ldap_errno(ld)));
+            printf("%s %s\n", (!bherror || LDAP_SECURITY_ERROR(squid_ldap_errno(ld))) ? "ERR" : "BH", ldap_err2string(squid_ldap_errno(ld)));
```
yadij: Please do this as a conversion from printf() statements in this while-loop to the provided SEND_*() macros in helper/protocol_defines.h which are the official helper API interface for Squid-3+.

This particular line change would then look like this:
```suggestion
            const auto e = squid_ldap_errno(ld);
            if (LDAP_SECURITY_ERROR(e)) {
                SEND_ERR(HLP_MSG(ldap_err2string(e)));
            } else {
                SEND_BH(HLP_MSG(ldap_err2string(e)));
            }
```

yadij:
OK to test

yadij:
Sorry, I missed that HLP_MSG can only take static strings. You can omit that macro from the lines taking ldap_err2string() output as text.

amishmm:
Done. Removed HLP_MSG.

rousskov:
I polished PR title/description to obey commit message [rules](https://github.com/measurement-factory/anubis#readme) and simplify. Hopefully, Anubis will notice and update the labels. Please adjust further if needed.

rousskov:
Thank you for fixing formatting -- I accidentally used an incorrect line length setting when doing the first path.

PR #348, opened 2019-01-06
Fix BodyPipe/Sink memory leaks associated with auto-consumption

rousskov:
Auto-consumption happens (and could probably leak memory) in many cases,
but this leak was exposed by an eCAP service that blocked or replaced
virgin messages.

The BodySink job termination algorithm relies on body production
notifications. A BodySink job created after the body production had
ended can never stop and, hence, leaks (leaking the associated BodyPipe
object with it). Such a job is also useless:  If production is over,
there is no need to free space for more body data!  This change avoids
creating such leaking and useless jobs.

### DIFF
--- a/src/BodyPipe.cc
+++ b/src/BodyPipe.cc
@@ -286,7 +286,7 @@ BodyPipe::expectNoConsumption()
         abortedConsumption = true;
 
         // in case somebody enabled auto-consumption before regular one aborted
-        if (mustAutoConsume)
+        if (mustAutoConsume && theProducer.valid())
             startAutoConsumption();
     }
 }
@@ -313,15 +313,16 @@ BodyPipe::consume(size_t size)
     postConsume(size);
 }
 
-// In the AutoConsumption  mode the consumer has gone but the producer continues
-// producing data. We are using a BodySink BodyConsumer which just discards the produced data.
+// In the AutoConsumption mode, the consumer is gone, but the producer continues
+// to produce data. We use a BodySink BodyConsumer to discard that data.
 void
 BodyPipe::enableAutoConsumption()
 {
     mustAutoConsume = true;
     debugs(91,5, HERE << "enabled auto consumption" << status());
-    if (!theConsumer && theBuf.hasContent())
+    if (!theConsumer && theProducer.valid() && theBuf.hasContent())
         startAutoConsumption();
+    // else already consuming, nobody needs consumption, or nothing to consume
 }
 
 // start auto consumption by creating body sink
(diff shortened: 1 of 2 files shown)


src/BodyPipe.cc
```diff
@@ -313,23 +312,28 @@ BodyPipe::consume(size_t size)
...
-BodyPipe::startAutoConsumption()
+BodyPipe::startAutoConsumptionIfNeeded()
 {
-    Must(mustAutoConsume);
-    Must(!theConsumer);
+    const auto startNow =
+        mustAutoConsume && // was enabled
+        !theConsumer && // has not started yet
+        theProducer.valid() && // still useful (and will eventually stop)
+        theBuf.hasContent(); // has something to consume right now
```
yadij: This condition looks suspicious to me. Is it absolutely certain that content needing to be absorbed has *always* got some portion in the buffer when this function is called?

I mean, since all the I/O calls and callback chains are triggered by variable amount of bytes arriving there is a non-zero chance that the particular last-known byte was at the Mime and payload boundary. What happens in that case?

yadij: I am not sure enough of the BodyPipe API to phrase it in specifics, which is why this is only a review comment/question.

What I see is that after this patch calls to enableAutoConsumption() do not always setup an auto-consume theConsumer. Have you checked and confirmed that *all* possible I/O activity on the pipe result in startAutoConsumptionIfNeeded() being used to start the auto-consume of new bytes arriving?
 When mustConsume==true we must never use bytes in that pipe for anything else than discard/consume.

rousskov: > after this patch calls to enableAutoConsumption() do not always setup an auto-consume theConsumer

Yes, both before and after the patch, the start of auto-consumption is conditional on many things. In fact, shrinking the set of cases where auto-consumption starts is what fixed the leak (which is the goal/focus of this PR).

> Have you checked and confirmed that _all_ possible I/O activity on the pipe result in startAutoConsumptionIfNeeded() being used to start the auto-consume of new bytes arriving?

There are problems with your question that make it difficult to answer in a meaningful way. Please let me know if you want me to detail those problems. Meanwhile, I can offer answers to similar but different questions that might address your underlying concerns.

> Have you evaluated each startAutoConsumptionIfNeeded() invocation for correctness? Do you believe each of them does not make things worse? Does at least one of them fix the leak?

The answer to each of the above question is "yes, of course".

> When mustConsume==true we must never use bytes in that pipe for anything else than discard/consume.

I assume you are talking about mustAutoConsume. I am sorry, but mustAutoConsume does not mean "never use bytes in that pipe". mustAutoConsume may be true when some external consumer is still using those bytes. The enableAutoConsumption() method does not assume or require that theConsumer is already nil. IIRC, it may not be nil in some cases.

The mustAutoConsume data member means that BodyPipe must start or continue consuming when there is production without a consumer. The `startNow` conditions, now centralized in startAutoConsumptionIfNeeded(), check that invariant. Very similar (but sometimes insufficiently strict and often implicit) conditions were present in the unpatched code as well.

yadij: Okay, thank you. That makes the situation a bit clearer.

PR #349, opened 2019-01-09
Fix OpenSSL builds that define OPENSSL_NO_ENGINE

neheb:
Even with ENGINE support disabled, OpenSSL provides the openssl/engine.h
header. We have to use OPENSSL_NO_ENGINE to detect ENGINE support.

### DIFF
--- a/src/ssl/support.cc
+++ b/src/ssl/support.cc
@@ -485,7 +485,7 @@ Ssl::Initialize(void)
 
     SQUID_OPENSSL_init_ssl();
 
-#if HAVE_OPENSSL_ENGINE_H
+#if !defined(OPENSSL_NO_ENGINE)
     if (::Config.SSL.ssl_engine) {
         ENGINE_load_builtin_engines();
         ENGINE *e;


rousskov:
OK to test

rousskov:
@neheb, I polished your PR description (and future commit message) to clarify the change scope in the title (for those who do not know that ENGINE is an OpenSSL thing), to add a missing period in the body, and to obey Squid commit message [rules](https://github.com/measurement-factory/anubis#commit-message). Please adjust further as necessary.

src/ssl/support.cc
```diff
@@ -485,7 +485,7 @@ Ssl::Initialize(void)
 
     SQUID_OPENSSL_init_ssl();
 
-#if HAVE_OPENSSL_ENGINE_H
+#ifndef OPENSSL_NO_ENGINE
```
yadij: Please use #if instead. The #ifndef check has problems sometimes, so for consistency and a long-term guarantee of build success we only allow it in header protection checks.

```suggestion
#if !defined(OPENSSL_NO_ENGINE)
```

neheb:
Ah sorry about that. I'm just used to doing so as the other projects that I contribute to (systemd and openwrt) recommend doing so.

rousskov:
No problem at all. We have started with those recommendations as well but have evolved.

PR #354, opened 2019-01-14
Bug 4914: Do not call setsid() in --foreground mode

marcosfrm:
Squid executed in --foreground is always a process group leader. Thus,
setsid(2) is unnecessary and always fails (with EPERM) for such Squids.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1866,6 +1866,7 @@ GoIntoBackground()
         exit(EXIT_SUCCESS);
     }
     // child, running as a background daemon
+    Must(setsid() > 0); // ought to succeed after fork()
 }
 
 static void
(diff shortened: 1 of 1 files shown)


rousskov:
OK to test

rousskov:
@marcosfrm, please do not resolve change requests initiated by others. Only the reviewer can authoritatively tell whether their request is fully addressed and needs no further work, even when the author applies changes suggested by the reviewer.

Clicking the "resolve conversation" button does not trigger GitHub notifications, so the reviewer remains in the dark that their request may have been addressed.

When/if a reviewer finds that a change was marked as resolved by another person, they realize that they cannot rely on any past resolution marks reported by GitHub, including change requests they resolved _themselves_. And when a PR has many change requests, a single request resolved by another person can easily double or even triple reviewer work!

GitHub lacks a good reviewing interface. Reviewers and authors suffer from this. The best way for authors to mark addressed change requests (that I know about) is to add a "Done at [SHA]" or similar comment to the change request thread.

PR #355, opened 2019-01-15
Bug 4915: Detect IPv6 loopback binding errors

yadij:
Systems which have been partially 'IPv6 disabled' may allow
sockets to be opened and used but missing the IPv6 loopback
address.

Implement the outstanding TODO to detect such failures and
disable IPv6 support properly within Squid when they are found.

This should fix bug 4915 auth_param helper startup and similar
external_acl_type helper issues. For security such helpers are
not permitted to use the machine default IP address which is
globally accessible.

### DIFF
--- a/src/ip/tools.cc
+++ b/src/ip/tools.cc
@@ -60,7 +60,7 @@ Ip::ProbeTransport()
     // Test for IPv6 loopback/localhost address binding
     Ip::Address ip;
     ip.setLocalhost();
-    if (ip.isIPv6()) { // false on --disable-ipv6 builds
+    if (ip.isIPv6()) { // paranoid; always succeeds if we got this far
         struct sockaddr_in6 sin;
         ip.getSockAddr(sin);
         if (bind(s, reinterpret_cast<struct sockaddr *>(&sin), sizeof(sin)) != 0) {
--- a/src/ip/tools.cc
+++ b/src/ip/tools.cc
@@ -10,6 +10,7 @@
 
 #include "squid.h"
 #include "Debug.h"
+#include "ip/Address.h"
 #include "ip/tools.h"
 
 #if HAVE_UNISTD_H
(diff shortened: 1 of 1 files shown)


src/ip/tools.cc
```diff
@@ -55,8 +56,21 @@ Ip::ProbeTransport()
...
-    //  needs to be safe to contact and with guaranteed working IPv6 at the other end.
+
+    // Test for IPv6 loopback/localhost address binding
+    Ip::Address ip;
+    ip.setLocalhost();
+    if (ip.isIPv6()) { // false on --disable-ipv6 builds
+        struct sockaddr_in6 sin;
+        ip.getSockAddr(sin);
+        if (bind(s, reinterpret_cast<struct sockaddr *>(&sin), sizeof(sin)) != 0) {
+            debugs(3, DBG_CRITICAL, "WARNING: BCP 177 violation. Detected non-functional IPv6 loopback.");
```
rousskov: Since "non-functional IPv6 loopback" does not necessarily break anything, perhaps this should be `DBG_IMPORTANT`? I am just asking and not even suggesting a change here. Your call.

src/ip/tools.cc
```diff
@@ -55,8 +56,21 @@ Ip::ProbeTransport()
     debugs(3, 2, "Missing RFC 3493 compliance - attempting split IPv4 and IPv6 stacks ...");
     EnableIpv6 |= IPV6_SPECIAL_SPLITSTACK;
 #endif
-    // TODO: attempt to use the socket to connect somewhere ?
-    //  needs to be safe to contact and with guaranteed working IPv6 at the other end.
+
+    // Test for IPv6 loopback/localhost address binding
+    Ip::Address ip;
+    ip.setLocalhost();
+    if (ip.isIPv6()) { // false on --disable-ipv6 builds
```
rousskov: AFAICT, this check will always succeed because EnableIpv6 is always true here, and ip.setLocalhost() sets an IPv6 address when EnableIpv6 is true. Am I missing something?

The `false on --disable-ipv6 builds` comment implies that `USE_IPV6` has some effect on setLocalhos() and/or  isIPv6(), but I do not see that effect. Is it hidden somewhere deeper than the code of those two methods?

src/ip/tools.cc
yadij: I picked critical because this system is claiming IPv6 to be enabled (socket succeeded) but lacking any NIC registered as loopback. There are quite a number of connections which use loopback/localhost for security purposes. The IPv4 localhost is less secure than IPv6 loopback in VM and container environments. As shown in the bug report though it might be intentional so the message is a warning and reversion to IPv4 anyway.

src/ip/tools.cc
yadij: Just me being extra paranoid. Since admin removing ::1 allocation is an edge case already I do not want this test to suddenly start disabling IPv6 if we or someones patch does anything in future that changes that situation (like looking up a local definition of localhost or patching setLocalhost() to an IPv4 address).
 With this the test will skip unless it is working with at least a valid IPv6 address. Even if it is not strictly ::1.

src/ip/tools.cc
rousskov: FWIW, I do not see the connection between some systems using loopback for security purposes and Squid claiming that not using loopback is a critical problem (presumably for Squid). Clearly, Squid could check lots of BCPs, but (thankfully) does not, because Squid is a proxy, not a BCP compliance validator.

Said that, I am resolving this minor comment. It is your call.

src/ip/tools.cc
rousskov: If I interpret your answer correctly, then I was not missing any hidden dependencies, and the source code comment is indeed lying -- isIPv6() will _not_ return `false on --disable-ipv6 builds`. Please replace that comment. I suggest:

```suggestion
    if (ip.isIPv6()) { // paranoid; always succeeds if we got this far
```

rousskov: @yadij, please do not resolve change requests made by others.

PR #359, opened 2019-02-05
Add support for buffer-size= to UDP logging

mahdi1001:
Allow admin control of buffering for log outputs written to UDP
receivers using the buffer-size= parameter.

buffer-size=0byte disables buffering and sends UDP packets
immediately regardless of line size.

When non-0 values are used lines shorter than the buffer may be
delayed and aggregated into a later UDP packet.

Log lines larger than the buffer size will be sent immediately
and may trigger delivery of previously buffered content to
retain log order (at time of send, not UDP arrival).

To avoid truncation problems known with common recipients
the buffer size remains capped at 1400 bytes.

### DIFF
--- a/src/log/ModUdp.cc
+++ b/src/log/ModUdp.cc
@@ -166,7 +166,7 @@ logfile_mod_udp_open(Logfile * lf, const char *path, size_t bufsz, int fatal_fla
         path += 2;
     }
     strAddr = xstrdup(path);
-    char *flush_time = (char *) strchr(strAddr, ','); 
+    char *flush_time = (char *) strchr(strAddr, ',');
     if (flush_time) {
         *flush_time = '\0';
         ++flush_time;
--- a/src/log/ModUdp.cc
+++ b/src/log/ModUdp.cc
@@ -148,14 +148,6 @@ logfile_mod_udp_open(Logfile * lf, const char *path, size_t bufsz, int fatal_fla
         path += 2;
     }
     strAddr = xstrdup(path);
-    char *flush_time = (char *) strchr(strAddr, ',');
-    if (flush_time) {
-        *flush_time = '\0';
-        ++flush_time;
-        int flushperiod = atoi(flush_time);
-        /* Start the flush event */
-        eventAdd("logfile_mod_udp_flush", logfileUdpFlushEvent, lf, flushperiod, 1);
-    }
     if (!GetHostWithPort(strAddr, &addr)) {
         if (lf->flags.fatal) {
             fatalf("Invalid UDP logging address '%s'\n", lf->path);
--- a/src/log/ModUdp.cc
+++ b/src/log/ModUdp.cc
@@ -109,22 +109,22 @@ logfile_mod_udp_rotate(Logfile *, const int16_t)
 {
 }
 
-/*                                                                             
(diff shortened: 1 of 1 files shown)


mahdi1001:
Hi Yadij 
Squid don't send log to udp if packet len Less than 1400 byte. Never send!!!
This is a bug in udp module and best practice for this bug use event for flush periodical.
Please merge this code to master.
Thank you for your attention to this matter

yadij:
Best practice is actually to fix the bug rather than force small packets out to broken receivers.

The code in logfile_mod_udp_open() which forces buffered logs to be on for UDP loggers has a note saying that it is doing so to force the UDP packets to stay within MTU. see the end of logfile_mod_udp_open(). The logic there looks broken, fixing it to obey the buffer-size= option and only use 1400 as a default would be better.

Despite that, the short log entries should currently be accumulated into a buffer and all sent together when the accumulated length reaches 'bufsz' or more bytes. see logfile_mod_udp_writeline second if statement which performs the flush.

yadij:
Please try a build of Squid without the flush events and just changing the lines at the end of logfile_mod_udp_open() to do this:
`
if (bufsz > 1400)
    bufsz = 1400;
`
instead of just:
`bufsz = 1400;`

That change alone should let you use the buffer-size=N option on your access_log line to set the desired UDP packet size without any extra performance cost. Where 0 means no buffering at all.
eg.
> access_log udp://192.0.2.1:1234 buffer-size=0 squid

mahdi1001:
OK I change it.

yadij:
Just to confirm, did you test this works for you?

mahdi1001:
Yes. This option "buffer-size=0bytes logformat=squid " is correct.
Can I change buffer-size=0bytes to  buffer-size=0B in cash_cf.cc?
Let me change and refactor this code?

yadij:
OK to test

eduard-bagdasaryan:
Jenkins [failed](https://build.squid-cache.org/job/5-pr-auto/compiler=gcc,label=d-ubuntu-cosmic/255/console) the build of the staged 07d4ea3 . Though it is unclear why it has not updated the build status to red.

yadij:
The status seems to have been set for (https://build.squid-cache.org/job/5-pr-test/1021/) which was a success.

yadij:
Kicking the 5-pr-auto test again failed and still not updated this PR status. I have altered the Jenkins config to see if another node will work better.

eduard-bagdasaryan:
> The status seems to have been set for (https://build.squid-cache.org/job/5-pr-test/1021/) which was a success.

No, that is a different status. The "5-pr-test" Jenkins job is responsible for PR branch testing, while "5-pr-auto"  tests staged commits (which finally go to master). In our case the problematic job is "5-pr-auto".

> Kicking the 5-pr-auto test again failed and still not updated this PR status.

Unfortunately, just rebuilding was no use because the problem was different. The error message " Cannot retrieve Git metadata for the build" relates to an open Jenkins [bug](https://issues.jenkins-ci.org/browse/JENKINS-54249?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel). I applied the workaround described there: rollback three Git-related plugins back to their previous versions.

mahdi1001:
If possible merge to v4 branch, please

PR #363, opened 2019-02-10
Unit empty string == bytes

mahdi1001:
Add empty string check for byte unit.

### DIFF
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -1306,7 +1306,8 @@ parseBytesOptionValue(size_t * bptr, const char *units, char const * value)
 static size_t
 parseBytesUnits(const char *unit)
 {
-    if (!strncasecmp(unit, B_BYTES_STR, strlen(B_BYTES_STR)))
+    if (!strncasecmp(unit, B_BYTES_STR, strlen(B_BYTES_STR)) ||
+            !strncasecmp(unit, "", strlen(unit)))
         return 1;
 
     if (!strncasecmp(unit, B_KBYTES_STR, strlen(B_KBYTES_STR)))


yadij:
parseBytesUnits() should only ever be called with a unit string. So this looks like a wrong fix to the recently uncovered problems with config parsing of units.

Since we have a mixed situation of config setting having optional vs mandatory vs no units, and differing placement regards to whitespaces the function calling parseBytesUnits() should detect where the units come from.

IMHO, we should probably add generic support for bytes parameters using the string "none" as 0. This will need some planning discussion in squid-dev mailing list though to get buy-in from the other devs since "none" is already special cased in many directives with value -1.

rousskov:
> we should probably add generic support for bytes parameters using the string "none" as 0

FWIW, I do not understand what `generic support for bytes parameters using the string "none" as 0` means. The only interpretation I can imagine here is that admins would be able to write `foo=none` instead of `foo=0` (or equivalent) but that does not really make sense -- `none` is clearly worse than `0` in this context -- so you probably meant something else.

mahdi1001:
For example :
Buffer-size=256  is invalid. You should entered "Buffer-size=256byte" completly.
I just fixed this problem.

rousskov:
> Buffer-size=256 is invalid.

And, IMO, should stay invalid. Squid may need better error reporting, but it is best to require explicit units for non-zero size values where possible. It is not difficult to type "bytes", and there have been costly configuration bugs where admins assumed a wrong default size unit for some of the configuration options.

yadij:
By the "none" support I mean adjusting the parse_access_log function to special case "buffer-size=none" to set the cl->bufferSize=0 akin to turning "buffered_logs off" for just this one output. Having a bufferSize of 0 is already special-cased in the UDP writeline() logic to skip buffer use.

rousskov:
IMO `none` is clearly worse than `0` in this context -- I see no advantages in adding support for `none` sizes where we can support `0` sizes.

PR #364, opened 2019-02-12
Bug 4919: master commit b599471 leaks memory

rousskov:
Restored the natural order of the following two notifications:
* BodyConsumer::noteMoreBodyDataAvailable() and
* BodyConsumer::noteBodyProductionEnded() or noteBodyProducerAborted().

Commit b599471 unintentionally reordered those two notifications. Client
kids (and possibly other BodyConsumers) relied on the natural order to
end their work. If an HttpStateData job was done with the Squid-to-peer
connection and only waiting for the last adapted body bytes, it would
get stuck and leak many objects. This use case was not tested during
b599471 work.

### DIFF
--- a/src/BodyPipe.cc
+++ b/src/BodyPipe.cc
@@ -397,12 +397,14 @@ BodyPipe::postAppend(size_t size)
     thePutSize += size;
     debugs(91,7, HERE << "added " << size << " bytes" << status());
 
-    if (!mayNeedMoreData())
-        clearProducer(true); // reached end-of-body
-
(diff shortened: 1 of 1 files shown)


rousskov:
The branch commit https://github.com/squid-cache/squid/pull/364/commits/8bbdb59a9c92edc6873b56b14dd14e787455ab4b message details the bug mechanism, but those details felt to be too low-level to be included in the PR description (and the official commit). Feel free to copy them back if you think they should be included.

PR #368, opened 2019-02-17
Bug 4919: master commit b599471 leaks memory (#364)

squidadm:
Restored the natural order of the following two notifications:
* BodyConsumer::noteMoreBodyDataAvailable() and
* BodyConsumer::noteBodyProductionEnded() or noteBodyProducerAborted().

v4 commit dd9d539 (ported from master commit b599471) unintentionally
reordered those two notifications. Client kids (and possibly other
BodyConsumers) relied on the natural order to end their work. If an
HttpStateData job was done with the Squid-to-peer connection and only
waiting for the last adapted body bytes, it would get stuck and leak
many objects. This use case was not tested during b599471 work.


yadij:
OK to test

PR #371, opened 2019-02-21
Made Coverity happier after Bug 4864 fix

rousskov:
... by addressing the following false-positive:
CID 1442832: Null pointer dereferences (FORWARD_NULL)

The Comm::IsConnOpen() test guarantees that connManager is not nil. An
explicit assert() might reduce bugs when this code is refactored.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -971,6 +971,7 @@ FwdState::usePinned()
     ++n_tries;
     request->flags.pinned = true;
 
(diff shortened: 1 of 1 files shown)


rousskov:
```
CID 1442832:  Null pointer dereferences  (FORWARD_NULL)
    Passing null pointer "connManager" to "pinnedAuth", which dereferences it.
```

PR #373, opened 2019-02-22
Fixed squidclient authentication after 4b19fa9 (Bug 4843 pt2)

rousskov:
* squidclient -U sent Proxy-Authorization instead of Authorization.
  Code duplication bites again.

* squidclient -U and -u could sent random garbage after the correct
  [Proxy-]Authorization value as exposed by Coverity CID 1441999: Unused
  value (UNUSED_VALUE). Coverity missed this deeper problem, but
  analyzing its report lead to discovery of the two bugs fixed here.

Also reduced authentication-related code duplication.

### DIFF
--- a/tools/squidclient/squidclient.cc
+++ b/tools/squidclient/squidclient.cc
@@ -183,7 +183,7 @@ class Authorization
 {
 public:
     Authorization(const char *aHeader, const char *aDestination):
-        header(aHeader), destination(aDestination), user(nullptr), password(nullptr) {}
+        header(aHeader), destination(aDestination) {}
 
     /// finalizes and writes the right HTTP header to the given stream
     void commit(std::ostream &os);
--- a/tools/squidclient/squidclient.cc
+++ b/tools/squidclient/squidclient.cc
@@ -188,7 +188,7 @@ public:
     /// finalizes and writes the right HTTP header to the given stream
     void commit(std::ostream &os);
 
-    std::string header = nullptr; ///< HTTP header name to send
+    std::string header; ///< HTTP header name to send
     std::string destination = nullptr; ///< used when describing password
     const char *user = nullptr; ///< user name to encode and send
     const char *password = nullptr; ///< user password to encode and send
--- a/tools/squidclient/squidclient.cc
+++ b/tools/squidclient/squidclient.cc
@@ -22,6 +22,7 @@ using namespace Squid;
 /** \endcond */
 #endif
 
+#include <cassert>
 #include <cerrno>
 #include <csignal>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


rousskov:
Quoting the (largely irrelevant) Coverity problem report here for the record:

CID 1441999: Unused value (UNUSED_VALUE) assigned_value: Assigning value to blen here, but that stored value is overwritten before it can be used.
```
blen += base64_encode_final(&ctx, pwdBuf+blen);
```

tools/squidclient/squidclient.cc
```diff
@@ -177,6 +178,56 @@ shellUnescape(char *buf)
...
+class Authorization
+{
+public:
+    Authorization(const char *aHeader, const char *aDestination):
+        header(aHeader), destination(aDestination), user(nullptr), password(nullptr) {}
+
+    /// finalizes and writes the right HTTP header to the given stream
+    void commit(std::ostream &os);
+
+    std::string header = nullptr; ///< HTTP header name to send
```
yadij: std::string are complex objects, not raw-pointers. So they do not need explicit initialization.
```suggestion
    std::string header; ///< HTTP header name to send
```

tools/squidclient/squidclient.cc
```diff
@@ -177,6 +178,56 @@ shellUnescape(char *buf)
...
+    std::string destination = nullptr; ///< used when describing password
```
yadij:
```suggestion
    std::string destination; ///< used when describing password
```

tools/squidclient/squidclient.cc
```diff
@@ -177,6 +178,56 @@ shellUnescape(char *buf)
     *d = '\0';
 }
 
+/// [Proxy-]Authorization header producer
+class Authorization
+{
+public:
+    Authorization(const char *aHeader, const char *aDestination):
+        header(aHeader), destination(aDestination), user(nullptr), password(nullptr) {}
```
yadij: user and password are C++11 initialized. They should not need to be repeat-initialized here.

tools/squidclient/squidclient.cc
```diff
@@ -177,6 +178,56 @@ shellUnescape(char *buf)
     *d = '\0';
 }
 
+/// [Proxy-]Authorization header producer
+class Authorization
```
yadij: Per our style guide this class should be in .h/.cc files of its own. Not a blocker, but it would be nice not to have to come back and shuffle that around later.

tools/squidclient/squidclient.cc
```diff
@@ -177,6 +178,56 @@ shellUnescape(char *buf)
...
+    base64_encode_init(&ctx);
+    const auto bcapacity = base64_encode_len(strlen(user) + 1 + strlen(password));
+    const auto buf = new char[bcapacity];
+
+    size_t bsize = 0;
+    bsize += base64_encode_update(&ctx, buf, strlen(user), reinterpret_cast<const uint8_t*>(user));
+    bsize += base64_encode_update(&ctx, buf+bsize, 1, reinterpret_cast<const uint8_t*>(":"));
+    bsize += base64_encode_update(&ctx, buf+bsize, strlen(password), reinterpret_cast<const uint8_t*>(password));
+    bsize += base64_encode_final(&ctx, buf+bsize);
+    assert(bsize <= bcapacity); // paranoid and late but better than nothing
```
yadij: The Base64 code does more timely detection and assert for this overflow case. If we cannot do the check in advance (and produce an error rather than assert) there is no point in having this line.

tools/squidclient/squidclient.cc
rousskov: This helper class is not meant to be reusable elsewhere (without lots of additional work). No (valid) guidelines can require its isolation.

tools/squidclient/squidclient.cc
rousskov: > The Base64 code does more timely detection and assert for this overflow case

AFAICT, it does not, and it cannot -- the old and ugly base64_encode*() API does not know where the buffer ends. Fixing that is outside this PR scope IMO.

> there is no point in having this line.

This is just a paranoid assert() that might detect unlikely [future] bugs, but nothing more. If you insist on removing it for some reason, I will.

tools/squidclient/squidclient.cc
rousskov: Applied.

tools/squidclient/squidclient.cc
rousskov: Applied.

tools/squidclient/squidclient.cc
rousskov: Fixed.

rousskov:
@yadij, AFAICT, all of your change requests have been addressed except these two:
1. https://github.com/squid-cache/squid/pull/373#discussion_r259221731
1. https://github.com/squid-cache/squid/pull/373#discussion_r259222431

tools/squidclient/squidclient.cc
yadij: The buffer is allocated from the base64_encode_len() output.  The base64_encode_* functions each assert that their individual chunk of input is encoded to an output strictly fitting within the number of bytes base64_encode_len() would estimate for that input. This detects both buffer overflows and encoding corruption (ignored by this 'paranoid' assert).

The only bugs we could add here are by violating that guarantee trying to "fix" the API usage.

yadij:
Checked, and approved despite some non-blocker disagreements.

tools/squidclient/squidclient.cc
rousskov: I stand by my statement of fact but am not going to waste time discussing base64 APIs and this bug-fixing PR code any further.

PR #375, opened 2019-02-22
Fixed build after a56fcf0 (Reuse helpers after timeout)

rousskov:
In all tested environments, something else #included `<map>`,
apparently, but that old explicit #include should not have been removed.

Also fixed STL #include order.


rousskov:
> what is exactly broken?

The code used `std::map` without including `<map>`.

yadij:
Oh I see, not a dependency of each other.

PR #377, opened 2019-02-23
Add support for buffer-size= to UDP logging #359

mahdi1001:
Allow admin control of buffering for log outputs written to UDP
receivers using the buffer-size= parameter.

buffer-size=0byte disables buffering and sends UDP packets
immediately regardless of line size.

When non-0 values are used lines shorter than the buffer may be
delayed and aggregated into a later UDP packet.

Log lines larger than the buffer size will be sent immediately
and may trigger delivery of previously buffered content to
retain log order (at time of send, not UDP arrival).

To avoid truncation problems known with common recipients
the buffer size remains capped at 1400 bytes.


rousskov:
OK to test

PR #378, opened 2019-02-27
protection against buggy 304 Not Modified replies

bertranddemiddelaer:
When Squid tries to refresh its cache after its local cached content is expired, some buggy upstream servers reply "304 Not Modified" status with a "Content-Length" header set to 0.

After that, Squids updates its response headers to the client with the all the ones he got from the upstream server but "Warning".

Squid should protect itself from those buggy inputs, otherwise it will send back an inconsistent reply to the client (Content-Length set to 0 followed by the previously cached body).

### DIFF
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -273,14 +273,20 @@ HttpHeader::updateWarnings()
 }
 
 bool
-HttpHeader::skipUpdateHeader(const Http::HdrType id) const
+HttpHeader::skipUpdateHeader(const Http::HdrType id, bool update_content_length) const
 {
     // RFC 7234, section 4.3.4: use fields other from Warning for update
-    return id == Http::HdrType::WARNING;
+    if (id == Http::HdrType::WARNING)
+        return true;
+
+    if (update_content_length == false && id == Http::HdrType::CONTENT_LENGTH)
+        return true;
+
+    return false;
 }
 
 bool
-HttpHeader::update(HttpHeader const *fresh)
+HttpHeader::update(HttpHeader const *fresh, bool update_content_length)
 {
     assert(fresh);
     assert(this != fresh);
@@ -298,7 +304,7 @@ HttpHeader::update(HttpHeader const *fresh)
     while ((e = fresh->getEntry(&pos))) {
         /* deny bad guys (ok to check for Http::HdrType::OTHER) here */
 
-        if (skipUpdateHeader(e->id))
+        if (skipUpdateHeader(e->id, update_content_length))
             continue;
 
         if (e->id != Http::HdrType::OTHER)
@@ -311,7 +317,7 @@ HttpHeader::update(HttpHeader const *fresh)
     while ((e = fresh->getEntry(&pos))) {
         /* deny bad guys (ok to check for Http::HdrType::OTHER) here */
 
-        if (skipUpdateHeader(e->id))
+        if (skipUpdateHeader(e->id, update_content_length))
             continue;
 
         debugs(55, 7, "Updating header '" << Http::HeaderLookupTable.lookup(e->id).name << "' in cached entry");
--- a/src/HttpHeader.h
+++ b/src/HttpHeader.h
@@ -82,7 +82,7 @@ public:
     /* Interface functions */
     void clean();
     void append(const HttpHeader * src);
-    bool update(HttpHeader const *fresh);
+    bool update(HttpHeader const *fresh, bool update_content_length=true);
     void compact();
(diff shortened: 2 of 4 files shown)


yadij:
This change would directly violate a MUST level requirement in the RFC section the code your are editing referenced:
" 
 If a stored response is selected for update, the cache MUST:
...
  *  use other header fields provided in the 304 (Not Modified)
      response to replace all instances of the corresponding header
      fields in the stored response.
"

That includes the Content-Length header. Server 304 responses such as described are valid. Their purpose is to explicitly inform the proxy that the payload content has now been reduced to 0 bytes of length by the origin (such as at the end of a cacheable but temporary "live stream" media). The proxy is expected to discard the payload data and return only the updated response headers to clients until future revalidation update.

bertranddemiddelaer:
> This change would directly violate a MUST level requirement in that RFC section the code you are editing referenced:
> "
> If a stored response is selected for update, the cache MUST:
> ...
> 
> * use other header fields provided in the 304 (Not Modified)
>   response to replace all instances of the corresponding header
>   fields in the stored response.
>   "
> 
> That includes the Content-Length header. Server 304 responses such as described are valid. Their purpose is to explicitly inform the proxy that the payload content has now been reduced to 0 bytes of length by the origin (such as at the end of a cacheable but temporary "live stream" media). The proxy is expected to discard the payload data and return only the updated response headers to clients until future revalidation update.



> This change would directly violate a MUST level requirement in that RFC section the code you are editing referenced:
> "
> If a stored response is selected for update, the cache MUST:
> ...
> 
> * use other header fields provided in the 304 (Not Modified)
>   response to replace all instances of the corresponding header
>   fields in the stored response.
>   "
> 
> That includes the Content-Length header. Server 304 responses such as described are valid. Their purpose is to explicitly inform the proxy that the payload content has now been reduced to 0 bytes of length by the origin (such as at the end of a cacheable but temporary "live stream" media). The proxy is expected to discard the payload data and return only the updated response headers to clients until future revalidation update.

I think that if the status of the answer is 304 "Not Modified", this is not consistent to grab a different content length than the one previously cached.

I understand your reply, but right now squid sends back to the http client one answer containing "Content-Length: 0" and the content previously cached, which is not consistent at all. I understand that we should follow the RFC by the book, but how can we handle an upstream server that does not do it correctly ? IMHO Squid should protect itself from buggy replies. I mean, my only way to solve this situation is to modify squid, as I can not fix the root cause.

bertranddemiddelaer:
For information, this behavior has been noticed on the following upstream server: https://kubernetes-charts.storage.googleapis.com/

yadij:
As I mentioned the proxy should technically not be sending the payload if it is relaying on the Content-Length:0 header.

The Squid bug here is the continued existence of the payload bytes, not the use of server provided C-L value. That is already being tracked by https://bugs.squid-cache.org/show_bug.cgi?id=4882. In there I mention that the proxy does have the option of retrying with a full fetch if we conclude the server 304 was broken or in any other way inconsistent with the object it tried to revalidate. If you want to take on implementing that please do. Design details should be discussed in squid-dev first.

Please notice that URL you provide produces objects which are explicitly marked as *not cacheable*. It is to be expected that odd things break when caching is forced. The downloadable contents within that domain which are marked as cacheable do produce working 304 responses.

xeron:
I think there's also some related information in https://bugs.squid-cache.org/show_bug.cgi?id=4606.

We've had squid crashes under load when some endpoint was returning `304` with `C-L: 0`, but it was v3.5.x, so I'm not sure it still applies.

hno:
Content-length is a bit different than other headers being both an entity
and a transport header.

A 304 with a different content-length from the cached entity is either
a) Broken (most likely). Especially true if the request had a strong
conditional. And a 304 with content-lenght 0 is very much in this category.
b) If the request only had If-Modified-Since and no strong conditional then
the server clock is doing bad things.

A server that replaces the entity with a 0 length entity creates a new
entity for the URL, and should have at least a new ETag (MUST NOT be the
same) or later Last-Modified compared to the previous version.

We have three choices

a) Accept the 304 as a not-modified for the cached entity, update the
cached entity headers, except for content-length.
b) Discard the cached entity.
c) Discard the 304 and NOT update the cached entity. The 304 may still be
served to the client if the conditional came from the client. If the
conditional did not come from the client then we have to retry the request
here without adding conditionals verifying our cached entity.

Plus notifying the server vendor that their server is odd.

Whateer we do ee MUST NOT update the cached entity with the new
content-length.

Regards
Henrik

PR #379, opened 2019-02-28
Bug 4928: Cannot convert non-IPv4 to IPv4

yadij:
… when reaching client_ip_max_connections

The client_ip_max_connections limit is checked before the TCP
dst-IP is located for the newly received TCP connection. This
leaves Squid unable to fetch the NFMARK or similar details later
on (they do not exist for [::]).

Move client_ip_max_connections test later in the TCP accept
process to ensure dst-IP is known when the error is produced.

### DIFF
--- a/src/comm/TcpAcceptor.cc
+++ b/src/comm/TcpAcceptor.cc
@@ -382,15 +382,6 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
     details->fd = sock;
     details->remote = *gai;
 
-    if ( Config.client_ip_max_connections >= 0) {
-        if (clientdbEstablished(details->remote, 0) > Config.client_ip_max_connections) {
-            debugs(50, DBG_IMPORTANT, "WARNING: " << details->remote << " attempting more than " << Config.client_ip_max_connections << " connections.");
-            Ip::Address::FreeAddr(gai);
-            PROF_stop(comm_accept);
-            return Comm::COMM_ERROR;
-        }
-    }
-
     // lookup the local-end details of this new connection
     Ip::Address::InitAddr(gai);
     details->local.setEmpty();
@@ -403,6 +394,14 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
     }
     details->local = *gai;
     Ip::Address::FreeAddr(gai);
+
+    if ( Config.client_ip_max_connections >= 0) {
+        if (clientdbEstablished(details->remote, 0) > Config.client_ip_max_connections) {
+            debugs(50, DBG_IMPORTANT, "WARNING: " << details->remote << " attempting more than " << Config.client_ip_max_connections << " connections.");
+            PROF_stop(comm_accept);
+            return Comm::COMM_ERROR;
+        }
+    }
 
     /* fdstat update */
     // XXX : these are not all HTTP requests. use a note about type and ip:port details->
--- a/src/comm/TcpAcceptor.cc
+++ b/src/comm/TcpAcceptor.cc
@@ -282,16 +282,7 @@ Comm::TcpAcceptor::acceptOne()
     ConnectionPointer newConnDetails = new Connection();
     const Comm::Flag flag = oldAccept(newConnDetails);
 
-    /* Check for errors */
-    if (!newConnDetails->isOpen()) {
-
-        if (flag == Comm::NOMESSAGE) {
-            /* register interest again */
-            debugs(5, 5, HERE << "try later: " << conn << " handler Subscription: " << theCallSub);
-            SetSelect(conn->fd, COMM_SELECT_READ, doAccept, this, 0);
-            return;
-        }
-
+    if (flag == Comm::COMM_ERROR) {
         // A non-recoverable error; notify the caller */
         debugs(5, 5, HERE << "non-recoverable error:" << status() << " handler Subscription: " << theCallSub);
         if (intendedForUserConnections())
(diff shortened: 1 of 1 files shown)


eduard-bagdasaryan:
> Addressing this problem will also fix the probably buggy placement of intendedForUserConnections() check: That check and associated logging should happen when client_ip_max_connections check fails.

IIRC intendedForUserConnections() helps to distinguish two TcpAcceptor use cases: clients connections (we want to log them) and server connections, e.g., FTP server connections (we do not want to log them). Why should we limit this logging only when client_ip_max_connections is exceeded? Currently, it logs whenever COMM_ERROR is observed (which is correct behavior IMO).

rousskov:
> Why should we limit this logging only when client_ip_max_connections is exceeded?

We should not. I do not think such limiting has been suggested.

> Currently, it logs whenever COMM_ERROR is observed 

Unfortunately, that is not what is happening AFAICT: We do _not_ logAcceptError() when newConnDetails isOpen(), despite `COMM_ERROR`. In other words, the existing intendedForUserConnections() check does not cover `COMM_ERROR` cases where newConnDetails isOpen() [probably because the surrounding code misleads the developer into thinking that the isOpen() `if` statement is where all oldAccept() errors are handled].

qb-syn:
Hey, thanks for the first patch and the following diagsnostic.

I've applied the and tested out succesfully, the server does not crash anymore when reaching 
client_ip_max_connections.
Other, it lead to another bug (from `cache.log`) which I'm not sure is already covered in the above discussion:

```
squid[6033]: WARNING: 10.0.3.1:45810 attempting more than 2 connections.
squid[6033]: BUG 3556: FD 1023 is not an open socket.
```

Also, looking at the process FD its seems that the socket is in CLOSE-WAIT:

```
$ lsof -p 6033 | grep -cF 'IPv4'
1007
```
```
$ ss -peanut | grep -F 'fd=999'
tcp    CLOSE-WAIT  117     0             10.0.3.4:3128          10.0.3.1:45760   users:(("squid",pid=6033,fd=999)) uid:13 ino:757643 sk:158 -->
```

yadij:
@qb-syn after addressing Alex requests it looks to me like there is a fourth bug. The COMM_ERROR flag being passed when the limit was encountered would result in the listening port being closed and possibly the worker shutdown. Can you please try the updated patch to see if this other issue disappears now?

src/comm/TcpAcceptor.cc
```diff
@@ -440,6 +408,34 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
     }
 #endif
 
+    details->nfConnmark = Ip::Qos::getNfConnmark(details, Ip::Qos::dirAccepted);
+
+    if (Config.client_ip_max_connections >= 0) {
+        if (clientdbEstablished(details->remote, 0) > Config.client_ip_max_connections) {
```
rousskov: This call will create a client_db record even though we do not plan to accept this connection into Squid. The new zero-counter record will eventually be garbage-collected, but its creation just wastes resources when they are already tight. And when then limit is not reached, we are calling clientdbEstablished() at least twice for the same connection, which is confusing -- the connection is established/accepted only once.

I suggest either adjusting clientdbEstablished() so that it does not create a client_db record when delta is zero OR, better, adding a read-only clientdbCount() function with similar functionality that should be used _instead_ of the current hack of calling clientdbEstablished() with zero delta. The supposed-to-be-read-only ACL code will benefit from this adjustment as well.

This suggested improvement is outside this PR scope, and, naturally, I do not insist on it. On the other hand, if implementing it in this PR is easier/cheaper, I certainly would not object to enlarging this PR scope a bit. Your call.

yadij: That one I think should be in a followup PR.

rousskov: Sounds good to me.

qb-syn:
Hey, thanks for the patch.

I did run some tests from yadij:bug4928 and the original bug seems gone for good now, thanks.

However I still have other error in the log file of the daemon:

```
squid[25031]: WARNING: 10.0.3.4:57124 attempting more than 2 connections.
squid[25031]: BUG 3556: FD 16 is not an open socket.
squid[25031]: WARNING: 10.0.3.4:57120 attempting more than 2 connections.
squid[25031]: BUG 3556: FD 17 is not an open socket.
```

But I'm not sure if this was supposed to be addressed by the fixes.

Tell me if you need more data

rousskov:
@eduard-bagdasaryan, please check the git "rejected tag" [failure](https://semaphoreci.com/squid/squid/branches/auto/builds/711) on Semaphore CI. Restarting the test does [not](https://semaphoreci.com/squid/squid/branches/auto/builds/712) help. Is this our fault? Is there a workaround?

eduard-bagdasaryan:
> Is this our fault?
We started experience these bugs after recent SemaphoreCI [update](https://semaphoreci.com/renderedtext/semaphore/changelog/changes/154) (on March 5th),  similar problems are in PR330 and PR357.
SemaphoreCI reports git errors:

```
! [rejected]            M-staged-PR330 -&gt; M-staged-PR330  (would clobber existing tag)
! [rejected]            M-staged-PR357 -&gt; M-staged-PR357  (would clobber existing tag)
```

I found an [explanation](https://issues.jenkins-ci.org/browse/JENKINS-55284?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel) of this new Git behavior, changed  since Git 2.20. In short, 'git fetch' now fails by default, if it sees that an old tag points now to different revision. Re-assigning the same tag for the staged commit (after a failure) is what Anubis currently does.

>  Is there a workaround?

I could not find in SemaphoreCI log the Git command causing the error. Anyway, I think we should configure it to get the fresh copy to resolve this problem. That is how Jenkins is configured.

rousskov:
@eduard-bagdasaryan, thank you for useful research! I moved the merging-related discussion to https://github.com/measurement-factory/anubis/issues/29

rousskov:
I temporary made Semaphore CI tests optional in hope to work around this problem. I will merge one ready PR manually in hope that Anubis will resume merging other ready PRs after that.

rousskov:
> I will merge one ready PR manually in hope that Anubis will resume merging other ready PRs after that.

I will not -- there are no M-cleared-for-merge PRs right now. @yadij, if you believe this (or any other PR in a similar all-green-except-Semaphore-git-failure state) is ready, please label it accordingly. If that is not enough to restart merging, I will merge one such labeled PR manually.

PR #381, opened 2019-03-08
Fixed bootstrapping on Debian

qb-syn:
I did had some difficulties to run the `bootstrap.sh` in order to build squid-cache from source on a Debian box. It seems to me that the Debian libtool packages is naming the binary 'libtoolize' which makes `find_variant` reporting an error.

As its seems lsb_release is required to run this script, I'm using it to detect if the Linux system is Debian or not and if, change the `LIBTOOL_BIN` path.

Seems to work on Debian, but I'm unsure on the repercussion on other distro.

Is this going ok or not?

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -8,6 +8,7 @@
 
 To build and install the Squid Cache, type:
 
+        % ./bootstrap.sh
 	% ./configure --prefix=/usr/local/squid
         % make all
         % make install
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -93,6 +93,11 @@ then
   LIBTOOL_BIN="glibtool"
 else
   LIBTOOL_BIN="libtool"
+  lsb_release_id="$(lsb_release -si)"
+  if [ -n "${lsb_release_id}" ] \
+         && [ "${lsb_release_id}" = "Debian" ]; then
+      LIBTOOL_BIN="libtoolize"
+  fi
 fi
 
 # Adjust paths of required autool packages
@@ -138,7 +143,7 @@ do
 	    # Bootstrap the autotool subsystems
 	    bootstrap aclocal$amver $acincludeflag
 	    bootstrap autoheader$acver
-	    bootstrap_libtoolize ${LIBTOOL_BIN}ize${ltver}
+	    bootstrap_libtoolize ${LIBTOOL_BIN}${ltver}
 	    bootstrap automake$amver --foreign --add-missing --copy -f
 	    bootstrap autoconf$acver --force
 	fi ); then


INSTALL
```diff
@@ -8,6 +8,7 @@
 
 To build and install the Squid Cache, type:
 
+        % ./bootstrap.sh
```
rousskov: INSTALL instructions are not for developers. Non-developers should not run bootstrap.sh -- it is unnecessary, likely to fail in their environments, complicates support, and failures may be difficult to recover from because they damage files extracted from the official tarball.
```suggestion
```

bootstrap.sh
```diff
@@ -93,6 +93,11 @@ then
   LIBTOOL_BIN="glibtool"
 else
   LIBTOOL_BIN="libtool"
+  lsb_release_id="$(lsb_release -si)"
+  if [ -n "${lsb_release_id}" ] \
```
rousskov: If this `-n` expression works (i.e. does not lead to shell errors when the lsb_release_id variable has an empty value, then the following `Debian` equality test will work as well. And if this `-n` expression does not work, then the following equality test will not work as well. Thus, you do not need the `-n` part of the test AFAICT.

bootstrap.sh
```diff
@@ -93,6 +93,11 @@ then
   LIBTOOL_BIN="glibtool"
 else
   LIBTOOL_BIN="libtool"
```
rousskov: When adding more special cases to this `if` statement, we should make them all follow the same pattern:

```
if [ `uname -s 2>/dev/null` = 'Darwin' ]
then
  LIBTOOL_BIN="glibtool"
elif [`lsb_release -si 2>/dev/null` = 'Debian' ]
then
  LIBTOOL_BIN="libtoolize"
else
  LIBTOOL_BIN="libtool"
fi
```

bootstrap.sh
```diff
@@ -138,7 +143,7 @@ do
 	    # Bootstrap the autotool subsystems
 	    bootstrap aclocal$amver $acincludeflag
 	    bootstrap autoheader$acver
-	    bootstrap_libtoolize ${LIBTOOL_BIN}ize${ltver}
+	    bootstrap_libtoolize ${LIBTOOL_BIN}${ltver}
```
rousskov: Would not this break bootstrap where it was working before? For example, this used to get `glibtoolize` on Darwin but now will only get `glibtool`. Why would this continue to work on Darwin?

rousskov: Semaphore CI build [failure](https://semaphoreci.com/squid/squid/branches/pull-request-381/builds/1) increase my concerns about the `ize` suffix removal.

rousskov:
@qb-syn, what ./bootstrap.sh error were you getting on Debian before this PR?

qb-syn:
@rousskov I did get:

```
$ ./bootstrap.sh
WARNING: Cannot find libtool version .
./bootstrap.sh: 1: ./bootstrap.sh: libtool: not found
```

But, as this file is for dev only I will withdraw my PR which does not bring any goods.

rousskov:
> as this file is for dev only I will withdraw my PR which does not bring any goods.

I do not understand your logic. PRs that benefit Squid developers are certainly welcomed! If bootstrapping fails on Debian, we should fix it. Your PR needs more work, but if the problem you encountered is real (as opposed to some local misconfiguration), then please consider finishing that work.

yadij:
Debian separated the libtool package into two parts. The libtoolize binary is in a package called libtool, and a 'libtool' binary in a libtool-bin package. 

You can silence the warning by installing that libtool-bin package, but it is *only* used to display the libtool version number. So may be sufficient to check for libtoolize existing, but im not sure about other OS.

PR #257, opened 2018-07-17
Fixed assert( !isOpen(conn->fd) ) on log rotate.

vel21ripn:

PR #382, opened 2019-03-19
Bug 4796: comm.cc !isOpen(conn->fd) assertion when rotating logs

rousskov:
Squid abandoned cache.log file descriptor maintenance, calling fd_open()
but then closing the descriptor without fd_close(). If the original file
descriptor value was reused for another purpose, Squid would either hit
the reported assertion or log a "Closing open FD" WARNING (depending on
the new purpose). The cache.log file descriptor is closed on log
rotation and reconfiguration events.

This short-term solution avoids assertions and WARNINGs but sacrifices
cache.log listing in fd_table and, hence, mgr:filedescriptors reports.

The correct long-term solution is to properly maintain descriptor meta
information across cache.log closures/openings, but doing so from inside
of debug.cc is technically difficult due to linking boundaries/problems.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1151,8 +1151,6 @@ mainInitialize(void)
 
     _db_init(Debug::cache_log, Debug::debugOptions);
 
-    fd_open(fileno(debug_log), FD_LOG, Debug::cache_log);
-
     debugs(1, DBG_CRITICAL, "Starting Squid Cache version " << version_string << " for " << CONFIG_HOST_TYPE << "...");
     debugs(1, DBG_CRITICAL, "Service Name: " << service_name);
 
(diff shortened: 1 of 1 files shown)


rousskov:
This is a short-term fix by @vel21ripn that was mentioned in the abandoned/closed PR #257.

rousskov:
@squid-cache/commiters, If there are no last-minute objections, I will clear this for commit. See [earlier](https://github.com/squid-cache/squid/pull/382#pullrequestreview-215957395) comment for rationale.

rousskov:
> Instead of fully erasing, can you comment it out and add a reference to the bug report please.

Done at fb9f02b.

PR #383, opened 2019-03-19
When using OpenSSL, trust intermediate CAs from trusted stores

rousskov:
According to [1], GnuTLS and NSS do that by default.

Use case: Chrome and Mozilla no longer trust Semantic root CAs _but_
still trust several whitelisted Semantic intermediate CAs[2]. Squid
built with OpenSSL cannot do that without X509_V_FLAG_PARTIAL_CHAIN.

[1] https://www.openldap.org/lists/openldap-devel/201506/msg00012.html
[2] https://wiki.mozilla.org/CA/Additional_Trust_Changes#Symantec

### DIFF
--- a/src/security/PeerOptions.cc
+++ b/src/security/PeerOptions.cc
@@ -287,6 +287,7 @@ Security::PeerOptions::createClientContext(bool setOptions)
         updateContextNpn(t);
         updateContextCa(t);
         updateContextCrl(t);
+        updateContextTrust(t);
     }
 
     return t;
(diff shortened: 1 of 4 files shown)


rousskov:
Initial customer feedback and alleged GnuTLS/NSS _default_ behavior suggest that we should always enable the proposed behavior. And with automatic intermediate certificate fetching in v4, there is no need to populate the trusted store with intermediate certificates that complete chains, decreasing the negative impact of this no-configuration support.

FWIW, the Curl project is an example where somebody [proposed](https://curl.haxx.se/mail/lib-2015-11/0079.html) to change the default, but the resulting discussion stalled, with several objections. AFAICT, curl still has no such support, neither by default nor via a configuration option, but I would not be surprised if the recent changes related to Semantic CAs force their hand.

rousskov:
> It may be better to set the flag before updating the context CRL, so if the CRL was signed by one of these intermediaries it does not fail to verify.

If my reading of OpenSSL code is correct, then simply adding CRL to the context (i.e. calling X509_STORE_add_crl()) does not result in any verifications. I suspect it will stay that way because validating CRLs during context configuration would create a chicken-and-egg problem -- CRLs _and_ many other context options/parameters affect validation (not to mention that it would waste a lot of CPU cycles on validating something that may never be actually used).

PR #384, opened 2019-03-20
Fix build broken by missing <map> in 36c774

eduard-bagdasaryan:
In all tested environments, something else #included `<map>`.

### DIFF
--- a/src/proxyp/Elements.cc
+++ b/src/proxyp/Elements.cc
@@ -11,28 +11,53 @@
 #include "proxyp/Elements.h"
 #include "sbuf/Stream.h"
 
+#include <algorithm>
 #include <limits>
-
-const ProxyProtocol::FieldMap ProxyProtocol::PseudoHeaderFields = {
-    { SBuf(":version"), ProxyProtocol::Two::htPseudoVersion },
-    { SBuf(":command"), ProxyProtocol::Two::htPseudoCommand },
-    { SBuf(":src_addr"), ProxyProtocol::Two::htPseudoSrcAddr },
-    { SBuf(":dst_addr"), ProxyProtocol::Two::htPseudoDstAddr },
-    { SBuf(":src_port"), ProxyProtocol::Two::htPseudoSrcPort },
-    { SBuf(":dst_port"), ProxyProtocol::Two::htPseudoDstPort }
-};
+#include <map>
 
 namespace ProxyProtocol {
+namespace Two {
+
+/// a mapping between pseudo header names and ids
+typedef std::map<SBuf, FieldType> FieldMap;
+static const FieldMap PseudoHeaderFields = {
+    { SBuf(":version"), htPseudoVersion },
+    { SBuf(":command"), htPseudoCommand },
+    { SBuf(":src_addr"), htPseudoSrcAddr },
+    { SBuf(":dst_addr"), htPseudoDstAddr },
+    { SBuf(":src_port"), htPseudoSrcPort },
+    { SBuf(":dst_port"), htPseudoDstPort }
+};
+
+} // namespace Two
+
 static Two::FieldType NameToFieldType(const SBuf &);
 static Two::FieldType IntegerToFieldType(const SBuf &);
+
 } // namespace ProxyProtocol
+
+
+const SBuf &
+ProxyProtocol::PseudoFieldTypeToFieldName(const Two::FieldType fieldType)
+{
+    // reverse PseudoHeaderFields so that we can look up names by FieldType
+    typedef std::vector<SBuf> PsuedoFieldNames;
+    static PsuedoFieldNames psuedoFieldNames;
+    if (psuedoFieldNames.empty()) {
+        std::for_each(Two::PseudoHeaderFields.begin(), Two::PseudoHeaderFields.end(),
+            [](const Two::FieldMap::value_type &item) {
+                psuedoFieldNames.at(item.second) = item.first;
+            });
+    }
+    return psuedoFieldNames.at(fieldType - Two::htPseudoBegin);
+}
 
 /// FieldNameToFieldType() helper that handles pseudo headers
 ProxyProtocol::Two::FieldType
 ProxyProtocol::NameToFieldType(const SBuf &name)
 {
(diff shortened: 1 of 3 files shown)


src/proxyp/Elements.h
```diff
@@ -11,6 +11,8 @@
 
 #include "sbuf/SBuf.h"
 
+#include <map>
```
rousskov: It is rather unfortunate that we have to include <map> in what-is-supposed-to-be elementary header file just so that proxyp/Message.cc can iterate pseudo header names. proxyp/Message.cc actually does not care how the pseudo headers are stored. Can you think of a better alternative?

yadij: No matter what type is used, it needs to be included because it is used in two different .cc. That is the bug being fixed here.

rousskov: Addressing my concern would require changing what type is used in at least one of those .cc files.

rousskov: Untested (beyond building) faf92c2b64e8b62513b548a67c32207c94aa77c8 addresses my concern by hiding PseudoHeaderFields type inside proxyp/Elements.cc. If you like the idea, please retest and re-review. Otherwise, please revert and commit the original fix.

src/proxyp/Elements.cc
```diff
@@ -11,28 +11,53 @@
...
 static Two::FieldType NameToFieldType(const SBuf &);
 static Two::FieldType IntegerToFieldType(const SBuf &);
+
 } // namespace ProxyProtocol
 
+
+const SBuf &
+ProxyProtocol::PseudoFieldTypeToFieldName(const Two::FieldType fieldType)
+{
+    // reverse PseudoHeaderFields so that we can look up names by FieldType
```
yadij: "reverse" is the wrong term here. This is generating a new mapping, not changing the contents of PseudoHeaderFields.

src/proxyp/Elements.cc
```diff
@@ -11,28 +11,53 @@
...
+    typedef std::vector<SBuf> PsuedoFieldNames;
```
yadij: This typedef is unnecessary code.

src/proxyp/Elements.cc
```diff
@@ -11,28 +11,53 @@
...
+    static PsuedoFieldNames psuedoFieldNames;
```
yadij: s/psuedo/pseudo/ and all the other typos.

src/proxyp/Elements.cc
rousskov: Addressed in 90e929a? Changed to "flip" in hope to guess what you consider the "right" term to be.

FWIW, "reverse" is a popular and meaningful term here, probably based on "reverse lookup" which is a well-known technical term. The code reverses the roles of map keys and map values to optimize reverse lookups. Lots of online discussions refer to this operation as reversing the map. I cannot find a better term for that operation ("transpose" is not it).

src/proxyp/Elements.cc
rousskov: Left as is because naming non-trivial types is good practice: Naming makes some compiler errors clearer, often simplifies working with a debugger, may reduce future code changes, and may reduce instances where copy-pasting folks incorrectly duplicate types because they forget to name them (in contexts where the type is actually spelled out multiple times).

I will refactor to remove this declaration if you insist on this change. Do you insist on this change?

src/proxyp/Elements.cc
rousskov: Fixed in b21fed9.

src/proxyp/Elements.cc
yadij: The PseudoHeaderFields is a 2-dimensional container. It has two directions which can be reversed / flipped / swapped. The comment does not make it clear which is happening *and* implies PseudoHeaderFields (as the singular subject) is being changed somehow. So one has to understand both the code and mapping terminology to read the comment properly - which makes it confusing / broken.

The improvements that I can think of right now are:
 * name the static local pseudoHeaderFieldsByName and remove the comment.
  That makes the code speak for itself as being some way to initialize the _new_ mapping.

* have the comment say the code action, not the abstraction.
```suggestion
    // invert mapping from PseudoHeaderFields so that we can look up names by FieldType
```

src/proxyp/Elements.cc
yadij: good != best. And the nature of its goodness is debatable. If it were a long function or a deeply nested template definition - sure typedef has merits. This is not one of those cases.

FYI, there have been some serious conversations starting this past year in Debian about large software (eg Firefox, Apache, Bind, Chrome) becoming non-compilable due to symbol table bloat slowing the compiler and linker down. You may recall some times in the past when Squid itself hit these issues in the BuildFarm VMs. To void spreading that further IMO we should only typedef when it is actively useful. With C++11 theoretically we should not need to use it at all.

rousskov: I am interpreting your comment as "yes, I insist on this change" answer. Please let me know if have misinterpreted it.

src/proxyp/Elements.cc
rousskov: @eduard-bagdasaryan, after fixing the runtime problems, please use the comment wording and variable naming that Amos likes to address this change request.

src/proxyp/Elements.cc
rousskov: @eduard-bagdasaryan, after fixing runtime errors, please remove this typedef to address this change request.

eduard-bagdasaryan: Removed at 87f38f.

src/proxyp/Elements.cc
eduard-bagdasaryan: Fixed at 87f38f. The code around this comment was refactored and I removed it.

rousskov:
@yadij, it looks like all your change requests have been addressed (including made obsolete by other changes). GitHub currently shows you as voting for this code acceptance, but I am not sure whether that state is a misleading artifact of multiple comments -- your last review did not sound like acceptance to me. Thus, I am not going to clear this PR for merge so that you can either request more changes or clear it yourself.

PR #387, opened 2019-04-05
Made Coverity happier after f5e1794 (Peering for SslBump)

kinkie:
CID 1444393: Uninitialized members (UNINIT_CTOR)

From the current code point of view, this is a false positive because
the private constructor which lacked initialization was only used from
other constructors that do initialize httpStatus.


kinkie:
Ok to test

rousskov:
I polished PR description to obey title length limits and to improve consistency with recent similar fixes. Please adjust further as needed, of course.

PR #389, opened 2019-04-07
Release Notes: Q tag is not supported by linuxdoc

yadij:


yadij:
Unescaped quote seems to work. Thank you.

I'm not sure when the change to SGML happened exactly. These showed up with linuxdoc upgrades more strictly validating the latest SGML syntax.

PR #390, opened 2019-04-08
Check user-provided PATH for krb5-config

yadij:
When MIT or Heimdal Keberos libraries are installed at a custom
location there may be several krb5-config installed. The one
located at the user-provided path (if any) needs to have preference.


kinkie:
ok to test

yadij:
Passed 10 days without objections. Tester response was that this worked.

PR #392, opened 2019-04-11
V4 next backports

squidadm:


yadij:
OK to test

xeron:
What about this change? https://github.com/squid-cache/squid/pull/370

I'd love to see it backported. We're currenty applying it as a patch to 4.6 and running in production.

yadij:
This set was the first pass for patches that were easy to apply with a clear need in v4. The PR #370 patch and some others were not quite a clear-cut (its meta said it started in v5). Thanks for the feedback, I will likely add it to the next port cycle.

PR #393, opened 2019-04-12
Bug 4912: same-name notes being appended instead of replaced

amishmm:
When auth helper replies with clt_conn_tag=foo (or any KV pair) and
then later in future replaces it with clt_conn_tag=bar. The new pair
was getting appended instead of getting replaced.

i.e. notes would contain two KV pairs:
clt_conn_tag=foo
clt_conn_tag=bar

This fixes it and also fixes:
https://bugs.squid-cache.org/show_bug.cgi?id=4912

### DIFF
--- a/src/Notes.cc
+++ b/src/Notes.cc
@@ -355,6 +355,15 @@ NotePairs::appendNewOnly(const NotePairs *src)
 }
 
 void
+NotePairs::replaceOrAddOrAppend(const NotePairs *src, const std::vector<SBuf> &appendKeys)
+{
+    for (auto e: src->entries)
+        if (std::find(appendKeys.begin(), appendKeys.end(), e->name()) == appendKeys.end())
+            remove(e->name());
+    append(src);
+}
+
+void
 NotePairs::replaceOrAdd(const NotePairs *src)
 {
     for (auto e: src->entries)
--- a/src/Notes.cc
+++ b/src/Notes.cc
@@ -355,11 +355,12 @@ NotePairs::appendNewOnly(const NotePairs *src)
 }
 
 void
-NotePairs::replaceOrAddOrAppend(const NotePairs *src, const std::vector<SBuf> &appendKeys)
+NotePairs::replaceOrAddOrAppend(const NotePairs *src, const NotePairs::Names &appendables)
 {
-    for (auto e: src->entries)
-        if (std::find(appendKeys.begin(), appendKeys.end(), e->name()) == appendKeys.end())
+    for (const auto e: src->entries) {
+        if (std::find(appendables.begin(), appendables.end(), e->name()) == appendables.end())
             remove(e->name());
+    }
     append(src);
 }
 
--- a/src/Notes.h
+++ b/src/Notes.h
@@ -200,6 +200,11 @@ public:
     void append(const NotePairs *src);
 
     /// Replace existing list entries with the src NotePairs entries.
+    /// Do not replace some special keys but append them instead.
+    /// Entries which do not exist in the destination set are added.
+    void replaceOrAddOrAppend(const NotePairs *src, const std::vector<SBuf> &appendKeys);
+
+    /// Replace existing list entries with the src NotePairs entries.
     /// Entries which do not exist in the destination set are added.
     void replaceOrAdd(const NotePairs *src);
 
--- a/src/Notes.h
+++ b/src/Notes.h
@@ -191,6 +191,7 @@ public:
         SBuf theValue;
     };
     typedef std::vector<Entry::Pointer> Entries;      ///< The key/value pair entries
+    typedef std::vector<SBuf> Names;
 
     NotePairs() {}
     NotePairs &operator=(NotePairs const &) = delete;
@@ -200,9 +201,9 @@ public:
     void append(const NotePairs *src);
 
     /// Replace existing list entries with the src NotePairs entries.
-    /// Do not replace some special keys but append them instead.
+    /// Do not replace but append entries named in the appendables
     /// Entries which do not exist in the destination set are added.
-    void replaceOrAddOrAppend(const NotePairs *src, const std::vector<SBuf> &appendKeys);
+    void replaceOrAddOrAppend(const NotePairs *src, const Names &appendables);
 
     /// Replace existing list entries with the src NotePairs entries.
     /// Entries which do not exist in the destination set are added.
--- a/src/auth/basic/UserRequest.cc
+++ b/src/auth/basic/UserRequest.cc
@@ -168,7 +168,7 @@ Auth::Basic::UserRequest::HandleReply(void *data, const Helper::Reply &reply)
 
     // add new helper kv-pair notes to the credentials object
     // so that any transaction using those credentials can access them
-    r->auth_user_request->user()->notes.appendNewOnly(&reply.notes);
+    r->auth_user_request->user()->notes.replaceOrAdd(&reply.notes);
 
     /* this is okay since we only play with the Auth::Basic::User child fields below
      * and do not pass the pointer itself anywhere */
--- a/src/auth/basic/UserRequest.cc
+++ b/src/auth/basic/UserRequest.cc
@@ -168,8 +168,8 @@ Auth::Basic::UserRequest::HandleReply(void *data, const Helper::Reply &reply)
 
     // add new helper kv-pair notes to the credentials object
     // so that any transaction using those credentials can access them
-    static const std::vector<SBuf> appendKeys = { SBuf("group"), SBuf("tag") };
-    r->auth_user_request->user()->notes.replaceOrAddOrAppend(&reply.notes, appendKeys);
+    static const NotePairs::Names appendables = { SBuf("group"), SBuf("tag") };
+    r->auth_user_request->user()->notes.replaceOrAddOrAppend(&reply.notes, appendables);
 
     /* this is okay since we only play with the Auth::Basic::User child fields below
      * and do not pass the pointer itself anywhere */
--- a/src/auth/basic/UserRequest.cc
+++ b/src/auth/basic/UserRequest.cc
@@ -168,7 +168,8 @@ Auth::Basic::UserRequest::HandleReply(void *data, const Helper::Reply &reply)
 
     // add new helper kv-pair notes to the credentials object
     // so that any transaction using those credentials can access them
-    r->auth_user_request->user()->notes.replaceOrAdd(&reply.notes);
+    static const std::vector<SBuf> appendKeys = { SBuf("group"), SBuf("tag") };
+    r->auth_user_request->user()->notes.replaceOrAddOrAppend(&reply.notes, appendKeys);
 
     /* this is okay since we only play with the Auth::Basic::User child fields below
      * and do not pass the pointer itself anywhere */
--- a/src/auth/digest/UserRequest.cc
+++ b/src/auth/digest/UserRequest.cc
@@ -327,7 +327,8 @@ Auth::Digest::UserRequest::HandleReply(void *data, const Helper::Reply &reply)
 
     // add new helper kv-pair notes to the credentials object
     // so that any transaction using those credentials can access them
-    auth_user_request->user()->notes.appendNewOnly(&reply.notes);
+    static const NotePairs::Names appendables = { SBuf("group"), SBuf("nonce"), SBuf("tag") };
+    auth_user_request->user()->notes.replaceOrAddOrAppend(&reply.notes, appendables);
     // remove any private credentials detail which got added.
     auth_user_request->user()->notes.remove("ha1");
 
--- a/src/auth/negotiate/UserRequest.cc
+++ b/src/auth/negotiate/UserRequest.cc
@@ -278,7 +278,8 @@ Auth::Negotiate::UserRequest::HandleReply(void *data, const Helper::Reply &reply
 
     // add new helper kv-pair notes to the credentials object
     // so that any transaction using those credentials can access them
-    auth_user_request->user()->notes.appendNewOnly(&reply.notes);
+    static const NotePairs::Names appendables = { SBuf("group"), SBuf("tag") };
+    auth_user_request->user()->notes.replaceOrAddOrAppend(&reply.notes, appendables);
     // remove any private credentials detail which got added.
     auth_user_request->user()->notes.remove("token");
 
--- a/src/auth/ntlm/UserRequest.cc
+++ b/src/auth/ntlm/UserRequest.cc
@@ -272,7 +272,8 @@ Auth::Ntlm::UserRequest::HandleReply(void *data, const Helper::Reply &reply)
 
     // add new helper kv-pair notes to the credentials object
     // so that any transaction using those credentials can access them
-    auth_user_request->user()->notes.appendNewOnly(&reply.notes);
+    static const NotePairs::Names appendables = { SBuf("group"), SBuf("tag") };
+    auth_user_request->user()->notes.replaceOrAddOrAppend(&reply.notes, appendables);
     // remove any private credentials detail which got added.
     auth_user_request->user()->notes.remove("token");
 


amishmm:
Note that currently this PR just fixes basic authentication. But similar code exists in digest, negotiate, ntlm schemes.

Since I dont use / am not aware of those auth schemes, I have not touched the code.

amishmm:
Mailing list reference: (@rousskov 's reply)
http://lists.squid-cache.org/pipermail/squid-users/2018-December/019816.html

Earlier commit which was supposed to do it but apparently did not:
https://github.com/squid-cache/squid/commit/457857fe7cf51037cd9e54e86c0985391d7ea594

Quote from it:
> Also after this patch the notes comming from helpers replaces any existing
> note values.


The reason this commit did not fix it is because UpdateRequestNotes() called replaceOrAdd() which removed duplicate keys in destination notes but did not check for duplicate keys in source notes.

So if source had duplicates it would get duplicated in destination too.

And duplicate source notes were coming from helper code - that code always appended the new KV pair by calling appendNewOnly(). That code matched on unique key=value pair instead of just unique key. Hence creating duplicate keys which populated to request notes eventually.

So replacing appendNewOnly() with replaceOrAdd() in helper code itself - fixes the bug.

yadij:
This breaks auth helpers needing to supply multiple kv-pairs for the same named key. For example; a user can have multiple groups (thus group=), multiple connections (thus clt_conn_tag=), and via those multiple concurrent auth tokens (thus token= and nonce=).

What we need is a defined extension to the helper protocol to separate the append vs replace intention. I have been thinking a "key+=value" for add and/or "key:=value" for replace. If the default for regular "key=value" is changed to the replace you are wanting we will need backward compatibility handling of the existing keys (mentioned above) which require append behaviour.

rousskov:
I am traveling for a few days and cannot review this immediately. Please do 
not jump to interface conclusions just yet.


IIRC, the original/intended design was based on careful analysis of the 
options and didn't break anything. Perhaps multivalued annotations were 
supposed to be supported via CSVs? Not sure. Christos may remember more. 
Please ping him.

amishmm:
> This breaks auth helpers needing to supply multiple kv-pairs for the same named key. For example; a user can have multiple groups (thus group=), multiple connections (thus clt_conn_tag=),

But if you see clt_conn_tag documentation: http://www.squid-cache.org/Doc/config/url_rewrite_program/

> clt_conn_tag=TAG ... A helper may update the TAG during subsequent requests be returning a new kv-pair.

Also @rousskov 's reply in mailing list as well as comment in the commit mentioned above suggests that KV pairs by helpers are supposed to be replaced instead of appended.

> and via those multiple concurrent auth tokens (thus token= and nonce=).

Yes thats why I didnt touch other types of helpers. As I dont have idea how they worked.

amishmm:
> Perhaps multivalued annotations were supposed to be supported via CSVs?

Yes I think that would be much easier way out which supports both cases.

If helper wants multiple values it could return:
group=foo,bar,...

If helper wanted to overwrite it could return:
group=baz

yadij:
> > This breaks auth helpers needing to supply multiple kv-pairs for the same named key. For example; a user can have multiple groups (thus group=), multiple connections (thus clt_conn_tag=),
> 
> But if you see clt_conn_tag documentation: http://www.squid-cache.org/Doc/config/url_rewrite_program/
> 
> > clt_conn_tag=TAG ... A helper may update the TAG during subsequent requests be returning a new kv-pair.
> 

The URL-rewrite interface is applied per-transaction. The annotations it produces are added directly to the transaction state and die with that transaction. The clt_conn_tag in particular is updated on the *single* connection the request arrived on.

The auth interface annotations are stored as part of the credentials, not a transaction. A clt_conn_tag received on that interface is added to the credentials and updated on the connection which supplied those particular credentials.
 ... also every other request that uses those credentials in future has its connection updated with the tag. Until the user is logged out / garbage collected.

Discarding duplicates of that particular clt_conn_tag key, or your policy_ custom key may be fine. But throwing away all same-name keys will break other helper setups that use the current behaviour.


> Also @rousskov 's reply in mailing list as well as comment in the commit mentioned above suggests that KV pairs by helpers are supposed to be replaced instead of appended.
> 

Between helpers or helper lookups, what Alex stated is expected. But as I said above username scope exceeds transaction and even connection scope.

> > and via those multiple concurrent auth tokens (thus token= and nonce=).
> 
> Yes thats why I didnt touch other types of helpers. As I dont have idea how they worked.

One of the problems with this patch currently is that each helper binary also works differently. So by altering the whole interface behaviour around note storage you are affecting other helpers behaviour. Most specifically helpers producing group= tags.

But this still has not solved the stated append behaviour completely - since other interfaces can still append key_= tags to the total set of annotations on a username. All you have stopped here is the Basic auth username cache recording the policy history of credentials.

> > Perhaps multivalued annotations were supposed to be supported via CSVs?
> 
> Yes I think that would be much easier way out which supports both cases.
> 
> If helper wants multiple values it could return:
> group=foo,bar,...
> 
> If helper wanted to overwrite it could return:
> group=baz

This overlooks the interaction between auth being done by auth_param and external_acl_type helpers. They may be queried in either order, and have different knowledge levels of groups to assign to the same username. Due to lookup caching and queuing either one may be omitted lookup entirely on a transaction. So which helpers CSV list gets used?

amishmm:
Ok to be honest I have failed to understand much of what you said. (as I do not know much about different places where these notes are used)

But if I understood correctly, atleast ctl_conn_tag should be updated with new value (replacing old value instead of appending the new value).

But when I am doing the testing - I see that clt_conn_tag also have multiple values too.

I can modify the PR in such a way that all KV's will be kept (even if duplicate) except clt_conn_tag which will be replaced. Is it fine?

yadij:
Lets see if this clarifies:
* A request can only arrive in one connection - one clt_conn_tag.
* A username may be used on multiple connections simultaneously - multiple clt_conn_tag.

amishmm:
> Lets see if this clarifies:
> 
>     * A request can only arrive in one connection - one clt_conn_tag.
> 
>     * A username may be used on multiple connections simultaneously - multiple clt_conn_tag.

Yes but I would expect that each request / each connection will have single clt_conn_tag.

By connection I assume --- connection to squid port (SYN) - till the end of connection (FIN)

In that connection there may be multiple requests by way of persistent connection? So an external acl or auth helper may get called multiple times? But if external acl or auth helper changes clt_conn_tag - i would expect that it gets replaced instead of getting appended.

Sorry if conceptually I have misunderstood.

amishmm:
Ok let me give simple example of what I am trying to do.

1st connection)
curl -x 127.0.0.1:8080 http://yahoo.com
Auth helper called and returns OK clt_conn_tag=foo

2nd connection)
curl -x 127.0.0.1:8080 http://google.com
Auth helper called and returns OK clt_conn_tag=bar

But what I see is that %note{clt_conn_tag} returns "foo,bar"

This is the bug (atleast I consider it as bug) I am trying to fix. When a next connection changes clt_conn_tag I want it to be updated and not appended.

yadij:
> Ok let me give simple example of what I am trying to do.
> 
> 1st connection)
> curl -x 127.0.0.1:8080 -U admin:one http://yahoo.com
> Auth helper called and returns OK clt_conn_tag=foo
> 
> 2nd connection)
> curl -x 127.0.0.1:8080 -U admin:one http://google.com
> Auth helper called and returns OK clt_conn_tag=bar
> 

Were the credentials still valid at the time of the second request?
 if so why was the helper called at all?

What is the difference between the strings "admin:one" and "admin:one" and "admin:one" that they should have different tags?

amishmm:
How to reproduce:

squid.conf

```
auth_param basic program /usr/lib/squid/test_auth
auth_param basic credentialsttl 5 seconds
acl proxyauthport myportname 3128
acl proxypass proxy_auth REQUIRED
http_access deny proxyauthport !proxypass

logformat testlog %ru %note{clt_conn_tag}
access_log daemon:/var/log/squid/proxyauth.log logformat=testlog all
```

Sample helper (accepts any username / password sets clt_conn_tag as a counter).
/usr/lib/squid/test_auth

```
#!/usr/bin/perl
$|=1;
$count = 0;
while (<>) {
    ($user, $password) = split;
    $user =~ s/%(..)/pack("H*", $1)/ge;
    $password =~ s/%(..)/pack("H*", $1)/ge;
    ++$count;
    print "OK clt_conn_tag=$count\n";
}
```

Now call repeatedly: (with 6 second interval to avoid caching)
curl -x 127.0.0.1:3128 -U admin:one http://google.com

cat /var/log/squid/proxyauth.log

```
http://google.com/ 1
http://google.com/ 1,2
http://google.com/ 1,2,3
http://google.com/ 1,2,3,4
http://google.com/ 1,2,3,4,5
```

amishmm:
> Were the credentials still valid at the time of the second request?
> if so why was the helper called at all?

Please see my post just above (for testing i set caching as 5 seconds only)

> What is the difference between the strings "admin:one" and "admin:one" and "admin:one" that they should have different tags?

In my scenario clt_conn_tag represents the users policy. (which is picked from a database). So if users policy changes from foo to bar after 5 seconds, clt_conn_tag should also change from foo to bar.

In above example assume that policy changes from name "1" to name "2"

rousskov:
I suspect that at least some of the confusion (and bugs?) come from mixing two different concepts:

1. Custom annotations like `clt_conn_tag=foo` and `bar=baz`. These annotations are simple concepts unrelated to authentication. They annotate the current transaction (by default) or the client TCP connection (`clt_conn_tag` and `annotate_client`). They should overwrite/replace old values (by default). If a helper wants to preserve the old value, the helper can use CSV or another helper-specific multivalue syntax compatible with Squid's whitespace-based (IIRC) helper response parser. We can also enlarge support for `+=` syntax that  `annotate_client` already supports (but there may be no pressing need to do that outside  `annotate_client` scope). These annotation principles apply to authentication helpers, external ACL helpers, eCAP adapters, the `note` directive, `annotate_transaction` ACL, etc. -- anything that can annotate.
1. Authentication-related values like `ttl=0` or `group=foo` returned by some authentication helpers. These things have authentication-related definitions, lifetimes, storage, etc. Squid "knows" what they mean. They are not custom annotations from concept 1 above.

For example, the `clt_conn_tag` feature simply tags/annotates a TCP client connection. It can be used for _faking_ authentication, but it is a concept 1 annotation, not a concept 2 feature.  @amishmm has demonstrated that this annotation currently persists across connections. That annotation persistence is a bug -- the feature was not designed (and is not documented) to cross connection boundaries.

Since @yadij is talking about user names, credentials, and similar authentication concepts in the context of this PR, I assume that he thinks that this PR breaks authentication (i.e., the PR breaks some aspects of concept 2 support). If he is right, then the PR needs to be adjusted, of course, but we should still do the right thing with regard to concept 1 support -- annotations should not persist across transactions (by default) or client connections (in a few well-documented exceptional connection annotating cases).

Authentication is not my specialty, but I hope the above classification would help us reach consensus.

amishmm:
I can modify PR which replaces the value if key name is clt_conn_tag or ends with _ (underscore). Otherwise for any other keyname it will append it (like in current squid)

Is it fine?

rousskov:
> I can modify PR which replaces the value if key name is clt_conn_tag or ends with _ (underscore). Otherwise for any other keyname it will append it (like in current squid)

I am not sure, but since many helpers are likely to use custom key=value pairs with key names _not_ ending in underscore, and since most reserved-for-Squid key=value pairs cannot be meaningfully appended, it feels like we should:

1. Append to `group`, provided it had been appended to before `clt_conn_tag` was added , and Amos confirms that this appending is the right default. Discuss otherwise.
1. If there are any other `group`-like keys with values that were appended before and should still be appended, then append their values as well. I do not know of any, but I may have missed or misclassified some.
1. Overwrite values for all other keys, including `clt_conn_tag`, `tag`, and all keys that end with `_`. Overwriting is the right default and also the documented/expected behavior.

amishmm:
@yadij waiting for your inputs / thoughts.

amishmm:
Pinging @yadij again. Because I wouldnt be able to modify PR without exact clarity on what needs to be done. Thank you.

yadij:
I think the simplest thing would be an optional parameter to Notes::replaceOrAdd() which takes a whitelist of key names that skip the remove(). Then you can choose to update only the Basic auth to call replaceOrAdd() here, or update all the relevant interfaces.

If you do want to go beyond Basic auth the keys that AFAIK are being actively used as multiples are:
* Basic:  "group", "tag"
* Digest: "group", "nonce", "tag"
* NTLM / Negotiate:  "group", "tag"
* external ACL: "group", "tag"
* URL-rewrite: "group", "tag"

The other helpers AFAICS do not use any multi-value keys, and the other multi-value keys (eg token, ha1) are handled specially outside the cached notes list. That latter is why I do not think we can centralize the whitelist any further than this per-interface level.

amishmm:
As per suggestion by @yadij - I have added a new function for NotePairs called replaceOrAddOrAppend() where you  can specify which keys to append instead of replace.

Currently used only by basic auth helper. (group and tag are appended and everything else is replaced)

If it is OKed then I will try to fix other helpers too (as much as possible for me).

Please have a look.

CC: @rousskov

amishmm:
First of all - you made me learn vectors!! So this is my first ever attempt with vectors.

Also I forced-push as its a complete rewrite.

Please have a look. Thank you.

(I am actually working on 4.7 branch. And code seems to be working fine on that. But the vector code for master branch is a lot different. Hope that semaphoreci would not show any compilation failure)

src/auth/basic/UserRequest.cc
```diff
@@ -168,7 +168,8 @@ Auth::Basic::UserRequest::HandleReply(void *data, const Helper::Reply &reply)
 
     // add new helper kv-pair notes to the credentials object
     // so that any transaction using those credentials can access them
-    r->auth_user_request->user()->notes.appendNewOnly(&reply.notes);
+    static const std::vector<std::string> appendKeys = { "group", "tag" };
```
yadij: Make this a vector of SBuf and initialize with { SBuf("group"), SBuf("tag") }

src/Notes.cc
```diff
@@ -354,6 +354,15 @@ NotePairs::appendNewOnly(const NotePairs *src)
     }
 }
 
+void
+NotePairs::replaceOrAddOrAppend(const NotePairs *src, const std::vector<std::string> &appendKeys)
+{
+    for (auto e: src->entries)
+        if (std::find(appendKeys.begin(), appendKeys.end(), const_cast<SBuf &>(e->name()).c_str()) == appendKeys.end())
```
yadij: The cast here is trying to go from char* to a reference of a complex object - that is wrong. Ideally the vector and the e->name() would use the same type and avoid any conversion at all. But our old String class does not do early initialization properly so until the notes member is converted to SBuf (dont do that here) this has to be a little bit messy.

Once the vector definition is using SBuf you should be able to use StringToSBuf(e->name()) to convert safely. That may need sbuf/StringConvert.h included to build, but try without that header first.

src/auth/basic/UserRequest.cc
amishmm: Done!

src/Notes.cc
amishmm: Done! Also sbuf/StringConvert.h is already included in Notes.cc.

Just wondering if StringToSBuf() needs free() somewhere? But then I have no idea about vectors. Do they get freed auto-magically?

yadij: All these objects handle their own memory.

Sorry I steered you a bit wrong earlier. It looks like the StringToSBuf() appears not to be needed at all.

amishmm: Oh yes! Since I was working on 4.7 branch, where name and value of NotePairs are String variables. So it needed StringToSBuf().

But I didnt notice that in master branch, name and value of NotePairs are already migrated to SBuf.

Fixed now. Fingers crossed!

amishmm:
Just incase my comment was lost in code, posting here that the suggested changes have been made.

rousskov:
OK to test

src/Notes.cc
```diff
@@ -354,6 +354,15 @@ NotePairs::appendNewOnly(const NotePairs *src)
     }
 }
 
+void
+NotePairs::replaceOrAddOrAppend(const NotePairs *src, const std::vector<SBuf> &appendKeys)
+{
+    for (auto e: src->entries)
```
rousskov:
```suggestion
    for (const auto e: src->entries) {
```

src/Notes.cc
```diff
@@ -354,6 +354,15 @@ NotePairs::appendNewOnly(const NotePairs *src)
     }
 }
 
+void
+NotePairs::replaceOrAddOrAppend(const NotePairs *src, const std::vector<SBuf> &appendKeys)
```
rousskov:
```suggestion
NotePairs::replaceOrAddOrAppend(const NotePairs *src, const Names &appendables)
```

src/Notes.h
```diff
@@ -199,6 +199,11 @@ class NotePairs: public RefCountable
     /// Append the entries of the src NotePairs list to our list.
     void append(const NotePairs *src);
 
+    /// Replace existing list entries with the src NotePairs entries.
+    /// Do not replace some special keys but append them instead.
```
rousskov:
```suggestion
    /// Do not replace but append entries named in the appendables.
```

src/Notes.h
```diff
@@ -199,6 +199,11 @@ class NotePairs: public RefCountable
...
+    /// Entries which do not exist in the destination set are added.
+    void replaceOrAddOrAppend(const NotePairs *src, const std::vector<SBuf> &appendKeys);
```
rousskov:
```suggestion
    typedef std::vector<SBuf> Names;
    void replaceOrAddOrAppend(const NotePairs *src, const Names &appendables);
```

src/Notes.cc
```diff
@@ -354,6 +354,15 @@ NotePairs::appendNewOnly(const NotePairs *src)
...
+{
+    for (auto e: src->entries)
+        if (std::find(appendKeys.begin(), appendKeys.end(), e->name()) == appendKeys.end())
+            remove(e->name());
+    append(src);
```
rousskov:
```suggestion
    }
    append(src);
```

src/auth/basic/UserRequest.cc
```diff
@@ -168,7 +168,8 @@ Auth::Basic::UserRequest::HandleReply(void *data, const Helper::Reply &reply)
 
     // add new helper kv-pair notes to the credentials object
     // so that any transaction using those credentials can access them
-    r->auth_user_request->user()->notes.appendNewOnly(&reply.notes);
+    static const std::vector<SBuf> appendKeys = { SBuf("group"), SBuf("tag") };
```
rousskov:
```suggestion
    static const NotePairs::Names appendKeys = { SBuf("group"), SBuf("tag") };
```

src/Notes.cc
amishmm: I will change this to "const" but please note that there are other similar functions which dont use "const". (although I will not make changes there unless you tell me to do so)

src/Notes.cc
amishmm: I will add braces here but please note that there are other similar functions which dont use braces for  "for" loop. (toString, findFirst, hasPair etc.) (although I will not make changes there unless you tell me to do so)

src/Notes.cc
amishmm: Done.

src/Notes.cc
amishmm: Done.

src/Notes.h
amishmm: Done

src/Notes.h
amishmm: Done

src/Notes.cc
amishmm: Done

src/auth/basic/UserRequest.cc
amishmm: Done

amishmm:
I have implemented changes requested by @rousskov here: https://github.com/squid-cache/squid/pull/393#pullrequestreview-239832043

I have also implemented this bug fix to other AUTH helpers suggested by @yadij here: https://github.com/squid-cache/squid/pull/393#issuecomment-491481060

I have not implemented it for "external ACL" and "URL-rewrite" because I could not find time to locate the exact code.

Please review and approve if alright. Thank you.

src/Notes.cc
rousskov: Thank you. New and changed code should use const whenever possible. Old code should not be changed just for the sake of adding the missing const.

src/Notes.cc
rousskov: Thank you. Multiline loop bodies in new and changed code should use braces in most cases for clarity and safety sake. Old code should not be changed just for the sake of adding the missing braces.

Most rules have some exceptions, especially when it comes to such vague things as clarity and safety. It is difficult and probably both impractical and unnecessary to codify everything exactly, but there are many "easy" cases like this one where it is fairly clear what the right thing to do is. Thank you for doing the right thing.

rousskov:
@yadij, I have no further comments because @amishmm has addressed all my change requests. Please feel free to clear for merge when you are happy with the code.

amishmm:
Thank you for merge... will this land in squid 4.x version or I will have to wait till 5.x version?

I can backport if required.

yadij:
I am inclined to hold for v5 due to the obvious behavioural changes.

tonchikk:
Hi Team, would you be able document changes committed under this PR in 
https://www.squid-cache.org/Doc/config/auth_param/
https://www.squid-cache.org/Doc/config/acl/
https://wiki.squid-cache.org/Features/AddonHelpers

It was very surprising finding on Squid logic that some pre0defined "tags" are persistent, others are overwritten

rousskov:
> Hi Team, would you be able document changes committed under this PR in [auth_param](https://www.squid-cache.org/Doc/config/auth_param/) [acl](https://www.squid-cache.org/Doc/config/acl/) [AddonHelpers](https://wiki.squid-cache.org/Features/AddonHelpers)

Partially done in annotate_transaction ACL documentation and, via reference to the former, in annotate_client ACL documentation. That documentation was added in 75d47340, more than a year before this _bug-fixing_ PR was merged AFAICT. Here is the current text:

```
Annotations can be specified using replacement and addition
formats. The key=value form replaces old same-key annotation
value(s). The key+=value form appends a new value to the old
same-key annotation. Both forms create a new key=value
annotation if no same-key annotation exists already.
```

> It was very surprising finding on Squid logic that some predefined "tags" are persistent, others are overwritten

I agree that some of Squid annotations handling aspects are difficult to predict. While I am skeptical that documentation alone can fully address that problem, I do believe that Squid documentation is missing a high-level overview of several key concepts, including annotations and helpers. That overview should probably be added to the beginning of cf.data.pre (i.e. squid.conf.documented source). Configuration directive descriptions and even wiki pages can then refer to that concept documentation as needed.

FWIW, I would not hold my breath for the mythical "Team" to improve documentation. Those improvements will not happen until somebody like you posts a PR (or hires somebody to do it) as detailed [elsewhere](https://wiki.squid-cache.org/SquidFaq/AboutSquid#how-to-add-a-new-squid-feature-enhance-of-fix-something).

PR #394, opened 2019-04-19
Bug 4823: assertion failed: "lowestOffset () <= target_offset"

eduard-bagdasaryan:
This assertion could be triggered by various swapout failures for
ufs/aufs/diskd cache_dir entries.

The bug was caused by 4310f8b change related to storeSwapOutFileClosed()
method. Before that change, swapout failures resulted in
StoreEntry::swap_status set to SWAPOUT_NONE, preventing
another/asserting iteration of StoreEntry::swapOut().

This fix adds SWAPOUT_FAILED swap status for marking swapout failures
(instead of reviving and abusing SWAPOUT_NONE), making the code more
reliable.

Also removed storeSwapOutFileNotify() implementation.  We should not
waste time on maintaining an unused method that now contains conflicting
assertions: swappingOut() and !hasDisk().

### DIFF
--- a/src/Store.h
+++ b/src/Store.h
@@ -119,6 +119,8 @@ public:
     bool swappingOut() const { return swap_status == SWAPOUT_WRITING; }
     /// whether the entire entry is now on disk (possibly marked for deletion)
     bool swappedOut() const { return swap_status == SWAPOUT_DONE; }
+    /// whether we failed to write this entry to disk
+    bool swapoutFailed() const { return swap_status == SWAPOUT_FAILED; }
     void swapOutFileClose(int how);
     const char *url() const;
     /// Satisfies cachability requirements shared among disk and RAM caches.
--- a/src/enums.h
+++ b/src/enums.h
@@ -57,7 +57,11 @@ typedef enum {
     SWAPOUT_WRITING,
     /// StoreEntry is associated with a complete (i.e., fully swapped out) disk store entry.
     /// Guarantees the disk store entry existence.
-    SWAPOUT_DONE
+    SWAPOUT_DONE,
+    /// StoreEntry is associated with an unusable disk store entry.
+    /// Swapout attempt has failed. The entry should be marked for eventual deletion.
+    /// Guarantees the disk store entry existence.
+    SWAPOUT_FAILED
 } swap_status_t;
 
 typedef enum {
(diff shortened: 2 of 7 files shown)


eduard-bagdasaryan:
This is a v4 port of master 02ba66.

eduard-bagdasaryan:
This PR failed to proceed with Anubis because it requires that "PR approval" check was "required" for v4 (as for master). It is currently "optional". @yadij, do you know why it was configured differently? Should we make this check "required" in GitHub v4 settings?

yadij:
I'm not sure I follow you. This PR was not reviewed nor marked as M-cleared-for-merge by any of us. So should not have been a consideration for merge yet anyway.

eduard-bagdasaryan:
Each branch has several checks configured as "required" on "Branch protection rules" GitHub page. Our master's "PR approval" check is required, whereas our "v4.0" "PR approval" is not. This difference troubles Anubis: it expects this check to be always "required". That is why it marked this PR with "M-failed-other" label. 
So either:
* "PR approval" check is not configured properly for v4 and we need to switch it to "required". I do not have enough permissions to do this myself.
* "PR approval" check is not "required" for some reason and then we need to fix Anubis to accept such setups.

What do you think?

yadij:
Ah, okay. I don't recall why it was optional, so have updated the settings to be consistent.

rousskov:
When PR approval is required, v4 changes cannot be automatically merged based on maintainer's (i.e. @yadij) vote alone. Compared to manual merge, this creates excessive merging delays. The best solution I can think of is to teach Anubis to treat release branches like v4 specially, but that requires significant design and non-trivial development work, so I cannot promise a quick delivery. Please review my initial design sketch at https://github.com/measurement-factory/anubis/issues/32

eduard-bagdasaryan:
There is one more problem causing Anubis to fail this PR: we need to switch off "Require pull request reviews before merging" checkbox on v4 "Branch protection rule" GitHub page. IIRC this is how master is configured. I temporarily added "WIP" prefix here until it is fixed.

rousskov:
> we need to switch off "Require pull request reviews before merging" checkbox on v4 "Branch protection rule" GitHub page

Done. I think it "off" is the right setting because Anubis tracks votes via the (required) "PR approval" check. GitHub built-in approval rules are not flexible enough to accommodate Squid Project requirements.

PR #395, opened 2019-04-24
Bug 4942: --with-filedescriptors does not do anything

yadij:
SQUID_CHECK_MAXFD has been unconditionally overwriting any
user-defined limit with an auto-detected limit from the build
machine. The change causing this was an incomplete fix for
bug 3970 added to v3.3 and later releases.

Fixing that problem has two notable side effects:

* the user-defined value now has the FD property checks applied
  to it (multiple of 64, too-few, etc). This means warnings will
  start to appear in build logs for a number of custom
  configurations. We should expect an increase in questions
  about that.

* builds which have previously been passing in outrageous values
  will actually start to use those values as the SQUID_MAXFD
  limit. This may result in surprising memory consumption or
  performance issues. Hopefully the warnings and new messages
  displaying auto-detected limit separate from the value used
  will reduce the admin surprise, but may not.

This PR also includes cleanup of the autoconf syntax within the
SQUID_CHECK_MAXFD macro and moves the ./configure warnings about
possible issues into that check macro.

### DIFF
--- a/acinclude/os-deps.m4
+++ b/acinclude/os-deps.m4
@@ -248,9 +248,16 @@ SQUID_STATE_ROLLBACK(maxfd)
 
 AC_MSG_NOTICE([Default number of filedescriptors: $squid_filedescriptors_num])
 
-if test `expr $squid_filedescriptors_num % 64` != 0; then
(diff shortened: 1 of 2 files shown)


yadij:
10 days no vetos, and tester reports it working fine.

PR #397, opened 2019-04-28
Fixed startup assertions when using a src_as or dst_as ACL

eduard-bagdasaryan:
Specify the port for generated whois URLs when requesting AS numbers.

### DIFF
--- a/src/anyp/Uri.cc
+++ b/src/anyp/Uri.cc
@@ -466,6 +466,12 @@ AnyP::Uri::touch()
     authorityWithPort_.clear();
 }
 
+void
+AnyP::Uri::defaultPort()
+{
+    port(getScheme().defaultPort());
+}
+
 SBuf &
 AnyP::Uri::authority(bool requirePort) const
 {
--- a/src/anyp/Uri.cc
+++ b/src/anyp/Uri.cc
@@ -466,12 +466,6 @@ AnyP::Uri::touch()
     authorityWithPort_.clear();
 }
 
-void
-AnyP::Uri::defaultPort()
-{
-    port(getScheme().defaultPort());
-}
-
 SBuf &
 AnyP::Uri::authority(bool requirePort) const
 {
--- a/src/anyp/Uri.h
+++ b/src/anyp/Uri.h
@@ -82,6 +82,8 @@ public:
 
     void port(unsigned short p) {port_=p; touch();}
     unsigned short port() const {return port_;}
+    /// initialize the port with the default port number for this URL's scheme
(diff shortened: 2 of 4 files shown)


yadij:
This does not make much sense. The default port should never need to be set for any scheme.

eduard-bagdasaryan:
> The default port should never need to be set for any scheme.

I would not have set the port value here if Uri::parse() had been called in this context. Without calling parse() method the port value remains '0'. I agree that we should not provide the port number explicitly, since UriScheme::defaultPort() is aware of the mapping between schemes and ports. I fixed this at 96d802.

src/anyp/Uri.h
```diff
@@ -82,6 +82,8 @@ class Uri
 
     void port(unsigned short p) {port_=p; touch();}
     unsigned short port() const {return port_;}
+    /// initialize the port with the default port number for this URL's scheme
```
rousskov:
```suggestion
    /// reset the port to the default port number for the current scheme
```

src/anyp/Uri.cc
```diff
@@ -466,6 +466,12 @@ AnyP::Uri::touch()
     authorityWithPort_.clear();
 }
 
+void
+AnyP::Uri::defaultPort()
+{
+    port(getScheme().defaultPort());
```
rousskov: Please consider inlining this implementation if you have not already.

rousskov:
I also polished the PR description to focus on the primary effect/problem rather than the implementation specifics. Please check that I did not misrepresent anything.

rousskov:
> The default port should never need to be set for any scheme.

@yadij, it sounds like you are thinking about cases where a URL is parsed. In this PR case, the URL object is generated/built/assembled from pre-parsed components instead. Based on Eduard's analysis, the existing URL assembly API does not handle that case well as far as the (default) port is concerned, leading to assertions. The proposed fix makes sense and does not preclude future URL API improvements (which should be welcomed but are outside this particular PR scope).

Another viable in-scope solution would be to immediately set the default URL port whenever the URL scheme is set in the object constructor. That may result in setting the port twice though (because of problems with the old parsing and possibly some URL assembly code). Assembling the port explicitly (as in the current PR) follows the overall "specify each part" style for the URL assembly code, so I do not think it should be banned.

I hope the discussion so far addresses your initial concern. If not, the PR will be on a slow burner for a while if you do want to request changes, of course.

yadij:
LGTM now.

Though if you want to go that far I think the constructor could set the default port to the default when UriScheme is the only parameter.

PR #398, opened 2019-05-02
Supply ALE with HttpReply before checking http_reply_access

eduard-bagdasaryan:
Before this fix, Squid warned "ALE missing HttpReply object",
because ALE::reply was initialized too late. This problem affected
both HTTP and FTP code paths.

To avoid these problems, ALE::reply is initialized early with the
available (received) response.

Also:

* Fixed logging control (1xx) responses in case these responses
  are final. Before this fix, '-' was logged.

* All reply headers ('<h') were logged with a single CR separator,
  instead of CRLF.

* Supply ALE for send_hit ACL. cb36505 already covered many (but not
  all) similar cases.


src/AccessLogEntry.cc
```diff
@@ -70,6 +70,14 @@ AccessLogEntry::syncNotes(HttpRequest *req)
         assert(notes == req->notes());
 }
 
+void
+AccessLogEntry::setReply(HttpReply *rep)
```
yadij: If you want to properly fix the potential locking issues please consider making the reply member an HttpReplyPointer instead of adding this alternative way to set the member. I do not insist on this.

eduard-bagdasaryan:
> How certain are you that every code path where replies are read from servers calls client-side cloneReply() before using access controls?

No, there is no such guarantee in this solution, focused on fixing http_reply_access. I assume that http_reply_access is the first which is checked when passing the adapted reply to the client side, so I think that this solution should cover all access checks performed after adaptation completion, i.e., after getting an adapted reply (which ALE::reply holds). A full solution, including early checks such as adaptation_access, as you noted, would require setting ALE::reply even earlier, storing there a virgin reply.

src/client_side_reply.cc
```diff
@@ -1656,6 +1656,8 @@ clientReplyContext::cloneReply()
 
     /* do header conversions */
     buildReplyHeader();
+
+    http->al->setReply(reply);
```
rousskov: This setReply() argument points to a to-be-sent response rather than the (adapted) received response, right? Is that what we want in ALE? Your new ALE::reply member documentation seems to imply that we want the _received_ (and possibly adapted) response instead. Please check and adjust this code or the new data member description.

src/adaptation/icap/ModXact.cc
```diff
@@ -1338,11 +1338,7 @@ void Adaptation::Icap::ModXact::finalizeLogInfo()
     al.adapted_request = adapted_request_;
     HTTPMSGLOCK(al.adapted_request);
 
-    if (adapted_reply_) {
-        al.reply = adapted_reply_;
-        HTTPMSGLOCK(al.reply);
-    } else
-        al.reply = NULL;
+    al.setReply(adapted_reply_);
```
rousskov: Here, we set ICAP ALE::reply way after the corresponding ICAP transaction is over. AFAICT, this is conceptually wrong because that same ALE may be used a lot earlier for various ACL-driven directives. We should be populating ALE fields as we learn ALE information instead. Do you agree?

If yes, then I do _not_ think this PR should fix this problem (because a lot of other fields are already set in this too-late method, and we should probably fix all or most of them at once), but please add an XXX comment to mark this problem. For example:

```suggestion
    // XXX: This reply (and other ALE members!) may have been needed earlier.
    al.setReply(adapted_reply_);
```

src/AccessLogEntry.cc
eduard-bagdasaryan: Done at c5aaabc.

yadij:
Looks like this now just needs @rousskov review re-check on the ICAP comments.

rousskov:
> Looks like this now just needs @rousskov review re-check on the ICAP comments.

My biggest concern was:

> Is it possible to set ALE::reply when we actually get the reply, rather than rely (AFAICT) on a side effect of doing an access control directive check? I may be missing something, but the current PR code looks fragile -- there is no clearly guaranteed/intentional ALE::reply setting.

@eduard-bagdasaryan, was that concern addressed by the recent branch changes? Or is it an invalid concern?

src/client_side.cc
```diff
@@ -387,6 +387,9 @@ ClientHttpRequest::logRequest()
         al->http.code = al->reply->sline.status();
         al->http.content_type = al->reply->content_type.termedBuf();
     } else if (loggingEntry() && loggingEntry()->mem_obj) {
+        // possibly unreachable code, since al->reply is always
```
rousskov: This comment confuses me: `possibly unreachable` means we are not sure, but `always initialized` appears to imply that we are sure that there is no possibility of reaching this code. I suggest removing the whole comment, but perhaps I missed or forgot the reason for adding it.

src/cf.data.pre
```diff
@@ -4697,13 +4697,16 @@ DOC_START
 	    RESPONSE
 
 		[http::]<Hs	HTTP status code received from the next hop
```
rousskov: If the following is accurate, please use it:

```suggestion
		[http::]<Hs	HTTP status code of the last received virgin reply
```

If it is wrong, please try to find the right definition of this %code using the being-formalized [terminology](https://wiki.squid-cache.org/ResponseTerms).

src/client_side_reply.cc
```diff
@@ -1656,6 +1656,8 @@ clientReplyContext::cloneReply()
 
     /* do header conversions */
     buildReplyHeader();
+
+    http->al->reply = reply;
```
rousskov: Please move this initialization higher, to give all ACLs access to sent reply, including reply_header_add ACLs.

src/AccessLogEntry.h
```diff
@@ -177,7 +177,7 @@ class AccessLogEntry: public RefCountable
     SBuf lastAclData; ///< string for external_acl_type %DATA format code
 
     HierarchyLogEntry hier;
-    HttpReply *reply = nullptr;
+    HttpReplyPointer reply; ///< virgin 1xx reply or a reply a to-be-sent to client
```
rousskov: Similar to another comment in this review, please use the being-formalized terminology. For example:
```suggestion
    HttpReplyPointer reply; ///< the last sent reply
```

src/cf.data.pre
```diff
@@ -4697,13 +4697,16 @@ DOC_START
...
-		[http::]>Hs	HTTP status code sent to the client
+		[http::]>Hs	HTTP status code of the last sent reply.
 
-		[http::]<h	Reply header. Optional header name argument
-				as for >h
+		[http::]<h	The header of the last sent reply. If that last
```
rousskov: I would remove the `If...` sentence. The [terminology](https://wiki.squid-cache.org/ResponseTerms) already covers this nuance (and probably does an overall better job of doing it).

src/cf.data.pre
```diff
@@ -4697,13 +4697,16 @@ DOC_START
...
+				reply was an interim reply (e.g., HTTP 100 Continue),
+				then the interim reply header is logged.
+				Optional header name argument as for >h.
 
-		[http::]mt	MIME content type
+		[http::]mt	MIME content type of the last sent reply.
```
rousskov: Can we be more specific here? What is a `MIME content type` in this context? The contents of the Content-Type response header? Something more complex?

src/clients/Client.cc
```diff
@@ -522,6 +526,7 @@ Client::blockCaching()
         // That method lacks HttpRequest and may be called too many times.
         ACLFilledChecklist ch(acl, originalRequest().getRaw());
         ch.reply = const_cast<HttpReply*>(entry->getReply()); // ACLFilledChecklist API bug
+        ch.al = fwd->al;
```
rousskov: Please do not separate the `ch.reply` assignment above from its HTTPMSGLOCK() below.

src/format/Format.cc
```diff
@@ -357,7 +357,7 @@ sslErrorName(Security::ErrorCode err, char *buf, size_t size)
 static const Http::Message *
 actualReplyHeader(const AccessLogEntry::Pointer &al)
 {
-    const Http::Message *msg = al->reply;
+    const Http::Message *msg = al->reply.getRaw();
```
rousskov: auto?

src/AccessLogEntry.cc
```diff
@@ -131,3 +129,14 @@ AccessLogEntry::effectiveVirginUrl() const
     return nullptr;
 }
 
+const char *
+AccessLogEntry::replyHeaders() const
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

rousskov:
I finally have a specific suggestion for pointing the reader of squid.conf and C++ code descriptions to the new Squid terminology [document](https://wiki.squid-cache.org/ResponseTerms) being created during this PR: I suggest inventing a _single_ _short_ but _unique_ and _suggestive_ combination of letters that we will use as a label/reference/pointer. With time, it would be trivial for a reader to google that label and find the terminology document on Squid wiki, regardless of that wiki page final location. `SeeSqdTrms` is one such combination. With that suggestion, we get:

```C++
    HttpReplyPointer reply; ///< sent reply (SeeSqdTrms)
```

We can also use `CsquidTerms` but its CaseFlow is less consistent and it probably requires more guessing to read correctly. Another version is `InSqdTrms`:

```C++
    HttpReplyPointer reply; ///< sent reply (InSqdTrms)
```

All of the above labels were no found by google (which is my proxy for defining uniqueness) at the time of this message posting. There are a few other variations that satisfy uniqueness, shortness, and suggestiveness criteria, of course. @squid-cache/commiters, if you have a strong preference, please post.

Alternatively, we could replace general English words like `sent reply` with labels like sqtSentReply. It is not as readable, and it may be more difficult to find documentation for a given `sqt*` term, but this alternative reduces confusion where a `sent reply` was not actually _sent_ -- in some cases, it could be a virgin or adapted response received by Squid that did not get it to the Squid sending code, for example.


```C++
    HttpReplyPointer reply; ///< sqtSentReply
```

Please note that this terminology is going to be repeatedly used in many tight contexts where repeatedly providing a wiki URL (instead of a short label) will look ugly. And simply providing a wiki URL in some introduction section is unlikely to work well because that section/URL will be overlooked by most readers, and they would not know that they need to look for some definition after reading that `%<h` is defined as `sent reply`.

eduard-bagdasaryan:
> I finally have a specific suggestion for pointing the reader of squid.conf and C++

As an alternative, we could just refer to Squid wiki (assuming that it would be trivial to do the remaining search within the wiki).

```
HttpReplyPointer reply; ///< sent reply (see wiki terms)
```

To facilitate this search, we would need creating a hierarchical structure of term pages.

rousskov:
> As an alternative, we could just refer to Squid wiki

Let's merge the best of the two ideas into one:

```C++
HttpReplyPointer reply; ///< sent reply (SeeSquidWiki)
```

`SeeSquidWiki` already has 9 hits on Google right now, but I think that is acceptable. If we want zero hits, we can use more cryptic `See_Sqd_Wiki` (the shorter `SeeSqdWiki` version has 2 million hits in an unquoted search).

> (assuming that it would be trivial to do the remaining search within the wiki)

It depends on whether the reader knows that she should search for something like "squid terminology". It would be very difficult to find the definitions while searching for common words like "sent" or "received" on the wiki. We semi-address this problem by merging the two ideas as summarized above -- searching for `SeeSquidWiki` should work very well on wiki. We can (and probably should) even make a wiki page with that name/URL.

> To facilitate this search, we would need creating a hierarchical structure of term pages.

A hierarchy will not help IMO -- "nobody" reads Table Of Contents and similar indexes on wiki pages, and there are too many very different concepts that our wiki has to cover anyway (the top-level index will be long). And once the reader finds the terminology page, it would be easy to find the term she needs there (and a ToC/hierarchy can help at _that_ level if our list of terms grows long).

rousskov:
@eduard-bagdasaryan, if you do not see any objections here soon, please proceed with refactoring PR changes to use the `SeeSquidWiki` label for well-defined terms.

eduard-bagdasaryan:
> proceed with refactoring PR changes to use the SeeSquidWiki label

Done at 5f8055f.

src/client_side.cc
```diff
@@ -427,7 +427,7 @@ ClientHttpRequest::logRequest()
     if (request) {
         SBuf matched;
         for (auto h: Config.notes) {
-            if (h->match(request, al->reply, NULL, matched)) {
+            if (h->match(request, al->reply.getRaw(), NULL, matched)) {
```
yadij: nullptr on touched lines please.

src/AccessLogEntry.h
```diff
@@ -177,7 +179,8 @@ class AccessLogEntry: public RefCountable
     SBuf lastAclData; ///< string for external_acl_type %DATA format code
 
     HierarchyLogEntry hier;
-    HttpReply *reply = nullptr;
+    HttpReplyPointer reply; ///< sent reply (SeeSquidWiki)
```
yadij: All this fussing over the terminology seems a bit unnecessary to me. Some of it appears to be derived from the fact that this _single_ member is not really suitable for top-level ALE placement relative to the logformat needs and the 'h' and 'ha' codes specifically have not yet been updated to use it.

We really have three types of response being juggled;
 1) the "reply from server or peer" which really should be at ALE::HttpDetails::serverReply
 2) the "adapted reply" - partially processed, maybe adapted from (1) but not yet sent as (3).
   - this one maybe could be top-level, but in a member named better than just "reply". or for now stored as (3).
 3) the "reply sent to the client" which really should be at ALE::HttpDetails::clientReply

The "<h" is (1), "<ha" is (3).

eduard-bagdasaryan: > The "<h" is (1), "<ha" is (2) / (3).

To clarify: we don't have '<ha'.
'<h' logs sent replies (as it logged before the PR changes).

yadij: That was part of my point. Much of this mind-bending to re-define the common terms you guys seem to have been doing is working around that. After it gets added there is likely very little remainder to worry about.

eduard-bagdasaryan: > The "<h" is (1), "<ha" is (2) / (3).

Ideally, the reply terminology and log format codes should correlate with their request counterparts. However, '>ha' stands for 'received request after adaptation/redirection' and usually differs from the request, whereas you suggest that '<ha' should keep the sent reply( if available). So, in order to simplify/unify this task, we would need to have three logformat codes for request (virgin/adapted/sent) and three for reply. Currently we have only two for request(virgin/adapted) and one for reply (sent). Do you think we should add the remaining(missing) logformat codes and (if yes) do it in this PR?

rousskov: > Amos: After it gets added there is likely very little remainder to worry about.

Regardless of how many %codes we add or have, we still need to explain what those codes mean. At least some of those %codes (and fields/methods) will require lengthy, non-trivial explanations because their computation method is _naturally_ complex. For example, many admins want to log the "latest reply available", which is a complex notion regardless of how many (other) %codes we have/add.

So yes, there will be quite a bit to "worry about" even after adding more %codes. And, ideally, deciding what %codes to add and thinking about how to describe them should be done at the same time, not in sequence!

Your ad-hoc attempt to suggest what should be done instead is a good illustration of the natural problem complexity -- the suggestion immediately spawned a discussion full of controversial ideas and misunderstandings (all using inadequate terminology). If we want to revise message %codes, a _comprehensive_ solution should be carefully thought out outside a largely unrelated PR comment thread.

> Eduard: Do you think we should add the remaining(missing) logformat codes and (if yes) do it in this PR?

The answer to this question is simple: We should remove documentation changes from this PR if their presence triggers "add more %codes" change requests. In this case, improving documentation simply becomes too much side work. We should cut our losses and just focus on fixing the bug, leaving the existing documentation as useless as it is right now.

@yadij, if you do not like the idea of adding SeeSquidWiki references, we will remove those controversial documentation improvements from this PR. Is that your preference?

Documentation and %codes issues can be revisited in the future, of course.

----

The only viable alternative to so called "redefining of common terms" is to invent uncommon ones.
* If we continue to restrict this PR from adding new %codes, then I do not see any good uncommon terms that we should use instead. Referring to wiki where the current status quo can be documented and where the terminology can be developed/perfected is an imperfect but viable solution here IMO. However, we do not have to change documentation at all if there is no consensus regarding this SeeSquidWiki idea.
* If we add new %codes, then, for _some_ sets of new %codes, those %codes _themselves_ can be used as "uncommon terms". However, doing this right is too much optional work. We will not do it in this PR.

yadij: > Amos: After it gets added there is likely very little remainder to worry about.
> 
> Regardless of how many %codes we add or have, we still need to explain what those codes mean. At least some of those %codes (and fields/methods) will require lengthy, non-trivial explanations because their computation method is naturally complex. For example, many admins want to log the "latest reply available", which is a complex notion regardless of how many (other) %codes we have/add.

Years of experience with the request-path format codes belies those assertions. Request handling has just as much processing applied and the descriptions are still mostly one-liners which are easily read and understood without having to resort to diagrams and pages teaching admin definitions. The exceptions have been where you chose to add new un-common terms.

The major relevant difference between the two right now is that the response code paths are lacking a lot of the cleanup which happened in the request pathways already. Which is what my initial comment was saying, there are things here which are already planned out to change. When it happens that change can be accompanied by a better description. It is a waste of time re-doing the description now when things are still a mess ...

> So yes, there will be quite a bit to "worry about" even after adding more %codes. And, ideally, deciding what %codes to add and thinking about how to describe them should be done at the same time, not in sequence!

By that reason ideally the existing %code should not have the apparently arbitrary description change we see going on here.

The %codes I mentioned in my comment (and other missing %codes too) have already been selected when the existing terms were designed. The label reservations can be found in ByteCode.h as commented out entries symmetrical with request handling codes (though the Token.cc "<ha" part seems to be missing now):
``

    /* response header details pre-adaptation */
    LFT_REPLY_HEADER,
    LFT_REPLY_HEADER_ELEM,
    LFT_REPLY_ALL_HEADERS,

    /* response header details post-adaptation */
    /* LFT_ADAPTED_REPLY_HEADER, */
    /* LFT_ADAPTED_REPLY_HEADER_ELEM, */
    /* LFT_ADAPTED_REPLY_ALL_HEADERS, */
``

All that remains is the code cleanup in the reply handling to provide the pointers to ALE differently, and implementation of the %code's using the resulting ALE members. Sadly the incremental changes leading to that end kept getting stalled in review over not being important enough in small chunks.


> Your ad-hoc attempt to suggest what should be done instead 

Is not ad-hoc at all. It is the result of a) the previous round of discussions the then-core team had on this topic, and b) years of study and months of drawn out discussions about the design for ALE and MasterTransaction contents. The docs terms are well established and have been used by all the logformat additions up until this point. Now I see the two of you without any other consultation beginning a complete re-write to naming things in a way that looks quite at odds with the previous long-term plans.

> If we want to revise message %codes, a comprehensive solution should be carefully thought out outside a largely unrelated PR.

Agreed.

> Eduard: Do you think we should add the remaining(missing) logformat codes and (if yes) do it in this PR?
> 
> The answer to this question is simple: We should remove documentation changes from this PR if their presence triggers "add more %codes" change requests. In this case, improving documentation simply becomes too much side work. We should cut our losses and just focus on fixing the bug, leaving the existing documentation as useless as it is right now.

> @yadij, if you do not like the idea of adding SeeSquidWiki references, we will remove those controversial documentation improvements from this PR. Is that your preference?

It is, yes.

eduard-bagdasaryan: > we will remove those controversial documentation improvements

Done at 7bd1b13.

rousskov: > Years of experience with the request-path format codes belies those assertions.

That is your opinion, not a fact. My years of experience have been different. Glad we may have found a way to move forward despite these differences in opinion. Are you OK with the current documentation changes in this PR? If not, we should remove them (which is what I meant earlier but @eduard-bagdasaryan has removed just the SeeSquidWiki references).

src/client_side.cc
eduard-bagdasaryan: Done at 3318fe2.

src/AccessLogEntry.h
```diff
@@ -177,7 +179,8 @@ class AccessLogEntry: public RefCountable
     SBuf lastAclData; ///< string for external_acl_type %DATA format code
 
     HierarchyLogEntry hier;
-    HttpReply *reply = nullptr;
+    HttpReplyPointer reply; ///< sent reply
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done at 064bf97.

src/cf.data.pre
```diff
@@ -4696,14 +4696,14 @@ DOC_START
 
 	    RESPONSE
 
-		[http::]<Hs	HTTP status code received from the next hop
-		[http::]>Hs	HTTP status code sent to the client
+		[http::]<Hs	HTTP status code of the final virgin reply
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done at 064bf97.

yadij:
Thank you.

src/AccessLogEntry.cc
```diff
@@ -131,3 +129,10 @@ AccessLogEntry::effectiveVirginUrl() const
     return nullptr;
 }
 
+void
+AccessLogEntry::replyHeaders(MemBuf &mb) const
```
rousskov: I think we should change the name because we changed the API to return void.

```suggestion
AccessLogEntry::packReplyHeaders(MemBuf &mb) const
```

src/AccessLogEntry.h
```diff
@@ -61,6 +61,9 @@ class AccessLogEntry: public RefCountable
 
     void syncNotes(HttpRequest *request);
 
+    /// string representation of the reply headers
```
rousskov: Whatever that is, I do not think we are still using "string representation" in this context. If the packing code automatically masks some sensitive headers, then I suggest:

```suggestion
    /// dump all reply headers, masking sensitive headers for logging
```

If the packing code dumps everything as is:

```suggestion
    /// dump all reply headers (for sending or risky logging)
```

Adjust as needed, of course.

src/client_side.cc
```diff
@@ -3665,6 +3665,10 @@ ConnStateData::finishDechunkingRequest(bool withSuccess)
 void
 ConnStateData::sendControlMsg(HttpControlMsg msg)
 {
+    Http::StreamPointer context = pipeline.front();
```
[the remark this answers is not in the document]
rousskov: I believe this is the last unresolved change request from my previous review.

src/client_side_request.cc
```diff
@@ -2018,6 +2018,9 @@ ClientHttpRequest::handleAdaptedHeader(Http::Message *msg)
         storeEntry()->replaceHttpReply(new_rep);
         storeEntry()->timestampsSet();
 
+        if (al)
```
rousskov: AFAICT, ClientHttpRequest constructor unconditionally creates ALE to initialize `al`. If that is true and assuming `al` is never cleared later, please remove this guard. I also see lots of code that dereferences `ClientHttpRequest::al` without checking.

There is also code that does check `ClientHttpRequest::al` for being nil, but we should not blindly duplicate that mistake (if it is a mistake).

src/AccessLogEntry.cc
eduard-bagdasaryan: Done.

src/AccessLogEntry.h
eduard-bagdasaryan: Done.

src/client_side.cc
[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

src/client_side_request.cc
eduard-bagdasaryan: Done.

PR #410, opened 2019-05-25
Bug 4953: to_localhost does not include ::

yadij:
Some OS treat unspecified destination address as an implicit
localhost connection attempt. Add ::/128 alongside the
to_localhost 0.0.0.0/32 address to let admin forbid these
connections when DNS entries wrongly contain [::].

Also, adjust ::1 to ::1/128 to match IPv4 range-based definition
and clarify that IPv6 localhost is /128 rather than /127.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -1003,7 +1003,7 @@ ENDIF
 DEFAULT: all src all
 DEFAULT: manager url_regex -i ^cache_object:// +i ^https?://[^/]+/squid-internal-mgr/
 DEFAULT: localhost src 127.0.0.1/32 ::1
-DEFAULT: to_localhost dst 127.0.0.0/8 0.0.0.0/32 ::1
+DEFAULT: to_localhost dst 127.0.0.0/8 0.0.0.0/32 ::1/128 ::/128
 DEFAULT: CONNECT method CONNECT
 DEFAULT_DOC: ACLs all, manager, localhost, to_localhost, and CONNECT are predefined.
 DOC_START


rousskov:
This is not my area of expertise, but, FWIW, the PR description does not tell me what bug this PR fixes: What are the symptoms? What kind of Squid behavior changes can I expect? The description also does not seem to match PR changes (which do more than add one new masked IP address -- they also add a mask to an old IP address).

Finally, `to prevent when localhost connections are forbidden` sounds very awkward. If possible, please simplify/rephrase to clarify what that phrase means.

yadij:
The referenced bug report should be doing most of the problem description. Updated anyway.

rousskov:
> The referenced bug report should be doing most of the problem description.

IMO, a master commit should have all the essentials, including a clear independently-usable message. There are rare exceptions to that rule of thumb, but I see not reason to make this PR exceptional. 

Thank you for clarifying the description.

rousskov:
@yadij, it looks like you committed this PR manually but you did not use the right description. You also used the wrong commit (or let GitHub create the wrong commit). The right commit was https://github.com/squid-cache/squid/commit/7a4ab323b7acfa1bf0c9128957609928976a6802 AFAICT.

Please do not commit manually unless absolutely necessary. If you think it is necessary to commit manually, please discuss the necessity first, to make sure we are all on one page regarding that necessity. Emergencies and other exceptional situations aside, we should not do manual commits, especially without a discussion.

yadij:
This was the result of GitHub squash+merge button. I think the commit ID changed due to the branch not being based on masters' current HEAD at the time.

As for why the manual kick: the LetsEncrypt update failure caused a series of side effects down into Jenkins and Anubis being told there was no job state change. The whole system needed master to change to invalidate all the broken build results at once and trigger a shift to different tests (on new hashes at Jenkins end). I chose this as the PR with least risk to trigger that.

rousskov:
> I think the commit ID changed due to the branch not being based on masters' current HEAD at the time.

The commit has not "changed" -- GitHub merge button does not know what the right commit is and always creates a brand new untested commit, usually with the wrong commit message. This is one of the reasons to avoid manual mergers.

> the LetsEncrypt update failure caused a series of side effects down into Jenkins and Anubis being told there was no job state change.

That does not compute for me: squid-cache.org site state cannot have a long-term effect on Anubis (if it does, it is probably a bug that we should fix). I know little about Jenkins, but virtually any Jenkins problem can be dealt with without a commit to master -- the official repository is not subservient to Jenkins. There should be no cases where a commit to master is required to fix Jenkins state.

PR #411, opened 2019-05-26
ESI: convert parse exceptions into 500 status response 

yadij:
Produce a valid HTTP 500 status reply and continue operations when
ESI parser throws an exception. This will prevent incomplete ESI
responses reaching clients on server errors. Such responses might
have been cacheable and thus corrupted, albeit corrupted consistently
and at source by the reverse-proxy delivering them.


src/esi/Esi.cc
```diff
@@ -1256,8 +1256,17 @@ ESIContext::parse()
 
         PROF_start(esiParsing);
 
-        while (buffered.getRaw() && !flags.error)
-            parseOneBuffer();
+        try {
+            while (buffered.getRaw() && !flags.error)
+                parseOneBuffer();
+
+        } catch (const TextException &ex) {
```
rousskov: Please merge the two catch clauses into one and use CurrentException to set the error message: See SquidMainSafe() and StoreEntry::checkDisk() for catching examples. And ToSBuf() can be used to convert some error message prefix and CurrentException into SBuf.

src/esi/Expression.cc
```diff
@@ -280,6 +284,7 @@ evalor(stackmember * stack, int *depth, int whereAmI, stackmember * candidate)
 
     srv.precedence = 1;
 
+    Must2(*depth < STACK_DEPTH_LIMIT-1, "expression too complex");
```
rousskov: Please move the `stack` addition code (i.e. this Must2() and the increment/assignment below it) into a new dedicated AddToStack() or some such function to avoid code duplication. Same for all other similar places in the PR, of course.

Ideally, we should refactor expression storage instead, but such refactoring probably goes beyond what you want to do in this PR.

src/esi/Expression.cc
```diff
@@ -215,6 +218,7 @@ evalnegate(stackmember * stack, int *depth, int whereAmI, stackmember * candidat
     /* copy down */
     --(*depth);
 
+    Must2(*depth < STACK_DEPTH_LIMIT-1, "expression too complex");
```
rousskov: This Must() looks odd because we neither _increase_ `depth` nor add stack elements here. We are actually removing one stacked element AFAICT. Please double check.

src/esi/Expression.cc
rousskov: The code below this (and similar) Must2() does no need `*depth` to be smaller than `STACK_DEPTH_LIMIT-1`. Being smaller than `STACK_DEPTH_LIMIT` is enough AFAICT.
```suggestion
    Must2(*depth < STACK_DEPTH_LIMIT, "ESI expression too complex");
```

src/esi/Expression.cc
yadij: Good catch. Thanks. This was supposed to handle the underflow cases.

src/esi/Expression.cc
yadij: The expression code relies on the stack[depth] member being an initialized stackmember object in some places. While it looks like STACK_DEPTH_LIMIT is enough, some test cases hit a segfault in later evaluation code. So I am being a bit cautious for these initial new exceptions.

rousskov: I do not recommend adding exceptions that do not match the seemingly associated code because it makes the reader question both the exception and the code. If some other code needs a different (e.g., more strict) exception, add two exceptions. If we should check a seemingly unrelated condition for some obscure reason, add a comment explaining it. Hopefully, we can do the former and, hence, do not need to do the latter.

src/esi/Esi.cc
yadij: I think you have missed the purpose of these. If we use the unified catch and re-throw like those examples then any malformed ESI response will halt the entire Squid process.

Catching the TextException is done to convert _only_ the errors caught by our parse code into ESI error pages. All other exceptions are more serious and need to be re-thrown for that more fatal handling by SquidMainSafe() itself.

src/esi/Expression.cc
yadij: IMO this is resolved with the addition of stackpush() function requested earlier. That function is clearly documented as ensuring depth value is valid before *and after* the push.

yadij: Indeed. I have added stackpush() to match the stackpop() already present. Replacing the stack with std::stack is left for later.

src/esi/Esi.cc
rousskov: > If we use the unified catch and re-throw ...

I did not suggest doing that. I suggested that we use a "unified catch". See SquidMainSafe() and StoreEntry::checkDisk() for _catching_ examples, not for (re)throwing examples. Rethrowing is what happens after the exception is caught. It is a part of exception handling. ESI exception handling needs are different from those of SquidMainSafe() and StoreEntry::checkDisk().

There should be no rethrow in this ESI context at all AFAICT -- we handle all exceptions by (terminating the parsing loop and) setting the error message. The catch-and-rethrow clause in the current (and reviewed earlier) PR code does nothing -- it should not be there at all (in its current shape). Sorry I did not say that explicitly. I wanted to stay positive.

Rule of thumb: Do not catch exceptions that you do not intend to (at least partially) handle.

src/esi/Expression.cc
rousskov: Yes, this is now resolved, at least partially. If I find specific places in the new code that need further work, I will mark them as such.

src/esi/Esi.cc
```diff
@@ -930,7 +930,7 @@ void
 ESIContext::addStackElement (ESIElement::Pointer element)
 {
     /* Put on the stack to allow skipping of 'invalid' markup */
-    Must(parserState.stackdepth < 10);
+    Must2(parserState.stackdepth < 10, "Too many nested elements");
```
rousskov: I do not insist on this change, but it would be more logical and clear to move this Must() to where we actually add the element to the stack using the being-checked `parserState.stackdepth` index. In its current location, this Must() incorrectly implies that we are protecting the `parserState.top()->addElement()` rather than `parserState.stack[parserState.stackdepth]` expression.

src/esi/Esi.cc
```diff
@@ -1188,7 +1188,7 @@ ESIContext::addLiteral (const char *s, int len)
     assert (len);
     debugs(86, 5, "literal length is " << len);
     /* give a literal to the current element */
-    Must(parserState.stackdepth < 10);
+    Must2(parserState.stackdepth < 10, "Too many nested elements");
```
rousskov: Why check `parserState.stackdepth` that this method code does not use itself? Perhaps this check belongs to those addElement() implementations that actually increment `parserState.stackdepth` and/or use it as an array index?

src/esi/Expression.cc
```diff
@@ -97,6 +100,16 @@ stackpop(stackmember * s, int *depth)
     cleanmember(&s[*depth]);
 }
 
+/// push item onto the stack.
+/// Throws an exception with the given message if depth is not referencing
+/// a valid location of the stack both before and after the push.
+static void
+stackpush(stackmember *stack, stackmember &item, int *depth, const char *message)
+{
+    Must2(*depth >= 0 && *depth < STACK_DEPTH_LIMIT-1, message);
```
rousskov: This check seems too conservative because the `depth` parameter can reach `STACK_DEPTH_LIMIT`, right?

Also the first check does not seem to be related to the `message` being passed to this method.

```suggestion
    assert(*depth >= 0);
    Must2(*depth < STACK_DEPTH_LIMIT, message);
```

src/esi/Expression.cc
```diff
@@ -97,6 +100,16 @@ stackpop(stackmember * s, int *depth)
     cleanmember(&s[*depth]);
 }
 
+/// push item onto the stack.
+/// Throws an exception with the given message if depth is not referencing
+/// a valid location of the stack both before and after the push.
+static void
+stackpush(stackmember *stack, stackmember &item, int *depth, const char *message)
```
rousskov: Overall, it is usually a bad idea to pass context specifics to inner methods just so that they can throw a more detailed error message. Context reporting or annotation, if it is really needed, should be done by the caller that knows what operation it is processing. I doubt we need to be very precise here, but if we do, we can probably store the last operation name in some place where the exception handler can find it.

I do not insist on removing (context from) the `message` parameter but please consider it.

src/esi/Expression.cc
```diff
@@ -215,6 +228,7 @@ evalnegate(stackmember * stack, int *depth, int whereAmI, stackmember * candidat
     /* copy down */
     --(*depth);
 
+    Must2(*depth >= 0 && *depth < STACK_DEPTH_LIMIT && whereAmI >= 0 && whereAmI < STACK_DEPTH_LIMIT, "negate expression too complex");
```
rousskov: The last component of this check feels too late and too lax because we have already accessed `stack[whereAmI + 1]` a few lines above. It also feels unnecessary because `whereAmI` is `*depth - 2` and depth has been validated by the earlier components.

src/esi/Esi.cc
rousskov: Can we avoid duplicating this magic `10`context? Giving it a name may also explain why it is not just `STACK_DEPTH_LIMIT`.

src/esi/Expression.cc
yadij: You _required_ that these *two* lines be extracted to de-duplicate - one of those lines was the caller producing this caller-specific thrown message. Without this parameter the de-duplication was a pointless waste of time and effort.

src/esi/Expression.cc
yadij: Remember that the key issue behind this change is that the depth can be remotely controlled. Both push and pop can be manipulated by the server response.

src/esi/Esi.cc
yadij: The exception string generated by these Must is displayed publicly. That means the caller / location of the Must itself matters for admin debugging how exactly the ESI response is broken.

src/esi/Esi.cc
yadij: Well, it is a different stack. But sure, done.

src/esi/Expression.cc
yadij: Fixed, there are now two Must, one for each parameter checked.

src/esi/Esi.cc
yadij: These TextException thrown by the parser are being placed into error pages displayed to the client. If we catch all exceptions and display the text from system exceptions we will be displaying machine internal address locations and OS internal details in an essentially public output.

src/esi/Expression.cc
yadij: BY that aI meant that we can split this into two MUST, but using assert is not an option.

src/esi/Expression.cc
rousskov: To save time, I am going to ignore your comment and resolve this change request because this change request is not blocking, and your comment appears to misinterpret some requirements to arrive at an absurd conclusion.

src/esi/Esi.cc
rousskov: > If we catch all exceptions and display the text from system exceptions...

There are no "system exceptions" and "not-system exceptions" in Squid exception types. There are also no "sensitive exceptions" and "not-sensitive exceptions" in Squid exception types. This PR does not protect Squid from leaking sensitive info by treating exceptions of type TextExceptions differently from other exception types.

Today, there is no valid reason to treat TextExceptions differently than all other exception types in this context. And there is virtually never a valid reason to add code that does nothing. My change request fixes both problems (without changing Squid security posture).

We can, of course, discuss whether or how Squid exception throwing and/or catching code should protect sensitive info, but that discussion is not specific to this particular code and, most likely, should not happen in this PR.

src/esi/Esi.cc
rousskov: I agree that Must() location matters, for many reasons.

This particular Must() is wrong because it checks something that the surrounding code does not care about. If you are using the wrong location because it relays some useful information to the debugging admin, then, most likely, there is a better way to relay that information. In other words, I am not convinced that the best way to relay ESI response breakage details is to put a `Must(foo)` in a place where `foo` is unused.

src/esi/Expression.cc
rousskov: > we can split this into two

Yes, and we should. Please do that.

> using assert is not an option

Whether the first condition should be an assert() or Must() depends on whether Squid has already checked/ensured that depth does not go negative. If we are confident that Squid has, then this check can and should be an assert(). Otherwise, this check should be a Must(). Assertions are overall better (where they meet their usage preconditions), but Must()s are virtually always acceptable, so if you are not sure about `depth` checks in callers, then you can use a Must() instead of an assert here.

src/esi/Expression.cc
rousskov: > Fixed, there are now two Must

The conditions have changed, but the same rule of thumb applies to the remaining ones: One Must() or assert() per ANDed condition. It may especially be useful here, where one condition is driven by the immediate user input while the other one should probably never fail (or, at the very least, is violated by ESIs processed earlier). For example:

```
    Must2(*depth >= 0, "negate expression too complex");
    Must2(*depth < ESI_STACK_DEPTH_LIMIT, ...); // too big even after we decremented it!
```

src/esi/Esi.cc
yadij: The ESI parsing includes XML parsing sub-calls to libxml2 or libexpat. They include sub-calls to stdlib for memory allocations and other such things which may trigger; for example an alloc failure exception - which contains address locations for memory blocks.
 If we catch such exceptions and transmit in error page texts to remote hosts that is a leak of ASLR information - which is very sensitive and immediate CVE assignment when spotted.

Since those exceptions are coming from third-party code and/or transitory system activity we cannot assume their absence or safety. Only our TextException messages are certain not to have any sensitive information (because I/we have checked for that) - and if they do that is a bug we can fix without loosing the needed info.

rousskov: > Only our TextException messages are certain not to have any sensitive information (because I/we have checked for that)

There are lower-level mistakes in your response, but I will try to stay focused on the big picture in hope to reduce time waste: "I have checked" arguments about specific TextException messages cannot address this change request because it is not about specific exceptions. C++ exception catchers are based on types, not individual exception objects. No mortal, not even you, can "check" what specific exception objects this code will catch tomorrow. And as far as types are concerned, the TextException class, just like std::runtime_error, is not designed to be "not-sensitive" -- TextException objects may contain sensitive information.

There are several valid ways to address this change request, including these two:

1. Treat all exceptions the same. Do not copy their specifics to the response (because those specifics might contain sensitive info and because they are not really meant/designed for sharing with end-users). Report all caught exceptions to cache.log (with a rate or count limit for level-1 reports), where a capable admin can find them.
2. Allow throwers to mark some exceptions as "safe to share with end users" (or something like that). This can be done using a general mechanism (to be reused in other contexts) or something designed specifically for ESI parsing (can carry context-specific but still non-sensitive info). Add marked exceptions to the response. Report all (other?) exceptions to cache.log (with a rate or count limit for level-1 reports), where a capable admin can find them.

I recommend the first option because it is simpler and focuses on the primary problem this PR is meant to solve.

Please let me know if you need my help with fleshing out one of the above options.

yadij:
After looking through PR 460 and how Christos has implemented the pinningError(),  I think the approach so far here has been completely wrong. The conditions producing info intended for the client should maybe throw an ERR_ESI instead of the TextException generated by Must(). Will look into implementing that in the coming days.

rousskov:
> The conditions producing info intended for the client should maybe throw an ERR_ESI instead of the TextException generated by Must().

Throwing ErrorState objects would be an example of option 2 in my earlier [summary](https://github.com/squid-cache/squid/pull/411#discussion_r314986861). And Squid already has committed code throwing non-TextException exceptions for similar purposes. It is, of course, a valid approach.

The key in deciding what to throw is in the assessment of the catcher(s) needs. If a catcher should do X for one group of exceptions and Y for all other exceptions, then throwing something special is warranted for the first group. As you know from my earlier summary, I do not recommend dragging these classification problems (and associated expanded handling) into _this_ specific PR, but I cannot stop you -- both options in my summary are valid.

Please do not forget to properly handle regular exceptions even if you introduce special ones. See my earlier [change request](https://github.com/squid-cache/squid/pull/411#discussion_r287635374) for details.

yadij:
I believe d2ebcefa8450483cec5d2b5729a2e9f75d70aef1 meets all the required changes. Please re-review

yadij:
One thing to note perhapse is that I have gone with SBuf for Esi::ErrorDetail type instead of an ErrorState page because the stack code (correctly) does not have access to enough things to create that object. After we replace the custom stack with an STL container the Esi::Error function can be aligned to the generic design - out of scope for now to reduce patch size.

src/esi/Expression.cc
```diff
@@ -97,6 +98,20 @@ stackpop(stackmember * s, int *depth)
     cleanmember(&s[*depth]);
 }
 
+/// push item onto the stack.
```
rousskov: We can make the function description at least slightly useful by mentioning that the item is copied.

```suggestion
/// copies `item` onto the stack
```

src/esi/Esi.cc
```diff
@@ -930,13 +930,18 @@ void
...
+    if (parserState.stackdepth < 0)
+        throw Esi::Error("ESI elements stack error, probable error in ESI template");
+
     assert (!failed());
     debugs(86, 5, "ESIContext::addStackElement: About to add ESI Node " << element.getRaw());
 
     if (!parserState.top()->addElement(element)) {
-        debugs(86, DBG_IMPORTANT, "ESIContext::addStackElement: failed to add esi node, probable error in ESI template");
-        flags.error = 1;
+        throw Esi::Error("ESIContext::addStackElement failed, probable error in ESI template");
```
rousskov: Ideally, this and the other, similar, post-addElement() `throw` should be moved inside addElement(), making the latter void. I do not insist on this change because it is not trivial -- addElement() has multiple overrides -- and can be considered outside this PR scope.

src/esi/Esi.cc
```diff
@@ -1256,8 +1258,20 @@ ESIContext::parse()
...
+
+        } catch (Esi::ErrorDetail &errMsg) { // FIXME: non-const for c_str()
+            debugs(86, DBG_IMPORTANT, errMsg);
+            setError();
+            setErrorMessage(errMsg.c_str());
+
+        } catch (...) {
+            setError();
+            setErrorMessage("ESI parse, unknown exception");
+            debugs(86, 5, "ESI parse error: " << CurrentException);
```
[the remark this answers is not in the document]
yadij: true, done.

src/esi/Esi.h
```diff
@@ -18,5 +23,14 @@ extern CSD esiStreamDetach;
 extern CSS esiStreamStatus;
 int esiEnableProcessing (HttpReply *);
 
+namespace Esi
+{
+
+typedef SBuf ErrorDetail;
+/// prepare an Esi::ErrorDetail for throw on ESI parser internal errors
+inline Esi::ErrorDetail Error(const char *msg) { return ErrorDetail(msg); }
```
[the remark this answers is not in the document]
yadij: That is to prepare for when the object is not just an SBuf. The page state object has some more complicated setup needed, the Error() function would do all that without changing the callers.

src/esi/Expression.cc
```diff
@@ -97,6 +98,20 @@ stackpop(stackmember * s, int *depth)
...
+/// Throws an exception with the given message if depth is not referencing
```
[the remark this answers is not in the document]
yadij: You requested earlier that those things get documented. Fine by me not to add these comments at all - the function name covers the brief line part.

Removed the docs.

src/esi/Expression.cc
```diff
@@ -97,6 +98,20 @@ stackpop(stackmember * s, int *depth)
...
+/// a valid location of the stack both before and after the push.
+static void
+stackpush(stackmember *stack, stackmember &item, int *depth)
+{
+    if (*depth < 0)
+        throw Esi::Error("ESIExpression stack has negative size");
+    if (*depth >= ESI_STACK_DEPTH_LIMIT-1)
```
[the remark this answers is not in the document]
yadij: if *depth == ESI_STACK_DEPTH_LIMIT then the increment will make depth point to an index exceeding the array bounds. When that last push was for a symbol mid-expression the addmember() for the next item in that expression calls dumpstack() and does an array out-of-bounds access.

src/esi/Esi.cc
```diff
@@ -1256,8 +1258,20 @@ ESIContext::parse()
...
+                parseOneBuffer();
+
+        } catch (Esi::ErrorDetail &errMsg) { // FIXME: non-const for c_str()
+            debugs(86, DBG_IMPORTANT, errMsg);
+            setError();
+            setErrorMessage(errMsg.c_str());
+
+        } catch (...) {
+            setError();
+            setErrorMessage("ESI parse, unknown exception");
```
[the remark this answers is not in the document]
yadij: 'unclassified' just adds more questions, like what the classifications) are and where they are documented. I have dropped the 'unknown' part.

src/esi/Esi.cc
```diff
@@ -1256,8 +1258,20 @@ ESIContext::parse()
...
         PROF_start(esiParsing);
 
-        while (buffered.getRaw() && !flags.error)
-            parseOneBuffer();
+        try {
+            while (buffered.getRaw() && !flags.error)
+                parseOneBuffer();
+
+        } catch (Esi::ErrorDetail &errMsg) { // FIXME: non-const for c_str()
+            debugs(86, DBG_IMPORTANT, errMsg);
```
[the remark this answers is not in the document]
yadij: These are errors the client will be seeing but the admin may be able to fix, or pass on to the web dev to fix. The other calls may need to be made exceptions or raised debug level, but I am leaving that out of scope.

src/esi/Esi.cc
yadij: Those are the reasons I had for not doing it in the multiple implementations. Leaving out of scope.

src/esi/Esi.cc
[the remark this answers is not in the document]
yadij: OK, done.

src/esi/Expression.cc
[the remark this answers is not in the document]
rousskov: > You requested earlier that those things get documented.

I did not.

src/esi/Expression.cc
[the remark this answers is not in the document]
rousskov: > `if (*depth >= ESI_STACK_DEPTH_LIMIT-1)`
> if *depth == ESI_STACK_DEPTH_LIMIT then the increment will ...

I am asking whats wrong with depth being `ESI_STACK_DEPTH_LIMIT-1` before the push. 
Your answer is about depth being `ESI_STACK_DEPTH_LIMIT` before the push.

Your new source code comment and the old PR code imply that depth must not become `ESI_STACK_DEPTH_LIMIT` after the push, but the comment does not explain what is wrong with depth becoming `ESI_STACK_DEPTH_LIMIT` after the push. The comment talks about `depth` "pointing" somewhere, but depth itself does not point. If depth becomes `ESI_STACK_DEPTH_LIMIT` after the push, the last element stored in the stack is still using a valid (last) position in the stack storage AFAICT.

Something is wrong here: Either we do not need to protect the last storage slot (and, hence, the `-1` in the condition and the new comment should be removed), OR we must protect it for some valid but still unknown reason (which requires clarification in comment(s) and/or code).

> the addmember() [...] calls dumpstack() and does an array out-of-bounds access.

In PR code:

* addmember() itself never accesses more than depth array elements
* dumpstack() itself never accesses more than depth array elements

Can you detail the problem? I do not see it in the two functions you mentioned. Does the stack depth become invalid somewhere, somehow? Does some code accesses an array element at `depth` (or beyond)? Something else?

src/esi/Esi.cc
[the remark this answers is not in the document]
rousskov: > 'unclassified' just adds more questions, like what the classifications are and where they are documented

Those are good/valid questions to ask! And a future curious developer can answer them, narrowing down the error cause in the process (or even adding the currently missing documentation). This PR may help with providing a clue via a simple source code comment. The current `parse error` may be vague enough to prevent an average reader from asking good questions, but that is _not_ a good thing.

Nevertheless, I am resolving this minor change request to save time.

src/esi/Esi.cc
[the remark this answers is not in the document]
rousskov: > The other calls may need to [...] raised debug level, but I am leaving that out of scope.

You have no reasons to leave out of scope the level of a debugs() statement that this PR adds. The exception reporting levels are currently inconsistent -- some exceptions are reported at level 1 and some at level 5. All that reporting code was added (or at least seriously altered) in this PR. Please make that new reporting consistent (or document why it should be different).

rousskov:
I found two unresolved change requests: [1](https://github.com/squid-cache/squid/pull/411#discussion_r361699468) and [2](https://github.com/squid-cache/squid/pull/411#discussion_r361702616). 

I also disagreed with the outcome of some of the resolved change requests but those can be ignored.

src/esi/Expression.cc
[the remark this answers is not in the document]
yadij: I have re-run the static analysis and the error has gone away. Maybe an artifact from the Must.

src/esi/Esi.cc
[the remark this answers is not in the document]
yadij: So you would rather have all these errors silenced for the admin when they should be highlighted for fixing ASAP?

What I am leaving out of scope is the expansion of the *other* debugs which are wrongly quiet today.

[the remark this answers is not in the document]
rousskov: > So you would rather have all these errors silenced for the admin when they should be highlighted for fixing ASAP?

I would rather have a consistent approach to error reporting rather than inconsistent new debugs() calls justified by ever-changing hand-waving. In the context of level-0/1 debugging, I am using the corresponding rule of thumb:

Admin should only be informed about important events that admins have control over such as (suspected) serious Squid bugs, misconfigurations, attacks, and similar kid/instance-wide failures or state changes. Individual transaction failures (that do not qualify as important events enumerated earlier) should be reflected in access.log records instead. This rule reduces chances that admins stop paying attention to important log messages because the logs are dominated by irrelevant-to-the-admin noise.

Furthermore, important events (as defined above) that may be logged frequently (e.g., those that may be triggered by incoming data) should have a reporting limit. In other words, Squid should stop reporting them after a while or, better, rate-limit such reports. This secondary rule both prevents DoS attacks via excessive logging and makes sure that important messages do not become noise themselves.

```C++
         try {
             ...
         } catch (Esi::ErrorDetail &errMsg) { // FIXME: non-const for c_str()
            // DBG_IMPORTANT because these are syntax bugs the ogirins' dev can fix
            debugs(86, DBG_IMPORTANT, "ERROR: " << errMsg);
            ...
        } catch (...) {
            // level-5 because these are OS issues not easily fixed
```


Now I am applying these rules of thumb to the current PR code quoted above:

* Transaction-specific ESI syntax errors that the admin usually has no direct control over are reported as DBG_IMPORTANT but probably should not be. IIRC, ESI is enabled by default, even on non-accelerating Squids. If that is true, then the fact that "origin developers" can fix the problem is irrelevant -- there is virtually always somebody somewhere who can fix the problem! What is relevant here is that the admin probably cannot fix this problem, and that the problem has transaction scope.

* Unclassified errors, possibly including Squid code bugs detected via Must() failures, are not reported as DBG_IMPORTANT. It is debatable whether all unclassified problems should be reported just because some of them could be important Squid bugs. However, if we do decide to report transaction-specific ESI syntax errors (the first catch clause), then we should be eager to report unclassified errors that may have much wider scope/impact and may be under admin control.

Going forward, I see three ways to address the current problems:

1. Report classified problems at level 2 (or higher) because they are transaction-level problems outside admin control. Report unclassified problems at `DBG_IMPORTANT` because some of them can be Squid bugs. Add a hard-coded limit to `DBG_IMPORTANT`reports.
2. Report both classified and unclassified problems at level 2 (or higher). Let origin admins deal with their ESI problems and assume that all unclassified errors are benign transaction-specific problems. To use this option, we need to convince ourselves that unclassified errors are (usually) unimportant. You may already know enough about ESI code to give examples that will confirm that.
3. Report both classified and unclassified problems at `DBG_IMPORTANT`. Add a hard-coded limit(s) to those reports. To use this option, we need to convince ourselves that the classified errors are important events under admin control. I am not convinced they are.

I suspect option 1 is the best way forward but I am happy to evaluate arguments supporting option 2 and even option 3.

P.S. And if the `ogirins' dev can fix` comment survives these changes, please spellcheck it. And the `OS issues` phrasing would have to be adjusted -- we cannot know what all of those exceptions are and should not pretend otherwise. What "OS issues" did you have in mind?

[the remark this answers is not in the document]
yadij: We seem to be disagreeing on the "outside admin control" aspect here. 
* All the client receiving the errors can do is contact this admin.
* The site dev can only see what the client sees. Which is a somewhat obfuscated view of the error.
* The admin is the person with most control in these situations - they have access to the proxy debug, and usually a CDN contract agreement + contact details to pass that info to the dev who can implement the fix.

I agree with option 1, making the protocol WARNING and ERROR level-2 messages to display alongside the protocol I/O and slightly higher than the (level 3+) parse activity. If we agree on that as a general policy for all protocols I will update the *touched* lines here.  I still do not want to bloat this PR with more scope creep just for polishing [this *is* a security related PR after all]. Will do a followup PR that brings the other error/warning/exception handing inline with policy and update the debugs() documentation.

yadij:
@rousskov as far as I can see this is still waiting on your response to that final review thread. Not sure why you bumped it back to author.

Shall I make the change to make *only* touched lines meet the new policy (cleanup PR later for everywhere else) or we continue discussion about why ESI is special  ?

rousskov:
> Not sure why you bumped it back to author.

AFAICT, I bumped it after responding on that thread. 

I did not realize you have responded since then. Give me a few hours to re-review.

src/esi/Esi.cc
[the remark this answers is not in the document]
rousskov: I am reordering your response in hope to make progress faster:

> I agree with option 1, 

Great! Please let me know if you want me to implement it to minimize misunderstandings regarding what "option 1" means. I am glad that you liked that option, and would certainly welcome your matching implementation, but I am concerned that your option 1 summary (i.e. `making the protocol WARNING and ERROR level-2 messages`) may not match how I would interpret that option.

> making the protocol WARNING and ERROR level-2 messages

If "protocol messages" are those corresponding to the first `catch` clause in the PR, then (they are what I called "classified problems" in my option1 description, and) we are in agreement here. Level 2 is an acceptable debugging level for them.

However, please note that my option1 also included the second part: Report unclassified problems (i.e. the second `catch` clause) at `DBG_IMPORTANT`. If you are OK with that, than we are in a complete agreement!

> If we agree on that as a general policy for all protocols I will update the _touched_ lines here.

Changing the PR-touched lines is all I am asking for.

> I still do not want to bloat this PR with more scope creep just for polishing

I am not asking for changing any debugs() lines other than those already touched/moved/added by this PR.

Again, I can make those option1 changes if you want me to.

The rest is about future work and further analysis...

----

> Will do a followup PR that brings the other error/warning/exception handing inline with policy and update the debugs() documentation.

Thank you. Before you dive into that development, let's make sure we do agree what those changes should look like. While I think my proposal was rather specific, your seemingly abstract reaction to it does not give me confidence that we are on the same page.

* If you do agree with the two-paragraph rule of thumb that I proposed in my previous [comment](https://github.com/squid-cache/squid/pull/411#discussion_r362064319) and polished below, then perhaps we can start with a single small adjustment or two?
* Otherwise, let's discuss how we can adjust those rules to reach consensus (before making many changes outside this PR?).


----

> We seem to be disagreeing on the "outside admin control" aspect here.

I think the disagreement may be more along the definition of "control". You may be talking "control" as in "can report the problem details to the site admin". I meant "control" as in "views the problem as something he or she should be involved with _and_ can adjust squid.conf or change Squid version to fix the problem". In retrospect, "control over a problem" was a poor expression of the idea.

Here is an adjusted version of the rules of thumb proposed in my previous [comment](https://github.com/squid-cache/squid/pull/411#discussion_r362064319):

When reporting various errors, bugs, and problems:

1. Level-0/1 messages should be used for important events affecting the Squid instance as a whole (e.g., (suspected) serious Squid bugs, misconfigurations, and attacks).
2. Individual transaction failures (that do not qualify as important events in item 1) should be reflected in access.log records.
3. Level-0/1 messages that may be logged frequently (e.g., those that may be triggered by incoming data) should have a reporting limit. In other words, Squid should stop reporting them after a while or, better, rate-limit such reports.

This rule of thumb reduces chances that admins stop paying attention to important cache.log messages because the cache.log is dominated by irrelevant-to-the-admin noise and/or relevant-but-overwhelming info. The rule also prevents DoS attacks via excessive cache.log logging.

N.B. Folks may never agree on whether some particular debugs() line qualifies as item1 problem -- there are too many vague terms in item1 definition. The only solution to _that_ problem (that I know of) is making level-0/1 reporting configurable at individual message level. Factory has a patch for that; we are polishing it for the official submission.

rousskov:
> Give me a few hours to re-review.

[Done](https://github.com/squid-cache/squid/pull/411#discussion_r382656921).

src/esi/Esi.cc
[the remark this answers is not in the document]
yadij: I believe option1 is now implemented as of commit ca3a52c. If there are any other changes you were expecting to see please do implement as a sub-branch so we can compare.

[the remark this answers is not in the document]
rousskov: > I believe option1 is now implemented as of commit ca3a52c.

Thank you. The ball is on my side.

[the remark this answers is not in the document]
rousskov: I agree that option 1 is now implemented. Resolving this change request.

src/esi/Esi.cc
```diff
@@ -1256,8 +1259,24 @@ ESIContext::parse()
...
+
+        } catch (Esi::ErrorDetail &errMsg) { // FIXME: non-const for c_str()
+            // level-2: these are protocol/syntax errors from upstream
+            debugs(86, 2, "WARNING: ESI syntax error: " << errMsg);
+            setError();
+            setErrorMessage(errMsg.c_str());
+
+        } catch (...) {
+            // DBG_IMPORTANT because these are local issues the admin needs to fix
+            static FadingCounter logEntries(); // TODO: set horizon less than infinity
```
rousskov: This overly specific construction probably causes this build [error](https://semaphoreci.com/squid/squid/branches/pull-request-411/builds/18):

```
src/esi/Esi.cc:1275:28: error: request for member ‘count’ in ‘logEntries’, which is of non-class type ‘FadingCounter()’
             if (logEntries.count(1) < 100)
```

I suggest to KISS:

```suggestion
            static FadingCounter logEntries; // TODO: set horizon less than infinity
```

The whole FadingCounter API should be adjusted if we are going to use it for limiting debugs() statements, but those adjustments are outside this PR scope. If you prefer to replace this FadingCounter with a basic `unsigned int` or `size_t` counter, please feel free to do so without triggering another review!

yadij: Agreed on the need for adjustments to the fader class. The primary one is as noted in the TODO.

FTR; my reasoning is this:
 If we use a simple counter and stop reporting past a threshold we could hit problems with one error happening N times filling the counter, then silence some other errors. That can still happen with the fader but self-corrects as the fading happens, allowing the suppressed problems to be reported at some point.
 Making a more performant fader (if we can) would be good not just in debugs() case but for the existing users of the FadingCounter class. So TODO on a redesign is better than a second implementation or a bad counter.

src/esi/Esi.cc
yadij: resolved with the the exception model

src/esi/Esi.cc
yadij: resolved with the the exception model

src/esi/Expression.cc
yadij: resolved with the the exception model

PR #414, opened 2019-06-04
ICAP stalls large HTTP transactions

eduard-bagdasaryan:
Calling virginBodySending.plan() has nothing to do with ICAP trailers
support. These (expensive!) plans are necessary if and only if we may
echo the virgin message back to Squid.

Since 6fbd6e, ICAP code always planned to echo the virgin body, even
when no echoing was needed. Echoing requires buffering the virgin
message body, so these plans stalled responses with bodies exceeding
~65K buffer capacity when, for example, the ICAP server did not respond
until getting the entire request body.

### DIFF
--- a/src/adaptation/icap/ModXact.cc
+++ b/src/adaptation/icap/ModXact.cc
@@ -1542,11 +1542,8 @@ void Adaptation::Icap::ModXact::makeAllowHeader(MemBuf &buf)
     const bool allow204 = allow204in || allow204out;
     const bool allow206 = allow206in || allow206out;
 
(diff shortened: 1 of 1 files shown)


rousskov:
To clarify: I removed "enabled ICAP trailers" from the title because while the bug was related to a condition from the ICAP trailers code, it is not really possible to "enable" ICAP trailers (they are always supported), and, at the end, trailers support as such was not the (direct) cause of the problem AFAICT.

eduard-bagdasaryan:
I am ok with the updated PR description.

PR #416, opened 2019-06-08
Update HttpHeader::getAuth to SBuf

yadij:
Replace the fixed-size buffer for decoding base64 tokens with an
SBuf to avoid decoder issues on large inputs.

Update callers to SBuf API operations for more efficient memory
management.


src/HttpHeader.cc
```diff
@@ -1289,43 +1289,46 @@ HttpHeader::getContRange() const
...
     for (; field && xisspace(*field); ++field);
 
     if (!*field)        /* no authorization cookie */
-        return NULL;
+        return nil;
 
-    static char decodedAuthToken[8192];
+    const auto fieldLen = strlen(field);
+    SBuf result;
+    char *decodedAuthToken = result.rawAppendStart(BASE64_DECODE_LENGTH(fieldLen)+1);
```
rousskov: We are no longer using c-strings here so there is no need to reserve a place for the terminator. I have not checked, but even if the calling code is converting our SBuf `result` into a c-string, we should not reserve extra space to accommodate that conversion because that conversion is temporary and is not a part of our API.

```suggestion
    char *decodedAuthToken = result.rawAppendStart(BASE64_DECODE_LENGTH(fieldLen));
```

src/cache_manager.cc
```diff
@@ -244,20 +245,20 @@ CacheManager::ParseHeaders(const HttpRequest * request, Mgr::ActionParams &param
     // TODO: use the authentication system decode to retrieve these details properly.
 
     /* base 64 _decoded_ user:passwd pair */
-    const char *basic_cookie = request->header.getAuth(Http::HdrType::AUTHORIZATION, "Basic");
+    const SBuf basic_cookie(request->header.getAuthToken(Http::HdrType::AUTHORIZATION, "Basic"));
```
rousskov: If possible, please use `auto` here and in a similar place elsewhere in the PR.

```suggestion
    const auto basic_cookie(request->header.getAuthToken(Http::HdrType::AUTHORIZATION, "Basic"));
```

yadij: Done

PR #418, opened 2019-06-14
Add the NO_TLSv1_3 option to available  tls-options values

chtsanti:
... also fix the deprecated sslversion option to exclude tls v1.3 from
allowed protocols where required.

### DIFF
--- a/src/security/PeerOptions.cc
+++ b/src/security/PeerOptions.cc
@@ -221,13 +221,6 @@ Security::PeerOptions::updateTlsVersionLimits()
             add = ":-VERS-TLS1.0:-VERS-TLS1.1:-VERS-TLS1.3";
 #endif
             break;
-        case 7:
-#if USE_OPENSSL
-            add = ":NO_SSLv3:NO_TLSv1:NO_TLSv1_1:NO_TLSv1_2";
-#elif USE_GNUTLS
-            add = ":-VERS-TLS1.0:-VERS-TLS1.1:-VERS-TLS1.2";
-#endif
-            break;
         default: // nothing
             break;
         }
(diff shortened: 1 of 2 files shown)


src/security/PeerOptions.cc
```diff
@@ -195,30 +195,37 @@ Security::PeerOptions::updateTlsVersionLimits()
...
         case 6:
 #if USE_OPENSSL
-            add = ":NO_SSLv3:NO_TLSv1:NO_TLSv1_1";
+            add = ":NO_SSLv3:NO_TLSv1:NO_TLSv1_1:NO_TLSv1_3";
 #elif USE_GNUTLS
-            add = ":-VERS-TLS1.0:-VERS-TLS1.1";
+            add = ":-VERS-TLS1.0:-VERS-TLS1.1:-VERS-TLS1.3";
+#endif
+            break;
+        case 7:
```
yadij: Please do not add new cases here. The purpose of back-compat code is purely to map the old options values to the new option implementation. Adding new undocumented behaviours is not a good idea.

src/security/PeerOptions.cc
```diff
@@ -401,6 +408,11 @@ static struct ssl_option {
 #else
     { "NO_TLSv1_2", 0 },
 #endif
+#if SSL_OP_NO_TLSv1_3
+    {
+        "NO_TLSv1_3", SSL_OP_NO_TLSv1_3
+    },
+#endif
```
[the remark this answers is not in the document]
chtsanti: Only two of the defined options uses this #else case, and the code which uses the ssl_options does not case about.

The commit f0b01376 adds this else case.

src/security/PeerOptions.cc
chtsanti: Removed with the commit f0b01376

chtsanti:
@yadij 
commit f0b01376 should handle your concerns, please re-review.

PR #420, opened 2019-06-21
Bug 4966: Lower cache_peer hostname

uhliarik:
When parsing entries from /etc/hosts file, they are all lowered
(see bug 3040). If cache_peer hostname is uppercase, it will
lead to DNS resolution failure. Lowering cache_peer host fixes
this issue.

This change may expose broken Squid configurations that
incorrectly relied on non-lowercase peer host names to
bypass Squid's "is this cache_peer different from me?"
check. Though such configurations should encounter
forwarding loop errors later anyway.

### DIFF
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -2139,6 +2139,7 @@ parse_peer(CachePeer ** head)
 
     CachePeer *p = new CachePeer;
     p->host = xstrdup(host_str);
+    Tolower(p->host);
     p->name = xstrdup(host_str);
     p->type = parseNeighborType(token);
 


uhliarik:
https://bugs.squid-cache.org/show_bug.cgi?id=4966

yadij:
OK to test

rousskov:
It would be good to document, in the PR description, that this change may expose broken Squid configurations that incorrectly relied on non-lowercase peer host names to bypass Squid's "is this cache_peer different from me?" check:

```
neighbors_init(void)
{
    const char *me = getMyHostname();
    ...
            if (0 != strcmp(thisPeer->host, me))
                continue;
            ...
                debugs(15, DBG_IMPORTANT, "WARNING: Peer looks like this host");
                ...
                neighborRemove(thisPeer);
```

yadij:
FTR: updated the description as requested and clearing this to go in as-is. I will take on the API redesign in a separate cleanup PR.

PR #421, opened 2019-06-21
Send Connection:close with the known-last request on a connection

eduard-bagdasaryan:
Squid did not send Connection:close request header on a to-be-closed
HTTP/1.1 connection, violating RFC 7230 Section 6.6 "SHOULD send"
requirement.

### DIFF
--- a/src/http.cc
+++ b/src/http.cc
@@ -1926,10 +1926,10 @@ HttpStateData::httpBuildRequestHeader(HttpRequest * request,
         delete cc;
     }
 
-    /* maybe append Connection: keep-alive */
-    if (flags.keepalive) {
-        hdr_out->putStr(Http::HdrType::CONNECTION, "keep-alive");
-    }
+    // Always send Connection because older servers need explicit "keep-alive"
+    // while modern servers need explicit "close", and we do not always know
+    // the server expectations.
+    hdr_out->putStr(Http::HdrType::CONNECTION, flags.keepalive ? "keep-alive" : "close");
 
     /* append Front-End-Https */
     if (flags.front_end_https) {
--- a/src/http.cc
+++ b/src/http.cc
@@ -1926,8 +1926,8 @@ HttpStateData::httpBuildRequestHeader(HttpRequest * request,
         delete cc;
     }
 
-    // Always send Connection because older servers need explicit "keep-alive"
-    // while modern servers need explicit "close", and we do not always know
+    // Always send Connection because HTTP/1.0 servers need explicit "keep-alive"
+    // while HTTP/1.1 servers need explicit "close", and we do not always know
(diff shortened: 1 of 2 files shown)


src/http.cc
```diff
@@ -1926,10 +1926,10 @@ HttpStateData::httpBuildRequestHeader(HttpRequest * request,
         delete cc;
     }
 
-    /* maybe append Connection: keep-alive */
-    if (flags.keepalive) {
-        hdr_out->putStr(Http::HdrType::CONNECTION, "keep-alive");
-    }
+    // Always send Connection because older servers need explicit "keep-alive"
```
yadij:
```suggestion
    // Always send Connection because HTTP/1.0 servers need explicit "keep-alive"
```

src/http.cc
```diff
@@ -1926,10 +1926,10 @@ HttpStateData::httpBuildRequestHeader(HttpRequest * request,
...
+    // while modern servers need explicit "close", and we do not always know
```
yadij:
```suggestion
    // while HTTP/1.1 servers need explicit "close", and we do not always know
```

src/http/StateFlags.h
```diff
@@ -16,7 +16,7 @@ class StateFlags
 {
 public:
     unsigned int front_end_https = 0; ///< send "Front-End-Https: On" header (off/on/auto=2)
-    bool keepalive = false;
+    bool keepalive = false; ///< whether to keep the connection persistent
```
yadij: Is this the right default still? false is the default for HTTP/1.0 semantics, while implicit true would match HTTP/1.1 requirements. I suspect this would need a lot more code change so do not insist on it, but *if* we know for sure that this needs to change now would be a good time to add a TODO.

eduard-bagdasaryan: There is a relatively complex/unobvious logic of flags.keepalive initialization in HttpStateData::sendRequest(). However, it is clear that 'keepalive' cannot have a single default value(covering all cases). For example, for (peer && peer->stats.n_keepalives_sent >= 10) precondition the default value should be 'false', but for (peer==null) precondition it should be 'true'. I think that the proper requirement should be that  "flags.keepalive must be initialized before its first usage". But we cannot do that by means of a boolean flag.

yadij: Thank you.

rousskov: > I think that the proper requirement should be that "flags.keepalive must be initialized before its first usage". But we cannot do that by means of a boolean flag.

Exactly! _And_ a `false` value is a "safer" default because the wrong `true` default may stall some transactions (which is worse than performance implications of the wrong `false` default).

eduard-bagdasaryan:
The wording looks correct.

PR #422, opened 2019-06-21
Fix handling of tiny invalid responses

eduard-bagdasaryan:
Squid converted any invalid response shorter than 4 bytes into an
invalid "HTTP/1.1 0 Init" response (with those received characters and a
CRLFCRLF suffix as a body). In some cases (e.g., with ICAP RESPMOD), the
resulting body was not sent to the client at all.

Now Squid handles such responses the same way it handles any non-HTTP/1
(and non-ICY) response, converting it into a valid HTTP/200 response
with an X-Transformed-From:HTTP/0.9 header and received bytes as
a message body.


rousskov:
@yadij, do you plan to commit this fix to v4? If that is your plan, then what is the current ETA and do you need help with porting or testing these changes to/in v4?

PR #423, opened 2019-06-24
Prevent TLS transaction stalls by preserving flags.read_pending

eduard-bagdasaryan:
DoSelect() cleared flags.read_pending just like it clears the read
handler, but that was wrong: Unlike the read handler, the "pending read"
state is not managed by the read _handler_. It is managed by the reading
_method_. DoSelect() calls read handler so it has the right to clear the
handler before that call -- the called handler is given an opportunity
to set another read handler as needed. DoSelect() does not call the
reading method.

In most cases, the reading method is called by the reading handler so
the code appears to "work". However, if the reading handler has to defer
the read, it will not call the reading method. Deferred reads are
naturally ignorant about the high-level read_pending flag so they cannot
(and should not) attempt to preserve that flag. The flag and the
associated "I have buffered data to give to the reader" state get lost.
When the deferred read is kicked, it starts waiting for socket I/O.

The loss of the flag usually leads to (minor) response time increases
because the peer socket will have more data available for reading,
triggering a call to the read handler which (this time) calls the
reading method which supplies the previously buffered data. Thus, the
code usually appears to "work" even when the bug is triggered.

However, when the buffered data contained the _last_ response byte, the
peer socket often remains idle (not ready for reading), the transaction
gets stalled and eventually timeouts/aborts without delivering the last
byte(s) to the client. Such bugs have been observed in the wild.

This change delegates read_pending management to the reading method.
When the method is not called (e.g., because reading is deferred), the
flag remains set and continues to reflect the current reading state.

When the reading method is "downgraded" from tls_read_method() to
default_read_method(), the new code clears flags.read_pending to avoid
endless "this FD is ready for reading" looping. The downgrade itself is
a dangerous operation because it may result in the loss of the already
buffered data. I hope Squid does not lose data when downgrading (and
added comments to justify that hope), but I cannot guarantee that. This
change itself does not introduce data losses.

I also made fde I/O methods private to improve our chances of preventing
a similar bug if future code changes carelessly reset them.

### DIFF
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -3058,11 +3058,10 @@ ConnStateData::splice()
 {
     // normally we can splice here, because we just got client hello message
 
-    if (fd_table[clientConnection->fd].ssl.get()) {
-        // Restore default read methods
-        fd_table[clientConnection->fd].read_method = &default_read_method;
-        fd_table[clientConnection->fd].write_method = &default_write_method;
-    }
+    // fde::ssl/tls_read_method() probably reads from our own inBuf. If so, then
+    // we should not lose any raw bytes when switching to raw I/O here.
+    if (fd_table[clientConnection->fd].ssl.get())
+        fd_table[clientConnection->fd].useDefaultIo();
 
     // XXX: assuming that there was an HTTP/1.1 CONNECT to begin with...
     // reset the current protocol to HTTP/1.1 (was "HTTPS" for the bumping process)
--- a/src/fde.cc
+++ b/src/fde.cc
@@ -6,10 +6,11 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-/* DEBUG: none          FDE */
+/* DEBUG: section 05    Comm */
 
 #include "squid.h"
 #include "comm/Read.h"
+#include "Debug.h"
 #include "fd.h"
 #include "fde.h"
 #include "globals.h"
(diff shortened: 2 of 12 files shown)


eduard-bagdasaryan:
Note that this patch was developed by @rousskov , I have merely ported it to master.

rousskov:
I changed the PR title so that admins/maintainers have a better chance of understanding the change impact.

src/client_side.cc
```diff
@@ -3058,11 +3058,10 @@ ConnStateData::splice()
 {
     // normally we can splice here, because we just got client hello message
 
-    if (fd_table[clientConnection->fd].ssl.get()) {
-        // Restore default read methods
-        fd_table[clientConnection->fd].read_method = &default_read_method;
-        fd_table[clientConnection->fd].write_method = &default_write_method;
-    }
+    // fde::ssl/tls_read_method() probably reads from our own inBuf. If so, then
```
yadij: The uncertainty here worries me a bit. Please at least add a TODO for someone else to follow up on which states what needs to be done to get surety.

src/fde.cc
```diff
@@ -6,10 +6,11 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-/* DEBUG: none          FDE */
+/* DEBUG: section 05    Comm */
```
yadij: Please leave the "FDE" label, this is not comm.cc or libcomm even though related. We can have multiple labels and it helps document the variety of things the section covers.

src/client_side.cc
rousskov: I share your worries, but, FWIW, I cannot formulate a useful TODO here. IMO, the current comment (including its intended uncertainty) reflects the current (shallow) understanding of this code well. Anybody with enough expertise to resolve this uncertainty ought to be able to do so without an explicit TODO.

PR #425, opened 2019-06-27
Rename allow_t to Acl::Answer

josepjones:
Enact the TODO item asking to rename class to Acl::Answer

### DIFF
--- a/src/acl/Acl.h
+++ b/src/acl/Acl.h
@@ -23,6 +23,8 @@ class ConfigParser;
 
 namespace Acl {
 
+class Answer;
+
 /// the ACL type name known to admins
 typedef const char *TypeName;
 /// a "factory" function for making ACL objects (of some ACL child type)
@@ -109,14 +111,14 @@ typedef enum {
 } aclMatchCode;
 
 /// \ingroup ACLAPI
-/// ACL check answer; TODO: Rename to Acl::Answer
-class allow_t
+/// ACL check answer
+class Acl::Answer
 {
 public:
-    // not explicit: allow "aclMatchCode to allow_t" conversions (for now)
-    allow_t(const aclMatchCode aCode, int aKind = 0): code(aCode), kind(aKind) {}
+    // not explicit: allow "aclMatchCode to Acl::Answer" conversions (for now)
+    Answer(const aclMatchCode aCode, int aKind = 0): code(aCode), kind(aKind) {}
 
-    allow_t(): code(ACCESS_DUNNO), kind(0) {}
+    Answer(): code(ACCESS_DUNNO), kind(0) {}
 
     bool operator ==(const aclMatchCode aCode) const {
         return code == aCode;
@@ -126,7 +128,7 @@ public:
         return !(*this == aCode);
     }
 
-    bool operator ==(const allow_t allow) const {
+    bool operator ==(const Answer allow) const {
         return code == allow.code && kind == allow.kind;
     }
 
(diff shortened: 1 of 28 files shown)


src/acl/Checklist.h
```diff
@@ -228,7 +228,7 @@ class ACLChecklist
     bool asyncCaller_; ///< whether the caller supports async/slow ACLs
     bool occupied_; ///< whether a check (fast or non-blocking) is in progress
     bool finished_;
-    allow_t allow_;
+    Acl::Answer allow_;
```
josepjones: I wasn't sure, but maybe the variable name should also be renamed?

maybe something like this?
```
Acl::Answer answer;
```

yadij: Yes please do.

src/acl/Acl.h
```diff
@@ -109,14 +111,14 @@ typedef enum {
 } aclMatchCode;
 
 /// \ingroup ACLAPI
-/// ACL check answer; TODO: Rename to Acl::Answer
-class allow_t
+/// ACL check answer
+class Acl::Answer
```
yadij: Squid code style is to use an explicit namespace wrapper around the entire code block, as you can see around the code at lines 24-35 above. This helps our automated code processing tools get namespace attributions correct.

src/acl/Acl.h
```diff
@@ -23,6 +23,8 @@ class ConfigParser;
 
 namespace Acl {
 
+class Answer;
```
yadij: This predefine should be (and is already) done by acl/forward.h instead. The other code in here should be too, but out of scope for this PR.

So for here, just remove the 2 new lines.

src/acl/Checklist.cc
```diff
@@ -370,9 +370,9 @@ ACLChecklist::fastCheck()
 void
 ACLChecklist::calcImplicitAnswer()
 {
-    const allow_t lastAction = (accessList && cbdataReferenceValid(accessList)) ?
-                               accessList->lastAction() : allow_t(ACCESS_DUNNO);
-    allow_t implicitRuleAnswer = ACCESS_DUNNO;
+    const Acl::Answer lastAction = (accessList && cbdataReferenceValid(accessList)) ?
```
yadij: Please take advantage of the change to use C++11 "auto" instead of an explicit type name for lastAction.
```suggestion
    const auto lastAction = (accessList && cbdataReferenceValid(accessList)) ?
```

There are quite a few other places which used to have explicit allow_t which can also become auto now instead of Acl::Answer. The rule of thumb to use is that if the RHS of an assignment is a function, method or variable with the same type as the new variable on the LHS then you can use auto instead of repeating its base type name.

The tricky part here is that the ACCESS_* things have a different type so only use auto there if they are assigned using the type conversion wrapper Acl::Answer(ACCESS_FOO).

rousskov: > The rule of thumb to use is that if the RHS of an assignment is a function, method or variable with the same type as the new...

The rule of thumb is actually a lot simpler: If you can use `auto`, use `auto`. Same for `const`. These rules are applied to new and changed-for-other-reasons code. As is typical for rules of thumb, they come with exceptions, but they usually work fine.

src/acl/Checklist.h
rousskov: ... but with an underscore suffix to match the existing naming scheme for private data members: `answer_`.

rousskov:
OK to test

PR #427, opened 2019-06-29
Prevent truncation for large origin-relative domains

yadij:
The domain in a URL must obey hostname length restrictions after
append_domain is applied, not just MAX_URL for the normalized
absolute-URL.

### DIFF
--- a/src/anyp/Uri.cc
+++ b/src/anyp/Uri.cc
@@ -180,8 +180,8 @@ urlAppendDomain(char *host)
     /* For IPv4 addresses check for a dot */
     /* For IPv6 addresses also check for a colon */
     if (Config.appendDomain && !strchr(host, '.') && !strchr(host, ':')) {
-        const int64_t dlen = strlen(host);
-        const int64_t want = dlen + Config.appendDomainLen;
+        const uint64_t dlen = strlen(host);
+        const uint64_t want = dlen + Config.appendDomainLen;
         if (want > SQUIDHOSTNAMELEN - 1) {
             debugs(23, 2, "URL domain too large (" << dlen << " bytes)");
             return false;
--- a/src/anyp/Uri.cc
+++ b/src/anyp/Uri.cc
@@ -167,6 +167,20 @@ urlParseProtocol(const char *b)
     return AnyP::PROTO_NONE;
 }
 
+bool
+urlAppendDomain(char *host)
+{
+    if (Config.appendDomain && !strchr(host, '.') && !strchr(host, ':')) {
+        const auto dlen = strlen(host);
+        if (dlen > (SQUIDHOSTNAMELEN - Config.appendDomainLen - 1)) {
+            debugs(23, 2, "URL domain too large (" << dlen << " bytes)");
+            return false;
+        }
+        strncat(host, Config.appendDomain, SQUIDHOSTNAMELEN - dlen - 1);
+    }
+    return true;
+}
+
 /*
  * Parse a URI/URL.
  *
@@ -377,14 +391,8 @@ AnyP::Uri::parse(const HttpRequestMethod& method, const char *url)
     }
 
     /* For IPV6 addresses also check for a colon */
-    if (Config.appendDomain && strchr(foundHost, '.') == 0 && strchr(foundHost, ':') == 0) {
-        const auto dlen = strlen(foundHost);
-        if (dlen > (SQUIDHOSTNAMELEN - Config.appendDomainLen - 1)) {
-            debugs(23, DBG_IMPORTANT, MYNAME << ": URL domain too large (" << dlen << " bytes)");
-            return false;
-        }
-        strncat(foundHost, Config.appendDomain, SQUIDHOSTNAMELEN - dlen - 1);
-    }
+    if (!urlAppendDomain(foundHost))
+        return false;
 
     /* remove trailing dots from hostnames */
     while ((l = strlen(foundHost)) > 0 && foundHost[--l] == '.')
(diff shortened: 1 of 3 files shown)


src/anyp/Uri.cc
```diff
@@ -167,6 +167,20 @@ urlParseProtocol(const char *b)
     return AnyP::PROTO_NONE;
 }
 
+bool
+urlAppendDomain(char *host)
+{
+    if (Config.appendDomain && !strchr(host, '.') && !strchr(host, ':')) {
+        const auto dlen = strlen(host);
+        if (dlen > (SQUIDHOSTNAMELEN - Config.appendDomainLen - 1)) {
```
rousskov: This expression is both 
* more difficult to read/interpret than necessary and
* potentially dangerous with regard to unsigned values wrapping when (unvalidated) `Config.appendDomainLen` exceeds `SQUIDHOSTNAMELEN`

I hope this version addresses both problems:
```suggestion
        if (dlen + Config.appendDomainLen > SQUIDHOSTNAMELEN - 1) {
```

The `-1` crap is unfortunate, but I am guessing we have to live with it because some code assumes that `SQUIDHOSTNAMELEN` includes a c-string terminator. You can move that `1` to the left side if you prefer. IMHO, it is more logical on the right side because it is derived from SQUIDHOSTNAMELEN _meaning_ (is not dependent on input) and affects the _limit_, which is what the right side should be about.

N.B. We could also protect from `dlen + Config.appendDomainLen` overflows but, unlike underflows, those are just too unlikely to happen in real deployments. Your call though.

src/anyp/Uri.cc
```diff
@@ -377,8 +391,8 @@ AnyP::Uri::parse(const HttpRequestMethod& method, const char *url)
     }
 
     /* For IPV6 addresses also check for a colon */
```
rousskov: This comment should be moved into urlAppendDomain(), right? It seems to be about that `strchr(:)` check there. I think it is a good idea to keep it because it is not obvious why the code checks for a colon.

In fact, it is also not obvious that the code checks for a period to, in part, exclude IPv4 addresses! Your call whether to clarify that with another comment.

src/anyp/Uri.h
```diff
@@ -193,6 +193,7 @@ bool urlIsRelative(const char *);
 char *urlMakeAbsolute(const HttpRequest *, const char *);
 char *urlRInternal(const char *host, unsigned short port, const char *dir, const char *name);
 char *urlInternal(const char *dir, const char *name);
+bool urlAppendDomain(char *host); ///< apply append_domain config to the given hostname
```
rousskov: I think it is important to document buffer size expectations here. It would not hurt to specify the return value meaning, but I do not insist on that.

```suggestion
/// Appends configured append_domain to hostname, assuming
/// the given buffer is at least SQUIDHOSTNAMELEN bytes long.
/// \returns false if and only if there is not enough space to append
bool urlAppendDomain(char *hostname);
```

src/anyp/Uri.cc
```diff
@@ -167,6 +167,20 @@ urlParseProtocol(const char *b)
...
+bool
+urlAppendDomain(char *host)
+{
+    if (Config.appendDomain && !strchr(host, '.') && !strchr(host, ':')) {
+        const auto dlen = strlen(host);
+        if (dlen > (SQUIDHOSTNAMELEN - Config.appendDomainLen - 1)) {
+            debugs(23, 2, "URL domain too large (" << dlen << " bytes)");
+            return false;
+        }
+        strncat(host, Config.appendDomain, SQUIDHOSTNAMELEN - dlen - 1);
```
rousskov: Either we trust our math or we do not:

```suggestion
        strncat(host, Config.appendDomain, Config.appendDomainLen);
```

src/anyp/Uri.cc
yadij: The (SQUIDHOSTNAMELEN - Config.appendDomainLen) underflow is the less likely to be problematic. That is only run-time variable with the append_domain length. Which is highly unlikely any admin will exceed (and we can validate on configure if you want to be completely certain). Whereas the suggested (dlen+ N) form is dependent on arbitrary remote input. Using an explicit int64 though should be sufficient until this function is dropped entirely by the SBuf refactoring.

src/anyp/Uri.cc
yadij: We used that way back. But static analysis showed that the terminator can be elided if the to-be-copied amount does not account for it. In this case that would result in random heap or stack memory being copied into the URL strings between the appended portion and the buffer end terminator (assuming we remember to put that in).

So current strncat() practice is to indicate the available buffer space and let the function put the terminator in the right place for us if it happens early. Yes we still need the -1 hack for end-of-buffer termination case, but that is better than mixing in sprinkles of +1 hacks into the math too.

src/anyp/Uri.cc
yadij: done

src/anyp/Uri.h
yadij: Done, with extra mention of the dotless TLD assumption.

src/anyp/Uri.cc
rousskov: I will not waste time arguing which of the two unlikely events is more likely. Fixing the expression while preventing sum overflows using 64bit integers is, naturally, fine with me. If possible, please avoid unsigned-to-signed conversion though: Use `uint64_t` instead of `int64_t`.

src/anyp/Uri.cc
rousskov: I do not insist on this change, so I am resolving this request to save time on discussing problems with your facts/arguments.

PR #437, opened 2019-07-16
Detail validation errors for loaded-from-file certificate chains

ghulands:
This just adds logging of the openssl failure reason for
checking the certificate chain.

PR #438, opened 2019-07-16
Detail validation errors for loaded-from-file certificate chains

ghulands:


rousskov:
There is a similar PR against v4: #437 

@ghulands, PR title/description will become official commit title/description. I edited them to reflect what should be committed. See https://wiki.squid-cache.org/MergeProcedure#Pull_Request for more info.

rousskov:
OK to test

src/security/KeyData.cc
```diff
@@ -119,14 +119,15 @@ Security::KeyData::loadX509ChainFromFile()
...
+            const auto checkCode = X509_check_issued(ca, latestCert.get());
+            if (checkCode == X509_V_OK) {
                 debugs(83, DBG_PARSE_NOTE(3), "Adding issuer CA: " << nameStr);
                 // OpenSSL API requires that we order certificates such that the
                 // chain can be appended directly into the on-wire traffic.
                 latestCert = CertPointer(ca);
                 chain.emplace_front(latestCert);
             } else {
-                debugs(83, DBG_PARSE_NOTE(2), "Ignoring non-issuer CA from " << certFile << ": " << nameStr);
+                debugs(83, DBG_PARSE_NOTE(2), "Ignoring non-issuer CA from " << certFile << ": " << nameStr << ". Reason: " << X509_verify_cert_error_string(checkCode));
```
rousskov: I checked and discovered that X509_verify_cert_error_string() does not print the error code. To ease triage in complicated cases, I think we should print both the numeric code and the error string, but I do not insist on this change.

```suggestion
                debugs(83, DBG_PARSE_NOTE(2), certFile << ": Ignoring non-issuer CA " << nameStr << ": " << X509_verify_cert_error_string(checkCode) << " (" << checkCode << ")");
```

PR #442, opened 2019-07-18
Refactor RefCount tests to use cppunit

josepjones:


kinkie:
Ok to test

src/Makefile.am
```diff
@@ -773,7 +773,7 @@ squid.conf.default squid.conf.documented: cf_parser.cci
 cf_parser.cci: cf.data cf_gen$(EXEEXT)
 	./cf_gen$(EXEEXT) cf.data $(srcdir)/cf.data.depend
 
-# The cf_gen_defines.cci is auto-generated and does not exist when the 
+# The cf_gen_defines.cci is auto-generated and does not exist when the
```
yadij: unrelated whitespace change
```suggestion
# The cf_gen_defines.cci is auto-generated and does not exist when the 
```

src/Makefile.am
```diff
@@ -851,7 +851,7 @@ CLEANFILES += cf.data squid.conf.default squid.conf.documented \
 test_tools.cc: $(top_srcdir)/test-suite/test_tools.cc
 	cp $(top_srcdir)/test-suite/test_tools.cc .
 
-# stock tools for unit tests - library independent versions of dlink_list 
+# stock tools for unit tests - library independent versions of dlink_list
```
yadij: unrelated whitespace change.
```suggestion
# stock tools for unit tests - library independent versions of dlink_list 
```

src/tests/testRefCount.cc
```diff
@@ -41,91 +42,123 @@ class AlsoRefCountable : public RefCountable, public _ToRefCount
...
-        ToRefCount anotherObject(new _ToRefCount);
+    ToRefCount anObject;
+    ToRefCount anotherObject(new _ToRefCount);
 
-        if (anObject == anotherObject)
-            exit(1);
+    CPPUNIT_ASSERT(anObject != anotherObject);
 
-        anotherObject = NULL;
+    anotherObject = NULL;
```
yadij: Please use nullptr instead of NULL on any touched lines.

And please also add a check on the Instances state being correct before this operation. Several of the other tests need the same change. To be sure the post-operation state is a valid test of the expected change the pre-operation state need to be confirmed.

PR #445, opened 2019-07-28
RFC 7230: server MUST reject messages with BWS after field-name

yadij:
Obey the RFC requirement to reject HTTP requests with whitespace
between field-name and the colon delimiter. Rejection is
critical in the presence of broken HTTP agents that mishandle
malformed messages.

Also obey requirement to always strip such whitespace from HTTP
response messages. The relaxed parser is no longer necessary for
this response change.

For now non-HTTP protocols retain the old behaviour of removal
only when using the relaxed parser.

### DIFF
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1421,7 +1421,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
 
     if (name_len > 65534) {
         /* String must be LESS THAN 64K and it adds a terminating NULL */
-        debugs(55, DBG_IMPORTANT, "WARNING: rejecting due to header name of " << name_len << " bytes (" << Raw('value_start', value_start) << ")");
+        debugs(55, 2, "found header name of " << name_len << " bytes (" << Raw("value_start", value_start, name_len) << ")");
         return NULL;
     }
 
@@ -1452,7 +1452,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
             --name_len;
 
         if (!name_len) {
-            debugs(55, DBG_IMPORTANT, "WARNING: rejecting due to header with name of " << name_len << " bytes");
+            debugs(55, 2, "found header with only whitespace for name");
             return NULL;
         }
     }
@@ -1487,7 +1487,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
 
     if (field_end - value_start > 65534) {
         /* String must be LESS THAN 64K and it adds a terminating NULL */
-        debugs(55, DBG_IMPORTANT, "WARNING: rejecting due to '" << theName << "' header of " << (field_end - value_start) << " bytes");
+        debugs(55, 2, "WARNING: found '" << theName << "' header of " << (field_end - value_start) << " bytes");
         return NULL;
     }
 
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1425,17 +1425,21 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, http_hdr_
         return NULL;
     }
 
-    // check for BSP following header field-name
+    /*
+     * RFC 7230 section 3.2.4:
+     * "No whitespace is allowed between the header field-name and colon.
+     * ...
+     *  A server MUST reject any received request message that contains
+     *  whitespace between a header field-name and colon with a response code
+     *  of 400 (Bad Request).  A proxy MUST remove any such whitespace from a
+     *  response message before forwarding the message downstream."
+     */
     if (xisspace(field_start[name_len - 1])) {
 
-        // RFC 7230 section 3.2.4
-        // - server MUST reject any request message with this whitespace
-        // - a proxy MUST remove this whitespace from HTTP responses
-        //
-        // for now also let relaxed parser remove this BSP from any non-HTTP messages
         if (msgType == hoRequest)
             return nullptr;
 
+        // for now, also let relaxed parser remove this BWS from any non-HTTP messages
         const bool stripWhitespace = (msgType == hoReply) ||
                                      Config.onoff.relaxed_header_parser;
         if (!stripWhitespace)
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -469,7 +469,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             break;      /* terminating blank line */
         }
 
-        auto *e = HttpHeaderEntry::parse(field_start, field_end, owner);
+        const auto e = HttpHeaderEntry::parse(field_start, field_end, owner);
         if (!e) {
             debugs(55, warnOnError, "WARNING: unparseable HTTP header field {" <<
                    getStringPrefix(field_start, field_end-field_start) << "}");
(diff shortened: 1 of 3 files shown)


src/HttpHeader.cc
```diff
@@ -1424,19 +1421,40 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end)
...
+
         debugs(55, Config.onoff.relaxed_header_parser <= 0 ? 1 : 2,
                "NOTICE: Whitespace after header name in '" << getStringPrefix(field_start, field_end-field_start) << "'");
 
         while (name_len > 0 && xisspace(field_start[name_len - 1]))
             --name_len;
 
-        if (!name_len)
+        if (!name_len) {
+            debugs(55, DBG_IMPORTANT, "WARNING: rejecting due to header with name of " << name_len << " bytes");
```
rousskov: An average admin reading this message in cache.log would not know _what_ Squid is rejecting and what rejecting that thing actually _means_. Both answers are probably caller/context-dependent so this message itself is probably misplaced. I do not insist on fixing this in the PR because this PR does not introduce this problem and because the fix is probably not trivial.

The suggestion below just makes the wording less awkward without addressing the problems mentioned above. If you use it here, please apply to other similar messages touched by this PR, of course.

```suggestion
            debugs(55, DBG_IMPORTANT, "WARNING: cannot parse message headers due to a " << name_len << "-byte long header name");
```

src/HttpHeader.cc
```diff
@@ -1469,7 +1487,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end)
 
     if (field_end - value_start > 65534) {
         /* String must be LESS THAN 64K and it adds a terminating NULL */
-        debugs(55, DBG_IMPORTANT, "WARNING: ignoring '" << theName << "' header of " << (field_end - value_start) << " bytes");
+        debugs(55, DBG_IMPORTANT, "WARNING: rejecting due to header of " << (field_end - value_start) << " bytes");
```
rousskov: I would still report the header name here because many admins may find it useful (especially the few admins that are going to find this message itself useful). If needed, use the Raw interface for reporting the name. I do not insist on restoring name reporting.

src/HttpHeader.cc
```diff
@@ -1407,7 +1404,7 @@ HttpHeaderEntry::~HttpHeaderEntry()
 
 /* parses and inits header entry, returns true/false */
 HttpHeaderEntry *
-HttpHeaderEntry::parse(const char *field_start, const char *field_end)
+HttpHeaderEntry::parse(const char *field_start, const char *field_end, http_hdr_owner_type &msgType)
```
rousskov: The new API should not imply that parse() modifies msgType.

```suggestion
HttpHeaderEntry::parse(const char *field_start, const char *field_end, http_hdr_owner_type msgType)
```

or

```suggestion
HttpHeaderEntry::parse(const char *field_start, const char *field_end, const http_hdr_owner_type msgType)
```

src/HttpHeader.cc
```diff
@@ -469,15 +469,12 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             break;      /* terminating blank line */
         }
 
-        HttpHeaderEntry *e;
-        if ((e = HttpHeaderEntry::parse(field_start, field_end)) == NULL) {
+        auto *e = HttpHeaderEntry::parse(field_start, field_end, owner);
```
rousskov: If we do not reassign e, then please use `const`:

```suggestion
        const auto e = HttpHeaderEntry::parse(field_start, field_end, owner);
```

src/HttpHeader.cc
yadij: Done.

src/HttpHeader.cc
yadij: This begins a cascade of side effects via push_back inside addEntry() needing the type of HttpHeaderEntry::entries vector to match the const'ness of this auto. Fixing the const'ness is going to have to be left until later.

src/HttpHeader.cc
yadij: Okay. Done

src/HttpHeader.cc
yadij: 'cannot parse' would be wrong for this message. The header is just being rejected as invalid according to the HTTP requirements. The parser itself could continue on to other headers if we want to make that happen - out of scope now.

I am inclined to make this and the below output both level 2 so that the message only appears in the context of admin debugging why an HTTP message is being rejected. Do you agree?

yadij:
> Please adjust this PR description to explain why we are now rejecting requests with spaces after fields names. Simply referring to an "RFC requirement" is not sufficient IMO because our relaxed parser should and does violate RFC requirements. We ought to have a more serious/compelling motivation for this change than just following RFC requirements.

What are you looking for here? I think these being a MUST requirement should be enough justification. The added code docs and RFC reference lead one to the specific details without having to lookup the commit message.

The relaxed parser should never be violating RFC requirements. That is what the --enable-http-violations is for. Relaxed is for abnormal syntax where the spec leaves interpretation up to the implementation, or explicitly outlines a tolerant handling for it. If you can point out code using relaxed parser to directly violate the spec in a --disable-http-violations build that is a bug we need to fix.

src/HttpHeader.cc
rousskov: > 'cannot parse' would be wrong for this message

You are being too literal -- "parsing" can (and, at the admin level, usually should) be viewed as all the stages of internalizing a message, including RFC requirements checks, but please use whatever phrasing you think would be better than the current broken one.

> I am inclined to make this and the below output both level 2

Yes, that would be a step forward, especially if the corresponding access log entry makes it clear that the message was rejected (whatever that means). I am a bit worried about ICAP messages that do not have an enabled-by-default access log -- it may be difficult for the admin to know that something went wrong specifically when talking to the ICAP service, but we can hope that the side effects would be visible enough in most cases (and enhance this reporting after specific complains).

rousskov:
> I think these being a MUST requirement should be enough justification.

I do not. And I suspect that admins that will start complaining about this change will agree with me. If you can make these violations conditional on `--enable-http-violations`, then I would support that. If you cannot, then use the reason _why_ you cannot to justify this change.

yadij:
Very well:
 By placing whitespace between the field name and colon of an HTTP request any client can perform the following actions:
 * open an arbitrary two-way tunnel to any server using only a GET or POST request.
 * inject content into the Squid cache.
 * corrupt arbitrary existing cache entries,
 * scan the internal structure of a CDN or ISP using Squid.
You want the 0-day code published too?

These are currently the behaviour when relaxed parser is enabled (on by default) but can be disabled by turning it off. Placing it behind --enable-http-violations would leave these as the default behaviour of nearly all Squid but remove the config option to disable, since you and others insist that option remain as set by default on builds.

rousskov:
It feels like my basic concern is agitating you for some unknown to me reason, but I have to ask to understand: Assuming Squid recognizes and removes that extra space before interpreting and forwarding header fields, do those horrors require something else, like a second broken proxy or a broken origin server? I do not see how an extra space _alone_ can do that damage if all agents are aware of the problem and avoid misinterpreting header fields with extra spaces.

yadij:
Did you get the squid-bugs report from regis with the details?  All that is required is that Squid tolerates whitespace before the colon. Either by "fixing" it or by ignoring the header entirely.

yadij:
rev 840246fdeb3922c405e77d4b4909e7df37d7bf43 should resolves these last polish requests. I have left the change to class Raw API to minimize this patch size, will do that in a followup cleanup PR shortly.

rousskov:
@yadij, this PR is on your side.

PR #446, opened 2019-07-28
Take cacert dates into account when generating certiticate

HeisSpiter:


src/ssl/gadgets.cc
```diff
@@ -479,18 +479,24 @@ static bool buildCertificate(Security::CertPointer & cert, Ssl::CertificatePrope
...
-    // We should get caCert notBefore and notAfter fields and do not allow
-    // notBefore/notAfter values from certToMimic before/after notBefore/notAfter
-    // fields from caCert.
-    // Currently there is not any way in openssl tollkit to compare two ASN1_TIME
-    // objects.
+    ASN1_TIME *notBefore = NULL;
+    ASN1_TIME *notAfter = NULL;
+
+    // Get times from cacert
+    notBefore = X509_getm_notBefore(properties.signWithX509.get());
```
rousskov: The modern C++ way of doing this is

```suggestion
    const auto notBefore = X509_getm_notBefore(properties.signWithX509.get());
```

... but I suggest renaming this variable to clarify that it refers to CA/signing certificate rather than the certificate being mimicked:


```suggestion
    const auto caNotBefore = X509_getm_notBefore(properties.signWithX509.get());
```

Besides other advantages, this version avoids `NULL` which is banned from new/changed Squid code.

However, based on the code further below, properties.signWithX509 may be nil. I would not test how OpenSSL deals with nil pointers and check for that explicitly in Squid code. That probably pushes us one step back:

```suggestion
    ASN1_TIME *notBefore = nullptr;
    if (const auto caCert = properties.signWithX509.get())
        caNotBefore = X509_getm_notBefore(caCert);
```

src/ssl/gadgets.cc
```diff
@@ -479,18 +479,24 @@ static bool buildCertificate(Security::CertPointer & cert, Ssl::CertificatePrope
...
+    notAfter = X509_getm_notAfter(properties.signWithX509.get());
```
rousskov: Here and elsewhere in this review, please see change requests for `notBefore`. They are symmetric.

This particular assignment can go inside the same `if` statement as the assignment for `caNotBefore`.

src/ssl/gadgets.cc
```diff
@@ -479,18 +479,24 @@ static bool buildCertificate(Security::CertPointer & cert, Ssl::CertificatePrope
...
+
     ASN1_TIME *aTime = NULL;
     if (!properties.setValidBefore && properties.mimicCert.get())
         aTime = X509_getm_notBefore(properties.mimicCert.get());
     if (!aTime && properties.signWithX509.get())
```
rousskov: You want to check the presence of the limit in the CA certificate, not the presence of the CA certificate itself.

```suggestion
     if (!aTime && caNotBefore)
```

... but since assigning nil is OK here, perhaps we can simplify further:

```suggestion
     if (!aTime)
         aTime = caNotBefore; // may also be nil
```

src/ssl/gadgets.cc
```diff
@@ -479,18 +479,24 @@ static bool buildCertificate(Security::CertPointer & cert, Ssl::CertificatePrope
...
-        aTime = X509_getm_notBefore(properties.signWithX509.get());
+        aTime = notBefore;
 
     if (aTime) {
+        // If time is before cacert, fix it
+        if (notBefore && ASN1_TIME_compare(aTime, notBefore) == -1)
```
rousskov: It is more logical (and faster) to make this check earlier and only when aTime was set from the very beginning, not here.

```suggestion
```

rousskov: IIRC, Squid still supports OpenSSL versions that lack ASN1_TIME_compare(). If I am right, then this PR has to limit the new adjustment to build environments where ASN1_TIME_compare() is actually available. In your adjusted ./configure.ac, please do _not_ use an OpenSSL version check. Use ASN1_TIME_compare() availability check instead.

HeisSpiter:
Thanks for your review. I've taken all your remarks into account and pushed a new version of the patch addressing them all.

src/ssl/gadgets.cc
```diff
@@ -479,16 +479,28 @@ static bool buildCertificate(Security::CertPointer & cert, Ssl::CertificatePrope
...
     ASN1_TIME *aTime = NULL;
-    if (!properties.setValidBefore && properties.mimicCert.get())
+    if (!properties.setValidBefore && properties.mimicCert.get()) {
         aTime = X509_getm_notBefore(properties.mimicCert.get());
-    if (!aTime && properties.signWithX509.get())
-        aTime = X509_getm_notBefore(properties.signWithX509.get());
+
+#if HAVE_LIBCRYPTO_ASN1_TIME_COMPARE
+        // If time is before cacert, fix it
+        if (caNotBefore && ASN1_TIME_compare(aTime, caNotBefore) == -1)
```
rousskov: AFAICT by looking at OpenSSL sources, ASN1_TIME_compare() will do strange, unwanted things when aTime is nil. The old code implied that aTime can be nil here. I suggest adding a simple static ASN1_TIME_compare() wrapper that will return -2 when at least one of its ASN1_TIME pointer arguments is nil and use that wrapper here:

```suggestion
        if (xASN1_TIME_compare(aTime, caNotBefore) == -1)
```

As an added bonus, you will move `HAVE_LIBCRYPTO_ASN1_TIME_COMPARE` inside that wrapper, reducing code duplication:

```
/// ASN1_TIME_compare() wrapper that
/// \returns -2 when one of the arguments is nil or ASN1_TIME_compare() is missing
int
ASN1_TIME_compare(ASN1_TIME *a, ASN1_TIME *b)
{
    if (!a || !b)
        return -2;
#if HAVE_LIBCRYPTO_ASN1_TIME_COMPARE
    return ASN1_TIME_compare(a, b); // also returns -2 on errors
#else  
    return -2; // OpenSSL v1.0 lacks ASN1_TIME_compare()
#endif
}
```

The `if` statement in the wrapper is left outside HAVE_LIBCRYPTO_ASN1_TIME_COMPARE check on purpose -- to avoid unused function parameter warnings from the compiler.

The wrapper can be added right above the buildCertificate() function you are editing. As a static helper function, it does not need to go into any header file.

HeisSpiter: Integrated! Thanks :-)

src/ssl/gadgets.cc
```diff
@@ -459,6 +459,20 @@ addAltNameWithSubjectCn(Security::CertPointer &cert)
     return result;
 }
 
+/// ASN1_TIME_compare() wrapper that
+/// \returns -2 when one of the arguments is nil or ASN1_TIME_compare() is missing
+static int
+ASN1_TIME_compare(ASN1_TIME *a, ASN1_TIME *b)
```
rousskov:
```suggestion
xASN1_TIME_compare(ASN1_TIME *a, ASN1_TIME *b)
```

src/ssl/gadgets.cc
```diff
@@ -479,16 +493,26 @@ static bool buildCertificate(Security::CertPointer & cert, Ssl::CertificatePrope
...
+
     ASN1_TIME *aTime = NULL;
-    if (!properties.setValidBefore && properties.mimicCert.get())
+    if (!properties.setValidBefore && properties.mimicCert.get()) {
         aTime = X509_getm_notBefore(properties.mimicCert.get());
-    if (!aTime && properties.signWithX509.get())
-        aTime = X509_getm_notBefore(properties.signWithX509.get());
+
+        // If time is before cacert, fix it
+        if (caNotBefore && ASN1_TIME_compare(aTime, caNotBefore) == -1)
```
rousskov:
```suggestion
        if (xASN1_TIME_compare(aTime, caNotBefore) == -1)
```

src/ssl/gadgets.cc
```diff
@@ -497,10 +521,17 @@ static bool buildCertificate(Security::CertPointer & cert, Ssl::CertificatePrope
...
 
     aTime = NULL;
-    if (!properties.setValidAfter && properties.mimicCert.get())
+    if (!properties.setValidAfter && properties.mimicCert.get()) {
         aTime = X509_getm_notAfter(properties.mimicCert.get());
-    if (!aTime && properties.signWithX509.get())
-        aTime = X509_getm_notAfter(properties.signWithX509.get());
+
+        // If time is after cacert, fix it
+        if (caNotAfter && ASN1_TIME_compare(aTime, caNotAfter) == 1)
```
rousskov:
```suggestion
        if (xASN1_TIME_compare(aTime, caNotAfter) == 1)
```

rousskov:
BTW, you do not need to squash and force-push your changes during review. [Anubis](https://github.com/measurement-factory/anubis#readme) will squash your changes automatically when it is time to merge your branch into the official repository. For details, please see [Merge Procedure](https://wiki.squid-cache.org/MergeProcedure), especially its [Pull Request](https://wiki.squid-cache.org/MergeProcedure#Pull_Request) section.

HeisSpiter:
Everything got updated conforming to your last review :-)

rousskov:
OK to test

src/ssl/gadgets.cc
```diff
@@ -479,16 +493,26 @@ static bool buildCertificate(Security::CertPointer & cert, Ssl::CertificatePrope
...
+
     ASN1_TIME *aTime = NULL;
-    if (!properties.setValidBefore && properties.mimicCert.get())
+    if (!properties.setValidBefore && properties.mimicCert.get()) {
         aTime = X509_getm_notBefore(properties.mimicCert.get());
-    if (!aTime && properties.signWithX509.get())
-        aTime = X509_getm_notBefore(properties.signWithX509.get());
+
+        // If time is before cacert, fix it
+        if (xASN1_TIME_compare(aTime, caNotBefore) == -1)
```
yadij: It looks to me like this action is misplaced. The time fix is only supposed to be done if properties.setValidBefore is true. In this current scope it is false.

Same applies for the properties.setValidAfter statements.

rousskov: > The time fix is only supposed to be done if properties.setValidBefore is true.

The current placement gave me a pause as well, but I think the PR code matches documentation:

```
setValidBefore
    Sets the "Not Before" property to the "Not Before" property of
    the CA certificate used to sign generated certificates.
```

When properties.setValidBefore is true, we do not need to check which timestamp is earlier. Instead, we must force aTime to become caNotBefore. And that is what the old code (unchanged by this PR) already does.

When properties.setValidBefore is false, then we are allowed (and, hence, should) check the mimicked certificate property and, if needed, fix it. That is what the PR code does now.

HeisSpiter: > It looks to me like this action is misplaced. The time fix is only supposed to be done if properties.setValidBefore is true. In this current scope it is false.

This is not my understanding. When setValidBefore is set, it forces to used CA timestamp, and that one only. And thus, timestamp is valid.
But when setValidBefore is not set, nothing says the generated certificate can have invalid before date.
And this is the behaviour this PR fixes: it makes sure the certificate is always within the CA time range.
But I'm OK to come over my changes and to change documentation instead.

yadij: Alex:
> The current placement gave me a pause as well, but I think the PR code matches documentation

As far as the specific option *current* behaviour matching its docs, I agree. If that was the ideal behaviour, then the TODO can just be erased. My reading of the TODO was that there is a way to mimic better by checking the timestamps. Better mimic meaning more transparency of the mimic'd cert - ie *less* altered cert values.

The issue bothering me is that this change currently is a step towards client-first problems and away from the transparency that mimic was intended to provide. Below I outline a pseudo code sequence that meets the TODO wording while increasing the cases where cert generated is valid.


HeisPSpter:
> It looks to me like this action is misplaced. The time fix is only supposed to be done if properties.setValidBefore is true. In this current scope it is false.
> 
> This is not my understanding. When setValidBefore is set, it forces to used CA timestamp, and that one only. And thus, timestamp is valid.

This matches my understanding of what the flag does in Squid right now. But that does not guarantee validity of the value injected. The CA cert may have a bungled setup with far future/past timestamp or something weird in that field.

IMO what should be checking here is that the CA cert timestamp being used leaves the generated cert at least as valid as the mimic'd one. I.e. no screwing up TLS by injecting bad CA details.

> But when setValidBefore is not set, nothing says the generated certificate can have invalid before date.

Nothing says it can be set to the CA one when setting is *disabled*.

The point of mimic is to pass on to the client as much of the cert untouched as possible - so problems can be decided there instead of in Squid. Imagine a origin cert that said it was valid since 1973 for TLS/1.1+ use only. Right now this change would erase that potential problem signal from client view in exactly the case we want the client to be aware. It may (or not) be a part of malware signature - but only the client can decide that properly and has upgrade cycle fast enough to change its mind if things change overnight.

> And this is the behaviour this PR fixes: it makes sure the certificate is always within the CA time range.

That "always" is IMO the problem.

We should be using this time range comparison to increase the cases where Squid _avoids_ changing the cert fields when the CA timestamp limits are fine with the value being mimic'd.

This is howI expected to see this TODO implemented:
>> If setValidBefore is true {
>>   If caNotBefore < mimicCert.notBefore; then
>>      use the mimicCert.notBefore value (can be transparent, yay);
>>   else
>>    use the caNotBefore to fix errors
>> } else
>>  use the mimicCert.notBefore value to preserve errors

rousskov: > Below I outline a pseudo code sequence that meets the TODO wording while increasing the cases where cert generated is valid.

Thank you for providing that sketch!

```
If setValidBefore is true {
    If caNotBefore < mimicCert.notBefore; then
        use the mimicCert.notBefore value 
```

The above part of the sketch contradicts setValidBefore documentation and, IIRC, setValidBefore intended purpose. We can, of course, add another mimicking algorithm option that will do something like what you sketched, but doing so would be outside this PR scope.

Please keep in mind that the TODO-like comment in sources predates setValidBefore functionality. Hence, the comment context (at the time of its writing) was equivalent to the current context with setValidBefore false. The PR changes live in that context; they do what the comment authors wanted (while not doing what setValidBefore does not want).

Please also note that setValidBefore is not a global boolean option. It is an ACL-controlled option. If somebody wants to fine-tune setValidBefore application, they can do that using ACLs. In some situations, an admin may still need a new certificate adaptation algorithm (e.g., to limit the range of mimicked before/after dates or even to specify a specific date) or a new ACL (e.g., to match suspicious before/after dates in the being-mimicked certificate), but they already can limit setValidBefore application quite a bit, making that option safe and simple to use.

Finally, as I said earlier, the setValidBefore _name_ is unfortunate. The name does not reflect what that algorithm does. IIRC, the customer wanted to fix an invalid date of the real origin certificate, and that is probably where we got the "valid" part of the name from, but it is misleading because it incorrectly implies that the algorithm _finds_ and _fixes_ invalid dates. In reality, it does not find them (ACLs do), and it does not have a notion of a "fix" or even "correct" date (it just blindly sets the date to the CA date).

If you use a better name for setValidBefore, your sketch would become:

```
If set_NotBeforeDate_to_CA's_NotBeforeDate is true {
    ...
        use the mimicCert.notBefore value 
```

... which, of course, does not match what the algorithm promises to do.

----

P.S. FWIW, I have no feedback regarding your other, informal comments addressed to me because there are too many very different ways for me to interpret them.

yadij: good point on the setValidX intention not matching their names implied meaning. If we leave that case as-is for now, we are still left with the problem of mimic supposed to be preserving errors for the client to decide.

Can we locate the info about whether these fields are valid or invalid on the server link? then time compare would allow us to ensure that state is preserved. That may mean generating an intentionally invalid time if our CA would normally cause the mimicCert value to be considered valid.

rousskov: > If we leave that case as-is for now, we are still left with the problem of mimic supposed to be preserving errors for the client to decide.

Squid is currently generating _invalid_ certificate chains when the validity range of Squid's signing certificate contradicts the validity range of the certificate being mimicked. I know of two cases here that are worth looking at:

1. The validity range of the certificate being mimicked is "good" (i.e. the range belongs to the validity range of the original/server signing certificate and/or satisfy whatever conditions its is supposed to satisfy). The PR works as intended in this case.
2. The validity range of the certificate being mimicked is "bad" (e.g., the dates go beyond the range of the original/server signing certificate). AFAICT, you are talking about this case now. I agree that it would be better to (continue to) mimic a "bad" range in this case. The PR currently "fixes" instead of mimicking that bad range, at least for some "bad" cases.

> Can we locate the info about whether these fields are valid or invalid on the server link?

I do not know what you mean, but if nobody comes up with an authoritative definition/algorithm, then I suggest to define "good" or "bad" validity dates in the certificate being mimicked based on the validity dates in the original signing certificate (or the certificate being used by Squid to validate the certificate being mimicked):

* If a validity date of the certificate being mimicked goes outside the validity range of the server signing certificate, that date is "bad". Squid should mimic that problem by using a date that also lies outside Squid's signing certificate validity range. Where possible, Squid should use the same date as in the certificate being mimicked.
* If a validity date of the certificate being mimicked is inside the validity range of the server signing certificate, that date is "good". Squid should mimic that "lack of a problem" by using a date that is also inside Squid's signing certificate validity range. Where possible, Squid should use the same date as in the certificate being mimicked. This is what the PR is doing now IIRC.

If you agree with the above algorithm, then this PR should be adjusted to detect bad (out of range) date(s) and preserve their problems by generating a certificate with similarly bad date(s).

yadij: I am thinking of different definitions for good vs bad. But agreed on the algorithm itself.

For those definitions we have two CA. Lets look at an example when they have completely non-overlapping date ranges to clarify good vs bad:

 A) our custom CA with NotBefore 2017-01-01 and NotAfter 2017-12-31
 B) origin server CA with NotBefore 2018-01-01 and NotAfter 2018-12-31

When the origin server cert was issued within 2018 it would have been "good" according to (B) but our (A) declared it "bad". We should adjust our generated date to somewhere within the 2017 range so it validates as "good" against our CA.

When the origin server cert was issued in 2017 it would have been "bad" according to (B) but our (A) declared it "good". We need to adjust our generated date to somewhere outside the 2017 range so it validates as "bad".

When the origin cert as issued in 2019 it would have been "bad" according to both. So no particular need to do anything. If anything is done, make sure it stays outside the (A) CAs 2017 "good" range so it validates as "bad" for the client.

rousskov: Glad we agree on the algorithm. @HeisSpiter, can you make the necessary adjustments to implement the first bullet of that [algorithm](https://github.com/squid-cache/squid/pull/446#discussion_r311574086). You have already implemented the second bullet.

----

> Amos: I am thinking of different definitions for good vs bad. 

You have not provided yours, so I cannot tell whether they are better or worse than mine. If I infer from your example, then they seem identical except you define goodness/badness relative to a given CA while I define goodness/badness relative to the signing CA. My definition is

* A certificate is good if its validity period overlaps the validity period of its signing certificate.

Using my definition, we can summarize the algorithm as: 

* A mimicking Squid should keep good certificates good and bad certificates bad while making as few date field changes as possible.


I think my approach is simpler and more natural: Our CA cannot really make the certificate to be mimicked "bad" -- there is no relationship between those two certificates and nobody outside Squid will see them side-by-side. Our CA can only make the mimicked certificate (i.e. the certificate Squid signs) "bad".

HeisSpiter: @rousskov, unless I'm totally wrong, while in certificate generator, you don't have access to the CA which has signed the certificate you're trying to mimic.
So, I'm not sure how we can preserve the fact that the original certificate has invalid not after/not before attributes.

rousskov: Since Squid validates the server certificate, Squid and/or OpenSSL has access to the signing CA at some point. I suspect that access happens _before_ Squid starts to mimic the server certificate. Unfortunately, I do not remember where exactly that happens and cannot tell you how to pass the good/bad determination(s) (or the signing certificate itself) to the generating code.

If you cannot find the answer, please ping @chtsanti. He may not be available for a couple of weeks, but he knows these code areas better than I do.

PR #449, opened 2019-08-05
Bug 4978: eCAP crash after using MyHost().newRequest()

RemmargorP:
Since commit 8babada, Squid was using a c_str() result after its
std::string toString() source went out of scope.

### DIFF
--- a/src/adaptation/ecap/MessageRep.cc
+++ b/src/adaptation/ecap/MessageRep.cc
@@ -204,8 +204,7 @@ Adaptation::Ecap::RequestLineRep::uri(const Area &aUri)
 {
     // TODO: if method is not set, AnyP::Uri::parse will assume it is not connect;
     // Can we change AnyP::Uri::parse API to remove the method parameter?
-    auto uriBuffer = aUri.toString();
-    const bool ok = theMessage.url.parse(theMessage.method, uriBuffer.c_str());
+    const auto ok = theMessage.url.parse(theMessage.method, aUri.toString().c_str());
     Must(ok);
 }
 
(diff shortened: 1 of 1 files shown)


src/adaptation/ecap/MessageRep.cc
```diff
@@ -204,8 +204,8 @@ Adaptation::Ecap::RequestLineRep::uri(const Area &aUri)
 {
     // TODO: if method is not set, AnyP::Uri::parse will assume it is not connect;
     // Can we change AnyP::Uri::parse API to remove the method parameter?
-    const char *buf = aUri.toString().c_str();
-    const bool ok = theMessage.url.parse(theMessage.method, buf);
+    std::string uriBuffer = aUri.toString();
```
yadij: Please use auto instead of std::string to avoid implicit type casting and unnecessary copies (two re-allocate cycles is bad enough).

RemmargorP: fixed

rousskov: > Please use auto instead of std::string to avoid implicit type casting and unnecessary copies (two re-allocate cycles is bad enough).

FYI: Yes, we should use `auto` when we can, including here, but there was no implicit cast here, and `auto` does not eliminate any copies in this context AFAICT: Area::toString() returns std::string which was the type used in the original PR code.

rousskov:
OK to test.

yadij:
@RemmargorP does Alex edited version still resolve the issue you had that prompted this?

rousskov:
I edited the PR description again after realizing that there is a Bugzilla bug report for this PR.

RemmargorP:
> @RemmargorP does Alex edited version still resolve the issue you had that prompted this?

Yes, his version works too. 

Thanks for such fast replies!

PR #450, opened 2019-08-06
Fix openSUSE ./configure-time Linux PAM authentication detection

AdamMajer:
Recently, openSUSE Factory enabled -Wreturn-type in CFLAGS which
resulted in ./configure failing to detect Linux type of PAM
authentication.


yadij:
OK to test

PR #451, opened 2019-08-07
Suggest disallowing access to link-local IP addresses

art-bjss:
The Capital One hack was caused by a misconfigured proxy (no suggestion that it was Squid, 
of course) that allowed external access to the EC2 instance metadata endpoint at 
http://169.254.169.254/latest/meta-data/

It seems to me there's no reason for Squid to allow access to this IP address which is covered
 by the link-local IP range.

The `to_localhost` restriction is commented out by default so this PR will have no effect on the
 default configuration.


yadij:
Further to the reason Alex outlined this subnet is included in the localnet ACL definition which admin are advised to edit to match their actual LAN ranges.

Admin need to ensure their policy rules account for what they want to do with traffic from any IP range(s) they use. That is no more (or less) important for link-local than any of the other possible LAN ranges.

PR #454, opened 2019-08-08
Fixed MemStore::copyFromShm() debugging

eduard-bagdasaryan:
* Level-9 is reserved for data dumps.

* The "mem-loaded all" message was printing -1 instead of the
  accumulated object size. It also deserves a lower debugging level
  because it happens at most once per transaction.


rousskov:
I forced-pushed to remove recently added master [changes](https://github.com/squid-cache/squid/commit/a9e52ad461809514afba4d6eb18f55da49813fe7) from this PR branch. Those master changes were probably merged to work around the Semaphore CI `diff --check` [failure](https://semaphoreci.com/squid/squid/branches/pull-request-454/builds/1). I undid them to test whether the fixed `diff --check` works. Looks like It does, but I will continue to keep an eye on it. Once all the other checks are completed, Anubis should merge this PR into master normally.

PR #455, opened 2019-08-09
Add local FS to systemd startup dependencies

yadij:
The squid -z startup process requires that disk FS are
running if any cache_dir are configured.

### DIFF
--- a/tools/systemd/squid.service
+++ b/tools/systemd/squid.service
@@ -8,7 +8,7 @@
 [Unit]
 Description=Squid Web Proxy Server
 Documentation=man:squid(8)
-After=network.target network-online.target nss-lookup.target
+After=network.target network-online.target nss-lookup.target local-fs.target
 
 [Service]
 Type=forking


rousskov:
This is not my area of expertise, so I do not plan on voting here, but I know that Squid configuration files sometimes refer to pieces stored on remote file systems. If, for some strange reason, we want to go down the slippery slope of requiring anything that might be needed in some cases (but is not needed by default), then perhaps Squid startup should depend on non-local FS as well?

marcosfrm:
Unnecessary. `squid.service` uses the implicit `DefaultDependencies=yes`. Hence `local-fs.target` dependency is already there.

See:
https://www.freedesktop.org/software/systemd/man/systemd.service.html#Default%20Dependencies
https://www.freedesktop.org/software/systemd/man/systemd.special.html#sysinit.target
https://www.freedesktop.org/software/systemd/man/bootup.html#System%20Manager%20Bootup

marcosfrm:
For the remote filesystem case, I would left to the sysadmin expand squid.service:

```
systemctl edit squid.service
```

Then add:

```
[Unit]
After=remote-fs.target
```

yadij:
Thank you for that info marcos. It seems the issue the user was having must be somewhere else than the local-fs being loaded.

marcosfrm:
Hi Amos. I assume it is https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=932593 .

`systemctl show local-fs.target` will help debug the issue. Anyway, it probably sould be assigned to Debian's systemd package.

PR #456, opened 2019-08-11
Ignore malformed Host header in intercept and reverse proxy mode

yadij:


src/http/one/Parser.cc
```diff
@@ -256,6 +256,11 @@ Http::One::Parser::getHeaderField(const char *name)
         // prevent buffer overrun on char header[];
         p.chop(0, sizeof(header)-1);
 
+        // XXX: currently only used for pre-parse Host header, ensure valid domain[:port] or ip[:port]
```
rousskov: I see no reason why this specific XXX should be added to master instead of being addressed internally in this PR. There are several ways to address it AFAICT, including:

1. The best solution I can think of is to adjust this method to return `Optional<SBuf>`, based on the _minimum_ subset of features from C++17 [optional](https://en.cppreference.com/w/cpp/utility/optional) template. Please let me know if you need my help adding such Optional. That template will be generally useful in many similar contexts (and will eventually be replaced with the STL template).
2. A simpler, C-style solution would be to adjust this method API to return bool and fill an SBuf header field value argument (when returning true).

In both cases, a new getHostHeader() method can then use the adjusted getHeaderFIeld() API to return a validated c-string host value or nil. That new method will contain the code you wrote for this PR, without the XXX part.

src/http/one/Parser.cc
```diff
@@ -256,6 +256,11 @@ Http::One::Parser::getHeaderField(const char *name)
...
+        static const CharacterSet hostValid = CharacterSet("host",":[].-_") + CharacterSet::ALPHA + CharacterSet::DIGIT;
```
rousskov: If possible, please use auto.

I also suggest renaming hostValid to hostChars for consistency sake (many such sets are named using fooChars and similar patterns) and to avoid the implication that all those chars are valid in a host name.

```suggestion
        static const auto hostChars = CharacterSet("host",":[].-_") + CharacterSet::ALPHA + CharacterSet::DIGIT;
```

rousskov:
I suggest using the following text for the PR title to focus more on the effect of the changes. PR description can then become empty.

`Ignore malformed Host header in intercept and reverse proxy mode`

src/http/one/Parser.cc
yadij: I do not want to waste anyone time getting into anything complicated about perfecting this API. It is going to have a quite different design and requirements when we do the proper Tokenizer updates for the mime parse ("soon") - all this functions code and anything we do here is going to be thrown away at that time.

I considered having the callers pass in their validChars characterset, to have the API remain header-agnostic. But with it only being Host for all callers and erasure the destiny of this code it did not seem worth even going that far. The XXX is to highlight that anyone changing this function or its internals is going to have to adjust these to account for the new requirements - this *will* break things for them otherwise (lacking comon punctuation characters, etc).

src/http/one/Parser.cc
yadij: Done

src/http/one/Parser.cc
yadij: I have just renamed the function getHostHeaderField and removed the input parameter to avoid implication that it does anything but find Host header. That should be enough to avoid the issues the XXX was for without getting into more complicated things.

PR #460, opened 2019-08-14
Fix "BUG: Lost previously bumped from-Squid connection"

chtsanti:
FwdState assumed that PeerSelector always returns the connection pinned
by previous SslBump steps. That assumption was wrong in two cases:
 
1. The previously pinned connection got closed.
2. PeerSelector policies prevent pinned connection reuse. For example,
    connection destination is now denied by cache_peer_access rules.

PeerSelector now returns a PINNED selection even if a pinned connection
cannot or should not be used. The initiator is now fully responsible for
checking the returned connection usability, including the new
ConnStateData::pinning::peerAccessDenied flag. Unusable pinned
connection is now treated as any other fatal (for the transaction)
forwarding problem rather than an internal BUG.

The above changes do not change traffic on the wire but remove bogus
level-1 BUG messages from cache.log.

We also polished which error page is returned depending on the pinning
validation problem: ERR_ZERO_SIZE_OBJECT is returned when the validation
failed because of the peer disappearance or to-server connection
closure. Other cases use ERR_CANNOT_FORWARD. Eventually, these errors
can be detailed further to distinguish various problems. We may also
want to generalize ERR_CONFLICT_HOST and/or ERR_FORWARDING_DENIED to
make them applicable in this context.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -558,28 +558,6 @@ FwdState::noteDestination(Comm::ConnectionPointer path)
 
     debugs(17, 3, path);
 
-    if (request->pinnedConnection()) {
-        // This is a previously pinned connection which not allowed to be used
-        // any more probably because of peer selection policy.
-
-        if (!healthyPinnedConnection()) {
-            // client connection is going to be closed soon.
-            stopAndDestroy("pinned connection is closing");
-            return;
-        }
-
-        // We have a pinned connection and we should check if reconnect/retry
-        // is allowed for this type of pinned connections.
-        if (!retriablePinned()) {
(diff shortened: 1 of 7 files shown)


rousskov:
I added `S-waiting-for-author` for conflict resolution.

chtsanti:
I rebase again to latest master branch a70e75b7.
I hope it is ok.

> Do feel free to add a new error template for when the CANNOT_FORWARD is not fitting well. Though creating an error page is rather a lot of work, so avoiding might be better when there is definitely no client to deliver it to.

The patch does not add any new error template. The CANNOT_FORWARD is enough to describe the problem at least for now (error while trying to forward through a pinned connection).

PR #462, opened 2019-08-15
Fix string truncation in getnameinfo()

yadij:
strncpy can leave strings unterminated. Use xstrncpy instead
of relying on callers to terminate their buffers.


compat/getnameinfo.cc
```diff
@@ -24,6 +24,8 @@
  *
  *    13-Jan-2015 : Various fixed for C++ and MinGW native build
  *
+ *    16-Aug-2019 : Replace strncpy with xstrncpy for string termination
```
rousskov: I do not recommend adding fresh entries to this stale history because it creates a false impression that this history is accurate and being maintained. The best thing is to remove the entire history blob, but doing so is probably outside this PR scope.

yadij: Strange as it may seem the history here is accurate and up to date for the manual changes. The only omissions are whitespace formatting and copyright updates made by scripts.

rousskov: I see at least one non-trivial git change record (quoted below) that is missing or misrepresented in the current manually-maintained history. And scripted "copyright updates" may be important in some cases. And even if the current "manual" history were perfect, we should not be maintaining it anyway! We should (continue to) use git for change records, of course.

----

```
commit 55d7d5e
Date:   Sun Dec 21 08:28:17 2014 -0800

Windows: fix getaddrinfo, getnameinfo, inet_ntop and inet_pton detection

These API symbols are not always defined as functions, and in varying
locations. AC_REPLACE_FUNCS cannot handle that kind of complexity so we
must use AC_CHECK_DECL instead and provide the sequence of #include
necessary to identify their existence.
```

yadij: That is the original import commit - changes referred to are in our code to *use* this files code. If you think that "git add", #include squid.h or the #ifdef's around this file constitute logic change then I will add it, but I do not.

rousskov: There are lower-level mistakes in your response, but I will try to stay focused on the big picture in hope to reduce time waste: To maintain change history, we should use git rather than source code files themselves. Please do not update this in-file history snapshot.

yadij: Fine. Done.

PR #463, opened 2019-08-20
Supply ALE for note ACL

eduard-bagdasaryan:
The following cache.log WARNING is a symptom of this bug:

    ACL is used in context without an ALE state. Assuming mismatch.

Commit cb36505 already covered many (but not all) similar cases.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -6779,6 +6779,7 @@ DOC_START
 	    logformat myFormat ... %{key}note ...
 
 	This clause only supports fast acl types.
+	See http://wiki.squid-cache.org/SquidFaq/SquidAcl for details.
 DOC_END
 
 NAME: relaxed_header_parser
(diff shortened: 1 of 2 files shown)


rousskov:
v4 has the same bug

src/cf.data.pre
```diff
@@ -6777,6 +6777,8 @@ DOC_START
 
 	    note key value acl ...
 	    logformat myFormat ... %{key}note ...
+
+	This clause only supports fast acl types.
```
yadij: Please include the wiki reference line as other uses of that statement do.

yadij:
Ping @eduard-bagdasaryan, this looks important enough to try merging before the next v4 release (just a few days away). I'd rather no muddy the authorship for a simple reference addition, but will do if still necessary to complete the PR tomorrow.

eduard-bagdasaryan:
> Please include the wiki reference line

Done at 8124eb6.

PR #464, opened 2019-08-20
Fix cross-compilation

ffontaine:
Do not check check file descriptor maximum value through AC_RUN_IFELSE
when cross-compiling as this will raise an error

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/acinclude/os-deps.m4
+++ b/acinclude/os-deps.m4
@@ -231,7 +231,7 @@ int main(int argc, char **argv) {
 	fprintf (fp, "%d\n", i & ~0x3F);
 	return 0;
 }
-  ]])],[squid_filedescriptors_limit=`cat conftestval`],[],[])
+  ]])],[squid_filedescriptors_limit=`cat conftestval`],[],[:])
   dnl Microsoft MSVCRT.DLL supports 2048 maximum FDs
   AS_CASE(["$host_os"],[mingw|mingw32],[squid_filedescriptors_limit="2048"])
   AC_MSG_RESULT($squid_filedescriptors_limit)


rousskov:
OK to test

rousskov:
Was Squid cross-compilation completely broken before these changes? Or does this AC_RUN_IFELSE() actually work OK in some cross-compilation cases but not in others?

ffontaine:
I found this issue when bumping squid from 4.6 to 4.8 in buildroot. In buildroot, we are forcing file descriptors to 1024 (with --with-filedescriptors=1024) since the addition of squid in version 3.0 back in 2009. 
Without this option, when cross-compiling, indeed the default will be 256 but I did not want to change this part without your feedback. Before this bump, everything was building fine, the issue has been introduced with https://github.com/squid-cache/squid/commit/6afc7f3d5d579eca3676bb1b46e9c3a987fa3bda especially:

`]])],[squid_filedescriptors_num=`cat conftestval`],[squid_filedescriptors_num=256],[squid_filedescriptors_num=256])`

has been replaced by:
`
]])],[squid_filedescriptors_limit=`cat conftestval`],[],[])`

In other words, the _action-if-cross-compiling_ from the AC_RUN_IFELSE call has been replaced from `[squid_filedescriptors_num=256]` to its default action `[]` which is exiting with an error. More information can be found on AC_RUN_IFELSE here : https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Runtime.html

I do agree that my PR could be enhanced and I would be happy to update it.

yadij:
This test is intended to leave squid_filedescriptors_limit unset in the event that the test fails or cannot be run for any reason. It should never halt the build.


Does replacing both those "[]" with "[:]" resolve your cross-compile issue?
if yes, then that would be a better solution for this PR to implement.

ffontaine:
OK, I updated the PR has requested.

PR #465, opened 2019-08-21
Bug 4918: Crashes when using OpenSSL prior to v1.0.2

cotequeiroz:
The implementation of x509_get0_signature() replacement in 24b30fd was
based on OpenSSL v1.1.0 where `signature` and `sig_alg` members of
`x509_st` structure stopped being raw pointers and became structures.
The mismatch caused segfaults when using OpenSSL versions that lacked
x509_get0_signature() -- anything earlier than OpenSSL v1.0.2.

    // OpenSSL < v1.1.0
    struct x509_st {
        X509_CINF *cert_info;
        X509_ALGOR *sig_alg;
        ASN1_BIT_STRING *signature;
        ...
    }

    // OpenSSL >= v1.1.0
    struct x509_st {
        X509_CINF cert_info;
        X509_ALGOR sig_alg;
        ASN1_BIT_STRING signature;
        ...
    }

A C-style reinterpreting cast hid the type mismatch from the compilers
and reviewers.

Tested with OpenSSL v1.0.1f. The types of the two data members were
checked back to OpenSSL v0.9.6.

Signed-off-by: Eneas U de Queiroz <cotequeiroz@gmail.com>

### DIFF
--- a/compat/openssl.h
+++ b/compat/openssl.h
@@ -177,9 +177,9 @@ extern "C" {
     X509_get0_signature(ASN1_BIT_STRING **psig, X509_ALGOR **palg, const X509 *x)
     {
         if (psig)
-            *psig = (ASN1_BIT_STRING *)&x->signature;
+            *psig = x->signature;
         if (palg)
-            *palg = (X509_ALGOR *)&x->sig_alg;
+            *palg = x->sig_alg;
     }
 #endif
 


rousskov:
@cotequeiroz, please check the updated PR description. It obeys the formatting rules, but I removed reverence to `openssl < 1.0.2` because that specific version seemed to contradict the rest of the explanation. The structure change happened in OpenSSL v1.1.0, right? Please adjust the description as needed.

rousskov:
OK to test.

cotequeiroz:
> 
> 
> @cotequeiroz, please check the updated PR description. It obeys the formatting rules, but I removed reverence to `openssl < 1.0.2` because that specific version seemed to contradict the rest of the explanation. The structure change happened in OpenSSL v1.1.0, right? Please adjust the description as needed.

It's all good.

>C-style reinterpret cast bites again!

The compiler showed it, but having a mess of `const` vs. no-const arguments between different openssl versions clouded my judgment on this one. :see_no_evil:

rousskov:
> I did not notice the title. The `X509_get0_signature` call was added in version 1.0.2, so it only fails prior to that. The `struct` changed in 1.1.0, but the shim is only used before 1.0.2.

Aha, that explains why I thought there is an "extra" info in your original writeup. Thank you for fixing the PR title and detailing this caveat! I added the missing version information to the PR description. Hopefully, we both are on the same page now. The PR should be auto-merged soon.

PR #466, opened 2019-09-05
Size how much printable data gets printed by Raw

yadij:
An arbitrary default of 256 bytes has been chosen. Callers can use the
whole() method to dump the entire data if necessary.

We size the number of characters to print rather than the total
data size of the output. Encodings (e.g., hex()) and decorations (e.g. labels and
truncation indicators) may result in more output characters than the
data size. For example, a 100-octet buffer sized to 10 octets would be
printed using 20 hex printables.

### DIFF
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1539,8 +1539,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
 
     if (name_len > 65534) {
         /* String must be LESS THAN 64K and it adds a terminating NULL */
-        // TODO: update this to show proper name_len in Raw markup, but not print all that
-        debugs(55, 2, "ignoring huge header field (" << Raw("field_start", field_start, 100) << "...)");
+        debugs(55, 2, "ignoring huge " << Raw("field-name", field_start, name_len).atMost(100));
         return nullptr;
     }
 
@@ -1584,8 +1583,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
      */
     for (const char *pos = field_start; pos < (field_start+name_len); ++pos) {
         if (!CharacterSet::TCHAR[*pos]) {
-            debugs(55, 2, "found header with invalid characters in " <<
-                   Raw("field-name", field_start, min(name_len,100)) << "...");
+            debugs(55, 2, "found invalid characters in " << Raw("field-name", field_start, name_len).atMost(100));
             return nullptr;
         }
     }
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -486,7 +486,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
     int warnOnError = (Config.onoff.relaxed_header_parser <= 0 ? DBG_IMPORTANT : 2);
 
     assert(header_start && header_end);
-    debugs(55, 7, "parsing hdr: (" << this << ")" << std::endl << getStringPrefix(header_start, hdrLen));
+    debugs(55, 3, "parsing hdr: (" << this << ")" << std::endl << Raw("", header_start, hdrLen).minLevel(3).whole());
     ++ HttpHeaderStats[owner].parsedCount;
 
     char *nulpos;
@@ -613,8 +613,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
 
     if (clen.headerWideProblem) {
         debugs(55, warnOnError, "WARNING: " << clen.headerWideProblem <<
-               " Content-Length field values in" <<
-               Raw("header", header_start, hdrLen));
+               " Content-Length field values in header (hdr: " << this << ")");
     }
 
     String rawTe;
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -534,7 +534,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             if (memchr(this_line, '\r', field_end - this_line)) {
                 hasBareCr = "bare CR";
                 debugs(55, warnOnError, "WARNING: suspicious CR characters in " <<
-                       Raw("field", field_start, field_end-field_start).asHex().minLevel(warnOnError));
+                       Raw("field", field_start, field_end-field_start).hex().minLevel(warnOnError));
 
                 if (Config.onoff.relaxed_header_parser) {
                     char *p = (char *) this_line;   /* XXX Warning! This destroys original header content and violates specifications somewhat */
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1528,7 +1528,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
 
     if (name_len > 65534) {
         /* String must be LESS THAN 64K and it adds a terminating NULL */
-        debugs(55, 2, "ignoring huge " << Raw("field-name", field_start, name_len));
+        debugs(55, 2, "ignoring field with huge " << Raw("field-name", field_start, name_len).minLevel(2));
         return nullptr;
     }
 
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1539,7 +1539,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
 
     if (name_len > 65534) {
         /* String must be LESS THAN 64K and it adds a terminating NULL */
-        debugs(55, 2, "ignoring huge " << Raw("field-name", field_start, name_len).atMost(100));
+        debugs(55, 2, "ignoring huge " << Raw("field-name", field_start, name_len));
         return nullptr;
     }
 
@@ -1583,7 +1583,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
      */
     for (const char *pos = field_start; pos < (field_start+name_len); ++pos) {
         if (!CharacterSet::TCHAR[*pos]) {
-            debugs(55, 2, "found invalid characters in " << Raw("field-name", field_start, name_len).atMost(100));
+            debugs(55, 2, "found invalid character at position " << (pos-field_start) << " in " << Raw("field-name", field_start, name_len));
             return nullptr;
         }
     }
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -124,16 +124,6 @@ httpHeaderMaskInit(HttpHeaderMask * mask, int value)
     memset(mask, value, sizeof(*mask));
 }
 
-/** handy to printf prefixes of potentially very long buffers */
-static const char *
-getStringPrefix(const char *str, size_t sz)
-{
-#define SHORT_PREFIX_SIZE 512
-    LOCAL_ARRAY(char, buf, SHORT_PREFIX_SIZE);
-    xstrncpy(buf, str, (sz+1 > SHORT_PREFIX_SIZE) ? SHORT_PREFIX_SIZE : sz);
-    return buf;
-}
-
 void
 httpHeaderInitModule(void)
 {
@@ -492,7 +482,8 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
     char *nulpos;
     if ((nulpos = (char*)memchr(header_start, '\0', hdrLen))) {
         debugs(55, DBG_IMPORTANT, "WARNING: HTTP header contains NULL characters {" <<
-               getStringPrefix(header_start, nulpos-header_start) << "}\nNULL\n{" << getStringPrefix(nulpos+1, hdrLen-(nulpos-header_start)-1));
+               Raw("", header_start, nulpos-header_start).minLevel(DBG_IMPORTANT).whole() <<
+               "}\nNULL\n{" << Raw("", nulpos+1, hdrLen-(nulpos-header_start)-1).minLevel(DBG_IMPORTANT));
         clean();
         return 0;
     }
@@ -532,7 +523,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
                     if (cr_only) {
                         debugs(55, DBG_IMPORTANT, "SECURITY WARNING: Rejecting HTTP request with a CR+ "
                                "header field to prevent request smuggling attacks: {" <<
-                               getStringPrefix(header_start, hdrLen) << "}");
+                               Raw("", header_start, hdrLen).minLevel(DBG_IMPORTANT).whole() << "}");
                         clean();
                         return 0;
                     }
@@ -542,8 +533,8 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             /* Barf on stray CR characters */
             if (memchr(this_line, '\r', field_end - this_line)) {
                 hasBareCr = "bare CR";
-                debugs(55, warnOnError, "WARNING: suspicious CR characters in HTTP header {" <<
-                       getStringPrefix(field_start, field_end-field_start) << "}");
+                debugs(55, warnOnError, "WARNING: suspicious CR characters in " <<
+                       Raw("field", field_start, field_end-field_start).asHex().minLevel(warnOnError));
 
                 if (Config.onoff.relaxed_header_parser) {
                     char *p = (char *) this_line;   /* XXX Warning! This destroys original header content and violates specifications somewhat */
@@ -559,8 +550,8 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             }
 
             if (this_line + 1 == field_end && this_line > field_start) {
-                debugs(55, warnOnError, "WARNING: Blank continuation line in HTTP header {" <<
-                       getStringPrefix(header_start, hdrLen) << "}");
+                debugs(55, warnOnError, "WARNING: Blank continuation line in " <<
+                       Raw("field", field_start, field_end-field_start).whole().minLevel(warnOnError));
                 clean();
                 return 0;
             }
@@ -569,7 +560,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         if (field_start == field_end) {
             if (field_ptr < header_end) {
                 debugs(55, warnOnError, "WARNING: unparsable HTTP header field near {" <<
-                       getStringPrefix(field_start, hdrLen-(field_start-header_start)) << "}");
+                       Raw("", field_start, hdrLen-(field_start-header_start)).minLevel(warnOnError) << "}");
                 clean();
                 return 0;
             }
@@ -579,10 +570,9 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
 
         const auto e = HttpHeaderEntry::parse(field_start, field_end, owner);
         if (!e) {
-            debugs(55, warnOnError, "WARNING: unparsable HTTP header field {" <<
-                   getStringPrefix(field_start, field_end-field_start) << "}");
-            debugs(55, warnOnError, " in {" << getStringPrefix(header_start, hdrLen) << "}");
-
+            debugs(55, warnOnError, "WARNING: unparsable HTTP header " <<
+                   Raw("field", field_start, field_end-field_start).minLevel(warnOnError) <<
+                   " in hdr: " << this);
             clean();
             return 0;
         }
@@ -1562,8 +1552,8 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
         if (!stripWhitespace)
             return nullptr; // reject if we cannot strip
 
-        debugs(55, Config.onoff.relaxed_header_parser <= 0 ? 1 : 2,
-               "WARNING: Whitespace after header name in '" << getStringPrefix(field_start, field_end-field_start) << "'");
+        debugs(55, Config.onoff.relaxed_header_parser <= 0 ? DBG_IMPORTANT : 2,
+               "WARNING: Whitespace after field-name in '" << Raw("field", field_start, field_end-field_start).whole().minLevel(DBG_IMPORTANT));
 
         while (name_len > 0 && xisspace(field_start[name_len - 1]))
             --name_len;
@@ -1588,8 +1578,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
     }
 
     /* now we know we can parse it */
-
-    debugs(55, 9, "parsing HttpHeaderEntry: near '" <<  getStringPrefix(field_start, field_end-field_start) << "'");
+    debugs(55, DBG_DATA, Raw("field", field_start, field_end-field_start));
 
     /* is it a "known" field? */
     Http::HdrType id = Http::HeaderLookupTable.lookup(field_start,name_len).id;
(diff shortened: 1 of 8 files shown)


src/Debug.h
```diff
@@ -184,8 +184,8 @@ void _db_rotate_log(void);
 class Raw
 {
 public:
-    Raw(const char *label, const char *data, const size_t size):
-        level(-1), label_(label), data_(data), size_(size), useHex_(false), useGap_(true) {}
+    Raw(const char *label, const char *data, const size_t size, const size_t limit = 0):
```
rousskov: Please remove the limit from the constructor arguments. Raw uses dedicated methods for configuration options like that (e.g., hex(), gap(), and minLevel()). Please add such a method if at least one caller needs it. I suggest calling it atMost(), avoiding an overly generic "limit".

src/debug.cc
```diff
@@ -876,7 +876,8 @@ Raw::printHex(std::ostream &os) const
     const auto savedFill = os.fill('0');
     const auto savedFlags = os.flags(); // std::ios_base::fmtflags
     os << std::hex;
-    std::for_each(data_, data_ + size_,
+    auto len = (limit_ > 0 && limit_ < size_ ? limit_ : size_);
```
rousskov: Please add a printableSize() or similar method to avoid duplicating this code.

src/Debug.h
```diff
@@ -214,6 +214,7 @@ class Raw
     const char *label_; ///< optional data name or ID; triggers size printing
     const char *data_; ///< raw data to be printed
     size_t size_; ///< data length
+    size_t limit_; ///< do not print more than this amount of the data
```
rousskov:
```suggestion
    size_t limit_; ///< do not print more data (0 means unlimited)
```

src/debug.cc
```diff
@@ -902,8 +903,10 @@ Raw::print(std::ostream &os) const
         if (data_) {
             if (useHex_)
                 printHex(os);
-            else
-                os.write(data_, size_);
+            else {
+                auto len = (limit_ > 0 && limit_ < size_ ? limit_ : size_);
+                os.write(data_, len);
```
rousskov:
```suggestion
                os.write(data_, printableSize());
```

src/debug.cc
```diff
@@ -876,7 +876,8 @@ Raw::printHex(std::ostream &os) const
     const auto savedFill = os.fill('0');
     const auto savedFlags = os.flags(); // std::ios_base::fmtflags
     os << std::hex;
-    std::for_each(data_, data_ + size_,
+    auto len = (limit_ > 0 && limit_ < size_ ? limit_ : size_);
+    std::for_each(data_, data_ + len,
```
rousskov:
```suggestion
    std::for_each(data_, data_ + printableSize(),
```

src/debug.cc
```diff
@@ -902,8 +903,10 @@ Raw::print(std::ostream &os) const
         if (data_) {
             if (useHex_)
                 printHex(os);
-            else
-                os.write(data_, size_);
+            else {
+                auto len = (limit_ > 0 && limit_ < size_ ? limit_ : size_);
+                os.write(data_, len);
+            }
```
rousskov: Please provide a visual clue to the reader that the data was truncated:

```suggestion
        if (printableSize() < size_)
            os << "[...]";
```

You can even print the number of truncated bytes: `[... N more bytes ...]`.

rousskov (changes requested):
> A quick check indicates most current callers will not find a limit useful - so the initial default is no limiting.

I believe we should emphasize/prioritize logging safety/sanity _and_ flag large printers by specifying a large but still reasonable limit as the default (instead of no limit). There should be very few cases where printing more than, say, 1000 bytes of data is desirable/intentional. The default should reflect _that_ principle (rather than the fact that most Raw users either print small strings or deliberately print potentially large strings). Selecting a reasonable default limit now may also help minimize the time wasted on arguing about specific limits applied to individual borderline callers.

Please add a whole() method to be used by "large" printers to remove the default limit discussed above.

src/debug.cc
yadij:
```suggestion
```

src/debug.cc
```diff
@@ -913,8 +914,10 @@ Raw::print(std::ostream &os) const
         if (data_) {
             if (useHex_)
                 printHex(os);
-            else
-                os.write(data_, size_);
+            else {
+                auto len = (limit_ > 0 && limit_ < size_ ? limit_ : size_);
```
yadij, line 918:
```suggestion
```

src/debug.cc
```diff
@@ -913,8 +914,10 @@ Raw::print(std::ostream &os) const
...
+                os.write(data_, len);
+            }
```
yadij, line 920:
```suggestion
```

src/debug.cc
```diff
@@ -913,8 +914,10 @@ Raw::print(std::ostream &os) const
         if (data_) {
             if (useHex_)
                 printHex(os);
-            else
-                os.write(data_, size_);
+            else {
```
yadij, line 917:
```suggestion
            else
```

src/Debug.h
yadij: done in dc3814d

src/debug.cc
yadij: Done in 33cdf16

src/Debug.h
```diff
@@ -214,10 +217,12 @@ class Raw
 
 private:
     void printHex(std::ostream &os) const;
+    size_t printableSize() const { return (limit_ > 0 && limit_ < size_ ? limit_ : size_); }
```
rousskov, line 220: AFAICT, after the other changes, this method should be removed while limit_ is renamed to printableSize_. It is cheaper and clearer to compute this value correctly once than to call this method two-three times in a row. This change will also clarify what `limit_` is to the code reader (without the need to consult the member declaration).

BTW, the atMost() method name should stay the same. From the caller point of view, that method sets a limit which may (and often will!) exceed `size_`. The _implementation_ of that method should set `printableSize_` correctly, of course.

rousskov (changes requested):
The primary API problem identified in my previous [review](https://github.com/squid-cache/squid/pull/466#pullrequestreview-284241071) still needs to be addressed. The rest is polishing, including another API addition request from the previous review. I tried to itemize all these changes in the inlined requests now. Please let me know if you want me to finish all these changes instead.

> TODO full review of all current uses of class Raw to see which can be limited for better logs

I think this review should be done during this PR. We may decide not to change any other callers, of course, but I see no reason to postpone this review, especially since there are probably fewer than 30 occurrences to review.

The word "full" may imply some complex analysis. I do not think anything like that is needed right now. The focus of this review should be on cases that may have printed more than the default print size -- should they call whole() now? Or specify some larger-than-the-default limit?

Besides making this PR work complete, the review of current Raw users may inspire a different default limit than the currently proposed semi-arbitrary 1000 bytes.

rousskov:
> Alex: We may decide not to change any other callers, of course

The "other" part was wrong -- this PR currently changes no callers. However, there is at least one caller that should be changed -- HttpHeaderEntry::parse(). The need for a printing limit in that caller [triggered](https://github.com/squid-cache/squid/pull/445#discussion_r316909763) this PR! Completing that TODO item may identify other callers, of course.

rousskov:
@yadij, I was surprised to see this PR closed. Limiting Raw output size is clearly a useful feature, needed in existing and future code. Both of us spent quite a bit of time on this PR. Can you detail why you decided to close this PR, please?

yadij:
Small 'oops' cleaning up my local repo. It seems to have removed branches and closed PRs which I only had a copy on my company testing rig. Trying to fix these now.

src/debug.cc
```diff
@@ -914,7 +914,9 @@ Raw::print(std::ostream &os) const
             if (useHex_)
                 printHex(os);
             else
-                os.write(data_, size_);
+                os.write(data_, printableSize());
+            if (printableSize() < size_)
+                os << "[..." << (size_ - printableSize()) << " bytes...]";
```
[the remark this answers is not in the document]
yadij: "more" is implied by the '...', so not adding this extra word.

src/Debug.h
yadij: Done in f06e16fe5

src/Debug.h
```diff
@@ -189,11 +189,14 @@ class Raw
...
 public:
     Raw(const char *label, const char *data, const size_t size):
-        level(-1), label_(label), data_(data), size_(size), useHex_(false), useGap_(true) {}
+        level(-1), label_(label), data_(data), size_(size), limit_(0), useHex_(false), useGap_(true) {}
 
     /// limit data printing to at least the given debugging level
     Raw &minLevel(const int aLevel) { level = aLevel; return *this; }
 
+    /// print no more than n bytes of data (0 means unlimited)
+    Raw &atMost(const size_t n) { limit_ = n; return *this; }
```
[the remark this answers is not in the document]
yadij: Done in 6a29f9520

src/Debug.h
```diff
@@ -189,11 +189,14 @@ class Raw
 {
 public:
     Raw(const char *label, const char *data, const size_t size):
-        level(-1), label_(label), data_(data), size_(size), useHex_(false), useGap_(true) {}
+        level(-1), label_(label), data_(data), size_(size), limit_(0), useHex_(false), useGap_(true) {}
```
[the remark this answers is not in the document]
yadij: Done in 6a29f9520

rousskov:
I have updated PR title/description to better match the current PR code. Primary changes:

* Updated example to match the current code.

* Removed stale TODO: Since the new functionality is now enabled by default, the "full review" of all current uses to add limits is no longer necessary. We should still review current uses during this PR, but that work does not need a TODO in the PR description.

* Re-formatted.

Please adjust further as needed. I will post a review shortly.

rousskov (changes requested):
I think we are getting close to being done here. These remaining Raw changes are just follow-up adjustments after the default-to-limit and zero-is-not-special changes in the earlier reviews. I am happy to do this round myself if you prefer; just let me know.

Related Raw-using code fixes that I know about are:

* Downloader::CbDialer::print()
* "ignoring huge header field" in HttpHeaderEntry::parse() that [triggered](https://github.com/squid-cache/squid/pull/445#discussion_r316909763) this PR
* "found header with invalid characters" in HttpHeaderEntry::parse()
* `REQUEST:` and `RESPONSE:` whole-header dumping code; the ones using Raw will need `.whole()` adjustments, right?

Since we seem to have settled on the API, now is probably a good time to add those changes to the PR. I can do these as well if you prefer; just let me know.

src/base/Raw.cc
```diff
@@ -24,17 +24,19 @@ Raw::print(std::ostream &os) const
 
     // finalize debugging level if no level was set explicitly via minLevel()
     const int finalLevel = (level >= 0) ? level :
-                           (size_ > 40 ? DBG_DATA : Debug::SectionLevel());
+                           (printableSize_ > 40 ? DBG_DATA : Debug::SectionLevel());
```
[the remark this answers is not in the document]
yadij: done in [1af1167](https://github.com/squid-cache/squid/pull/466/commits/1af1167a77b3623c5fd9e7a3e8f0c4011ba4aafe)

src/base/Raw.h
```diff
@@ -20,12 +21,18 @@
...
-    Raw(const char *label, const char *data, const size_t size):
-        level(-1), label_(label), data_(data), size_(size), useHex_(false), useGap_(true) {}
+    Raw(const char *label, const char *data, const size_t size) :
+        label_(label), data_(data), size_(size) { atMost(size); }
 
     /// limit data printing to at least the given debugging level
     Raw &minLevel(const int aLevel) { level = aLevel; return *this; }
 
+    /// print no more than n bytes of data
+    Raw &atMost(const size_t n) { printableSize_ = std::min(n, printableSize_); return *this; }
```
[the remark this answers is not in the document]
yadij: done in [1d0445b](https://github.com/squid-cache/squid/pull/466/commits/1d0445b5a9fce57242f5bf5b35c597e4e32e2777)

yadij:
> I think we are getting close to being done here. These remaining Raw changes are just follow-up adjustments after the default-to-limit and zero-is-not-special changes in the earlier reviews. I

In [2d45af8](https://github.com/squid-cache/squid/pull/466/commits/2d45af89f781daaa76f6877e1b283c505f679cd9) I have gone through all the calls to `Raw("` and adjusted the ones I think need to use a different limit from the default 256 being added here.   I also dropped a few `debugs()` lines in FTP where the buffer content would be printed twice in a row at different levels.

Please check if I missed any you think need changing, or disagree on keeping any of the old 100-byte limits.

kinkie (dismissed):
lgtm

src/base/Raw.cc
```diff
@@ -23,18 +23,22 @@ Raw::print(std::ostream &os) const
         return os;
 
     // finalize debugging level if no level was set explicitly via minLevel()
+    const auto printLimit = std::min(size_, printableSize_);
```
rousskov, line 26: We never print fewer (or more) bytes than this variable, so this variable is not really a "limit" but "size":


```suggestion
    const auto printableDataSize = std::min(size_, printableDataSizeLimit_);
```

The rest of this method code will need to be updated to use the new variable name, of course.

src/base/Raw.h
```diff
@@ -20,12 +21,18 @@
 class Raw
 {
 public:
-    Raw(const char *label, const char *data, const size_t size):
-        level(-1), label_(label), data_(data), size_(size), useHex_(false), useGap_(true) {}
+    Raw(const char *label, const void *data, const size_t size) :
```
rousskov:
```suggestion
    Raw(const char *const label, const void *const data, const size_t size):
```

src/base/Raw.h
```diff
@@ -42,16 +49,17 @@ class Raw
...
 
-    const char *label_; ///< optional data name or ID; triggers size printing
-    const char *data_; ///< raw data to be printed
-    size_t size_; ///< data length
-    bool useHex_; ///< whether hex() has been called
-    bool useGap_; ///< whether to print leading space if label is missing
+    const char *label_ = nullptr; ///< optional data name or ID; triggers size printing
+    const void *data_ = nullptr; ///< raw data to be printed
+    size_t size_ = 0; ///< data length
+    size_t printableSize_ = 256; ///< do not print more data
```
rousskov, line 59: We may print less, so this is not a size but a limit (on size):


```suggestion
    size_t printableDataSizeLimit_ = 256; ///< do not print more data
```

N.B. Since we often print more than data and future Raw API may allow limiting/controlling total print size, let's be more specific here by saying "data size" rather than "size". However,  I do not insist on that. Here is a variant without "data":

```suggestion
    size_t printableSizeLimit_ = 256; ///< do not print more data
```

src/HttpHeader.cc
```diff
@@ -1539,8 +1539,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
 
     if (name_len > 65534) {
         /* String must be LESS THAN 64K and it adds a terminating NULL */
-        // TODO: update this to show proper name_len in Raw markup, but not print all that
-        debugs(55, 2, "ignoring huge header field (" << Raw("field_start", field_start, 100) << "...)");
+        debugs(55, 2, "ignoring huge " << Raw("field-name", field_start, name_len).atMost(100));
```
rousskov: We are ignoring the whole header field, not just its name, so let's not change the existing level-2 explanation:


```suggestion
        debugs(55, 2, "ignoring huge header field with " << Raw("field-name", field_start, name_len).atMost(100));
```

or

```suggestion
        debugs(55, 2, "ignoring huge header field: " << Raw("field-name", field_start, name_len).atMost(100));
```

src/HttpHeader.cc
```diff
@@ -1584,8 +1583,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
      */
     for (const char *pos = field_start; pos < (field_start+name_len); ++pos) {
         if (!CharacterSet::TCHAR[*pos]) {
-            debugs(55, 2, "found header with invalid characters in " <<
-                   Raw("field-name", field_start, min(name_len,100)) << "...");
+            debugs(55, 2, "found invalid characters in " << Raw("field-name", field_start, name_len).atMost(100));
```
rousskov: Since we are changing this line, I would report the character position as well. Something like this may work (untested):


```suggestion
            debugs(55, 2, "found invalid character at position " << (pos-field_start) << " in " << Raw("field-name", field_start, name_len).atMost(100));
```

I do not insist on this improvement.

src/base/Raw.h
```diff
@@ -42,16 +49,17 @@ class Raw
...
+    bool useHex_ = false; ///< whether hex() has been called
```
rousskov, line 60: Current PR mistreats one case where we used to print the whole thing but are now going to print the first 256 characters: `Tunnel server PUSH Payload` dump in tunnel.cc. I support limiting that dump in this PR (by default), like current PR code does, but if we do start to limit it, then we should adjust the corresponding `Tunnel client PUSH Payload` dump a few lines below, to keep both dumps in sync as far as this limit is concerned (at least).

rousskov, line 60: Please adjust `clen.headerWideProblem` dump in `HttpHeader.cc` to print more than 256 characters. That limit is too short for many HTTP headers and, hence, may hide the problem we are trying to report. I can suggest `10*1024`, but I do not insist on any specific number.

rousskov, line 60: Please adjust `CurrentException` and `proto=urn` dumps in `Uri.cc` to restore printing of the whole URI (or a lot more than the first 256 chars). The exception-reporting case is especially important IMO.

rousskov, line 60: Please address `ErrorPage::BuildErrorPrinter::print()` TODO that was waiting for this "Add support for prefix printing to Raw" PR.

src/base/Raw.h
```diff
@@ -11,6 +11,7 @@
 #ifndef SQUID_SRC_BASE_RAW_H
 #define SQUID_SRC_BASE_RAW_H
 
+#include <algorithm>
```
rousskov: AFAICT, this addition is no longer needed (in this header file):

```suggestion
```

It may be needed in `Raw.cc` now.

rousskov (changes requested):
> Please check if I missed any you think need changing

Done. I left a few specific change requests (all tied to an unrelated `bool useHex_` line).

> or disagree on keeping any of the old 100-byte limits.

We should drop them IMO: They were as arbitrary as the PR-proposed 256-byte limit, they were not really context-specific, 256-byte limit works OK for those cases, and I see no good reason to keep two arbitrary limits here. However, I do not insist on dropping those two `atMost(100)` calls.

I also removed the no longer true "removed the private method printHex() unimplemented since 675b840" claim from PR description.

src/base/Raw.cc
yadij: Several problems with this request:

 1) this algorithm is not limited to "printable" bytes.

 2) This is the limit bounding output of the action. Derived from but not limited by the object size.
    - The object itself may be larger, with more "printable" bytes that are not output.
    - The object may technically be smaller than `size_` provided by the caller claims.

 3) `asHex` feature means that we cannot consider this to be the "printable" size/amount of output either.

rousskov: I do not insist on any specific name -- feel free to propose something better while still addressing my concern. For example, `reportableDataSize` and `reportableDataSizeLimit_` could also work (but I think "printable" is better here).

> this algorithm is not limited to "printable" bytes

The word "printable" has several interpretations. IMHO, the proposed name matches one of them, and humans would understand (with high probability) which meaning is used _in this specific context_.

> This is the limit bounding output of the action

It is not a limit; it is the _exact_ size -- we never send fewer data bytes to the printing code; we send exactly that many data bytes.

> The object itself may be larger ...
> [the caller may lie about the object size]

Yes, but those facts are not relevant to this particular variable.

> `asHex` feature means that we cannot consider this to be the "printable" size/amount of output either.

Yes, and that is why we should not call this variable "outputSize" or something like that. Suggested "printableDataSize" does _not_ imply that formatted/decorated output is going to have the same size.

src/base/Raw.h
yadij: Done in [ca89cdb](https://github.com/squid-cache/squid/pull/466/commits/ca89cdbcf4ececd37dd732982525816c95a827d0)

src/HttpHeader.cc
yadij: This is not the entire header field, just the "field-name" syntax token.

src/HttpHeader.cc
yadij: done in [88d8e2a](https://github.com/squid-cache/squid/pull/466/commits/88d8e2ac2c3ac0a42733925cfb5001ee61d71a9d)

src/base/Raw.h
yadij: Agreed and done in 137fd293d7.

Also, I am wondering WTF the `in` local has type `SBuf * const in` instead of `const SBuf &` ?

src/HttpHeader.cc
rousskov: > This is not the entire header field, just the "field-name" syntax token.

The `debugs()` line in question talks about ignoring X. In the official code, X is "header field". In PR code, X became "field-name", and that is wrong -- the method acts on the entire field, not just its name. Hence my change request to restore official wording.

That official wording itself is problematic because this method does not actually "ignore" X and because documenting _caller_ actions is usually a bad idea (and at least one caller does not ignore X either). Fixing _that_ problem is outside this PR scope, but if you do want to fix it, then replace "ignoring huge" with just "huge". After removing "ignore", we do not need to say "header field" anymore:

```C++
        debugs(55, 2, "huge" << Raw("field-name", field_start, name_len));
```

src/base/Raw.h
rousskov: > Also, I am wondering WTF the `in` local has type `SBuf * const in` instead of `const SBuf &` ?

I cannot explain that extra address operator. AFAICT, it was introduced in commit 6c420975 and then preserved in commit fcc444e3. It is not needed, but if you polish that declaration to remove the address operator (and switch to `auto`) in this PR, then this PR will have to touch more lines unrelated to this PR scope. Your call.

yadij: This should be resolved by c3432e2597. Rather than printing the whole header on every warning/error, we can reference the previously output "hdr: this" log ID.

yadij: done in 4778dcd592

rousskov: > This should be resolved by c3432e2597. Rather than printing the whole header on every warning/error, we can reference the previously output "hdr: this" log ID.

FWIW, there is no such commit in this PR right now.

yadij: Done in 169a7b24e3

src/HttpHeader.cc
yadij: Oh, I see now. Thanks.

yadij: Resolved in c1d13c4e47.

src/base/Raw.h
yadij: Aye, will do separately.

yadij: pushed now

src/base/Raw.cc
yadij: So, using your definitions described here to re-write the title and description.

squid-anubis:
Cannot create a git commit message from PR title and description.

Error while parsing PR description body: `the line is too long 81>72`

Problematic parser input:

    data size of the output. Encodings (e.g., hex()) and decorations (e.g. labels and
Please see PR [title](https://github.com/measurement-factory/anubis#pr-title) and [description](https://github.com/measurement-factory/anubis#pr-description) formatting requirements for more details.

This message was added by Anubis bot. Anubis will add a new message if the error text changes. Anubis will remove M-failed-description label when there are no corresponding failures to report.

PR #470, opened 2019-09-10
Make configure check compatible with OpenSSL 1.1

fxcoudert:
OpenSSL 1.1 [deprecated the `SSL_library_init` function](https://www.openssl.org/docs/man1.1.0/man3/SSL_library_init.html), and this function is not present if OpenSSL is compiled without deprecated support (which is the case of several distributions, such as [Homebrew](https://github.com/Homebrew/homebrew-core)).

The `AC_CHECK_LIB` should therefore check for a function that is present in `libssl` for both OpenSSL 1.0 and 1.1. `SSL_CTX_new` is a good candidate.

I can confirm that this patch fixes the build with OpenSSL 1.1, and does not break build with OpenSSL 1.0, on macOS 10.14.

---

Another possibility would be to have _two_ tests, one for OpenSSL 1.0 and one for OpenSSL 1.1. If this is what is desired, the [OpenSSL wiki](https://wiki.openssl.org/index.php/Library_Initialization#Autoconf) suggests using:

```
     FOUND_SSL_LIB="no"
     AC_CHECK_LIB(ssl, OPENSSL_init_ssl, [FOUND_SSL_LIB="yes"])
     AC_CHECK_LIB(ssl, SSL_library_init, [FOUND_SSL_LIB="yes"])
     AS_IF([test "x$FOUND_SSL_LIB" = xno], [AC_MSG_ERROR([library 'ssl' is required for OpenSSL])])
```

but that is more complex.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1335,7 +1335,7 @@ if test "x$with_openssl" = "xyes"; then
     AC_CHECK_LIB(crypto,[CRYPTO_new_ex_data],[LIBOPENSSL_LIBS="-lcrypto $LIBOPENSSL_LIBS"],[
       AC_MSG_ERROR([library 'crypto' is required for OpenSSL])
     ],$LIBOPENSSL_LIBS)
-    AC_CHECK_LIB(ssl,[SSL_library_init],[LIBOPENSSL_LIBS="-lssl $LIBOPENSSL_LIBS"],[
+    AC_CHECK_LIB(ssl,[SSL_CTX_new],[LIBOPENSSL_LIBS="-lssl $LIBOPENSSL_LIBS"],[
       AC_MSG_ERROR([library 'ssl' is required for OpenSSL])
     ],$LIBOPENSSL_LIBS)
   ])


fxcoudert:
>  squid-anubis added the M-failed-description label

What is expected of me?

> Can one of the admins verify this patch?

This pull request addresses @rousskov's comments on another (stuck) PR: https://github.com/squid-cache/squid/pull/333

rousskov:
> What is expected of me?

Please see https://github.com/measurement-factory/anubis#pull-request-labels

rousskov:
OK to test

rousskov:
@fxcoudert, thank you for fixing this problem! Please check edited PR title and description. If you agree with them, please say so, and they will become the official commit message. Otherwise, please adjust them as needed. I will not clear this PR for commit until your confirmation.

FWIW, your discussion about an alternative implementation (that I have deleted from the PR description) was _not_ wrong. It is just not appropriate for the commit message because, from the commit point of view, there is no reason to entertain clearly inferior solutions. I bet the only reason you entertained that more complex solution is because its direction was mentioned in PR #333 but that PR discussion was misdirected itself AFAICT. There are many obviously worse solutions. The commit message does not need to explain obvious things.

fxcoudert:
I'm OK with title and description. I didn't know you use the PR body as a commit message (I think I actually have never seen a project set up that way), so I just wanted to give a detailed justification of why I thought this was the best way to go.

rousskov:
> I'm OK with title and description.

Great, thank you. The PR should be auto-merged soon.

> I think I actually have never seen a project set up that way

Yes, the other projects that use merge bots usually force developers to cleanup development branches instead. Since our bot automatically squashes submitted changes, we both free the developers from (occasionally disrupting/unwelcomed) cleanup and have a much better control over the text of the commit message (which most developers, especially new contributors, naturally get wrong anyway). Our approach is not without cons, of course, but I think it works overall better for the Squid Project.

One of the things we currently lack is an automated message for new contributors explaining what is going on and what to expect.

PR #471, opened 2019-09-12
Add %master_xaction logformat code

eduard-bagdasaryan:
Currently, knowing master transaction ID can be very helpful in triage,
especially when dealing with flash crowds on busy proxies. Upcoming
changes will also tie many current "anonymous" level-0/1 messages to
logged transactions via this ID.


rousskov:
> it may be optimal to followup this with a code to log the kid identifier

That is already supported via the `${process_number}` macro. I use it all the time in SMP tests.

PR #473, opened 2019-09-13
Fix build broken by 6c880a1

eduard-bagdasaryan:


rousskov:
@kinkie, if you have time, could you please check why this bug was missed by Jenkins? Do all the build nodes lack libecap installation? Are we missing `--enable-ecap` somewhere? Something else?

yadij:
Found the problem. The enable-ecap check in configure.ac has a clause setting "no, implicitly" - thus bypassing the library auto-detection logic. Working on fixing that now.

rousskov:
> fixing that

FYI: IIRC, that was considered as desired behavior when the checks were added because enabling ecap support creates additional attack vectors via arbitrary loadable modules. Back then, the decision was that such risks should not be created silently. We can revisit that decision, of course.

As far as _tests_ of stable Squid versions is concerned, ecap support should be enabled _explicitly_ on nodes where such testing is supposed to take place.

PR #474, opened 2019-09-15
Bug 4796: comm.cc !isOpen(conn->fd) assertion when rotating logs

eduard-bagdasaryan:
This long-term solution overrides the short-term fix at 2cd72a2. Now,
debug.cc correctly maintains meta information associated with its file
descriptors.

IMO, the correct place for calling _db_init() is just after locking the
PID file because we want to log ASAP, but cache.log is a common resource
that requires protection. Thus, the two old _db_init() calls were both
excessive and misplaced:

* The first call happens too early, allowing another Squid instance to
  pollute cache.log with messages unrelated to the cache.log-owning
  instance (e.g., a "FATAL: Squid is already running" message when
  attempting to start another instance).

* The second call happens too late, missing earlier debugs() that ought
  to be written into cache.log (e.g., debugs() in comm_init()).

Fixing _db_init() calls led to adjustments like moving mainSetCwd() to
be called prior to the relocated _db_init(). However, these changes
should not affect chroot-sensitive code such as UseConfigRunners().

Some early debugs() messages are no longer written into cache.log:

* Exception messages like "Squid is already running" emitted by another
  Squid instance. This is an improvement: Messages about other instances
  do not belong to the cache.log locked by the running instance.

* Messages (mostly errors and warning) from "finalizeConfig" callers
  (e.g., "WARNING: mem-cache size is too small..."). This loss is
  regrettable. Long-term, these messages should be reported after
  configuration is finalized (TODO). Delayed reporting will also help
  when Squid starts rejecting invalid reconfigurations.

* Messages from a few early enter_suid()/leave_suid() calls, such as
  "enter_suid: PID" and "leave_suid: PID". This is an improvement: These
  debugging messages pollute cache.log.

* A few early SquidMain() messages emitted between parseConfigFile() and
  StartUsingConfig() (e.g., "Doing post-config initialization"). This is
  an improvement: These debugging messages pollute cache.log.

Also removed outdated 'TEST_ACCESS' hack for simplicity sake.

Also marked an old XXX: Chrooted SMP master process does not db_init().


rousskov:
@eduard-bagdasaryan, I have edited the PR description for length. Please check that I have not dropped something important (or misrepresented anything).

src/main.cc
```diff
@@ -1443,12 +1434,57 @@ ConfigureCurrentKid(const CommandLine &cmdLine)
...
+    } else if (Config.chroot_dir) {
+        RunConfigUsers();
+        enter_suid();
+        // TODO: don't we need to RunConfigUsers() in the configured
+        // chroot environment?
+        mainSetCwd();
+        leave_suid();
+        ConfigureDebugging();
+    } else {
+        ConfigureDebugging();
```
yadij: I am a little suspicious that this is different from the other else clause. This might put the debug log in whatever home directory, /tmp or such the init is using.

It seems to me they should both contain the same sequence - and that sequence should set the CWD as early as possible.

rousskov: It is a known fact that this sequence looks suspicious, for several reasons. Eduard's claim, AFAICT, is that essentially the same suspicious sequence is implemented in the official code. @yadij, do you have reasons to believe that the official sequence is different?

Please note that we should refrain from changing the existing sequence in _this_ PR (to the extent possible), regardless of whether that behavior is deemed right or wrong. The added TODOs already reflect known problems with the sequences. If you just want to add more TODOs, please say so explicitly.

> This might put the debug log in whatever home directory, /tmp or such the init is using.

Yes, it will do that in some cases, AFAICT. Eduard's claim, AFAICT, is that it would happen in the official code as well.
* If that claim is accurate, this PR can be merged without changes.
* If that claim is false, _and_ it is possible to preserve the official behavior without drastic changes, then this PR should be fixed to do so.
* If that claim is false but drastic changes would be required to preserve the official behavior, then the behavior change should be explained/documented. Depending on the details in that explanation/documentation, we would then decide whether this PR should be made conditional on a follow up PR that fixes the sequences (and changes the official behavior again) to be what we want them to be.

rousskov: > Amos (on [squid-dev](http://lists.squid-cache.org/pipermail/squid-dev/2019-October/009521.html)): I want to be very sure of the sequence being identical or better before
> this goes in.

@yadij, everybody wants that AFAICT, but it is not clear whether you are actively working on validating PR's "identical or better" claim, or you are just waiting for the PR author to do something that will assure you that those claims are correct. The PR is (or looks) stuck because of this uncertainty regarding your position. Please add a PR comment to resolve this uncertainty.

We should not rely on me _guessing_ that I should not clear this PR for merge until something happens on your side. This guessing and double checking wastes time and, sooner or later, I will guess incorrectly. AFAICT, we have more than enough mechanisms to prevent this guessing and associated stuck PRs. Please use them.

rousskov:
@yadij, FYI: This PR is stuck on your side. Please clarify what your position is as detailed in https://github.com/squid-cache/squid/pull/474#discussion_r334145663

rousskov:
@yadij, this PR is stuck on your side for about three months. Please clarify what your position is as detailed in https://github.com/squid-cache/squid/pull/474#discussion_r334145663. For example, f I should disregard your squid-dev comment instead of waiting for you to clarify your position, please say so.

src/fde.h
```diff
@@ -52,6 +52,9 @@ class fde
 {
 
 public:
+
+    static void Init();
```
yadij: Missing doxygen comments for this public function. Simple as it looks there are complexities like the dependency that Squid_MaxFD being correct before this function is called, and that it asserts.

src/main.cc
```diff
@@ -1443,12 +1434,57 @@ ConfigureCurrentKid(const CommandLine &cmdLine)
...
     RunRegisteredHere(RegisteredRunner::useConfig);
 }
 
+static void
+StartUsingConfig()
+{
+    setMaxFD();
+    fde::Init();
+    const auto skipCwdAdjusting = IamMasterProcess() && InDaemonMode();
+    if (skipCwdAdjusting) {
```
yadij: Auditing the old vs new startup process I find that this clause can be true when using "-z --foreground" - which will result in the chroot environment being ignored, possibly with wrong paths used or incorrect path errors produced.

src/main.cc
```diff
@@ -1443,12 +1434,57 @@ ConfigureCurrentKid(const CommandLine &cmdLine)
...
+        ConfigureDebugging();
+        RunConfigUsers();
+    } else if (Config.chroot_dir) {
+        RunConfigUsers();
+        enter_suid();
```
yadij: Now that mainSetCwd has been removed from all the odd locations these SUID operations should be moved inside.

src/main.cc
yadij: I have now reviewed the call sequence and my suspicion was correct. One of these clause orderings is brand new code sequence. 

The order of RunConfigUsers() before ConfigureDebugging() is what the old code used to do. The first if-block and this final else block are new code. If that new sequence has been tested and works, then great - but we should keep the chroot sequence in sync.

 NP: doing the proposed shuffling of SUID bits into mainSetCwd() and then also moving this "if (Config.chroot_dir)" check to allow mainSetCwd to act as a no-op when irrelevant should resolve this confusing oddity. We would then have a much simpler:
``
 if (!skipCwdAdjusting) {
    mainSetCwd()
}
 ConfigureDebugging();
 RunConfigUsers();
``

PS. For completeness I found that the few code paths which skip or do not call this function at all are still doing what they used to. The technically unrelated move of setMaxFD and removal of TEST_ACCESS look fine.

src/main.cc
eduard-bagdasaryan: > this clause can be true when using "-z --foreground" - which will result in the chroot environment being ignored

I think this is not correct. With these arguments, both patched and unpatched Squids spawn a worker and then call mainSetCwd(). chroot is ignored for the parent process, but it was ignored in unpatched Squid too.

src/main.cc
eduard-bagdasaryan: > The order of RunConfigUsers() before ConfigureDebugging() is what the old code used to do. 

To begin with, both of these methods are new: the old code could not use them. Basically, they correspond to the old StartUsingConfig() and _db_init() respectively, with the difference that ConfigureDebugging() does a little more, opening standard I/O streams. What do we see in the old code? When chroot is not configured, _db_init() is invoked always before StartUsingConfig() from parseConfigFile() internals:
```
ParseConfigFileOrThrow(const char *file_name) {
    [...]
    if (!Config.chroot_dir) {
        leave_suid();
        setUmask(Config.umask);
        _db_init(Debug::cache_log, Debug::debugOptions);
        enter_suid();
    }
}
```
This order is preserved in the StartUsingConfig().

src/main.cc
eduard-bagdasaryan: I am afraid that a couple of 'odd locations' still persist, (see one in mainInitialize()). I think we should postpone your proposal until future refactoring, which would clean up that mess.

src/main.cc
eduard-bagdasaryan: I also run some tests confirming that mainSetCwd() is called before creating swap directories when running "-z --foreground", i.e, these directories should be created in the correct environment.

src/main.cc
yadij: Oh I see. Missed that second one. It does seem quite out of place there.

src/fde.h
rousskov: > Missing doxygen comments for this public function.

Fixed in 3792efe:

```C++
// TODO: Merge with comm_init() to reduce initialization order dependencies.
/// Configures fd_table (a.k.a. fde::Table).
/// Call once, after learning the number of supported descriptors (i.e.
/// setMaxFD()) and before dereferencing fd_table (e.g., before Comm I/O).
static void Init();
```

> Also, we are using the Foo::Instance() pattern elsewhere to avoid globals. A TODO to mark this for future fix would be good.

The initialization complexities need to be hidden and better managed indeed -- I added a TODO -- but we probably should not use the singleton pattern _here_ because fd_table is not the only thing that needs proper (e.g., in the right moment and in the right order) configuration. We could, of course, move _all_ Comm accesses behind Comm::Instance(). Placing the whole module in a singleton _feels_ wrong to me now, but we should at least consider that idea when fixing Comm initialization.

rousskov:
@yadij, it looks like all your change requests have been addressed. You have resolved one of them. Please check whether the changes and discussion resolve the [last](https://github.com/squid-cache/squid/pull/474#discussion_r378620521) [two](https://github.com/squid-cache/squid/pull/474#discussion_r378612237).

src/main.cc
yadij: Testing run-time last night shows that the order is indeed the same in that case, though much later than before (as intended?).

@rousskov , Now that I have the code marked up to display these I'd like to re-run the whole set of permutation checks with backtraces to double-check my notes on the eyeball audit. But if I do not get to that within 5 days feel free to clear this for merge.

rousskov: > if I do not get to that within 5 days feel free to clear this for merge.

Sounds good. And, even if it takes you a bit longer than 5 days, I am sure that we can fix any in-scope bugs your tests expose.

The primary value of this PR, AFAICT, is not the bug it is already fixing but the cleanup of the relevant code that, among other things, exposes several problems that should probably be addressed (separately).

PR #475, opened 2019-09-17
Bug 4987: restore support for the "cache_object://" URI scheme

chtsanti:
Broken by commit 6c880a1.


chtsanti:
> Please wrap this with USE_HTTP_VIOLATIONS.
> 
Done, commit c7a5acde

chtsanti:
This PR is blocked on M-failed-description
I think this is because of title. IT is ok to change it to something like the following?
>   Bug 4987: broken parsing of squid URI scheme "cache_object://"

rousskov:
I rephrased the title to make it shorter

PR #476, opened 2019-09-19
Set ALE::reply to the 200 (Connection established)

eduard-bagdasaryan:
... thus addressing a TODO. Lack of reply may cause
"ACL is used in context without an HTTP response" errors in some
contexts. These contexts (if any) should be fixed separately.

### DIFF
--- a/src/HttpReply.cc
+++ b/src/HttpReply.cc
@@ -117,6 +117,14 @@ HttpReply::pack() const
 }
 
 HttpReply *
+HttpReply::MakeConnectionEstablished() {
+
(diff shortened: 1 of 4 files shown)


src/HttpReply.cc
```diff
@@ -116,6 +116,14 @@ HttpReply::pack() const
     return mb;
 }
 
+HttpReply *
```
yadij: The al->reply being set from this return result stores HttpReplyPointer. Please just return one of those instead of this unique_ptr dance with raw-pointers.

eduard-bagdasaryan: Done at d8c6441.

PR #477, opened 2019-09-19
Supply ALE to request_header_access/reply_header_access

eduard-bagdasaryan:
Commit cb36505 already covered many (but not all) similar cases.

Also supply HttpReply to cache_peer_access. Lack of reply
caused "ACL is used in context without an HTTP response" bug.


eduard-bagdasaryan:
@yadij , if you plan to port these changes into v4, please consider a [patch](https://patch-diff.githubusercontent.com/raw/measurement-factory/squid/pull/53.diff) which is applicable to a recent c311939 revision. This patch contains both PR476 and PR477 fixes.

PR #478, opened 2019-09-19
Fix the SQUID_CC_REQUIRE_ARGUMENT autoconf function

DimitryAndric:
Inside AC_DEFUN(), autoconf replaces `$1` with the first argument of the
function.  In this case, the first argument is a variable name. To get
the _value_ of that variable, one has to use `$$1`.

One known effect of this fix (in many build environments) is the
disappearance of the following annoying extra error when a build fails
for some other reason:

    unrecognized command line option -Wno-deprecated-register

### DIFF
--- a/acinclude/compiler-flags.m4
+++ b/acinclude/compiler-flags.m4
@@ -41,7 +41,7 @@ AC_DEFUN([SQUID_CC_REQUIRE_ARGUMENT],[
     SAVED_CFLAGS="$CFLAGS"
     SAVED_CXXFLAGS="$CXXFLAGS"
     AC_COMPILE_IFELSE([AC_LANG_PROGRAM($3,$4)],[$1=no],[],[$1=no])
-    if test "x$1" != "xno" ; then
+    if test "x$$1" != "xno" ; then
       CFLAGS="$CFLAGS $2"
       CXXFLAGS="$CXXFLAGS $2"
       AC_COMPILE_IFELSE([AC_LANG_PROGRAM($3,$4)],[$1=yes],[$1=no],[$1=no])


rousskov:
OK to test

rousskov:
@DimitryAndric, can you give an example of what did _not_ work because of this bug? In other words, what does this PR fix from a Squid admin point of view (not from the developer point of view)?

DimitryAndric:
> @DimitryAndric, can you give an example of what did _not_ work because of this bug? In other words, what does this PR fix from a Squid admin point of view (not from the developer point of view)?

For a Squid admin, nothing should have changed, as this only influences the configuration stage of the Squid build.  What happens there is that the configure script is testing for a certain feature of the compiler, and incorrectly concludes that the feature doesn't work, while in reality it does work.

Maybe there could be cases where, due to this issue, Squid would not build on a particular system, but I guess it is theoretical.  For me, it just caused a warning suppression flag `-Wno-deprecated-register` to be added, while that was unnecessary.

rousskov:
Many admins build Squid from sources...

> it just caused a warning suppression flag `-Wno-deprecated-register` to be added

OMG! You made my day. That bogus flag has bothered me for so long, but I could not quickly figure out where it is coming from. Thanks a lot for fixing that! I have adjusted the PR description (and, hence, the future commit message) to add that info.

DimitryAndric:
> OMG! You made my day. That bogus flag has bothered me for so long

Haha, indeed I was also annoyed by it!  That is why I had to look into the configure script to figure out where it came from.  It looks like there at least three compiler flags tested in this way, but not all of these seem to apply to my system (which is FreeBSD).

PR #479, opened 2019-09-20
Fix URN response exception handling

eduard-bagdasaryan:
CID 1453857: Error handling issues (UNCAUGHT_EXCEPT)

Due to various Store deficiencies, storeUnregister() might call swapout
code which Broadcast()s and throws Ipc::OneToOneUniQueue::ItemTooLarge.


rousskov:
I added an explanation how this cleanup code can lead, at least in theory, to the code that throws Ipc::OneToOneUniQueue::ItemTooLarge.

PR #480, opened 2019-09-24
Smarter auth_param utf8 handling, including CP1251 support

Sergey-Kirpa:
When forwarding authentication credentials to authentication helpers:

* With auth_param utf8 parameter: Squid assumed that the received
  credentials are encoded with Latin-1 (and re-encoded them with UTF-8).
  This assumption is wrong for Internet Explorer running with CP1251
  regional settings. Now Squid uses HTTP Accept-Language request header
  to guess the received credentials encoding (Latin-1, CP1251, or UTF-8)
  and converts the first two encodings into UTF-8.

* Without auth_param utf8 parameter: No changes. Squid sends credentials
  in their original encoding, only applying RFC 1738 escaping on top.

Chrome and Firefox should not be affected because they always use UTF-8
encoding when sending authentication credentials.


src/auth/basic/Config.cc
```diff
@@ -148,8 +149,106 @@ authenticateBasicStats(StoreEntry * sentry)
         basicauthenticators->packStatsInto(sentry, "Basic Authenticator Statistics");
 }
 
+inline size_t
+inlineUTF8SequenceLengthNonASCII(char b0)
```
rousskov: Please avoid specifying inlining preference in the function name. Focus on what the function does. BTW, it would be nice to add brief descriptions to these new helper functions.

src/auth/basic/Config.cc
```diff
@@ -148,8 +149,106 @@ authenticateBasicStats(StoreEntry * sentry)
...
+}
+
+/*
+ * Returns whether a UTF-8 string is legal or not.
+ */
+static bool
+isLegalUTF8String(const char *source, const char *sourceEnd) {
+    while (source != sourceEnd) {
+        int length = inlineUTF8SequenceLength(*source);
+        if (length > sourceEnd - source || !isLegalUTF8((const unsigned char*) source, length))
```
rousskov: Avoid C casts, even in reinterpret_cast cases:

```suggestion
        if (length > sourceEnd - source || !isLegalUTF8(reinterpret_cast<const unsigned char*>(source), length))
```

src/auth/basic/Config.cc
```diff
@@ -148,8 +149,106 @@ authenticateBasicStats(StoreEntry * sentry)
...
+    return true;
+}
+
+/*
+ * Returns whether a UTF-8 string is legal or not.
+ */
+static bool
+isLegalUTF8String(const char *source, const char *sourceEnd) {
+    while (source != sourceEnd) {
+        int length = inlineUTF8SequenceLength(*source);
```
rousskov: To avoid silent type conversion bugs like this one and to help the reader/compiler:
* Please use `const` in every variable declaration that allows it.
* Please use `auto` in every variable declaration that allows it.

src/auth/basic/Config.cc
```diff
@@ -148,8 +149,106 @@ authenticateBasicStats(StoreEntry * sentry)
...
+static bool
+isCP1251EncodingAllowed(const char *httpLangHeader)
+{
+    if (httpLangHeader == NULL)
+        return false;
+
+    const char *p = httpLangHeader;
+
+    while (*p) {
+	if ((strncmp(p, "ru", 2) == 0
```
rousskov: Missing comment explaining the relationship between two seemingly random country codes and CP1251.

src/auth/basic/Config.cc
```diff
@@ -148,8 +149,106 @@ authenticateBasicStats(StoreEntry * sentry)
...
+	     || strncmp(p, "uk", 2) == 0)
+	    && !xisalpha(p[2]))
+	    return true;
+
+	if ((p = strchr(p, ',')) == NULL)
```
rousskov: Please use existing Squid APIs such as strListGetItem() for iterating list header fields.

rousskov:
OK to test

yadij:
You may find it better to add support for Squid to advertise charset="UTF-8" (see RFC 7235, 7616 and 7617) when the auth_param utf8 option is configured. Have it selectively skip the custom utf8 conversion when the client responds with UTF-8 credentials.

Sergey-Kirpa:
> You may find it better to add support for Squid to advertise charset="UTF-8" (see RFC 7235, 7616 and 7617) when the auth_param utf8 option is configured. Have it selectively skip the custom utf8 conversion when the client responds with UTF-8 credentials.

Unfortunately, this will not work. In my case, the response is Proxy-Authenticate. The charset="UTF-8" is defined only for WWW-Authenticate response, and not for Proxy-Authenticate.

yadij:
> The charset="UTF-8" is defined only for WWW-Authenticate response, and not for Proxy-Authenticate.

This is not true. charset is defined as a parameter of the scheme itself. It is usable in both server and proxy auth headers whenever that scheme parameter list is sent.

The issue would be determining when the credential response is UTF-8. This PR adding auto-detect logic already has to handle this problem.

src/auth/basic/Config.cc
```diff
@@ -176,6 +295,18 @@ Auth::Basic::Config::decodeCleartext(const char *httpAuthHeader)
     if (base64_decode_update(&ctx, &dstLen, reinterpret_cast<uint8_t*>(cleartext), srcLen, eek) && base64_decode_final(&ctx)) {
         cleartext[dstLen] = '\0';
 
+        if (!isLegalUTF8String(cleartext, cleartext + dstLen)) {
+            SBuf str;
+
+            if (isCP1251EncodingAllowed(request))
```
yadij: This is converting the nominally 'is allowed' test function into a 'force to happen' test because the else clause Latn-1 conversion is skipped whenever cp1251 is merely _allowed_.

As Alex mentioned relying on languages is somewhat problematic. While this may work for you it is not guaranteed to be like that everywhere. At least for sr-* we have sr-Latn alphabet translations for our error pages because that alphabet is actively being used by large populations (plural, some outside the country itself).

The same might be true for any of the languages, just we do not have evidence supporting.  That means when the cp-1251 does not succeed we still may need to try the Latin-1 conversion.

Sergey-Kirpa:
> > The charset="UTF-8" is defined only for WWW-Authenticate response, and not for Proxy-Authenticate.
> 
> This is not true. charset is defined as a parameter of the scheme itself. It is usable in both server and proxy auth headers whenever that scheme parameter list is sent.

Perhaps we are talking about different things. I mean that these responses are different, as shown in the "Directives" section:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authenticate#Directives
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate#Directives

In my case, the response is always Proxy-Authenticate. Even if I set charset="UTF-8", Internet Explorer will ignore it anyway.

Sergey-Kirpa:
> HTTP is explicitly ASCII based and Squid is supposed to be receiving ASCII (Latin-1) credentials from the headers - yes that is somewhat bad, but HTTP is old and contains systems relying on that behaviour.

You're right. HTTP is explicitly ASCII. But the "credentials" string is base64 encoded.
According to the RFC 7617, user identifiers and passwords containing characters outside the US-ASSIA are allowed. However, there may be compatibility issues.

src/auth/SchemeConfig.cc
```diff
@@ -51,7 +51,7 @@ Auth::SchemeConfig::CreateAuthUser(const char *proxy_auth, AccessLogEntry::Point
         config->keyExtras->assemble(rmb, al, 0);
     }
 
-    return config->decode(proxy_auth, rmb.hasContent() ? rmb.content() : NULL);
+    return config->decode(proxy_auth, al->request, rmb.hasContent() ? rmb.content() : NULL);
```
yadij: Use nullptr instead of NULL on touched lines please.

src/auth/basic/Config.cc
```diff
@@ -32,6 +32,7 @@
 #include "Store.h"
 #include "util.h"
 #include "wordlist.h"
+#include "errorpage.h"
```
yadij: Includes need to be in alphabetical order, exception being the squid.h file.

yadij:
Just some polish remaining. I will approve when that is in.

src/auth/basic/Config.cc
```diff
@@ -176,6 +306,18 @@ Auth::Basic::Config::decodeCleartext(const char *httpAuthHeader)
     if (base64_decode_update(&ctx, &dstLen, reinterpret_cast<uint8_t*>(cleartext), srcLen, eek) && base64_decode_final(&ctx)) {
         cleartext[dstLen] = '\0';
 
+        if (utf8 && !isLegalUTF8String(cleartext, cleartext + dstLen)) {
+            SBuf str;
```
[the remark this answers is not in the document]
Sergey-Kirpa: Thank! You helped find my typo.

src/auth/basic/Config.cc
```diff
@@ -148,8 +154,132 @@ authenticateBasicStats(StoreEntry * sentry)
...
+        return 2;
+    if ((b0 & 0xF0) == 0xE0)
+        return 3;
+    if ((b0 & 0xF8) == 0xF0)
+        return 4;
+    return 0;
+}
+
+/*
+ * Returns the length of UTF-8 sequence.
```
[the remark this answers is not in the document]
Sergey-Kirpa: You're right. This is an unusual behavior that is misleading. Zero return is a special case when an invalid code point was specified. I corrected the description and added a comment for this case.

src/auth/basic/Config.cc
```diff
@@ -148,8 +154,132 @@ authenticateBasicStats(StoreEntry * sentry)
...
+ * definition of UTF-8 goes up to 4-byte sequences.
+ */
+static bool
+isLegalUTF8(const unsigned char* source, int length)
+{
+    unsigned char a;
+    const unsigned char* srcptr = source + length;
+    switch (length) {
+        default: return false;
+        // Everything else falls through when "true"...
```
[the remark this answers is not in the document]
Sergey-Kirpa: I signed up for Coverity but I still do not understand how it works. Perhaps I need to add a test project consisting only of this function to see what warnings will be generated.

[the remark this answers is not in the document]
rousskov: Coverity free scan service is rather restricted/limited. You may not have enough access rights to add projects/branches/builds. YMMV, but I could not figure out how to add a given branch or commit for scanning.

If nobody else can (temporary) add your branch or merge commit, then I think the best strategy here is to review Coverity [documentation](https://scan5.coverity.com/doc/en/cov_checker_ref.html#static_checker_MISSING_BREAK) for fall-through switch statements, follow their instructions for disabling checkers via source code comments, and then watch for new Coverity defects when the code gets officially merged. If necessary, follow up with a small PR fixing new defects.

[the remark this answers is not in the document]
rousskov: > review Coverity [documentation](https://scan5.coverity.com/doc/en/cov_checker_ref.html#static_checker_MISSING_BREAK) for fall-through switch statements, follow their instructions for disabling checkers via source code comments

If I am interpreting that documentation correctly, you may not even need special comments in some cases (e.g., the `Start and end on the same line.` case).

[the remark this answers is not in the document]
Sergey-Kirpa: Everything is fine. I checked in a test project. Coverity scanner does not see any defects in this function. No corrections are needed.

src/auth/basic/Config.cc
yadij: FTR the Cp1251ToUtf8 was updated to not have error cases other than NULL input. The output may be garbage but there is no longer a need for and should not be any followup conversion now.

rousskov:
> @yadij requested a review from @rousskov 

Acknowledged. I hope to be able to re-review this PR within a few days.

yadij:
> @yadij, can you suggest a better location for the moved files, either within src/ or, preferably, within one of the existing convenience libraries? Or should we create a new src/utf/ or src/char-encodings/ convenience library?

Since it needs SBuf we cannot leave it in lib/ with the other encoders and libbase is no better than libsquid in regards to dependencies. If you are going to insist on it having a different location than libsquid then src/auth/libauth.la would be the best alternative IMO.

rousskov:
> Since it needs SBuf we cannot leave it in lib/ with the other encoders

`lib/` is not the right place for code unrelated to compatibility wrappers and alike, regardless of SBuf usage.

> and libbase is no better than libsquid in regards to dependencies.

`libbase` is not the right place for encoding-handling code, especially code that very few Squid areas use.

> If you are going to insist on it having a different location than libsquid then src/auth/libauth.la would be the best alternative IMO.

`libauth` feels like a strange choice for encoding-handling code that is not (or should not be) specific to authentication but it is better than libsquid.

Please note that I am OK with leaving the files in `src/` if we cannot find (and do not want to create) a better place. The vast majority of files in `src/` are not in libsquid, of course.

yadij:
@Sergey-Kirpa it seems src/auth/ is the location to ink the .cc fie into. Any future PR extending the encoding to other uses can revisit that choice later.

rousskov:
> it seems src/auth/ is the location

No objection from me.

Sergey-Kirpa:
> @Sergey-Kirpa, please retest the current code and, if you can, check that Coverity is still happy about it.

Coverity - OK (No defects found).

rousskov:
@Sergey-Kirpa, I polished the PR description a little, but it still does not tell the reader what has _changed_. Can you describe the change in Squid functionality (_not_ code) in your own words? For example, perhaps Squid previously assumed that all credentials are X-encoded, and now Squid can auto-detect most Y- and Z-encoded credentials as well (with X remaining the default/assumed encoding).

Sergey-Kirpa:
I updated the PR description.

rousskov:
> I updated the PR description.

Excellent, thank you! I polished it a little. Please check that I did not misrepresent anything and did not lose any important info. Adjust as needed, of course. See [Anubis](https://github.com/measurement-factory/anubis#commit-message) for formatting requirements.

One potentially important aspect that you should double check is whether the changes apply to Basic authenticaiton only. You original text implied that, but cf.data.pre appears to imply that the utf8 parameter applies to both Basic and Digest authentication (at least):

	"keep_alive" on|off
		...
		For Basic and Digest this parameter is ignored.

	"utf8" on|off
		...
		For NTLM and Negotiate this parameter is ignored.

Who is right? Your original PR description or cf.data.pre? I can adjust the PR description further if you confirm that the changes are specific to Basic authentication only.

Sergey-Kirpa:
> I can adjust the PR description further if you confirm that the changes are specific to Basic authentication only.

Yes, these changes apply only to the Basic authentication. The Digest authentication scheme will work without changes. I'm not sure if this feature is needed for the Digest authentication. If it needs to be done too, I can do it.

rousskov:
> I'm not sure if this feature is needed for the Digest authentication.

I am also not sure, but I suspect that Squid should be consistent across all schemes that support the `utf8` configuration parameter. I know that Squid currently applies latin1_to_utf8() when sending the username component (and only that component) to the Digest authentication helper. See Auth::Digest::UserRequest::startHelperLookup().

This PR should either properly extend encoding support to Digest or document that the support is limited to the Basic scheme (for some valid reason).

@Sergey-Kirpa, AFAICT, your work was motivated by IE using non-latin1 encoding for Basic authentication. Can you tell what encoding IE uses for Digest authentication?

@yadij, do you know whether the encoding concerns apply to _Digest_ helper requests? And, if yes, whether that encoding support should apply to any components other than the username component?

yadij:
> @yadij, do you know whether the encoding concerns apply to _Digest_ helper requests? And, if yes, whether that encoding support should apply to any components other than the username component?

My understanding of the issue is that it is only relevant to data input received from the UA popups. If the UA is not correctly re-encoding we are filling that processing gap.

For Digest the username should be the only user input value. Password may also be affected, but that we can't do anything about. At worst it would mean completely non-working Digest auth for that UA.

rousskov:
@Sergey-Kirpa, my interpretation of @yadij answer is that we should add encoding support to Digest (username) as well. If you agree, please add the necessary code. I can help with adjusting `cf.data.pre` documentation.

src/auth/digest/Config.cc
```diff
@@ -807,8 +808,15 @@ Auth::Digest::Config::decode(char const *proxy_auth, const char *aRequestRealm)
         switch (t) {
         case DIGEST_USERNAME:
             safe_free(username);
-            if (value.size() != 0)
-                username = xstrndup(value.rawBuf(), value.size() + 1);
+            if (value.size() != 0) {
+                const auto v = value.termedBuf();
+		if (utf8 && !isValidUtf8String(v, v + value.size())) {
```
yadij: Please remove tabs from code indentation. More lines below too

```suggestion
        if (utf8 && !isValidUtf8String(v, v + value.size())) {
```

src/auth/toUtf.cc
```diff
@@ -0,0 +1,174 @@
...
+ * Copyright (C) 1996-2019 The Squid Software Foundation and contributors
+ *
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#include "squid.h"
+#include "sbuf/SBuf.h"
+#include "toUtf.h"
```
yadij:
```suggestion
#include "auth/toUtf.h"
```

src/auth/digest/Config.cc
```diff
@@ -807,8 +808,15 @@ Auth::Digest::Config::decode(char const *proxy_auth, const char *aRequestRealm)
         switch (t) {
         case DIGEST_USERNAME:
             safe_free(username);
-            if (value.size() != 0)
-                username = xstrndup(value.rawBuf(), value.size() + 1);
+            if (value.size() != 0) {
+                const auto v = value.termedBuf();
+                if (utf8 && !isValidUtf8String(v, v + value.size())) {
+                    auto str = isCP1251EncodingAllowed(request) ? Cp1251ToUtf8(v) : Latin1ToUtf8(v);
+                    username = xstrndup(str.rawContent(), str.length() + 1);
```
rousskov: This xstrndup() call is buggy: xstrndup() has a terrible API that leads to both buggy-looking code like `xstrndup(n_characters, n+1)` and actual bugs like feeding non-terminated rawContent() output to xstrndup() which calls termination-relying length() on its input!

```suggestion
                    username = xstrdup(str.c_str());
```

1. The easiest fix is to use `xstrdup(str.c_str())` as depicted by the suggestion above.
2. An arguably better solution would be to add and use an SBuf::c_strdup() method that has no API and implementation flaws of xstrndup().
3. The arguably best solution would be to fix xstrndup() itself, but that is a lot of work because it requires fixing its API and, hence, reviewing (and possibly adjusting) all 30+ existing callers.

I recommend using either solution 1 or solution 2. Your call. FWIW, I can (slowly) help with solution 2.

P.S. Yes, there is a similar ugly old xstrndup() call that uses rawBuf(). That call only looks buggy because String::rawBuf() output happens to be zero-terminated. And yes, the String::rawBuf() documentation kind of implies otherwise.

rousskov:
FYI: You do not need to rebase the PR branch just to keep it up to date. [Anubis](https://github.com/measurement-factory/anubis#readme) will essentially do that for you when the time comes. You can rebase if you have to rebase for other reasons, of course, keeping in mind that forced pushes often confuse GitHub as far as individual change request state is concerned.

src/auth/digest/Config.cc
```diff
@@ -807,8 +808,14 @@ Auth::Digest::Config::decode(char const *proxy_auth, const char *aRequestRealm)
         switch (t) {
         case DIGEST_USERNAME:
             safe_free(username);
-            if (value.size() != 0)
+            if (value.size() != 0) {
+                const auto v = value.rawBuf();
+                if (utf8 && !isValidUtf8String(v, v + value.size())) {
+                    auto str = isCP1251EncodingAllowed(request) ? Cp1251ToUtf8(v) : Latin1ToUtf8(v);
+                    value.limitInit(str.rawContent(), str.length());
```
yadij: Jenkins tests are currently failing on this line with:  error: 'class String' has no member named 'limitInit'

```suggestion
                    value = SBufToString(str);
```

rousskov:
> I will post cf.data.pre changes.

This is still on my side. I have not forgotten and hope to work on this later next week. Feel free to take over this part, of course.

rousskov:
> I will post cf.data.pre changes.

Done in 467be37.

rousskov:
@Sergey-Kirpa, please look through my last three commits to make sure I did not screw anything up:

* 525e625 fixup: Undo 09ae483. Uniformity is dangerous for this code.
* 92075ad fixup: SourceFormat Enforcement
* 467be37 fixup: Synced cf.data.pre documentation

@yadij, this PR is on your side now.

Sergey-Kirpa:
> @Sergey-Kirpa, please look through my last three commits to make sure I did not screw anything up:

Everything looks fine.

PR #482, opened 2019-09-27
Docs: update master branch to v6

yadij:


yadij:
There are always more changes in flight "any day now". We need to make progress with v5. It is over a year "late" now already and meets all the criteria for a first beta.

rousskov:
> There are always more changes in flight "any day now".

Yes, but incoming changes are not such a big problem. One of the key problems here is that there are old changes blocked by you for no good reason. The number of such changes grows, and there is no recourse. I hope to find a way out of this ridiculous situation by making the actions you see as important conditional on fulfilling the reviewer responsibilities you tend to assume and then ignore.

> We need to make progress with v5. It is over a year "late" now already and meets all the criteria for a first beta.

While I do not care much about made-up deadlines and arbitrary criteria, I am willing to play along as long as we are making reasonable progress in master. Currently, we are not. Let's clear the backlog and then branch. Whatever v5-worthy changes we can get into master by then will automatically go into v5 with no overhead. AFAICT, if we cooperate, the existing blocked PRs and most of the changes already announced on squid-dev can be in master (and v5 can then be branched) before this year ends.

yadij:
Now that v5 has branched this needs merging to produce the v6 daily snapshots tracking master commits. Please remove the block on this PR.

PR #483, opened 2019-09-27
Report context of level-0/1 cache.log messages

rousskov:
Most runtime level-0/1 cache.log messages do not carry information
sufficient to identify the transaction that caused the problem. Admins
are forced to guess the transaction based on message timestamp and, if
they are lucky to get one, request URL. The associated triage and
mitigation delay costs are often significant and can be huge, especially
when administering busy proxies in complex deployment environments.

The focus of this change is associating level-0/1 cache.log messages
with access.log records, but the same API is also used for associating
master transactions with (sections of) debugging cache.log messages.

Since level-0/1 messages are rare, association code usually wastes
resources. This performance overhead is reduced by saving pointers to
the existing transaction information (e.g., ALE). ALE gives access to
HttpRequest and MasterXaction (where available), is available in
contexts where HttpRequest and MasterXaction do not exist, and (unlike
HttpRequest) represents the whole master transaction rather than its
(often repeated) component.

CodeContext::Current() represents the current master transaction (or a
similar "primary processing task" context). A new context is created
when the master transaction (or a similar "primary processing task")
starts. Context changes usually happen in low-level transaction-unaware
callback-calling "context switching" code such as DoSelect().

The vast majority of AsyncCalls, including AsyncCall-based callbacks,
should run in their creator's context. This association is easily
automated. The primary difficulty is in handling C-style typeless calls
that prohibit context storage and restoration automation:

* In our design, the "context switching" code is ultimately responsible
  for associating the being-saved callback with the current code context
  and for restoring CodeContext::Current() when calling that callback.

* An alternative design would task the higher-level callback creator and
  callback recipient with saving/restoring CodeContext::Current(). That
  design is inferior because there are a lot more callback creators and
  recipients than "context switchers". That alternative would require a
  lot more manual changes.

The code context remains unknown if the context creator is not
instrumented to set CodeContext::Current(). TODO: Instrument ICP query
listener. Check for others.

The code context gets forgotten if the context switcher dealing with
C-style callbacks does not remember and restore CodeContext::Current().
TODO: Instrument remaining DoSelect()s, event.cc, as well as DNS, ICP,
HTCP, and IPC listeners/handlers. Check for others.

This change covers epoll DoSelect(), TcpAcceptor, ConnStateData, and SMP
disk I/O (IpcIoFile). It already annotates several level-0/1 messages
and significantly improves complex debugging. The remaining
instrumentation TODOs are likely to use similar techniques.

Squid might report wrong context until all context switchers are
instrumented, but the vast majority of uninstrumented cases result in
benign loss of context knowledge rather than mis-attribution. No design
guarantees correct attribution until all C-style callbacks are gone.

TODO: Remove legacy ctx_enter()/exit() debugging that covers very little
while suffering from worse mis-attribution problems.

Also log "-" instead of the made-up method "NONE".

### DIFF
--- a/src/AccessLogEntry.cc
+++ b/src/AccessLogEntry.cc
@@ -49,13 +49,16 @@ AccessLogEntry::getLogClientIp(char *buf, size_t bufsz) const
 SBuf
 AccessLogEntry::getLogMethod() const
 {
+    static const SBuf dash("-");
     SBuf method;
     if (icp.opcode)
         method.append(icp_opcode_str[icp.opcode]);
     else if (htcp.opcode)
         method.append(htcp.opcode);
-    else
+    else if (http.method)
         method = http.method.image();
+    else
+        method = dash;
     return method;
 }
 
--- a/src/AccessLogEntry.cc
+++ b/src/AccessLogEntry.cc
@@ -120,14 +120,14 @@ AccessLogEntry::~AccessLogEntry()
 #endif
 }
 
-std::ostream &
-AccessLogEntry::briefCodeContext(std::ostream &os) const
+ScopedId
+AccessLogEntry::codeContextGist() const
 {
     if (request) {
         if (const auto &mx = request->masterXaction)
-            os << mx->id;
+            return mx->id.detach();
     }
-    return os;
+    return ScopedId();
 }
 
 std::ostream &
--- a/src/AccessLogEntry.cc
+++ b/src/AccessLogEntry.cc
@@ -120,6 +120,48 @@ AccessLogEntry::~AccessLogEntry()
 #endif
 }
 
+std::ostream &
+AccessLogEntry::briefCodeContext(std::ostream &os) const
+{
+    if (request) {
+        if (const auto &mx = request->masterXaction)
+            os << mx->id;
+    }
+    return os;
+}
+
+std::ostream &
+AccessLogEntry::detailCodeContext(std::ostream &os) const
+{
+    // TODO: Consider printing all instead of the first most important detail.
+
+    if (request) {
+        if (const auto &mx = request->masterXaction)
+            return os << Debug::Extra << "current master transaction: " << mx->id;
+    }
+
+    // provide helpful details since we cannot identify the transaction exactly
+
+    if (tcpClient)
+        return os << Debug::Extra << "current from-client connection: " << tcpClient;
+    else if (!cache.caddr.isNoAddr())
+        return os << Debug::Extra << "current client: " << cache.caddr;
+
+    const auto optionalMethod = [this,&os]() {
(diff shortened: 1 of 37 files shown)


rousskov:
> Also log "-" instead of the made-up method "NONE".

We can extract this small change into a separate PR. The change was necessary for this PR (to avoid reporting unknown methods to cache.log), but it may be a good idea to merge it separately because it affects access.log records. You can see this change in isolation at https://github.com/squid-cache/squid/pull/483/commits/5e0fec48a122689fd372d8aa7e555fc7f442a3e2

Please let me know if you want this to be extracted.

rousskov:
Here is an example of transaction association with one of the level-0/1 messages:

```
2019/09/27 09:30:58.887| WARNING: HTTP: Invalid Response: Bad header encountered from ...
    current master transaction: master63
```

The reported transaction ID can be easily associated with the access-logged record by adding %master_xaction to logformat:

```
logformat xsquid xsquid kid${process_number}/mx%master_xaction %err_code/%err_detail ...
```

This output example is a good illustration for the following user interface question: 

* @squid-cache/commiters, Should Squid report as many transaction details as it can when reporting context associated with a level-0/1 cache.log message?

The current implementation attempts to report enough info to associate the message with an access.log record (assuming `%master_xaction` was added to logformat) and stops there. The alternative would be to report more transaction details. For example, instead of reporting just the master transaction ID, Squid could, in many cases, report the request method, URL, TCP client address, and other ALE details.

From code point of view, the difference is usually small: Removing `return` from an `if haveX` statement may be enough in many cases. For example, see https://github.com/squid-cache/squid/pull/483/files#diff-ba6d6228d5ea9a29828df53119db821aR135

yadij:
Please submit 5e0fec48a122689fd372d8aa7e555fc7f442a3e2 as a separate PR. It changes access.log rather than cache.log this PR is apparently about.

rousskov:
> Please submit 5e0fec4 as a separate PR.

Done: #486

rousskov:
@yadij, I will merge these changes soon if there are no last-minute objections.

> @squid-cache/commiters, should Squid report as many transaction details as it can when reporting context associated with a level-0/1 cache.log message?

The [question](https://github.com/squid-cache/squid/pull/483#issuecomment-535995013) still stands, but the code can be adjusted after the merge, so it is not a blocker. On the other hand, the sooner we answer that question, the fewer changes we (and Squid admins) would have to make long-term.

P.S. We have discovered one nice side-effect of this PR changes while working on a completely unrelated Factory project: The new "transaction scope" API enables measuring transaction-level CPU usage, which can be extremely helpful in narrowing down suspects when a busy production Squid suddenly starts using 100% CPU cycles (and then magically recovers).

PR #484, opened 2019-09-27
Add GeneratingCONNECT step for the existing at_step ACL

chtsanti:
The new step allows admins to customize CONNECT request generation using
request_header_access and request_header_replace. As detailed below,
matching the request method does not work for those purposes.

request_header_access controls what HTTP request headers Squid _sends_.
However, this directive, like most other ACL-driven Squid directives,
uses the received request, _not_ the being-formed to-be-sent request.
This can be confusing, but it is the correct design/behavior.

When Squid is about to send a CONNECT request to a cache peer, what
_received_ request should request_header_access ACLs look at? There are
several ways to answer that question:

1. The received CONNECT request (or, for intercepted TLS connections,
   its faked equivalent). This is how the current CONNECT-sending code
   works when establishing the tunnel for a not-yet-bumped client
   connection. Problems:
    
    1. The CONNECT request received by Squid was not meant for the cache
       peer. It was meant specifically for this Squid. While it may have
       info useful for request_header_access checks, it is conceptually
       wrong to think of whats happening as "forwarding of that received
       CONNECT to the cache peer". Unlike GET requests, the CONNECT
       request that this Squid will send to the cache peer is dedicated
       to the Squid-peer connection. It is generated, not forwarded.
    
    2. That CONNECT request may have been received a long time ago, and
       Squid may have forwarded many bumped GET requests since then. It
       feels strange to consult such an old/"disconnected" message.
2. The received (and bumped) GET request. This is how the current
   CONNECT-sending code works when establishing the tunnel for an
   already bumped client connection. Problem:
    
    1. Squid is about to send a generated CONNECT request, not to
       forward a received GET request (the latter will happen later,
       after the CONNECT transaction). The two requests may differ a
       lot. Using a GET request when generating a CONNECT request is
       confusing: "Why is my CONNECT method ACL does not match when
       Squid sends a CONNECT request?!"

3. No request. Problems:
    
    1. Some old configurations that use request-specific ACLs with
       request_header_access will generate runtime "missing request"
       warnings and may fail to work correctly.
    
    2. Extra admin work is required to store request-specific
       information as connection annotations that request_header_access
       ACLs can still access.
    
    3. Conceptually, there is a request that Squid is establishing this
       CONNECT tunnel for. Squid will access-log that request. Hiding
       that information from ACLs feels odd/wrong. And some of that info
       would still be accessible to ACLs (via ALE/etc.). This hiding
       does not really hide all of the details.

Our solution preserves what received request Squid is looking at. Items
1 and 2 above still apply (depending on the configuration and on the
request being processed), with their unique problems intact. Those
problems are not as bad as the problems associated with the item 3!

The at_step ACL was added for SslBump but, IIRC, we knew that it may
eventually cover other request processing steps. Generating a CONNECT
request is one of those other steps.

### DIFF
--- a/src/acl/AtStep.cc
+++ b/src/acl/AtStep.cc
@@ -16,24 +16,19 @@
 #include "ssl/ServerBump.h"
 
 int
-ACLAtStepStrategy::match (ACLData<int> * &data, ACLFilledChecklist *checklist)
+ACLAtStepStrategy::match (ACLData<XactionSteps> * &data, ACLFilledChecklist *checklist)
 {
     Must(checklist->request);
     Must(checklist->request->masterXaction);
-    if (checklist->request->masterXaction->generatingConnect && data->match(ACLAtStepData::atStepGeneratingConnect))
+    if (checklist->request->masterXaction->generatingConnect && data->match(xaStepGeneratingConnect))
         return 1;
 
 #if USE_OPENSSL
-    static std::map<Ssl::BumpStep, ACLAtStepData::AtStepValues> BumpAtStepMap= {
-        {Ssl::bumpStep1, ACLAtStepData::atStepSslBump1},
-        {Ssl::bumpStep2, ACLAtStepData::atStepSslBump2},
-        {Ssl::bumpStep3, ACLAtStepData::atStepSslBump3}
-    };
     Ssl::ServerBump *bump = NULL;
     if (checklist->conn() != NULL && (bump = checklist->conn()->serverBump()))
-        return data->match(BumpAtStepMap[bump->step]);
+        return data->match(bump->step);
     else
-        return data->match(BumpAtStepMap[Ssl::bumpStep1]);
+        return data->match(xaStepTlsBump1);
 #endif
     return 0;
 }
--- a/src/acl/AtStep.cc
+++ b/src/acl/AtStep.cc
@@ -25,10 +25,10 @@ ACLAtStepStrategy::match(ACLData<XactionStep> * &data, ACLFilledChecklist *check
     if (checklist->conn() != NULL && (bump = checklist->conn()->serverBump()))
         return data->match(bump->step);
     else
-        return data->match(xstepTlsBump1);
+        return data->match(XactionStep::tlsBump1);
 #endif
 
-    if (data->match(xstepGeneratingConnect)) {
+    if (data->match(XactionStep::generatingConnect)) {
         if (!checklist->request)
             return 0; // we have warned about the missing request earlier
 
--- a/src/acl/AtStep.cc
+++ b/src/acl/AtStep.cc
@@ -13,7 +13,9 @@
 #include "acl/FilledChecklist.h"
 #include "client_side.h"
 #include "http/Stream.h"
+#if USE_OPENSSL
 #include "ssl/ServerBump.h"
+#endif
 
 int
 ACLAtStepStrategy::match (ACLData<XactionSteps> * &data, ACLFilledChecklist *checklist)
--- a/src/acl/AtStep.cc
+++ b/src/acl/AtStep.cc
@@ -22,7 +22,7 @@ ACLAtStepStrategy::match (ACLData<XactionStep> * &data, ACLFilledChecklist *chec
 {
     Must(checklist->request);
     Must(checklist->request->masterXaction);
-    if (checklist->request->masterXaction->generatingConnect && data->match(xaStepGeneratingConnect))
+    if (checklist->request->masterXaction->generatingConnect && data->match(xstepGeneratingConnect))
         return 1;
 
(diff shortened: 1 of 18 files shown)


src/acl/AtStep.cc
```diff
@@ -8,25 +8,40 @@
...
-}
+    if (checklist->conn() != NULL &&
+        (bump = checklist->conn()->serverBump()) &&
+        data->match(bump->step))
+        return 1;
+#endif
+
+    if (checklist->request && data->match(XactionStep::generatingConnect)) {
+        if (!checklist->request->masterXaction)
+            debugs(28, DBG_IMPORTANT, "at_step GeneratingCONNECT ACL is missing master transaction info. Assuming mismatch.");
```
rousskov: The `Assuming mismatch` part a lie -- the code may still return 1 which means it does not simply "assume mismatch". If this check cannot be moved lower, then we have to remember that `generatingConnectMatchedWithoutMx`, wait until we know the outcome of the tlsBump1 match() below, and only then, _if_ that outcome is falsy and generatingConnectMatchedWithoutMx is true, display this warning (and return 0).

The above suggestion assumes that the match(x) and match(y) can both report a match for the same `data` object. Otherwise, it does not make sense to wait until that second match() call before returning 0.

BTW, please prefix this with `WARNING: `, for consistency (and clarity) sake.

src/acl/AtStep.cc
```diff
@@ -8,25 +8,40 @@
...
-#endif /* USE_OPENSSL */
+#if USE_OPENSSL
+    // We need the following to cover the case of bumping at SslBump1 step
+    // where the connStateData::serverBump() is not build yet.
+    // The following also has the meaning that if no bumping preformed
+    // or a client-first bumping is applied then the request is remaining
+    // at SslBump1 bumping processing step.
+    return data->match(XactionStep::tlsBump1);
+#endif
+}
```
rousskov: Missing return when not `USE_OPENSSL`.

src/acl/AtStep.cc
chtsanti: > The Assuming mismatch part a lie -- the code may still return 1 which means it does not simply "assume mismatch".

I saw it but I consider it no huge problem, because the generatingConnect may missmatch but the step1 may match, in the case they are both used at the same acl line, but the admin still needs to now that the generatingConnect is problematic here.

Anyway the code is fixed now.

> If this check cannot be moved lower, 

The check can be moved under the check of tlsBump1 I was wrong about. Now the related code does the following:
 - If there is a ServerBump structure retrieve the bumping step from here.  If the acl matches the current bumping step return 1;
 - else assume that we are at bumping step1 and check if acl matches the bumping step1
 - else check if the acl matches the generatingConnect. etc.

Commit ece98b98.

src/acl/AtStep.cc
```diff
@@ -8,25 +8,46 @@
...
+    if (data->match(XactionStep::tlsBump1))
+        return 1;
+#endif
+
+    if (data->match(XactionStep::generatingConnect)) {
+        if (!checklist->request)
+            return 0; // we have warned about the missing request earlier
+
+        if (!checklist->request->masterXaction) {
+            debugs(28, DBG_IMPORTANT, "at_step GeneratingCONNECT ACL is missing master transaction info. Assuming mismatch.");
```
rousskov: If there are no known cases in Squid code where a request might be missing a master transaction, let's prefix this with a `BUG:`.

```suggestion
            debugs(28, DBG_IMPORTANT, "BUG: at_step GeneratingCONNECT ACL is missing master transaction info. Assuming mismatch.");
```

chtsanti: commit 3d5a7c02

src/acl/AtStep.cc
```diff
@@ -8,25 +8,46 @@
...
+        (bump = checklist->conn()->serverBump()) &&
+        data->match(bump->step))
+        return 1;
+
+    // We need the following to cover the case of bumping at SslBump1 step
+    // where the connStateData::serverBump() is not build yet.
+    // The following also has the meaning that if no bumping preformed
+    // or a client-first bumping is applied then the request is remaining
+    // at SslBump1 bumping processing step.
+    if (data->match(XactionStep::tlsBump1))
```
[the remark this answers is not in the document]
chtsanti: Fixed in commit e6180454
I used your template.

chtsanti:
@rousskov 

I made the fixes you are requested. Please re-review

rousskov:
I polished the code a bit (572130e and 8de1873) but did not test those changes. If you are happy with the end result, please feel free to clear for merge -- I do not plan on reviewing this further.

chtsanti:
I rebased branch to latest master.
I add the flag M-cleared-for-merge.
I hope it is ok.

PR #487, opened 2019-09-30
Bug 4989: Leaking StoreEntry objects on Cache Digest rebuilds

craiggowing:
When writing a newly generated Cache Digest to cache, Squid relied on a
cache key collision to purge the old digest entry. Since 4310f8b, the
collision resolution method -- forcePublicKey() -- leaked an idle (i.e.
lock_count=0) digest entry. If Squid still had unlocked entries lying
around, then the problem could extend to clashes unrelated to Digests.

Until 4310f8b, StoreEntry::forcePublicKey() called setPrivateKey()
before releasing the old entry. That explicit call was wasteful in many
cases, but, unbeknownst to its removal authors, it allowed release() to
destroy an idle Cache Digest entry by effectively disabling the
ENTRY_SPECIAL hack in StoreEntry::locked().

This change removes the ENTRY_SPECIAL hack in StoreEntry::locked(),
addressing an old TODO. The two ENTRY_SPECIAL creators (icons and Cache
Digests) now lock their entries to prevent their unwanted destruction.

Also explicitly release the old Cache Digest entry (instead of relying
on the implicit key collision) to avoid the unchecked assumption that
the Cache Digest key never changes.

### DIFF
--- a/src/Store.h
+++ b/src/Store.h
@@ -128,7 +128,7 @@ public:
     /// TODO: Rename and make private so only those two methods can call this.
     bool checkCachable();
     int checkNegativeHit() const;
-    int locked() const;
+    int locked() const { return lock_count; }
     int validToSend() const;
     bool memoryCachable(); ///< checkCachable() and can be cached in memory
 
--- a/src/mime.cc
+++ b/src/mime.cc
@@ -436,7 +436,8 @@ MimeIcon::created(StoreEntry *newEntry)
     e->flush();
     e->complete();
     e->timestampsSet();
-    e->unlock("MimeIcon::created");
+    // MimeIcons are only loaded once, prevent accidental destruction
+    // e->unlock("MimeIcon::created");
     debugs(25, 3, "Loaded icon " << url_);
 }
 
--- a/src/store.cc
+++ b/src/store.cc
@@ -1250,24 +1250,6 @@ storeLateRelease(void *)
     eventAdd("storeLateRelease", storeLateRelease, NULL, 0.0, 1);
 }
 
-/* return 1 if a store entry is locked */
-int
-StoreEntry::locked() const
-{
-    if (lock_count)
-        return 1;
-
-    /*
-     * SPECIAL, PUBLIC entries should be "locked";
-     * XXX: Their owner should lock them then instead of relying on this hack.
-     */
-    if (EBIT_TEST(flags, ENTRY_SPECIAL))
-        if (!EBIT_TEST(flags, KEY_PRIVATE))
-            return 1;
-
-    return 0;
-}
-
 bool
 StoreEntry::validLength() const
 {
--- a/src/store.cc
+++ b/src/store.cc
@@ -1201,12 +1201,8 @@ StoreEntry::release(const bool shareable)
 
     if (locked()) {
         releaseRequest(shareable);
-        /* "releaseRequest" may allow some entries such as cache digests to be
-         * unlocked immediately, so we should retest and release if possible. */
-        if (locked()) {
-            PROF_stop(storeRelease);
-            return;
-        }
+        PROF_stop(storeRelease);
+        return;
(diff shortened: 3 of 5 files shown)


craiggowing:
There is further description and discussion of this in bugzilla. @rousskov suggested there may be a better way of doing this rather than checking "locked()" twice, but for now it does seem to fix the problem with a relatively small change.

craiggowing:
I've updated the PR and the digest is now being unlocked during the start of the next rebuild as suggested. It seems to work well, no leaks detected and the Mime objects all seem to be intact. "mgr:objects" output is showing the MimeIcons all present with 1 active lock, as well as a single digest also with 1 active lock.

craiggowing:
I initially thought the unlock at the beginning of storeDigestRewriteStart wouldn't be a problem since it goes straight into storeDigestRewriteResume which sets the public key immediately, but I suppose it is not accounting for cases where the rebuild_lock is set. I still don't fully understand why it would be released at the point it is unlocked since the release flag shouldn't be set at this point and it should already be public.

Deferring it until after the new entry has the public key set does seem the safer option now since the release flag would just be set and the release would then happen on the unlock instead.

rousskov:
> I still don't fully understand why it would be released at the point it is unlocked since the release flag shouldn't be set at this point and it should already be public.

We cannot be certain that the release flag is not set -- _we_ did not set it when unlocking, but somebody else could do that (earlier). The only certainty here is that our lock protects the old entry.

> Deferring it until after the new entry has the public key set does seem the safer option now since the release flag would just be set and the release would then happen on the unlock instead.

Agreed. We can even unlock right _after_ making the new entry public -- if the latter fails/throws (for any reason) there will still be an old entry to serve others until Squid regenerates the digest again.

BTW, we should probably release() the old entry before unlocking it. We should not rely on the keys to clash and cause the release. The code computing the entry key might depend on some configuration values that may have changed or whatever. We know the old entry is stale and should explicitly release it when we ready to replace it. Here is a sketch:

```C++
// the calling order here preserves the old entry during exceptions
e->setPublicKey();
if (const auto oldEntry = sd_state.publicEntry) {
    oldEntry->release();
    sd_state.publicEntry = nullptr;
    oldEntry->unlock("storeDigestRewriteFinish"); // may delete oldEntry
}
assert(e->locked());
sd_state.publicEntry = e;
```

craiggowing:
I've updated the branch to unlock and release after the public key is first set on the new store entry. I don't think we should be setting the public entry during storeDigestRewriteResume since this can be called multiple times and could unexpectedly unlock and release the current store entry, so I'm setting the public entry in the state only during storeDigestRewriteFinish. I'm just building and testing this at the moment.

craiggowing:
@rousskov thanks for your clear explanation. I initially thought there may have been a race condition where this could incorrectly free the StoreEntry between the "storeDigestRewriteStart" event and from "storeDigestRebuildFinish", but I can now see that won't happen since "storeDigestRewriteStart" bails out if the rebuild lock is set and depends on the call from "storeDigestRebuildFinish". I've updated the branch accordingly.

I will hold off on the refactor since the rewrite lock _is_ the StoreEntry, so moving the creation to "storeDigestRewriteResume" would mean adapting the mechanism for getting "storeDigestRebuildFinish" to call "storeDigestRewriteResume" once done. Currently it does it if the lock is set or not. I suppose it would just be a case of changing the lock to an int like the other lock is and tracking the current StoreEntry object separately.

rousskov:
OK to test

craiggowing:
> Please review the updated PR description and a small improvement at [6464ed0](https://github.com/squid-cache/squid/commit/6464ed0ab0ade61a88d7afb91dae2f221cdca32c).
> 
> If you are happy, this PR can go in IMO.

Thanks, both look good and I've retested on v4.8 which works fine.

PR #488, opened 2019-10-01
Cleanup: rename String::limitInit() to assign()

josepjones:


src/esi/VarState.cc
```diff
@@ -282,7 +282,7 @@ void
 ESIVarState::addVariable(char const *name, size_t len, Variable *aVariable)
 {
     String temp;
-    temp.limitInit (name, len);
+    temp.assign (name, len);
```
rousskov: Please fix formatting while changing this line.

```suggestion
    temp.assign(name, len);
```

src/String.cc
```diff
@@ -228,7 +228,7 @@ String::substr(String::size_type from, String::size_type to) const
     Must(to > from);
 
     String rv;
-    rv.limitInit(rawBuf()+from,to-from);
+    rv.assign(rawBuf()+from,to-from);
```
rousskov: Please fix formatting when changing this line:

```suggestion
    rv.assign(rawBuf()+from, to-from);
```

josepjones:
I think starting a wiki page for prioritizing TODO/XXX sound like a reasonable approach. I'll look at getting my account setup and add a page.

rousskov:
OK to test

yadij:
FYI: https://wiki.squid-cache.org/RoadMap/Tasks is the wiki page to extend if you want to list the TODOs, \\todo, BUG, and FIXME, ... and any equivalents you find that are not requiring specialist skills. One meta TODO is to sort through and remove obsolete ones; another to define a meaning to those labels and document it - relabeling any inconsistent with that.

Keep in mind that we are adding and removing them all the time in the code, and difficulty is relative to ones skills and experience. So a fully curated list is likely a lot of work/time to take on.

rousskov:
> a fully curated list is likely a lot of work/time to take on

Agreed. Moreover, such work would be largely wasted right now: We do not have an army of volunteers eager and able to take on a long list of to-dos. If a new contributor needs help prioritizing 5-20 existing items that all seem attractive to _them_, that is easy to do once in a while and can be very useful in maximizing contributor's impact. That's what I offered [earlier](https://github.com/squid-cache/squid/pull/488#pullrequestreview-295752143). On the other hand, consistently maintaining 20 such items (when nobody is using that list) is costly and mostly wasteful.

rousskov:
@kinkie, can you kick Jenkins, please? It appears to be stuck in "about to shut down" state.

kinkie:
Done. The arm jobs were stuck and were preventing the restart.

rousskov:
Thank you! That reboot or restart seemed to revive Anubis as well.

PR #489, opened 2019-10-07
Reforward CONNECT after TLS handshake failure with peer

eduard-bagdasaryan:
When Squid received a CONNECT request and attempted to establish a
secure connection to an SSL cache_peer, it did not try the next
available destination after a TLS negotiation failure.

Why retry such TLS negotiation failures? Peer B may not have the same
misconfiguration that led to the negotiation failure when talking to
peer A. Admins may improve fault tolerance by using pools of different
peers or pools of identical peers that are reconfigured one by one. And
FwdState already does this -- flags.connected_okay is not set until TLS
negotiation is successful.


src/tunnel.cc
```diff
@@ -1016,8 +1067,11 @@ void
 TunnelStateData::connectedToPeer(Security::EncryptorAnswer &answer)
 {
     if (ErrorState *error = answer.error.get()) {
-        answer.error.clear(); // sendError() will own the error
-        sendError(error, "TLS peer connection error");
+        saveError(error);
+        answer.error.clear();
+        if (CachePeer *p = server.conn->getPeer())
```
rousskov:
```suggestion
        if (const auto p = server.conn->getPeer())
```

src/tunnel.cc
```diff
@@ -1016,8 +1067,11 @@ void
...
+            peerConnectFailed(p);
```
rousskov: Can this logic be moved to Security::BlindPeerConnector() or somesuch so that the callers do not need to remember to do this?

src/tunnel.cc
```diff
@@ -1016,8 +1067,11 @@ void
...
+        server.conn->close();
```
rousskov: If I understand the intent correctly, I would clarify:

```suggestion
        server.conn->close(); // tunnelServerClosed() will retry as needed
```

src/tunnel.cc
```diff
@@ -625,6 +661,20 @@ TunnelStateData::Connection::write(const char *b, int size, AsyncCall::Pointer &
     Comm::Write(conn, b, size, callback, free_func);
 }
 
+void
+TunnelStateData::Connection::setCloseHandler(CLCB *handler, TunnelStateData *tunnelState)
+{
+    assert(handler);
+    assert(tunnelState);
+    if (!Comm::IsConnOpen(conn)) {
```
rousskov: Why is this check needed? It seems to contradict what the method is promising to do. Sooner or later the caller may incorrectly rely on the closer being set. If the check is necessary, perhaps the method should be renamed to setCloseHandlerIfOpen()? Otherwise, this condition may become an assertion (or we could rely on a similar assertion in comm_remove/add_close_handler() if there is one in those functions).

src/tunnel.cc
```diff
@@ -625,6 +661,20 @@ TunnelStateData::Connection::write(const char *b, int size, AsyncCall::Pointer &
...
+        closer = nullptr;
+        return;
+    }
+    if (closer)
```
rousskov: If we may be resetting an old handler, then let's rename this method into resetCloseHandler() to reduce caller surprises.

src/tunnel.cc
```diff
@@ -625,6 +661,20 @@ TunnelStateData::Connection::write(const char *b, int size, AsyncCall::Pointer &
     Comm::Write(conn, b, size, callback, free_func);
 }
 
+void
+TunnelStateData::Connection::setCloseHandler(CLCB *handler, TunnelStateData *tunnelState)
```
rousskov: If it does not cause a lot of trouble, please use an AsyncCall instead of raw CLCB. Comm will convert CLCB into an AsyncCall for us, but that conversion cannot know the handler name.

src/tunnel.cc
```diff
@@ -259,61 +266,90 @@ static CTCB tunnelTimeout;
...
-    if (!tunnelState->server.writer) {
+    if (!tunnelState->server.writer)
         tunnelState->server.conn->close();
-        return;
+}
+
+/// Server close handler applied on the TLS connection being established.
+/// If TLS handshake failures, retries connection to another destination.
+static void
+serverClosed(const CommCloseCbParams &params)
```
rousskov: If this handler is specific to TLS closures, I would rename it to tlsServerClosed(), handleTlsServerClosure(), or similar.

eduard-bagdasaryan:
> Please reexamine new and rewritten static functions -- should some (or all) of them be TunnelStateData methods?

Done at 05e3b1a. I decided to convert them to methods (though these static methods do not need access to TunnelStateData private data) for simplicity sake.

src/tunnel.cc
rousskov: > if BlindPeerConnector is created, peerConnectFailed() will be called twice.

This problem seems unrelated to this PR. I would add an XXX to FwdState::connectedToPeer(). Here is a sketch:

```
// XXX: Required for PeekingPeerConnector but duped by BlindPeerConnector.
```

rousskov:
> I decided to convert them to methods (though these static methods do not need access to TunnelStateData private data)

I think you did the right thing. I sense the implied conflict with the "Make a function a member only if it needs direct access to the representation of a class" [guideline](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-member), but we have to keep in mind that most, if not all TunnelStateData data members should not be public but currently are (in part because there are legacy non-member functions that access them). We may be able to solve this problem for good when deduplicating TunnelStateData and FwdState. Your decision to use methods here is the right step towards that anticipated final solution.

eduard-bagdasaryan:
> add an XXX to FwdState::connectedToPeer()

Done at 10fdab4.

src/tunnel.cc
```diff
@@ -922,7 +916,7 @@ TunnelStateData::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
 void
 TunnelStateData::notePeerReadyToShovel()
 {
-    server.resetCloseHandlerIfOpen(tunnelServerClosed, "tunnelServerClosed");
+    server.resetCloseHandler(tunnelServerClosed, "tunnelServerClosed");
```
eduard-bagdasaryan: I do not see an 'explicit check' in this path: tunnelEstablishmentDone() -> notePeerReadyToShovel(). The connection should be open when connectedToPeer() is called ( and tunnelEstablishmentDone() is scheduled), but between this scheduling and calling the method it may close already.

src/tunnel.cc
```diff
@@ -323,30 +323,35 @@ TunnelStateData::handleClientClosure()
...
+        return;
     }
+
+    if (!tunnelState->hasError()) {
+        const auto anErr = new ErrorState(ERR_CANNOT_FORWARD, Http::scInternalServerError,
+                tunnelState->request.getRaw(), tunnelState->al);
+        tunnelState->saveError(anErr);
+    } // else use actual error from last connection attempt
+
+    tunnelState->handleServerClosure();
```
eduard-bagdasaryan: > We
need to stop only when there are no destinations to try _and_ no more
destinations are expected

The difference I see between the old code and the new one is where we check and save the error by saveError(): previously it was done if subscribed is true, now - if subscribed is false. If my understanding is correct, the old code could work incorrectly because we saved the error before all destinations have been tried and noteDestinationsEnd() was called: during checking the last destination noteDestinationsEnd() could see the saved error and do preliminary abort by sendError(). However, the same problem can happen because we also save the error in connectedToPeer(), i.e., before tlsServerClosed().

src/tunnel.cc
rousskov: I agree that this code looks suspicious (at least), on several levels.

> The connection ... may close already.

If the connection is closed, why was not TunnelStateData close handler called? Did TunnelStateData have no handler when the connection started to close? Or do you mean that the connection is closing, and a TunnelStateData previous/current close handler call has been scheduled/queued already?

Moreover, the location of this reset call looks rather strange. Why are we resetting the close handler here (rather than earlier or later)? What state change does calling this method signify (that forces us to reset the handler)?

src/tunnel.cc
rousskov: > the old code could work incorrectly because we saved the error before all destinations have been tried

AFAICT, the old code was wrong because it did premature error handling/processing (i.e. checking whether there was a saved error and creating a new made-up error if there was not) and initiated tunnel termination (i.e. calling handleServerClosure()) instead of simply waiting for more destinations to try. In other words, the old code treated the tunnel as unusable/dead/broken when there was no reason for doing that.

> during checking the last destination noteDestinationsEnd() could see the saved error

FWIW, I do not think that specific series of events was possible -- noteDestinationsEnd() does not check for saved errors if destinationsFound is true. The noteDestination() method naturally makes destinationsFound true. And without at least one noteDestination() call, there can be no tlsServerClosed(), right?

eduard-bagdasaryan: > (i.e. checking whether there was a saved error and creating a new made-up error

Agreed. I thought that we need to save the error from the previous (unsuccessful) attempt so that we could use it if something went wrong (and no more errors are available). Probably this situation is unlikely, and each connection attempt is either successful or produces an error.

> and initiated tunnel termination (i.e. calling handleServerClosure()) instead of simply waiting for more destinations to try.

Disagreed, this is not how the old code worked. Basically, you changed my if/elses into if/returns, but the logic remains the same.

Before:

```
 if (!tunnelState->destinations->empty()) {
    // start connecting
 } else if (tunnelState->subscribed) {
   // create an error if needed and wait
 } else {
  // cleanup
 }
```

After:

```
if (!tunnelState->destinations->empty()) {
  // start connecting
  return;
}
if (tunnelState->subscribed) {
  // wait
  return;
}
// save the error if needed
// cleanup

```

> And without at least one noteDestination() call, there can be no tlsServerClosed(), right?

I think you are right.

src/tunnel.cc
eduard-bagdasaryan: > Or do you mean that the connection is closing, and a TunnelStateData previous/current close handler call has been scheduled/queued already?

Yes, this 'previous/current' is tlsServerClosed() handler, which is called on closures before the tunnel is fully established.

> Moreover, the location of this reset call looks rather strange.

Since tunnelServerClosed() handler is responsible for handling tunnel closures, I think we need to reset to tunnelServerClosed() handler when the tunnel is fully established. Also (according to the code) I think that it is fully established somewhere inside tunnelEstablishmentDone(), after we checked that the answer is possible. Assuming this, I reset the handler inside notePeerReadyToShovel() which is called in this case.

rousskov: > we need to reset to tunnelServerClosed() handler when the tunnel is fully established

I think that is one of the cases (for some definition of "fully established"). The other case is when the tunnel has failed in such a way that another attempt to establish a connection with the server is impossible or prohibited (for any reason). The common notion among these cases is that we are looking for a "point of no return".

I believe we reach that point of no return in two places:

1. Where connectDone() calls notePeerReadyToShovel() -- this handles to-origin use case.
2. Where connectedToPeer() discovers a _positive_ answer -- this handles to-peer use case. Note that this to-peer use cases also calls notePeerReadyToShovel(), but that happens later, well after the point of no return.

Placing closure handler reset in notePeerReadyToShovel() handles the first use case well, but it is the wrong spot because it causes problems for the second use case. The nature of those problems would vary from bugs to double-reset overheads, depending on how the rest of the code is changed, but those details are unimportant.

Branch commit f433d15 moves the close handler reset from notePeerReadyToShovel() to the two places enumerated above. Unfortunately, that change appears to expose more problems. Please see commit message TODO and commit XXXs.

src/tunnel.cc
rousskov: > I thought that we need to save the error from the previous (unsuccessful) attempt so that we could use it if something went wrong (and no more errors are available).

You thought correctly, but connection closure itself is not an error. If we discovered an error earlier, then it was already saved. Otherwise, (i.e. this is some kind of external closure that we did not initiate), there is no information we _can_ save to explain the being-closed connection -- we do not know why it got closed or who closed it. All we can do is to try the next destination.

If there are no other destinations to try (or retries are prohibited), then we reuse the previously saved error or (if there was not one saved), generate a generic "we are out of paths" error.

If you are OK with the above explanation and the current tlsServerClosed() code, then I believe this conversation thread can be resolved because the code already reflects what was said here.

----

> Disagreed, this is not how the old code worked

You are right. I probably misplaced the handleServerClosure() call when describing the diff. Sorry. Only the error creation was wrong, not the handleServerClosure() call placement.

src/tunnel.cc
eduard-bagdasaryan: > Please see commit message TODO and commit XXXs.

I tried to address them at 3a9481f.

>  This solution requires determining whether the being-closed connection can be retried. If that determination is easy

I could not find how to determine this by means of available methods/flags so I had to add a new TunnelStateData::retriable flag for this purpose. There is a similar HappyConnOpener::retriable_ already.

src/tunnel.cc
```diff
@@ -1016,11 +1082,16 @@ void
...
-        return;
+        saveError(error);
+        answer.error.clear();
+        Must(server.monitorsClosures()); // closure notification (is already pending or) will happen
+        if (IsConnOpen(server.conn))
+            server.conn->close();
+        return; // tunnelServerClosed() will retry as needed
     }
 
+    retriable = false;
```
rousskov: Should not TunnelStateData retry Http::Tunneler failures? FwdState does retry them, right?

src/tunnel.cc
```diff
@@ -250,75 +269,110 @@ class TunnelStateData: public PeerSelectionInitiator
...
+/// handles closures of the Squid-to-server connection
 static void
-tunnelClientClosed(const CommCloseCbParams &params)
+tunnelServerClosed(const CommCloseCbParams &params)
 {
-    TunnelStateData *tunnelState = (TunnelStateData *)params.data;
-    debugs(26, 3, HERE << tunnelState->client.conn);
-    tunnelState->client.conn = NULL;
-    tunnelState->client.writer = NULL;
+    const auto tunnelState = reinterpret_cast<TunnelStateData *>(params.data);
```
rousskov: It looks like the guts of this rewritten function (i.e. everything below the reinterpret_cast) should become TunnelStateData::serverClosed(). There is a similar method in FwdState already.

src/tunnel.cc
```diff
@@ -1242,13 +1313,15 @@ switchToTunnel(HttpRequest *request, Comm::ConnectionPointer &clientConn, Comm::
...
     TunnelStateData *tunnelState = new TunnelStateData(context->http);
+    tunnelState->retriable = false;
 
     // tunnelStartShoveling() drains any buffered from-client data (inBuf)
     fd_table[clientConn->fd].useDefaultIo();
 
     request->hier.resetPeerNotes(srvConn, tunnelState->getHost());
 
     tunnelState->server.conn = srvConn;
+    tunnelState->server.resetCloseHandler(tunnelServerClosed, "tunnelServerClosed", tunnelState);
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

src/tunnel.cc
eduard-bagdasaryan: Done.

src/tunnel.cc
eduard-bagdasaryan: Done.

eduard-bagdasaryan:
Please mark this PR with 'waiting-for-reviewer'.

eduard-bagdasaryan:
> The last branch commit (9d0dd77) promises to retry on Tunneler failures ...

I tried to address your concerns at c94fd17.

src/tunnel.cc
```diff
@@ -1015,10 +1121,11 @@ TunnelStateData::connectToPeer()
 void
 TunnelStateData::connectedToPeer(Security::EncryptorAnswer &answer)
 {
-    if (ErrorState *error = answer.error.get()) {
-        answer.error.clear(); // sendError() will own the error
-        sendError(error, "TLS peer connection error");
```
eduard-bagdasaryan: @chtsanti , I have a question you probably may help with. Should not we call sendError() here only if clientExpectsConnectResponse() check returns true? For example, TunnelStateData::tunnelEstablishmentDone() already does this check before calling sendError() which makes sense. I assume that this is a bug and fixed it in the branch, but I would like you to confirm that my decision was correct.

chtsanti: Yep, it does not make sense to send an HTTP error if the client does not "speak" HTTP.

In other places too the sendError is used without check the clientExpectsConnectResponse() before. 

I must note that the sendError also updates the TunnelStateData::*status_ptr which is used to log the status code in squid access log. This is must updated even if we do not want to send the error to the client. The TunnelStateData::tunnelEstablishmentDone() update it even if the sendError is not called.

A solution is to replace the TunnelStateData::sendError(error) calls with a TunnelStateData::sendErrorOrClose(error) which updates always the status_ptr, and checks clientExpectsConnectResponse() before send the error or close the connection.

I do not know if this is should fixed inside this PR. In any case inside TunnelStateData::retryOrBail when the sendError is not called we should update the status_ptr to the error->httpStatus.

eduard-bagdasaryan: >  This is must updated even if we do not want to send the error to the client.

TunnelStateData::*status_ptr points to al->http.code, which is logged as ">Hs" and described as "HTTP status code sent to the client". However, in our case, we do not send the HTTP error response and, hence, the status code to the client. Something is wrong here: either the code not updating 'status_ptr' or the documentation lying about the sent error code. What do you think?

rousskov: We must update `*status_ptr` as Christos mentioned.

Related %codes documentation is indeed imprecise (e.g. `sent` really means `sent or would have been sent if ...` in most cases), but, as you know, recent attempts at improving documentation precision faced significant non-technical hurdles, so I would leave it as is.

chtsanti: The only I can say is that the status code is useful to be logged even if it is not really sent to the client. In any case if not updated still we may log something which is not the final error. For example:
  - TunnelStateData::tunnelEstablishmentDone called with error.
  - TunnelStateData::tunnelEstablishmentDone updates status_ptr and call retryOrBail
  - At the next try the TunnelStateData connected OK but calls Encryptor which fails
  - The retryOrBail does not send the error to client and does not update the status_ptr 
  - Squid logs the status of the first failed transaction, which maybe is not the same as the second one

Maybe there are other similar cases too.
My opinion is that if needed we should update the  >Hs documentation. But I think squid users expects this behavior even if not documented.

eduard-bagdasaryan: Done at a63d463.

rousskov:
@yadij, if you were waiting for the PR code to settle down, then I think it is ready for your review.

src/tunnel.cc
```diff
@@ -430,16 +433,16 @@ TunnelStateData::retryOrBail(const char *context)
     // sendNewError() and sendSavedErrorOr(), used in "error detected" cases.
     const auto error = savedError ? savedError : new ErrorState(ERR_CANNOT_FORWARD,
             Http::scInternalServerError, request.getRaw(), al);
-    if (Comm::IsConnOpen(client.conn) && clientExpectsConnectResponse()) {
-        sendError(error, context);
-        return;
-    }
+    const auto canSendError = Comm::IsConnOpen(client.conn) && !client.dirty &&
```
eduard-bagdasaryan: I am trying to understand whether the new Connection::dirty is really required: can we use the existing 'retriable' flag here? It becomes 'false' in notePeerReadyToShovel(), whereas 'durty' I guess becomes true a bit later, somewhere in tunnelStartShoveling().

rousskov: I rejected the idea of reusing `retriable` (i.e. `dirty == !retriable`) because the two states have different (and largely orthogonal) semantics. Regardless of what the code does (not do) now, it may be possible, in theory, to retry the connection to the _next_ hop after the connection to the _previous_ hop becomes dirty. In other words, I can image some future code having true `retriable` (yes, can try another peer) and true `dirty` (no, can no longer respond with an error).

The above logic does not necessarily imply that `retriable` is falsified too early today. Squid may not be able and/or may not be _willing_ to retry Squid-to-peer failures after notePeerReadyToShovel(). Imagine, for example, a write timeout for the first chunk of data going from Squid to the peer. We could retry after that error (at least in some cases) because no client data was sent to the next hop, but there is currently no code to support such retries.

eduard-bagdasaryan:
>  if the current code still works well

I have not found problems with the code. I reran tests which were OK.

src/tunnel.cc
```diff
@@ -361,6 +393,61 @@ TunnelStateData::Connection::~Connection()
...
+
+    /* bail */
+
+    if (request)
+        request->hier.stopPeerClock(false);
+
+    // TODO: Add sendSavedErrorOr(err_type type, Http::StatusCode, context).
+    // Then, the remaining method code (below) should become the common part of
+    // sendNewError() and sendSavedErrorOr(), used in "error detected" cases.
+    const auto error = savedError ? savedError : new ErrorState(ERR_CANNOT_FORWARD,
```
eduard-bagdasaryan: It looks like the created error is never deleted if canSendError is false.

rousskov: Agreed. Please fix. I suspect calling saveError(error) if savedError is nil would be sufficient, but I did not check the details. The TODO above this line is still out of scope.

eduard-bagdasaryan: Fixed at d38e99c.

src/tunnel.cc
```diff
@@ -672,15 +781,40 @@ tunnelTimeout(const CommTimeoutCbParams &io)
...
+        comm_remove_close_handler(conn->fd, closer);
+        closer = nullptr;
         conn->close();
+        conn = nullptr;
+        return;
+    }
+
+    if (closer) {
+        Must(conn);
+        Must(fd_table[conn->fd].closing()); // explains false IsConnOpen()
```
eduard-bagdasaryan: This code looks problematic to me:: both true conditions !Comm::IsConnOpen(conn) and conn!=null imply that fd==-1, but we cannot use this as an array index.

rousskov: `conn && !IsConnOpen(conn)` does not imply `fd==-1` because Connection::isOpen() tests (among other things) `fd_table[fd].flags.open`.

However, I agree that we should add another Must() to make sure that conn->fd is set. There may be similar problems in some other fde::closing() calls (inside and/or outside this PR). Please check and fix (the ones inside this PR).

eduard-bagdasaryan: Here are these checks:

```
Comm::IsConnOpen(const Comm::ConnectionPointer &conn) { return conn != NULL && conn->isOpen(); }

bool Connection::isOpen() const { return (fd >= 0); }

```

I do not see here fd_table[fd].flags.open check.

src/tunnel.cc
```diff
@@ -894,6 +1021,7 @@ TunnelStateData::noteConnection(HappyConnOpener::Answer &answer)
         syncHierNote(answer.conn, request->url.host());
         saveError(error);
         answer.error.clear(); // savedError has it now
+        // XXX: should honor clientExpectsConnectResponse() before replying. Also check other occurrences.
         sendError(savedError, "tried all destinations");
```
yadij: Please enact this XXX before merging. It looks like the correct actions will be performed by:
```suggestion
        retriable = false;
        retryOrBail("tried all destinations");
```

eduard-bagdasaryan: You are right that we could easily fix this specific case but, unfortunately, these 'other occurrences' do exist (see, e.g., TunnelStateData::noteDestinationsEnd()). So, we would have to move then this XXX there but the same question arises again. I think it would be better to resolve all these cases in a separate dedicated PR.

src/tunnel.cc
```diff
@@ -672,15 +781,40 @@ tunnelTimeout(const CommTimeoutCbParams &io)
...
+TunnelStateData::Connection::closeQuietly()
+{
+    assert(!writer);
+
+    if (Comm::IsConnOpen(conn)) {
+        Must(closer);
+        comm_remove_close_handler(conn->fd, closer);
+        closer = nullptr;
         conn->close();
+        conn = nullptr;
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Do you mean that it would be better to call just conn->close(), do not remove the handler (tunnelServerClosed()) so that it would perform all retry/cleanup (including retryOrBail())?
In this case, roughly, the call sequence would be:
`retryOrBail() ->conn->close()->tunnelServerClosed()->retryOrBail()->retry`
Currentlry, it is
`retryOrBail() ->closeQuietly()->retry`
, which is simpler. In other words, since we close the connection ourselves, we can eliminate the callback call and do all cleanup at once.

src/tunnel.cc
rousskov: @eduard-bagdasaryan, you are right. I must have looked at isOpen() instead of IsConnOpen(). The different meaning of these functions is probably the source of some old bugs, but we are not going to address that in this PR, of course. I am glad you spotted this new bug! I hope branch commit b722bc4 addresses your concern. Please check and adjust as needed.

> @yadij: In its current placement this if-statement block should not be doing anything related to conn, only cancel and discard the closer.

And it should also clear the `conn` because the tunnel is completely done working with that connection now. The `conn` member is not nil inside the `if (closer)` statement (when Comm::IsConnOpen() is false), because when the tunnel calls conn->close(), it does not clear the `conn` member. This `conn` clearing is probably not critical because we usually check Comm::IsConnOpen() instead of checking the `conn` pointer itself, but this clearing is the right action here.

When the closeQuietly() method returns, the affected tunnel side should no longer be associated with any Connection object, and no Connection callbacks should reach the tunnel. Please keep in mind that this method is not a connection closing callback. It is a method designed to get rid of the connection (closing it if needed) without allowing the previously assigned closing callbacks (if any) to reach the tunnel.

src/tunnel.cc
rousskov: > it would be better to resolve all these cases in a separate dedicated PR.

@eduard-bagdasaryan, agreed. This fix is not directly related to this PR and requires non-trivial likely-to-cascade changes. Since those changes affect traffic on the wire, a dedicated description would help admins to asses whether the fix is related to problems they suffer from. The fix deserves its own PR.

In general, we should avoid penalizing folks for reporting bugs. If a developer adds an XXX to some old code, and that XXX _can be_ interpreted as lying outside the PR scope, we should not press the developer into addressing that XXX. Otherwise, developers will simply stop adding valuable XXXs and TODOs, hurting everybody!

src/tunnel.cc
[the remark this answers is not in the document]
rousskov: I suspect that @yadij means that `conn` should not be set to nil after calling `comm->close()` in this "get rid of the connection" method. I disagree. The `conn` member represents tunnel's association with the current to-server connection, not the "last" to-server connection. Thus, the member should be cleared when the current connection is gone.

> The non-FD data stored there may be needed for logging, debugs or error display later.

The method caller has decided to get rid of the current to-server connection. If the caller wanted to preserve connection information, it could have done so (e.g., in an error or ALE object). Such preservation is not _this_ method business.

There are known serious problems with Comm::Connection API(*), but we are not going to solve those problems here. Fortunately, this PR can accomplish its goals without solving those API problems.

----

(*) including Connection use for storing things not associated with any existing connection and the fact that `isOpen(conn->fd) != IsOpen(conn)` for some Connections.

src/tunnel.cc
eduard-bagdasaryan: I think your b722bc4 solution fixed the problem.

eduard-bagdasaryan:
This should be switched back into the 'waiting for reviewer' state.
@yadij , are all your remarks addressed now?

src/tunnel.cc
yadij: Sending a plain-text error page to a client which is not speaking plain-text HTTP. The client may be speaking some unknown protocol which can be corrupted in unpredictable ways by the error page response.

We should not be accepting a PR which actively *adds* cases of traffic corruption. Especially when that protection already existed with the previous sendError() call and is being deleted by this PR. It should have been moved along with the sendError() call.

src/tunnel.cc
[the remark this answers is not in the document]
yadij: @rousskov so where can you see the logging getting its information about the connection last tried when the Pointer to that data is nil?
 Also, how is retry going to re-open the connection when there is no record of where it was going?

src/tunnel.cc
yadij: > Please keep in mind that this method is not a connection closing callback.

... yet it is called "close", documented as closing the connection, and calls close at every opportunity.

> It is a method designed to get rid of the connection (closing it if needed) without allowing the previously assigned closing callbacks (if any) to reach the tunnel.

The design of Comm::Connection is that code with that requirements should cancel their closer callback and set their Pointer to nullptr. Not call close(). Anything sharing the conn object will then be free to continue using it, and if there are no sharing the conn will close itself cleanly.

There is one oddity with tunnels *if* the connection has possibly been used (dirty?) or the client was told it got opened (200 status message) then is *also* has to be closed for safety. That is a feature/promise of the tunnel mechanism - *not* the Comm::Connection API.

IMO this method should look like this:
``
{
    assert(!writer);

    if (closer) {
        // cancel in case it is already scheduled
        closer->cancel("no longer needed by tunnel");
    }

    if (Comm::IsConnOpen(conn)) {
        Must(closer);
        comm_remove_close_handler(conn->fd, closer);
    }

    conn->close();
    conn = nullptr;
    closer = nullptr;
}
``

yadij: Maybe with a check for conn being nullptr at the end there if that case can ever happen.

src/tunnel.cc
eduard-bagdasaryan: > Especially when that protection already existed with the previous sendError() 

This is incorrect: the old sendError() lacked such protection.

src/tunnel.cc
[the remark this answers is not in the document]
rousskov: > where can you see the logging getting its information about the connection last tried

Access-logging code should get its information from ALE (and, where really necessary, HttpRequest and similar high-level legacy locations). Certainly not from TunnelStateData::Connection that it cannot even access because the TunnelStateData class is private to tunnel.cc!

> ... when the Pointer to that data is nil?

For logging code, it does not matter whether TunnelStateData::server.conn is nil. That code has no access to that pointer. And even if it did have access, that pointer is _not_ a pointer to "the connection last tried". It is the pointer to the "currently connected connection", as documented. 
Naturally, this pointer should be nil when there is no such connection. I see no other reasonable way to use Connection pointers in code that goes through multiple connections.

> how is retry going to re-open the connection when there is no record of where it was going?

If/when TunnelStateData supports reforwarding requests to the same destination, it will probably get that reforwarding address from the TunnelStateData::destinations collection. FwdState works that way already.

src/tunnel.cc
rousskov: > Alex: It is a method designed to get rid of the connection (closing it if needed) without allowing the previously assigned closing callbacks (if any) to reach the tunnel.

> Amos: The design of Comm::Connection is that code with that requirements should cancel their closer callback and set their Pointer to nullptr. Not call close().

@yadij, the first 2-3 paragraphs of your [comment](https://github.com/squid-cache/squid/pull/489#discussion_r376959427) reflect some kind of miscommunication -- I think we have been talking about different code areas and/or different method requirements while thinking that we are discussing the same thing. I suggest exiting that part of the discussion with the following basic rule of thumb:

* When an owner of a Comm::Connection object decides that it is time to close the underlying transport connection, that owner should call Comm::Connection::close().

Both the PR code and your suggestion (quoted below in its polished version, per your [comment](https://github.com/squid-cache/squid/pull/489#discussion_r376961468)) comply with the above rule.

----

> Amos: IMO this method should look like this:


```C++
if (closer) {
    // cancel in case it is already scheduled
    closer->cancel("no longer needed by tunnel");
}

if (Comm::IsConnOpen(conn)) {
    Must(closer);
    comm_remove_close_handler(conn->fd, closer);
}

if (conn) {
    conn->close();
    conn = nullptr;
}

closer = nullptr;
```

AFAICT, both the PR code and the above suggestion accomplish the same thing. Your suggestion focuses on "clearing" or "getting rid of" seemingly independent connection artifacts -- closer, closer, and conn (and closer again). Those objects are actually dependent on each other and their dependencies are important for understanding and writing correct code. The code also has small problems (discussed/fixed later in this comment).

Overall, I prefer the PR version because it focuses on separating the two important but not obvious use cases (the first two bullets below), simplifying code logic by removing several hidden/dangerous dependencies or misleading if statements. It also allows us to add useful assertions. The third use case is trivial, but we should add a comment to document it _explicitly_, like we already do with the other two cases. All three use cases are mutually exclusive and all focus on the same set of {TunnelStateData,Comm}::Connection objects:

* Connection-owning TunnelStateData did not call conn->close() before this closeQuietly() call.
* Connection-owning TunnelStateData called conn->close() before this closeQuietly() call.
* TunnelStateData does not own the connection (at the time of this closeQuietly() call).


There are also several small problems with the suggested code quoted above:

1. The code manually cancels the callback and then calls comm_remove_close_handler() which also (naturally) cancels the same callback. Ideally, such repeated cancellations should be avoided.
2. The code introduces a new state -- `closer` is still set but the callback has already been canceled. Yes, this state is short-lived, but, ideally, the `closer` pointer should be cleared immediately after we get rid of the callback, reducing the chances of future separation bugs, removing surprises during exception-triggered cleanup, and simplifying object state. There should be no third state -- either we are asking Comm to call us back or we are not.
3. The code implies that IsConnOpen() might be true even without `closer`. That implication is wrong -- we should never have an open connection without a closing callback! It may seem like a minor complication -- it _is_ possible to get rid of each object independently -- but the tunneling code will become more complex if it does not maintain/rely on that simpler state. This is similar to adding an innocent-looking `if (foo)` statement where `foo` cannot be nil: Sooner or later, the code will start partially handing nil foos, nil foos might even become possible (breaking legacy code), and, in a few years, it would become very difficult to untangle that mess. The rule of thumb here is that a connection owner must maintain a closing callback -- the two objects are tied together. (Eventually, we will add APIs to better express connection ownership)
4. The code implies that conn might be true even without `closer`. Same reasoning as in item 3.


The above problems can be fixed, distilling the essential differences between PR and the suggested code. For example:

Item 1 can be addressed by avoiding the split of the callback cancellation code, moving it all inside the dedicated `closer` check:

```C++
if (closer) {
    if (Comm::IsConnOpen(conn))
        comm_remove_close_handler(conn->fd, closer);
    else
        closer->cancel("no longer needed by tunnel");
}

if (conn) {
    conn->close();
    conn = nullptr;
}

closer = nullptr;
```

Note that the above essentially fixes item 3 as well, which we can now express by adding a Must():

```C++
if (closer) {
    if (Comm::IsConnOpen(conn))
        comm_remove_close_handler(conn->fd, closer);
    else
        closer->cancel("no longer needed by tunnel");
} else {
    // open connection requires a closer
    Must(!Comm::IsConnOpen(conn));
}

if (conn) {
    conn->close();
    conn = nullptr;
}

closer = nullptr;
```

Similarly, we can fix item 4 by moving the close() call and the `conn` nullification to where `conn` is present. The Must() we just added above can be strengthened accordingly:

```C++
if (closer) {
    Must(conn);
    if (Comm::IsConnOpen(conn)) {
        comm_remove_close_handler(conn->fd, closer);
        conn->close();
        conn = nullptr;
    } else {
        closer->cancel("no longer needed by tunnel");
        conn = nullptr;
    }
} else {
    Must(!conn); // connection requires a closer
}

closer = nullptr;
```

Now we can address item 2 by nullifying `closer` ASAP:

```C++
if (closer) {
    if (Comm::IsConnOpen(conn)) {
        comm_remove_close_handler(conn->fd, closer);
        closer = nullptr;
        conn->close();
        conn = nullptr;
    } else {
        closer->cancel("no longer needed by tunnel");
        closer = nullptr;
        conn = nullptr;
    }
} else {
    Must(!conn); // connection requires a closer
}
```

To simplify the comparison with the PR code, let's convert the above to use the if-return style:

```C++
if (closer) {
    if (Comm::IsConnOpen(conn)) {
        comm_remove_close_handler(conn->fd, closer);
        closer = nullptr;
        conn->close();
        conn = nullptr;
        return;
    }

    closer->cancel("no longer needed by tunnel");
    closer = nullptr;
    conn = nullptr;
    return;
}

Must(!conn); // connection requires a closer
```

Finally, let's add comments that document the two (non-obvious) states (these comments are already present in this PR) and the third state (the PR should get the same `we do not own a connection` comment):

```C++
if (closer) {
    if (Comm::IsConnOpen(conn)) {
        // We did not call conn->close().
        comm_remove_close_handler(conn->fd, closer);
        closer = nullptr;
        conn->close();
        conn = nullptr;
        return;
    }

    // We called conn->close() followed by this closeQuietly() call.
    // Our close callback has already been scheduled but not yet fired.
    closer->cancel("no longer needed by tunnel");
    closer = nullptr;

    // fd_table[old conn->fd].closing() ought to be true here,
    // but we cannot check that because conn->close() clears conn->fd

    conn = nullptr;
    return;
}

// we do not own a connection
Must(!conn); // connection requires a closer
```

Now it becomes clear that the fixed version uses nested `if` statements to detect exactly the same two states that the PR code detects without nesting. The fixed version also uses a rather unusual for Squid top-level condition -- `closer`; most of Squid code starts by handling the Comm::IsConnOpen(conn) case at the top level.

IMO, the PR version is (a little) better than the code above because it is a little simpler and follows Squid conventions a little better. Do you insist on going with the above version instead? If yes, why?

src/tunnel.cc
yadij: Sorry for not being clear, I mean the code sequence in tunnelEstablishmentDone() protecting sendError() calls:
```
 if (!clientExpectsConnectResponse()) {
        // closing the non-HTTP client connection is the best we can do
        debugs(50, 3, server.conn << " closing on CONNECT-to-peer error");
        server.closeIfOpen();
        return;
    }
...
    sendError(error, "tunneler returns error");
```

The if-condition preventing error delivery is gone in PR code paths. This XXX is about the new PR code re-opening old bugs by removal of that condition.

IMO it is reasonable to expect this PR to avoid adding at least the easily solved occurrences (like this marked one) to minimize new bugs. Ones that are found to be not so simple can be deferred with an XXX as Alex says - at least then "Also check other occurrences" part is done and all remaining bugs marked so nobody mistakenly copies the bad code around before that deferred fix is done.

yadij:
So I have looked through the open discussions and closed the ones I am sure are resolved (some were old points by Alex).

Right now we have:
* one question by @eduard (https://github.com/squid-cache/squid/pull/489#pullrequestreview-328440583)  that needs resolving or continuing discussion.
* one request for XXX handling (https://github.com/squid-cache/squid/pull/489/files#r361874698) I have clarified hopefully what I would like to see happen there. It should not take much.

There are some open discussions about conn usage. These are polish and design principle discussions, not blocking on this PR. Once we are happy with the above two points this can be marked for merge.

src/tunnel.cc
eduard-bagdasaryan: > The if-condition preventing error delivery is gone in PR code paths. 

If your concern is only about tunnelEstablishmentDone() - I see no problem there, because retryOrBail() takes care of not calling sendError() if clientExpectsConnectResponse() returns false.

src/tunnel.cc
eduard-bagdasaryan: Ok.

src/tunnel.cc
yadij: That is *one* path where the now-deleted condition code enters retryOrBail() and reaches sendError(). There are many paths through retryOrBail() which do not reach that safe sendError() - as intended most of those exit and loop back into logic where other errors may need sending. eg. follow a winding path around to hit this XXX marked sendError() or any of the "other occurrences" alluded to.
 Thus; all those logic paths have this newly added "regression" bug.

So ... what about all the other sendError() occurances? please do at least locate and XXX marking part so the "Also check other occurrences" text can be erased. If possible fix so they do not need that period of annoying XXX behaviour.

rousskov: Sigh. This discussion and associated waste and delays is exactly why every time I want to document an accidentally discovered out-of-scope problem, I tell myself "Stop, Amos will make you regret this good deed!". I am a fool for hoping that "the next time it will be different". Again and again, we are being penalized for discovering and documenting out-of-scope problems.

rousskov: FTR, Factory will address this XXX. Since the XXX is unrelated to this PR, there will be a separate PR addressing it.

rousskov:
> * one question by @eduard ([#489 (review)](https://github.com/squid-cache/squid/pull/489#pullrequestreview-328440583))  that needs resolving or continuing discussion.

AFAICT, that change request is now [marked](https://github.com/squid-cache/squid/pull/489#discussion_r429898047) by Eduard as resolved.

> * one request for XXX handling (https://github.com/squid-cache/squid/pull/489/files#r361874698) I have clarified hopefully what I would like to see happen there. It should not take much.

If there are reasons to believe that this PR will make things significantly worse in some supported environments, then merging of this PR should be _delayed_ until the old bug is fixed (in a separate PR). Otherwise, this PR should be merged without artificial delays.

> Once we are happy with the above two points this can be marked for merge.

Do you have any reasons to believe that this PR will make things significantly worse in some supported environments? If yes, can you give an example?

FWIW, I could not tease out an answer from your previous comments. AFAICT, the undisputed/uncorrected comments were discussing unspecified code paths rather than specific use cases. They did not make it clear (to me) whether an explosion of new problems is likely here.

yadij:
I have done a simple review of most sendError() calls myself. All the ones checked were changed from unreachable without this PR to almost guaranteed to occur when:

either
* a peer has 2+ addresses:
or 
* multiple peers found for use in the peer selection destinations

plus any of
 * peer is non-secure misconfigured with any ssl or tls settings
 * reverse-proxy with originserver setting
 * peer requires TLS cipher(s) not supported by the current config.

rousskov:
> Please remember all I am asking is to mark them with XXX/TODO. 

Thank you for clarifying that.

> It should not be an arduous task to do that.

It is not arduous at all! Done at d54ecef.

eduard-bagdasaryan:
> carefully resolve these conflicts

Done at be346b3.

rousskov:
> You can mark as cleared for merge once the conflicts are resolved. I do not think this needs another review from me.

@yadij, the conflict resolution changes (be346b3) were not trivial. I will not clear this PR for a couple of days in case you want to re-review.

src/tunnel.cc
```diff
@@ -319,7 +316,6 @@ void
 TunnelStateData::serverClosed()
 {
     server.noteClosure();
-    retryOrBail("server closed");
 }
```
rousskov, line 319: @eduard-bagdasaryan, it looks like the retryOrBail() call in #489 branch was removed in branch commit be346b3 (when resolving conflicts with master). When you review #957 that effectively restores this call, please see whether you can recall why the call was removed -- if that removal was not an accident, then we may need to do more in that PR!

eduard-bagdasaryan: I believe that this removal was a mistake. I could not recollect exactly why I removed it, but perhaps I got confused by TunnelStateData::retriable, which is reset before the TunnelStateData::serverClosed() handler is assigned and makes impossible the 'retry' path, but still requires the 'bail' functionality.

PR #490, opened 2019-10-07
Send HTTP/500 (Internal Server Error) when lacking peers

eduard-bagdasaryan:
... instead of sending HTTP/503 (Service Unavailable) in
tunneling cases.

Tunneling code reported 503 while regular forwarding code reported 500
errors under identical "no peers to try" circumstances.

For consistency sake, we need to use one of these codes in both places.
We considered and rejected 503 because RFC7231 limits its usage to
several specific cases, not including 'lacking peers' case.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -445,7 +445,7 @@ FwdState::useDestinations()
 
         debugs(17, 3, HERE << "Connection failed: " << entry->url());
         if (!err) {
-            const auto anErr = new ErrorState(ERR_CANNOT_FORWARD, Http::scInternalServerError, request, al);
+            const auto anErr = new ErrorState(ERR_CANNOT_FORWARD, Http::scServiceUnavailable, request, al);
             fail(anErr);
         } // else use actual error from last connection attempt
 
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -445,7 +445,7 @@ FwdState::useDestinations()
 
         debugs(17, 3, HERE << "Connection failed: " << entry->url());
         if (!err) {
-            const auto anErr = new ErrorState(ERR_CANNOT_FORWARD, Http::scServiceUnavailable, request, al);
+            const auto anErr = new ErrorState(ERR_CANNOT_FORWARD, Http::scInternalServerError, request, al);
             fail(anErr);
         } // else use actual error from last connection attempt
 
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -1078,7 +1078,7 @@ TunnelStateData::noteDestinationsEnd(ErrorState *selectionError)
         if (savedError)
             return sendError(savedError, "all found paths have failed");
 
-        return sendError(new ErrorState(ERR_CANNOT_FORWARD, Http::scServiceUnavailable, request.getRaw(), al),
+        return sendError(new ErrorState(ERR_CANNOT_FORWARD, Http::scInternalServerError, request.getRaw(), al),
                          "path selection found no paths");
     }
     // else continue to use one of the previously noted destinations;


yadij:
> No single code can adequately cover all outcomes here

500 is the generic server error - it covers all possible issues with Squid internals or upstream servers.


> 500 (Internal Server Error) is more precise/focused, but since we do
not actually know whether this was a side effect of a
misconfiguration or earlier connectivity problems (or both!), we
should be using a more general status code.

This does not make sense as a reason to use 503. The HTTP x00 codes are always "more general" status than any other with the same X digit. It is a reason to potentially change the tunnel code to use 500 instead of 503 if it is no longer reliably specific to the 503 meaning.

An actual reason to use 503 is that it allows defined recovery behaviour from the client through the Retry-After header. AFAIK Squid does not do that though.

rousskov:
@yadij, it is not clear to me what you were trying to accomplish with your PR comment. Do you want FwdState to return 500 instead? Do you want us to adjust the PR description to use some other explanation for why tunneling code should return 503 (now done)? Something else? Please be more specific.

> The HTTP x00 codes are always "more general" status than any other with the same X digit

I do not want to argue about this at all, but I want to leave the following explanation for the original "more general" statement in the PR description. We do not need to agree on this. The explanation itself can be removed completely if needed. In hope to make progress, I removed the "more general" statement from the PR description.

I bet many admins would consider "Internal Server Error" more precise/focused than "Service Unavailable". These phrases and RFC code descriptions simply do not allow for nice "503 is 500 with more details" nesting that an "always more general" definition would imply. The RFC does designate 500 as the _default_ meaning for unknown 5xx codes, but many may consider 500's "unexpected condition" meaning and especially its "Internal Error" phrase as not very "general".

yadij:
> @yadij, it is not clear to me what you were trying to accomplish with your PR comment.

I want to have a good clear reason for making this 500 a 503, or making the other 503 a 500. So far the rational (1) is quite weak, and (2) has statements which are false (apparently because of misunderstanding the RFC details).

> 
> > The HTTP x00 codes are always "more general" status than any other with the same X digit
> 
> I do not want to argue about this at all, but I want to leave the following explanation for the original "more general" statement in the PR description. We do not need to agree on this. The explanation itself can be removed completely if needed. In hope to make progress, I removed the "more general" statement from the PR description.
> 

I think we had best "argue" about this. A decision made for visibly incorrect reason is not made better by hiding the reasoning. Correcting the reasoning should lead to a better result.


> I bet many admins would consider "Internal Server Error" more precise/focused than "Service Unavailable".

This is not for admin. The status code being changed is for automation handling the response message. 503 can legitimately enter an infinite loop of retries that 500 is not supposed to. Our choice of status can push clients towards broken behaviour or towards better working - depending on how closely the network conditions match the criteria of the status sent.

FYI: the reason phrase is now deprecated and may not be sent at all.

> These phrases and RFC code descriptions simply do not allow for nice "503 is 500 with more details" nesting that an "always more general" definition would imply.

If I am interpreting that statement right you are looking at the relationship backwards. 503 is a specific set of network error state(s) - it is wrong to use it in other error situations.  500 can be used in all 503 error conditions, plus many others - making it more general / generic.

per RFC7231 section 6.6.1 (status 500)
 "the server encountered an unexpected condition that prevented it from fulfilling the request."
 - this is extremely generic. No detail about the error is indicated for the client to work with. Only that something happened and that the client request itself was apparently not at fault (or it would be 4xx).
 
per RFC7231 section 6.6.4 (status 503)
 "the server is currently unable to handle the request due to a temporary overload or scheduled maintenance"
 - this is more specific. Only problems resulting from overload or pending shutdown should lead to this status. Specific handling by the client is indicated, with headers the server can send to help optimize the client handling.

 --> The error cases cited in (2) include not finding a peer to relay through which is outside the overload or shutting down cases. So 503 is apparently wrong for use here.

eduard-bagdasaryan:
> The error cases cited in (2) include not finding a peer to relay through which is outside the overload or shutting down cases. So 503 is apparently wrong for use here.

The RFC section you cited has a note below, explaining that 503's may be used in situations other from overloading:

> Note: The existence of the 503 status code does not imply that a server has to use it when becoming overloaded. Some servers might simply refuse the connection.

yadij:
> > The error cases cited in (2) include not finding a peer to relay through which is outside the overload or shutting down cases. So 503 is apparently wrong for use here.
> 
> The RFC section you cited has a note below, explaining that 503's may be used in situations other from overloading:
> 
> > Note: The existence of the 503 status code does not imply that a server has to use it when becoming overloaded. Some servers might simply refuse the connection.

No, that is stating that a server may TCP RST/FIN instead of delivering a 503 message. Likewise the server MAY send a 500 status instead of a 503.

rousskov:
If Amos wants 500, let's just use 500. Consistency is far more important here than the specific error code. Arguing with Amos about the meaning of intentionally vague RFC passages is very unlikely to lead to consensus. This particular issue is not worth spending hours and hours on consensus building, especially since the success is unlikely.

@yadij, please pick one code that _you_ think is better, for any reason. The current inconsistency is clear and must be addressed. If you do not have a preference, let's merge this PR as is. I have removed all attempts to justify the current choice in the PR description.

> I think we had best "argue" about this.

We did. We failed to reach consensus. There are so many vague terms and notions being thrown around, that it is very unlikely that we will ever reach it. Let's fix the inconsistency and move on.

The admins that found this inconsistency are OK with either 500 or 503, as long as the inconsistency is fixed. If somebody finds a good reason why we picked the wrong code (whatever it is), they can revisit this decision using specific arguments from their use case.

> A decision made for visibly incorrect reason is not made better by hiding the reasoning. Correcting the reasoning should lead to a better result.

Agreed, but correcting one's reasoning is not feasible and/or too costly in this case. It is not worth the gains. Let's just use one consistent status code that you are most comfortable with, whatever it is. If you do not have a preference, let's merge as is to save time. You can adjust the PR description as needed, of course.

eduard-bagdasaryan:
> If Amos wants 500, let's just use 500

Done at 8f528a.
Please label with 'waiting-for-reviewer'.

rousskov:
> Please label with 'waiting-for-reviewer'.

The current `S-waiting-for-committer` label seems accurate: The reviewer (me) is still OK with your changes, but I am uncomfortable clearing the PR for merging because Amos appeared to dislike this PR at some point in the past. @yadij, please revisit this PR and clear it for merging, request changes, or indicate that you are no longer interested in this PR. In the latter case, I will clear it for merging.

PR #491, opened 2019-10-10
Hash Digest noncedata

squidcontrib:
These commits together
1. Hash the noncedata for Digest nonces before encoding,
   to match the documentation.
2. Encode Digest nonces using hex, rather than base64.

### DIFF
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -21,7 +21,6 @@
 #include "auth/Gadgets.h"
 #include "auth/State.h"
 #include "base/LookupTable.h"
-#include "base64.h"
 #include "cache_cf.h"
 #include "event.h"
 #include "helper.h"
@@ -115,11 +114,8 @@ authDigestNonceEncode(digest_nonce_h * nonce)
     SquidMD5Update(&Md5Ctx, reinterpret_cast<const uint8_t *>(&nonce->noncedata), sizeof(nonce->noncedata));
     SquidMD5Final((unsigned char *) H, &Md5Ctx);
 
-    nonce->key = xcalloc(base64_encode_len(sizeof(H)), 1);
-    struct base64_encode_ctx ctx;
-    base64_encode_init(&ctx);
-    size_t blen = base64_encode_update(&ctx, reinterpret_cast<char*>(nonce->key), sizeof(H), reinterpret_cast<const uint8_t*>(H));
-    blen += base64_encode_final(&ctx, reinterpret_cast<char*>(nonce->key)+blen);
+    nonce->key = xcalloc(sizeof(HASHHEX), 1);
+    CvtHex(H, reinterpret_cast<char *>(nonce->key));
 }
 
 digest_nonce_h *
@@ -154,12 +150,12 @@ authenticateDigestNonceNew(void)
      *
      * Now for my reasoning:
      * We will not accept a unrecognised nonce->we have all recognisable
-     * nonces stored. If we send out unique base64 encodings we guarantee
+     * nonces stored. If we send out unique encodings we guarantee
      * that a given nonce applies to only one user (barring attacks or
      * really bad timing with expiry and creation).  Using a random
      * component in the nonce allows us to loop to find a unique nonce.
      * We use H(nonce_data) so the nonce is meaningless to the reciever.
-     * So our nonce looks like base64(H(timestamp,pointertohash,randomdata))
+     * So our nonce looks like hex(H(timestamp,pointertohash,randomdata))
      * And even if our randomness is not very random we don't really care
      * - the timestamp and memory pointer also guarantee local uniqueness
      * in the input to the hash function.
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -89,7 +89,7 @@ DigestFieldsLookupTable(DIGEST_INVALID_ATTR, DigestAttrs);
  */
 
 static void authenticateDigestNonceCacheCleanup(void *data);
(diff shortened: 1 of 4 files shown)


yadij:
OK to test

src/auth/digest/Config.cc
```diff
@@ -108,11 +108,14 @@ authDigestNonceEncode(digest_nonce_h * nonce)
...
-    nonce->key = xcalloc(base64_encode_len(sizeof(digest_nonce_data)), 1);
-    struct base64_encode_ctx ctx;
-    base64_encode_init(&ctx);
-    size_t blen = base64_encode_update(&ctx, reinterpret_cast<char*>(nonce->key), sizeof(digest_nonce_data), reinterpret_cast<const uint8_t*>(&(nonce->noncedata)));
-    blen += base64_encode_final(&ctx, reinterpret_cast<char*>(nonce->key)+blen);
+    SquidMD5_CTX Md5Ctx;
+    HASH H;
+    SquidMD5Init(&Md5Ctx);
+    SquidMD5Update(&Md5Ctx, reinterpret_cast<const uint8_t *>(&nonce->noncedata), sizeof(nonce->noncedata));
+    SquidMD5Final((unsigned char *) H, &Md5Ctx);
```
yadij: Please avoid C-style casts in his C++ code. Also the H parameter here is an uint8_t[] in the md5.h API.

src/auth/digest/Config.cc
```diff
@@ -108,11 +108,14 @@ authDigestNonceEncode(digest_nonce_h * nonce)
...
+
+    nonce->key = xcalloc(sizeof(HASHHEX), 1);
+    CvtHex(H, reinterpret_cast<char *>(nonce->key));
```
yadij: The hash_link::key is a void* type. If you have to cast at all a static_cast should be sufficient.

squidcontrib: I changed it to `static_cast<char *>`. Not casting gave an error:
```
Config.cc: In function ‘void authDigestNonceEncode(digest_nonce_h*)’:
Config.cc:118:22: error: invalid conversion from ‘void*’ to ‘char*’ [-fpermissive]
     CvtHex(H, nonce->key);
               ~~~~~~~^~~
```

src/auth/digest/Config.cc
squidcontrib: I changed this to a `reinterpret_cast<uint8_t *>`.

The API in [md5.h](https://github.com/squid-cache/squid/blob/SQUID_4_8/include/md5.h#L63) specifies `uint8_t[16]`, but in [md5.c](https://github.com/squid-cache/squid/blob/SQUID_4_8/lib/md5.c#L129-L130) it's `unsigned char[16]`.

```
SQUIDCEXTERN void SquidMD5Final(uint8_t digest[16], struct SquidMD5Context *context);
```

```
void
SquidMD5Final(unsigned char digest[16], struct SquidMD5Context *ctx)
```

squidcontrib:
I just pushed [another change](https://github.com/squidcontrib/squid/commit/5e3d96aad8709eb7b10b0de4867a77f16cc5526b) to rename identifiers: `UserRequest.nonceb64` to `UserRequest.noncehex` and `authenticateDigestNonceNonceb64` to `authenticateDigestNonceNonceHex`. I missed these earlier because I only searched for "base64", not "b64". Strangely, the [documentation comment](https://github.com/squid-cache/squid/blob/SQUID_4_8/src/auth/digest/UserRequest.h#L47) for `UserRequest.nonceb64` was already a hex string.

src/auth/digest/Config.cc
yadij: Hmm, okay, that can be fixed later then.

src/auth/digest/Config.cc
```diff
@@ -329,18 +332,18 @@ authenticateDigestNonceNonceb64(const digest_nonce_h * nonce)
 }
 
 static digest_nonce_h *
-authenticateDigestNonceFindNonce(const char *nonceb64)
+authenticateDigestNonceFindNonce(const char *noncehex)
 {
     digest_nonce_h *nonce = NULL;
 
-    if (nonceb64 == NULL)
+    if (noncehex == NULL)
```
yadij:
```suggestion
    if !(noncehex)
```

src/auth/digest/Config.cc
```diff
@@ -329,18 +332,18 @@ authenticateDigestNonceNonceb64(const digest_nonce_h * nonce)
...
         return NULL;
 
-    debugs(29, 9, "looking for nonceb64 '" << nonceb64 << "' in the nonce cache.");
+    debugs(29, 9, "looking for noncehex '" << noncehex << "' in the nonce cache.");
 
-    nonce = static_cast < digest_nonce_h * >(hash_lookup(digest_nonce_cache, nonceb64));
+    nonce = static_cast < digest_nonce_h * >(hash_lookup(digest_nonce_cache, noncehex));
 
-    if ((nonce == NULL) || (strcmp(authenticateDigestNonceNonceb64(nonce), nonceb64)))
+    if ((nonce == NULL) || (strcmp(authenticateDigestNonceNonceHex(nonce), noncehex)))
```
yadij:
```suggestion
    if (!nonce || strcmp(authenticateDigestNonceNonceHex(nonce), noncehex) != 0)
```

src/auth/digest/UserRequest.cc
```diff
@@ -23,7 +23,7 @@
 #include "SquidTime.h"
 
 Auth::Digest::UserRequest::UserRequest() :
-    nonceb64(NULL),
+    noncehex(NULL),
```
yadij:
```suggestion
    noncehex(nullptr),
```

PR #495, opened 2019-10-17
Fix build after ccfbe8f (Report context of cache.log messages)

rousskov:
In some environments (e.g., Ubuntu 14 without libcppunit-dev), GCC
correctly complained that it does not know how to print "[unknown]",
probably because none of the other headers imported `<ostream>` before
importing base/InstanceId.h.

### DIFF
--- a/src/base/InstanceId.cc
+++ b/src/base/InstanceId.cc
@@ -1,0 +1,25 @@
+/*
+ * Copyright (C) 1996-2019 The Squid Software Foundation and contributors
+ *
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#include "squid.h"
+#include "base/InstanceId.h"
+
+#include <iostream>
+
+std::ostream&
+operator <<(std::ostream &os, const ScopedId &id)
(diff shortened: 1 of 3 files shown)


yadij:
Is there no way we can do this as a template member? I do not insist on this, but it would be great to use these IDs in helpers etc without pulling libbase.la dependency into them.

rousskov:
> Is there no way we can do this as a template member? 

Do _what_ as a template member? ScopeId is not and cannot be a templated class. The print/shift _operator_ cannot be a member IIRC.
Feels increading wrong to d
If you are asking to keep the print/shift operator inlined, then yes, we can easily do that, at the expense of explicitly adding `#include <iostream>` to InstanceId.h and, hence, to lots of other Squid header files. Such "leaking" of a large implementation code like iostream into header files is considered "wrong", of course, but we can violate those norms if really needed.

> it would be great to use these IDs in helpers etc without pulling libbase.la dependency into them.

Perhaps libbase and similar low-level primitives should be made available to helpers that need them? Depriving increasingly complex helpers from increasingly reusable base code feels increasingly wrong. IIRC, we have discussed this in SBuf context, but I do not remember if we reached any consensus.

yadij:
Sorry, misread what was going on in the patch. LGTM

PR #498, opened 2019-10-21
Improved handling of cachemgr.cgi parameters

aaron-costello:
Increased browser safety for those that support this header.

### DIFF
--- a/tools/cachemgr.cc
+++ b/tools/cachemgr.cc
@@ -245,6 +245,8 @@ auth_html(const char *host, int port, const char *user_name)
     if (fp == NULL)
         printf("X-Error: message=\"Unable to open config %s\"", DEFAULT_CACHEMGR_CONFIG);
 
+    printf("X-XSS-Protection: 1\r\n");
+    
     printf("Content-Type: text/html\r\n\r\n");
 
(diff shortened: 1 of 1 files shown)


yadij:
I do not think this is a good solution to the problem of parameter handling. The header relies on additional configuration at multiple places - at least the client browser settings are vital to it having any effect at all and outside the control of any interested admin.

aaron-costello:
@yadij I understand that viewpoint. However, there continues to be XSS issues affecting the cachemgr.cgi as the previous fixes weren't sufficient. I do not have experience with advanced filtering HTML characters appropriately in C/CPP, but a revision of the fixes in [429](https://github.com/squid-cache/squid/pull/429) is needed as it is by passable.

yadij:
That PR 429 was scoped to the authentication credentials, it is not relevant to any others. If XSS issues remain elsewhere we need them to be fixed, not simply hidden from view.

FYI: The cachemgr reports may contain arbitrary URLs, snippets of object contents from cache, or be replaced in whole with proxy error pages containing dynamic scripts and links to objects in the proxy origin rather than the CGI origin. So approaches like this header that universally prevent these features working are going to break some subset of working installation.

I will followup to the email thread with a suggestion about the parameter you mentioned already. If you have found others, please mention them in the thread and we can look at each as a separate case.

aaron-costello:
In that case since the other parameters appear to be sanitized effectively, it would be simplest to remove the verbose error that's being displayed. I believe "Unknown Host" is a clear enough error without reflecting the parameter too.

PR #501, opened 2019-10-28
Fix expiration of self-signed generated certs to be 3 years

jdef:
Generated self-signed certificates lifetime was incorrectly set to 
356*3 days, which is approximately 27 days shorter than 3 years.

The fixed problem does not affect Squids using "sslproxy_cert_adapt
setValidAfter" and Squids using a configured signing CA certificate.

### DIFF
--- a/src/ssl/gadgets.cc
+++ b/src/ssl/gadgets.cc
@@ -504,7 +504,7 @@ static bool buildCertificate(Security::CertPointer & cert, Ssl::CertificatePrope
     if (aTime) {
         if (!X509_set1_notAfter(cert.get(), aTime))
             return false;
-    } else if (!X509_gmtime_adj(X509_getm_notAfter(cert.get()), 60*60*24*356*3))
+    } else if (!X509_gmtime_adj(X509_getm_notAfter(cert.get()), 60*60*24*365*3))
         return false;
 
     int addedExtensions = 0;


yadij:
OK to test

rousskov:
@jdef, could you please detail under what conditions this bug was triggered and what its effects were? I could not guess quickly based on the code changes and Squid admins need more info to understand whether the bug affects them and, if yes, what will change with this fix.

jdef:
sure, `http_port` docs mention 3 years for generated certs. 1 year is 365 days, not 356 days.
```
	   generate-host-certificates[=<on|off>]
			Dynamically create SSL server certificates for the
			destination hosts of bumped CONNECT requests.When 
			enabled, the cert and key options are used to sign
			generated certificates. Otherwise generated
			certificate will be selfsigned.
			If there is a CA certificate lifetime of the generated 
			certificate equals lifetime of the CA certificate. If
			generated certificate is selfsigned lifetime is three 
			years.
```

rousskov:
@jdef, thank you for pointing me to the related documentation. Please check that the adjusted PR title/description match the fix. Adjust as needed. PR title/description will become an official commit message when this PR is merged.

rousskov:
We may also want to document that previously cached generated certificates will continue to be served with the wrong expiration time (if that is indeed the case).

rousskov:
> @jdef, thank you for pointing me to the related documentation. Please check that the adjusted PR title/description match the fix. Adjust as needed. PR title/description will become an official commit message when this PR is merged.

@jdef, just FYI: I am not clearing this PR for merge to give you a chance to validate the adjusted description.

jdef:
Looks good to me

PR #504, opened 2019-11-02
cachemgr.cgi: Add validation for hostname parameter

aaron-costello:
Prevention of HTML/invalid chars in host param

PR #506, opened 2019-11-05
cachemgr.cgi: Add validation for hostname parameter (#504)

squidadm:
Prevention of HTML/invalid chars in host param


yadij:
OK to test

PR #508, opened 2019-11-10
Docs: mention --foreground in squid -z man page

marcosfrm:

PR #509, opened 2019-11-11
Happy Eyeballs: Do not wait for already exhausted spares

eduard-bagdasaryan:
ResolvedPeers::findSpareOrNextPeer() never returned the next peer. That
bug made doneWithSpares() return false when the next peer was present,
even though no spares were expected. The exact effects of this bug are
not known, but wrong false doneWithSpares() outcome may force Squid to
wait for more spare addresses that would never come, possibly slowing
down transactions.

### DIFF
--- a/src/ResolvedPeers.cc
+++ b/src/ResolvedPeers.cc
@@ -65,14 +65,11 @@ ResolvedPeers::findSpareOrNextPeer(const Comm::Connection &currentPeer)
     const auto familyToAvoid = ConnectionFamily(currentPeer);
     // Optimization: Also stop at the first mismatching peer because all
     // same-peer paths are grouped together.
-    auto found = std::find_if(paths_.begin(), paths_.end(),
+    return std::find_if(paths_.begin(), paths_.end(),
     [peerToMatch, familyToAvoid](const Comm::ConnectionPointer &conn) {
         return peerToMatch != conn->getPeer() ||
                familyToAvoid != ConnectionFamily(*conn);
     });
-    if (found != paths_.end() && peerToMatch == (*found)->getPeer())
-        return found;
-    return paths_.end();
 }
 
 Comm::ConnectionPointer


rousskov:
Your branch name implies that some destinations could be missed, but the original PR description did not state that explicitly (and was rather vague as far as observable bug outcome is concerned). The branch name is _not_ important, but documenting how the bug affected Squid behavior is.

For example, we may need to fix our tests if buggy Squid could, in fact, miss some forwarding paths. If an additional delay is the only known problem, then tests may be OK (we can detect when Squid moves too fast, but we do not know how to reliably detect when Squid is a bit slower because factors outside the test control often slow Squid down).

eduard-bagdasaryan:
It is difficult for me to foretell all the consequences of this bug just looking at the code, which is complex and filled with many conditions. It looks like ResolvedPeers::doneWithSpares() is the only caller of findSpareOrNextPeer() which depends on 'next peer' outcome. For example, it could sometimes give wrong 'false' instead of 'true' (when some of the destinations have not been analyzed yet). This wrong result would confuse either maybeGivePrimeItsChance() or checkForNewConnection(), probably resulting in unexpected delays (until the notification of the next available destination) - as you suggested. Could this problem lead to a worse case, when both these methods are not called and some of the available destinations are ignored? An open question still.

rousskov:
I will rephrase the question: Do you have a specific test/use case that exposed this bug? If yes, how did Squid misbehaved in that case? Did Squid miss a destination or just delayed its use?

I am not asking for a comprehensive analysis of possible outcomes of this bug -- just for a disclosure of the outcomes you have already observed (if any). If you have found this bug by working on the code and have no corresponding test/use case, that is perfectly fine too, of course.

eduard-bagdasaryan:
> If you have found this bug by working on the code and have no corresponding test/use case

Exactly.

rousskov:
OK, I documented our lack of knowledge in the PR description and cleared the PR for merge.

rousskov:
@eduard-bagdasaryan, if is there a way to kick Jenkins without making new commits, please do that. Jenkins started to get stuck a few days ago. I killed stuck builds but my attempt to rebuild them failed. I hope you know a Jenkins trick that can work here.

yadij:
Click the DETAILS link from here, then the RETRY button on the Jenkins page which comes up.

There are two caveats:
* you must be logged in to Jenkins for that button to appear.
* the test is queued behind all other PR-test jobs. It will not always start immediately.

rousskov:
> RETRY button on the Jenkins page

Thank you. I probably tried the REBUILD button before asking the question. The button I tried did not work for me. I will try to remember to try the RETRY button next time this happens.

rousskov:
Five Jenkins tests [failed](https://build.squid-cache.org/job/5-pr-test/1491/). AFAICT, all of the failed tests failed for reasons unrelated to this PR. Did somebody upgrade Jenkins hosts again, triggering new build failures in old/official code? PR #508 has a couple of similar PR-unrelated failures but 508 [test](https://build.squid-cache.org/job/5-pr-test/1490/) lacks `d-ubuntu-disco` and `d-ubuntu-eoan` nodes. Do we need to mark more Jenkins nodes/configurations as "failure OK"?

yadij:
The failures need fixing. The eoan sysctl.h one I have a patch for already and will PR shortly. The clang ones are a bit more tricky to test since we dont have PR 154 merged yet.

rousskov:
> The failures need fixing. The eoan sysctl.h one I have a patch for already and will PR shortly. The clang ones are a bit more tricky to test since we dont have PR 154 merged yet.

If you are implying that this PR should wait for some other PRs to fix these build failures, then I think that would be wrong because this PR has nothing to do with the failures. If you are implying some other course of action, please clarify.

yadij:
Are you suggesting we commit untested code rather than fixing the tests first?

rousskov:
> Are you suggesting we commit untested code rather than fixing the tests first?

I suggest that we commit hand-picked tested-enough code despite known-to-be-irrelevant build failures. Doing so does not preclude us from fixing whatever needs to be fixed, of course.

However, if you prefer to stall ready-for-merge PRs while we resolve these new build failures, I can play along because there are no emergency PRs right now AFAICT. We just need a plan we all can agree on and follow to make progress. Its specifics are less important.

PR #510, opened 2019-11-13
Add http_port sslflags=CONDITIONAL_AUTH

trapexit:
Enabling this flag removes SSL_VERIFY_FAIL_IF_NO_PEER_CERT from the
SSL_CTX_set_verify callback. Meaning a client certificate verify
occurs iff provided.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -2365,8 +2365,10 @@ DOC_START
 				immediately, but wait until acl processing
 				requires a certificate (not yet implemented).
 			    CONDITIONAL_AUTH
-				Do not fail auth if client certificate is
-				not provided.
+				Request a client certificate during the TLS
+				handshake, but ignore certificate absence in
+				the TLS client Hello. If the client does
+				supply a certificate, it is validated.
 			    NO_SESSION_REUSE
 				Don't allow for session reuse. Each connection
 				will result in a new SSL session.
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -2364,6 +2364,9 @@ DOC_START
 				Don't request client certificates
 				immediately, but wait until acl processing
 				requires a certificate (not yet implemented).
+			    CONDITIONAL_AUTH
+				Do not fail auth if client certificate is
+				not provided.
 			    NO_SESSION_REUSE
 				Don't allow for session reuse. Each connection
 				will result in a new SSL session.
--- a/src/security/PeerOptions.cc
+++ b/src/security/PeerOptions.cc
@@ -539,7 +539,7 @@ Security::PeerOptions::parseOptions()
 /**
  * Parses the TLS flags squid.conf parameter
  */
-long
+Security::ParsedPortFlags
 Security::PeerOptions::parseFlags()
 {
     if (sslFlags.isEmpty())
@@ -547,7 +547,7 @@ Security::PeerOptions::parseFlags()
 
     static struct {
         SBuf label;
-        long mask;
+        ParsedPortFlags mask;
     } flagTokens[] = {
         { SBuf("NO_DEFAULT_CA"), SSL_FLAG_NO_DEFAULT_CA },
         { SBuf("DELAYED_AUTH"), SSL_FLAG_DELAYED_AUTH },
@@ -565,9 +565,9 @@ Security::PeerOptions::parseFlags()
     ::Parser::Tokenizer tok(sslFlags);
     static const CharacterSet delims("Flag-delimiter", ":,");
 
-    long fl = 0;
+    ParsedPortFlags fl = 0;
     do {
-        long found = 0;
+        ParsedPortFlags found = 0;
         for (size_t i = 0; flagTokens[i].mask; ++i) {
             if (tok.skip(flagTokens[i].label)) {
                 found = flagTokens[i].mask;
--- a/src/security/PeerOptions.cc
+++ b/src/security/PeerOptions.cc
@@ -569,6 +569,7 @@ Security::PeerOptions::parseFlags()
     do {
         ParsedPortFlags found = 0;
         for (size_t i = 0; flagTokens[i].mask; ++i) {
+            // XXX: skips FOO in FOOBAR, missing merged flags and trailing typos
             if (tok.skip(flagTokens[i].label)) {
                 found = flagTokens[i].mask;
                 break;
@@ -587,7 +588,7 @@ Security::PeerOptions::parseFlags()
 
     if ((fl & (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH)) ==
         (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH))
-        fatal("ERROR: DONT_VERIFY_PEER and CONDITIONAL_AUTH are mutually exclusive");
+        throw TextException("DONT_VERIFY_PEER and CONDITIONAL_AUTH are mutually exclusive", Here());
 
     return fl;
 }
--- a/src/security/PeerOptions.cc
+++ b/src/security/PeerOptions.cc
@@ -552,6 +552,7 @@ Security::PeerOptions::parseFlags()
         { SBuf("NO_DEFAULT_CA"), SSL_FLAG_NO_DEFAULT_CA },
         { SBuf("DELAYED_AUTH"), SSL_FLAG_DELAYED_AUTH },
         { SBuf("DONT_VERIFY_PEER"), SSL_FLAG_DONT_VERIFY_PEER },
+        { SBuf("CONDITIONAL_AUTH"), SSL_FLAG_CONDITIONAL_AUTH },
         { SBuf("DONT_VERIFY_DOMAIN"), SSL_FLAG_DONT_VERIFY_DOMAIN },
         { SBuf("NO_SESSION_REUSE"), SSL_FLAG_NO_SESSION_REUSE },
 #if X509_V_FLAG_CRL_CHECK
@@ -583,6 +584,10 @@ Security::PeerOptions::parseFlags()
         } else
             fl |= found;
     } while (tok.skipOne(delims));
+
+    if ((fl & (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH)) ==
+        (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH))
+        fatal("ERROR: DONT_VERIFY_PEER and CONDITIONAL_AUTH are mutually exclusive");
 
     return fl;
 }
--- a/src/security/PeerOptions.cc
+++ b/src/security/PeerOptions.cc
@@ -19,6 +19,8 @@
 #include "ssl/support.h"
 #endif
 
+#include <bitset>
+
 Security::PeerOptions Security::ProxyOutgoingConfig;
 
 Security::PeerOptions::PeerOptions()
@@ -586,9 +588,17 @@ Security::PeerOptions::parseFlags()
             fl |= found;
     } while (tok.skipOne(delims));
 
-    if ((fl & (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH)) ==
-        (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH))
-        throw TextException("DONT_VERIFY_PEER and CONDITIONAL_AUTH are mutually exclusive", Here());
+    const auto mutuallyExclusive =
+        SSL_FLAG_DONT_VERIFY_PEER|
+        SSL_FLAG_DELAYED_AUTH|
+        SSL_FLAG_CONDITIONAL_AUTH;
+    typedef std::bitset<sizeof(decltype(fl))> ParsedPortFlagBits;
+    if (ParsedPortFlagBits(fl & mutuallyExclusive).count() > 1) {
+        if (fl & SSL_FLAG_CONDITIONAL_AUTH)
+            throw TextException("CONDITIONAL_AUTH is not compatible with NO_DEFAULT_CA and DELAYED_AUTH flags", Here());
+        debugs(83, DBG_PARSE_NOTE(DBG_IMPORTANT), "WARNING: Mixtures of incompatible TLS flags" <<
+               " are deprecated and will become a fatal configuration error");
+    }
 
     return fl;
 }
--- a/src/security/ServerOptions.cc
+++ b/src/security/ServerOptions.cc
@@ -430,20 +430,13 @@ Security::ServerOptions::updateContextClientCa(Security::ContextPointer &ctx)
             return;
         }
 
-        if (parsedFlags & SSL_FLAG_DELAYED_AUTH) {
-            debugs(83, 9, "Not requesting client certificates until acl processing requires one");
-            SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_NONE, nullptr);
-        } else {
-            debugs(83, 9, "Requiring client certificates.");
-            Ssl::SetupVerifyCallback(ctx);
-        }
+        Ssl::ConfigurePeerVerification(ctx, parsedFlags);
 
         updateContextCrl(ctx);
         updateContextTrust(ctx);
 
     } else {
-        debugs(83, 9, "Not requiring any client certificates");
-        SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_NONE, NULL);
+        Ssl::DisablePeerVerification(ctx);
     }
 #endif
 }
--- a/src/ssl/support.cc
+++ b/src/ssl/support.cc
@@ -413,7 +413,7 @@ Ssl::ConfigurePeerVerification(Security::ContextPointer &ctx, const Security::Pa
         mode = SSL_VERIFY_PEER|SSL_VERIFY_FAIL_IF_NO_PEER_CERT;
     }
 
-    SSL_CTX_set_verify(ctx.get(),mode,(mode != SSL_VERIFY_NONE) ? ssl_verify_cb : nullptr);
+    SSL_CTX_set_verify(ctx.get(), mode, (mode != SSL_VERIFY_NONE) ? ssl_verify_cb : nullptr);
 }
 
 void
@@ -1330,3 +1330,4 @@ BIO *Ssl::BIO_new_SBuf(SBuf *buf)
 }
 
 #endif /* USE_OPENSSL */
+
--- a/src/ssl/support.cc
+++ b/src/ssl/support.cc
@@ -390,7 +390,7 @@ ssl_verify_cb(int ok, X509_STORE_CTX * ctx)
 }
 
 void
-Ssl::ConfigurePeerVerification(Security::ContextPointer &ctx, int flags)
+Ssl::ConfigurePeerVerification(Security::ContextPointer &ctx, const Security::ParsedPortFlags flags)
 {
     int mode;
 
@@ -559,7 +559,7 @@ Ssl::InitServerContext(Security::ContextPointer &ctx, AnyP::PortCfg &port)
 }
 
 bool
-Ssl::InitClientContext(Security::ContextPointer &ctx, Security::PeerOptions &peer, long fl)
+Ssl::InitClientContext(Security::ContextPointer &ctx, Security::PeerOptions &peer, Security::ParsedPortFlags fl)
 {
     if (!ctx)
         return false;
--- a/src/ssl/support.cc
+++ b/src/ssl/support.cc
@@ -390,9 +390,40 @@ ssl_verify_cb(int ok, X509_STORE_CTX * ctx)
 }
 
 void
-Ssl::SetupVerifyCallback(Security::ContextPointer &ctx)
+Ssl::ConfigurePeerVerification(Security::ContextPointer &ctx, int flags)
 {
-    SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, ssl_verify_cb);
+    int mode;
+
+    if ((flags & (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH)) ==
+        (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH))
+        throw TextException("ERROR: DONT_VERIFY_PEER and CONDITIONAL_AUTH are mutually exclusive", Here());
+
+    if (flags & SSL_FLAG_DONT_VERIFY_PEER) {
+        debugs(83, DBG_IMPORTANT, "SECURITY WARNING: Peer certificates are not verified for validity!");
+        debugs(83, DBG_IMPORTANT, "UPGRADE NOTICE: The DONT_VERIFY_PEER flag is deprecated. Remove the clientca= option to disable client certificates.");
+        mode = SSL_VERIFY_NONE;
+    }
+    else if (flags & SSL_FLAG_DELAYED_AUTH) {
+        debugs(83, DBG_PARSE_NOTE(3), "not requesting client certificates until ACL processing requires one");
+        mode = SSL_VERIFY_NONE;
+    }
+    else if (flags & SSL_FLAG_CONDITIONAL_AUTH) {
+        debugs(83, DBG_PARSE_NOTE(3), "will request the client certificate but ignore its absense");
+        mode = SSL_VERIFY_PEER;
+    }
+    else {
+        debugs(83, DBG_PARSE_NOTE(3), "Requiring client certificates.");
+        mode = SSL_VERIFY_PEER|SSL_VERIFY_FAIL_IF_NO_PEER_CERT;
+    }
+
+    SSL_CTX_set_verify(ctx.get(),mode,(mode != SSL_VERIFY_NONE) ? ssl_verify_cb : nullptr);
+}
+
+void
+Ssl::DisablePeerVerification(Security::ContextPointer &ctx)
+{
+    debugs(83, DBG_PARSE_NOTE(3), "Not requiring any client certificates");
+    SSL_CTX_set_verify(ctx.get(),SSL_VERIFY_NONE,nullptr);
 }
 
 // "dup" function for SSL_get_ex_new_index("cert_err_check")
@@ -579,13 +610,7 @@ Ssl::InitClientContext(Security::ContextPointer &ctx, Security::PeerOptions &pee
 
     MaybeSetupRsaCallback(ctx);
 
-    if (fl & SSL_FLAG_DONT_VERIFY_PEER) {
-        debugs(83, 2, "SECURITY WARNING: Peer certificates are not verified for validity!");
-        SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_NONE, NULL);
-    } else {
-        debugs(83, 9, "Setting certificate verification callback.");
-        Ssl::SetupVerifyCallback(ctx);
-    }
+    Ssl::ConfigurePeerVerification(ctx, fl);
 
     return true;
 }
@@ -1308,4 +1333,3 @@ BIO *Ssl::BIO_new_SBuf(SBuf *buf)
 }
 
 #endif /* USE_OPENSSL */
-
--- a/src/ssl/support.cc
+++ b/src/ssl/support.cc
@@ -394,10 +394,7 @@ Ssl::ConfigurePeerVerification(Security::ContextPointer &ctx, const Security::Pa
 {
     int mode;
 
-    if ((flags & (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH)) ==
-        (SSL_FLAG_DONT_VERIFY_PEER|SSL_FLAG_CONDITIONAL_AUTH))
-        throw TextException("ERROR: DONT_VERIFY_PEER and CONDITIONAL_AUTH are mutually exclusive", Here());
-
+    // assume each flag is exclusive; flags creator must check this assumption
     if (flags & SSL_FLAG_DONT_VERIFY_PEER) {
         debugs(83, DBG_IMPORTANT, "SECURITY WARNING: Peer certificates are not verified for validity!");
         debugs(83, DBG_IMPORTANT, "UPGRADE NOTICE: The DONT_VERIFY_PEER flag is deprecated. Remove the clientca= option to disable client certificates.");
--- a/src/ssl/support.h
+++ b/src/ssl/support.h
@@ -86,7 +86,8 @@ bool InitServerContext(Security::ContextPointer &, AnyP::PortCfg &);
 bool InitClientContext(Security::ContextPointer &, Security::PeerOptions &, long flags);
 
 /// set the certificate verify callback for a context
-void SetupVerifyCallback(Security::ContextPointer &);
+void ConfigurePeerVerification(Security::ContextPointer &, int);
+void DisablePeerVerification(Security::ContextPointer &);
 
 /// if required, setup callback for generating ephemeral RSA keys
 void MaybeSetupRsaCallback(Security::ContextPointer &);
--- a/src/ssl/support.h
+++ b/src/ssl/support.h
@@ -83,10 +83,10 @@ typedef RefCount<CertValidationResponse> CertValidationResponsePointer;
 bool InitServerContext(Security::ContextPointer &, AnyP::PortCfg &);
 
 /// initialize a TLS client context with OpenSSL specific settings
-bool InitClientContext(Security::ContextPointer &, Security::PeerOptions &, long flags);
+bool InitClientContext(Security::ContextPointer &, Security::PeerOptions &, Security::ParsedPortFlags);
 
 /// set the certificate verify callback for a context
-void ConfigurePeerVerification(Security::ContextPointer &, int);
+void ConfigurePeerVerification(Security::ContextPointer &, const Security::ParsedPortFlags);
 void DisablePeerVerification(Security::ContextPointer &);
 
 /// if required, setup callback for generating ephemeral RSA keys
--- a/src/security/PeerOptions.h
+++ b/src/security/PeerOptions.h
@@ -98,7 +98,7 @@ private:
     bool optsReparse = true;
 
 public:
-    ParsedPortFlags parsedFlags = 0;   ///< parsed value of sslFlags
+    ParsedPortFlags parsedFlags = 0; ///< parsed value of sslFlags
 
     std::list<Security::KeyData> certs; ///< details from the cert= and file= config parameters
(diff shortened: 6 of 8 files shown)


src/security/PeerOptions.cc
```diff
@@ -552,6 +552,7 @@ Security::PeerOptions::parseFlags()
         { SBuf("NO_DEFAULT_CA"), SSL_FLAG_NO_DEFAULT_CA },
         { SBuf("DELAYED_AUTH"), SSL_FLAG_DELAYED_AUTH },
         { SBuf("DONT_VERIFY_PEER"), SSL_FLAG_DONT_VERIFY_PEER },
+        { SBuf("CONDITIONAL_VERIFY"), SSL_FLAG_CONDITIONAL_VERIFY },
```
yadij: For consistency with the DELAYED_AUTH flag this should probably be named OPTIONAL_AUTH.

yadij:
OK to test

src/ssl/support.cc
```diff
@@ -395,6 +395,12 @@ Ssl::SetupVerifyCallback(Security::ContextPointer &ctx)
     SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, ssl_verify_cb);
 }
 
+void
+Ssl::SetupConditionalVerifyCallback(Security::ContextPointer &ctx)
```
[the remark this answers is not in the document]
rousskov: Actually, please remove all existing SSL_CTX_set_verify() calls outside SetupVerifyCallback() by adding a `flags` parameter to Ssl::SetupVerifyCallback(). I would also rename SetupVerifyCallback() to SetupPeerVerification() or ConfigurePeerVerification() but that is optional.

src/security/PeerOptions.cc
trapexit: Alright. I wasn't sure about the name. "optional" like NGINX feels odd given it makes it sound like the verification is optional rather than being based on existence... but the SSL flag felt too verbose.

rousskov: I agree that `_AUTH` suffix is a good idea and that "optional" is a bit misleading. I can suggest `CONDITIONAL_AUTH` or even `IF_RECEIVED_CERTIFICATE_AUTH`. Pick what you think works best.

src/ssl/support.cc
```diff
@@ -390,9 +390,17 @@ ssl_verify_cb(int ok, X509_STORE_CTX * ctx)
 }
 
 void
-Ssl::SetupVerifyCallback(Security::ContextPointer &ctx)
+Ssl::SetupPeerVerification(Security::ContextPointer &ctx, bool verify, bool fail_if_no_peer_cert)
 {
-    SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, ssl_verify_cb);
+  int mode;
+
+  mode = SSL_VERIFY_NONE;
```
yadij: No need for the C-style initialization and empty line above.
```suggestion
    int mode = SSL_VERIFY_NONE;
```

Please also indent the lines below properly: 4-spaces per level of indentation, and one space between "if" and their condition brackets.

src/ssl/support.h
```diff
@@ -86,7 +86,7 @@ bool InitServerContext(Security::ContextPointer &, AnyP::PortCfg &);
 bool InitClientContext(Security::ContextPointer &, Security::PeerOptions &, long flags);
 
 /// set the certificate verify callback for a context
-void SetupVerifyCallback(Security::ContextPointer &);
+void SetupPeerVerification(Security::ContextPointer &, bool, bool);
```
rousskov: You have decided to use two boolean parameters instead of one flags parameter. Correct usage (and reading/review) of these two boolean parameters requires familiarization with the SetupPeerVerification() API details (well, in your particular implementation, it actually requires learning about that API _implementation_, but that problem is trivial to fix if we stay with this design). It is easy to accidentally swap them and/or copy the wrong old parameter when writing new code because many developers and reviewers are unlikely to notice such mistakes. Such mistakes have happened in the past with similar APIs.

Is it very difficult to use a single flags parameter instead? If not, please do that.

src/security/ServerOptions.cc
```diff
@@ -432,18 +432,21 @@ Security::ServerOptions::updateContextClientCa(Security::ContextPointer &ctx)
 
         if (parsedFlags & SSL_FLAG_DELAYED_AUTH) {
             debugs(83, 9, "Not requesting client certificates until acl processing requires one");
-            SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_NONE, nullptr);
+            Ssl::SetupPeerVerification(ctx, false, false);
+        } else if(parsedFlags & SSL_FLAG_CONDITIONAL_AUTH) {
+            debugs(83, 9, "Allowing client certificates.");
```
rousskov: A few polishing touches here:

1. I have not checked, but I suspect that client certificates are, technically, always "allowed" -- the server cannot stop the client from supplying more information. Let's see if we can improve this wording. I have a specific suggestion below.
2. Please use debugs() level 5 here, despite the inconsistency wit similar debugs() statements nearby. Level 9 is reserved for data dumps in new code. 
3. Also, please do not format routine low-level debugging as a sentence, for brevity and consistency sake (again, the other nearby debugs() statements are wrong, and it is overall better not to mimic them here).

```suggestion
            debugs(83, 5, "will request the client certificate but ignore its absence");
```

src/ssl/support.cc
```diff
@@ -390,9 +390,17 @@ ssl_verify_cb(int ok, X509_STORE_CTX * ctx)
...
+  if(verify)
+    mode |= SSL_VERIFY_PEER;
```
rousskov: My reading of OpenSSL [docs](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_verify.html) suggests that SSL_VERIFY_PEER and SSL_VERIFY_FAIL_IF_NO_PEER_CERT should not be added to SSL_VERIFY_NONE, even it it happens to work in your tests:

> If the mode is SSL_VERIFY_NONE none of the other flags may be set.

src/ssl/support.cc
```diff
@@ -390,9 +390,17 @@ ssl_verify_cb(int ok, X509_STORE_CTX * ctx)
...
 
 void
-Ssl::SetupVerifyCallback(Security::ContextPointer &ctx)
+Ssl::SetupPeerVerification(Security::ContextPointer &ctx, bool verify, bool fail_if_no_peer_cert)
 {
-    SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, ssl_verify_cb);
+  int mode;
+
+  mode = SSL_VERIFY_NONE;
+  if(verify)
```
rousskov: My reading of OpenSSL [docs](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_verify.html) suggests that SSL_VERIFY_FAIL_IF_NO_PEER_CERT should have SSL_VERIFY_PEER set as well (emphasis mine):

> SSL_VERIFY_PEER [...] the server sends a client certificate request to the client. The certificate returned (if any) is checked. If the verification process fails, the TLS/SSL handshake is immediately terminated with an alert message containing the reason for the verification failure. The behaviour can be controlled by the **additional** SSL_VERIFY_FAIL_IF_NO_PEER_CERT, SSL_VERIFY_CLIENT_ONCE and SSL_VERIFY_POST_HANDSHAKE flags.

If I am right, then fixing this properly would require changing this function boolean parameters (and/or their handling) to make impossible (or reject) an invalid `(false, true)` combinations. I am not suggesting a specific fix because I hope that those boolean parameters will be gone as discussed in another change request.

src/cf.data.pre
```diff
@@ -2364,6 +2364,9 @@ DOC_START
 				Don't request client certificates
 				immediately, but wait until acl processing
 				requires a certificate (not yet implemented).
+                            CONDITIONAL_AUTH
+                                Do not fail auth if client certificate is
```
rousskov: Please see if you can keep the weird formatting consistent by _copying_ the DELAYED_AUTH blob above and preserving copied leading spacing in the new blob. I am sorry we have to do this. Someday, somebody will convert `cf.data.pre` to a more developer-friendly format.

src/ssl/support.h
trapexit: I'm trying to limit the API to limit the possibility of bad values needing some sort of error handling which would require a more significant change. How about an enum so at least there would be warnings if misused? If using flags might as well just use the SSL flags.

rousskov: I hope there is some misunderstanding here. I reordered your statements to clarify my suggestion first.

> If using flags might as well just use the SSL flags.

I am not sure what you mean by "SSL flags" but my suggestion was (and still is) to pass `parsedFlags`. As is. Instead of two post-flags-analysis booleans.  SetupPeerVerification() itself will perform all the necessary analysis to form the right OpenSSL mode parameter. Sorry if that was not clear. AFAICT, the code will look much simpler after that change and addressing other problems mentioned in my review may become easier as well.

> I'm trying to limit the API to limit the possibility of bad values needing some sort of error handling

The flags-based API I have proposed does not imply any unnecessary error handling AFAICT.

> How about an enum so at least there would be warnings if misused?

I would not recommend that for two reasons: 

1. We may end up with code that converts Squid configuration flags into internal enums _and_ code that converts those internal enums to OpenSSL configuration flags. Hopefully, we do not need such back-and-forth conversions!
1. Enum itself is a poor medium for representing nested/dependent flags (as illustrated by OpenSSL enum-like API effectively causing a few other PR problems marked in my review).

trapexit: I mean the literal SSL_VERIFY mode flags. I was asked to abstract the call to SSL_CTX_set_verify.  I'm confused as to what that means now. There are only 3 legitimate values... SSL_VERIFY_NONE, SSL_VERIFY_PEER, and SSL_VERIFY_PEER|SSL_VERIFY_PEER_IF_NO_PEER_CERT. The boolean API style didn't protect against just SSL_VERIFY_PEER_IF_NO_PEER_CERT because it is, while a bug, apparently a legit client mode value as mentioned in the man page.

The SSL_VERIFY_* flags are a mix of actual flags and enums in practice. NONE = 0.

I see only 4 ways to pass in the necessary information. 1 function with 3 different argument styles (booleans, enums, or bit flags) and using 3 different functions. Bitflags are what existed prior and I was asked to remove the direct SSL_CTX_set_verify calls. Booleans make it harder to tell what's going on. Separate functions is what I had the first time and were rejected. And enums are also said to be no good.

Enums or 3 different functions are the safest means to limit the selection. If the suggestion is simply to pass around SSL_VERIFY mode flags raw I'm not understanding the request for the abstraction since there really isn't any. At most there would be the ability for uniform error handling which would be a different kind of change and I'd rather not mix restructuring with improvements. If we create new bit flags to provide an abstraction then we'll need more conversion code than enums would.

Please just tell me precisely the function signature you want.

trapexit: Additionally, `parsedFlags` isn't available in support.cc. So I'd need a different interface or recreate the necessary flags to call the function.

trapexit: How should I handle the `clientCaStack` conditional affecting the verification mode? I either have to have the logging outside the SetupPeerVerfication call which would mean duplicating the conditional logical; adding the clientCaStack to the SetupPeerVerification signature; make logging optional and treat that as a special case where I log prior and fake the flags?

rousskov: > Passing the config flags through to underlying API seems ugly to me, but if that's what you want. OK.

Yes, that is the best _overall_ option for now AFAICT. If you prefer, you can think of it as the lesser evil: Each option has advantages and disadvantages, but the problems with all the other (posted or discussed) options are worse than problems with this option AFAICT.

> especially given it references a bunch of other SSL behaviors unrelated to peer verification

This is not a valid concern from the code architecture point of view -- the flags are currently an integer but they are essentially a configuration object and should be viewed as such for the purpose of this discussion. Each piece of code that gets a configuration object (or a relevant portion of that object) can take what it needs from it. In fact, it is easy to imagine a situation where a function like SetupPeerVerification() needs more than just peer verification flags to work correctly.

When calling SetupPeerVerification(), we could mask parsedFlags to leave just the flags currently used by SetupPeerVerification(), but I see no value in that extra effort/complication.  We could even convert the flags into an object of some new dedicated configuration class, but, again, I see no value in that extra effort/complication.

Another related important architectural aspect here is that, all other factors being equal, it is best to interpret peer validation configuration in one place. The current code splits/spreads that interpretation across two different functions. So far, we have discovered problems with that split (anonymous boolean parameters are evil and their derivation may obscure lack of error checking), but we have not discovered a valid reason to split processing in the first place.

rousskov:
Please note that you do not need to merge `master` into your PR branch just to avoid GitHub warnings that the PR branch is out of sync. If you need the merge for _other_ reasons (e.g., to resolve conflicts, for deployment reasons, etc.), then you can do that, of course. Otherwise, let [Anubis](https://github.com/measurement-factory/anubis/) do the (final) merging for you and minimize noise/distractions during review.

src/ssl/support.h
rousskov: > How should I handle the `clientCaStack` conditional affecting the verification mode? 

Here are a few sketches. Please pick the one you dislike the least and adjust as needed.

### Lazy/minimum:

```C++
    if (clientCaStack) {
        ...
        SetupPeerVerification(ctx.get(), parsedFlags);

        updateContextCrl(ctx);
        updateContextTrust(ctx);
    } else {
        // TODO: Reject configuration flags that require a client CA
        SetupPeerVerification(ctx.get(), SSL_FLAG_NONE);
    }
```

For the second SetupPeerVerification() above, the SSL_FLAG_NONE (with a value of zero) will need to be added to the list of the already `#define`d configuration flags.

### Detects misconfigurations:

```C++
    if (clientCaStack) {
        ...
        ConfigurePeerVerification(*ctx, parsedFlags);

        updateContextCrl(ctx);
        updateContextTrust(ctx);
    } else {
        DisablePeerVerification(*ctx, parsedFlags);
    }
```

### Compact misconfiguration detection:

```C++
    if (clientCaStack) {
        ...
        updateContextCrl(ctx);
        updateContextTrust(ctx);
    }
    ConfigurePeerVerification(ctx.get(), clientCaStack, parsedFlags);
```

In this last sketch, ConfigurePeerVerification()'s second parameter is a pointer to the optional CA. The function just uses it to know whether peer validation should be allowed or denied. It can even be a `void *` pointer! I would not make it a boolean for the reasons similar to "two anonymous booleans are evil" , but it can be a boolean if really needed.

trapexit: > Yes, that is the best overall option for now AFAICT.

Enums seems the clearest and simplest to me but so be it.

> they are essentially a configuration object and should be viewed as such for the purpose of this discussion.

I understand that. I'm debating the use of app config for directly setting lower level API state. Just as I wouldn't pass ACL config into the peer verification calls I wouldn't want to pass any other non-peer verification config. It makes using it in other places more difficult / clumsy. See below.

How should I resolve the problem with clientCaStack in ::updateContextClientCa? 

If I move to using the config flags for SetupPeerVerification and move the logging as well there isn't a clean way to manage when clientCaStack = nullptr. 

1) Add clientCaStack to the SetupPeerVerification signature and do the check twice?
2) Fake flags to get SSL_VERIFY_NONE set and double log? or move the logging outside SetupPeerVerification?
3) Add a new SSL_FLAG to indicate clientCaStack = nullptr?

None of these are pretty.

trapexit: I don't see how the 'lazy' works. The behavior is that if clientCaStack is not null then verify peer is enabled unless DONT_VERIFY_PEER is set. Having SSL_FLAG_NONE be 0 is indistinguishable from the current default. It'd require swapping the DONT to DO or making NONE nonzero.

I'll just do the "detects" one.

rousskov: > [Making the 'lazy' version work would] require ... making NONE nonzero.

Agreed. SSL_FLAG_NONE must not be zero for the lazy sketch to work. Fortunately, as a new flag, it can be anything (and can be renamed to something like `SSL_FLAG_NO_CA_CONFIGURED` to avoid the customary association of NONE with zero). Sorry I missed that detail when sketching the solutions.

The introduction of a special non-zero flag makes the "lazy" option even worse than it first looked.

> I'll just do the "detects" one.

Thank you. The misconfiguration detecting sketches require more development work, but they also improve Squid ability to detect bogus configurations and configuration typos.

src/ssl/support.cc
```diff
@@ -390,9 +390,35 @@ ssl_verify_cb(int ok, X509_STORE_CTX * ctx)
...
 {
-    SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, ssl_verify_cb);
+    int mode;
+
+    if(flags & SSL_FLAG_DONT_VERIFY_PEER) {
+        debugs(83, 2, "SECURITY WARNING: Peer certificates are not verified for validity!");
+        mode = SSL_VERIFY_NONE;
+    }
+    else if(flags & SSL_FLAG_DELAYED_AUTH) {
+        debugs(83, 9, "Not requesting client certificates until acl processing requires one");
```
yadij: Please use a level 3-6 for messages that may be of use to admin, but not important enough to display by default. As Alex already mentioned 9 is reserved for huge data dumps.

You can also use DBG_PARSE_NOTE() to inform admin doing a config test what is going to happen during production.
```suggestion
        debugs(83, DBG_PARSE_NOTE(3), "not requesting client certificates until ACL processing requires one");
```

src/ssl/support.cc
```diff
@@ -390,9 +390,35 @@ ssl_verify_cb(int ok, X509_STORE_CTX * ctx)
...
 
 void
-Ssl::SetupVerifyCallback(Security::ContextPointer &ctx)
+Ssl::ConfigurePeerVerification(Security::ContextPointer &ctx, int flags)
 {
-    SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_PEER | SSL_VERIFY_FAIL_IF_NO_PEER_CERT, ssl_verify_cb);
+    int mode;
+
+    if(flags & SSL_FLAG_DONT_VERIFY_PEER) {
+        debugs(83, 2, "SECURITY WARNING: Peer certificates are not verified for validity!");
```
yadij: Security alerts should be at a high level (eg DBG_IMPORTANT for ones that are only warning) so they get seen.

I am keen for the DONT_VERIFY_PEER flag to be erased. It is already removed from the documentation about these listening ports. So now this code is changing please add a debug statement about the deprecation.
```suggestion
        debugs(83, DBG_IMPORTANT, "SECURITY WARNING: Peer certificates are not verified for validity!");
        debugs(83, DBG_IMPORTANT, "UPGRADE NOTICE: The DONT_VERIFY_PEER flag is deprecated. Remove the clientca= option to disable client certificates.");
```

src/ssl/support.cc
trapexit: I simply copied what was already there.

src/ssl/support.cc
trapexit: Can we please separate improvements & refactoring from the feature enhancement? This would be a whole lot simpler if the debug level and message fixes were done separately.

yadij:
Thank you for addressing the changes I pointed at directly. There are still some debugs levels in the ConfigurePeerVerification() and DisablePeerVerification() which need to be made to match the one I highlighted.

src/ssl/support.h
trapexit: Shouldn't bad config be picked up in the config parser? Not down in the usage of said config? By this point in the code there should be no misconfiguration.

trapexit:
Why can't this cleanup be a separate patch. It merely complicates this process. Also, I'm not a Squid developer. I don't know what the best debug level policies are. I'm just using what's already there to keep things clean. In the time it's taking to go back and forth about things unrelated to the feature someone in the know could have submitted the debugs cleanup patch a 100 times over.

src/ssl/support.h
rousskov: > Shouldn't bad config be picked up in the config parser?

In the ideal world: the answer would be "no": The parser should complain about syntax errors, and the errors we are dealing here are about semantics. Moreover, these errors are partially dependent on the outcomes of such complex operations as loading a certificate chain from a disk file. The latter dependencies/complexities explain why these checks are not done in the parsing code today, even though a lot of semantics checks are misplaced there.

> By this point in the code there should be no misconfiguration.

Yes, you sensed the right problem, but we are not going to solve it in this PR AFAICT. Squid configuration has several huge architectural flaws that we are slowly addressing, including the lack of clear separation between configuration parsing/validation and configuration application. When we are done, the placement of the code you are working on now will probably feel more natural, but even if it is misplaced, it would be easy to move.

You are already going the extra mile to improve the quality of this code you touch. Loading you with the herculean task of separating SSL configuration loading from SSL configuration application would be rather unfair IMO, especially since doing so will probably require changes going well beyond the code affected by the feature you are interested in.

trapexit: I should have said "config system" rather than "parser." I meant just that the config code should be doing all syntax and data validation. Hence my opposition to passing the raw config bit flags into the runtime API. If the interest is in separating these further I'm not following why furthering it is desired.

> but we are not going to solve it in this PR AFAICT. 

Nor is that what I'm trying to accomplish. In fact I already feel this PR has turned into more than it should. Fixing debug messages and whatnot should, IMO, be a separate PR and had I been asked to look into doing so out of band I'd have been happy to do so but doing it as a prerequisite for accepting a feature patch frustrates things.

rousskov:
> Why can't this cleanup be a separate patch.

It can be. Historically, the vast majority of folks contributing Squid features "disappear" after the feature is committed -- they are not coming back to polish things up. Since there is no army of Squid developers cleaning up the mess, our strong preference (some may even say "policy") is to require high quality submissions even though it sometimes hurts the clarity of the change itself.

Naturally, there is a limit to what has to be polished. Unfortunately, it is impossible to define the boundaries precisely. Taking your particular case as an example, I think that fixing the new API was a fair request (and I would be willing to help with the fixes if needed) while asking you to come up with better replacements of old/moved debugging messages was pushing it a bit too far:

> Also, I'm not a Squid developer. I don't know what the best debug level policies are. 

Exactly! Personally, I think that such polishing should be done by experienced Squid developers, either by providing specific/exact replacements as change request suggestions or by committing those fixes directly to the PR branch.

The extent to which you cooperate with reviewer request is your call, of course. You have to balance the extra time/work you spend addressing vague requests with the risk that the PR will be blocked until somebody else volunteers to do those polishing touches (assuming the reviewer does not bog down and does not implement the changes they want themselves).

This friction is unfortunate, but we have not found a better way to move forward with the historical baggage and the human resources we have.

trapexit:
> This friction is unfortunate, but we have not found a better way to move forward with the historical baggage and the human resources we have.

I don't agree this friction is necessary. It's generally bad practice to combine feature changes with refactoring. It muddies the purpose of the commit and makes reviewing and cherry picking harder. We aren't talking about a half baked feature. This patch is extremely small and concise. The API discussion is relevant but everything else about this PR is extraneous. The easy way to not have friction would be to debate the feature in context of the existing code and style (or any yet committed restructuring) and then afterwards bring up the other stuff and ask if the committer would be willing to help out cleaning given my new familiarity with that area of the code. Rather than complicating both the patch and dialog with a new developer to the project with conflating a number of concerns. Some of which are systemic to the platform and couldn't be addressed here... and some which could have been addressed by the reviewers in far less time in a separate PR.

rousskov:
> > This friction is unfortunate, but we have not found a better way to move forward with the historical baggage and the human resources we have.
 
> I don't agree this friction is necessary. 

I did not say it is _necessary_ :-). I only documented its (unfortunate) existence. We have not found a way to remove that friction without sacrificing code quality. Most of that friction, IMHO, is due to a human factor, but some of its sources are fundamental (e.g., lack of resources and absence of a central authority).

> It's generally bad practice to combine feature changes with refactoring.

Yes, of course. We are often combining them not because it is good practice but because separating them results in an even worse outcome (on average) -- the feature is committed while the refactoring part never materializes.

> The easy way to not have friction would be to debate the feature in context of the existing code and style (or any yet committed restructuring) and then afterwards bring up the other stuff and ask if the committer would be willing to help out cleaning given my new familiarity with that area of the code.

Yes, when possible, polishing should be done as the last step. In the ideal world, that last step would be in a separate commit. In reality, a lot of PR bugs are not even visible (to mortals) without related polishing _and_ a lot of developers would not agree to polish _after_ their feature is committed. Typical responses include "my employer does not want to pay for polishing and thinks the project should be thankful we shared our changes in the first place".

Please note that I am not trying to justify _all_ polishing requests -- some of them are wrong, and many should be done by the requestor instead. I am just trying to explain that, given the specifics of this project, you may see more polishing requests than in a typical/average open source project.

trapexit:
Is having a client CA and not wanting verification not a valid config? Turning off all verification because no CA is provided seems maybe generally what some would want but it seems to mimic the SSL API and behavior more closely to keep/add flags for verification (or not).

Regarding implementation. If your end goal is cleaning things up then why did I pass raw config parse flags into the SSL runtime wrapper? My enum API was a clean separation from the config space and the constrained runtime space. Seems to me that any validations should be done in Security::ServerOptions. Not in Ssl::ConfigurePeerVerification().

I'm sorry but I'm getting impatient with all this. I'm not going to disappear into the night. I just want to get the feature upstreamed. This has turned into far more work than I really wanted to invest right now. Can we please agree to some simple feature patch and I pinky swear I'll take a crack at some cleanup when I have the time. If not I'm going to just let this sit a while because I really need to focus on other things.

rousskov:
> Is having a client CA and not wanting verification not a valid config?

I think that config would be OK if a client CA is required to solicit a client certificate (whether that certificate is validated is a separate issue). I am more concerned about the opposite combination -- no client CA and wanting (some) verification. That combination would indicate a clear misconfiguration and would lead to improper Squid operation (that may be difficult to notice), right? Similarly, requesting no verification at all and some verification for the same https_port would not make sense and may silently misfire. Can the new code throw if it discovers a self-contradicting combination? Or is there some fundamental reason why Squid should ignore such misconfigurations?

FWIW, detecting misconfigurations (or rejecting that idea for some valid reason) is my last significant concern. I can polish the remaining bits (wrong flags type and formatting) myself if needed.

> If your end goal is cleaning things up then why did I pass raw config parse flags into the SSL runtime wrapper? My enum API was a clean separation from the config space and the constrained runtime space

I can explain, but I suggest not to revisit this topic to save time. I know that you do not have much time left for this PR, and the current flags passing implementation is acceptable to both of us AFAICT.

trapexit:
> no client CA and wanting (some) verification. That combination would indicate a clear misconfiguration and would lead to improper Squid operation (that may be difficult to notice), right

Agreed. But prior it was suggested and requested that I log that DONT_VERIFY_PEER would be going away in favor of not setting clientca=. If the flag is just going to be removed anyway it doesn't make much sense to put a check against it now just to remove it later. Especially given there isn't a check now.

> I can explain, but I suggest not to revisit this topic to save time.

Well if you want me to clean things up later... I'd rather not mess them up more now which feels like what I've been doing.

rousskov:
> > no client CA and wanting (some) verification. That combination would indicate a clear misconfiguration and would lead to improper Squid operation (that may be difficult to notice), right
 
> Agreed. But prior it was suggested and requested that I log that DONT_VERIFY_PEER would be going away in favor of not setting clientca=. If the flag is just going to be removed anyway it doesn't make much sense to put a check against it now just to remove it later. Especially given there isn't a check now.

Agreed -- doing this just for SSL_FLAG_DONT_VERIFY_PEER would be wasteful. However, DONT_VERIFY_PEER is not the only flag that can conflict with the lack of client CA AFAICT. Here are all the potential candidates: SSL_FLAG_DELAYED_AUTH, SSL_FLAG_DONT_VERIFY_PEER, SSL_FLAG_VERIFY_CRL, SSL_FLAG_VERIFY_CRL_ALL, and the new SSL_FLAG_CONDITIONAL_AUTH. If any of them is given without a client CA, then Squid should complain and throw, especially if the new SSL_FLAG_CONDITIONAL_AUTH is used.

Also, the combination of SSL_FLAG_DELAYED_AUTH and SSL_FLAG_CONDITIONAL_AUTH is a misconfiguration, right?

The bottom line:

* The new ConfigurePeerVerification() code establishes an undocumented and silently-applied hierarchy of flags via the if-else-if-else... construct. Squid should complain/throw instead.
* The new DisablePeerVerification() code silently ignores admin-set peer validation options. Squid should complain/throw instead.

You can state that it is not *PeerVerification() job to validate the configuration. I would disagree on two levels: First, the complex relationship between flags and client CA (at least) make it difficult to validate the configuration at PeerOptions::parseFlags() time. And second, you have not added that validation to PeerOptions::parseFlags() either.

The only semi-valid argument here, AFAICT, is "the old code did not validate, so I am not going to validate either". The problem with that argument is that adding validation _after_ a new flag is deployed is much more difficult because validation may break "working" configs.

Adding a bit of validation is a fair price for new feature acceptance IMHO. If you disagree, then I will _not_ insist on it in this "there are already lots of similar unvalidated flags" case if others approve the PR without that validation.

yadij:
Alex: FYI I plan to add validation for all these flags with the GnuTLS support patches renaming the squid.conf options. So it is not a major issue to leave these all consistently "silently ignored" for now.
At most it would be a warning anyway, since there is no danger from having a useless setting configured.

trapexit:
> DONT_VERIFY_PEER is not the only flag that can conflict with the lack of client CA AFAICT

But you are talking about adding new functionality into a segment of the code that doesn't exist today. I agree completely that Squid should have better config validation. It's terrible today. I also completely disagree that I, now, in this PR, should be putting those checks in.

> You can state that it is not *PeerVerification() job to validate the configuration. I would disagree on two levels: First, the complex relationship between flags and client CA (at least) make it difficult to validate the configuration at PeerOptions::parseFlags() time. And second, you have not added that validation to PeerOptions::parseFlags() either.

Difficult how? There is nothing complex about the relationship. If client ca is not set the validation stuff is not valid or at least pointless. And if I can articulate the logic downstream surely I can do so upstream. We aren't talking about runtime issues (the client ca info itself should be valid for things to get that far). We are talking about things that are config issues. Static data validation is fine. Any runtime problems are a separate topic. There is a clear contract for the SSL_CTX_set_verify flags (including some documented bugs). There is a fixed number of permutations of those flags Squid as a whole would allow. Checking that in the parser is trivial. There is already some checks in the parser. Why would you move that to multiple locations? The only reason it's possible to have the flag values wrong in both places right now is because the flags are being passed raw, as an int, into an API that should be limiting it's input.

> The problem with that argument is that adding validation after a new flag is deployed is much more difficult because validation may break "working" configs.

That's already a problem. There is discussion of removing a flag all together. You going to ignore it? That's a far bigger issue IMO. If the config is going to suddenly be breaking *anything* that wasn't a bug it should be a major release bump.

I'm not opposed to config verification. I'm opposed to *introducing new validation* as was suggested. I'm not interested in the waxing of Squid's issues on the whole. I am interested in solving my problem of Squid limiting my client cert verification strategies. If you would like me to take a crack at other verification in the same space as what I'm currently involved in... fine... but I'm not interested in doing it right now in this PR. If you're going to treat every new contributor as some fly by night bomb thrower then you're going to get fewer repeat contributors.

> Adding a bit of validation is a fair price for new feature acceptance IMHO.

As I said before: The problem isn't asking me to do something unrelated to my PR. It's the way it has been gone about I have a problem with.

trapexit:
If you want to establish general rules and expectations for contributors then may I recommend a "CONTRIBUTING" file in the root source. To be honest I probably wouldn't have noticed it prior to this PR but at least you'll have something in writing so it doesn't need to be debated in a thread like this.

https://help.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors

rousskov:
> Alex: FYI I plan to add validation for all these flags with the GnuTLS support patches renaming the squid.conf options. So it is not a major issue to leave these all consistently "silently ignored" for now.

If history is to repeat itself, then when you actually do that PR, we will not kill Squid on dangerous misconfiguations because doing so would break existing deployments. That is why, as I have already said, the best time to add proper configuration checks is when a new configuration option is added.

> At most it would be a warning anyway, since there is no danger from having a useless setting configured.

I already gave [examples](https://github.com/squid-cache/squid/pull/510#issuecomment-561784468) of dangerous misconfigurations so I am not going to repeat them, but please reread them when you start adding checks. The primary problem here is not about useless settings.

rousskov:
> If you're going to treat every new contributor as some fly by night bomb thrower then you're going to get fewer repeat contributors.

> It's the way it has been gone about I have a problem with.

@trapexit, I am sorry that my documentation of the background issues (or something else in my messages) have antagonized you. I was only trying to explain _why_ you are seeing what you are seeing. You are not responsible for the ugly problems we have to deal with. I do appreciate your contribution. I hope me spending significant time on improving your PR is a sufficient proof of the sincerity of that statement.

> at least you'll have something in writing so it doesn't need to be debated in a thread like this

I agree that this thread is the wrong place to debate policies. It would be great to carefully document and justify the primary expectations (and then automatically refer new contributors to that documentation). Unfortunately, that is currently impossible (beyond the already [documented](https://wiki.squid-cache.org/MergeProcedure) basics) because there is no meaningful consensus regarding how Squid PRs should be handled and no authority to dictate a single course. This is one of the ugly problems we have to deal with.

I am no longer asking you to add validation to this PR. If, despite all the mistreatment, you decide to come back to work on validation, I would be happy to discuss the best approach on the squid-dev@ mailing list or elsewhere.

rousskov:
Going forward, my next step is to polish the PR code so that it can be merged. Please give me a few days to do that.

rousskov:
> my next step is to polish the PR code so that it can be merged

Done. @trapexit, please review my changes and adjust if needed. I tried to document the reasons behind them in individual commit messages. I was not quite sure about duplicated code removed in the second part of https://github.com/squid-cache/squid/pull/510/commits/c872737024f8fc7c8884ad34ab312ac09392beea. Unfortunately, it looks like you squashed your intermediate changes, and I could not find the reason behind that duplication.

Once you are happy with the code, please let us know so that we can mark the PR as ready for merging.

P.S. You do not need to squash the changes.

rousskov:
> @trapexit, please review my changes and adjust if needed. I tried to document the reasons behind them in individual commit messages. I was not quite sure about duplicated code removed in the second part of [c872737](https://github.com/squid-cache/squid/commit/c872737024f8fc7c8884ad34ab312ac09392beea). Unfortunately, it looks like you squashed your intermediate changes, and I could not find the reason behind that duplication.
> 
> Once you are happy with the code, please let us know so that we can mark the PR as ready for merging.
> 
> P.S. You do not need to squash the changes.

@trapexit, just a kind reminder that the ball is on your side for this PR. Please see my December 2019 [message](https://github.com/squid-cache/squid/pull/510#issuecomment-565644071) (quoted above) for details.

rousskov:
> @trapexit, just a kind reminder that the ball is on your side for this PR. Please see my December 2019 [message](https://github.com/squid-cache/squid/pull/510#issuecomment-565644071) (quoted above) for details.

@yadij, it looks like @trapexit is not interested in reviewing my changes. I do not even remember what they were, but the PR has been stuck in this state for a while. If you think the current PR code should go in, please feel free to clear it for merging.

PR #511, opened 2019-11-18
Fix detection of sys/sysctl.h

yadij:
Make sure we test the EUI specific headers using same flags
chosen for final build operations. This should make the
test detect the header as unavailable if the user options
would make the compiler #warning be a fatal error later.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1109,6 +1109,10 @@ AC_ARG_ENABLE(eui,
 SQUID_YESNO([$enableval],[--disable-eui expects no arguments])
 ])
 if test "x${enable_eui:=yes}" = "xyes" ; then
+  SQUID_STATE_SAVE(LIBEUI)
+  # GLIBC 2.30 deprecates sysctl.h. Test with the same flags that (may) break includes later.
+  CFLAGS=$SQUID_CFLAGS
+  CXXFLAGS=$SQUID_CXXFLAGS
   case "$squid_host_os" in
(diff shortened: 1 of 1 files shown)


kinkie:
clang is failing on modern Debian-based distros on some unique_ptr abuse in src/security/ServerOptions.h:38

I am OK with this PR, will approve as soon as build succeeds.

yadij:
The clang issue is being fixed in a separate PR, so the build will never succeed automatically. We are going to have to merge one to break the deadlock - ideally this simpler one.

kinkie:
I'm gladly approving it, but can't do it on the github interface due to the build failure.

rousskov:
> I'm gladly approving it, but can't do it on the github interface due to the build failure.

Build failures have nothing to do with PR approvals: The two areas (i.e. automated tests and manual reviews) are independent. One can approve a PR that failed all build tests.

> yadij merged commit 195ec84 into squid-cache:master ... 1 of 3 checks passed

@yadij, please do not abuse your access rights to manually commit master changes when an automated commit is possible, and there is no emergency. In situations like this one, we still can (and, hence, should) use Anubis to merge the PR. When the PR is otherwise ready to go in, we just need to temporary change configuration to ignore Jenkins failures.

Another option here would be to revert whatever Jenkinks changes resulted in test breakage. I do not know how easy that would be in this specific case, but, in general, a temporary Jenkins fix/reversal is the preferred way forward.

PR #513, opened 2019-11-19
Fix Security::ServerOptions copy-constructor

yadij:
clientCaStack uses a std::unique_ptr which cannot be copied.

### DIFF
--- a/src/security/ServerOptions.h
+++ b/src/security/ServerOptions.h
@@ -35,6 +35,7 @@ public:
         // is more secure to have only a small set of trusted CA.
         flags.tlsDefaultCa.defaultTo(false);
     }
+    ServerOptions(const ServerOptions &o): ServerOptions() { *this = o; }
     ServerOptions &operator =(const ServerOptions &);
     ServerOptions(ServerOptions &&o) { this->operator =(o); }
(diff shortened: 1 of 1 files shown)


src/security/ServerOptions.h
```diff
@@ -35,7 +35,6 @@ class ServerOptions : public PeerOptions
         // is more secure to have only a small set of trusted CA.
         flags.tlsDefaultCa.defaultTo(false);
     }
-    ServerOptions(const ServerOptions &) = default;
```
rousskov: If copying is needed, you may be able to trivially implement the copy constructor via the existing copy assignment operator:

```C++
ServerOptions(const ServerOptions &other): ServerOptions() { *this = other; }
```

rousskov:
> clientCaStack uses a std::unique_ptr which cannot be copied implicitly.

If the above sentence stays in the adjusted PR description, please remove the word "implicitly" because std::unique_ptr cannot be copied at all (neither explicitly not implicitly).

src/security/ServerOptions.h
yadij: hmm, github wont let me apply this change suggestion directly. But done.

PR #514, opened 2019-11-19
Bug 5009: Build failure with older clang libc++

rousskov:
Older clang libc++ implementations correctly reject implicit usage of an
explicit (in C++11) std::map copy constructor with "chosen constructor
is explicit in copy-initialization" errors. The same code becomes legal
in C++14[1], so newer libc++ implementation allow implicit usage (even
in C++11), but there is no need for copy-initialization here at all.
Evidently, libstdc++ has never declared constructors explicit.

The bug was seen with Apple clang in Xcode 5.1.1 (roughly upstream clang
3.4) and Xcode 6.2 (roughly upstream clang 3.5), both using libc++.

[1] http://cplusplus.github.io/LWG/lwg-defects.html#2193

### DIFF
--- a/src/acl/Options.cc
+++ b/src/acl/Options.cc
@@ -246,14 +246,14 @@ Acl::ParseFlags(const Options &options, const ParameterFlags &flags)
 const Acl::Options &
 Acl::NoOptions()
 {
-    static const Options none = {};
+    static const Options none;
     return none;
 }
 
 const Acl::ParameterFlags &
 Acl::NoFlags()
 {
-    static const ParameterFlags none = {};
+    static const ParameterFlags none;
     return none;
 }
 


jmroot:
I would say "older" rather than "newer" in the commit message. Newer libc++ versions have the C++14 change and so aren't affected. Otherwise LGTM, thanks!

rousskov:
> I would say "older" rather than "newer" in the commit message. Newer libc++ versions have the C++14 change and so aren't affected. Otherwise LGTM, thanks!

Aha, that makes the situation even simpler and explains why I could not reproduce the problem with a newer clang installation. Thank you! I have adjusted the PR title/description accordingly and am clearing this PR for the official merge.

jmroot:
> I have adjusted the PR title/description accordingly

The PR is updated but not the actual commit message.

rousskov:
> The PR is updated but not the actual commit message.

Thank you for staying vigilant! There is no actual commit message yet. [Anubis](https://github.com/measurement-factory/anubis/) will automatically create it (on an auto branch) when it starts merging this PR into master. Passing all tests (against PR branch) is a precondition for that creation. Somebody ~broke~updated Jenkins tests, and when you wrote your comment, they were failing. These problems may have been resolved already or will be resolved soon. Just give it a bit more time.

jmroot:
OK, thanks for clarifying the process.

PR #516, opened 2019-11-27
Do not stall if xactions overwrite a recently active cache entry

eduard-bagdasaryan:
After the last transaction that cached or read the reply R1 ended, its
Transients entry T1 was not freed. Subsequent requests (with different
public keys) could occupy the same shared memory and rock slots (purging
unlocked R1), preventing Squid from attaching a T1-derived StoreEntry to
the cache. Another request for R1 would receive T1 and stall because its
worker W1 kept waiting for a notification from another worker W2,
incorrectly assuming that W2 exists and is going to fetch R1 for W1.
That request was aborted after a timeout.

A Transients entry represents active transaction(s). Broadcasts stop
when there are no transactions to inform. We must remove idle (i.e.
unlocked) Transients entries to avoid feeding new transactions with
stale info. We now do that when unlocking a Transients entry and also
double check that a found unattached Transients entry has a writer.

### DIFF
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -251,6 +251,13 @@ Transients::addEntry(StoreEntry *e, const cache_key *key, const Store::IoStatus 
     }
 }
 
+bool
+Transients::hasWriter(const StoreEntry &e) {
+    if (!e.hasTransients() || !e.mem_obj)
+        return false;
+    return map->peekAtWriter(e.mem_obj->xitTable.index);
+}
+
 void
 Transients::noteFreeMapSlice(const Ipc::StoreMapSliceId)
 {
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -253,7 +253,7 @@ Transients::addEntry(StoreEntry *e, const cache_key *key, const Store::IoStatus 
 
 bool
 Transients::hasWriter(const StoreEntry &e) {
-    if (!e.hasTransients() || !e.mem_obj)
+    if (!e.hasTransients())
         return false;
     return map->peekAtWriter(e.mem_obj->xitTable.index);
 }
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -157,7 +157,7 @@ Transients::get(const cache_key *key)
     if (StoreEntry *oldE = locals->at(index)) {
         debugs(20, 3, "not joining private " << *oldE);
         assert(EBIT_TEST(oldE->flags, KEY_PRIVATE));
-        map->closeForReading(index);
+        map->closeForReadingAndFreeIdle(index);
         return nullptr;
     }
 
@@ -283,11 +283,6 @@ Transients::completeWriting(const StoreEntry &e)
 {
     assert(e.hasTransients());
     assert(isWriter(e));
-    // The not freed Transients entry may eventually become 'unattached',
-    // so that it cannot be anchored anymore. Though we handle this situation,
-    // there are some risks the entry metadata becomes stale and, hence, we
-    // should drop such entries. TODO: should we drop this entry immediately to
-    // avoid such risks?
     map->switchWritingToReading(e.mem_obj->xitTable.index);
     e.mem_obj->xitTable.io = Store::ioReading;
 }
@@ -338,7 +333,7 @@ Transients::disconnect(StoreEntry &entry)
             map->abortWriting(xitTable.index);
         } else {
             assert(isReader(entry));
-            map->closeForReading(xitTable.index);
+            map->closeForReadingAndFreeIdle(xitTable.index);
         }
         locals->at(xitTable.index) = nullptr;
         xitTable.index = -1;
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -252,7 +252,8 @@ Transients::addEntry(StoreEntry *e, const cache_key *key, const Store::IoStatus 
 }
 
 bool
-Transients::hasWriter(const StoreEntry &e) {
+Transients::hasWriter(const StoreEntry &e)
+{
     if (!e.hasTransients())
         return false;
     return map->peekAtWriter(e.mem_obj->xitTable.index);
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -283,6 +283,11 @@ Transients::completeWriting(const StoreEntry &e)
 {
     assert(e.hasTransients());
     assert(isWriter(e));
+    // The not freed Transients entry may eventually become 'unattached',
+    // so that it cannot be anchored anymore. Though we handle this situation,
+    // there are some risks the entry metadata becomes stale and, hence, we
+    // should drop such entries. TODO: should we drop this entry immediately to
+    // avoid such risks?
     map->switchWritingToReading(e.mem_obj->xitTable.index);
     e.mem_obj->xitTable.io = Store::ioReading;
 }
--- a/src/Transients.h
+++ b/src/Transients.h
@@ -85,6 +85,9 @@ public:
     bool isReader(const StoreEntry &) const;
     /// whether the entry is in "writing to Transients" I/O state
     bool isWriter(const StoreEntry &) const;
+    /// whether there is a writer into the Transients entry, this
+    /// StoreEntry is attached to
+    bool hasWriter(const StoreEntry &);
 
     static int64_t EntryLimit();
 
--- a/src/Transients.h
+++ b/src/Transients.h
@@ -85,8 +85,7 @@ public:
     bool isReader(const StoreEntry &) const;
     /// whether the entry is in "writing to Transients" I/O state
     bool isWriter(const StoreEntry &) const;
-    /// whether there is a writer into the Transients entry, this
-    /// StoreEntry is attached to
+    /// whether we or somebody else are in the "writing to Transients" I/O state
     bool hasWriter(const StoreEntry &);
 
     static int64_t EntryLimit();
--- a/src/Transients.h
+++ b/src/Transients.h
@@ -85,7 +85,7 @@ public:
     bool isReader(const StoreEntry &) const;
     /// whether the entry is in "writing to Transients" I/O state
     bool isWriter(const StoreEntry &) const;
-    /// whether we or somebody else are in the "writing to Transients" I/O state
+    /// whether we or somebody else is in the "writing to Transients" I/O state
     bool hasWriter(const StoreEntry &);
 
     static int64_t EntryLimit();
--- a/src/ipc/StoreMap.cc
+++ b/src/ipc/StoreMap.cc
@@ -253,6 +253,18 @@ Ipc::StoreMap::peekAtReader(const sfileno fileno) const
     return NULL;
 }
 
+const Ipc::StoreMap::Anchor *
+Ipc::StoreMap::peekAtWriter(const sfileno fileno) const
+{
+    const Anchor &s = anchorAt(fileno);
+    if (s.writing())
+        return &s; // immediate access by lock holder so no locking
+    if (s.reading())
+        return nullptr;
+    assert(false); // must be locked for reading or writing
+    return nullptr;
+}
+
 const Ipc::StoreMap::Anchor &
 Ipc::StoreMap::peekAtEntry(const sfileno fileno) const
 {
--- a/src/ipc/StoreMap.cc
+++ b/src/ipc/StoreMap.cc
@@ -247,10 +247,8 @@ Ipc::StoreMap::peekAtReader(const sfileno fileno) const
     const Anchor &s = anchorAt(fileno);
     if (s.reading())
         return &s; // immediate access by lock holder so no locking
-    if (s.writing())
-        return NULL; // the caller is not a read lock holder
-    assert(false); // must be locked for reading or writing
-    return NULL;
+    assert(s.writing()); // must be locked for reading or writing
+    return nullptr;
 }
 
 const Ipc::StoreMap::Anchor *
@@ -259,9 +257,7 @@ Ipc::StoreMap::peekAtWriter(const sfileno fileno) const
     const Anchor &s = anchorAt(fileno);
     if (s.writing())
         return &s; // immediate access by lock holder so no locking
-    if (s.reading())
-        return nullptr;
-    assert(false); // must be locked for reading or writing
+    assert(s.reading()); // must be locked for reading or writing
     return nullptr;
 }
 
--- a/src/ipc/StoreMap.cc
+++ b/src/ipc/StoreMap.cc
@@ -450,6 +450,23 @@ Ipc::StoreMap::closeForReading(const sfileno fileno)
     debugs(54, 5, "closed entry " << fileno << " for reading " << path);
 }
 
+void
+Ipc::StoreMap::closeForReadingAndFreeIdle(const sfileno fileno)
+{
+    Anchor &s = anchorAt(fileno);
+    assert(s.reading());
+
+    if (!s.lock.unlockSharedAndSwitchToExclusive()) {
+        debugs(54, 5, "closed entry " << fileno << " for reading " << path);
+        return;
+    }
+
+    assert(s.writing());
+    assert(!s.reading());
+    freeChain(fileno, s, false);
+    debugs(54, 5, "closed idle entry " << fileno << " for reading " << path);
+}
+
 bool
 Ipc::StoreMap::openForUpdating(Update &update, const sfileno fileNoHint)
 {
--- a/src/ipc/StoreMap.cc
+++ b/src/ipc/StoreMap.cc
@@ -453,7 +453,7 @@ Ipc::StoreMap::closeForReading(const sfileno fileno)
 void
 Ipc::StoreMap::closeForReadingAndFreeIdle(const sfileno fileno)
 {
-    Anchor &s = anchorAt(fileno);
+    auto &s = anchorAt(fileno);
     assert(s.reading());
 
     if (!s.lock.unlockSharedAndSwitchToExclusive()) {
--- a/src/ipc/StoreMap.h
+++ b/src/ipc/StoreMap.h
@@ -262,6 +262,9 @@ public:
     /// only works on locked entries; returns nil unless the slice is readable
     const Anchor *peekAtReader(const sfileno fileno) const;
 
+    /// only works on locked entries; returns nil unless the slice is writable
+    const Anchor *peekAtWriter(const sfileno fileno) const;
+
     /// only works on locked entries; returns the corresponding Anchor
     const Anchor &peekAtEntry(const sfileno fileno) const;
 
--- a/src/ipc/StoreMap.h
+++ b/src/ipc/StoreMap.h
@@ -291,6 +291,8 @@ public:
     const Anchor *openForReadingAt(const sfileno fileno);
     /// closes open entry after reading, decrements read level
     void closeForReading(const sfileno fileno);
+    /// same as closeForReading() but also frees the entry if it is unlocked
+    void closeForReadingAndFreeIdle(const sfileno fileno);
 
     /// writeable slice within an entry chain created by openForWriting()
     Slice &writeableSlice(const AnchorId anchorId, const SliceId sliceId);
--- a/src/ipc/StoreMap.h
+++ b/src/ipc/StoreMap.h
@@ -259,10 +259,12 @@ public:
     /// undoes partial update, unlocks, and cleans up
     void abortUpdating(Update &update);
 
-    /// only works on locked entries; returns nil unless the slice is readable
+    /// the caller must hold a lock on the entry
+    /// \returns nullptr unless the slice is readable
     const Anchor *peekAtReader(const sfileno fileno) const;
 
-    /// only works on locked entries; returns nil unless the slice is writable
+    /// the caller must hold a lock on the entry
+    /// \returns nullptr unless the slice is writeable
     const Anchor *peekAtWriter(const sfileno fileno) const;
 
     /// only works on locked entries; returns the corresponding Anchor
--- a/src/ipc/StoreMap.h
+++ b/src/ipc/StoreMap.h
@@ -267,7 +267,8 @@ public:
     /// \returns nullptr unless the slice is writeable
     const Anchor *peekAtWriter(const sfileno fileno) const;
 
-    /// only works on locked entries; returns the corresponding Anchor
+    /// the caller must hold a lock on the entry
+    /// \returns the corresponding Anchor
     const Anchor &peekAtEntry(const sfileno fileno) const;
 
     /// free the entry if possible or mark it as waiting to be freed if not
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -355,6 +355,8 @@ Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
         const bool found = anchorToCache(entry, inSync);
         if (found && !inSync)
             throw TexcHere("cannot sync");
+        else if (!found && !transients->hasWriter(entry)) // cannot anchor right now and will not be able in the future
+            throw TexcHere("unattached transients entry");
     }
 }
 
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -334,7 +334,7 @@ Store::Controller::find(const cache_key *key)
             return entry;
         } catch (const std::exception &ex) {
             debugs(20, 2, "failed with " << *entry << ": " << ex.what());
-            entry->release(true);
+            entry->release();
             // fall through
         }
     }
@@ -356,20 +356,16 @@ Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
         if (found && !inSync)
             throw TexcHere("cannot sync");
         if (!found) {
-            const auto hasWriter = transients->hasWriter(entry);
-            if (!hasWriter) {
-                // prevent others from falling into the same trap
-                stopSharing(entry);
-            }
-
             // !found should imply hittingRequiresCollapsing() regardless of writer presence
             if (!entry.hittingRequiresCollapsing()) {
                 debugs(20, DBG_IMPORTANT, "BUG: missing ENTRY_REQUIRES_COLLAPSING for " << entry);
                 throw TexcHere("transients entry missing ENTRY_REQUIRES_COLLAPSING");
             }
 
-            if (!hasWriter)
+            if (!transients->hasWriter(entry)) {
+                // prevent others from falling into the same trap
                 throw TexcHere("unattached transients entry missing writer");
+            }
         }
     }
 }
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -355,8 +355,12 @@ Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
         const bool found = anchorToCache(entry, inSync);
         if (found && !inSync)
             throw TexcHere("cannot sync");
-        else if (!found && !transients->hasWriter(entry)) // cannot anchor right now and will not be able in the future
+        else if (!found && !transients->hasWriter(entry)) {
+            // Cannot anchor right now and will not be able in the future.
+            // Prevent others from falling into the same trap.
+            stopSharing(entry);
             throw TexcHere("unattached transients entry");
+        }
     }
 }
 
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -356,18 +356,14 @@ Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
         if (found && !inSync)
             throw TexcHere("cannot sync");
         if (!found) {
-            if (transients->hasWriter(entry)) {
-                Transients::EntryStatus entryStatus;
-                transients->status(entry, entryStatus);
-                if (!entryStatus.collapsed) {
-                    debugs(20, DBG_IMPORTANT, "BUG: missing ENTRY_REQUIRES_COLLAPSING for " << entry);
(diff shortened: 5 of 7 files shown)


src/Transients.cc
```diff
@@ -251,6 +251,13 @@ Transients::addEntry(StoreEntry *e, const cache_key *key, const Store::IoStatus
     }
 }
 
+bool
+Transients::hasWriter(const StoreEntry &e) {
```
yadij:
```suggestion
Transients::hasWriter(const StoreEntry &e)
{
```

eduard-bagdasaryan: Done.

eduard-bagdasaryan:
> Is that lack of Transients entry deletion a bug or desirable behavior?

We have three options here:
* Immediately remove the unlocked Transients entry (i.e., when writer and readers are gone). This is how Squid behaved recently (probably v4 still uses this approach).
* Keep the unlocked Transients entry but allow to get it only if either it was 'anchored' to a cache or there is cache writer (which is going to create that cache). PR516 implements this approach.
* Same as (2) but remove the 'troubled' Transients entry just after it could not be obtained (due to 'anchoring' failure or lack of writer).

The second approach probably has an advantage over the first one since it reuses the Transients entry. But it has to pay for it by an extra check, happening each time we hit that 'troubled' entry. The third approach is a better alternative, IMO.

> Please confirm my understanding.

By 'unattached' Trasients entry I meant a Transients entry that cannot be 'anchored' anymore (because those cache slots were overwritten). Probably there is a better terminology for that (I used and rejected the 'orphaned' term previously).

> If the writer disappears before creating a cache entry ...

I think that this situation will be handled as if this entry is 'collapsed'. So it should receive notifications since (I hope) that collapsed workers receive notifications if the writer prematurely aborts.

rousskov:
> I think that this situation will be handled as if this entry is 'collapsed'. So it should receive notifications since (I hope) that collapsed workers receive notifications if the writer prematurely aborts.

Your answer raises these partially overlapping questions:

1. I assume that the presence of a Transients entry that cannot be attached right now but already has a writer implies a collapsed forwarding situation. When we create a local StoreEntry object based on such a Transients entry, do we set `ENTRY_REQUIRES_COLLAPSING` so that the requestor can confirm that they accept collapsing risks?
2. Should this PR prohibit returning a StoreEntry object for unattached Transient entries when Collapsed Forwarding is disabled? Right now, we say "if Transients entry has a writer, then we can return a StoreEntry because the requester will eventually get cached content" but that appears to imply collapsing, which could be disabled...

eduard-bagdasaryan:
> When we create a local StoreEntry object based on such a Transients entry, do we set ENTRY_REQUIRES_COLLAPSING so that the requestor can confirm that they accept collapsing risks?

This flag (and others as well) are copied into the local StoreEntry form the shared Transients entry (i.e., from anchor->basics.flags). The shared Transients flags, in turn, are copied from the local StoreEntry, created from scratch (i.e, on becoming writer). If collapsing was enabled, ENTRY_REQUIRES_COLLAPSING flag should be set in the local StoreEntry by that time. However, the Transients entry we are considering here is of another kind: it has been received some time ago (and since that time its cache was overwritten), and hence, it should not have ENTRY_REQUIRES_COLLAPSING by the time of the request.

> Should this PR prohibit returning a StoreEntry object for unattached Transient entries 

Taking into account the previous answer, I think that preserving such a Transients entry is useless because it would never become writer again. In other words, the 'IsWriter()' condition I added here will never be 'true'. So, this PR unwillingly 'prohibited' returning StoreEntry for such unattached Transient entries. However, we would need to polish this solution, using something like  (3) [option](https://github.com/squid-cache/squid/pull/516#issuecomment-559579283) from my previous sugestion or modify it by means of (1) option.

rousskov:
> Alex: Should this PR prohibit returning a StoreEntry object for unattached Transient entries when Collapsed Forwarding is disabled?

```C++
if (!found && !transients->hasWriter(entry)) // cannot anchor right now and will not be able in the future
    throw TexcHere("unattached transients entry");
```

> Eduard: the 'IsWriter()' condition I added here will never be 'true'. So, this PR unwillingly 'prohibited' returning StoreEntry for such unattached Transient entries.

I think you meant the hasWriter() condition. I think it will be true for collapsed transactions. All other transactions will throw and, hence, not get a StoreEntry object -- a good/desirable outcome for those other transactions.

What may be missing here is a confirmation that a `!found` but `hasWriter()` StoreEntry object has an `ENTRY_REQUIRES_COLLAPSING` flag set. AFAICT, a missing flag would indicate a Squid bug. If you agree, then we could probably salvage the situation (if that bug is detected) by throwing and returning nil instead of a StoreEntry object (after reporting the problem to the admin).

> we would need to polish this solution, using something like (3) [option](https://github.com/squid-cache/squid/pull/516#issuecomment-559579283) from my previous sugestion or modify it by means of (1) option.

AFAICT, we should add check(s) for unattached entries regardless of whether we delete idle Transient entries:

* If we delete idle Transient entries, then detecting (presumably aborted right after we created the StoreEntry object) writer-less unattachable entries earlier will reduce transaction overheads. Affected transactions would have received a notification later, but there is probably no good reason to have them wait for that notification if we can easily detect the aborted entry at find() time.
* If we do not delete idle Transient entries, then checking `!found && !hasWriter()` is _required_ to avoid the bugs that this PR is fixing.

A new check for the `ENTRY_REQUIRES_COLLAPSING` flag maintenance bugs is probably a good idea in both cases. If nothing else, it would eliminate the suspicion that transactions get stack because this flag is not properly propagated across find().

When we discover a bad entry, we should delete it so that others will not rediscover it, falling into the same trap. Deletion is easy/fast IIRC. Thus, I think we can eliminate option 2 from your earlier [list](https://github.com/squid-cache/squid/pull/516#issuecomment-559579283) of options. This leaves us with option 1 and option 3.

If all options have the same set of find() checks, do you think the third [option](https://github.com/squid-cache/squid/pull/516#issuecomment-559579283) is still better than the first? What makes it (or the first option) the best in your opinion?

eduard-bagdasaryan:
> do you think the third option is still better than the first

I think this depends on cache sizes and other environment specifics. For example, the third has an advantage that we do not recreate the still anchorable Transients entry each time we hit it. However, if on average the hit rate is low before the cache is overwritten, the first approach is preferable, I think.

eduard-bagdasaryan:
> A new check for the ENTRY_REQUIRES_COLLAPSING flag maintenance bugs is probably a good idea in both cases.

Done at cdb21a1.

> Same as (2) but remove the 'troubled' Transients entry just after it could not be obtained (due to 'anchoring' failure or lack of writer).

Done at e5ba792.

src/ipc/StoreMap.cc
```diff
@@ -253,6 +253,18 @@ Ipc::StoreMap::peekAtReader(const sfileno fileno) const
...
 
+const Ipc::StoreMap::Anchor *
+Ipc::StoreMap::peekAtWriter(const sfileno fileno) const
+{
+    const Anchor &s = anchorAt(fileno);
+    if (s.writing())
+        return &s; // immediate access by lock holder so no locking
+    if (s.reading())
+        return nullptr;
+    assert(false); // must be locked for reading or writing
```
yadij: Is this StoreMap class used exclusively by Transients table? if not, this assert is a bit risky - the non-transient caches using it may just have these pointing at regular cache entries with nobody using them at this time.

Same logic applies for peekAtReader()

src/ipc/StoreMap.h
```diff
@@ -262,6 +262,9 @@ class StoreMap
     /// only works on locked entries; returns nil unless the slice is readable
     const Anchor *peekAtReader(const sfileno fileno) const;
 
+    /// only works on locked entries; returns nil unless the slice is writable
```
yadij: This is incomplete. The comment inside the definition indicates that this may *only* be called by the lock holder itself. Other objects/code/threads/workers may not call this method regardless of its lock status.

src/Transients.h
```diff
@@ -85,6 +85,9 @@ class Transients: public Store::Controlled, public Ipc::StoreMapCleaner
     bool isReader(const StoreEntry &) const;
     /// whether the entry is in "writing to Transients" I/O state
     bool isWriter(const StoreEntry &) const;
+    /// whether there is a writer into the Transients entry, this
```
yadij: This comment is very confusing. It is easier to read the code than to decipher the Americanization of the compound noun split between beginning and end of the sentence.
Perhapse use the same style as the isWriter() documentation?
```suggestion
    /// whether the StoreEntry is attached anywhere for "writing to Transients" I/O
```

src/Transients.h
```diff
@@ -85,6 +85,9 @@ class Transients: public Store::Controlled, public Ipc::StoreMapCleaner
...
+    /// StoreEntry is attached to
```
yadij:
```suggestion
```

src/ipc/StoreMap.h
eduard-bagdasaryan: There may be several lock holders: a writer and some readers. The description explains that this method may be called only _after_ the entry was opened for writing or reading, i.e., acquired such a lock.

rousskov: > Amos: Other objects/code/threads/workers may not call this method regardless of its lock status.

What you are saying is essentiallyIn SMP world, one cannot tell whether an entry is locked without locking it... implied by the existing description: In SMP world, one cannot tell whether an entry is locked without locking it. Thus, "other objects/code/threads/workers" (i.e. those without a lock) must not call this method because it only works on locked entries, and those "others" _cannot tell_ whether the entry is locked. Those others can only check whether the entry was locked at some point in the past, and that is not enough to satisfy this method precondition.

If you prefer, we can replace "only works on locked entries" with "the caller must hold a lock on the entry" or something like that. This replacement should be done for both peekAt*() methods. Due to its length, it should probably be moved to a second description line. @yadij, would this replacement address your concern?

src/ipc/StoreMap.cc
rousskov: > Eduard: The assertion below does check this requirement.

... and that assertion requires an `s.reading()` check above it, answering the other question in Amos' comment:

> Amos: It is also not clear why peeking at the writer requires any info about a reader.

src/Transients.h
rousskov: Eduard's replacement is OK (after `s/are/is/`). Here are more options to consider if Amos does not like Eduard's replacement:

`whether isWriter() is true for some SMP kid`
`whether there is an SMP kid which isWriter()`
`whether there is an SMP kid for which isWriter() is true`

@yadij, does the latest committed version or any of the above work well enough for you?

src/ipc/StoreMap.cc
yadij: Okay, so is there any reason other than this assert to why this method is tied to s.reading() ?

This makes more sense to me (assuming you really need the assert at all):

> if (s.writing()) return &s; // immediate access by lock holder so no locking
> assert(s.reading()); // must be locked for reading or writing
> return nullptr;

So the reader status is not relevant (always nullptr if no writer), you get the assert info (if needed), and optimized builds are not slowed down by checking read locks.

src/ipc/StoreMap.h
yadij: > Those others can only check whether the entry was locked at some point in the past, and that is not enough to satisfy this method precondition.

That is the answer to my question then. Only the lock holder may call this.

> 
```suggestion
    /// the caller must hold a lock on the entry
    /// \returns nullptr unless the slice is writable
```

src/ipc/StoreMap.h
```diff
@@ -262,6 +262,9 @@ class StoreMap
     /// only works on locked entries; returns nil unless the slice is readable
```
yadij:
```suggestion
    /// the caller must hold a lock on the entry
    //// \returns nullptr unless the slice is readable
```

src/ipc/StoreMap.cc
eduard-bagdasaryan: Done.

src/ipc/StoreMap.h
eduard-bagdasaryan: Done.

src/ipc/StoreMap.h
eduard-bagdasaryan: Done.

src/Transients.h
yadij: Latest commit is fine.

src/store/Controller.cc
```diff
@@ -355,6 +355,21 @@ Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
         const bool found = anchorToCache(entry, inSync);
         if (found && !inSync)
             throw TexcHere("cannot sync");
+        if (!found) {
+            if (transients->hasWriter(entry)) {
+                Transients::EntryStatus entryStatus;
+                transients->status(entry, entryStatus);
+                if (!entryStatus.collapsed) {
```
rousskov: Why are we checking flags in shared memory rather than calling entry.hittingRequiresCollapsing()? For the caller, it is that latter flag that makes all the difference, right? AFAICT, unknown bugs notwithstanding, Transients::get() should have set that flag for these `!found` StoreEntries.

src/ipc/StoreMap.h
```diff
@@ -259,9 +259,14 @@ class StoreMap
...
-    /// only works on locked entries; returns nil unless the slice is readable
+    /// the caller must hold a lock on the entry
+    /// \returns nullptr unless the slice is readable
     const Anchor *peekAtReader(const sfileno fileno) const;
 
+    /// the caller must hold a lock on the entry
+    /// \returns nullptr unless the slice is writeable
+    const Anchor *peekAtWriter(const sfileno fileno) const;
+
     /// only works on locked entries; returns the corresponding Anchor
```
rousskov: Stale comment format. Please update the description to stay in sync with the other two peeking method descriptions.

src/ipc/StoreMap.h
```diff
@@ -259,9 +259,14 @@ class StoreMap
     /// undoes partial update, unlocks, and cleans up
     void abortUpdating(Update &update);
 
-    /// only works on locked entries; returns nil unless the slice is readable
+    /// the caller must hold a lock on the entry
+    /// \returns nullptr unless the slice is readable
     const Anchor *peekAtReader(const sfileno fileno) const;
 
+    /// the caller must hold a lock on the entry
```
rousskov: I have not checked, but I suspect that placing the `caller must hold a lock on the entry` line on top will make _that_ phrase as the brief method description in Doxygen. That would be unfortunate because the line does not summarize what the method is about. If I am right, consider moving that line down, after the `\returns` line _if_ doing so would solve the problem -- I do not know whether Doxygen is smart enough to use the `\returns` line as the brief method description.

eduard-bagdasaryan: No, reordering these lines did not do the trick. I experimented also with different related Doxygen options but could not get 'returns' line in the brief description. The only option I know to overcome this - revert back to a single line description without 'returns' tag.

src/ipc/StoreMap.h
eduard-bagdasaryan: Done.

src/ipc/StoreMap.h
rousskov: > The only option I know to overcome this - revert back to a single line description without 'returns' tag.

OK, let's leave the descriptions as they are now. Further improvements are probably not worth the time spent fighting Doxygen and reviewers.

src/Transients.cc
```diff
@@ -275,6 +283,11 @@ Transients::completeWriting(const StoreEntry &e)
 {
     assert(e.hasTransients());
     assert(isWriter(e));
+    // The not freed Transients entry may eventually become 'unattached',
+    // so that it cannot be anchored anymore. Though we handle this situation,
+    // there are some risks the entry metadata becomes stale and, hence, we
+    // should drop such entries. TODO: should we drop this entry immediately to
+    // avoid such risks?
```
rousskov: AFAICT, we are concerned that an idle unattached Transient entry loses its (presumed) ability to become attached to a cache store because another, unrelated transaction purges the corresponding store entry. The Controller must (and now does) prevent such a no-longer-attachable Transient entry from being given to Store clients, but its metadata might become stale, making Controller task more difficult/riskier.

However, Transients::completeWriting() is not the place that deals with idle entries -- Store::Controller::handleIdleEntry() is. _This_ method does not create idle entries. If the corresponding StoreEntry is public, it needs this Transients entry -- we cannot "drop this entry immediately". Otherwise, we should probably rely on handleIdleEntry() being called soon rather than preempt or duplicate its future actions.

Given your existing use cases (that this PR fixes), can you tell why the Transients entry was not freed when it became idle? Or did it become idle _after_ its disk entry was purged, because the Transients entry somehow lost attachment to its disk entry (while remaining busy)?

eduard-bagdasaryan: I placed this comment in Transients::completeWriting() because old Squids (pre 4310f8b) used this method to mark the Transients entry for removal:

```
Transients::completeWriting(const StoreEntry &e)
{
    if (e.mem_obj && e.mem_obj->xitTable.index >= 0) {
    [...]
        map->freeEntry(e.mem_obj->xitTable.index); // just marks the locked entry
    [...]
    }
}
```

Once the Transients entry is 'marked', it is not available for the public anymore. Eventually, such 'marked' entry may become unlocked (idle) but this is probably irrelevant here. Perhaps I need to change the comment's 'drop entry' to 'mark entry' for better understanding.

rousskov: Marking the Transients entry in Transients::completeWriting() does not make sense to me: There is nothing wrong with the Transients entry, so why make it inaccessible to new concurrent transactions?

Right now, I see two options for going forward:

1. Remove this comment. The fixes already committed should prevent the known bug from happening again. This is sufficient.
2. Understand what the underlying issue is and add a corresponding TODO to document and address the underlying cause (assuming we do not want to address it in this PR for some valid reason). This is ideal.

It feels like we have or should have enough information to move from the "sufficient" to "ideal" outcome here. That is why I am asking why the Transients entry was not deleted in your use case(s) after it became idle. Did we not call handleIdleEntry() or did handleIdleEntry() decide to keep the idle Transients entry? Why? Or perhaps it never became idle but still lost its lock on the disk (or memory) cache entry? Something else?

And when did the StoreEntry corresponding to the TransientsEntry lost its association with the disk (or memory) store? Is that loss expected/normal?

Somewhere, there is a conflict between our expectations and actual Squid behavior. This bug is the outcome of that conflict. The current fix does not expose which of our basic expectations does not match the code. The fix correctly reacts to a problematic situation, but there may be more we can and should do to either avoid that situation in more cases or adjust our expectations. If we have a use case that can reproduce the original problem, it would be great to use that use case to pin-point the source of that conflict (at least)...

eduard-bagdasaryan: > Did we not call handleIdleEntry()

We call handleIdleEntry().

> did handleIdleEntry() decide to keep the idle Transients entry?  Why? 

Yes, it keeps the idle Transients entry.
Transients::handleIdleEntry() calls destroyStoreEntry() and then StoreEntry::destroyMemObject(). Eventually Transients::disconnect() is called. Since we have 'reader' entry there, StoreMap::closeForReading() is called which just releases the reader lock without marking or freeing the slot. If the entry were 'writer' there (old Squid behavior), StoreMap::abortWriting() would be called, freeing (or marking) the slot.

> And when did the StoreEntry corresponding to the TransientsEntry lost its association with the disk (or memory) store? 

It loses its association with the disk in Rock::SwapDir::disconnect() (called from destroyStoreEntry()) and with memory in MemStore::disconnect(), called via Controller from StoreEntry::destroyMemObject(). That loss seems expected: we disconnect  the StoreEntry from caches before deleting it.

rousskov: > Eventually Transients::disconnect() is called. Since we have 'reader' entry there, StoreMap::closeForReading() is called which just releases the reader lock without marking or freeing the slot.

Aha, this is good news! The underlying "stale entry" problem now boils down to the fact that Transients are reusing a caching-oriented StoreMap which does not release idle valid entries by default. Going forward, I see two options:

1. Add StoreMap::closeForReadingAndFreeIdle() and call it instead of calling closeForReading() in Transients.cc. I see two such closeForReading() calls but my code copy may be stale.
2. Just add a `TODO: s/closeForReading/closeForReadingAndFreeIdle/g` comment to the "we are reading" case of Transients::disconnect().

If it does not have some hidden dangers, I lean towards option 1 because it allows us to get rid of these likely-to-become-stale idle Transients. What do you think?

----

closeForReadingAndFreeIdle() is a long name, but the operation, assuming it is possible to implement at all, has to be atomic from the other workers point of view. Thus, we cannot split it in two. We could add a `freeIfIdle` option to closeForReading() instead, but I think it is best to be more explicit here -- the method code when that option is turned on will probably be rather different from the basic closeForReading() code. If possible, closeForReadingAndFreeIdle() should reuse closeForReading(), of course.

eduard-bagdasaryan: > closeForReadingAndFreeIdle() is a long name, but the operation, assuming it is possible to implement at all, has to be atomic from the other workers point of view.

Note that some of the existing StoreMap methods are not 'atomic' when solving similar problems, e.g., StoreMap::abortWriting().  This method checks an atomic condition and performs an action, depending on that condition:

```
if (!s.lock.readers) {
    freeChain(fileno, s, false);
```
If, for example, lock.readers==0 and freeChain() is going to start but another worker has just attached as a reader, it will face a problem, since the slice chain it is going to read is being freed now.

rousskov: > If ... another worker has just attached as a reader

You probably missed the following at the start of StoreMap::abortWriting().

```C++
    s.lock.appending = false; // locks out any new readers
```

abortWriting() is only called for locked-for-writing entries so `lock.appending` is required to allow any concurrent readers.

> some of the existing StoreMap methods are not 'atomic'

IIRC, there are no known bugs in this area.

eduard-bagdasaryan: No, StoreMap::abortWriting() seems to be correct: new readers are filtered out, making 'appending' false.

eduard-bagdasaryan: Here is a draft of the suggested closeForReadingAndFreeIdle():

```
void
Ipc::StoreMap::closeForReadingAndFreeIdle(const sfileno fileno)
{
    closeForReading(fileno);
    // free if idle
    if (lock.lockExclusive()) {
        assert(!s.writing());
        assert(!s.reading());
        Anchor &s = anchorAt(fileno);
        freeChain(fileno, s, false); 
    }
}

```

rousskov: Is there a way to avoid leaving the "unlocked entry" gap between closeForReading() and lockExclusive() without adding new/complex APIs? I know that the gap does not lead to "bugs", but we should strive for reliable handling of _locked_ entries, and the current draft implementation is not reliable.

eduard-bagdasaryan: Done at 96b229f, please review. If we are happy with this solution, I'll undo the fix already applied 
to Controller::allowSharing(), except for that useful 'hittingRequiresCollapsing()' check.

eduard-bagdasaryan: > I'll undo the fix already applied to Controller::allowSharing(),

Done at ffc8ea76.

src/store/Controller.cc
```diff
@@ -355,6 +355,13 @@ Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
         const bool found = anchorToCache(entry, inSync);
         if (found && !inSync)
             throw TexcHere("cannot sync");
+        if (!found) {
+            assert(transients->hasWriter(entry));
```
eduard-bagdasaryan: I hesitated when deciding whether we should assert here or report to admin. However, since we now remove idle Transients entries, there should be no situation when both '!found' and !hasWriter() conditions are true: if the writer disappears before delivering the response, it should 'mark' the entry (making it unavailable). Similarly, for a cached response without writer, 'found' should be true.

rousskov: We cannot assert that `hasWriter` _stays_ true _after_ a false `found` was computed.

eduard-bagdasaryan: Fixed at a9cfe77.

src/store/Controller.cc
```diff
@@ -357,11 +357,12 @@ Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
...
-                // Cannot anchor right now and will not be able in the future.
-                // Prevent others from falling into the same trap.
+                // prevent others from falling into the same trap
                 stopSharing(entry);
-                throw TexcHere("unattached transients entry");
-            } else if (!entry.hittingRequiresCollapsing()) {
+                throw TexcHere("unattached transients entry missing writer");
+            }
+
+            if (!entry.hittingRequiresCollapsing()) {
```
rousskov: Should we move this BUG check up, before we may bail due to the just-disappeared writer? It feels like the lack of a hittingRequiresCollapsing() flag for a `!found` entry is a bug, regardless of whether the writer has disappeared after the Transients::get() call or not. Am I missing something here? Perhaps this move requires some other pending changes related to the hittingRequiresCollapsing() flag?

eduard-bagdasaryan: > whether the writer has disappeared after the Transients::get() call or not.

Also, there could be no writer at Transients::get() time, I think we cannot distinguish these two situations. If there was no writer, hittingRequiresCollapsing() remains false and there is no bug here. That is why I check for a writer first.  After that,  we can assert that the existing writer implies hittingRequiresCollapsing().

rousskov: I can rephrase my concern as the following expectation: `!found` should imply hittingRequiresCollapsing(). This expectation is the strongest upon successfully leaving allowSharing(), but it would probably be best to preserve that invariant at false anchorToCache() exit time as well. My expectation may be wrong, but, if it is wrong, then, unless I am missing something obvious, the code should not suggest such a wrong expectation in the first place.

eduard-bagdasaryan: >  there could be no writer at Transients::get() time

I think this hypothetical situation could happen in a case of a bug in the newly added StoreMap::closeForReadingAndFreeIdle() and other routines responsible for that cleanup. Anyway, even if such a situation happens, checking for hittingRequiresCollapsing() first should reveal that problem (as well as the previous hasWriter() check), so I do not mind moving the 'BUG' check up (done at c531911). I had to split the !isWriter condition into two: I think we should always stopSharing() in this case, even if hittingRequiresCollapsing() check fails too.

rousskov: I see two red flags in the current implementation, and addressing the second concern will create the third red flag:

1. The current (stopSharing(), check hittingRequiresCollapsing()) order implies that an unshareable entry may still be collapsed on. While the current implementation of those two methods may allow for any call order, that is likely a side effect of a hidden problem with one or both methods rather than a fundamental invariant that is likely to survive future code adjustments. To avoid this implication, we should check hittingRequiresCollapsing() (or remember its value) _before_ calling stopSharing().
2. Calling stopSharing() from startSharing() implies that we want to undo (some of the) startSharing() effects. I think this is a good idea, but it looks strange that we only stopSharing() upon discovery of some, but not necessarily all sharing problems. To address this concern, consider moving the stopSharing() call into find(), where allowSharing() is called, and where exceptions are already handled.
3. When handling exceptions, find() keeps the entry shareable (i.e. it calls release() with a true flag). In this PR, we want to call stopSharing() when throwing an exception. Please double check whether that release() parameter in find() should actually be false. And, if it should be false, please check whether we should call stopSharing() from StoreEntry::release(false) instead of doing that directly from the find()'s `catch` clause.

eduard-bagdasaryan: I tried to address your concerns at c859658.

> we should check hittingRequiresCollapsing() (or remember its value) before calling stopSharing()

I agree: since after stopSharing() has been called no more 'hitting' is possible on this entry, 'collapsing' simply does not have any sense.

> consider moving the stopSharing() call into find(), where allowSharing() is called

This concern looks reasonable: for symmetry/simplicity sake it is better to call stopSharing()/allowSharing() at one level, than one from another.

> Please double check whether that release() parameter in find() should actually be false

That release() parameter effectively sets the StoreEntry::shareableWhenPrivate flag value. I think it should be false, at least for the two cases in question:
  * unexpected !hittingRequiresCollapsing() indicate an invalid (and unrecoverable) entry state and BUG report does warn about that. We should stop sharing this entry even for already attached readers.
  * unexpected !hasWriter means that no future data notifications will happen and, hence, no sharing for already attached readers is possible (for otherwise, they will hang forever).

We can probably formulate a general rule that any exception in Controller::allowSharing() scope indicates that the entry is not suitable for any sharing, including 'private sharing'.


> whether we should call stopSharing() from StoreEntry::release(false)

We should not call stopSharing() from there because it is already performed (implicitly): stopSharing() effectively means Transients::evictCached(), whereas release() calls Controller::evictCached(), which evicts the entry from all caches.

rousskov: > We should not call stopSharing() from release() because it is already performed (implicitly): [both call] evictCached()

Yes, release() appears to call evictCached(), even on the releaseRequest() path, because our StoreEntry should have a key -- we call addReading() earlier in allowSharing().

I think this PR should be content with the `release(false)` call. Compared to release(false), stopSharing() looks like a "milder" sharing-focused action that may not do enough to fully get rid of the problematic entry. There may be more, deeper problems here, but no obvious ones that fall into this PR scope. I am resolving this change request.

rousskov:
The [failed build](https://build.squid-cache.org/job/5-pr-test/1625/compiler=clang,label=d-ubuntu-bionic/console) looked like a Jenkins/etc. bug to me, so I restarted it (queued as built 1​63​7?). I do not know whether that restart will advance this PR. If somebody knows that it is not correct or sufficient, then please do whatever is needed to advance this PR.

PR #518, opened 2019-12-10
Detail TLS and CONNECT cache_peer negotiation failures

chtsanti:
Before PeerConnector and Tunneler were introduced, FwdState and
TunnelStateData naturally owned their to-server connection. When CONNECT
and TLS negotiation were outsourced, we kept that ownership to minimize
changes and simplify negotiation code. That was wrong because FwdState
and TunnelStateData, as connection owners, had to monitor for connection
closures but could not distinguish basic TCP peer closures from complex
CONNECT/TLS negotiation failures that required further detailing. The
user got generic error messages instead of details known to negotiators.

Now, Ssl::PeerConnector and Http::Tunneler jobs own the connection they
work with and, hence, are responsible for monitoring it and, upon
successful negotiation, returning it to the initiators. In case of
problems, these jobs send detailed errors to the initiators instead.

Passing connection ownership to and from a helper job is difficult
because the connection may be either closed or begin to close (e.g. by
shutdown) while the callback is pending without working close handlers.
Many changes focus on keeping Connection::fd in sync with Comm.

Also improved tunnel.cc mimicking of (better) FwdState code: Partially
open connections after Comm::ConnOpener failures are now closed, and
Http::Tunneler failures are now retried.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -807,7 +807,11 @@ FwdState::establishTunnelThruProxy(const Comm::ConnectionPointer &conn)
                                             Http::Tunneler::CbDialer<FwdState>(&FwdState::tunnelEstablishmentDone, this));
     HttpRequest::Pointer requestPointer = request;
     const auto tunneler = new Http::Tunneler(conn, requestPointer, callback, connectingTimeout(serverConnection()), al);
-    tunneler->usesPconn_ = true; // TODO: Replace this hack with proper Connection-Pool association
+
+    // TODO: Replace this hack with proper Comm::Connection-Pool association
+    // that is not tied to fwdPconnPool and can handle disappearing pools.
+    tunneler->noteFwdPconnUse = true;
+
 #if USE_DELAY_POOLS
     Must(serverConnection()->getPeer());
     if (!serverConnection()->getPeer()->options.no_delay)
@@ -882,7 +886,7 @@ FwdState::secureConnectionToPeerIfNeeded(const Comm::ConnectionPointer &conn)
         else
 #endif
             connector = new Security::BlindPeerConnector(requestPointer, conn, callback, al, sslNegotiationTimeout);
-        connector->usesPconn_ = true;
+        connector->noteFwdPconnUse = true;
         AsyncJob::Start(connector); // will call our callback
         return;
     }
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -841,15 +841,13 @@ void
 FwdState::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
 {
     ErrorState *error = nullptr;
-
     if (!answer.positive()) {
-        // TODO: Reuse to-peer connections after a CONNECT error response.
-
         error = answer.squidError.get();
         Must(error);
         answer.squidError.clear(); // preserve error for fail()
-    }else if (!Comm::IsConnOpen(answer.conn) || fd_table[answer.conn->fd].closing()) {
-        // connection gone before we are called.
+        Must(!Comm::IsConnOpen(answer.conn));
+    } else if (!Comm::IsConnOpen(answer.conn) || fd_table[answer.conn->fd].closing()) {
+        // the connection could get closed while our callback was queued
         serverConnection(serverConn);
         closeServerConnection("closing connection");
         error = new ErrorState(ERR_CANNOT_FORWARD, Http::scServiceUnavailable, request, al);
@@ -866,7 +864,6 @@ FwdState::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
         error = new ErrorState(ERR_CONNECT_FAIL, Http::scBadGateway, request, al);
         closeServerConnection("found early data after CONNECT response");
     }
-
     if (error) {
         fail(error);
         retryOrBail();
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -847,22 +847,21 @@ FwdState::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
         answer.squidError.clear(); // preserve error for fail()
         Must(!Comm::IsConnOpen(answer.conn));
     } else if (!Comm::IsConnOpen(answer.conn) || fd_table[answer.conn->fd].closing()) {
-        // the connection could get closed while our callback was queued
-        serverConnection(serverConn);
-        closeServerConnection("closing connection");
+        // The socket could get closed while our callback was queued.
+        // We close Connection here to sync Connection::fd.
+        answer.conn->close();
         error = new ErrorState(ERR_CANNOT_FORWARD, Http::scServiceUnavailable, request, al);
     } else if (!answer.leftovers.isEmpty()) {
         // This should not happen because TLS servers do not speak first. If we
         // have to handle this, then pass answer.leftovers via a PeerConnector
         // to ServerBio. See ClientBio::setReadBufData().
-        syncWithServerConn(answer.conn, request->url.host(), false);
         static int occurrences = 0;
         const auto level = (occurrences++ < 100) ? DBG_IMPORTANT : 2;
         debugs(17, level, "ERROR: Early data after CONNECT response. " <<
                "Found " << answer.leftovers.length() << " bytes. " <<
-               "Closing " << serverConnection());
+               "Closing " << answer.conn);
         error = new ErrorState(ERR_CONNECT_FAIL, Http::scBadGateway, request, al);
-        closeServerConnection("found early data after CONNECT response");
+        answer.conn->close();
     }
     if (error) {
         fail(error);
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -80,6 +80,14 @@ PconnPool *fwdPconnPool = new PconnPool("server-peers", nullptr);
 
 CBDATA_CLASS_INIT(FwdState);
 
+/// emulate AsyncJob call protections
+#define FwdStateEnterThrowingCode() try {
+#define FwdStateExitThrowingCode() } \
+    catch (...) { \
+    debugs (17, 2, "exception: " << CurrentException); \
+    retryOrBail(); \
+    }
+
 class FwdStatePeerAnswerDialer: public CallDialer, public Security::PeerConnector::CbDialer
 {
 public:
@@ -758,6 +766,8 @@ FwdState::handleUnregisteredServerEnd()
 void
 FwdState::noteConnection(HappyConnOpener::Answer &answer)
 {
+    FwdStateEnterThrowingCode();
+
     calls.connector = nullptr;
     connOpener.clear();
 
@@ -797,6 +807,8 @@ FwdState::noteConnection(HappyConnOpener::Answer &answer)
     }
 
     secureConnectionToPeerIfNeeded(conn);
+
+    FwdStateExitThrowingCode();
 }
 
 void
@@ -825,6 +837,8 @@ FwdState::establishTunnelThruProxy(const Comm::ConnectionPointer &conn)
 void
 FwdState::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
 {
+    FwdStateEnterThrowingCode();
+
     ErrorState *error = nullptr;
 
     if (!answer.positive()) {
@@ -859,6 +873,8 @@ FwdState::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
     }
 
     secureConnectionToPeerIfNeeded(answer.conn);
+
+    FwdStateExitThrowingCode();
 }
 
 /// handles an established TCP connection to peer (including origin servers)
@@ -978,6 +994,8 @@ FwdState::syncHierNote(const Comm::ConnectionPointer &server, const char *host)
 void
 FwdState::connectStart()
 {
+    FwdStateEnterThrowingCode();
+
     debugs(17, 3, *destinations << " to " << entry->url());
 
     Must(!request->pinnedConnection());
@@ -1011,6 +1029,8 @@ FwdState::connectStart()
     destinations->notificationPending = true; // start() is async
     connOpener = cs;
     AsyncJob::Start(cs);
+
+    FwdStateExitThrowingCode();
 }
 
 /// send request on an existing connection dedicated to the requesting client
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -80,6 +80,8 @@ PconnPool *fwdPconnPool = new PconnPool("server-peers", nullptr);
 
 CBDATA_CLASS_INIT(FwdState);
 
+// TODO: Extract destination-specific handling from FwdState so that it can be
+// surrounded with a single try/catch,retry block written without these macros.
 /// emulate AsyncJob call protections
 #define FwdStateEnterThrowingCode() try {
 #define FwdStateExitThrowingCode() } \
@@ -766,8 +768,6 @@ FwdState::handleUnregisteredServerEnd()
 void
 FwdState::noteConnection(HappyConnOpener::Answer &answer)
 {
-    FwdStateEnterThrowingCode();
-
     calls.connector = nullptr;
     connOpener.clear();
 
@@ -807,13 +807,13 @@ FwdState::noteConnection(HappyConnOpener::Answer &answer)
     }
 
     secureConnectionToPeerIfNeeded(conn);
-
-    FwdStateExitThrowingCode();
 }
 
 void
 FwdState::establishTunnelThruProxy(const Comm::ConnectionPointer &conn)
 {
+    FwdStateEnterThrowingCode();
+
     AsyncCall::Pointer callback = asyncCall(17,4,
                                             "FwdState::tunnelEstablishmentDone",
                                             Http::Tunneler::CbDialer<FwdState>(&FwdState::tunnelEstablishmentDone, this));
@@ -831,14 +831,14 @@ FwdState::establishTunnelThruProxy(const Comm::ConnectionPointer &conn)
 #endif
     AsyncJob::Start(tunneler);
     // and wait for the tunnelEstablishmentDone() call
+
+    FwdStateExitThrowingCode();
 }
 
 /// resumes operations after the (possibly failed) HTTP CONNECT exchange
 void
 FwdState::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
 {
-    FwdStateEnterThrowingCode();
-
     ErrorState *error = nullptr;
 
     if (!answer.positive()) {
@@ -873,8 +873,6 @@ FwdState::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
     }
 
     secureConnectionToPeerIfNeeded(answer.conn);
-
-    FwdStateExitThrowingCode();
 }
 
 /// handles an established TCP connection to peer (including origin servers)
@@ -897,6 +895,7 @@ FwdState::secureConnectionToPeerIfNeeded(const Comm::ConnectionPointer &conn)
     const bool needTlsToOrigin = !p && request->url.getScheme() == AnyP::PROTO_HTTPS && !clientFirstBump;
 
     if (needTlsToPeer || needTlsToOrigin || needsBump) {
+        FwdStateEnterThrowingCode();
         HttpRequest::Pointer requestPointer = request;
         AsyncCall::Pointer callback = asyncCall(17,4,
                                                 "FwdState::ConnectedToPeer",
@@ -911,6 +910,7 @@ FwdState::secureConnectionToPeerIfNeeded(const Comm::ConnectionPointer &conn)
             connector = new Security::BlindPeerConnector(requestPointer, conn, callback, al, sslNegotiationTimeout);
         connector->noteFwdPconnUse = true;
         AsyncJob::Start(connector); // will call our callback
+        FwdStateExitThrowingCode();
         return;
     }
 
@@ -994,8 +994,6 @@ FwdState::syncHierNote(const Comm::ConnectionPointer &server, const char *host)
 void
 FwdState::connectStart()
 {
-    FwdStateEnterThrowingCode();
-
     debugs(17, 3, *destinations << " to " << entry->url());
 
     Must(!request->pinnedConnection());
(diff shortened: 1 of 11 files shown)


src/tunnel.cc
```diff
@@ -995,38 +1047,89 @@ tunnelStart(ClientHttpRequest * http)
...
+    connectedToPeer(answer.conn);
+}
+
+void
+TunnelStateData::connectedToPeer(const Comm::ConnectionPointer &conn)
+{
+    advanceDestination("establish tunnel thru proxy", conn, [this,&conn] {
+        establishTunnelThruProxy(conn);
+    });
+    // XXX: In tests, if establishTunnelThruProxy() throws and
```
rousskov: @chtsanti, please try to reproduce this XXX. Why do we not get the expected outcome here?

src/FwdState.cc
```diff
@@ -871,25 +909,35 @@ FwdState::secureConnectionToPeerIfNeeded()
...
+
+    // if not encrypting just run the post-connect actions
+    successfullyConnectedToPeer(conn);
+}
+
+/// encrypts an established TCP connection to peer (including origin servers)
+void
+FwdState::secureConnectionToPeer(const Comm::ConnectionPointer &conn)
+{
+    { // TODO: Remove this diff reduction before commit
```
rousskov: We should not forget to remove this, but let's not remove it yet.

src/tunnel.cc
```diff
@@ -995,38 +1047,89 @@ tunnelStart(ClientHttpRequest * http)
...
+
+    connectedToPeer(conn);
+}
+
+/// encrypts an established TCP connection to peer
+void
+TunnelStateData::secureConnectionToPeer(const Comm::ConnectionPointer &conn)
+{
+    { // TODO: Remove this diff reduction before commit
+        { // TODO: Remove this diff reduction before commit
```
rousskov: We should not forget to remove these two, but let's keep them for now.

src/FwdState.cc
```diff
@@ -793,20 +823,28 @@ FwdState::noteConnection(HappyConnOpener::Answer &answer)
...
+FwdState::establishTunnelThruProxy(const Comm::ConnectionPointer &conn)
 {
     AsyncCall::Pointer callback = asyncCall(17,4,
                                             "FwdState::tunnelEstablishmentDone",
                                             Http::Tunneler::CbDialer<FwdState>(&FwdState::tunnelEstablishmentDone, this));
     HttpRequest::Pointer requestPointer = request;
-    const auto tunneler = new Http::Tunneler(serverConnection(), requestPointer, callback, connectingTimeout(serverConnection()), al);
+    const auto tunneler = new Http::Tunneler(conn, requestPointer, callback, connectingTimeout(serverConnection()), al);
+
+    // TODO: Replace this hack with proper Comm::Connection-Pool association
```
yadij: Can you clarify for me why we need a pool here?
It sounds like the tunneler needs to be a pool owner. That is easy enough to implement since a pool is just a container class which may be empty at any time.

src/tunnel.cc
chtsanti: In my tests I am seeing that this is not correct, the message will appear after the last owner of the connection object, the AccessLogEntry::hier::tcpServer, be destroyed.
I removed this XXX comment in my commit 6eab3c8.

However while working for this I found other problems with one crash fixed in commit b5d3dcf enough serious. I added also some other fixes please review.

src/FwdState.cc
chtsanti: The fwdPconnPool is used only for FwdState.cc but not for tunnel.cc, however both are using the Http::Tunneler. 
If the fwdPconnPool is used then the Http::Tunneler should update its statistics in the case of errors.
We have a similar flag for the HappyConnOpener the HappyConnOpener::allowPconn_ to tell to HappyConnOpener whether to use fwdPconnPool or not.

Maybe it is not difficult to fix this inconsistency, but I believe it is required a different PR for this.

rousskov: A proper solution is far from trivial. IIRC, we have even attempted to implement it during this or another recent project, and the changes quickly grew out of control. This TODO should be addressed in a dedicated PR.

src/security/PeerConnector.cc
```diff
@@ -543,13 +550,37 @@ Security::PeerConnector::bail(ErrorState *error)
...
+Security::PeerConnector::sendSuccess()
+{
+    callBack();
+    disconnect();
+}
+
+void
+Security::PeerConnector::disconnect()
+{
+    if (!Comm::IsConnOpen(serverConnection()))
```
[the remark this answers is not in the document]
chtsanti: > This check is wrong (see Tunneler::disconnect() for details). AFAICT, 06f1381 (which added this check) describes a real bug (even though the commit message refers to a non-existent PeerConector::retryOrBail() method), 

Sorry I mean PeerConnector::bail() method.

> but the root cause of that bug is not inside disconnect(), and the correct solution for that bug is different.

I decided to make this fix (commit 06f1381), because of two reasons:
- Inside Security::PeerConnector::start there is the following
`
if (!Comm::IsConnOpen(serverConn) || fd_table[serverConn->fd].closing())
        return bail(new ErrorState(ERR_CONNECT_FAIL, Http::scBadGateway, request.getRaw(), al));
`
This is implies that the serverConn maybe be closed for an external reason.
- Having in my mind the possibility where an external user closes the Comm::Connection object we are owning, setting the Comm::Connection::fd to -1. T

Both of the above are only possible if we have a bug, because the PeerConnector is the only owner of Comm::Connection object. For example I entered this state forcing an exception after PeerConnector created but before AsyncJob::Start is called, where in practice there is not any code which can cause an exception. We can have the fd_table[serverConn->fd].closing() but (hopefully) not closed the Comm::Connection.

But because there is the check Comm::IsConnOpen in PeerConnector::start I made the check inside disconnect before use the comm_remove_close_handler for consistency and protect squid from bugs in other places.

> I do not know whether the following change will be enough to address the root cause, but PeerConnector::connectionClosed (i.e. the connection closure callback) must nullify closeHandler (and possibly the server connection itself!).

It does it. But uses the PeerConnector::bail (which calls disconnect) to do it.
Probably we can do it with a cleaner way however requires many changes in current code.
It is better to not try it now, but if you think this is required I can try.

> Removing PeerConnector::connectionClosed(), moving its functionality inside PeerConnector::commCloseHandler() is probably a good idea as well -- why spread the same functionality across two methods

commit 2a53a4663

src/clients/HttpTunneler.cc
```diff
@@ -341,13 +351,49 @@ Http::Tunneler::bailWith(ErrorState *error)
...
+{
+    assert(answer().positive());
+    callBack();
+    disconnect();
+}
+
+void
+Http::Tunneler::disconnect()
+{
+    if (!Comm::IsConnOpen(connection))
```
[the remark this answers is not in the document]
chtsanti: > We must reach that disconnected state even if the connection is already closing.

The most possible is that the disconnect name is not good, we should rename it to be "stopMonitoringServer" or something like that.  Because we are not actually disconnecting from connection here, the connection object may continue used by Tunneler after this call is called.

Please also check my comments bellow for PeerConnector::disconnect method. More or less are valid for this method to.

src/tunnel.cc
```diff
@@ -995,38 +1043,87 @@ tunnelStart(ClientHttpRequest * http)
...
+    // TODO: Extract destination-specific handling from TunnelStateData so that
+    // all the awkward, limited-scope advanceDestination() calls can be replaced
+    // with a single simple try/catch,retry block.
+    try {
+        startStep();
+        // now wait for the step callback
+    } catch (...) {
+        debugs (26, 2, "exception while trying to " << stepDescription << ": " << CurrentException);
+        closePendingConnection(conn, "connection preparation exception");
+        saveError(new ErrorState(ERR_CANNOT_FORWARD, Http::scInternalServerError, request.getRaw(), al));
```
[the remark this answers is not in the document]
chtsanti: commit 81916b6a72

src/security/PeerConnector.cc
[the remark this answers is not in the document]
rousskov: > I decided to make this fix (commit 06f1381), because of two reasons:

> Security::PeerConnector::start ... implies that the serverConn maybe be closed for an external reason.

Yes, it can be. External connection closures is one of the primary reasons why Squid still has a connection closure handler concept (and why this PR adds a start() check).

> Having in my mind the possibility where an external user closes the Comm::Connection object we are owning,

The external connection closure is possible but ...

>  setting the Comm::Connection::fd to -1.

Assume this part is impossible. We are the only Connection owner. Only we can initiate setting Comm::Connection::fd to -1. Without that assumption, it is virtually impossible to write reasonable Connection-handling job code. If you know of any specific place that violates this assumption for PeerConnector, please let me know.

> Both of the above are only possible if we have a bug

The first situation (i.e. external connection closure) is actually possible in some contexts today. One example is the shutdown sequence that initiates closures for connections that shutdown code does not own. Another example is ModPoll's DoSelect(). There may be more. I do not think PeerConnector is directly exposed to the shutdown problem _today_, but, as long as such external closures exist in Squid, I think it is best to assume that PeerConnector (and any other Connection-owning job) is or will be exposed to an external closure.

> because there is the check Comm::IsConnOpen in PeerConnector::start I made the check inside disconnect for consistency and protect squid from bugs in other places.

Consistency is good, but you cannot apply it to the `if` statement condition alone. You have to extend the principle to the code that handles the condition: The start() method does not just quit after detecting an external connection closure! Instead, start() correctly proceeds with what the PeerConnector job is supposed to do (in this case, it means returning an error to the job initiator). Similarly, disconnect() should not just quit -- it should complete what it was asked to do (i.e., prevent any Comm callbacks from reaching the job). Currently, disconnect() violates its promise to disconnect.

I suspect that if you fix that disconnect() violation, then the condition inside that method may itself disappear (or change), but that is a secondary speculation.

> Alex: PeerConnector::connectionClosed must nullify closeHandler.
> Christos: It does it [via] PeerConnector::bail() which calls disconnect().

You are missing the point: disconnect() _removes_ an active closure handler _registration_ (if any). That is _not_ what the closure handler code must do! The closure handler code must clear the handler member (i.e. make closeHandler nil). That is the only way to restore the correct job state. This restoration must happen before calling virtually any other job method because other methods usually assume that the job state is correct. It would be insanely difficult to write correct code if internal methods cannot rely on the job state being valid.

It is possible, even likely, that the closure handler must do _more_ (after syncing the job state with the external changes), of course, but I am focusing on this single critical aspect for now.

> Probably we can do it with a cleaner way however requires many changes in current code.

I doubt the latter is true, but we have to do the right thing here regardless of the cost. Not nullifying closeHandler in a closure handler method is a bug. Fixing that bug is in this PR scope -- this PR works (in part) on correct handling of Connection ownership and connection closure handler is an important part of that ownership lifetime.

src/clients/HttpTunneler.cc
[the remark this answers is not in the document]
rousskov: > we are not actually disconnecting from connection here

We are disconnecting from Comm (i.e. making sure that Comm does not call our job). I am comfortable with the current method name, but am happy to consider better alternatives. For example, if you would prefer disconnectFromComm(), please use that instead.

src/security/PeerConnector.cc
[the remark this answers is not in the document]
chtsanti: > >  Security::PeerConnector::start ... implies that the serverConn maybe be closed for an external reason.
 
> Yes, it can be. External connection closures is one of the primary reasons why Squid still has a connection closure handler concept (and why this PR adds a start() check).

> >    Having in my mind the possibility where an external user closes the Comm::Connection object we are owning,

> The external connection closure is possible but ...

 >>  setting the Comm::Connection::fd to -1.

> Assume this part is impossible. We are the only Connection owner. Only we can initiate setting Comm::Connection::fd to -1.

Saying that the "Comm::Connection is closed" or "external user closes the Comm::Connection object" I mean that the Comm::Connection::close is called for this object which sets the Connection::fd to -1. The Comm::IsConnOpen actually  only checks if the Connection::fd is not -1.
This is why I am saying that the Comm::Connection may "closing" (because an external user closed the fd, and we can expect the comm close handler clear us), but is difficult to be closed (have the Comm::Connection::fd to -1)

> Without that assumption, it is virtually impossible to write reasonable Connection-handling job code. If you know of any specific place that violates this assumption for PeerConnector, please let me know.

I agree but we had such bugs in the past. This patch between others also fixes a such bug, because the Comm::Connection object shared with FwdStateData which closed the connection object.
Currently I do not know any specific place which can violate this assumption. But we may have it in the future if an exception thrown after the PeerConnector object is created but before the AsyncJob::Start is executed.  Because of a bug.

> I doubt the latter is true, but we have to do the right thing here regardless of the cost. Not nullifying closeHandler in a closure handler method is a bug. Fixing that bug is in this PR scope -- this PR works (in part) on correct handling of Connection ownership and connection closure handler is an important part of that ownership lifetime.

OK I will try to fix this.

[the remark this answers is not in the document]
rousskov: Thank you for detailing your terminology. I hope we are on the same page now. I am looking forward to your next update.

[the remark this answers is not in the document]
chtsanti: I commited the patches e9d742e9 a5277e0 daec389f and db3e4ea35 which are related to this.

To answer on Alex questions:

> I do not know whether the following change will be enough to address the root cause, but PeerConnector::connectionClosed (i.e. the connection closure callback) must nullify closeHandler (and possibly the server connection itself!).

This is not enough, we are going to have crash inside PeerConnector::disconnect while the commUnsetConnTimeout be called.

> If necessary, we can remember the timeout callback and then nullify that as well, in PeerConnector::connectionClosed() and in PeerConnector::commTimeoutHandler() callbacks.

I did not do that. Doing this will not resulted to a better code. To avoid code duplication (in commCloseHandler and commTimeoutHandler) a new method required which will looked very similar with disconnect method. The PeerConnector::bail method which is called by commCloseHandler, commTimeoutHandler  and on errors should do the required work without any problem.

Before proceed I must recall that closing the fd is different than closing the Comm::Connection object. An fd can be closing but its Comm::Connection object still considered as open, until the related commClose handlers are called.

My major problem is who can close the server Comm::Connection object. There are classes like the ConnStateData class where its Comm::Connection object can be closed by anyone inside squid and we are expecting that the commClose handler will clear its state. To handle this we have many checks into this class for Comm::IsConnOpen(clientConnection) to check if an external closed our Comm::Connection object.

For the Security::PeerConnector and the Http::Tunneler, thinks looks simpler we have not to do this. We can assume that we are the only owners of our Comm::Connection object and this object can be closed only  by us because of an error or because the commClose handler is called.
Internally we should always call PeerConnector::bail and Http::Tunneler::bailWith methods to clear our state (remove handlers, respond to callers etc) and close the Comm::Connection object.  Moreover these methods allow/require sending an error/reason for the closed connection to the caller, which in practice is required for these classes.

Please also read the comments in my commits.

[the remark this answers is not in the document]
rousskov: > I must recall that closing the fd is different than closing the Comm::Connection object

Yes, this unfortunate but important fact is probably the source of several bugs. I appreciate you reminding me of that fact and, AFAICT, I agree with your logic surrounding Connection ownership. I made three untested polishing commits on top of your changes: f63ce51, 625f70a, and e61ac40. If you are OK with them, we can resolve this change request.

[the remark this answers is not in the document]
chtsanti: Just the commit 625f70a is not enough to fix squid crashes. Unfortunately the
PeekingPeerConnector::clientConn can be closed, while the request->clientConnectionManager is still valid, for example after a (lifetime?) timeout.

I make the commit 7471528 which replaces the assertion with an "if" check and also add a similar check inside  PeekingPeerConnector::noteNegotiationDone where the switchToTunnel is called which assumes that the clientConn is open.
I hope this is ok.

src/ssl/PeekingPeerConnector.cc
```diff
@@ -215,6 +218,10 @@ Ssl::PeekingPeerConnector::noteNegotiationDone(ErrorState *error)
     if (!request->clientConnectionManager.valid() || !fd_table[serverConnection()->fd].ssl)
         return;
 
+    // Abort if client connection is closed
+    if (!Comm::IsConnOpen(clientConn))
+        return;
```
rousskov: This check seems premature. Most of the code below does not use clientConn and may update transaction state. I suggest that we check clientConn only when we actually need/use it. In this method, that happens only when calling switchToTunnel(). The latter assumes an open connection -- we should add a Must() or assert() to switchToTunnel().

I am also concerned about the "do nothing and hope that the caller will figure it out" strategy. The comment says "abort" but we are aborting nothing. We simply do not do what the caller may expect us to do. Perhaps we should throw instead of returning?

src/security/PeerConnector.cc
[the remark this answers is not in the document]
rousskov: I am OK with your 7471528 changes in principle and am resolving this old change request. The next review will focus on the remaining/current problem.

src/ssl/PeekingPeerConnector.cc
chtsanti: > This check seems premature. Most of the code below does not use clientConn and may successfully update transaction state with useful (e.g., logged server datails) information. I suggest that we check clientConn only when we actually need/use it.

Yep.


>  In this method, that happens only when we decide to call switchToTunnel(). As you know, the latter assumes an open client connection -- we should add the corresponding Must() or assert() to switchToTunnel(). Let's move this check just above the switchToTunnel() call?

Yes, I added this "Must". Moreover the clientConn does not needed in switchToTunnel function it can retrieve it from request->clientConnectionManager->clientConnection. The same for PeekingPeerConnector.  But this PR is already fixes many-many thinks.

> I am also concerned about the implicit "do nothing and hope that the caller will figure it out" strategy. The source code comment says "abort", but we are aborting nothing. We simply do not do what the caller may expect us to do. We could call mustStop() before returning, but it feels like we should throw instead of successfully returning. What do you think?

You are right. My only problem is the DBG_IMPORTANT debug message about a BUG inside PeerConnector::swanSong which will be printed in this case. 
Is it ok to replace this DBG_IMPORANT with an other to level 2 or 3?

rousskov: > My only problem is the DBG_IMPORTANT debug message about a BUG inside PeerConnector::swanSong which will be printed in this case. Is it ok to replace this DBG_IMPORANT with an other to level 2 or 3?

We do not have a formalized policy about this (yet), but I am leaning towards the following: In new code: Must()s are similar to asserts -- the code should not use Must()s as a reaction to a known/anticipated problem (e.g., a client connection closure). If we follow this principle, then the `BUG` message in swanSong() should stay at `DBG_IMPORTANT` -- it indicates an unexpected problem (that may result in memory leaks or be otherwise mishandled).

Please investigate whether noteNegotiationDone() can call bail() and _then_ throw? The former would be proper handling of the anticipated problem and the latter would be a proper action for a method that cannot complete as promised.

chtsanti: > Please investigate whether noteNegotiationDone() can call bail() and then throw

Yep this is looks that works well.  I was stuck thinking of using the "Must" clause.
Commits 8d2a60c5 and 456ad0ab.

rousskov: Thank you for addressing my concern.

chtsanti:
> Thank you for addressing my previous concerns. I know of only one problem with the current code.

commits 806a1bca, 5cfa25e11, 686c2c93a and bf842d5c548

In all of this cases now there is a check:
 `if (!Comm::IsConnOpen(answer.conn) || fd_table[answer.conn->fd].closing())`
I think this is should actually be  
 `if (Comm::IsConnOpen(answer.conn) && fd_table[answer.conn->fd].closing())`
and later after all the ifs add a check "Must(Comm::IsConnOpen(answer.conn))". This is because is impossible to receive a closed answer.conn object if there is not error. But the existing form already used and I did not change it. Also we do not know if in the future there answers with  closed answer.conn objects.

> If the connection got closed during answer delivery:
> 
>     * a direct Must() fails in connectDone()

> ...
> If the connection got closed during answer delivery:
> 
>     * a direct Must() fails in syncWithServerConn();
> 
> ...
The answer.conn Comm::Connection object can not be closed here (Comm::IsConnOpen(answer.conn) is always true) however the  fd_table[answer.conn->fd].closing() can be true. The problem here is that there is not any active comm-close handler when the fd socket was closed and any comm-close handler we are going to add after the socket start closing will not be used. We will never informed that the socket is closed and most possible squid will crash later when we are going to read/write to a closing socket or maybe try to use a fd/socket used by an other new Comm::Connection object.

I think I found an inconsistency in PeekingPeerConnector/PeerConnector answers when splicing is selected. The Security::EncryptorAnswer object when the FwdState::connectedToPeer called will have attached a Comm::Connection object which is also used by a TunnelStateData object (because we spliced and builded one).

I did not found a good ways to fix the above. I think there are  two easy solutions:
1) One which throws after splicing. But I do not like, exceptions are slow should used only for errors.
2) One which will check in Security::PeerConnector::sendSuccess() if the serverConn is not null and only then call disconnect. It has its own problems.

But in the other hand is not important, we can ignore it for now, or add an XXX/TODO inside FwdState::connectedToPeer where the answer.tunneled case is handled.

chtsanti:
Alex I believe I handled all remaining issues into this branch.
Please re-review.

rousskov:
> Please re-review.

Just FYI: In the future, you can (and should) click the "request review" icon in the reviewers section of the PR page (top right corner). This allows the reviewer to quickly find PRs awaiting their reviews using GitHub PR search interface.

I clicked that icon for this PR. The ball is on my side.

src/clients/HttpTunneler.cc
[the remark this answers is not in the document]
chtsanti: I think this is resolved with commit a378a06

src/tunnel.cc
chtsanti: commit 62df15ad

src/FwdState.cc
chtsanti: commit 62df15ad

rousskov:
@yadij, please let me know if you need help porting this large but important change back to v5.

PR #519, opened 2019-12-10
Fix request URL generation in reverse proxy configurations

guidovranken:


src/client_side.cc
```diff
@@ -1148,26 +1148,28 @@ prepareAcceleratedURL(ConnStateData * conn, const Http1::RequestParserPointer &h
...
-                snprintf(thost, sizeof(thost), "%s:%d",host, vport);
-                host = thost;
+            const auto lastColonPos = host.rfind(':');
+            if (lastColonPos != SBuf::npos) {
+                if (*host.rbegin() != ']') {
+                    // address with a port (IPv4, IPv6, or domain name)
+                    host.chop(0, lastColonPos); // truncate until the last colon
+                    host.appendf(":%d", vport);
+                }
+                // else: IPv6 address without a port
```
rousskov: Please check whether this case needs to be handled (differently).

yadij: good catch. yes the appendf line should be out a few scopes. That makes the new code much simpler.

yadij: I have fixed this in 0a4a970

rousskov:
I adjusted PR title to be less developer-specific. I adjusted PR description because the Squid Project uses commit Author field (which should be set to @guidovranken by Anubis), instead of `Signed-off-by` pseudo-headers.

@guidovranken, please feel free to add yourself to `CONTRIBUTORS` (IIRC that will eventually happen semi-automatically based on the Author field, but I do not know how often that synchronization happens).

Due to these potentially sensitive changes, I removed `M-cleared-for-merge` label to give others a chance to weigh in before the PR is auto-merged.

rousskov:
OK to test

yadij:
LGTM

PR #520, opened 2019-12-17
Fix shared memory size calculation on 64-bit systems

eduard-bagdasaryan:
Since commit 2253ee0, the wrong type (uint32 instead of size_t) was used
to calculate the PagePool::theLevels size. theLevels memory (positioned
by different and correct code) did not overlap with the raw pages
buffer, but the raw pages buffer could, in some cases, be 32 bits short,
placing the last 4 bytes of the last page outside of allocated memory.

In practice, shared memory allocations are page-aligned, and the
difference in 4 bytes was probably compensated by the extra allocated
bytes in most (or perhaps even all) cases.

### DIFF
--- a/src/ipc/mem/PageStack.cc
+++ b/src/ipc/mem/PageStack.cc
@@ -122,7 +122,7 @@ Ipc::Mem::PageStack::sharedMemorySize() const
 size_t
 Ipc::Mem::PageStack::SharedMemorySize(const uint32_t, const unsigned int capacity, const size_t pageSize)
 {
-    const size_t levelsSize = PageId::maxPurpose * sizeof(std::atomic<Ipc::Mem::PageStack::Value>);
+    const size_t levelsSize = PageId::maxPurpose * sizeof(Levels_t);
     const size_t pagesDataSize = capacity * pageSize;
     return StackSize(capacity) + LevelsPaddingSize(capacity) + levelsSize + pagesDataSize;
 }
--- a/src/ipc/mem/PageStack.cc
+++ b/src/ipc/mem/PageStack.cc
@@ -122,7 +122,7 @@ Ipc::Mem::PageStack::sharedMemorySize() const
 size_t
 Ipc::Mem::PageStack::SharedMemorySize(const uint32_t, const unsigned int capacity, const size_t pageSize)
 {
-    const size_t levelsSize = PageId::maxPurpose * sizeof(Levels_t);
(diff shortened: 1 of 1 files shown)


yadij:
You mention code duplication in the description. Can you please add a comment to the duplicated code sections (all known) mentioning that they need to be in sync. If they do not have to be kept in sync, then it is not a code duplication problem but a dev copy-paste mistake.

Also, can we use decltype() instead of the typedef name to further clarify what this calculation is about?

eduard-bagdasaryan:
> You mention code duplication in the description. Can you please add a comment to the duplicated code sections (all known) mentioning that they need to be in sync

It is my assumption that in 2253ee0 this code was mistakenly duplicated (or copy-pasted, which is the same for me), probably from `const Ipc::Mem::PageStack::Value Writable = 0;` declaration. I am not insisting on that and if you wish, we can be more specific and change the phrasing into 'due to a mistake done at 2253ee0' or something like that. Also, note that after this fix there are no places to 'keep in sync' and no comments are needed.

> Also, can we use decltype() instead of the typedef name to further clarify what this calculation is about?

Sorry, I do not see how we can improve with decltype. We need to warrant that PagePool::theLevels type and PageStack::Levels_t are the same, but this is already guaranteed by PagePool class implementation.

rousskov:
> Eduard: we can be more specific and change the phrasing into 'due to a mistake done at 2253ee0' or something like that.

We can just say `Since 2253ee0, ...` or even nothing at all (especially if Amos has doubts about that 2253ee0 attribution). I wholeheartedly agree that the exact classification of this bug (e.g., "copy-paste", "type duplication by copy-paste", "creative code duplication", "misnamed type", or whatever) is not important at all. We do not need to spend time debating this!

@yadij, I edited the PR description to mention that specific commit instead of the disputed "duplication", but please remove that attribution if you do not like it for any reason.

> Amos: Can you please add a comment to the duplicated code sections (all known) mentioning that they need to be in sync

> Eduard: after this fix there are no places to 'keep in sync' and no comments are needed

Agreed. The fix removes the duplication (or whatever it was). No "keep in sync" comments are needed/possible.

> Amos: can we use decltype() instead of the typedef name to further clarify what this calculation is about?

We cannot use decltype(*theLevels) or similar because PagePool::theLevels data member is not accessible in the fixed PageStack::SharedMemorySize() method. This and several similar problems are the consequences of poor split of responsibilities between PageStack and PagePool classes, but that known problem is outside this PR scope.

PR #521, opened 2019-12-18
Fix FTP buffers handling

chtsanti:
Fix the parsing of the received listing from FTP services.
Also relaxed size/filename grammar used for DOS listings: Tolerate
multiple spaces between the size and the filename.


yadij:
I think we are going in a slightly wrong direction here. The long term destination we want to achieve is converting the I/O into an SBuf buffer and parsing with Tokenizer.

Is there a bug closed with this half-measure change? if so please reference it properly in the title/description. Otherwise this change should be held for the Squid-6 cycle.

PR #522, opened 2019-12-18
Fix server_cert_fingerprint on cert validator-reported errors

chtsanti:
The server_cert_fingerprint ACL mismatched when sslproxy_cert_error
directive was applied to validation errors reported by the certificate
validator because the ACL could not find the server certificate.

### DIFF
--- a/src/acl/FilledChecklist.h
+++ b/src/acl/FilledChecklist.h
@@ -89,7 +89,10 @@ public:
 
     /// SSL [certificate validation] errors, in undefined order
     const Security::CertErrors *sslErrors;
-    /// The peer certificate
+
+    /// Peer certificate being checked by ssl_verify_cb() and by
+    /// Security::PeerConnector class. In other contexts, the peer
+    /// certificate is retrieved via ALE or ConnStateData::serverBump.
     Security::CertPointer serverCert;
 
     AccessLogEntry::Pointer al; ///< info for the future access.log, and external ACL
--- a/src/acl/ServerCertificate.cc
+++ b/src/acl/ServerCertificate.cc
@@ -25,7 +25,7 @@ ACLServerCertificateStrategy::match(ACLData<MatchType> * &data, ACLFilledCheckli
     if (checklist->serverCert)
         cert = checklist->serverCert;
     else if (checklist->al && Comm::IsConnOpen(checklist->al->hier.tcpServer)) {
-        auto ssl = fd_table[checklist->al->hier.tcpServer->fd].ssl.get();
+        const auto ssl = fd_table[checklist->al->hier.tcpServer->fd].ssl.get();
         cert.resetWithoutLocking(SSL_get_peer_certificate(ssl));
     } else if (checklist->conn() != NULL && checklist->conn()->serverBump())
         cert = checklist->conn()->serverBump()->serverCert;
--- a/src/acl/ServerCertificate.cc
+++ b/src/acl/ServerCertificate.cc
@@ -27,7 +27,7 @@ ACLServerCertificateStrategy::match(ACLData<MatchType> * &data, ACLFilledCheckli
     else if (checklist->al && Comm::IsConnOpen(checklist->al->hier.tcpServer)) {
         const auto ssl = fd_table[checklist->al->hier.tcpServer->fd].ssl.get();
         cert.resetWithoutLocking(SSL_get_peer_certificate(ssl));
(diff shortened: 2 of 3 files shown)


src/acl/FilledChecklist.h
```diff
@@ -89,7 +89,10 @@ class ACLFilledChecklist: public ACLChecklist
 
     /// SSL [certificate validation] errors, in undefined order
     const Security::CertErrors *sslErrors;
-    /// The peer certificate
+
+    /// Peer certificate being checked by ssl_verify_cb() and by
+    /// Security::PeerConnector class. In other contexts, the peer
+    /// certificate is retrieved via ALE or ConnStateData::serverBump.
     Security::CertPointer serverCert;
```
yadij: Can we move this into the ALE instead? it makes sense to store the serverCert there as soon as it is known regardless of its validity - exactly for cases like this.

yadij: The ICAP related details should be stored in ALE::icap sub-class and the upstream server/peer cert in the ALE class itself. Ideally that would be the MasterXaction, but still two certs to handle even then.

The architectural issue I can see is that we have to adjust ACLs to check both locations for relevance, and eventually add flags to target which data source to match against. I do not insist on the latter part here.

rousskov: > The architectural issue I can see is that we have to adjust ACLs

You are thinking in the right direction but oversimplifying the issue.

Yes, the ACLs can be adjusted, but simply "checking both locations for relevance" is not going to work well because, without flags or equivalent, the same ACL will not be able to tell which certificate, HTTP or ICAP, is the "right" one for the current context. In the final/fixed design, either certificate can be the "right" one, depending on the admin intent (expressed, for example, via ACL namespaces or flags).

And there is more than just ACLs. For example, logformat %code expansion deals with similar problems.

These architectural issues require serious thinking and discussion, but those activities are outside this PR scope.

> it would be nice to have done

Yes, but not in this small bug-fixing PR. Thank you for not insisting on that change.

rousskov: @yadij, you have not answered my question: Do you really want Christos to move this special certificate field into ALE which already provides access to a certificate? Doing so will not solve any problems AFAICT, and will make this PR larger, but if you insist on that change, it can be done.

chtsanti: If we re going to save the ICAP server certificate in a new ALE field, still the ACLServerCertificateStrategy should informed with a way which certificate to use, the ICAP or HTTP peer, depending the context it is called. So If we remove the FilledChecklist::serverCert we need a flag here, eg a FilledCheckList::context which will get the values HttpContext or IcapContext.

Alternate solution maybe is to use different acls for ICAP and HTTP server certificates. For example use an  http_server_cert_fingerprint and an icap_server_cert_fingerprint acls.

rousskov: > If we re going to save the ICAP server certificate in a new ALE field, still the ACLServerCertificateStrategy should informed with a way which certificate to use, the ICAP or HTTP peer, depending the context it is called

* Yes, if we set the new ALE field permanently (i.e. once when the ICAP certificate becomes known).
* No, if we only set the new ALE field temporary, in contexts where it must be used (i.e. just like what this PR does today, but using the new ALE field).

Neither is better than what this PR does today IMHO. That is why I asked @yadij whether he insists on moving the field to ALE. Hopefully, this change will not be necessary at all, but we have to wait for Amos to respond.

> An alternate solution can use different acls for ICAP and HTTP server certificates

Yes, but doing that _properly_ requires developing a comprehensive solution with changes going well beyond this bug-fixing PR. Adding a couple of ICAP-specific ACLs without a long-term strategy is likely to make things worse in the long run.

yadij: > @yadij, you have not answered my question: Do you really want Christos to move this special certificate field into ALE which already provides access to a certificate?

It would be ideal, but I do not insist on it.

rousskov: > Amos: It would be ideal, but I do not insist on it.

I suspect moving this special old field into ALE would create more problems than it would solve; the ideal long-term solution lies elsewhere. Christos does not seem to be excited about the move either. You do not insist on the bove. Thus, I do not think we should do it.

@yadij, AFAICT, you have not requested any other changes. Please dismiss your negative review or request other changes to move this PR forward.

yadij:
v5 now restricted to security issues.

PR #523, opened 2019-12-19
Fixed prohibitively slow search for new SMP shm pages

eduard-bagdasaryan:
The original Ipc::Mem::PageStack algorithm used an optimistic linear
search to locate the next free page. Measurements showed that, in
certain cases, that search could take seconds on busy caches, iterating
over millions of page index items and effectively stalling all workers
while showing 100% CPU utilization.

The new code uses a deterministic stack. It is still lock-free. The spin
loops around stack head pointer updates are expected to quit after at
most few iterations, even with a large number of workers. These loops do
not have ABA update problems. They are not spin locks.

### DIFF
--- a/src/ipc/mem/PageStack.cc
+++ b/src/ipc/mem/PageStack.cc
@@ -25,15 +25,15 @@ static_assert(sizeof(Ipc::Mem::PageStackStorageSlot::Pointer) ==
 void
 Ipc::Mem::PageStackStorageSlot::take()
 {
-    const auto next = nextOrMarker.exchange(TakenPage);
-    assert(next != TakenPage);
+    const auto nxt = nextOrMarker.exchange(TakenPage);
+    assert(nxt != TakenPage);
 }
 
 void
-Ipc::Mem::PageStackStorageSlot::put(const PointerOrMarker expected, const Pointer next)
+Ipc::Mem::PageStackStorageSlot::put(const PointerOrMarker expected, const Pointer nxt)
 {
-    assert(next != TakenPage);
-    const auto old = nextOrMarker.exchange(next);
+    assert(nxt != TakenPage);
+    const auto old = nextOrMarker.exchange(nxt);
     assert(old == expected);
 }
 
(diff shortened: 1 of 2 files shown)


src/ipc/mem/PageStack.cc
```diff
@@ -15,128 +15,133 @@
...
-const Ipc::Mem::PageStack::Value Writable = 0;
-
-Ipc::Mem::PageStack::PageStack(const uint32_t aPoolId, const unsigned int aCapacity, const size_t aPageSize):
-    thePoolId(aPoolId), theCapacity(aCapacity), thePageSize(aPageSize),
-    theSize(theCapacity),
-    theLastReadable(prev(theSize)), theFirstWritable(next(theLastReadable)),
-    theItems(aCapacity)
+/* Ipc::Mem::PageStackStorageSlot */
+
+// We are using uint32_t for Pointer because PageId::number is uint32_t.
```
yadij: This check does not care about what the type size is, so long as they are matching.

IMO this text should be at the place the Pointer type is defined. It may also be useful there to indicate "larger" means over 16TB or whatever N*4 GB is for page size N.

src/ipc/mem/PageStack.h
```diff
@@ -45,39 +83,36 @@ class PageStack
...
     size_t sharedMemorySize() const;
 
     /// shared memory size required only by PageStack, excluding
     /// shared counters and page data
-    static size_t StackSize(const unsigned int capacity);
+    static size_t StackSize(const PageCount capacity);
     size_t stackSize() const;
 
     /// \returns the number of padding bytes to align PagePool::theLevels array
     static size_t LevelsPaddingSize(const unsigned int capacity);
```
yadij: The type here should probably be using PageCount like all other capacity_ related items.

eduard-bagdasaryan: Done.

src/ipc/mem/PageStack.cc
eduard-bagdasaryan: Done.

yadij:
v5 now restricted to security issues.

PR #525, opened 2019-12-20
Centralized PagePool/PageStack ID generation

eduard-bagdasaryan:
Easy-to-find-in-cache.log and predictable/stable stack IDs for shared
memory pages and/or index slot numbers are very useful when debugging
cache metadata corruption issues because they allow to track related
(e.g. same-stack) operations across huge SMP logs.


eduard-bagdasaryan:
> giving a dedicated name to the ID type

Done at f320be3

PR #527, opened 2019-12-30
Invalidate cache when last chunk missing from upstream

DrDaveD:
See [bug 4735](https://bugs.squid-cache.org/show_bug.cgi?id=4735).  This implement's @rousskov's suggested fix, which worked perfectly in my testing.

PR #528, opened 2019-12-30
Bug 4735: Truncated chunked responses cached as whole

DrDaveD:
Mark responses received without the last chunk as responses that have
bad (and, hence, unknown) message body length (i.e. ENTRY_BAD_LENGTH).
If they were being cached, such responses will be released and will stop
being shareable.

### DIFF
--- a/src/http.cc
+++ b/src/http.cc
@@ -1505,6 +1505,9 @@ HttpStateData::processReplyBody()
 
         case COMPLETE_NONPERSISTENT_MSG:
             debugs(11, 5, "processReplyBody: COMPLETE_NONPERSISTENT_MSG from " << serverConnection);
+            if (flags.chunked && !lastChunk)
+                entry->lengthWentBad("missing last-chunk");
+
             serverComplete();
             return;
         }


DrDaveD:
This PR is a replacement of PR #527 for the master branch.

rousskov:
OK to test

DrDaveD:
The updated title & description look good.

The squid-to-client response both before and after the fix is truncated as it should be.

I hope this will get backported to v4.

yadij:
> The squid-to-client response both before and after the fix is truncated as it should be.

Does that "as it should be" mean the client connection is closed when the last byte is written/delivered?

DrDaveD:
> > The squid-to-client response both before and after the fix is truncated as it should be.
> 
> Does that "as it should be" mean the client connection is closed when the last byte is written/delivered?

Yes and it does not send the last zero to indicate a successfully completed transfer.

rousskov:
@DrDaveD, thanks a lot for checking/confirming! This and your other cleared PRs should be auto-merged (including the necessary rebasing) in a few days, once the objections timeout expires and the tests complete. You should not need to do anything else.

PR #531, opened 2019-12-31
Support plugin-style scripts for source format enforcement

yadij:
Allow the source-maintenance script to run arbitrary code or sub-scripts
to perform enforcement of Squid code style and content.

Code placed in the scripts/maintenance/ sub-folder MUST meet the
following criteria:

 * be self-executable,
 * receive filename of the code file to be touched as one and only
   command-line parameter,
 * always dump the file contents to stdout (with or without edits),
 * not depend on any other code in this sub-folder being run first.


scripts/maintenance.awk
```diff
@@ -0,0 +1,14 @@
...
+##
+## Squid software is distributed under GPLv2+ license and includes
+## contributions from numerous individuals and organizations.
+## Please see the COPYING and CONTRIBUTORS files for details.
+##
+
+#
+# This script performs easily automated code updates for
+# removal or renaming of deprecated/obsolete syntax.
+# It is run by maintenance script over .h, .c, .cc, and .cci files
```
rousskov: Please mention that the script is only applied to git-tracked files (i.e. generated files are not covered).

scripts/maintenance.awk
```diff
@@ -0,0 +1,14 @@
...
+#
+
+{ print }
```
rousskov: I do not understand why we should commit a script that does nothing. Please explain (probably here, in a PR comment, to start with).

yadij: This is the infrastructure part of scripting away the annoyances like HERE or NULL. Once this is in we can have lots of more targeted discussions about how to replace particular code patterns on an individual basis.

I chose to do this part as a standalone PR so that we can discuss this script change in terms of "how" to do changes rather than get bogged down into another long discussion about "what" of any particular change.

scripts/source-maintenance.sh
```diff
@@ -66,7 +66,8 @@ for FILENAME in `git ls-files`; do
 	#
 	# Code Style formatting maintenance
 	#
-        if test "${ASVER}"; then
+	awk -f ./scripts/maintenance.awk "${FILENAME}" >"${FILENAME}.awkbak" && mv "${FILENAME}.awkbak" "${FILENAME}"
```
rousskov: We should iterate through individual scripts named using a hard-coded pattern (e.g. `./scripts/maintenance-cpp-*` or `./scripts/maintenance/cpp-*`) rather than dumping all adjustments into one script.

The script execution command should be made interpreter-neutral (rather than forcing AWK usage for every adjustment script). The script itself should supply its interpreter name (and invocation parameters).

Finally, existing script(s) should be adjusted to use the new approach.

rousskov: Please also avoid using the backup-associated `bak` extension for newly created files. Something like `.next` or `.future` would be a lot more clear/natural.

scripts/maintenance.awk
```diff
@@ -0,0 +1,15 @@
+## Copyright (C) 1996-2019 The Squid Software Foundation and contributors
```
rousskov: After the "apply all scripts that are named using a hard-coded pattern" change request is implemented, this do-nothing file will not be needed in _this_ PR.

kinkie:
On Wed, 1 Jan 2020 at 17:59, Alex Rousskov <notifications@github.com> wrote:

>
> Generalizing the application of the source code formatting/adjusting
> scripts is probably a good idea. I left specific change requests in that
> direction.
>
> I am not yet convinced that HERE/NULL-removal scripts belong to the
> official Squid repository. By committing them, we essentially promise that
> they work reliably for all custom Squid code. IIRC, none of the known
> removal scripts are that good and, more importantly, making them that good
> is a waste of resources. After the official code is stripped from HERE and
> NULLs: Relative to the long-term nature of the official repository files,
> few people will need to apply conversion scripts to few branches for a
> short period of time.
>
I disagree here, I see no such promise. It’s up to the users of the script
to ensure they work for the specific use case.

> Instead of committing those temporary scripts to the official Squid
> repository, I suspect that we should automatically *check* that the new
> code does not contain HEREs, NULLs, and such. That check can mishandle a
> few esoteric cases without causing much harm. And it does not need to be in
> the official Squid repository.
>
It would be a manual process, and we will anyway be stuck in the middle of
the river for years, as shown by the last few years.
I'm in favour of biting the bullet and just going.

scripts/maintenance.awk
yadij: Done

scripts/source-maintenance.sh
yadij: How do you feel about adding dozens upon dozens of files containing one line of code each and which developers are then asked to run manually over their code? Sounds like a lot of annoyance for the sake of having this one PR meet your personal definition of perfection.

The reason for choosing AWK here over things like perl is that the rules are essentially just a regex pattern with some action associated.

Also, note that each of these files needs the shell command to parse and process the entire source code input file. Having two scripts would mean twice the processing for each file in the repository, three scripts thrice the processing and so on increasing for the number of individual changes being performed _ever_.

rousskov: > How do you feel about adding dozens upon dozens of files containing one line of code each and which developers are then asked to run manually over their code?

Feels like a straw man argument: Nobody suggested adding dozens of files or manual execution.

> The reason for choosing AWK here over things like perl is that the rules are essentially just a regex pattern with some action associated.

We should not be discussing advantages of one language over the other until we have the rules  that the scripts will implement, especially when each script can select its own "best" interpreter.

> Also, note that each of these files needs the shell command to parse and process the entire source code input file. Having two scripts would mean twice the processing for each file in the repository ...

Yes, the processing speed will decrease with the number of scripts, but we already execute a few scripts per file, and I do not expect us to add a lot more scripts.

rousskov:
> @rousskov: I am not yet convinced that HERE/NULL-removal scripts belong to the official Squid repository. By committing them, we essentially promise that they work reliably for all custom Squid code.

> @kinkie: I see no such promise. It’s up to the users of the script to ensure they work for the specific use case.

I see two problems with this logic:

1. It sounds like you are saying that Squid Project distributes two sets of programs: One set consists of programs where users may have a reasonable expectation that the program does what it is supposed to do, and another set where the user should be super careful with running the program from the set because it may mangle user's uncommitted source files, publish private keys, shut down the server, etc. I wonder where you see the line between those two sets? In other words, how can a user tell that a Project-distributed program is not reliable and requires case-by-case review/analysis?
2. The interface here is not an execution of an _individual_ source code adjusting script. It is an execution of `scripts/source-maintenance.sh` or, eventually, something like `make pretty`. It is unreasonable, IMO, to expect the user to carefully examine that target beyond a brief description of what it does. We _want_ (well, I cannot speak for others, so we _should_ want) developers to run `make pretty` (or equivalent) often and without fear or substantial time investment into safety analysis. That desire/recommendation implies, IMO, a certain level of confidence in the scripts general applicability and safety. The NULL/HERE removal scripts I have seen were meant for a different use case (i.e., careful "one-time" execution by folks who know what they are doing and can be warned), which allowed us to make those scripts much simpler.

> It would be a manual process,

I am not sure which process you are referring to, but Amos proposes adding a script to remove HEREs and NULLs. That script will be executed by `scripts/source-maintenance.sh`. No manual process beyond the execution of that script itself (which should be further automated behind a `make pretty` or a similar target).

> and we will anyway be stuck in the middle of the river for years,

Why?! The idea here, AFAICT, is to remove all NULLs and all HEREs from the supported branches in a few commits. If done right, this would take less than a week, even with all the special cases that should be done by hand. 

A developer can run the same script against their custom branches if they want to reduce future merge conflicts (and most developers either should not care about special cases or can fix them by cherry-picking official commits designated to those special cases). From a developer point of view, that will also take a few days.

And if some developer sits on their custom branch for a few years without applying these changes, then they are clearly not integrating with recent official code and hence are not a concern.

In summary, I see no "stuck for years" situation if we manage this correctly. This, BTW, is one of the reasons why I suggested removing NULLs and HEREs now, before branching v5. It would effectively remove the pain from two branches (v5 and master) at once, with no synchronization effort.

> as shown by the last few years.

During the last few years we were manually removing NULLs and HEREs in new/modified code lines. That painful process is the opposite of what the script-driven process should do, so it is not a valid comparison. To end these pains, various developers, including Amos and I, proposed scripts that would auto-remove most NULLs and HEREs. This specific PR adds a placeholder to integrate those scripts with `scripts/source-maintenance.sh`. I question the need for that integration, but I (still) support removing NULLs and HEREs using scripts (and possibly a little bit of manual effort for a few special cases in the official repositories).

rousskov:
Perhaps we can attack the same problem from another direction: Let's remove NULLs and HEREs from v4 and master using a pair of stand-alone scripts and then see what happens. _If_ there is sufficient demand for integrating those scripts with `scripts/source-maintenance.sh`, then we will integrate them.

scripts/source-maintenance.sh
yadij: > > How do you feel about adding dozens upon dozens of files containing one line of code each and which developers are then asked to run manually over their code?
> 
> Feels like a straw man argument: Nobody suggested adding dozens of files or manual execution.
> 

You suggested breaking this one script down into individual scripts applied using a pattern. That mess is the natural mid-to-long term outcome if each maintenance change needs its own script.

> > The reason for choosing AWK here over things like perl is that the rules are essentially just a regex pattern with some action associated.
> 
> We should not be discussing advantages of one language over the other until we have the rules that the scripts will implement, especially when each script can select its own "best" interpreter.
> 

Fair point.

> > Also, note that each of these files needs the shell command to parse and process the entire source code input file. Having two scripts would mean twice the processing for each file in the repository ...
> 
> Yes, the processing speed will decrease with the number of scripts, but we already execute a few scripts per file, and I do not expect us to add a lot more scripts.

This counter argument depends on that "feeling" about there never being many scripts. I expect there to be a great many scripts added if we split these up into multiple. With additional work some of the existing scripts may actually be removable, but the future will see about that.

yadij: Done using .new

yadij:
> > @rousskov: I am not yet convinced that HERE/NULL-removal scripts belong to the official Squid repository. By committing them, we essentially promise that they work reliably for all custom Squid code.
> 
> > @kinkie: I see no such promise. It’s up to the users of the script to ensure they work for the specific use case.
> 
> I see two problems with this logic:
> 
>     1. It sounds like you are saying that Squid Project distributes two sets of programs: One set consists of programs where users may have a reasonable expectation that the program does what it is supposed to do, and another set where the user should be super careful with running the 

We promise that the scripts/source-maintenance.sh does what we intend it to, with the source open for review if anyone has doubts about what that is. We do *not* promise that nothing gets mangled in the sources (custom or official). So even we have to read the patch it creates before commit. Right now that is done by me manually before triggering the Format Enforcement PRs to official.

> 
>     2. The interface here is not an execution of an _individual_ source code adjusting script.

We already expect people running the script to review its results. Nothing changes in that regard.

Outside of large PRs being committed with lots of formatting changes that output should be small or nil and restricted to the devs own touched code. There is a small change here in that these large PR events may be increased slightly by addition of new policy rules to the AWK script. We control that with the PR review process though, and the rule addition occurs first so we can use that as an easily identifiable rebase point for old code with generic how-to to reduce the pain.

> 
> > It would be a manual process,
> 
> I am not sure which process you are referring to, but Amos proposes adding a script to remove HEREs and NULLs.

Not exactly. I am proposing a script to do that *kind* of painful code edit less painfully. They are just the obvious examples.


> That script will be executed by `scripts/source-maintenance.sh`. No manual process beyond the execution of that script itself (which should be further automated behind a `make pretty` or a similar target).

That is not true. There is always manual review of its output needed (as mentioned above).


> 
> > and we will anyway be stuck in the middle of the river for years,
> 
> Why?! The idea here, AFAICT, is to remove all NULLs and all HEREs from the supported branches in a few commits. If done right, this would take less than a week, even with all the special cases that should be done by hand.
> 
> A developer can run the same script against their custom branches if they want to reduce future merge conflicts (and most developers either should not care about special cases or can fix them by cherry-picking official commits designated to those special cases). From a developer point of view, that will also take a few days.
> 
> And if some developer sits on their custom branch for a few years without applying these changes, then they are clearly not integrating with recent official code and hence are not a concern.
> 
> In summary, I see no "stuck for years" situation if we manage this correctly.

That was exactly the argument you used 9 years ago to argue for the painful manual way forward. Yet here we are, less than 1/7th of necessary removals completed. Somewhere between 56 and infinity years to go.

> This, BTW, is one of the reasons why I suggested removing NULLs and HEREs now, before branching v5. It would effectively remove the pain from two branches (v5 and master) at once, with no synchronization effort.

This is the argument I used at the v3.4, v3.5 and v4 branching time. Which you countered with how much work it would be for people with custom code forked off our repository. Irony.

Anyhow, this script PR is not for HERE/NULL removal. It is about reducing pain for that *type* of change. Even if we block again on those particular macros there are other code snippets that can use it for much smaller updates.

scripts/source-maintenance.sh
rousskov: > You suggested breaking this one script down into individual scripts applied using a pattern.

I did not.

rousskov:
> That was exactly the argument you used 9 years ago to argue for the painful manual way forward.

You are misrepresenting what happened 9 years ago.

rousskov:
> This is the argument I used at the v3.4, v3.5 and v4 branching time. Which you countered with how much work it would be for people with custom code forked off our repository. Irony.

You are misrepresenting what your NULL/HERE removal suggestion was and, hence, why it was rejected.

rousskov:
> We promise that the scripts/source-maintenance.sh does what we intend it to.

I agree. AFAICT, kinkie did not. My two points were meant to convince him to change (or clarify) his opinion. He has not done it yet, so we are still blocked on that key disagreement.

> We do not promise that nothing gets mangled in the sources.

I disagree. The first promise is meaningless without this second component. "We promise that this guillotine cures your headache" is not a meaningful promise in my world (thought I admit that a person overly focused on literal/dictionary interpretations may find such a promise meaningful). By requiring developers to run a script, we implicitly promise that it will not do serious irreversible harm. The proposed addition is incompatible with that promise because it may remove or mangle a source file (which may have no copies).

> So even we have to read the patch it creates before commit.

Sure, but reversing a patch cannot recover lost modifications because the patch is generated against committed/staged sources rather than sources that existed before the recommended script invocation. We should promise that the last step in the following sequence is possible:

1. git clone (state A)
2. edit src/main.cc (to arrive at state B)
3. run "make pretty" or equivalent (creating state C)
4. easily reverse "make pretty" changes, if any (back to state B)

The above sketch is not a complete/final specification, but a good starting point (more on that below).

> Kinkie: It’s up to the users of the script to ensure they work for the specific case

> Alex: 2. The interface here is not an execution of an _individual_ source code adjusting script. It is an execution of `scripts/source-maintenance.sh` or, eventually, something like `make pretty`. It is unreasonable, IMO, to expect the user to carefully examine that target beyond a brief description of what it does....

> Amos: We already expect people running the script to review its results. Nothing changes in that regard.

You are missing the context/point of my second bullet. Kinkie implied that users are responsible for figuring out whether the script is safe to run in _their_ use case -- we make no promises, especially about their custom code. My point was that it would be unreasonable to task the user with such an analysis because we are talking not just about (currently non-existent) small awk script but the entire source code mangling behemoth. I am talking about impracticality of the pre-run analysis (implicitly required by the no-promise approach). You are talking about post-run review of damages.

The related "manual process" discussion is similarly centered around the implication that users are going to apply some small manual action and, hence, can be tasked with making sure that the action script will do what they want. Your followup there appears to be missing that context/point -- we are talking past each other. I am skipping that part...

As for "Nothing changes in that regard", we do (or at least _should_) expect a change: Developers should eventually be required (or at least strongly encouraged) to run `make pretty` or equivalent. Today, there is no such requirement (and some of us actually _discourage_ running the source code formatting script and resist attempts to make it more portable or user-friendly).

This specific PR does not change the policy, of course, but one of the arguments attempting to justify this PR existence, AFAICT, is that other developers would be able to remove NULLs/HEREs (and alike) from their custom sources. That argument would be dead on arrival without the implication that we will start encouraging (or even requiring) developers to format their sources.

Please do not get me wrong though: I think we should require proper source code formatting, and I welcome changes that would get us closer to that goal. The infrastructure currently sketched by this PR is a small step in the opposite direction. However, before fixing that problem, we have to agree that the formatting script should either avoid making irreversible changes to sources or should refuse to run on uncommitted/unstaged sources (i.e. relying on git to preserve any custom code before it is formatted). This intent needs further formalization/detalization than the 4-step sketch above, but until there is consensus that we promise no harm (for some reasonable definition of harm), there is no point in detailing this further!

scripts/source-maintenance.sh
yadij: > > You suggested breaking this one script down into individual scripts applied using a pattern.
> 
> I did not.

The first sentence of your review comment was:
> We should iterate through individual scripts named using a hard-coded pattern ... rather than dumping all adjustments into one script.

Does that not mean placing the planned adjustments from this script into individual files and iterating through those?

rousskov: This continuing misunderstanding is driven, in part, by the lack of (shared) specifics regarding the future application of the proposed source code adjustment script(s). At first, there were no specifics at all. Then you mentioned HERE/NULL removal but did not like me saying that this is what the script(s) will do -- it is just an example of what it could do. So the future remains fuzzy, but here is my understanding of where we are going:

Squid will continue to make a few source code adjustments. Those adjustments can be naturally split into several scripts. We have at least one such script today. Those scripts will all be executed against each git-controlled C++-related file (for now). The names of those scripts will be hard-coded (as in the current PR) or identified using a simple filename pattern (as I have suggested). Please note that my primary concerns are not about that difference.

yadij: Very well. The proposed long-term contents of this script would be rules along the lines of these:

```
# FooDisplay macro is deprecated
/FooDisplay/ { gsub("FooDisplay","foo->print"); }

# BarStateData got renamed to Foo::Bar
/BarStateData/ { gsub("BarClass","Foo::Bar"); }

# NULL is deprectated by nullptr in C++11 code
# auto-remove common cases
/NULL/ {
  if ( FILENAME ~ /(src\/|tools\/|\.cci?)$/ ) {
    gsub(/[ ]=[ ]?NULL[;]/, " = nullptr;");
  }
}

# HERE macro is deprecated
# auto-remove common cases
/HERE/ {
  gsub("<< HERE \"", "\"");
  gsub(", HERE\)", ", MYNAME\)");
}
```

With rules added as we agree on changes (discussion in the PR adding the rule), and removed when we think it no longer needed (eg HERE actually deleted, so code containing it wont even build)

rousskov: If your plans materialize, we will have at least two scripts, one (existing) formatting sources and one (future) removing deprecated stuff. In that case, the iteration I suggested would iterate over those two scripts. We can hard-code that iteration (essentially like this PR is currently doing, but polished to address my other concerns):

```
apply maintenance/remove-deprecated.awk
apply maintenance/format-sources.pl
```

or generalize it (naming each of the two scripts to match some suitable shell pattern):

```
for script in maintenance/cpp-*
do
    apply $script
done
```

There are (obvious?) pros and cons for each approach. I suggested the latter because it makes adding dummy scripts unnecessary and makes this PR useful for generalizing safe script execution. I can live with hard-coding if others prefer or you insist on that.

I hope that this clarifies my change request.

rousskov: >  I can live with hard-coding if others prefer or you insist on that.

... but for this "provide infrastructure for future scripts" PR to be useful, even with hard-coding, it most likely should still follow the (first) `apply ...` approach sketched above, even if it will apply just one (existing) script. With that approach, there will be no reason to add a dummy script execution -- it would be trivial to add one more `apply ...` line later if/as needed.

kinkie:
There's a number of topics that emerged in the conversation, I'll try to touch them all with my viewpoint in LIFO order:
- project-supporting tools are mostly for our own benefit; as such they offer very little promise. As an example, their interface is often ugly but not worth improving. People who are uncomfortable with their output should generally just not use them
- IMO these do belong to the source repository. Anything that we use to build and improve the project, and that has the slimmest chance of being reuseable, belongs there.
- this is especially true now where the main hurdle is that a large scale change would have to be replicated through some long-running branches
- I generally hope that anything we cook will not cause any major damage; let's not optimize for the unlikely case. I am confident that the time spent in this conversation already exceeds any time needed to fix breakages
- I like the 4-step process, especially if automated

rousskov:
> There's a number of topics that emerged in the conversation, I'll try to touch them all with my viewpoint 

FWIW, I probably disagree with your assertions in the first three bullets, unsure what the unstated implications of the fourth bullet are, and glad we may have an agreement on the basic principles behind the 4-step sketch (i.e. your last bullet). BTW, the same basic principles went into the recent #565.

I believe that further discussion should be moved to IRC or a voive meeting because we are not making (enough) progress here and the disagreements appear to be rather vast. It would be good to narrow down the scope for that meeting agenda though, but we can try without it, especially if it is a voice meeting.

yadij:
To make some progress on this and the other PRs waiting on it I am switching to Alex design.

scripts/source-maintenance.sh
```diff
@@ -66,7 +86,11 @@ for FILENAME in `git ls-files`; do
 	#
 	# Code Style formatting maintenance
 	#
-        if test "${ASVER}"; then
+	for SCRIPT in `git ls-files scripts/maintenance/`; do
+		${SCRIPT} "${FILENAME}" >"${FILENAME}.new" &&
+			updateifchanged "${FILENAME}" "${FILENAME}.new" "by ${SCRIPT}"
+	done
```
rousskov: The approach in this PR does not protect unstaged changes from being wiped out by source-maintenance.sh. We should not accept such risks IMO, especially when we are adding generic "plugin" support. I think we should adopt scripts/spell-check.sh approach: Refuse to adjust any files in a repository with unstaged changes. It is trivial to implement. If that simple approach has problems in source-maintenance.sh context, let's discuss.

scripts/source-maintenance.sh
```diff
@@ -40,6 +40,26 @@ fi
 COPYRIGHT_YEARS=`date +"1996-%Y"`
 echo "s/1996-2[0-9]+ The Squid Software Foundation and contributors/${COPYRIGHT_YEARS} The Squid Software Foundation and contributors/g" >>boilerplate_fix.sed
 
+updateifchanged()
```
rousskov: Nit: Please use camelCase or snake_case for new names.

yadij: done

scripts/source-maintenance.sh
```diff
@@ -40,6 +40,26 @@ fi
...
+{
+original="$1"
```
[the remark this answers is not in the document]
yadij: done

scripts/source-maintenance.sh
```diff
@@ -40,6 +40,26 @@ fi
...
+updated="$2"
+message="$3"
+
+if test -e "${original}" -a -e "${updated}"; then
+	md51=`cat "${original}" | tr -d "\n \t\r" | ${MD5}`;
+	md52=`cat "${updated}" | tr -d "\n \t\r" | ${MD5}`;
+	if test "${md51}" != "${md52}" ; then
+		echo "NOTICE: File ${original} changed: ${message}"
```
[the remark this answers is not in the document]
yadij: IIRC it was because whitespace only changes were quite common when these scripts were added. I am retaining it here for consistency with the existing notices.

scripts/source-maintenance.sh
```diff
@@ -40,6 +40,26 @@ fi
...
+		mv "${updated}" "${original}"
+	else
+		rm -f "${updated}"
+	fi
+else
+	echo "BUG: invalid filename '$1' or '$2'. Skipped update for ${message}."
```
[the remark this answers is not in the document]
yadij: done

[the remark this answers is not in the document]
yadij: As to your question. Two reasons;

* The MD5 tools produce output regardless of whether a file exists. So we get a false-positive detection of change if the .new does not exist.

* There is no point doing all the work of sub-shell creation and MD5 calculation if either file is non-existent. This being the lowest level of post-processing it adds up to a significant portion of the script running time doing useless work.

scripts/source-maintenance.sh
```diff
@@ -40,6 +40,26 @@ fi
...
+original="$1"
+updated="$2"
+message="$3"
+
+if test -e "${original}" -a -e "${updated}"; then
+	md51=`cat "${original}" | tr -d "\n \t\r" | ${MD5}`;
+	md52=`cat "${updated}" | tr -d "\n \t\r" | ${MD5}`;
+	if test "${md51}" != "${md52}" ; then
+		echo "NOTICE: File ${original} changed: ${message}"
+		mv "${updated}" "${original}"
```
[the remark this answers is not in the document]
yadij: The actions in the main script are very modular and error in one place do not preclude following sections from being run. So I will not be adding -e at this time, nor likely for a while.

scripts/source-maintenance.sh
yadij: This request is effectively a "new feature request" for the maintenance script. Outside the scope of this PR.

The source-maintenance.sh as a whole is intentionally capable of reformatting any file in the repository, including staged ones. Specifically so that files can be staged, have the formatting enforcement run to check them before merge.  "Blowing away" a staged file shows up in the resulting diff/patch as clearly as the same action on a committed file - that case has been tested for long ago.

rousskov: I will reorder your paragraphs because there might simply be a misunderstanding here. The added emphasis is mine. I hope that the second set of paragraphs (i.e. "... including staged ones") will clarify things.

> The source-maintenance.sh as a whole is intentionally capable of reformatting _any_ file in the repository

Agreed.

I am not sure whether the following minor aspect is a part of the bigger misunderstanding, but just in case: If there are valid reasons for doing so, I will happily agree to grandfather existing source-maintenance.sh modification avenues, applying the suggested no-unstaged-changes restriction just to _plugins_. The difference between "the restriction applies to all source-maintenance.sh changes" and "the restriction applies to changes by plugins" is not critically important to me for this PR. It would be simpler to apply the restriction globally, but if there are valid reasons for a narrower application, they can be accommodated AFAICT.

> ..., including _staged_ ones. Specifically so that files can be staged, have the formatting enforcement run to check them before merge. "Blowing away" a staged file shows up in the resulting diff/patch as clearly as the same action on a committed file - that case has been tested for long ago.

Exactly! We seem to be in an agreement here! Perhaps you thought that I am asking you to prohibit applying changes to staged files? I am _not_ asking for that, and asking for that would be _wrong_. There is nothing wrong with modifying a staged file. It is a safe operation.

To reiterate this change request: I think we should prohibit script-driven modification of files with _unstaged_ changes, and I am suggesting accomplishing that by applying the restriction to the repository as a whole (as already implemented in the official scripts/spell-check.sh):

```shell
if ! git diff --quiet; then
    echo "There are unstaged changes. This script may modify sources."
    echo "Stage changes to avoid permanent losses when things go bad."
    exit 1
fi
```

> This request is effectively a "new feature request" for the maintenance script. Outside the scope of this PR.

I disagree on both counts: IMO, the requested core functionality is _required_ to properly support "running of arbitrary code" -- a stated PR goal. The new-out-of-scope-feature argument could be valid if we were just polishing the existing script to do better what it already does. This PR adds a new powerful mechanism. I can support the addition of that mechanism if it has certain safety guarantees. Fortunately, adding those guarantees is very simple (see above for a sketch).

BTW, I was just told that avoiding script-changing unstaged files is a popular/common approach in git days. We are not inventing anything new or doing anything very controversial here! It is a natural solution for mitigating a real risk.

scripts/source-maintenance.sh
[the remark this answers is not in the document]
rousskov: Thank you for explaining your rationale. Armed with that understanding, I can advance this change request further: Please replace this detection of non-whitespace changes with detection of any changes (using something like `diff -q $original $updated`). Please see below for justification.

The actual reasons why the existing/official code removes whitespace are very different from "whitespace only changes were quite common [and should be ignored]":

1. Commit 1afc2b9 introduced that `tr|md5` trick in order to detect astyle bugs. You even mentioned the trick in that commit message. Astyle was not supposed to modify anything but the whitespace so our trick worked well. Only whitespace changes are (silently) allowed. Non-whitespace changes are undone (with a warning).
2. Commit 987ca90 reused the trick to warn about poorly ordered `#include`s while not warning about (but still applying!) whitespace-only changes. All changes are applied. Non-whitespace changes generate a warning.

In both cases, applying the proposed PR logic would result in bugs -- each of those special existing cases needs a _different_ treatment. As for future plugins, some of them will likely need a different default treatment as well. In summary, the current PR default cannot be justified by the old code and is unlikely to be correct for at least some future code. It should be changed.

I hope the above detailed explanation clarifies why a simpler "any changes count" default should be used instead. Ideally, the plugin should signal its high-level checking needs somehow, but this change request is _not_ about that advanced feature.

scripts/source-maintenance.sh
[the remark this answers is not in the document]
rousskov: I consider this change request addressed, thank you. IIRC, there is another (more general) change request about quitting on errors where that more general discussion can take place.

As for those "two reasons", I have doubts about them -- we would not get to updateIfChanged at all if the source file did not exist (unstaged changes are prohibited and the plugin would have failed) or if the output file was not created (leading to the short-circuited `plugin && updateIfChanged` command failure) -- but perhaps it is not important to reach an agreement on this interpretation. I am mentioning these doubts here primarily because they are a good illustration why quitting on the first error is a good default for most scripts -- it eliminates or reduces these and many similar concerns/uncertainties/risks. But that discussion is for a different change request...

scripts/source-maintenance.sh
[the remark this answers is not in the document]
rousskov: > The actions in the main script are very modular

I agree.

> and error in one place do not preclude following sections from being run. 

IMO, we should not keep modifying the same file after the first modification has failed. 

Yes, every modification is a separate "module", but that does not make dealing with cascading errors simple enough. We could build a blacklist of failed files to preclude _their_ further modification, but it is much better, overall, to stop on the first failure instead IMO, especially when some files may depend on others.

[rant] Imagine a preprocessor that gives the compiler a known-to-be-broken preprocessed source file (because the problem will not be visible to the compiler) and the compiler that discovers more errors but gives the linker that known-to-be-broken object file just because the linker would be able to successfully link that object into an executable. The resulting executable is completely not what the developer wanted, but it "runs"! In some cases, an experienced developer would quickly find the preprocessor error and correct it, but in many other cases, hours or days will be wasted trying to understand why the newly deployed program version brought the server down... [/rant]

> So I will not be adding -e at this time, nor likely for a while.

I do not think I should insist on adding a global `set -e` at this point, especially if there are valid reasons why it would be difficult to add (none known to me at this point, but they could exist).

However, I do insist that the new code checks for errors and quits on the first one. If you find that the best way to accomplish that is a (local) `set -e`, great, but there may be better solutions out there, and I am always open to finding the best solution.

rousskov:
FYI: I have updated my earlier [review](https://github.com/squid-cache/squid/pull/531#pullrequestreview-381159739), resolving all the addressed change requests. It took a lot of time so I cannot promise a quick return, but please request my review when it is time for another round. Thank you.

scripts/source-maintenance.sh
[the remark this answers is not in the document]
yadij: Okay. Done.

scripts/source-maintenance.sh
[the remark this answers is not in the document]
yadij: IMO the plugin maintenance scripts should never depend on each other. So we should be able to just skip any which goes wrong and nothing will break.

scripts/source-maintenance.sh
yadij: I am doing this, but only because it is not worth arguing over. I stand by it being feature creep on the current PR scope.

yadij:
Changes in b18e1b1 since your last audit should be the last of the issues. If you are happy now please mark for merge.

scripts/source-maintenance.sh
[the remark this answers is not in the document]
rousskov: > IMO the plugin maintenance scripts should never depend on each other. 

I agree. `Never` might be too strong of a word here (there might be some reasonable exceptions), but it is a valid design principle even if we do allow exceptions, so we do not need to agree on the `never` bit as far as this change request is concerned.

> So we should be able to just skip any which goes wrong and nothing will break.

I see no logical connection between the "independence of plugins" principle and the desire to pile modifications on top of modifications that went wrong. "Nothing will break" is clearly false -- something has gone wrong already. The question in this discussion is not whether something breaks (that is a necessary _precondition_) but whether it is a good idea to pile up modifications on top of failures.

The current PR already contains some improvements in this area. In hope to make progress, I am going to resolve this "outdated" change request and focus on error handing of specific commands in the updated script. Perhaps we can find a way to agree on specific modifications (to continue running after some failures) even if we cannot agree on whether the script should quit on the first failure.

scripts/source-maintenance.sh
```diff
@@ -66,7 +89,11 @@ for FILENAME in `git ls-files`; do
 	#
 	# Code Style formatting maintenance
 	#
-        if test "${ASVER}"; then
+	for SCRIPT in `git ls-files scripts/maintenance/`; do
+		${SCRIPT} "${FILENAME}" >"${FILENAME}.new" &&
+			updateIfChanged "${FILENAME}" "${FILENAME}.new" "by ${SCRIPT}"
```
rousskov, line 94: As you know, I would like us to avoid piling modifications on top of failures:

```suggestion
			updateIfChanged "${FILENAME}" "${FILENAME}.new" "by ${SCRIPT}" || return
```

If we cannot agree on that, I suggest making error ignorance configurable (with a conservative default), similar to `make -k`. I can contribute the code necessary to make that happen.

scripts/source-maintenance.sh
```diff
@@ -40,7 +46,24 @@ fi
...
+	message="$3"
+
+	test -e "${original}" -a -e "${updated}" || return
+	md51=`cat "${original}" | ${MD5}`;
+	md52=`cat "${updated}" | ${MD5}`;
+	if test "${md51}" != "${md52}" ; then
+		echo "NOTICE: File ${original} changed: ${message}"
+		mv "${updated}" "${original}" || return
+	else
+		rm -f "${updated}"
```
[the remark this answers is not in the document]
yadij: The case you are worried about is one which indicated a major fault in the filesystem and security permissions. Modifying the suggestion to use exit rather than return.

scripts/source-maintenance.sh
yadij: Accepting that the rm set of changes is fatal to the entire maintenance process now, the "failures" that remain are just minor nits where skipping the script is acceptable.

There should *not* be any dependency between these scripts. Specifically because we cannot predetermine which order they will be run. ie. If script A depends on script B; then they should be combined into one, or A calls B and only A placed in the scripts/maintenance/ directory for auto-execution.

scripts/source-maintenance.sh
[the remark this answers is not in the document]
rousskov: I consider this change request addressed, thank you.

> The case you are worried about is one which indicated a major fault in the filesystem and security permissions.

... or just the same developer running `sudo scripts/source-maintenance.sh` in another window/session. Or a plugin script creating its own output file with the wrong permissions. Etc... 

I have no energy to object to a sudden and partial classification of command errors as "deserving an immediate death inside a low-level function" even though I consider such classification an overreaction and do not like that its implementation leaves the code reader guessing.

scripts/source-maintenance.sh
rousskov: > the "failures" that remain are just minor nits where skipping the script is acceptable

I disagree with that assertion. Applying modifications after any plugin failure, especially after a failure that leaves partial output _may_ make triaging the situation a lot more difficult. It all depends on the specifics of the failure and on whether the failure was even noticed. We cannot reliably predict those things.

I can easily imagine cases where continuing is harmful _and_ cases where continuing is desirable. Moreover, I am willing to contribute proper support for both. Is not resistance to that solution counter-productive?

> There should not be any dependency between these scripts.

Yes, we agree on plugin independence -- no need to remind me of that principle in this context. The principle is a good one (at least for now), but it does not address my concern.

yadij: If you wish to put the work in yourself feel free to do so. I do not think this issue is serious enough to spend any more time on.

Please note that "return" is silent so if there is an actual *problem* happening something else is needed.

rousskov: > If you wish to put the work in yourself feel free to do so.

Done at f50627e.

rousskov:
All of my specific change requests have been addressed, thank you. AFAICT, the remaining steps are:

1. Your review of my f50627e commit. Please adjust the code as needed, of course. I will review those adjustments as usual.

2. We need a reason to add this general plugin mechanism. Currently, I do not see any good applications, but I trust that you already have them. We do _not_ need PRs for them now, just a general agreement that at least one such PR is coming and is likely to be acceptable. I suggest starting by listing a few representative examples that we can evaluate. I fear that either we fail to agree on any plugin acceptability OR that we find acceptable plugins, but they would need a different interface (e.g., all of the existing source modification snippets cannot use the interface this PR adds AFAICT).

scripts/source-maintenance.sh
```diff
@@ -40,7 +72,48 @@ fi
 COPYRIGHT_YEARS=`date +"1996-%Y"`
 echo "s/1996-2[0-9]+ The Squid Software Foundation and contributors/${COPYRIGHT_YEARS} The Squid Software Foundation and contributors/g" >>boilerplate_fix.sed
 
-srcformat ()
+# executes the specified command
+# in KeepGoing mode, remembers errors and hides them from callers
+run_()
```
yadij, line 77: Some shells require a space before these brackets.
```suggestion
run_ ()
```

The underscore may also cause problems as it is a reserved character.

yadij:
As for #2 the reason is that you insisted we add arbitrary code support instead of just declarative AWK rules (https://github.com/squid-cache/squid/pull/531#discussion_r362334698).

scripts/source-maintenance.sh
rousskov: I applied the suggestion (9a6ad2d3d2dda82a4bb5cce1c1b72317d4ee60d5).

FWIW, I did not find a confirmation regarding underscore being prohibited in function names (POSIX [allows](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_05) [it](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tag_06_01)). I added that underscore to emphasize that this is an internal helper function and to minimize chances of clashes with programs and built-ins without using long names. Please feel free to rename, of course.

rousskov:
> the reason is that you insisted we add arbitrary code support instead of just declarative AWK rules

That was the reason you have implemented the general plugin support, yes, but that is not a valid justification for committing that support: Just because B is better than A does not mean we should commit B. There needs to be a good reason for committing either. I do not think we need to agree on this point though -- if you have a plan to add at least one or, better, two specific plugins, then the principle agreement on _that_ plan is sufficient to move forward with merging this PR. I am just clarifying the logic behind my request for the principal agreement on plugins here; I am _not_ trying to continue this part of the argument!

> There are at least two PRs coming to add scripts to this plugin directory. One that moves the sort-includes script there, and one that adds nullptr enforcement.

If you manage to move the existing "includes sorting" script into plugins directory, it would be sufficient to justify adding this general mechanism. When I looked at that script, it was impossible to move it without changing its functionality:

1. The current plugin support cannot warn about changing includes order like the existing sorting script does.
2.The current plugin support will warn about whitespace-only changes unlike the existing sorting script (that does not warn about whitespace-only changes).

Are you going to drop and add those warnings during the move? Or are you going to enhance the plugin support so that it can differentiate between whitespace-only and non-whitespace changes?

As for nullptr enforcement, are you going to exclude C headers and sources from being subject to these plugins? There are currently 600+ NULLs in C sources alone. AFAICT, C does not have `nullptr`.

yadij:
> > the reason is that you insisted we add arbitrary code support instead of just declarative AWK rules
> 
> That was the reason you have implemented the general plugin support, yes, but that is not a valid justification for committing that support: Just because B is better than A does not mean we should commit B. There needs to be a good reason for committing either. I do not think we need to agree on this point though -- if you have a plan to add at least one or, better, two specific plugins, then the principle agreement on _that_ plan is sufficient to move forward with merging this PR. 


I am not sure I understand what you were asking for. If there is something you want to be added to the description, please state clearly what you are wanting.


> > There are at least two PRs coming to add scripts to this plugin directory. One that moves the sort-includes script there, and one that adds nullptr enforcement.
> 
> If you manage to move the existing "includes sorting" script into plugins directory, it would be sufficient to justify adding this general mechanism. When I looked at that script, it was impossible to move it without changing its functionality:
> 

The sort-includes.pl script and how it is called is what this PR was modeled on. It can be moved as-is into the scripts/maintenance/ directory. All the audit requested changes to this PR apply equally to how that script was being called - including the arguments for warning on whitespace change, diff replacement of MD5, and not touching unstaged files.

> 
> As for nullptr enforcement, are you going to exclude C headers and sources from being subject to these plugins? There are currently 600+ NULLs in C sources alone. AFAICT, C does not have `nullptr`.

Of course. Only C++ files will have the nullptr enforcement applied. I will leave more details on that to the relevant followup PRs.

rousskov:
> Alex: When I looked at that script, it was impossible to move it without changing its functionality: ...

> Amos: All the audit requested changes to this PR apply equally to how that script was being called

AFAICT, you imply that you expect that a future PR will change how includes are sorted in order to migrate sorting to the new plugin interface. That change is OK with me; I wanted to make sure that you realize that the existing sorting code is not compatible with the new plugin interface.

I would prefer to have an agreement on multiple anticipated plugin examples, but one example is sufficient to justify merging this PR, so I am resolving my last change request.

Please note that there is currently no agreement regarding NULL fixing, the current plugin code applies to C files, and the current plugin API makes it difficult for a plugin to distinguish C from C++ sources. I agree that we can discuss these problems separately. Hopefully, we will find a way to reach an agreement there.

scripts/source-maintenance.sh
```diff
@@ -21,6 +21,38 @@
...
+SeenErrors=0
+
+
+if ! git diff --quiet; then
+	echo "There are unstaged changes. This script may modify sources."
+	echo "Stage changes to avoid permanent losses when things go bad."
+	exit 1
+fi
+
 # On squid-cache.org we have to use the python scripted md5sum
```
kinkie, line 56: This whole squid-cache.org block is rotten and is by now unnecessary. In fact, it hasn't worked for years.

scripts/source-maintenance.sh
```diff
@@ -66,7 +139,10 @@ for FILENAME in `git ls-files`; do
 	#
 	# Code Style formatting maintenance
 	#
-        if test "${ASVER}"; then
+	for SCRIPT in `git ls-files scripts/maintenance/`; do
```
kinkie, line 142: Nit: why git ls-files and not just listing the files?

rousskov: This may not be a complete or even valid answer to your question, but:

* `git ls-files` ignores files ignored by git. This is very helpful when editors and such create helper/temporary files inside our directory.
* AFAICT, `git ls-files` has better (than `ls -R`) support for listing files in _sub_-directories. We might group scripts in sub-directories of the `scripts/maintenance` directory.

scripts/source-maintenance.sh
yadij: I know.  It is out of scope though for this feature change.

scripts/source-maintenance.sh
yadij: Pretty much what Alex said plus the benefit that it ignores objects which are not in the repository. So unintentional files are not going to be arbitrarily executed.

yadij:
Was the author tag just supposed to indicated this was waiting on my reply to kinkie? or something else?

rousskov:
> Was the author tag just supposed to indicated this was waiting on my reply to kinkie? or something else?

The flag indicates that the ball is on your side. IIRC, when I set this flag, this PR was awaiting your response to Francesco and a clearance for merging. I will reset it to `S-waiting-for-committer` now.

PR #536, opened 2020-01-03
Fix Squid V4-Bug 4992

ssg-squid:
mgr:objects response stalls fix.
Missing "state->sentry->flush();" statement at line no 355 of src/stat.cc, was causing halt state in mgr:objects command.


rousskov:
I polished PR title and description to follow Squid Project conventions. This adjustment should remove the M-failed-description label.

PR #539, opened 2020-01-09
Bug 5016: systemd thinks Squid is ready before Squid listens

marcosfrm:
Use systemd API to send start-up completion notification if built
with libsystemd support. New configure option --with-systemd
can be used to force enable or disable the feature (default:
auto-detect on Linux platforms).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2102,6 +2102,51 @@ esac
 AC_SUBST(LDAPLIB)
 AC_SUBST(LBERLIB)
 
+AC_ARG_WITH(systemd,
+  AS_HELP_STRING([--without-systemd],
+                 [Do not use systemd API to send start-up completion
+                  notification. Default: auto-detect]), [
+case "$with_systemd" in
+  yes|no)
+    : # Nothing special to do here
+    ;;
+  *)
+    if test ! -d "$withval" ; then
+      AC_MSG_ERROR([--with-systemd path does not point to a directory])
+    fi
+    SYSTEMD_PATH="-L$with_systemd/lib"
+    CPPFLAGS="-I$with_systemd/include $CPPFLAGS"
+  esac
+])
+AH_TEMPLATE(USE_SYSTEMD,[systemd support is available])
+if test "x$with_systemd" != "xno" -a "x$squid_host_os" = "xlinux"; then
+  SQUID_STATE_SAVE(squid_systemd_state)
+
+  # User may have provided a custom location for systemd. Otherwise...
+  LIBS="$LIBS $SYSTEMD_PATH"
+
+  # auto-detect using pkg-config
+  PKG_CHECK_MODULES(SYSTEMD,[libsystemd],,[
+    # systemd < 209
+    PKG_CHECK_MODULES(SYSTEMD,[libsystemd-daemon],,[:])
+  ])
+
+  AC_CHECK_HEADERS(systemd/sd-daemon.h)
+
+  SQUID_STATE_ROLLBACK(squid_systemd_state) #de-pollute LIBS
+
+  if test "x$with_systemd" = "xyes" -a "x$SYSTEMD_LIBS" = "x"; then
+    AC_MSG_ERROR([Required systemd library not found])
+  fi
+  if test "x$SYSTEMD_LIBS" != "x" ; then
+    CXXFLAGS="$SYSTEMD_CFLAGS $CXXFLAGS"
+    AC_DEFINE(USE_SYSTEMD,1,[systemd support is available])
+  else
+    with_systemd=no
+  fi
+fi
+AC_MSG_NOTICE([systemd library support: ${with_systemd:=auto} ${SYSTEMD_PATH} ${SYSTEMD_LIBS}])
+
 AC_ARG_ENABLE(forw-via-db,
   AS_HELP_STRING([--enable-forw-via-db],[Enable Forw/Via database]), [
   SQUID_YESNO([$enableval],[unrecognized argument to --enable-forw-via-db: $enableval])
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -3488,8 +3488,13 @@ clientListenerConnectionOpened(AnyP::PortCfgPointer &s, const Ipc::FdNoteId port
     // Subsequent sd_notify() calls, including calls during reconfiguration,
     // do nothing because the first call parameter is 1.
     // XXX: Send the notification only after opening all configured ports.
-    if ((opt_foreground || opt_no_daemon) && (sd_notify(1, "READY=1") < 0))
-        debugs(1, DBG_IMPORTANT, "WARNING: failed to send start-up notification to systemd");
+    if (opt_foreground || opt_no_daemon) {
+        const auto result = sd_notify(1, "READY=1");
+        if (result < 0) {
+            debugs(1, DBG_IMPORTANT, "WARNING: failed to send start-up notification to systemd" <<
+                   Debug::Extra << "sd_notify() error: " << xstrerr(-result));
+        }
+    }
 #endif
 }
 
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -139,6 +139,10 @@
 #include <cmath>
 #include <limits>
 
+#if HAVE_SYSTEMD_SD_DAEMON_H
+#include <systemd/sd-daemon.h>
+#endif
+
 #if LINGERING_CLOSE
 #define comm_close comm_lingering_close
 #endif
@@ -3478,6 +3482,15 @@ clientListenerConnectionOpened(AnyP::PortCfgPointer &s, const Ipc::FdNoteId port
            << s->listenConn);
 
     Must(AddOpenedHttpSocket(s->listenConn)); // otherwise, we have received a fd we did not ask for
+
+#if USE_SYSTEMD
+    // When the very first port opens, tell systemd we are able to serve connections.
+    // Subsequent sd_notify() calls, including calls during reconfiguration,
+    // do nothing because the first call parameter is 1.
+    // XXX: Send the notification only after opening all configured ports.
+    if ((opt_foreground || opt_no_daemon) && (sd_notify(1, "READY=1") < 0))
+        debugs(1, DBG_IMPORTANT, "WARNING: failed to send start-up notification to systemd");
+#endif
 }
 
 void
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -581,6 +581,7 @@ squid_LDADD = \
 	$(EPOLL_LIBS) \
 	$(MINGW_LIBS) \
 	$(KRB5LIBS) \
+	$(SYSTEMD_LIBS) \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 
--- a/tools/systemd/squid.service
+++ b/tools/systemd/squid.service
@@ -11,12 +11,13 @@ Documentation=man:squid(8)
 After=network.target network-online.target nss-lookup.target
 
 [Service]
-Type=forking
+Type=notify
 PIDFile=/var/run/squid.pid
 ExecStartPre=/usr/sbin/squid --foreground -z
-ExecStart=/usr/sbin/squid -sYC
+ExecStart=/usr/sbin/squid --foreground -sYC
 ExecReload=/bin/kill -HUP $MAINPID
 KillMode=mixed
+NotifyAccess=all
 
 [Install]
 WantedBy=multi-user.target


marcosfrm:
Since we are only interested in start-up completion, to avoid needless work by libsystemd (in case of multiple ports), I have set `sd_notify()`'s first parameter to `1`. After the first call it will return early doing nothing.

src/client_side.cc
```diff
@@ -3478,6 +3482,14 @@ clientListenerConnectionOpened(AnyP::PortCfgPointer &s, const Ipc::FdNoteId port
            << s->listenConn);
 
     Must(AddOpenedHttpSocket(s->listenConn)); // otherwise, we have received a fd we did not ask for
+
+#if USE_SYSTEMD
+    // at least one port is open, tell systemd we are able to serve connections
+    // subsequent sd_notify() calls will not be processed because first parameter is non-zero
+    // XXX: send the notification only when all configured ports are open
+    if (opt_foreground || opt_no_daemon)
```
yadij: This code path is shared by startup and reconfigure. A local static counter would avoid sending a burst on reconfigure as well.
```suggestion
    static int portsSeen = 0;
    if (++portsSeen == NHttpSockets && (opt_foreground || opt_no_daemon))
```

NP: this does not resolve the XXX because NHttpSockets only points at the number of sockets that started the opening process. The two counters may match on every port, or at a subset of points during startup. This will be resolved when HttpPortList becomes a proper STL container we can check length() on.

yadij:
FYI: I am not marking this as testable by prbot to reduce builds before the configure.ac is settled.

rousskov:
IMHO, _if it is not difficult_, we should enable this notification support (by default) if libsystemd is available (instead of requiring an explicit ./configure option). This is not my area of expertise, but "enabling systemd integration when possible" sounds like a natural default to me, and Squid, in general, tends to enable the features it can enable (rather than requiring an explicit ./configure option for every feature).

yadij:
@roussokv, the configure.ac changes I requested include that.

marcosfrm:
systemd provides `.pc` files since day one. Can I realy on `PKG_CHECK_MODULES` only and make `--with-systemd` a boolean?

marcosfrm:
Copy and paste from gnutls configure section. Please take a look.

rousskov:
> systemd provides .pc files since day one. Can I rely on [it]

IMO, you should! I was actually going to add such a comment when I saw your PR, but decided to wait until that part of the PR becomes more stable :-). Please note that I do not know/remember for sure what "relying on it" means exactly (the part of your question that I removed in my quote). I am only commenting on the "require and rely on systemd package config files" idea itself.

configure.ac
```diff
@@ -2102,6 +2102,51 @@ esac
...
+])
+AH_TEMPLATE(USE_SYSTEMD,[systemd support is available])
+if test "x$with_systemd" != "xno" -a "x$squid_host_os" = "xlinux"; then
+  SQUID_STATE_SAVE(squid_systemd_state)
+
+  # User may have provided a custom location for systemd. Otherwise...
+  LIBS="$LIBS $SYSTEMD_PATH"
+
+  # auto-detect using pkg-config
+  PKG_CHECK_MODULES(SYSTEMD,[libsystemd],,[
```
yadij: Does checking for [libsystemd >= 209] work?  Just checking for [libsystemd] will provide a match against older versions so long as the named library exists.

marcosfrm: Yes, `[libsystemd >= 209]` works. The library changed name in systemd 209. Perhaps require >= 209 is good idea, because squid.service has `KillMode=mixed`, only available starting with this version.

yadij: Okay, if there is definitely never a libsystemd with version prior to 209 then both criteria are met by the check you have now.

marcosfrm:
> hough you may want to remove the <209 version support if they cannot cope with the squid.service file contents we already have.

squid.service will work with systemd < 209. Their policy is when an option is not recognized ignore it and use the defaults: `KillMode=control-group`. It means *all* Squid processes will receive SIGTERM at shutdown, not only the main pid.

marcosfrm:
First message is the commit message, right? Can I update it (--with-systemd now)? What about documentation?

rousskov:
> First message is the commit message, right?

PR title/[description](https://github.com/squid-cache/squid/pull/539#issue-361208617) is the commit [message](https://github.com/measurement-factory/anubis#commit-message).

> Can I update it (--with-systemd now)?

You should. In general, it is a good idea to keep PR title/description up to date. Reviewers may forget to check that metadata.

> What about documentation?

This PR should update any relevant in-repository documentation. Are you aware of any missing updates? Or are you talking about some external-to-repository documentation like Squid wiki? The latter can be updated after the feature is officially committed (to minimize updates).

src/client_side.cc
```diff
@@ -3478,6 +3482,15 @@ clientListenerConnectionOpened(AnyP::PortCfgPointer &s, const Ipc::FdNoteId port
            << s->listenConn);
 
     Must(AddOpenedHttpSocket(s->listenConn)); // otherwise, we have received a fd we did not ask for
+
+#if USE_SYSTEMD
+    // at least one port is open, tell systemd we are able to serve connections
+    // subsequent sd_notify() calls will not be processed because first parameter is non-zero
+    // XXX: send the notification only when all configured ports are open
+    static int portsSeen = 0;
+    if (++portsSeen == NHttpSockets && (opt_foreground || opt_no_daemon))
```
rousskov: The `++portsSeen == NHttpSockets` condition is very complex. It is unlikely that somebody will understand what it really means or why it is here without spending significant time on researching the answer. Can we simply remove it?

```suggestion
    // When the very first port opens, tell systemd we are able to serve connections.
    // Subsequent sd_notify() calls, including calls during reconfiguration,
    // do nothing because the first call parameter is 1.
    // XXX: Send the notification only after opening all configured ports.
    if (opt_foreground || opt_no_daemon)
```

src/client_side.cc
```diff
@@ -3478,6 +3482,15 @@ clientListenerConnectionOpened(AnyP::PortCfgPointer &s, const Ipc::FdNoteId port
...
+        (void)sd_notify(1, "READY=1");
```
rousskov: Please report sd_notify() failures via debugs(). I am not sure what the right level is -- either `DBG_IMPORTANT` or `2`. If it is easy to do, make the level dependent on whether Squid was started from systemd.

src/client_side.cc
rousskov: This is not my area of expertise so please feel free to dismiss the following question if the answer is obvious to an expert: Why do we need to exclude "daemonized" Squid instances? In other words, why do we need the `opt_foreground || opt_no_daemon` check here? I understand that the `opt_foreground || opt_no_daemon` check kind of matches the suggested `ExecStart` command, but I am wondering whether it is a good idea to allow _other_ ExecStart commands as well. For example, I am thinking about a custom ExecStart line that starts (in foreground, as usual) a custom script that, in turn, starts Squid (normally, i.e. in the background). AFAICT, _without_ the `opt_foreground || opt_no_daemon` check, this custom setup will work fine because Squid would still notify the grand-grand-parent systemd process.

configure.ac
```diff
@@ -2102,6 +2102,51 @@ esac
...
+  *)
+    if test ! -d "$withval" ; then
+      AC_MSG_ERROR([--with-systemd path does not point to a directory])
+    fi
+    SYSTEMD_PATH="-L$with_systemd/lib"
+    CPPFLAGS="-I$with_systemd/include $CPPFLAGS"
+  esac
+])
+AH_TEMPLATE(USE_SYSTEMD,[systemd support is available])
+if test "x$with_systemd" != "xno" -a "x$squid_host_os" = "xlinux"; then
```
rousskov: This is not my area of expertise so please feel free to dismiss the following question if the answer is obvious to an expert: Why do we need to exclude non-Linux environments? In other words, should not the libsystemd check drive the decision here rather than the spelling of the host OS name? I understand that systemd may be specific to Linux-like OSes (today), but that does not seem like a good enough reason -- many OS features are or were OS-specific, but if we can directly test the feature support itself, should not we?

src/client_side.cc
marcosfrm: It was Amos suggestion. I can remove it if you both agree.

src/client_side.cc
marcosfrm: OK.

src/client_side.cc
marcosfrm: > This is not my area of expertise so please feel free to dismiss the following question if the answer is obvious to an expert: Why do we need to exclude "daemonized" Squid instances?

Type=notify cannot work with binaries that fork()/exit() at startup; once the parent call exit(), it is considered terminated.

> For example, I am thinking about a custom ExecStart line that starts (in foreground, as usual) a custom script that, in turn, starts Squid (normally, i.e. in the background). AFAICT, _without_ the `opt_foreground || opt_no_daemon` check, this custom setup will work fine because Squid would still notify the grand-grand-parent systemd process.

The problem is process monitoring by systemd.

I gave it a shot and, indeed, it works, only because PIDFile= is set.

But there is a race: if the shell process exits *before* daemonized Squid writes the pid file, it will fail.

configure.ac
marcosfrm: No strong reason. To not perturb BSD folks with (for now) useless libraries checks? hehe

hno:
Squid sytemd service should imho always use -N option and let systemd
manage the process session. There is no reason to daemonise itself when
running from systemd.

marcosfrm:
> According to the Semaphore [test](https://semaphoreci.com/squid/squid/branches/pull-request-539/builds/2), the "enable by default if and only if enabling works in ./configure environment" idea was not implemented correctly -- the build without ./configure options fails when there is no libsystemd-daemon package.
> 
> ```
> checking for SYSTEMD... no
> checking for SYSTEMD... no
> configure: error: Package requirements (libsystemd-daemon) were not met:
> 
> No package 'libsystemd-daemon' found
> ```
> 
> Please let me know if I misinterpreted that test failure.
> 
> Please also see a few inlined change requests/questions.

Missing second fallback:

```autoconf
  # auto-detect using pkg-config
  PKG_CHECK_MODULES(SYSTEMD,[libsystemd],,[
    # systemd < 209
    PKG_CHECK_MODULES(SYSTEMD,[libsystemd-daemon],,[:])
  ])
```

marcosfrm:
> Squid sytemd service should imho always use -N option and let systemd manage the process session. There is no reason to daemonise itself when running from systemd.

We are using `--foreground`. `-N` works too, but SMP features are disabled.

src/client_side.cc
marcosfrm: Scratch that. If the shell process does not exit, the main pid will be the shell, breaking everything (signal delivery, shutdown). This kind of setup is very fragile.

rousskov:
> We are using --foreground. -N works too, but SMP features are disabled.

Yes, @hno knowledge about Squid invocation is a bit outdated :-): `--foreground` is the right way to avoid daemonization these days. `-N` is for debugging/etc.

src/client_side.cc
rousskov: > I can remove it if you both agree.

Sounds good -- you are in charge of making that agreement happen. My current arguments are reflected in the change request. FWIW, I will withdraw my change request if it is wrong, of course.

rousskov: Update: I found two more problems with the `++portsSeen == NHttpSockets` condition:

First of all, Squid does not call sd_notify() at all if the last of the configured ports fails to open (but the first one(s) have opened). This was one of the problems I warned about in my [item 2](https://bugs.squid-cache.org/show_bug.cgi?id=5016#c6)  -- correctly tracking when the last port that could open was opened "requires a bit of development to track the number of ports that failed to open". The simpler first item on that list is what would happen if you remove the portsSeen condition.

```
2020/01/13 14:57:36.188 kid1| Accepting HTTP Socket connections at conn3 local=[::]:3131 remote=[::] FD 29 flags=1
2020/01/13 14:57:36.188 kid1| XXX: opt_foreground=1 opt_no_daemon=0 portsSeen=1 NHttpSockets=2
...
2020/01/13 14:57:36.191 kid4| commBind Cannot bind socket FD 14 to [::]:22: (98) Address already in use
```

I added the `XXX` debugging before the sd_notify() call to illustrate that portsSeen does not reach NHttpSockets=2 when the last port fails to open. In this specific test, two listening ports were configured, and the second port (in the squid.conf order) was already occupied at Squid start time. Squid pretends to run fine in this case, so systemd has to be notified (my footnote [1](https://bugs.squid-cache.org/show_bug.cgi?id=5016#c6) talks more about that pretending).

Second, the condition does not match its `at least one port is open, tell systemd` documentation. The condition only becomes true only if _all_ configured ports opened successfully, not if at least one port was opened.

rousskov: I appreciate you investigating/explaining this issue and am marking this specific change request as addressed. Thank you.

FWIW, it feels wrong to disable notifications because simple Squid wrappers are too fragile to benefit from them. Other, more complex, wrappers could still benefit from such notifications AFAICT. However, since this is not my expertise area, and you are already spending a lot of time on critical PR changes, I am OK with ignoring this minor concern until somebody actually needs notifications from a daemonized Squid.

configure.ac
rousskov: OK, I would remove this OS spelling check in this case, but it is your call. I am marking this change request as addressed.

marcosfrm:
New version.

src/client_side.cc
rousskov: Thank you for fixing this in d5f4a31. I am resolving my change request.

src/client_side.cc
rousskov: Thank you for adding debugs(). I took the liberty to add reporting of the actual error in 069d4ff. I am resolving this change request.

rousskov:
@marcosfrm, the build failure on platforms lacking libsystemd has been resolved in d5f4a31, thank you!

```
checking for SYSTEMD... no
checking for SYSTEMD... no
checking systemd/sd-daemon.h usability... no
checking systemd/sd-daemon.h presence... no
checking for systemd/sd-daemon.h... no
configure: systemd library support: no  
```

The current ./configure output on those platforms is not ideal (repeated "checking for SYSTEMD" lines and probably no distinction between explicit `--without-systemd` and "auto" mode on the `systemd library support` line), but it is usable, and, I am guessing, improving it further is difficult.

It would be nice to use level-2 messages for sd_notify() failures when _not_ executed under systemd control, but I am guessing it is hard to detect the latter condition. I was hoping that checking for something like a set `INVOCATION_ID` environment variable would work, but this is not my area of expertise.

@yadij, this PR is back to your side. If you are happy with the current code after the latest changes, please clear it for merging.

rousskov:
OK to test

PR #540, opened 2020-01-10
Add %>connection_id logformat code

chtsanti:
Knowing client connection ID can be very helpful in triage, especially
when dealing with multiple requests associated with a problematic
connection (e.g., two access log records associated with a single TLS
connection bumping failure).


chtsanti:
Incomplete implementation.
I am closing and I will re-open when it is ready.

PR #541, opened 2020-01-14
Log the number of request sending attempts

eduard-bagdasaryan:
Squid retries certain request forwarding failures. When the problem is
persistent, and there are many peers (or the origin server has many
IPs), many retries are possible. Absence of Squid-server transaction log
hides these retries from admins. The new logformat code helps discover
and triage problems, especially when Squid retries many times.


src/HappyConnOpener.h
```diff
@@ -195,6 +191,11 @@ class HappyConnOpener: public AsyncJob
     void sendSuccess(const Comm::ConnectionPointer &conn, bool reused, const char *connKind);
     void sendFailure();
 
+    /// the total number of connection opening attempts
+    int tries() const { return ale ? ale->requestAttempts : 0; }
+    /// increments the connection opening attempts
+    void incTries() { if (ale) ale->requestAttempts++; }
```
eduard-bagdasaryan: We could simplify by removing the 'ale' check (assuming that ALE is always non-nil) but HappyConnOpener does not have this requirement. This decision depends on whether both Tunnel and FwdState callers pass a non-nil ALE here, which is not clear because they use ALE inconsistently - sometimes checking for its presence and sometimes not.

src/FwdState.cc
```diff
@@ -1290,7 +1288,7 @@ FwdState::logReplyStatus(int tries, const Http::StatusCode status)
 bool
 FwdState::exhaustedTries() const
 {
-    return n_tries >= Config.forward_max_tries;
+    return tries() >= Config.forward_max_tries;
```
rousskov: This change affects reforwarding logic AFAICT. Consider a situation where the destination has two IP addresses, IPv4 and IPv6. Let's assume forward_max_tries is set to 2. The first address is slow to respond, so Happy Eyeballs makes two TCP connection establishment attempts. Let's say TCP/IPv4 connection wins after all, and Squid sends a request to the IPv4 address. If the server response needs to be reforwarded, FwdState will check exhaustedTries() and

* Before this change: FwdState reforwards the request to the TCP/IPv6 connection, right?
* After this change: FwdState does not reforward the request.

Are there any flaws in the above analysis? If there are not, what do you propose as a fix?

eduard-bagdasaryan: > Before this change: FwdState reforwards the request to the TCP/IPv6 connection, right?

No, I think that it won't reforward. If HappyConnOpener makes two connection attempts (as in your example), HappyConnOpener::n_tries becomes 2 and then passed over to FwdState::n_tries. If, after that, reforward is required, FwdState::exhaustedTries() returns false and no reforward is performed.
Basically, the changes introduced in this PR only substitute one counter with another (i.e, HappyConnOpener::n_tries with AccessLogEntry::requestAttempts), while leaving the forwarding behavior unchanged.

rousskov: Aha, so we are essentially dealing with buggy official code. The changes in this PR do not introduce the bug we discovered. The question is: Are they making the existing bug _worse_ by making the future fix harder?

I think there are at least answers to that question:

1. "yes" or "maybe": These changes essentially equate connecting attempts with forwarding attempts when, in fact, the two classes of attempts are distinct. The distinction is in increment _timing_ -- the second counter value is bounded by the first one, but the second counter is incremented _later_, and that delay is meaningful rather than accidental. FwdState reforwarding checks should use the second, delayed counter. We should not merge the two counters together.
2. "no, but this PR should remove the broken FwdState (and tunnel.cc) checks": One ALE-based counter is fine if FwdState relies on HappyConnOpener to observe the limit. If HappyConnOpener gave FwdState a connection, then the limit has already been observed (if not, then this is a HappyConnOpener bug unrelated to this PR).

The second answer would make everything simpler. Please check whether it is possible to fully outsource forward_max_tries checks to HappyConnOpener? In other words, are there some fundamental reasons why FwdState (and tunnel.cc) must maintain their own attempts counter (or equivalent)? For example, are there any connections that FwdState does not receive from HappyConnOpener _and_ that are subject to the forward_max_tries limit?

eduard-bagdasaryan: > are there any connections that FwdState does not receive from HappyConnOpener 

'pinned' connections are separate and FwdState::usePinned() increments the counter.

rousskov: Yes, but, IIRC, there can be at most one PINNED connection, and PINNED destination(s) are tried first, before everything else. If that recollection is correct, and assuming forward_max_tries cannot be legitimately used to deny the very first forwarding attempt, then PINNED connections do not need forward_max_tries checks.

eduard-bagdasaryan: > PINNED destination(s) are tried first, before everything else

It looks like that is true. However, failed pinned connections may be retried (FwdState::pinnedCanRetry()) and we need to check for forward_max_tries before picking another destination. We could do this probably in HappyConnOpener before attempting to connect (e.g., somewhere in HappyConnOpener::maybeOpenAnotherPrimeConnection()) but this looks awkward - in this case, we do not need to create HappyConnOpener at all.

rousskov: I am not sure which solution will be less awkward, but I think the discussion about the location of the missing check is a bit premature/misplaced. AFAICT, we agree that HappyConnOpener (and/or surrounding code) has a bug -- it may reach the forward_max_tries limit but then hide the last (spare) connection from the requestor, effectively depriving the requestor from the last connection attempt.

If we agree on that, then I think that bug should be fixed in a separate PR. This PR should assume that the bug is fixed. We can either pause this PR (until that other bug is fixed) or, with that assumption in mind, adjust (remove the duplicated check from FwdState/tunnel.cc) and merge this PR. If the latter does not make things worse, then I would prefer it because it allows us to make progress _and_ may make fixing the bug easier. What do you think?

eduard-bagdasaryan: There is another counter for the same purpose: PeerSelector::foundPaths. It is incremented in PeerSelector::handlePath() in both pinned and regular cases and checked against forward_max_tries in PeerSelector::wantsMoreDestinations(). It looks like this check should be enough to cover forward_max_tries condition since it is performed earlier than similar/equivalent checks in FwdState and HappyConnOpener. In other words, if foundPaths reached its maximum, there are no new destinations and HappyConnOpener should not bother about it. If so, we can remove forward_max_tries checks from all places, except PeerSelector. Meanwhile, in this case, we can go on with this PR: AccessLogEntry::requestAttempts is merely for attempts counting, without any validation purpose.

rousskov: PeerSelector::foundPaths is not sufficient for forward_max_tries enforcement because FwdState (and, in the future, tunnel.cc) may retry the same found path twice. Look for the retryPath() call. You are right that PeerSelector::foundPaths should not exceed forward_max_tries, but, due to pconn races, FwdState may reach the attempts limit even before trying all the found paths.

rousskov: I am still not sure whether a single ALE-based counter is what we need long-term. I would not be surprised if fixing the lost destination bug discovered during this PR [review](https://github.com/squid-cache/squid/pull/541#discussion_r366872547) requires splitting that counter somehow. I recommend pausing this PR until we know what that fix looks like. I am marking this PR as "WIP" for now.

eduard-bagdasaryan:
Since there are not open change requests here - I think we shoud remove S-waiting-for-author until we discuss/agree on possible solution(s).

rousskov:
> Since there are not open change requests here - I think we shoud remove S-waiting-for-author until we discuss/agree on possible solution(s).

Sure, removing. We do not have a label that can be used to mark PRs waiting for other PRs (an increasingly common case) or other development work in general (this PR, for example, does not have the corresponding PR yet IIRC). Perhaps we should add `S-waiting-for-prerequisite`? Or an even more general `S-waiting-for-other` (with a risk that it will be abused instead of the more specific labels like `S-waiting-for-distro` and `S-waiting-for-reviewer`)?

eduard-bagdasaryan:
> Perhaps we should add S-waiting-for-prerequisite? Or an even more general S-waiting-for-other

IMO both 'prerequisite' and 'other' are quite general - e.g., 'reviewer' can fall into these categories. I would use something more specific, such as `S-waiting-for-other-pr`, assuming that the 'other-pr' may either exist already or be posted in a foreseeable future.

rousskov:
> I would use something more specific, such as `S-waiting-for-other-pr`, assuming that the 'other-pr' may either exist already or be posted in the foreseeable future.

I like your "future PR" excuse! I plan to add `S-waiting-for-pr` and use it here[^1]. CCing @squid-cache/commiters for any objections or better ideas.

[^1]:  The `-other-` part would be pretty much noise in this context IMO. Obviously, we would not label a PR because it waits for itself.

rousskov:
> I plan to add `S-waiting-for-PR` and use it here. CCing @squid-cache/commiters for any objections or better ideas.

Done.

src/FwdState.cc
rousskov: >  I recommend pausing this PR until we know what that fix looks like.

The proposed fix is now posted as #1093. This PR should still wait (now for that PR landing).

rousskov: _Status update_: PR 1093 was merged. We need to resume this PR work. @eduard-bagdasaryan, if you think that we should resolve current merge conflicts before finalizing the fix design or if they are easy to resolve, please resolve them.

src/HappyConnOpener.h
rousskov: You are right. While TunnelStateData always has ALE, FwdState is not guaranteed to have it now and did not have it in some past cases (e.g., see commit 1ce66e2).

src/FwdState.cc
rousskov: I am resolving this old thread. The "making future bug fix hard" concern I raised here was addressed by PR 1093 (which fixed the bug this discussion exposed).

rousskov (approved):
@eduard-bagdasaryan, I merged recent master changes into this old PR branch, polished PR description and text/code a little, and then discovered a design problem. I addressed that problem in branch commit 88b7f62. Please review the latest state of this PR branch (i.e. commit 620a3ba).

rousskov:
@yadij, just FYI: I think this old PR is ready to go in. I plan to clear it for merging on 2023-01-05 or so, unless somebody stops me.

PR #543, opened 2020-01-14
Preserve caller context across Store data delivery

rousskov:
StoreEntry::invokeHandlers() sends a recently loaded response data to
the waiting store_clients. During concurrent cache hits (including, but
not limited to collapsed ones), the response can be loaded into Store by
one transaction and delivered to several different transactions (i.e.
store_clients). This Store "hit sharing service" must restore the
context of the transactions it serves.

### DIFF
--- a/src/StoreClient.h
+++ b/src/StoreClient.h
@@ -9,6 +9,7 @@
 #ifndef SQUID_STORECLIENT_H
 #define SQUID_STORECLIENT_H
 
+#include "base/forward.h"
 #include "dlink.h"
 #include "StoreIOBuffer.h"
 #include "StoreIOState.h"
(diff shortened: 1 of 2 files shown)


rousskov:
@eduard-bagdasaryan, please check why Anubis marked this with `M-failed-other`.

rousskov:
> @eduard-bagdasaryan, please check why Anubis marked this with M-failed-other. And I suspect there are other PRs in this failed state. Perhaps this is related to the buildmaster migration.

For the record: Eduard has triaged the problem. To replicate PR test statuses in the merge commit test statuses, Anubis requires a certain field in the GitHub response. Jenkins does not provide that field (to GitHub) until the corresponding Jenkins test _starts_. The problem existed before the buildmaster migration, but Jenkins tests were running faster (and/or there were fewer tests), so the problem was rarely triggered. We will now teach Anubis to disregard the missing field when (or until) copying Jenkins test statuses.

PR #546, opened 2020-01-16
Add %transport::>connection_id logformat code

chtsanti:
Knowing client transport connection ID can be very helpful in triage,
especially when dealing with multiple requests associated with a
problematic connection (e.g., two access log records associated with a
single TLS connection bumping failure).

The "transport" namespace was added in hope to reduce confusion when/if
other connection_id %codes are added. The shorter and more familiar
"tcp::" namespace was rejected because we want configurations using the
new %code to continue to work well after Squid gains HTTP/3 support, and
HTTP/3 uses UDP for transport connections. Other UDP-based protocols may
have similar "connection" concepts.


rousskov:
> Since this is specifically the TCP connection ID I think this should be prefixed with "tcp::" as we are trying to do for the other protocol-specific format codes.

AFAICT, the proposed logformat `%code` is protocol-agnostic -- the logged ID can be for _any_ client connection. IIRC, Squid does not currently support any client connections that are not backed by TCP, but if such support is added, then the _same_ `%>connection_id` logformat code should be used for them.

In other words, if there is a (non-TCP) client connection that the PR code does not cover, then perhaps the PR code should be adjusted (to cover those non-TCP client connections) instead of adding the %code namespace to match the PR code.

@yadij, is there a Squid-supported non-TCP from-client connection that the PR code does not cover?

chtsanti:
Looks that the jenkins build test is stuck.

yadij:
Alex: @yadij, is there a Squid-supported non-TCP from-client connection that the PR code does not cover?

There are transactions over UDP that are supported (WCCP, SNMP). You insisted on using the "connection" terminology for TLS, so there are nested TLS and HTTPS connections as well.


Christos: Looks that the jenkins build test is stuck.

Just taking a long time. With the new build machine we can test only two PR commit per day. For now I have been manually optimizing the queue by cancelling the outdated pr-test jobs when two happen for the same PR. The test for commit b67d202 was queued for over a day, but started already and has an ETA of about 3.5 hrs to completion.

chtsanti:
> Alex: @yadij, is there a Squid-supported non-TCP from-client connection that the PR code does not cover?
> 
> There are transactions over UDP that are supported (WCCP, SNMP). You insisted on using the "connection" terminology for TLS, so there are nested TLS and HTTPS connections as well.

Just a note, some types of UDP transactions like those used for htcp and icp include message ids or request numbers, which can be used to emulate connection ids.

> Christos: Looks that the jenkins build test is stuck.
> 
> Just taking a long time. 

ok

rousskov:
> Alex: is there a ... connection that the PR code does not cover?

> There are transactions over UDP that are supported (WCCP, SNMP).

Yes, but those transactions do not operate on top of any _connection_, so I am not sure why you mention them in your answer to my question about connections.

> there are nested TLS and HTTPS connections as well.

Yes, and also HTTP connections. However, all those connections are already covered by this PR AFAICT -- their connection ID is already being logged by this PR code. Thus, again, I do not know why you mention them when answering a question about connections that a _not covered_.

Adding a namespace like `tcp::` for `%X` is useful when a single access log entry may have two different Xs (in two different namespaces). AFAICT, there is no second namespace where `%>connection_id` would have a different logged value today. I can imagine scenarios where that would change in the foreseeable future (see below) but not where that change would make using a low-level `tcp::` namespace now a good idea.

> Amos: I think this should be prefixed with "tcp::" as we are trying to do for the other protocol-specific format codes.

This PR does not add a protocol-specific %code. Today, this %code identifies a client connection regardless of the protocol(s) used for that connection. If that unexpectedly changes tomorrow, we will have a much better idea what namespaces distinguish the two client connections best. It may not be `tcp::`!

For example, when we add support for UDP-based HTTP/3, we will still want to log an internal connection ID for HTTP/3 transactions. In that future code, the internal connection ID will be influenced by HTTP/3 protocol frames rather than the Connection::id field used in this PR, but that is an internal implementation detail that does not affect the %code namespace. Folks logging/using `%>connection_id` today would reasonably expect to get a similarly usable Squid connection ID value in their logs when Squid starts receiving HTTP/3 traffic.

At HTTP/3 support time, it may also make sense to also add an `%http3::>connection_id` field to log the optional low-level non-monotonic HTTP/3 protocol field, of course. And we may want to come up with some optional namespace for the then-existing `%>connection_id` to better distinguish the two %codes, but that optional namespace will certainly not be `tcp::` because HTTP/3 connections are not TCP-based (but still need a Squid connection ID introduced in this PR).

We could add a `squid::` or `internal::` namespace _now_, but that sounds premature at best -- many (most?) access log fields are internal Squid-specific concepts; that namespace should be the assumed default rather than a required-to-be-spelled-out exception.

yadij:
> 
> > Alex: is there a ... connection that the PR code does not cover?
> 
> > Amos: There are transactions over UDP that are supported (WCCP, SNMP).
> 
> Yes, but those transactions do not operate on top of any _connection_, so I am not sure why you mention them in your answer to my question about connections.
> 

In the context of their protocols they are "connections" which this general (no-namespace) form of the code should cover.

> > Amos: there are nested TLS and HTTPS connections as well.
> 
> Yes, and also HTTP connections. However, all those connections are already covered by this PR AFAICT -- their connection ID is already being logged by this PR code.

Disagree, as this is implemented only a TCP connection ID is what gets logged.

Consider that one TCP connection (ID 1) can start TLS connection (ID 2&1), then stop midway and change into HTTP (ID 3&1) then Upgrade into TLS/1.2 (ID 4&1) then HTTP(S) (ID 5&4&1), and so on dynamically - depending only on whether the current protocol has a mechanism to switch protocols or signal its end without a TCP closure. By the definition of "connection" you had use agreeing on months back these would all need different IDs - this PR logs them all as one for the TCP layer. So lets be clear that the TCP connection is tracked by this code output not any of the others (yet).

I think we are talking about the same thing at the architecture level - the generic/non-namespace form of the code should be implemented to log relevant protocol data on a best-effort to identify the most accurate and/or reliable. With namespace variants (implemented, or reserved) for each specific data source.

That to me implies that the proper form of the generic %code display is something more complex than implemented here (eg the n&n&n list I used above). The simplest thing we can do is use the namespaced code which represents specifically the limited current implementation - no change to PR scope, and minimal diff change.


>  Thus, again, I do not know why you mention them when answering a question about connections that a _not covered_.

I interpreted your "covered" as meaning protocols that were no assigned unique IDs for their "connection".  I listed the protocols I'm aware of that are not having their connection IDs (if any) being logged by this generic log code.
If all you want is an ID to track "anything happening on X" then %master_xaction already "covers" all that by being generated one per client TCP connection. The PR description made me think something better was needed - though perhapse the real answer is you need to extend master_xaction code already.

I expect the generic code is very unsuitable as currently defined (single integer) for what is actually needed (eg the nested connections) so I am requesting that we implement the namespaced form of the code here and leave the PR scope as-is. There really is no pressing need to burn the generic name now.

> For example, when we add support for UDP-based HTTP/3, we will still want to log an internal connection ID for HTTP/3 transactions. In that future code, the internal connection ID will be influenced by HTTP/3 protocol frames rather than the Connection::id field used in this PR, but that is an internal implementation detail that does not affect the %code namespace. Folks logging/using `%>connection_id` today would reasonably expect to get a similarly usable Squid connection ID value in their logs when Squid starts receiving HTTP/3 traffic.
> 

(I am trying to avoid examples for protocols we have not yet implemented. WCCP seems equivalent to the HTTP/3 situation [client being the router] and is relevant now).

> At HTTP/3 support time, it may also make sense to also add an `%http3::>connection_id` field to log the optional low-level non-monotonic HTTP/3 protocol field, of course. And we may want to come up with some optional namespace for the then-existing `%>connection_id` to better distinguish the two %codes, but that optional namespace will certainly not be `tcp::` because HTTP/3 connections are not TCP-based (but still need a Squid connection ID introduced in this PR).

Lets not have any more weird implicit namespaces. One is more than enough trouble for newbies.

> 
> We could add a `squid::` or `internal::` namespace _now_, but that sounds premature at best -- many (most?) access log fields are internal Squid-specific concepts; that namespace should be the assumed default rather than a required-to-be-spelled-out exception.

You make a good case there _against_ adding more implicit namespaces.

rousskov:
I am reordering the conversation in order to start with the most promising bits that may render the rest of the disagreement irrelevant:

> Amos: Consider that one TCP connection (ID 1) can start TLS connection (ID 2&1), then stop midway and change into HTTP (ID 3&1) then Upgrade into TLS/1.2 (ID 4&1) then HTTP(S) (ID 5&4&1),

To treat TCP connection IDs separately from TLS connection IDs separately from HTTP connection IDs, etc., your example is better worded like this:

> A TCP connection (tcpConnId 1) can start TLS connection (tlsConnId 1), then stop midway and change into HTTP (httpConnId 1), then Upgrade into TLS (tlsConnId 2) then HTTP (httpConnId 2). A single logged transaction may be related to _multiple_ connection IDs (e.g., tcpConnId 1 and tlsConnId 2 and httpConnId 2).

I think this is a very useful example that illustrates both the validity of the current PR (all those IDs have the same value if there is no nesting) and the limitations of the current PR (the IDs diverge after meaningful nesting is supported). Thank you for inventing this example.

This PR does not support separate namespaces for different protocols, but one could certainly add such support later if needed. Such namespace separation becomes really useful when the same protocol is nested (e.g., HTTP inside HTTP or TLS inside TLS) _and_ the protocol transactions from different nesting layers can be reordered/mixed (e.g., a transaction from http1 connection happens after some transaction from http11 connection where http11 is nested inside http1).

Today, Squid does not support meaningful same-protocol nesting -- no TLS inside TLS, and HTTP CONNECT or Upgrade effectively ends the previous layer of HTTP before irreversibly switching to the next layer. This is why a single ID works well for all layers now. All of that may change, of course.

We can add a connection_id namespace today to ease anticipated introduction of meaningful nested connections support, but `tcp::` is the wrong namespace for HTTP/3 reasons discussed earlier. Unless you propose something better, we can add a `transport::` namespace (that will accommodate both TCP-based connections supported today and UDP-based connections that Squid will have to support tomorrow). Would adding a `transport::` (or a similar TCP-neutral) namespace address your concern?

N.B. If you do not want this new namespace to be _required_, then we do not need to add it now. We can wait for somebody to add support for non-transport connection IDs (with required namespaces) first. I believe that your idea of reserving unscoped %code for a combination of IDs is best implemented via a new, _plural_ %code like `%connection_ids`.


> So lets be clear that the TCP connection is tracked by this code output not any of the others (yet).

In hope to make progress, let's separate "this code" (that we seem to agree on) from the proposed interface (where you want changes that I think need further adjustments):

* We agree that the code in this PR tracks the TCP connection ID. When HTTP/3 is supported, more code will be added to track UDP-based connections.
* I hope we can agree that the interface in this PR represents the transport connection ID. When HTTP/3 is supported, the logformat `%namespace::code` introduced by _this_ PR should continue to work for HTTP/3 connections as it does for HTTP/1 connections.


> I expect the generic code is very unsuitable as currently defined (single integer) for what is actually needed (eg the nested connections) so I am requesting that we implement the namespaced form of the code here and leave the PR scope as-is. There really is no pressing need to burn the generic name now.

While I disagree that multiple IDs should be logged under a singular `%connection_id` name, I am _not_ opposed to adding a namespace to `%connection_id` now, so I hope we can make progress in that direction.

----

> If all you want is an ID to track "anything happening on X" then %master_xaction already "covers" all that by being generated one per client TCP connection

No, `%master_xaction` does not group GET, CONNECT, and bumped POST (all on the same TCP connection) together. Each must have its own master transaction ID. IIRC, `%master_xaction` does not even group two SslBump steps together, but I may be wrong about that. Master transaction groups a single client GET and all the associated ICAP RESPMOD requests together (for example).

> Amos: In the context of their protocols they are "connections" which this general (no-namespace) form of the code should cover.

*If* those protocols have a concept of a connection, *and* that concept is reflected in Squid usage of those protocols, then those connections should probably be covered by this logformat %code.

I am not a WCCP or SNMP expert, but
*  I see no mention of "connection" in WCCP v2 [specs](https://tools.ietf.org/html/draft-mclaggan-wccp-v2rev1-00), on WCCP [wiki](https://en.wikipedia.org/wiki/Web_Cache_Communication_Protocol). Instead of dealing with connections, WCCP seems focused on routing individual packets, regardless of whether that packet transport is UDP or TCP.
* I see no mention of "connection" in SNMP architecture [specs](https://tools.ietf.org/html/rfc3411). SNMP transport [specs](https://tools.ietf.org/html/rfc5590) mention transport connections, but, according to your earlier "UDP transactions" comment, Squid does not support such SNMP transports.

In summary, I see no reason why WCCP and SNMP would be relevant the proposed %code (today). However, if there is some connection ID that WCCP or SNMP code in Squid should contribute to the %code logged in this PR, please say so.

yadij:
> N.B. If you do not want this new namespace to be _required_, then we do not need to add it now. We can wait for somebody to add support for non-transport connection IDs (with required namespaces) first. I believe that your idea of reserving unscoped %code for a combination of IDs is best implemented via a new, _plural_ %code like `%connection_ids`.
> 

We seen to both understand the future needs now. I asked for it to be added, so yes I think right now it should be required. Future PR can make the generic when other protocol support is added.


> > So lets be clear that the TCP connection is tracked by this code output not any of the others (yet).
> 
> In hope to make progress, let's separate "this code" (that we seem to agree on) from the proposed interface (where you want changes that I think need further adjustments):
> 
>     * We agree that the code in this PR tracks the TCP connection ID. When HTTP/3 is supported, more code will be added to track UDP-based connections.
> 

Agreed.


>     * I hope we can agree that the interface in this PR represents the transport connection ID. When HTTP/3 is supported, the logformat `%namespace::code` introduced by _this_ PR should continue to work for HTTP/3 connections as it does for HTTP/1 connections.
> 

I can agree to it being the transport connection. But that does not quite match the description you gave ... HTTP being a transfer protocol would not have a connection - which is fine, it has MasterXaction::id.

That word will need adding to the cf.data.pre text to clarify for users and remind future us.


> 
> > I expect the generic code is very unsuitable as currently defined (single integer) for what is actually needed (eg the nested connections) so I am requesting that we implement the namespaced form of the code here and leave the PR scope as-is. There really is no pressing need to burn the generic name now.
> 
> While I disagree that multiple IDs should be logged under a singular `%connection_id` name, I am _not_ opposed to adding a namespace to `%connection_id` now, so I hope we can make progress in that direction.
> 

I think the scope should be there for now. Especially if we disagree on representing the nesting into the display value for non-namespace %code.

> > If all you want is an ID to track "anything happening on X" then %master_xaction already "covers" all that by being generated one per client TCP connection
> 
> No, `%master_xaction` does not group GET, CONNECT, and bumped POST (all on the same TCP connection) together. Each must have its own master transaction ID. IIRC, `%master_xaction` does not even group two SslBump steps together, but I may be wrong about that. Master transaction groups a single client GET and all the associated ICAP RESPMOD requests together (for example).
> 

Hmm. Okay, thanks for clarifying.


> > Amos: In the context of their protocols they are "connections" which this general (no-namespace) form of the code should cover.
> 
> _If_ those protocols have a concept of a connection, _and_ that concept is reflected in Squid usage of those protocols, then those connections should probably be covered by this logformat %code.
> 
> I am not a WCCP or SNMP expert, but
> 
>     * I see no mention of "connection" in WCCP v2 [specs](https://tools.ietf.org/html/draft-mclaggan-wccp-v2rev1-00), on WCCP [wiki](https://en.wikipedia.org/wiki/Web_Cache_Communication_Protocol). Instead of dealing with connections, WCCP seems focused on routing individual packets, regardless of whether that packet transport is UDP or TCP.
> 

It is not mentioned explicitly, but there is a essentially a background channel (connection) between Squid and each router relaying hash information about acceptible client IP ranges for TCAM tables.


>     * I see no mention of "connection" in SNMP architecture [specs](https://tools.ietf.org/html/rfc3411). SNMP transport [specs](https://tools.ietf.org/html/rfc5590) mention transport connections, but, according to your earlier "UDP transactions" comment, Squid does not support such SNMP transports.
> 

SNMP needs a multi-packet UDP connection/stream for large responses. Currently we do not send the required ID, but AFAIK that is all that is necessary. I have not added it because I'm not certain of the encoding into the packets we already send. SNMPv3 (future) will have authenticated session ("connection").

If we are going with the "transport connection" scope. That removes these upper protocols and shifts it back to a UDP transport ID (not implemented).

rousskov:
> I can agree to it being the transport connection.

Glad we found a mutually acceptable solution. Christos and/or I will adjust the PR accordingly.

> But that does not quite match the description you gave

Yes, I was describing a different solution, not the one we now settled on. This PR is not going to log the HTTP connection ID, and that is OK with me because the use case driving this PR can be addressed with the transport connection ID. If somebody decides to add HTTP connection ID logging, they would need to define where/whether the original HTTP connection ends for cases with nested HTTP connections. Thankfully, we do not need to answer that question for this PR.

rousskov:
> Christos and/or I will adjust the PR accordingly.

Done at https://github.com/squid-cache/squid/pull/546/commits/1195f5ffd87e1783145ba59c96d467bd4aa159fe.

@yadij, the ball is on your side.

PR #549, opened 2020-01-29
Remove pointer from the input of Digest nonce hashes

squidcontrib:
This is a follow-up to #491 (b20ce97), which hashed what was previously
revealed as plaintext. Removing the pointer from the input to the hash
removes the possibility that someone could recover a pointer by
reversing a hash. Having the pointer as input was not adding anything:
Squid remembers all outstanding nonces, so it really only requires
uniqueness, which is already guaranteed by the
authenticateDigestNonceFindNonce loop.

### DIFF
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -157,10 +157,10 @@ authenticateDigestNonceNew(void)
      * really bad timing with expiry and creation).  Using a random
      * component in the nonce allows us to loop to find a unique nonce.
      * We use H(nonce_data) so the nonce is meaningless to the reciever.
-     * So our nonce looks like hex(H(timestamp,pointertohash,randomdata))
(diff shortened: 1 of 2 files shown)


yadij:
OK to test

rousskov:
I added commit SHA to the PR description to provide another (and arguably more unique and Squid-specific) reference to the previous change than a GitHub-specific PR number. Most modern Squid commits use SHAs to refer to other commits.

I have no objections for this change going in, but I have not studied the code enough to approve it. If this PR should be fast-tracked, please let me know. Otherwise, it should be auto-merged in about 10 days.

yadij:
Your choice. It fully resolves the CVE-2019-18679 issue we only hardened previously. But there will not be a release tomorrow, so either process is sufficient.

PR #550, opened 2020-01-31
Fix request URL generation in reverse proxy configurations

squidadm:


yadij:
OK to test

rousskov:
FYI: Semaphore CI build [failure](https://semaphoreci.com/squid/squid/branches/pull-request-550/builds/1) should be ignored: Squid v3 does not support `--foreground` that the test harness is using.  I told GitHub that Semaphore tests are optional for v3.5. Unfortunately, I do not see a way to tell Semaphore _not_ to test v3.5 pull requests.

If this becomes a real problem, we can enhance the test harness, but I do not think the extra work is worth it at this point -- we should spend scarce resources on testing newer branches, and most functionality tests will fail for v3 (or would have to be excluded) anyway.

yadij:
As can the Debian and Ubuntu failures. Only CentOS nodes still represent the environments v3.5 is shipped with.

PR #551, opened 2020-01-31
WIP: Preserve caller context across event.h callbacks

eduard-bagdasaryan:
eventAdd*() API queues callbacks without saving the call context. Thus,
EventScheduler::checkEvents() schedules event callbacks (via AsyncCall
API) without restoring the original recipient context. We must
preserve the code context across events.

### DIFF
--- a/src/event.cc
+++ b/src/event.cc
@@ -95,6 +95,7 @@ ev_entry::ev_entry(char const * aName, EVH * aFunction, void * aArgument, double
     when(evWhen),
     weight(aWeight),
     cbdata(haveArg),
+    codeContext(CodeContext::Current()),
     next(NULL)
 {
 }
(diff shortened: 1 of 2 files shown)


rousskov:
FYI: This PR may need to be reworked to avoid leaking transaction context onto unrelated service code (and, in some cases, preserving unrelated transaction context inside that service forever). Most eventAdd() events are not transaction-specific, but many eventAdd() calls are executed in a transaction context. For example, think of some _periodic_ service cleanup event triggered by the _first_ transaction using that service. I asked Eduard to dig deeper and will review once his investigation is over.

yadij:
I suspect the simplest change would be to make events use a sub-context such as "event Foo triggered by CurrentContext()".

rousskov:
> The vast majority of the existing eventAdd() and eventAddIsh() callers are context-free global services and should stay that way. Eduard is now making good progress with identifying the few eventAdd() callers that need special treatment. This PR will change to restrict context preservation to those few special cases.

No special cases left AFAICT. Closing this stale PR as currently unnecessary. We may come back to this later if we decide to add "service code context" support for various global services.

PR #554, opened 2020-02-05
Fixed SMP kids iteration during SNMP and cache manager queries

rousskov:
The kids iterator became stale when strands were sort()ed in 8822ebe.


rousskov:
These changes are trivial but they have not been tested yet.

rousskov:
The changes survived basic tests during bug [5022](https://bugs.squid-cache.org/show_bug.cgi?id=5022) [triage](https://bugs.squid-cache.org/show_bug.cgi?id=5022#c5). They do not fix that bug. I adjusted this PR description accordingly.

.... And then I discovered that there is no bug here -- the C++ standard apparently [guarantees](https://stackoverflow.com/a/3885638) that sort() does not invalidate container iterators:

> Unless otherwise specified, ... passing a container as an argument to a library function shall not invalidate iterators to ... objects within that container.

I am closing this PR as unnecessary. Adding a source code comment to document the above may be a good idea.

PR #557, opened 2020-02-18
test-builds.sh: add possibility to force build directory

kinkie:
Having one directory per layer when doing a test build increases ccache churn
which harms build farm build times.
Adding argument to test-builds.sh to specify a specific build directory will enable increased ccache efficiency and reduced build times.

On a clean checkout for a full test build, this change
increases ccache hit ratio from 11.5% to 46.5%, and
decreases ccache size from 2 to 1.8gb.

### DIFF
--- a/test-builds.sh
+++ b/test-builds.sh
@@ -33,6 +33,11 @@ while [ $# -ge 1 ]; do
 	verbose="yes"
 	shift
 	;;
+    --builddir)
+    shift
+    builddir="$1"
+    shift
+    ;;
     --keep-going)
 	keepGoing="yes"
 	shift
@@ -69,20 +74,20 @@ buildtest() {
     opts=$1
     action=$2
     layer=`basename ${opts} .opts`
-    btlayer="bt${layer}"
-    log=${btlayer}.log
+    builddir="${builddir:-bt${layer}}"
+    log=bt${layer}.log
     echo "TESTING: ${layer}"
-    if test -e ${btlayer}; then
-	chmod -R 777 ${btlayer};
+    if test -e ${builddir}; then
+	chmod -R 777 ${builddir};
     fi
-    rm -f -r ${btlayer} || ( echo "FATAL: Failed to prepare test build sandpit." ; exit 1 )
-    mkdir ${btlayer}
+    rm -f -r ${builddir} || ( echo "FATAL: Failed to prepare test build sandpit." ; exit 1 )
+    mkdir ${builddir}
     if test "${verbose}" = "yes" ; then
-        ls -la ${btlayer}
+        ls -la ${builddir}
     fi
     {
 	result=255
-	cd ${btlayer}
+	cd ${builddir}
 	if test -e $top/test-suite/buildtest.sh ; then
 	    $top/test-suite/buildtest.sh "${action}" ${opts} 2>&1
 	    result=$?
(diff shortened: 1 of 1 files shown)


yadij:
I do not think this is a good idea. The layers are in separate directories in the first place to prevent outdated objects being shared between builds. From what I understand of ccache hashing it still does not by default include enough details to reliably distinguish between objects in our complex build environment.

For example; a common complaint used to be strange build errors about wrong enum values/ranges when HTCP was enabled or disabled in production builds. That turned out to be build objects from one test sequence being incorrectly used for a second test run. We were seeing successful build tests and users non-compilable code in apparently the same environments.

By testing as separate directories we prove that any repeat of those errors is not due to problems at our end - the farm catches those early.  I do still see occasional complaints in squid-users from new users, but a simple "delete the directory and rebuild" solves that - and I imagine a lot more are self-resolved now.

kinkie:
On Tue, Feb 18, 2020 at 11:36 PM Alex Rousskov <notifications@github.com>
wrote:

> *@rousskov* commented on this pull request.
>
> Will accidentally pressing Enter while typing test-builds.sh --builddir ~/
> erase one's home directory?
>

Yes.
We can put some safeguards in place such as barfing if the supplied path is
absolute.
But let's keep in mind that test-builds.sh is not meant for the general
public.

-- 
    Francesco

rousskov:
> We can put some safeguards in place such as barfing if the supplied path is absolute.

That would not be enough IMO: `test-builds.sh --builddir ../../<accidental ENTER>`.

> But let's keep in mind that test-builds.sh is not meant for the general public.

I hope I will not lose a lot of karma points by saying it is **my** directories I am worried about (among other things) :-).

rousskov:
> Amos: We were seeing successful build tests and users non-compilable code in apparently the same environments.

I think Amos makes a very important point. We should be testing what admins are a lot more likely to experience, even if it takes longer.

If we need to optimize the average test build speed, there are much better ways to do that than relying on obscure compiler-level caches IMO. Here is just one data point to illustrate: Running build-tests.sh takes only ~40 minutes on Semaphore CI today (and that can be reduced to ~10 minutes if we really need to speed it up).

If CI speed optimization is indeed the only known use case here, then I recommend shelving the `--builddir` idea (until there is a valid use case for it).

yadij:
> If CI speed optimization is indeed the only known use case here, then I recommend shelving the `--builddir` idea (until there is a valid use case for it).

Agreed.

The worst delays IME are from a) having the layers in the first place [1], and b) units needing to link against huge amounts of irrelevant objects (eg testHttpRequest)[2].

[1] @rbtcollins initial Squid-3 design plan to have dynamic libraries instead of --enable/--disable would be a great help here to drop the ESI and Auth layers completely (replace with daft run-time checks instead of full rebuild). We already have the dynamic library environment requirements in place for eCAP, just need to restructure the build process to make that code loadable libraries and later some refactoring to reduce the main code calling into them (Runners and AsyncJobs?). This would cut 40% of the testing time.

[2] I got 1min reduction per layer with simply the addition of a few stubs in the previous unit tests cleanup. There are a half dozen larger tests which spend 1-2 minutes each waiting on linker still. So finishing off that project should get us down around 35min (~10% reduction) and improve Makefile.am size and readability a lot.

PS. These are the long-term maintenance projects on my TODO list related to test time savings. If you want to discuss further and/or take on the task that should be planned on squid-dev or #squiddev.

rousskov:
> dynamic libraries instead of --enable/--disable

I am not sure we should go that far. I would at least consider compiling everything in and using squid.conf to turn things on/off as needed. However, let's not discuss this here. If anybody wants to change these parts of the current architecture, let's discuss elsewhere first.

P.S. Removing `#ifdefs` will help with build times, but that reduction is only a (good) side effect. The primary motivation here should be code quality and reduction of development overheads.

rousskov:
@kinkie, if we have convinced you that this kind of change should not be done (now) because it does not have a known good use case, please close this PR. If you see flaws in our logic, the ball is on your side to correct them.

PR #560, opened 2020-02-26
Preserve caller context across tunnelDelayed*Read

eduard-bagdasaryan:
tunnel.cc code approximates delay pools functionality using event.h API
that is not meant for transaction-specific events. We (temporary) add a
transaction context data member to TunnelStateData until the class
switches to using transaction-specific deferred reads API.

### DIFF
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -168,6 +168,8 @@ public:
     HappyConnOpenerPointer connOpener; ///< current connection opening job
     ResolvedPeersPointer destinations; ///< paths for forwarding the request
     bool destinationsFound; ///< At least one candidate path found
+    // TODO: remove after fixing deferred reads in TunnelStateData::copyRead()
+    CodeContext::Pointer codeContext; ///< our creator context
 
     // AsyncCalls which we set and may need cancelling.
     struct {
@@ -318,7 +320,8 @@ TunnelStateData::TunnelStateData(ClientHttpRequest *clientRequest) :
     startTime(squid_curtime),
     waitingForConnectExchange(false),
     destinations(new ResolvedPeers()),
-    destinationsFound(false)
+    destinationsFound(false),
+    codeContext(CodeContext::Current())
 {
     debugs(26, 3, "TunnelStateData constructed this=" << this);
     client.readPendingFunc = &tunnelDelayedClientRead;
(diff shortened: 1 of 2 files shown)


src/tunnel.cc
```diff
@@ -168,6 +168,8 @@ class TunnelStateData: public PeerSelectionInitiator
     HappyConnOpenerPointer connOpener; ///< current connection opening job
     ResolvedPeersPointer destinations; ///< paths for forwarding the request
     bool destinationsFound; ///< At least one candidate path found
+    // TODO: remove after fixing deferred reads in TunnelStateData::copyRead()
```
yadij: I don't think this comment is very clear what is needed to fix the TODO.
 What is eg "proper" ? just _doing_ deferred reads at all ?

eduard-bagdasaryan:
> What is stopping deferred reads being added now instead of this workaround?

Development overheads. We need to add some functionality to make this work: a method similar to MemObject::delayRead(), DeferredReadManager and place DeferredReadManager::kickReads() calls in proper places. I would say that this deserves a separate PR.

src/tunnel.cc
eduard-bagdasaryan: The point is that once we fix the tunnel delay code with DeferredRead, removing the misused event, there should be no need to keep the context TunnelStateData because there should be no context loss when delaying reads.

rousskov:
> Amos: What is stopping deferred reads being added now instead of this workaround?

> Eduard: Development overheads. 

To be more precise, it is a _combination_ of two factors:

1. Solving the discovered bugs is hard, for many reasons, including those mentioned by Eduard. Even the authors of the current official workaround added to TunnelStateData::copyRead() in 9fe3d31 did not fix the underlying bug despite a bug-fixing focus of their work.
2. This specific PR is not about fixing delay pools. It is about supplying transaction context. Any other bug fixes are out of scope here.

If fixing delay pools were easy OR if this PR was about fixing delay pool bugs, then we would have fixed those delay pools bugs, but the above combination tells me that we made the right call to stay away.

FWIW, Factory may be able to work on fixing the delay pool bugs as well, but it will probably take several months. Meanwhile, with this and other small improvements, we can and should approach 100% transaction context coverage soon, and that is a valuable milestone on its own. Waiting for the proper delay pools fix instead would be wrong in this situation IMO.

src/tunnel.cc
rousskov: > I don't think this comment is very clear what is needed to fix the TODO.

@yadij, please reread the TODO comment. It is very clear about what needs to be done here -- the data member should be removed. The comment also provides a precondition for that removal -- fixing a bug in TunnelStateData::copyRead(). The details of that bug are documented in that method. They should not be documented _here_, in this TODO.

>  What is eg "proper" ? just doing deferred reads at all ?

This TODO does not use the word "proper". You are probably referring to the PR description, which contained the following sentence:

> Future adjusting this scope by proper DeferredRead operations should eliminate events thus making TunnelStateData::codeContext unnecessary (TODO).

The wording can be polished and the word "proper" can be removed (now done), but this was a (minor) English problem, not a PR scope/implementation problem. Please see if the adjusted [wording](https://github.com/squid-cache/squid/pull/560#issue-380369669) addresses your concern.

yadij: The new wording is clear. Thank you.

rousskov:
Same Jenkins problems here AFAICT. Hopefully they will clear up after recent fixes by @kinkie (thank you!).

PR #561, opened 2020-02-26
Ban reserved annotations in "note", "adaptation_meta" directives

chtsanti:
Squid defines a list of names used internally for exchanging name=value
pairs with various helpers. When Squid receives a name=value pair with a
reserved name, Squid stores it as if it was any other annotation; the
information can be checked with a "note" ACL and logged with %note. An
admin who configures a custom annotation with the same reserved name may
see strange/unexpected results such as seemingly corrupted access.log
record fields and mismatching ACLs.

Squid already prohibits reserved annotation names in
annotate_transaction and annotate_client ACLs. This change adds the
missing protection to the "note" and adaptation_meta directives.


src/Notes.cc
```diff
@@ -124,6 +125,30 @@ Note::toString(const char *sep) const
...
+    "log",
+    "tag",
+    "ttl",
+    "ha1",
+    "rewrite-url",
+    "url"
+};
+const Notes::KeysList Notes::BlackList(std::begin(NotesBlackListStr), std::end(NotesBlackListStr));
+
+Notes::Notes(const char *aDescr, const Notes::KeysList *metasBlacklist, bool allowFormatted): descr(aDescr), blacklisted(Notes::BlackList), formattedValues(allowFormatted)
```
yadij: This being a long line, please use multi-line form for the initialization list entries:

```suggestion
Notes::Notes(const char *aDescr, const Notes::KeysList *metasBlacklist, bool allowFormatted) :
        descr(aDescr),
        blacklisted(Notes::BlackList),
        formattedValues(allowFormatted)
```

src/Notes.h
```diff
@@ -109,12 +109,13 @@ class Notes : public RefCountable
...
     typedef RefCount<Notes> Pointer;
+    typedef std::vector<SBuf> KeysList;
     typedef std::vector<Note::Pointer> NotesList;
     typedef NotesList::iterator iterator; ///< iterates over the notes list
     typedef NotesList::const_iterator const_iterator; ///< iterates over the notes list
 
-    Notes(const char *aDescr, const char **metasBlacklist, bool allowFormatted = true): descr(aDescr), blacklisted(metasBlacklist), formattedValues(allowFormatted) {}
-    Notes(): descr(nullptr), blacklisted(nullptr), formattedValues(false) {}
+    explicit Notes(const char *aDescr, const KeysList *metasBlacklist = nullptr, bool allowFormatted = true);
+    explicit Notes();
```
yadij: Since you are re-arranging the constructors you may as well convert this class to C++11 initialization and use "= default" here instead of adding a new stub file.

src/Notes.cc
chtsanti: commit de7ed8eb

src/Notes.h
chtsanti: If I remember correctly, I got strange results when I tried to initialize  Notes::blacklist with the static Notes::BlackList member  as C++11 class member , I think when the Adaptation::Config::metaHeaders initialized inside adaptation/Config.cc file. 
No compile errors but the Notes::blacklist member appeared to have strange  contains. I did not check thoroughly I just reverted to the classic constructor. I considered it as initialization order problem but now I am thinking that maybe the problem was just (local) linking problems.

Do you believe it is important to check and use C++11 here?

rousskov: @chtsanti, It sounds like you experiences the static initialization order fiasco: In general, one should not initialize static Y in y.cc using static X in x.cc. The code will compile but may not work in some builds because the initialization code in y.cc may run _before_ the initialization code in x.cc.

I am not sure, but AFAICT, a similar initialization bug was still in the PR code. It may became hidden from your tests after you shoveled the code around, but I saw no guarantee that the Adaptation::Config::metaHeaders global (that uses Notes) would always be initialized _after_ the Notes::BlackList global (that Notes use).

I fixed that in branch commit b00ac92.

rousskov: @yadij, I got rid of stub_Notes.cc in branch commit d5fa335, hopefully addressing your change request.

yadij: Yes that removes the issue I had and more.

rousskov:
@yadij, thank you for re-reviewing this quickly. I am leaving it for @chtsanti to clear this PR for merging once he reviews (and, if necessary, adjusts) my polishing changes.

chtsanti:
It is not important and I can not remember any more but I think the initial trigger for the initialization order problem was a "make check" failure (not compile but test run). The metaHeaders in adaptation code probably it does not have any problem. And unfortunately even if I initially I was mark it as a problem I wanted to looked more on it, I forgot it later. It is easy to make a make bug.

I think the Notes::BlackList() now can be converted to a normal static member again, but not important.

I believe some work in the icmp sources and related libraries worth, to remove the need for these "copied_SOURCES" hack  and the need for stub files.

>  I am leaving it for @chtsanti to clear this PR for merging once he reviews (and, if necessary, adjusts) my polishing changes.

Removed.

yadij:
> 
> > I am leaving it for @chtsanti to clear this PR for merging once he reviews (and, if necessary, adjusts) my polishing changes.
> 
> Removed.

@chtsanti I think Alex meant setting the M-cleared-for-merge label

PR #562, opened 2020-02-27
WIP: Bug 5021 - Add a script to fix spelling errors with codespell

mrumph:
[Bug 5021] Add a script to fix spelling errors with codespell

Run "scripts/spell-check.sh" from the Squid tree root directory.
For details, see bug report 5021:
- https://bugs.squid-cache.org/show_bug.cgi?id=5021


rousskov:
> From Bugzilla: I wasn't sure how to do separate pull requests for each commit.

You use a separate git branch for each PR/work. Git branches are "cheap". 

To untangle this, I suggest moving the first commit from your PR branch (i.e. 0678534) into a separate/new branch. The following worked in my tests (that are based on [GitHints](https://wiki.squid-cache.org/GitHints) setup):

```
git checkout upstream/master # official code
# ... You are in 'detached HEAD' state...
git cherry-pick 0678534 # your first/primary commit (scripting)
git checkout -b bug-5021-scripting # creates a branch with the above change
git push --set-upstream origin bug-5021-scripting # publish in your repository
... now submit bug-5021-scripting as a pull request ...
```

You could repeat the above for the second commit (https://github.com/squid-cache/squid/pull/562/commits/2c7a572708ac1e7fa7073b0b30fa35f753eb6af7), creating a new branch and PR, OR, if you prefer, you can delete the first commit from this PR branch. I think the former is easier, but I am not a git or GitHub guru.

At any rate, do not worry about the current wrong PR state -- we can figure it out!

INSTALL
```diff
@@ -17,7 +17,7 @@ To run a Cache, you will need to:
 	1. customize the squid.conf configuration file:
 	      % vi /usr/local/squid/etc/squid.conf
 
-	2. Initalise the cache:
+	2. Initialise the cache:
```
rousskov: @mrumph, is there a way to tell codespell to prefer one spelling over the other? This is not critical or even important, but most Squid sources use the (American) "initialize" variant rather than the (British) "initialise" variant. I do not think codespell should change one for the other (or, at the very least, that deserves a separate discussion that I do not want to have now/here), but when fixing a spelling mistake, can we tell codespell to use the prevailing (in Squid) variant?

```shell
$ git grep -i initialise | wc -l
88
$ git grep -i initialize | wc -l
416
```

mrumph: One possibility (but not really desirable for a number of reasons) is to supply our own dictionary with the -D option on codespell.

rousskov: If that custom dictionary will _replace_ the standard one (rather than just overwrite a few entries), then I agree that `-D` is not a good solution. Let's ignore this problem.

yadij:
Closing this now that the work has been split into PR #565 and PR #578

PR #565, opened 2020-03-04
Bug 5021: Add a script to fix spelling error with codespell

mrumph:
Running scripts/spell-check.sh from the Squid source tree root directory
should fix many spelling errors in a subset of git-controlled files.

This commit excludes the spelling fixes themselves.


mrumph:
This pull request is related to pull request #562.
It isolates the first commit, so that it can be reviewed separately.
The commit simply adds a spell check script and a white list of words to allow.

scripts/spell-check.sh
```diff
@@ -0,0 +1,76 @@
...
+if test "${CODESPELL_LOC}" = ""; then
+    echo "This script requires codespell which was not found."
+    exit 1
+fi
+
+CODESPELL_VER=`codespell --version 2>&1`
+echo "The codespell version is ${CODESPELL_VER}."
+
+UNSTAGED_CHANGES=`git diff | wc -l`
+if test "${UNSTAGED_CHANGES}" != "0"; then
```
rousskov: Does the suggestion below work? If we can use a simpler, more reliable interface, we should. I am also concerned that comparing raw `wc` output with a literal string may not always work as intended.

```suggestion
if ! git diff --quiet; then
```

scripts/spell-check.sh
```diff
@@ -0,0 +1,76 @@
...
+## contributions from numerous individuals and organizations.
+## Please see the COPYING and CONTRIBUTORS files for details.
+##
+
+#
+# This script runs codespell against selected files.
+#
+# Use -q or --quiet option to run quietly
+#
+
```
rousskov: Have you considered something like `set -e` to avoid cascading or undetected errors causing problems, especially if this script is used in a "make" target or another script.

scripts/spell-check.sh
```diff
@@ -0,0 +1,76 @@
...
+for FILENAME in `git ls-files`; do
+    # skip subdirectories, git ls-files is recursive
+    test -d $FILENAME && continue
+
+    case ${FILENAME} in
+
+    doc/*.txt|doc/*/*.txt)
+        ;;	    
+
+    *.h|*.c|*.cc|*.cci|*.pl|*.sh|*.pre|*.pl.in|*.pm|*.dox|*.html|*.txt|*.sql|errors/templates/ERR_*|INSTALL|README|QUICKSTART)
```
rousskov: I am not a shell expert. Is there a trick that would place these patterns one per line to improve readability and maintenance?

scripts/spell-check.sh
```diff
@@ -0,0 +1,76 @@
...
+echo "The codespell version is ${CODESPELL_VER}."
+
+UNSTAGED_CHANGES=`git diff | wc -l`
+if test "${UNSTAGED_CHANGES}" != "0"; then
+    echo "There are unstaged changes. Stage these first to prevent conflict."
+    exit 1
+fi	
+
+WHITE_LIST=scripts/codespell-whitelist.txt
+if test ! -f "${WHITE_LIST}"; then
```
rousskov: What happens if this test is removed and `codespell-whitelist.txt` is missing? If codespell quits with an error, then perhaps we can remove this check? Please note that you may need to surround `${WHITE_LIST}` in quotes where you call codespell so that codespell gets an empty string as a parameter value.

scripts/spell-check.sh
```diff
@@ -0,0 +1,76 @@
...
+
+    *.h|*.c|*.cc|*.cci|*.pl|*.sh|*.pre|*.pl.in|*.pm|*.dox|*.html|*.txt|*.sql|errors/templates/ERR_*|INSTALL|README|QUICKSTART)
+        #
+        # Run codespell against specific file
+        #
+        if test "${QUIET}" = "0"; then
+            echo "Running codespell for ${FILENAME}"	
+        fi
+ 	codespell -d -q 3 -w -I ${WHITE_LIST} ${FILENAME}
+        if test "$?" != "0"; then
```
rousskov:
```suggestion
 	if ! codespell -d -q 3 -w -I "${WHITE_LIST}" ${FILENAME}; then
```

scripts/spell-check.sh
```diff
@@ -0,0 +1,76 @@
...
+    case $1 in
+        -q | --quiet )
+            QUIET=1
+            ;;
+    esac
+    shift
+done
+
+#
+# Scan for file-specific actions
```
rousskov: Stale/copied comment?

scripts/spell-check.sh
```diff
@@ -0,0 +1,76 @@
...
+#
+
+CODESPELL_LOC=`which codespell`
+if test "${CODESPELL_LOC}" = ""; then
+    echo "This script requires codespell which was not found."
+    exit 1
+fi
+
+CODESPELL_VER=`codespell --version 2>&1`
+echo "The codespell version is ${CODESPELL_VER}."
```
rousskov: I think we can simplify this quite a bit and avoid relying on "which". Something along these lines, perhaps?

```suggestion
echo -n "Codespell version: "
if ! codespell --version; then
    echo "This script requires codespell which was not found."
    exit 1
fi
```

rousskov: I doubt we should add the `-q` option (too much work for too little gain), but if you do add that option, then this output ought to be covered by it IMO.

scripts/spell-check.sh
rousskov: BTW, the current code would proceed when run in a directory _not_ controlled by git. That case is common for scripts executed in sources unpacked from a tarball. Proceeding in such a case is relatively safe because `git ls-files` will also fail later.

One could argue that it is better to stop sooner because, for example, future script changes may add support for applying codespell to files or directories supplied on the command line. The suggested `! git diff --quiet` check has that "exit ASAP" effect. One could argue also that the echoed `There are unstaged changes` message is a bit odd in that case, but it is possible to view all files in such a directory as effectively unstaged (and git's own `Not a git repository` message would help clarify things).

rousskov:
@mrumph, I adjusted your PR description (and the future commit message) to follow Squid Project style and satisfy automated formatting checks. If you are curious what automated checks and merge actions are applied to Squid PRs, please see [CheckList](https://wiki.squid-cache.org/MergeProcedure) and [Anubis](https://github.com/measurement-factory/anubis#readme).

rousskov:
> I will come back after reviewing the results of applying this script

I looked through the actual spelling changes (2c7a572708ac1e7fa7073b0b30fa35f753eb6af7) and was surprised to see no truly bad ones. Either codespell is just terrific at avoiding problematic areas or @mrumph did an excellent job whitelisting them! I had one minor question at https://github.com/squid-cache/squid/pull/562/files#r388061762 but that wrinkle can be ignored if needed.

I also plan to run the script myself. If there are no red flags there, and after this script-focused PR is cleaned up, I would be happy to support merging this PR, committing the script results (in another dedicated PRs), and automating these checks.

scripts/spell-check.sh
mrumph: This was copied from another script but seemed appropriate.
Removed in next commit.

scripts/spell-check.sh
mrumph: Added "set -e" in next commit.

scripts/spell-check.sh
mrumph: Applied in next commit.

scripts/spell-check.sh
mrumph: The case statement in bash scripts is not as flexible for this as the switch statement in C.
But using backslashes to break up the line does seem to work.
Instead of one per line, I suggest using small related groups of file types.
I will submit one version of this in the commit, then we can go from there.

scripts/spell-check.sh
mrumph: Without this test, codespell gives the following message and the dumps a long usage page before exiting the script.
"ERROR: cannot find ignore-words file: scripts/xxcodespell-whitelist.txt"
The current result seems cleaner. 
I will wait for a response before removing the test.
And yes, the quotes are needed in case of an empty string (the script goes wild without them).

scripts/spell-check.sh
mrumph: Applied in the next comment.

scripts/spell-check.sh
mrumph: Removed the -q option in the next commit.

mrumph: Applied in next commit.

scripts/spell-check.sh
rousskov: > ... dumps a long usage page before exiting the script.

I agree that your current solution is appropriate due to this noisy usage dump. I consider this change request resolved. Thank you.

rousskov:
Status update: I have resolved the change requests that I consider fully addressed by https://github.com/squid-cache/squid/pull/565/commits/bc08d0fc0656b8189aabfbdba13e5a7e82f2171d. Thank you! I need a bit more time to finalize two or three outstanding issues and test the script. I may have a chance to do that this weekend.

rousskov:
I am getting the following error when testing the current PR with codespell v1.14.0 on Ubuntu 19.04:

```
src/acl/external/LDAP_group/ext_ldap_group_acl.cc:422: tread  ==> tread, thread
codespell failed for src/acl/external/LDAP_group/ext_ldap_group_acl.cc
```

The important (but invisible) part here is that the problematic source code file does not have a word `tread` in it. It has a `\tread` sequence which codespell misinterprets. Our script stops, which is a good thing, but we kind of got lucky here. @mrumph, I assume you did not get this problem in your tests. Do you know why? Perhaps you use a different codespell version?

The core of the problem could be related to https://github.com/codespell-project/codespell/pull/1422 but I am not sure.

scripts/spell-check.sh
rousskov: Thank you. I polished this further in 9b3822b and now consider this change request addressed.

scripts/spell-check.sh
rousskov: Thank you. I polished this further in 5523abb. I consider this change request addressed, but if you insist on keeping the output I removed in that commit, I will restore it if another reviewer backs you up. FWIW, I do not think it is a good idea to be so noisy by default -- that noise often obscures (more) important messages.

rousskov:
Status update: I made a few adjustments and have resolved all the outstanding change requests. I also started to test the script and found one [issue](https://github.com/squid-cache/squid/pull/565#issuecomment-602327068) that we will have to resolve, one way or another, before this PR is merged. I requested more information in hope to be able to suggest the best way forward.

@mrumph, please check my recent changes (up to branch commit f44a864) to make sure that you more-or-less agree with them and that I did not miss anything important. If you are learning git: `git show SHA` will show you the changes in the given commit. `git diff SHA` will show you the changes since the given commit.

mrumph:
For the dictionary entry "tread  ==> tread, thread", I didn't see a hit on this, but there were a few other words that I added to the white list  because of being preceded by "\n" or "\t".
Perhaps my default dictionary does not include that entry.
I was using the codespell installed on Ubuntu 18.04.1 LTS (GNU/Linux 4.15.0-29-generic x86_64).

mrumph:
All of the commits by Rousskov look good to me.

rousskov:
> I was using the codespell installed on Ubuntu 18.04.1 LTS

What does your `codespell --version` say?

mrumph:
My codespell version is 1.16.0, but this replaced an earlier 1.8 version.
For this issue, we should be able to resolve this by adding "tread" to the white list.

rousskov:
> For this issue, we should be able to resolve this by adding "tread" to the white list.

Done at 37649dc.

rousskov:
OK to test

rousskov:
@yadij, do we want to include the script (and its whitelist file) in the "make dist" tarball? If yes, we will need to adjust scripts/Makefile.am before this PR goes in. I am guessing that the spellchecking script should _not_ be included because we do not include scripts/source-maintenance.sh, but you know better.

Also, if you want this script to be called from scripts/source-maintenance.sh, please say so. I think it should eventually be called from there (because it is a part of source code maintenance), but I do not know whether adding it _now_ is a good idea (because it may affect your packaging scripts).

mrumph:
So using "$@" will limit the list of files to a specific list of directories.
That seems like a useful feature to add.
Also, it might be good to add some usage comment at the beginning of the script that it uses the parameters in this way.

mrumph:
Sorry.  The $@ can only be used as a list of directories if it is not quoted.
So `git ls-files "$@"` should probably be `git ls-files $@`. 
Try the difference between:
git ls-files scripts tools
and:
git ls-files "scripts tools"

yadij:
> @yadij, do we want to include the script (and its whitelist file) in the "make dist" tarball?

FTR: criteria for that answer is whether the test/check is verifying or altering how the end-user code builds and operates. From what I have seen so far this is just internal textual changes, so no we only need the script itself in the repo. The tarballs get its output.

> 
> Also, if you want this script to be called from scripts/source-maintenance.sh, please say so. I think it should eventually be called from there (because it is a part of source code maintenance), but I do not know whether adding it _now_ is a good idea (because it may affect your packaging scripts).

We will need to arrange its call like the astyle one either way - ensuring that we only have a manually verified tool version used. So the build machines do not fight over trivial version output differences. If that is done before this gets merged it should be fine adding now.

rousskov:
> Sorry. The $@ can only be used as a list of directories if it is not quoted.

What makes you think that?


Please double check. In my tests, quoted `$@` works exactly the way one would want it to work -- quoting individual parameters rather than the whole parameter list -- it is some kind of shell magic. Try, for example, the command below and you should see that the first (non-existent) directory with a space in its name is correctly skipped while the `doc` directory is processed.

```shell
sh -x ./scripts/spell-check.sh "src " 'doc' 2>&1 | less
```

> Try the difference between:
> git ls-files scripts tools
> and:
> git ls-files "scripts tools"

Your example above does not use the `$@` magic so it does not count.

Even if I am right, we can still remove quoting because Squid sources do not use filenames with special characters. Please let me know if you insist on that removal, and I will do it.

rousskov:
@yadij, thank you for sharing your insights! Subject to @mrumph [blessing](https://github.com/squid-cache/squid/pull/565#issuecomment-603244247), I plan to clear this PR after documenting anonymous parameters per @mrumph [request](https://github.com/squid-cache/squid/pull/565#issuecomment-602878326).

> We will need to arrange its call like the astyle one either way - ensuring that we only have a manually verified tool version used. So the build machines do not fight over trivial version output differences. If that is done before this gets merged it should be fine adding now.

I hope we can avoid running this new script (and, long term, running `source-maintenance.sh`) on the build machines by controlling repository input during automated and centralized PR acceptance testing, just like we already ban bad whitespace. If my hopes materialize, there will be no need for locking down the codespell version. If they do not, then we will add version control as a small separate PR (while attempting to implement it in a more flexible way than it has been done for astyle).

rousskov:
> it might be good to add some usage comment at the beginning of the script that it uses the parameters in this way.

Agreed. Done in 94a2043. Please review that commit and 21a54ee that polishes error output.

@mrumph, with these changes, I am waiting for your blessing to clear this PR for merging OR your request for more adjustments (e.g., unquoting `"$@"`).

mrumph:
"Please double check. In my tests, quoted $@ works exactly the way one would want it to work -- quoting individual parameters rather than the whole parameter list -- it is some kind of shell magic. 
Even if I am right, we can still remove quoting because Squid sources do not use filenames with special characters. Please let me know if you insist on that removal, and I will do it."

First of all, I don't "insist" on anything.  I am only trying to help the project.
Yes.  I was mistaken on how "$@" works in a script.  I've verified your magic.
The spell check script looks good to me as you have changed it.

rousskov:
> First of all, I don't "insist" on anything. I am only trying to help the project.

Sorry, I did not mean to imply anything negative: As the primary author, you have a well-deserved right to pick one of the several options approved by the reviewer. I did not want to impose (technically optional) quoting if you happen to dislike it for some reason, especially since I am not a shell scripting expert...

>  The spell check script looks good to me as you have changed it.

Glad we are on the same page! I am clearing the PR for merging. Other reviewers may block it at the last minute, but I still want to thank you for investing so much into this valuable improvement. I especially appreciate your prompt reaction to my questions and changes.

rousskov:
@mrumph, now that this script is in, please either submit a new PR with the actual spelling fixes done by the script OR adjust your old PR #562 to contain nothing but the spelling fixes done by the committed/official script, whichever is easier for you.

PR #566, opened 2020-03-05
Bug 5030: Negative responses are never cached

DrDaveD:
Negative caching was blocked by checkCachable().

Since 3e98df2, Squid cached ENTRY_NEGCACHED entries in memory cache
only. Back then, storeCheckSwapable() prevented what later became
ENTRY_NEGCACHED entries from going to disk. The design was obscured by
8350fe9 that renamed storeCheckSwapable() to storeCheckCachable().

Commit 97754f5 violated that (obscured) design by adding a
checkCachable() call to StoreEntry::memoryCachable(), effectively
blocking ENTRY_NEGCACHED entries from the memory cache as well. That
call should have been added, but checkCachable() should not have denied
caching rights to ENTRY_NEGCACHED -- the corresponding check should have
been moved into StoreEntry::mayStartSwapOut().

By removing ENTRY_NEGCACHED from checkCachable(), we now allow
ENTRY_NEGCACHED entries into both memory and disk caches, subject to all
the other checks. We allow ENTRY_NEGCACHED to be cached on disk because
negative responses are fundamentally no different than positive ones:
HTTP allows caching of 4xx and 5xx responses expiring in the future.
Hopefully, the increased disk cache traffic will not be a problem.

### DIFF
--- a/src/store.cc
+++ b/src/store.cc
@@ -1028,8 +1028,8 @@ storeCheckCachableStats(StoreEntry *sentry)
                       store_check_cachable_hist.no.not_entry_cachable);
     storeAppendPrintf(sentry, "no.wrong_content_length\t%d\n",
                       store_check_cachable_hist.no.wrong_content_length);
-    /* for backward compatibility */
-    storeAppendPrintf(sentry, "no.negative_cached\t0\n");
+    storeAppendPrintf(sentry, "no.negative_cached\t%d\n",
+                      0); // TODO: Remove this backward compatibility hack.
     storeAppendPrintf(sentry, "no.missing_parts\t%d\n",
                       store_check_cachable_hist.no.missing_parts);
     storeAppendPrintf(sentry, "no.too_big\t%d\n",
--- a/src/store.cc
+++ b/src/store.cc
@@ -899,7 +899,7 @@ struct _store_check_cachable_hist {
         int non_get;
         int not_entry_cachable;
         int wrong_content_length;
-        int negative_cached;
+        int expired_negative_cached;
         int too_big;
         int too_small;
         int private_key;
@@ -910,6 +910,7 @@ struct _store_check_cachable_hist {
 
     struct {
         int Default;
+        int unexpired_negative_cached;
     } yes;
 } store_check_cachable_hist;
 
@@ -980,8 +981,13 @@ StoreEntry::checkCachable()
             debugs(20, 2, "StoreEntry::checkCachable: NO: wrong content-length");
             ++store_check_cachable_hist.no.wrong_content_length;
         } else if (EBIT_TEST(flags, ENTRY_NEGCACHED)) {
-            debugs(20, 3, "StoreEntry::checkCachable: NO: negative cached");
-            ++store_check_cachable_hist.no.negative_cached;
+            if (squid_curtime < expires) {
+                debugs(20, 3, "StoreEntry::checkCachable: YES: unexpired negative cached");
+                ++store_check_cachable_hist.yes.unexpired_negative_cached;
+                return 1;
+            }
+            debugs(20, 3, "StoreEntry::checkCachable: NO: expired negative cached");
+            ++store_check_cachable_hist.no.expired_negative_cached;
             return 0;           /* avoid release call below */
         } else if (!mem_obj) {
             // XXX: In bug 4131, we forgetHit() without mem_obj, so we need
@@ -1033,8 +1039,8 @@ storeCheckCachableStats(StoreEntry *sentry)
                       store_check_cachable_hist.no.not_entry_cachable);
     storeAppendPrintf(sentry, "no.wrong_content_length\t%d\n",
                       store_check_cachable_hist.no.wrong_content_length);
-    storeAppendPrintf(sentry, "no.negative_cached\t%d\n",
-                      store_check_cachable_hist.no.negative_cached);
+    storeAppendPrintf(sentry, "no.expired_negative_cached\t%d\n",
+                      store_check_cachable_hist.no.expired_negative_cached);
     storeAppendPrintf(sentry, "no.missing_parts\t%d\n",
                       store_check_cachable_hist.no.missing_parts);
     storeAppendPrintf(sentry, "no.too_big\t%d\n",
@@ -1047,6 +1053,8 @@ storeCheckCachableStats(StoreEntry *sentry)
                       store_check_cachable_hist.no.too_many_open_files);
     storeAppendPrintf(sentry, "no.too_many_open_fds\t%d\n",
                       store_check_cachable_hist.no.too_many_open_fds);
+    storeAppendPrintf(sentry, "yes.unexpired_negative_cached\t%d\n",
+                      store_check_cachable_hist.yes.unexpired_negative_cached);
     storeAppendPrintf(sentry, "yes.default\t%d\n",
                       store_check_cachable_hist.yes.Default);
 }
@@ -1374,6 +1382,7 @@ StoreEntry::negativeCache()
 #else
         expires = squid_curtime;
 #endif
+    debugs(20, 6, "expires = " << expires << ", curtime = " << squid_curtime);
     EBIT_SET(flags, ENTRY_NEGCACHED);
 }
 
--- a/src/store.cc
+++ b/src/store.cc
@@ -899,7 +899,6 @@ struct _store_check_cachable_hist {
         int non_get;
         int not_entry_cachable;
         int wrong_content_length;
-        int expired_negative_cached;
         int too_big;
         int too_small;
         int private_key;
@@ -909,8 +908,8 @@ struct _store_check_cachable_hist {
     } no;
 
     struct {
+        int negative_cached;
         int Default;
-        int unexpired_negative_cached;
     } yes;
 } store_check_cachable_hist;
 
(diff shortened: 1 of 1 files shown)


DrDaveD:
Good points, Alex.  I believe I addressed them all.

src/store.cc
```diff
@@ -1033,8 +1033,6 @@ storeCheckCachableStats(StoreEntry *sentry)
                       store_check_cachable_hist.no.not_entry_cachable);
     storeAppendPrintf(sentry, "no.wrong_content_length\t%d\n",
                       store_check_cachable_hist.no.wrong_content_length);
-    storeAppendPrintf(sentry, "no.negative_cached\t%d\n",
```
yadij: Please be aware the removal of this UI value constitutes a breaking change since it may be depended on by third-party tools. As such it prevents backporting the bug fix to v5 or v4. For now we would be better adding an XXX and hard-coding this to display a '0' value in the UI.

src/store.cc
```diff
@@ -1009,6 +1005,10 @@ StoreEntry::checkCachable()
         } else if (fdNFree() < RESERVED_FD) {
             debugs(20, 2, "StoreEntry::checkCachable: NO: too many FD's open");
             ++store_check_cachable_hist.no.too_many_open_fds;
+        } else if (EBIT_TEST(flags, ENTRY_NEGCACHED)) {
+            debugs(20, 3, "StoreEntry::checkCachable: YES: negative cached");
```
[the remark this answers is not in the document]
rousskov: Please do not remove the redundant `StoreEntry::checkCachable: ` debug output. Consistency is far more important here than redundancy -- all related debugging in this method should follow the same format.

src/store.cc
rousskov: > As such it prevents backporting the bug fix to v5 or v4.

I do not see how a removal in v6 _prevents_ backporting the fix to v5 and v4. Yes, backporting becomes a bit more difficult -- the old line should probably be preserved in some form -- but it is still possible to backport the fix, is it not?

src/store.cc
```diff
@@ -1009,6 +1005,10 @@ StoreEntry::checkCachable()
...
+            ++store_check_cachable_hist.yes.negative_cached;
+            return 1;
```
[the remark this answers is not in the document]
DrDaveD: I'd be happy to do that, but similar to Alex's comments on the debug message, isn't it more important to have consistency in the function than to have one 'true' with all the other 1s and zeros returned in the function?  Would you like me to change them all?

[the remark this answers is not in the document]
rousskov: While I disagree with this change request (on the same vague consistency grounds), to save time and keep making progress, my recommendation is to change any touched return values. Unlike debugging/output changes in the other change request, here, it will address Amos requests while causing only mild inconsistencies.

Please do not change untouched `return` lines!

src/store.cc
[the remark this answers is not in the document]
DrDaveD: How about we change them all for v6 and not backport that change.  To make backporting easier, I propose that I submit a separate PR to change the debug messages and return values in this function, and then don't backport that PR.

src/store.cc
DrDaveD: If you want me to make backporting easy, I can leave the messages and counters untouched in this PR and then include all counter changes and updated messages in my second PR that doesn't get backported.

DrDaveD: Or go ahead and change the counters in this PR and only leave the storeCheckCachableStats the same until the second PR.

rousskov: > I can leave the messages and counters untouched in this PR and then include all counter changes and updated messages in my second PR that doesn't get backported.

> Or change the counters in this PR and only leave the storeCheckCachableStats the same until the second PR.

I doubt butchering this PR to accommodate future porting needs is the best approach, but we can do it one step at a time:

1. Get this PR to a point where it can be merged, with all the changes necessary for the PR to be self-contained and fully addressing all the in-scope problems.
2. Decide whether to split this PR for some valid reason.

If you think that we are done with step 1 already, then please wait for my review. I did not have a chance to check the PR after your latest changes.

DrDaveD: I think this PR is fully ready for v6 as-is, so please do review when you have a chance.

src/store.cc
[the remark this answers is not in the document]
yadij: That would be fine.

src/store.cc
rousskov: Assuming the PR code is correct, this `if` statement looks out of place (at best) or even unnecessary:

1. Its addition does not change the method outcome. Its only effect is a side-effect.

2. Its only justification could be the collection of statistics about negatively cached objects (e.g., what percentage of cached objects were negatively cached), but this method is the wrong place to collect such statistics -- this method does not (fully) determine which objects will be cached. The statistics in this method is supposed to be about key decisions made in this method, but, per item 1 above, this `if` statement only pretends to make a decision.

Assuming the PR code is correct, I suggest removing this if statement and the corresponding statistics. If somebody wants to add a new "the number/ratio of cached objects that were negatively cached" and/or a new "the number/ratio of responses that were eligible for negative caching" statistics, they can certainly do that. If you want to add that statistics, you can do that in this PR, but I do not think it is _required_ for this PR acceptance. I have not researched the best place(s) to add such statistics -- it may not be easy to find the right place(s).

src/store.cc
```diff
@@ -1374,7 +1374,9 @@ StoreEntry::negativeCache()
 #else
         expires = squid_curtime;
 #endif
-    EBIT_SET(flags, ENTRY_NEGCACHED);
+    debugs(20, 6, "expires = " << expires << ", curtime = " << squid_curtime);
+    if (expires > squid_curtime)
+        EBIT_SET(flags, ENTRY_NEGCACHED);
```
rousskov: I do not insist on this change, but I think it would be better to move the new debugs() inside the `if` statement to both reduce noise and to ease detection of `ENTRY_NEGCACHED` entries. Reporting the number of seconds left may also be useful. Here is a sketch:

```suggestion
    
    if (expires > squid_curtime) {
        EBIT_SET(flags, ENTRY_NEGCACHED);
        debugs(20, 6, "expires = " << expires << " +" << (expires-squid_curtime) << ' ' << *this);
    }
```

src/store.cc
DrDaveD: I restored the no.negative_cached message to hardcode a '0', for easiest backporting.

src/store.cc
DrDaveD: The yes.negative_cached stat and corresponding code is removed.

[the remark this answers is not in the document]
DrDaveD: This section has been removed.

DrDaveD:
I believe I have addressed all comments, please re-review.

src/store.cc
yadij: This is fine by me.

yadij:
Regarding the FIXME in description. AIUI these are fine being cached anywhere because in HTTP/1.1 the 4xx and 5xx can be cached when Expires header (or Cache-Control) says expiry in future.

The status alone is not a reason to keep out of any store. They should *default* to already expired / non-cacheable (done by StoreEntry::negativeCache()). But when expires is in future for any valid reason they are just another cacheable response.

DrDaveD:
I replaced the FIXME with an explanation for why negative cached objects may be on disk.

yadij:
re-wrapped the description text (no changes) to avoid the failed tag.

rousskov:
> We allow ENTRY_NEGCACHED to be cached on disk because negative cached objects are fundamentally no different than positively cached objects. In HTTP/1.1 the 4xx and 5xx can be cached when the Expires header (or Cache-Control) says expiry is in the future.

This explains that the protocol allows Squid to cache negative responses on disk. What needs to be explained is why we are changing the "do not cache negative responses on disk" policy. Clearly, the original feature authors thought that those responses should _not_ be cached on disk -- they specifically added code to prevent such caching. It is very unlikely that those authors were driven by some misinterpreted protocol considerations. It is far more likely that they did not want to create (expensive) disk traffic for (short-lived) negative responses or had some other "optimization" in mind. The current explanation does not tell the reader why we want to change that long-standing policy/optimization.

I can propose the following explanation: `For simplicity sake, Squid should apply a uniform default disk admission policy to all cachable responses, negative and positive. If an admin does not want to cache negative responses on disk, then they should use FIXME to customize that default.`

Unfortunately, I am not sure how to fix the FIXME in the above paragraph -- I cannot think of an existing mechanism that an admin can use to restore the old default behavior (without prohibiting memory caching of negative responses). Did I miss it? If such a mechanism does not exist, do you still think that we should change the disk admission policy to allow negative responses?

DrDaveD:
I think it's ok to allow disk caching without an option to have the old behavior.  I don't see a good reason for it to be different than positive responses.   The caching might not be short-lived, it depends on what Expires or Cache-control: max-age or negative_ttl is set to.  If anything maybe there should be options to disable disk caching for any response that expires in less than a configurable amount of time or is smaller than a configurable size.

rousskov:
OK to test

rousskov:
[I submitted the following text when dismissing my negative review a few minutes ago. I do not know why GitHub has not (as of now) showed this text or the dismissal event itself. The review was successfully dismissed though.]

> I think it's ok to allow disk caching without an option to have the old behavior. I don't see a good reason for it to be different than positive responses.

While I agree with you in principle, a Squid admin that gets a lot of negative responses may feel differently when those responses start going to disk, overwhelming the disk subsystem and/or reducing disk hit ratio, etc.

If we were starting from scratch, I would insist on treating all cachable responses the same, but we are not starting from scratch. We have to take care of the existing deployments. We _can_ change old defaults, but making changes that cannot be undone via configuration should be done only when really necessary. I am not convinced the simplicity benefits outweigh the risks here.

I have adjusted the last paragraph in the PR description to better (IMHO) reflect the current decision, and I am removing me blocking vote. If Amos is happy to clear this PR for merging despite the now-disclosed risk of inconveniencing innocent admins, I will _not_ object.

PR #567, opened 2020-03-11
Happy Eyeballs: Do not discard viable reforwarding destinations

eduard-bagdasaryan:
When HappyConnOpener starts opening two connections, both destinations
are removed from the shared destinations list. As soon as one connection
(X) succeeded, the other destination (Y) was essentially forgotten. If
FwdState, after using X, decided to reforward the request, then the
request was never reforwarded to Y. We now return Y to the list.

Also abort the still-active ConnOpener job upon the termination of its
"parent" HappyConnOpener job. Without an explicit termination, those
abandoned child jobs wasted OS and Squid resources while the OS was
trying to open the requested connection. They would terminate on a
timeout or when the connection was finally opened (and discarded).

Waiting for the child ConnOpener job to end is a useful optimization,
but it remains a TODO. It requires complex accumulation avoidance logic.

Also fixed retrying via FwdState::fail() which could reorder addresses.
It incorrectly assumed that only the prime connections were retried.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -470,7 +470,7 @@ FwdState::fail(ErrorState * errorState)
     if (pconnRace == racePossible) {
         debugs(17, 5, HERE << "pconn race happened");
         pconnRace = raceHappened;
-        destinations->retryPath(serverConn);
+        destinations->retryPrimePath(serverConn);
     }
 
     if (ConnStateData *pinned_connection = request->pinnedConnection()) {
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -470,7 +470,7 @@ FwdState::fail(ErrorState * errorState)
     if (pconnRace == racePossible) {
         debugs(17, 5, HERE << "pconn race happened");
         pconnRace = raceHappened;
-        destinations->retryPrimePath(serverConn);
+        destinations->retryPath(serverConn);
     }
 
     if (ConnStateData *pinned_connection = request->pinnedConnection()) {
--- a/src/HappyConnOpener.cc
+++ b/src/HappyConnOpener.cc
@@ -486,7 +486,7 @@ void
 HappyConnOpener::cancelPrimeAttempt(Attempt &attempt, const char *reason)
 {
     Must(attempt);
-    destinations->retryPrimePath(attempt.path);
+    destinations->retryPath(attempt.path);
     attempt.cancel(reason);
 }
 
@@ -496,7 +496,7 @@ void
 HappyConnOpener::cancelSpareAttempt(Attempt &attempt, const char *reason)
 {
     Must(attempt);
-    destinations->retrySparePath(attempt.path);
+    destinations->retryPath(attempt.path);
     attempt.cancel(reason);
 }
 
--- a/src/HappyConnOpener.cc
+++ b/src/HappyConnOpener.cc
@@ -480,8 +480,7 @@ HappyConnOpener::sendSuccess(const Comm::ConnectionPointer &conn, bool reused, c
     callback_ = nullptr;
 }
 
-/// if the given spare attempt is in progress, cancels it and re-inserts
-/// the path into the candidates list
+/// cancels the being in-progress attempt and re-inserts the path into the candidates list
 void
 HappyConnOpener::cancelAttempt(Attempt &attempt, const char *reason)
 {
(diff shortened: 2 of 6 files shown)


rousskov:
@kinkie, if you have not already, could you please check the following Jenkins [error](https://build.squid-cache.org/job/5-pr-test/1710/COMPILER=gcc,OS=centos-7,label=docker-build-host/console)?

```
16:31:01 /tmp/jenkins5435838095719442546.sh: 2: .: Can't open /home/jenkins/lib/jenkins/helper_functions.sh
```

eduard-bagdasaryan:
> need to return that spare into its rightful place

Fixed.

src/FwdState.cc
```diff
@@ -470,7 +470,7 @@ FwdState::fail(ErrorState * errorState)
     if (pconnRace == racePossible) {
         debugs(17, 5, HERE << "pconn race happened");
         pconnRace = raceHappened;
-        destinations->retryPath(serverConn);
+        destinations->retryPrimePath(serverConn);
```
rousskov: Technically, `serverConn` may be a spare address (from Happy Eyeballs point of view) that was connected first (before the prime address got connected or after the prime address failed to connect). I am concerned that by assuming that this is a prime address, we may reorder destination addresses for the same peer (if the peer still has addresses in both families left).

eduard-bagdasaryan: I agree that the retrying address may be 'spare' here but I doubt we can distinguish 'spare' from 'prime' at this point because this classification is wholly HappyConnOpener's business. We could not rely on grouping addresses of the same families in 'destinations' (i.e., place 'serverConn' into 'destinations' near other addresses of the same family). For example, if 'destinations' contains only IPv4 addresses (for the given destination) and we are retrying IPv6 address it is unclear whether to place it before or after IPv4 addresses (i.e., whether IPv6 was 'prime' or 'spare' when it was recently fetched to FwdState). It seems that the only way to overcome this difficulty is to store prime/spare attribute somewhere within the Connection itself.
Note that this problem is old and unrelated to the PR changes.

rousskov: Given the small amount/low complexity of the current PR changes, the likely significant overlap in final solutions, and the related nature of both bugs (i.e. discarding viable destinations and reordering viable destinations), I think we should fix both bugs in this PR.

> It seems that the only way to overcome this difficulty is to store prime/spare attribute somewhere within the Connection itself.

I am sure there are better solutions than polluting innocent and already overburdened Connection. I am also worried about preserving the "spare or prime" classification beyond a single HappyConnOpener instance lifetime -- the very last address for a peer ought to become "prime" if it reenters Happy Eyeballs on a retry, even if it was originally extracted as "spare".

----

How about _never removing_ candidates from the  ResolvedPeers::paths_  _and_ storing a boolean "available" marker (together with Connection) in ConnectionList::value_type?

The new ConnectionList::value_type can be called ResolvedPeerPath, for example (I do not recommend ResolvedPeer because that would falsely imply that we are storing a CachePeer pointer).

The existing ResolvedPeers methods will never return unavailable items. ResolvedPeers::extractFound() will set the new marker to false. The ResolvedPeers::retry*() methods will reset the marker back to true after searching for a matching Connection (using a raw pointer match -- the Connection object itself could not have changed because it is refcounted and never created in the contexts we are interested in, right?).

If needed, we can optimize candidate extraction searches by caching the offset of the last available candidate seen. Without that optimization, one could argue that the lack of "shift" in this solution may create a potentially long linear search. The counter argument is that very similar searches are already possible (e.g., when a spare address is behind a very long list of not-yet-tried prime addresses). I probably would not rush with this error-prone optimization unless we think it is needed.

eduard-bagdasaryan: Your idea looks good. Also, if we never remove elements and only append them into ResolvedPeers::paths_ , we can use std::vector instead of std::list as a container type.

eduard-bagdasaryan: Sorry, the container type is 'vector' already. I misinterpreted 'ConnectionList' type name...

eduard-bagdasaryan: > How about never removing candidates from the ResolvedPeers::paths_ and storing a boolean "available" marker 

Done (03f0268-09e4d7e).

eduard-bagdasaryan: I found bugs in the implementation and fixed them (up to 0d359e5).

rousskov:
I plan to clear this PR for merging within a few days.

PR #564, opened 2020-03-04
Supply ALE to request_header_add/reply_header_add

chtsanti:
Supply ALE to request_header_add and reply_header_add ACLs that need it
(e.g., external, annotate_client, and annotate_transaction ACLs). Fixes
"ACL is used in context without an ALE state" errors when external ACLs
are used in the same context (other ACLs do not yet properly disclose
that they need ALE).

Also provides HTTP reply to reply_header_add ACLs.

This is a Measurement Factory project.

PR #568, opened 2020-03-13
Supply ALE to request_header_add/reply_header_add

chtsanti:
Supply ALE to request_header_add and reply_header_add ACLs that need it
(e.g., external, annotate_client, and annotate_transaction ACLs). Fixes
"ACL is used in context without an ALE state" errors when external ACLs
are used in the same context (other ACLs do not yet properly disclose
that they need ALE).

Also provides HTTP reply to reply_header_add ACLs.

This is a v4 port of master commit b6f3c19a9.

### DIFF
--- a/src/HttpHeaderTools.cc
+++ b/src/HttpHeaderTools.cc
@@ -477,6 +477,12 @@ httpHdrAdd(HttpHeader *heads, HttpRequest *request, const AccessLogEntryPointer 
 {
     ACLFilledChecklist checklist(NULL, request, NULL);
 
+    checklist.al = al;
+    if (al && al->reply) {
+        checklist.reply = al->reply;
+        HTTPMSGLOCK(checklist.reply);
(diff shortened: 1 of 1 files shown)


rousskov:
I edited PR description to remove the unqualified (and unnecessary) reference to a GitHub-specific PR #564. We should be using commit SHAs instead (and the description already had that commit SHA).

This v4 PR is on Amos side.

yadij:
Ignoring clang internal error on Ubuntu (GCC works fine on same OS).

rousskov:
@yadij, please note that when we merge manually (which we should not do, even on non-master branches), we become responsible for crafting the commit message based on the PR description -- GitHub cannot do that for us. The best way to bypass test failures may be to (temporary) declare the failing test optional in the GitHub branch protection configuration so that Anubis can properly merge the PR.

PR #576, opened 2020-03-24
Bug 5036: capital 'L's in logs when daemon queue overflows

DrDaveD:
Varying number of capital 'L's have been observed in access logs
written by the daemon module.  The 'L's corresponded with the
cache log message:

  "queue is too large; some log messages have been lost."

This was caused by the fact that the 'L' command byte was
buffered even when the queue was too large to accept messages.

Deferring buffering the command byte until just before the
message itself is buffered.

### DIFF
--- a/src/log/ModDaemon.cc
+++ b/src/log/ModDaemon.cc
@@ -47,7 +47,6 @@ static void logfile_mod_daemon_append(Logfile * lf, const char *buf, int len);
 struct _l_daemon {
     int rfd, wfd;
     char eol;
-    char command_written;
     pid_t pid;
     int flush_pending;
     dlink_list bufs;
@@ -227,7 +226,6 @@ logfile_mod_daemon_open(Logfile * lf, const char *path, size_t, int)
     ll = static_cast<l_daemon_t*>(xcalloc(1, sizeof(*ll)));
     lf->data = ll;
     ll->eol = 1;
-    ll->command_written = 0;
     {
         Ip::Address localhost;
         args[0] = "(logfile-daemon)";
@@ -301,12 +299,13 @@ logfile_mod_daemon_writeline(Logfile * lf, const char *buf, size_t len)
         return;
     }
 
-    if (!ll->command_written) {
+    /* Are we eol? If so, prefix with our logfile command byte */
+    if (ll->eol) {
 	char tb[2];
 	tb[0] = 'L';
 	tb[1] = '\0';
 	logfile_mod_daemon_append(lf, tb, 1);
-	ll->command_written = 1;
+	ll->eol = 0;
     }
 
     /* Append this data to the end buffer; create a new one if needed */
@@ -316,9 +315,10 @@ logfile_mod_daemon_writeline(Logfile * lf, const char *buf, size_t len)
 static void
 logfile_mod_daemon_linestart(Logfile * lf)
 {
-    l_daemon_t *ll = static_cast<l_daemon_t *>(lf->data);
-    assert(ll->eol == 1);
-    ll->eol = 0;
+    /* 
+     * Sending the starting command byte has been moved to
+     * logfile_mod_daemon_writeline
+     */
 }
 
 static void
@@ -326,11 +326,7 @@ logfile_mod_daemon_lineend(Logfile * lf)
 {
     l_daemon_t *ll = static_cast<l_daemon_t *>(lf->data);
     logfile_buffer_t *b;
-    assert(ll->eol == 0);
     ll->eol = 1;
-    if (!ll->command_written)
-	return;
-    ll->command_written = 0;
     /* Kick a write off if the head buffer is -full- */
     if (ll->bufs.head != NULL) {
         b = static_cast<logfile_buffer_t*>(ll->bufs.head->data);
--- a/src/log/ModDaemon.cc
+++ b/src/log/ModDaemon.cc
@@ -349,3 +349,4 @@ logfile_mod_daemon_flush(Logfile * lf)
         return;
     }
 }
+
--- a/src/log/ModDaemon.cc
+++ b/src/log/ModDaemon.cc
@@ -301,10 +301,7 @@ logfile_mod_daemon_writeline(Logfile * lf, const char *buf, size_t len)
 
     /* Are we eol? If so, prefix with our logfile command byte */
     if (ll->eol) {
-	char tb[2];
-	tb[0] = 'L';
-	tb[1] = '\0';
-	logfile_mod_daemon_append(lf, tb, 1);
+	logfile_mod_daemon_append(lf, "L", 1);
 	ll->eol = 0;
     }
 
@@ -351,4 +348,3 @@ logfile_mod_daemon_flush(Logfile * lf)
         return;
     }
 }
-
--- a/src/log/ModDaemon.cc
+++ b/src/log/ModDaemon.cc
@@ -312,6 +312,7 @@ logfile_mod_daemon_writeline(Logfile * lf, const char *buf, size_t len)
 static void
 logfile_mod_daemon_linestart(Logfile * lf)
 {
+    l_daemon_t *ll = static_cast<l_daemon_t *>(lf->data);
     assert(ll->eol == 1);
     // logfile_mod_daemon_writeline() sends the starting command
 }
--- a/src/log/ModDaemon.cc
+++ b/src/log/ModDaemon.cc
@@ -320,6 +320,8 @@ logfile_mod_daemon_lineend(Logfile * lf)
 {
     l_daemon_t *ll = static_cast<l_daemon_t *>(lf->data);
     logfile_buffer_t *b;
+    if (ll->eol == 1) // logfile_mod_daemon_writeline() wrote nothing
+        return;
     ll->eol = 1;
     /* Kick a write off if the head buffer is -full- */
     if (ll->bufs.head != NULL) {
--- a/src/log/ModDaemon.cc
+++ b/src/log/ModDaemon.cc
@@ -47,6 +47,7 @@ static void logfile_mod_daemon_append(Logfile * lf, const char *buf, int len);
 struct _l_daemon {
     int rfd, wfd;
     char eol;
+    char command_written;
     pid_t pid;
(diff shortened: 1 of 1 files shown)


DrDaveD:
I thought of using `eol`, but it appeared to me that the purpose of the flag was to catch coding errors of mismatched linestart/lineend calls, and I didn't feel empowered to change that purpose.  Would we want to assert failure if lineend is called twice with no writeline between? That is what would happen if we don't change lineend.  At a minimum I don't think we would want to have `assert(ll->eol == 1)` in writeline, unless writeline is only allowed to be called once with each lineend (I'm not saying you suggested that, but it's a difference from current behavior).

Having said that, if you and Amos want me to change it as you suggest, I'll be happy to do that.

DrDaveD:
Actually, if `eol` became the flag used for this purpose, the assert would have to come out of linend because otherwise it would crash immediately after a buffer overflow.

rousskov:
I cannot tell for sure whether your interpretation of my sketch matches mine. I also cannot guarantee that my sketch will point you in the correct direction, even when interpreted correctly. I am fairly optimistic about the sketched solution, and I am usually right about these things, but I _do_ make mistakes, and this could be one of them.

It would be easier for me to implement the correct fix than to properly address your specific concerns about the sketch. Unfortunately, I cannot do that right now. It is, of course, your call whether to try implementing the sketch.

P.S. The wrong `Are we eol?...` source code comment that your PR has removed essentially matches the _intent_ behind my sketch. I suspect the original author of this Squid2 code planned to do something similar but those plans did not materialize.

rousskov:
> Actually, if `eol` became the flag used for this purpose, the assert would have to come out of linend because otherwise it would crash immediately after a buffer overflow.

Yes, that assert will become wrong. If we did nothing before calling lineend, lineend should do nothing as well. I adjusted the third item in the sketch. Please keep in mind that it is just a sketch -- it is likely to get low-level details like this wrong.

DrDaveD:
I made the requested changes, please review.

src/log/ModDaemon.cc
```diff
@@ -298,29 +298,34 @@ logfile_mod_daemon_writeline(Logfile * lf, const char *buf, size_t len)
...
+     */
 }
 
 static void
 logfile_mod_daemon_lineend(Logfile * lf)
 {
     l_daemon_t *ll = static_cast<l_daemon_t *>(lf->data);
     logfile_buffer_t *b;
-    assert(ll->eol == 0);
     ll->eol = 1;
```
rousskov: Would it be possible to do nothing if `ll->eol` is still true? If yes, please do that instead of just removing the no-longer-valid assertion. Here is a sketch:

```C++
logfile_buffer_t *b;
if (ll->eol == 1) // logfile_mod_daemon_writeline() wrote nothing
    return;
```

Technically, the C-style `b` declaration should be moved into the `if` statement instead of being left where it is, but that polishing change would obscure the PR diff too much IMO. Thus, I do not recommend moving it. I wrote it above to show the placement of the new `if` statement (replacing the old assertion).

src/log/ModDaemon.cc
```diff
@@ -298,29 +298,34 @@ logfile_mod_daemon_writeline(Logfile * lf, const char *buf, size_t len)
...
-    char tb[2];
-    assert(ll->eol == 1);
-    ll->eol = 0;
-    tb[0] = 'L';
-    tb[1] = '\0';
-    logfile_mod_daemon_append(lf, tb, 1);
+    /* 
+     * Sending the starting command byte has been moved to
+     * logfile_mod_daemon_writeline
+     */
```
rousskov: Let's document the resulting code rather than the changes that lead to it:

```suggestion
    // logfile_mod_daemon_writeline() sends the starting command
```

This will also fix a CI [failure](https://semaphoreci.com/squid/squid/branches/pull-request-576/builds/2).

----

P.S. You may receive two copies of this change request. AFAICT, GitHub ate my first version, so I am resubmitting it.

src/log/ModDaemon.cc
```diff
@@ -298,29 +298,34 @@ logfile_mod_daemon_writeline(Logfile * lf, const char *buf, size_t len)
...
+    /* Append this data to the end buffer; create a new one if needed */
     logfile_mod_daemon_append(lf, buf, len);
 }
 
 static void
 logfile_mod_daemon_linestart(Logfile * lf)
 {
-    l_daemon_t *ll = static_cast<l_daemon_t *>(lf->data);
-    char tb[2];
-    assert(ll->eol == 1);
```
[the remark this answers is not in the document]
DrDaveD: I don't see how that can be done, unless we're sure that it is not allowed to do more than one logfile_mod_daemon_writeline between each call to logfile_mod_daemon_lineend.

src/log/ModDaemon.cc
DrDaveD: Done

src/log/ModDaemon.cc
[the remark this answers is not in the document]
yadij: Simply keeping the assert as-was should be all that is necessary. If it triggers there is a bug to fix.

The valid state sequence is:
 linestart [ writeline writeline ... ] lineend 

Once a line has been started with "L" any call to linestart() is not valid until lineend() has put the terminator on.

yadij:
I have had to shorten the title a lot to meet our infrastructure requirements. Also formatted the description a bit better.

src/log/ModDaemon.cc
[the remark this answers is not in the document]
DrDaveD: Oh!  I was thinking you meant to leave it with sending the 'L', my bad.  Ok, done.

src/log/ModDaemon.cc
```diff
@@ -298,29 +298,31 @@ logfile_mod_daemon_writeline(Logfile * lf, const char *buf, size_t len)
...
+    }
+
+    /* Append this data to the end buffer; create a new one if needed */
     logfile_mod_daemon_append(lf, buf, len);
 }
 
 static void
 logfile_mod_daemon_linestart(Logfile * lf)
 {
-    l_daemon_t *ll = static_cast<l_daemon_t *>(lf->data);
```
yadij: This line also needs restoring to fix the build errors

DrDaveD: Of course, I'm sorry I didn't notice that myself.  Done.

rousskov:
OK to test

rousskov:
The latest Jenkins error for the latest staged commit that I see is:

```
FATAL: Could not checkout d2b024ed75a1ca232c01a2f47375957ba6c1203d
fatal: reference is not a tree: d2b024ed75a1ca232c01a2f47375957ba6c1203d
```

AFAICT, the error is correct -- that staged commit was replaced by the staged commit from PR #579 . However, that other staged [commit](https://github.com/squid-cache/squid/commit/8e52f54668c5d702dc4d3d05c1b6daa5b0750522) (or its successor) was having similar Jenkins [problems](https://build.squid-cache.org/job/5-pr-auto/COMPILER=gcc,OS=ubuntu-bionic,label=docker-build-host/588/console):

```
FATAL: Could not checkout 8e52f54668c5d702dc4d3d05c1b6daa5b0750522
fatal: reference is not a tree: 8e52f54668c5d702dc4d3d05c1b6daa5b0750522
```

This switcheroo may be perfectly "normal" if Anubis had a legitimate reason to abandon the staged commit for this PR for the staged commit of another PR. However, I would like to double check that there is no Anubis problem here that leads to something like an infinite staging loop, with several PRs competing for the single auto spot. @eduard-bagdasaryan, is it possible to explain why Anubis abandoned this older PR 576 in favor of the younger PR 579?

rousskov:
And as I was writing the previous comment, this PR was staged for testing. It may be important to note that the last staged commit (4e48cb7) of the competing PR has failed with the usual error but still has unfinished Jenkinks tests and is still labeled `M-waiting-staging-checks`. After reloading both GitHub pages, I see two `M-waiting-staging-checks` PRs which I did not think was possible! @eduard-bagdasaryan, please check.

eduard-bagdasaryan:
> However, that other staged commit (or its successor) was having similar Jenkins problems:

I am afraid that these are not Jenkins problems because I experience the same "reference is not a tree" when trying to check out the commit locally. I stopped Anubis from creating numerous auto commits until I can figure out what is going on.

eduard-bagdasaryan:
I found the problem: someone increased the number of checks passed on Jenkins for a staged commit (8 nodes now, and 8x2=16 checks). Anubis verifies that the actual number of statuses <=staging_checks, (which is 6 now) and aborts PR otherwise (the cause of M-failed-other).  We need to keep these Jenkins and Anubis configuration parameters in sync.

rousskov:
> someone increased the number of checks passed on Jenkins for a staged commit (8 nodes now, and 8x2=16 checks).

I assume that you mean that somebody adjusted Jenkins configuration and the result (or side effect) of that change was adding more Jenkins tests/checks for the auto branch. Please correct my understanding if it is wrong.

> We need to keep these Jenkins and Anubis configuration parameters in sync.

Assuming the increase in the number of Jenkins tests is desirable, would you be able to adjust Anubis configuration to match the current Jenkins setup? Please let me know if you need my help with that.

eduard-bagdasaryan:
> adjust Anubis configuration to match the current Jenkins setup

Already done.

eduard-bagdasaryan:
> (8 nodes now, and 8x2=16 checks).

There are actually 15 Jenkins checks: 'ubuntu-xenial' for 'clang' is disabled.

rousskov:
@eduard-bagdasaryan, thank you for addressing this problem!

@DrDaveD, your changes are now official and will be available in v6 (at least). Thank you!

eduard-bagdasaryan:
> We need to keep these Jenkins and Anubis configuration parameters in sync.

As I wrote above, currently we have 15 [configured]( https://build.squid-cache.org/job/5-pr-auto/603/) Jenkins tests for our 'auto' branch. This parameter is important for Anubis and is manually configured. @kinkie , @yadij , could you please inform me about such Jenkins configuration changes? Note that only the number of active checks matters.

PR #577, opened 2020-03-25
Fix clang-6.0 build on FreeBSD 12 [-Wformat]

kinkie:
Cast and print std::chrono:nanoseconds::count() as intmax_t because its
actual type might be larger than the 64 bytes expected by %PRId64.

The problem was not, technically, specific to clang or FreeBSD.

The error message says "long" because %PRId64 is %ld on platforms where
"long" is 64 bits:

    format specifies type 'long' but the argument has type [...]
    'long long' [-Wformat]

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3012,7 +3012,6 @@ AC_CHECK_TYPE([bool])
 #need the defines for PRId64
 AC_CHECK_SIZEOF(int64_t)
 AC_CHECK_SIZEOF(long)
-AC_CHECK_SIZEOF(long long)
 #need the defines for PRIuSIZE
 AC_CHECK_SIZEOF(size_t)
 #need the define for overflow checks
--- a/configure.ac
+++ b/configure.ac
@@ -3012,6 +3012,7 @@ AC_CHECK_TYPE([bool])
 #need the defines for PRId64
 AC_CHECK_SIZEOF(int64_t)
 AC_CHECK_SIZEOF(long)
+AC_CHECK_SIZEOF(long long)
 #need the defines for PRIuSIZE
 AC_CHECK_SIZEOF(size_t)
 #need the define for overflow checks
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -3072,8 +3072,8 @@ free_time_msec(time_msec_t * var)
 static void
 dump_time_nanoseconds(StoreEntry *entry, const char *name, const std::chrono::nanoseconds &var)
 {
-    // This assumes that nothing in C++ is larger than long long. TODO: Avoid printf().
-    storeAppendPrintf(entry, "%s %lld nanoseconds\n", name, static_cast<long long>(var.count()));
+    // std::chrono::nanoseconds::rep is unknown a priori so we cast to (and print) the largest supported integer
+    storeAppendPrintf(entry, "%s %jd nanoseconds\n", name, static_cast<intmax_t>(var.count()));
 }
 
 static void
@@ -5110,4 +5110,3 @@ free_on_unsupported_protocol(acl_access **access)
 {
     free_acl_access(access);
 }
-
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -3072,7 +3072,7 @@ free_time_msec(time_msec_t * var)
 static void
 dump_time_nanoseconds(StoreEntry *entry, const char *name, const std::chrono::nanoseconds &var)
 {
-    storeAppendPrintf(entry, "%s %" PRId64 " nanoseconds\n", name, var.count());
+    storeAppendPrintf(entry, "%s %" PRId64 " nanoseconds\n", name, static_cast<int64_t>(var.count()));
 }
 
 static void
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -3072,7 +3072,8 @@ free_time_msec(time_msec_t * var)
 static void
 dump_time_nanoseconds(StoreEntry *entry, const char *name, const std::chrono::nanoseconds &var)
 {
-    storeAppendPrintf(entry, "%s %" PRId64 " nanoseconds\n", name, static_cast<int64_t>(var.count()));
+    // This assumes that nothing in C++ is larger than long long. TODO: Avoid printf().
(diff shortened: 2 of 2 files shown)


configure.ac
```diff
@@ -3012,6 +3012,7 @@ AC_CHECK_TYPE([bool])
 #need the defines for PRId64
 AC_CHECK_SIZEOF(int64_t)
 AC_CHECK_SIZEOF(long)
+AC_CHECK_SIZEOF(long long)
```
yadij: You are not adding any SIZEOF_LONG_LONG macros to the code. Is this actually needed?

The other ones are for the libcompat definition:
``
#elif SIZEOF_INT64_T > SIZEOF_LONG
#define PRId64 "lld"
``

src/cache_cf.cc
```diff
@@ -3072,7 +3072,7 @@ free_time_msec(time_msec_t * var)
 static void
 dump_time_nanoseconds(StoreEntry *entry, const char *name, const std::chrono::nanoseconds &var)
 {
-    storeAppendPrintf(entry, "%s %" PRId64 " nanoseconds\n", name, var.count());
+    storeAppendPrintf(entry, "%s %" PRId64 " nanoseconds\n", name, static_cast<int64_t>(var.count()));
```
yadij: What type is FreeBSD using for this value?

FYI, it already smells that a counter is using a signed type. Most likely this should be displaying PRIu64 or similar.

Please at least add a comment indicating this cast is a hack and if possible for which OS version. So we can purge it in future when no longer needed.

kinkie:
```
cache_cf.cc:3075:68: error: format specifies type 'long' but the argument has type 'std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000000> >::rep' (aka 'long long') [-Werror,-Wformat]
    storeAppendPrintf(entry, "%s %" PRId64 " nanoseconds\n", name, var.count());

```
interestingly enough, both long and long long are 64-bit, but clang on freebsd automatically convert.
switching to PRIu64 doesn't help:
```
depbase=`echo cache_cf.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`; ccache clang++90 -DHAVE_CONFIG_H -DDEFAULT_CONFIG_FILE=\"/home/jenkins/squid/install/etc/squid.conf\"  -DDEFAULT_SQUID_DATA_DIR=\"/home/jenkins/squid/install/share\"  -DDEFAULT_SQUID_CONFIG_DIR=\"/home/jenkins/squid/install/etc\"    -I.. -I../include  -I../lib -I../src  -I../include -I/usr/local/include   -I../libltdl -I../src -I../libltdl   -I/usr/local/include/libxml2 -Werror -Qunused-arguments  -D_REENTRANT -I/usr/local/include/libxml2 -g -O2 -I/usr/local/include -MT cache_cf.o -MD -MP -MF $depbase.Tpo -c -o cache_cf.o cache_cf.cc && mv -f $depbase.Tpo $depbase.Po
cache_cf.cc:3075:68: error: format specifies type 'unsigned long' but the argument has type 'std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000000> >::rep' (aka 'long long') [-Werror,-Wformat]
    storeAppendPrintf(entry, "%s %" PRIu64 " nanoseconds\n", name, var.count());
                                 ~~~~~~~~~                         ^~~~~~~~~~~
1 error generated.
```

rousskov: A slightly better fix may be:

```suggestion
    // This assumes that nothing in C++ is larger than long long. TODO: Avoid printf().
    storeAppendPrintf(entry, "%s %lld nanoseconds\n", name, static_cast<long long>(var.count()));
```

The above may be slightly better because there is nothing larger than `long long` in today's C++. The current PR code might, in theory, truncate the count() value on rare platforms where `long long` is 128 bits rather than 64. All this is rather theoretical, of course, so I do not insist on this change.

rousskov:
I took the liberty to polish the PR description to detail what is going on here because the error itself and its combination with the fix was very confusing to me. Please undo or adjust further as needed.

src/cache_cf.cc
kinkie: according to the freebsd documentation, lld should work. I'm accepting your suggestion, althought I agree the scenario is very theoretical

kinkie: We used to support some pretty bizarre architectures, but by now there's just a few that are relevant.
I used the "Commit suggestion" button in github. No idea were that line came from

rousskov: > I used the "Commit suggestion" button in github. No idea were that line came from

Must be a GitHub bug then. I [re-added](https://github.com/squid-cache/squid/pull/577/commits/82c70e9b2996ba749a1ef34e55279c23f42efc8a) the line that GitHub removed.

AFAICT, the ball is fully on @yadij side now.

kinkie:
@yadij , as the changes you requested are in, would you mind approving?

src/cache_cf.cc
```diff
@@ -3072,7 +3072,8 @@ free_time_msec(time_msec_t * var)
 static void
 dump_time_nanoseconds(StoreEntry *entry, const char *name, const std::chrono::nanoseconds &var)
 {
-    storeAppendPrintf(entry, "%s %" PRId64 " nanoseconds\n", name, var.count());
+    // This assumes that nothing in C++ is larger than long long. TODO: Avoid printf().
+    storeAppendPrintf(entry, "%s %lld nanoseconds\n", name, static_cast<long long>(var.count()));
```
rousskov: Sorry. While syncing the PR description with the PR code, I realized that we may be able to do even better:

```suggestion
    // std::chrono::nanoseconds::rep is unknown a priori so we cast to (and print) the largest supported integer
    storeAppendPrintf(entry, "%s %jd nanoseconds\n", name, static_cast<intmax_t>(var.count()));
```

The above should work in all cases as long as the underlying printf() supports `%jd` which was added to [C++11](https://en.cppreference.com/w/cpp/io/c/fprintf). Again, I do not know how portable that is in Squid-relevant environments.

If you accept this suggestion, please adjust the PR description accordingly. If you reject it, please check that you are OK with the current PR description.

rousskov: There [is](https://en.cppreference.com/w/c/types/integer) also `PRIdMAX`. I do not know whether it is any more portable than raw `%jd`.

PR #583, opened 2020-03-29
Bug 5038: Introduce allow_small_pids option

larsks:
The allow_small_pids option permits commands like "squid -k rotate" to
operate correctly even when the squid process is PID 1 (as is commonly
found when running in containerized environments). The option
defaults to "off", preserving the existing behavior unless it is
explicitly enabled.

### DIFF
--- a/src/Instance.cc
+++ b/src/Instance.cc
@@ -80,7 +80,7 @@ GetOtherPid(File &pidFile)
 
     debugs(50, 7, "found PID " << rawPid << " in " << TheFile);
 
-    if (rawPid <= 1)
+    if ((rawPid <= 1 && !Config.onoff.allow_small_pids) || (rawPid < 1))
         throw TexcHere(ToSBuf("Bad ", TheFile, " contains unreasonably small PID value: ", rawPid));
     const auto finalPid = static_cast<pid_t>(rawPid);
     if (static_cast<int64_t>(finalPid) != rawPid)
--- a/src/SquidConfig.h
+++ b/src/SquidConfig.h
@@ -277,6 +277,7 @@ public:
     } Netdb;
 
     struct {
+        int allow_small_pids;
         int log_udp;
         int res_defnames;
         int anonymizer;
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -2828,6 +2828,18 @@ DOC_START
 	See http://wiki.squid-cache.org/SquidFaq/SquidAcl for details.
 DOC_END
 
+NAME: allow_small_pids
+TYPE: onoff
+DEFAULT: off
+LOC: Config.onoff.allow_small_pids
+DOC_START
+    If allow_small_pids is off (the default), "squid -k" actions will
+    fail if squid has a pid of 1. This prevents erroneously sending
+    signals to init. If you are running squid in a containerized 
+    environment where it is actually pid 1, set allow_small_pids on
+    to permit "squid -k" commands to operate as intended.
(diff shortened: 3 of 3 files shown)


squidadm:
FTR: the PID check is to prevent services triggering signals directly on the PID 1 init system daemon. It is dangerous to allow high level applications like Squid, with network access to be signalling there.

I believe we should be automatically checking for container situations and allowing signals only if PID 1 is confirmed to be a Squid. Not leaving it to configuration.

yadij:
FTR: the PID check is to prevent services triggering signals directly on the PID 1 init system daemon. It is dangerous to allow high level applications like Squid, with network access to be signalling there.

I believe we should be automatically checking for container situations and allowing signals only if PID 1 is confirmed to be a Squid. Not leaving it to configuration.

larsks:
> FTR: the PID check is to prevent services triggering signals directly on the PID 1 init system daemon.

Yes, I mentioned that specifically in the documentation included in this patch.

> I believe we should be automatically checking for container situations and allowing signals only if PID 1 is confirmed to be a Squid.

That's great if the check is relatively generic, but if it means being aware of specific container runtimes, I think it's better just to leave this up to the operator.

rousskov:
I do not fully understand the threat model behind the existing check, but if Squid is trying to protect the system by refusing to signal the init daemon, perhaps Squid should also protect the system by refusing to _run_ as an init daemon? Clearly, Squid is not meant to be an init daemon... If Squid never runs with PID=1, then there will be no need for an additional configuration option!

I bet container creators can (and probably should) set things up so that Squid is not the first process that runs. We can also teach Squid to immediately fork another process if Squid is started with PID=1.

larsks:
> I bet container creators can (and probably should) set things up so that Squid is not the first process that runs. We can also teach Squid to immediately fork another process if Squid is started with PID=1.

This all seems like a lot of extra to work, involving coordinating with the developers of multiple container runtimes *and* the people designing container images. Why not just use a simple configuration option that someone can toggle? It's simple, and it doesn't change the default behavior.

rousskov:
> Why not just use a simple configuration option that someone can toggle?

... because every configuration option increases long-term development and maintenance costs. If everybody who wanted a configuration option got one, Squid would be completely unusable today.

Please note that the "if you can accomplish the desired effect without adding a configuration option, consider avoiding the configuration option" principle does not mean that the option you proposed should not be added! It only means that we should carefully consider the alternative solutions (if any). It also implies that we may want to add a more general "running in a container" option so that one option can tune multiple behavioral aspects.

To avoid misunderstanding, I am not voting for or against your PR. I am seeing a strange exception in Squid code and a strange configuration option disabling that strange exception. Both could be doing the right thing, but I am hoping there is an overall better solution here. If there is not, then your configuration option (or something similar) should be accepted.

patrickscottbest:
My use case for this requirement in containers is a reload of the configuration.  What are the ramifications of "kill -HUP 1".  Is there a different signal that should be used to reconfigure?

rousskov:
@larsks, sorry I missed your [question](https://github.com/squid-cache/squid/pull/583#issuecomment-667584860). I did not get a GitHub notification for that 2020 comment :-(.

FWIW, SIGHUP is the only signal that Squid uses for reconfiguration. I believe that does not really matter here though (for the reasons I detailed earlier). In summary, if we want to protect the OS from using Squid as an init daemon, then we should do that even if a different signal is used for reconfiguration.

PR #585, opened 2020-03-31
Fix auth digest refcount integer overflow

desbma-s1n:
This fixes a possible overflow of the nonce reference counter in the 
digest authentication scheme, found by security researchers 
@synacktiv.

It changes `references` to be an 64 bits unsigned integer. This makes 
overflowing the counter impossible in practice.

### DIFF
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -96,9 +96,6 @@ static void authenticateDigestNonceDelete(digest_nonce_h * nonce);
 static void authenticateDigestNonceSetup(void);
 static void authDigestNonceEncode(digest_nonce_h * nonce);
 static void authDigestNonceLink(digest_nonce_h * nonce);
-#if NOT_USED
-static int authDigestNonceLinks(digest_nonce_h * nonce);
-#endif
 static void authDigestNonceUserUnlink(digest_nonce_h * nonce);
 
 static void
@@ -291,20 +288,9 @@ authDigestNonceLink(digest_nonce_h * nonce)
 {
     assert(nonce != NULL);
     ++nonce->references;
+    assert(nonce->references != 0); // no overflows
     debugs(29, 9, "nonce '" << nonce << "' now at '" << nonce->references << "'.");
 }
-
-#if NOT_USED
-static int
-authDigestNonceLinks(digest_nonce_h * nonce)
-{
-    if (!nonce)
-        return -1;
-
-    return nonce->references;
-}
-
-#endif
 
 void
 authDigestNonceUnlink(digest_nonce_h * nonce)
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -43,7 +43,6 @@
  */
 #include "mem/Pool.h"
 
-#include <climits>
 #include <random>
 
 static AUTHSSTATS authenticateDigestStats;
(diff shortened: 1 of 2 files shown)


rousskov:
OK to test

yadij:
Typically a nonce reference count should not exceed 7 (client HTTP request, server HTTP request, ICAP request, ICAP response, HTTP cache, auth credentials cache, transaction ALE) - plus or minus a few transaction state objects.

2^16 is far larger than this particular counter should ever reach, even assuming the admin sets some reasonably large re-use count for nonce'. I think we would be better served adding overflow checks rather than increasing the limit like this.  That can be achieved by converting this struct to a class and adding a RefCount Pointer rather than this ad-hoc reference counting (which is long overdue anyway).

If you wish to take that on please do. It is relatively simple work, though time consuming to find all the code lines needing tweaks for the new ref-count API.

rousskov:
> I think we would be better served adding overflow checks rather than increasing the limit like this.

What makes overflow checks superior to a never-overflowing counter?

> If you wish to take that on please do.

FWIW, I am not interested in doing that work at this time, but can undo my polishing commits if they are in the way.

yadij:
> > I think we would be better served adding overflow checks rather than increasing the limit like this.
> 
> What makes overflow checks superior to a never-overflowing counter?

I this case; detection of suspicious activity.

> 
> > If you wish to take that on please do.
> 
> FWIW, I am not interested in doing that work at this time, but can undo my polishing commits if they are in the way.

They are not a problem either way.

FTR; If none wants to take on the full update I am okay with this going in as-is.

desbma-s1n:
> 2^16 is far larger than this particular counter should ever reach, even assuming the admin sets some reasonably large re-use count for nonce'.

We have developed a proof of concept internally.
We send a lot a requests to cause the overflow of the signed integer which is undefined behavior but typically causes the integer to wrap around to -2^16.
Then the code skips the decrementations because of the `< 0` check in `authDigestNonceUnlink`. We can carefully control when the counter reaches 0 again, to free the nonce while a request is still in flight, which can be exploited to cause a use after free.

> Please see if you agree with them and revert or adjust further as needed.

Looks good to me although the 64 bits are probably overkill.

To reach such value would require a number of open file descriptors that is simply not possible.
By default squid is compiled with `with-filedescriptors=65536`, that number can be increased so is for example the default Linux ulimit but can any OS really handle 2^32 file descriptors?

desbma-s1n:
> If you accept my changes, we will need to adjust the PR description accordingly. That description will become the future commit message (and needs reformatting anyway).

I have updated the description as requested.

> Unlike your initial version, I suspect that my change will increase the nonce structure size a bit, but I do not think that increase is important. If it is important, then the whole structure should be refactored to reduce waste anyway.

On 64 bits architectures, it should not change the struct layout, as both the field before and after are already 64bits wide.

rousskov:
> Alex: What makes overflow checks superior to a never-overflowing counter?

> Amos: detection of suspicious activity.

IMO, comparable detection is possible and, in fact, is easier to do correctly (as well as report more useful details) when there are _no_ overflows.

> Amos: Typically a nonce reference count should not exceed 7

> Author: We send a lot a requests to cause the overflow of the signed integer

Clearly, the possible range of this particular counter is rather uncertain/mysterious. Unless there is a provable certainty that this counter cannot exceed the number of file descriptors, even in the presence of transaction leaks and other "minor" bugs in the surrounding code, I think it is best to go with a 64-bit counter that cannot overflow even under those special circumstances. Fewer things to worry about...

With Amos blessing, I am clearing this PR for merging. If automated tests pass, the PR will be auto-merged in a few days. @desbma-s1n, thanks again for working on this bug!

PR #586, opened 2020-03-31
Restore PURGE miss replies to be "404 Not Found", not "0 Init"

chtsanti:
Since commit 6956579, PURGE requests resulted in invalid HTTP responses
with zero status code when Store lacked both GET and HEAD entries with
the requested URI.

Also adjusted Http::StatusLine packing code to avoid generating similar
invalid responses in the future (and to attract developer attention to
their presence in the code logic with a BUG message).

### DIFF
--- a/src/http/StatusLine.cc
+++ b/src/http/StatusLine.cc
@@ -48,14 +48,16 @@ Http::StatusLine::packInto(Packable * p) const
     assert(p);
 
     auto packedStatus = status();
+    auto packedReason = reason();
+
     if (packedStatus == Http::scNone) {
         static unsigned int reports = 0;
         if (++reports <= 100)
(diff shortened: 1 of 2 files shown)


src/http/StatusLine.cc
```diff
@@ -47,6 +47,17 @@ Http::StatusLine::packInto(Packable * p) const
 {
     assert(p);
 
+    auto packedStatus = status();
+    auto packedReason = reason();
+
+    if (packedStatus == Http::scNone) {
+        static unsigned int reports = 0;
+        if (++reports <= 100)
+            debugs(57, DBG_IMPORTANT, "BUG: Generated response lacks status code");
```
xeron, line 56: https://bugs.squid-cache.org/show_bug.cgi?id=5057

PR #589, opened 2020-04-03
Remove dlink from HttpHdrSc

yadij:
Resolves Coverity issue 1461128 Resource leak. Which is a false positive
due to analysis inability to track dlink pointers.

### DIFF
--- a/src/HttpHdrSc.h
+++ b/src/HttpHdrSc.h
@@ -11,7 +11,6 @@
 
 #include "http/forward.h"
 #include "HttpHdrScTarget.h"
-#include "mem/PoolingAllocator.h"
 #include "SquidString.h"
 
 #include <list>
--- a/src/HttpHdrSc.h
+++ b/src/HttpHdrSc.h
@@ -31,8 +31,6 @@ public:
     void updateStats(StatHist *) const;
     HttpHdrScTarget * getMergedTarget(const char *ourtarget); // TODO: make const?
     void setMaxAge(char const *target, int max_age);
-    void addTarget(HttpHdrScTarget *t);
-    void addTargetAtTail(HttpHdrScTarget *t);
 
 private:
     HttpHdrScTarget * findTarget (const char *target);
--- a/src/HttpHdrSc.h
+++ b/src/HttpHdrSc.h
@@ -11,7 +11,7 @@
 
 #include "http/forward.h"
 #include "HttpHdrScTarget.h"
-#include "mem/forward.h"
+#include "mem/PoolingAllocator.h"
 #include "SquidString.h"
 
 #include <list>
@@ -35,7 +35,7 @@ public:
 private:
     HttpHdrScTarget * findTarget (const char *target);
 
-    std::list<HttpHdrScTarget> targets;
+    std::list<HttpHdrScTarget, PoolingAllocator<HttpHdrScTarget>> targets;
 };
 
 /* Http Surrogate Control Header Field */
--- a/src/HttpHdrSc.h
+++ b/src/HttpHdrSc.h
@@ -9,23 +9,16 @@
 #ifndef SQUID_HTTPHDRSURROGATECONTROL_H
 #define SQUID_HTTPHDRSURROGATECONTROL_H
 
-#include "dlink.h"
+#include "http/forward.h"
+#include "HttpHdrScTarget.h"
 #include "mem/forward.h"
 #include "SquidString.h"
 
-class HttpHdrScTarget;
+#include <list>
+
 class Packable;
 class StatHist;
 class StoreEntry;
-
-typedef enum {
-    SC_NO_STORE,
-    SC_NO_STORE_REMOTE,
-    SC_MAX_AGE,
-    SC_CONTENT,
-    SC_OTHER,
-    SC_ENUM_END /* also used to mean "invalid" */
-} http_hdr_sc_type;
 
 /* http surogate control header field */
 class HttpHdrSc
@@ -33,10 +26,6 @@ class HttpHdrSc
     MEMPROXY_CLASS(HttpHdrSc);
 
 public:
-    HttpHdrSc(const HttpHdrSc &);
-    HttpHdrSc() {}
-    ~HttpHdrSc();
-
     bool parse(const String *str);
     void packInto(Packable * p) const;
     void updateStats(StatHist *) const;
@@ -45,10 +34,10 @@ public:
     void addTarget(HttpHdrScTarget *t);
     void addTargetAtTail(HttpHdrScTarget *t);
 
-    dlink_list targets;
 private:
     HttpHdrScTarget * findTarget (const char *target);
 
+    std::list<HttpHdrScTarget> targets;
 };
 
 /* Http Surrogate Control Header Field */
--- a/src/HttpHdrScTarget.h
+++ b/src/HttpHdrScTarget.h
@@ -10,8 +10,10 @@
 #define SQUID_HTTPHDRSURROGATECONTROLTARGET_H
 
 #include "defines.h" //for bit mask operations
-#include "HttpHdrSc.h"
+#include "http/forward.h"
+#include "SquidString.h"
 
+class HttpHdrSc;
 class Packable;
 class StatHist;
 class StoreEntry;
@@ -30,11 +32,11 @@ public:
     static const int MAX_AGE_UNSET=-1; //max-age is unset
     static const int MAX_STALE_UNSET=0; //max-stale is unset
 
-    HttpHdrScTarget(const char *target_):
+    explicit HttpHdrScTarget(const char *target_) :
         mask(0), max_age(MAX_AGE_UNSET), max_stale(MAX_STALE_UNSET),target(target_) {}
-    HttpHdrScTarget(const String &target_):
+    explicit HttpHdrScTarget(const String &target_) :
         mask(0), max_age(MAX_AGE_UNSET), max_stale(MAX_STALE_UNSET),target(target_) {}
-    HttpHdrScTarget(const HttpHdrScTarget &t):
+    explicit HttpHdrScTarget(const HttpHdrScTarget &t) :
         mask(t.mask), max_age(t.max_age), max_stale(t.max_stale),
         content_(t.content_), target(t.target) {}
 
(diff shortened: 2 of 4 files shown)


src/HttpHdrSc.h
```diff
@@ -45,10 +34,10 @@ class HttpHdrSc
     void addTarget(HttpHdrScTarget *t);
     void addTargetAtTail(HttpHdrScTarget *t);
 
-    dlink_list targets;
 private:
     HttpHdrScTarget * findTarget (const char *target);
 
+    std::list<HttpHdrScTarget> targets;
```
rousskov, line 40: Thank you for working on this Coverity defect! FWIW, Factory is looking through the recent Coverity report, focusing on defects introduced by our changes.

This is a partial PR review. IMO, we should not add memory allocations (for std::list element wrappers) when parsing or copying typical headers. I also doubt that we should reduce allocation overheads by adding an allocator to std::list because I do not see a lot of operations that a list is optimized for.

I have not analyzed all the details, but I would start by considering mimicking c3b51d6 changes: Using std::vector with a Squid-specific allocator. Most likely, the initial capacity of that vector should not be as large as the one in HttpHeader::entries. If most HttpHdrSc::targets use cases have zero or one target, then I would investigate whether popular std::vector implementations would do the right thing by _default_, so that we do not have to reserve any specific capacity.

yadij: > Thank you for working on this Coverity defect! FWIW, Factory is looking through the recent Coverity report, focusing on defects introduced by our changes.
> 

From what I have seen many of these double-free issues are false-positives due to the analysis inability to track allocate/deallocate cycles properly in RefCount.

> This is a partial PR review. IMO, we should not add memory allocations (for std::list element wrappers) when parsing or copying typical headers. I also doubt that we should reduce allocation overheads by adding an allocator to std::list because I do not see a lot of operations that a list is optimized for.
> 

I was hoping to make this list a set of Pointers. But have not yet tracked down all the places those need to be used to prevent double-free and such refcount issues. It may also be counter productive to move to Pointer for coverity until we fix those RefCount false postives better.

> I have not analyzed all the details, but I would start by considering mimicking [c3b51d6](https://github.com/squid-cache/squid/commit/c3b51d64909335642672ca01dc6da9526187d36d) changes: Using std::vector with a Squid-specific allocator.

Minimizing alloc is why list is used here. vector is worse because majority of push operations are to the ends - but do mix both front/back so we cannot optimize towards one specific end. I have optimized with move semantics as much as I can see so far (if you can find more uses, welcome), so we are not hitting double allocs which STL would be doing otherwise.

The commit you referred to needed one because it was a raw-pointer array being stored. These classes are MEMPROXY already, so using Squid pools by default without need of a special allocator.

> Most likely, the initial capacity of that vector should not be as large as the one in HttpHeader::entries. If most HttpHdrSc::targets use cases have zero or one target, then I would investigate whether popular std::vector implementations would do the right thing by _default_, so that we do not have to reserve any specific capacity.

There is not much information about use-cases. I have seen only a few installations and there was a mix of different numbers up to 5 targets across those. I imagine you are right about the 0-1, but not sure enough to want optimizing for it. In the case of 0, there should be no HttpHdrSc obect at all - so nothing to optimize.

rousskov: > Alex: we should not add memory allocations (for std::list element wrappers) 

> Amos: Minimizing alloc is why list is used here.  ... These classes are MEMPROXY already, so using Squid pools by default without need of a special allocator.

I cannot find a way to reconcile your response with how std::list works because `std::list<T>` allocates memory for _each_ element T _wrapper_. Not all elements at once (like vector). And not each element T (like the official code). Naturally, those internal STL allocations cannot use Squid pools by default. Moreover, MEMPROXY API itself is not compatible with STL Allocators API. AFAIK, the only Squid pooling allocator that we can use with std::list is PoolingAllocator (which we added in the same commit I have mentioned earlier -- c3b51d6).

* dlist: one MEMPROXY allocation of each HttpHdrScTarget (without re-allocations)
* std::list: one raw allocation of each HttpHdrScTarget wrapper (without re-allocations)
* std::list with PoolingAllocator: one pooled allocation of each HttpHdrScTarget wrapper (without re-allocations)
* std::vector: one raw allocation for a few HttpHdrScTargets (with re-allocations as necessary)
* std::vector with PoolingAllocator:: one pooled allocation for a few HttpHdrScTargets (with re-allocations as necessary)

The first bullet is the official code. The second bullet is the PR I reviewed.

In summary: PR removes HttpHdrScTarget pooling. We can fix that performance regression by using a PoolingAllocator (from c3b51d6), but before we proceed with std::list as the replacement container, we should consider all viable container options.

> Alex: If most HttpHdrSc::targets use cases have zero or one target, then I would investigate whether popular std::vector implementations would do the right thing by _default_, so that we do not have to reserve any specific capacity.
 
> Amos: There is not much information about use-cases. I have seen only a few installations and there was a mix of different numbers up to 5 targets across those. I imagine you are right about the 0-1, but not sure enough to want optimizing for it. In the case of 0, there should be no HttpHdrSc obect at all - so nothing to optimize.

Looks like we are in agreement on this: If we anticipate 0 or 1 targets in most cases, _and_ we decide to go with std::vector as the storage container,  _and_ std::vector does the right thing for 0 and 1 elements by default, then we do not need to manually reserve that vector capacity. The default would work just fine.

rousskov: @yadij, I am not sure, but it looks like you asked for my review before you saw my [response](https://github.com/squid-cache/squid/pull/589#discussion_r403623773) on this thread. I see no changes addressing this change request (as of fa83143). Please let me know if I missed something.

src/HttpHdrScTarget.h
```diff
@@ -30,11 +32,11 @@ class HttpHdrScTarget
...
-    HttpHdrScTarget(const char *target_):
+    explicit HttpHdrScTarget(const char *target_) :
         mask(0), max_age(MAX_AGE_UNSET), max_stale(MAX_STALE_UNSET),target(target_) {}
-    HttpHdrScTarget(const String &target_):
+    explicit HttpHdrScTarget(const String &target_) :
         mask(0), max_age(MAX_AGE_UNSET), max_stale(MAX_STALE_UNSET),target(target_) {}
-    HttpHdrScTarget(const HttpHdrScTarget &t):
+    explicit HttpHdrScTarget(const HttpHdrScTarget &t) :
         mask(t.mask), max_age(t.max_age), max_stale(t.max_stale),
         content_(t.content_), target(t.target) {}
```
rousskov, line 41: If the existing code copies HttpHdrScTargets, then please explicitly default this copy constructor (if possible):

```suggestion
    explicit HttpHdrScTarget(const HttpHdrScTarget &) = default;
```

If the existing code does not copy HttpHdrScTargets, then please delete this copy constructor:

```suggestion
    explicit HttpHdrScTarget(const HttpHdrScTarget &) = delete; // avoid accidental string copies
```

src/HttpHdrSc.h
rousskov: @yadij, FYI: I see no changes addressing this old change request as of 13097c7. Please let me know if I missed something. See the following comment for a good summary of the remaining work for this change request: https://github.com/squid-cache/squid/pull/589/files#r403623773

yadij: Done in 17da1eb03

src/HttpHdrScTarget.h
yadij: Done in 8ed967f14

src/HttpHdrSc.h
rousskov: FYI: There is currently no such commit in this PR branch.

yadij: FYI: shortly after both our comments github started showing it.

src/HttpHdrScTarget.h
```diff
@@ -10,8 +10,10 @@
 #define SQUID_HTTPHDRSURROGATECONTROLTARGET_H
 
 #include "defines.h" //for bit mask operations
-#include "HttpHdrSc.h"
+#include "http/forward.h"
+#include "SquidString.h"
 
+class HttpHdrSc;
```
rousskov, line 16: If possible, please move this forward declaration to `http/forward.h`. IMO, it is OK that this class is not in `src/http/` yet.

src/HttpHdrScTarget.h
```diff
@@ -30,13 +32,8 @@ class HttpHdrScTarget
...
 
-    HttpHdrScTarget(const char *target_):
-        mask(0), max_age(MAX_AGE_UNSET), max_stale(MAX_STALE_UNSET),target(target_) {}
-    HttpHdrScTarget(const String &target_):
-        mask(0), max_age(MAX_AGE_UNSET), max_stale(MAX_STALE_UNSET),target(target_) {}
-    HttpHdrScTarget(const HttpHdrScTarget &t):
-        mask(t.mask), max_age(t.max_age), max_stale(t.max_stale),
-        content_(t.content_), target(t.target) {}
+    explicit HttpHdrScTarget(const char *target_) : target(target_) {}
+    explicit HttpHdrScTarget(const String &target_) : target(target_) {}
```
rousskov, line 37: If the existing code copies HttpHdrScTargets, then please explicitly default the copy constructor (if possible):

```suggestion

    explicit HttpHdrScTarget(const HttpHdrScTarget &) = default;

```

If the existing code does not copy HttpHdrScTargets, then please explicitly delete this copy constructor:

```suggestion

    explicit HttpHdrScTarget(const HttpHdrScTarget &) = delete; // avoid accidental string copies

```

The latter is the preferred solution if it is viable, of course.

src/HttpHdrScTarget.h
rousskov: Not done. I refreshed this old request at https://github.com/squid-cache/squid/pull/589#discussion_r454734913 and now resolving this outdated one.

src/HttpHdrScTarget.h
yadij: Done in 46054e7

src/HttpHdrScTarget.h
yadij: Done in e8df08

rousskov:
I pushed one minor polishing touch (commit 97f907e). I did not review the details of this PR. I have no objections to this PR going in if you are comfortable with the changes.

PR #591, opened 2020-04-04
Add editorconfig file

kinkie:


kinkie:
On Sat, Apr 4, 2020 at 4:59 PM Alex Rousskov <notifications@github.com>
wrote:

>
> IMHO, the official repository should not contain configuration files for
> highly divergent tools where a specific tool is not prescribed/required for
> Squid development or deployment:
>


>
>    1. Making our favorite tool configurations official would effectively
>    introduce noise for the majority of developers using *different* tools
>    ("divergent" implies that there will be a lot of variation in this area).
>    2. Committing a given tool configuration can complicate things for
>    developers that use a different configuration of the *same* tool. It
>    is often easier to supply an external project-specific configuration than
>    to overwrite a project-specific configuration that came from the official
>    repository.
>
> The spec for this settings file is that it only applies to subdirectories
of its location. In other words, it's only applied to Squid (
https://editorconfig.org/).

From this point of view, your argument about developers using different
settings of the same tool is actually in favour of the change, not against.

If the developer's tool of choice doesn't support that settings file, it
will ignore it, so at worst there will be no harm in having it.
At best, since these settings are mandated by our coding guidelines,
adding them will make it less likely that PRs need extra review cycles due
to formatting issues, saving developer and reviewer time .

In general, I can agree with you. In the specific case, I see no practical
drawback, and that's why I believe this PR can be helpful and not harmful.

-- 
    Francesco

rousskov:
> The spec for this settings file is that it only applies to subdirectories of its location.

Please interpret my original change request as if I know that fact (which is true).

> From this point of view, your argument about developers using different settings of the same tool is actually in favour of the change, not against.

AFAICT, you are assuming that your configuration is exactly what every developer using `VS` (whatever that is) wants to use for Squid. I speculate that, in many cases, that assertion is likely to fail.

Also, please note that my objections consisted of two primary arguments. Even if every `VS` user wants this exact configuration, my other would still apply.

> If the developer's tool of choice doesn't support that settings file, it will ignore it, so at worst there will be no harm in having it.

There is always "harm" in adding files to the repository: Every file is an overhead. We have to justify that overhead. You may not see harm in adding a file that you use, but if I add 100 files that I use, you may start to feel differently. Also, I am certain that sooner or later somebody will want to change the file you are adding -- we assume responsibility for maintaining it. That responsibility and that maintenance is a part of that "harm" or "overhead".

> At best, since these settings are mandated by our coding guidelines, adding them will make it less likely that PRs need extra review cycles due to formatting issues, saving developer and reviewer time .

That may be true, but these settings control very little for a subset of developers, and there are better (including more universal) ways to ensure proper code formatting.

> In the specific case, I see no practical drawback

That is why I have outlined the particular drawbacks that I see. Your response discussed some positives but did not address those drawbacks AFAICT.

kinkie:
> > From this point of view, your argument about developers using different settings of the same tool is actually in favour of the change, not against.
> 
> AFAICT, you are assuming that your configuration is exactly what every developer using `VS` (whatever that is) wants to use for Squid. I speculate that, in many cases, that assertion is likely to fail.

VS = "Visual Studio Code"
As far as I can tell, the version I am proposing to merge covers the full specification of the "standard", and this part is mandated by our coding guidelines. In other words, I'm trying to save churn in code review, although the time spent in this discussion is already starting to hit diminishing returns on that.

But we can also have  a quick show of hands on the sentiment among currently active developers, and masure who would be harmed, helped or neutral from this change. 

> Also, please note that my objections consisted of two primary arguments. Even if every `VS` user wants this exact configuration, my other concern would still apply.
> 
> > If the developer's tool of choice doesn't support that settings file, it will ignore it, so at worst there will be no harm in having it.
> 
> There is always "harm" in adding files to the repository: Every file is an overhead. We have to justify that overhead. You may not see harm in adding a file that you use, but if I add 100 files that I use, you may start to feel differently. Also, I am certain that sooner or later somebody will want to change the file you are adding -- we assume responsibility for maintaining it. That responsibility and that maintenance is a part of that "harm" or "overhead".

We have different views about this point, and that's fine.
There already are hundreds of files in our source that have no interest to me, for instance the whole ESI subsystem. I can afford not to care about it.

> > At best, since these settings are mandated by our coding guidelines, adding them will make it less likely that PRs need extra review cycles due to formatting issues, saving developer and reviewer time .
> 
> That may be true, but these settings control very little for a subset of developers, and there are better (including more universal) ways to ensure proper code formatting.

Well, right  now the universal way is to run a code formatter script that is tied to a very specific version of a very specific tool, and we are relying on manual review and added churn. I'm not exactly happy about this state of things, if it can be improved.


> > In the specific case, I see no practical drawback
> 
> That is why I have outlined the particular drawbacks that I see. Your response discussed some positives but did not address those drawbacks AFAICT.

I'll try to do it very explicitly:
1. "introduce noise for the majority of developers using different tools" . We already are, under the form of code formatting guidelines, and source-maintenance.sh / formatter.pl / astyle-2.04 . Enabling a subset of developers to do the right thing earlier on will reduce churn, and for everubody else it's a net neutral. If one single person has a practical example of harm to show, I'll immediately cancel the PR and be quiet on the topic.
2. "can complicate things for developers that use a different configuration of the same tool" . Sure, someone could choose to use completely different formatting guidelines - use tabs instead of spaces for instance - and fight against their tools to get code accepted by us, and then see it forcibly changed in whatever small part escaping manual review by the source formatter.

The point I'm trying to make is that the drawbacks you suggest are - absent contrary proof - theoretical, whereas introducing this change has a neutral to positive (however small) practical effect.

.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+end_of_line = lf
+insert_final_newline = true
+
+[*.{c,h,cc}]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+
+[Makefie]
```
rousskov, line 14:
```suggestion
[Makefile]
```

.editorconfig
```diff
@@ -0,0 +1,15 @@
+# editorconfig for VS code
+
+root = true
```
rousskov, line 3: If we do not have to disable higher-level configurations, let's not disable them.

```suggestion
```

.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.{c,h,cc}]
```
rousskov, line 8: Please see if you can make this list more comprehensive by reviewing _all_ filename extensions used by Squid sources. The rules below apply to most Squid sources with Makefile and cf.data.pre being notable exceptions (that I recall).

.editorconfig
```diff
@@ -0,0 +1,15 @@
+# editorconfig for VS code
```
rousskov, line 1: `editorconfig` is obvious
`VS` is unnecessary restrictive -- editorconfig is used by lots of different tools, including non-Microsoft tools AFAICT
`code` is misleading (this is a configuration file for the VS editor/IDE; the code is ours)

```suggestion
# https://editorconfig.org/
```
 or

```suggestion
```

kinkie:
> I am trying to find a way forward here, so I am reordering your comments to start with the most promising one and bypassing the ones that are unlikely to help us make progress.

Thanks for taking the time, let's try to move this forward

> > Well, right now the universal way is to run a code formatter script that is tied to a very specific version of a very specific tool, and we are relying on manual review and added churn. I'm not exactly happy about this state of things, if it can be improved.
> 
> I bet _nobody_ is happy about this state, and we should consider improvements, of course. Your "saving developer and reviewer time" (from an earlier comment) is a valid "pro" argument for accepting editor configurations. Unfortunately, it got lost in the noise of other, invalid arguments (e.g., "I see no harm for you, and it helps me" or "adding noise is OK because we have a lot of that already") and attempts to reject my valid "cons" arguments. I regret that I got sucked into defending valid arguments instead of trying harder to find other arguments to support your use case.

Better ways to comunicate and come to an agreement will definitely help and reduce the fustration levels. I appreciate you investing the time to lay the groundwork here.
Generally speaking, I see two levels of arguments in support:
- idea X improves the state of things for all
- idea X improves the state of things for some, and it doesn't regress for others
- I am not sure I used the argument "adding noise is OK because we have plenty" - if that's how it came across, I apologise as it wasn't my intention

> Perhaps we can find a version of admission criteria that allow this small improvement without creating significant problems in other "please use configuration files for the tool(s) that I happen to use today" areas. How about this draft:
> 
> The official repository may contain self-activating configuration files for 3rd party tools that are usually required for for Squid development or deployment (e.g. compilers, editors, linters, and code formatters). In order to reduce admission and maintenance overheads, each such configuration file must meet the following admission criteria:
> 
> 1. The corresponding tool should be commonly used for Squid development or deployment (e.g., a `.vimrc` configuration file may be OK but `.babelrc` would not be).
> 2. The configuration file should only contain Squid-applicable directives that prescribe behavior that all tool users can agree on (e.g., configuring spaces for C++ code indentation is OK but switching editor console to a specific time zone would be wrong).
> 3. The configuration file may explicitly overwrite higher-level settings but should not disable support for higher-level settings (e.g., if a user-specific configuration file sets option X that is not explicitly configured in the Squid configuration file, that option X should continue to be enabled for that user).
> 4. It should be easy (and usually unnecessary) to keep the configuration file in sync with repository changes (e.g., listing filename extensions is OK but enumerating lots of source file paths would be too much).
> 5. At least one core developer should be willing to maintain the configuration file long-term. When the last maintenance commitment is withdrawn, the configuration file may be removed.

Broadly speaking, it sounds reasonable. There are some tweaks that would make sense IMO:
1. I think non-regression might make sense here. Any such auxiliary file must not affect adversely users of other tools. I agree some sort  of relevance would be helpful but how to define it?
2. how about rewording to "any such configuration should contain directives implementing commonly accepted best practices"?
3. I'm not sure I understand what you mean. I think it means "each developer should be able to override shared directives"?
4. I agree
5. I agree

> Do you think the above would work well in general and for this PR specifically? Any adjustments?
> 
> Needless to say, we can adjust the rules as we gain more experience.

yes.

> Please note that the rules do not apply to configuration examples, only to files that are implicitly used by the corresponding tools (what I called "self-activating") because they are specially named to be used that way.

Agreed

[...]

> FWIW, the "commit what does not immediately hurt others" idea has been used by the Squid Project for many years. We still have not fully recovered from its negative side effects. I am certain that it is the wrong approach for a production-quality product such as Squid.

I generally trust that other developers know what they're doing; I think we would gain by allowing ourselves to move faster and improve developer efficiency, even if at the expense of some performance if needed. You seem to have different priorities, and that's of course OK. I wish we would find an easier way to come to a common understanding, and I appreciate your attempt at doing that in this PR

kinkie:
On the matter of the proposed criteria:
https://editorconfig.org/#download is the list of editors that support the directives (without or with plugins). 
1. The most relevant ones are Visual Studio Code (which I'm starting to use) and GitHub without plugins, and Atom, Eclipse, Emacs, NetBeans, Notepad++, Sublime Text, Vim with plugins / configs.
2. the proposed config only implements mandatory format directives according to our guidelines, so checks
3. this is editor-specific, I can't be sure. I would expect this to be the case
4. I'll review other extensions as per your suggestions, it does and will check
5. that is me, so check

rousskov:
TLDR: We may be converging on the acceptance rules for self-executed files. This round may be the last one needed before they can be finalized for Amos' review. I will submit an updated PR review separately. Assuming broad agreement and no new surprises, this PR can then be merged. The other parts of the discussion below might improve the overall PR review process but are not directly related to this PR.

----

> Generally speaking, I see two levels of arguments in support:
> * idea X improves the state of things for all
> * idea X improves the state of things for some, and it doesn't regress for others

In my experience, very few ideas in this world belong to the above ideal categories. There is virtually always a downside for somebody or something. The presence of those "cons" is not the actual problem though IMO: Most humans naturally try to _defend_ or _push through_ their idea X, as opposed to _evaluating_ X on its merits. And when one is pushing through his idea, he is going to (consciously or otherwise) ignore, reject, or diminish its cons while inflating its pros. A natural human reviewer reaction to that is to excessively attack the inflated pros and overemphasize the neglected cons. A natural submitter reaction to that is to fight back... The situation rapidly moves further away from finding the best solution to shouting about increasingly irrelevant issues.

I see the majority of Squid initial submissions or their initial reviews hitting the above obstacle. Unfortunately. Perhaps this cannot be fully fixed, but I hope that it can be improved.

> 1. I think non-regression might make sense here. Any such auxiliary file must not affect adversely users of other tools.

I agree that "no adverse effect on other tools" is a good rule. Let's add it as "item 6" (for now, to avoid renumbering -- its position is not meant as a sign of low importance).

> I agree some sort  of relevance would be helpful but how to define it?

We cannot define it precisely. Checking off this item requires making a judgement call. That's not ideal, of course, but it is the best we can do for this item AFAICT.

> 2. how about rewording to "any such configuration should contain directives implementing commonly accepted best practices"?

Implementing "commonly accepted best practices" would be a mistake IMO because commonly accepted best practices vary widely -- John may be convinced that `else[ ]if` is such practice while Alice is sure that `else[\n]if` is what everybody should be using these days. The rules we are discussing here are not supposed to change or supersede Squid styling rules. To avoid style wars (in the context of adding editor configuration files), the official Squid styling rules have to be used to draw the lines (and everything outside those official rules should not be in the configuration file so that developers can tweak those aspects the way they want).

N.B. Nothing in the above is meant to imply that the official styling rules are both well-understood and cannot be changed. Improving official styling rules is out of this discussion scope though.

> 3. I'm not sure I understand what you mean. I think it means "each developer should be able to override shared directives"?

It is kind of the opposite: Each developer should be able to continue using their existing (more global in scope) directives as long as those directives do not clash with the officially committed ones. For example, if a configuration file in my home directory sets the editor time zone to UTC, the committed configuration file should not prevent that time zone configuration from taking effect (assuming that the committed configuration file contains no time zone configuration of its own).

The (now gone) `root=true` setting in the proposed file is an example of what we should probably avoid (to still allow developers to customize their editing experience).

It would be even better to allow developers to overwrite committed settings using higher-level configuration files, but that is probably technically _virtually impossible_ (and hopefully unnecessary) in most cases.

> I generally trust that other developers know what they're doing; I think we would gain by allowing ourselves to move faster and improve developer efficiency,

In my reality, most Squid pull requests that I review are objectively problematic. I see overwhelming objective evidence against accepting pull requests without a careful review. Also, most of the problems I am able to detect can be addressed relatively quickly -- these reviews do not slow down the average "project speed" by much.

This particular pull request is exceptional because it introduces a whole new category of supported files, but even here we are making progress with arguably reasonable speed -- in 16 calendar days, we were able to fix the PR, substantially improve the proposed changes, and possibly reach an agreement on the rules for the new category. Could we structured the initial arguments better to improve our efficiency? Yes! Is this PR a good example of a case where the "know what they are doing" reviewers should look the other away while the "know what he is doing" developer is quickly merging the stuff he wants? I really doubt that!

Some could argue that the reviewers should not simply "look the other away" -- they should, instead, just trust the developer to "know what he or she is doing" during the review. Unfortunately, I do not know what adjustments that would necessitate in my reviews. For example, none of my mistakes in _this_ PR review stemmed from that mistrust AFAICT.

The problems in many pull requests are a lot more serious than what we are dealing with in this PR, of course, and there are existing Squid bugs in "efficiently committed" but unreviewed (or poorly reviewed) code that have major long-term impact on the ongoing Squid development (and Squid image).

> I think we would gain by allowing ourselves to move faster and improve developer efficiency, even if at the expense of some performance if needed. You seem to have different priorities,

I am all for decreasing performance if that decrease is indeed "needed". No sane person can be against a required performance decrease AFAICT. I am against decreasing performance if simple changes can prevent that decrease, if the gains are not worth the decrease, etc. It is difficult for me to quickly respond to the "your unspecified priorities slow us down" argument, but perhaps that is not the argument you wanted to make?

kinkie:
> Thank you for addressing my earlier change requests!
> 
> I have a new important question (and a half): Do all current files that will be subject to the proposed editorconfig rules already comply with those editorconfig rules?

I hand-checked samples for each category, and built the rules according what I saw in these files. I didn't get 100% coverage of all existing files, so I can't guarantee we are not being inconsistent.

> * If yes, how did you check that? Perhaps you ran some tool that reformats sources based on the editorconfig rules? I could not find anything like that for Linux (to avoid asking this question)...

I found eclint (https://www.npmjs.com/package/eclint), but it is unuseable due to noise, as it barfs on any line that is not aligned on a multiple of tabspaces boundary.

> * Otherwise, what happens when a developer modifies a single line in a file that contains many violations of the editorconfig rules? Will the editor fix _all_ the violations in this case, introducing a lot of out-of-scope formatting changes? If yes, then we would have to comment out all rules that apply to files not covered by existing automated formatting!

I tried Visual Studio Code, and it doesn't reformat the whole file when applying editorconfig, it only uses it for new lines. I cannot test all the editors we would possibly use, but I'm very willing to remove the editorconfig file if some relevant editor misbehaves in this regard.

.editorconfig
```diff
@@ -0,0 +1,35 @@
...
+[configure.ac]
+indent_style = space
+indent_size = 2
+
+[*.awk]
+indent_style = space
+indent_size = 4
+
+[*.{pl,pl.in,pm}]
+indent_style = tab
```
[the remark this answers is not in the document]
kinkie: > > I hand-checked samples for each category, and built the rules according what I saw in these files.
> 
> FWIW, searching for four spaces in the beginning of a line shows many `indent_style = tab` violations for Perl files:
> 
> ```shell
> git ls-files | egrep '[.](pl|pm)(.in)?$' | xargs grep '^    ' | wc -l
> 863 # lines
> ```
> 
> ```shell
> git ls-files | egrep '[.](pl|pm)(.in)?$' | xargs grep -m1 '^    ' | wc -l
> 33 # files
> ```
> 
> In fact, the number of Perl files using spaces (33) is about the same as the number of Perl files using tabs (37)! There is a large overlap between the two sets -- the total number of Perl files is 44 -- but that does not matter here. This finding makes me wonder if we should use `indent_style=space` for Perl files instead: Since the current sources are inconsistent, we should use a better default -- spaces -- for the same reasons we are using spaces for C++ files _and_ because we are already using spaces for C++ and other file types.

I'm fine with the idea. I'd like to head @yadij 's voice before implementing the change.


> A separate PR replacing tabs with spaces in all Perl files would be very welcomed. If you do not want to volunteer such a PR, then we should probably disable the indentation rules for now:

I can do it, no problem.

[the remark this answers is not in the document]
rousskov: FTR: #660 reformats Perl files.

rousskov:
> kinkie requested a review from rousskov 6 days ago

Bouncing back to you because the PR is not in a reviewable state -- lots of unrelated changes.

BTW, it is not "normal" for PR branches to routinely get into this bad state -- I do not know what you do to your git branches, so I cannot offer any specific fixes/suggestions, but _normally_ PR branches do not get into such a state. I hope that following [GitHints](https://wiki.squid-cache.org/GitHints) does not cause this state. If it does, we need to fix something.

kinkie:
I had rebased without squashing first, to retain comments. I have now squashed and force-updated

.editorconfig
```diff
@@ -0,0 +1,36 @@
...
+
+[*.{c,h,hh,cc,cci}]
+indent_style = space
+indent_size = 4
+
+[Makefile]
+indent_style = tab
+
+[*.am]
+indent_style = tab
```
rousskov, line 17: I suggest merging these two blocks. They are essentially about the same kind of files.

.editorconfig
```diff
@@ -0,0 +1,36 @@
...
+indent_style = space
+indent_size = 4
+
+[*.{pl,pl.in,pm}]
+indent_style = space
+indent_size = 4
+
+[*.m4]
+indent_style = space
+indent_size = 4
```
rousskov, line 33: I suggest merging all these identical rules with the C/C++ block above. This would send a pretty clear signal that this is the formatting we prefer "by default" (i.e. unless there are good reasons to use something else).

.editorconfig
kinkie: They would have to be represented by the same glob, something like
[Makefile{,.am}]
I have some concerns on readability, I'd prefer to keep them separate

.editorconfig
kinkie: I see your point. At the same time, it would be a pretty long, "kitchen sink" style of extensions and languages; I find separating them by language to more readable.

OTOH, there is a bug in the *.m4 line, which can furthermore be merged with *.ac. I'll do that.

kinkie:
This PR has passed slow-burner time, and none of the suggestions are blockers.
I'm merging it, we can iterate in the future in case improvements are warranted.

PR #592, opened 2020-04-05
Reduce Pointer refcounting in forwarding

yadij:
Remove temporary Pointer copy which confused Coverity scan
into thinking there is a double-free. Should resolve Issue #1461170
and maybe others.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -1353,7 +1353,7 @@ aclFindNfMarkConfig(acl_nfmark * head, ACLChecklist * ch)
 }
 
 void
-getOutgoingAddress(HttpRequest * request, Comm::ConnectionPointer conn)
+getOutgoingAddress(HttpRequest * request, const Comm::ConnectionPointer &conn)
 {
     // skip if an outgoing address is already set.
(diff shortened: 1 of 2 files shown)


yadij:
getOutgoingAddress() is far too complex an operation for Comm::Connection to be doing. It is supposed to be kept as close to a state storage object as possible - whereas this function involves ACL checks and other access to transaction state.

Also, "outgoing" is not a concept Connection object is aware of. Depending on how the connection is being used either remote or local maybe "outgoing".

Also, Connection was intended to store the IP addresses and FD for any IP based protocol. That means the caller is the one with knowledge of which state and ACLs are to be applied (eg, tcp_outgoing_* is not relevant to UDP protocols).

Removing the Coverity complaint and the data copy is as far as I want to go in this PR.

yadij:
Cleared for merging under the 10-day criteria.

rousskov:
Just in case this suggestion was missed in my earlier comment: I suggest changing "allocations" in the PR title to "refcounting". "Allocations" usually refer to on-heap allocations and they are not affected by this PR. This PR reduces refcounting overheads. I (still) do not insist on this change.

PR #593, opened 2020-04-06
Convert LRU map into a CLP map

yadij:
Revamped LruMap implementation, adding support for value-specific TTLs,
value-specific memory consumption estimates, memory pooling for value
wrappers, and polishing code. Renamed the result to ClpMap.

Fixed memory leak of generated fake SSL certificates in Squids
configured with dynamic_cert_mem_cache_size=0 (the default is 4MB).

Controversially changed sslcrtvalidator_program cache and TTL defaults:

* After we started accounting for key lengths in validator cache
  entries, the old 2048 bytes default size became not just ridiculously
  small but insufficient to cache the vast majority of cache entries
  because an entry key contains a PEM certificate sent to the validator
  (among other validation parameters). A PEM certificate usually
  consumes at least a few kilobytes. Living the old 2KB default would
  essentially mean that we are disabling caching (by default).

* And if we fix one default, it makes sense to fix the other as well.
  The old 60 second TTL default felt too conservative. Most certificate
  invalidations last days, not seconds. We picked one hour because that
  value is still fairly conservative, used as the default by several
  other Squid caches, and should allow for decent cache hit ratio.

Dropped support for the undocumented "sslcrtvalidator_program ttl=-1"
hack that meant unlimited TTL. Large TTL values should be used instead.
TODO: The option parser should be revamped to reject too-large values.


src/base/ClpMap.h
```diff
@@ -0,0 +1,253 @@
...
+        ~Entry() {delete value;}
+    private:
+        Entry(Entry &);
+        Entry & operator = (Entry &);
+    public:
+        Key key; ///< the key of entry
+        EntryValue *value = nullptr; ///< A pointer to the stored value
+        time_t date = 0; ///< The date the entry created
+    };
+    typedef std::list<Entry *> Queue;
```
rousskov, line 47: To avoid unnecessary raw memory allocations, this type needs adjustments. I have not done the research necessary to find the best type. The usual suspects include:

1. `std::list<Entry>` with an allocator
1. `std::list<Entry*>` with an allocator (with a justification for doubling the number of allocations/dereferences: `STL list item wrapper->entry pointer->stuffWeActuallyNeed`)
1. an invasive list of pointers to Entry

Please also document the order this Queue is maintaining.

src/base/ClpMap.h
```diff
@@ -0,0 +1,253 @@
...
+    typedef typename std::list<Entry *>::iterator QueueIterator;
+
+    /// key:queue_item mapping for fast lookups by key
+    typedef std::map<Key, QueueIterator> Map;
```
rousskov, line 51: Why do we need to order QueueIterators by their keys here? Maintaining that order usually costs a lot of extra CPU cycles (`O(log(n))` instead of `O(1)` of std::unordered_map). What makes this use case special?

yadij: Just that this is the way LruMap did things. This container has the same properties that one had. The adjustments made by this PR only affect the overall maximum N - by reducing it in most installations, though for some increasing it.

src/base/ClpMap.h
yadij: I have done some research. It lends towards using dqueue for the final storage. My primary goal is getting this CLP mapping available for use within Squid. If you want to add the storage optimization as part of this PR though I am open to that.

FTR:

* The required operations are: push to head, keyed search, pop random, pop tail - in that order of commonality.

* For the operations most commonly performed on caches the choice we have is between list, dequeue, vector, map. Other containers I'm aware of lack the ability to perform some required operation.

* From the (circa 2012) speed comparisons I was able to find std::dqueue outperforms list on almost all fronts, and vector on the primary push to head case. No data on std::map there - I imagine that since moves are done via pop+push the lookup hash management is a killer cost.

  * push to head - list and dqueue are equal with essentially zero time.

  * keyed search - vector or dqueue pretty close.

  * remove random - list is fastest by a huge margin which grows as object sizes increase. second best is dqueue. vector gets very, very bad.

  * remove tail - list and dqueue and vector are equal with essentially zero time.

src/base/ClpMap.h
rousskov: Understood, thank you. It is my understanding that this PR replaces LruMap rather than enhances it. If my understanding is correct, we should evaluate ClpMap design. This particular design aspect looks questionable to me (for the reasons I have stated in the change request). If you share my concern, please address it. Otherwise, let's discuss why my concern is invalid.

src/base/ClpMap.h
rousskov: > It lends towards using dqueue for the final storage

What exactly do you mean by `dqueue`? std::deque?

> My primary goal is getting this CLP mapping available for use within Squid. If you want to add the storage optimization as part of this PR though I am open to that.

* If you are adding a new class, it should have adequate performance for the intended use cases. My usual [principles](https://github.com/measurement-factory/squid-notes/blob/start/review.md) of requiring PR cleanup would apply to every class aspect in this case, including storage.
* If you are enhancing LruMap with new functionality, then the PR acceptance criteria would be different because the PR would only be responsible for new/changed code.

Based on the PR description and the diff, it feels like you are doing the former, but I am ready to re-evaluate this PR if your intent is the latter. Please clarify this PR scope/intent.

I will come back to this change request once `dqueue` definition and PR scope are clear.

yadij: This is functionality upgrade to LruMap.

rousskov: > This is functionality upgrade to LruMap.

I interpret "functionality upgrade to LruMap" as enhancing LruMap by adding new LruMap functionality and leaving the old LruMap code/functionality unchanged to the extent possible. This PR is not, by default, responsible for improving existing LruMap code/storage/etc. -- it reuses what it can safely reuse. The unusable PR diff is just an unfortunate side effect of premature renaming. Please correct me if my interpretation is wrong.

> It lends towards using dqueue for the final storage.

What exactly do you mean by dqueue? std::deque?

Also, please clarify what you mean be "it" or "it lends" in this context. My change request identified unnecessary raw memory allocations. You responded with the above phrase. I do not see a way to interpret "lends towards using dqueue for the final storage" as justifying those unnecessary raw memory allocations, but perhaps you were commenting on something else?

yadij: WTF, my reply to you went in as an edit rather than a quote. sorry about that.

yadij: >> It lends towards using dqueue for the final storage.
>
> What exactly do you mean by dqueue? std::deque?

Yes.

> Also, please clarify what you mean be "it" or "it lends" in this context.

The benchmarks I found when researching the STL container performances. That showed list as very sub-par in the most common operations performed by our Lru/ClpMap template.

 > My change request identified unnecessary raw memory allocations. You responded with the above phrase. I do not see a way to interpret "lends towards using dqueue for the final storage" as justifying those unnecessary raw memory allocations, but perhaps you were commenting on something else?

If we are going to widely change the storage related code of this template to optimize the existing internal storage. Then we should move to the container with optimal performance at the same time. Which is not the std::list currently used.

src/base/ClpMap.h
yadij: The rename does imply replacement, but the implemented logic for now is still LRU in the queue mechanism.

The rename is to remove the implication that the queue type has to be an LRU. Resolving that sticking point in our previous disagreements. The LRU portion can be replaced in future with any other queue ordering and the type remains ClpMap.

src/base/ClpMap.h
rousskov: It feels like your evaluation in earlier [comment](https://github.com/squid-cache/squid/pull/593/files#r404478061) was not for the type we are discussing here -- Queue. This (currently undocumented) type is for ensuring the right LRU removal order. Queue does not participate in item _searches_. We have Map for that. For Queue, the operations are:

* push_back() -- adding a new item to the cache;
* pop_front() -- purging the next capacity victim from the cache;
* erase(iterator) -- removing a caller-requested item from the cache.

std::deque, std::list, and a custom invasive list (all with an allocator) have similar (amortised) small costs for the first two operations. std::deque loses a bit on the erase() speed, but that is not why std::deque has to be disqualified. The problem with std::deque is that its erase() _invalidates all iterators_ (that are stored in Map). We cannot afford that kind of damage because we have to keep Queue and Map in sync!

This leaves us with std::list with an allocator (item 1 on my [earlier](https://github.com/squid-cache/squid/pull/593/files#r404089699) list) and a custom invasive list (item 3 on my earlier list). I recommend the former because it is an existing, well-known class with overheads similar to a custom one (AFAICT). Please adjust the PR code accordingly.

yadij: Allocator added in commit 52b7adb8abbe686e2bb6856f267b01c7842e7187

yadij:
The Allocator issue is resolved now I believe. If you agree please resolve the thread(s) on that request.

Regarding Map; I am not sure what exactly you are wanting to be done here. IIUI, you were asking why it is used. I have answered that question - it was existing code and I do not fully understand the why. As such I am reluctant to remove it right now.

If you insist on Map removal please state that clearly and I will do so. Otherwise I think this PR is as good for merge.

src/base/ClpMap.h
```diff
@@ -0,0 +1,254 @@
...
+        ~Entry() {delete value;}
+    private:
+        Entry(Entry &);
+        Entry & operator = (Entry &);
+    public:
+        Key key; ///< the key of entry
+        EntryValue *value = nullptr; ///< A pointer to the stored value
+        time_t date = 0; ///< The date the entry created
+    };
+    typedef std::list<Entry *, PoolingAllocator<Entry *> > Queue;
```
rousskov, line 48: N.B. I am moving the old [change request](https://github.com/squid-cache/squid/pull/593#discussion_r404089699) here to keep it together with other change requests in this review.

> Alex: This leaves us with std::list with an allocator (item 1 on my [earlier](https://github.com/squid-cache/squid/pull/593/files#r404089699) list) and a custom invasive list (item 3 on my earlier list). I recommend the former

> Amos: Allocator added in commit 52b7adb

That commit implements item 2 on my [earlier](https://github.com/squid-cache/squid/pull/593/files#r404089699) list minus the justification. Not the recommended item 1 and not even the barely tolerated item 3. As its description says, item 2 doubles allocations and, hence, needs justification. If implementing item 2 was not an accident, then please justify those double allocations. Otherwise, let's go with item 1 instead:

```suggestion
    typedef std::list<Entry, PoolingAllocator<Entry> > Queue;
```
Needless to say, the rest of the code would need to be adjusted accordingly.

FWIW, I really do not understand why this is so hard. I am spending a lot of time describing specific (significant) problems and offering specific (simple) solutions. Addressing such change requests should not require so many painful iterations. If there is something I should do differently, please let me know.

src/base/ClpMap.h
rousskov: Here is a specific suggestion that would move us towards addressing this old change request:

```suggestion
    /// key:queue_item mapping for fast lookups by key
    typedef std::unordered_map<Key, QueueIterator, ..., ..., PoolingAllocator<MapItem> > Map;
```

... but the std::unordered_map template will need a hash and equality functors that I cannot just type in from memory here. This is the primary reason why I hesitated providing an exact code line after you confirmed that we do not need the O(log(n)) search overhead here.

BTW, the MapPair type declared below should probably be declared above. I recommend calling it MapItem because the "pair" is an unimportant implementation detail that may change.

src/base/ClpMap.h
rousskov: Moved to https://github.com/squid-cache/squid/pull/593#discussion_r420145859

src/base/ClpMap.h
yadij: These are pointers to avoid a full re-allocation of the Entry object on every HIT. The LRU storage requires moving it to the front of the index whenever it is used (see ::findEntry() logic).

rousskov: > These are pointers to avoid a full re-allocation of the Entry object on every HIT

AFAICT, we do not need to prioritize cache hits over misses/additions: We should use std::list<Entry,...> and its [splice()](https://en.cppreference.com/w/cpp/container/list/splice) method to avoid _both_ an extra allocation on every cache miss/add operation and re-allocation on every cache hit operation.

yadij: I get what these two Map and Queue containers are doing now.
 * Map for fast Key-based lookups to fetch the data
 * Queue for LRU behaviour management

We do need both, since the types suitable for LRU do not have fast fetch and the types suitable for fast fetch cannot be used for LRU operations. However, it does not matter which one the Entry actually gets allocated by.

I think it would be best to emplace the Entry in Map where the iterators are more stable and make Queue the one storing iterators. That also makes the implementation match the roles index and storage member names imply for those containers.

rousskov: > I get what these two Map and Queue containers are doing now.

Glad to hear that. I just wish that such understanding would be a precondition for any optional refactoring rather than a side effect of many painful reviews.

> I think it would be best to emplace the Entry in Map where the iterators are more stable and make Queue the one storing iterators.

I am not sure what you mean by "more stable", but if Map uses std::unordered_map (as discussed in another change request), then we have to allocate Entry object wrappers in the std::list/Queue because inserting a new element into an unordered map may invalidate all map iterators AFAICT.

yadij: By stable I mean in regards to memory movements. Queue contents are being moved around regularly, thus the pointers to avoid reallocating Entry objects constantly. Map contents are added and stay there until deleted.

You are right about the iterator invalidation. We can minimize that with emplace() instead of insert() - but rehash remains a problem. Queue would have to stay as Entry*, but would be references to the Map content instead of a secondary allocation hanging off Queue. The docs for emplace() state those remain valid even if a rehash happens.

rousskov: > Queue contents are being moved around regularly, thus the pointers to avoid reallocating Entry objects constantly.

std::list does not reallocate items when moving them using std::list::splice(). Thus, there is no reallocation whether the list item is Entry or a pointer to Entry. When moving, the std::list<T> simply changes next/previous pointers in T wrappers.

Compared to storing an Entry in std::list<Entry>, allocating an Entry and then storing a pointer to that allocated Entry in std::list<Entry*> results in an extra memory allocation (and an extra dereference overhead). There are more comparisons like that in the list further below.

> Map contents are added and stay there until deleted.

This alleged lack of movement in Map is a misconception: The primary difference between Map and Queue (as far as item movement is concerned) is that Queue item movements are explicit (i.e. done by our code) while Map item movements are implicit (i.e. done by STL code). In ClpMap, both containers move their items as needed. Those needs differ, but they exist for both containers.

> We can minimize [iterator invalidation] with emplace() instead of insert()

Not sure what you mean. AFAICT, emplace() and insert() have identical effect on iterator invalidation.

> Queue would have to stay as Entry*, but would be references to the Map content instead of a secondary allocation hanging off Queue

That "secondary allocation hanging off Queue" is unavoidable: As we have discussed many times already, std::list<T> does not allocate Ts. It allocates T's wrappers. Thus, an std::list that stores pointers to X, performs exactly the same number of allocations as an std::list that stores X.

Here are the allocations for a few designs that I can think of, including three inferior ones. The "Described" entries represent my attempts to express in C++ what you described in the paragraph quoted above.

1. Requested: `std::list<Entry>` plus `std::unordered_map<Key, QueueIterator>` design allocates one `Entry` wrapper (plus one node for the map pair) -- 2 allocations total
2. Implemented: `std::list<Entry*>` plus `std::map<Key, QueueIterator>` design allocates one `Entry*` wrapper and one `Entry` (plus one node for the map pair) -- 3 allocations total
3. Described(?): `std::list<Entry*>` plus `std::map<Key, Entry>` allocates one `Entry*` wrapper (plus one node for the map pair) -- 2 allocations total
4. Described(?): `std::list<Entry*>` plus `std::map<Key, Entry*>` allocates one `Entry*` wrapper and one `Entry` (plus one node for the map pair) -- 3 allocations total

The above list does _not_ summarize all the pros/cons discussed elsewhere, just allocations per cache entry. Taken everything into account, the first design wins. If I missed a superior design, please specify it explicitly.

yadij: What I have been implementing (bcf090f3287d9c5fa8c42dd5a7c23563f4ef6d71) is #3 - to get 2 allocations.

rousskov: > What I have been implementing (bcf090f) is 3 - to get 2 allocations.

Why not item 1, to get the same 2 allocations but _without_ the problems introduced by item 3, such as the current linked list scan with a cost of O(n)?!

yadij: One list search of O(n) here   vs  N list searches of O(1) + N map searches of O(1) on the sub-call to trim and del,  plus an extra layer of pointer dereference on every operation accessing the stored object;  map iterator -> list iterator -> Entry

And +40 extra lines of diff against master towards code obscurity.

I am not sure what other problems you are talking about. The things we have discussed as problematic are all needing to be taken care of in either implementation.

rousskov: > One list search of O(n) here vs N list searches of O(1) + N map searches of O(1)

Please explain where you see "N list searches of O(1) + N map searches of O(1)" for item 1. AFAICT, a correct implementation of `std::list<Entry>` plus `std::unordered_map<Key, QueueIterator>` requires:

* add/miss without capacity purging: an O(1) search, an Entry wrapper allocation, and a map node allocation
* capacity purge: an O(1) search and two deallocations
* delete by key: an O(1) search plus two deallocations
* hit: an O(1) search

The reviewed implementation based on `std::list<Entry*>` plus `std::map<Key, Entry>` uses (at least):

* add/miss without capacity purging: an O(1) search, an `Entry*` wrapper allocation, and a map node allocation
* capacity purge: an O(1) search and two deallocations
* delete by key: an O(1) search and a full O(n) scan plus two deallocations
* hit: an O(1) search and a full O(n) scan

The important difference is in "delete by key" and "hit" operations.

BTW, the O(1) search in the capacity purge can probably be removed by further changes. We may be able to store a map iterator in the list.

> I am not sure what other problems you are talking about.

The one I recall right now is storing a pointer to a _part_ of an allocated object (the Entry pointer in the list points to a part of an allocated object; the allocated object is the map node in this case). This is confusing, dangerous, may cause Coverity false positives and, worse, false negatives, etc.

rousskov: I added a minor correction and clarification to the previous https://github.com/squid-cache/squid/pull/593#discussion_r422642304. Look for the `Edit:` phrases. These details are unimportant but I adjusted them in hope to reduce review iterations in case we get distracted by them.

yadij: > > One list search of O(n) here vs N list searches of O(1) + N map searches of O(1)
> 
> Please explain where you see "N list searches of O(1) + N map searches of O(1)" for item 1. 
> AFAICT, a correct implementation of `std::list<Entry>` plus `std::unordered_map<Key, QueueIterator>` requires:
> 
>     * add/miss without capacity purging: an O(1) search, an Entry wrapper allocation, and a map node allocation
> 
>     * capacity purge: an O(1) search and two deallocations
> 

trim() is a O(2) search [of list] to find which entries to purge - with N repeats of the search. The second search for each N is in the sub-call to del().

This trim search is an O(2) because it is just several calls to end(). If there were actual list walking it would be worse.


> 
> The reviewed implementation based on `std::list<Entry*>` plus `std::map<Key, Entry>` uses (at least):
.
> 
>     * hit: an O(1) search and a O(n) list search (Edit: no full scan since [a7de1f4](https://github.com/squid-cache/squid/commit/a7de1f4a395009173abe6efadbf5724e371db52e)).

Ah. I missed that O(n) search with the focus on allocations, it is worse than the extra trim() operations. That convinces me.

rousskov: > trim() is a O(2) search [of list] to find which entries to purge - with N repeats of the search. 

I am writing the following paragraph in hope to avoid similar distractions in the future:

To avoid discussing and comparing needlessly complex formulas, we should focus on the set of basic operations that every approach has to implement and apply. In this particular context, that basic operation is locating and removing a single victim (rather than locating and removing enough victims to free a given number of bytes). In both discussed approaches (items 1 and 3), it has the same cost -- O(1) plus two deallocations. O(1) search is comprised of two O(1) components: trivially locating the first list item and searching the hash to remove the corresponding mapped entry. Needless to say, 2*O(1) is the [same](https://en.wikipedia.org/wiki/Big_O_notation#Multiplication_by_a_constant) as O(1).

> [linear search] is worse than the extra trim() operations.

There are no _extra_ trim() operations AFAICT. The cost of locating and removing a victim is the same for both approaches.

Please note that I am describing the _fundamental_ costs of data structures chosen for the CLP map. Poor implementation may increase those costs further, of course, but this change request is not about those implementation bugs or deficiencies. I have not even reviewed the implementation yet. The correct choice of primary data structures can (and should) be made based on fundamentals.

Said that, if something in the current implementation contradicts the stated fundamental expectations for the recommended approach, fixing the implementation before requesting another review round will reduce the number of review iterations.

> That convinces me.

The ball is on your side. Once this old change request is addressed, I will start reviewing the details of the PR (which are likely to change quite a bit by then).

yadij: Done at 3928dfd

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+{
+    if (ttl == 0)
+        return false;
+
+    if (memLimit() == 0)
+        return false;
+
+    del(key);
+
+    const auto wantSz = memoryCountedFor(key, t);
```
rousskov, line 190: I recommend renaming wantSz to wantSpace. Our trim() is already using that name; it clarifies the true meaning of this variable.

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+    /// The current size of the map
+    size_t memoryUsed() const {return memUsed_;}
+    /// The number of stored entries
+    int entries() const {return entries_;}
+private:
+    ClpMap(ClpMap const &);
+    ClpMap & operator = (ClpMap const &);
+
+    bool expired(const Entry &e) const;
+    void trim(size_t wantSpace = 0);
```
rousskov, line 88: After other changes, the strange zero default would be gone:

```suggestion
    void trim(size_t wantSpace);
```

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+DefaultMemoryUsage(const EntryValue *e)
+{
+    return sizeof(*e);
+}
+
+/**
+ * A Map for caching data by Capacity, Lifetime, and Priority (CLP)
+ * Unlike other Map containers data is;
+ * - added only if it fits within a predetermined memory limit (Capacity),
+ * - gets expired based on TTL (Lifetime), and a fading Priority Queue.
```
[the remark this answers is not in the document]
yadij: The Priority part is the LRU.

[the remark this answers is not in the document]
rousskov: Please do not edit my GitHub comments. I also do not recommend editing others' comments. The gains are not worth it, and the outcome is often very confusing and may be misleading.

> The Priority part is the LRU.

Thank you for clarifying that. Here is the adjusted suggestion:

```suggestion
 * - becomes hidden (and may eventually be deleted) after its TTL expires (Lifetime),
 * - purged in LRU order (Priority).
```

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+    Storage data;
+
+    /// index of stored data by key
+    KeyMapping index;
+
+    /// TTL to use if none provided to add(). 0 to disable caching.
+    int defaultTtl = std::numeric_limits<int>::max;
+    size_t memLimit_ = 0; ///< The maximum memory to use
+    size_t memUsed_ = 0;  ///< The amount of memory currently used
+    int entries_ = 0;     ///< The stored entries
```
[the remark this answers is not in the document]
yadij: Done in 9cd838d3c

src/base/ClpMap.h
yadij: Done in 9cd838d3c

src/base/ClpMap.h
yadij: Done in 9cd838d3c

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+
+template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *)>
+void
+ClpMap<Key, EntryValue, MemoryUsedByEV>::trim(size_t wantSpace)
+{
+    while (memLimit() < (memoryUsed() + wantSpace)) {
+        auto i = data.end();
+        --i;
+        if (i != data.end()) {
+            del(i->key);
```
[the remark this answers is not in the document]
yadij: Using rbegin() always produces an iterator. When the list is empty that iterator is rend().
The del() method uses find() and will always find the end iterator in the object (at the rend() position). Resulting in an infinite loop removing a non-existent entry.

We could use loop, but not avoid the complexity of checking the iterator value before deletion.
```suggestion
        auto i = data.rbegin();
        if (i != data.rend()) {
            del(i->key);
```

src/base/ClpMap.h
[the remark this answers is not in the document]
rousskov: No 9cd838d3c yet.

src/base/ClpMap.h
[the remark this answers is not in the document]
rousskov: > Using rbegin() always produces an iterator. When the list is empty that iterator is rend().

In the correct implementation, the list will never be empty in the middle of trim(). You can assert() that. There is no need to walk on tiptoes with iterators when the code is solid. It only hurts code understanding (and, in some cases, performance). Parts of the LruMap code were written in the "I am not sure what invariants I can rely on so I am going to check anything I can think of just in case something went wrong somewhere or I misunderstood something" style. We do not want to replicate that mistake in the new class.

BTW, _inlining_ trim()'s code, as I already suggested elsewhere, would make the "trim is guaranteed to succeed [because it must not be called otherwise]" principle more direct/obvious than using a method ever can, even if you describe the method with a beautiful 20-lline comment box documenting all the warnings and preconditions ;-).

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *)>
+bool
+ClpMap<Key, EntryValue, MemoryUsedByEV>::expired(const ClpMap::Entry &entry) const
+{
+    return entry.expires < squid_curtime;
+}
+
+template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *)>
+bool
+ClpMap<Key, EntryValue, MemoryUsedByEV>::del(const KeyMapIterator &i)
```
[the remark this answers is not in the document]
yadij: Done in c775ad2fd

src/base/ClpMap.h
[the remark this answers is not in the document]
yadij: I agree with the concept. But changing these types here will will have a large impact and is out of scope for this PR. These are the types we currently have to work with in general within Squid.

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+    size_t memLimit() const {return memLimit_;}
+    /// The free space of the map
+    size_t freeMem() const { return (memLimit() > memoryUsed() ? memLimit() - memoryUsed() : 0);}
+    /// The current size of the map
+    size_t memoryUsed() const {return memUsed_;}
+    /// The number of stored entries
+    int entries() const {return entries_;}
+private:
+    ClpMap(ClpMap const &);
+    ClpMap & operator = (ClpMap const &);
```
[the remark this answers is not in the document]
yadij: Done in 38f0ec390

src/ssl/helper.cc
```diff
@@ -212,7 +212,7 @@ void Ssl::CertValidationHelper::Init()
 
     //WARNING: initializing static member in an object initialization method
     assert(HelperCache == NULL);
-    HelperCache = new Ssl::CertValidationHelper::LruCache(ttl, cache);
+    HelperCache = new CacheType(ttl, cache);
```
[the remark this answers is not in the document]
yadij: Done in efccf99fc

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+    size_t memoryCountedFor(const Key &, const EntryValue *);
+
+    /// The {key, value, ttl} tuples.
+    /// Currently stored and maintained in LRU sequence.
+    Storage data;
+
+    /// index of stored data by key
+    KeyMapping index;
+
+    /// TTL to use if none provided to add(). 0 to disable caching.
```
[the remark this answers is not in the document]
yadij: Done in 11c1286be08a46e072721aad2aa5d7c7f70ba443

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+ClpMap<Key, EntryValue, MemoryUsedByEV>::add(const Key &key, EntryValue *t)
+{
+    return add(key, t, defaultTtl);
+}
+
+template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *)>
+bool
+ClpMap<Key, EntryValue, MemoryUsedByEV>::add(const Key &key, EntryValue *t, int ttl)
+{
+    if (ttl == 0)
```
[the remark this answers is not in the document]
yadij: Done in ac65b764d

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+        return false;
+
+    if (memLimit() == 0)
+        return false;
+
+    del(key);
```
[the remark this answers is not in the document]
yadij: Irrelevant after ac65b764d

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+{
+    i = index.find(key);
+    if (i == index.end()) {
+        return;
+    }
+
+    auto &e = (*i).second;
+    if (!expired(*e)) {
+        if (e != data.begin())
+            data.splice(data.begin(), data, e, std::next(e));
```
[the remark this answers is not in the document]
yadij: Done in b741c2a01

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *)>
+void
+ClpMap<Key, EntryValue, MemoryUsedByEV>::findEntry(const Key &key, KeyMapIterator &i)
+{
+    i = index.find(key);
+    if (i == index.end()) {
+        return;
+    }
+
+    auto &e = (*i).second;
```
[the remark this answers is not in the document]
yadij: Done in 579b91ddb

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+ClpMap<Key, EntryValue, MemoryUsedByEV>::setMemLimit(size_t aSize)
+{
+    const auto oldLimit = memLimit_;
+    if (aSize > 0)
+        memLimit_ = aSize;
+    else
+        memLimit_ = 0;
+
+    if (oldLimit > memLimit_)
+        trim();
```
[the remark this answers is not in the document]
yadij: Done with a slight variant. Using a trim() first that "wants" free space enough to swallow any size reduction in memory capacity before the capacity change takes effect.

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+    /// Add an entry to the map with specific TTL
+    bool add(const Key &key, EntryValue *t, int ttl);
+    /// Delete an entry from the map
+    bool del(const Key &key);
+    /// (Re-)set the maximum size for this map
+    void setMemLimit(size_t aSize);
+    /// The available size for the map
+    size_t memLimit() const {return memLimit_;}
+    /// The free space of the map
+    size_t freeMem() const { return (memLimit() > memoryUsed() ? memLimit() - memoryUsed() : 0);}
```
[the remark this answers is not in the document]
yadij: Done in 9e274ee46

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+    else
+        memLimit_ = 0;
+
+    if (oldLimit > memLimit_)
+        trim();
+}
+
+template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *)>
+void
+ClpMap<Key, EntryValue, MemoryUsedByEV>::findEntry(const Key &key, KeyMapIterator &i)
```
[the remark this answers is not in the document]
yadij: Done in 47a2dd39b

src/base/ClpMap.h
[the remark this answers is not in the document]
yadij: Done in 9fa0e446b

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+    EntryValue *get(const Key &key);
+    /// Add an entry to the map
+    bool add(const Key &key, EntryValue *t);
+    /// Add an entry to the map with specific TTL
+    bool add(const Key &key, EntryValue *t, int ttl);
+    /// Delete an entry from the map
+    bool del(const Key &key);
+    /// (Re-)set the maximum size for this map
+    void setMemLimit(size_t aSize);
+    /// The available size for the map
```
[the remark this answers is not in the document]
yadij: Done in 67072b76a

src/ssl/context_storage.cc
```diff
@@ -46,8 +46,7 @@ void Ssl::CertificateStorageAction::dump (StoreEntry *sentry)
         LocalContextStorage & ssl_store_policy(*(i->second));
         stream << ssl_store_policy.memLimit() / 1024 << delimiter;
         stream << ssl_store_policy.entries() << delimiter;
-        stream << SSL_CTX_SIZE / 1024 << delimiter;
-        stream << ssl_store_policy.size() / 1024 << delimiter;
+        stream << ssl_store_policy.memoryUsed() / 1024 << delimiter;
```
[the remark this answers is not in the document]
yadij: Why show this column at all when it has a constant hard-coded value of "1" ?

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+
+template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *)>
+bool
+ClpMap<Key, EntryValue, MemoryUsedByEV>::del(const KeyMapIterator &i)
+{
+    if (i != index.end()) {
+        auto &e = (*i).second;
+        const auto sz = memoryCountedFor(e->key, e->value);
+        data.erase(e);
+        index.erase(i);
```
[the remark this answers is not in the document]
yadij: When index.erase(i) is called both i and e become invalid, so it must be called after data.erase(e).

src/base/ClpMap.h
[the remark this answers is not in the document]
yadij: Okay. Done as of 5d500de6b now that the other code is actually maintaining those required memory invariants.

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+    KeyMapIterator i;
+    findEntry(key, i);
+    return del(i);
+}
+
+template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *)>
+void
+ClpMap<Key, EntryValue, MemoryUsedByEV>::trim(size_t wantSpace)
+{
+    while (memLimit() < (memoryUsed() + wantSpace)) {
```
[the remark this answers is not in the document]
yadij: Done in 5d500de6b

src/base/ClpMap.h
```diff
@@ -0,0 +1,248 @@
...
+    if (ttl == 0)
+        return false;
+
+    if (memLimit() == 0)
+        return false;
+
+    del(key);
+
+    const auto wantSz = memoryCountedFor(key, t);
+    if (wantSz >= memLimit())
```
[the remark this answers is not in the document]
yadij: Done in 5d500de6b

rousskov:
@yadij, you may have missed the following question in my last [review](https://github.com/squid-cache/squid/pull/593#pullrequestreview-414771782) before bouncing the ball to my side: 

> BTW, there are many minor fixes that should be done before this file is merged. Do you want me to slowly itemize them on GitHub or should, I commit them after this round?

rousskov:
AFAICT, this is the highest priority review project on my side right now. Current ETA: +48 hours.

@yadij, please answer the earlier [question](https://github.com/squid-cache/squid/pull/593#issuecomment-644264333).

yadij:
> @yadij, you may have missed the following question in my last [review](https://github.com/squid-cache/squid/pull/593#pullrequestreview-414771782) before bouncing the ball to my side:
> 
> > BTW, there are many minor fixes that should be done before this file is merged. Do you want me to slowly itemize them on GitHub or should, I commit them after this round?

If I take that question literally and they are *fixes* for more bugs, then please itemize them first. If you just mean polish and style cleanup, then it may be easier for you to apply and we discuss afterward.

rousskov:
Unfortunately, I wasted too much time on #680 and missed the previously allocated review window. Current ETA: 2020-06-30.

rousskov:
I resumed working on this PR. I hope to post at least an updated review within 24 hours.

src/base/ClpMap.h
```diff
@@ -0,0 +1,229 @@
...
+public:
+    class Entry
+    {
+    public:
+        Entry(const Key &aKey, EntryValue *t, int ttl): key(aKey), value(t), expires(squid_curtime+ttl) {}
+        ~Entry() {delete value;}
+        Entry(const Entry &) = delete;
+        Entry & operator = (const Entry &) = delete;
+        Entry(Entry &&) = default;
+        Entry & operator = (Entry &&) = default;
```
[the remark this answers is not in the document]
yadij: Sorry. The original C++11 submission docs put it into my head that move is to be a {memmove(D,S); memset(S,0); } sequence. Having some trouble escaping that info. Hopefully this great example will help.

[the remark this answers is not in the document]
yadij: Removed move semantics in d5b233c22

src/base/ClpMap.h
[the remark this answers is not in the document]
yadij: Oh, doh! sorry. I see what you mean now. Done in 61b570fc1

src/base/ClpMap.h
```diff
@@ -0,0 +1,229 @@
...
+    typedef std::list<Entry, PoolingAllocator<Entry> > Storage;
+    typedef typename Storage::iterator StorageIterator;
+
+    /// Key:Entry* mapping for fast lookups by key
+    typedef std::pair<Key, StorageIterator> MapItem;
+    /// key:queue_item mapping for fast lookups by key
+    typedef std::unordered_map<Key, StorageIterator, std::hash<Key>, std::equal_to<Key>, PoolingAllocator<MapItem> > KeyMapping;
+    typedef typename KeyMapping::iterator KeyMapIterator;
+
+    ClpMap(int aTtl, size_t aSize) : defaultTtl(aTtl) { assert(aTtl >= 0); setMemLimit(aSize); }
```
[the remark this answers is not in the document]
yadij: I guess you are referring to making ttl a std::chrono based type. Yes, that is a future project on its own.

ctor split done in 60c9cb986

rousskov:
AFAIK, here are the remaining to-do items:

- [ ] Fix build with GCC v4.8. PoolingAllocator probably requires more GCC compatibility declarations: `_Node_allocator_type has no member named 'construct'`

- [ ] Fix PR description. LruMap also supported total capacity limits. As far as primary functionality is concerned, we are only adding per-entry TTLs and per-entry sizes here.
- [ ] General polish. I have already fixed a few, but there are still oddities, inconsistencies, bad names, and other minor problems (inherited from LruMap and new) that should be fixed in this replacement.

I plan to chip away at these during the next ten days or so, committing small/isolated changes as soon as they become ready after pulling in any changes you might commit.

rousskov:
**Status update**: I pushed a few polishing changes (up to commit 7750188). While working on the next polishing change, I discovered an old bug (caused by the dangerous map API that I was documenting). I will research whether we can address the problem within the next 24 hours, but I have to pause now. I do not think there will be a lot of polishing work left after that.

src/base/ClpMap.h
```diff
@@ -32,10 +32,13 @@ template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *)
 class ClpMap
 {
 public:
+    /// maximum desired entry caching duration (a.k.a. TTL), in seconds
+    using Ttl = int;
```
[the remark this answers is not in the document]
rousskov: > Why "using" instead of typedef?

Personally, I find type alias declarations more readable (many developers share that view) and a bit more "logical" (we are not really defining a new type but using an alias for an existing one). Alias declarations are also more universal (i.e. cover more use cases) so if we want to standardize on one approach, it makes sense to use alias declarations.

> there appears to be no need for "using"

Yes, both `using` and `typedef` are legal in this context from C++ point of view.

> when the rest of Squid code has typedef

Since [2017](https://github.com/squid-cache/squid/commit/e99fa72) (at least), Squid uses both -- there are more than 20 non-templated alias declarations in master today. Thus, this code line does not introduce a new inconsistency. Moreover, when I find the time to work on it, one of the [modernization](http://lists.squid-cache.org/pipermail/squid-dev/2020-April/009569.html) changes might remove the remaining 1000 typedefs.

rousskov:
> I discovered an old bug (caused by the dangerous map API that I was documenting). I will research whether we can address the problem within the next 24 hours

Fixed (commit e3bc1d0). I also found and fixed another bug (commit 7e6488c). The polishing changes are almost over as well. I hope to finish everything within 24 hours.

src/base/ClpMap.h
```diff
@@ -223,10 +223,13 @@ ClpMap<Key, Value, MemoryUsedBy>::erase(const IndexIterator &i)
 {
     assert(i != index_.end());
     const auto entryPosition = i->second;
+
     const auto sz = entryPosition->memCounted;
+    assert(memUsed_ >= sz);
+    memUsed_ -= sz;
```
yadij, line 229: Moving this above the erase() violates the invariant that memUsed_ >= memory currently used for the duration of how long it takes to perform the erasures. Since those operations are far from atomic and may throw that introduces a memory leak opportunity.

src/base/ClpMap.h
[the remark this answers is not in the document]
yadij: Very well. If you are wanting to make a formal switch to remove typedef I agree.

src/base/ClpMap.h
```diff
@@ -44,8 +44,11 @@ class ClpMap
     explicit ClpMap(const size_t aCapacity) { setMemLimit(aCapacity); }
     ClpMap(size_t aCapacity, Ttl aDefaultTtl);
     ~ClpMap() = default;
-    ClpMap(ClpMap const &) = delete;
-    ClpMap & operator = (ClpMap const &) = delete;
+
+    // copying disabled because it may be expensive for large maps
+    // moving (implicitly) disabled for simplicity sake
```
yadij, line 48: IMO these should be above their respective operators.

src/base/ClpMap.h
```diff
@@ -17,18 +17,24 @@
...
-    return sizeof(*e);
+    return sizeof(e);
 }
 
-/// An in-memory cache enforcing three primary policies:
-/// Capacity: The memory used by cached entries has a configurable limit;
-/// Lifetime: Entries are hidden (and may be deleted) after their TTL expires;
-/// Priority: Capacity victims are purged in LRU order.
-template <class Key, class EntryValue, size_t MemoryUsedByEV(const EntryValue *) = DefaultMemoryUsage>
+/// An in-RAM associative container enforcing three primary caching policies:
```
yadij, line 26: "in-RAM" assumes that RAM is the only form of memory. This is becoming less and less true for modern machines. If you want to be specific this had better be "in-NAND" or "in-crystal".

yadij:
I am seeing you undo several changes you argued intensely for me to implement ... changes to invariants protecting del() and re-addition of assert() and if() you argued were unnecessary waste of cycles. naming of EntryValue, MemoryUsedByEV, and aFoo parameters.

Something changed your mind on those?

src/base/ClpMap.h
yadij: Updating the memory value before the erase() leads Squid to thinking it can add more entries when it should not. That is a memory leak behaviour.

I still believe it is best to erase in the order which minimizes potential side effects without having to rely on implementation details of the particular container used:
1) drop index entry. Leaves all data entry positions unaffected. Worst-case one entry has no index record.
2) drop data entry. Index is already updated successfully. Releases memory. Worst case memory accounting over-estimates usage.
3) update memory data. State becomes completely consistent again.

Yes, there are still bad cases with invariants. However with this ordering they all "fail safe" by preventing Squid consuming resources.

rousskov:
While polishing documentation, I realized that this PR essentially disables sslcrtvalidator_program caching. Here is a quote from branch commit 3479ada:

> This PR does not change how sslcrtvalidator_program response size is estimated -- it is still 1024 bytes. However, we did change how the key size is estimated. It used to be zero, but now I see values of 5K and 10K in my primitive tests! That already exceeds the default cache size. Before this PR, Squid could cache two responses by default. Now, it most likely will cache zero.

N.B. Please note that we multiply the key size by two, essentially.

This significant growth of cache entry size estimate will affect non-default configurations as well.

@yadij, do you think we should leave this as is, to be resolved in some other PR (e.g., because the defaults were pretty bad already, and the response calculation remains broken)?

BTW, please feel free to adjust or remove my commit 3479ada changes. I do not insist on them.

yadij:
> This significant growth of cache entry size estimate will affect non-default configurations as well.
> 
> @yadij, do you think we should leave this as is, to be resolved in some other PR (e.g., because the defaults were pretty bad already, and the response calculation remains broken)?

Increasing the cache default size seems reasonable to me. Fixing the calculation is still  independent from the cache default size anyway, so that being out of scope is not as relevant. Maybe a cache default of 64KB if there was a reason for it to be that small or 256KB to scale it similarly to other helpers.

> 
> BTW, please feel free to adjust or remove my commit [3479ada](https://github.com/squid-cache/squid/commit/3479adaf355a272d9da79a7db261889554d65244) changes. I do not insist on them.

I am okay wit the texts. The only issue I see is indentation style. When the option name extends into the first description lines column I have bumped the description down to start on next line.


In regards to 64cad20103421cb3a38c846dbe912f1aa97572bd the methods there have already stopped working as of GCC-5. They are only used by GCC-4.8 builds.

rousskov:
> In regards to 64cad20 the methods there have already stopped working as of GCC-5.

I assume that by "stop working" you mean "not needed". I am not aware of any cases where those extra declarations cause trouble. I expect those cases to appear when we migrate to C++17 or higher.

> They are only used by GCC-4.8 builds.

Those are the ones we know about, but I do not think we should use compiler-version-specific code until there is a compelling reason to use it. I do not see such a reason here and, perhaps more importantly, I would expect that reason to extend to more than one extra declaration in that class (i.e. to cover more than just the construct() method). That is why I removed the preprocessor conditional added in branch commit 90cb84e.

rousskov:
> Increasing the cache default size seems reasonable to me. Maybe [...] 256KB to scale it similarly to other helpers.

I found external_acl_type cache (262144 _entries_ by default which is usually at least 64MB if the docs are not lying). Do you know what other helper caches are currently in use? It would be good to know the status quo before making the decision here.

> Fixing the calculation is still independent from the cache default size anyway, so that being out of scope is not as relevant.

It is not that simple IMO: If fixing the calculation will have a significant effect on the default (i.e. will change how a reasonable default would look like), then there is some relevance. We may not want to change the default twice in a row...

On the other hand, we can try to approach the same problem from the other end. We can postulate that the default cache size should be determined primarily by the amount of memory that a typical deployment can easily sacrifice even if the cache is barely used. The default will never satisfy some deployments with tiny memory or some high-performance instances, but that is not what the default should target anyway. Those special environments naturally require special tuning.

If we follow this approach, I would suggest a 64MB default. It is small enough to be ignored if wasteful and large enough to accommodate 1-5 thousand entries before and after the context size calculation fix, making the cache effective. It can also be viewed as similar to the external ACL cache (see above). If you or others think that 64MB is too much, then perhaps we can settle on 16MB (enough for a few hundred entries before and after the fix)?

And if we change this default, I would suggest to change the TTL default as well. The current 60 second TTL feels too short for an average deployment. Certificate invalidations normally take days; we do not need to be so conservative IMO. I would suggest a 1 hour default, the same as for the external ACL cache.

rousskov:
> I am okay wit the texts. The only issue I see is indentation style. Elsewhere, when the option name extends into the first description lines column (tab indentation) I have bumped the description down to start on next line.

Can you point me to two directives with similarly large descriptions that have the "right" formatting please? I will reformat to mimic them. I am happy to reformat, but I am not sure what the above description means exactly, and there are too many inconsistencies to take any random directive as an example...

src/base/ClpMap.h
```diff
@@ -0,0 +1,260 @@
...
+}
+
+template <class Key, class Value, uint64_t MemoryUsedBy(const Value &)>
+uint64_t
+ClpMap<Key, Value, MemoryUsedBy>::MemoryCountedFor(const Key &k, const Value &v)
+{
+    // approximate calculation (e.g., containers store wrappers not value_types)
+    const auto storageSz = sizeof(typename Entries::value_type) + k.length() + MemoryUsedBy(v);
+    const auto indexSz = sizeof(typename Index::value_type) + k.length();
+    return storageSz + indexSz;
```
rousskov, line 191: @yadij, the above counts the key length twice. The current code nearly assumes SBuf use (by not optimizing key copies and by calling length()). When Key is SBuf, there are no copies, so we should count the key length _once_. Should i change this to count the key once? Or byte the bullet and outsource key size calculation to the map creator?

This is not a minor concern because certificate validator keys, for example, are often 5-50KB in size (because they contain PEM certificates)!

src/base/ClpMap.h
rousskov: > Updating the memory value before the erase() leads Squid to thinking it can add more entries when it should not.

No, it does not lead to that. Or, to be more precise, there is actually no "updating X before Y" here. The updates of all three interlinked components (memory consumption counter, search index, and storage) should be viewed as a single/atomic operation. From the invariants point of view, the order inside that operation does not matter. Moreover, no order can preserve correct state between component updates.

> I still believe it is best to erase in the order which minimizes potential side effects without having to rely on implementation details of the particular container used:

I do not think we can _meaningfully_ compare potential side effects of major invariant violations.

> 1. drop index entry

The "every stored entry can be found by its key" invariant is violated here.

> 2. drop data entry

The "every stored entry is accounted for in the memory consumption counter" invariant is violate here.

> 3. update memory data

Sanity restored.

> However with this ordering they all "fail safe" by preventing Squid consuming resources.

There is no true "fail safe" here. Each violation above may lead to innocent entries getting purged from the cache, consuming extra CPU cycles and other resources on obtaining fresh copies of that information.

Again, if you think it is better to reorder the steps, you should reorder them. I think the current order is the best because it preserves map state when an assertion is triggered.

Moreover, if you want to add an try/catch wrapper (around the code that should not throw unless we hit undefined behavior areas where all bets are off), I would not object.

src/base/ClpMap.h
yadij: I am okay with reducing to one key.length(). But would place a code comment in with that change to indicate why we are able to ignore one of the two keys.

yadij:
> > In regards to [64cad20](https://github.com/squid-cache/squid/commit/64cad20103421cb3a38c846dbe912f1aa97572bd) the methods there have already stopped working as of GCC-5.
> 
> I assume that by "stop working" you mean "not needed". I am not aware of any cases where those extra declarations cause trouble. I expect those cases to appear when we migrate to C++17 or higher.
> 

I mean the code is "dead code" to more recent compilers. Compilers will not complain about that since we are using libraries and they cannot guarantee that no library API symbol is unused by all possible code linking to it.

The trouble I am concerned about here is future code maintenance and technical debt hanging onto this once it is truly useless.  So I would prefer to wrap the definitions in the (__cplusplus <= 201103L) to make it perfectly clear on the future date we move to deprecating C++11 code that it can be safely erased. I did not wrap the pre-existing members due to uncertainty about the specifics of their usage in C++11 and later compilers. If you can clarify them as not being needed for a particular C++ version, then IMO they should all be wrapped in the same #if clause to elide.

> > They are only used by GCC-4.8 builds.
> 
> Those are the ones we know about, but I do not think we should use compiler-version-specific code until there is a compelling reason to use it.

Aye, which is why I used the language-version specific check instead of compiler version. The comment you already have there covers the compiler version detail adequately to explain why they were added. It does not help much with the future removal decision - a protocol version check does that part nicely and lets us detect build errors today while the issue is fresh in our heads if we get the condition wrong (eg C++17 limit instead of C++14).




> > Increasing the cache default size seems reasonable to me. Maybe [...] 256KB to scale it similarly to other helpers.
> 
> I found external_acl_type cache (262144 _entries_ by default which is usually at least 64MB if the docs are not lying). Do you know what other helper caches are currently in use? It would be good to know the status quo before making the decision here.

Beyond the validator and external ACL  there is AuthUser::CredantialsCache, digest_nonce_cache, acl_proxy_auth_match_cache, aclMatchUserMaxIP user-IP cache, TLS session cache, error details list cache, MimeIcons list cache.

At a bit of a stretch there is also ipcache, fqdncache, netdb, clientdb and wccp2 caches list. But those I would expect to have much larger defaults.


> 
> > Fixing the calculation is still independent from the cache default size anyway, so that being out of scope is not as relevant.
> 
> It is not that simple IMO: If fixing the calculation will have a significant effect on the default (i.e. will change how a reasonable default would look like), then there is some relevance. We may not want to change the default twice in a row...
> 


> And if we change this default, I would suggest to change the TTL default as well. The current 60 second TTL feels too short for an average deployment. Certificate invalidations normally take days; we do not need to be so conservative IMO. I would suggest a 1 hour default, the same as for the external ACL cache.


Okay.  64MB limit and 60min TTL Seems reasonable assuming that an "average deployment" is a server with GB's of memory available.




> > I am okay wit the texts. The only issue I see is indentation style. Elsewhere, when the option name extends into the first description lines column (tab indentation) I have bumped the description down to start on next line.
> 
> Can you point me to two directives with similarly large descriptions that have the "right" formatting please? I will reformat to mimic them. I am happy to reformat, but I am not sure what the above description means exactly, and there are too many inconsistencies to take any random directive as an example...

auth_param, icap_servcie, ecap_service, uri_whitespace are all fairly clean and consistent with the indent style for option lists.

rousskov:
> Amos: In regards to 64cad20 the methods there have already stopped working as of GCC-5.

> Alex: I assume that by "stop working" you mean "not needed".

> Amos: I mean the code is "dead code" to more recent compilers.

Sigh. FYI: The existing workaround declarations are _used_ by _all_ compilers, recent and old. Here is PoolingAllocator [test](https://cpp.godbolt.org/z/KrzPYz) output with GCC v10:

```
construct() was used!
destroy() was used!
```

Compliant C++11 compilers/STLs automatically provide a _default_ replacement if needed while GCC v4 fails the build instead. 

> Amos: Compilers will not complain about that since we are using libraries and they cannot guarantee that no library API symbol is unused by all possible code linking to it.

Good C++17 compilers and linting tools should complain about outdated approaches being used. Whether they can detect such usage in _our_ specific case, I am not sure. However, MS VC++ compiler, for example, already warns about similar (but simpler) use cases (the ones that can be disabled with `_SILENCE_CXX17_OLD_ALLOCATOR_MEMBERS_DEPRECATION_WARNING`). I agree that there is no guarantee that we will eventually get a warning for our code here. I added a reference to C++17 in commit aa9ca42 to provide even more clues that this code needs attention as we upgrade.


> Alex: I do not think we should use compiler-version-specific code until there is a compelling reason to use it.

> Amos: Aye, which is why I used the language-version specific check instead of compiler version.

The problem here is not specific to the language version. The existing code [works](https://cpp.godbolt.org/z/az5r76) fine with C++20, for example.

> The comment you already have there [...] does not help much with the future removal decision - a protocol version check does that part nicely and lets us detect build errors today while the issue is fresh in our heads if we get the condition wrong (eg C++17 limit instead of C++14).

Currently, there are no build errors even with C++20 (regardless of whether destruct() is declared).

Hopefully, the current comment covering all declarations and mentioning C++17 is enough to move forward beyond this issue.

rousskov:
I did not find much consistency among the following caches, unfortunately:

* external_acl_type cache -- configurable defaults: 262144 entries (at least 64MB), 1 hour
* proxy_match_cache -- unlimited size, unlimited TTL
* digest_nonce_cache -- 7921 entries, unlimited TTL
* aclMatchUserMaxIP -- configurable number of entries, unlimited TTL
* TLS session cache -- configurable defaults: 2 MB (shm), 5 minutes
* AuthUser::CredentialsCache -- unlimited size, configurable 1 hour TTL

Disclaimer: I only did a quick check. Some of the above may be wrong.

> 64MB limit and 60min TTL Seems reasonable assuming that an "average deployment" is a server with GB's of memory available.

Will change.

rousskov:
> Amos: I am okay wit the texts. The only issue I see is indentation style. Elsewhere, when the option name extends into the first description lines column (tab indentation) I have bumped the description down to start on next line.

> Alex: Can you point me to two directives with similarly large descriptions that have the "right" formatting please?

> Amos: auth_param, icap_servcie, ecap_service, uri_whitespace are all fairly clean and consistent with the indent style for option lists.

All of the above have whitespace inconsistencies, but I hope that my branch commit 0e7b5f5 guessed the indentation style you like. If not, please adjust the style yourself.

src/base/ClpMap.h
rousskov: > I am okay with reducing to one key.length().

Done at d9ea1e5.

rousskov:
I have updated PR description. @yadij, please check it and adjust further if needed. Please also look through the code as a whole. If there are no more adjustments that you want to do, are there any changes you would like _me_ to work on?

src/base/ClpMap.h
```diff
@@ -0,0 +1,266 @@
...
+    // Both storage and index store keys, but we count keySz once, assuming that
+    // copying a Key does not consume more memory. This assumption holds for
+    // Key=SBuf, but, ideally, we should be outsourcing this decision to another
+    // configurable function, storing each key once, or hard-coding Key=SBuf.
+    const auto keySz = k.length();
+
+    // approximate calculation (e.g., containers store wrappers not value_types)
+    const auto storageSz = sizeof(typename Entries::value_type) + MemoryUsedBy(v);
+    const auto indexSz = sizeof(typename Index::value_type);
+    return keySz + storageSz + indexSz;
```
rousskov, line 197: MemoryCountedFor() sums uint64_t values without checking for overflows. I suggest returning zero in overflow cases. The caller should be adjusted to reject the corresponding values. Since we always add a few sizeof()s, the zero return value cannot happen naturally. @yadij, any better ideas? Would you be able to add this code?

FWIW, I doubt we should throw (and try/catch) on overflows because I suspect that doing it correctly would require introducing a special exception value and quite a bit of other custom code.

We could also return uint64::max() on overflows, which feels natural, but that large value, unlike zero, is, technically, a possible/valid value for this function.

yadij: Agreed on zero.

yadij:
> I have updated PR description. @yadij, please check it and adjust further if needed. Please also look through the code as a whole.

Done. I have no issues beyond the disagreement on erase() internals. That is not important enough to me to block merge.

> If there are no more adjustments that you want to do, are there any changes you would like _me_ to work on?

I am working on some final polish to the sslcrtvalidator_program directive parser and will get to the MemOrUsedBy() overflow detection next. If you can review those two patches meet your quality standards it would be good.

src/base/ClpMap.h
yadij: Done in 260e3bbfc

src/base/ClpMap.h
```diff
@@ -0,0 +1,276 @@
...
+
+    /// The number of currently stored entries, including expired ones
+    size_t entries() const { return entries_.size(); }
+
+private:
+    /// the keeper of cache entry Key, Value, and caching-related entry metadata
+    class Entry
+    {
+    public:
+        Entry(const Key &aKey, const Value &v, const Ttl ttl): key(aKey), value(v), expires(squid_curtime+ttl) {}
```
rousskov, line 92: The `expires` sum will overflow when folks use `ttl=infinity`, right?

yadij: Yes. Good catch Thank you.

Fixed in b55e5e36e (could not use the SafeSum because time_t is signed).

rousskov: I hope to finish working on this within 24 hours.

rousskov: Still working on this, sorry. Should finish within the next 24 hours.

rousskov: Fixed(?) at 2c941cb. Please test.

yadij: Thank you. Looks good to me, though signed overflows have never been one of my strong points.

I will spend the next day running this code and see if anything shows up before clearing for merge.

yadij:
@rousskov, this code has now been facing 1.4K req/sec steadily with randomized key/value inputs large enough to push the capacity limit with occasional TTL expiry on top - for the past 12hrs and showing no issues. If you want to mark it as approved now I will leave it running the tests and clear for Anubis tomorrow.

yadij:
Overnight test completed successful.

src/base/ClpMap.h
yadij: FTR, my understanding of "fail safe" definition for erase sequences like this is that during the "insanity" period other callers using the public API get(key) / find(key) / add(key) / del(key) should never be faced with risk of following an invalid iterator/pointer within the container. That means the index such callers interact with to find things needs to never point to missing/erased entry - even if some fluke occurrence causes the erase method to exit during "insanity".


Resolving this conversation, since we are now nitpicking edge cases that likely never occur.

src/base/ClpMap.h
```diff
@@ -0,0 +1,284 @@
...
+    // optimization: avoid del() search, MemoryCountedFor() in always-empty maps
+    if (memLimit() == 0)
+        return false;
+
+    del(key);
+
+    if (ttl < 0)
+        return false; // already expired; will never be returned by get()
+
+    const auto memoryRequirements = MemoryCountedFor(key, v);
```
rousskov, line 219: This add() method is the only place where we use MemoryUsedBy template parameter today. The code that calls add() probably knows what entry it is adding, so it could supply the sizing information to add(). The current MemoryUsedBy() default value is unused IIRC. If we ever add support for dynamic size changes this single sizing function is not going to be enough anyway. Should we remove the MemoryUsedBy parameter and add the corresponding size parameter instead?

rousskov:
@yadij, I approved the PR, but removed the `M-cleared-for-merge` label to give you a chance to review my two commits (c3146b0 and 4a1bca6) and consider another [change](https://github.com/squid-cache/squid/pull/593#discussion_r460451485).

src/base/ClpMap.h
rousskov: > my understanding of "fail safe" definition for erase sequences like this is that during the "insanity" period other callers using the public API

There must be no other callers during the critical section execution. It is virtually impossible to write correct code if there are such callers. If there is a possibility of such reentrant behavior that is not already covered by the existing assertions, then we should add more assertions.

rousskov:
I also pushed 6838cd1, primarily so that GitHub sends a proper notification to Semaphore CI which was blocking build tests. Something went wrong with the previous GitHub notification; I am not sure, but I think GitHub sent a notification for the wrong/old branch commit instead. This last push seemed to help -- Semaphore CI is building the right/latest commit now.

src/base/ClpMap.h
yadij: Tempting, but many of the caches that will be migrated to using this will only need the default calculation. So for now I will leave it in and we can address the differences(if any) for dynamic entries when that happens.

yadij:
Thank you very much for all the time spent on this.

PR #599, opened 2020-04-12
Maintenance: Move sort-includes.pl to scripts/maintenance/

yadij:
We will no longer get `changed #include order` notices but will get
`changed: by maintenance/sort-includes.pl` notices, including for
whitespace-only changes.


yadij:
Depends on PR #531

rousskov:
> after trying clang-format out, I'm inclined to favour us adopting that and dropping this.

What are clang-format rules for ordering `#include` files?

rousskov:
FYI: I marked this PR as a "draft" PR. We should continue to use our manual `WIP:` prefix as well (for now) because Anubis does not know about draft PRs (yet) -- it is a recently added GitHub feature. Both markers should be removed when the PR stops being a WIP/draft PR.

rousskov:
I described notification changes in the PR description. Please check that it matches reality.

kinkie:
> > after trying clang-format out, I'm inclined to favour us adopting that and dropping this.
> 
> What are clang-format rules for ordering `#include` files?

IncludeCategories:
  - Regex:           '^"squid.h"$'
    Priority:        -1
  - Regex:           '^"'
    Priority:        2
  - Regex:           '.*'
    Priority:        3
IncludeIsMainRegex: '(stub_)?$'
SortIncludes:    true

rousskov:
> Alex: What are clang-format rules for ordering #include files?

> Francesco: [See configuration]

@kinkie, FYI: I saw that configuration before asking the question. The configuration is not sufficient for me to understand why clang-format has changed Squid includes the way it did for [compat/xstring.cc](https://github.com/squid-cache/squid/compare/master...kinkie:clang-format#diff-b73ad949015aac149ab87ce6ebc9ffe6L10), for example. I suspect `SortIncludes` uses different rules (inside each priority group) than we currently do.

We can, of course, redefine the correct order as "whatever this particular configuration of this particular clang-format version does". However, before we do that, it would be nice to at least attempt to discover what that actually means.

Against my best judgement, I researched this a little instead of waiting for you to supply the missing details. Clang-format [documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) for `IncludeCategories` says that "The main header for a source file automatically gets category 0". This means that your current configuration changes the current include order by placing that "main header" right below `squid.h`. We might be able to avoid this change by using negative categories for everything, but I have not tested that workaround.

Moreover, I suspect that clang-format comparison is case sensitive. I am guessing that could explain `include/unitTestMain.h` [changes](https://github.com/squid-cache/squid/compare/master...kinkie:clang-format#diff-eb18ae0860db5bcf8f8caa0e627e80bdR22), for example. I did not find the confirmation for that idea in the clang-format documentation, and clang-format _ignores_ case sensitivity when sorting `using` declarations, for example (see `SortUsingDeclarations`).

kinkie:
> > Alex: What are clang-format rules for ordering #include files?
> 
> > Francesco: [See configuration]
> 
> @kinkie, FYI: I saw that configuration before asking the question. The configuration is not sufficient for me to understand why clang-format has changed Squid includes the way it did for [compat/xstring.cc](https://github.com/squid-cache/squid/compare/master...kinkie:clang-format#diff-b73ad949015aac149ab87ce6ebc9ffe6L10), for example. I suspect `SortIncludes` uses different rules (inside each priority group) than we currently do.

Aha! I had misunderstood what you were asking for, sorry.


> We can, of course, redefine the correct order as "whatever this particular configuration of this particular clang-format version does". However, before we do that, it would be nice to at least attempt to discover what that actually means.

Lol! No, that's not what's happening here.
TBH, I had missed that changed line, but here's my undertstanding of why it's happening:
clang-format automatically applies a priority 0 to what it considers 'the main include'. The main include is the one having the stem of the implementation file, possibly modified by the IncludeIsMainRegex option.

> Against my best judgement, I researched this a little instead of waiting for you to supply the missing details. Clang-format [documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) for `IncludeCategories` says that "The main header for a source file automatically gets category 0". This means that your current configuration changes the current include order by placing that "main header" right below `squid.h`. We might be able to avoid this change by using negative categories for everything, but I have not tested that workaround.

I can look into that. Or we might decide that the clang-format behaviour is intuitive and adopt it.

> Moreover, I suspect that clang-format comparison is case sensitive. I am guessing that could explain `include/unitTestMain.h` [changes](https://github.com/squid-cache/squid/compare/master...kinkie:clang-format#diff-eb18ae0860db5bcf8f8caa0e627e80bdR22), for example. I did not find the confirmation for that idea in the clang-format documentation, and clang-format _ignores_ case sensitivity when sorting `using` declarations, for example (see `SortUsingDeclarations`).

Running a test with input file:
```
#include "foo.h"
#include "bar.h"
#include "a/foo.h"
#include "b/foo.h"
#include "a/bar.h"
#include "b/bar.h"
#include "a/b/foo.h"
#include "a/c/bar.h"
#include "Foo.h"
```
yields as result:
```
#include "Foo.h"
#include "a/b/foo.h"
#include "a/bar.h"
#include "a/c/bar.h"
#include "a/foo.h"
#include "b/bar.h"
#include "b/foo.h"
#include "bar.h"
#include "foo.h"
```
From which I desume:
- sorting is case-sensitive
- slashes are ignored in defining the sort order
This behaviour doesn't seem to be configureable

This is consistent with the behaviour you observed in include/unitTestMain.h

rousskov:
> sorting is case-sensitive

Yes, probably.

> slashes are ignored in defining the sort order

I see the opposite in your test example (e.g., `bar.h` is below `b/foo.h` which would be viewed as  `bfoo.h` and would go lower if slashes were ignored), and I would be surprised if slashes are unnaturally ignored. Slash is probably treated as any other character.

> we might decide that the clang-format behaviour is intuitive and adopt it.

I would not object to switching to this new sorting order if there are good reasons to switch.

FWIW, IMHO, there relatively little "intuitive" about the concept of the "main" include file and especially the requirement to have it/them on the second place(s) while the remaining files are sorted alphabetically. The existence of multiple complicated clang-format options related to the "main file" concept seem to support my opinion. I bet the concept has technical merits in some esoteric cases, and I bet many projects adopted similar rules (without properly evaluating them) long time ago, forcing clang-format to support them.

We should keep in mind that "main files" concept complicates manual sorting a bit -- one cannot just select the `#include` statements below `squid.h` in any file and click on the "sort lines" command. Messy naming of Squid source files and overly literal interpretation of documentation would make this new rule particularly difficult to codify correctly. I was _not_ kidding about using the "whatever this particular configuration of this particular clang-format version does" rule instead of the usual English prose.

kinkie:
Is this (only loosely related) PR the best place to be discussing clang-format adoption and pros and cons?

> > sorting is case-sensitive
> 
> Yes, probably.
> 
> > slashes are ignored in defining the sort order
> 
> I see the opposite in your test example (e.g., `bar.h` is below `b/foo.h` which would be viewed as `bfoo.h` and would go lower if slashes were ignored), and I would be surprised if slashes are unnaturally ignored. Slash is probably treated as any other character.
> 
> > we might decide that the clang-format behaviour is intuitive and adopt it.
> 
> I would not object to switching to this new sorting order if there are good reasons to switch.

Squid is currently working fine, so it's not possible to come with any hard evidence. My main observation is that it may be more intuitive to keep interface and implementation more closely coupled than we do now.
I'm not going to insist on it however, and if we decide to adopt clang-format I'll try the approach you suggested to defeat clang-format's default behaviour.

> FWIW, IMHO, there relatively little "intuitive" about the concept of the "main" include file and especially the requirement to have it/them on the second place(s) while the remaining files are sorted alphabetically. The existence of multiple complicated clang-format options related to the "main file" concept seem to support my opinion. I bet the concept has technical merits in some esoteric cases, and I bet many projects adopted similar rules (without properly evaluating them) long time ago, forcing clang-format to support them.

Clang-format can support a number of different coding styles out of the box, so I don't think historic reasons are a key factor. This is however just speculation on my part and I'm not interested in exploring it.

> We should keep in mind that "main files" concept complicates manual sorting a bit -- one cannot just select the `#include` statements below `squid.h` in any file and click on the "sort lines" command. Messy naming of Squid source files and overly literal interpretation of documentation would make this new rule particularly difficult to codify correctly. I was _not_ kidding about using the "whatever this particular configuration of this particular clang-format version does" rule instead of the usual English prose.

I am not  a fan of being locked to a particular version of a tool, like we are with astyle right now.
clang-format covers a much bigger spectrum of code aspects, and as such it may be interesting for us.
As an exercise, I've tried formatting the sources with version 7, 8, 9, and 10 of clang-format (6.0 is quite ancient), released over the course of about 18 months. I've had  to remove a couple of configuration options compared to the proposed format file to be compatible with clang-format 7.

Diffstat are:
v7 -> v8: 3 files changed, 6 insertions(+), 6 deletions(-)
v8 -> v9: no changes
v9 -> v10:  6 files changed, 14 insertions(+), 14 deletions(-)

yadij:
* case-sensitivity
  The test performed is broken. The slash used is ASCII smaller than all alphanumerical characters. It should sort before letters in both case sensitive and insensitive systems.

A proper test needs to use upper and lower alphabet characters. Non-alpha characters sort based on locale settings, not case sensitivity.

* include order policy

The reason we have the current order is to reduce patch diff's from random developers. So any order that is easy for developers to anticipate/use for added includes is fine. All that is needed is consistent automatic mistake correction.

* tool version lock-in

The issue with tool differences is not whether version X+1 changes things version X missed. But whether any version X undoes changes version Y makes.

If that undo behaviour happens,  we are limited to particular "safe" version(s). That is the issue with astyle today.


IMO, the demonstrated ability to configure slightly better code formatting than astyle is one pro in favour of clang-format. So long as we do not get a regression in the other behaviours it would be reasonable to change. BUT, if there is a regression this is not enough of a benefit to change.

The removal of those end-of-file empty lines is a regression. Several of the tools that interact with Squid code either automatically add them in or complain with errors if there is not at least one CRLF after the final code line. That is a hard blocker for me with the current clang-format output.

rousskov:
> Kinkie: Is this (only loosely related) PR the best place to be discussing clang-format adoption and pros and cons?

Probably not. Your squid-dev RFC thread would be a more appropriate place.
 
> Alex: I would not object to switching to this new sorting order if there are good reasons to switch.

> Kinkie: Squid is currently working fine, so it's not possible to come with any hard evidence. 

I do not see the relevance of the above assertion. We strive to make Squid headers independent, and most of them probably are. Thus, virtually any order (with `squid.h` being first) should "work fine" from Squid building point of view (or can be easily fixed if needed).

> My main observation is that it may be more intuitive to keep interface and implementation more closely coupled than we do now.

Yes, I am sure that many would feel that the interface-first order is more intuitive. My primary concern is that it is very difficult to define "interface" precisely in many real cases, especially in Squid environment, for humans and programs alike. The more rules require human interpretation, and the more complex that interpretation is, the more wasteful conflicts we have. From conflict-avoidance and automation point of view, "[case-insensitive] alphabetical order" is a much better rule than "interface file(s) first and then [case-sensitive] alphabetical order".

> I am not a fan of being locked to a particular version of a tool, like we are with astyle right now.

There is a significant difference between the formatting rule being based on version X of the tool and being forced to use version X. When everything is fully automated, a developer can use _any_ tool version as long as the result complies with the rule. For stable, well-written, and properly configured tools, upgrades (or reasonable version differences) are not a problem at all. Astyle was not such a tool when we started using it. Based on your clang-format experiments, clang-format is not quite there yet either (or our configuration does not cover some formatting aspects that it should cover).

> clang-format covers a much bigger spectrum of code aspects, and as such it may be interesting for us.

Yes, it is an argument for switching to clang-format if we are sure that modern astyle cannot control those formatting aspects. To avoid misunderstanding: There is no doubt in my mind that clang-format has the potential to be much better than astyle (because of what kind of information a clang-based tool has access to). I just do not know how much of that potentially available information clang-format is using _now_ and what the current clang-format implementation quality is. The potential is certainly there!

> Amos: The issue with tool differences is not whether version X+1 changes things version X missed. But whether any version X undoes changes version Y makes.

I would not formulate it like this. IMO, the primary "issue" here is any surprise/uncontrolled change of official Squid sources. Once we fully automate formatting, we do not want Squid sources to suddenly change because somebody ran "sudo apt full-upgrade" on the CI box. And before that full automation goal is reached, we want to reduce surprises and reformatting during PR review/merging process.

Again, this does not mean that others cannot use other tool versions. It means that if they do, they may face formatting changes down the road. We want to minimize that.

kinkie:
> > Kinkie: Is this (only loosely related) PR the best place to be discussing clang-format adoption and pros and cons?
> 
> Probably not. Your squid-dev RFC thread would be a more appropriate place.

Let's do so: I'll create a new PR with the config file only, and we can discuss over that. I'll take care of quoting the conversation.

PR #601, opened 2020-04-12
Update SPONSORS

kinkie:


yadij:
Thanks for the update. As per discussion at the last board meeting I'm doing these updates as part of the next release prep. It would have been done already but was held up the recent Anubis issues to be resolved. Have resumed work now and there should be a release-prep PR to supersede this in the next few days.

kinkie:
Abandoning, as Amos has this in his sights.

PR #605, opened 2020-04-16
Preserve half-closed transaction context

eduard-bagdasaryan:
The code monitoring half-closed connections (with half_closed_clients
"on") did not save and did not restore transaction contexts.

### DIFF
--- a/src/comm.cc
+++ b/src/comm.cc
@@ -1639,6 +1639,7 @@ commStartHalfClosedMonitor(int fd)
     debugs(5, 5, HERE << "adding FD " << fd << " to " << *TheHalfClosed);
     assert(isOpen(fd) && !commHasHalfClosedMonitor(fd));
     (void)TheHalfClosed->add(fd); // could also assert the result
+    fd_table[fd].codeContext = CodeContext::Current();
     commPlanHalfClosedCheck(); // may schedule check if we added the first FD
 }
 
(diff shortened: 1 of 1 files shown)


src/comm.cc
```diff
@@ -1666,10 +1667,12 @@ commHalfClosedCheck(void *)
         Comm::ConnectionPointer c = new Comm::Connection; // XXX: temporary. make HalfClosed a list of these.
         c->fd = *i;
         if (!fd_table[c->fd].halfClosedReader) { // not reading already
-            AsyncCall::Pointer call = commCbCall(5,4, "commHalfClosedReader",
-                                                 CommIoCbPtrFun(&commHalfClosedReader, NULL));
-            Comm::Read(c, call);
-            fd_table[c->fd].halfClosedReader = call;
+            CallBack(fd_table[c->fd].codeContext, [c] {
```
rousskov, line 1670: Can we avoid refcounting overheads here and help the compiler to inline this lambda?

```suggestion
            CallBack(fd_table[c->fd].codeContext, [&c] {
```

Sorry I did not realize that `c` is a smart pointer when looking at this code earlier.

rousskov:
```
15:09:30 docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```

@kinkie, please kick the daemon if it needs to be kicked.

src/comm.cc
eduard-bagdasaryan: Done.

PR #607, opened 2020-04-17
Preserve caller context in commHandleWriteHelper()

eduard-bagdasaryan:
This event handler resumes write operations for descriptors,
queued due to delay pool restraints. Before this fix, the enqueuing
code did not save and the dequeuing code did not restore transaction
contexts.


yadij:
Accepted under 10-day criteria.

PR #611, opened 2020-04-20
Bug 5041: Missing Debug::Extra breaks build on hosts with systemd

rousskov:
Master commit 6fa8c66 (i.e. Bug 5016 fix) relied on Debug::Extra added
by master commit (ccfbe8f) that was not ported to v4. The port of the
former commit lacked the required piece of the latter commit, breaking
v4 build since v4 commit 62f38c1.

The problem is invisible on hosts without a systemd package (that Squid
can find/use) and with Squids explicitly ./configured --without-systemd.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2162,6 +2162,7 @@ if test "x$with_systemd" != "xno" -a "x$squid_host_os" = "xlinux"; then
   fi
   if test "x$SYSTEMD_LIBS" != "x" ; then
     CXXFLAGS="$SYSTEMD_CFLAGS $CXXFLAGS"
+    LDFLAGS="$SYSTEMD_LIBS $LDFLAGS"
     AC_DEFINE(USE_SYSTEMD,1,[systemd support is available])
   else
     with_systemd=no
--- a/src/Debug.h
+++ b/src/Debug.h
@@ -99,6 +99,10 @@ public:
 
     /// configures the active debugging context to write syslog ALERT
     static void ForceAlert();
+
+    /// prefixes each grouped debugs() line after the first one in the group
+    static std::ostream& Extra(std::ostream &os) { return os << "\n    "; }
+
 private:
     static Context *Current; ///< deepest active context; nil outside debugs()
 };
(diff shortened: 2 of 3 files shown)


rousskov:
I hope this helps, but I cannot test this fix quickly because my build system lacks the right package.

yadij:
Can you add --without-systemd to the minimal build layer tests please. Those were omitted as well and should have allowed the build farm to catch this.

rousskov:
> --without-systemd [in] the minimal build layer tests ... should have allowed the build farm to catch this.

Done in b164f12 but, AFAICT, `--without-systemd` will not allow us to catch this bug. What will allow us to catch it is adding the systemd package to at least one of the Jenkins nodes. I [asked](https://bugs.squid-cache.org/show_bug.cgi?id=5041#c1) @kinkie to look into that.

Please also note that this PR fixes a v4 bug -- there is no build problem in master. Somebody will need to make the corresponding minimal build layer changes (which are correct/welcomed) in master as well. Or let me know and I can undo them in this PR.

yadij:
Testing this patch today brought up another change that is necessary to pass the unit tests once kinkie updates the farm libraries.

diff --git a/configure.ac b/configure.ac
index 9d1a38c4f..281d237bc 100644
--- a/configure.ac
+++ b/configure.ac
@@ -2162,6 +2162,7 @@ if test "x$with_systemd" != "xno" -a "x$squid_host_os" = "xlinux"; then
   fi
   if test "x$SYSTEMD_LIBS" != "x" ; then
     CXXFLAGS="$SYSTEMD_CFLAGS $CXXFLAGS"
+    LDFLAGS="$SYSTEMD_LIBS $LDFLAGS"
     AC_DEFINE(USE_SYSTEMD,1,[systemd support is available])
   else
     with_systemd=no

rousskov:
> Testing this patch today brought up another change that is necessary to pass the unit tests once kinkie updates the farm libraries.

```diff
+    LDFLAGS="$SYSTEMD_LIBS $LDFLAGS"
```

Do you want me to add that change to the v4 PR branch? BTW, please feel free to commit anything to that branch yourself if that will speed things up!

yadij:
Yes please. I just get errors about not being authorized to commit.

rousskov:
> Yes please.

Done at faa2686.

PR #613, opened 2020-04-21
Cleanup: remove xusleep() hacks

yadij:
usleep() has been obsolete for a long time. C++11 provides better and
portable waiting mechanism we can use easily instead of these hacks
trying to replicate and extend usleep().


src/base/File.cc
```diff
@@ -331,7 +332,7 @@ File::lock(const FileOpeningConfig &cfg)
                    " more time(s) after a failure: " << ex.what());
         }
         Must(attemptsLeft); // the catch statement handles the last attempt
-        xusleep(cfg.RetryGapUsec);
+        std::this_thread::sleep_for(std::chrono::microseconds(cfg.RetryGapUsec));
```
kinkie, line 335: Would it make sense to have an inline xusleep making this call less verbose?

yadij: I would like to eventually have the object members store the std::chrono duration types. But that is a lot more work and the scope here is just removing the xusleep hacks.

yadij:
Waiting on a clean Jenkins result before clearing this one for merge. FreeBSD (GCC 4.8) was suddenly having issues with linking static variables passed to the new template.

src/base/File.cc
yadij, line 335: Clang on CentOS 7 is having problems with this line:
```
/usr/bin/../lib/gcc/x86_64-redhat-linux/4.8.5/../../../../include/c++/4.8.5/chrono:263: undefined reference to `FileOpeningConfig::RetryGapUsec'
```

I am inclined to ignore this for Squid-6 change. CentOS 7 is EOL in 4 months, Squid-6 is not even going to be in beta by then, and GCC builds fine on the same system.

rousskov: I doubt it is a good idea to ignore linking errors just because CentOS 7 is going to be EOL. It is possible that this or a similar issue will appear in other environments we do not know about yet. Is it possible to understand what is going on (i.e. find the cause of the failure and investigate whether there are fixes/workarounds) before we decide ignore this error and effectively remove CentOS 7 from CI tests?

rousskov: BTW, is it normal/expected that a clang build error refers to GCC `#include` files? Perhaps we can solve the problem by avoiding such a hybrid clang+gcc use?

yadij: I expect so. Being STL definitions they should be portable headers.

@kinkie, can you confirm nothing special was needed or done to install clang on that node? Just a package install like any user would do.

rousskov: > I expect so. Being STL definitions they should be portable headers.
 
In general, I do not see why a GCC implementation of STL interfaces has to be portable to other compilers. Such implementations may use compiler-specific tricks. On the other hand, as I have said [elsewhere](https://github.com/squid-cache/squid/pull/587#issuecomment-619204004), AFAICT, clang does support, in principle, building with GCC (or other) STL headers. However, clang [documentation](https://clang.llvm.org/get_started.html) mentions `weird link-time errors relating to inline functions`.

If we have to pick one option out of two, then it may be better to use clang-provided headers for clang builds than to use GCC-provided headers for clang builds, regardless of whether both options should work in principle.

kinkie: > I expect so. Being STL definitions they should be portable headers.
> 
> @kinkie, can you confirm nothing special was needed or done to install clang on that node? Just a package install like any user would do.

The install scripts are at https://github.com/kinkie/dockerfiles .

The images to test builds on prem, if you want to, are available from the docker hub as squidcache/buildfarm:<arch>-<distro>-<version> (e.g. squidcache/buildfarm:aarch64-ubuntu-focal)

yadij:
Leaving this PR open for merge in future. Pending when the CentOS 7 clang build environment stops being an issue.

kinkie:
updating and triggering a refresh of the jenkins build

yadij:
Rebased to resolve conflicts and kick fresh build tests.

Removed M-waiting-for-distro as Squid-6 requirement for C++17 support mandates support for the C++11 functionality this requires.

rousskov (approved):
Still LGTM.

PR #616, opened 2020-04-24
Add missing libsystemd build logic

yadij:
* "Minimum build" test should run --without-systemd

* Add -lsystemd to linker flags

Completes the changes added in GitHub PR 539 (commit
6fa8c66435d55a2e713db0222cdca3a9dccf5bbe).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2008,22 +2008,7 @@ esac
 AC_SUBST(LDAPLIB)
 AC_SUBST(LBERLIB)
 
-AC_ARG_WITH(systemd,
-  AS_HELP_STRING([--without-systemd],
-                 [Do not use systemd API to send start-up completion
-                  notification. Default: auto-detect]), [
-case "$with_systemd" in
-  yes|no)
-    : # Nothing special to do here
-    ;;
-  *)
-    if test ! -d "$withval" ; then
-      AC_MSG_ERROR([--with-systemd path does not point to a directory])
-    fi
-    SYSTEMD_PATH="-L$with_systemd/lib"
-    CPPFLAGS="-I$with_systemd/include $CPPFLAGS"
-  esac
-])
+SQUID_AUTO_LIB(systemd,[systemd API for start-up notification],[SYSTEMD])
 AH_TEMPLATE(USE_SYSTEMD,[systemd support is available])
 if test "x$with_systemd" != "xno" -a "x$squid_host_os" = "xlinux"; then
   SQUID_STATE_SAVE(squid_systemd_state)
@@ -2046,7 +2031,7 @@ if test "x$with_systemd" != "xno" -a "x$squid_host_os" = "xlinux"; then
   fi
   if test "x$SYSTEMD_LIBS" != "x" ; then
     CXXFLAGS="$SYSTEMD_CFLAGS $CXXFLAGS"
-    LDFLAGS="$SYSTEMD_LIBS $LDFLAGS"
+    LDFLAGS="$SYSTEMD_PATH $SYSTEMD_LIBS $LDFLAGS"
     AC_DEFINE(USE_SYSTEMD,1,[systemd support is available])
   else
     with_systemd=no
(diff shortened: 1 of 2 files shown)


marcosfrm:
This causes overlinking, no?

`-lsystemd` is added already by `PKG_CHECK_MODULES` (its internal `AC_SUBST`), setting the flag in `src/Makefile`'s `squid_LDADD`.

rousskov:
> This causes overlinking, no?

I have not tested it, but I suspect that you are right. To fix this overlinking, we should either remove `$SYSTEMD_LIBS` from `LDFLAGS` or remove `$SYSTEMD_LIBS` from `squid_LDADD`. I have not studied this, but I suspect the former is better because the latter would still supply an unused library to other executables.

@yadij, please consider fixing this if you agree that there is a problem here (ignoring my specific suggestions if they are incorrect).

@marcosfrm, FWIW, I suspect that your explanation below is wrong:

> -lsystemd is added already by PKG_CHECK_MODULES, setting the flag in src/Makefile's squid_LDADD.

AFAICT, the side effects of `PKG_CHECK_MODULES` are undone by  `SQUID_STATE_ROLLBACK`. The actual cause of this overlinking is the `$SYSTEMD_LIBS` component inside the following configure.ac line (added by this PR):

```
LDFLAGS="$SYSTEMD_PATH $SYSTEMD_LIBS $LDFLAGS"
```

I also find the current explicit inclusion of `$SYSTEMD_PATH` suspicious because pkg-config [documentation](https://people.freedesktop.org/~dbn/pkg-config-guide.html) does not mention `X_PATH`. I would expect that `-L` would be a part of `X_LIBS` just like `pkg-config --libs` includes `pkg-config --libs-only-L`.

BTW, similar arguments apply to the old `CXXFLAGS` adjustment AFAICT: We should not pollute the global flags just because one of the executables needs `SYSTEMD_CFLAGS`.

Disclaimer: This is not my area of expertise.

yadij:
> > This causes overlinking, no?
> 
> I have not tested it, but I suspect that you are right. To fix this overlinking, we should either remove `$SYSTEMD_LIBS` from `LDFLAGS` or remove `$SYSTEMD_LIBS` from `squid_LDADD`. I have not studied this, but I suspect the former is better because the latter would still supply an unused library to other executables.
> 
> @yadij, please consider fixing this if you agree that there is a problem here (ignoring my specific suggestions if they are incorrect).

LDFLAGS was polluted due to many of the unit test binaries pulling in the client_side.cc code and thus needing the linkage. During review (record of which no longer exists in github!) you declared it a better thing to link globally than have the patch adding SYSTEMD_LIBS to each relevant *_LDADD list.

I am not worried by this because of the --as-needed flag usually built with drops unnecessary over-linkage. Plus the future configure.ac cleanup plans will remove/reduce this type of thing from many libraries which share this situation.

> 
> @marcosfrm, FWIW, I suspect that your explanation below is wrong:
> 
> > -lsystemd is added already by PKG_CHECK_MODULES, setting the flag in src/Makefile's squid_LDADD.
> 
> AFAICT, the side effects of `PKG_CHECK_MODULES` are undone by `SQUID_STATE_ROLLBACK`. The actual cause of this overlinking is the `$SYSTEMD_LIBS` component inside the following configure.ac line (added by this PR):
> 

No, he is right. The squid_LDADD list in Makefile explicitly adds $SYSTEMD_LIBS. Then LDFLAGS adds it a second time to the squid binary compile line.

rousskov:
> During review you declared it a better thing to link globally than have the patch adding SYSTEMD_LIBS to each relevant *_LDADD list.

I do not recall making such a declaration. You might be thinking about the following thread, but there my suggestion was the opposite of both "link globally" and "[manually] add to each relevant *_LDADD": https://github.com/squid-cache/squid/pull/587#discussion_r415330675

In case of SYSTEMD_LIBS, I do not know of a better option than to "manually add to each relevant *_LDADD" because there is no convenience library that can depend on SYSTEMD_LIBS.

> The squid_LDADD list in Makefile explicitly adds $SYSTEMD_LIBS. Then LDFLAGS adds it a second time to the squid binary compile line.

Yes, we are in agreement here. I think the second part is a mistake caused by the ./configure line I have pointed to.

PR #620, opened 2020-04-30
SslBump: Disable OpenSSL TLSv1.3 support for older TLS traffic

chtsanti:
This change fixes stalled peeked-at during step2 connections from IE11
and FireFox v56 running on Windows 10 (at least), producing "Handshake
with SSL server failed" cache.log errors with this OpenSSL detail:

`SSL routines:ssl_choose_client_version:inappropriate fallback`

Disabling TLS v1.3 support for older TLS connections is required
because, in the affected environments, OpenSSL detects and, for some
unknown reason, blocks a "downgrade" when a server claims support for
TLS v1.3 but then accepts a TLS v1.2 connection from an older client.

Based on master commit cd29a421.

### DIFF
--- a/src/anyp/ProtocolVersion.h
+++ b/src/anyp/ProtocolVersion.h
@@ -40,6 +40,9 @@ public:
     unsigned int major;    ///< major version number
     unsigned int minor;    ///< minor version number
 
+    /// whether the version is "known" (e.g., has been parsed or explicitly set)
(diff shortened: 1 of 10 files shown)


chtsanti:
This is the v5 version of the #588 PR.
In the cases does not apply cleanly to v5.

PR #623, opened 2020-05-03
Bug 5046: FreeBSD lacks open(2) O_DSYNC flag

yadij:
ext_session_acl built with TrivialDB uses O_DSYNC to ensure
thread-safe manipulation of data within the TDB files in Squid
multi-process environment.

FreeBSD lacks this flag entirely. Use the O_SYNC flag as a
backup, which apparently provides file-level synchronization.
It is not clear whether this flag will prevent duplicate keys or
record overwrites in the case of process write race collisions.

NP: this appears to be FreeBSD specific. Other BSD either define
O_DSYNC or lack support for these POSIX flags entirely.

### DIFF
--- a/src/acl/external/session/ext_session_acl.cc
+++ b/src/acl/external/session/ext_session_acl.cc
@@ -144,7 +144,6 @@ static void init_db(void)
     db = tdb_open(db_path, 0, TDB_CLEAR_IF_FIRST, O_CREAT|O_DSYNC, 0666);
 #endif
     if (!db) {
-    db = tdb_open(db_path, 0, TDB_CLEAR_IF_FIRST, , 0666);
         fprintf(stderr, "FATAL: %s: Failed to open session db '%s'\n", program_name, db_path);
         shutdown_db();
         exit(EXIT_FAILURE);
--- a/src/acl/external/session/ext_session_acl.cc
+++ b/src/acl/external/session/ext_session_acl.cc
@@ -137,15 +137,14 @@ static void init_db(void)
         }
     }
 #elif USE_TRIVIALDB
-#if _SQUID_FREEBSD_
+#if _SQUID_FREEBSD_ && !defined(O_DSYNC)
     // FreeBSD lacks O_DSYNC, O_SYNC is closest to correct behaviour
-    const auto openFlags = O_CREAT|O_SYNC;
-#else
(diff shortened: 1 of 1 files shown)


src/acl/external/session/ext_session_acl.cc
```diff
@@ -137,7 +137,13 @@ static void init_db(void)
...
     }
 #elif USE_TRIVIALDB
-    db = tdb_open(db_path, 0, TDB_CLEAR_IF_FIRST, O_CREAT|O_DSYNC, 0666);
+#if _SQUID_FREEBSD_
+    // FreeBSD lacks O_DSYNC, O_SYNC is closest to correct behaviour
+    const auto openFlags = O_CREAT|O_SYNC;
+#else
+    const auto openFlags = O_CREAT|O_DSYNC;
+#endif
+    db = tdb_open(db_path, 0, TDB_CLEAR_IF_FIRST, openFlags, 0666);
```
rousskov, line 146: We do not have to duplicate the `O_CREAT` part of the opening flags logic, right?

```suggestion
    const auto syncFlags = O_SYNC;
#else
    const auto syncFlags = O_DSYNC;
#endif
    db = tdb_open(db_path, 0, TDB_CLEAR_IF_FIRST, O_CREAT|syncFlags, 0666);
```

yadij: We could, but AFAICT they way I did it allows the compiler to more easily deduce whether the parameter is constexpr.

timp87:
Works for me.

src/acl/external/session/ext_session_acl.cc
rousskov: > they way I did it allows the compiler to more easily deduce whether the parameter is constexpr.

Your call, but I bet modern compilers will optimize both versions the same way: https://godbolt.org/z/2YJqg5

> One similar alternative would be a local #define for O_DSYNC instead of the local constant.

That sounds like the best option to me!

And just in case it was not obvious from the lack of a negative review with a change request: I do _not_ insist on any of these changes.

yadij: Okay, Done.

rousskov: Thank you. Approving in anticipation of the double-tdb_open() typo fix.

PR #632, opened 2020-05-13
Fix PoolingAllocator build errors with older GCCs

rousskov:
error: no class template named rebind in class PoolingAllocator

GCC v4.8.4 (at least) does not fully implement C++11 Allocator-related
APIs, forcing the developer to manually provide Allocator traits that
are supposed to come automatically via std::allocator_traits.

The problem may only manifest itself when using a custom allocator with
types such as std::map<T> and std::list<T> that allocate wrapper
objects instead of Ts. For example, std::vector is unaffected.


rousskov:
In my tests, this PR fixes GCC v4.8 builds of code using PoolingAllocator with std::map and std::list. There is no such code in master yet, but several pending refactoring attempts and new developments need it. Semaphore CI uses GCC v4.8.4 and is affected.

PR #633, opened 2020-05-14
Fix sending of unknown validation errors to cert. validator

chtsanti:
Squid may be compiled with an OpenSSL release introducing X509
validation errors that Squid does not have the names for. Send their
integer codes.

Also sync Squid certificate verification errors with OpenSSL v1.1.1g.


yadij:
Updated the comment to say "does not have the names for" since this code does *not* appear to send codes for any error if squid has the name, either hard coded or configured by admin.

The PR title indicates this is fixing some bug. But there is no mention of what the problem actually is, nor how this PR changes behaviour - the code is already being sent and there are no visible changes to the helper protocol in this PR.
 Also, if a user has a cert validator which needs to get this patch; what symptom(s) are they going to be complaining about?

Please point out if I have missed anything, but the only logic change I see is an increase to the set of error codes with known names and/or marked as 'optional'.  That alone is not "fixing" anything for unknown future codes, nor changing *how* they are sent to the helper (as title currently claims).

chtsanti:
> Thank you for fixing this! Before clearing this for merging, in the PR description, please mention which OpenSSL version you synced the errors with (instead of just saying "latest").

I used OpenSSL-1.1.1d but it is the same for OpenSSL-1.1.1g(now latest)

PR #634, opened 2020-05-14
Reduced startup time with large rock cache_dirs

eduard-bagdasaryan:
... addressing an old TODO.

Before scanning the disks to find the actual entries, the old rock
cache_dir initialization code had to populate its index as if all disk
slots were available and then remove all the added entries. Those two
wasteful operations took ~1.5 seconds for a 200GB disk before the
PageStack ABA bug was fixed in a586085. With the tree-based fix, that
time increased to ~15 seconds. The delay is completely gone now,
reducing the total index initialization time (for a 200GB disk) down to
a second.

### DIFF
--- a/src/ipc/mem/PageStack.h
+++ b/src/ipc/mem/PageStack.h
@@ -126,7 +126,7 @@ public:
         size_t pageSize = 0; ///< page size, used to calculate shared memory size
         PageCount capacity = 0; ///< the maximum number of pages
 
-        /// whether a newly created PageStack should contain capacity IDs
+        /// whether a newly created PageStack should be prefilled with PageIds
         bool createFull = false;
     };
 
--- a/src/ipc/mem/PageStack.h
+++ b/src/ipc/mem/PageStack.h
@@ -116,10 +116,24 @@ public:
     /// the number of (free and/or used) pages in a stack
     typedef unsigned int PageCount;
 
-    PageStack(const PoolId aPoolId, const PageCount aCapacity, const size_t aPageSize);
+    // XXX: poolId, pageSize look misplaced due to messy separation of PagePool
+    // (which should support multiple Segments but does not) and PageStack
+    // (which should not calculate the Segment size but does) duties.
+    /// PageStack construction and SharedMemorySize calculation parameters
+    class Config {
+    public:
+        uint32_t poolId = 0; ///< pool ID
+        size_t pageSize = 0; ///< page size, used to calculate shared memory size
+        PageCount capacity = 0; ///< the maximum number of pages
 
-    PageCount capacity() const { return capacity_; }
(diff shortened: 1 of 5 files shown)


src/ipc/mem/PageStack.h
```diff
@@ -116,10 +116,24 @@ class PageStack
...
+    // (which should support multiple Segments but does not) and PageStack
+    // (which should not calculate the Segment size but does) duties.
+    /// PageStack construction and SharedMemorySize calculation parameters
+    class Config {
+    public:
+        uint32_t poolId = 0; ///< pool ID
+        size_t pageSize = 0; ///< page size, used to calculate shared memory size
+        PageCount capacity = 0; ///< the maximum number of pages
+
+        /// whether a newly created PageStack should contain capacity IDs
```
yadij, line 129: What are "capacity IDs"? they do not seem to be defined.

eduard-bagdasaryan: These are IDs below the requested capacity. There is some explanation in IdSet::makeFullBeforeSharing().

eduard-bagdasaryan:
> what role the "createFull" flag has

This flag allows initial PageStack configuring either with all pages free (true - for MemStore) or full (false - for Rock).

>  the difference between old and new algorithms

Before a586085 (the tree-based PageStack implementation), Rock::SwapDirRr::create() method popped all elements from the newly created PageStack (the loop below the TODO). However, I assume that this loop was fast - it just ran across the page array without any linear/binary search. After a586085, the loop was slowed down by the binary search. This PR eliminates this popping loop at all - and substitutes it with a much faster IdSet::makeFullBeforeSharing().
In other words, the benefit of this change became evident after a586085 (see the PR preamble timings). It is less evident compared with pre-a586085, because the wasteful loop was fast enough.

rousskov:
> the wasteful loop was fast enough

Yes, it was fast enough to leave its optimization as a TODO. Wasting a couple of startup seconds was still annoying. This PR removes that waste for good.

src/ipc/mem/PageStack.h
rousskov: Actually, there are no "capacity IDs".

There are many ways to parse createFull description. The correct way is "whether a newly created PageStack should contain n IDs (where n is equal to Config::capacity)". In branch commit eb6fd09, I adjusted the phrase in hope to decrease the number of ways it can be misparsed. Please feel free to adjust further, of course.

rousskov:
@yadij, please earmark this PR (commit f670688) and #587 (commit a586085) fixes for Squid v5 inclusion. Please let me know if you need help with backporting.

Squid v4 still suffers from the bug (badly) fixed in v5 commit c3408a3, so v4 does not need these followup fixes.

PR #636, opened 2020-05-17
Fix Linux build with glibc v2.30 which deprecated sys/sysctl.h

rousskov:
error: #warning "The <sys/sysctl.h> header is deprecated and
    will be removed." [-Werror=cpp]

Squid only uses sysctl() on FreeBSD and friends so let's not #include
sys/sysctl.h in other environments. The existing compat/shm.cc already
uses a similar condition for including this header.

Glibc recommends using /proc instead of the sysctl.h they deprecated,
but that is probably not an option for FreeBSD anyway.

### DIFF
--- a/src/eui/Eui48.cc
+++ b/src/eui/Eui48.cc
@@ -45,8 +45,10 @@ struct arpreq {
 /* required by Solaris */
 #include <sys/sockio.h>
 #endif
+#if _SQUID_FREEBSD_ || _SQUID_NETBSD_ || _SQUID_OPENBSD_ || _SQUID_DRAGONFLY_ || _SQUID_KFREEBSD_
 #if HAVE_SYS_SYSCTL_H
 #include <sys/sysctl.h>
+#endif
 #endif
 #if HAVE_NET_ROUTE_H
 #include <net/route.h>


rousskov:
I tested this fix on Linux (where the build used to fail). I cannot test on FreeBSD/etc. at the moment, but since the added `#include` condition is identical to the one where sysctl() is actually [used](https://github.com/squid-cache/squid/blob/a6bed9ab54c2dfdc79dca7eb51d4eb8b9a026efb/src/eui/Eui48.cc#L342), I hope the change does not break FreeBSD/etc.

rousskov:
Glibc release notes mentioning sysctl.h deprecation are available [elsewhere](https://lwn.net/Articles/795127/).

yadij:
The configure.ac logic is supposed to prevent HAVE_SYS_SYSCTL_H being defined on systems where the file is not usable. IMO the fix needs to be done to that code.

rousskov:
IMO, this build-fixing PR is acceptable because it is common for Squid to condition `#include` on header usage in the code following that `#include`. One could even argue that making such includes conditional is a good approach (although I am sure it can be abused if applied too aggressively/literally).

Is this PR the best solution? Probably not. I know it is the best solution I can volunteer right now. I do not insist on this PR acceptance and would welcome any better solution. It is your call.

----

> The configure.ac logic is supposed to prevent HAVE_SYS_SYSCTL_H being defined on systems where the file is not usable.

After reading your comment, I checked what happens to `HAVE_SYS_SYSCTL_H` in my environment and discovered that it is defined because the standard `AC_CHECK_HEADERS` function does not use our custom compiler flags such as `-Werror`. 

```
configure:7969: strict error checking enabled: yes
...
configure:18780: checking for sys/sysctl.h
configure:18780: g++ -c -g -O2 -march=native  conftest.cpp >&5
/usr/include/x86_64-linux-gnu/sys/sysctl.h:21:2: warning: #warning "The <sys/sysctl.h> header is deprecated and will be removed." [-Wcpp]
configure:18780: $? = 0
configure:18780: result: yes
...
configure:39366: BUILD EXTRA C++ FLAGS: -Wall ... -Werror ...
```

That `g++` line 18780 lacks `-Werror`. The same problem applies to other header checks, of course, but I did not see any other ignored warnings in my environment. Hopefully, somebody will investigate that further and fix (without breaking other code that might implicitly depend on this bug). I cannot volunteer to do that right now.

Moreover, since Squid does not set `DISTCHECK_CONFIGURE_FLAGS`, the flags on that `g++` line may _change_ when `./configure` is executed again by `make distcheck`. I think that is another bug, but I do not have the time to fight for its recognition. That change is what triggered the discovery of the build error in my environment and this PR.

rousskov:
Closing after discovering #511 and Squid bug [4985](https://bugs.squid-cache.org/show_bug.cgi?id=4985). The committed fix (master commit 195ec84) is not the best either IMO, but I was not given a chance to review it back then and do not have the time to improve it now.

PR #637, opened 2020-05-18
Move ClientActiveRequests feature to dedicated source files

kinkie:


src/dlink.h
```diff
@@ -25,10 +25,13 @@ class dlink_list
 public:
     dlink_node *head = nullptr;
     dlink_node *tail = nullptr;
+    void add(void *data, dlink_node *node);
+    void addAfter(void *data, dlink_node *node, dlink_node *reference);
+    void addTail(void *data, dlink_node *node);
+    void moveToFront(dlink_node *node);
+    void remove(dlink_node *node);
```
rousskov, line 32: >  Add class methods to dlink_list

Please remove these changes from this PR.

IMO, adding whole-list-manipulation methods to an individual list element is a step in the wrong direction because it obscures the different roles those two concepts (i.e. the whole list and an individual element) play. It makes wrong-role mistakes (e.g., supplying an inner node instead of the list head) easier to make and more difficult to spot during review.

We can (and probably should) convert the dlink_list (i.e., invasive list with nodes that may not belong to any list) API to modern C++. However, doing so requires a different approach and should be done in a stand-alone PR. An ideal PR would also minimize existing caller changes if such minimization is possible, but I have not research that possibility. If you decide to write such a dedicated PR, I recommend posting the proposed class declarations first along with a couple of representative user code adjustments, before syncing the rest of Squid code.

P.S. I did not review individual method implementations.

src/ClientActiveRequests.h
```diff
@@ -0,0 +1,16 @@
...
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID_CLIENTACTIVEREQUESTS_H
+#define SQUID_CLIENTACTIVEREQUESTS_H
+
+#include "dlink.h"
+
+extern dlink_list ClientActiveRequests;
```
rousskov, line 14: If you know the answer, please document what ClientActiveRequests are.

src/Makefile.am
```diff
@@ -234,6 +234,8 @@ squid_SOURCES = \
 	carp.cc \
 	cbdata.cc \
 	cbdata.h \
+	ClientActiveRequests.h \
+	ClientActiveRequests.cc \
```
rousskov, line 238: This list is inconsistent, so I do not insist on this change, but I believe this list was meant to be alphabetized:

```suggestion
	ClientActiveRequests.cc \
	ClientActiveRequests.h \
```

Same for other lists.

src/Makefile.am
```diff
@@ -3112,6 +3122,8 @@ tests_testEventLoop_SOURCES = \
 	carp.h \
 	tests/stub_carp.cc \
 	cbdata.cc \
+	ClientActiveRequests.h \
+	ClientActiveRequests.cc \
```
rousskov, line 3126: Does testEventLoop and other similar cases actually use the ClientActiveRequests structure? If not, can it be provided via a stub file rather than explicitly adding the ClientActiveRequests concept inside unrelated test cases?

src/Makefile.am
rousskov, line 238: On the surface, adding two source files just to declare a global variable feels like an overkill. Have you considered adding that global to the existing files instead (e.g., client_side_request.{cc,h})?

src/stat.cc
```diff
@@ -11,6 +11,7 @@
 #include "squid.h"
 #include "CacheDigest.h"
 #include "CachePeer.h"
+#include "ClientActiveRequests.h"
```
rousskov, line 14: AFAICT, we should move statClientRequests() (and its cache manager registration) from `stat.cc` into the new ClientActiveRequests home instead. I think this move should be done in any case, but I only insist on this move if you insist on adding new source files for ClientActiveRequests -- if you want to treat ClientActiveRequests as a stand-alone "feature", let's not yank an important piece of it out.

src/client_side_reply.cc
```diff
@@ -1905,9 +1906,7 @@ clientReplyContext::SendMoreData(void *data, StoreIOBuffer result)
 void
 clientReplyContext::makeThisHead()
 {
-    /* At least, I think that's what this does */
```
rousskov, line 1908: If you have not studied this subject to confirm that move-to-front is what the original code did (before the code was refactored and the comment was added), I would leave this valuable (but poorly worded) comment -- it illustrates that the refactoring author did not know for sure what this code is supposed to do. Future refactoring or a bug triage should study and address this uncertainty.

src/dlink.h
kinkie: Maybe there is a misunderstanding. These class methods are essentially aliases for the free ones; they are so far only syntactic sugar. I noticed that it is a common pattern for dlink_list users is to have the dlink_node as a data member and to use it e.g. for O(1) LRU.
In code:
```
struct foo {
  dlink_node dl;
}
// ..
void foo::some_method() {
  // free function
  dlink_add(this, &dl, &some_list);
  // member function, same effect
  some_list.add(this, &dl);
}
```

src/ClientActiveRequests.h
kinkie: Ok

src/Makefile.am
kinkie: We are not very consistent there at all. Changing, but something more generic may be needed.
I'll see if I can cook up some script or similar.

src/Makefile.am
kinkie: Maybe; I'll check it.

src/Makefile.am
kinkie: It us used by both request and reply. I believe having an own file increases clarity and consistency; it used to be in dlink_list.cc :(

src/dlink.h
rousskov: Please interpret my change request as if I knew that "These class methods are essentially aliases for the free ones". I did not check, but that was indeed my assumption for all the methods but moveToFront(). That single exception is irrelevant here.

There are often many ways to achieve the same assembly code. In most non-trivial cases, very few of those ways are actually the right thing to do at the C++ level. The proposed way is not the right thing (for the reasons stated earlier), and the invasive list API problem itself deserves a dedicated PR (even if you disagree with those earlier reasons).

src/Makefile.am
rousskov: > I believe having an own file increases clarity and consistency

Yes, but it is still an _overkill_. The global in question is closely tied to the ClientHttpRequest. The global does not deserve its own pair of source code files.

Many folks, myself included, think that even C++ classes do not always deserve their own pair of source code files. The best granularity balances well-known overheads of source code files and well-known overheads caused by separating related declarations with well-known overheads caused by grouping unrelated declarations together. Dedicating two new source code files to a global closely tied to ClientHttpRequest clearly does not achieve the right balance IMO.

rousskov: > I'll see if I can cook up some script or similar.

Your call. FWIW, I would not spend time on marginal improvements here -- any decent editor can sort a few lines alphabetically, which is sufficient for most cases. You are right that the current lists is grossly inconsistent and terrible from maintenance point of view, but I suspect that the proper solution lies in a complete revamping on how we manage these lists, and the solution space might even include abandoning automake.

src/dlink.h
kinkie: Ok, removing and converting moveToFront to free function

src/Makefile.am
kinkie: You are correct. At the same time, this would be inconvenient for stubs; they follow a good naming convention and it'd be nice to sort them close to the header files they refer to.

kinkie: I'm of two minds here, but I won't go into the philosophical question.
I think we can agree that that symbol doesn't belong with dlink. Unfortunately it's tied to both request and reply, or I would agree with you.
In a pure c++ world we may have request and reply inherit from a common class and do the bookkeeping there RAII-style, but it would still be a couple of additional files

src/stat.cc
kinkie: done

src/client_side_reply.cc
kinkie: sure

src/Makefile.am
rousskov: Yes, that so called "good" naming convention is not so good for the purpose of listing dependencies in a consistent human-friendly way. We have to pick our poison.

Again, I know that this area is currently very messy, and I do not insist on this minor mess-reducing change.

kinkie:
On Mon, May 18, 2020 at 10:39 PM Alex Rousskov <notifications@github.com>
wrote:

> *@rousskov* commented on this pull request.
> ------------------------------
>
> In src/Makefile.am
> <https://github.com/squid-cache/squid/pull/637#discussion_r426907599>:
>
> > +	ClientActiveRequests.h \
> +	ClientActiveRequests.cc \
>
> Yes, that so called "good" naming convention is not so good for the
> purpose of listing dependencies in a consistent human-friendly way. We have
> to pick our poison.
>
> Again, I know that this area is currently very messy, and I do not insist
> on this minor mess-reducing change.
>
Yes. It would be out ofd scope for this PR anyway.

-- 
    Francesco

src/Makefile.am
rousskov: > I think we can agree that that symbol doesn't belong with dlink

Yes, of course.

> Unfortunately it's tied to both request and reply

Replies need this global because, AFAICT, somebody wanted the cache manager to list recently active requests first, but that presentation order has nothing to with the proper place for _defining_ this global. I can imagine ten more places in the code where some equivalent of makeThisHead() would be called, all of them outside of client_side_reply.cc. That does not mean ClientActiveRequests is tied to all of those ten places or somehow deserves two dedicated source files just because many places want to reorder cache manager output.

One cannot properly use ClientActiveRequests global without knowing what ClientHttpRequest is, and the ClientHttpRequest object lifetime defines the lifetime of the corresponding ClientActiveRequests member. The global and the type are very strongly tied together.

> In a pure c++ world we may have request and reply inherit from a common class...

FWIW, the whole idea of splitting a single transaction into ClientHttpRequest and clientReplyContext types was probably a mistake in itself, but that is not something we can fix in this PR. Just ignore the fact that clientReplyContext reaches into ClientHttpRequest internals (as if the two were the same object) and place ClientActiveRequests where its objects are defined -- with ClientHttpRequest.

If you want to go a step further, move makeThisHead() into a global TouchClientRequest() function and hide ClientHttpRequest inside client_side_request.cc.

src/Makefile.am
kinkie: Stub done. I also extended stat.cc stub and used it in all tests

src/Makefile.am
kinkie: > If you want to go a step further, move makeThisHead() into a global TouchClientRequest() function and hide ClientHttpRequest inside client_side_request.cc.

If you don't mind, I'll do it in a followup PR

src/Makefile.am
rousskov: I hope that stub file is not actually needed, but I am resolving _this_ change request because its primary problem has been addressed. I will deal with the new stub file in a later review.

src/dlink.cc
```diff
@@ -78,3 +76,14 @@ dlinkDelete(dlink_node * m, dlink_list * list)
...
+dlinkMoveToFront(dlink_node *node, dlink_list *list)
+{
+    if (node->next)
+        node->next->prev = node->prev;
+    if (node->prev)
+        node->prev->next = node->next;
+    node->prev = nullptr;
+    node->next = list->head;
+    list->head = node;
+}
```
rousskov, line 89: This new undocumented global function is buggy -- it is missing a `list->tail` update (at least). Please remove this function and go back to using dlinkDelete()/dlinkAdd() (instead of spending even more time on enhancing the dlink_list API that should not exist).

FWIW, I have never seen anybody being able to write doubly-linked list functions without errors on the first try. I am sure such superhumans exist, but we are not them ;-). Please do not make me re-review this function. I almost missed the known bug after the two scans and can miss more bugs. Please remove this function instead.

rousskov:
After addressing the latest [review](https://github.com/squid-cache/squid/pull/637#pullrequestreview-418564862) concerns, please update PR title and description to reflect the current PR changes.

src/ClientActiveRequests.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include "dlink.h"
+
+/// Active requests table to be rendered in cache manager
+extern dlink_list ClientActiveRequests;
+
+class StoreEntry;
+void statClientRequests(StoreEntry *);
+
+#endif /* SQUID_CLIENTACTIVEREQUESTS_H */
```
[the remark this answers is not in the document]
kinkie: done to all touched files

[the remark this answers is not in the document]
kinkie: Interestingly, formater.pl doesn't seem to ensure an empty line at EOF. I'll make a note to add that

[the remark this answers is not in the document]
rousskov: > Interestingly, formater.pl doesn't seem to ensure an empty line at EOF. I'll make a note to add that

FWIW, in my environment, `formater.pl` does add an empty line to `ClientActiveRequests.h`, as expected:

```shell
$ ./scripts/formater.pl src/ClientActiveRequests.h
$ git diff | tail -n 4
```

```diff
-#endif /* SQUID_CLIENTACTIVEREQUESTS_H */
\ No newline at end of file
+#endif /* SQUID_CLIENTACTIVEREQUESTS_H */
+
```

kinkie:
@rousskov , I'm sorry but in all the long history of this PR I have lost track of what is still missing for approval.
I apologise for asking you to spend more time on this, but I'm at a loss.
The long history means also a lot of rebases which muds the waters.

src/tests/stub_stat.cc
```diff
@@ -15,4 +15,10 @@
 
 class StoreEntry;
 const char *storeEntryFlags(const StoreEntry *) STUB_RETVAL(NULL)
-
+int stat5minClientRequests(void) STUB_RETVAL(0)
+int statMemoryAccounted(void) STUB_RETVAL(0)
+class StatCounters;
```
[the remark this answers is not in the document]
kinkie: fixed

src/tests/stub_ClientActiveRequests.cc
```diff
@@ -0,0 +1,9 @@
+#include "squid.h"
+#include "ClientActiveRequests.h"
+
+#define STUB_API "ClientActiveRequests.cc"
+#include "tests/STUB.h"
+
+dlink_list ClientActiveRequests;
+class StoreEntry;
```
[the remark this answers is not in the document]
kinkie: re-fixed it. It must have not been pushed. Coming up

src/ClientActiveRequests.cc
```diff
@@ -0,0 +1,98 @@
...
+
+        storeAppendPrintf(s, "\n");
+    }
+}
+
+void
+ClientActiveRequestsInit()
+{
+    Mgr::RegisterAction("active_requests",
+                        "Client-side Active Requests",
```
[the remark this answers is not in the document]
kinkie: Changing

src/ClientActiveRequests.cc
```diff
@@ -0,0 +1,88 @@
...
+#include "fde.h"
+#include "http/Stream.h"
+#include "ip/forward.h"
+#include "Store.h"
+#include "util.h"
+
+dlink_list ClientActiveRequests;
+
+void
+statClientRequests(StoreEntry *s)
```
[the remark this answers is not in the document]
kinkie: done

kinkie:
I think I have resolved all pending issues.

src/ClientActiveRequests.cc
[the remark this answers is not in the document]
rousskov: I see no changes, and GitHub agrees.

src/ClientActiveRequests.h
```diff
@@ -0,0 +1,20 @@
...
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID_CLIENTACTIVEREQUESTS_H
+#define SQUID_CLIENTACTIVEREQUESTS_H
+
+#include "dlink.h"
+
+/// Active requests table to be rendered in cache manager
+extern dlink_list ClientActiveRequests;
```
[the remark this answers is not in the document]
rousskov: The "overexposing" `#include` that this change request is about (and that the suggested change was removing) is still there.

src/ClientActiveRequests.cc
```diff
@@ -0,0 +1,98 @@
...
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#include "squid.h"
+#include "ClientActiveRequests.h"
+#include "client_side_request.h"
+#include "fde.h"
+#include "http/Stream.h"
+#include "ip/forward.h"
```
rousskov, line 14: It is unusual for a `.cc` file to include a `forward.h` file because the latter is usually needed for forward declarations. There are exceptions, of course, but please double check that this new `.cc` file needs this specific `#include`. If some other header file cannot compile in this context because that other header file is missing `ip/forward.h`, then this `.cc` file is not the place to fix that bug.

rousskov (changes requested):
I adjusted the PR title and description to better, IMO, describe the current PR state. This PR does not contain much of an dlink_list API cleanup or true/significant ClientActiveRequests refactoring -- the PR just moves the code to where it belongs (which is a great improvement, do not get me wrong, it is just not what the PR metadata is describing). The PR does not contain dlink changes that the PR description was talking about. Please adjust further if needed.

Other than addressing the specific change requests (most of them are old), please reformat the changed (including new) sources if you can.

src/ClientActiveRequests.cc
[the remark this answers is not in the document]
rousskov: BTW, if you cannot see the [original](https://github.com/squid-cache/squid/pull/637/files#r438771448) change request in the default PR view, click on the [Files](https://github.com/squid-cache/squid/pull/637/files) view. It is clearly shown there (for me).

src/ClientActiveRequests.h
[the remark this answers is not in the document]
rousskov: BTW, if you cannot see the [original](https://github.com/squid-cache/squid/pull/637/files#r438804499) change request in the default PR view, click on the [Files](https://github.com/squid-cache/squid/pull/637/files) view. It is clearly shown there (for me).

kinkie:
Abandoning. Unentangling from the effects of PR 906 is equivalent to redoing the work

PR #638, opened 2020-05-18
WIP: fast shutdown sequence

yadij:
Add a 'shutdown tick' to the shutdown period where registered
runners can be polled for early completion of their shutdown.

Also, flag which FD are open client connections so shutdown
process finish can be triggered as soon as no clients are
waiting for transaction completion.

Turns shutdown_timeout into a proper timeout instead of a
minimum delay for shutdown.

### DIFF
--- a/src/base/RunnersRegistry.h
+++ b/src/base/RunnersRegistry.h
@@ -73,6 +73,12 @@ public:
     /// Meant for cleanup and state saving that may require other modules.
     virtual void startShutdown() {}
 
+    /// Called once per second after receiving a shutdown request and before
+    /// stopping the main loop. At least one main loop iteration is guaranteed
+    /// after this call.
+    /// Meant for cleanup and state saving that may require other modules.
+    virtual void checkShutdown() {}
+
     /// Called after shutdown_lifetime grace period ends and before stopping
     /// the main loop. At least one main loop iteration is guaranteed after
     /// this call.
--- a/src/comm.cc
+++ b/src/comm.cc
@@ -845,6 +845,7 @@ comm_close_complete(const FdeCbParams &params)
     fde *F = &fd_table[params.fd];
     F->ssl.reset();
     F->dynamicTlsContext.reset();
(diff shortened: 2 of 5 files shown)


yadij:
> I believe Squid should let modules (that need a graceful shutdown) tell Squid how much time they still need before the endingShutdown() broadcast,

That does not seem realistic in the desired final context where a shutdown which might end earlier than the known timeout period. Any value that the shutdown even function knows is subject to change on next tick. eg it could tell a runner 29sec to go then discover nothing actually needs to be running and exit after only 1 sec.
 Which brings to the second issue..

> instead of checking low-level descriptors every second in main.cc. In other words, this should be a reusable push mechanism instead of a low-level pull/poll mechanism.

... how to get the information that shutdownFinish can be called early to the event engine?

If there is a better way I am interested. I have already given a lot of thought to;
 * changing Runners API to use non-void return values for ticker runners (too much complexity irrelevant to other runner hooks)
 * some form of global container storing reference-ish to client connections (too much performance loss per-transaction).
 * having client connections register a runner to get the tick notice and set a global flag (call/notice to every client conn each tick is worse performance than this fd_table scan)

The fd_table can be quite large for scanning. But there are several optimization built into the current implementation to minimize that:
* runner(s) the register for tick notice called first to reduce the table size ASAP. eg pruning the idle and persistent connection sets, closing helpers down to a minimum idle set, etc.
* loop exits on first detected client connection. One is enough to stay waiting.
* table scanned backwards to avoid re-scanning listener, helper, and unused FD entries repeatedly.


> I would be happy to discuss specifics, either while this PR is still work-in-progress or after this PR matures out of its WIP state. Your call.

Lets get the design settled before WIP goes away if you have time. This is fairly low-priority though. I posted this PR mainly to give people who care a patch they can use and get the ball rolling on better design ideas.

rousskov:
> Lets get the design settled before WIP goes away if you have time. This is fairly low-priority though.

OK, I will detail the design at a lower priority.

PR #647, opened 2020-05-27
Maintenance: Remove FIXME and \todo labels

yadij:
Update old ad-hoc FIXME labels and experimental use of doxygen \todo
listings to use XXX or TODO labels for consistency sake.

Approximate current definitions (to be formalized elsewhere):
 * XXX - a problem, often a serious one
 * TODO - a problem, often a minor one
 * BUG - a problem that is worth informing admin about

Existing TODO, XXX, and BUG labels may not fit the above definitions.

Some old marks were edited or purged.

### DIFF
--- a/src/DescriptorSet.cc
+++ b/src/DescriptorSet.cc
@@ -37,13 +37,12 @@ DescriptorSet::~DescriptorSet()
 bool
 DescriptorSet::add(int fd)
 {
-    Must(0 <= fd);
-    Must(fd < capacity_);
+    assert(0 <= fd && fd < capacity_); // TODO: replace with Must()
 
     if (has(fd))
         return false; // already have it
 
-    Must(size_ < capacity_);
+    assert(size_ < capacity_); // TODO: replace with Must()
     const int pos = size_;
     ++size_;
     index_[fd] = pos;
@@ -55,21 +54,20 @@ DescriptorSet::add(int fd)
 bool
 DescriptorSet::del(int fd)
 {
-    Must(0 <= fd && fd < capacity_);
+    assert(0 <= fd && fd < capacity_); // TODO: here and below, use Must()
 
     if (!has(fd))
         return false; // we do not have it
 
-    Must(!empty());
+    assert(!empty());
     const int delPos = index_[fd];
-    Must(0 <= delPos);
-    Must(delPos < capacity_);
+    assert(0 <= delPos && delPos < capacity_);
 
     // move the last descriptor to the deleted fd position
     // to avoid skipping deleted descriptors in pop()
     const int lastPos = size_-1;
     const int lastFd = descriptors_[lastPos];
-    Must(delPos <= lastPos); // may be the same
+    assert(delPos <= lastPos); // may be the same
     descriptors_[delPos] = lastFd;
     index_[lastFd] = delPos;
 
--- a/src/DescriptorSet.cc
+++ b/src/DescriptorSet.cc
@@ -37,12 +37,13 @@ DescriptorSet::~DescriptorSet()
 bool
 DescriptorSet::add(int fd)
 {
-    assert(0 <= fd && fd < capacity_); // \todo: replace with Must()
+    Must(0 <= fd);
+    Must(fd < capacity_);
 
     if (has(fd))
         return false; // already have it
 
-    assert(size_ < capacity_); // \todo: replace with Must()
+    Must(size_ < capacity_);
     const int pos = size_;
     ++size_;
     index_[fd] = pos;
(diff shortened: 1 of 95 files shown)


rousskov:
Thank you for this cleanup.

FWIW, I was told that FIXME is more modern or more "universally recognizable" than XXX, but I could not quickly find any proof of that opinion. Without such a proof, I do not think we should switch to FIXME because our code uses XXX a lot more often already.

> Some old entries that seem no longer applicable have been purged.

Unfortunately, the above and other out-of-scope changes mean that I will need more time to review the details.

rousskov:
I adjusted the PR description to reflect my recollection of what the labels should mean [in new code]. The earlier description was just too far from my perception of reality. I agree that we do not have to define these markings precisely here. Unfortunately, the PR is _changing_ some of these labels, so at least some superficial agreement on those definitions is probably a good idea.

src/DescriptorSet.cc
```diff
@@ -37,12 +37,13 @@ DescriptorSet::~DescriptorSet()
 bool
 DescriptorSet::add(int fd)
 {
-    assert(0 <= fd && fd < capacity_); // \todo: replace with Must()
+    Must(0 <= fd);
+    Must(fd < capacity_);
```
rousskov, line 41: @yadij, would you mind removing this and similar `s/assert/Must/` changes from this PR, so that we do not have to argue about them and their details? Combining comment changes with any "actual code" changes makes the scope of this PR a lot more complex and complicates the PR review quite a bit!

yadij: Done in 117408db0d48d25c18ad97034e6b666cd7cf33f9

yadij: NP: the other notable logic change is mem/Pool.h and mem/Pool.cc everything else is documentation.

yadij:
> FWIW, I was told that FIXME is more modern or more "universally recognizable" than XXX, but I could not quickly find any proof of that opinion. Without such a proof, I do not think we should switch to FIXME because our code uses XXX a lot more often already.

The only place I have encountered XXX is here in Squid. My experience with other projects is that they either use FIXME or TODO depending on the maintainers preference. Also just one used. Not a mixture.

If it were up to me I would use just TODO for all of them and not split by severity since that is largely subjective and will change over time. eg items in gopher were once high severity, today not so much.

I do not want to block this cleanup on more policy discussion though. The scope is primarily at removing the \todo which we already agreed to drop. FIXME was included as scope creep to improve consistency since there were so very few uses.

rousskov:
> The only place I have encountered XXX is here in Squid. My experience with other projects is that they either use FIXME or TODO[...]. Not a mixture.

FYI: Here are approximate counters from a few projects:

|Project|TODO|FIXME|XXX|
|:--|--:|--:|--:|
|glibc|46|64|352|
|Squid|671|29|512|
|FreeBSD|2662|2029|12251|
|Linux|3633|3329|2608|
|rust|163|2706|22|

> If it were up to me I would use just TODO for all of them and not split by severity since that is largely subjective and will change over time. eg items in gopher were once high severity, today not so much.

I find it useful to know what the author thought of the problem when she recorded it, but I agree that would could settle on one to avoid arguing about the right category in PR reviews. I probably would use FIXME as that single choice because it does not imply that the following is a list of things to do and because of the poorly substantiated "all the cool projects are using it these days" argument. However, I am not sure changing 1000 lines is worth the benefits in this case.

> I do not want to block this cleanup on more policy discussion though. The scope is primarily at removing the \todo which we already agreed to drop. FIXME was included as scope creep to improve consistency since there were so very few uses.

I agree on all counts. We are just abusing GitHub for this side discussion.

src/esi/Include.cc
```diff
@@ -335,9 +335,8 @@ ESIInclude::ESIInclude(esiTreeParentPtr aParent, int attrcount, char const **att
              */
             debugs(86, 5, "ESIIncludeNew: Requesting alternate '" << attr[i+1] << "'");
 
-            assert (alt.getRaw() == NULL); /* TODO: FIXME */
+            assert(!alt.getRaw());
```
rousskov, line 338: If we start out of scope manual polishing, why stop midway?

```suggestion
            assert(!alt);
```

yadij: Reverted in 6871c39

src/ip/Address.cc
```diff
@@ -936,9 +936,9 @@ Ip::Address::getSockAddr(struct sockaddr_storage &addr, const int family) const
     struct sockaddr_in *sin = NULL;
 
     if ( family == AF_INET && !isIPv4()) {
-        // FIXME INET6: caller using the wrong socket type!
+        // XXX: caller using the wrong socket type!
         debugs(14, DBG_CRITICAL, HERE << "Ip::Address::getSockAddr : Cannot convert non-IPv4 to IPv4. from " << *this);
-        assert(false);
+        assert(family == AF_INET && !isIPv4());
```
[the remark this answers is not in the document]
yadij: Reverted in 6871c39

src/esi/VarState.cc
```diff
@@ -805,32 +805,31 @@ ESIVariableProcessor::~ESIVariableProcessor()
...
-/* XXX FIXME: this should be comma delimited, no? */
 void
 ESIVarState::buildVary (HttpReply *rep)
 {
     char tempstr[1024];
     tempstr[0]='\0';
 
     if (flags.language)
-        strcat (tempstr, "Accept-Language ");
+        strcat(tempstr, "Accept-Language, ");
```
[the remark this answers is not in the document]
yadij: Reverted in 6871c39

src/mem/Pool.h
```diff
@@ -143,21 +139,18 @@ class MemPools
...
      *
-     \par
-     * Should be called relatively often, as it sorts chunks in suitable
-     * order as to reduce free memory fragmentation and increase chunk
-     * utilisation.
-     * Suitable frequency for cleanup is in range of few tens of seconds to
-     * few minutes, depending of memory activity.
-     *
-     \todo DOCS: Re-write this shorter!
-     *
```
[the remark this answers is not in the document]
yadij: Reverted in 6871c39

src/mgr/BasicActions.h
```diff
@@ -93,7 +93,7 @@ class OfflineToggleAction: public Action
     OfflineToggleAction(const CommandPointer &cmd);
 };
 
-/// Registers profiles for the actions above; \todo move elsewhere?
+/// Registers profiles for the actions above.
```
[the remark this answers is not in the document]
yadij: Done in f1532ff

PR #648, opened 2020-05-29
V4 next backports

squidadm:

PR #650, opened 2020-06-01
V4 next backports

squidadm:


yadij:
OK to test

rousskov:
To avoid misunderstanding: I am blocking this PR to get your attention because [it](https://github.com/squid-cache/squid/pull/648#pullrequestreview-421197410) is a serious issue, and you have not responded to my previous comment about it. I will unblock once you acknowledge that you are aware of the problem, even if you disagree that the problem deserves the fix in this PR.

yadij:
Thank you. Reverting in this update. Though note the commit messages from the v5 PR explicitly state "These loops do not have ABA update problems".

rousskov:
> Though note the commit messages from the v5 PR explicitly state "These loops do not have ABA update problems".

Yes, I know. This bug was disclosed on squid-bugs in my email dated 2020-03-07.

PR #651, opened 2020-06-01
Release notes prep for 5.0.3 and 4.12

yadij:


ChangeLog
```diff
@@ -1,3 +1,9 @@
+Changes in squid-5.0.3 (04 Jun 2020):
+
+	- Bug 5046: FreeBSD lacks open(2) O_DSYNC flag
+	- ... and several debug improvements
+	- ... and all fixes from 4.12
```
rousskov, line 5: IMHO, we should list changes without referring to other Squid releases because the current approach is 

* inconvenient for the reader (who has to follow one or more links to get the complete picture for each release)

* often arguably incorrect or misleading (e.g., one of the `fixes from 4.12` referenced on this line actually belongs to the already released `Changes in squid-5.0.2` set rather than the currently documented `Changes in squid-5.0.3` set; look for "Bug 5030").

yadij:
> Is this a draft PR?

Kind of. With the v4 backports incomplete when it was written some of the commits were not finalized about which version they would be listed for.


> If not, I would expect entries for some (or all) of these recent v5 commits:
> 
>     * [8eece63](https://github.com/squid-cache/squid/commit/8eece637e89fcb3276ec5ba5b481c7a63fca54c2): Detail TLS and CONNECT cache_peer negotiation failures
> 

This is under "several debug improvements" since it seems the only user visible change is improved error page descriptions.

>     * [66f8515](https://github.com/squid-cache/squid/commit/66f85158b1947105716a4a27868b974d2635441b): Fix keyblock use for Heimdal in kerberos_ldap_group helper

This is in 4.12 under "some compile fixes".

I try not to detail individual changes to documentation, compile errors, or debug outputs unless there is a bug number to mention closing.

> BTW, please note that v5.0.2 (i.e. [3953412](https://github.com/squid-cache/squid/commit/3953412ebd17d99eb6bab35695880d4b09c4ac2d)) is missing a git tag.

Thanks. I will be updating the tags for v5 and v4 branches after this release is in.

ChangeLog
yadij: This is an indicator for when changes in the stable release sync up with the beta release. The previous few have been done in sync for security fixes, but normally the different release requirements result in the series not matching up for periods of time. That it helps to cut down changelog size and multiple entries for commits is a useful addition.

You can see examples of this in the v4.0.12 to 4.0.14 period (a normal 4 vs 8 week difference), and the 4.0.21 to 4.2 period (the EOL security bump on deprecated v3.5).

Complete lists of every little commit can be found in our changesets page and the repository log.

rousskov: > This is an indicator for when changes in the stable release sync up with the beta release.

It might be a useful indicator for the select few initiated, but it does not offset the negatives enough IMHO.

> Complete lists of every little commit can be found

I know, but this problem is not the "lack of complete lists of every little commit".

yadij:
@eduard, why is anubis insisting that this failed after 1b14ee7 ?

rousskov:
> @eduard, why is anubis insisting that this failed after 1b14ee7?

AFAICT, Anubis tells us that the last staged commit (c7ababf) failed the tests. That conclusion is correct. The red flag here is not the `M-failed-staging-checks` label, but the lack of a new staged commit after the PR branch was modified (by PR commit 1b14ee7). I speculate that Anubis just did not have a chance to stage this PR because there are other, higher priority PRs that should be staged/merged (e.g. PR 635).

If my speculation is accurate, then we probably want to change Anubis so that it does not _restore_ the human-removed `M-failed-staging-checks` label until it creates a new staged request. @eduard-bagdasaryan, do you agree?

eduard-bagdasaryan:
> The red flag here is not the M-failed-staging-checks label, but the lack of a new staged commit after the PR branch was modified (by PR commit 1b14ee7).

Anubis did not create a new stage commit for this PR because it assumed that the staged c7ababf is still 'fresh'. This freshness is calculated, comparing SHAs of 'tree' objects of the staged commit and the current 'merge' commit. It turned out that the last PR branch change (1b14ee7) was a no-change, so the 'merge' commit tree SHA did not change.
So it is not possible to restart staging by means of such PR branch touches. However, this will happen after another PR takes control (i.e., becomes staged and merged or failed).

rousskov:
> This freshness is calculated, comparing SHAs of 'tree' objects of the staged commit and the current 'merge' commit.

Very interesting. Thank you for correcting my theory! I did not notice that the last branch commit was empty.

Using the code tree SHA is a must, but is there something that Anubis can use as an _additional_ sign of human-perceived PR changes? The bot did the right thing here as far as the bytes in the repository are concerned, but Anubis should keep the "appearances" in mind because that is what most humans react to. Without introducing Anubis state, can we add any existing GitHub signal to the staged commit freshness decision to handle the case of an empty commit better?

I understand that we can add another label as such a signal, but we should avoid adding labels where a label-free solution is available.

rousskov:
> to handle the case of an empty commit better

Here, "better" means "realize that the staged commit is stale". I assume that such a "stale staged commit" decision would already result in the right human-observable effect -- the `M-failed-staging-checks` will not be restored.

eduard-bagdasaryan:
> can we add any existing GitHub signal to the staged commit freshness decision to handle the case of an empty commit better

For example, we can compare 'date' properties of the 'merge' and 'staged' commits: the newer 'merge' commit will indicate that the 'staged' commit is stale.

rousskov:
> we can compare 'date' properties of the 'merge' and 'staged' commits

Sounds good to me. I created an internal issue for this to-do.

hno:
It could in theory use the git commit hash. This is a hash of the tree, and
it's history.

Den fre 5 juni 2020 19:11Alex Rousskov <notifications@github.com> skrev:

rousskov:
> It could in theory use the git commit hash

It takes two to tango:

* The SHA of the merge commit (created by GitHub and not visible to ordinary humans browsing GitHub pages) itself is useless -- it is unique.
* Same uniqueness problem for the SHA of the staged commit (created by [Anubis](https://github.com/measurement-factory/anubis/blob/master/README.md#pull-request-merging-algorithm) and visible on GitHub pages).
* Both merge and staged commits contain the PR source _tree_ SHA. Anubis _is_ already using those SHAs to detect stale staged commits, but our working theory is that the tree SHA does not _change_ when an empty commit is added to the PR branch -- our use case here.

We are looking for another way (i.e. an additional condition) to detect a mismatch between PR state and the staged commit. The dates of merge and staged commits, as [suggested](https://github.com/squid-cache/squid/pull/651#issuecomment-639614535) by Eduard, may work OK in most cases. Ideally, it would be something more reliable (like another SHA or some kind of commit counter), but we do not want to add housekeeping noise to the staged commit itself (because it will become the official commit if everything goes OK).

hno:
> It takes two to tango:
>
>    - The SHA of the merge commit (created by GitHub and not visible to
>    ordinary humans browsing GitHub pages) itself is useless -- it is unique.
>
>
The commit hash is a unique identifier and is unique for that commit in the
source history. More unique than hash of the sources and timestamp, but in
practice only marginally unless I miss something fundamental about what
Anubis is doing.

>
>    - Same uniqueness problem for the SHA of the staged commit (created by
>    Anubis
>    <https://github.com/measurement-factory/anubis/blob/master/README.md#pull-request-merging-algorithm>
>    and visible on GitHub pages).
>    - Both merge and staged commits contain the PR source *tree* SHA.
>    Anubis *is* already using those SHAs to detect stale staged commits,
>    but our working theory is that the tree SHA does not *change* when an
>    empty commit is added to the PR branch -- our use case here.
>
>
An empty commit is only adding history. Code is the same but history has
additional information.

An empty commit creates a new commit hash, and stores the associated commit
message.

A fast forward merge creates nothing, just updates the branch pointer.

What I am perhaps not understanding is what is the use of the source tree
hash to identify identical trees in different commits if we anyway need to
identify the commit state?

>
>
> We are looking for another way (i.e. an additional condition) to detect a
> mismatch between PR state and the staged commit. The dates of merge and
> staged commits, as suggested
> <https://github.com/squid-cache/squid/pull/651#issuecomment-639614535> by
> Eduard, may work OK in most cases. Ideally, it would be something more
> reliable (like another SHA or some kind of commit counter), but we do not
> want to add housekeeping noise to the staged commit itself (because it will
> become the official commit if everything goes OK.
>

A GitHub pull request is a kind of git branch. For all practical purposes
you can view the referenced git history as a branch. Any changes in the
branch creates a new commit hash. Added discussions changes nothing (source
wise). Normally one adds commits to the head of the branch to fix issues
detected but if one rebases the branch then all commits from the changed
point onwards on the rebases branch gets new commit hashes.

Regards
Henrik

rousskov:
> What I am perhaps not understanding is what is the use of the source tree hash to identify identical trees in different commits if we anyway need to identify the commit state?

Sorry, I do not understand the question, possibly because it is based on a false assumption. I will start with the assumption and then try to describe what is going on from two or three different points of view (with some overlaps).

Regarding that "identical trees in different commits" assumption: The source trees may no longer be identical -- that is one of the freshness criteria that Anubis is checking. And using the tree SHA for comparison is probably influenced by GitHub APIs -- we need some way to compare two source trees, and (IIRC) GitHub provides the tree SHA that we can conveniently use for that purpose.

As you know, a git commit consists of many parts, including the commit message, author, date, and the git tree object. When Anubis creates a new commit (called "staging commit" in the current Anubis docs, but I usually call it "staged commit"), Anubis uses GitHub [commit creation](https://developer.github.com/v3/git/commits/#create-a-commit) API. The git tree object SHA is a required parameter in that API.

Anubis needs to decide whether the staged commit is "fresh" (i.e. still matches the PR state). To do that, Anubis iterates over commit parts (that Anubis uses to create staged commits) and checks that each PR-derived part is the same as the current staged commit part. These parts include a commit message and a commit source tree. The messages are probably compared as strings. The trees are compared using their SHAs. IIRC, those SHAs are already computed and provided by GitHub. When PR sources change, Anubis notices that the parts no longer match and abandons the stale staged commit (eventually creating a new one).

> Any changes in the branch creates a new commit hash

Yes, but Anubis does not work directly with the PR branch (yet). Anubis works with the GitHub-created merge commit (naturally derived from that PR branch). As you probably know, GitHub [merge commit](https://developer.github.com/v3/pulls/#get-a-single-pull-request) is _not_ a part of any branch. If @eduard-bagdasaryan analysis is correct, then some PR branch changes do not affect the source tree hash of the GitHub merge commit, and, hence, are invisible to the current Anubis algorithm that tries to detect PR changes: The PR branch has changed, but the source tree hash of the GitHub merged commit did not.

AFAICT, Eduard's theory (or perhaps even a verified fact?) is that, after a no-source-changes commit to the PR branch, GitHub merge commit gets a fresh date field (and the old source tree hash). Anubis can notice that the GitHub merge commit has a newer date than the staged commit and conclude that the staged commit is probably stale.

PR #652, opened 2020-06-03
Bug 5051: Some collapsed revalidation responses never expire

DrDaveD:
Since negative caching support was repaired in master commit 91870bf, it
has been found to last indefinitely when cache revalidation happens.

New revalidation requests were collapsing on a negatively cached
response forever because handleIMS() logic does not validate response
freshness (still assuming that the reply came in response to the current
request even though that assumption could be false since collapsed
revalidation support was added in master commit 1a210de).

Clearing the ENTRY_REQUIRES_COLLAPSING flag when hitting the negatively
cached collapsed revalidaiton response for the first time works around
this "lack of freshness check" problem. The same solution existed in the
official code for positive responses. However, this solution is partial
and unreliable because there is no guarantee that the clearing code will
be reached (and reached while the cached response is still fresh).

Also added additional partial protections against collapsing on entries
abandoned by their writers, including idle hittingRequiresCollapsing()
StoreEntry objects.

Also fixed a tiny race condition missed in master commit d1d3b4d which
addressed a much bigger (and more frequent) problem. I am not aware of
any real-world cases where this race condition surfaced, but they would
probably manifest in unwarranted failures to collapse.

### DIFF
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -377,6 +377,10 @@ clientReplyContext::sendClientUpstreamResponse()
 {
     StoreIOBuffer tempresult;
     removeStoreReference(&old_sc, &old_entry);
+
+    if (collapsedRevalidation)
+        http->storeEntry()->clearPublicKeyScope();
+
     /* here the data to send is the data we just received */
     tempBuffer.offset = 0;
     old_reqsize = 0;
@@ -490,10 +494,6 @@ clientReplyContext::handleIMSReply(StoreIOBuffer result)
             http->logType.update(LOG_TCP_REFRESH_MODIFIED);
             debugs(88, 3, "origin replied " << status <<
                    ", replacing existing entry and forwarding to client");
-
-            if (collapsedRevalidation)
-                http->storeEntry()->clearPublicKeyScope();
-
             sendClientUpstreamResponse();
         }
     }
@@ -503,10 +503,6 @@ clientReplyContext::handleIMSReply(StoreIOBuffer result)
         http->logType.update(LOG_TCP_REFRESH_FAIL_ERR);
         debugs(88, 3, "origin replied with error " << status <<
                ", forwarding to client due to fail_on_validation_err");
-
-        if (collapsedRevalidation)
-            http->storeEntry()->clearPublicKeyScope();
-
         sendClientUpstreamResponse();
     } else {
         // ignore and let client have old entry
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -457,6 +457,9 @@ clientReplyContext::handleIMSReply(StoreIOBuffer result)
     const auto &new_rep = http->storeEntry()->mem().freshestReply();
     const auto status = new_rep.sline.status();
 
+    // XXX: Disregard stale incomplete (i.e. still being written) borrowed (i.e.
+    // not caused by our request) IMS responses. That new_rep may be very old!
+
     // origin replied 304
     if (status == Http::scNotModified) {
         http->logType.update(LOG_TCP_REFRESH_UNMODIFIED);
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -503,6 +503,10 @@ clientReplyContext::handleIMSReply(StoreIOBuffer result)
         http->logType.update(LOG_TCP_REFRESH_FAIL_ERR);
         debugs(88, 3, "origin replied with error " << status <<
                ", forwarding to client due to fail_on_validation_err");
+
+        if (collapsedRevalidation)
+            http->storeEntry()->clearPublicKeyScope();
+
         sendClientUpstreamResponse();
     } else {
         // ignore and let client have old entry
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -329,6 +329,26 @@ Store::Controller::hasReadableDiskEntry(const StoreEntry &e) const
     return swapDir->hasReadableEntry(e);
 }
 
+/// flags problematic ephemeral peek() results before we share them
+void
+Store::Controller::checkPeeked(const StoreEntry &entry) const
+{
+    // The "hittingRequiresCollapsing() has an active writer" checks below
+    // protect callers from getting stuck and/or from using a stale revalidation
+    // reply. However, these protections are not reliable because the writer may
+    // disappear at any time and/or without a trace. Collapsing adds risks...
+    if (entry.hittingRequiresCollapsing()) {
+        if (entry.hasTransients()) {
+            // Too late to check here because the writer may be gone by now, but
+            // Transients do check when they setCollapsingRequirement().
+        } else {
+            // a local writer must hold a lock on its writable entry
+            if (!(entry.locked() && entry.isAccepting()))
+                throw TextException("no local writer", Here());
+        }
+    }
+}
+
 StoreEntry *
 Store::Controller::find(const cache_key *key)
 {
@@ -337,6 +357,7 @@ Store::Controller::find(const cache_key *key)
             if (!entry->key)
                 allowSharing(*entry, key);
             checkTransients(*entry);
+            checkPeeked(*entry);
             entry->touch();
             referenceBusy(*entry);
             return entry;
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -380,6 +380,8 @@ Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
     addReading(&entry, key);
 
     if (entry.hasTransients()) {
+        // store hadWriter before computing `found`; \see Transients::get()
+        const auto hadWriter = transients->hasWriter(entry);
         bool inSync = false;
         const bool found = anchorToCache(entry, inSync);
         if (found && !inSync)
@@ -391,7 +393,7 @@ Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
                 throw TextException("transients entry missing ENTRY_REQUIRES_COLLAPSING", Here());
             }
 
-            if (!transients->hasWriter(entry)) {
+            if (!hadWriter) {
                 // prevent others from falling into the same trap
                 throw TextException("unattached transients entry missing writer", Here());
             }
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -296,6 +296,14 @@ Store::Controller::dereferenceIdle(StoreEntry &e, bool wantsLocalMemory)
             keepInStoreTable = wantsLocalMemory || keepInStoreTable;
     }
 
+    if (e.hittingRequiresCollapsing()) {
+        // If we were writing this now-locally-idle entry, then we did not
+        // finish and should now destroy an incomplete entry. Otherwise, do not
+        // leave this idle StoreEntry behind because handleIMSReply() lacks
+        // freshness checks when hitting a collapsed revalidation entry.
+        keepInStoreTable = false; // may overrule fs decisions made above
+    }
+
     return keepInStoreTable;
 }
 
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -330,10 +330,12 @@ Store::Controller::hasReadableDiskEntry(const StoreEntry &e) const
     return swapDir->hasReadableEntry(e);
 }
 
-/// flags problematic ephemeral peek() results before we share them
+/// flags problematic entries before find() commits to finalizing/returning them
 void
-Store::Controller::checkPeeked(const StoreEntry &entry) const
+Store::Controller::checkFoundCandidate(const StoreEntry &entry) const
 {
+    checkTransients(entry);
+
     // The "hittingRequiresCollapsing() has an active writer" checks below
     // protect callers from getting stuck and/or from using a stale revalidation
     // reply. However, these protections are not reliable because the writer may
@@ -357,8 +359,7 @@ Store::Controller::find(const cache_key *key)
         try {
             if (!entry->key)
                 allowSharing(*entry, key);
-            checkTransients(*entry);
-            checkPeeked(*entry);
+            checkFoundCandidate(*entry);
             entry->touch();
             referenceBusy(*entry);
             return entry;
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -278,7 +278,8 @@ Store::Controller::dereferenceIdle(StoreEntry &e, bool wantsLocalMemory)
 
     bool keepInStoreTable = false; // keep only if somebody needs it there
 
-    /* Notify the fs that we're not referencing this object any more */
+    // Notify the fs that we are not referencing this object any more. This
+    // should be done even if we overwrite keepInStoreTable afterwards.
 
     if (e.hasDisk())
         keepInStoreTable = swapDir->dereference(e) || keepInStoreTable;
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -162,13 +162,26 @@ Transients::get(const cache_key *key)
         return nullptr;
     }
 
+    // store hadWriter before checking ENTRY_REQUIRES_COLLAPSING to avoid racing
+    // the writer that clears that flag and then leaves
+    const auto hadWriter = map->peekAtWriter(index);
+    if (!hadWriter && EBIT_TEST(anchor->basics.flags, ENTRY_REQUIRES_COLLAPSING)) {
+        debugs(20, 3, "not joining abandoned entry " << index);
+        map->closeForReadingAndFreeIdle(index);
+        return nullptr;
+    }
+
     StoreEntry *e = new StoreEntry();
     e->createMemObject();
     e->mem_obj->xitTable.index = index;
     e->mem_obj->xitTable.io = Store::ioReading;
     anchor->exportInto(*e);
-    const bool collapsingRequired = EBIT_TEST(anchor->basics.flags, ENTRY_REQUIRES_COLLAPSING);
-    e->setCollapsingRequirement(collapsingRequired);
+
+    if (EBIT_TEST(anchor->basics.flags, ENTRY_REQUIRES_COLLAPSING)) {
+        assert(hadWriter);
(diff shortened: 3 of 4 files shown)


yadij:
OK to test

src/client_side_reply.cc
```diff
@@ -503,6 +503,10 @@ clientReplyContext::handleIMSReply(StoreIOBuffer result)
         http->logType.update(LOG_TCP_REFRESH_FAIL_ERR);
         debugs(88, 3, "origin replied with error " << status <<
                ", forwarding to client due to fail_on_validation_err");
+
+        if (collapsedRevalidation)
+            http->storeEntry()->clearPublicKeyScope();
+
         sendClientUpstreamResponse();
```
rousskov: > collapsed revalidation was not finalized when there was a negative response

... and is still not finalized in several other cases!

0. My minor PR commit 55cb158 covers all future sendClientUpstreamResponse() cases, leaving two existing areas exposed:

1. We probably need to handle sendClientOldEntry() cases. We probably do not want future collapsed revalidation clients to accidentally collapse on the entry that is being rejected by the code calling sendClientOldEntry() now. That future client should start its own revalidation attempt instead of collapsing on the stale entry created by an old attempt. To accomplish that, we need to mark the entry so that the already collapsed clients can still use it in their handleIMSReply() code while new clients cannot collapse on it. I believe releaseRequest(true) would do that. What do you think?

2. Consider cases where a collapsed revalidation did not reach handleIMSReply() at all. For example, imagine a client that created a collapsed revalidation entry and then disconnected or died. How to protect other clients from hitting that essentially stale entry forever, keeping bug 5051 alive? Unfortunately, I cannot think of a better solution than to add a collapsing timeout -- startCollapsingOn() would return false for collapsing-required entries that were around for longer than X seconds. Can you think of a better solution? If not, what X value would you recommend? We can hard-code X, for now, but if there is an existing configuraton directive that we can reuse for X, we should use that directive.

FWIW, bug 5051 should probably be described as "Squid does not check expiration of collapsed revalidation entries" rather than "Squid does not check expiration of negatively-cached collapsed-revalidation entries" -- negatively-cached collapsed-revalidation entries is just one (albeit common) manifestation of this bug. 

Our commit 1a210de essentially introduced a new class of publicly discoverable entries, but there is no code to check whether those entries have expired. For regular (i.e. not collapsable-on internal-revalidation) entries, such checks probably live in clientReplyContext::cacheHit(). Not all existing cacheHit() checks apply to (collapsed) revalidation, but some probably do.

DrDaveD: I really don't understand much about how collapsed revalidation is implemented.  I don't know what clearing a public key scope has to do with it.  I do see that clearPublicKeyScope() calls forcePublicKey() which calls clearPrivate() which sets shareableWhenPrivate=false, and that releaseRequest() calls setPrivateKey() which can set the same value.  releaseRequest() is called from abort(), so does that already handle the missing cases you are concerned about?

I don't understand why we wouldn't be seeing much more problems if all those cases you list are broken.  Even with the commit you already added to this PR, there's an extra call to  clearPublicKeyScope() when there is a 304 Not Modified response.  We get those all the time, so I can't imagine that collapsed revalidation is not getting properly cleared in that case.  And sendOldClient() calls restoreState(); does that end up also clearing collapsed revalidation from the storage?

rousskov: > releaseRequest() is called from abort(), so does that already handle the missing cases you are concerned about?

No, that only covers cases where the entry is abort()ed. I am concerned about completed entries.

> Even with the commit you already added to this PR, there's an extra call to clearPublicKeyScope() when there is a 304 Not Modified response. We get those all the time, so I can't imagine that collapsed revalidation is not getting properly cleared in that case.

Yes, that "extra" call did not fix any bugs in the current code. My polishing change avoids similar bugs in _future_ sendClientUpstreamResponse() callers. It is quite common for two calls that must go together to get separated during code refactoring, or for one of the calls to be forgotten. The latter is exactly what happened when we added the first clearPublicKeyScope() call. That is why it is a good idea to keep these two calls together.

> sendClientOldEntry() calls restoreState(); does that end up also clearing collapsed revalidation from the storage?

I was hoping that it does, and I did check it before posting my review.  AFAICT, it does not: restoreState() restores the current transaction state -- to use the old entry -- but simply unlocks the new entry. Due to old code design problems, it is difficult to be sure here, especially without looking at the logs, but I think the code will eventually end up in Store::Controller::handleIdleEntry() which does not seem to treat collapsed revalidation entries specially. If they are cachable, they may be left alone in the cache. I wonder whether we should add another clearPublicKeyScope() to that method because an idle entry cannot require collapsing (in a non-SMP environment)!


> I don't understand why we wouldn't be seeing much more problems if all those cases you list are broken.

That is a good question. There are actually not so many cases that remain allegedly buggy:

1. sendClientOldEntry() cases: There are probably only two relevant ones (the others do not deal with a cachable response). You can probably reproduce the first one -- "origin replied ... but with an older date header" --  using a custom script, but such cases are very rare in the wild. IIRC, that code was added primarily to make compliance tests happy. The second case would require a false failOnValidationError setting so it does not apply to your environment. 

2. A lack of handleIMSReply() call. To trigger this, one may have to kill the client-serving "job" after the negative response was cached but before that response was delivered to handleIMSReply(). I suspect it would be very difficult to time the client connection closure just right to accomplish this, so I am not very surprised that you do not see such cases. The "job" cleanup code may end up in Store::Controller::handleIdleEntry(). See my earlier paragraph about that method.

I wish we could clearPublicKeyScope() much earlier, when the response was received. It would solve all these problems before they start! Unfortunately, simply doing that would destroy the old being-revalidated entry -- there can be only one entry with the same public key (and default key scope).

I wonder whether we should move (part of) the logic that invalidates the old entry from handleIMSReply() to the code that caches the new entry. If we can make the right/informed decision about old entry preservation when the new entry is about to be cached, then it would be OK to invalidate the old entry early, of course. Unfortunately, I cannot volunteer to do the legwork for this non-trivial change right now.

If you would rather not do it either, then I can add the simpler changes (discussed in my review and this comment) and have you test them before this PR is merged. I will also add a TODO comment for the long-term fix described in the above paragraph.

How do you want to proceed?

DrDaveD: Let's go with the simpler changes and the TODO, and I will test them.

rousskov: Agreed. The ball is on my side.

rousskov: FYI: I started working on this earlier today. I hope to finish within 24 hours.

rousskov: Status update: I thought I was done adding the promised changes, but I am getting surprising test results now. I need more time to convince myself that these problems are _not_ related to this specific bug (and, hence, should be addressed separately). I hope to finish by Monday.

rousskov: The near-minimal changes discussed on this thread are now completed (as of branch commit cd35900).

rousskov: > I need more time to convince myself that these problems are not related to this specific bug (and, hence, should be addressed separately).

I plan to submit a small PR dedicated to those other problems. They are not related to stale collapsed revalidation responses that this PR is working on.

rousskov:
@DrDaveD, please pull in my changes and retest the current PR branch version (up to branch commit cd35900), as [planned](https://github.com/squid-cache/squid/pull/652#discussion_r436028291).

@eduard-bagdasaryan, please review these branch changes, especially my commit cd35900 (which is based on branch commit 8ffc91d logic and references your commit d1d3b4d in #516 ). Did I miss anything important, both in the code that partially addresses bug 5051 and in comments discussing the difficulties with developing a (currently missing) complete/reliable solution for stale collapsed revalidation responses.

Thank you both!

src/store/Controller.cc
```diff
@@ -296,6 +296,14 @@ Store::Controller::dereferenceIdle(StoreEntry &e, bool wantsLocalMemory)
             keepInStoreTable = wantsLocalMemory || keepInStoreTable;
     }
 
+    if (e.hittingRequiresCollapsing()) {
+        // If we were writing this now-locally-idle entry, then we did not
+        // finish and should now destroy an incomplete entry. Otherwise, do not
+        // leave this idle StoreEntry behind because handleIMSReply() lacks
+        // freshness checks when hitting a collapsed revalidation entry.
```
yadij, line 303: Should this be a XXX or TODO to add those checks?
I do not insist on this, but it sounds to me like those checks would be useful in general.

yadij:
@eduard, is there something broke in Anubis that it is still doing staging checks on this PR after we revoked the earlier M-cleared-for-merge flag?   Is anything going to break (eg wrong commit merged) if we re-add that flag when the staging success is already set?

eduard-bagdasaryan:
> after we revoked the earlier M-cleared-for-merge flag

I do not see in the history that M-cleared-for-merge was ever set/unset.

> if we re-add that flag when the staging success is already set

If we add this label now, the already built and ready (staged) a3f3405 should be merged.

yadij:
Okay, thank you. Just waiting on DrDave's tests and your review then.

src/store/Controller.cc
```diff
@@ -296,6 +296,14 @@ Store::Controller::dereferenceIdle(StoreEntry &e, bool wantsLocalMemory)
...
+        keepInStoreTable = false; // may overrule fs decisions made above
+    }
```
eduard-bagdasaryan, line 305: I would move this block above, just below 'ENTRY_SPECIAL' check.

src/store/Controller.cc
```diff
@@ -321,6 +329,26 @@ Store::Controller::hasReadableDiskEntry(const StoreEntry &e) const
...
+/// flags problematic ephemeral peek() results before we share them
+void
+Store::Controller::checkPeeked(const StoreEntry &entry) const
+{
+    // The "hittingRequiresCollapsing() has an active writer" checks below
+    // protect callers from getting stuck and/or from using a stale revalidation
+    // reply. However, these protections are not reliable because the writer may
+    // disappear at any time and/or without a trace. Collapsing adds risks...
+    if (entry.hittingRequiresCollapsing()) {
+        if (entry.hasTransients()) {
```
eduard-bagdasaryan, line 341: I think that we should have already checked these has_writer-related conditions in Transients::get() - so there is no need to repeat those checks here. If so, then, basically, checkPeeked() covers only the 'local writer' case. Then I would move the 'else' block below into a separate method and call this new method from Store::Controller::peekAtLocal() - somewhere below checkTransients() call.

src/store/Controller.cc
rousskov: We cannot move this block up -- the "fs" code needs to be notified about this entry _and_ may decide to keep it in store_table. I hinted at that order by referencing "fs decisions made above" in the assignment comment. There is also a `Notify the fs` comment earlier in the method.Perhaps those hints are too subtle. Does the comment adjustment in 8ec3f30 address your concern?

src/store/Controller.cc
rousskov: > we should have already checked these has_writer-related conditions in Transients::get()

Yes and no: Yes, we should have already checked these conditions in Transients::get(), but that may have been done long time ago (and for a different HTTP request). A lot could change since then -- and that is what the "Too late" comment is talking about. Also:

* Transients::get() is not called for cache entries that already have local StoreEntry objects in store_table. 
* Failed Transients::get() is not relevant for MemStore::get() and Rock::SwapDir::get() cases -- those are handled by allowSharing() and transients->monitorIo(). There _are_ problems with the latter path, but I decided to handle them outside this PR because they are not specific to stale collapsed revalidation entries.

Finally, I speculate that we will add more checks to this new method eventually.

Thus, I think it is best to keep this method general and exclude/document impossible-to-check cases explicitly, like this PR does.

src/store/Controller.cc
eduard-bagdasaryan: > We cannot move this block up

I think you are right. I hoped that we could somehow refactor the code for the path when e.hittingRequiresCollapsing() is true - in this case, keepInStoreTable flag should not be recalculated. But it is difficult to do that since dereference() methods are responsible for two tasks at once - 'dereferencing' and returning a boolean constant (false for Rock and MemStore and true for Ufs). This refactoring would require splitting dereference() into two methods (e.g., dereference() and keepInStoreTable()). I do not think it is worth doing here.

rousskov: We are on the same page then, thank you.

src/store/Controller.cc
eduard-bagdasaryan: >  but that may have been done long time ago 

I am not against additional checks (by means of Controller::checkPeeked()), but I wonder why should we call it only after 'long time has passed', trying to recalculate the context then (i.e., to understand what Store produced the StoreEntry) and finally realize that we can not complete this checking in some cases (as the comment notes). 

> Transients::get() is not called for cache entries that already have local StoreEntry objects in store_table.

Agreed, but in this case, we would do that collapsing check in Controller::peekAtLocal().

> those are handled by allowSharing() and transients->monitorIo().

This also looks a little strange - checkPeeked() is called after allowSharing(), which may have an effect on conditions, calculated in checkPeeked(). In other words, checkPeeked() may be probably renamed because it checks something produced by several methods (including peek()).

rousskov: > I wonder why should we call it only after 'long time has passed', trying to recalculate the context then (i.e., to understand what Store produced the StoreEntry) and finally realize that we can not complete this checking in some cases (as the comment notes).

"Long-time-after" is not exactly what is happening in this new checking code though. The new code is executed for every viable find() candidate. Some are freshly created. Some were around for a long time. If we can write a generic check that applies to both categories, we should. In this particular (SMP) case, we cannot -- we can only check freshly created entries (and we do check them, at creation time). In the non-SMP case, we can write a single/generic check. IMO, the current structure of the method and comments highlight this difference.

> in this case, we would do that collapsing check in Controller::peekAtLocal().

Yes, but peekAtLocal() may not be a good place for additional consistency checks because, besides find(), that method is used in evictIfFound() which should be rather aggressive -- if there is doubt, it should release the entry, not pretend that it does not exist like find() does.

> In other words, checkPeeked() may be probably renamed because it checks something produced by several methods (including peek()).

I agree but do not have good suggestions (unfortunately). I doubt something like checkPeekedAndPublic() or checkSharable() would be better. FWIW, I started by inlining this code, but it produced too much noise in the already non-trivial logic, so I extracted it into a method. I think this code deserves its own method, but checkPeeked() is the best name I can come up with. I know it is far from perfect. Suggestions welcomed.

Would you prefer checkFoundCandidate()?

rousskov: > not pretend that it does not exist like find() does.

Well, find() also releases doubtful entries so that distinction is misleading. We could add some checks to peekAtLocal() if needed, but if we can centralized checks, it is better to centralize them.

eduard-bagdasaryan: > Would you prefer checkFoundCandidate()?

Or checkCandidate() - since we are already inside find() method.

> if we can centralized checks, it is better to centralize them

Ok.

DrDaveD:
@rousskov Unfortunately it is very difficult for me to test as-is because I only use squid4, and this does not port cleanly back to squid4.  I can't figure out how to port it either, because a function called from Transient::get() does not even exist: setCollapsingRequirement().   There's another change in Store::Controller::allowSharing() that moves a call to transients->hasWriter(), but the v4 version of that function does not do that call at all.  

It does apply cleanly to v5, by the way.  I suppose I could try seeing if I can compile v5 and test that, but in any case it wouldn't get any testing on production servers.  Could you instead make a comparable v4 version of the PR?

I think it is unfortunate that pr #566 was just released with squid-4.12, because other people that have negative caching and collapsed forwarding enabled could also start hitting this.

rousskov:
> Could you instead make a comparable v4 version of the PR?

IMO, for v4, we should use your branch as of commit 55cb158f9643b95e84f88a39349b2a0ca2c9ea53. I am not sure v4 even has the right scaffolding for the Transients-driven changes, and I do not have enough free time to figure it out right now. If other collapsed revalidation problems (targeted by the latest changes) surface in v4, we will have to deal with them (or ask folks to upgrade). That v4-targeting PR can be submitted after this PR is merged.

If you cannot test v5 or master, we will have to rely on the existing (still minimal) testing for this PR. We should add automated tests targeting collapsed revalidation specifically, but we are still waiting for somebody to sponsor that work.

DrDaveD:
Ok I will first test v4 as of 55cb158 on top of the 4.12 release.  Should we have a separate PR for that?   I will also see if I can compile & test v5 with this PR.

rousskov:
> Should we have a separate PR for that?

Yes, that v4-targeting PR should be created after this PR is merged.

> I will also see if I can compile & test v5 with this PR.

Thank you. I will post another polishing change to address Eduard's concern. Give me a minute before testing v5.

rousskov:
> I will post another polishing change to address Eduard's concern. Give me a minute before testing v5.

Done at 3c6f947.

src/store/Controller.cc
rousskov: I used checkFoundCandidate() in 3c6f947. The checks (or their interpretation) kind of depends on being done in the find() context...

DrDaveD:
@rousskov I did successfully build squid-5.0.3.  However, I could not reproduce the problem there.  I also tested it with this patch applied, and saw no difference.

rousskov:
> I did successfully build squid-5.0.3. However, I could not reproduce the problem there. I also tested it with this patch applied, and saw no difference.

@DrDaveD, thank you for reporting these results. I suggest merging this PR even though you could not reproduce the problem in v5. I suspect the original problem or a similar one can be reproduced with more effort, and even if it cannot be, the changes in this PR make the code better (and may prevent other similar problems in the future). Finally, v4 does need a fix that partially overlaps with this PR, and it is best to keep the code in sync (with all other factors being equal).

@yadij, if you agree, please clear this PR for merging.

DrDaveD:
Done in #683

src/client_side_reply.cc
rousskov: > I plan to submit a small PR dedicated to those other problems

Done: #686 .

PR #654, opened 2020-06-05
make the way version.h is included standard

kinkie:
include guards for version.h are in squid.h
Change this to use the standard approach


rousskov:
I am abstaining. I do not know what `sub-packages` are. Perhaps this is related to how Squid is packaged? I am asking @yadij to double check this.

kinkie:
That line was last changed 12 years ago, and the HAVE_SQUID symbol is defined nowhere. It's dead code

rousskov:
I was worried that something that packages Squid defines `HAVE_SQUID` to provide custom versions (despite `snmp.h` including `versions.h` unconditionally). That is why I asked Amos to double check. I did not mean to imply that there is something wrong.

kinkie:
No worries, I am perfectly fine with double checking, I was sharing context

yadij:
IIRC the HAVE_SQUID macro was for the kerberos helpers and purge tool when they were first bundled. The wrapper is no longer used by them, so it can go now.

include/version.h
```diff
@@ -22,3 +25,4 @@
 #define APP_FULLNAME  PACKAGE "/" VERSION
 #endif
 
+#endif /* SQUID_VERSION_H */
```
yadij, line 28: git complains, missing newline before EOF
```suggestion
#endif /* SQUID_VERSION_H */

```

kinkie:
@yadij mind unblocking this PR after the change? Thanks!

PR #655, opened 2020-06-05
Remove leakfinder (--enable-leakfinder)

kinkie:
There are better leak finding tools for most use cases now. The removed
code can be manually added back for experiments in other use cases. The
original leakfinder design required code modifications anyway.


yadij:
Why restore enable_leakfinder?  those bits appear to no longer be used in the code as well.

kinkie:
--enable-leakfinder enables the build of src/{LeakFinder.h,cc}, it's not dead code.
I'm not convinced it has been used in a long tme, but I can't rule it out right now.

kinkie:
I'm happy to remove everything; since this was implemented tooling has gone a long way. Updating the PR.

rousskov:
I adjusted PR description to clarify our reasons and fix formatting. I also changed the title to add a well-known ./configure string that some folks may want to search for in `git log`. Please review these changes before clearing this PR for merging.

kinkie:
LGTM. Clearing for merge

PR #29, opened 2017-07-22
Maintenance for git

yadij:
Update the repository ignored files list for git, also include a few entries that should have been added long ago.

PR #659, opened 2020-06-06
Add .gitignore file

kinkie:


kinkie:
This iteration is specific to the tools I use; but it would make sense in my opinion to list here the files we never want to add to our sources, such as for SPONSORS

.gitignore
```diff
@@ -0,0 +1,3 @@
+.vscode
+.editorconfig
```
rousskov, line 2: Are you abandoning #591 that adds `.editorconfig`? AFAICT, #591 ball is on your side; it may be better to advance that PR before this one. I do not insist on removing this entry from this PR, just pointing out a conflict between your open PRs.

.gitignore
```diff
@@ -0,0 +1,3 @@
...
+.bzr
```
rousskov, line 3: I am just curious: Why do you have that file or directory in a git repository?

.gitignore
kinkie: No, I am not, this is a side-step to have one for myself, while waiting for the global one to be accepted (or not).
I''ve submitted PR#660, cleaning up PERL code, which can be a step forward for PR#591

rousskov:
> I think we shouldn't obsess over a policy

I agree. I think we should not obsess over anything.

> a .gitignore file can very well be our policy.

I do not understand what you mean, especially in a context where my proposed additions to .gitignore have been effectively rejected in #29 . If you mean the "core developers can add whatever entries they want to  be ignored in their setup" policy, then I doubt such a policy is a good idea, but we can use that as a starting point if you prefer.

FWIW, I have not heard any valid arguments against my [original](https://github.com/squid-cache/squid/pull/29#issuecomment-317583443) proposal to list all untracked files that can be generated in a working directory by Squid build command(s). I think we can expand that to typically-embedded configuration `.files` of common Squid development tools. That combination will take care of two out of three entries proposed in this PR (and I can add many more from [my](https://gist.github.com/rousskov/c10fe5587034294cebdaa2415847bdd2) own `.gitignore`).

Needless to say, if you officially add your personal Squid `.gitignore`, I will have to abandon mine, and vice versa, so we should either find an inclusive official compromise or do nothing (officially). A middle-ground official solution would be worse than none here.

> Specific tools configuration file (as in this case), and autogenerated files feels like a first good start.

I agree. The second part is what I proposed in #29 .

> None of this is necessary, but it can avoid boring review iterations by preventing trivial mistakes.

I agree.

.gitignore
kinkie: We don't. The objective is to ensure we don't add it by mistake

kinkie:
Second iteration tried. Feel free to push to this PR with any amendments you think are warranted, let's iterate in code instead of words

.gitignore
rousskov: Understood, thank you!

FWIW, I do not think we should list files just because they should not be added to the repository. The list of such files is nearly infinite. Fortunately, policies suggested so far did not include such files IIRC.

Sorry for using words. I will try the [suggested](https://github.com/squid-cache/squid/pull/659#issuecomment-640605278) "iterate in code" approach from now on (for this PR) -- I noticed that suggestion after writing the above.

rousskov:
> Feel free to push to this PR with any amendments you think are warranted, let's iterate in code instead of words

OK, please wait for my push.

kinkie:
i think adding the generated binaries is a bit overkill, but that's my only observation on the additions.
If you feel strongly about it, we can proceed, but otherwise I'd remove them before merging

yadij:
My objection to **all auto-generated files** being listed remains strong. Developers, distributors, and users all need to be able to see when their checkout has files that will interfere with the build process, or vice versa be altered by next build. An out-of-tree build directory (eg the btlayer-* pattern) is **best practice** for development - especially when doing iterative work that needs build objects to stay.

kinkie:
@yadij what is your proposal then? To not have a .gitignore at all? Or to have what in there?

rousskov:
> Francesco: i think adding the generated binaries is a bit overkill

FWIW, I do not understand the reasoning behind that statement.

> Amos: Developers, distributors, and users all need to be able to see when their checkout has files that will interfere with the build process, or vice versa be altered by next build.

And they are able to see that equally well with or without the generated files being in .gitignore. Naturally, their setups and commands will depend, in part, on .gitignore.

Each developer can devise an approach that works best for them. This is the question of the best _default_ approach rather than the question of enabling or disabling some fundamental "ability".

rousskov:
> Francesco: If you feel strongly about it, we can proceed

To avoid misunderstanding, I "feel strongly" about developing and following a good `.gitignore` membership policy. I hope my proposal, based on a basic "keep git status clean to see repository changes" principle, sets an example. I am ready to support (or abstain from blocking) other good policies (and their matching `.gitignore`s).

FWIW, a bad `.gitignore`, especially the one created without good membership rules, is a short- and long-term inconvenience/overhead (at least for me). That is the primary reason I am still wasting time on this sad repetition of the #29 fiasco.

yadij:
> @yadij what is your proposal then? To not have a .gitignore at all? Or to have what in there?

Not having .gitignore at all has proven better so far than ad-hoc or a controversial policy.

rousskov:
> Not having .gitignore at all has proven better so far than ad-hoc or a controversial policy.

For me, having no official .gitignore is better than having a bad official .gitignore. Just one more annoyance to work around.

I speculate that for most new Squid developers, having an official .gitignore based on my simple policy would be better than having no official .gitignore because, with that .gitignore, they will see a clean `git status -sb` that would immediately expose uncommitted/forgotten sources, saving a lot of time and embarrassment. I am sure there are exceptions though, so I do not have much hope in convincing Amos that it is a good policy -- there will always be some poorly specified esoteric use case that I would not be able to exclude as clearly invalid. We have been down that path in #29 .

kinkie:
I'm closing this PR without merge; it's quite obvious we won't reach any sort of consensus. Sorry for stirring this hornet's nest

PR #665, opened 2020-06-10
Source Format Enforcement

squidadm:
Does not contain astyle-driven changes.

Makefile changes are what commit 2a06115 mistakenly claimed to contain.


kinkie:
It probably needs to be rebased past the merge of refactor-leakfinder

lib/profiler/xprof_type.h
```diff
@@ -11,79 +11,77 @@
...
+XPROF_StringAllocAndFill,
+XPROF_StringAppend,
+XPROF_StringClean,
+XPROF_StringInitBuf,
+XPROF_StringReset,
+XPROF_write,
+XPROF_xcalloc,
+XPROF_xmalloc,
+XPROF_xrealloc,
+  XPROF_LAST } xprof_type;
```
rousskov, line 85: What happened to this generated file formatting? I think it became worse, one several counts. Perhaps it was formatted before and now is not? This formatting change seems odd in a fully-automated commit. Is there a regression or an old problem we should fix before this goes in?

src/tests/Stub.list
```diff
@@ -12,6 +12,7 @@ STUB_SOURCE= tests/STUB.h \
 	tests/stub_cache_cf.cc \
```
rousskov, line 12: @kinkie, it looks like this generated Makefile include file does not go through the formatting script that you added in 2a06115. IMO, we should change the file extension to `.am` (which we already use for other Makefile include files) so that it is properly formatted. If you volunteer to fix this, please also check that this file is generated _before_ Makefiles are formatted. Thank you!

lib/profiler/xprof_type.h
yadij: The astyle version on master had a bump to 2.05.1 so it no longer gets run by the automation. It was being relied on to fix this up along with any dev added mistakes.

My initial checks of the new astyle are that is fixes a few things better, but also makes some files worse. I'll be working on both these in the next day or so.

rousskov: Understood. I think "format enforcement" changes should use the official astyle version. Furthermore, a switch from one official astyle version to another deserves a dedicated PR. The state of "astyle version on master [box]" does not affect the former and is dependent on the latter.

yadij:
FTR: scope of this PR is to run the scripts/sourcemaintenance.sh on master and merge changes the automation performs.

xprof_type.h is broken by the astyle version issues so manually reverting alterations to that specific file until PR #669 merged.

lib/profiler/xprof_type.h
yadij: Just to be clear this PR does *not* include any astyle changes. The version issue prevented astyle being run for this enforcement update - which is why the xprof file shows changes.

src/tests/Stub.list
yadij: This is being handled in other PR.

lib/profiler/xprof_type.h
rousskov: > Just to be clear this PR does not include any astyle changes.

... which, IMO, is a problem that should be addressed before merging this PR.

> The version issue prevented astyle being run for this enforcement update

AFAICT, the "version issue" is not some insurmountable problem that has to be ignored. Instead of hiding and now ignoring it, we should fix it before this PR is merged.

FWIW, in my experience, it is fairly straightforward easy to build astyle from source if needed.

PR #669, opened 2020-06-18
Maintenance: fix syntax of profiler/xprof_type.h

yadij:
Generate xprof_type.h compliant with Squid coding
guidelines to prevent syntax collisions when astyle
is not available to enforce correctness.

### DIFF
--- a/scripts/source-maintenance.sh
+++ b/scripts/source-maintenance.sh
@@ -312,11 +312,13 @@ echo "#define _PROFILER_XPROF_TYPE_H_"
 echo "/* AUTO-GENERATED FILE */"
 echo "#if USE_XPROF_STATS"
 echo "typedef enum {"
-echo "  XPROF_PROF_UNACCOUNTED,"
-grep -R -h "PROF_start.*" ./* | grep -v probename | sed -e 's/ //g; s/PROF_start(/XPROF_/; s/);/,/' | sort -u
-echo "  XPROF_LAST } xprof_type;"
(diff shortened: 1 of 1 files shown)


scripts/source-maintenance.sh
```diff
@@ -312,11 +312,13 @@ echo "#define _PROFILER_XPROF_TYPE_H_"
 echo "/* AUTO-GENERATED FILE */"
 echo "#if USE_XPROF_STATS"
 echo "typedef enum {"
-echo "  XPROF_PROF_UNACCOUNTED,"
-grep -R -h "PROF_start.*" ./* | grep -v probename | sed -e 's/ //g; s/PROF_start(/XPROF_/; s/);/,/' | sort -u
-echo "  XPROF_LAST } xprof_type;"
+echo "    XPROF_PROF_UNACCOUNTED,"
+grep -R -h "PROF_start.*" ./* | grep -v probename | sed -e 's/ //g; s/PROF_start(/XPROF_/; s/);/,/; s/XPROF_/    XPROF_/' | sort -u
```
rousskov, line 316: IMHO, adding indentation when replacing the opening sequence -- `PROF_start(` -- would be a bit clearer:

```suggestion
grep -R -h "PROF_start.*" ./* | grep -v probename | sed -e 's/ //g; s/PROF_start(/    XPROF_/; s/);/,/' | sort -u
```

I do not insist on this untested change.

yadij: Tested and done in 1531a6d

PR #670, opened 2020-06-18
Maintenance: rename .list auto-generated files to .am

yadij:
The contents are automake script. Also this allows
the source format enforcement to correctly detect
that and enforce our automake code guidelines.


scripts/source-maintenance.sh
```diff
@@ -378,7 +378,7 @@ do
 	echo -n "	${f}"
 done
 echo " "
-)| sed s%src/%%g >src/tests/Stub.list
+)| sed s%src/%%g >src/tests/Stub.am
```
rousskov, line 381:
```suggestion
) | sed s%src/%%g > src/tests/Stub.am
```

I do not insist on this "(arguably) clearer/safer formatting" change.

scripts/source-maintenance.sh
```diff
@@ -320,65 +319,31 @@ echo "#endif"
 ) >lib/profiler/list
 mv lib/profiler/list lib/profiler/xprof_type.h
 
+amFileGenerate ()
```
rousskov, line 322: I suggest renaming this function to avoid Yoda words order and emphasize that the function does not create any files:

```suggestion
printAmFile ()
```

or 

```suggestion
printAutomakefile ()
```

rousskov:
@kinkie, did you try to restore @eduard-bagdasaryan's workaround for Jenkins build failures like the [one](https://build.squid-cache.org/job/5-pr-test/2308/COMPILER=gcc,OS=centos-7,label=docker-build-host/console) this PR suffers from? It was lost during Jenkins migration IIRC. Please reach out to Eduard if you need assistance beyond his old email about that workaround (using two Jenkins plugins IIRC).

scripts/source-maintenance.sh
yadij: Done in 2516080

PR #677, opened 2020-06-21
Fix "make check" with GCC-10

yadij:
Latest GCC-10 release on some OS (Debian initially) no
longer locates files in "." by default. Fix Makefile hack
pulling tests/ files into tools/ for stub linking.

Also, Squid #includes under src/ are supposed to always
use path relative to src/. Fix wrong STUB.h references.


yadij:
Moving tools/ into src/ is a larger discussion that I think we will have a lot of disagreements over. So yes well outside of scope for this PR. I would also like to defer that discussion until a lot more of the files currently at the top level of src/ have been refactored into SourceLayout convenience libraries.

PR #678, opened 2020-06-21
Turn asserts to Must checks in DescriptorSet

kinkie:


kinkie:
I am confused. How can I distinguish a TODO to work on from one that is not to be worked on?
I feel I'm wasting my time going on like this

rousskov:
> How can I distinguish a TODO to work on from one that is not to be worked on?

IMO, making the right decision usually requires understanding of the underlying issue (which may have changed since the comment was added) and judgement calls. AFAICT, this is similar to any other development project, except it is possible that comments become stale faster than code, and there is no explicit external "stakeholder" that can motivate and guide your design choices. From that point of view, addressing old code TODOs is riskier than doing "regular" development!

This is not required in any shape of form, but if you want my support for a given change in advance of that change, then I recommend talking to me _before_ doing the bulk of the development. Alternatively, if you want my recommendations on what TODOs to work on, I may be able to suggest a few. Needless to say, Amos may have a completely different position on any given TODO...

In this particular case, a possibly stale TODO hints at a much bigger problem. It is not that there is no problem to solve here. There is, and an important one! Unfortunately, the problem differs from what the literal interpretation of a possibly stale TODO implies.

> I feel I'm wasting my time going on like this

I share your concern (and it is not just your time being wasted, of course). On the other hand, some of your recent PRs _were_ useful (thank you!). Hopefully, we can find the way to eliminate or drastically reduce the waste.

PR #679, opened 2020-06-22
Refactror AccessLogEntry

kinkie:
Move inner classes to toplevel
Rename some fields
Remove some TODOs
Reshuffle destructrors to appropriate class

### DIFF
--- a/src/AccessLogEntry.cc
+++ b/src/AccessLogEntry.cc
@@ -171,3 +171,9 @@ AccessLogEntry::packReplyHeaders(MemBuf &mb) const
         reply->packHeadersUsingFastPacker(mb);
 }
 
+#if ICAP_CLIENT
+AccessLogEntryIcapDetails::~AccessLogEntryIcapDetails() {
+    HTTPMSGUNLOCK(reply);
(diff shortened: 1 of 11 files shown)


yadij:
FYI the TODO triggering this PR is not as simple as it seems. The long-term plan is to replace ALE with MasterXaction - not just extracting the internal class definitions, but also adding a sane hierarchy for sub-transactions (so eg a bumped HTTPS transaction can back-reference to its parent CONNECT state), re-typing for better memory management, and re-naming for more accurate uses.

I have a branch https://github.com/yadij/squid/commits/arc-transport which performs the next steps towards that end goal - making the MXA class able to spawn children for sub-transactions. Any assistance testing and extending that work is welcome.

rousskov:
> The long-term plan is to replace ALE with MasterXaction

And even that long-term plan may be wrong or misinterpreted.This is a very difficult code area, with several concurrent projects improving the related code and trying to find the right solution for several big outstanding problems. I do not recommend "polishing" ALE right now.

kinkie:
abandoning

PR #680, opened 2020-06-23
Bug 5048: ResolvedPeers.cc:35: "found != paths_.end()" assertion

eduard-bagdasaryan:
This bug was caused by a4d576d. ResolvedPeers::retryPath() used
connection object pointers to find the original path location under the
incorrect assumption that the Connection pointers never change. That
assumption was wrong for persistent connections: ResolvedPeers stores a
half-baked Connection object rather than the corresponding open
Connection object that the unfortunate caller got from fwdPconnPool and
passed to ResolvedPeers::retryPath().

While working on a fix, we discovered a second reason why the pointer
comparison was a bad idea (and why a simpler fix of comparing addresses
rather than pointers was also a bad idea): The peer selection code
promises to deliver unique destinations, but we now suspect that, under
presumably rare circumstances, it may deliver duplicates. This broken
promise is now an out-of-scope XXX in PeerSelector::addSelection().

Squid now eliminates the search (and the previously documented concern
about its slow speed) by remembering Connection's position in the
destination array. The position is remembered in a smart Connection
pointer (compatible with Comm::ConnectionPointer) so that the rest of
the code (that does not really care about all this) is preserved largely
unchanged. Most changes are just renaming the pointer type.

Also freshened a FwdState call and comment made stale by 25b0ce4.

Also marked an out-of-scope problem in maybeGivePrimeItsChance() caller.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -483,7 +483,10 @@ FwdState::fail(ErrorState * errorState)
     if (pconnRace == racePossible) {
         debugs(17, 5, HERE << "pconn race happened");
         pconnRace = raceHappened;
-        destinations->retryPath(serverConn);
+        if (destinationReceipt) {
+            destinations->returnPath(destinationReceipt);
+            destinationReceipt = nullptr;
+        }
     }
 
     if (ConnStateData *pinned_connection = request->pinnedConnection()) {
@@ -504,6 +507,7 @@ FwdState::unregister(Comm::ConnectionPointer &conn)
     comm_remove_close_handler(conn->fd, closeHandler);
     closeHandler = NULL;
     serverConn = NULL;
+    destinationReceipt = nullptr;
 }
 
 // \deprecated use unregister(Comm::ConnectionPointer &conn) instead
@@ -789,6 +793,8 @@ FwdState::advanceDestination(const char *stepDescription, const Comm::Connection
 void
 FwdState::noteConnection(HappyConnOpener::Answer &answer)
 {
+    assert(!destinationReceipt);
+
     calls.connector = nullptr;
     connOpener.clear();
 
@@ -802,14 +808,25 @@ FwdState::noteConnection(HappyConnOpener::Answer &answer)
         Must(!Comm::IsConnOpen(answer.conn));
         answer.error.clear(); // preserve error for errorSendComplete()
     } else if (!Comm::IsConnOpen(answer.conn) || fd_table[answer.conn->fd].closing()) {
+        // We do not know exactly why the connection got closed, so we play it
+        // safe, allowing retries only for persistent (reused) connections
+        if (answer.reused) {
+            destinationReceipt = answer.conn;
+            assert(destinationReceipt);
+        }
         syncHierNote(answer.conn, request->url.host());
         closePendingConnection(answer.conn, "conn was closed while waiting for noteConnection");
         error = new ErrorState(ERR_CANNOT_FORWARD, Http::scServiceUnavailable, request, al);
+    } else {
+        assert(!error);
+        destinationReceipt = answer.conn;
+        assert(destinationReceipt);
+        // serverConn remains nil until syncWithServerConn()
     }
 
     if (error) {
         fail(error);
-        retryOrBail(); // will notice flags.dont_retry and bail
+        retryOrBail();
         return;
     }
 
@@ -848,7 +865,7 @@ FwdState::establishTunnelThruProxy(const Comm::ConnectionPointer &conn)
                                             "FwdState::tunnelEstablishmentDone",
                                             Http::Tunneler::CbDialer<FwdState>(&FwdState::tunnelEstablishmentDone, this));
     HttpRequest::Pointer requestPointer = request;
-    const auto tunneler = new Http::Tunneler(conn, requestPointer, callback, connectingTimeout(serverConnection()), al);
+    const auto tunneler = new Http::Tunneler(conn, requestPointer, callback, connectingTimeout(conn), al);
 
     // TODO: Replace this hack with proper Comm::Connection-Pool association
     // that is not tied to fwdPconnPool and can handle disappearing pools.
@@ -999,6 +1016,7 @@ FwdState::syncWithServerConn(const Comm::ConnectionPointer &conn, const char *ho
 {
     Must(IsConnOpen(conn));
     serverConn = conn;
+    // no effect on destinationReceipt (which may even be nil here)
 
     closeHandler = comm_add_close_handler(serverConn->fd,  fwdServerClosedWrapper, this);
 
@@ -1043,6 +1061,7 @@ FwdState::connectStart()
     err = nullptr;
     request->clearError();
     serverConn = nullptr;
+    destinationReceipt = nullptr;
 
     request->hier.startPeerClock();
 
@@ -1073,6 +1092,8 @@ FwdState::usePinned()
     debugs(17, 7, "connection manager: " << connManager);
 
     try {
+        // TODO: Refactor syncWithServerConn() and callers to always set
+        // serverConn inside that method.
         serverConn = ConnStateData::BorrowPinnedConnection(request, al);
         debugs(17, 5, "connection: " << serverConn);
     } catch (ErrorState * const anErr) {
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -484,7 +484,7 @@ FwdState::fail(ErrorState * errorState)
         debugs(17, 5, HERE << "pconn race happened");
         pconnRace = raceHappened;
         if (destinationReceipt) {
-            destinations->returnPath(destinationReceipt);
+            destinations->reinstatePath(destinationReceipt);
             destinationReceipt = nullptr;
         }
     }
--- a/src/HappyConnOpener.cc
+++ b/src/HappyConnOpener.cc
@@ -18,7 +18,6 @@
 #include "neighbors.h"
 #include "pconn.h"
 #include "PeerPoolMgr.h"
-#include "ResolvedPeers.h"
 #include "SquidConfig.h"
 
 CBDATA_CLASS_INIT(HappyConnOpener);
@@ -455,7 +454,7 @@ HappyConnOpener::makeError(const err_type type) const
 
 /// \returns pre-filled Answer if the initiator needs an answer (or nil)
 HappyConnOpener::Answer *
-HappyConnOpener::futureAnswer(const Comm::ConnectionPointer &conn)
+HappyConnOpener::futureAnswer(const PeerConnectionPointer &conn)
 {
     if (callback_ && !callback_->canceled()) {
         const auto answer = dynamic_cast<Answer *>(callback_->getDialer());
@@ -469,7 +468,7 @@ HappyConnOpener::futureAnswer(const Comm::ConnectionPointer &conn)
 
 /// send a successful result to the initiator (if it still needs an answer)
 void
-HappyConnOpener::sendSuccess(const Comm::ConnectionPointer &conn, bool reused, const char *connKind)
+HappyConnOpener::sendSuccess(const PeerConnectionPointer &conn, const bool reused, const char *connKind)
 {
     debugs(17, 4, connKind << ": " << conn);
     if (auto *answer = futureAnswer(conn)) {
@@ -485,7 +484,7 @@ void
 HappyConnOpener::cancelAttempt(Attempt &attempt, const char *reason)
 {
     Must(attempt);
-    destinations->retryPath(attempt.path); // before attempt.cancel() clears path
+    destinations->returnPath(attempt.path); // before attempt.cancel() clears path
     attempt.cancel(reason);
 }
 
@@ -514,7 +513,7 @@ HappyConnOpener::noteCandidatesChange()
 
 /// starts opening (or reusing) a connection to the given destination
 void
-HappyConnOpener::startConnecting(Attempt &attempt, Comm::ConnectionPointer &dest)
+HappyConnOpener::startConnecting(Attempt &attempt, PeerConnectionPointer &dest)
 {
     Must(!attempt.path);
     Must(!attempt.connector);
@@ -530,13 +529,14 @@ HappyConnOpener::startConnecting(Attempt &attempt, Comm::ConnectionPointer &dest
 /// \returns true if and only if reuse was possible
 /// must be called via startConnecting()
 bool
-HappyConnOpener::reuseOldConnection(const Comm::ConnectionPointer &dest)
+HappyConnOpener::reuseOldConnection(PeerConnectionPointer &dest)
 {
     assert(allowPconn_);
 
     if (const auto pconn = fwdPconnPool->pop(dest, host_, retriable_)) {
         ++n_tries;
-        sendSuccess(pconn, true, "reused connection");
+        dest.finalize(pconn);
+        sendSuccess(dest, true, "reused connection");
         return true;
     }
 
@@ -546,7 +546,7 @@ HappyConnOpener::reuseOldConnection(const Comm::ConnectionPointer &dest)
 /// opens a fresh connection to the given destination
 /// must be called via startConnecting()
 void
-HappyConnOpener::openFreshConnection(Attempt &attempt, Comm::ConnectionPointer &dest)
+HappyConnOpener::openFreshConnection(Attempt &attempt, PeerConnectionPointer &dest)
 {
 #if URL_CHECKSUM_DEBUG
     entry->mem_obj->checkUrlChecksum();
@@ -583,9 +583,12 @@ HappyConnOpener::connectDone(const CommConnectCbParams &params)
     const bool itWasSpare = (params.conn == spare.path);
     Must(itWasPrime != itWasSpare);
 
+    PeerConnectionPointer handledPath;
     if (itWasPrime) {
+        handledPath = prime.path;
         prime.finish();
     } else {
+        handledPath = spare.path;
         spare.finish();
         if (gotSpareAllowance) {
             TheSpareAllowanceGiver.jobUsedAllowance();
@@ -595,7 +598,7 @@ HappyConnOpener::connectDone(const CommConnectCbParams &params)
 
     const char *what = itWasPrime ? "new prime connection" : "new spare connection";
     if (params.flag == Comm::OK) {
-        sendSuccess(params.conn, false, what);
+        sendSuccess(handledPath, false, what);
         return;
     }
 
@@ -605,7 +608,7 @@ HappyConnOpener::connectDone(const CommConnectCbParams &params)
     params.conn->close(); // TODO: Comm::ConnOpener should do this instead.
 
     // remember the last failure (we forward it if we cannot connect anywhere)
-    lastFailedConnection = params.conn;
+    lastFailedConnection = handledPath;
     delete lastError;
     lastError = nullptr; // in case makeError() throws
     lastError = makeError(ERR_CONNECT_FAIL);
@@ -717,11 +720,14 @@ HappyConnOpener::checkForNewConnection()
     // open a new prime and/or a new spare connection if needed
     if (!destinations->empty()) {
         if (!currentPeer) {
-            currentPeer = destinations->extractFront();
+            auto newPrime = destinations->extractFront();
+            currentPeer = newPrime;
             Must(currentPeer);
             debugs(17, 7, "new peer " << *currentPeer);
             primeStart = current_dtime;
-            startConnecting(prime, currentPeer);
+            startConnecting(prime, newPrime);
+            // XXX: if reuseOldConnection() in startConnecting() above succeeds,
+            // then we should not get here, and Must(prime) below will fail.
             maybeGivePrimeItsChance();
             Must(prime); // entering state #1.1
         } else {
--- a/src/HappyConnOpener.cc
+++ b/src/HappyConnOpener.cc
@@ -484,7 +484,7 @@ void
 HappyConnOpener::cancelAttempt(Attempt &attempt, const char *reason)
 {
     Must(attempt);
-    destinations->returnPath(attempt.path); // before attempt.cancel() clears path
+    destinations->reinstatePath(attempt.path); // before attempt.cancel() clears path
     attempt.cancel(reason);
 }
 
--- a/src/HappyConnOpener.cc
+++ b/src/HappyConnOpener.cc
@@ -726,7 +726,7 @@ HappyConnOpener::checkForNewConnection()
             debugs(17, 7, "new peer " << *currentPeer);
             primeStart = current_dtime;
             startConnecting(prime, newPrime);
-            // XXX: if reuseOldConnection() in startConnecting() above succeeds,
+            // TODO: if reuseOldConnection() in startConnecting() above succeeds,
             // then we should not get here, and Must(prime) below will fail.
             maybeGivePrimeItsChance();
             Must(prime); // entering state #1.1
--- a/src/ResolvedPeers.cc
+++ b/src/ResolvedPeers.cc
@@ -13,8 +13,6 @@
 #include "ResolvedPeers.h"
 #include "SquidConfig.h"
 
-/* ResolvedPeers */
-
 ResolvedPeers::ResolvedPeers()
 {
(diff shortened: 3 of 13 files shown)


src/FwdState.cc
```diff
@@ -802,14 +808,25 @@ FwdState::noteConnection(HappyConnOpener::Answer &answer)
...
+        assert(!error);
+        destinationReceipt = answer.conn;
+        assert(destinationReceipt);
+        // serverConn remains nil until syncWithServerConn()
     }
 
     if (error) {
         fail(error);
-        retryOrBail(); // will notice flags.dont_retry and bail
+        retryOrBail();
```
rousskov, line 829: FTR: The documentation for this comment removal is in unofficial commit 22f55cda04f2c7b15e8b2da83b1d3f52bba43c60 message, quoted here:

> Removed stale (and dangerously misleading) comment

> This branch changes did not make that comment stale -- FwdState may actually retry if "conn was closed while waiting for noteConnection", but the comment is in the way of correctly interpreting the related code that is being modified in this branch.

src/comm/Connection.h
```diff
@@ -137,12 +140,9 @@ class Connection: public CodeContext
...
 
-private:
-    /** These objects may not be exactly duplicated. Use copyDetails() instead. */
-    Connection(const Connection &c);
-
-    /** These objects may not be exactly duplicated. Use copyDetails() instead. */
-    Connection & operator =(const Connection &c);
+    // TODO: Declare and use (here and elsewhere) NonCopyable, NonMovable, etc.
+    /// no C++ copying/moving; see clone*() methods for custom/partial copies
+    Connection &operator =(Connection &&) = delete;
```
rousskov, line 145: FTR: The explanation behind this change is in the last paragraph of the unofficial commit 46f70173f1ab392bf3b8b6755f0fdfab8124e1e2 message, quoted here:

> I had to refactor undefined Connection copy constructors because their comments referred to the renamed copyDetails() and should now refer to the second cloning method as well. Modern code should also avoid making these methods private and should disable moving (where it is unsupported). Deleting the move assignment operator is not consistent with how we often delete the move constructor instead, but I found some references that seem to recommend this method now, including https://stackoverflow.com/a/48865077. I left a TODO that would refactor this properly, for all affected classes.

@kinkie, BTW, this may be a good TODO candidate for you: It is relatively compact, useful for future development, and arguably educational/interesting (if a bit challenging). The key for this TODO is to come up with the smallest set of reusable base classes that leaves no holes in the current (and clearly foreseeable future) Squid needs. Taking a clue from existing libraries that do this already (e.g., Boost?) may be a good idea, but blind copying of their APIs would probably miss some important Squid specifics. Locating existing Squid classes that should get these base classes instead of their ad hoc hacks may be a bit challenging as well.

src/comm/Connection.h
```diff
@@ -77,10 +77,13 @@ class Connection: public CodeContext
...
-     * This excludes the FD. The new copy will be a closed connection.
-     */
-    ConnectionPointer copyDetails() const;
+    /// Create a new (closed) IDENT Connection object based on our from-Squid
+    /// connection properties.
+    ConnectionPointer cloneIdentDetails() const;
+
+    /// Create a new (closed) Connection object pointing to the same destination
+    /// as this from-Squid connection.
+    ConnectionPointer cloneDestinationDetails() const;
```
rousskov, line 86: FTR: From the unofficial commit 46f70173f1ab392bf3b8b6755f0fdfab8124e1e2 message, quoted here with minor edits/updates for clarity:

> I renamed copyDetails() because a reader like me would be puzzled by the copyDestinationDetails() vs. copyDetails() choice: Which one do I want? The new pair of names makes the choice easier. ... Finally, I prefer "clone" for methods that create new objects. The "copy" prefix is better suited for methods that fill existing objects.

src/ResolvedPeers.h
```diff
@@ -44,19 +47,20 @@ class ResolvedPeers: public RefCountable
     /// add a candidate path to try after all the existing paths
     void addPath(const Comm::ConnectionPointer &);
 
-    /// re-inserts the previously extracted address into the same position
-    void retryPath(const Comm::ConnectionPointer &);
+    /// makes the previously extracted path available for extraction at its
+    /// original position
+    void returnPath(const PeerConnectionPointer &);
```
rousskov, line 52: FTR: The explanation behind this renaming is in the first paragraph of the unofficial commit 46f70173f1ab392bf3b8b6755f0fdfab8124e1e2 message, quoted here:

> I renamed retryPath() because ResolvedPeers do not retry anything. Also, if we keep retryPath(), then PeerConnection::returnable() should become retriable() and that would clash badly with other existing methods and concepts.

src/peer_select.cc
```diff
@@ -1083,6 +1083,8 @@ PeerSelector::addSelection(CachePeer *peer, const hier_code code)
         // There can be at most one PINNED destination.
         // Non-PINNED destinations are uniquely identified by their CachePeer
         // (even though a DIRECT destination might match a cache_peer address).
+        // XXX: We may still add duplicates because the same peer could have
+        // been removed from `servers` already (and given to the requestor).
         const bool duplicate = (server->code == PINNED) ?
                                (code == PINNED) : (server->_peer == peer);
```
rousskov, line 1089: FTR: This newly discovered XXX may result in occasional violation of the documented peer selection [algorithm](https://wiki.squid-cache.org/Features/LoadBalance#Overall_peer_selection_logic), but its effects, if any, should be very rare and mild in most environments.

src/HappyConnOpener.cc
```diff
@@ -717,11 +720,14 @@ HappyConnOpener::checkForNewConnection()
...
+            currentPeer = newPrime;
             Must(currentPeer);
             debugs(17, 7, "new peer " << *currentPeer);
             primeStart = current_dtime;
-            startConnecting(prime, currentPeer);
+            startConnecting(prime, newPrime);
+            // XXX: if reuseOldConnection() in startConnecting() above succeeds,
+            // then we should not get here, and Must(prime) below will fail.
             maybeGivePrimeItsChance();
             Must(prime); // entering state #1.1
```
rousskov, line 732: I speculate that this bug has no significant runtime effect (in most cases) because a successful startConnecting() sends the persistent connection to the requestor. The subsequent and immediate HappyConnOpener death (due to the Must failure) cannot hurt the requestor (or the already "done" HappyConnOpener job) in any way.

src/comm/Connection.h
yadij: The reason you gave me for the insistence on clone vs copy years ago was that copy had a loose definition that implied some details were not copied, whereas clone implied a complete and absolute duplicate of the original object.

Using "clone" methods to only duplicate parts of objects goes against that principle.

src/ResolvedPeers.h
yadij: "return" as a word by itself has multiple primary meanings that imply actions contrary to what you are intending here. IMO "retryPath" was better than this change (while still being bad, I agree on that). For lack of a single better word we may want it to be a descriptive few words instead. "resetUnusedPath()" or somewhat similar?

src/peer_select.cc
yadij: FYI: That description of the problem as being minor for me indicates that this is a TODO, rather than XXX importance.

src/HappyConnOpener.cc
yadij: again "no significant effect" == TODO,  not XXX.

src/comm/Connection.h
yadij: FTR: Please do ensure that all non-copy + non-move objects related to logged state data have a Pointer which can be copied and moved around as needed. Otherwise we end up with performance regressions for them.


@rousskov, why is move now being forbidden? move is completely safe for this object. Copy is only forbidden due to FD close races which are impossible for move semantics.

src/ResolvedPeers.cc
```diff
@@ -13,32 +13,29 @@
 #include "ResolvedPeers.h"
 #include "SquidConfig.h"
 
+/* ResolvedPeers */
+
```
yadij, line 17: Pointless addition. The namespace of every method documents the scope. Please remove.

```suggestion
```

src/ResolvedPeers.h
```diff
@@ -110,8 +114,53 @@ class ResolvedPeers: public RefCountable
...
+/// An invasive reference-counting Comm::Connection pointer that also keeps an
+/// (optional) ResolvedPeers position for the ResolvedPeers::returnPath() usage.
+/// Reference counting mechanism is compatible with Comm::ConnectionPointer.
+class PeerConnectionPointer
+{
+public:
+    using size_type = ResolvedPeers::size_type;
+
+    PeerConnectionPointer() = default;
+    PeerConnectionPointer(nullptr_t): PeerConnectionPointer() {} ///< implicit nullptr conversion
```
yadij, line 126: IMO this would be better written as an assignment operator so the RHS object does not need to go through allocate+deallocate cycles of a nil object just to perform "= nullptr;" assignments.

src/ResolvedPeers.h
eduard-bagdasaryan: We could try, for example, reAddPath() (symmetrical to the existing addPath()) or reinsertPath().

src/ResolvedPeers.h
eduard-bagdasaryan: If we add the assignment operator, we would need to add also copy/move constructor, destructor and move operator= , according to the [rule](https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)) of 5. Do you think it is worth burdening this simple class with that?

src/HappyConnOpener.cc
eduard-bagdasaryan: IHMO it would be strange to mark bugs (even rare ones) with TODOs.

src/comm/Connection.h
eduard-bagdasaryan: I do not mind renaming the 'clone' prefix to 'copy' in both these methods.

src/ResolvedPeers.cc
eduard-bagdasaryan: This cc module also has PeerConnectionPointer methods (with a corresponding separating comment). I think that If we follow this style, we should keep both comments or none.

yadij: Yes, it should not have either. This type of comment is an artifact of coding practices to document C and older language module boundaries. Those boundaries are self-documenting in modern languages like C++.

src/ResolvedPeers.cc
```diff
@@ -227,3 +225,19 @@ operator <<(std::ostream &os, const ResolvedPeers &peers)
     return os << peers.size() << (peers.destinationsFinalized ? "" : "+") << " paths";
 }
 
+/* PeerConnectionPointer */
+
```
yadij, line 229:
```suggestion
```

src/ResolvedPeers.h
yadij: I like reinsertPath()

@rousskov, do you agree on that naming or have a better one?

src/HappyConnOpener.cc
yadij: @eduard-bagdasaryan, do you mean leave this known issue unmarked? or fix it in this PR?

src/ResolvedPeers.h
yadij: That would be true for adding a default/copy/move assignment operator. They are all defined implicity anyway, otherwise the nullptr_t constructor would not work.

 But I mean an explicit nullptr_t receiving one, like so:
```suggestion
    PeerConnectionPointer &operator =(nullptr_t) { clear(); }
```

I assume that builds okay. If not then the ctor is fine. This is just a minor optimization to avoid some alloc+copy cycles.

src/HappyConnOpener.cc
eduard-bagdasaryan: No, I meant that a known/discovered bug should be better marked with XXX rather than with TODO. I do not think we should fix this unrelated problem in this PR.

src/comm/Connection.h
rousskov: > The reason you gave me for the insistence on clone vs copy years ago was that copy had a loose definition that implied some details were not copied, whereas clone implied a complete and absolute duplicate of the original object.

Bare clone() and bare copy() would indeed imply "full" replication, but we are not dealing with bare clone() and bare copy() method names here, only prefixes. The "clone" prefix is a better fit here because (as documented) we are creating a new object as opposed to filling in an existing one.

If "years ago" I said and meant something that contradicts the above, then I was wrong.

> I do not mind renaming the 'clone' prefix to 'copy' in both these methods.

I recommend using "clone", but if Amos insists on "copy" for some reason, then I am not going to fight about it.

src/ResolvedPeers.h
rousskov: I do not object to reinsertPath() if that is the best we can agree on. I do not like that reinsert implies that it was inserted before while it was actually added (somewhere), not inserted (implying addition "between" other paths).

IMO, reinstatePath() would be better.

src/peer_select.cc
rousskov: I speculate that it is minor in most environments, but it is a bug that may badly screw up some deployments. Said that, if you insist on changing this to TODO, I am not going to fight about it.

src/HappyConnOpener.cc
rousskov: I agree with Eduard -- most bugs should be marked with an XXX.

Unfortunately, it is difficult to define "bug" precisely, especially for folks that prefer literal interpretation of definitions. I tried to avoid that trap when sketching XXX/TODO differences in 9837567dd913854a4deddcc49043bfd7631ab63f by using the word "problem" for both. Clearly, that opened another attack vector!

src/ResolvedPeers.cc
rousskov: These visual "section" markings help navigate the files that contain multiple class definitions. The "inlined" difference in class names is not enough for that purpose.

Long time ago, Amos has banned repeated empty lines that served a similar section-delimiting purpose. Please leave us at least these marks! I do not insist on others adding them, but they help me and, I speculate, some other developers.

yadij: What is it delimiting though? ... that code following "/* Foo */"  are in a Foo:: namespace. One would have to be blind not to notice from the method names themselves, or the filename at least.

A comment like this has some utility delimiting sections of code which is *misplaced* and should be in another cc unit. Which is not the case for these comments AFAIK.

src/comm/Connection.h
yadij: I'm not particularly attached to either naming myself. Just wanted to point out the discontinuity with the arguments over why the original method was not allowed to be called cloneDetails().

src/ResolvedPeers.h
yadij: reinstate sounds good. I'm okay with that or reinsert. Over to you eduard to pick one.

src/peer_select.cc
yadij: Okay. Then please make the comment clarify why having a peer be tried twice composes being "a problem, often a serious one" (per our current ad-hoc policy).

I do not recall the previous behaviour of peer selection when there wold often be whole series of duplicate entries (dozens at times) being a big problem. All it did was lead to some connections being slower, and some user complaints about non-contactable sites when the origin was provably broken. IIRC, the de-dup logic was just a performance optimization not a bug fix.

src/HappyConnOpener.cc
yadij: I am judging the selection between them based on severity alone, per that ad-hoc policy description.

Eduard is saying he thinks it is serious -> XXX

Alex is saying it has "no significant effect" aka not serious at all.

What is anyone else to think?   The comment indicates serious (XXX) and code has a Must() which implies there is a handler already to cope with it cleanly (so, serious bug? or already managed?).

src/ResolvedPeers.h
rousskov: > IMO this would be better written as an assignment operator so the RHS object does not need to go through allocate+deallocate cycles of a nil object just to perform "= nullptr;" assignments.

Modern optimizing compilers do not actually create a temporary object in this context. It's all inlined. I double checked that expectation by looking at the generated assembly [code](https://godbolt.org/z/mxcBqm).

With some compilers, adding an explicit nullptr assignment operator will generate slightly different code so I cannot claim that adding that operator makes no difference at all in this case. The optimizations that a compiler can do will differ a bit. I speculate that the resulting performance difference is very small, but this analysis exceeds my assembly interpretation capabilities.

Adding a nullptr assignment operator will duplicate some code -- there is currently no duplication in this area because all "clearing" code reuses default member initialization instead of spelling out how those members should be "cleared".

If Amos wants an explicit nullptr assignment operator added after reading the above, I do not object to that addition despite the (minor) code duplication that addition will cause.

rousskov: Please note that the nullptr assignment operator should _not_ replace the existing from-nullptr constructor. The existing conversion constructor should remain because we want to continue to support both operations (i.e. construction and assignment), at least. Adding an explicit nullptr assignment operator is nothing but an optimization.

yadij: Hmm. okay since the alloc is being removed these days that clears my concern.

src/comm/Connection.h
rousskov: > why is move now being forbidden? 

The question implies that the move was allowed before this PR but is now forbidden. This PR does not change the Connection moving ban: It was forbidden before and it remains forbidden now. This PR accomplishes that in one line of code (instead of three), covers more cases at compile time (instead of linking time), and leads to better compiler error messages. Other than those improvements, there are no material changes here. 

The copy/move-banning code will become clearer when the TODO (affecting all such cases) is implemented. The same could have been said about the old code as well, of course.

> move is completely safe for this object.

I do not know what you mean exactly, but the default/generated move operations would be deadly for Connection (before and after these PR changes). This statement is true for nearly any object with a non-default destructor. In this particular case, that destructor will close the underlying connection (if it was open), leaving the moved copy with a stale `fd` value (among other problems).

> Copy is only forbidden due to FD close races which are impossible for move semantics.

Custom code is required to implement that move semantics. Connection lacks that code. While one could correctly implement a custom move constructor and move assignment operator for Connection, there is currently no good reason to do that because there is no code that would use that optimization. That may change, of course.

BTW, that custom moving code will have to overcome more obstacles than FD ownership, but that is not important here.

src/ResolvedPeers.cc
rousskov: Roughly speaking, these "section start" comments (not uncommon in Squid code) announce the start of a class-specific section in a file that has multiple classes. When I scroll through such a file, they help me notice that I crossed into another section (because they are visually different from any method start and most method descriptions). They have other, secondary benefits as well, such as highlighting classes that have only one or two small methods, which, sometimes, helps readers understand the class role in the code.

The first "section start" comment does not really delimit anything -- it announces the first section. The first section comment is primarily useful for consistency sake -- all sections in this file should have these comments -- but it is also marginally useful for warning the reader that this file has multiple sections.

This formatting trick has its cons, of course. For example, the section boundaries/names are difficult to define precisely for non-class code. Also, the section start comments duplicate information.

As most formatting conventions, it is impossible to formally prove that these comments should stay. If you insist on eliminating them for some reason, or if you are going to police every section start comment (or lack of thereof), then I am not going to fight for their preservation.

src/peer_select.cc
rousskov: > Then please make the comment clarify why having a peer be tried twice composes being "a problem, often a serious one"

You are asking me to clarify why missing a duplicate in the code that claims to "Bail on a duplicate" and computes a variable called "duplicate" is "a problem, often a serious one". Sigh.

Can we just delete this XXX instead? I do not want to waste even more time on this. If this XXX must stay to pass your review, then please suggest specific wording that would make you happy.

src/HappyConnOpener.cc
rousskov: > Eduard is saying he thinks it is serious -> XXX

Eduard did _not_ say that AFAICT. He said it is a bug. I agreed. We both think bugs should be marked with an XXX.

> Alex is saying it has "no significant effect" aka not serious at all.

I am not saying just "no significant effect". I said "no significant runtime effect (in most cases)". There are enough loopholes in that phrase to accommodate for serious effects.

Besides, the "problem, often a serious one" does not imply that all XXXs are serious problems.

If replacing the letters `XXX` with `TODO` would allow us to resolve this thread, we should replace it: The time wasted here has already negated any benefit of perfecting this comment!

src/peer_select.cc
eduard-bagdasaryan: Done at 31c8392 to facilitate the PR progress.

src/HappyConnOpener.cc
eduard-bagdasaryan: Done at 31c8392 to facilitate the PR progress.

src/ResolvedPeers.h
eduard-bagdasaryan: Done at ab76848.

src/ResolvedPeers.cc
yadij: Like I said its an antiquated coding style. Properly scoped units would only have one of these at the top saying the same thing as the filename - which makes it pointless. Exactly what we are seeing with this comment.

src/ResolvedPeers.cc
yadij: This comment has more reason to exist, since IMO it does declare the code following to be misplaced. Serving the purpose Alex cited for using these still. So I do not insist on this one being removed.

@eduard-bagdasaryan, up to you on this one.

src/ResolvedPeers.cc
eduard-bagdasaryan: Done (91a2607).

src/ResolvedPeers.cc
eduard-bagdasaryan: Left as is.

src/comm/Connection.h
yadij: The old code explicitly forbid (and documented) copy as not allowed. Leaving move up to whatever the compiler chose.

The new code says nothing about copy and forbids move. That is wrong since copy is the bad operation, move we should not care about.

src/HappyConnOpener.cc
rousskov: BTW, I have a crazy theory where the disagreement about the policy [sketch](https://github.com/squid-cache/squid/commit/9837567dd913854a4deddcc49043bfd7631ab63f) may be coming from.

> XXX: a problem, often a serious one

The word "often" in the above phrase applies to the entire set of XXXs in Squid, not every single XXX. In other words, "many XXXs mark serious problems" not "every XXX marks a problem that is often serious". Better wording to remove this alleged ambiguity is welcomed.

> a Must() which implies there is a handler already to cope with it cleanly

FYI: Squid Must()s do not carry such implications.

----

Please do not misinterpret this comment as a resurrection of this change request thread. I (still) do not object to the wrong label used for marking this bug. I am just desperately trying to avoid wasting so much time on _future_ markings.

src/comm/Connection.h
rousskov: I am sorry, but your statements contradict how C++ works. There is no "compiler choice" here now and there was not one before this change. The same things were and are prohibited/allowed.

rousskov: @eduard-bagdasaryan, to make progress in these sad circumstances, I suggest restoring the old `private` undefined method declarations and just updating their duplicated comments to reference the new `clone*()` methods names. And drop the TODO, just in case!

eduard-bagdasaryan: Done (dfc20a2).

rousskov: Thank you!

yadij:
It appears clang requires the std:: namespace prefix.
```
06:24:13 In file included from ../../../../src/FwdState.h:24:
06:24:13 ../../../../src/ResolvedPeers.h:126:27: error: unknown type name 'nullptr_t'; did you mean 'std::nullptr_t'?
06:24:13     PeerConnectionPointer(nullptr_t): PeerConnectionPointer() {} ///< implicit nullptr conversion
```

eduard-bagdasaryan:
> It appears clang requires the std:: namespace prefix

Fixed at 8d56022.

PR #681, opened 2020-06-24
Avoid FwdState::serverConn while establishing a CONNECT tunnel

rousskov:
Master/v6 commit 25b0ce4 missed two cases where serverConnection() was
used prematurely: One case requires --enable-delay-pools, and the other
does not actually dereference the prematurely used serverConnection().
Both cases establish a CONNECT tunnel through a cache_peer.


yadij:
What does "premature" mean in that description?

rousskov:
> What does "premature" mean in that description?

"before it is available/set"

PR #682, opened 2020-06-24
WIP: Upgrade ICAP protocol parser

yadij:
Use the new Parser model for message parsing based on HTTP/1
message format to perform initial parse of ICAP messages for
OPTIONS response messages.

Add (some) missing 93,2 debug tracing for ICAP traffic.

Polished Http1::StatusLine to remove some hard-coded HTTP
outputs and allow simpler more generic protocol label handling.

Add missing registration for ICAP as a protocol Squid can use.


yadij:
FTR: code changes have not yet been runtime tested to verify accuracy of the resulting internal representation for OPTIONS response messages, nor debug display of parsed objects.

yadij:
Marked as waiting for QA since ICAP testing still needed to see if there is any unexpected issues author and review cannot see. I will run it with c-icap echo service shortly, but third-party confirmation with a different service would be good.

rousskov:
I removed `S-waiting-for-QA` because, I assume, you will merge this PR when it is approved regardless of whether somebody volunteers to perform more ICAP tests than you are willing to run. That label is meant to declare PRs blocked on testing. Please restore that label if the PR is indeed blocked on those ICAP tests.

src/adaptation/icap/Launcher.h
```diff
@@ -83,18 +83,21 @@ class XactAbortInfo
...
-    HttpReply *icapReply;
-    bool isRetriable;
-    bool isRepeatable;
+    HttpRequestPointer icapRequest;
+    HttpReplyPointer icapReply;
+    bool isRetriable = false;
+    bool isRepeatable = false;
 
 private:
     XactAbortInfo &operator =(const XactAbortInfo &); // undefined
```
rousskov, line 103: A PR changing one copy constructor becomes naturally responsible for _all_ copying/moving methods. Please remove this customization as well:

```suggestion
```

src/adaptation/icap/OptXact.cc
```diff
@@ -105,17 +106,35 @@ bool Adaptation::Icap::OptXact::parseResponse()
...
     r->protoPrefix = "ICAP/"; // TODO: make an IcapReply class?
+    r->sline.set(replyParser->messageProtocol(), replyParser->messageStatus());
 
-    if (!parseHttpMsg(r.getRaw())) // throws on errors
-        return false;
+    // parse headers
+    if (!r->parseHeader(*replyParser)) {
+        r->sline.set(replyParser->messageProtocol(), Http::scInvalidHeader);
+        debugs(93, 2, "error parsing response headers mime block");
+    }
```
rousskov, line 130: AFAICT, this replacement of the parseHttpMsg() call is incorrect; it does not follow parseHttpMsg() API/invariants. I do not know whether we should keep parseHttpMsg() or not, but I do not want to write parsing code using change requests. Have you tried preserving parseHttpMsg() calls?

src/adaptation/icap/ResponseParser.cc
```diff
@@ -0,0 +1,45 @@
...
+        buf_ = tok.remaining(); // resume checkpoint
+        return parseResponseStatusAndReason(tok);
+
+    } else if (buf_.length() < Icap1magic.length() && Icap1magic.startsWith(buf_)) {
+        debugs(74, 7, Raw("valid ICAP/1 prefix", buf_.rawContent(), buf_.length()));
+        return 0;
+    }
+
+    // else, protocol error
+    assert(tok.atEnd());
```
rousskov, line 43: The overall logic here looks very generic, not specific to ICAP. Please avoid duplicating this code (if it is duplicated) or creating a protocol-specific code instead of generic one.

rousskov (changes requested):
This is a partial review. 

I do not think it is a good idea to upgrade just the OPTIONS parser. We should use the same parsing approach for all ICAP responses. I recommend switching this PR to Draft status until those changes are implemented and tested.

src/adaptation/forward.h
```diff
@@ -9,6 +9,8 @@
 #ifndef SQUID_ADAPTATION__FORWARD_H
 #define SQUID_ADAPTATION__FORWARD_H
 
+#include "adaptation/icap/forward.h"
```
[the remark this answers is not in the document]
yadij: adaptation/forward.h is supposed to be the forward declarations for all symbols provided by libadaptation.la. That library provides ICAP symbols.

src/adaptation/icap/OptXact.cc
yadij: AFAICT the parseHttpMsg() attempted to parse the buffer and chopped out the bytes which appeared to be a valid response (status-line + headers block). That is all performed by the new liens 112-113 above
```C++
    Must(replyParser->parse(readBuf)); // throws on parse failure OR incomplete data
    readBuf = replyParser->remaining(); // leave any unparsed bytes in the I/O buffer
```

This if-statement is only doing the post-processing of the mime headers already identified.

src/http/one/ResponseParser.h
```diff
@@ -50,10 +50,11 @@ class ResponseParser : public Http1::Parser
     /// \throws InsuffientInput and other exceptions on syntax and validation errors
     static void ParseResponseStatus(Tokenizer &, StatusCode &code);
 
-private:
-    int parseResponseFirstLine();
+protected:
+    virtual int parseResponseFirstLine();
```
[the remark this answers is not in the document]
yadij: The use of "first line" is intentional for several reasons:
 * the first line is not always a status-line for HTTP/1.0, 0.9, nor protocols based on older HTTP.
 * also symmetrical with the equivalent RequestParser "first line"
 * this is paired with a getter for "first line" detail provided by the generic Parser shared by all HTTP messages.

src/http/StatusLine.cc
```diff
@@ -65,11 +64,11 @@ Http::StatusLine::packInto(Packable * p) const
 
     /* local constants */
     /* AYJ: see bug 2469 - RFC2616 confirms stating 'SP characters' plural! */
-    static const char *Http1StatusLineFormat = "HTTP/%d.%d %3d %s\r\n";
+    static const char *Http1StatusLineFormat = "%s/%d.%d %3d %s\r\n";
```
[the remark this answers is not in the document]
yadij: done in fb861e5

src/adaptation/icap/ResponseParser.cc
yadij: The pattern is very similar to HTTP status-line parsing. But only the first if-statement is actually the same.

src/adaptation/icap/OptXact.cc
```diff
@@ -105,17 +106,36 @@ bool Adaptation::Icap::OptXact::parseResponse()
...
```
rousskov, line 131: I am resurrecting/continuing an out-of-date [change request](https://github.com/squid-cache/squid/pull/682/#discussion_r585753086) here to ease navigation. I resolved that change request.

> Alex: AFAICT, this replacement of the parseHttpMsg() call is incorrect; it does not follow parseHttpMsg() API/invariants. I do not know whether we should keep parseHttpMsg() or not, but I do not want to write parsing code using change requests. Have you tried preserving parseHttpMsg() calls?

> Amos: AFAICT the parseHttpMsg() attempted to parse the buffer and chopped out the bytes which appeared to be a valid response (status-line + headers block). That is all performed by the new liens 112-113 above

```C++
Must(replyParser->parse(readBuf)); // throws on parse failure OR incomplete data
readBuf = replyParser->remaining(); // leave any unparsed bytes in the I/O buffer
```

No, those lines do not seem to follow parseHttpMsg() API/invariants. For example, parseHttpMsg() has three outcomes (need-more-data, success, exception). The lines you mention only have two outcomes (success and exception). The surrounding code is unaware of this major change AFAICT, hence, is likely broken. BTW, the difference in outcomes explains why the Adaptation::Icap::OptXact::parseResponse() always returns true now, highlighting this part of the problem.

There are other potential parseHttpMsg() API/invariant changes that I have not investigated yet.

I am sure this all can be fixed, one way or another, but, again, I would very much prefer not to progress bug-by-bug here: It is very difficult and painful to write correct parsers via code review, for both the author and the reviewer. If you can largely preserve parseHttpMsg() calls, it may help a lot. The primary goals of this PR can be satisfied without significant changes to parseHttpMsg() API/invariants AFAICT. And if they cannot be, then we may have to split this work in two phases (more pain), to carefully control the switch from parseHttpMsg() to something better.

rousskov (changes requested):
Thank you for addressing some of my concerns. This partial review updates the remaining old change requests.

PR #687, opened 2020-06-29
Bug 5054: mark dns_v4_first as obsolete in cf.data.pre

DrDaveD:
The dns_v4_first directive is ignored since commit fd9c47d.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -10028,21 +10028,9 @@ DOC_START
 DOC_END
 
 NAME: dns_v4_first
-TYPE: onoff
-DEFAULT: off
-LOC: Config.dns.v4_first
+TYPE: obsolete
 DOC_START
-	With the IPv6 Internet being as fast or faster than IPv4 Internet
-	for most networks Squid prefers to contact websites over IPv6.
-
-	This option reverses the order of preference to make Squid contact
-	dual-stack websites over IPv4 first. Squid will still perform both
-	IPv6 and IPv4 DNS lookups before connecting.
-
-	WARNING:
-	  This option will restrict the situations under which IPv6
-	  connectivity is used (and tested). Hiding network problems
-	  which would otherwise be detected and warned about.
+	Remove this line. Squid performs a 'Happy Eyeballs' algorithm, doing v4 first is no longer relevant.
 DOC_END
 
 NAME: ipcache_size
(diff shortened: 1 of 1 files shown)


DrDaveD:
Ok, done.  Checking the other variables listed in the release notes as being removed directives, I see that wasn't done for `clientside_mark`, but maybe that's an exception because it belongs with acls.  `collapsed_forwarding_shared_entries_limit` is not marked as obsolete at all either, but I guess that's because it has become a silent alias for `shared_transient_entries_limit`.

I see another couple of directives `dns_v4_fallback` and `balance_on_multiple_ip` that say they were removed because of "Happy Eyeballs" as well, but they're under the "Removed in 3.2" section.  I don't understand how that happened.  Looking at the [balance_on_multiple_ip option description](http://www.squid-cache.org/Doc/config/balance_on_multiple_ip/) I also don't understand what it has to do with Happy Eyeballs.  It also claims it was in versions 2.6 through 3.5, not removed in 3.2.

yadij:
OK to test

rousskov:
@DrDaveD, just to partially explain why options related to "Happy Eyeballs" disappear at different times: Many commits, spanning more than 8 years added Happy Eyeballs-related functionality. For example, b65c1b4 dated 2012 removed dns_v4_fallback as a part of the 18b305d work.

Unfortunately, Squid lacks the scaffolding to automatically detect configuration options that are no longer backed by the code so we get these stale options floating around until a kind soul like you fixes them. Thank you.

src/cf.data.pre
```diff
@@ -181,6 +181,13 @@ DOC_START
 	This option is not yet supported by Squid-3.
 DOC_END
 
+# Options removed in 5.x
+NAME: dns_v4_first
+TYPE: obsolete
+DOC_START
+	Remove this line. Squid performs a 'Happy Eyeballs' algorithm, doing v4 first is no longer relevant.
```
rousskov, line 188: Preferring IPv4 over IPv6 addresses is still relevant (to Squid admins -- the intended audience of this documentation). We just do not have the code to enable some form of such preferential treatment. I suggest the following adjustment to avoid misleading or upsetting admins, but you may want to get Amos consent before applying this suggestion.

```suggestion
	Remove this line. Squid no longer supports preferential treatment of DNS A records.
```

@yadij, any objections to this change?

rousskov:
I adjusted PR description to be more precise/informative.

src/cf.data.pre
yadij: fine by me

yadij:
Updating branch to kick Anubis state after Jenkins restarted (previously queued staging tasks will never happen).

PR #690, opened 2020-07-05
Report SMP store queues state (mgr:store_queues)

eduard-bagdasaryan:
The state of two Store queues is reported: Transients notification queue
(a.k.a. Collapsed Forwarding queue) and SMP I/O request queue (used for
communication with rock diskers). Each worker and disker kid reports its
view of that kid's incoming and outgoing queues state, including a small
sample (up to 7 items) of queued messages. These kid-specific reports
are YAML-compliant.

With the exception of a field labeled "other", each queue report is
self-consistent despite accessing data shared among kids -- the reported
combination of values did exist at the snapshot collection time.

The special "other" field represents a message counter maintained by the
other side of the queue. In most cases, that field will be close to its
correct value, but, due to modifications by the other process of a
non-atomic value, it may be virtually anything. We think it is better to
report (without officially naming) this field because it may be useful
in triage regardless of these caveats. Making the counter atomic just
for these occasional reports is not worth the performance overheads.

Also fixed testStore linking problems (on some platforms) that were
caused by the wrong order of libdiskio and libipc dependencies.


rousskov:
FTR. Here is a sample report from one of the kids (using an earlier version of this code, but the output should be the same):

```
by kid1 {
Transients queues:
  kid1 receiving from kid1: { size: 0, capacity: 1024, other: 0, popIndex: 0 }
  kid1 receiving from kid2: { size: 0, capacity: 1024, other: 0, popIndex: 0 }
  kid1 receiving from kid3: { size: 0, capacity: 1024, other: 0, popIndex: 0 }
  kid1 receiving from kid4: { size: 0, capacity: 1024, other: 0, popIndex: 0 }

  kid1 sending to kid1: { size: 0, capacity: 1024, pushIndex: 0, other: 0 }
  kid1 sending to kid2: { size: 0, capacity: 1024, pushIndex: 0, other: 0 }
  kid1 sending to kid3: { size: 0, capacity: 1024, pushIndex: 0, other: 0 }
  kid1 sending to kid4: { size: 0, capacity: 1024, pushIndex: 0, other: 0 }

SMP disk I/O queues:
  kid1 receiving from kid5: { size: 0, capacity: 1024, other: 0, popIndex: 0 }

  kid1 sending to kid5: { size: 597, capacity: 1024, pushIndex: 597, other: 0, items: [
    { id: 1, offset: 11452416, size: 555, page: sh_page1.12210, command: w, start: 1593901516.835811, elapsed: 3.703715, errno: 0 }, # [0]
    { id: 2, offset: 101138432, size: 555, page: sh_page1.12208, command: w, start: 1593901516.847892, elapsed: 3.691634, errno: 0 }, # [1]
    { id: 3, offset: 70270976, size: 555, page: sh_page1.12206, command: w, start: 1593901516.857768, elapsed: 3.681758, errno: 0 }, # [2]
    # ... 591 items not shown ...
    { id: 595, offset: 81739776, size: 555, page: sh_page1.11042, command: w, start: 1593901520.521518, elapsed: 0.018008, errno: 0 }, # [594]
    { id: 596, offset: 11960320, size: 555, page: sh_page1.11040, command: w, start: 1593901520.526292, elapsed: 0.013234, errno: 0 }, # [595]
    { id: 597, offset: 78839808, size: 555, page: sh_page1.11037, command: w, start: 1593901520.535968, elapsed: 0.003558, errno: 0 }, # [596]
  ]}
} by kid1
```

src/store.cc
```diff
@@ -122,6 +125,18 @@ Store::Stats(StoreEntry * output)
     Root().stat(*output);
 }
 
+/// reports the current state of Store-related queues
+static void
+StatQueues(StoreEntry *e)
```
yadij, line 130: Please use Packable* as the parameter type for newly added report functions.

yadij:
@rousskov, thanks, beat my review request for the output :)

src/store.cc
rousskov: @yadij, FYI: Mgr::RegisterAction() does not support functions that accept Packable, only StoreEntry.

eduard-bagdasaryan:
Semaphore build was [failed](https://semaphoreci.com/squid/squid/branches/pull-request-690/builds/4), but the log does show any problem.

rousskov:
> Semaphore build was failed, but the log does show any problem.

To see (the signs of) the problem, you need to download the log. See the download link at the top when you scroll to the very top of the log:

```
collect2: error: ld returned 1 exit status
make[5]: *** [tests/testRock] Error 1
collect2: error: ld returned 1 exit status
make[5]: *** [tests/testUfs] Error 1
make[4]: *** [check-am] Error 2
make[3]: *** [check-recursive] Error 1
make[2]: *** [check] Error 2
make[1]: *** [check-recursive] Error 1
make: *** [distcheck] Error 1
Build Failed.
```

The full log is not very detailed. It is not yet clear (to me) whether it is Squid, Semaphore CI, or our Semaphore CI configuration fault. It may be a combination of problems. However, solving those problems, whatever they are, is outside this PR scope.

Jenkins [log](https://build.squid-cache.org/job/5-pr-test/2389/COMPILER=gcc,OS=ubuntu-focal,label=docker-build-host/console) may be more helpful:

```
09:40:59 .../src/DiskIO/IpcIo/IpcIoFile.cc:527: multiple definition of `IpcIoFile::StatQueue(std::ostream&)';
    src/tests/stub_IpcIoFile.cc:16: first defined here
```

eduard-bagdasaryan:
Here is a screenshot of cachemanager.cgi page:


cachemanager_cgi

eduard-bagdasaryan:
@yadij , it looks like both your requests were addressed.

src/store.cc
```diff
@@ -1285,6 +1304,7 @@ storeRegisterWithCacheManager(void)
     Mgr::RegisterAction("store_io", "Store IO Interface Stats", &Mgr::StoreIoAction::Create, 0, 1);
     Mgr::RegisterAction("store_check_cachable_stats", "storeCheckCachable() Stats",
                         storeCheckCachableStats, 0, 1);
+    Mgr::RegisterAction("store_queues", "SMP Transients and Caching Queues", StatQueues, 0, 1);
```
yadij, line 1307: Ideally this would be using an Action class to manage the report. See Mgr::StoreIoAction for example.

That would allow cleaner separation of the report logic into src/store/ or src/mgr/ and setting of the report Content-Type to "text/yaml" correctly.

src/ipc/Queue.h
```diff
@@ -410,6 +425,92 @@ OneToOneUniQueue::push(const Value &value, QueueReader *const reader)
...
+    for (uint32_t i = 0; i < n; ++i) {
+        // XXX: Throughout this C++ header, these overflow wrapping tricks work
+        // only because theCapacity currently happens to be a power of 2 (e.g.,
+        // the highest offset (0xF...FFF) % 3 is 0 and so is the next offset).
+        const auto pos = (offset++ % theCapacity) * theMaxItemSize;
+        Value value;
+        memcpy(&value, theBuffer + pos, sizeof(value));
+        os << "    { ";
+        value.stat(os);
+        os << " }, # [" << skippedCount + i << "]\n"; // item's queue position
```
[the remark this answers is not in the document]
rousskov: Let's drop this YAML comment to avoid wasting time on review. The value is helpful for (some) humans, but it can be computed from the already posted data if it is desperately needed. IIRC, you should be able to simplify the method profile if you remove this YAML comment.

> reserved [] characters

FWIW, YAML [allows](https://yaml.org/spec/1.2/spec.html#id2780069) those characters inside YAML comments.

src/ipc/Queue.h
```diff
@@ -410,6 +425,92 @@ OneToOneUniQueue::push(const Value &value, QueueReader *const reader)
...
+        const auto secondSampleOffset = std::max(sampleSize, count - sampleSize);
+        const auto secondSampleSize = std::min(sampleSize, count - sampleSize);
+
+        // but first we print a sample separator, unless there are no items
+        // between the samples or the separator hides the only unsampled item
+        const auto bothSamples = sampleSize + secondSampleSize;
+        if (bothSamples + 1U == count)
+            statRange<Value>(os, start, sampleSize, 1);
+        else if (count > bothSamples)
+            os << "    # ... " << (count - bothSamples) << " items not shown ...\n";
```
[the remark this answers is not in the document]
rousskov: Samples are useful when diagnosing some problems. For example, they can tell humans whether the whole queue got stuck for a long time, whether two kids have a compatible/expected view of the queue that links them, and whether some queues are overflown while showing messages that Squid already reported as timed out.

src/store.cc
eduard-bagdasaryan: This Mgr::StoreIoAction (and other similar Actions) is a kind of 'aggregatable' Action. For example, such actions redefine their virtual collect() for this aggregation. On the other hand, StatQueues is not aggregatable and that is why we should use the existing FunAction for it (which is selected in our case).

rousskov: > See Mgr::StoreIoAction for example

I would not use that class as an example. Like most class-based Actions, StoreIoAction is aggregatable() while the new report is not (and should not be). If we somehow decide that duplicating code is a good idea, then we should duplicate Mgr::FunAction instead, calling the reporting method instead of the OBJH handler.

> That would allow cleaner separation of the report logic into src/store/ or src/mgr/

AFAICT, the separation of Queue reporting methods from the Queue classes has nothing to do with the reporting methods (not) being Action class methods: We can put Queue reporting functions into src/store/ or src/mgr/, without making them Action class methods.

There is at least one serious reason to reject moving Queue reporting functions away from Queue: This PR places Queue reporting together with Queue implementation to allow the intricate reporting code to safely access sensitive Queue data members (in shared memory) without the overheads/risks of exposing those private members. Having that safe/convenient access is especially important here because there is not just one Queue class but several layers of classes that the reporting code has to navigate. @yadij, what different placement principle do you suggest that would overwrite the current "convenience/safety of access" placement principle?

> setting of the report Content-Type to "text/yaml" correctly

That would be a mistake: While individual kid reports are valid YAML, the response itself is not. We need to change how Squid wraps/separates individual kid responses. After that, we can start using text/yaml for valid YAML responses. That global enhancement is not trivial and is outside this project scope. IMO, it should be done when adding proper YAML generation support -- a stand-alone project!

@yadij, do the above arguments against using an Action class address you change request?

src/tests/stub_IpcIoFile.cc
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (C) 1996-2020 The Squid Software Foundation and contributors
+ *
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#include "squid.h"
+
+#if USE_DISKIO
```
rousskov, line 11: Can we use the more specific `HAVE_DISKIO_MODULE_IPCIO` here?

rousskov:
I polished PR description to emphasize that YAML compliance applies to kid-specific reports, to document the surprising Makefile.am change (branch commit d3dcff9), and improve layout. Please adjust further as needed.

src/ipc/Queue.h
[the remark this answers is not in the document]
eduard-bagdasaryan: Done (d07fe48).

src/tests/stub_IpcIoFile.cc
eduard-bagdasaryan: Done (e382f47).

eduard-bagdasaryan: Done (af443a7).

src/ipc/Queue.h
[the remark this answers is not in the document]
yadij: So this is a debug output, not a management report ?

src/store.cc
yadij: My understanding of Action is that we are trying to move the cachemgr API over to Actions and FunAction is a wrapper for legacy C code.

Yes StoreIoAction may not have been the best example. This new report would initially only need to override Action::dump() with the logic currently inside the StatQueues() function being registered.

If the current output is not validating as YAML properly it should be made to do so. Using standard formats properly instead of in-house custom plain-text is the "good design practice" I am aiming at here. I don't particularly care *which* standard format right now since we do not have a formal policy on that.

eduard-bagdasaryan: > This new report would initially only need to override Action::dump()

I am afraid that this is insufficient because our action is not aggregatable, i.e., it sends the response directly to HTTP client. In FunAction, this logic is implemented in Mgr::FunAction::respond(), where the work is delegated to Mgr::Filler() and Mgr::StoreToCommWriter(). I am not sure that it is a good idea to copy and paste this logic into a new hypothetical class. We could probably refactor this area, implementing a new base class for non-aggregatable actions so that the descendants would only need overriding their 'dump()' as you noted. But I think this refactoring is out of scope - it should be done as a separate work.

src/ipc/Queue.h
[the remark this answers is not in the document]
eduard-bagdasaryan: Not sure what does the 'management report' mean in this context, but IIRC there were there options:
1. Show no samples at all.
2. Show all samples.
3. Show only a few samples.

We selected (3) as a compromise.

src/store.cc
rousskov: > My understanding of Action is that we are trying to move the cachemgr API over to Actions and FunAction is a wrapper for legacy C code

The above understanding does not reflect the current reality well. FunAction is a wrapper for non-aggregating function-initiated reports. Yes, FunAction is used for legacy reports that were first implemented using C functions, but bare functions are not second class citizens in C++. If something can be equally well implemented using a function and a class, the former should be preferred because function is a simpler concept. There is nothing wrong with using functions in modern C++, of course! This PR is using a simple StatQueues() function to jump-start queue reports. I see no problems with that solution. Creating a whole new class just to wrap those eight lines of code would make things worse, not better!

The existing function-vs-class separation between FunAction and aggregating Actions is a bit of a red herring: I am pretty sure that some of those aggregating Actions should have been implemented using a simple function as well (with a common Action-based class that does the rest). There is some code duplication among those aggregating Actions. The initial refactoring solved many legacy problems and added SMP support, but we probably missed some good opportunities. Such defects are common for complex cleanup tasks, especially when they do not go through proper reviews. Things are clearer now, and we may do another refactoring pass eventually.

> If the current output is not validating as YAML properly it should be made to do so.

Yes, but not here/now. None of the non-aggregating reports are validating as YAML today (because of how kid-specific reports are separated/wrapped). Fixing that old problem is outside this project scope because the fix affects _existing_ reports (and automation that uses them) and will benefit from being combined with other YAML-related improvements. What is in scope for this (and any similar future) PR is generating YAML-compliant per-kid output. This PR does that.

rousskov: > I don't particularly care which standard format right now since we do not have a formal policy on that.

This is an out of scope discussion, but my review of the old squid-dev threads and improved understanding of YAML shows that we have only two viable options left:

1. YAML (with formatting manipulators ensuring the right syntax).
2. Configurable output format (including YAML) based on the generic formatting manipulators [ideas](http://www.squid-cache.org/mail-archive/squid-dev/201404/0009.html).

Thus, until we have the right formatting manipulators, new aggregated reports and new kid-specific parts of non-aggregated reports should use YAML. This PR follows that principle.

----

* http://lists.squid-cache.org/pipermail/squid-dev/2015-August/002953.html
* http://www.squid-cache.org/mail-archive/squid-dev/201404/0009.html
* http://www.squid-cache.org/mail-archive/squid-dev/201110/0309.html
* http://www.squid-cache.org/mail-archive/squid-dev/201101/0146.html

src/ipc/Queue.h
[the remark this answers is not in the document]
rousskov: > So this is a debug output, not a management report ?

I cannot answer this question because I do not know what you mean by "debug output" and "management report". You know what "this" is because you see the code and the [output](https://github.com/squid-cache/squid/pull/690#issuecomment-653958250) it produces, so I suspect this is some kind of an indirect question designed to expose some perceived flaw. Please consider rephrasing the question in a more direct/specific way.

FWIW, I think there is nothing wrong with adding samples to cache manager reports. I personally find them useful and sometimes even insightful when looking at these reports during bug triage. Moreover, showing thousands of entries (by default) has been proven to be a _bad_ idea because generating the corresponding report usually blocks Squid for a long time, overflows buffers, and leads to other problems/[bugs](https://github.com/squid-cache/squid/pull/536). If somebody wants a full queue snapshot, they can enhance this code to produce one when explicitly requested via a cache manager request parameter.

src/store.cc
yadij: @rousskov, I prefer to use YAML yes - due to its relatively human-readable nature. But reports are not restricted to that. Any class inheriting from Action can override contentType() to deliver other standardized formats.

@eduard-bagdasaryan, the class Action API has two sections - collate() + respond() are part of the per-worker API, and dump() + contentType() which I am suggesting be used are for non-aggregation reports like this one.

src/ipc/Queue.h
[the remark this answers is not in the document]
yadij: I agree there is nothing wrong with reporting this data. Just trying to get an understanding of the use-cases here.

FWIW; I meant "debug output" for reports only used when debugging or troubleshooting a problem. Which would naturally have a much, much lower frequency of use than reports used routinely to see what the traffic is doing (eg info report) - that later being the "management report"'s.

Since this appears to be in the former debug group I wondered why the need to compromise and show so few entries. That is an output more suited to summaries for management reports. Debug would seem to be more useful getting the full data.  From your responses I understand that only the beginning and ending of the queue is useful for the debug.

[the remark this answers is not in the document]
rousskov: Thank you for resolving this change request. While I disagree with your analysis and do not recommend using the proposed terminology, I do not think we need to reach consensus on the anticipated reporting frequency or the terminology used to describe the purpose(s) of this cache manager report. We can revisit the terminology issue later if such a need arises, of course.

src/store.cc
rousskov: > FunAction restricts the report to text/plain 

which is what the new report is. Again, the new report (as a whole) is not YAML and cannot be YAML (within this PR scope). Future refactoring can make function-initiated reports, like this one, to produce YAML _without_ converting them into classes.

> and require a refactoring when the class is made to send a non-custom format.

That is a speculation, not a fact. I can easily imagine scenarios where no such refactoring would be necessary or, to be more precise, where the amount of refactoring would be the same (or smaller) if this report is initiated by a function rather than a class. Again, you are basing your suggestion on a false premise that a class would be better than the function here. In this context, the reverse is true. A function is the simplest and overall best solution here.

> This being a brand new report (debug feature) I would like to see it polished up to fully before being added.

Your request is not a polishing request. The changes you want will make the code _worse_, not better. Adding a class just to convert a function into a method would result in more code duplication, increased complexity, more special cases, more ink, etc. All the given reasons for adding a new class were either factually wrong (e.g., "the new class will return responses with YAML content type") or unsubstantiated speculation where other experts see a different picture (i.e. "it will require less refactoring in the future"). If I missed any valid reason where a class would work better than the function, please let me know.

> the class Action API has two sections - collate() + respond() are part of the per-worker API, and dump() + contentType() [...] for non-aggregation reports like this one.

There are no such "sections" in the Action API. The collect() method is not paired with respond(), contentType() is not paired with dump(), respond() is used for all reports, etc. The API is fairly complex because it needs to solve several non-trivial problems for two partially overlapping categories of reports in SMP and non-SMP environments. I am sure it can be improved as well -- as discussed earlier, it probably mistreats function-based report initiators (an API mistake that probably contributes to misunderstandings like the ones expressed on this change request thread).

However, all of that complexity should not really matter here. A simple initiation function is the best (and already implemented!) solution for this report. It is simple, it follows a common well-known pattern, it produces the best result possible today, it does not duplicate code, it is unlikely to increase future refactoring costs. Converting it into a new class will have the opposite effect on most of these categories.

PR #691, opened 2020-07-16
Detail client closures of CONNECT tunnels during TLS handshake

chtsanti:
... and improve detailing of other errors.

Many admins cannot triage TLS client failures, and even Squid developers
often cannot diagnose TLS problems without requiring detailed debugging
logs of failing transactions. The problem is especially bad for busy
proxies where debugging individual transactions is often impractical.

We enhance existing error detailing code so that more information is
logged via the existing %err_code/%err_detail logformat codes.
Propagating low-level error details required significant enhancements
and refactoring. We also built initial scaffolding for better error
detailing by GnuTLS-driven code and documented several key
error-handling APIs, exposing a few out-of-scope problems.

Also checkLogging() once, after consuming unparsed input attributed to a
transaction: Due to fake CONNECT requests, from-client read errors, and
possibly other complications, we may have a transaction that did not
consume every input byte available to it. That transaction is still
responsible for reporting those unparsed bytes (e.g., by logging the
number of bytes read on a connection and the number of parsed bytes).

Also fixed passing wrong (errno vs. size) or stale (requested vs. read)
I/O size to connFinishedWithConn(); now shouldCloseOnEof(). The bad
value was "correct" (i.e. zero) in many cases, obscuring the bug.


chtsanti:
Below is a sample of `%err_code/` and `%err_detail` records:

* Squid timed out while peeking at (expected) TLS client Hello:

       ERR_REQUEST_START_TIMEOUT/-

* Squid timed out while peeking at (partial) TLS client Hello:

       ERR_REQUEST_PARSE_TIMEOUT/-

* Squid got a non-TLS handshake while peeking at TLS client Hello:

       ERR_PROTOCOL_UNKNOWN/EXCEPTION=18686e4e

   To retrieve more details about this error someone can use the squid utility 'calc-must-ids.sh':

       # calc-must-ids.sh 18686e4e
       ./src/security/Handshake.cc:158: 0x18686E4E Must(type >= ctChangeCipherSpec && type <= ctApplicationData);

* The client closed its TCP connection while Squid was accepting
  a TLS connection on https_port:

       ERR_SECURE_ACCEPT_FAIL/SQUID_TLS_ERR_ACCEPT+TLS_IO_ERR=5

* Squid+OpenSSL TLS negotiation failure with a TLS client:

       ERR_SECURE_ACCEPT_FAIL/SQUID_TLS_ERR_ACCEPT+TLS_LIB_ERR=1408F09B+TLS_IO_ERR=1
    Someone can retrieve more details about the OpenSSL library error using 'openssl strerr':

       # openssl errstr 1408F09B
       error:1408F09B:SSL routines:ssl3_get_record:https proxy request

* Squid+GnuTLS TLS negotiation failure with a TLS client:

       ERR_SECURE_ACCEPT_FAIL/SQUID_TLS_ERR_ACCEPT+GNUTLS_E_UNEXPECTED_PACKET

* Squid+OpenSSL TLS negotiation failure with a TLS server

       ERR_SECURE_CONNECT_FAIL/
       SQUID_ERR_SSL_HANDSHAKE+TLS_LIB_ERR=1408F10B+TLS_IO_ERR=1
       X509_V_ERR_CERT_HAS_EXPIRED+broken_cert

* Squid+GnuTLS TLS negotiation with a non-TLS server:

       ERR_SECURE_CONNECT_FAIL/SQUID_TLS_ERR_CONNECT+GNUTLS_E_UNEXPECTED_PACKET

rousskov:
Here is my TODO/XXX classification for this PR. I compiled it for a slightly earlier version so there might be some minor discrepancies with the current code.

To be decided at the beginning of the review process after gathering feedback from others:

* `return ToSBuf("SYSERR=", errorNo); // TODO: Use errno=... instead?`
* `return ToSBuf("EXCEPTION=", std::hex, exceptionId); // TODO: Use exception=... instead?`
* `TODO: Either move to Security::Io or remove/restrict the Io namespace.`

To be done at the end of the review process (to minimize diff noise):

* `TODO: Remove this file before merging this PR.`
* `TODO: Remove this file before merging this PR.`

Out of scope:

* `TODO: No ConnStateData::tunnelOnError() on this forward-proxy code`
* `TLS handshakes on non-bumping https_port. TODO: Discover these`
* `TODO: throw when nonBlockingCheck() callbacks gain job protections`
* `TODO: Remove. Make sure there is always a suitable ALE instead.`
* `TODO: Support compilation of unterminated/SBuf bodies.`
* `TODO: Replace type, xerrno and detail with Error while adding a virtual`
* `TODO: Replace high-level ERR_get_error() calls with a new std::ostream`
* `TODO: After dropping OpenSSL v1.1.0 support, this and Security::Connect() can`
* `TODO: Honor result.important when working in a reverse proxy role?`
* `SBuf buf(err_code()); // TODO: Upgrade err_code() to return SBuf.`
* `TODO: Convert this and sslErrorName() in Format.cc to TLS_ERR=<HEX>.`
* `TODO: Save/detail all stacked errors by always flushing stale ones.`
* `TODO: Avoid gnutls_*() calls if debugging is off.`
* `TODO: Avoid gnutls_*() calls if debugging is off.`

yadij:
I also would prefer the lower case key names. Will need to review to have an opinion on the namespace.

rousskov:
> Amos: I also would prefer the lower case key names.

Changed in commits 6cfdb21 and 57efccb. Please note that the official code uses `ALLCAPS` for similar %err_detail record fields (and a different spelling for errno -- `SYSERR`). We are already changing %err_detail quite a bit (by supporting a list of error details or, to be more accurate, compounded details) so perhaps we have enough ammunition to justify this change. However, I remain unsure about it.

> Will need to review to have an opinion on the namespace.

@yadij, If I do not hear otherwise from you within a day or two, I will remove the existing/official `Io` namespace. I am [still](https://github.com/squid-cache/squid/pull/691#issuecomment-659513148) pretty sure it is the right thing to do long-term, and this important PR has been waiting for this decision for more than two weeks already.

src/tests/stub_libsecurity.cc
```diff
@@ -31,12 +31,30 @@ std::ostream &Security::operator <<(std::ostream &os, const Security::EncryptorA
...
+
 #include "security/KeyData.h"
 namespace Security
 {
 void KeyData::loadFromFiles(const AnyP::PortCfg &, const char *) STUB
 }
 
+#include "security/ErrorDetail.h"
+Security::ErrorDetail::ErrorDetail(ErrorCode, const CertPointer &, const CertPointer &, const char *) STUB
+#if USE_OPENSSL
```
yadij, line 47: The library specific #if cases should never be exposed externally by libsecurity.

This one should be able to be removed by OpenSSL code defining LibErrorCode properly for the generic API users.

src/error/Detail.h
```diff
@@ -0,0 +1,91 @@
...
+    /// suitable for error pages and other output meant for human consumption
+    virtual SBuf verbose(const HttpRequestPointer &) const = 0;
+};
+
+/// creates a new NamedErrorDetail object with a unique name
+/// \see NamedErrorDetail::Name for naming restrictions
+ErrorDetail::Pointer MakeNamedErrorDetail(const char *name);
+
+/// system call failure detail based on standard errno(3)/strerror(3) APIs
+class SysErrorDetail: public ErrorDetail {
```
yadij, line 41: Please split all these specialized classes (SysErrorDetail, ExceptionErrorDetail, NamedErrorDetail) into separate files named after the class contained. Since all this code is in a convenience library there is no need to bunch them together like this.

Please also inline the one-liner methods wherever possible.

src/Makefile.am
```diff
@@ -543,6 +563,7 @@ squid_LDADD = \
 	DiskIO/libdiskio.la \
 	comm/libcomm.la \
 	anyp/libanyp.la \
+	error/liberror.la \
```
yadij, line 566: libsecurity and SSL_LIBS content depends on liberror yes? That means liberror should be listed later.

src/Makefile.am
```diff
@@ -31,8 +31,30 @@ LOADABLE_MODULES_SOURCES = \
 	LoadableModules.cc \
 	LoadableModules.h
 
-SUBDIRS	= mem base anyp helper dns ftp parser comm eui acl format clients sbuf servers fs repl store DiskIO proxyp
-DIST_SUBDIRS = mem base anyp helper dns ftp parser comm eui acl format clients sbuf servers fs repl store DiskIO proxyp
+SRC_DIRS = \
```
yadij, line 34: Please undo this variable addition. The whole sequence does need cleaning up. But not as part of this PR.

rousskov:
> Alex: I suggest removing the Io namespace because ...
> Alex: If ..., I will remove the existing/official Io namespace.
> Amos: Regarding the Io namespace, I agree it is good here in context of the full design.

FYI: I am interpreting the above as "agree to remove" rather than "keep a namespace that is good here" and will proceed with removing that namespace as [planned](https://github.com/squid-cache/squid/pull/691#issuecomment-668819747).

> Amos: For now I have skipped requesting C++11 initialization of new/changed classes where there appears to only be an explicit constructor initializing to a caller-defined value. What do you think about making C++11 initialization a policy requirement even in those classes?

I think such a policy would be overall harmful and, in rare cases, impractical (and even impossible) to support.

In general: Where multiple ways of writing code are supported by the language, we should not require exclusive use of one specific way just because it is "new". All features, especially the new ones, are usually usable in many cases but are designed for a specific _subset_ -- they solve a particular set of problems. Requiring the use of such a feature in _all_ usable cases is essentially requiring misuse.

As far as non-static data member initialization is concerned: The natural/minimal policy here goes something like "all members must be initialized during object construction". To reduce the number of initialization-related bugs, I recommend an _additional_ restriction:

* Do not initialize class members using assignment operator in a constructor.

The combination of the natural/minimal policy and the above restriction means that for every object creation way, each non-static data member must be initialized either via the default member initializer or via the member initializer list of the constructor.

FWIW, my restriction is a variant of the C++ [Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-initialize) "prefer initialization to assignment" recommendation (if you interpret the latter correctly).

Further blanket regulation like "always use default member initializer" would harm code clarity (and cannot be applied to some cases for technical reasons). There are cases where all members should be initialized via default initializers and cases where none of them should be. There are also cases where the best design calls for a mixture of initialization approaches. If I am forced to come up with any additional blanket regulation, I can only suggest something obvious(?) like

* All class constructors should use internally consistent approach to member initialization.

... but even that vague recommendation can be abused to cripple good designs. I can imagine a class having two sets of constructors, each internally consistent, but substantially different across the sets (for some very valid/good design reason).

> The benefit I see is only for debugging objects which have issues during construction or implicit copy.

I am not quite sure what benefits you see and/or what context you are talking about, but AFAICT, default member initializers make debugging a bit _harder_ because the debugger cannot easily step through the explicit constructor source code when stepping through object initialization instructions. That small problem is not the reason to shy away from them, of course. They are very useful in many use cases.

src/LogTags.h
```diff
@@ -11,6 +11,20 @@
...
 #include "CollapsingHistory.h"
 
+/// Things that may happen to a transaction while it is being
+/// processed according to its LOG_* category. Logged as _SUFFIX(es).
+/// Unlike LOG_* categories, these flags may not be mutually exclusive.
+class LogTagsErrors {
+public:
+    LogTagsErrors(): ignored(false), timedout(false), aborted(false) {}
+
+    void update(const LogTagsErrors &other);
```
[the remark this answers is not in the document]
chtsanti: commit 1fe36470

src/LogTags.h
```diff
@@ -11,6 +11,20 @@
 
 #include "CollapsingHistory.h"
 
+/// Things that may happen to a transaction while it is being
+/// processed according to its LOG_* category. Logged as _SUFFIX(es).
+/// Unlike LOG_* categories, these flags may not be mutually exclusive.
+class LogTagsErrors {
+public:
+    LogTagsErrors(): ignored(false), timedout(false), aborted(false) {}
```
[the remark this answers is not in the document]
chtsanti: commit 1fe36470

src/LogTags.h
```diff
@@ -11,6 +11,20 @@
 
 #include "CollapsingHistory.h"
 
+/// Things that may happen to a transaction while it is being
+/// processed according to its LOG_* category. Logged as _SUFFIX(es).
+/// Unlike LOG_* categories, these flags may not be mutually exclusive.
+class LogTagsErrors {
```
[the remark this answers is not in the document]
chtsanti: commit 8bf5ebb

src/base/forward.h
```diff
@@ -15,6 +15,8 @@ class CallDialer;
 class CodeContext;
 class ScopedId;
 
+template<class C> class RefCount;
+
```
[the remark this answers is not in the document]
chtsanti: commit 82728dd087

src/error/Detail.cc
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+#include "squid.h"
+#include "error/Detail.h"
+#include "HttpRequest.h"
+#include "sbuf/SBuf.h"
+#include "sbuf/Stream.h"
+
+/// details an error by tying it to a uniquely named circumstance
+class NamedErrorDetail: public ErrorDetail {
```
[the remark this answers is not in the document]
chtsanti: commit cb67f9b58

src/LogTags.h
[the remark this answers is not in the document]
yadij: With the constructor being defaulted now you can remove its definition entirely and let the compiler use the defaults. If you are adding it here to disable implicit definition of other operators then it is better to explicitly "=delete" those operators.

src/error/Detail.cc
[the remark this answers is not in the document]
yadij: The other child classes are not yet fixed as of cb67f9b

yadij:
> > Alex: I suggest removing the Io namespace because ...
> > Alex: If ..., I will remove the existing/official Io namespace.
> > Amos: Regarding the Io namespace, I agree it is good here in context of the full design.
> 
> FYI: I am interpreting the above as "agree to remove" 

Yes, correct.


[ policy discussion below, largely unrelated to this PR ]

> 
> > Amos: For now I have skipped requesting C++11 initialization of new/changed classes where there appears to only be an explicit constructor initializing to a caller-defined value. What do you think about making C++11 initialization a policy requirement even in those classes?
> 
> I think such a policy would be overall harmful and, in rare cases, impractical (and even impossible) to support.

Okay, we sort-of align there. My view differs in reasoning, but the outcome is same.

> 
> In general: Where multiple ways of writing code are supported by the language, we should not require exclusive use of one specific way just because it is "new". All features, especially the new ones, are usually usable in many cases but are designed for a specific _subset_ -- they solve a particular set of problems. Requiring the use of such a feature in _all_ usable cases is essentially requiring misuse.
> 
> As far as non-static data member initialization is concerned: The natural/minimal policy here goes something like "all members must be initialized during object construction". To reduce the number of initialization-related bugs, I recommend an _additional_ restriction:
> 
>     * Do not initialize class members using assignment operator in a constructor.
> 
> 
> The combination of the natural/minimal policy and the above restriction means that for every object creation way, each non-static data member must be initialized either via the default member initializer or via the member initializer list of the constructor.

I think "do not" is going too far. Some members need complicated initialization logic. That is sometimes suitable for placing in the constructor body.

> 
> FWIW, my restriction is a variant of the C++ [Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-initialize) "prefer initialization to assignment" recommendation (if you interpret the latter correctly).
> 

Ack. "prefer" is good. It allows for the complex cases while eliminating trivial assignments.


> Further blanket regulation like "always use default member initializer" would harm code clarity (and cannot be applied to some cases for technical reasons). There are cases where all members should be initialized via default initializers and cases where none of them should be. There are also cases where the best design calls for a mixture of initialization approaches. If I am forced to come up with any additional blanket regulation, I can only suggest something obvious(?) like
> 
>     * All class constructors should use internally consistent approach to member initialization.
> 

Since C++11 initialization was added to the language this policy creates unnecessary verbosity in code. The policy I would like to use is:
 - trivial initialization to a constant should use default member-assignment in the class definition. This is the basic value all ctor can rely on existing when they are called.
 - non-constant values should be set in ctor initializer list, usually from parameters.
 - complex logic should be in the ctor body (if cannot be avoided entirely).

IME these cover all cases for ctor design while allowing some leeway for QA to determine the definition of "complex" for the third item on a case-by-case basis.


> 
> ... but even that vague recommendation can be abused to cripple good designs. I can imagine a class having two sets of constructors, each internally consistent, but substantially different across the sets (for some very valid/good design reason).
> 
> > The benefit I see is only for debugging objects which have issues during construction or implicit copy.
> 
> I am not quite sure what benefits you see and/or what context you are talking about, but AFAICT, default member initializers make debugging a bit _harder_ because the debugger cannot easily step through the explicit constructor source code when stepping through object initialization instructions. That small problem is not the reason to shy away from them, of course. They are very useful in many use cases.

The initializers which cannot be stepped through are not part of the custom ctor logic sequence being debugged. They can/should be treated as constant values set in memory before the constructor is called. The benefit is not having to step through a period where they are uninitialized/random values that could screw with the logic behaviour.

rousskov:
> Alex: Do not initialize class members using assignment operator in a constructor.

> I think "do not" is going too far. Some members need complicated initialization logic. That is sometimes suitable for placing in the constructor body.

Given the number of initialization bugs, I am not sure we want to be vague or permissive here, but I do not insist on any specific wording at this point.

Needless to say, complicated initialization logic can and often should be encapsulated using a function that returns the computed initialization value. Or there is a suitable nil-like initialization value that the (complicated) assignment logic can overwrite later. The remaining rare exceptions are already covered by [doc2](https://github.com/measurement-factory/squid-notes/blob/start/README.md#doc2-there-are-no-absolute-principles) -- "There are no absolute principles" -- principle as far as I am concerned.
 
> The policy I would like to use is:
> 
> * trivial initialization to a constant should use default member-assignment in the class definition. This is the basic value all ctor can rely on existing when they are called. 
> * non-constant values should be set in ctor initializer list, usually from parameters. 
> * complex logic should be in the ctor body (if cannot be avoided entirely).

Beyond C++ restrictions on default member initializers, initial value constness should not be driving these design decisions IMO. Designing constructors is usually driven by far more important considerations here than the ephemeral difference between `5` and `SomeDefaultLevel` (for the lack of a better example). The interesting cases are usually about providing a consistent/natural/intuitive set of initializations _across_ class constructors that minimizes mistakes and helps with maintenance.

> * complex logic should be in the ctor body (if cannot be avoided entirely).

Having no initialization until constructor body often means the variable is left uninitialized for some time while that constructor body code is running. Not good, regardless of how complex that belated initialization is.

chtsanti:
> > Amos: For now I have skipped requesting C++11 initialization of new/changed classes where there appears to only be an explicit constructor initializing to a caller-defined value. What do you think about making C++11 initialization a policy requirement even in those classes?
> 
> I think such a policy would be overall harmful and, in rare cases, impractical (and even impossible) to support.

The c++11 in-class initialization can not be used when you need to define pod or trivial classes for example.

> As far as non-static data member initialization is concerned: The natural/minimal policy here goes something like "all members must be initialized during object construction". To reduce the number of initialization-related bugs, I recommend an _additional_ restriction:
> 
>     * Do not initialize class members using assignment operator in a constructor.
> 

I am against this rule too, because maybe there are reasons for initializing members using assignment operator in a constructor for example simple "if" conditions.

rousskov:
> The c++11 in-class initialization can not be used when you need to define pod or trivial classes for example.

Yes, C++11 has several use cases that would be "impossible" to support with default member initializers.

> there are reasons for initializing members using assignment operator in a constructor for example simple "if" conditions.

Simple `if` conditions is a bad example: We have tertiary `?'` operator for that. For more complex cases, there are functions (returning the initialization value).

src/LogTags.h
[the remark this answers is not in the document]
chtsanti: commit 63e5b5365e

> If you are adding it here to disable implicit definition of other operators then it is better to explicitly "=delete" those operators.

I did not add it for this reason and I am not seeing any reason for declaring as delete some operations.

src/Makefile.am
```diff
@@ -3071,6 +3094,7 @@ tests_testEvent_LDADD = \
 	ip/libip.la \
 	fs/libfs.la \
 	anyp/libanyp.la \
+	error/liberror.la \
```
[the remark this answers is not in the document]
chtsanti: commit 2c8da03355

src/err_detail_type.h
```diff
@@ -9,46 +9,9 @@
...
-
-    if (errDetailId < ERR_DETAIL_MAX)
-        return err_detail_type_str[errDetailId-ERR_DETAIL_START+2];
-
-    if (errDetailId >=ERR_DETAIL_EXCEPTION_START)
-        return "EXCEPTION";
-
-    return "UNKNOWN";
-}
+// TODO: Remove this file before merging this PR.
```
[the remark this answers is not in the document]
chtsanti: commit 76a0408cd
Probably more work is required in the mk-string-arrays.awk file.

src/err_type.h
```diff
@@ -9,102 +9,9 @@
...
-
-inline
-const char *
-errorTypeName(err_type err)
-{
-    if (err < ERR_NONE || err >= ERR_MAX)
-        return "UNKNOWN";
-    return err_type_str[err];
-}
+// TODO: Remove this file before merging this PR.
```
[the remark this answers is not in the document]
chtsanti: commit 76a0408cd

src/error/Makefile.am
```diff
@@ -0,0 +1,32 @@
...
+noinst_LTLIBRARIES = liberror.la
+
+liberror_la_SOURCES = \
+	forward.h \
+	Error.cc \
+	Error.h \
+	Detail.cc \
+	Detail.h \
+	\
+	$(BUILT_SOURCES)
```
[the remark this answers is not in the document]
chtsanti: commit 765e1c0487

src/error/Makefile.am
```diff
@@ -0,0 +1,32 @@
...
+
+CLEANFILES += \
+	$(BUILT_SOURCES)
+
+nodist_liberror_la_SOURCES = \
```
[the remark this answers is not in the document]
chtsanti: commit 765e1c0487

src/Makefile.am
chtsanti: commit 0b25e37359

src/Makefile.am
chtsanti: commit 0575ca35a9

src/error/Makefile.am
```diff
@@ -0,0 +1,31 @@
...
+## Squid software is distributed under GPLv2+ license and includes
+## contributions from numerous individuals and organizations.
+## Please see the COPYING and CONTRIBUTORS files for details.
+##
+
+include $(top_srcdir)/src/Common.am
+include $(top_srcdir)/src/TestHeaders.am
+
+categories.cc: forward.h $(top_srcdir)/src/mk-string-arrays.awk
+	$(AWK) -f $(top_srcdir)/src/mk-string-arrays.awk < $(srcdir)/forward.h > $@ || ($(RM) -f $@ && exit 1)
```
yadij, line 12:
```suggestion
	$(AWK) -f $(top_srcdir)/src/mk-string-arrays.awk ipath=error/ < $(srcdir)/forward.h > $@ || ($(RM) -f $@ && exit 1)
```

src/mk-string-arrays.awk
```diff
@@ -73,8 +73,10 @@ codeSkip == 1		{ next }
 }
 
 END {
-	if (sbuf) print "#include \"sbuf/SBuf.h\""
-	print "#include \"" nspath type ".h\""
+	if (sbuf) {
+            print "#include \"sbuf/SBuf.h\""
+	    print "#include \"" nspath type ".h\""
+        }
```
yadij, line 79:
```suggestion
	if (sbuf) print "#include \"sbuf/SBuf.h\""
	print "#include \"" ipath nspath type ".h\""
```

yadij:
Above two suggestions should fix the mk-string-arrays.awk logic in context of the current error/forward.h contents.

FYI: nspath is not related to the SBuf usage. It is a path automatically added for convenience libraries which are correctly using namespace for their symbol definitions. If you update the classes in liberror to be within an Error:: namespace (per our coding guidelines they should be) you can drop these changes to the AWK script.

rousskov:
> Amos: If you update the classes in liberror to be within an Error:: namespace (per our coding guidelines they should be) ...

We need to proceed carefully here to avoid strange/unwanted outcomes driven by imperfect guidelines. If some coding guideline _requires_ all liberror classes to be in Error namespace, then that guideline _itself_ may need more work: Not all names in subdirectories should be in namespaces and not all subdirectories need a namespace.

src/error/Makefile.am
chtsanti: Does not fix the issue.
Please check my commit 424cb777ae, it it is an acceptable solution. 
Commit comment:
> Previews "Remove the uneeded err_detail_type.h ..." commit changed
    mk-string-arrays.awk to add an include file with array definition
    only when this array is an array of SBufs. The reasons are:
      a) I (wrongly?) believed that only in the case of SBus is needed.
      b) It computed the include file name using the enum name. In the case of
         error/categories.cc the include file (error/forward.h) does not match
         the enum name (err_type)
 > This patch fixes mk-string-arrays.awk to accept an include file name
    as parameter and add it to #include files list if given

src/mk-string-arrays.awk
chtsanti: commit 424cb77

src/error/Detail.cc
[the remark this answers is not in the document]
chtsanti: commit 7c383f65dc

src/error/Detail.cc
```diff
@@ -0,0 +1,94 @@
...
+public:
+    /// briefly describes the error circumstances
+    /// must not contain characters that require quoting in access logs or HTML
+    typedef SBuf Name;
```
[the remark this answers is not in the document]
chtsanti: > Is there actually a real need for this to be typedef'd ? 

I have not any opinion on this. I will wait @rousskov opinion before remove it.

src/errorpage.cc
```diff
@@ -962,18 +963,13 @@ ErrorState::compileLegacyCode(Build &build)
...
-            detail->useRequest(request.getRaw());
-            const String &errDetail = detail->toString();
-            if (errDetail.size() > 0) {
-                const auto compiledDetail = compileBody(errDetail.termedBuf(), false);
-                mb.append(compiledDetail.rawContent(), compiledDetail.length());
-                do_quote = 0;
-            }
+            auto rawDetail = detail->verbose(request);
+            // TODO: Support compilation of unterminated/SBuf bodies.
+            const auto compiledDetail = compileBody(rawDetail.c_str(), false);
```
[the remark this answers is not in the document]
chtsanti: commit 3a88ef0ea3
Personally I believe it is good idea to always allocate one more byte in SBuf just for C-string termination  if required.

src/tests/stub_libsecurity.cc
chtsanti: We have many `#if USE_OPENSS` and `#if USE_GNUTLS` in squid code outside libsecurity.
Moreover the stub_libsecurity.cc is a dummy implementation of libsecurity. Why not to have this `#if` here?

src/error/Detail.h
chtsanti: Finally I did not split.
There are source files which need more than one of class definitions this is will result to more files included.
Looks ok  include them from "error/forward.h" so the uses include only this file?
In the other hand these classes are very small, I am not seeing any reason to spread them into separate files.

yadij: By that logic all Squid code would be in one huge file. Other code combining code is not a good reason for all code to do so. It needs to be considered case-by-case and I believe this is a case where they should be split.

src/tests/stub_libsecurity.cc
yadij: The existing #if outside of libsecurity are primarily legacy code that we are still needing to make generic.

The problem with having this if in the stub is that it wraps the entire definition. Exposing the fact that the class is not abstracted cleanly. As I said a clean abstraction would define LibErrorCode for OpenSSL code and this ctor could use it.

src/errorpage.cc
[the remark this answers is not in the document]
yadij: It is not the length of buffer that causes the reallocation. It is the need to cow() to prevent the terminator injection corrupting any SBuf which shares and overlaps the memory buffer.

src/error/Makefile.am
yadij: It is not.  The AWK logic should be:

 * when enum is an array of SBuf strings SBuf.h must be included by the generated file.
> 	if (sbuf) print "#include \"sbuf/SBuf.h\""

 * when ifile is provided include the given file
>	if (ifile != "") print "#include \"" ifile "\""

 * otherwise just include the header the enum is defined in
>	else print "#include \"" nspath type ".h\""


Also, you are using the ifile parameter only for http/MethodTypes.h when this PR is needing it for error/Details.h

yadij:
> > Amos: If you update the classes in liberror to be within an Error:: namespace (per our coding guidelines they should be) ...
> 
> We need to proceed carefully here to avoid strange/unwanted outcomes driven by imperfect guidelines. If some coding guideline _requires_ all liberror classes to be in Error namespace, then that guideline _itself_ may need more work: Not all names in subdirectories should be in namespaces and not all subdirectories need a namespace.


Under our SourceLayout model each convenience library is intended to be a distinct component which has its own namespace.  If all the content of a library is in global namespace we need to reconsider why it is in a dedicated library at all.

src/tests/stub_libsecurity.cc
chtsanti: My opinion is that we should consider the Security::ErrorDetail constructors  as internal to security namespace.  They are used only internally to generated errors. 
My sense is that their definition required in stub_libsecurity just because there is reference inside client_side.cc code,  inside `#if USE_OPENSSL` block, which is must moved under security namespace. When this is done these ErrorDetail constructors could be removed from stub_libsecurity.cc.

If required I think we can hide these two constructors by just build two new classes, one for each TLS toolkit which has only one constructor. But I am not seeing any strong reason for this. The code outside security and Ssl namespaces only interested for ErrorDetail::brief() and ErrorDetail::verbose() virtual methods, this is the abstract interface.

> As I said a clean abstraction would define LibErrorCode for OpenSSL code and this ctor could use it.

This branch adds a clear definition of LibErrorCode for OpenSSL. But GnuTLS and OpenSSL are different toolkits with their own philosophy and we can not have always the absolute equivalent of each other.
In Security::ErrorDetail constructor we need to pas an io error. The io error in GnuTLS is expressed by a LibErrorCode, but OpenSSL has a different set of integer io error codes.

I must also note that the previous try inside Squid to define a common error  code which is used by both OpenSSL and GnuTLS was the ErrorCode, but it was buggy.

But If you have suggestions I will like to hear about.

src/error/Detail.h
chtsanti: commit f62b229a6f

> By that logic all Squid code would be in one huge file

It is not the same.  Splitting STL exceptions will result to about 10 new include files instead of just including "exception". 

In any case, and finally, this change in Squid does not looks bad.

src/error/Makefile.am
chtsanti: I am ok with your suggestion, I fix it in commit aa65598e7.

However still I prefer my form:
 - I think make more sense a tool command line parameter which has a strict job. Eg "add this file in include files list".
 - Only the generated http/MethodType.cc needs to include the http/MethodType.h file
 - The ifile can be extended to pass more than one files to include:  `mk-string-arrays.awk ifile="sbuf/SBuf.h;http/MethodType.h" < $(srcdir)/MethodType.h`. It will cover more future cases.

rousskov:
> [SourceLayout](https://wiki.squid-cache.org/action/diff/Features/SourceLayout): 
>> **Problem**: Can we remove Foo prefix from FOO/FooSomething.h file names?
>> **Proposed Solution**: Yes, carefully. File name should match the primary class declared or defined in that file. Directory name should match the (lowercased) namespace used by classes in that directory. We should move from PROTOFoo to PROTO::Foo classes.

> Amos: Under our SourceLayout model each convenience library is intended to be a distinct component which has its own namespace.

The namespace bit in your assertion looks like a misunderstanding/misinterpretation to me. There is no blanket Project policy that requires each convenience library to have its own namespace AFAICT! Yes, most FooX classes should be moved into a foo directory and into a Foo namespace (becoming Foo::X). That does not imply that all X classes need to become Y::X.

The quoted Proposed Solution has to be interpreted in the context of the quoted Problem that it is solving! When I [wrote](https://wiki.squid-cache.org/action/diff/Features/SourceLayout?action=diff&rev1=2&rev2=3) the initial version of that proposed solution, we were [discussing](http://www.squid-cache.org/mail-archive/squid-dev/200802/0144.html) protocol-specific FooX classes, where namespaces are a natural (and likely the best) solution (let me know if I should detail why I think that way). General-purpose classes like the error reporting scaffolding in liberror is a completely different animal. They were not the subject of the original "Can we remove Foo prefix?" question, and they most likely should not be moved into a namespace.

> If all the content of a library is in global namespace we need to reconsider why it is in a dedicated library at all.

Anybody may reconsider any specific design, of course. I hope we still agree that we should be moving towards `src/` without source code files (i.e. just subdirectories and a few metafiles). Where should general-purpose error-related scaffolding go? Today, I do not see a better solution than placing that code into src/error/; those file have enough commonality, specificity, and volume to deserve a dedicated directory IMO. I welcome better ideas, of course, especially when they are born during a search for a solution to a real-world problem.

chtsanti:
I believe I responded in all requests for changes.
There are open:
 - I need @rousskov opinion about removing a "typedef SBuf Name"  (which is used only in one place, but I know Alex loves such typedefs)
 - The request for removing a `#if USE_OPENSSL .. #elif USE_GNUTLS` inside stub_libsecurity.cc needs more discussion. Requires to use more or less the same  error details/parameters for two different TLS libraries with a little different philosophy.

@rousskov and @yadij please re-review.

rousskov:
@chtsanti, thank you for making progress and enumerating the remaining to-dos. I know some change requests raised red flags for me too. I hope to get to this PR very soon.

rousskov (commented):
@chtsanti has [identified](https://github.com/squid-cache/squid/pull/691#issuecomment-696249346) two unaddressed change requests. I have committed the change requested in one of them (commit 937d57a). I also explained why the other change request should not result in any code changes IMO.

@yadij, the ball is on your side.

rousskov:
@kinkie, please check [Jenkins](https://build.squid-cache.org/job/5-pr-test/2738/COMPILER=clang,OS=centos-7,label=docker-build-host/console):

```
12:21:56 docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```

rousskov:
I committed a few changes to fix the build after merging master into the PR branch. Semaphore CI is happy now. Jenkins is itself broken, as usual.

@yadij, I do not think any of these build-fixing changes are controversial, but commit 771655979 changes were not exactly trivial so you may want to look at them if you are not going to re-review the whole PR.

rousskov:
@kinkie, please check [Jenkins](https://build.squid-cache.org/job/5-pr-test/COMPILER=clang,OS=centos-7,label=docker-build-host/2740/console):

```
14:45:19 docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```

yadij (approved):
Sorry for the delay. Latest code looks fine now. Thank you very much.

rousskov:
Jenkins is still [failing](https://build.squid-cache.org/job/5-pr-test/2744/COMPILER=gcc,OS=centos-7,label=docker-build-host/console) with the same error:

```
16:13:05 docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```

@eduard-bagdasaryan, do you know how to start that docker daemon on Jenkins (if that is what is needed). I have [asked](https://github.com/squid-cache/squid/pull/691#issuecomment-742669283) @kinkie, but he is probably busy with more important stuff.

eduard-bagdasaryan:
Well, I started docker service, let's see if it helps.

rousskov:
@eduard-bagdasaryan, it helped. Thank you.

PR #692, opened 2020-07-20
Bug 5065: url_rewrite_program documentation update

patrickscottbest:


src/cf.data.pre
```diff
@@ -5685,6 +5685,7 @@ DEFAULT: none
 DOC_START
 	Specify the location of the executable URL rewriter to use.
 	Since they can perform almost any function there isn't one included.
+	URL rewrites occur after acl evaluations.
```
rousskov, line 5688: It would indeed be useful to document the timing of those rewrites, but the "after acl evaluations" part is not clear because Squid evaluates ACLs at many different points of a transaction lifetime. Can you suggest a more precise wording?

patrickscottbest: My research or knowledge isn't thorough enough to describe this in relation to all the acl execution points, other than to say something like "URL rewrites occur after " http_access acl's".  
In other words , http://before.com is evaluated against an http_access acl prior to it's transformation to http://after.com.  

A statement like that would have saved me a little time in experimentation.

If i'm still off the mark, or you think it's just unfit, i would digress on 5688, and be perfectly happy with the one-character grammar change.  Thank you!

yadij: For now you could add a reference to see https://wiki.squid-cache.org/ProgrammingGuide/Architecture#Transaction_Processing for details on transaction sequencing.

patrickscottbest: Will do.

yadij: Any update? next release happens end of the week so earlier this gets completed and merged the better.

yadij:
Updated description to meet Anubis requirements.

yadij:
OK to test

src/cf.data.pre
patrickscottbest: I have updated the text.  Thank you.

yadij (approved):

PR #694, opened 2020-07-23
Bug 5133: OpenSSL 3.0 support 

yadij:
This TLS update includes:

* Fix build with OpenSSL v3.
* Refactor RSA key generation to avoid deprecated RSA_*() APIs.
* Refactor DH parameter and key config to avoid deprecated DH_*() APIs.
* Refactor ECDH key creation to avoid deprecated EC_*() APIs.
* Deprecate ssl_engine support in builds with OpenSSL v1-.
* Disable ssl_engine support in builds OpenSSL v3+.

We deprecated/removed ssl_engine support (as summarized in the last two
bullets above) without providing an OpenSSL Providers-based alternative
because of the following factors:

1. We do not have the resources to update ssl_engine code to build
   (without deprecation warnings) with OpenSSL v3 when the feature is
   unused.

2. We do not have the resources to create an OpenSSL v3 Provider-based
   replacement for ssl_engine code that uses deprecated Engine APIs.

3. OpenSSL v3 deprecated Engine support (triggering deprecation warnings
   in applications that use Engine APIs with OpenSSL v3). Since Squid
   default builds use -Werror, doing nothing would break such builds.

4. Squid ssl_engine does not appear to be a popular feature.


yadij:
@christos, this EVP change to cert generation seems to work fine for me with libssl1.1.  But when I build against libssl3 the first "make check" test of default squid.conf parsing hangs when initializing the default "untrusted" server cert Squid uses for error responses.

Is there anything you can see I'm doing wrong or can think of elsewhere that may be leading to the hang?

Does this code work for you with OpenSSL 1.1 ?
 i.e. is it good enough to merge the RSA upgrade for older OpenSSL and treat as a bug in v3.0?

If we cant track this down ourselves I will try to get in touch with OpenSSL dev about a possible v3 bug at their end.

src/ssl/gadgets.cc
```diff
@@ -9,35 +9,28 @@
...
 
-    if (!EVP_PKEY_assign_RSA(pkey.get(), (rsa.get())))
-        return NULL;
+    /* Generate key */
+    Security::PrivateKeyPointer pkey(EVP_PKEY_new());
+    if (pkey) {
+        auto *foo = pkey.get();
+        if (EVP_PKEY_keygen(rsa.get(), &foo) <= 0)
+            return nullptr;
+    }
```
chtsanti, line 32: I do not think that this part of code can work. Without studying a lot the related OpenSSL api my sense is that you need something like the following:

  `EVP_PKEY *pkey = NULL;`
  `EVP_PKEY_keygen(rsa.get(), pkey);`
  `return pkey;`

chtsanti:
> @christos, this EVP change to cert generation seems to work fine for me with libssl1.1. But when I build against libssl3 the first "make check" test of default squid.conf parsing hangs when initializing the default "untrusted" server cert Squid uses for error responses.
> 

The "make check" works for me using openSSL-1.1.1 and openSSL-3.0.0alpha5 releases.

> Is there anything you can see I'm doing wrong or can think of elsewhere that may be leading to the hang?

My sense is that the CreateRsaPrivateKey is not correctly implemented. But I do not know if it can hang squid.

> 
> Does this code work for you with OpenSSL 1.1 ?
> i.e. is it good enough to merge the RSA upgrade for older OpenSSL and treat as a bug in v3.0?
> 
> If we cant track this down ourselves I will try to get in touch with OpenSSL dev about a possible v3 bug at their end.

---

I must also note that when squid trying to use an openSSL-3.0.0 which is not compiled with OPENSSL_NO_ENGINE flag can not compile. It will fail inside Ssl::Initialize, support.cc file inside OPENSSL_NO_ENGINE related block of code

src/ssl/gadgets.cc
yadij: Thanks. Done that.

satya-rajesh:
Tried with openssl-3.0.0-alpha17, observing the below failures with squid 5.0.5. Seems openssl 3.0.0 has made some changes for below in alpha17.

In file included from ../../compat/openssl.h:42,
                 from ../../src/security/Context.h:15,
                 from ../../src/security/forward.h:13,
                 from ../../src/security/PeerOptions.h:14,
                 from PeerOptions.cc:16:
PeerOptions.cc:330:5: error: missing binary operator before token "1"
  330 | #if SSL_OP_TLS_ROLLBACK_BUG
      |     ^~~~~~~~~~~~~~~~~~~~~~~
PeerOptions.cc:335:5: error: missing binary operator before token "1"
  335 | #if SSL_OP_ALL
      |     ^~~~~~~~~~
PeerOptions.cc:370:5: error: missing binary operator before token "1"
  370 | #if SSL_OP_CIPHER_SERVER_PREFERENCE
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
PeerOptions.cc:380:5: error: missing binary operator before token "1"
  380 | #if SSL_OP_NO_SSLv3
      |     ^~~~~~~~~~~~~~~
PeerOptions.cc:385:5: error: missing binary operator before token "1"
  385 | #if SSL_OP_NO_TLSv1
      |     ^~~~~~~~~~~~~~~
PeerOptions.cc:392:5: error: missing binary operator before token "1"
  392 | #if SSL_OP_NO_TLSv1_1
      |     ^~~~~~~~~~~~~~~~~
PeerOptions.cc:399:5: error: missing binary operator before token "1"
  399 | #if SSL_OP_NO_TLSv1_2
      |     ^~~~~~~~~~~~~~~~~
PeerOptions.cc:406:5: error: missing binary operator before token "1"
  406 | #if SSL_OP_NO_TLSv1_3
      |     ^~~~~~~~~~~~~~~~~
PeerOptions.cc:413:5: error: missing binary operator before token "1"
  413 | #if SSL_OP_NO_COMPRESSION
      |     ^~~~~~~~~~~~~~~~~~~~~
PeerOptions.cc:418:5: error: missing binary operator before token "1"
  418 | #if SSL_OP_NO_TICKET
      |     ^~~~~~~~~~~~~~~~
make[4]: *** [Makefile:904: PeerOptions.lo] Error 1

sergiodj:
FWIW, I have backported the patches from this PR, built and tested squid 5.2 (actually, `5.2-1ubuntu1`) that we have in Ubuntu Jammy (development version), and everything seems OK.  We're using the same autopkgtests from the Debian squid package, and they're all passing.  This is not a "real world scenario", but I thought I'd leave this feedback here to help move this PR forward.

FTR, we will ship Jammy with OpenSSL 3, which has just migrated to jammy-release a few minutes ago.

src/security/PeerOptions.cc
```diff
@@ -297,130 +297,130 @@ static struct ssl_option {
 
 } ssl_options[] = {
 
-#if SSL_OP_NETSCAPE_REUSE_CIPHER_CHANGE_BUG
+#if defined(SSL_OP_NETSCAPE_REUSE_CIPHER_CHANGE_BUG)
```
rousskov, line 300: This and similar changes seem to contradict OpenSSL v3 code that does not (AFAICT) support these options but defines these constants:
https://github.com/openssl/openssl/blob/0fcf2351ecff5db21cba431704e4da631b74904a/include/openssl/ssl.h.in#L425-L433

This change confuses our supported option searching code in Security::PeerOptions::parseOptions() which assumes that options listed here have non-zero values. That code may still "work" after this change but that is essentially an accident that we should not rely on.

Could you please clarify why these preprocessor conditions should be changed?

src/security/PeerOptions.cc
```diff
@@ -512,7 +512,7 @@ Security::PeerOptions::parseOptions()
 
     }
 
-#if SSL_OP_NO_SSLv2
+#if defined(SSL_OP_NO_SSLv2)
     // compliance with RFC 6176: Prohibiting Secure Sockets Layer (SSL) Version 2.0
     op = op | SSL_OP_NO_SSLv2;
```
rousskov, line 517: Similar preprocessor change concern here: After this PR change, this "Prohibiting Secure Sockets Layer" code prohibits nothing when SSL_OP_NO_SSLv2 is zero, misleading the reader. Please clarify why this preprocessor condition should be changed. Referring to another change request explanation, it it applies, is fine, of course.

src/security/ServerOptions.cc
```diff
@@ -372,9 +376,36 @@ Security::ServerOptions::loadDhParams()
...
     parsedDhParams.resetWithoutLocking(dhp);
+
+#else // OpenSSL 3.0+
+    EVP_PKEY *pkey = nullptr;
+    const char *type = "DH";
+    if (!eecdhCurve.isEmpty())
+        type = "EC";
+    // XXX: use the eecdhCurve name when generating the EVP_KEY object. or at least verify it matches the loaded params.
+
+    if (auto *dctx = OSSL_DECODER_CTX_new_for_pkey(&pkey, "PEM", nullptr, type, OSSL_KEYMGMT_SELECT_ALL, nullptr, nullptr)) {
```
rousskov, line 388: Is that XXX a regression (i.e. we were properly handing eecdhCurve before these changes) or just a missing feature (i.e. we did not use eecdhCurve name before these changes either)? This change request does not request any code changes. I just do not know enough about this code to guess the answer by reading the XXX comment itself.

src/ssl/support.cc
```diff
@@ -559,7 +559,11 @@ Ssl::VerifyCallbackParameters::At(Security::Connection &sconn)
 }
 
 // "dup" function for SSL_get_ex_new_index("cert_err_check")
-#if SQUID_USE_CONST_CRYPTO_EX_DATA_DUP
+#if OPENSSL_VERSION_MAJOR >= 3
+static int
+ssl_dupAclChecklist(CRYPTO_EX_DATA *, const CRYPTO_EX_DATA *, void **,
+                    int, long, void *)
+#elif SQUID_USE_CONST_CRYPTO_EX_DATA_DUP
```
rousskov, line 566: AFAICT, this change implies that SQUID_USE_CONST_CRYPTO_EX_DATA_DUP-setting code needs to be updated (or, better, the whole thing should be refactored to move all this mess to ./configure code). Otherwise, falsy SQUID_USE_CONST_CRYPTO_EX_DATA_DUP in the PR code incorrectly implies that SSL_get_new_ex_index() dup callback accepts non-const `CRYPTO_EX_DATA *`. I do not insist on these changes though.

rousskov (changes requested):
Thank you for working on OpenSSL v3 support.

I am concerned about proliferation of version-specific checks (`#if OPENSSL_VERSION_MAJOR...`). Have you considered using feature tests instead? I believe the latter is our usual/default preference, but perhaps this use case is exceptional? If it is exceptional, it would be good to document why in the PR description.

rousskov:
I adjusted PR description a little to make it more readable and improve markdown rendering. Please adjust or undo as needed.

sergiodj:
Hi,

The development version of Ubuntu (Jammy) is already shipping openssl 3.0, and for this reason squid is FTBFSing there currently.  I've been keeping an eye on this PR for a while, because eventually it will need to be backported to squid in Ubuntu.  I don't feel entirely comfortable backporting this as-is because of the requests for changes that have been made.  I've been wondering if you have any estimate on the completion of this PR.  Thanks in advance.

rousskov:
> I've been wondering if you have any estimate on the completion of this PR.

I cannot speak for @yadij (the primary PR author), but as a reviewer, I will do my best to move quickly on this PR. Other development projects are currently stuck waiting for this PR to land.

I know there was no activity for a month since my [review](https://github.com/squid-cache/squid/pull/694#pullrequestreview-832080109), but it was during the holiday season, so I would not read too much into it. In the extreme case, if Amos does not have the time to move this PR forward, then Factory can step in and finish it, but I would very much prefer for Amos to finish what he started, of course. Given all these factors, my estimate is that this PR should take anywhere from a few days to a few months to land. I hope the increased interest will push it closer to a few days :-). Thank you for updating this PR!

sergiodj:
Thanks for the reply, @rousskov.  I'll continue monitoring this PR and decide what to do next month, when Ubuntu enters the Feature Freeze phase.

configure.ac
```diff
@@ -1299,14 +1299,22 @@ AC_MSG_NOTICE([GnuTLS library support: ${with_gnutls:=auto} ${LIBGNUTLS_PATH} ${
...
+  CPPFLAGS="$LIBOPENSSL_CFLAGS $CPPFLAGS"
+  LIBS="$LIBS $LIBOPENSSL_PATH"
+  PKG_CHECK_MODULES([LIBOPENSSL],[openssl >= 3],[with_openssl="yes"],[:])
+  SQUID_STATE_ROLLBACK(squid_openssl3_state)
+])
+## OpenSSL < 3 is default disable due to licensing issues on some OS
 if test "x$with_openssl" = "xyes"; then
   CPPFLAGS="$LIBOPENSSL_CFLAGS $CPPFLAGS"
   AC_CHECK_HEADERS( \
     openssl/asn1.h \
```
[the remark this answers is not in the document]
hno: My understanding is that it is just a bit of juggling to simulate --with-openssl, results in with_openssl="yes" as if --with-openssl had been specified on the command line. No change wrt AC_CHECK_HEADERS and other actions.

But I am a bit confused by the LIBS&CFLAGS juggling, and how it works with a non-standard path --with-openssl=/path/to/openssl-install but original code is pretty much as confusing.

And also the PKG_CHECK_MODULES is invoked twice, which should be harmless but not sure it's intentional.

But is this at all working? The SQUID_OPTIONAL_LIB() already sets with_openssl="no" for the default case I think

  AS_IF([test "x$withval" = "x"],[m4_translit([with_$1], [-+.], [___])=no])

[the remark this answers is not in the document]
rousskov: > My understanding is that it is just a bit of juggling to simulate --with-openssl, results in with_openssl="yes" as if --with-openssl had been specified on the command line. No change wrt AC_CHECK_HEADERS and other actions.

I agree. I misread the diff because the code essentially overwrites user configuration. Sorry. I consider my change request resolved. @yadij, I will keep it open because Henrik's comment discusses other issues, but please feel free to close this thread.

I also agree that there are other problems here. I did not want to dive into them until my higher-level concern was addressed. I recommend opening dedicated change request(s) for those concerns rather than piggybacking them to this addressed change request. For example, The misnamed SQUID_OPTIONAL_LIB() is documented to target default-disabled libraries. If the documented "Default-enable OpenSSL if pkg-config finds OpenSSL v3+." goal of this PR remains unchanged, then OpenSSL should no longer be considered a "default-disabled" library: Since v3+ is the "stable present and future", we should make our primary decisions based on that, and then take care of any "temporary exceptions" like v1-.

The original code is terrible. This PR does not have to improve it, but we should not make things worse, of course.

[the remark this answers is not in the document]
hno: I'd suggest dropping this fragment, and work on openssl3 enabled by default in another follow-up change after this PR. It's unrelated to adding support.

rousskov:
@yadij, you are probably busy with more urgent things. Do you want Factory to finish this important OpenSSL work? You will be able to review the result, of course. I can even use a dedicated branch/PR if you prefer. Please let me know either way. Thank you.

src/security/ServerOptions.cc
yadij: This is a regression that needs to be resolved before this PR is suitable for use by anyone wanting EC ciphers.

src/security/PeerOptions.cc
yadij: This is intended to solve two problems;
  1) compile warning/error with opensslv3 about some of the option definitions being non-constexpr, and
  2) backward compatibility. So that Squid built with libssl3 can take old config files using the deprecated options from libssl1.x

src/cf.data.pre
```diff
@@ -3057,6 +3057,8 @@ DEFAULT: none
 DOC_START
 	The OpenSSL engine to use. You will need to set this if you
 	would like to use hardware SSL acceleration for example.
+
+	Note: OpenSSL 3.0 and newer do not provide Engine support.
```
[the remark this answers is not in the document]
yadij: The ENGINE API is not available in the libssl3 I am building/testing against. On top of that I have never seen this option used by anyone. So for now I have opted to deprecate the whole feature.

If anyone really wants to have an ENGINE with libssl3 they will need to refactor all the code making use of this config option and the variables it sets up.

configure.ac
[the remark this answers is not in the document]
yadij: Dropped in[0f14f83](https://github.com/squid-cache/squid/pull/694/commits/0f14f83d5e83e524291f4ba8ea2fac83e84722a3)

src/ssl/support.cc
yadij: This change is about parameter 3 (```void *```) becoming a ```void **`` and parameter 2 being cemented as the ```const CRYPTO_EX_DATA``` for all OpenSSL 3+.

Older OpenSSL versions require non-const ```CRYPTO_EX_DATA```. Except some release had backported patches of partial v3 API - adding only the ```const CRYPTO_EX_DATA```. This latter group is detected by SQUID_USE_CONST_CRYPTO_EX_DATA_DUP.

src/security/ServerOptions.cc
yadij: It turns out the issue was OSSL_KEYMGMT_SELECT_ALL causing errors when irrelevant (to EC) details were missing from the parameters file. Patch fixing that and polishing debugs output coming soon as the test-builds.sh completes here.

OpenSSL auto-detects the curve name etc from the file contents. It is not clear why earlier versions needed the curve name to be provided by the load functions (validation maybe?). I'm reducing this comment to a TODO for later.

src/cf.data.pre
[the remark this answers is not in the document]
rousskov: > The ENGINE API is not available in the libssl3 I am building/testing against.

The limitations specific to the library you use should not affect the decision regarding deprecation of a Squid feature that may be used by others. Perhaps you can rephrase the above statement in more general terms (e.g., OpenSSL v3 that comes by default with Ubuntu X disables ENGINE API support)? A _general_ statement about the necessary library availability could influence our decision to deprecate, of course.

> I have never seen this option used by anyone. So for now I have opted to deprecate the whole feature.

You have seen it used by bug [4112](https://bugs.squid-cache.org/show_bug.cgi?id=4112) reporter (at least) AFAICT.

> If anyone really wants to have an ENGINE with libssl3 they will need to refactor all the code making use of this config option and the variables it sets up.

The cost of support is a valid factor in our decision making process, of course. Can you give an example of Squid code that will need to be changed in order to preserve support for OpenSSL engine selection via `ssl_engine` directive in Squid built with OpenSSL v3? My superficial interpretation of OpenSSL v3 migration [documentation](https://www.openssl.org/docs/man3.0/man7/migration_guide.html) is that there should be no Squid changes as long as the OpenSSL library is built with that deprecated API/code enabled.

----

Assuming the continued support is indeed costly (which is to be determined above), one could address this change request by making the following changes:

1. Adjust PR description to document that we are deprecating OpenSSL Engine support (not because it is removed in OpenSSL v3 but) because it is deprecated in OpenSSL v3, popular distributions ship OpenSSL v3 without Engine support (if that is true), and we do not have the resources to update the corresponding Squid code to remain compatible with OpenSSL v3 (if the code needs updating).
2. Add deprecation warning (for OpenSSL before v3) to squid.conf parser or interpreter. You have already adjusted the code to fail with OpenSSL v3+.
3. Replace the current (incorrect) claim in cf.data.pre with something along these lines:

```suggestion
	Not supported in builds with OpenSSL v3 or newer.
```

There are other ways to address this change request, of course. The above is just an example that illustrates the missing pieces if we decide to deprecate `ssl_engine` support starting with OpenSSL v3 builds.

src/security/PeerOptions.cc
rousskov: > compile warning/error with opensslv3 about some of the option definitions being non-constexpr

Does this "compile warning/error" problem apply to the line this change request is attached to? It does not seem to apply (because `SSL_OP_NETSCAPE_REUSE_CIPHER_CHANGE_BUG` is defined to be `0x0` so the unchanged code should have compiled, but perhaps I am missing something.

If there are two classes of PR changes here (i.e., PR changes necessary to avoid "compile warning/error with opensslv3" and all others), then let's use this change request for the first class.

> backward compatibility. So that Squid built with libssl3 can take old config files using the deprecated options from libssl1.x

We do not want that kind of backward compatibility IMO: Squid should not silently swallow unsupported options it knows about. A configuration using `SSLREF2_REUSE_CERT_TYPE_BUG` should either stop working after the upgrade due to a fatal configuration error or, at the very least, log an error that the configured option is no longer supported.

src/security/PeerOptions.cc
rousskov: > OpenSSLv3 makes the definition of all options an expression instead of a constant value

OpenSSL v3 defines `SSL_OP_NO_SSLv2` to be `0x0`. `0x0` is a constant value AFAICT. Moreover, `SSL_OP_NO_SSLv2` was already set to zero (using a different spelling) in OpenSS v1.1!

> it is either desirable, or at worst harmelss, to set this option whenever the macro is #defined.

When `SSL_OP_NO_SSLv2` is defined to a falsy value `0x0` (as it is in OpenSSL v3), it is _not_ desirable to "set this option" (IMO) because
* doing so misleads the reader that into thinking that this "Prohibiting Secure Sockets Layer v2" code prohibits something when, in fact, it does not;
* it increases the PR footprint without any runtime or code reading time benefit (that I can see).

In other words, the official code is better than the proposed PR change. Please revert this and similar changes or explain why they are needed in this PR.

rousskov (changes requested):
This is not a comprehensive review, but I have updated a few old unresolved change requests to help you make progress. My [offer](https://github.com/squid-cache/squid/pull/694#issuecomment-1035507074) to take this PR off your hands still stands.

src/security/PeerOptions.cc
yadij: The options defined [1] to ```0x0``` are in the backward compatibility group. For slightly older libssl releases where they were/maybe set to a macro.

[1] the definition value is specific to the OpenSSL code you looked at. The const/macro nature and resulting value can and does change between releases of OpenSSL.

rousskov: > the definition value is specific to the OpenSSL code you looked at. The const/macro nature and resulting value can and does change between releases of OpenSSL.

I looked at more than 200 OpenSSL tagged releases (all I could find). All of them define `SSL_OP_NETSCAPE_REUSE_CIPHER_CHANGE_BUG`  to be a constant (`0x00000008L`, `0x0`, or `0x0U`). In OpenSSL v3, it is defined to be `0x0`. These facts support my change request: There are known cases where the proposed change makes things worse and no known cases where it makes things better. Thus, unless you have specific evidence to the contrary, this (and equivalent) changes should be undone.

If there are some similarly-looking PR changes that do not follow the above pattern of constant macro values, then they may need to be treated differently, of course, but that would be a discussion for another piece of code.

src/security/PeerOptions.cc
rousskov: I also checked other tagged OpenSSL releases (all I 200+ could find), and, AFAICT, all those that define `SSL_OP_NO_SSLv2` macro set it to a constant value (`0x0`, `0x00000000U`, or `0x01000000L`).

src/security/PeerOptions.cc
yadij: Package: libssl-dev 3.0.3-5
```
In file included from ../../../compat/openssl.h:44,
                 from ../../../src/security/Context.h:15,
                 from ../../../src/security/forward.h:14,
                 from ../../../src/security/PeerOptions.h:14,
                 from ../../../src/security/PeerOptions.cc:16:
../../../src/security/PeerOptions.cc:330:5: error: missing binary operator before token "1"
  330 | #if SSL_OP_TLS_ROLLBACK_BUG
      |     ^~~~~~~~~~~~~~~~~~~~~~~
../../../src/security/PeerOptions.cc:335:5: error: missing binary operator before token "1"
  335 | #if SSL_OP_ALL
      |     ^~~~~~~~~~
../../../src/security/PeerOptions.cc:370:5: error: missing binary operator before token "1"
  370 | #if SSL_OP_CIPHER_SERVER_PREFERENCE
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../src/security/PeerOptions.cc:380:5: error: missing binary operator before token "1"
  380 | #if SSL_OP_NO_SSLv3
      |     ^~~~~~~~~~~~~~~
../../../src/security/PeerOptions.cc:385:5: error: missing binary operator before token "1"
  385 | #if SSL_OP_NO_TLSv1
      |     ^~~~~~~~~~~~~~~
../../../src/security/PeerOptions.cc:392:5: error: missing binary operator before token "1"
  392 | #if SSL_OP_NO_TLSv1_1
      |     ^~~~~~~~~~~~~~~~~
../../../src/security/PeerOptions.cc:399:5: error: missing binary operator before token "1"
  399 | #if SSL_OP_NO_TLSv1_2
      |     ^~~~~~~~~~~~~~~~~
../../../src/security/PeerOptions.cc:406:5: error: missing binary operator before token "1"
  406 | #if SSL_OP_NO_TLSv1_3
      |     ^~~~~~~~~~~~~~~~~
../../../src/security/PeerOptions.cc:413:5: error: missing binary operator before token "1"
  413 | #if SSL_OP_NO_COMPRESSION
      |     ^~~~~~~~~~~~~~~~~~~~~
../../../src/security/PeerOptions.cc:418:5: error: missing binary operator before token "1"
  418 | #if SSL_OP_NO_TICKET
      |     ^~~~~~~~~~~~~~~~
```

rousskov: > Treating some `SSL_OP_*` macros as `#if X` and others as `#if defined(X)` is inconsistent and odd looking code. It also increases our maintenance burden in future - requiring us to track OpenSSL (and its distributors) patching of libssl API definitions for these macros.

Finally we are starting to make progress, thank you!

```C
#define SSL_OP_BIT(n)  ((uint64_t)1 << (uint64_t)n)
...
# define SSL_OP_TLS_ROLLBACK_BUG                         SSL_OP_BIT(23)
...
# define SSL_OP_NETSCAPE_REUSE_CIPHER_CHANGE_BUG         0x0
```

So we have a group of macros that use constants (including `SSL_OP_NETSCAPE_REUSE_CIPHER_CHANGE_BUG` that triggered this change request) and a group of macros that use constant expressions (including `SSL_OP_TLS_ROLLBACK_BUG` in the error output you have now shared). A compiler does not allow us to process macros in both categories using `#if` conditionals. The logic represented by those macros does not allow us to process macros in both categories using `#ifdef` conditionals alone. A given macro may travel from one category to another (or might even disappear if we do not cut support for old OpenSSL versions soon enough). We want to be consistent and future-proof, so we need to develop a single solution that covers all two (or three) categories.

I suggest using this approach:

1. Use `#if defined(SSL_OP_X)` to test the existence of an OpenSSL `SSL_OP_X` macro (during preprocessing). The PR already contains all the necessary changes for this AFAICT.
2. Use `if (SSL_OP_X)` or equivalent to test whether the corresponding OpenSSL feature/option/etc. is actually enabled/supported by OpenSSL (during compilation). The current PR is missing those changes, but they are fairly easy to add.

For this specific change request, we need to adjust the ssl_options loop in Security::PeerOptions::parseOptions() so that it uses Optional for `value` (instead of misinterpreting zero `value` as "not found").

src/security/PeerOptions.cc
rousskov: Based on the suggestion in the other change request:

```suggestion
    if (SSL_OP_NO_SSLv2)
        op = op | SSL_OP_NO_SSLv2;
```

Yes, the added `if` does not have an effect on `op` final value, but it improves consistency of treating `SSL_OP_X` values and highlights an important caveat: being `#defined` is not enough for `SSL_OP_X` to be supported. If you prefer to be more explicit about it, we can do this:

```suggestion
    if (SSL_OP_NO_SSLv2) // may be #defined but not supported
        op = op | SSL_OP_NO_SSLv2;
```

src/security/PeerOptions.cc
```diff
@@ -297,130 +297,130 @@ static struct ssl_option {
```
rousskov, line 297: Please sync the ssl_option `value` type with OpenSSL v3 [changes](https://github.com/openssl/openssl/commit/f04bb0bce490de847ed0482b8ec9eabedd173852) to become uint64_t instead of long. Please try to avoid duplicating that code/type in Squid code -- either name it (e.g., ssl_option::Value) or, if there are only a couple of occurrences in low-level code, try using `decltype` for the second occurrence.

rousskov (changes requested):

src/security/PeerOptions.cc
yadij: Done in 3815b94d14858c503ad67adcaec8d98b1c1d5b24

src/security/PeerOptions.cc
yadij: > Finally we are starting to make progress, thank you!
> 

Does this mean you finally accept that I know what I'm doing here?



> I suggest using this approach:
> 
>     1. Use `#if defined(SSL_OP_X)` to test the existence of an OpenSSL `SSL_OP_X` macro (during preprocessing). The PR already contains all the necessary changes for this AFAICT.
> 

Indeed. The code you objected to has been doing this from the start.

>     2. Use `if (SSL_OP_X)` or equivalent to test whether the corresponding OpenSSL feature/option/etc. is actually enabled/supported by OpenSSL (during compilation). The current PR is missing those changes, but they are fairly easy to add.

The existing code in Security::PeerOptions::parseOptions() already implements this is a far simpler way than you are proposing.

src/security/PeerOptions.cc
yadij: Done in [3815b94](https://github.com/squid-cache/squid/pull/694/commits/3815b94d14858c503ad67adcaec8d98b1c1d5b24)

kinkie:
Everyone, please let's up the priority of merging this.
As of now, Squid will not build in strict mode on Centos stream 9, Debian unstable, Fedora 36 and rawhide, Opensuse tumbleweed and Ubuntu jammy

rousskov (changes requested):

yadij:
> > As of now, Squid will not build in strict mode on Centos stream 9, Debian unstable, Fedora 36 and rawhide, Opensuse tumbleweed and Ubuntu jammy
> 
> Does this PR code build (in strict mode) in those environments?

For Debian and Ubuntu yes it does.

Since the issues are with libssl API vs Squid I assume the others are the same. At least close enough that we can rely on those communities and Jenkins (after node upgrades) to find any remaining nits.

src/security/PeerOptions.cc
```diff
@@ -293,134 +293,134 @@ Security::PeerOptions::createClientContext(bool setOptions)
 /// set of options we can parse and what they map to
 static struct ssl_option {
     const char *name;
-    long value;
+    uint64_t value;
```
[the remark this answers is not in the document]
yadij: As you can see in 3815b94d14858c503ad67adcaec8d98b1c1d5b24 there is exactly **one** use of this type in Squid. It is not duplicated at all, let alone "badly".

src/security/PeerOptions.cc
yadij: The OP request:
 > Could you please clarify why these preprocessor conditions should be changed?

Has been answered already several times.

As for the subsequent request to "use Optional for value". That is requesting a behavioural change in Squid unrelated to the bug being fixed by this PR. I am leaving it out of scope here.

rousskov (changes requested):

src/security/PeerOptions.cc
yadij: The current official code treats missing array entries as ```value=0``` and produces the ```"ERROR: Unknown TLS option```.

This PR allows 0-value to exist in the array. Regardless of that existence their parsed ```value=0``` property stays unchanged and they produce the ```"ERROR: Unknown TLS option```. Exactly as if they were not listed in the array at all.

AIUI, your request is to use Optional to perform some **new** error handling (eg different debugs message or throw) for the 0-value vs undefined option name. That is unnecessary.

doc/release-notes/release-6.sgml
```diff
@@ -63,7 +63,9 @@ This section gives an account of those changes in three categories:
 <sect1>Changes to existing directives<label id="modifieddirectives">
 <p>
 <descrip>
-	<p>There have been no directives changed.
+	<tag>ssl_engine</tag>
+	<p>OpenSSL 3.0.0 removes the Engine feature. This directive is
+	   only supported when Squid is built for older OpenSSL versions.
```
rousskov, line 69: N.B. This is a copy the last [comment](https://github.com/squid-cache/squid/pull/694#discussion_r769077272
) from an earlier change request that was silently "resolved" for some (unknown to me) reason. I crossed out item 3 below because that change has been done, but we still have not even established whether engine support should be deprecated/disabled (a prerequisite for those to-do items). I will now re-resolve that outdated change request so that we have only one conversation to worry about.

> The ENGINE API is not available in the libssl3 I am building/testing against.

The limitations specific to the library you use should not affect the decision regarding deprecation of a Squid feature that may be used by others. Perhaps you can rephrase the above statement in more general terms (e.g., OpenSSL v3 that comes by default with Ubuntu X disables ENGINE API support)? A _general_ statement about the necessary library availability could influence our decision to deprecate, of course.

> I have never seen this option used by anyone. So for now I have opted to deprecate the whole feature.

You have seen it used by bug [4112](https://bugs.squid-cache.org/show_bug.cgi?id=4112) reporter (at least) AFAICT.

> If anyone really wants to have an ENGINE with libssl3 they will need to refactor all the code making use of this config option and the variables it sets up.

The cost of support is a valid factor in our decision making process, of course. Can you give an example of Squid code that will need to be changed in order to preserve support for OpenSSL engine selection via `ssl_engine` directive in Squid built with OpenSSL v3? My superficial interpretation of OpenSSL v3 migration [documentation](https://www.openssl.org/docs/man3.0/man7/migration_guide.html) is that there should be no Squid changes as long as the OpenSSL library is built with that deprecated API/code enabled.

----

Assuming the continued support is indeed costly (which is to be determined above), one could address this change request by making the following changes:

1. Adjust PR description to document that we are deprecating OpenSSL Engine support (not because it is removed in OpenSSL v3 but) because it is deprecated in OpenSSL v3, popular distributions ship OpenSSL v3 without Engine support (if that is true), and we do not have the resources to update the corresponding Squid code to remain compatible with OpenSSL v3 (if the code needs updating).
2. Add deprecation warning (for OpenSSL before v3) to squid.conf parser or interpreter. You have already adjusted the code to fail with OpenSSL v3+.
3. ~Replace the current (incorrect) claim in cf.data.pre with something along these lines: "Not supported in builds with OpenSSL v3 or newer."~

There are other ways to address this change request, of course. The above is just an example that illustrates the missing pieces if we decide to deprecate `ssl_engine` support starting with OpenSSL v3 builds.

rousskov (changes requested):

src/cf.data.pre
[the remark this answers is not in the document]
rousskov: Resolving as moved to https://github.com/squid-cache/squid/pull/694#discussion_r892425293

src/security/ServerOptions.cc
```diff
@@ -372,9 +376,36 @@ Security::ServerOptions::loadDhParams()
...
+        type = "EC";
+
+    if (auto *dctx = OSSL_DECODER_CTX_new_for_pkey(&pkey, "PEM", nullptr, type, 0, nullptr, nullptr)) {
+        if (auto *in = fopen(dhParamsFile.c_str(), "r")) {
+            if (OSSL_DECODER_from_fp(dctx, in) == 1) {
+
+                /* pkey is created with the decoded data from the bio */
+                Must(pkey);
+                parsedDhParams.resetWithoutLocking(pkey);
+                // TODO: verify that the loaded parameters match the value in eecdhCurve
```
rousskov, line 394: The official code uses DH_check() for that. Can we just do the same (or something similar) here? Sorry if you have already explained why this TODO is not implemented in this PR, breaking OpenSSL v1 feature parity -- please feel free to refer me to that explanation.

rousskov (changes requested):
Another partial review, now focusing on another area of the code.

src/security/ServerOptions.cc
yadij: Please do.

What needs ding is verifying that parsedDhParams object contains EC parameters that map to the Curve named in eecdhCurve string.

Last I looked there was a complications:
 * some parameters have multiple Curve names. So a reverse mapping check may not result in the original name.
 *  it is valid to load a file which has custom parameters that do not map to any of the well-known Curve names. So we may need to define a custom eecdhCurve string to skip the check for these edge cases.

src/security/ServerOptions.cc
```diff
@@ -372,9 +376,36 @@ Security::ServerOptions::loadDhParams()
...
     parsedDhParams.resetWithoutLocking(dhp);
+
+#else // OpenSSL 3.0+
+    EVP_PKEY *pkey = nullptr;
+    const char *type = "DH";
+    if (!eecdhCurve.isEmpty())
+        type = "EC";
+
+    if (auto *dctx = OSSL_DECODER_CTX_new_for_pkey(&pkey, "PEM", nullptr, type, 0, nullptr, nullptr)) {
+        if (auto *in = fopen(dhParamsFile.c_str(), "r")) {
```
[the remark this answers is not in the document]
yadij: Done in [0cf59cf](https://github.com/squid-cache/squid/pull/694/commits/0cf59cff679738c366372a0e3c729f78b879570e)

src/security/ServerOptions.cc
```diff
@@ -372,9 +376,36 @@ Security::ServerOptions::loadDhParams()
...
+
+            } else {
+                debugs(83, DBG_IMPORTANT, "WARNING: Failed to decode " << type << " parameters '" << dhParamsFile << "'");
+            }
+            fclose(in);
+        }
+        OSSL_DECODER_CTX_free(dctx);
+
+    } else {
+        debugs(83, DBG_IMPORTANT, "WARNING: no suitable potential decoders found for " << type << " parameters");
```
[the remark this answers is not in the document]
yadij: Thanks. Fixed in [14e21ad](https://github.com/squid-cache/squid/pull/694/commits/14e21ad13d332c878a7de03d0cc1b59e3758be12)

src/security/ServerOptions.cc
rousskov: OK, this change request is on my side.

doc/release-notes/release-6.sgml
yadij: > N.B. This is a copy the last [comment](https://github.com/squid-cache/squid/pull/694#discussion_r879784328) from an earlier [change request](https://github.com/squid-cache/squid/pull/694#discussion_r769077272) that was silently "resolved" for some (unknown to me) reason. 

github says that thread was:
> rousskov marked this conversation as resolved.

src/security/PeerOptions.cc
yadij: I do not see the issue you are ...

 * What you describe as a "new problem" is the same behaviour which has **always** occured on every Squid + OpenSSL upgrade. When value==0 for any option Squid rejects the squid.conf text. This is true regardless of how value became 0.

 * This PR is the implementation which **avoids** producing "new errors".

Can you please point me at the exact line sequence where you are seeing these alleged "new problems" ?

doc/release-notes/release-6.sgml
yadij: Do you find this working clearer?
```
 OpenSSL 3.0.0 replaces the Engine feature Providers (not yet supported by Squid).
 This directive is only supported when Squid is built for older OpenSSL versions.
```

src/security/ServerOptions.cc
rousskov: Status update: I have the validation code for DH parameters, but there is a problem with the surrounding code that should be fixed before the validation code can run. I am polishing that fix as well and hope to commit all the changes in 24-48 hours.

doc/release-notes/release-6.sgml
rousskov: > github says that "rousskov marked this conversation as resolved"

Yes, IIRC, I unresolved the old (already incorrectly resolved at that time) thread, but decided to move the conversation here instead, and re-resolved the old thread after the move. What you are seeing is the latest state. Unfortunately, these GitHub pages do not reflect the full resolution history (AFAICT), just its outcome.

> Do you find this working clearer?

As far as clarity is concerned, a word "with" appears to be missing after "feature" in that proposed text, but, more importantly, I do not understand your resistance to telling folks the truth about OpenSSL deprecation of the Engine feature. The feature is still available in OpenSSL v3. It is deprecated. We decided to drop its support now because, in part, keeping that support required significant[^1] work. There is nothing wrong with any of that if it is all true! We do not need to hide any of those (potentially important to some Squid admins) facts behind vague phrasing. We can and should be open/direct here IMO.

[^1]: We have not yet established that the amount of work to keep Engines support is significant, but I assume, for the sake of the argument, that we will establish that.

yadij: > I do not understand your resistance to telling folks the truth about OpenSSL deprecation of the Engine feature. The feature is still available in OpenSSL v3. It is deprecated.

The API used by Squid to load an ENGINE is not deprecated. It is **gone**.

AFAICT, what has been deprecated but still works are the APIs for ENGINES to pass data/object pointers to-from libssl. **after** being refactored to load through the Provider API old ENGINE's can continue to use the deprecated APIs to exchange data.  None of that has anything to do with Squid.

None of these OpenSSL-internal technicalities are of any consequence to Squid admin.

kinkie: Would it make sense to deprecate the whole config option instead of dealing with the complexity?
If it's an useless internal technicality, we could ignore-and-warn

rousskov: > Would it make sense to deprecate the whole config option instead of dealing with the complexity?

Yes, that is item 2 on the list in this change request (after the complexity claim has been substantiated).

The cost of satisfying the preconditions for that deprecation has been artificially inflated by lack of cooperation, but we are making progress and will get there eventually.

>  If it's an useless internal technicality, we could ignore-and-warn

Justification of a (known to be used in the past) feature deprecation is not a useless technicality. We ought to be clear about such a deprecation (internally and externally). I have been [asking](https://github.com/squid-cache/squid/pull/694#discussion_r769077272) for that clarity since December 2021, but, again, we are making progress and will eventually get there.

src/security/ServerOptions.cc
rousskov: > Alex: we should at least try to preserve the feature parity with OpenSSL v1 in this context.

Done in e794b46. I am resolving this change request. While working on this changes, I discovered and fixed a few related problems. Please see commit messages for details, especially 75bd3aa and f2830c5.

> Amos: What needs doing is verifying that parsedDhParams object contains EC parameters that map to the Curve named in eecdhCurve string

AFAICT, all of that (and associated complications) are outside this PR scope because OpenSSL v1 code was not verifying any of that either.

src/security/ServerOptions.cc
```diff
@@ -372,9 +376,36 @@ Security::ServerOptions::loadDhParams()
...
+        if (auto *in = fopen(dhParamsFile.c_str(), "r")) {
+            if (OSSL_DECODER_from_fp(dctx, in) == 1) {
+
+                /* pkey is created with the decoded data from the bio */
+                Must(pkey);
+                parsedDhParams.resetWithoutLocking(pkey);
+                // TODO: verify that the loaded parameters match the value in eecdhCurve
+
+            } else {
+                debugs(83, DBG_IMPORTANT, "WARNING: Failed to decode " << type << " parameters '" << dhParamsFile << "'");
```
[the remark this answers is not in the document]
rousskov: For the record, my (partially implemented) suggestion in this change request was wrong. OSSL_DECODER_CTX_new_for_pkey() API is even worse than its documentation implies! I hope I fixed our code in branch commit f2830c59ba15e989ff09f941b6065ba3aacc8415.

src/security/forward.h
```diff
@@ -170,11 +185,11 @@ class KeyData;
 class KeyLog;
 
 #if USE_OPENSSL
-typedef long ParsedOptions;
+typedef uint64_t ParsedOptions;
 #elif USE_GNUTLS
 typedef std::shared_ptr<struct gnutls_priority_st> ParsedOptions;
 #else
-class ParsedOptions {}; // we never parse/use TLS options in this case
+typedef uint64_t ParsedOptions;
```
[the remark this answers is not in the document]
yadij: You requested that ssl_options::value use the correct type for parsed options. Compiler does not permit the empty-class to be used for a struct member.

src/ssl/support.cc
```diff
@@ -653,8 +657,8 @@ Ssl::Initialize(void)
 
     SQUID_OPENSSL_init_ssl();
 
-#if !defined(OPENSSL_NO_ENGINE)
     if (::Config.SSL.ssl_engine) {
+#if !defined(OPENSSL_NO_ENGINE) && OPENSSL_VERSION_MAJOR < 3
```
yadij, line 661: Remove this version check to see why ```ssl_engine``` directive is deprecated.

```suggestion
#if !defined(OPENSSL_NO_ENGINE)
```

doc/release-notes/release-6.sgml
yadij: As stated already several times there is no OpenSSL API replacement to fix these:
```
../../../src/ssl/support.cc: In function 'void Ssl::Initialize()':
../../../src/ssl/support.cc:662:36: error: 'void ENGINE_load_builtin_engines()' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
  662 |         ENGINE_load_builtin_engines();
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
In file included from ../../../src/ssl/support.h:30,
                 from ../../../src/AccessLogEntry.h:32,
                 from ../../../src/acl/FilledChecklist.h:12,
                 from ../../../src/ssl/support.cc:18:
/usr/include/openssl/engine.h:358:28: note: declared here
  358 | OSSL_DEPRECATEDIN_3_0 void ENGINE_load_builtin_engines(void);
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../src/ssl/support.cc:664:31: error: 'ENGINE* ENGINE_by_id(const char*)' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
  664 |         if (!(e = ENGINE_by_id(::Config.SSL.ssl_engine)))
      |                   ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../../../src/ssl/support.h:30,
                 from ../../../src/AccessLogEntry.h:32,
                 from ../../../src/acl/FilledChecklist.h:12,
                 from ../../../src/ssl/support.cc:18:
/usr/include/openssl/engine.h:336:31: note: declared here
  336 | OSSL_DEPRECATEDIN_3_0 ENGINE *ENGINE_by_id(const char *id);
      |                               ^~~~~~~~~~~~
../../../src/ssl/support.cc:667:32: error: 'int ENGINE_set_default(ENGINE*, unsigned int)' is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
  667 |         if (!ENGINE_set_default(e, ENGINE_METHOD_ALL)) {
      |              ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
In file included from ../../../src/ssl/support.h:30,
                 from ../../../src/AccessLogEntry.h:32,
                 from ../../../src/acl/FilledChecklist.h:12,
                 from ../../../src/ssl/support.cc:18:
/usr/include/openssl/engine.h:708:27: note: declared here
  708 | OSSL_DEPRECATEDIN_3_0 int ENGINE_set_default(ENGINE *e, unsigned int flags);
      |                           ^~~~~~~~~~~~~~~~~~
```

yadij: I have now also reviewed the Providers API to see how it differs from ENGINE.

The conclusion I have come to is that we have two options:

 1)  in squid.conf add a new ```providers=`` option to each of the directives which create a TLS/SSL context (*_port, cache_peer, tls_outgoing_options) . Since the set of Providers loaded into an interface specific context determines what ```tls-options=``` are available and how they work on TLS connections using that context (eg ciphers available).

2) leave Provider support out of Squid entirely and encourage admin use openssl.conf to configure them. This IMO is better overall since a) engines and providers are mostly useful to add a hardware device or enable FIPS mode at the library level, and b) it reduces our own work and code supporting the feature.

kinkie: I would favour option 2, because it's simpler to implement and maintain, and it is more future-proof, as it will leave openssl in charge of configuring their own settings. At most we should reference this in our documentation

src/security/forward.h
```diff
@@ -101,9 +101,24 @@ typedef std::list<Security::CertPointer> CertList;
 typedef std::list<Security::CrlPointer> CertRevokeList;
 
 #if USE_OPENSSL
+CtoCpp1(EVP_PKEY_free, EVP_PKEY *)
+typedef Security::LockingPointer<EVP_PKEY, EVP_PKEY_free_cpp, HardFun<int, EVP_PKEY *, EVP_PKEY_up_ref> > PrivateKeyPointer;
+#elif USE_GNUTLS
+typedef std::shared_ptr<struct gnutls_x509_privkey_int> PrivateKeyPointer;
```
kinkie, line 107: nit: is there any reason for using typedef instead of using?

kinkie (commented):
Assuming the code works - I don't know openssl well enough to assess it, I see no reason not to ship this

src/security/forward.h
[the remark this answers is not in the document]
rousskov: > You requested that ssl_options::value use the correct type for parsed options.

AFAICT, `ssl_options` are specific to `USE_OPENSSL` code. I am asking about the change to the `#else` code line that is only compiled when `USE_OPENSSL` and `USE_GNUTLS` are falsy.

> Compiler does not permit the empty-class to be used for a struct member.

What makes you think that? I cannot think of a reason to prohibit that, and it is allowed in my quick [test](https://cpp.godbolt.org/z/YbPYj68qs).

src/security/PeerOptions.cc
rousskov: > Can you please point me at the exact line sequence where you are seeing these alleged "new problems" ?

Sure, the following `if` statement assumes that `!value` means "not found" when, in fact, it no longer means that in the PR code. In the PR code `!value` means "either not found or found but not supported".

https://github.com/squid-cache/squid/blob/c25648fb4eca515f8a6a535cc93c2ac2eccc292b/src/security/PeerOptions.cc#L490-492

I did not check for other occurrences, but you should.

rousskov (changes requested):
Updated my open change requests. Please let me know if you want me to address them.

doc/release-notes/release-6.sgml
yadij: In effort to make some progress a recap:

Alex:
> what would it take to preserve legacy Engines support while using OpenSSL v3.0?

Alex:
> Can you give an example of Squid code that will need to be changed in order to preserve support for OpenSSL engine selection via ssl_engine directive in Squid built with OpenSSL v3?


Amos:
> there is no OpenSSL API replacement to fix these:
> `OSSL_DEPRECATEDIN_3_0 void ENGINE_load_builtin_engines()`
> `OSSL_DEPRECATEDIN_3_0 ENGINE *ENGINE_by_id()`
> `OSSL_DEPRECATEDIN_3_0 int ENGINE_set_default()`
> 

Alex:
> I am not sure why you are stating the above. Your statement does not seem to invalidate or address this change request.

You elided the important part of what I quoted ... The fact that those symbols use produce fatal build errors.

src/security/forward.h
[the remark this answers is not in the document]
yadij: Odd. There was a compile error when I first changed the code. It seems to be gone now so reverted in [9b8eb3d](https://github.com/squid-cache/squid/pull/694/commits/9b8eb3d2aa345a9185abdc23c82c668e6a31561e)

src/security/PeerOptions.cc
yadij: Bit of a stretch there (none of the array options match hex numeric format). But okay, changed in [5a875b2](https://github.com/squid-cache/squid/pull/694/commits/5a875b21d2ae69d423239cdce0ba170a9c76671f)

src/ssl/gadgets.cc
```diff
@@ -421,7 +414,11 @@ mimicExtensions(Security::CertPointer & cert, Security::CertPointer const &mimic
     // XXX: Add PublicKeyPointer. In OpenSSL, public and private keys are
     // internally represented by EVP_PKEY pair, but GnuTLS uses distinct types.
     const Security::PrivateKeyPointer certKey(X509_get_pubkey(mimicCert.get()));
-    const auto rsaPkey = EVP_PKEY_get0_RSA(certKey.get()) != nullptr;
+#if OPENSSL_VERSION_MAJOR < 3
+    const auto rsaPkey = bool(EVP_PKEY_get0_RSA(certKey.get()));
```
rousskov, line 418: I think the original code relays the intent better. There is no need to change it here AFAICT:

```suggestion
    const auto rsaPkey = EVP_PKEY_get0_RSA(certKey.get()) != nullptr;
```

src/ssl/gadgets.cc
```diff
@@ -421,7 +414,11 @@ mimicExtensions(Security::CertPointer & cert, Security::CertPointer const &mimic
...
+#else
+    const auto rsaPkey = EVP_PKEY_is_a(certKey.get(), "RSA");
```
rousskov, line 420: To make the type of both rsaPkey variables `bool`, I suggest:

```suggestion
    const auto rsaPkey = EVP_PKEY_is_a(certKey.get(), "RSA") == 1;
```

or similar. I do not insist on this change.

rousskov (changes requested):
During this review round, I looked through the entire PR and opened a few (small) new change requests. I also resolved or updated old change requests. My offer to promptly address all open change requests still stands.

kinkie:
@yadij, @rousskov , can we please prioritise getting this through over everything else? It's blocking a key feature on any modern Linux distro unless the admin wishes to go jump through hoops

rousskov:
> @yadij, @rousskov , can we please prioritise getting this through over everything else?

FWIW, I have been doing exactly what you are asking for, including providing prompt reviews, contributing fixes, and spinning off helper PRs. My offer to address all known issues still stands.

kinkie:
> > @yadij, @rousskov , can we please prioritise getting this through over everything else?
> 
> FWIW, I have been doing exactly what you are asking for, including providing prompt reviews, contributing fixes, and spinning off helper PRs. My offer to address all known issues still stands.

Thank you for that. I just noticed a week had passed since the last comment. @yadij . I'm happy to jump in if you believe there is any benefit in me doing so

src/security/PeerOptions.cc
```diff
@@ -481,14 +482,16 @@ Security::PeerOptions::parseOptions()
...
             }
 
             // Special case.. hex specification
             ::Parser::Tokenizer tmp(option);
-            if (!value && tmp.int64(hex, 16, false) && tmp.atEnd()) {
+            if (!found && tmp.int64(hex, 16, false) && tmp.atEnd()) {
                 value = hex;
+                found = true;
             }
         }
```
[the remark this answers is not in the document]
yadij: FWIW, I am not against the idea of making this array better and will probably waive through good quality PRs doing it. I really just think it is complexity, not *necessary* for this PR scope.

src/ssl/gadgets.cc
```diff
@@ -48,33 +48,26 @@ ThrowErrors(const char * const problem, const int savedErrno, const SourceLocati
                         where);
 }
 
-EVP_PKEY * Ssl::createSslPrivateKey()
+static EVP_PKEY *
+CreateRsaPrivateKey()
```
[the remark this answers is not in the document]
yadij: I am okay with that change but unlikely to be able to do it anytime soon. If you of you could take it on that would be great.

src/security/forward.h
yadij: @rouskov what do you mean by thumbs down on my answer? its true.

yadij:
I am away from home with covid at present. This is still top-priority for me, but all I can do directly is use github web API or email until thats over.

@kinkie fresh ideas about what the release notes file should say might break that cycle. I have also indicated the latest review items that need coding I cannot do right now in those threads. Updates there would be useful progress.

src/ssl/gadgets.cc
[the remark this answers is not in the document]
rousskov: Sure, will do.

src/security/PeerOptions.cc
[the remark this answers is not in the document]
rousskov: This change request is not about the array. It is about the (value, found) pair (which is essentially added by this PR). Addressing this change request will not make the code more complex.

src/security/forward.h
rousskov: > what do you mean by thumbs down on my answer? its true.

While the statement in your answer may be correct, the answer does not address the problem implied by the change request. In this specific case, this is a very minor potential problem in a long-suffering PR, so I did not do more about it, but the overall flaw is very common -- folks are focusing on defending their actions/code/words instead of focusing on improving the PR. In my estimates, our PR process efficiency is lower than 30% -- most of the time spent on PRs is wasted. This amount of waste is ridiculous IMO.

I hope my [EasierDoneThanSad](https://squid.zulipchat.com/#narrow/stream/314059-core/topic/Easier.20Done.20.28Than.20Said.29/near/286106470) experiment with Francesco will improve things, but it cannot solve many of the problems, and I suspect you are [against](https://github.com/squid-cache/squid/pull/864#issuecomment-1100425996) that kind of collaboration anyway.

src/ssl/gadgets.cc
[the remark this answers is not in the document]
rousskov: Done in 94601d6.

src/security/PeerOptions.cc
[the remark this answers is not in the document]
rousskov: I would be happy to address this change request, demonstrating that the improved code will remain simple. Just let me know if you want me to do that. I still do not insist on this change.

doc/release-notes/release-6.sgml
```diff
@@ -111,7 +113,9 @@ This section gives an account of those changes in three categories:
 <sect1>Changes to existing options<label id="modifiedoptions">
 <p>
 <descrip>
-	<p>There have been no options changed.
+	<tag>--with-openssl</tag>
+	<p>OpenSSL features are now auto-enabled when OpenSSL 3.0.0
+	   or later is detected during build.
```
rousskov, line 118: AFAICT, branch commit 0f14f83d5e83e524291f4ba8ea2fac83e84722a3 removed the improvement summarized here. We should undo this SGML change AFAICT:

```suggestion
	<p>There have been no options changed.
```

rousskov (changes requested):
I think there are only three issues remain. Let me know if I should address them.

src/security/forward.h
yadij: Done in [d4a42e7](https://github.com/squid-cache/squid/pull/694/commits/d4a42e7ecfe4f517704031f6e7827ccb21512d08)

yadij:
As of [d4a42e7](https://github.com/squid-cache/squid/pull/694/commits/d4a42e7ecfe4f517704031f6e7827ccb21512d08) all the requests that have been insisted on are resolved.

There is just https://github.com/squid-cache/squid/pull/694#discussion_r913881575 which I leave to @rousskov to submit a PR making the change. Since that is not being insisted on include AFAICT this PR is ready to go in.

rousskov:
> AFAICT this PR is ready to go in.

I have no objections and have cleared this PR for merging.

> @[yadij](https://github.com/yadij) yadij requested a review from [chtsanti](https://github.com/chtsanti) 9 months ago
> @[yadij](https://github.com/yadij) yadij requested review from [hno](https://github.com/hno) 27 days ago

However, I will not dismiss @yadij review requests to avoid creating a dangerous precedent. @yadij, please cancel those requests if you believe those reviews are no longer required. Anubis will not merge while those review requests are pending.

wohlford:
Squid 5.7 (05 Sep 2022) officially contains support for OpenSSL 3.0.

PR #695, opened 2020-07-25
Fix handling of overly long squid.conf line continuations

ihsinme:
The addition of a check that excludes the possibility of an integer overflow with a subsequent impact on the integrity of memory in the heap, with incorrect use of \\ characters in the configuration file.  
the / character makes it possible to infinitely increase the length of the processed string, which led to an integer overflow and subsequent heap overflow

### DIFF
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -431,7 +431,7 @@ parseOneConfigFile(const char *file_name, unsigned int depth)
     const int orig_config_lineno = config_lineno;
     char *token = NULL;
     char *tmp_line = NULL;
-    int tmp_line_len = 0;
+    size_t tmp_line_len = 0;
     int err_count = 0;
     int is_pipe = 0;
 
@@ -515,10 +515,10 @@ parseOneConfigFile(const char *file_name, unsigned int depth)
         const char* append = tmp_line_len ? skip_ws(config_input_line) : config_input_line;
 
         size_t append_len = strlen(append);
-        if(tmp_line_len+append_len >= std::numeric_limits<int>::max())
-            fatalf("Long line in config file");
-        tmp_line = (char*)xrealloc(tmp_line, tmp_line_len + append_len + 1);
-
+        if (const auto combinedSize = Sum(tmp_line_len, append_len, 1U))
+            tmp_line = static_cast<char*>(xrealloc(tmp_line, combinedSize.value()));
+        else
+            throw TextException("Exceedingly long line continuation", Here());
         strcpy(tmp_line + tmp_line_len, append);
 
         tmp_line_len += append_len;
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -515,7 +515,8 @@ parseOneConfigFile(const char *file_name, unsigned int depth)
         const char* append = tmp_line_len ? skip_ws(config_input_line) : config_input_line;
 
         size_t append_len = strlen(append);
-
+        if(tmp_line_len+append_len >= std::numeric_limits<int>::max())
+            fatalf("Long line in config file");
         tmp_line = (char*)xrealloc(tmp_line, tmp_line_len + append_len + 1);
 
         strcpy(tmp_line + tmp_line_len, append);


src/cache_cf.cc
```diff
@@ -515,7 +515,8 @@ parseOneConfigFile(const char *file_name, unsigned int depth)
         const char* append = tmp_line_len ? skip_ws(config_input_line) : config_input_line;
 
         size_t append_len = strlen(append);
-
+        if(tmp_line_len+append_len >= std::numeric_limits<int>::max())
+            fatalf("Long line in config file");
         tmp_line = (char*)xrealloc(tmp_line, tmp_line_len + append_len + 1);
```
rousskov, line 520: Please fix the `tmp_line_len` type to become `size_t` and then use the recently added (commit 7224761) Sum() to properly handle overflows. Here is an untested sketch for the second part:

```suggestion
        if (const auto combinedSize = Sum(tmp_line_len, append_len, 1U))
            tmp_line = static_cast<char*>(xrealloc(tmp_line, combinedSize.value()));
        else
            throw TextException("Exceedingly long line continuation", Here());
```

rousskov:
```
cache_cf.cc:533:45: error: invalid initialization of reference of type ‘int&’ from expression of type ‘size_t {aka long unsigned int}’
         ProcessMacros(tmp_line, tmp_line_len);
```

FYI: To fix this build [failure](https://semaphoreci.com/squid/squid/branches/pull-request-695/builds/2), you will need to fix the type of the second ProcessMacros() parameter and also adjust `len` declaration in gen_default(). Other adjustments may be needed as well. It is best to test before pushing changes into the public pull request.

To fix the other build failure, you will need to rebase your branch on top of the current master (at least). As you know, Sum() was added only recently. If you are not familiar with git, you might find [GitHints](https://wiki.squid-cache.org/GitHints) useful.

rousskov:
> which led to an integer overflow and subsequent heap overflow

How does that heap overflow manifest itself to a Squid admin that starts Squid with a configuration file that has excessively long continuation lines? What will that admin see on the terminal or in the logs? Think of an admin who suspects (but does not know for sure) that she is suffering from this bug or is googling for a bug solution using observable symptoms. Providing such symptoms in the commit message would allow such an admin to drastically narrow down the fixes she needs to test.

If we cannot offer a representative "what happens when the bug is triggered" sample for some good reason, that is OK! I am just hoping we can.

ihsinme:
FYI: To fix this build failure, you will need to fix the type of the second ProcessMacros() parameter and also adjust len declaration in gen_default(). Other adjustments may be needed as well. It is best to test before pushing changes into the public pull request.

sorry for the build failure, I relied on your suggestion and had no practical judgment. tomorrow I will try to put together a working code on your proposal.

ihsinme:
>What will that admin see on the terminal or in the logs?  

maybe I don't understand you correctly, but my correction is aimed at eliminating the error and the admin will not see anything.
by giving an example of a vulnerable configuration, I can unnecessarily confuse the reader, including because of the need to prepare different configs for different environments.
if you see an incomprehensible moment, perhaps point to it, I will explain it.

rousskov:
> my correction is aimed at eliminating the error and the admin will not see anything.

Yes, I know.

> by giving an example of a vulnerable configuration

Please do not give examples of vulnerable configurations.

Please give an example of Squid output (or some other artifact) when unfixed/official code tries to process a vulnerable configuration. What happens (from the Squid admin point of view) when the bug is triggered? Depending on how those artifacts look, we will either include a polished example in the future commit message or just leave it as a GitHub comment. Do not worry about the format for now.

> if you see an incomprehensible moment, perhaps point to it, I will explain it.

I see a potential lack of information, rather than something one can point to. I hope the above clarifies what information I am asking for.

ihsinme:
Using the Sum function will involve a deeper change, so it will take me a few days to change the typing of the affected functions and to evaluate how they work. if you think that changes are needed more quickly, I suggest returning to my first revision within the framework of the specified PR.

Regarding the display in case of erroneous operation, if we are talking about an accidental overflow, then a classic segmentation fault occurs, on most operating systems the overflow is prevented by the __strcpy_chk function and the program will be stopped with a description of buffer overflow detected

rousskov:
> Using the Sum function will involve a deeper change, so it will take me a few days

Please take your time.

> Regarding the display in case of erroneous operation, if we are talking about an accidental overflow, then a classic segmentation fault occurs, on most operating systems the overflow is prevented by the __strcpy_chk function and the program will be stopped with a description of buffer overflow detected

The signs of a segmentation fault are well-known, so we do not need to record an example for that IMO. I will try to remember to update the PR description to mention segmentation faults as one possible outcome.

If you can provide a specific output example for a "description of buffer overflow detected", please do so; it is understood that this will be just one example and other OSes/environments will have different (or no) output. If you cannot provide such an example, it is not the end of the world -- I just was under impression that you have seen it while working on this issue and can easily reproduce it to copy-paste into the PR description. Perhaps I misunderstood.

PR #696, opened 2020-07-25
Simplify parsing of domain names from the /etc/hosts file

ihsinme:
For different lengths of domain names, access situations outside
the allocated buffer buf2 are possible. At the same time, the
fix eliminates the possibility of losing the terminator after
executing the strncpy function.

### DIFF
--- a/src/tools.cc
+++ b/src/tools.cc
@@ -972,11 +972,11 @@ parseEtcHosts(void)
             debugs(1, 5, "etc_hosts: got hostname '" << lt << "'");
 
             /* For IPV6 addresses also check for a colon */
-            if (Config.appendDomain && !strchr(lt, '.') && !strchr(lt, ':')) {
+            if (Config.appendDomain && strlen(Config.appendDomain)<sizeof(buf2)-2 && !strchr(lt, '.') && !strchr(lt, ':')) {
                 /* I know it's ugly, but it's only at reconfig */
-                strncpy(buf2, lt, sizeof(buf2)-1);
-                strncat(buf2, Config.appendDomain, sizeof(buf2) - strlen(lt) - 1);
-                buf2[sizeof(buf2)-1] = '\0';
+                memset(buf2,0,sizeof(buf2));
+                strncpy(buf2, lt, sizeof(buf2)-strlen(Config.appendDomain)-2);
+                strncat(buf2, Config.appendDomain,  sizeof(buf2) - strlen(buf2) );
                 host = buf2;
             } else {
                 host = lt;
--- a/src/tools.cc
+++ b/src/tools.cc
@@ -970,8 +970,8 @@ parseEtcHosts(void)
             *nt = '\0';
             debugs(1, 5, "etc_hosts: got hostname '" << lt << "'");
 
-            (void)urlAppendDomain(lt);
-            host = lt;
+            if(urlAppendDomain(lt))
+                host = lt;
             if (ipcacheAddEntryFromHosts(host, addr) != 0) {
                 /* invalid address, continuing is useless */
                 hosts.clear();
--- a/src/tools.cc
+++ b/src/tools.cc
@@ -911,7 +911,6 @@ void
 parseEtcHosts(void)
 {
     char buf[1024];
-    char buf2[512];
     char *nt = buf;
     char *lt = buf;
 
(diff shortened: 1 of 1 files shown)


src/tools.cc
```diff
@@ -972,11 +972,11 @@ parseEtcHosts(void)
             debugs(1, 5, "etc_hosts: got hostname '" << lt << "'");
 
             /* For IPV6 addresses also check for a colon */
-            if (Config.appendDomain && !strchr(lt, '.') && !strchr(lt, ':')) {
+            if (Config.appendDomain && strlen(Config.appendDomain)<sizeof(buf2)-2 && !strchr(lt, '.') && !strchr(lt, ':')) {
```
yadij, line 975: In latest Squid we have a helper function to handle this append directive more safely.

```suggestion
            (void)urlAppendDomain(lt);
```

src/tools.cc
```diff
@@ -972,11 +972,11 @@ parseEtcHosts(void)
...
                 /* I know it's ugly, but it's only at reconfig */
-                strncpy(buf2, lt, sizeof(buf2)-1);
-                strncat(buf2, Config.appendDomain, sizeof(buf2) - strlen(lt) - 1);
-                buf2[sizeof(buf2)-1] = '\0';
+                memset(buf2,0,sizeof(buf2));
+                strncpy(buf2, lt, sizeof(buf2)-strlen(Config.appendDomain)-2);
+                strncat(buf2, Config.appendDomain,  sizeof(buf2) - strlen(buf2) );
                 host = buf2;
             } else {
                 host = lt;
```
yadij, line 982: After the if-statement replacement all this becomes:

```suggestion
            host = lt;
```

... and remove the } below, which for some reason github will not let me add to the suggested patch.

yadij:
There is also internalHostname() performing an unsafe append in src/internal.cc that can be replaced with:
> (void)urlAppendDomain(host);

ihsinme:
your proposed changes are limited to 256 bytes, which probably requires revision and local buffers with sizes of 512 and 1024 bytes. additionally, I'm not sure if the 256 cutoff applies correctly to domain names.
therefore I propose to accept my suggestions as an overflow solution.

yadij:
The buffers at present are designed to be a large amount oversized for the values stored. The append function is designed to enforce the name limitation found throughout Squid which the resulting names will have to fit within anyway. With correct detection and enforcement of overflows we no longer need the excess buffer memory to cope with mistakes.

NP: 256 is just the default for SQUIDHOSTNAMELEN if there is no system provided limitation, or alternative provided during build. So the actual value may be larger even than the 512 those buffers cope with.

ihsinme:
applying the append function as you suggest will severely limit the size of the domains processed. If this is acceptable, I am ready to agree and accept your changes.

yadij:
It is acceptable. People having issues with this limit need to extend it for other parts of Squid to operate properly on the resulting domains. Best to have it handled early on startup/reconfigure than later during client transactions.

ihsinme:
I was only able to make changes with a new fix.

yadij:
OK to test

src/tools.cc
```diff
@@ -971,17 +970,8 @@ parseEtcHosts(void)
...
-                /* I know it's ugly, but it's only at reconfig */
-                strncpy(buf2, lt, sizeof(buf2)-1);
-                strncat(buf2, Config.appendDomain, sizeof(buf2) - strlen(lt) - 1);
-                buf2[sizeof(buf2)-1] = '\0';
-                host = buf2;
-            } else {
-                host = lt;
-            }
-
+            (void)urlAppendDomain(lt);
```
rousskov, line 973: Why are we ignoring the false result of the urlAppendDomain() call? Does not it indicate a problem worth handling here?

ihsinme: urlAppendDomain is a complete replacement for ifelse. false response is unnecessary at this point.

rousskov: AFAICT, your "we are preserving the old behavior" answer assumes that the old code correctly ignored failures. Let's forget what the (buggy) old code did for a second. Can you explain why ignoring failures is correct in this context? As a Squid administrator, I would expect Squid to _at least_ warn me that my `append_domain` directive was not honored. What's wrong with those expectations?

ihsinme: my fix aims to eliminate the overflow as soon as possible. and we have been discussing for 3 days how best to fix it. It seems to me that the correct practice is to quickly fix the vulnerable section of the code and then calmly develop and improve the code with discussion.

rousskov: AFAICT, there is no rush with this fix -- there is no emergency here. Please take your time to "calmly develop" a quality fix. As for the best practices, they vary from project to project, and even from reviewer to reviewer. If you want to know my personal views, you can find them [elsewhere](https://github.com/measurement-factory/squid-notes/blob/start/review.md).

yadij: FWIW, the append function already handles reporting that domain too long and by how much. The false result is for the caller if it wants to do any further handling eg error, fatal, context message explaining what the domain is being used for, etc.

rousskov: > the append function already handles reporting that domain too long and by how much

FYI: urlAppendDomain() reports nothing to Squid administrators. Its level-2 debugging is irrelevant in the context of this change request.

It is possible that urlAppendDomain() itself should do more than it does now (e.g., throw an exception), but I have not checked all the details to recommend a specific solution. In many cases like this one, it is best to fix the current caller first and then look at whether all callers or the function should be adjusted to unify handling. Both steps should be done in one PR, but splitting this work into two steps may help new Squid developers to make steady progress.

And just to avoid misunderstand, this change request is _not_ limited to reporting. I suspect problematic `/etc/host` should be treated as a fatal configuration problem. Admins can use hosts_file to supply a working alternative if they cannot fix the original.

PR #699, opened 2020-07-28
Do not stall while debugging a scan of an empty store_table

eduard-bagdasaryan:
Non-SMP Squid and each SMP kid allocate a store_table hash. With large
caches, some allocated store_table may have millions of buckets.

Recently we discovered that it is almost impossible to debug SMP Squid
with a large but mostly empty disk cache because the disker registration
times out while store_table is being built -- the disker process is
essentially blocked on a very long debugging loop.

The code suspends the loop every 500 entries (to take care of tasks like
kid registration), but there are no pauses when scanning millions of
empty hash buckets where every bucket prints two debug lines.

Squid now does not report empty store_table buckets explicitly. When
dealing with large caches, the debugged process may still be blocked for
a few hundred milliseconds (instead of many seconds) while scanning the
entire (mostly empty) store_table. Optimizing that should be done as a
part of the complex "store search" API refactoring.

### DIFF
--- a/src/store/LocalSearch.cc
+++ b/src/store/LocalSearch.cc
@@ -91,7 +91,6 @@ Store::LocalSearch::copyBucket()
 {
     /* probably need to lock the store entries...
      * we copy them all to prevent races on the links. */
-    debugs(47, 3, "Store::LocalSearch::copyBucket #" << bucket);
     assert (!entries.size());
     hash_link *link_ptr = NULL;
     hash_link *link_next = NULL;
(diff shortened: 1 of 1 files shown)


yadij:
Fixed description line wrapping, no text changes.

rousskov:
> Fixed description line wrapping, no text changes.

FYI: The wrapping was probably fine. I rewrapped at 72 columns. The problem was a unicode dash character: `–` instead of ASCII `--`. I fixed that.

rousskov:
@kinkie, please check a Jenkins optimization [problem](https://build.squid-cache.org/job/5-pr-test/2444/COMPILER=clang,OS=debian-stable,label=docker-build-host/console):

```
fatal: destination path 'support-tools' already exists and is not an empty directory.
```

If possible, please restart the test after addressing that problem, to get an all-green PR.

rousskov:
> If possible, please restart the test after addressing that problem, to get an all-green PR.

FTR: I restarted the test in hope that this problem will go away on its own.

PR #701, opened 2020-07-31
Forbid obs-fold and bare CR whitespace in framing header fields

yadij:
Header folding has been used for various attacks on HTTP proxies and
servers. RFC 7230 prohibits sending obs-fold (in any header field) and
allows the recipient to reject messages with folded headers. To reduce
the attack space, Squid now rejects messages with folded Content-Length
and Transfer-Encoding header field values. TODO: Follow RFC 7230 status
code recommendations when rejecting.

Bare CR is a CR character that is not followed by a LF character.
Similar to folding, bare CRs has been used for various attacks. HTTP
does not allow sending bare CRs in Content-Length and Transfer-Encoding
header field values. Squid now rejects messages with bare CR characters
in those framing-related field values.

When rejecting, Squid informs the admin with a level-1 WARNING such as

    obs-fold seen in framing-sensitive Content-Length: ...

### DIFF
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -372,6 +372,9 @@ HttpHeader::parse(const char *buf, size_t buf_len, bool atEnd, size_t &hdr_sz, H
     return -1;
 }
 
+// XXX: callers treat this return as boolean.
+// XXX: A better mechanism is needed to signal different types of error.
+//      lexicon, syntax, semantics, validation, access policy - are all (ab)using 'return 0'
 int
 HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthInterpreter &clen)
 {
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -496,7 +496,8 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         if (lines > 1 || hasBareCr) {
             const auto framingHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
             if (framingHeader) {
-                debugs(55, warnOnError, "WARNING: " << (hasBareCr ? hasBareCr : "obs-fold") << " seen in framing-sensitive " << e->name << ": " << e->value);
+                if (!hasBareCr) // already warned about bare CRs
+                    debugs(55, warnOnError, "WARNING: " << "obs-fold in framing-sensitive " << e->name << ": " << e->value);
                 delete e;
                 PROF_stop(HttpHeaderParse);
                 clean();
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1773,3 +1773,4 @@ HttpHeader::removeConnectionHeaderEntries()
             refreshMask();
     }
 }
+
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -443,7 +443,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
 
             /* Barf on stray CR characters */
             if (memchr(this_line, '\r', field_end - this_line)) {
-                hasBareCr = "bare-CR";
+                hasBareCr = "bare CR";
                 debugs(55, warnOnError, "WARNING: suspicious CR characters in HTTP header {" <<
                        getStringPrefix(field_start, field_end-field_start) << "}");
 
@@ -1773,4 +1773,3 @@ HttpHeader::removeConnectionHeaderEntries()
             refreshMask();
     }
 }
-
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -400,11 +400,11 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         const char *field_start = field_ptr;
         const char *field_end;
 
-        int seenLines = 0;
+        size_t lines = 0;
         do {
             const char *this_line = field_ptr;
             field_ptr = (const char *)memchr(field_ptr, '\n', header_end - field_ptr);
-            ++seenLines;
+            ++lines;
 
             if (!field_ptr) {
                 // missing <LF>
@@ -488,8 +488,8 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             return 0;
         }
 
-        const bool frameHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
-        if (seenLines > 1 && frameHeader) {
+        const auto framingHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
+        if (lines > 1 && framingHeader) {
             debugs(55, warnOnError, "WARNING: obs-fold seen in " << e->name << ": " << e->value);
             delete e;
             PROF_stop(HttpHeaderParse);
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -498,7 +498,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         if (lines > 1 || hasBareCr) {
             const auto framingHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
             if (framingHeader) {
-                debugs(55, warnOnError, "WARNING: " << (lines > 1 ? "obs-fold" : hasBareCr) << " seen in " << e->name << ": " << e->value);
+                debugs(55, warnOnError, "WARNING: " << (hasBareCr ? hasBareCr : "obs-fold") << " seen in framing-sensitive " << e->name << ": " << e->value);
                 delete e;
                 PROF_stop(HttpHeaderParse);
                 clean();
@@ -1775,4 +1775,3 @@ HttpHeader::removeConnectionHeaderEntries()
             refreshMask();
     }
 }
-
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -400,9 +400,11 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         const char *field_start = field_ptr;
         const char *field_end;
 
+        int seenLines = 0;
         do {
             const char *this_line = field_ptr;
             field_ptr = (const char *)memchr(field_ptr, '\n', header_end - field_ptr);
+            ++seenLines;
 
             if (!field_ptr) {
                 // missing <LF>
@@ -481,6 +483,15 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
                    getStringPrefix(field_start, field_end-field_start) << "}");
             debugs(55, warnOnError, " in {" << getStringPrefix(header_start, hdrLen) << "}");
 
+            PROF_stop(HttpHeaderParse);
+            clean();
+            return 0;
+        }
+
+        const bool frameHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
+        if (seenLines > 1 && frameHeader) {
+            debugs(55, warnOnError, "WARNING: obs-fold seen in " << e->name << ": " << e->value);
+            delete e;
             PROF_stop(HttpHeaderParse);
             clean();
             return 0;
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1774,3 +1774,4 @@ HttpHeader::removeConnectionHeaderEntries()
             refreshMask();
     }
 }
+
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -497,7 +497,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             const auto framingHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
             if (framingHeader) {
                 if (!hasBareCr) // already warned about bare CRs
-                    debugs(55, warnOnError, "WARNING: " << "obs-fold in framing-sensitive " << e->name << ": " << e->value);
+                    debugs(55, warnOnError, "WARNING: obs-fold in framing-sensitive " << e->name << ": " << e->value);
                 delete e;
                 PROF_stop(HttpHeaderParse);
                 clean();
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -488,13 +488,15 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             return 0;
         }
 
-        const auto framingHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
-        if (lines > 1 && framingHeader) {
-            debugs(55, warnOnError, "WARNING: obs-fold seen in " << e->name << ": " << e->value);
-            delete e;
-            PROF_stop(HttpHeaderParse);
-            clean();
-            return 0;
+        if (lines > 1) {
+            const auto framingHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
+            if (framingHeader) {
+                debugs(55, warnOnError, "WARNING: obs-fold seen in " << e->name << ": " << e->value);
+                delete e;
+                PROF_stop(HttpHeaderParse);
+                clean();
+                return 0;
+            }
         }
 
         if (e->id == Http::HdrType::CONTENT_LENGTH && !clen.checkField(e->value)) {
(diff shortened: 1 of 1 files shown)


src/HttpHeader.cc
```diff
@@ -486,6 +488,15 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             return 0;
         }
 
+        const bool frameHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
+        if (seenLines > 1 && frameHeader) {
+            debugs(55, warnOnError, "WARNING: obs-fold seen in " << e->name << ": " << e->value);
+            delete e;
+            PROF_stop(HttpHeaderParse);
+            clean();
+            return 0;
```
rousskov, line 497: We should not return 0 here -- the header is parseable -- but it is difficult to fix this properly because this PR was split from PR #702.  Assuming you insist on keeping the two PRs separate, I will insist on this PR being merged first with an XXX to be properly addressed in PR #702 (where the next generation of the framing field interpreter class can handle this nicely):

```suggestion
            return 0; // XXX: Signal invalid framing instead of unparsable header
```

src/HttpHeader.cc
```diff
@@ -400,9 +400,11 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         const char *field_start = field_ptr;
         const char *field_end;
 
+        int seenLines = 0;
```
rousskov, line 403: Nit: I think "lines seen" sounds a bit more natural, but I do not insist on this change.

```suggestion
        int linesSeen = 0;
```

or

```suggestion
        int lines = 0;
```

rousskov, line 403: Nit: Let's use a more fitting and safer size_t for counting things in a continues memory block:

```suggestion
        size_t lines = 0;
```

src/HttpHeader.cc
```diff
@@ -486,6 +488,15 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             return 0;
         }
 
+        const bool frameHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
```
rousskov, line 491: Nit: I think "framing header" sounds more standard/accurate. I do not insist on this change.

```suggestion
        const auto framingHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
```

or even 

```suggestion
        const bool framing = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
```

src/HttpHeader.cc
```diff
@@ -486,6 +488,15 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
...
+        if (seenLines > 1 && frameHeader) {
```
rousskov, line 492: I would optimize this a little (at the expense of code simplicity) because the vast majority of header fields are not about framing and only have one line:

```C++
if (seenLines > 1) {
    const auto framing = ...
    if (framing) {
        ...
    }
}

File: src/HttpHeader.cc (hunk 1 above)
Comment (yadij) [reply src/HttpHeader.cc:497]: The protocol violations introduced to non-HTTP/1.1 protocols by banning obs-fold entirely instead of taking the "or replace" criteria are not related to and cannot be fixed by the additions in PR 702.

Comment (rousskov) [reply src/HttpHeader.cc:497]: Sorry, I have no idea what protocol violations you are talking about, and why any of this matters for this change request.

Comment (yadij) [reply src/HttpHeader.cc:497]: Returning "Invalid Request" error in protocols where folding is a valid mechanism is a violation of those protocols. There are currently 3 non-HTTP/1.1 protocols using this parser where that will happen. Resolving this conflict of requirements will require the next level of ParserNG to be implemented - which is far too big a change to make a prerequisite for this and/or PR 702 (given that they need to backport).

File: src/HttpHeader.cc (hunk 2 above)
Comment (yadij) [reply src/HttpHeader.cc]: Done in 1a826d532

Comment (yadij) [reply src/HttpHeader.cc]: Done in 1a826d532

File: src/HttpHeader.cc (hunk 3 above)
Comment (yadij) [reply src/HttpHeader.cc]: Done in 1a826d532

File: src/HttpHeader.cc (hunk 1 above)
Comment (yadij) [reply src/HttpHeader.cc:497]: Returning "Invalid Request" error in protocols where folding is a valid mechanism is a violation of those protocols. There are currently 3 non-HTTP/1.1 protocols using this parser where that will happen.

Resolving this conflict of requirements will require the next level of ParserNG to be implemented - which is far too big a change to make a prerequisite for this and/or PR 702 (given that the need to backport).

File: src/HttpHeader.cc (hunk 4 above)
Comment (yadij) [reply src/HttpHeader.cc]: Done in 6ae264393

File: src/HttpHeader.cc (hunk 1 above)
Comment (rousskov) [reply src/HttpHeader.cc:497]: > Returning "Invalid Request" error in protocols where folding is a valid mechanism is a violation of those protocols. There are currently 3 non-HTTP/1.1 protocols using this parser where that will happen.

I do not see how your comment applies to my change request. I am asking to add an XXX for the new code. You are responding with a statement that implies that the new code violates some protocols. If you are right, that is a yet another reason to add an XXX! 

> Resolving this conflict of requirements will require the next level of ParserNG to be implemented - which is far too big a change to make a prerequisite for this and/or PR 702 (given that they need to backport).

Let's argue about that in PR 702, after it gets this XXX from this PR. It is very difficult to have a reasonable conversation about the future of the code in a different PR while discussing a specific ("add an XXX") change request in another PR. The XXX itself is valid and should be added. I suggest postponing other discussions until we have only one PR (with a lot more relevant-to-the-discussion code) to deal with.

Please do not misinterpret this comment as an agreement with your statements. Fortunately, no such agreement is necessary to add that XXX.

File: src/HttpHeader.cc (hunk 4 above)
Comment (rousskov) [reply src/HttpHeader.cc]: > Done in 6ae264393

No such commit in the pushed branch. The last commit is currently 1a826d5.

**Comment by rousskov** [comment]
> Please cover the `suspicious CR characters in HTTP header` case as well. I cannot be sure about this unreadable loop overall, but it looks like the suspicious CR characters removal code can be executed while seenLines remains at 1. You can just set a boolean variable to react to this case together with the folded framing header case.

Please do not forget about this (primary!) change [request](https://github.com/squid-cache/squid/pull/701#pullrequestreview-459486108).

File: src/HttpHeader.cc (hunk 4 above)
Comment (yadij) [reply src/HttpHeader.cc]: Odd. It is showing up for me - page refresh?

**Comment by yadij** [comment]
That is a significant change to the intended scope of this PR. But sure will add a flag to that case too.

File: src/HttpHeader.cc (hunk 1 above)
Comment (yadij) [reply src/HttpHeader.cc]: Adding the XXX. But FYI PR 702 does not contain the next generation of Parser - all that PR has is another layer of partial improvement (hacks) to cope with current problems (targeted at v5 and v4 usage). The actual next-gen proposal is PR 703 (targeted at v6).

Comment (yadij) [reply src/HttpHeader.cc]: XXX Done in 9d4d1f555. If I still misunderstood what you are requesting please suggest specific wording.

File: src/HttpHeader.cc (hunk 4 above)
Comment (rousskov) [reply src/HttpHeader.cc]: Resolving. FWIW, I see nothing odd about it: I wrote my comment at HH:12. GitHub [event](https://github.com/squid-cache/squid/settings/hooks/14998733) `cb151f00-d5d9-11ea-88da-e10f90b7ced5` announced the push of your commit 6ae2643 about 20 minutes later, around HH:36. FWIW, the corresponding GitHub push event `1f7fa800-d4cf-11ea-92e2-c3899794ada9` for the previous commit (1a826d5) went out more than a day earlier.

File: src/HttpHeader.cc (hunk 1 above)
Comment (rousskov) [reply src/HttpHeader.cc]: > But FYI PR 702 does not contain the next generation of Parser

Yes, I know. I believe that all we should do to address the underlying problem that spawned PR 701 and PR 702 (so far) can be done without upgrading/rewriting a lot of parsing code.

> targeted at v5 and v4 usage

We should fix the problem in master/v6 and then discuss what to do in v4 and what to do in v5. I certainly agree that v5 and/or v4 fixes may be different, but we should always start from the top -- that is how we have agreed to modify Squid.

If you want to treat this particular case as an exception, I can (reluctantly) agree to that as well, but we should postpone the current set of PRs and do a PR against v4 in that case. You would be able to commit virtually anything into v4, even if I think that more should be done in master/v6. Once that v4 PR is merged, we will pretty much ignore v4 when resuming our master/v6 work (instead of arguing whether every change is appropriate for master because of vague v4 porting concerns).

The decision about v5 should be done once master/v6 changes are known. If master/v6 changes are small/safe/surgical enough, then they should be merged into v5 as is IMO, because v5 is still very unstable right now -- there is no reason to create this significant difference for the sake of v5 stability -- and because future parsing work in the related areas would significantly benefit from this v5 exposure. Otherwise, v5 can get v4 changes (or even something in-between).

> If I still misunderstood what you are requesting please suggest specific wording.

This XXX is not related to the method API. It is related to a specific code snippet added by this PR. I _have_ [suggested](https://github.com/squid-cache/squid/pull/701#discussion_r463853974) specific XXX location and wording already!

File: src/HttpHeader.cc [hunk 5]
```diff
@@ -486,6 +495,17 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             return 0;
         }
 
+        if (lines > 1 || hasBareCr) {
+            const auto framingHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
+            if (framingHeader) {
+                debugs(55, warnOnError, "WARNING: " << (lines > 1 ? "obs-fold" : hasBareCr) << " seen in " << e->name << ": " << e->value);
```
rousskov, line 501: I would rewrite this code to replace `hasBareCr` with a `problem` or a similar general variable and set the value of that variable to "suspicious CR", "obs-fold", and "obs-fold with suspicious CR" as needed. It would simplify the code, improve diagnostics, and reduce code duplication. This particular debugs() line would become:

```suggestion
                debugs(55, warnOnError, "WARNING: " << problem << " seen in framing-sensitive " << e->name << ": " << e->value);
```

However, if you would rather keep problem-dedicated marks/variables, then let's at least reduce the duplication and simplify a little:

```suggestion
                debugs(55, warnOnError, "WARNING: " << (hasBareCr ? hasBareCr : "obs-fold") << " seen in framing-sensitive " << e->name << ": " << e->value);
```

src/HttpHeader.cc
```diff
@@ -444,6 +452,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
                     char *p = (char *) this_line;   /* XXX Warning! This destroys original header content and violates specifications somewhat */
 
                     while ((p = (char *)memchr(p, '\r', field_end - p)) != NULL) {
+                        hasBareCr = "bare-CR";
```
rousskov, line 455: The "suspicious CR characters" condition has been established by the first memchr() call above already. There is no need to (re)set hasBareCr in a loop, for every such character.

The following suggestion looks huge, but all it does is moving the hasBareCr assignment line _before_ the loop-related code.

```suggestion
                    hasBareCr = "bare-CR";
                    char *p = (char *) this_line;   /* XXX Warning! This destroys original header content and violates specifications somewhat */

                    while ((p = (char *)memchr(p, '\r', field_end - p)) != NULL) {
```

src/HttpHeader.cc
```diff
@@ -400,9 +403,12 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         const char *field_start = field_ptr;
         const char *field_end;
 
+        const char *hasBareCr = nullptr;
```
rousskov, line 406: Depending on how you decide to address the other change requests, this should become at least

```suggestion
        const char *hasSuspiciousCr = nullptr;
```

or (my preference):

```suggestion
        const char *problem = nullptr;
```

src/HttpHeader.cc
yadij: I see. The problem is that these cases are *not* invalid framing. They are perfectly valid syntax - just *suspicious* due to being unnecessary existing. It is a somewhat arbitrary policy decision to reject these messages instead of continuing with the status-quo ignore behaviour.

The same new mechanism to deliver signals to callers should be usable for all the other non-error situations this method handles. Thus my choice of placement and wording for the XXX. I intend to resolve that in PR 703 for v6 release before it moves out of WIP/Draft.

src/HttpHeader.cc
```diff
@@ -423,6 +429,8 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
                     for (const char *p = this_line; p < field_end && cr_only; ++p) {
                         if (*p != '\r')
                             cr_only = false;
+                        else
+                            hasBareCr = "bare-CR";
                     }
```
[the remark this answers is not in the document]
yadij: Not quite. This loop detects when the entire line consist only of a series of CR*LF. If cr_only becomes true there is a rejection immediately due to that. Otherwise it has found a pattern of CR* foo CRLF.

That prefix CR is bare-CR so does meet our policy requirement. Technically we might use this to short-circuit the memchr() test later as we already know bare-CR exists, but I have not gone that far since common case will not get here anyway.

src/HttpHeader.cc
yadij: Fixed in 697651bd5

src/HttpHeader.cc
yadij: The comments in this old parser use the wording "bare CR". IIRC that is based on security issue documentation from many years ago. Since this is all going to be replaced before to long anyway I am leaving this with the existing terminology.

src/HttpHeader.cc
rousskov: > The problem is that these cases are not invalid framing. They are perfectly valid syntax... It is a somewhat arbitrary policy decision to reject these messages instead of continuing with the status-quo ignore behaviour.

I agree with the quoted summary, but I do not see "the problem" with that summary.

We should make a policy decision and then, if necessary, adjust the code to follow it.

* If our decision is to ban folded framing headers, then we should adjust the official code accordingly, eventually merging this PR.
* If the decision is to continue to allow folded framing headers, then this PR should be closed without merging.

My recommendation is (still) the former, but I do not insist on it. We can ban folded framing headers after the next CVE that will abuse folded framing headers (instead of speculating now that there will be such a CVE in the future).

The same logic applies to other things this PR bans in framing headers (currently "suspicious CRs" IIRC).

src/HttpHeader.cc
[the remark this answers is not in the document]
rousskov: > Alex: Please double check this change. AFAICT, it is buggy and should be removed.
> Amos: Not quite.

The change was buggy. Thank you for removing it. I am resolving this change request.

As for your explanation of what the cr_only loop does, I believe the "Otherwise it has found a pattern of CR foo CRLF" and "prefix CR" parts are inaccurate or misleading, but perhaps you were talking about some other loop/code, so I am not going to spend time on discussing those details in this now-addressed change request.

src/HttpHeader.cc
rousskov: > The comments in this old parser use the wording "bare CR".

What do you mean by "comments in this old parser"? I cannot find the word "bare" in the master:src/HttpHeader.cc.

While researching this issue, I discovered that "bare CR" is a technical term that essentially means CR not followed by LF. This term is used (without definition) by RFC 7230 and RFC 7231, for example, where its intended meaning is possible to discern from the context. The existing src/HttpHeader.cc code does not use this term, but we _can_ start using it. Thus, I am resolving this change request.

src/HttpHeader.cc
```diff
@@ -437,6 +443,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
 
             /* Barf on stray CR characters */
             if (memchr(this_line, '\r', field_end - this_line)) {
+                hasBareCr = "bare-CR";
```
rousskov, line 446: The `bare CR` term does not use a dash. Let's drop it to better match RFCs and other documentation:

```suggestion
                hasBareCr = "bare CR";
```

src/HttpHeader.cc
```diff
@@ -486,6 +493,17 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
             return 0;
         }
 
+        if (lines > 1 || hasBareCr) {
+            const auto framingHeader = (e->id == Http::HdrType::CONTENT_LENGTH || e->id == Http::HdrType::TRANSFER_ENCODING);
+            if (framingHeader) {
+                debugs(55, warnOnError, "WARNING: " << (hasBareCr ? hasBareCr : "obs-fold") << " seen in framing-sensitive " << e->name << ": " << e->value);
```
rousskov, line 499: Forgot to mention: We have already warned about bare CR. I do not insist on reducing this WARNING noise in this PR, but it is easy to reduce if you want to:

```suggestion
                if (!hasBareCr) // already warned about bare CRs
                    debugs(55, warnOnError, "WARNING: " << "obs-fold in framing-sensitive " << e->name << ": " << e->value);
```

A better version would warn about both obs-fold and bare CR if both were found, but you have rejected similar improvements earlier in this PR, so I am not going to revise them here.

yadij:
Okay. Anything else?

rousskov:
I fixed formatting and polished debugs(). Please check the last two commits. I have no objections to this PR going in.

PR #702, opened 2020-07-31
Improve Transfer-Encoding handling

yadij:
Reject messages containing Transfer-Encoding header with coding other
than chunked or identity. Squid does not support other codings.

For simplicity and security sake, also reject messages where
Transfer-Encoding contains unnecessary complex values that are
technically equivalent to "chunked" or "identity" (e.g., ",,chunked" or
"identity, chunked").

RFC 7230 formally deprecated and removed identity coding, but it is
still used by some agents.

### DIFF
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -531,7 +531,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         // and clen state becomes irrelevant
 
         if (rawTe == "chunked") {
-            teChunked_ = true;
+            ; // NP: leave header present for chunked() method
         } else if (rawTe == "identity") { // deprecated. no coding
             delById(Http::HdrType::TRANSFER_ENCODING);
         } else {
@@ -1777,3 +1777,4 @@ HttpHeader::removeConnectionHeaderEntries()
             refreshMask();
     }
 }
+
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1779,3 +1779,4 @@ HttpHeader::removeConnectionHeaderEntries()
             refreshMask();
     }
 }
+
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -510,8 +510,8 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
 
     String rawTe;
     if (clen.prohibitedAndIgnored()) {
-        // NOTE: prohibitedAndIgnored() includes trailer header blocks
-        //       being parsed as a case to forbid/ignore these headers.
+        // prohibitedAndIgnored() includes trailer header blocks
+        // being parsed as a case to forbid/ignore these headers.
 
         // RFC 7230 section 3.3.2: A server MUST NOT send a Content-Length
         // header field in any response with a status code of 1xx (Informational)
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -533,7 +533,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         // and clen state becomes irrelevant
 
         if (rawTe == "chunked") {
-            ; // NP: leave header present for chunked() method
+            ; // leave header present for chunked() method
         } else if (rawTe == "identity") { // deprecated. no coding
             delById(Http::HdrType::TRANSFER_ENCODING);
         } else {
@@ -1779,4 +1779,3 @@ HttpHeader::removeConnectionHeaderEntries()
             refreshMask();
     }
 }
-
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -181,6 +181,7 @@ HttpHeader::operator =(const HttpHeader &other)
         update(&other); // will update the mask as well
         len = other.len;
         conflictingContentLength_ = other.conflictingContentLength_;
+        teUnsupported_ = other.teUnsupported_;
     }
     return *this;
 }
@@ -229,6 +230,7 @@ HttpHeader::clean()
     httpHeaderMaskInit(&mask, 0);
     len = 0;
     conflictingContentLength_ = false;
+    teUnsupported_ = false;
     PROF_stop(HttpHeaderClean);
 }
 
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -506,6 +506,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
                Raw("header", header_start, hdrLen));
     }
 
+    String rawTe;
     if (clen.prohibitedAndIgnored()) {
         // RFC 7230 section 3.3.2: A server MUST NOT send a Content-Length
         // header field in any response with a status code of 1xx (Informational)
@@ -513,11 +514,29 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         // such Content-Lengths.
         if (delById(Http::HdrType::CONTENT_LENGTH))
             debugs(55, 3, "Content-Length is " << clen.prohibitedAndIgnored());
-    } else if (chunked()) {
+
+        // RFC 7230 section 3.3.1 has the same criteria forbid Transfer-Encoding
+        if (delById(Http::HdrType::TRANSFER_ENCODING)) {
+            debugs(55, 3, "Transfer-Encoding is " << clen.prohibitedAndIgnored());
+            teUnsupported_ = true;
+        }
+
+    } else if (getByIdIfPresent(Http::HdrType::TRANSFER_ENCODING, &rawTe)) {
         // RFC 2616 section 4.4: ignore Content-Length with Transfer-Encoding
         // RFC 7230 section 3.3.3 #3: Transfer-Encoding overwrites Content-Length
         delById(Http::HdrType::CONTENT_LENGTH);
         // and clen state becomes irrelevant
+
+        if (rawTe == "chunked") {
+            teChunked_ = true;
+        } else if (rawTe == "identity") { // deprecated. no coding
+            delById(Http::HdrType::TRANSFER_ENCODING);
+        } else {
+            // This also rejects multiple encodings until we support them properly.
+            debugs(55, warnOnError, "WARNING: unsupported Transfer-Encoding used by client: " << rawTe);
+            teUnsupported_ = true;
+        }
+
     } else if (clen.sawBad) {
         // ensure our callers do not accidentally see bad Content-Length values
         delById(Http::HdrType::CONTENT_LENGTH);
@@ -1755,4 +1774,3 @@ HttpHeader::removeConnectionHeaderEntries()
             refreshMask();
     }
 }
-
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -508,6 +508,9 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
 
     String rawTe;
     if (clen.prohibitedAndIgnored()) {
+        // NOTE: prohibitedAndIgnored() includes trailer header blocks
+        //       being parsed as a case to forbid/ignore these headers.
+
         // RFC 7230 section 3.3.2: A server MUST NOT send a Content-Length
         // header field in any response with a status code of 1xx (Informational)
         // or 204 (No Content). And RFC 7230 3.3.3#1 tells recipients to ignore
--- a/src/HttpHeader.h
+++ b/src/HttpHeader.h
@@ -159,8 +159,9 @@ public:
     int hasListMember(Http::HdrType id, const char *member, const char separator) const;
     int hasByNameListMember(const char *name, const char *member, const char separator) const;
     void removeHopByHopEntries();
-    bool chunked() const { return teChunked_; };             ///< whether message uses chunked Transfer-Encoding
-    bool unsupportedTe() const { return teUnsupported_; }; ///< whether message uses unsupported Transfer-Encoding
+    inline bool chunked() const; ///< whether message uses chunked Transfer-Encoding
+    /// whether message used an unsupported (or invalid) Transfer-Encoding
+    bool unsupportedTe() const { return teUnsupported_; }
 
     /* protected, do not use these, use interface functions instead */
     std::vector<HttpHeaderEntry*, PoolingAllocator<HttpHeaderEntry*> > entries; /**< parsed fields in raw format */
@@ -184,7 +185,6 @@ protected:
 private:
     HttpHeaderEntry *findLastEntry(Http::HdrType id) const;
     bool conflictingContentLength_; ///< found different Content-Length fields
-    bool teChunked_ = false;        ///< found Transfer-Encoding:chunked
     bool teUnsupported_ = false;    ///< found unsupported Transfer-Encoding value(s)
 };
 
@@ -194,6 +194,14 @@ int httpHeaderParseQuotedString(const char *start, const int len, String *val);
 SBuf httpHeaderQuoteString(const char *raw);
 
 void httpHeaderCalcMask(HttpHeaderMask * mask, Http::HdrType http_hdr_type_enums[], size_t count);
+
+inline bool
+HttpHeader::chunked() const
+{
+    return has(Http::HdrType::TRANSFER_ENCODING) &&
+           hasListMember(Http::HdrType::TRANSFER_ENCODING, "chunked", ',');
+}
+
 void httpHeaderInitModule(void);
 
 #endif /* SQUID_HTTPHEADER_H */
--- a/src/HttpHeader.h
+++ b/src/HttpHeader.h
@@ -200,8 +200,7 @@ void httpHeaderCalcMask(HttpHeaderMask * mask, Http::HdrType http_hdr_type_enums
 inline bool
 HttpHeader::chunked() const
 {
-    return has(Http::HdrType::TRANSFER_ENCODING) &&
-           hasListMember(Http::HdrType::TRANSFER_ENCODING, "chunked", ',');
+    return has(Http::HdrType::TRANSFER_ENCODING);
 }
 
 void httpHeaderInitModule(void);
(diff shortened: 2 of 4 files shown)


src/HttpHeader.h
```diff
@@ -191,14 +194,6 @@ int httpHeaderParseQuotedString(const char *start, const int len, String *val);
 SBuf httpHeaderQuoteString(const char *raw);
 
 void httpHeaderCalcMask(HttpHeaderMask * mask, Http::HdrType http_hdr_type_enums[], size_t count);
-
-inline bool
-HttpHeader::chunked() const
-{
-    return has(Http::HdrType::TRANSFER_ENCODING) &&
```
rousskov, line 198: I am very worried that we are replacing this always-fresh method with a cached-by-parse() value while manually adding Transfer-Encoding headers to HttpHeader, completely bypassing parse(). I suggest keeping this method and just changing its implementation to a single has() call because the parser will remove non-chunked Transfer-Encoding values and we never add non-chunked Transfer-Encoding values.

This change request supersedes teChunked polishing change requests.

src/HttpHeader.cc
```diff
@@ -522,18 +522,37 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
...
     if (clen.prohibitedAndIgnored()) {
         // RFC 7230 section 3.3.2: A server MUST NOT send a Content-Length
         // header field in any response with a status code of 1xx (Informational)
         // or 204 (No Content). And RFC 7230 3.3.3#1 tells recipients to ignore
         // such Content-Lengths.
         if (delById(Http::HdrType::CONTENT_LENGTH))
             debugs(55, 3, "Content-Length is " << clen.prohibitedAndIgnored());
-    } else if (chunked()) {
+
+        // RFC 7230 section 3.3.1 has the same criteria forbid Transfer-Encoding
```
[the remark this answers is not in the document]
yadij: The missing GET/HEAD criteria cases are already XXX marked in the ContentLengthInterpreter class.

The addition of trailer as a place where these header are forbidden is I believe an oversight in the RFC texts. This handling seems correct to me for that case. Will add a note comment about that to the caller.

I am not seeing any other ways that method can return true/false result different to the RFC criteria.

src/HttpHeader.h
```diff
@@ -183,6 +184,8 @@ class HttpHeader
 private:
     HttpHeaderEntry *findLastEntry(Http::HdrType id) const;
     bool conflictingContentLength_; ///< found different Content-Length fields
+    bool teChunked_ = false;        ///< found Transfer-Encoding:chunked
+    bool teUnsupported_ = false;    ///< found unsupported Transfer-Encoding value(s)
```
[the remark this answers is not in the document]
yadij: The naming of this is intentional. The parser setting the flag is not detecting "bad" - it is detecting unsupported, of which "bad" is just a (quite minor) subset. Future improved parser should be able to detect and handle "bad" without setting this flag.

As you know from earlier discussion the I consider the empty-header to be semantically identical to "identity" coding. We are intentionally calling that case "unsupported".

src/HttpHeader.h
yadij: I do not believe there is ever a valid case for adding/changing these framing headers to values different from those seen by the parser. Doing so would mean the payload is different to the one the header represents.

src/HttpHeader.h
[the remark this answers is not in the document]
rousskov: These headers are "bad" from Squid point of view. If you can find a better word, please use that instead. The point is that we want to include unsupported, malformed, invalid, etc. headers. Not just the valid-but-unsupported ones.

src/HttpHeader.h
rousskov: I will try to rephrase: The existence of an API for adding a Transfer-Encoding:chunked header is incompatible with the chunked() implementation in the PR. The official implementation does not have that problem. We should not create such a problem without a compelling reason, especially when Squid already uses that API. I do not see such a reason.

To see cases where these framing headers are added without going through the parser use `git grep putStr.*TRANSFER_ENCODING`.

src/HttpHeader.h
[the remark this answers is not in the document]
yadij: My view is that we do not support identifying "bad" vs "valid". So *everything* except the two exact accepted headers is unsupported right now. To call something bad we have to have parsed the syntax to determine that it is bad.

src/HttpHeader.h
yadij: So what exactly did you mean then when you argued for implementing this function as a flag test?

rousskov: I was thinking about a (correct) implementation that would not raise these red flags. Such an implementation may require changing some putStr() callers as well. However, that idea was meant for the rather different Transfer-Encoding parsing code that no longer exists.

Given the current PR code, returning has() is probably the best approach for chunked(): It is simple, reliable, fast enough, and does not conflict with Squid code that adds chunked encoding (without any caller modifications). It would require deleting unsupported Transfer-Encoding headers, but deleting them is probably a good idea anyway (just like we already delete bad Content-Length headers and Transfer-Encoding:identity headers).

src/HttpHeader.cc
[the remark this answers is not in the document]
rousskov: > I am not seeing any other ways that method can return true/false result different to the RFC criteria.

It sounds like you are focusing on the details of the method implementation but forgetting that the method itself is a part of the class that is completely _unused_ for Transfer-Encoding header fields: ContentLengthInterpreter is only used for Content-Length fields. Please reread my original change [request](https://github.com/squid-cache/squid/pull/702/files#r463819674) with that fact in mind.

> Added a note comment about that in b1f897d .

FYI: The new note/comment is misplaced: prohibitedAndIgnored() _caller_ should not care why or how that "prohibit and ignore" decision was made. The same can be said about the old/existing `RFC 7230 section 3.3.2: A server MUST NOT` comment. All this information should most likely go into ContentLengthInterpreter instead of its users. Some of it may be there already. This caller should just apply the already made decision.

src/HttpHeader.h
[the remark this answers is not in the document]
rousskov: If we cannot find (or agree on) a word that would describe the applicable use cases well, let's at least clarify what it going on in the field description, which currently represents a useless "foo is foo" tautology.

```suggestion
    bool teUnsupported_ = false; ///< unsupported and/or invalid Transfer-Encoding
```

> Future improved parser should be able to detect and handle "bad" without setting this flag.

If/when that happens, the new field name would become very appropriate and the description as well as the code using this field will need to be adjusted accordingly. Until then, we should be mindful that the field covers both unsupported valid codings and malformed ones.

src/HttpHeader.cc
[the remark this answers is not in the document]
yadij: Yes I am focusing on the implementation for this PR. I looked into the interpreter usage and saw that the request-method / reply-status criteria used by prohibitedAndIgnored() are set by callers of the parser. So this particular check can be used regardless of whether the Content-Length header exists, it is the other checks that depend on C-L header existence.

If the caller lacks the method/status info for this check it cannot pass them in for T-E filtering anyway - so I think this is sufficient (for now) to prune away T-E headers whenever this check returns true.

src/HttpHeader.h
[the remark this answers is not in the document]
yadij: Okay, that sounds like a good compromise to me. Updating the documentation now to be more verbose about the multiple input case types that are currently under the category "unsupported".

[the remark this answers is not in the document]
yadij: Good catch. Thanks. Added the missing copy/clone in upcoming polish commit. The other uses of the example flag are covered by the C++11 initialization this new flag uses.

src/HttpHeader.h
yadij: Okay. If we are going to assume putStr() API usage, that caller may set a non-chunked coding so I am leaving the check for "chunked" it used to do (in aff8e63).

That leaves some potentially nasty issues with custom unsupported/invalid T-E not being detected. But since it is coming from an "internal" source I think we can (should?) leave that as a problem for the admin or ICAP/eCAP service dev to ensure never happens - for now at least.

src/client_side.cc
```diff
@@ -1666,13 +1664,8 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
...
     }
 
-    if (request->header.chunked()) {
-        chunked = true;
-    } else if (request->header.has(Http::HdrType::TRANSFER_ENCODING)) {
-        const String te = request->header.getList(Http::HdrType::TRANSFER_ENCODING);
-        // HTTP/1.1 requires chunking to be the last encoding if there is one
-        unsupportedTe = te.size() && te != "identity";
-    } // else implied identity coding
+    bool chunked = request->header.chunked();
```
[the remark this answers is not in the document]
yadij: Done in f1b48b45d

src/HttpHeader.cc
```diff
@@ -547,7 +549,7 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         // and clen state becomes irrelevant
 
         if (rawTe == "chunked") {
-            teChunked_ = true;
+            ; // NP: leave header present for chunked() method
```
rousskov, line 552: Nit: "NP" is not a common abbreviation. Most readers will need to look it up and/or will misinterpret it. IIRC, most of the existing Squid code misuses "NP" to mean "note please". AFAICT, the most common definition (outside of Squid) for NP is "no problem". We should avoid NP in any context (in favor of more common/understood variants). It is unnecessary here.

```suggestion
            ; // leave header present for chunked() method
```

src/HttpHeader.h
```diff
@@ -194,6 +196,14 @@ int httpHeaderParseQuotedString(const char *start, const int len, String *val);
 SBuf httpHeaderQuoteString(const char *raw);
 
 void httpHeaderCalcMask(HttpHeaderMask * mask, Http::HdrType http_hdr_type_enums[], size_t count);
+
+inline bool
+HttpHeader::chunked() const
+{
+    return has(Http::HdrType::TRANSFER_ENCODING) &&
+           hasListMember(Http::HdrType::TRANSFER_ENCODING, "chunked", ',');
```
rousskov: AFAICT, this method does not need its second component anymore. Please move its defitinion to the declaration and just return `has(Http::HdrType::TRANSFER_ENCODING)`.

src/HttpHeader.cc
```diff
@@ -522,18 +524,40 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
                Raw("header", header_start, hdrLen));
     }
 
+    String rawTe;
     if (clen.prohibitedAndIgnored()) {
+        // NOTE: prohibitedAndIgnored() includes trailer header blocks
+        //       being parsed as a case to forbid/ignore these headers.
+
```
rousskov, line 531: Nit: This code is not the place where we should document what prohibitedAndIgnored() means or what cases it covers. I suggest just dropping this comment (and the following new line -- GitHub does not support that via suggestions). If you prefer, let's improve ContentLengthInterpret API documentation to detail this.

rousskov:
> I have only two minor nits left for now. 

And an HttpHeader::chunked() [change](https://github.com/squid-cache/squid/pull/702#discussion_r465172095).

src/HttpHeader.h
rousskov: > [Amos](https://github.com/squid-cache/squid/pull/702#discussion_r464808252): that caller may set a non-chunked coding

There are no such callers today, and I see no plausible scenario where they would be allowed in the future. If you want to protect us from (more) future bugs, you can (by adjusting putStr() implementation and adding a new method dedicated to setting Transfer-Encoding), but I do not think those protections are _necessary_. And even if they are deemed desirable, it is OK to leave them outside this "surgical" PR IMO because the problem is unlikely to manifest itself and does not affect Squid performance or functionality. Those hypothetical non-trivial changes are only about protecting future us from rather unlikely coding mistakes.

yadij: My understanding of your argument against using a flag was that there might become such users on grounds that the API to set arbitrary headers exists. The corollary to that (theoretical) situation is the ability to set non-chunked codings into the header.

And yes there are such callers today. The httpHdrMangleList() logic is one.  Leaving that second argument in place covers what I expect to be the most common situations without going so far as to increase scope (and bonus: it shrinks the patch).

src/HttpHeader.cc
yadij: Interesting interpretation of "NP:". This is a en_NZ / en_AU 'ism - perhapse since we have other far more common phrases meaning "no problem". Accepting since the Squid documentation is being enforced to be en_US.

src/HttpHeader.cc
yadij: This is here to mark specifically prohibitedAndIgnored() implications on the following delById() calls extending beyond the RFC requirements also commented here. It is unrelated to any other uses of prohibitedAndIgnored().

src/HttpHeader.h
rousskov: > My understanding of your argument against using a flag was that there might become such users on grounds that the API to set arbitrary headers exists

My argument against the initial (incorrect) flag-based implementation attempt was that existing Squid code makes that flag stale. No future/hypothetical/admin-dependent problems. The suggested code was simply broken.

The same problem does not exist for a has()-based implementation. For a has()-based implementation to become broken in a similar fashion, we would have to add new code (and new functionality) that does not look like anything we have today. That is why I am comfortable with not addressing that future (unlikely) problem in this PR. Please note that I am not against addressing that problem in this PR either -- I do not recommend doing it here for the reasons already [stated](https://github.com/squid-cache/squid/pull/702#discussion_r465207738), but I would not object to (properly) addressing it if your risk assessment is different.

> And yes there are such callers today. The httpHdrMangleList() logic is one.

I do not see how httpHdrMangleList() would break a has()-based implementation. If somebody misconfigures Squid to add custom Transfer-Encoding headers, then we are not responsible for the outcome, _and_ the Squid code would "work" as badly with the `chunked` member check as without it (if it makes sense to compare sets of broken use cases at all -- I doubt it does).

src/HttpHeader.cc
rousskov: > This is here to mark specifically prohibitedAndIgnored() implications on the following delById() calls extending beyond the RFC requirements

The source code comment about the RFC requirements suffers from exactly the same problem -- it is misplaced. This PR makes things arguably worse by adding another misplaced comment to an existing misplaced comment (using that existing misplaced comment as the hidden/undisclosed reason/excuse). See my intial [suggestion](https://github.com/squid-cache/squid/pull/702#discussion_r465177177) for two ways to avoid aggravating this problem.

Please note that I am just explaining my (optional) change suggestion. I am not blocking this PR because it adds a misplaced comment.

rousskov:
I adjusted the PR description, primarily to better detail the (controversial) decision to reject unnecessary complex header values that evaluate to the two supported codings. Please adjust further if needed.

rousskov:
> Will wait for PR 701 going in so that its XXX can be addressed as well.

I found official HttpHeader::parse(...clen) code that also contains the same problem -- mistreating parsable dangerous header as unparsable one. Thus, I am not going to insist on that (old/existing) problem to be fixed in this pair of "surgical" PRs. We just need to resolve the remaining explicit change requests (one of which is optional).

src/HttpHeader.cc
rousskov: Also, if you insist on adding this misplaced comment, please at least remove its `NOTE: ` prefix. It is just noise  -- virtually all comments are "please note: ..." code annotations.

yadij: Removed the "NOTE:" prefix in 50643ac0d.

src/HttpHeader.h
yadij: Fine. Done in 2fc34818a

kinkie:
Latest jenkins build failed on commit hash not being found; next commit will fix it

rousskov:
> Latest jenkins build failed on commit hash not being found; next commit will fix it

I pushed a trivial change (commit 7439beb) in hope to fix Jenkins.

rousskov:
FTR, Squid's strict handling of Transfer-Encoding values rejects the following (slightly abridged) server responses containing two `Transfer-Encoding: chunked` header fields:

```
HTTP/1.1 200 OK
Date: Thu, 24 Sep 2020 11:56:13 GMT
Server: JBoss-EAP/7
Expires: Thu, 01 Jan 1970 00:00:00 GMT
Cache-Control: no-cache
Cache-Control: no-store
X-Powered-By: Undertow/1
X-Powered-By: JSP/2.3
Pragma: no-cache
Transfer-Encoding: chunked
Content-Type: text/html;charset=UTF-8
Vary: Accept-Encoding
Connection: close
Cache-Control: no-cache
Content-Encoding: gzip
Transfer-Encoding: chunked
```

The above response header can be interpreted as a violation of the following RFC 7230 MUST-level rquirement: `A sender MUST
   NOT apply chunked more than once to a message body (i.e., chunking an
   already chunked message is not allowed).` The message body itself is probably not double-chunked, but I did not check, and it should not really matter.

FWIW, at this time, Factory does not plan to change Squid to treat such messages specially -- they will be rejected to mitigate HTTP framing-related attacks. However, if this and similar problems become too frequent, we will need to do more to improve compatibility. For now, let's just collect information about these use cases.

yadij:
> The above response header can be interpreted as a violation of the following RFC 7230 MUST-level rquirement: `A sender MUST NOT apply chunked more than once to a message body (i.e., chunking an already chunked message is not allowed).` The message body itself is probably not double-chunked, but I did not check, and it should not really matter.

It is definitively a violation. If chunked was applied twice that is forbidden. The header is a list header - this input means explicitly that chunked _was_ applied twice. Whether or not it was actually applied twice the sender is not performing HTTP sufficiently well to trust its message framing.

rousskov:
FTR, I am [updating](https://github.com/squid-cache/squid/pull/702#issuecomment-698948143) the list of known real-world benign cases that Squid (correctly!) rejects as of this PR:

1. WARNING: unsupported Transfer-Encoding used by client: `chunked, chunked`
1. WARNING: unsupported Transfer-Encoding used by client: `gzip, chunked`
1. WARNING: unsupported Transfer-Encoding used by client: `Binary`
1. WARNING: Leading garbage or empty value in `Content-Length[4]=null`

The above were seen on various secondary legitimate servers, including servers that belong to well-known well-resourced companies, organizations, and governments around the world.

I may update this comment as Squid admins continue to complain about these cases.

PR #705, opened 2020-08-05
Replaced X-Cache and X-Cache-Lookup headers with Cache-Status

yadij:
See https://tools.ietf.org/html/draft-ietf-httpbis-cache-header

Also switched to identifying Squid instance in the header using
unique_hostname(), fixing a bug affecting proxies that share the same
visible_hostname in a cluster. The Cache-Status field values will now
point to a specific proxy in such a cluster.

The new initial lookup reporting (formally X-Cache-Lookup)
implementation has several differences:

* The reporting of the lookup is now unconditional, just like the
  Cache-Status reporting itself. The dropped X-Cache-Lookup required
  --enable-cache-digests. We removed that precondition because
  Cache-Status already relays quite a bit of information, and sharing
  lookup details is unlikely to tilt the balance in most environments.
  The original lookup reporting code was conditional because the feature
  was added for Cache Digests debugging, not because we wanted to hide
  the info. Folks later discovered that the info is useful in many other
  cases. If we do want to hide this information, it should be done
  directly rather than via a (no) Cache Digests side effect.

* The initial lookup classification can no longer be overwritten by
  additional Store lookups. Official code allowed such rewrites due to
  implementation bugs. If we only report one lookup, the first lookup
  classification is the most valuable one and almost eliminates doubts
  about the the cache state at the request time. Ideally, we should also
  exclude various internal Store lookup checks that may hide a matching
  cache entry, but that exclusion is difficult to implement with the
  current needlessly asynchronous create() Store API.

* Lookup reporting now covers more use cases. The official code probably
  missed or mishandled various PURGE/DELETE use cases and did not
  distinguish absence of Store lookup because of CC:no-cache from other
  lookup bypass cases (e.g., errors). More work is probably needed to
  cover every lookup-avoiding case.

### DIFF
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -1562,8 +1562,8 @@ clientReplyContext::buildReplyHeader()
 #endif
 
     SBuf cacheStatus(uniqueHostname());
-    if (is_hit)
-        cacheStatus.append(";hit");
+    if (const auto hitOrFwd = http->logType.cacheStatusSource())
+        cacheStatus.append(hitOrFwd);
     if (firstStoreLookup_) {
         cacheStatus.append(";detail=");
         cacheStatus.append(firstStoreLookup_);
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -81,9 +81,6 @@ clientReplyContext::clientReplyContext(ClientHttpRequest *clientContext) :
     old_reqsize(0),
     reqsize(0),
     reqofs(0),
-#if USE_CACHE_DIGESTS
-    lookup_type(NULL),
-#endif
     ourNode(NULL),
     reply(NULL),
     old_entry(NULL),
@@ -1567,14 +1564,6 @@ clientReplyContext::buildReplyHeader()
                      (is_hit ? "hit" : "fwd=miss")
                      );
 
-#if USE_CACHE_DIGESTS
-    /* Append X-Cache-Lookup: -- temporary hack, to be removed @?@ @?@ */
-    httpHeaderPutStrf(hdr, Http::HdrType::X_CACHE_LOOKUP, "%s from %s:%d",
-                      lookup_type ? lookup_type : "NONE",
-                      getMyHostname(), getMyPort());
-
-#endif
-
     const bool maySendChunkedReply = !request->multipartRangeRequest() &&
                                      reply->sline.protocol == AnyP::PROTO_HTTP && // response is HTTP
                                      (request->http_ver >= Http::ProtocolVersion(1,1));
@@ -1718,10 +1707,6 @@ clientReplyContext::identifyFoundObject(StoreEntry *newEntry)
       */
     if (r->flags.noCache || r->flags.noCacheHack())
         ipcacheInvalidateNegative(r->url.host());
-
-#if USE_CACHE_DIGESTS
-    lookup_type = e ? "HIT" : "MISS";
-#endif
 
     if (!e) {
         /** \li If no StoreEntry object is current assume this object isn't in the cache set MISS*/
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -1562,10 +1562,9 @@ clientReplyContext::buildReplyHeader()
         Auth::UserRequest::AddReplyAuthHeader(reply, request->auth_user_request, request, http->flags.accel, 0);
 #endif
 
-    httpHeaderPutStrf(hdr, Http::HdrType::CACHE_STATUS, "%s;%s%s",
+    httpHeaderPutStrf(hdr, Http::HdrType::CACHE_STATUS, "%s;%s",
                      uniqueHostname(),
-                     (is_hit ? "hit" : "fwd=miss"),
-                     (collapsedRevalidation != crNone ? ";collapsed" : "")
+                     (is_hit ? "hit" : "fwd=miss")
                      );
 
 #if USE_CACHE_DIGESTS
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -1562,7 +1562,6 @@ clientReplyContext::buildReplyHeader()
         Auth::UserRequest::AddReplyAuthHeader(reply, request->auth_user_request, request, http->flags.accel, 0);
 #endif
 
-    /* Append Cache-Status */
     httpHeaderPutStrf(hdr, Http::HdrType::CACHE_STATUS, "%s;%s%s",
                      uniqueHostname(),
                      (is_hit ? "hit" : "fwd=miss"),
@@ -2348,4 +2347,3 @@ clientBuildError(err_type page_id, Http::StatusCode status, char const *url,
 
     return err;
 }
-
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -2346,3 +2346,4 @@ clientBuildError(err_type page_id, Http::StatusCode status, char const *url,
 
     return err;
 }
+
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -931,6 +931,8 @@ clientReplyContext::purgeAllCached()
 void
 clientReplyContext::created(StoreEntry *newEntry)
 {
+    detailStoreLookup(newEntry ? "match" : "mismatch");
+
     if (lookingforstore == 1)
         purgeFoundGet(newEntry);
     else if (lookingforstore == 2)
@@ -1559,10 +1561,14 @@ clientReplyContext::buildReplyHeader()
         Auth::UserRequest::AddReplyAuthHeader(reply, request->auth_user_request, request, http->flags.accel, 0);
 #endif
 
-    httpHeaderPutStrf(hdr, Http::HdrType::CACHE_STATUS, "%s;%s",
-                     uniqueHostname(),
-                     (is_hit ? "hit" : "fwd=miss")
-                     );
+    SBuf cacheStatus(uniqueHostname());
+    cacheStatus.append(is_hit ? ";hit" : ";fwd=miss");
+    if (firstStoreLookup_) {
+        cacheStatus.append(";detail=");
+        cacheStatus.append(firstStoreLookup_);
+    }
+    // TODO: Remove c_str() after converting HttpHeaderEntry::value to SBuf
+    hdr->putStr(Http::HdrType::CACHE_STATUS, cacheStatus.c_str());
 
     const bool maySendChunkedReply = !request->multipartRangeRequest() &&
                                      reply->sline.protocol == AnyP::PROTO_HTTP && // response is HTTP
@@ -1686,6 +1692,8 @@ clientReplyContext::identifyStoreObject()
         lookingforstore = 5;
         StoreEntry::getPublicByRequest (this, r);
     } else {
+        // "external" no-cache requests skip Store lookups
+        detailStoreLookup("no-cache");
         identifyFoundObject(nullptr);
     }
 }
@@ -1768,6 +1776,18 @@ clientReplyContext::identifyFoundObject(StoreEntry *newEntry)
     debugs(85, 3, "default HIT " << *e);
     http->logType.update(LOG_TCP_HIT);
     doGetMoreData();
+}
+
+/// remembers the very first Store lookup classification, ignoring the rest
+void
+clientReplyContext::detailStoreLookup(const char *detail)
+{
+    if (!firstStoreLookup_) {
+        debugs(85, 7, detail);
+        firstStoreLookup_ = detail;
+    } else {
+        debugs(85, 7, "ignores " << detail << " after " << firstStoreLookup_);
+    }
 }
 
 /**
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -1562,9 +1562,12 @@ clientReplyContext::buildReplyHeader()
         Auth::UserRequest::AddReplyAuthHeader(reply, request->auth_user_request, request, http->flags.accel, 0);
 #endif
 
-    /* Append X-Cache */
-    httpHeaderPutStrf(hdr, Http::HdrType::X_CACHE, "%s from %s",
-                      is_hit ? "HIT" : "MISS", getMyHostname());
+    /* Append Cache-Status */
+    httpHeaderPutStrf(hdr, Http::HdrType::CACHE_STATUS, "%s;%s%s",
+                     uniqueHostname(),
+                     (is_hit ? "hit" : "fwd=miss"),
+                     (collapsedRevalidation != crNone ? ";collapsed" : "")
+                     );
 
 #if USE_CACHE_DIGESTS
     /* Append X-Cache-Lookup: -- temporary hack, to be removed @?@ @?@ */
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -1562,7 +1562,8 @@ clientReplyContext::buildReplyHeader()
 #endif
 
     SBuf cacheStatus(uniqueHostname());
-    cacheStatus.append(is_hit ? ";hit" : ";fwd=miss");
+    if (is_hit)
+        cacheStatus.append(";hit");
     if (firstStoreLookup_) {
         cacheStatus.append(";detail=");
         cacheStatus.append(firstStoreLookup_);
--- a/src/client_side_reply.h
+++ b/src/client_side_reply.h
@@ -129,11 +129,17 @@ private:
     void purgeAllCached();
     void forgetHit();
     bool blockedHit() const;
+    void detailStoreLookup(const char *detail);
 
     void sendBodyTooLargeError();
     void sendPreconditionFailedError();
     void sendNotModified();
     void sendNotModifiedOrPreconditionFailedError();
+
+    /// Classification of the initial Store lookup.
+    /// This very first lookup happens without the Vary-driven key augmentation.
+    /// TODO: Exclude internal Store match bans from the "mismatch" category.
+    const char *firstStoreLookup_ = nullptr;
 
     StoreEntry *old_entry;
     /* ... for entry to be validated */
--- a/src/client_side_reply.h
+++ b/src/client_side_reply.h
@@ -90,10 +90,6 @@ public:
     size_t reqsize;
     size_t reqofs;
     char tempbuf[HTTP_REQBUF_SZ];   ///< a temporary buffer if we need working storage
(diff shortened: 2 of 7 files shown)


yadij:
Clearing fro merge under the 10-day criteria.

rousskov:
Just to avoid misunderstanding: I still need to review this PR. The `M-cleared-for-merge` label is not a problem but should be (automatically) ignored until then.

src/client_side_reply.cc
```diff
@@ -1562,9 +1562,12 @@ clientReplyContext::buildReplyHeader()
         Auth::UserRequest::AddReplyAuthHeader(reply, request->auth_user_request, request, http->flags.accel, 0);
 #endif
 
-    /* Append X-Cache */
-    httpHeaderPutStrf(hdr, Http::HdrType::X_CACHE, "%s from %s",
-                      is_hit ? "HIT" : "MISS", getMyHostname());
+    /* Append Cache-Status */
+    httpHeaderPutStrf(hdr, Http::HdrType::CACHE_STATUS, "%s;%s%s",
+                     uniqueHostname(),
+                     (is_hit ? "hit" : "fwd=miss"),
```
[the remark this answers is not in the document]
yadij: sf-token is ABNF for the value of the key=value pair when the key is "fwd".

The layout of recent IETF drafts is getting a bit strange with examples separated from the definitions, but see section 3 for clarification on the syntax.

src/client_side_reply.cc
```diff
@@ -1562,9 +1562,12 @@ clientReplyContext::buildReplyHeader()
...
+                     (collapsedRevalidation != crNone ? ";collapsed" : "")
```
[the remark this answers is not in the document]
yadij: Noted. Since this is turning into more work than initially expected I will drop it for now. The more complicated addition can be done later.

src/client_side_reply.cc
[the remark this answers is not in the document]
yadij: For this initial support I am sticking to the values easily available in existing code and leaving all details that require logic changes to other parts of Squid out of scope.

src/client_side_reply.cc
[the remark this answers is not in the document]
yadij: Removed in 5abbd831f

src/http/RegisteredHeadersHash.gperf
```diff
@@ -95,7 +96,6 @@ Vary, Http::HdrType::VARY, Http::HdrFieldType::ftStr, HdrKind::ListHeader|HdrKin
 Via, Http::HdrType::VIA, Http::HdrFieldType::ftStr, HdrKind::ListHeader|HdrKind::GeneralHeader
 Warning, Http::HdrType::WARNING, Http::HdrFieldType::ftStr, HdrKind::ListHeader|HdrKind::ReplyHeader
 WWW-Authenticate, Http::HdrType::WWW_AUTHENTICATE, Http::HdrFieldType::ftStr, HdrKind::ListHeader|HdrKind::ReplyHeader
-X-Cache, Http::HdrType::X_CACHE, Http::HdrFieldType::ftStr, HdrKind::ReplyHeader
```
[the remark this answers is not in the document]
yadij: X-Cache will be tracked under "other" for caches which do not send it. I do not think the stats on this header existing or not existing have any meaning.

yadij:
> Please cite the draft (or RFC ID if this header was indeed formally standardized) this PR is based on. When reviewing this PR, I started with draft-nottingham-cache-header-04 but then found draft-nottingham-cache-header-05. We should be specific about our reference point, whatever it is.

https://tools.ietf.org/html/draft-ietf-httpbis-cache-header is the formal URL., it will always refer to the latest version of the document while it remains in draft status.

> 
> Should we drop the X-Cache-Lookup as well? The new Cache-Status header allows for reporting custom details, which is what X-Cache-Lookup does. Lookup detail reporting can still be conditional if needed for security reasons, of course (even though the current code pretty much assumes that it is not sensitive info).
> 

I skipped it because it reports HIT and MISS wrongly. But since you are okay with it going now I have removed it in 48fbfbe340eca1d032696e0b7681c790fa52aac1

rousskov:
> PR description: Cache-Status header is a formally standardized header
> Alex (after quoting PR description: Please cite the draft this PR is based on.
> Amos: https://tools.ietf.org/html/draft-ietf-httpbis-cache-header is the formal URL.

Not here please... That was a change request for the PR _description_. We should cite the document the implementation is based on. And please use the exact draft version. The URL does not really matter in this case -- it is trivial to find a given IETF draft text by its name-version.

src/client_side_reply.cc
```diff
@@ -1562,17 +1559,10 @@ clientReplyContext::buildReplyHeader()
...
-#if USE_CACHE_DIGESTS
-    /* Append X-Cache-Lookup: -- temporary hack, to be removed @?@ @?@ */
-    httpHeaderPutStrf(hdr, Http::HdrType::X_CACHE_LOOKUP, "%s from %s:%d",
-                      lookup_type ? lookup_type : "NONE",
-                      getMyHostname(), getMyPort());
-
-#endif
+    httpHeaderPutStrf(hdr, Http::HdrType::CACHE_STATUS, "%s;%s",
+                     uniqueHostname(),
+                     (is_hit ? "hit" : "fwd=miss")
```
rousskov, line 1564: > Alex: Should we drop the X-Cache-Lookup as well?
> Amos: I skipped it because it reports HIT and MISS wrongly.

Hm... I cannot find a reasonable interpretation of that answer, but I am OK with X-Cache-Lookup removal in this PR. Please do not forget to add the removed _information_ to Cache-Status though!

FWIW, X-Cache-Lookup is not about a cache hit or miss. X-Cache-Lookup is about initial cache index query -- it is a triage aid that cannot imply a (future) cache hit because the cache checks some but not all the response matching criteria, and because a lot of things can still go wrong even after a full match; a negative (i.e. "not found") result implies a miss AFAICT.

An Optional boolean field (clientReplyContext::positiveCacheLookup?) may work as a lookup_type replacement.

src/http/RegisteredHeadersHash.gperf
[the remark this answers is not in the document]
rousskov: > I do not think the stats on [X-Cache] header existing or not existing have any meaning.

I disagree but do not have enough facts to push for the opposite conclusion. Thank you for documenting your rationale. I am resolving this issue.

src/client_side_reply.cc
rousskov, line 1564: > I am sticking to the values easily available in existing code and leaving all details that require logic changes to other parts of Squid out of scope.

If I find an easy way to report `fwd=stale` that does not require logic changes to other parts of Squid, would you accept it?

src/client_side_reply.cc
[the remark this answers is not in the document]
rousskov: > sf-token is ABNF for the value of the key=value pair when the key is "fwd".

I see nothing in the normative parts of the draft that would support that assertion. Yes, draft examples match your natural interpretation, but the normative parts seem to contradict it:

```
Each member of the list can also have parameters that describe that
cache's handling of the request. [...]
This specification defines these parameters:
   hit          = sf-boolean
   fwd          = sf-token
```

The above "assignments" are ABNF, right? Or is that some kind of new syntax introduced by the S-F-V draft?? I see nothing in the above text or ABNF about `fwd` defining the value of some key=value pair.

> see section 3 for clarification on the syntax.

I do not see anything relevant in section 3 of either draft. Can you quote some specific draft text that would correct/adjust the formal ABNF definitions that I quoted?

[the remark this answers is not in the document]
mnot: Alex, this is the same convention we use for HTTP Header fields; e.g., see [Date](https://httpwg.org/specs/rfc7231.html#header.date). 

If you have suggestions about how to make this clearer in the cache-status draft, please make suggestions over there -- but no-one else I've seen has had trouble with this.

src/client_side_reply.cc
yadij: sure

yadij: > FWIW, X-Cache-Lookup is not about a cache hit or miss.

That is absurd. The header quite literally says "HIT from proxy.example.org:3128" or "MISS from proxy.example.org:3128" with the words being based on a cache lookup HIT/MISS result.


> X-Cache-Lookup is about initial cache index query -- it is a triage aid that cannot imply a (future) cache hit because the cache checks some but not all the response matching criteria, and because a lot of things can still go wrong even after a full match; a negative (i.e. "not found") result implies a miss AFAICT.
> 
> An Optional boolean field (clientReplyContext::positiveCacheLookup?) may work as a lookup_type replacement.

Those additional criteria inclientReplyContext::identifyFoundObject() are possibly a good basis for selecting some of the other fwd= values this header could send. But transferring that information around involves changing the APIs and testing the result accuracy - which is out of the scope selected for this PR.

src/client_side_reply.cc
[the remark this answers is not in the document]
rousskov: > Alex, this is the same convention we use for HTTP Header fields; e.g., see [Date](https://httpwg.org/specs/rfc7231.html#header.date).

@mnot, RFC 7231 is more careful in distinguishing header _fields_ from header field _names_ from header field _values_:

1. Unlike the Cache-Status draft, RFC 7231 does _not_ have a section that says
```
This specification defines these header fields:
    Date     = HTTP-date
    Location = URI-reference
    ...
```
2. The Date specification text in RFC 7231 clearly says that the ABNF below is about the Date field _value_:
```
The field value is an HTTP-date, as defined in Section 7.1.1.1.
    Date = HTTP-date
``` 

In summary, RFC 7231 uses prose/ABNF to define header field names/values, which is perfectly fine. The Cache-Status draft says that it is defining _parameters_ while providing ABNF for parameter _values_. The problem is exacerbated by the fact that parameter syntax (as a whole) is defined in another document.

> If you have suggestions about how to make this clearer in the cache-status draft, please make suggestions over there -- but no-one else I've seen has had trouble with this.

Glad I was the first to alert you about the problem, but I currently have too many windmills to fight to accept your kind invitation. If you would like to improve the quality of the draft, I suggest replacing the ABNF that follows `This specification defines these parameters` text with a two-column "Name, Value ABNF" table (similar to what RFC 7231 Section [7.1](https://httpwg.org/specs/rfc7231.html#rfc.section.7.1) is using, but more explicit/direct, to minimize draft modifications).

FWIW, I am sure that any competent Cache-Status implementer can follow the draft Examples to understand the syntax that the daft wants to define. The "just see Examples" [advice](https://github.com/squid-cache/squid/pull/705#discussion_r472091561) kind of defeats the purpose behind S-F IMHO, but saves everybody a lot of time in the short term!

src/client_side_reply.cc
rousskov: > > FWIW, X-Cache-Lookup is not about a cache hit or miss.
 
> That is absurd.

It is a fact. Whether you consider it absurd may explain your misunderstanding but does not change the meaning of this header.

> The header quite literally says "HIT from proxy.example.org:3128" or "MISS from proxy.example.org:3128" with the words being based on a cache lookup HIT/MISS result.

While I agree that the choice of the overloaded `HIT` and `MISS` terms was a bit unfortunate for this context, they correctly convey the _lookup_ outcome. Please note that the header field name tells us that its value is about the _lookup_, not whether the transaction is classified as a cache hit or cache miss.

 
> > X-Cache-Lookup is about initial cache index query -- it is a triage aid that cannot imply a (future) cache hit because the cache checks some but not all the response matching criteria, and because a lot of things can still go wrong even after a full match; a negative (i.e. "not found") result implies a miss AFAICT.
> > An Optional boolean field (clientReplyContext::positiveCacheLookup?) may work as a lookup_type replacement.
 
> Those additional criteria inclientReplyContext::identifyFoundObject() are possibly a good basis for selecting some of the other fwd= values this header could send. But transferring that information around involves changing the APIs and testing the result accuracy - which is out of the scope selected for this PR.

This change request does not require any API changes. It requires a tiny adjustment of the existing code to preserve useful information currently relayed by the being-removed X-Cache-Lookup. Please let me know if I should implement this myself. I think I found a way to do that with even fewer changes than using Optional would require.

yadij: So do you know why it often says "HIT" on responses from proxies that have no cache to look up? and "MISS" on responses obviously pulled from cache? Under your definition of its meaning both should be impossible states.
The only case where I see it conveying truth and being different from X-Cache is when a MISS is happening due to Vary.

Yes, please implement this solution you have. It would be faster than debating this further.

rousskov: > do you know why it often says "HIT" on responses from proxies that have no cache to look up?

I can offer one (speculative!) suspect: An object with the same key was in the `store_table` at the time of the request. Unfortunately, Squid uses store_table for all transactions, regardless of whether caching is enabled. I hope to fix that old design bug in the foreseeable future.

> and "MISS" on responses obviously pulled from cache?

I know of one case where the initial lookup_type `HIT` value gets overwritten but it does not match the "pulled from cache" precondition in your question: When Squid finds a matching cached response with a Vary response header, identifyFoundObject() sets lookup_type to `HIT`. Squid then tries to find another response object using a Vary-augmented Store key. That attempt results in the second identifyFoundObject() call (which lookup_type authors could not expect because Squid did not know about Vary at the time). That second call may change the lookup_type `HIT` value to `MISS`.


> Under your definition of its meaning both should be impossible states.

... or Squid bugs or our insufficient understanding of the corresponding (very convoluted!) code. Investigating a couple of such cases may be quite useful. FWIW, I have not noticed any such cases, but I was not looking for them.

BTW, the X-Cache:MISS value is often lying in delay pools environments due to the known [bug](http://lists.squid-cache.org/pipermail/squid-users/2019-March/020388.html). It would be very nice to fix that bug _before_ merging Cache-Status, so that we know that all Cache-Status responses do not suffer from that annoying and dangerous problem! Please see my squid-bugs email for more details.


> Yes, please implement this solution you have.

Done in dda920d. Please review and adjust as needed. Here is sample output:


```
HTTP/1.1 503 Service Unavailable
X-Squid-Error: ERR_CONNECT_FAIL 111
Vary: Accept-Language
Cache-Status: protofroot;fwd=miss;detail=match
```

src/client_side_reply.cc
```diff
@@ -1562,17 +1561,14 @@ clientReplyContext::buildReplyHeader()
...
-
-#if USE_CACHE_DIGESTS
-    /* Append X-Cache-Lookup: -- temporary hack, to be removed @?@ @?@ */
-    httpHeaderPutStrf(hdr, Http::HdrType::X_CACHE_LOOKUP, "%s from %s:%d",
-                      lookup_type ? lookup_type : "NONE",
-                      getMyHostname(), getMyPort());
-
-#endif
+    SBuf cacheStatus(uniqueHostname());
+    cacheStatus.append(is_hit ? ";hit" : ";fwd=miss");
```
rousskov, line 1565: I suggest dropping `fwd=miss` reporting for now:

* Its current implementation will often lie or violate the Cache-Status definition for the `fwd` parameter _name_: The request may have been served by Squid without forwarding of any kind.
* Its current implementation will often lie or violate the Cache-Status definition for the `miss` parameter _value_: The request may have been forwarded even though Squid had a matching response.
* The parameter presence conveys no new information -- the lack of `hit` is sufficient -- but wastes resources.

I will come back to the `fwd=stale` change [request](https://github.com/squid-cache/squid/pull/705#discussion_r472322494) once this issue is agreed upon.

rousskov, line 1565: S-F parameter serialization [algorithm](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-header-structure-19#section-4.1.1.2) implies no SP after `;`. Our implementation follows that suggestion. AFAICT, _all_ Cache-Status draft examples do not follow that suggestion -- they use `;` SP sequences for all parameters. It could be simply for readability/presentation purposes.

I think we should follow the S-F draft here rather than examples (so no changes are necessary), but I wanted to bring your attention to this mismatch in case you disagree.

yadij:
Sorry, my repo resync pushed this branch without pulling in your patch. Can you re-push please?

src/client_side_reply.cc
yadij: Noticed.

rousskov:
> Can you re-push please?

Done. dda920d is the tip of the PR branch again. Sorry I could not get to it sooner.

yadij:
Thank you, updated properly this time.

src/client_side_reply.cc
yadij: I see. Okay that is acceptable.

yadij:
Okay. Over to you for the fwd=stale part now. Otherwise if you are happy to do that as a separate PR please approve.

yadij:
Thank you. That looks like a reasonable solution to me.

src/client_side_reply.cc
rousskov: Done in d9bb1a3. I have tested only a few outcomes. Please adjust as needed.

BTW, we are mishandling 412 errors, including reporting them as hits, but I do not think we should fix that bug in _this_ PR. This will probably be addressed with a proper fix for bugs [1000](https://bugs.squid-cache.org/show_bug.cgi?id=1000) and [2096](https://bugs.squid-cache.org/show_bug.cgi?id=2096).

```
HTTP/1.1 412 Precondition Failed
X-Squid-Error: ERR_PRECONDITION_FAILED 0
Age: 1598638217
Warning: 113 squid/6.0.0-VCS "This cache hit is still fresh and more than 1 day old"
Cache-Status: protofroot;hit;detail=match
```

src/client_side_reply.h
```diff
@@ -133,12 +129,18 @@ class clientReplyContext : public RefCountable, public StoreClient
...
 
     void sendBodyTooLargeError();
     void sendPreconditionFailedError();
     void sendNotModified();
     void sendNotModifiedOrPreconditionFailedError();
 
+    /// Classification of the initial Store lookup.
+    /// This very first lookup happens without the Vary-driven key augmentation.
+    /// TODO: Exclude internal Store match bans from the "mismatch" category.
+    const char *firstStoreLookup_ = nullptr;
```
rousskov, line 142: I think this field belongs to ALE. Please feel free to move it and the corresponding detailStoreLookup() method to ALE or ask me to do it. I do not insist on this change because it can be viewed as out of scope (it was not this PR that misplaced `lookup_type`, this field's orignal spelling).

rousskov:
I updated PR title/description to match the current PR state. Please adjust further if needed.

PR #706, opened 2020-08-06
Kid restart leads to persistent queue overflows, delays/timeouts

eduard-bagdasaryan:
WARNING: communication with ... may be too slow or disrupted...
    WARNING: abandoning ... I/Os
    ERROR: worker I/O push queue for ... overflow...
    ERROR: Collapsed forwarding queue overflow...

SMP queues rely on the shared memory QueueReader::popSignal flag to
reduce the number of UDS messages that queue readers and writers need to
send each other. If the flag is true but there is no corresponding "wake
up, you have new queued items to read" UDS message for the reader, the
reader may stall. This happens when the reader restarts (e.g., after
hitting an assertion) while the flag is true. A stalled queue reader
leads to delays and queue overflows:

* When the problem affects worker-disker queues, disk I/O delays under
  the hard-coded 7-second timeout are not reported to the admin but may
  affect user experience. Larger delays trigger level-1 WARNINGs. Push
  queue overflows trigger level-1 ERRORs.

* Transient worker-worker queue problems may stall concurrent
  transactions that are reading from the cache entry being written by
  another process. Overflows trigger level-1 ERRORs.

The restarted worker usually starts working just fine because it does
not expect any messages. A busy restarted worker may also appear to
continue working fine because workers always pop queued items before
pushing new ones -- as long as the worker queues new items, it will see
and pop responses to earlier requests, masking the problem. However, the
"stuck popSignal" problem never goes away: Squid only clears the flag
when receiving a notification, but sending new notifications is blocked
by that stuck flag.

Upon kid start, we now clear popSignal (to reflect the actual
communication state) and empty the queue (to reduce overflows). Since
commit 4c21861, any corresponding in-flight UDS queue notification is
ignored because it was sent to the previous process playing the same kid
role. The queue writer will see the false popSignal flag and send a new
notification when queuing a new item, preventing queue stalls.

Also properly ignore stale disker responses about cache_dirs we have not
opened yet, especially since we are now trying to empty the queues ASAP
after a restart, before Coordinator has a chance to inform us about
available diskers, populating the IpcIoFiles container. We already have
similar protection from stale UDS messages and from stale disker queue
messages about _opened_ cache_dirs ("LATE disker response to...").

Also report SMP queue flags in mgr:store_queues.


rousskov:
FTR: This PR was created prematurely by mistake. Ignore it until it exits the draft state. I hope that will happen soon. If we cannot get it out of that draft state soon, we will close it and open a new PR when the code is ready for the official review.

yadij:
The direction of this looks dodgy to me. What happens if a queue is constructed outside of instance startup?
For example; restarting a crashed worker - is this going to reset a lock held by some otherwise nicely operating worker.

rousskov:
> The direction of this looks dodgy to me

Please ignore this PR until it is ready for review.

rousskov (approved):
LGTM, thank you.

> Amos: restarting a crashed worker - is this going to reset a lock held by some otherwise nicely operating worker.

That old question may not be relevant to the current PR code, but I will answer it below just in case it is still relevant.

If you were thinking about QueueReader::popSignal, then the following info might help:

* It is best to think about QueueReader::popSignal as a boolean flag that reflects a specific condition: The queue writer has sent a "wake up -- there is new data in the queue" notification, but the queue reader has not received that notification yet.
* The condition and the flag correspond to a specific queue reader process. If there are multiple reader processes, then there are multiple flags.
* As far as a single N:M Queue API is concerned, there are multiple flags for each reader process (one per unidirectional queue). As far as the actual current N:M queue _implementation_ is concerned, there is just one flag for each reader process. This API/implementation difference explains why we have two flag resetting methods that currently do the same thing. Each caller uses the right API for that caller. Eventually, those method implementations may start to differ.
* When a kid restarts, from the queue point of view, the new kid process will never receive a pending notification (if any) destined to the previous kid process. That effective (or actual) notification loss can be treated as if the writer has not sent any notification (to the current kid process), making the condition above false. The official code left that condition stale, leading to problems.
* I do not want to imply that this PR contains excessive flag clearing code, but clearing this flag one extra time does not create functionality problems -- in the worst case, it leads to an extra UDS notification.

rousskov:
> rousskov added S-waiting-for-committer and removed S-waiting-for-author labels on Nov 1, 2021

I plan to merge this PR if there are no last-minute objections.

yadij:
FYI; last instruction on this was "Please ignore this PR until it is ready for review.".

src/CollapsedForwarding.cc
```diff
@@ -146,6 +151,16 @@ CollapsedForwarding::HandleNotification(const Ipc::TypedMsgHdr &msg)
     HandleNewData("after notification");
 }
 
+/// Handle messages in IPC queue without relying on the queue state being in
```
yadij, line 154: Please explicitly mention here which queue/messages will be "handled" in this manner.

Perhapse some parts of https://github.com/squid-cache/squid/pull/706#pullrequestreview-794271638 to make it clear why other worker process messages are (un)affected.

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -525,6 +533,19 @@ IpcIoFile::HandleNotification(const Ipc::TypedMsgHdr &msg)
         HandleResponses("after notification");
 }
 
+/// Handle messages in IPC queue without relying on the queue state being in
```
yadij, line 536: Same doc update as for CollapsedForwarding::HandleNewDataAtStart

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -525,6 +533,19 @@ IpcIoFile::HandleNotification(const Ipc::TypedMsgHdr &msg)
...
+/// sync with our state (because the queue may be reflecting the state of our
+/// killed predecessor). See also: CollapsedForwarding::HandleNewDataAtStart().
```
yadij, line 538:
```suggestion
/// killed predecessor).
/// \sa CollapsedForwarding::HandleNewDataAtStart()
```

src/CollapsedForwarding.cc
```diff
@@ -146,6 +151,16 @@ CollapsedForwarding::HandleNotification(const Ipc::TypedMsgHdr &msg)
...
+/// sync with our state (because the queue may be reflecting the state of our
+/// killed predecessor). See also: IpcIoFile::HandleMessagesAtStart().
```
yadij, line 156:
```suggestion
/// killed predecessor).
/// \sa IpcIoFile::HandleMessagesAtStart()
```

yadij (changes requested):
Just some doc clarifications. Code is good, so when updated this can be merged.

rousskov:
> FYI; last instruction on this was "Please ignore this PR until it is ready for review.".

FYI: This PR was marked as ready for review for 3 months and labeled as ready for merging since November 2021:

> eduard-bagdasaryan marked this pull request as ready for review 3 months ago
> rousskov added S-waiting-for-committer and removed S-waiting-for-author labels on Nov 1, 2021

src/CollapsedForwarding.cc
rousskov: Done in commit 75fa9f5874ba840998a76c43530d0e9571db46c8.

src/DiskIO/IpcIo/IpcIoFile.cc
rousskov: Same kind of update in commit 68b76da213296061e22b5a343b29466c8fa4ab3d.

Just like in CollapsedForwarding, there is no message discrimination or multiple queues here.

CollapsedForwarding queue and IpcIoFile queue differ, of course, but we should not be saying "IpcIoFile queue" inside IpcIoFile, especially when no other queue is being discussed. By default, everything is "IpcIoFile" inside IpcIoFile.

IpcIoFile::HandleMessagesAtStart() handles the disker (multi)queue. The same top-level code is used for each end of that queue. That is why there is an `if` statement inside IpcIoFile::HandleMessagesAtStart(). There is still only one (multi)queue here, and HandleMessagesAtStart() handles all (incoming) messages in that single queue.

src/DiskIO/IpcIo/IpcIoFile.cc
rousskov: Done in 68b76da213296061e22b5a343b29466c8fa4ab3d.

rousskov (commented):
@yadij, I tried to address all of your concerns. Please re-review.

src/CollapsedForwarding.cc
yadij: Sorry I was quite vague on the request there.  The comment documents the *what*, however reading that info raises huge red flags. Such as why state matters at all if the queue can be used like this (surely its therefore safer to use the queue without the problematic state). Leaving one with worrisome implications -  "relying on" implies some major importance that in turn implies major risks of this function ignoring things.

My request is that you add to that comment, part of the clarifications you gave that explain the *why* such an apparently dangerous thing is both necessary and actually safe to do. I am not sure that I completely grasp it yet, but something like the following might ally the implied danger:
```
/// This method helps restores sanity of shared memory after worker death.
///
/// Handle messages in IPC queue without relying on the queue state
/// (such as number of unhandled messages)
/// being in sync with our state. Because the queue may be reflecting the state
/// of our killed predecessor.
```

rousskov: How about the new current version:

```C++
/// Handle queued IPC messages for the first time in this process lifetime, when
/// the queue may be reflecting the state of our killed predecessor.
```

The biggest problem here is that we want to avoid documenting how the function accomplishes its intended purpose (and even what specific queue states may be possible) because that information is irrelevant to the caller (i.e. to the primary function documentation audience) and because it is complicated/nuanced (i.e. cannot be accurately described in a couple of words). Today, the function resets the state of notification-related flags. Tomorrow, it may also set new/special flags to coordinate restart recovery with the message sender. All that is not what the caller needs/wants to know. Fucntion description should not change when those internal details change.

The caller only needs to know two or three things:

1. The function handles queued messages (the name kind of says that already, but we can detail it a bit to avoid arguing about the need for this function description in the first place).
2. This function must be called upon process startup (the name says that already, but we can detail it a bit to avoid arguing about the need for this function description in the first place).
3. Why the regular HandleNewData() call is not sufficient in this context (technically, this can be viewed as optional/obvious info but it may help developers understand why this special function was added with less thinking).

I will update the other function docs once these ones are settled.

----

> reading that info raises huge red flags. Such as why state matters at all if the queue can be used like this (surely its therefore safer to use the queue without the problematic state).

I do not think those thoughts are huge red flags. They are just questions about complex IPC queue functionality that go beyond any single queue-related method:

Q: Why not remove notifications completely if we can ignore their state at startup?
A: We would not know when to call HandleNewData() during runtime.

Q: Why not reset the whole notification state always/unconditionally, inside HandleNewData()?
A: Squid would send a lot more (slow/expensive) IPC messages. Worker-specific notification state is a notification-reducing optimization.

Q: Does that notification-reducing optimization add risks?
A: While any state, especially shared state, increases risks of bugs (like the one this PR is fixing), notification-reducing optimization does not reduce queue reliability as far as bug-free implementation is concerned.

yadij: Okay. This reads much cleaner. Thank you for addressing this (and the explanations).

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -525,6 +533,20 @@ IpcIoFile::HandleNotification(const Ipc::TypedMsgHdr &msg)
         HandleResponses("after notification");
 }
 
+/// Handle incoming messages in IPC queue without relying on the queue state
+/// being in sync with our state (because the queue may be reflecting the state
+/// of our killed predecessor).
```
yadij, line 538: Still needs the same fear removal that CollapsedForwarding::HandleNewDataAtStart() had.

```suggestion
/// Handle queued IPC messages for the first time in this process lifetime, when
/// the queue may be reflecting the state of our killed predecessor.
```
or, being identical text and we could redux one of them to just use the \sa connection.

```suggestion
```

yadij (changes requested):
Just this outstanding request from initial review. It can be resolved same as 1de63bb.

src/DiskIO/IpcIo/IpcIoFile.cc
eduard-bagdasaryan: Done. For clarity, I decided to make a copy of the description here. I guess we could use \copydoc in this case, but that would render \sa useless.

rousskov: I am working on a way to use `\copydoc`.

eduard-bagdasaryan:
> Just this outstanding request from initial review

@yadij , I think your request had been addressed (up to 5edea59). If so, could you please let this PR move forward?

yadij (approved):
LGTM now. Sorry for the delay getting back to this.

rousskov:
I think we should attempt to backport these kid stability changes into v5, but I do not know how difficult that backport would be. @yadij, if you agree in principle, we can start working on that backport and let you know whether it is practical.

yadij:
> I think we should attempt to backport these kid stability changes into v5, but I do not know how difficult that backport would be. @yadij, if you agree in principle, we can start working on that backport and let you know whether it is practical.

@rousskov, agreed.

yadij:
@rousskov, ```git cherry-pick``` appears to work okay with only trivial conflicts on the exact location of some entirely new functions. I have included it in PR #1010 backports for our normal CI testing. If there are additional tests or delay you want I can pull it out before merge in approx 16hrs.

rousskov:
@yadij, please see https://github.com/squidadm/squid/commit/0d8a666e3f4205087c86b1fa2b4d7dbcf98fa0d9#commitcomment-69920778 for the known (minor) problems with this PR backport to v5. Your call how to handle them.

PR #708, opened 2020-08-07
WCCP: Fix GCC-10 -Wstringop-truncation failures

sergiodj:
I can only trigger these failures when I compile on s390x.

    In function 'char* strncpy(char*, const char*, size_t)', output
        may be truncated copying 8 bytes from a string of length 8

Signed-off-by: Sergio Durigan Junior <sergiodj@debian.org>

### DIFF
--- a/src/wccp2.cc
+++ b/src/wccp2.cc
@@ -49,7 +49,7 @@ static EVH wccp2AssignBuckets;
 
 /* Useful defines */
 #define WCCP2_NUMPORTS  8
-#define WCCP2_PASSWORD_LEN  8
+#define WCCP2_PASSWORD_LEN  8 + 1 /* + 1 for C-string NUL terminator */
 
 /* WCCPv2 Pakcet format structures */
 /* Defined in draft-wilson-wccp-v2-12-oct-2001.txt */
@@ -451,7 +451,7 @@ struct wccp2_service_list_t {
     size_t wccp_packet_size;
 
     struct wccp2_service_list_t *next;
-    char wccp_password[WCCP2_PASSWORD_LEN + 1];     /* hold the trailing C-string NUL */
+    char wccp_password[WCCP2_PASSWORD_LEN];     /* hold the trailing C-string NUL */
     uint32_t wccp2_security_type;
 };
 
@@ -519,8 +519,8 @@ wccp2_add_service_list(int service, int service_id, int service_priority,
     wccp2_update_service(wccp2_service_list_ptr, service, service_id,
                          service_priority, service_proto, service_flags, ports);
     wccp2_service_list_ptr->wccp2_security_type = security_type;
-    memset(wccp2_service_list_ptr->wccp_password, 0, WCCP2_PASSWORD_LEN + 1);
-    strncpy(wccp2_service_list_ptr->wccp_password, password, WCCP2_PASSWORD_LEN);
+    memset(wccp2_service_list_ptr->wccp_password, 0, WCCP2_PASSWORD_LEN);
+    xstrncpy(wccp2_service_list_ptr->wccp_password, password, WCCP2_PASSWORD_LEN);
     /* add to linked list - XXX this should use the Squid dlink* routines! */
     wccp2_service_list_ptr->next = wccp2_service_list_head;
     wccp2_service_list_head = wccp2_service_list_ptr;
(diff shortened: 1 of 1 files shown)


kinkie:
ok to test

src/wccp2.cc
```diff
@@ -553,17 +552,16 @@ static char
...
     struct wccp2_security_md5_t *ws;
 
     debugs(80, 5, "wccp2_update_md5_security: called");
 
     /* The password field, for the MD5 hash, needs to be 8 bytes and NUL padded. */
-    memset(pwd, 0, sizeof(pwd));
-    strncpy(pwd, password, sizeof(pwd));
-    pwd[sizeof(pwd) - 1] = '\0';
+    memset(pwd, 0, WCCP2_PASSWORD_LEN);
+    xstrncpy(pwd, password, WCCP2_PASSWORD_LEN);
```
yadij, line 564: The sizeof(pwd) were correct. Please revert the change to size parameters for both these calls.

```suggestion
    memset(pwd, 0, sizeof(pwd));  // prevent WCCP leaking data from stack area
    xstrncpy(pwd, password, sizeof(pwd));
```

src/wccp2.cc
```diff
@@ -629,9 +627,8 @@ wccp2_check_security(struct wccp2_service_list_t *srv, char *security, char *pac
     /* If execution makes it here then we have an MD5 security */
 
     /* The password field, for the MD5 hash, needs to be 8 bytes and NUL padded. */
-    memset(pwd, 0, sizeof(pwd));
-    strncpy(pwd, srv->wccp_password, sizeof(pwd));
-    pwd[sizeof(pwd) - 1] = '\0';
+    memset(pwd, 0, WCCP2_PASSWORD_LEN);
+    xstrncpy(pwd, srv->wccp_password, WCCP2_PASSWORD_LEN);
```
yadij, line 631: Again here, please leave the sizeof(pwd) as the size parameter.

src/wccp2.cc
```diff
@@ -519,8 +519,7 @@ wccp2_add_service_list(int service, int service_id, int service_priority,
     wccp2_update_service(wccp2_service_list_ptr, service, service_id,
                          service_priority, service_proto, service_flags, ports);
     wccp2_service_list_ptr->wccp2_security_type = security_type;
-    memset(wccp2_service_list_ptr->wccp_password, 0, WCCP2_PASSWORD_LEN + 1);
-    strncpy(wccp2_service_list_ptr->wccp_password, password, WCCP2_PASSWORD_LEN);
+    xstrncpy(wccp2_service_list_ptr->wccp_password, password, WCCP2_PASSWORD_LEN);
```
yadij, line 522: This will truncate 1 character off the end of passwords that fill the buffer because the n parameter to xstrncpy() includes the c-string terminator character.

For these conversions you need to use the length parameter from the memset() as buffer size instead of the pre-shortened one from the old strncpy().

src/wccp2.cc
```diff
@@ -2137,7 +2134,7 @@ parse_wccp2_service(void *)
     if ((t = ConfigParser::NextToken()) != NULL) {
         if (strncmp(t, "password=", 9) == 0) {
             security_type = WCCP2_MD5_SECURITY;
-            strncpy(wccp_password, t + 9, WCCP2_PASSWORD_LEN);
+            xstrncpy(wccp_password, t + 9, WCCP2_PASSWORD_LEN);
```
yadij, line 2137: Please use sizeof() for the size parameter here to prevent string truncation.
```suggestion
            xstrncpy(wccp_password, t + 9, sizeof(wccp_password));
```

src/wccp2.cc
sergiodj: Done.

src/wccp2.cc
sergiodj: Done.

src/wccp2.cc
sergiodj: You're correct, thanks.  Fixed.

src/wccp2.cc
sergiodj: Done.

sergiodj:
Sorry about the mistakes; I was experimenting with `strncat` before finding `xstrncpy`, and apparently got confused by the buffer size parameter.

I force-pushed the branch with the fixes now.  Please let me know what you think.

rousskov:
Thank you for fixing this bug! I have adjusted PR description to be more suitable for the anticipated automated [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message). Please polish further as needed.

yadij:
If you want to go further and polish this up better all the +1 can be moved to the initial #define statement. I do not insist on this change.
```
#define WCCP2_PASSWORD_LEN  8+1  /* +1 for c-string nul terminator */
```

sergiodj:
On Saturday, August 08 2020, Amos Jeffries wrote:

> If you want to go further and polish this up better all the +1 can be moved to the initial #define statement. I do not insist on this change.
> ```
> #define WCCP2_PASSWORD_LEN  8+1  /* +1 for c-string nul terminator */
> ```

Sure, no problem.  I force-pushed with the proposed change.  Let me know
what you think.

Thanks,

-- 
Sergio
GPG key ID: 237A 54B1 0287 28BF 00EF  31F4 D0EB 7628 65FC 5E36
Please send encrypted e-mail if possible
https://sergiodj.net/

yadij:
Cleared for merge under 10-day no objections policy.

PR #709, opened 2020-08-10
Do not send keep-alive in 101 (Switching Protocols) responses

rousskov:
... because it breaks clients using websocket_client[1] library and is
redundant in our HTTP/1.1 control messages anyway.

I suspect that at least some buggy clients are confused by a multi-value
Connection field rather than the redundant keep-alive signal itself, but
let's try to follow RFC 7230 Upgrade example more closely this time and
send no keep-alive at all.

[1] https://pypi.org/project/websocket_client/

### DIFF
--- a/src/servers/Http1Server.cc
+++ b/src/servers/Http1Server.cc
@@ -360,7 +360,8 @@ Http::One::Server::writeControlMsgAndCall(HttpReply *rep, AsyncCall::Pointer &ca
 
     if (switching && /* paranoid: */ upgradeHeader.size()) {
         rep->header.putStr(Http::HdrType::UPGRADE, upgradeHeader.termedBuf());
-        rep->header.putStr(Http::HdrType::CONNECTION, "upgrade, keep-alive");
+        rep->header.putStr(Http::HdrType::CONNECTION, "upgrade");
+        // keep-alive is redundant, breaks some 101 (Switching Protocols) recipients
     } else {
         rep->header.putStr(Http::HdrType::CONNECTION, "keep-alive");
     }


rousskov:
Folks suffering from this incompatibility problem have confirmed that the suggested change fixes it.

rousskov:
@yadij, this PR undoes your (officially committed) [suggestion](https://github.com/squid-cache/squid/pull/481#discussion_r395420083) so you may want to review it.

squidadm:
> Folks suffering from this incompatibility problem have confirmed that the suggested change fixes it.

Do we know how those folks have been testing the patch? live traffic or lab tests?

FTR: Clients failing due to multi-value Connection fields will not be solved by this, only the amount of cases reduced.

The cases my suggestion was for were not obeying HTTP/1.1 requirements about keep-alive based on the HTTP/1.1 label anyway, so breaking them is okay now we have a real-world traffic case needing to do so.

yadij:
> Folks suffering from this incompatibility problem have confirmed that the suggested change fixes it.

Do we know how those folks have been testing the patch? live traffic or lab tests?

FTR: Clients failing due to multi-value Connection fields will not be solved by this, only the amount of cases reduced.

The cases my suggestion was for were not obeying HTTP/1.1 requirements about keep-alive based on the HTTP/1.1 label anyway, so breaking them is okay now we have a real-world traffic case needing to do so.

rousskov:
> Do we know how those folks have been testing the patch? live traffic or lab tests?

No, I do not have direct contact with those folks. My _guess_ is that they are using live traffic at this point.

rousskov:
FTR: A relatively popular service, Google Voice at web.voice.telephony.goog, responds with `HTTP/1.1 400 Bad Request` to Squid Upgrade requests with `Connection: keep-alive' (and successfully establishes a websocket tunnel without that header).

PR #711, opened 2020-08-11
Ignore SMP queue responses made stale by worker restarts

eduard-bagdasaryan:
When a worker restarts (for any reason), the disker-to-worker queue may
contain disker responses to I/O requests sent by the previous
incarnation of the restarted worker process (the "previous generation"
responses). Since the current response:request mapping mechanism relies
on a 32-bit integer counter, and a worker process always starts counting
from 0, there is a small chance that the restarted worker may see a
previous generation response that accidentally matches the current
generation request ID.

For writing transactions, accepting a previous generation response may
mean unlocking a cache entry too soon, making not yet written slots
available to other workers that might read wrong content. For reading
transactions, accepting a previous generation response may mean
immediately serving wrong response content (that have been already
overwritten on disk with the information that the restarted worker is
now waiting for).

To avoid these problems, each disk I/O request now stores the worker
process ID. Workers ignore responses to requests originated by a
different/mismatching worker ID.

### DIFF
--- a/src/DiskIO/IpcIo/IpcIoFile.cc
+++ b/src/DiskIO/IpcIo/IpcIoFile.cc
@@ -92,10 +92,15 @@ operator <<(std::ostream &os, const IpcIo::Command command)
 /* IpcIoFile */
 
 IpcIoFile::IpcIoFile(char const *aDb):
-    dbName(aDb), diskId(-1), error_(false), lastRequestId(0),
+    dbName(aDb),
+    myPid(getpid()),
+    diskId(-1),
+    error_(false),
+    lastRequestId(0),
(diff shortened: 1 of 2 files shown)


rousskov:
@kinkie, could you please check why no Jenkins tests are started for this staged PR? The queue on Jenkins pr-auto [page](https://build.squid-cache.org/job/5-pr-auto/) appears to be stuck in a "Jenkins is about to shut down" state.

rousskov:
@yadij, please mark this fairly important bug-fixing PR for v5 inclusion or let me know if you do not intend to include it.

yadij:
This is on my list for v5 but has not applied cleanly. If you want to make a backport PR that would help.

rousskov:
Will do.

PR #719, opened 2020-08-28
Cleanup helper.h classes

yadij:
Polish up the classes in helper.h to use proper constructors as the
caller API for creation + initialization. Use C++11 initialization for
default values.

* no "virtual" in helper class destructor declaration could create
  memory leaks in future (poorly) refactored code; the gained protection
  is probably worth adding the (currently unused) virtual table to the
  helper class; resolves Coverity issue CID 1461141 (VIRTUAL_DTOR)
* missing Comm::Connection timers initialization on helper startup
* multiple initialization of values used for state accounting
* initialization of booleans to non-0 integer values
* initialization of char using numeric values
* missing mgr:filedescriptors description for helper sockets

### DIFF
--- a/src/helper.h
+++ b/src/helper.h
@@ -158,7 +158,9 @@ private:
 class HelperServerBase: public CbdataParent
 {
 public:
+    HelperServerBase(Ip::Address &, int aPid, void *aIpc, int rfd, int wfd);
     virtual ~HelperServerBase();
+
     /** Closes pipes to the helper safely.
      * Handles the case where the read and write pipes are the same FD.
      *
@@ -231,6 +233,9 @@ class helper_server : public HelperServerBase
     CBDATA_CHILD(helper_server);
 
 public:
+    helper_server(helper *hlp, int pid, void *hIpc, int rfd, int wfd);
+    virtual ~helper_server();
+
     uint64_t nextRequestId = 0;
 
     MemBuf *wqueue = nullptr;
@@ -251,7 +256,6 @@ public:
     typedef std::map<uint64_t, Requests::iterator> RequestIndex;
     RequestIndex requestsIndex; ///< maps request IDs to requests
 
-    virtual ~helper_server();
     /// Search in queue for the request with requestId, return the related
     /// Xaction object and remove it from queue.
     /// If concurrency is disabled then the requestId is ignored and the
@@ -281,7 +285,9 @@ class helper_stateful_server : public HelperServerBase
     CBDATA_CHILD(helper_stateful_server);
 
 public:
+    helper_stateful_server(statefulhelper *hlp, int pid, void *hIpc, int rfd, int wfd);
     virtual ~helper_stateful_server();
+
     void reserve();
     void clearReservation();
 
--- a/src/helper.h
+++ b/src/helper.h
@@ -160,7 +160,7 @@ private:
 class HelperServerBase: public CbdataParent
 {
 public:
-    HelperServerBase(Ip::Address &, int aPid, void *aIpc, int rfd, int wfd);
+    HelperServerBase(const char *desc, Ip::Address &, int aPid, void *aIpc, int rfd, int wfd);
     virtual ~HelperServerBase();
 
     /** Closes pipes to the helper safely.
--- a/src/helper.h
+++ b/src/helper.h
@@ -29,6 +29,9 @@
 #include <queue>
 #include <unordered_map>
 
+class CommTimeoutCbParams;
+class HelperServerBase;
+class MemBuf;
 class Packable;
 class wordlist;
 
@@ -44,7 +47,6 @@ public:
 };
 }
 
-class HelperServerBase;
 /**
  * Managers a set of individual helper processes with a common queue of requests.
  *
@@ -94,18 +96,18 @@ public:
     dlink_list servers;
     std::queue<Helper::Xaction *> queue;
     const char *id_name = nullptr;
-    Helper::ChildConfig childs;          ///< configuration settings for number running
+    Helper::ChildConfig childs; ///< configuration settings for number running
     int ipc_type = 0;
     Ip::Address addr;
-    unsigned int droppedRequests = 0;    ///< requests not sent during helper overload
-    time_t overloadStart = 0;            ///< when the helper became overloaded (zero if it is not)
+    unsigned int droppedRequests = 0; ///< requests not sent during helper overload
+    time_t overloadStart = 0; ///< when the helper became overloaded (zero if it is not)
     time_t last_queue_warn = 0;
     time_t last_restart = 0;
-    time_t timeout = 0;                  ///< requests timeout
-    bool retryTimedOut = false;          ///< whether the timed-out requests must retried
-    bool retryBrokenHelper = false;      ///< whether the requests must retried on BH replies
-    SBuf onTimedOutResponse;             ///< the response to use when helper response timedout
-    char eom = '\n';                     ///< the char which marks the end of (response) message
+    time_t timeout = 0; ///< requests timeout
+    bool retryTimedOut = false; ///< whether the timed-out requests must retried
+    bool retryBrokenHelper = false; ///< whether the requests must retried on BH replies
+    SBuf onTimedOutResponse; ///< the response to use when helper response timedout
+    char eom = '\n'; ///< the char which marks the end of (response) message
 
     struct _stats {
         int requests = 0;
@@ -150,7 +152,7 @@ private:
     void submit(const char *buf, HLPCB * callback, void *data, const Helper::ReservationId & reservation);
     bool trySubmit(const char *buf, HLPCB * callback, void *data, const Helper::ReservationId & reservation);
 
-    ///< reserved servers indexed by reservation IDs
+    /// reserved servers indexed by reservation IDs
     Reservations reservations;
 };
 
@@ -215,16 +217,13 @@ public:
     Requests requests; ///< requests in order of submission/expiration
 
     struct {
-        uint64_t uses = 0;     ///< requests sent to this helper
-        uint64_t replies = 0;  ///< replies received from this helper
-        uint64_t pending = 0;  ///< queued lookups waiting to be sent to this helper
+        uint64_t uses = 0; ///< requests sent to this helper
+        uint64_t replies = 0; ///< replies received from this helper
+        uint64_t pending = 0; ///< queued lookups waiting to be sent to this helper
         uint64_t releases = 0; ///< times release() has been called on this helper (if stateful)
         uint64_t timedout = 0; ///< requests which timed-out
     } stats;
 };
-
-class MemBuf;
-class CommTimeoutCbParams;
 
 // TODO: Rename to StatelessHelperServer and rename HelperServerBase to HelperServer.
 /// represents a single "stateless helper" process
--- a/src/helper.h
+++ b/src/helper.h
@@ -65,21 +65,8 @@ class helper
     CBDATA_CLASS(helper);
 
 public:
-    inline helper(const char *name) :
-        cmdline(NULL),
-        id_name(name),
-        ipc_type(0),
-        droppedRequests(0),
-        overloadStart(0),
-        last_queue_warn(0),
-        last_restart(0),
-        timeout(0),
-        retryTimedOut(false),
-        retryBrokenHelper(false),
-        eom('\n') {
-        memset(&stats, 0, sizeof(stats));
-    }
-    ~helper();
+    explicit helper(const char *name) : id_name(name) {}
+    virtual ~helper();
 
     /// \returns next request in the queue, or nil.
     Helper::Xaction *nextRequest();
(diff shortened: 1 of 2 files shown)


kinkie (approved):
LGTM

src/helper.h
```diff
@@ -245,27 +233,29 @@ class helper_server : public HelperServerBase
     CBDATA_CHILD(helper_server);
 
 public:
-    uint64_t nextRequestId;
+    helper_server(helper *hlp, int pid, void *hIpc, int rfd, int wfd);
```
rousskov, line 236: FYI: Ideally, we should avoid adding function calls with a large number of parameters. We should especially avoid function calls with easy-to-mix-up parameters (like rfd and wfd here). It is better to set each such members explicitly, by name. You may recall multiple bugs caused by interchangeable eventAdd() parameters.

It is a very difficult problem to solve properly, and I do not insist on fixing it in this context, but this PR does create it (while fixing many other problems).

src/helper.h
```diff
@@ -306,14 +298,14 @@ class helper_stateful_server : public HelperServerBase
...
-    statefulhelper *parent;
+    statefulhelper *parent = nullptr;
 
     // Reservations temporary lock the server for an exclusive "client" use. The
     // client keeps the reservation ID as a proof of her reservation. If a
     // reservation expires, and the server is reserved for another client, then
     // the reservation ID presented by the late client will not match ours.
     Helper::ReservationId reservationId; ///< "confirmation ID" of the last
-    time_t reservationStart; ///< when the last `reservation` was made
+    time_t reservationStart = 0;         ///< when the last "reservation" was made
```
rousskov, line 308: I do not know why that word was quoted as if it is a class member before, but I do not think we need to quote it at all. Other comments use "reservation" without quotes already.

```suggestion
    time_t reservationStart = 0; ///< when the last reservation was made
```

src/helper.h
```diff
@@ -144,8 +131,8 @@ class statefulhelper : public helper
 public:
     typedef std::unordered_map<Helper::ReservationId, helper_stateful_server *> Reservations;
 
-    inline statefulhelper(const char *name) : helper(name) {}
-    inline ~statefulhelper() {}
+    explicit statefulhelper(const char *name) : helper(name) {}
+    virtual ~statefulhelper() {}
```
rousskov, line 135: If possible:

```suggestion
    virtual ~statefulhelper() = default;
```

rousskov (dismissed):
Thank you for polishing this.

> and fix several bugs in the old sequences

Those were missing-initialization bugs without known traffic-affecting effects, right?

src/helper.h
```diff
@@ -103,29 +90,29 @@ class helper
...
-    char eom;   ///< The char which marks the end of (response) message, normally '\n'
+    unsigned int droppedRequests = 0;    ///< requests not sent during helper overload
+    time_t overloadStart = 0;            ///< when the helper became overloaded (zero if it is not)
+    time_t last_queue_warn = 0;
+    time_t last_restart = 0;
+    time_t timeout = 0;                  ///< requests timeout
+    bool retryTimedOut = false;          ///< whether the timed-out requests must retried
+    bool retryBrokenHelper = false;      ///< whether the requests must retried on BH replies
+    SBuf onTimedOutResponse;             ///< the response to use when helper response timedout
+    char eom = '\n';                     ///< the char which marks the end of (response) message
```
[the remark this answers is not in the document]
yadij: I have only aligned lines touched for other reasons. It happens that all these members are having assignment added or changes to their documentation text.

src/helper.h
yadij: The initialization bugs I referred to were members of parent class up the hierarchy failing to be initialized correctly (at all) - and others being double-assigned with different values from the caller vs constructor sequence. Having the initialization outside the constructor is exactly what lead to those existing.

Some of the stats values may have been noticed but the stats gathering logic blamed when it was the broken ctor sequence initializing a wrong value.

yadij:
> Thank you for polishing this.
> 
> > and fix several bugs in the old sequences
> 
> Those were missing-initialization bugs without known traffic-affecting effects, right?

Coverity found missing destructor - memory leak on helper exit.

While cleaning I found members being missed by the caller-based initialization or repeatedly initialized with calculation in between not caught by analysis tools because of the non-uninitialized state before the caller sets/overwrites the value.

src/helper.h
rousskov: > The initialization bugs I referred to were members of parent class up the hierarchy failing to be initialized correctly (at all) - and others being double-assigned with different values from the caller vs constructor sequence. Having the initialization outside the constructor is exactly what lead to those existing.

I disagree with you analysis: Lack of initialization is certainly a bug, but it is not caused by post-construction object configuration. And "double-assignment" is not a bug per se. However, this disagreement is currently unimportant AFAICT.

Configuring objects with a large number of parameters is a difficult problem to solve, and no C++11 solution (that I know of) is perfect. Your solution is acceptable IMO. I am resolving this thread.

src/helper.h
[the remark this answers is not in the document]
rousskov: > I have only aligned lines touched for other reasons.

I know. Please interpret this change request with that shared knowledge in mind. If there is something in my change request that seems meaningless or wrong, please let me know.

rousskov:
> Coverity found missing destructor - memory leak on helper exit.
> ... And the FDs not being registered as used with the fd_table.

It would be useful to enumerate those fixes in the PR description, especially the memory leak fix.

FWIW, I failed to find changes adding a missing destructor and changes adding missing fd_note() calls. Please consider rephrasing the description of the fixes if I was looking for the wrong things.

src/helper.h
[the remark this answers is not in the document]
yadij: Done in fecfd5c

rousskov:
I polished PR description to better explain that this PR does not fix a memory leak and to slightly improve second bullet phrasing. Please adjust further if needed.

chtsanti:
Squid crashes after this PR is merged.
It crashes while trying to use external_acls because the helper_server::wqueue is not initiased any more.
Also even if I fix this one the Security helpers, cert validator and cert generators are not working any more.

I am reopening this PR, please close it again if should not reopened.

rousskov:
> I am reopening this PR, please close it again if should not reopened.I am reopening this PR, please close it again if should not reopened.

Amos should probably open a new PR because this one has already been merged. I would support committing a 017a2d1 reversal (in another emergency PR) if Amos cannot start working on fixing the new bug right away. @yadij, what is your availability/preference?

yadij:
I have some time to look into it in the next 2 days. If I do not have something by then feel free to go ahead with a reversal PR in 3 days.

yadij:
Working on it now.

@chtsanti, can you provide me any details you have on detecting the security helpers not working?

chtsanti:
> Working on it now.
> 
> @chtsanti, can you provide me any details you have on detecting the security helpers not working?

Squid exists because all of the certificate generator helper exited. 
At the helper side the helper after sending its firt response (which looks correct), reads 0  bytes and exits.
At squid side while reading helper response reports errno=14 (EFAULT):
 ```Read.cc(146) HandleRead: FD 15, size -1, retval -1, errno 14```

yadij:
It sounds like an EOM value problem, but the eom is not being changed by this PR. I can replicate crash behaviour, but all I can see for that is the correct input being delivered by Squid to the helper, then the helper exiting. Squid receiving EFAULT on read() is expected in that situation.

The helper will need to have debug support added to properly diagnose. Working on that now.

chtsanti:
I am suggesting to revert commit 017a2d1f5.
The PR is closed and looks that we had already forgot the buggy commit in squid.

rousskov:
It has been 20 days since the problems were [reported](https://github.com/squid-cache/squid/pull/719#issuecomment-713668336).

I support commit 017a2d1 reversal, assuming @chtsanti is sure that commit triggered those problems. Reversal is a simple solution that would give @yadij time to address the problem without external pressure. Amos, any better suggestions?

----

Hopefully, the change addressing the problems can be done in separate commit(s) (inside the same PR branch) so that we do not have to review the original changes from scratch (and can clearly see what went wrong).

yadij:
FYI, my investigations have so far identified that the remaining bug is in the helper itself. When it receives "\1\r\n" sequence ending the first message from Squid the helpers message parser terminates the process. Every message line up to that one is apparently handled fine. All the code touched by this PR is apparently working fine.

It will be a few more days before I can do further work on this. The revert plan sounds good if revert is needed.

@christos, since you are more familiar with this helpers internal design perhapse you can find the issue before I can?

rousskov:
> When it receives "\1\r\n" sequence ending the first message from Squid the helpers message parser terminates the process

Does this PR change what the helper _receives_? I did not check the details, but the above `\1\r\n` sequence feels wrong to me -- the EOM symbol (i.e. `\1`) should be the last one in the message.

rousskov:
> > When it receives "\1\r\n" sequence ending the first message from Squid the helpers message parser terminates the process
 
> the above `\1\r\n` sequence feels wrong to me -- the EOM symbol (i.e. `\1`) should be the last one in the message.

```C++
    char eom = '\n'; ///< the char which marks the end of (response) message
```

Moreover, the `eom` field is documented as a _response_ delimiter. Does Squid actually send `\1` _to_ helpers? I do not see any code that would append eom to from-Squid helper messages. For example, Ssl::CrtdMessage::compose() uses chunked-like encoding for queries, with a terminating `\n` added in Ssl::Helper::sslSubmit() and Ssl::CertValidationHelper::sslSubmit(). No `\1`. Did I miss it?

chtsanti:
Squid does not send the \1\r\n sequence at the end of helper request.
The helpers which send multi-lined responses use the '\1' as termination char of the response (instead of '\n' char).

A "strace -p" to the helper shows that the helper receives correctly the request and writes correctly the responce, however imediately after writing its response make a zero read. The squid can not read the response and this is shows an initialization problem.

The problem is that the HelperServerBase::rbuf_sz is initialized to zero. Iniside HelperServerBase constructor while the HelperServerBase::rbuf initialized the rbuf_sz is set.  However re-initialized again later to zero.

The SSL helpers is not the only issue I had the helper_server::wqueue is not initialized correctly too, cause crashes to other helpers too.

The following patch solves the problems for me. However maybe there are other similar issues.

```
-- a/src/helper.cc
+++ b/src/helper.cc
@@ -131,9 +131,9 @@ HelperServerBase::HelperServerBase(const char *desc, Ip::Add
ress &ip, int aPid,
     addr(ip),
     readPipe(new Comm::Connection),
     writePipe(new Comm::Connection),
-    hIpc(aIpc),
-    rbuf(static_cast<char *>(memAllocBuf(ReadBufSize, &rbuf_sz)))
+    hIpc(aIpc)
 {
+    rbuf = static_cast<char *>(memAllocBuf(ReadBufSize, &rbuf_sz));
     readPipe->fd = rfd;
     readPipe->noteStart();
     fd_note(readPipe->fd, desc);
@@ -153,6 +153,7 @@ HelperServerBase::~HelperServerBase()
 
 helper_server::helper_server(helper *hlp, int aPid, void *aIpc, int rfd, int wfd) :
     HelperServerBase(hlp->cmdline->key, hlp->addr, aPid, aIpc, rfd, wfd),
+    wqueue(new MemBuf),
     parent(cbdataReference(hlp))
 {
}
```

rousskov:
@chtsanti, please post a reversal PR. We should not be chasing after various commit 017a2d1 problems _here_. It sucks that _you_ have to do this reversing legwork, but I do not see a better alternative.

> Iniside HelperServerBase constructor while the HelperServerBase::rbuf initialized the rbuf_sz is set

Yeah, that initialization code looked like a red flag to me. I wish I noticed it when looking at the original PR!

yadij:
@christos, that makes some sense. rbuf_sz is listed in the class after rbuf. So its allowed to be initialized (unset) second, though compiler is supposed to be able to recognize it being set already. Odd that the other helpers are not affected by this too.

(Side note: I guess this demonstrates that we should not forbid all assignment in constructors as the new policy proposals say, just avoid when possible.)

Thank you so much. And apologies for not being able to do more on this regression.

chtsanti:
> Side note: I guess this demonstrates that we should not forbid all assignment in constructors as the new policy proposals say, just avoid when possible

Well, if someone wants to forbit all assignments in constructors there are always ways to do it. For example in the case of HelperServerBase we can just remove the default assignment for rbuf_sz (with a comment that this member initialized where the rbuf is initialized). Does not looks correct that there are members with default assignment which also listed in constructor members initializers list.

Or we can use something like the following (to cover the case someone in the future adds default assignment for rbuf_sz):
```
HelperServerBase::HelperServerBase(const char *desc, Ip::Address &ip, int aPid, void *aIpc, int rfd, int wfd) :
...
rbuf(static_cast<char *>(memAllocBuf(ReadBufSize, &rbuf_sz))),
rbuf_sz(rbuf_sz)
{
...
```
But the safer is to initialize both rbuf and rbuf_sz inside constructor.

I suppose this is will be discussed when this PR posted again for merge.

rousskov:
> Amos: compiler is supposed to be able to recognize [`rbuf_sz`] being set already.

When compiling helpers.cc, the compiler cannot "recognize" what happens inside memAlllocBuf() defined in old_api.cc. The compiler can only see that rbuf_sz is not explicitly initialized so it correctly applies default initialization. I bet the compiler would do exactly the same even if it could see memAlllocBuf() definition -- C++ compilers are not supposed to chase pointers like that -- but the actual case we have to explain is easier due to the obvious lack of visibility.

>  Amos: I guess this demonstrates that we should not forbid all assignment in constructors as the new policy proposals say, just avoid when possible.

It sounds like you are misinterpreting my [proposal](https://github.com/squid-cache/squid/pull/691#issuecomment-671087734). I will quote it here to ease comparison:

* all members must be initialized during object construction
* do not initialize class members using assignment operator in a constructor

That proposal forbids _initialization_ by assignment. Other assignments in constructors are, of course, allowed.

To avoid further misunderstanding, in this context, I used the term "initialization" in its general/natural meaning to describe the first time a class member gets a specific value. If we use a stricter/narrower initialization [definition](https://en.cppreference.com/w/cpp/language/initialization) from C++ (which further restricts initialization to member _construction_ time), then we can condense the above two bullets into a single rule:

* all class members must be initialized to specific (i.e. _not_ indeterminate) values.

or, if you prefer,

* do not use class member initialization to indeterminate values.

The term "initialization to indeterminate values" is defined by [C++11|https://en.cppreference.com/w/cpp/language/default_initialization]. Look for the third bullet that starts with "otherwise, nothing is done:...".

rousskov:
> But the safer is to initialize both rbuf and rbuf_sz inside constructor.

It is safer to reset both rbuf and rbuf_sz inside the constructor body. They should be (default-)initialized to nullptr/0 prior to entering that body.

PR #720, opened 2020-08-30
Refactor ACLASN related objects

yadij:
* replace CbDataList of ASN with std::list
* replace calloc/malloc/free with new/delete
* replace C style initialization with C++11 initialization and constructors
* fix memory leak of temporary data objects (should fix Coverity Issue 434128)

### DIFF
--- a/src/acl/Asn.cc
+++ b/src/acl/Asn.cc
@@ -90,7 +90,12 @@ public:
 CBDATA_CLASS_INIT(ASState);
 
 /** entry into the radix tree */
-struct rtentry_t {
+class rtentry_t
+{
+    MEMPROXY_CLASS(rtentry_t);
+public:
+    rtentry_t() { memset(&e_nodes, 0, sizeof(e_nodes)); }
+
     struct squid_radix_node e_nodes[2];
     as_info *e_info = nullptr;
     m_ADDR e_addr;
@@ -396,10 +401,8 @@ asnAddNet(char *as_string, int as_number)
 
     debugs(53, 3, "asnAddNet: called for " << addr << "/" << mask );
 
-    rtentry_t *e = (rtentry_t *)xcalloc(1, sizeof(rtentry_t));
-
+    auto *e = new rtentry_t;
     e->e_addr.addr = addr;
-
     e->e_mask.addr = mask;
 
     rn = squid_rn_lookup(&e->e_addr, &e->e_mask, AS_tree_head);
@@ -413,7 +416,8 @@ asnAddNet(char *as_string, int as_number)
             debugs(53, 3, "asnAddNet: Warning: Found a network with multiple AS numbers!");
             asinfo->as_number.emplace_back(as_number);
         }
-        xfree(e);
+        delete e;
+
     } else {
         auto *asinfo = new as_info;
         asinfo->as_number.emplace_back(as_number);
@@ -422,8 +426,8 @@ asnAddNet(char *as_string, int as_number)
             e->e_info = asinfo;
         } else {
             delete asinfo;
-            xfree(e);
-            debugs(53, 3, "asnAddNet: Could not add entry.");
+            delete e;
+            debugs(53, 3, "could not add entry");
             return 0;
         }
     }
@@ -445,7 +449,7 @@ destroyRadixNode(struct squid_radix_node *rn, void *w)
             debugs(53, 3, "destroyRadixNode: internal screwup");
 
         delete e->e_info;
-        xfree(rn);
+        delete rn;
     }
 
     return 1;
@@ -490,9 +494,8 @@ ACLASN::dump() const
 {
     SBufList sl;
 
-    for (const auto &element : data) {
+    for (const auto &element : data)
         sl.push_back(ToSBuf(element));
-    }
 
     return sl;
 }
--- a/src/acl/Asn.cc
+++ b/src/acl/Asn.cc
@@ -85,9 +85,7 @@ class rtentry_t
 {
     MEMPROXY_CLASS(rtentry_t);
 public:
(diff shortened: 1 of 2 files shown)


kinkie:
looks good; were you able to test the new code?

src/acl/Asn.cc
```diff
@@ -410,48 +390,34 @@ asnAddNet(char *as_string, int as_number)
...
-    }
-
-    if (rn == 0) {      /* assert might expand to nothing */
-        xfree(asinfo);
-        delete q;
-        xfree(e);
-        debugs(53, 3, "asnAddNet: Could not add entry.");
-        return 0;
+        if ((rn = squid_rn_match(&e->e_addr, AS_tree_head))) {
+            e->e_info.emplace_back(as_number);
```
rousskov, line 413: AFAICT, this `e->e_info.emplace_back()` statement is the only case where the allocated `e` is not explicitly deleted in the PR version of asnAddNet(). Let's focus on this specific case. I cannot understand where `e` (i.e. its allocated memory) goes and what code will eventually free it. Knowing the answers is important for several reasons, but primarily because this PR _changes_ how rtentry_t needs to be deleted (and copied/allocated/etc.). Can you answer those two questions, please?

src/acl/Asn.cc
```diff
@@ -96,9 +81,14 @@ class ASState
...
-struct rtentry_t {
+class rtentry_t
+{
+    MEMPROXY_CLASS(rtentry_t);
+public:
+    rtentry_t() { memset(&e_nodes, 0, sizeof(e_nodes)); }
+
     struct squid_radix_node e_nodes[2];
-    as_info *e_info;
+    std::list<int> e_info;
```
rousskov, line 91: If we want to use MEMPROXY_CLASS() for this class, then we should be pooling these `int` allocations as well, right?

yadij: We could, but the two do not need to be identical.

src/acl/Asn.cc
yadij: As far as I am able to understand the radix code squid_rn_addroute() tries to absorb "e" into the AS_tree_head tree structure by setting the e->e_nodes member values.

All other cases "e" has remains owned by this function.

src/acl/Asn.cc
rousskov: I do not know what you mean by "identical" in this context, but adding pooling for rtentry_t is inconsistent with changing a primary rtentry_t component to always do unpooled memory allocations.

This is not my _primary_ concern, but if this PR survives, it should avoid adding these unpooled memory allocations, especially since the same PR tries to add pooling in other areas. I see no justification for this waste and inconsistency.

Rule of thumb: If you add an std::list, you probably need to add pooling for its member wrappers.

src/acl/Asn.cc
rousskov: > squid_rn_addroute() tries to absorb "e" into the AS_tree_head tree structure

AFAICT, squid_rn_addroute() is unaware of the actual `e` type so `e` absorption cannot work correctly for non-trivial rtentry_t objects. This PR makes rtentry_t non-trivial.

rousskov: Please note that I do _not_ know whether your [understanding](https://github.com/squid-cache/squid/pull/720#discussion_r482646904) regarding squid_rn_addroute() is correct. I am only saying that _if_ it is correct, then this PR is wrong.

src/acl/Asn.cc
yadij: Done in 74bad9504

src/acl/Asn.cc
yadij: If you are worried about entries not being delete()'d after they are absorbed into the tree there is the destroyRadixNode() walker which extracts entries and calls delete on them.

src/acl/Asn.cc
rousskov: AFAICT, 74bad95 misses the case this change request was attached to.

src/acl/Asn.cc
rousskov: > If you are worried about entries not being delete()'d after they are absorbed into the tree 

That is one of many things I am worried about. Absorption means taking full control. One cannot have full control of a C++ object without knowing its type unless that type is trivial. This PR makes rtentry_t non-trivial.

The existing/official radix API is badly broken, but we can assume that it more-or-less works or is largely unused because there is no stream of bug reports about the related functionality. Quality replacements of that API should be welcomed. Surgical bug fixes correctly using the existing API should be welcomed. This PR is currently neither.

> there is the destroyRadixNode() walker which extracts entries and calls delete on them.

destroyRadixNode() does not call `delete` on rtentry_t objects we are discussing here, but even if it were doing that, my concerns would remain because we have no guarantees that all these non-trivial objects were properly adsorbed and handled in the low-level radix code that separates our shiny type-specific `new` and `delete`.

Rough summary: Given an `absorb(void*)` API, we cannot change the type of the object (which address we pass to absorb()) from trivial to non-trivial (without changing the API). In most cases, we cannot change the type of the object at all -- even changing one trivial type to another trivial type would require changing the code behind absorb()! Absorbing usually requires understanding the object structure. AFAICT, that is the case with squid_rn_addroute(). Without compiler-checked type safety (which we do not have here due to forced casts), we cannot alter the types of objects passed through that absorbing API, especially in non-trivial ways.

yadij:
Closing due to veto.

rousskov:
> Closing due to veto.

That is a rather misleading way to explain what is wrong with this PR IMO! My change [request](https://github.com/squid-cache/squid/pull/720#discussion_r479816174) was based on a well-founded serious technical concern. That concern is very far from an arbitrary exercise of PR-blocking superpowers. Closing this PR may be the right decision, but my request is not the reason this PR should not be merged.

PR #721, opened 2020-09-04
Cleanup: use C++11 initialization in ConnStateData

yadij:
Also, update some method documentation doxygen syntax.

Resolves Coverity Issue #1231353 - preserveClientData_ member
uninitialized by any constructor sequence.

### DIFF
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -2176,35 +2176,11 @@ clientLifetimeTimeout(const CommTimeoutCbParams &io)
 
 ConnStateData::ConnStateData(const MasterXaction::Pointer &xact) :
     AsyncJob("ConnStateData"), // kids overwrite
-    Server(xact),
-    bodyParser(nullptr),
+    Server(xact)
 #if USE_OPENSSL
-    sslBumpMode(Ssl::bumpEnd),
-    tlsParser(Security::HandshakeParser::fromClient),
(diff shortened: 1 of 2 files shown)


src/client_side.cc
```diff
@@ -2176,35 +2176,11 @@ clientLifetimeTimeout(const CommTimeoutCbParams &io)
 
 ConnStateData::ConnStateData(const MasterXaction::Pointer &xact) :
     AsyncJob("ConnStateData"), // kids overwrite
-    Server(xact),
-    bodyParser(nullptr),
+    Server(xact)
 #if USE_OPENSSL
-    sslBumpMode(Ssl::bumpEnd),
-    tlsParser(Security::HandshakeParser::fromClient),
+    , tlsParser(Security::HandshakeParser::fromClient)
```
rousskov, line 2181: If possible, please move this initialization into the header file as well.

rousskov (commented):
Thank you for cleaning this up, including the comments on the changed-for-other-reasons lines!

It would be nice to list the previously uninitialized member(s) in the PR description.

src/client_side.cc
yadij: Not possible without adding move semantics. This is not a trivial C++11 assignment, but a call to a specific sub-constructor of the Parser class.

yadij:
Cleared for merge under the 10-day criteria.

PR #722, opened 2020-09-05
Return from handleIMSReply when verdict is decided

yadij:
Prevents nullptr de-reference after response object has
been decided and already started sending to client.

Resolves Coverity Scan Issue 1364722

### DIFF
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -448,6 +448,7 @@ clientReplyContext::handleIMSReply(StoreIOBuffer result)
         debugs(88, 3, "request to origin aborted '" << http->storeEntry()->url() << "', sending old entry to client");
         http->logType.update(LOG_TCP_REFRESH_FAIL_OLD);
         sendClientOldEntry();
+        return;
     }
 
     const auto oldStatus = old_entry->mem().freshestReply().sline.status();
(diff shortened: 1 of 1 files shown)


src/client_side_reply.cc
```diff
@@ -472,11 +473,13 @@ clientReplyContext::handleIMSReply(StoreIOBuffer result)
             // forward the 304 from origin
             debugs(88, 3, "origin replied 304, revalidating existing entry and forwarding 304 to client");
             sendClientUpstreamResponse();
+            return;
         } else {
```
kinkie, line 477: do we need the else clause if we return at the end of the if case?

kinkie (approved):
stamp with comments

src/client_side_reply.cc
```diff
@@ -504,13 +509,17 @@ clientReplyContext::handleIMSReply(StoreIOBuffer result)
...
     } else {
         // ignore and let client have old entry
         http->logType.update(LOG_TCP_REFRESH_FAIL_OLD);
         debugs(88, 3, "origin replied with error " <<
                status << ", sending old entry (" << oldStatus << ") to client");
         sendClientOldEntry();
+        return;
     }
+
+    // XXX: unreachable? or sending what to client?
```
rousskov, line 522: I see no reason to leave this important question unresolved in this PR. Unless I am missing something, we should remove all `else` clauses after `return` statements and then (naturally) remove this XXX. Long `else if` lists is rarely a good idea and it looks like it created real bugs in this particular case. I can do that if needed; just let me know.

rousskov (dismissed):

yadij:
I left the line shuffling out to clarify what logic changes were happening. Indentation corrected in 5230e37.

PR #723, opened 2020-09-07
Restored support for non-lowercase Transfer-Encoding values

rousskov:
... after "Improve Transfer-Encoding handling" commit f6dd87e.

Folks are reporting Chunked Transfer-Encoding values in real
traffic. HTTP requires case-insensitve treatment of codings.

### DIFF
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -551,9 +551,9 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         delById(Http::HdrType::CONTENT_LENGTH);
         // and clen state becomes irrelevant
 
-        if (rawTe.caseCmp("chunked")) {
+        if (rawTe.caseCmp("chunked") == 0) {
             ; // leave header present for chunked() method
-        } else if (rawTe.caseCmp("identity")) { // deprecated. no coding
+        } else if (rawTe.caseCmp("identity") == 0) { // deprecated. no coding
             delById(Http::HdrType::TRANSFER_ENCODING);
         } else {
             // This also rejects multiple encodings until we support them properly.
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -551,9 +551,9 @@ HttpHeader::parse(const char *header_start, size_t hdrLen, Http::ContentLengthIn
         delById(Http::HdrType::CONTENT_LENGTH);
         // and clen state becomes irrelevant
 
-        if (rawTe == "chunked") {
+        if (rawTe.caseCmp("chunked")) {
             ; // leave header present for chunked() method
-        } else if (rawTe == "identity") { // deprecated. no coding
+        } else if (rawTe.caseCmp("identity")) { // deprecated. no coding
             delById(Http::HdrType::TRANSFER_ENCODING);
         } else {
             // This also rejects multiple encodings until we support them properly.


rousskov:
Folks are reporting `Chunked` Transfer-Encoding values in real traffic, possibly from `Oracle-iPlanet-Web-Server/7.0`. HTTP requires case-insensitve treatment of codings, and we missed that in f6dd87e. The fix has not been tested yet.

yadij (approved):

yadij:
Polished title and description to clarify that this is a regression, and why it is needed.

rousskov:
Re-adjusted PR title to avoid consenting to a yet another PR/commit title scope without a proper discussion . Lots of Squid commits can be classified as fixing regressions. I do not think it is a good idea to emphasize that fact in the space-limited commit _title_.

I also removed unproven specific server accusation from the PR description. The commit message is a permanent and rather visible record. We should not pummel specific products in there unless we are sure they deserve to be mentioned (and mentioning them is important). I do not have enough evidence to be sure (and identifying a specific produict is not that important in this case).

rousskov:
The (`chunked` part of the) fix has been successfully tested.

yadij:
> Re-adjusted PR title to avoid consenting to a yet another PR/commit title scope without a proper discussion . Lots of Squid commits can be classified as fixing regressions. I do not think it is a good idea to emphasize that fact in the space-limited commit _title_.
> 

Changelog entry will be "Regression fix: ...." it helps avoid confusion if the commit title matches the changelog. Since the earlier commit was security related it is important to highlight this fix as a major behaviour change.

> I also removed unproven specific server accusation from the PR description. The commit message is a permanent and rather visible record. We should not pummel specific products in there unless we are sure they deserve to be mentioned (and mentioning them is important). I do not have enough evidence to be sure (and identifying a specific product is not that important in this case).

Why are you casting this "accusation" then? there is no need to view any product operating within the protocol specification limits as at fault for their design choices. We enforce upper case characters in field names when they are supposed to be lower cased nowdays.

I added it to the description because it is useful to know products needing unusual protocol cases. They can be used for testing or judging importance of those cases being retained in future changes. ie. if we had known of this in advance the earlier commit would have accounted for it.

rousskov:
> Changelog entry will be "Regression fix: ...." it helps avoid confusion if the commit title matches the changelog. 

I do not want to fight about Changelog entries, but the prefix should not be used there either IMO.

> Since the earlier commit was security related it is important to highlight this fix as a major behaviour change.

I do not see the relevance of the earlier commit relationship to security. I do not consider this change to be "major". Squid has accepted `Chunked` for many years (until a recent commit where that support was accidentally dropped) and will resume accepting `Chunked` after this PR. The bug is just a small and temporary hiccup IMO.


> Why are you casting this "accusation" then?

I reported suspected server to help admins suffering from this bug to confirm that they are suffering from it. Since I am not sure that server is indeed involved, I do not want it listed in a more "permanent" record. There is nothing more to that.

> there is no need to view any product operating within the protocol specification limits as at fault for their design choices.

Well, poor design choices can be criticized even when they obey specification limits, but we have better things to do than argue about that.

 
> I added it to the description because it is useful to know products needing unusual protocol cases. They can be used for testing or judging importance of those cases being retained in future changes.

Yes, of course. If I were sure that the said server was involved, then I would have listed it in the commit message myself.

> ie. if we had known of this in advance the earlier commit would have accounted for it.

If I realized that we are dropping case-insensitivity support, we would not drop it even without any specific use cases. Unfortunately, I missed that (implicit) change.

yadij:
> I do not see the relevance of the earlier commit relationship to security. I do not consider this change to be "major". Squid has accepted `Chunked` for many years (until a recent commit where that support was accidentally dropped) and will resume accepting `Chunked` after this PR. The bug is just a small and temporary hiccup IMO.
> 

Since the cause patch is security we have to assume everyone will be applying it to stable proxies ASAP. Making the regression of behaviour not such a minor hiccup as a regular bug would be with normal upgrade cycles.


> > Why are you casting this "accusation" then?
> 
> I reported suspected server to help admins suffering from this bug to confirm that they are suffering from it. Since I am not sure that server is indeed involved, I do not want it listed in a more "permanent" record. There is nothing more to that.
> 

How do you expect anyone to know this can be tested for in X years when the information is hidden away in a temporary message that can be removed at the whim of a commercial providers generosity?

I guess we are differing on interpretation of the statement. The way you wrote it looked to me like a statement of uncertain fact. We have a number of code sections labeled in source and/or in commit history with agents that do or could be used to test unusual behaviour until they becomes obsolete.

rousskov:
> Since the cause patch is security we have to assume everyone will be applying it to stable proxies ASAP. Making the regression of behaviour not such a minor hiccup as a regular bug would be with normal upgrade cycles.

All true but irrelevant to the issue of adding "Regression fix" to the title AFAICT. The commit message will, of course, refer to the commit that created the bug. I was even going to put that commit SHA in the title itself, but since commit messages are often not adjusted when commits are merged into versioned branches, I did not want to increase the confusion. Feel free to move the SHA into the title if you prefer that.

> How do you expect anyone to know this can be tested for in X years

I do not understand the question. I do not expect anybody to test Squid using the agent information in the commit message or PR description, especially in X years. There are already tests that check for proper handling of non-lowercase codings IIRC; we are just not applying them automatically (yet). If there are none, they are easy to create. Testing is not the issue here IMO.

> I guess we are differing on interpretation of the statement. The way you wrote it looked to me like a statement of uncertain fact.

I do not like the combination of "uncertain" and "fact", but yes, my statement has expressed my uncertainty regarding the actual agent generating `Chunked`.

> We have a number of code sections labeled in source and/or in commit history with agents that do or could be used to test unusual behaviour until they becomes obsolete.

There are much easier (and better) ways to test this bug. Knowing the agent is useful not because somebody would use that agent to test Squid. It is useful because somebody who has the agent in their deployment would know that they may be affected by the bug. Unfortunately, we do not know for sure what that agent is.

If you insist on mentioning the agent in the PR description, please restore that part of the PR description (or let me know and I will restore it). Compared to proliferation of vague title prefixes/scopes, this part of the disagreement is not that important to me.

PR #725, opened 2020-09-11
Bug 5076: WCCP Security Info incorrect

craiggowing:
When generating and validating WCCP2 Security Info use only an
8 byte password.

### DIFF
--- a/src/wccp2.cc
+++ b/src/wccp2.cc
@@ -584,7 +584,7 @@ wccp2_update_md5_security(char *password, char *ptr, char *packet, int len)
 
     SquidMD5Init(&M);
 
-    SquidMD5Update(&M, pwd, sizeof(pwd));
+    SquidMD5Update(&M, pwd, sizeof(pwd) - 1);  // Remove the trailing nul
 
     SquidMD5Update(&M, packet, len);
 
@@ -638,7 +638,7 @@ wccp2_check_security(struct wccp2_service_list_t *srv, char *security, char *pac
 
     SquidMD5Init(&M);
 
-    SquidMD5Update(&M, pwd, sizeof(pwd));
+    SquidMD5Update(&M, pwd, sizeof(pwd) - 1);  // Remove the trailing nul
 
     SquidMD5Update(&M, packet, len);
 
--- a/src/wccp2.cc
+++ b/src/wccp2.cc
@@ -584,7 +584,9 @@ wccp2_update_md5_security(char *password, char *ptr, char *packet, int len)
 
     SquidMD5Init(&M);
 
-    SquidMD5Update(&M, pwd, sizeof(pwd) - 1);  // Remove the trailing nul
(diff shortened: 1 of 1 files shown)


yadij (approved):

yadij:
OK to test

src/wccp2.cc
```diff
@@ -584,7 +584,7 @@ wccp2_update_md5_security(char *password, char *ptr, char *packet, int len)
 
     SquidMD5Init(&M);
 
-    SquidMD5Update(&M, pwd, sizeof(pwd));
+    SquidMD5Update(&M, pwd, sizeof(pwd) - 1);  // Remove the trailing nul
```
rousskov, line 587: I think your change is correct. I suggest to clarify and make it more resilient against future changes using something like this:

```suggestion
    static_assert(sizeof(pwd) - 1 == 8, "hashing exactly 8 (padded) octets, excluding storage-terminating NUL");
    SquidMD5Update(&M, pwd, sizeof(pwd) - 1);
```

or, if you prefer less code duplication:



```suggestion
    const auto passwordOctetsToHash = 8;
    static_assert(sizeof(pwd) == passwordOctetsToHash + 1, "password storage has 8 octets plus terminating NUL");
    SquidMD5Update(&M, pwd, passwordOctetsToHash);
```

Same for the other/duplicated code, of course.

yadij: FYI:
 * the size of pwd field may have different sizes in various WCCP versions (eg larger in v3). The magic number here is not just bad practice, but somewhat dangerous with all other pwd related code using the available size macro of sizeof.
 * the value in pwd is padded with multiple NUL's. Most of those are supposed to be passed to the hash algorithm. Only the extra one we tack on for c-string safety is not supposed to be hashed.

So, IMO if we need a static assert at all, it should be checking that sizeof(pwd) >1 bytes. To prevent trying a hash on 0 or negative amounts of data.

rousskov: > the size of pwd field may have different sizes in various WCCP versions

Hopefully, by the time we add support for any version where the size is indeed different, this code will disappear. For now, IMO, it is best to check that the size matches the WCCP version(s) we actually support.

> the value in pwd is padded with multiple NUL's.

The proposed static asset comments attempt to reflect that fact. Better wording welcomed, of course.

In case it is not obvious, I do no insist on these or similar changes. They are just reducing the chances that a similar bug resurfaces as the surrounding code changes.

craiggowing: I've updated the branch with the static assert as suggested, this source file seems specifically for WCCP v2.

> * the size of pwd field may have different sizes in various WCCP versions (eg larger in v3). 

On a side note we see more and more people using ITD as a replacement for WCCP. It seems to be what Cisco is pushing on some of their newer kit, so perhaps they don't intend to further develop the protocol?

rousskov (approved):
Thank you for addressing my concern. I will leave it for Amos to clear this PR for merging in case he disagrees with the changes done after his review.

yadij (approved):

PR #729, opened 2020-09-19
Update ipcache_size and cache_log documentation

gongxini:
The squid website's description of ipcache_size configuration is rather outdated. It says that 

> The ipcache_size defines the maximum number of DNS IP cache entries 

The description doesn't mention the performance impact of tunning the ipcache_size. While increasing the ipcache_size would increase the memory usage, each entry takes really small amounts of memory usage but can avoid DNS lookup. Since the amount of available main memory is enormous these days, it seems to me that we can increase the default value in most scenarios. 

Detail:
I run squid 4.1 on Ubuntu 18.04 as a proxy to visit other web server. The result shows that by increasing the ipcache_size to larger than the averge latency is about 54.458ms. If the ipcache_size is smaller than the total number of dns name, the latency is about 71.428ms. Thus, there is a 31% performance benefit by increasing the ipcache_size.  

The squid website's description about cache_log mentions that :

> The cache_log is squid administrative logging file. This is where general information about Squid behavior goes. You can increase the amount of data logged to this file and how often it is rotated with "debug_options".

The description doesn't clarify the performance impact of cache_log. While it is nice to cache the log information, cache_log would add extra I/O operation and thus would cause performance problem. Therefore, I think it would be great to elaborate the description about the preformance impact.  

Detail:
I run squid 4.1 on Ubuntu 18.04 as a proxy to visit other web server.  I test the cache_log with a heavy benchmarking. The result shows that the average latency is 51ms when enabling the cache_log and the average latency when disabling cache_log is about 43ms. Thus, disabling cache_log would gain about 24% performance benefit.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -5473,7 +5473,9 @@ DOC_START
 
 	This is where general information about Squid behavior goes. You can
 	increase the amount of data logged to this file and how often it is
-	rotated with "debug_options"
+	rotated with "debug_options".
+	Caching the general information would introduce extra I/O and may cause performance issue 
+	when the workload is heavy. The cache_log can be disabled by setting the configuration to cache_log /dev/null 
 DOC_END
 
 NAME: debug_options
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -10070,7 +10070,7 @@ TYPE: int
 DEFAULT: 1024
 LOC: Config.ipcache.size
 DOC_START
-	Maximum number of DNS IP cache entries.
+	Maximum number of DNS IP cache entries. A large ipcache size may improve performance because it avoid more address lookups but excessively large values would waste memory. Since the entry only takes very small amounts of memory and the default value is relatively small, the developer may need to set a large size in production.
 DOC_END
 
 NAME: ipcache_low


src/cf.data.pre
```diff
@@ -5473,7 +5473,9 @@ DOC_START
 
 	This is where general information about Squid behavior goes. You can
 	increase the amount of data logged to this file and how often it is
-	rotated with "debug_options"
+	rotated with "debug_options".
+	Caching the general information would introduce extra I/O and may cause performance issue 
+	when the workload is heavy. The cache_log can be disabled by setting the configuration to cache_log /dev/null
```
yadij, line 5478: This log is not optional. /dev/nulll does *not* disable it, and is a very bad idea.

By default this log receives only critical and important messages admin need to be aware of. Outputs here are controlled by debug_options directive. So performance issues should be mentioned in debug_options documentation.

src/cf.data.pre
```diff
@@ -10070,7 +10072,7 @@ TYPE: int
 DEFAULT: 1024
 LOC: Config.ipcache.size
 DOC_START
-	Maximum number of DNS IP cache entries.
+	Maximum number of DNS IP cache entries. A large ipcache size may improve performance because it avoid more address lookups but excessively large values would waste memory. Since the entry only takes very small amounts of memory and the default value is relatively small, the developer may need to set a large size in production.
```
yadij, line 10075: Caching data may improve performance. This should be obvious information anyone operating a proxy cache - so no need to mention.

Also performance gain is subjective and dependent on latency of the recursive resolver being used by Squid. So it is possible reduction is better than increase. This is just one of many settings that need network-specific testing before performance tuning, we document performance tuning in the wiki instead of repeating it in every relevant directive.

yadij (changes requested):
The texts could be improved, but the suggested changes are not appropriate.

yadij:
> The squid website's description of ipcache_size configuration is rather outdated. It says that

The description is up to date. This directive has not changed since last edit.


> The squid website's description about cache_log mentions that :
> 
> > The cache_log is squid administrative logging file. This is where general information about Squid behavior goes. You can increase the amount of data logged to this file and how often it is rotated with "debug_options".
> 
> The description doesn't clarify the performance impact of cache_log. While it is nice to cache the log information, cache_log would add extra I/O operation and thus would cause performance problem. Therefore, I think it would be great to elaborate the description about the preformance impact.

NOTE: cache_log is not a cache. It is *the* log for Squid.

> 
> Detail:
> I run squid 4.1 on Ubuntu 18.04 as a proxy to visit other web server. I test the cache_log with a heavy benchmarking. The result shows that the average latency is 51ms when enabling the cache_log and the average latency when disabling cache_log is about 43ms. Thus, disabling cache_log would gain about 24% performance benefit.

What you have tested here is not Squid performance with and without a log. You have tested the I/O speed of your disk drive (writing to a file) vs memory (writing to /dev/null).

src/cf.data.pre
rousskov: > Caching data may improve performance. This should be obvious information anyone operating a proxy cache - so no need to mention.

I agree with Amos.

FWIW, It would be good to mention the actual size (in bytes) of a typical IP cache entry. _That_ knowledge is specific to Squid implementation and may be relevant for those wanting very large IP caches or those running on a very tight memory budget. Unfortunately, I do not have that information to suggest a specific edit.

P.S. I see problems with parts of the second paragraph in the change request, but I do not think it is important to agree on those issues because the first paragraph (the one quoted above) is enough to reject the proposed Config.ipcache.size documentation update as "common/obvious knowledge".

rousskov:
> The result shows that the average latency is 51ms when enabling the cache_log and the average latency when disabling cache_log is about 43ms.

@gongxini, what debug_options did you use for that performance test?

> Amos: What you have tested here is not Squid performance with and without a log. You have tested the I/O speed of your disk drive (writing to a file) vs memory (writing to /dev/null).

Well, the two things are not mutually exclusive. A performance of a given Squid instance includes the performance of the disk it is using (if it is using a disk).

Testing Squid performance with debugging enabled is, technically, a valid performance test. We just should not overstate its significance -- most production Squids should run with `debug_options ALL,1` (i.e. default), of course. With that setting, redirecting cache_log to `/dev/null` should have no performance effect on a properly operating Squid. If it does have such an effect, there may be a problem somewhere.

gongxini:
> > The result shows that the average latency is 51ms when enabling the cache_log and the average latency when disabling cache_log is about 43ms.
> 
> @gongxini, what debug_options did you use for that performance test?
> 

I  test with the debug_option as 6 which the wiki says to store more detail of debug information. 

> > Amos: What you have tested here is not Squid performance with and without a log. You have tested the I/O speed of your disk drive (writing to a file) vs memory (writing to /dev/null).
> 
> Well, the two things are not mutually exclusive. A performance of a given Squid instance includes the performance of the disk it is using (if it is using a disk).
> 
> Testing Squid performance with debugging enabled is, technically, a valid performance test. We just should not overstate its significance -- most production Squids should run with `debug_options ALL,1` (i.e. default), of course. With that setting, redirecting cache_log to `/dev/null` should have no performance effect on a properly operating Squid. If it does have such an effect, there may be a problem somewhere.
I agree that we should use debug level 1 in production but it would be nice to mention the performance impact of other debug level in the document since it seems unclear to me how many overhead should I pay to increase the debug_option level.

src/cf.data.pre
gongxini: Thanks for your answer. I was search on google for how to diable the cache_log and some answer suggest that I can disable it with /dev/null. I am not very familiar with the detail implementation of cache_log. Why it is a bad idea to set it to /dev/null?

src/cf.data.pre
gongxini: I think documenting the size of a IP cache entry would to good for users to decide the cache size. When I am tunning this configuration,  I felt that tunning this configuration is not that obvious since the performance gain and the memory cost of increasing the ipcache_size is unclear to me. While a large cache would definilty improve the performacne but it requires extra memory.   I search on wiki and find the following two pages: https://wiki.squid-cache.org/Features/CacheManager/IpCache and https://wiki.squid-cache.org/ProgrammingGuide/IpCache, but none of them mentions about how to tune this configuration.  there any document that I miss?

gongxini:
> > The squid website's description of ipcache_size configuration is rather outdated. It says that
> 
> The description is up to date. This directive has not changed since last edit.
> 
Sorry, my word is a little bit confused. What I mean is that the document doesn't mention the performance impact of changing the configuration 

> > The squid website's description about cache_log mentions that :
> > > The cache_log is squid administrative logging file. This is where general information about Squid behavior goes. You can increase the amount of data logged to this file and how often it is rotated with "debug_options".
> > 
> > 
> > The description doesn't clarify the performance impact of cache_log. While it is nice to cache the log information, cache_log would add extra I/O operation and thus would cause performance problem. Therefore, I think it would be great to elaborate the description about the preformance impact.
> 
> NOTE: cache_log is not a cache. It is _the_ log for Squid.
> 

Sorry, it is a typo.

> > Detail:
> > I run squid 4.1 on Ubuntu 18.04 as a proxy to visit other web server. I test the cache_log with a heavy benchmarking. The result shows that the average latency is 51ms when enabling the cache_log and the average latency when disabling cache_log is about 43ms. Thus, disabling cache_log would gain about 24% performance benefit.
> 
> What you have tested here is not Squid performance with and without a log. You have tested the I/O speed of your disk drive (writing to a file) vs memory (writing to /dev/null).

rousskov:
> I test with the debug_option as 6

By design, production Squids should not run with debugging levels higher than 1. Higher levels slow Squid down a lot, especially when logging to a disk. Such elevated debugging levels are only meant for problem triage. Unless you are interested in measuring performance of the debugging code, avoid setting `debug_options` to anything other than `ALL,1` (the default) in your performance tests.

While I believe that the principles sketched above apply to most programs and are fairly common knowledge among sysadmins, I would not be violently against adding it to squid.conf.documented because I have seen quite a few sysadmins running with debugging needlessly turned on.

rousskov:
> it seems unclear to me how many overhead should I pay to increase the debug_option level.

In general, there is no need to pay anything: Level-2+ debugging information is useless (and is harmful) in regular production runs. A (properly configured) access log should have the transaction details you may need.

If you can get high-quality estimates of performance overheads of various debugging levels, we should publish that information on Squid wiki. Personally, I do not know how much performance overhead each level introduces. IIRC, most high-performance tests in our lab do not run well with levels above 2 or 3, even when logging to a memory-based partition. YMMV.

src/cf.data.pre
rousskov: > Why it is a bad idea to set cache_log to /dev/null?

... because you will not see potentially critical Squid warnings and errors, including those about misconfigurations and environmental problems (unless you properly setup/enable syslog recording of those messages).

src/cf.data.pre
yadij: That is difficult, because an ipcache entry is dynamic. Size depends on the number of IPs stored and the length of the domain name used to find them.

The only thing we can say for certain is that larger cache entry limit == more memory used. And that is pretty obvious.

FYI: https://wiki.squid-cache.org/SquidFaq/SquidMemory is where memory tuning for performance is documented.

rousskov: > That is difficult, because an ipcache entry is dynamic. Size depends on the number of IPs stored and the length of the domain name used to find them.

Once the sizes of individual components (s4, s6, and c) are known, I see no difficulty in providing a basic formula for the cache size (something along the lines of `n4*s4 + n6*s6 + nd*(sd + c)` but I have not checked the code to come up with the actual formula) so that inquisitive admins can approximate the average entry size for their use cases. And I hope that Cache Manager already reports the actual entries and/or their stats to assist with those calculations.

We can also provide a ballpark entry size estimate for a basic case with one IPv4 and one IPv6 address for some specific average-length domain name.

gongxini:
> > it seems unclear to me how many overhead should I pay to increase the debug_option level.
> 
> In general, there is no need to pay anything: Level-2+ debugging information is useless (and is harmful) in regular production runs. A (properly configured) access log should have the transaction details you may need.
> 
> If you can get high-quality estimates of performance overheads of various debugging levels, we should publish that information on Squid wiki. Personally, I do not know how much performance overhead each level introduces. IIRC, most high-performance tests in our lab do not run well with levels above 2 or 3, even when logging to a memory-based partition. YMMV.

Thank you for the explanation. I can run a comprehensive performance test on each debug level and provide the information.  Sorry for the late reply, I am very busy this weekday.

rousskov:
> I can run a comprehensive performance test on each debug level and provide the information.

If you _want_ to do it and can do it well, then please do it (including publishing the information on the wiki) -- high-quality information in this area is occasionally useful for Squid admins.

src/cf.data.pre
gongxini: Thanks for your answer! It is very useful.

yadij:
@gongxini, this PR is waiting on your decision whether you want to update the documentation with different texts in light of our comments or to close.

gongxini:
> @gongxini, this PR is waiting on your decision whether you want to update the documentation with different texts in light of our comments or to close.

Hi Amos, I did some experiments about the performance overhead for different option level and want to share on squid wiki page. I sent an email to ask for a squid wiki account last Friday but still wait for it. I think I can update the experiment information on wiki and close this one for one. Thanks for reminding me!

PR #734, opened 2020-10-02
Removed StoreClient::created() and improved PURGE code quality

eduard-bagdasaryan:
The StoreClient::created() callback method was probably added in hope to
make Store lookups asynchronous, but that functionality has not been
implemented, leaving convoluted and dangerous synchronous created() call
chains behind. Moreover, properly supporting asynchronous Store lookups
in modern code is likely to require a very different API.

Removal of created() allowed to greatly simplify PURGE processing,
eliminating some tricky state, such as `purging` and `lookingforstore`.

Also removed all Store::getPublic*() methods as no longer used.

### DIFF
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -951,6 +951,18 @@ clientReplyContext::purgeDoPurge()
 {
     auto firstFound = false;
     if (const auto entry = storeGetPublicByRequestMethod(http->request, Http::METHOD_GET)) {
+        // special entries are only METHOD_GET entries without variance
+        if (EBIT_TEST(entry->flags, ENTRY_SPECIAL)) {
+            http->logType.update(LOG_TCP_DENIED);
+            Ip::Address tmp_noaddr;
+            tmp_noaddr.setNoAddr(); // TODO: make a global const
+            auto err = clientBuildError(ERR_ACCESS_DENIED, Http::scForbidden, nullptr,
+                                        http->getConn() ? http->getConn()->clientConnection->remote : tmp_noaddr,
+                                        http->request, http->al);
+            startError(err);
+            entry->abandon(__FUNCTION__);
+            return;
+        }
         firstFound = true;
         if (!purgeEntry(*entry, Http::METHOD_GET))
             return;
@@ -1004,19 +1016,6 @@ bool
 clientReplyContext::purgeEntry(StoreEntry &entry, const Http::MethodType methodType, const char *descriptionPrefix)
 {
     debugs(88, 4, descriptionPrefix << Http::MethodStr(methodType) << " '" << entry.url() << "'" );
-
-    if (EBIT_TEST(entry.flags, ENTRY_SPECIAL)) {
-        http->logType.update(LOG_TCP_DENIED);
-        Ip::Address tmp_noaddr;
-        tmp_noaddr.setNoAddr(); // TODO: make a global const
-        auto err = clientBuildError(ERR_ACCESS_DENIED, Http::scForbidden, nullptr,
-                                    http->getConn() ? http->getConn()->clientConnection->remote : tmp_noaddr,
-                                    http->request, http->al);
-        startError(err);
-        entry.abandon(__FUNCTION__);
-        return false;
-    }
-
 #if USE_HTCP
     neighborsHtcpClear(&entry, http->request, HttpRequestMethod(methodType), HTCP_CLR_PURGE);
 #endif
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -1029,7 +1029,7 @@ clientReplyContext::purgeDoPurge()
 }
 
 void
-clientReplyContext::purgeEntry(StoreEntry &entry, const Http::MethodType &methodType, const char *descriptionPrefix)
+clientReplyContext::purgeEntry(StoreEntry &entry, const Http::MethodType methodType, const char *descriptionPrefix)
 {
     debugs(88, 4, descriptionPrefix << Http::MethodStr(methodType) << " '" << entry.url() << "'" );
 #if USE_HTCP
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -956,7 +956,7 @@ clientReplyContext::purgeDoPurge()
             http->logType.update(LOG_TCP_DENIED);
             Ip::Address tmp_noaddr;
             tmp_noaddr.setNoAddr(); // TODO: make a global const
-            auto err = clientBuildError(ERR_ACCESS_DENIED, Http::scForbidden, nullptr,
+            const auto err = clientBuildError(ERR_ACCESS_DENIED, Http::scForbidden, nullptr,
                                         http->getConn() ? http->getConn()->clientConnection->remote : tmp_noaddr,
                                         http->request, http->al);
             startError(err);
@@ -1591,7 +1591,7 @@ clientReplyContext::identifyStoreObject()
     // client sent CC:no-cache or some other condition has been
     // encountered which prevents delivering a public/cached object.
     if (!r->flags.noCache || r->flags.internal) {
-        auto e = storeGetPublicByRequest(r);
+        const auto e = storeGetPublicByRequest(r);
         identifyFoundObject(e, storeLookupString(bool(e)));
     } else {
         // "external" no-cache requests skip Store lookups
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -989,15 +989,15 @@ clientReplyContext::purgeRequest()
 void
 clientReplyContext::purgeDoPurge()
 {
-    bool firstFound = false;
-    if (auto entry = storeGetPublicByRequestMethod(http->request, Http::METHOD_GET)) {
+    auto firstFound = false;
+    if (const auto entry = storeGetPublicByRequestMethod(http->request, Http::METHOD_GET)) {
         firstFound = true;
         purgeEntry(*entry, Http::METHOD_GET);
     }
 
     detailStoreLookup(storeLookupString(firstFound));
 
-    if (auto entry = storeGetPublicByRequestMethod(http->request, Http::METHOD_HEAD))
+    if (const auto entry = storeGetPublicByRequestMethod(http->request, Http::METHOD_HEAD))
         purgeEntry(*entry, Http::METHOD_HEAD);
 
     /* And for Vary, release the base URI if none of the headers was included in the request */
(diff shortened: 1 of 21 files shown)


src/client_side_reply.cc
```diff
@@ -1122,6 +1088,22 @@ clientReplyContext::purgeDoPurgeHead(StoreEntry *newEntry)
     http->storeEntry()->complete();
 }
 
+void
+clientReplyContext::purgeEntry(StoreEntry *entry, const Http::MethodType &methodType, const char *descriptionPrefix)
+{
+    if (!entry) {
+        if (purgeStatus == Http::scNone)
+            purgeStatus = Http::scNotFound;
```
rousskov, line 1096: AFAICT, this is a PR bug: purgeStatus may become scNotFound prematurely and will stay that way despite matching entries being found later.

In general, I think we should avoid changing purgeStatus to an incorrect status that would have to be changed later. That implies not changing purgeStatus to scNotFound until we know that no matching entries have been found. IIRC, this was discussed earlier, and the current official code follows that principle.

rousskov (dismissed):
Thank you for polishing this code.

After my concern is addressed (see the inlined change request), I encourage you to go one step further, replacing the current purgeDoPurgeGet()-purgeDoPurgeHead() chain with a single purgeDoPurge() method, removing the `lookingforstore = 4` hack. AFAICT, we just need to replace pretending to be asynchronous getPublicByRequestMethod() call with a direct storeGetPublicByRequestMethod() call. This simplification may help reduce the "when do we finalize purgeStatus" concerns that keep bugging this code. 

I do not insist on this further polishing.

src/client_side_reply.cc
eduard-bagdasaryan: > this is a PR bug: purgeStatus may become scNotFound prematurely

Sorry, I don't see a bug here. Initially, purgeStatus is scNone. When clientReplyContext::purgeEntry() is called the first time, it becomes either scNotFound or scOkay. Further calls (if an entry was found) will make it scOkay.
Also, IMO, it is more reliable to keep purgeStatus 'up-to-date' at each step. In other words, we should not rely on some other code, that will probably be called in the future and set it to scNotFound (if nothing was found by then).

rousskov: > Further calls (if an entry was found) will make it scOkay.

Yes, you are right, sorry.

> Also, IMO, it is more reliable to keep purgeStatus 'up-to-date' at each step.

I agree. We disagree what "up to date" means (or what "purgeStatus" is). To me, in the _middle_ of the processing steps, the value scNotFound is not an "up to date" value for purgeStatus because we have not reached that conclusion at any point in time yet. However, I can certainly see the other point of view as well. I prefer avoiding scNotFound status until the end because doing so might help with exposing bugs where the processing sequence got interrupted early.

One way to avoid this problem itself is to replace purgeStatus with a purgedEntries counter. A counter would also carry more information for triage!

eduard-bagdasaryan:
>  replacing the current purgeDoPurgeGet()-purgeDoPurgeHead() chain with a single purgeDoPurge() 

I found this advice useful, done at 84408a5.

> This simplification may help reduce the "when do we finalize purgeStatus" concerns

Yes, it helped with it, done at 28f4155.

yadij:
FTR, when refactoring this PURGE method code please plan towards isolating its logic as much as possible so it can be dropped easily later. It is already just waiting on better HTCP tooling support for formal deprecation.

So, thank you for this work. It already goes a long way towards that goal.

Before I review, is this PR scope just cleanup/refactor of the PURGE method code? or there something like a bug fix to consider too.

rousskov:
> Before I review, is this PR scope just cleanup/refactor of the PURGE method code? or there something like a bug fix to consider too.

I am not aware of any bug fixes in this PR. However, I recommend waiting a few days before reviewing this PR: We have discussed the current status, and @eduard-bagdasaryan was going to make a couple of corrections. More importantly, Eduard was evaluating whether it makes sense to go a step or two further, eliminating StoreClient::created() itself (another old TODO).

yadij:
Okay, will leave code review a while.

For design review please consider;

PURGE is essentially idempotent. Repeating it should not be dangerous at all. There is also a (weak) information disclosure issue in regards to delivering a 4xx response when the PURGE is performed on an empty cache.

We could workaround all the issues by spawning an AsyncJob to do the lookup and erase action(s) and,
 * immediately reply to the client with 202 Accepted (no async completion handler), or
 * wait for completion and always respond 200 OK to client only when done.
 * maybe with a timeout handler delivering 1xx or 202 Accepted if it takes too long.
 * respond 405 Method Not Allowed if purge feature is disabled.

The 202 possibility allows for a slow background scan for variant entries.
The AsyncJob can/should be shared between PURGE method and HTCP CLR.

rousskov:
I see a lot of problems with your comment, on several levels, but I hope that we do not need to agree on any of those issues here/now because any related changes would be outside _this_ PR scope. If somebody wants to implement any of those changes in the future, they should discuss them first.

eduard-bagdasaryan:
> is this PR scope just cleanup/refactor of the PURGE method code

The initial intention was to refactor a few PURGE methods. However, when some clientReplyContext methods were removed and clientReplyContext::created() itself became empty, we decided to go a little further and address another TODO (about eliminating StoreClient::created(). Note that this PR scope is still just about refactoring.

eduard-bagdasaryan:
I am done with changes for now. @yadij , please re-review.

src/client_side_reply.cc
```diff
@@ -1031,74 +942,53 @@ clientReplyContext::purgeRequest()
...
-        purgeStatus = Http::scOkay;
+clientReplyContext::purgeDoPurge()
+{
+    auto firstFound = false;
+    if (const auto entry = storeGetPublicByRequestMethod(http->request, Http::METHOD_GET)) {
+        // special entries are only METHOD_GET entries without variance
+        if (EBIT_TEST(entry->flags, ENTRY_SPECIAL)) {
+            http->logType.update(LOG_TCP_DENIED);
+            Ip::Address tmp_noaddr;
+            tmp_noaddr.setNoAddr(); // TODO: make a global const
```
yadij, line 958: This TODO can be implemented easily:
```suggestion
            static Ip::Address tmp_noaddr = Ip::Address::v6_noaddr;
```

src/icp_v3.cc
```diff
@@ -49,39 +48,32 @@ doV3Query(int fd, Ip::Address &from, char *buf, icp_common_t header)
...
-    // TODO: StoreClients must either store/lock or abandon found entries.
-    //if (e)
-    //    e->abandon();
+    if (e)
+        e->abandon(__FUNCTION__);
 
-    delete this;
 }
 
+ICP3State::~ICP3State()
```
yadij, line 74: Since git thinks this code is being changed, please take the opportunity to inline this destructor with "=default".

yadij (changes requested):
Just two bits of polish after the latest changes.

src/client_side_reply.cc
rousskov: The above suggestion does not look like a good implementation of this out-of-scope TODO to me, but if we can indeed use Ip::Address::v6_noaddr in this context, then I would not violently object to properly implementing this TODO, even though it will also require changing clientBuildError() profile to accept Ip::Address::v6_noaddr.

eduard-bagdasaryan: >  I would not violently object to properly implementing this TODO

Done at 7c72c484.

src/icp_v3.cc
eduard-bagdasaryan: Done at 4149d8f0.

eduard-bagdasaryan (commented):
Done, please re-review.

yadij (approved):

src/icp_v2.cc
```diff
@@ -536,15 +502,40 @@ doV2Query(int fd, Ip::Address &from, char *buf, icp_common_t header)
...
+    state.url = xstrdup(url);
+    state.flags = flags;
+    state.rtt = rtt;
+    state.src_rtt = src_rtt;
+
+    const auto e = storeGetPublic(url, Http::METHOD_GET);
+
+    icp_opcode codeToSend;
+
+    if (e && state.confirmAndPrepHit(*e)) {
```
rousskov, line 517: Judging by the current use cases, confirmAndPrepHit(e) should stay private and we should add a public boolean isHit(void) method that will call storeGetPublic(), confirmAndPrepHit(), and `e->abandon()`, all without "leaking" `e` outside.

src/ip/Address.h
```diff
@@ -298,6 +298,10 @@ class Address
      */
     bool GetHostByName(const char *s);
 
+    /// \returns an object containing the specific IP case NO_ADDR (format-neutral).
```
rousskov, line 301: Let's simplify this and avoid duplicating (rather questionable!) details:

```suggestion
    /// \returns an Address for which isNoAddr() is true
```

or

```suggestion
    /// \returns an Address with true isNoAddr()
```

src/ip/Address.h
```diff
@@ -298,6 +298,10 @@ class Address
...
+    /// see isNoAddr() for more detail.
```
rousskov, line 302: I know you are copying this phrase, but I would polish it:

```suggestion
    /// \see isNoAddr() for more details
```

src/urn.cc
```diff
@@ -175,29 +174,17 @@ UrnState::start(HttpRequest * r, StoreEntry * e)
...
-    if (!e || (e->hittingRequiresCollapsing() && !startCollapsingOn(*e, false))) {
+    if (!urlEntry || (urlEntry->hittingRequiresCollapsing() && !startCollapsingOn(*urlEntry, false))) {
         urlres_e = storeCreateEntry(urlres, urlres, RequestFlags(), Http::METHOD_GET);
         sc = storeClientListAdd(urlres_e, this);
         FwdState::Start(Comm::ConnectionPointer(), urlres_e, urlres_r.getRaw(), ale);
-        // TODO: StoreClients must either store/lock or abandon found entries.
-        //if (e)
-        //    e->abandon();
+        if (urlEntry)
+            urlEntry->abandon(__FUNCTION__);
```
rousskov, line 184: Let's not leak a pointer to the destroyed urlEntry:

```suggestion
        if (urlEntry) {
            urlEntry->abandon(__FUNCTION__);
            urlEntry = nullptr;
        }
```

We probably should add a `StoreEntry::SafelyAbandon(StoreEntry *&e, const char * context)` static method that will implement it (similar to the existing safe_free() macro), but I do not insist on that because the other abandon() callers touched in this PR currently quit right after they invalidate their StoreEntry pointer.

rousskov (changes requested):
Excellent progress! I did not review low-level details but suggested a few minor adjustments.

src/urn.cc
eduard-bagdasaryan: Done.

> I do not insist on that

Left as is.

eduard-bagdasaryan (commented):
Done (c7b027e).

rousskov (approved):
It is difficult (for me) to be absolutely sure about all the low-level code adjustments, but I think this cleanup is very valuable overall and should be accepted.

rousskov:
@eduard-bagdasaryan, thank you for addressing my earlier concerns. Please double check my two recent commits:

* b2f8d14 fixup: Do not use raw opcode as an array index
* 6a6f631 Improved StoreClient API const-correctness

@yadij, there were a few non-trivial changes since you have approved this PR. I will not clear it for a couple of days to give you a chance to revisit it.

eduard-bagdasaryan:
> Please double check my two recent commits

These commits look OK.

yadij (approved):
LGTM

PR #737, opened 2020-10-16
support multi connection

muhanadali100:
why don't support multi connection cache


rousskov:
@muhanadali100, it sounds like you are trying to ask a question about Squid. The pull request interface is not intended for that so I am closing this pull request.

Your question may already be answered in Squid [FAQ](https://wiki.squid-cache.org/SquidFaq/AboutSquid#How_to_add_a_new_Squid_feature.2C_enhance.2C_of_fix_something.3F). If not, you may ask your question on the squid-users mailing [list](http://www.squid-cache.org/Support/mailing-lists.html#squid-users) instead. If you do, I would recommend rephrasing it to make it clear whether you are asking for a new feature to be added to Squid or complaining about a problem with the existing functionality.

PR #739, opened 2020-10-18
Fix std::ostream precision settings leak in Progress::print()

rousskov:
No functionality changes expected.

Broken since inception (commit 8ecbe78d).

Detected by Coverity. CID 1468264: API usage errors(STREAM_FORMAT_STATE)

### DIFF
--- a/src/store_rebuild.cc
+++ b/src/store_rebuild.cc
@@ -247,7 +247,7 @@ Progress::print(std::ostream &os) const
         const auto savedPrecision = os.precision(2);
         const auto percent = 100.0 * completed / goal;
         os << percent << "% (" << completed << " out of " << goal << ")";
-        os.precision(savedPrecision);
+       (void)os.precision(savedPrecision);
     } else if (!completed && !goal) {
         os << "nothing to do";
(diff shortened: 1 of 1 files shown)


src/store_rebuild.cc
```diff
@@ -244,9 +244,10 @@ void
 Progress::print(std::ostream &os) const
 {
     if (goal > 0) {
+        const auto savedPrecision = os.precision(2);
         const auto percent = 100.0 * completed / goal;
-        os << std::setprecision(2) << percent << "% (" <<
-            completed << " out of " << goal << ")";
+        os << percent << "% (" << completed << " out of " << goal << ")";
+        os.precision(savedPrecision);
```
eduard-bagdasaryan, line 250:
```suggestion
       (void)os.precision(savedPrecision);
```
since we ignore the return value. However, I am not sure that we maintain this rule in the code.

yadij: We do not enforce it. But it is good practice, so if you want to please do.

yadij (approved):

PR #740, opened 2020-10-19
Bug 5073: Compile error: index was not declared in this scope

yadij:
Use strchr(3) instead of a legacy POSIX.1-2001 index(3) API.

Also removed the index() implementation on MS Windows as no longer used.


rousskov (approved):
Thank you for fixing this.

I polished the PR description a bit to make it more succinct and supply an additional argument for index() removal. Please undo if you prefer your version.

PR #741, opened 2020-10-21
cachemgr.cgi: Fix regression caused by Bug 4957

brozs:
In ```make_pub_auth()```; the ```rfc1738_escape()``` function is not reentrant and
cannot be used in a single ```snprintf()``` call twice.

This fixes a problem where users can authenticate to the CGI interface successfully but cannot use any of the links in the menu (returns HTTP 403) and the following error is in the ```cache.log```:

```
kid1| CacheManager: unknown@local=127.0.0.1:3128 remote=127.0.0.1:47736 FD 10 flags=1: password needed for 'menu'
```
or
```
kid1| CacheManager: <username>@local=127.0.0.1:3128 remote=127.0.0.1:55484 FD 12 flags=1: incorrect password for 'menu'
```


tools/cachemgr.cc
```diff
@@ -1086,20 +1086,29 @@ static void
...
+
     safe_free(req->pub_auth);
     debug("cmgr: encoding for pub...\n");
 
     if (!req->passwd || !strlen(req->passwd))
         return;
 
+    /*rfc1738_escape is not reentrant */
+    rfc1738_username = xstrdup(rfc1738_escape(safe_str(req->user_name)));
+    rfc1738_passwd = xstrdup(rfc1738_escape(req->passwd));
```
yadij, line 1099: Please update this to C++11 style. This is not C code.

```suggestion
    auto *rfc1738_username = xstrdup(rfc1738_escape(safe_str(req->user_name)));
    auto *rfc1738_passwd = xstrdup(rfc1738_escape(req->passwd));
```

yadij (changes requested):

yadij:
OK to test

tools/cachemgr.cc
brozs: Just to double-check, this is the CGI code "cachemgr.cc" which seems to be C and "auto *" isn't used anywhere else in that code. Can you please confirm?

rousskov: All `.cc` files are treated as C++ code. In most cases, we use modern C++ features when we add or change code, and cachemgr.cc is one of those neglected code areas that is rarely enhanced or touched... You do not _have to_ use `auto` in this PR -- such polishing is optional -- but please do it if you can.

tools/cachemgr.cc
```diff
@@ -1086,20 +1086,29 @@ static void
...
     static char buf[1024];
+    char *rfc1738_username = NULL, *rfc1738_passwd = NULL;
+
     safe_free(req->pub_auth);
     debug("cmgr: encoding for pub...\n");
 
     if (!req->passwd || !strlen(req->passwd))
         return;
 
+    /*rfc1738_escape is not reentrant */
```
rousskov, line 1097: rfc1738_escape() is not reentrant indeed, but "reentrant" is not really the right technical term for describing this bug concern (even a perfectly reentrant function may still be unusuable in this debugs() context), and the comment itself just raises more questions IMO. Please remove that comment.

```suggestion
```

rousskov (commented):
Thank you for fixing this bug!

Since the caller of a function f() should not rely on any specific f() _implementation_ (beyond the known f() guarantees), every function call that uses an rfc1738_escape() as an _argument_ should be treated as a [UB](https://en.wikipedia.org/wiki/Undefined_behavior) or equivalent. Sooner or later, one of those calls will become buggy when f() (or the function f() calls) decides to call rfc1738_escape() itself... rfc1738_escape() should be rewritten to stop using an output buffer shared across calls. I do not insist on any related fixes in this PR though.

rousskov:
I polished the PR title/description to fix formatting and focus on the information that cannot be easily derived from the diff. Please adjust further as needed.

brozs:
Thank you for all the feedback. I have updated the commit based on your suggestions. Please let me know if I missed anything?

rousskov (approved):
Thank you for adjusting the fix! LGTM now.

yadij (approved):

PR #743, opened 2020-10-23
Avoid null pointer dereferences when dynamic_cast'ing to SwapDir

eduard-bagdasaryan:
Detected by Coverity. CID 1461158: Null pointer dereferences
(FORWARD_NULL).

When fixing these problems, we moved a few cache_dir iterations into
Disks.cc by applying the "Only Store::Disks can iterate cache_dirs"
design principle to the changed code. The Disks class is responsible for
maintaining (and will eventually encapsulate all the knowledge of) the
set of cache_dirs. Adjusted cache_cf.cc no longer depends on Disk.h.

### DIFF
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -62,9 +62,7 @@
 #include "SquidString.h"
 #include "ssl/ProxyCerts.h"
 #include "Store.h"
-#include "store/Disk.h"
 #include "store/Disks.h"
-#include "StoreFileSystem.h"
 #include "tools.h"
 #include "util.h"
 #include "wordlist.h"
@@ -642,19 +640,6 @@ configDoConfigure(void)
     /* init memory as early as possible */
     memConfigure();
     /* Sanity checks */
-
-    Config.cacheSwap.n_strands = 0; // no diskers by default
-    if (Config.cacheSwap.swapDirs == NULL) {
-        /* Memory-only cache probably in effect. */
-        /* turn off the cache rebuild delays... */
-        StoreController::store_dirs_rebuilding = 0;
-    } else if (InDaemonMode()) { // no diskers in non-daemon mode
-        for (int i = 0; i < Config.cacheSwap.n_configured; ++i) {
-            const RefCount<SwapDir> sd = Config.cacheSwap.swapDirs[i];
-            if (sd->needsDiskStrand())
-                sd->disker = Config.workers + (++Config.cacheSwap.n_strands);
-        }
-    }
 
     if (Debug::rotateNumber < 0) {
         Debug::rotateNumber = Config.Log.rotateNumber;
@@ -1853,17 +1838,8 @@ parse_http_header_replace(HeaderManglers **pm)
 static void
 dump_cachedir(StoreEntry * entry, const char *name, const Store::DiskConfig &swap)
 {
-    SwapDir *s;
-    int i;
-    assert (entry);
-
-    for (i = 0; i < swap.n_configured; ++i) {
-        s = dynamic_cast<SwapDir *>(swap.swapDirs[i].getRaw());
-        if (!s) continue;
-        storeAppendPrintf(entry, "%s %s %s", name, s->type(), s->path);
-        s->dump(*entry);
-        storeAppendPrintf(entry, "\n");
-    }
+    assert(entry);
+    Store::Disks::Dump(swap, *entry, name);
 }
 
 static int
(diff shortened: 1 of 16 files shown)


yadij (commented):
LGTM, just some polishing if you are okay with including it in-scope. I do not insist on these.

src/store/Disks.cc
```diff
@@ -46,12 +53,20 @@ objectSizeForDirSelection(const StoreEntry &entry)
     return minSize;
 }
 
+static SwapDir &
+SwapDirByIndex(const int i)
+{
+    const auto sd = INDEXSD(i);
```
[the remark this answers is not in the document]
rousskov: I do not think the INDEXSD hack should be completely removed in this PR because, even after the additional polishing in Disks.cc, some of the remaining hack uses (outside of Disks.cc) will need non-trivial API refactoring that is best done in a dedicated PR or two. We should keep `INDEXSD` as a marker for those places that still need significant work.

However, there is probably no good justification for keeping `INDEXSD` uses in Disks.cc (and where the call can be replaced with a StoreEntry::disk() call). Modernizing those trivial callers can be brought into in this PR scope without creating review problems AFAICT. @eduard-bagdasaryan, please do that. Just do not make any new disk-by-id APIs public. Keeping those APIs confined to Disks.cc will help control proliferation of callers that iterate cache_dirs outside of Disks.cc.

src/store/Disks.cc
```diff
@@ -165,9 +179,7 @@ Store::Disks::store(int const x) const
 SwapDir &
 Store::Disks::Dir(const int i)
 {
-    SwapDir *sd = INDEXSD(i);
-    assert(sd);
-    return *sd;
+    return SwapDirByIndex(i);
```
rousskov, line 182: Please add two asserts to make sure `i` is within the valid range. Most of the existing callers do not need that assert, but I see at least one special case where the cache_dir index is computed (rather than simply iterated) and should be double checked.

src/store/Disks.cc
```diff
@@ -363,17 +375,27 @@ Store::Disks::maxObjectSize() const
...
+        if (disk.needsDiskStrand()) {
+            assert(InDaemonMode());
+            disk.disker = Config.workers + (++Config.cacheSwap.n_strands);
+        }
+
         if (!disk.active())
             continue;
 
+        // update limits
+
```
rousskov, line 398: If the overall shape of this method remains intact, then please drop this comment: The code updating the limits starts earlier that this comment implies (before the loop). We probably cannot accurately mark configuration and limits sections without segregating/splitting the code using two loops.

src/store/Disks.cc
```diff
@@ -363,17 +375,27 @@ Store::Disks::maxObjectSize() const
...
     largestMaximumObjectSize = -1;
     secondLargestMaximumObjectSize = -1;
 
     for (int i = 0; i < Config.cacheSwap.n_configured; ++i) {
-        const auto &disk = Dir(i);
+        auto &disk = Dir(i);
+        if (disk.needsDiskStrand()) {
+            assert(InDaemonMode());
+            disk.disker = Config.workers + (++Config.cacheSwap.n_strands);
+        }
```
rousskov, line 392: :triangular_flag_on_post: I am concerned that I see no PR code that zeroes the always-incremented n_strands counter during _reconfiguration_. We used to reset that counter in configDoConfigure(). Please double check. If my concern is valid, please fix initialization, retest reconfigure, and add an XXX:

```suggestion
        // XXX: Do not pretend to support disk.disker changes during reconfiguration
        if (disk.needsDiskStrand()) {
            assert(InDaemonMode());
            disk.disker = Config.workers + (++Config.cacheSwap.n_strands);
        }
```

I do not think we can easily address that old XXX, so I do not recommend addressing it in _this_ PR.

src/store/Disks.cc
```diff
@@ -386,6 +408,70 @@ Store::Disks::updateLimits()
...
+
+    const auto fs = StoreFileSystem::FindByType(typeStr);
+    if (!fs) {
+        debugs(3, DBG_PARSE_NOTE(DBG_IMPORTANT), "ERROR: This proxy does not support the '" << typeStr << "' cache type. Ignoring.");
+        return;
+    }
+
+    const auto fsType = fs->type();
+
+    // check for the existing cache_dir
```
rousskov, line 430: I would document another old out-of-scope XXX:

```suggestion
    // check for the existing cache_dir
    // XXX: This code mistreats duplicated cache_dir entries (that should be fatal).
```

src/store/Disks.h
```diff
@@ -42,14 +42,19 @@ class Disks: public Controlled
     virtual void evictIfFound(const cache_key *) override;
     virtual int callback() override;
 
-    /// slowly calculate (and cache) hi/lo watermarks and similar limits
-    void updateLimits();
+    /// update configuration, including limits (re)calculation
+    void configure();
+    /// parses a single cache_dir configuration line
+    static void Parse(Store::DiskConfig &);
```
rousskov, line 48: If possible, please remove `Store::` inside `Store` in new/changed code. There are several occurrences.

rousskov (changes requested):
Thank you for polishing this code! You almost managed to remove more code than you added (`+187` and  `-180`), which is a rare accomplishment for a PR that does not remove any functionality ;-).

I see one red flag. The rest is minor. Please see inlined change requests for details.

eduard-bagdasaryan (commented):
Done (d915c33).

src/store/Disks.cc
rousskov: > is this about that after reconfiguration disk.disker may change but KidIdentifier remains the same, thus breaking methods where they are compared (e.g Store::Disk::active())?

Yes, although I was thinking about a somewhat more general problem: A reconfiguration might change needsDiskStrand() outcome -- there are no API guarantees that needsDiskStrand() result is sticky -- but a worker cannot start (or stop) using a disker in the middle of a transaction. I also doubt our code can handle a sudden n_strands increase. In other words, we probably do not support Disk::disker and cacheSwap.n_strands reconfiguration in general.

In the current code, Disk::disker is barely used, so specific problems may be minor or even non-existent, but, as we all know, code tends to change. This code dangerously mixes first-time configuration and reconfiguration code, sometimes on purpose and sometimes essentially accidentally. This mixture will continue to break code long-term, but fixing that problem is outside this PR scope. Your changes will just make that fix a bit easier (and will add more reminders about the need for such a fix).

src/store/Disks.cc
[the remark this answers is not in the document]
rousskov: > Alex: If there is no good reason to keep both SwapDirByIndex() and Dir(), please remove SwapDirByIndex().

> Eduard: SwapDirByIndex() is used in several global/static functions, such as storeDirWriteCleanLogs().

I think that is a "good reason" to keep both functions! Converting all those functions would create a lot of arguably out-of-scope work. I would not violently object to that work if others insist on it, but I recommend just adding a TODO to the SwapDirByIndex() declaration: "Remove when cache_dir-iterating functions are converted to Disks methods".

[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

rousskov (approved):
It is difficult (for me) to be sure about all the low-level code movements/adjustments, but I think this cleanup is a valuable step forward and should be accepted. Its scope is a bit fuzzy, but it is impossible to fully resolve the related known problems in a small PR, and we have no basis to insist on a full resolution of those old problems.

@yadij, please consider revisiting recent PR changes to make sure you still have no objections. Some of those changes should implement some of your earlier (optional) suggestions.

src/store/Disks.cc
```diff
@@ -386,6 +412,71 @@ Store::Disks::updateLimits()
...
+            return;
+        }
+    }
+
+    const int cacheDirCountLimit = 64; // StoreEntry::swap_dirn is a signed 7-bit integer
+    if (swap.n_configured >= cacheDirCountLimit)
+        throw TextException(ToSBuf("Squid cannot handle more than ", cacheDirCountLimit, " cache_dir directives"), Here());
+
+    // create a new cache_dir
+
```
yadij, line 461: unnecessary new empty line
```suggestion
```

src/cache_cf.cc
```diff
@@ -1853,17 +1838,8 @@ parse_http_header_replace(HeaderManglers **pm)
...
-
-    for (i = 0; i < swap.n_configured; ++i) {
-        s = dynamic_cast<SwapDir *>(swap.swapDirs[i].getRaw());
-        if (!s) continue;
-        storeAppendPrintf(entry, "%s %s %s", name, s->type(), s->path);
-        s->dump(*entry);
-        storeAppendPrintf(entry, "\n");
-    }
+    assert(entry);
+    Store::Disks::Dump(swap, *entry, name);
```
yadij, line 1842: The assert is unnecessary here.

Also since this is now just a one-liner calling a function from the legacy API dump*() name the whole static definition can be replaced by a #define dump_cachedir(X,Y,Z)  Store::Disks::Dump((Z), *(X), (Y)) in store/Disks.h alongside the free_cachedir() declaration.

src/cache_cf.cc
```diff
@@ -1981,84 +1957,11 @@ ParseAclWithAction(acl_access **access, const Acl::Answer &action, const char *d
...
-    swap->swapDirs[swap->n_configured] = StoreFileSystem::FileSystems().at(fs)->createSwapDir();
-
-    sd = dynamic_cast<SwapDir *>(swap->swapDirs[swap->n_configured].getRaw());
-
-    /* parse the FS parameters and options */
-    sd->parse(swap->n_configured, path_str);
-
-    ++swap->n_configured;
+    assert(swap);
+    Store::Disks::Parse(*swap);
```
yadij, line 1964: Same move to store/Disks.h as for dump_cachedir() can be made for this code.

yadij (approved):
I have a few small bits of polishing that might push this PR down into the realm of line removal. All optional, so marking approved now.

src/cache_cf.cc
rousskov: > the whole static definition can be replaced by a #define dump_cachedir(X,Y,Z) Store::Disks::Dump((Z), *(X), (Y)) in store/Disks.h alongside the free_cachedir() declaration.

A simple static definition inside a `.cc` file is much better than a `#define` in a public header. That free_cachedir() declaration was made public in c8f4eac for unit tests. If we are to clean this up further, we should model free_cachedir() after dump_cachedir() rather than half-mimic an already problematic free_cachedir() with a dump_cachedir() macro! However, I would not do that cleanup inside this PR.

src/cache_cf.cc
rousskov: > Same move to store/Disks.h as for dump_cachedir() can be made for this code.

IMO such a more would make things worse, not better, for the same "private static function is better than a public macro" reasons.

Long-term, configuration parser needs to be redesigned to remove the need for these wrappers (for class-based types), of course, but replacing localized static wrappers with public macros can only complicate that refactoring.

rousskov (commented):
Clearing for merge after partially addressing @yadij optional polishing requests.

PR #745, opened 2020-11-06
Source Format Enforcement

squidadm:


yadij:
OK to test

src/MemStore.cc
```diff
@@ -1011,7 +1011,7 @@ MemStoreRr::create()
 
     Ipc::Mem::PageStack::Config spaceConfig;
     spaceConfig.poolId = Ipc::Mem::PageStack::IdForMemStoreSpace(),
-    spaceConfig.pageSize = 0; // the pages are stored in Ipc::Mem::Pages
+                spaceConfig.pageSize = 0; // the pages are stored in Ipc::Mem::Pages
```
rousskov, line 1014: @yadij, please let me know if I should submit a PR to fix this typo first (to minimize the overall number of changes at the expense of redoing formatting in this PR). I thought I already did that, but looks like I forgot about this.

Otherwise, I will try to remember to fix this after this formatting PR is in.

yadij: We should be able to include it into this PR.

yadij: Done in 9021c3b

rousskov: Thank you. I am not a fan of including non-formatting changes in a formatting enforcement PR, but since this change should not have a runtime effect, I am not going to object.

rousskov (approved):
Approving under the assumption that these are all automated whitespace changes except for one manual `s/,/;/` typo fix.

yadij (approved):

PR #747, opened 2020-11-08
Squid-to-client write_timeout triggers client_lifetime timeout

rousskov:
Since commit 5ef5e5c, a socket write timeout triggers two things:
* reporting of a write error to the socket writer (as designed/expected)
* reporting of a socket read timeout to the socket reader (unexpected).

The exact outcome probably depends on the transaction state, but one
known manifestation of this bug is the following level-1 message in
cache.log, combined with an access.log record showing a
much-shorter-than-client_lifetime transaction response time.

    WARNING: Closing client connection due to lifetime timeout


yadij (approved):
I spy a bug in the semantics of the AlreadyTimedOut() function (its naming is inverse to what it returns, and missing documentation to clarify the closed case). But that should be fixed separately.

rousskov:
> I spy a bug in the semantics of the AlreadyTimedOut() function. But that should be fixed separately.

Yep.

PR #748, opened 2020-11-08
Transactions exceeding client_lifetime are logged as _ABORTED

rousskov:
... rather than timed out (_TIMEOUT).

To record the right cause of death, we have to call terminateAll()
rather than setting logType.err.timedout directly. Otherwise, when
ConnStateData::swanSong() calls terminateAll(0), it overwrites our
direct setting.


yadij (approved):
Thank you to whomever caught this one.

PR #752, opened 2020-11-17
Bug 5057: "Generated response lacks status code"

eduard-bagdasaryan:
... for responses carrying status-code with numerical value of 0.

Upon receiving a response with such a status-code (e.g., 0 or 000),
Squid reported a (misleading) level-1 BUG message and sent a 500
"Internal Error" response to the client.

This BUG message exposed a different/bigger problem: Squid parser
declared such a response valid, while other Squid code could easily
misinterpret its 0 status-code value as scNone which has very special
internal meaning.

A similar problem existed for received responses with status-codes that
HTTP WG considers semantically invalid (0xx, 6xx, and higher values).
Various range-based status-code checks could misinterpret such a
received status-code as being cachable, as indicating a control message,
or as having special for-internal-use values scInvalidHeader and
scHeaderTooLarge.

Unfortunately, HTTP/1 does not explicitly define how a response with a
status-code having an invalid response class (e.g., 000 or 600)
should be handled, but there may be an HTTP WG consensus that such
status-codes are semantically invalid:

https://lists.w3.org/Archives/Public/ietf-http-wg/2010AprJun/0354.html

Since leaking semantically invalid response status-codes into Squid code
is dangerous for response retries, routing, caching, etc. logic, we now
reject such responses at response parsing time.

Also fixed logging of the (last) received status-code (%<Hs) when we
cannot parse the response status-line or headers: We now store the
received status-code (if we can parse it) in peer_reply_status, even if
it is too short or has a wrong response class. Prior to this change,
%<Hs was either not logged at all or, during retries, recorded a stale
value from the previous successfully parsed response.

### DIFF
--- a/src/http/one/ResponseParser.cc
+++ b/src/http/one/ResponseParser.cc
@@ -12,6 +12,7 @@
 #include "http/ProtocolVersion.h"
 #include "parser/Tokenizer.h"
 #include "profiler/Profiler.h"
+#include "sbuf/Stream.h"
 #include "SquidConfig.h"
 
 const SBuf Http::One::ResponseParser::IcyMagic("ICY ");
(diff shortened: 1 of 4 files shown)


yadij (approved):
LGTM

src/http/one/ResponseParser.cc
```diff
@@ -100,6 +82,30 @@ Http::One::ResponseParser::parseResponseStatusAndReason(Tokenizer &tok, const Ch
...
+
+        // RFC 7230 Section 3.1.2 says status-code is exactly three DIGITs
+        if (code <= 99)
+            throw TextException(ToSBuf("status-code too short: ", code), Here());
+
+        // Codes with a non-standard first digit (a.k.a. response class) are
+        // considered semantically invalid per the following HTTP WG discussion:
+        // https://lists.w3.org/Archives/Public/ietf-http-wg/2010AprJun/0354.html
+        if (code >= 600)
+            throw TextException(ToSBuf("status-code from an invalid response class: ", code), Here());
```
rousskov, line 101: I hope this status-code semantics enforcement will not trigger a lot of compatibility problems. I think we _should_ be conservative here, at least as long as Squid itself is using zero/6xx status-codes internally and lacks a proper API for response class determination. I anticipate this to be one of the enforcement changes that will need a (configurable) bypass.

rousskov (approved):

src/http/one/ResponseParser.cc
yadij: This enforcement only being on received traffic, not on internal message values should be fine (and correct).

Old Squid used to leak those internal 6xx sometimes. IIRC it was fixed at some point early in 3.x. We should not have many (if any) installations still doing that.

Also some old roundcube versions used to transmit 6xx status in its XHR login traffic. I fixed that some while back too and there have been several major CVE issues triggering upgrades since. So hopefully not many of those left in existence either.

There was some software delivering " 0 " a year or two back. That may be an issue if it is still around. I cannot recall which it was though to check. If we want to be pedantic "000" is three digits (ie valid syntax per section 3.1.2) it is only invalid at semantic layer due to not having a defined meaning.

rousskov: > This enforcement only being on received traffic, not on internal message values should be fine (and correct).
 
My compatibility concern is about _received_ traffic. I mentioned internally used zero/6xx code because that use helps _justify_ the (conservative) changes in this PR. In other words, yes, we should be doing this, but we may have to provide workarounds in the future.

Bugs notwithstanding, internally-created response objects should not have invalid status codes -- invalid status codes should be used internally _outside_ of any response object. Thus, there should be no code that packs internal responses with invalid status codes and, hence, this parsing code should not see any internal responses with invalid status codes.
 
> There was some software delivering " 0 " a year or two back.

FWIW, July 2020 bug [report|https://bugs.squid-cache.org/show_bug.cgi?id=5057#c2] that triggered this PR was allegedly caused by software sending responses with zero status-codes.

PR #753, opened 2020-11-21
Reject different HTTP requests with unusual framing

yadij:
... and remove support for request_entities.

Squid now follows the following (approximate) rules when checking HTTP
request framing. The first matching rule wins.

* HTTP requests with a Transfer-Encoding:chunked header, including GET
  and HEAD requests with that header, are accepted. No changes here.

* HTTP requests with unsupported Transfer-Encoding values are rejected
  (Squid replies with HTTP 501 "Not Implemented"). No changes here.

* HTTP requests having conflicting Content-Length values are rejected
  (Squid replies with HTTP 400 "Bad Request"). No changes here.

* HTTP/1.0 and HTTP/0.9 POST and PUT requests without a valid
  Content-Length header are now rejected (Squid replies with HTTP 411
  "Length Required"). All of these were allowed before.

* HTTP/1.0 GET and HEAD requests with a Content-Length:0 header are now
  rejected (Squid replies with HTTP 400 "Bad Request"). All of these
  were allowed before.

* HTTP/1.0 GET and HEAD requests with a positive Content-Length header
  are now rejected (Squid replies with HTTP 400 "Bad Request"). All of
  these were allowed before if and only if the request_entities
  directive was explicitly set to "on".

There are no other framing-related HTTP request restrictions. Prior to
these changes, HTTP/1.1 GET and HEAD requests with a positive
Content-Length header were rejected unless the request_entities
directive was explicitly set to "on". The following configuration sketch
keeps rejecting those requests:

    acl getOrHead method GET HEAD
    acl withContentLength req_header Content-Length .
    http_access deny getOrHead withContentLength

The new restrictions were added due to possibility of cache corruption
attacks and other security issues related to HTTP request framing.

The request_entities directive was removed to simplify decision logic.

Some developers believe that these changes should be accompanied by
configuration options that allow admins to bypass (most of) the
previously absent restrictions. However, these developers do not know of
any important use cases that these changes break, and such cases may not
even exist. The authors insist on these security-driven changes.

### DIFF
--- a/src/HttpRequest.cc
+++ b/src/HttpRequest.cc
@@ -651,13 +651,13 @@ HttpRequest::canHandle1xx() const
     return true;
 }
 
-bool
+Http::StatusCode
 HttpRequest::canUseContentLength() const
 {
     // RFC 7230 Section 3.3.3 #4:
     // conflicting Content-Length(s) mean a message framing error
     if (header.conflictingContentLength())
-        return false;
+        return Http::scBadRequest;
 
     // HTTP/1.0 requirements differ from HTTP/1.1
     if (http_ver <= Http::ProtocolVersion(1,0)) {
@@ -668,7 +668,7 @@ HttpRequest::canUseContentLength() const
         //   A valid Content-Length is required on all HTTP/1.0 POST requests.
         // "
         if (m == Http::METHOD_POST)
-            return (content_length >= 0);
+            return (content_length >= 0 ? Http::scNone : Http::scLengthRequired);
 
         // RFC 1945 section 7.2:
         // "
@@ -677,7 +677,7 @@ HttpRequest::canUseContentLength() const
         // "
         // GET and HEAD do not define ('call for') an entity
         if (m == Http::METHOD_GET || m == Http::METHOD_HEAD)
-            return (content_length < 0);
+            return (content_length < 0 ? Http::scNone : Http::scBadRequest);
 
         // other methods are not defined in RFC 1945
     }
@@ -689,7 +689,7 @@ HttpRequest::canUseContentLength() const
     //   framing is independent of method semantics, even if the method does
     //   not define any use for a message body.
     // "
-    return true;
+    return Http::scNone;
 }
 
 bool
--- a/src/HttpRequest.cc
+++ b/src/HttpRequest.cc
@@ -667,7 +667,12 @@ HttpRequest::canUseContentLength() const
         // "
         //   A valid Content-Length is required on all HTTP/1.0 POST requests.
         // "
-        if (m == Http::METHOD_POST)
+        // RFC 1945 Appendix D.1.1:
+        // "
+        //   The fundamental difference between the POST and PUT requests is
+        //   reflected in the different meaning of the Request-URI.
+        // "
+        if (m == Http::METHOD_POST || m == Http::METHOD_PUT)
             return (content_length >= 0 ? Http::scNone : Http::scLengthRequired);
 
         // RFC 1945 section 7.2:
@@ -675,11 +680,16 @@ HttpRequest::canUseContentLength() const
         //   An entity body is included with a request message only when the
         //   request method calls for one.
         // "
-        // GET and HEAD do not define ('call for') an entity
+        // section 8.1-2: GET and HEAD do not define ('call for') an entity
         if (m == Http::METHOD_GET || m == Http::METHOD_HEAD)
+            return (content_length < 0 ? Http::scNone : Http::scBadRequest);
+        // appendix D1.1.2-4: DELETE, LINK, UNLINK do not define ('call for') an entity
+        if (m == Http::METHOD_DELETE || m == Http::METHOD_LINK || m == Http::METHOD_UNLINK)
             return (content_length < 0 ? Http::scNone : Http::scBadRequest);
 
         // other methods are not defined in RFC 1945
+        // assume they support an (optional) entity
+        return Http::scNone;
     }
 
     // RFC 7230 section 3.3
--- a/src/HttpRequest.cc
+++ b/src/HttpRequest.cc
@@ -652,6 +652,47 @@ HttpRequest::canHandle1xx() const
 }
 
 bool
+HttpRequest::canUseContentLength() const
+{
+    // RFC 7230 Section 3.3.3 #4:
+    // conflicting Content-Length(s) mean a message framing error
+    if (header.conflictingContentLength())
+        return false;
+
+    // HTTP/1.0 requirements differ from HTTP/1.1
+    if (http_ver <= Http::ProtocolVersion(1,0)) {
+        const auto m = method.id();
+
+        // RFC 1945 section 8.3:
+        // "
+        //   A valid Content-Length is required on all HTTP/1.0 POST requests.
+        // "
+        if (m == Http::METHOD_POST)
+            return (content_length >= 0);
+
+        // RFC 1945 section 7.2:
+        // "
+        //   An entity body is included with a request message only when the
+        //   request method calls for one.
+        // "
+        // GET and HEAD do not define ('call for') an entity
+        if (m == Http::METHOD_GET || m == Http::METHOD_HEAD)
+            return (content_length < 0);
+
+        // other methods are not defined in RFC 1945
+    }
+
+    // RFC 7230 section 3.3
+    // "
+    //   The presence of a message body in a request is signaled by a
+    //   Content-Length or Transfer-Encoding header field.  Request message
+    //   framing is independent of method semantics, even if the method does
+    //   not define any use for a message body.
+    // "
+    return true;
+}
+
+bool
 HttpRequest::parseHeader(Http1::Parser &hp)
 {
     Http::ContentLengthInterpreter clen;
--- a/src/HttpRequest.cc
+++ b/src/HttpRequest.cc
@@ -654,6 +654,14 @@ HttpRequest::canHandle1xx() const
 Http::StatusCode
 HttpRequest::checkEntityFraming() const
 {
+    // RFC 7230 section 3.3.1:
+    // "
+    //  A server that receives a request message with a transfer coding it
+    //  does not understand SHOULD respond with 501 (Not Implemented).
+    // "
+    if (header.unsupportedTe())
+        return Http::scNotImplemented;
+
     // RFC 7230 section 3.3.3 #3 paragraph 3:
     // Transfer-Encoding overrides Content-Length
     if (header.chunked())
--- a/src/HttpRequest.cc
+++ b/src/HttpRequest.cc
@@ -652,8 +652,13 @@ HttpRequest::canHandle1xx() const
 }
 
 Http::StatusCode
-HttpRequest::canUseContentLength() const
+HttpRequest::checkEntityFraming() const
 {
+    // RFC 7230 section 3.3.3 #3 paragraph 3:
+    // Transfer-Encoding overrides Content-Length
+    if (header.chunked())
+        return Http::scNone;
+
     // RFC 7230 Section 3.3.3 #4:
     // conflicting Content-Length(s) mean a message framing error
     if (header.conflictingContentLength())
--- a/src/HttpRequest.h
+++ b/src/HttpRequest.h
@@ -248,8 +248,9 @@ public:
 
     virtual void configureContentLengthInterpreter(Http::ContentLengthInterpreter &) {}
 
-    /// checks whether the Content-Length header (or lack of) is acceptible
-    Http::StatusCode canUseContentLength() const;
+    /// Check whether the message framing headers are valid.
+    /// \returns Http::scNone or an HTTP error status
+    Http::StatusCode checkEntityFraming() const;
 
     /// Parses request header using Parser.
     /// Use it in contexts where the Parser object is available.
--- a/src/HttpRequest.h
+++ b/src/HttpRequest.h
@@ -249,7 +249,7 @@ public:
     virtual void configureContentLengthInterpreter(Http::ContentLengthInterpreter &) {}
 
     /// checks whether the Content-Length header (or lack of) is acceptible
-    bool canUseContentLength() const;
+    Http::StatusCode canUseContentLength() const;
 
     /// Parses request header using Parser.
     /// Use it in contexts where the Parser object is available.
--- a/src/HttpRequest.h
+++ b/src/HttpRequest.h
@@ -248,6 +248,9 @@ public:
 
     virtual void configureContentLengthInterpreter(Http::ContentLengthInterpreter &) {}
 
+    /// checks whether the Content-Length header (or lack of) is acceptible
+    bool canUseContentLength() const;
+
     /// Parses request header using Parser.
     /// Use it in contexts where the Parser object is available.
     bool parseHeader(Http1::Parser &hp);
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -181,6 +181,13 @@ DOC_START
 	This option is not yet supported by Squid-3.
 DOC_END
 
+# Options removed in 6.x
+NAME: request_entities
+TYPE: obsolete
+DOC_START
+	Remove this line. HTTP/1.1 defines per-message control over GET/HEAD message body using headers.
+DOC_END
+
 # Options removed in 5.x
 NAME: dns_v4_first
 TYPE: obsolete
@@ -6652,22 +6659,6 @@ DOC_START
 
 	WARNING: If turned on this may eventually cause some
 	varying objects not intended for caching to get cached.
-DOC_END
-
-NAME: request_entities
-TYPE: onoff
-LOC: Config.onoff.request_entities
-DEFAULT: off
-DOC_START
-	Squid defaults to deny GET and HEAD requests with request entities,
-	as the meaning of such requests are undefined in the HTTP standard
-	even if not explicitly forbidden.
-
-	Set this directive to on if you have clients which insists
-	on sending request entities in GET or HEAD requests. But be warned
-	that there is server software (both proxies and web servers) which
-	can fail to properly process this kind of request which may make you
-	vulnerable to cache pollution attacks if enabled.
 DOC_END
 
 NAME: request_header_access
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -185,7 +185,7 @@ DOC_END
 NAME: request_entities
 TYPE: obsolete
 DOC_START
-	Remove this line. HTTP/1.1 defines per-message control over GET/HEAD message body using headers.
+	Remove this line. Squid now accepts GET/HEAD message body by default.
 DOC_END
 
 # Options removed in 5.x
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -185,7 +185,8 @@ DOC_END
 NAME: request_entities
 TYPE: obsolete
 DOC_START
-	Remove this line. Squid now accepts GET/HEAD message body by default.
+	Remove this line. Squid now accepts HTTP/1.1 requests with bodies.
+	To simplify UI and code, Squid rejects certain HTTP/1.0 requests with bodies.
 DOC_END
 
 # Options removed in 5.x
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -698,29 +698,6 @@ clientSetKeepaliveFlag(ClientHttpRequest * http)
     request->flags.proxyKeepalive = request->persistent();
 }
 
-/// checks body length of non-chunked requests
-static bool
-clientIsContentLengthValid(const HttpRequestPointer &r)
-{
-    // No Content-Length means this request just has no body, but conflicting
-    // Content-Lengths mean a message framing error (RFC 7230 Section 3.3.3 #4).
-    if (r->header.conflictingContentLength())
-        return false;
-
-    // RFC 7230 section 3.3 - Content-Length defines body exists regardless of method
-    if (r->content_length >= 0)
-        return true;
-
-    // We do not want a request entity on HTTP/1.0 GET/HEAD requests
-    // without a Content-Length header.
-    const auto m = r->method.id();
-    if (m == Http::METHOD_GET || m == Http::METHOD_HEAD)
-        return !(r->http_ver <= Http::ProtocolVersion(1,0));
-
-    // For other types of requests we don't care
-    return true;
-}
-
 int
 clientIsRequestBodyTooLargeForPolicy(int64_t bodyLength)
 {
@@ -1720,7 +1697,7 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
     }
 
     const auto chunked = request->header.chunked();
-    if (!chunked && !clientIsContentLengthValid(request)) {
+    if (!chunked && !request->canUseContentLength()) {
         clientStreamNode *node = context->getClientReplyContext();
         clientReplyContext *repContext = dynamic_cast<clientReplyContext *>(node->data.getRaw());
         assert (repContext);
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -700,7 +700,7 @@ clientSetKeepaliveFlag(ClientHttpRequest * http)
 
 /// checks body length of non-chunked requests
 static bool
-clientIsContentLengthValid(HttpRequest * r)
+clientIsContentLengthValid(const HttpRequestPointer &r)
 {
     // No Content-Length means this request just has no body, but conflicting
     // Content-Lengths mean a message framing error (RFC 7230 Section 3.3.3 #4).
@@ -1720,7 +1720,7 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
     }
 
     const auto chunked = request->header.chunked();
-    if (!chunked && !clientIsContentLengthValid(request.getRaw())) {
+    if (!chunked && !clientIsContentLengthValid(request)) {
         clientStreamNode *node = context->getClientReplyContext();
         clientReplyContext *repContext = dynamic_cast<clientReplyContext *>(node->data.getRaw());
         assert (repContext);
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1696,20 +1696,17 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
         return;
     }
 
-    const auto chunked = request->header.chunked();
-    if (!chunked) {
-        const auto clResult = request->canUseContentLength();
-        if (clResult != Http::scNone) {
-            auto *node = context->getClientReplyContext();
-            auto *repContext = dynamic_cast<clientReplyContext *>(node->data.getRaw());
-            assert(repContext);
-            conn->quitAfterError(request.getRaw());
-            repContext->setReplyToError(ERR_INVALID_REQ, clResult, request->method, nullptr, conn, request.getRaw(), nullptr, nullptr);
-            assert(context->http->out.offset == 0);
-            context->pullData();
-            clientProcessRequestFinished(conn, request);
-            return;
-        }
+    const auto frameStatus = request->checkEntityFraming();
+    if (frameStatus != Http::scNone) {
+        auto *node = context->getClientReplyContext();
+        auto *repContext = dynamic_cast<clientReplyContext *>(node->data.getRaw());
+        assert(repContext);
+        conn->quitAfterError(request.getRaw());
+        repContext->setReplyToError(ERR_INVALID_REQ, frameStatus, request->method, nullptr, conn, request.getRaw(), nullptr, nullptr);
+        assert(context->http->out.offset == 0);
+        context->pullData();
+        clientProcessRequestFinished(conn, request);
+        return;
     }
 
     clientSetKeepaliveFlag(http);
@@ -1727,6 +1724,7 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
 #endif
 
     /* Do we expect a request-body? */
+    const auto chunked = request->header.chunked();
     expectBody = chunked || request->content_length > 0;
     if (!context->mayUseConnection() && expectBody) {
         request->body_pipe = conn->expectRequestBody(
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -708,7 +708,7 @@ clientIsContentLengthValid(const HttpRequestPointer &r)
         return false;
 
     // RFC 7230 section 3.3 - Content-Length defines body exists regardless of method
-    if (r->content_length > 0)
+    if (r->content_length >= 0)
         return true;
 
     // We do not want a request entity on HTTP/1.0 GET/HEAD requests
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1697,18 +1697,19 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
     }
 
     const auto chunked = request->header.chunked();
-    if (!chunked && !request->canUseContentLength()) {
-        clientStreamNode *node = context->getClientReplyContext();
-        clientReplyContext *repContext = dynamic_cast<clientReplyContext *>(node->data.getRaw());
-        assert (repContext);
-        conn->quitAfterError(request.getRaw());
-        repContext->setReplyToError(ERR_INVALID_REQ,
-                                    Http::scLengthRequired, request->method, NULL,
-                                    conn, request.getRaw(), nullptr, nullptr);
-        assert(context->http->out.offset == 0);
-        context->pullData();
-        clientProcessRequestFinished(conn, request);
-        return;
+    if (!chunked) {
+        const auto clResult = request->canUseContentLength();
+        if (clResult != Http::scNone) {
+            auto *node = context->getClientReplyContext();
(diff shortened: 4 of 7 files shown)


src/cf.data.pre
```diff
@@ -181,6 +181,13 @@ DOC_START
 	This option is not yet supported by Squid-3.
 DOC_END
 
+# Options removed in 6.x
+NAME: request_entities
+TYPE: obsolete
+DOC_START
+	Remove this line. HTTP/1.1 defines per-message control over GET/HEAD message body using headers.
```
rousskov, line 188: Line way too long for cf.data.pre.

rousskov, line 188: This documentation is meant for admins that use an explicit `request_entities` setting today. The current text does not really help the admin to understand what has changed, what the effects of that change are, and how they can tune what they want to tune. Instead of just repeating RFC rules (that have not changed), we should explain why the directive was removed, why the default behavior was changed, and (hint at) how the old behavior can be restored/configured.

src/client_side.cc
```diff
@@ -657,28 +656,26 @@ clientSetKeepaliveFlag(ClientHttpRequest * http)
...
-    switch (r->method.id()) {
-
-    case Http::METHOD_GET:
+        return false;
 
-    case Http::METHOD_HEAD:
-        /* We do not want to see a request entity on GET/HEAD requests */
-        return (r->content_length <= 0 || Config.onoff.request_entities);
+    // RFC 7230 section 3.3 - Content-Length defines body exists regardless of method
+    if (r->content_length > 0)
```
rousskov, line 668: If we should exclude zero Content-Length, we should explain why it should be excluded. If it should be excluded, perhaps the name of the function (and its description) should be changed.

src/client_side.cc
```diff
@@ -657,28 +656,26 @@ clientSetKeepaliveFlag(ClientHttpRequest * http)
...
 
-    default:
-        /* For other types of requests we don't care */
-        return 1;
-    }
+    // We do not want a request entity on HTTP/1.0 GET/HEAD requests
+    // without a Content-Length header.
+    const auto m = r->method.id();
+    if (m == Http::METHOD_GET || m == Http::METHOD_HEAD)
+        return !(r->http_ver <= Http::ProtocolVersion(1,0));
```
rousskov, line 675: Please reword the comment to explain "why" we added this special case instead of just retelling what the code does in English.

src/client_side.cc
```diff
@@ -657,28 +656,26 @@ clientSetKeepaliveFlag(ClientHttpRequest * http)
 }
 
 /// checks body length of non-chunked requests
-static int
+static bool
 clientIsContentLengthValid(HttpRequest * r)
```
rousskov, line 660: Please consider making this an HttpRequest method since you are rewriting most of it, it only needs HttpRequest object to make the decision, and it feels out of place in client_side.cc (we deal with HTTP requests in other parts of Squid as well).

src/client_side.cc
```diff
@@ -657,28 +656,26 @@ clientSetKeepaliveFlag(ClientHttpRequest * http)
...
-        return 1;
-    }
+    // We do not want a request entity on HTTP/1.0 GET/HEAD requests
+    // without a Content-Length header.
+    const auto m = r->method.id();
+    if (m == Http::METHOD_GET || m == Http::METHOD_HEAD)
+        return !(r->http_ver <= Http::ProtocolVersion(1,0));
 
-    /* NOT REACHED */
+    // For other types of requests we don't care
```
rousskov, line 677: It is not clear to me what "do not care" means in this context. We are probably returning `true` for a specific reason. I suspect that reason is RFC requirements/rules. The answer may depend on the function name/purpose which might change during this PR.

src/client_side.cc
```diff
@@ -657,28 +656,26 @@ clientSetKeepaliveFlag(ClientHttpRequest * http)
 }
 
 /// checks body length of non-chunked requests
```
rousskov, line 658: > PR description: Versions 0.9 and 1.0 without a Content-Length are forbidden a body due to cache
corruption possibilities.

The function documentation and usage implies that this function does not apply to chunked requests. The PR description can be reasonably interpreted as if the changes do apply to chunked requests. Please resolve this conflict one way or another.

rousskov (changes requested):
It is not clear _why_ we are removing this configuration option. The PR description and/or cf.data.pre changes should explain what the PR is trying to accomplish but currently do not.

The existing clientIsContentLengthValid() name/purpose/scope/placement are rather confusing, on many levels. It would be nice to fix that in this PR (possibly by moving most of the clientIsContentLengthValid() code into HttpRequest::expectingBody() and making other adjustments), but I do not insist on that fix. However, if the current name/purpose/scope are preserved, then the PR should match them. I left a few specific change requests related to that mismatch.

I think we should improve wording in several places, but I will wait for the primary changes to settle before attempting to do that.

src/cf.data.pre
yadij: we don't have a specific formal limit for cf.data.pre, and this line is only displayed in cache.log

yadij: Over many years the directive has had less and less effect on traffic. With the correct use of Content-Length in HTTP/1.1 it does nothing.

rousskov: > we don't have a specific formal limit for cf.data.pre

The above assertion is hardly a valid excuse for adding excessively long lines. Many bad ideas are not formally prohibited.

> this line is only displayed in cache.log

It is also "displayed" in cf.data.pre which developers have to view and edit, often mimicking nearby informal formatting style. The fewer "can't see the whole line of text" inconveniences and "oh, do not look here, it is not a good example to mimic" exceptions we create for developers, the better.

Does Squid code (that logs obsolete directive documentation to cache.log) logs multiline documentation blobs? If not, then _that_ problem could be a valid excuse for using excessively long documentation lines.

FTR, while there are excessively long obsolete directive documentation lines already (e.g., `maximum_single_addr_tries`), there are also multiline obsolete directive documentation blobs (e.g., `header_access`).

rousskov: > Over many years the directive has had less and less effect on traffic.

Do you mean that the default setting of this directive rejected fewer and fewer legitimate messages? How do we know that?

> With the correct use of Content-Length in HTTP/1.1 it does nothing.

What do you mean? AFAICT, the directive has request rejection effect regardless of whether Content-Length is used correctly (whatever "correct usage" means).

src/client_side.cc
yadij: I believe it means the caching issues that are mentioned do not exist for other methods. Likely because they are not cacheable.

src/client_side.cc
yadij: Good point. Will do that on next update.

src/client_side.cc
yadij: The reason is vague reference in the direction of "cache corruption" in messages without Content-Length header in the RFC document. I've not found any specific CVE issue, though I presume the issue relates to when the broken Keep-Alive header was used (not an issue in HTTP/1.1). So the reason is as stated "we do not want" for those old versions - which are quite rare now.

src/client_side.cc
yadij: I disagree that is a reasonable interpretation of a line which specifically talks about versions where chunked is not valid.

src/client_side.cc
yadij: Done in f2a7292

src/client_side.cc
rousskov: > I disagree that is a reasonable interpretation of a line which specifically talks about versions where chunked is not valid.

I think you are implying that HTTP/1.0 messages with chunked encoding are invalid. IIRC, Squid considers HTTP/1.0 messages with chunked encoding _valid_, and I think Squid (including related documentation) should continue to do so.

src/client_side.cc
rousskov: Your response does not address my concern. FWIW, I will document the same concern from another angle: Right now, this PR code rejects HTTP/1.0 GET requests without Content-Length/Transfer-Encoding headers. I am asking you to explain why they should be rejected. I do not know which "vague reference ... in the RFC document" you are talking about, but this drastic rejection deserves a much better explanation than vague references to vague references.

FTR, after covering cases with Content-Length and Transfer-Encoding headers (irrelevant to the specific code being discussed in this particular change request), RFC 7230 says: `If this is a request message and none of the above are true, then the message body length is zero (no message body is present).`

The RFC position is clear/natural/simple -- if there is no signal that request has a body, then the request does not have a body. This position is compatible with real traffic/implementations because "nobody" sends Content-Length:0 with requests that do not usually have a body (like GET or DELETE) and nobody wants to add a lot of custom/complex code to handle a rare (and unnecessary) case of a request that is successfully terminated by its transport connection closure.

We may want to treat some exceptional cases specially, of course, but we need to explain why. I speculate that something went wrong when this PR tried to detect and/or treat some exceptional cases, but the existing documentation/comments do not help me guess what went wrong.

yadij: Oh. I think I see what you mean. I have a complete re-write of the sequence with better documentation coming shortly that should resolve this mistake and meet your other requests at the same time.

rousskov: Sounds good. Please request my review when you are done.

src/cf.data.pre
yadij: > > Over many years the directive has had less and less effect on traffic.
> 
> Do you mean that the default setting of this directive rejected fewer and fewer legitimate messages? How do we know that?

I am estimating from the repeating cycles of people coming to the IETF with specifications to upgrade or replace GET with something that has a body.

Plus the traffic is migrating to HTTP/2 and later which use implicit chunking in the form of DATA frames instead of Content-Length. That seems to be translating into an increase of chunked encoding when they fallback to 1.1. This whole section of code is not run when chunked is used.


> 
> > With the correct use of Content-Length in HTTP/1.1 it does nothing.
> 
> What do you mean? AFAICT, the directive has request rejection effect regardless of whether Content-Length is used correctly (whatever "correct usage" means).

As currently coded in master you are right. However that is counter to a SHOULD in the 1.1 spec - Content-Length with valid value means a body does exist. Once we fix Squid to obey that SHOULD there is no need for admin to turn this on.

yadij: > > Over many years the directive has had less and less effect on traffic.
> 
> Do you mean that the default setting of this directive rejected fewer and fewer legitimate messages? How do we know that?

I am estimating from the repeating cycles of people coming to the IETF with specifications to upgrade or replace GET with something that has a body.

Plus the traffic is migrating to HTTP/2 and later which use implicit chunking in the form of DATA frames instead of Content-Length. That seems to be translating into an increase of chunked encoding when they fallback to 1.1. This whole section of code is not run when chunked is used.


> 
> > With the correct use of Content-Length in HTTP/1.1 it does nothing.
> 
> What do you mean? AFAICT, the directive has request rejection effect regardless of whether Content-Length is used correctly (whatever "correct usage" means).

As currently coded in master you are right. However that is counter to a SHOULD in the 1.1 spec - Content-Length with valid value means a body does exist. Once we fix Squid to obey that SHOULD there is no need for admin to turn this on.

yadij:
Full refactor done in fc31d38.  Please re-review.

Note: one thing I am unsure of is the implications of obeying HTTP/1.0 regarding POST. If rejecting POST without C-L header is an issue I am okay wrapping that new clause with !USE_HTTP_VIOLATIONS.

Aoso, if necessary we can retain the directive. It would only apply to increasingly rare HTTP/1.0 messages though, so I would rather drop it now and have a smaller UI.

src/client_side.cc
yadij: done in fc31d38

src/client_side.cc
yadij: description updated earlier

src/client_side.cc
yadij: comment removed in fc31d38

src/client_side.cc
yadij: done in fc31d38

rousskov (changes requested):
Thank you for moving this PR forward.

Please see the inlined change requests for new and remaining code-related concerns.

The PR title and description [still](https://github.com/squid-cache/squid/pull/753#pullrequestreview-535973152) need to migrate from a collection of protocol facts to what has been changed (title) and why (description). Due to significant anticipated code changes, we can do another review iteration before focusing on fixing that PR metadata, but doing so might increase the number of iterations.

src/cf.data.pre
yadij: I am failing to see what you want changed. You asked for "the why" to be added to the line. A brief "why" is already there- that second sentence.

Do we really need to prefix every obsoletion which happens due to RFC compliance updates with "Squid is HTTP/1.1 software and obeys that spec now" ? A longer description placed in release notes often says that, and a more verbose expansion of the "why".

src/HttpRequest.h
```diff
@@ -249,6 +249,9 @@ class HttpRequest: public Http::Message
 
     virtual void configureContentLengthInterpreter(Http::ContentLengthInterpreter &) {}
 
+    /// checks whether the Content-Length header (or lack of) is acceptible
+    bool canUseContentLength() const;
```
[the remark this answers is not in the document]
yadij: Except, this method is not about whether the body is permissible. It is about whether the state of Content-Length (either present or absent) will cause a framing error. It is only relevant to to traffic relying on that Content-Length state for framing.

There is a rather large and complex amount of logic in client_side.cc following the call to this method which goes deep into the payload contents, format, syntax etc which would be relevant to a generic "checkBody()". Refactoring all of that is not in scope here.

src/client_side.cc
```diff
@@ -1682,7 +1656,7 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
     }
 
     const auto chunked = request->header.chunked();
-    if (!chunked && !clientIsContentLengthValid(request.getRaw())) {
+    if (!chunked && !request->canUseContentLength()) {
         clientStreamNode *node = context->getClientReplyContext();
         clientReplyContext *repContext = dynamic_cast<clientReplyContext *>(node->data.getRaw());
         assert (repContext);
```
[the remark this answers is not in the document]
yadij: Seems reasonable. Done in e151b66b8

src/HttpRequest.h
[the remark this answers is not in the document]
rousskov: > Except, this method is not about whether the body is permissible

I think it is or should be, but if you can suggest a better name/description, please do. The current ones are not going to work for the reasons I have stated in this change request. Attacking suggested fixes alone does not help with making progress. If you do not like those suggestions, please propose a better fix! I certainly do not insist on any specific name or wording.

> It is about whether the state of Content-Length (either present or absent) will cause a framing error.

Agreed. The name does not match this scope well (as I already noted). Also, the method currently does not cover all Content-Length cases, but after we move the `chunked` case inside the method, it will. At that point, the current method name will become even less appropriate, of course.

> There is a rather large and complex amount of logic in client_side.cc following the call to this method which goes deep into the payload contents, format, syntax etc  ... Refactoring all of that is not in scope here.

No such refactoring has been requested. This change request is about the being-adjusted method name, parameter, and description. The matching caller changes are very small, and the PR already changes the caller and the method, so we are not talking about some serious new refactoring here. If you prefer, I can make these changes myself.

src/cf.data.pre
rousskov: > I am failing to see what you want changed.

I want to change the documentation for the removed request_entities directive. The current text does not really help the admin to discover the effects of that change on a Squid instance that used the now-removed option, and to tune Squid configuration to restore or approximate the old behavior (if such tuning is still possible at all). The final text should do that (directly or, if really necessary, by reference to the detailed documentation elsewhere).

> A brief "why" is already there- that second sentence.

I cannot interpret the "HTTP/1.1 defines per-message control over GET/HEAD message body using headers" sentence as a reasonable answer to the "Why the request_entities directive I was using was removed?" question -- the cited specs have not changed, and the directive was actually documented to work around uncertainties in those specs.

And "why?" is not the only question we should answer. In fact, in _this_ context, that "why" question is secondary! The primary questions here are:

* what will change in Squid behavior after the upgrade (assuming the non-default setting was used by the admin prior to the upgrade) and
* how can the admin restore the old behavior (if they can do that at all). 

The "why" questions are primary for the PR description...

This is really basic "document the effect of the change on admins and how to undo it" stuff. We should not be arguing about this at all!

> Do we really need to prefix every obsoletion which happens due to RFC compliance updates with "Squid is HTTP/1.1 software and obeys that spec now"?

That text would be slightly better than the current PR text, IMO, but, in most cases, including this one, we should be more specific. The basic questions are the same for most removals. The answers should be specific to each removal. Most reasonable admins would not know what "Squid is now compliant" mean in the context of this option removal. Heck, some HTTP authors are still arguing what compliance itself means in this area! Squid behavioral changes are specific. We should document them (and any undoing workarounds).

> A longer description placed in release notes often says that, and a more verbose expansion of the "why".

Agreed, but it also sometimes does not. This PR does not contain such a longer description; the effects of this PR has changed quite a bit already and remain very far from obvious; the workarounds, if any, are not obvious either; so I think my request to document the current effects/workarounds is very reasonable.

src/HttpRequest.cc
```diff
@@ -668,6 +668,47 @@ HttpRequest::canHandle1xx() const
     return true;
 }
 
+bool
+HttpRequest::canUseContentLength() const
+{
+    // RFC 7230 Section 3.3.3 #4:
+    // conflicting Content-Length(s) mean a message framing error
+    if (header.conflictingContentLength())
```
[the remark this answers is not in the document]
yadij: The Content-Length header is entirely irrelevant when chunking is used. There are a lot of things that are validated earlier and prevent this function being called - we do not need to list them as being irrelevant to this functions internal clauses.

src/HttpRequest.cc
```diff
@@ -668,6 +668,47 @@ HttpRequest::canHandle1xx() const
...
+        // "
+        //   An entity body is included with a request message only when the
+        //   request method calls for one.
+        // "
+        // GET and HEAD do not define ('call for') an entity
+        if (m == Http::METHOD_GET || m == Http::METHOD_HEAD)
+            return (content_length < 0);
+
+        // other methods are not defined in RFC 1945
+    }
```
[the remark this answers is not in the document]
yadij: 1.0 differing from 1.1 does not preclude non-1.0 (aka "other") methods from having conditions defined in their own spec(s). RFC1945 conditions are only relevant to methods defined there AND only when used in HTTP/1.0 messages.

The fall-through is intentional to make default method handling be per the specification which Squid advertises as its version. Currently that being RFC7230.

src/cf.data.pre
yadij: TL;DR: updated the PR description and added release-notes text to this PR.

> > I am failing to see what you want changed.
> 
> I want to change the documentation for the removed request_entities directive. The current text does not really help the admin to discover the effects of that change on a Squid instance that used the now-removed option, and to tune Squid configuration to restore or approximate the old behavior (if such tuning is still possible at all). The final text should do that (directly or, if really necessary, by reference to the detailed documentation elsewhere).
> 

FTR: I believe the spec uncertainty was addressed by the RFC723x editorial work which added the text in RFC 7230 section 3.3 referenced by canUseContentLength() as reason for allowing entity in HTTP/1.1 GET messages.

> > A brief "why" is already there- that second sentence.
> 
> I cannot interpret the "HTTP/1.1 defines per-message control over GET/HEAD message body using headers" sentence as a reasonable answer to the "Why the request_entities directive I was using was removed?" question -- the cited specs have not changed, and the directive was actually documented to work around uncertainties in those specs.

Er, "Because using this directive means you want the behaviour which happens by default now".

> 
> And "why?" is not the only question we should answer. In fact, in _this_ context, that "why" question is secondary! The primary questions here are:
> 
>     * what will change in Squid behavior after the upgrade (assuming the non-default setting was used by the admin prior to the upgrade) and
> 
>     * how can the admin restore the old behavior (if they can do that at all).
> 

The answer to these is many sentences, even paragraphs of descriptive text in the commit description. Not something for a log entry which may be printed whenever Squid reconfigures. A one-liner is enough bloat for the log.

> 
> The "why" questions are primary for the PR description...

Yet this is a request requiring it be placed in the cache.log output. More PR description text I am okay with.

> 
> This is really basic "document the effect of the change on admins and how to undo it" stuff. We should not be arguing about this at all!

So we are to follow GNU project code submission requirements now? That "and how to undo it" is straight from their playbook.

> 
> > Do we really need to prefix every obsoletion which happens due to RFC compliance updates with "Squid is HTTP/1.1 software and obeys that spec now"?
> 
> That text would be slightly better than the current PR text, IMO, but, in most cases, including this one, we should be more specific. The basic questions are the same for most removals. The answers should be specific to each removal. Most reasonable admins would not know what "Squid is now compliant" mean in the context of this option removal. Heck, some HTTP authors are still arguing what compliance itself means in this area! Squid behavioral changes are specific. We should document them (and any undoing workarounds).
> 
> > A longer description placed in release notes often says that, and a more verbose expansion of the "why".
> 
> Agreed, but it also sometimes does not. This PR does not contain such a longer description; the effects of this PR has changed quite a bit already and remain very far from obvious; the workarounds, if any, are not obvious either; so I think my request to document the current effects/workarounds is very reasonable.

That would be cause for a request to update those relevant texts. eg make this PR update the Squid-6 release notes with a text we are both happy with. Not a request to dump lots of info into a log "one-liner".

src/HttpRequest.cc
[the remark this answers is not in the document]
rousskov: > There are a lot of things that are validated earlier and prevent this function being called - we do not need to list them as being irrelevant to this functions internal clauses.

Agreed. We should bring chunking decision into this function to clarify its internal logic, avoiding coding mistakes and reader questions/confusion. The rest should stay out.

src/HttpRequest.cc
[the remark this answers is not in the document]
rousskov: > 1.0 differing from 1.1 does not preclude non-1.0 (aka "other") methods from having conditions defined in their own spec(s). RFC1945 conditions are only relevant to methods defined there AND only when used in HTTP/1.0 messages.

It is technically inaccurate to call, say, the PUT or DELETE method "non-1.0" and to restrict generic RFC 1945 conditions to GET and HEAD. RFC 1945 does explicitly acknowledge the existence of other methods and even details PUT, DELETE, UNLINK, and LINK.

> The fall-through is intentional

Being intentional does not make it right. For code clarity and safety sake, I recommend following the sketch suggested in the change request, but there is another way to address my concern -- you can make the dangerous fall-through explicit. This will require a different comment to explain why we want to treat some HTTP/1.0 messages as if they were HTTP/1.1 messages _after_ handling version-agnostic requirements _and_ saying that HTTP/1.0 requirements differ from HTTP/1.1 requirements. I do not recommend this painful path, but I cannot stop you from trying to convince me that the dangerous fall-through is justifiable.

> to make default method handling be per ... RFC7230

It is not clear why we treat HTTP/1.0 GET specially, but do not treat, say, HTTP/1.0 PUT specially. HTTP/1.0 specs that we honor do not apply to just GETs and HEADs. They apply to all HTTP/1.0 messages. I see no good reason to introduce this complex fall-through logic (and the need to justify it), but if you insist on introducing it, then you have to justify it.

src/cf.data.pre
rousskov: I am resolving this change request. The updated description still needs work, but it is better and further polishing is probably best addressed in a new change request.

> Er, "Because using this directive means you want the behaviour which happens by default now".

You are supposed to tell that to the admin, and the earlier text was not doing that at all. I am saying this here in hope to avoid similar surprising discussions in the future, not to restart this discussion.

> So we are to follow GNU project code submission requirements now? That "and how to undo it" is straight from their playbook.

Strange question. We should require what is right/good/reasonable/etc. I am not familiar with the "GNU project code submission requirements", but if some of our requirements match some of theirs, I see absolutely no problem with that per se. FWIW, the stuff I am requesting is a reasonable minimum IMO.

> Yet this is a request requiring it be placed in the cache.log output.

It is not; you were just misinterpreting it that way. This change request is about providing essential information to the admin. There are many ways to do that.

> More PR description text I am okay with.

A PR description (i.e. commit message) is not something admins can easily find. Anything we want admins to know should be in squid.conf.documented, logs, and/or release notes. Commit messages are for developers.

> That would be cause for a request to update those relevant texts. eg make this PR update the Squid-6 release notes with a text we are both happy with. Not a request to dump lots of info into a log "one-liner".

There is no requirement for the corresponding log message to be limited to one line. It may be possible to limit this particular documentation to one line while addressing my concerns, but if doing so well requires five lines, we should use five lines.

src/HttpRequest.cc
```diff
@@ -668,6 +668,47 @@ HttpRequest::canHandle1xx() const
...
+            return (content_length >= 0);
+
+        // RFC 1945 section 7.2:
+        // "
+        //   An entity body is included with a request message only when the
+        //   request method calls for one.
+        // "
+        // GET and HEAD do not define ('call for') an entity
+        if (m == Http::METHOD_GET || m == Http::METHOD_HEAD)
+            return (content_length < 0);
```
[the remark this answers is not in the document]
yadij: In HTTP Content-Length:0 means there *is* an entity present. It just has 0 bytes of length. This distinction of that vs no entity (content_length < 0) is relevant for requests where the payload is a file or nested message etc. as happens in some HTTP extension specifications. Since this code is implementing the specification "only no-entity is valid" the condition is correct.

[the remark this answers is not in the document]
yadij: Putting aside the interpretation of the method (function) for now. There are several things going on.

First is how HTTP has a semantic distinction between 0-length and non-existing entity. The exact un-optimized form of this condition is:
```
return header.has(Http::CONTENT_LENGTH) == false; // absence of Content-Length is the only compliant state
```
However, the validity checking has already converted the header into content_length local member. So as an optimization we check that against -1 (no-entity).

content_length==0 is not one of the valid semantic states even if the on-wire bytes look safe. For that level of tolerance HTTP/1.1 or later is needed.

src/HttpRequest.cc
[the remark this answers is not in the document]
yadij: Attractive proposition but there is a problem with bringing the chunking test(s) inside. Specifically that complex logic later in the caller depends on the chunking test value.

The cleanest solution that meets all your current change requests appears to be adding output parameter(s) to this method. Which seems to me worse than a method scoped as purely Content-Length checks.

src/HttpRequest.h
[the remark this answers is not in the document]
yadij: > > Except, this method is not about whether the body is permissible
> 
> I think it is or should be, 

See the other thread about moving chunking tests. The decision about body permissiveness is made by the caller.

src/HttpRequest.cc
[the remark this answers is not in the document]
rousskov: > there is a problem with bringing the chunking test(s) inside

There is no problem. Just add a boolean `chunked` parameter. Except for the single `if` statement calling this method (which will be simplified), the caller code remains intact.

src/HttpRequest.h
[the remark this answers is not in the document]
yadij: You don't consider a const method with side effects being a "smell" ?

src/HttpRequest.cc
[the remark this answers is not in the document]
yadij: You don't consider a const method with side effects being a "smell" ?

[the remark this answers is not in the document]
rousskov: Adding a boolean `chunked` parameter (and the corresponding `if` statement returning zero) creates no side effects.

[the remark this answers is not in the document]
rousskov:
```C++
if (const auto error = request->checkBody(chunked)) {
    ... ERR_INVALID_REQ-setting code goes here ...
}
```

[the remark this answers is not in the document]
rousskov: Or, if you prefer, call header.chunked() twice, once from the caller and once from checkBody(void). IIRC, we have already made chunked() very fast in master/v6.

Either way, both the caller and checkBody() will be improved. The latter will be significantly improved.

src/HttpRequest.cc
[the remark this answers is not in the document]
yadij: Thanks for pointing out PUT, DELETE etc. I did missed the appendix.

Done in 0c94c29

src/HttpRequest.h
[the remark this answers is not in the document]
yadij: I do not like the name checkBody as it implies that we are checking the content of the body.

checkEntityFraming() seems more appropriate when C-L and T-E are both being validated. If you are okay with that name I will make that change and move the chunking check inside.

src/cf.data.pre
yadij: Resolved in59a3bd1

yadij:
FTR; the inclusion of DELETE, LINK, UNLINK changes may require leaving request_entities to allow those methods to enable* payloads for those methods. I am leaving that for now to see if wider testing shows that need still existing, or if we can get away with becoming strict-compliant in these increasingly rare 1.0 protocol traffic cases.

src/HttpRequest.h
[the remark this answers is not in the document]
rousskov: I have no strong objections against checkEntityFraming(). The "entity" part sounds unnecessary complex and possibly technically wrong to me -- I would prefer the simpler "body" -- but I can accept "entity" to make progress. Please proceed with any variation of those names.

src/HttpRequest.cc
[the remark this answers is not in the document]
yadij: Done in 56a86ec94e5563987adcf7fcaee97a1a3bfc0c40

src/HttpRequest.h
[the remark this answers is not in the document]
yadij: Done in 56a86ec

rousskov (dismissed):
The PR title and description still need work, but I keep delaying that polishing until the code becomes stable. There is one aspect that can be relevant to this review though:

```
To keep the old rejection behaviour use access policy:
...
 http_access deny fetch entity
```

And what if I want to keep my Squid instance old "allow" behavior for HTTP/1.0 messages? AFAICT, that is no longer possible. That is, the PR removes a configuration directive without giving the admins a workable replacement/alternative. AFAICT, the only justification we currently give (in the PR description) is "undefined behaviour and historic security issues by HTTP/1.0 agents", but that justification does not really apply to admins that explicitly enabled the now-removed directive after checking that the behavior in their environments was well-defined and secure. AFAICT, we are removing that configuration directive just to simplify things a bit.

Since I am not sure that we will harm innocent admins by removing this rarely used directive, I will not block this PR if you are sure that the removal is worth the risks. Your call.

src/client_side.cc
```diff
@@ -1722,15 +1694,13 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
...
+    const auto frameStatus = request->checkEntityFraming();
+    if (frameStatus != Http::scNone) {
+        auto *node = context->getClientReplyContext();
+        auto *repContext = dynamic_cast<clientReplyContext *>(node->data.getRaw());
+        assert(repContext);
         conn->quitAfterError(request.getRaw());
-        repContext->setReplyToError(ERR_INVALID_REQ,
-                                    Http::scLengthRequired, request->method, NULL,
-                                    conn, request.getRaw(), nullptr, nullptr);
+        repContext->setReplyToError(ERR_INVALID_REQ, frameStatus, request->method, nullptr, conn, request.getRaw(), nullptr, nullptr);
```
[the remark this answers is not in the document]
yadij: Done in 84b2d5f.

src/HttpRequest.cc
```diff
@@ -651,6 +651,70 @@ HttpRequest::canHandle1xx() const
...
+        // "
+        //   An entity body is included with a request message only when the
+        //   request method calls for one.
+        // "
+        // section 8.1-2: GET and HEAD do not define ('call for') an entity
+        if (m == Http::METHOD_GET || m == Http::METHOD_HEAD)
+            return (content_length < 0 ? Http::scNone : Http::scBadRequest);
+        // appendix D1.1.2-4: DELETE, LINK, UNLINK do not define ('call for') an entity
+        if (m == Http::METHOD_DELETE || m == Http::METHOD_LINK || m == Http::METHOD_UNLINK)
+            return (content_length < 0 ? Http::scNone : Http::scBadRequest);
```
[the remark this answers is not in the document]
yadij: I kind of agree with you. But am not familiar with the specifics of the security issues that apparently have been identified with bodies in 1.0 request messages. So for now would rather be conservative and strict.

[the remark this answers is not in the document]
yadij: The ammunition we have is that such requests have been identified (by others, so we have no specifics) as leading to security attacks that are quite serious in proxy and caching environments. To protect against those we are forbidding the Content-Length:0 hacks and requiring clients to better comply with HTTP (at least 1.0 if not current).

rousskov (commented):
I have adjusted PR title/description for clarity sake and to document the lack of consensus regarding these one-way changes. Please check carefully and adjust as needed.

My earlier [concern](https://github.com/squid-cache/squid/pull/753#pullrequestreview-589985216) about the lack of knobs to control some aspects of new rejections remains, but since I am not sure that we will harm innocent admins by removing this rarely used directive, I will not block this PR. If you are sure that these changes are worth the risks, I will not stand in the way of merging this PR.

yadij:
Thank you for the PR updates. I am okay with the new description text. As such this can now be merged under the 10+ day policy.

PR #755, opened 2021-01-06
Fix foreground cache_dir indexing

rousskov:
Foreground cache indexing (squid -F) was unusable in many environments:

* Non-SMP: Squid started listening for incoming connections and then
  ignored those connections while building the index, clients timed out.

* SMP: Workers started serving incoming connections, but diskers ignored
  I/O requests from workers, and workers reported disk I/O timeouts,
  significantly slowing down transactions. With enough traffic, disker
  queues overflowed as well.

Also, during foreground indexing, non-SMP Squid process or SMP diskers
were blocked and, hence, did not react immediately to signal-based
commands such as shutdown and reconfiguration. These commands took
effect only after the cache index was built.

Now Squid indexes all cache_dirs first (while responding to signals) and
starts listening for incoming connections second.

A new RegisteredRunner::useFullyIndexedStore event now activates
subscribers waiting for a fully indexed Store. Besides the obvious use
case of opening listening ports, we found and fixed two other cases:

* StoreDigest: `-F` now delays Store digest computation. Making an
  essentially empty digest and then replacing it with a very different
  digest of an indexed cache is likely to harm peers. Also, computing
  digests is an expensive, blocking activity (similar to cache indexing)
  that should not overlap with a similarly expensive, blocking cache
  indexing activity in `-F` mode. Arguably, `-F` should block _both_
  activities, but that change deserves a dedicated consideration/PR.

* PeerPoolMgr: `-F` now delays peer polling. Opening and maintaining
  connections, especially TLS connections, may require faster reaction
  times than a foreground indexing Squid can provide. Also, seeing our
  connections, polled peers may incorrectly assume that our instance is
  accepting connections and may send cache manager requests (or worse).

In SMP mode, cache_dirs are indexed by diskers. Detecting
useFullyIndexedStore condition in kids requires a new IPC message. In
`-F` mode, workers do not receive access to cache_dirs until those are
indexed. To prevent cache_dir discovery timeouts in workers, we also
added a new IPC message to extend the wait.

The SMP part of the implementation relies on new IPC messages:

* mtRebuildFinished: Each disker now sends a new IPC message notifying
  everybody (via Coordinator) that it has finished indexing. This allows
  SMP workers to detect useFullyIndexedStore conditions.

* mtForegroundRebuild: With '-F', Coordinator does not announce a
  cache_dir X to workers until foreground indexing of X is complete. To
  prevent cache_dir search timeouts in workers, indexing diskers sent
  periodic "I am still working" messages (via Coordinator). Workers keep
  waiting for mtStrandReady, rescheduling their search timeouts.


rousskov:
I am posting this unpolished Draft PR to ask the following two related questions:

1. To fix foreground indexing, we have to add a few IPC messages. We have discovered that the existing HereIamMessage and new IPC messages have the same underlying structure -- they only differ in the message type field. The current PR code replaces all those message types with a single StrandMessage type. We could reduce this PR diff by introducing StrandMessage in a dedicated PR. It will look a bit strange because that refactoring is not really needed without this PR. Should we extract those changes into a dedicated PR?

2. Testing discovered that IPC messages destined to kid A1 can be successfully delivered to kid A2 when kid A1 crashes and is restarted as A2. When this happens, Squid may crash and/or misbehave, often producing confusing logging/artifacts. The current PR does not address and ignores this problem. Should we fix it in this PR (where the problem was discovered and arguably made worse by introducing more IPC messages) or do another PR dedicated to this fix?

IMO, ideally, both changes should be extracted, but I am looking for @squid-cache/commiters opinions. I do not want to spend cycles on creating three PRs if others are happy with one.

yadij:
I'm in favour of splitting. It would be somewhat easier to review in smaller PR chunks. Your call though.

rousskov:
Thank you for a prompt and clear answer. We will probably split in two or three parts. Until then, please ignore this draft PR.

eduard-bagdasaryan:
>  split in two or three parts

I separated the first part as [PR756](https://github.com/squid-cache/squid/pull/756).

rousskov:
**Status update**:

> > split in two or three parts
>
> I separated the first part as [PR756](https://github.com/squid-cache/squid/pull/756).

PR #756 has been merged.

The second prerequisite pull request -- PR #771 -- has been merged as well.

We are now polishing this draft PR based on the merged changes.

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -447,23 +444,57 @@ IpcIoFile::canWait() const
...
+/// \param file an IpcIoFile to be queued
+void
+IpcIoFile::StartWaiting(const Pointer &file)
+{
+    if (file) {
+        auto expireTime = current_time;
+        expireTime.tv_sec += Timeout;
+        WaitingForOpen.emplace(expireTime, file);
+    }
+
```
rousskov, line 467: Please split this function into two: StartWaitingFor(file) and StartWaiting(void). The former should assert file existence and (after queuing) call the latter. The latter should continue to assert a non-empty queue.

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -447,23 +444,57 @@ IpcIoFile::canWait() const
...
-            return;
-        }
+    debugs(47, 7, "disker" << response.strand.kidId <<  " " << response.strand.tag << " is ready");
+    if (const auto file = StopWaiting(response.strand))
+        file->openCompleted(&response);
+    else
+        debugs(47, 4, "premature or late disker response");
+}
+
+/// reschedules IpcIoFile open request timeout for the head of the waiting queue
```
rousskov, line 457: This function does not (necessarily) REschedule a timeout -- it adds a new timeout. It also does not place the new timeout onto the _head_ of the queue -- the timeout will actually go to the _tail_ of the queue. Finally, the head/tail detail is unimportant for the caller -- they just want to schedule a timeout.

```suggestion
/// (re)start monitoring IpcIoFile open request timeout
```

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -447,23 +444,57 @@ IpcIoFile::canWait() const
...
-        if (response.strand.tag == (*i)->dbName) {
-            (*i)->openCompleted(&response);
-            WaitingForOpen.erase(i);
-            return;
-        }
+    debugs(47, 7, "disker" << response.strand.kidId <<  " " << response.strand.tag << " is ready");
+    if (const auto file = StopWaiting(response.strand))
+        file->openCompleted(&response);
+    else
+        debugs(47, 4, "premature or late disker response");
```
rousskov, line 454: "Premature" kind of implies that we have sent a request and received a response earlier than expected. AFAICT, this is the case where this kid did not send a request yet.

```suggestion
        debugs(47, 4, "foreign or late disker response");
```

or

```suggestion
        debugs(47, 4, "unsolicited or late disker response");
```

However, if that situation is no longer possible because our caller will protect us from foreign/unsolicited messages, then we can simplify:

```suggestion
        debugs(47, 4, "late disker response");
```

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -447,23 +444,57 @@ IpcIoFile::canWait() const
...
+        WaitingForOpen.emplace(expireTime, file);
+    }
+
+    assert(!WaitingForOpen.empty());
+    eventDelete(&IpcIoFile::OpenTimeout, nullptr); // deletes all OpenTimeout handlers
+    const auto interval = tvSubDsec(current_time, WaitingForOpen.begin()->first);
+    eventAdd("IpcIoFile::OpenTimeout", &IpcIoFile::OpenTimeout, nullptr, interval, 0, false);
+}
+
+/// removes the IpcIoFile specified by the response from the waiting queue
```
rousskov, line 474:
```suggestion
/// returns the matching IpcIoFile (after removing it from the wait queue) or nil
```

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -447,23 +444,57 @@ IpcIoFile::canWait() const
...
+IpcIoFile::Pointer
+IpcIoFile::StopWaiting(const Ipc::StrandCoord &responseStrand)
+{
+    const auto it = std::find_if(WaitingForOpen.begin(), WaitingForOpen.end(),
+    [&responseStrand](const FileWait &pair) { return pair.second->dbName == responseStrand.tag; });
+    if (it != WaitingForOpen.end()) {
+        const auto file = it->second;
+        WaitingForOpen.erase(it);
```
rousskov, line 482: If possible, please delete the IpcIoFile::OpenTimeout event when the queue becomes empty.

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -447,23 +444,57 @@ IpcIoFile::canWait() const
...
+void
+IpcIoFile::HandleStrandBusyResponse(const Ipc::StrandMessage &response)
+{
+    assert(opt_foreground_rebuild);
+    debugs(47, 7, "disker" << response.strand.kidId << " foreground rebuild is still in progress");
+    const auto file = StopWaiting(response.strand);
+    if (file) {
+        // reschedule open timeout
+        StartWaiting(file);
+    } // else: ignore a presumably late message
```
rousskov, line 497:
```suggestion
    if (const auto file = StopWaiting(response.strand))        
        StartWaiting(file); // schedule open timeout
    // else: ignore a presumably late or foreign response
```

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -538,18 +569,13 @@ IpcIoFile::StatQueue(std::ostream &os)
...
-            i != WaitingForOpen.end(); ++i) {
-        if (*i == ipcIoFile) {
-            (*i)->openCompleted(NULL);
-            WaitingForOpen.erase(i);
-            break;
-        }
+    while (!WaitingForOpen.empty() && current_time >= WaitingForOpen.begin()->first) {
+        const auto file = WaitingForOpen.begin()->second;
+        file->openCompleted(nullptr);
+        WaitingForOpen.erase(WaitingForOpen.begin());
```
rousskov, line 575: I think we ought to reverse the order here because some future openCompleted() might schedule some more timeouts, affecting the value of begin(). That is unlikely but theoretically possible and there is no reason to expose ourselves to those dangers.

```suggestion
        WaitingForOpen.erase(WaitingForOpen.begin());
        file->openCompleted(nullptr);
```

src/DiskIO/IpcIo/IpcIoFile.h
```diff
@@ -148,8 +150,12 @@ class IpcIoFile: public DiskFile
 
     static const double Timeout; ///< timeout value in seconds
 
-    typedef std::list<Pointer> IpcIoFileList;
-    static IpcIoFileList WaitingForOpen; ///< pending open requests
+    typedef std::multimap<timeval, Pointer, std::less<timeval> > WaitingFiles;
```
rousskov, line 153: Please add an allocator in case this code gets reused for something other than Rock diskers (e.g., for lots of individual files).

src/DiskIO/IpcIo/IpcIoFile.h
```diff
@@ -148,8 +150,12 @@ class IpcIoFile: public DiskFile
...
+    typedef WaitingFiles::value_type FileWait;
+    /// being open diskers, ordered by their absolute deadlines
+    static WaitingFiles WaitingForOpen;
+    static void StartWaiting(const Pointer &);
+    static IpcIoFile::Pointer StopWaiting(const Ipc::StrandCoord &);
```
rousskov, line 158:
```suggestion
    static Pointer StopWaiting(const Ipc::StrandCoord &);
```

rousskov (commented):
@eduard-bagdasaryan, please help me polish this code further. This is a partial review.

src/DiskIO/IpcIo/IpcIoFile.cc
rousskov: Ah, got it, thank you. I think we can omit that implementation detail from this method description. Resolving.

src/DiskIO/IpcIo/IpcIoFile.h
rousskov: ... but to the wrong map: This PR does not modify RequestMap so we should not touch it without a good reason. Please add the allocator to the WaitingFiles instead. When you do, please keep in mind the "missing `const`" bug recently fixed in 2081cef for another map allocator.

Please use `using X = Y` style for new/changed type definitions.

src/fs/rock/RockRebuild.h
```diff
@@ -126,6 +126,9 @@ class Rebuild: public AsyncJob, private IndependentRunner
     const bool resuming;
 
     static void Steps(void *data);
+
+    /// informs Coordinator that we are still foreground-rebuilding
+    void extendWait();
```
rousskov, line 131: Please move this declaration, _above_ data members, between sameEntry() and progressDescription(). Please move this _private_ method description to the method definition.

src/fs/rock/RockSwapDir.cc
```diff
@@ -867,6 +891,17 @@ Rock::SwapDir::writeCompleted(int errflag, size_t, RefCount< ::WriteRequest> r)
         CollapsedForwarding::Broadcast(*sio.e);
 }
 
+void
+Rock::SwapDir::remoteIndexingCompleted()
+{
+    assert(IamWorkerProcess());
+
+    if(!theFile || theFile->error())
```
rousskov, line 899:
```suggestion
    if (!theFile || theFile->error())
```

src/fs/rock/RockSwapDir.cc
```diff
@@ -867,6 +891,17 @@ Rock::SwapDir::writeCompleted(int errflag, size_t, RefCount< ::WriteRequest> r)
...
+        return; // the Rock db file is not open/initialized
```
rousskov, line 900: Please double check our logic here. I am concerned that we are not calling storeRebuildComplete() on these errors, even though the indexing has stopped. Without that call, `squid -F` will probably get stuck, and even a "regular" Squid instance may experience problems while waiting (forever) for indexing to finish.

src/store/Disk.h
```diff
@@ -102,9 +102,12 @@ class Disk: public Controlled
     char *path;
     int index;          /* This entry's index into the swapDirs array */
     int disker; ///< disker kid id dedicated to this SwapDir or -1
+    bool diskerReady; ///< notified by the disker that it is ready
+    bool diskerIndexed; ///< notified by the disker that it is indexed
     RemovalPolicy *repl;
     int removals;
     int scanned;
+    bool indexed;
```
rousskov, line 110: Let's document this new field, especially since its name overlaps with another new field. Our description should make it clear why we need both of these fields (and what the difference is).

src/store/Disk.h
```diff
@@ -102,9 +102,12 @@ class Disk: public Controlled
     char *path;
     int index;          /* This entry's index into the swapDirs array */
     int disker; ///< disker kid id dedicated to this SwapDir or -1
+    bool diskerReady; ///< notified by the disker that it is ready
+    bool diskerIndexed; ///< notified by the disker that it is indexed
```
rousskov, line 106: If possible, please report the new Disk flags in Store::Disk::stat().

rousskov (commented):
Another partial review.

eduard-bagdasaryan (commented):
Done (c82adc3).

PR #756, opened 2021-01-12
Deduplicating IPC strand messages

eduard-bagdasaryan:
No functionality changes other than minor debugging improvements.

* replaced identical (except for the message kind value) HereIamMessage
  and StrandSearchResponse classes with StrandMessage
* reduced code duplication with a new StrandMessage::NotifyCoordinator()
* split TypedMsgHdr::type() into unchecked rawType() and checked type()
* renamed and documented several Ipc::MessageType enum values

The above code improvements will help with adding more IPC messages.

### DIFF
--- a/src/DiskIO/IpcIo/IpcIoFile.h
+++ b/src/DiskIO/IpcIo/IpcIoFile.h
@@ -87,7 +87,7 @@ public:
     virtual bool ioInProgress() const;
 
     /// handle open response from coordinator
-    static void HandleOpenResponse(const Ipc::StrandSearchResponse &response);
+    static void HandleOpenResponse(const Ipc::StrandMessage &);
 
     /// handle queue push notifications from worker or disker
     static void HandleNotification(const Ipc::TypedMsgHdr &msg);
@@ -99,7 +99,7 @@ public:
 
 protected:
     friend class IpcIoPendingRequest;
-    void openCompleted(const Ipc::StrandSearchResponse *const response);
+    void openCompleted(const Ipc::StrandMessage *);
     void readCompleted(ReadRequest *readRequest, IpcIoMsg *const response);
     void writeCompleted(WriteRequest *writeRequest, const IpcIoMsg *const response);
     bool canWait() const;
--- a/src/ipc/Messages.h
+++ b/src/ipc/Messages.h
@@ -18,22 +18,28 @@ namespace Ipc
 
 /// message class identifier
 typedef enum { mtNone = 0, ///< unspecified or unknown message kind; unused on the wire
+
                mtRegisterStrand, ///< notifies about our strand existence
                mtStrandRegistered, ///< acknowledges mtRegisterStrand acceptance
+
                mtFindStrand, ///< a worker requests a strand from Coordinator
-               mtStrandReady, ///< a mtFindStrand answer: the strand exists and should be usable
+               mtStrandReady, ///< an mtFindStrand answer: the strand exists and should be usable
+
                mtSharedListenRequest,
                mtSharedListenResponse,
+
                mtIpcIoNotification,
+
                mtCollapsedForwardingNotification,
+
                mtCacheMgrRequest,
-               mtCacheMgrResponse
+               mtCacheMgrResponse,
+
 #if SQUID_SNMP
-               ,
                mtSnmpRequest,
-               mtSnmpResponse
+               mtSnmpResponse,
 #endif
-               ,
+
                mtEnd ///< for message kind range checks; unused on the wire
              } MessageType;
 
--- a/src/ipc/Messages.h
+++ b/src/ipc/Messages.h
@@ -18,7 +18,8 @@ namespace Ipc
 
 /// message class identifier
 typedef enum { mtNone = 0, ///< unspecified or unknown message kind; unused on the wire
-               mtStrandRegistration, ///< strand registration with Coordinator (also used as an ACK)
+               mtRegisterStrand, ///< notifies about our strand existence
+               mtStrandRegistered, ///< acknowledges mtRegisterStrand acceptance
                mtFindStrand, ///< a worker requests a strand from Coordinator
                mtStrandReady, ///< a mtFindStrand answer: the strand exists and should be usable
                mtSharedListenRequest,
--- a/src/ipc/Messages.h
+++ b/src/ipc/Messages.h
@@ -17,16 +17,23 @@ namespace Ipc
 {
 
 /// message class identifier
-typedef enum { mtNone = 0, mtRegistration,
-               mtStrandSearchRequest, mtStrandSearchResponse,
-               mtSharedListenRequest, mtSharedListenResponse,
+typedef enum { enumBegin_ = 0,
+               mtRegistration, ///< strand registration with Coordinator (also used as an ACK)
+               mtFindStrand, ///< a worker requests a strand from Coordinator
+               mtStrandReady, ///< a mtFindStrand answer: the strand exists and should be usable
+               mtSharedListenRequest,
+               mtSharedListenResponse,
                mtIpcIoNotification,
                mtCollapsedForwardingNotification,
-               mtCacheMgrRequest, mtCacheMgrResponse
+               mtCacheMgrRequest,
+               mtCacheMgrResponse
 #if SQUID_SNMP
                ,
-               mtSnmpRequest, mtSnmpResponse
+               mtSnmpRequest,
+               mtSnmpResponse
 #endif
+               ,
+               enumEnd_
              } MessageType;
 
 } // namespace Ipc;
--- a/src/ipc/Messages.h
+++ b/src/ipc/Messages.h
@@ -17,8 +17,8 @@ namespace Ipc
 {
 
 /// message class identifier
-typedef enum { enumBegin_ = 0,
-               mtRegistration, ///< strand registration with Coordinator (also used as an ACK)
+typedef enum { mtNone = 0, ///< unspecified or unknown message kind; unused on the wire
+               mtStrandRegistration, ///< strand registration with Coordinator (also used as an ACK)
                mtFindStrand, ///< a worker requests a strand from Coordinator
                mtStrandReady, ///< a mtFindStrand answer: the strand exists and should be usable
                mtSharedListenRequest,
@@ -33,7 +33,7 @@ typedef enum { enumBegin_ = 0,
                mtSnmpResponse
 #endif
                ,
-               enumEnd_
+               mtEnd ///< for message kind range checks; unused on the wire
              } MessageType;
 
 } // namespace Ipc;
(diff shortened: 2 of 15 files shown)


src/ipc/Messages.h
```diff
@@ -17,16 +17,23 @@ namespace Ipc
 {
 
 /// message class identifier
-typedef enum { mtNone = 0, mtRegistration,
-               mtStrandSearchRequest, mtStrandSearchResponse,
-               mtSharedListenRequest, mtSharedListenResponse,
+typedef enum { enumBegin_ = 0,
+               mtRegistration, ///< strand registration with Coordinator (also used as an ACK)
```
rousskov, line 21: There are two WholeEnum API-related problems here:

1. Per WholeEnum API, `enumBegin_` must match the first "real" value but currently does not. To solve this problem, we cannot just say `mtRegistration = enumBegin_` because we (correctly) use the zero value as an invalid/impossible one. AFAICT, the correct solution requires restoring mtNone.

2. The enumBegin_/enumEnd_ names are only allowed in scoped [enumeration](https://en.cppreference.com/w/cpp/language/enum) -- same names from different _unscoped_ enums would clash. I am not sure we should switch MessageType to scoped enumeration, but if we do want to switch, we should do it in _this_ PR. It would be nice to come up with some rules for when scoped enumeration is preferred. Are there any existing general (i.e. not Squid-specific) rules for that?

If we do not switch to scoped enums:

```C++
typedef enum {
               mtNone = 0, ///< unspecified or unknown message kind; unused on the wire
               mtRegistration, ///< strand registration with Coordinator (also used as an ACK)
...
               mtEnd ///< for message kind range checks; unused on the wire
             } MessageType;
```

src/ipc/Port.cc
```diff
@@ -79,6 +79,7 @@ void Ipc::Port::noteRead(const CommIoCbParams& params)
            " [" << this << ']');
     if (params.flag == Comm::OK) {
         assert(params.buf == buf.raw());
+        debugs(54, 6, "received message type: " << buf.rawType());
```
rousskov, line 82: Nitpick: "received" is implied by "noteRead" so we can make this shorter/simpler.

```suggestion
        debugs(54, 6, "message type: " << buf.rawType());
```

src/ipc/StrandCoord.cc
```diff
@@ -37,20 +38,35 @@ void Ipc::StrandCoord::pack(TypedMsgHdr &hdrMsg) const
...
+    messageType(msgType),
     strand(aStrand)
 {
 }
 
-Ipc::HereIamMessage::HereIamMessage(const TypedMsgHdr &hdrMsg)
+Ipc::StrandMessage::StrandMessage(const TypedMsgHdr &hdrMsg):
+    messageType(hdrMsg.type())
 {
     hdrMsg.checkType(mtRegistration);
```
rousskov, line 50: Forgotten leftovers from the old code?

```suggestion
    hdrMsg.checkType(hdrMsg.type());
```

Please `git grep` to double check that there are no other stale mtRegistration values in the PR branch. In fact, I would recommend renaming mtRegistration to mtStrandRegistration to eliminate such dangers for good, including during backporting!

src/ipc/TypedMsgHdr.h
```diff
@@ -43,7 +44,10 @@ class TypedMsgHdr: public msghdr
     /* message type manipulation; these must be called before put/get*() */
     void setType(int aType); ///< sets message type; use MessageType enum
     void checkType(int aType) const; ///< throws if stored type is not aType
-    int type() const; ///< returns stored type or zero if none
+    MessageType type() const; ///< converts rawType() to MessageType
```
rousskov, line 47: Just a question: Other than the `messageType(hdrMsg.type())` initialization/creation use case shown in this PR diff, what are other current type() users? I wonder whenter we can completely eliminate this method...

src/ipc/TypedMsgHdr.h
```diff
@@ -43,7 +44,10 @@ class TypedMsgHdr: public msghdr
...
+    /// received or set message kind; may not be a MessageType value
+    /// \returns 0 if none message kind has been received or set
```
rousskov, line 49: Please polish:

```suggestion
    /// \returns 0 if no message kind has been received or set
```

rousskov (changes requested):
Thank you for extracting these changes from #755!

I think I found at least one bug. There are also two design questions that we should answer before finalizing and merging a polished PR: switching to scoped enums and eliminating Ipc::StrandMessage::type(). Please see inlined change requests for details.

Please also check that the updated PR description, including the new "no functionality changes except..." claim, is correct. BTW, if we are changing numerical values of on-the-wire message type IDs, we should adjust that claim to also mention that change.

src/ipc/Messages.h
eduard-bagdasaryan: Since currently MessageType is not used with WholeEnum I think we should not bother switching to the scoped enum. Done with mtNote/mtEnd.

> two WholeEnum API-related problems 

Note that some existing Squid enums have these problems (e.g, icp_opcode or Http::HdrType).

src/ipc/TypedMsgHdr.cc
```diff
@@ -81,11 +81,13 @@ void Ipc::TypedMsgHdr::sync()
...
 
-int
+Ipc::MessageType
 Ipc::TypedMsgHdr::type() const
 {
-    Must(msg_iovlen == 1);
-    return data.type_;
+    const auto uncheckedType = rawType();
+    Must(uncheckedType >= MessageType::enumBegin_);
+    Must(uncheckedType <= MessageType::enumEnd_);
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

src/ipc/TypedMsgHdr.h
eduard-bagdasaryan: There are no other users. We could eliminate this method (moving its checks into StrandMessage::StrandMessage(), but it would look a little awkward IMO.

src/ipc/TypedMsgHdr.h
eduard-bagdasaryan: Done.

src/ipc/Port.cc
eduard-bagdasaryan: Done.

eduard-bagdasaryan (commented):
Done at 28462d5, please re-review.

src/DiskIO/IpcIo/IpcIoFile.h
```diff
@@ -99,7 +99,7 @@ class IpcIoFile: public DiskFile
 
 protected:
     friend class IpcIoPendingRequest;
-    void openCompleted(const Ipc::StrandSearchResponse *const response);
+    void openCompleted(const Ipc::StrandMessage *);
```
yadij, line 102: Dropped a const from the member signature here.

```suggestion
    void openCompleted(const Ipc::StrandMessage * const);
```

yadij (changes requested):

src/DiskIO/IpcIo/IpcIoFile.h
eduard-bagdasaryan: I dropped this const because it is ignored by the compiler. @rousskov suggests removing 'const' in such cases. Here is also an [article](https://abseil.io/tips/109) favoring this approach.

rousskov: Just to add one more bit of supporting information to Eduard's "this polishing is the right thing to do" response: IIRC, when @kinkie enabled more useful compiler checks, these "extra" `const` in function _declarations_ started to produce false positives. If really needed, I can try to find the corresponding errors/discussions in the long history of #667.

@yadij, please reconsider your change request.

@eduard-bagdasaryan, great [ToW](https://abseil.io/tips/109) find! I will be adding its rules of thumb to our squid-notes.

rousskov: > I will be adding its rules of thumb to our squid-notes.

Actually, this rule of thumb was already documented as [#style-const](https://github.com/measurement-factory/squid-notes/blob/start/thumbs.md#style-const). I now added a supporting ToW reference.

src/ipc/TypedMsgHdr.h
yadij: Perhapse this should be implemented as a TypedMsgHdr conversion operator:
```
operator MessageType() { ... }
```
then the caller does this more elegant code instead:
```
 MessageType(hdrMsg)
```

I also see there are 9 calls to rawType(). It seems a bit odd to be having a bounds checked getter, but mostly ignoring it in favour of a raw getter.

eduard-bagdasaryan: > Perhapse this should be implemented as a TypedMsgHdr conversion operator

We could do this but there is only one type() user - the benefit would be small over the fact that the implicit conversion is more difficult to control (i.e., avoid an unintended conversion).

> I also see there are 9 calls to rawType().

Right, because TypedMsgHdr is a low-level structure dealing with wire data. We should not throw upon receiving some garbage but simply discard it. There are several rawType() calls (as you noted), but they are confined within that low-level data analyzing code.

yadij: On one hand this data is supposed to be coming from another kid of this Squid process, so any "garbage" is a sign of malfunction or bugs. Silently ignoring that is a bad idea.

On the other hand, this is I/O going via mechanisms outside of Squid. There is a possibility of third-party applications joining in on the communications. Which means trusting it unconditionally is a bad idea.

The practical middle ground is to do the bounds checking and debugs() log any unexpected incidents before otherwise ignoring them.

eduard-bagdasaryan: >  Silently ignoring that is a bad idea.

I did not mean that we silently ignore it - in this case, a level-1 message is reported. See, e.g., Ipc::Coordinator::receive().

> The practical middle ground is to do the bounds checking and debugs()  

I don't think we need to do extra bounds checking in, e.g., Ipc::Coordinator::receive() because the values are already checked by the switch operator itself.

rousskov: I am resolving this change request because the question I asked has been answered. Thank you.

My underlying concern remains, and, while thinking about it, I will account for the ideas in the followup discussion in this change request. If I find a good way to address that concern, I will either commit the changes or post a new change request.

rousskov (approved):
I have removed both TypedMsgHdr::type() and StrandMessage::messageKind. @yadij, I believe these changes match the related [discussion](https://github.com/squid-cache/squid/pull/756#discussion_r555902287) you and Eduard had, but you may want to review commit 7705906 if you want to double check.

I also made a couple of polishing changes. @eduard-bagdasaryan, please review my changes and retest the final result. Once you are happy, I will clear this PR for merging.

eduard-bagdasaryan:
I reviewed the latest changes and ran a couple of tests. They look OK.

rousskov:
@eduard-bagdasaryan, thank you!

@yadij, I plan to clear this PR for merging in about 16 hours. See my previous [comment](https://github.com/squid-cache/squid/pull/756#pullrequestreview-574464891) for details about the changes after your approval.

yadij:
LGTM, marking for merge now.

yadij:
@eduard, can you check the staging_checks config of Anubis please? it seems to be stuck.

PR #757, opened 2021-01-14
Profiling: CPU timing implemented for MAC non-x86

devnexen:


lib/profiler/get_tick.h
```diff
@@ -67,6 +67,19 @@ get_tick(void)
...
+#include <mach/mach_time.h>
+
+static __inline hrtime_t
+get_tick(void)
+{
+    static mach_timebase_info_data_t base;
+    uint64_t ticks;
+    (void) mach_timebase_info (&base);
+    ticks = mach_absolute_time ();
+    return (hrtime_t)((ticks * base.numer) / base.denom);
```
rousskov, line 80: Please avoid division by zero. It might happen if mach_timebase_info() fails, but I probably would not restrict the check to that specific case.

lib/profiler/get_tick.h
```diff
@@ -67,6 +67,19 @@ get_tick(void)
...
 
+#elif defined(__APPLE__)
+#include <mach/mach_time.h>
+
+static __inline hrtime_t
+get_tick(void)
+{
+    static mach_timebase_info_data_t base;
+    uint64_t ticks;
+    (void) mach_timebase_info (&base);
```
rousskov, line 78: Static `base` suggests that we want to initialize `base` once and reuse it many times (e.g., because initialization is expensive). Unconditional mach_timebase_info() call suggests the opposite initialize-everytime approach (e.g., because initialization is expected to be done at compile time). Please pick the best single strategy and run with it.

rousskov, line 78:
```
Prefer to use the equivalent clock_gettime_nsec_np(CLOCK_UPTIME_RAW) in nanoseconds.
```

Apple [recommends](https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time) using clock_gettime_nsec_np() instead. If that recommendation is wrong or does not apply, please add a comment to explain why we disregard it. Otherwise, please follow that recommendation.

lib/profiler/get_tick.h
rousskov, line 80: How is Apple defining `hrtime_t`? Knowing that definition may help us avoid or polish this cast.

For example, some Linux-related [pages](https://www.unix.com/man-page/hpux/3C/gethrtime/) appears to define `hrtime_t` as `int64_t`.

rousskov (commented):
Thank you for contributing this improvement! Just to clarify: Have you successfully tested this code?

I do not know much about Apple development environments, unfortunately, but I have a couple of questions and suggestions.

I think we should convert this code to portable C++ but that can come once the primary changes are settled.

lib/profiler/get_tick.h
devnexen: mach_timebase_info in fact never fails even tough it does return a state.

lib/profiler/get_tick.h
devnexen: Right that was because mach_timebase_info is the historical api but in the case of the M1 indeed makes less sense since it s brand new anyway.

lib/profiler/get_tick.h
rousskov: FYI: Apple [documentation](https://developer.apple.com/documentation/driverkit/3433733-mach_timebase_info) implies otherwise. Perhaps it never fails today but might start failing tomorrow?

rousskov: Ah, I see that you removed that function. Never mind then. I am resolving this thread.

lib/profiler/get_tick.h
```diff
@@ -67,6 +67,15 @@ get_tick(void)
     return regs;
 }
 
+#elif defined(__APPLE__)
+#include <time.h>
+
+static __inline hrtime_t
+get_tick(void)
+{
+    return clock_gettime_nsec_np (CLOCK_UPTIME_RAW);
```
rousskov, line 76: > mach_timebase_info is the historical api but in the case of the M1 indeed makes less sense since it's brand new anyway.

AFAICT, clock_gettime_nsec_np() is a relatively new API. Should we check for its availability (during ./configure) before using it here, to avoid breaking builds on Apple platforms without clock_gettime_nsec_np()?

Alternatively, perhaps we can use a C++11 API that is guaranteed by the standard to be available? Have you considered [high_resolution_clock](https://en.cppreference.com/w/cpp/chrono/high_resolution_clock) or its friends? FWIW, Squid is already using high_resolution_clock.

rousskov (commented):
Thank you for addressing my earlier specific concerns. I have only one new question now before we can apply (minor) C++ polishing.

lib/profiler/get_tick.h
devnexen: Yes I m aware of the chrono's api, but it is only for Mac M1 ARM64's case here, otherwise old Mac X86 cases fall into the first category.

rousskov: Sorry, I do not know how to reconcile your response with my questions or concerns.

Perhaps you are implying that older Apple environments, where clock_gettime_nsec_np() may not available, will be handled by earlier `#if` and `#elif` clauses? I do not see how negating all those earlier conditions would guarantee clock_gettime_nsec_np() availability on an Apple platform.

Perhaps more importantly, the PR changes are clearly not specific to the "Mac M1 ARM64's case". That may be the case you are interested in, and there is nothing wrong with that, of course, but the the proposed code does not reflect your interest. Needless to say, the more generic you can make your changes, the better -- I am not implying that you should add more `#if` conditions! I am only saying that you cannot use the missing conditions as if they were there.

I am not familiar with Apple conventions, and there may be some Apply-specific logic at play here, but I doubt it is a good idea to leave our reliance on any such conventions implicit.

devnexen: I added some comments and made sure it s the ARM case...

rousskov: Thank you for clarifying your position/code.

In Squid project experience, the approach of guessing foo() availability by checking for the environment Bar where foo() should be available often leads to build failures. Sooner or later, Bar starts missing foo(). In most cases, it is much better to check whether foo() is available than to check whether the build happens in the environment Bar where foo() should be available.

You may be able to convince others to accept these (allegedly) "Mac M1 line from 2020"-specific changes. I am not going to block them just because they violate the above rule of thumb.

If you want _my_ vote, please switch from "Mac M1 line from 2020" checks (at compile time) to "clock_gettime_nsec_np is available" checks (at ./configure) time. Those checks would not even be specific to Apple AFAICT... If you need help implementing those, please ask!

BTW, you have [said](https://github.com/squid-cache/squid/pull/757/files#r557698383) that you are familiar with the C++ chrono API (which requires no checks!). Is there a good reason to avoid that checkless API here? I apologize if you have already answered that question, and I just failed to extract/grok that answer.

devnexen: Ok I ve switched to configure time check.

lib/profiler/get_tick.h
```diff
@@ -67,6 +67,15 @@ get_tick(void)
     return regs;
 }
 
+#elif defined(HAVE_CLOCK_GETTIME_NSEC_NP)
+#include <time.h>
```
rousskov, line 71: To avoid arguing about the necessity for `HAVE_TIME_H` checks (which are probably optional but still used by some Squid code):

```suggestion
#include <ctime>
```

lib/profiler/get_tick.h
```diff
@@ -67,6 +67,15 @@ get_tick(void)
...
+
+static __inline hrtime_t
```
rousskov, line 73: Nitpick: C++ uses `inline`:

```suggestion
static inline hrtime_t
```

lib/profiler/get_tick.h
```diff
@@ -67,6 +67,15 @@ get_tick(void)
...
+get_tick(void)
```
rousskov, line 74: Nitpick: Squid/C++ style favors `()`

```suggestion
get_tick()
```

lib/profiler/get_tick.h
```diff
@@ -67,6 +67,15 @@ get_tick(void)
...
+{
+    return clock_gettime_nsec_np (CLOCK_UPTIME_RAW);
```
rousskov, line 76: Nitpick: Squid usually does not separate the function name from its arguments:

```suggestion
    return clock_gettime_nsec_np(CLOCK_UPTIME_RAW);
```

rousskov, line 76: If possible (I have not checked!), let's use a more standard constant:

```suggestion
    return clock_gettime_nsec_np (CLOCK_MONOTONIC_RAW);
```

This change request was inspired, in part, by a similar LVM [change](https://reviews.llvm.org/rG8bec8927134f116eb11ddea9db78b1a6241884c9).

rousskov (changes requested):
Excellent progress, thank you! I hope we are getting very close to the final version here. Most of the changes I am requesting now are fixing minor problems and polish style.

BTW, if you would like to add your name (and, optionally) email to `CONTRIBUTORS`, please feel free to do so in this PR. IIRC, that addition may happen automatically during major releases, but a manual addition is faster and, more importantly, allows you to explicitly control which name and email are used.

Several of the change requests in this review necessarily depend on each other. You may want to skim all of them before addressing individual ones.

I understand that making many changes to add such a seemingly simple functionality can be rather annoying. If you prefer, I can do this latest round of changes myself. It will just take some time, and I will need you to retest the final result in your environment. Needless to say, I appreciate you patiently fixing this code so far and would prefer if you continue to do this legwork.

lib/profiler/get_tick.h
devnexen: sure ... btw all of these style changes can be automated with clang format setting possibly ?

yadij:
OK to test

yadij (approved):
LGTM

acinclude/os-deps.m4
```diff
@@ -800,6 +800,8 @@ AC_DEFUN([SQUID_CHECK_FUNCTIONAL_CPU_PROFILER],[
 // okay
 #elif defined(_M_IX86) && defined(_MSC_VER) /* x86 platform on Microsoft C Compiler ONLY */
 // okay
+#elif defined(HAVE_CLOCK_GETTIME_NSEC_NP) && defined(CLOCK_MONOTONIC_RAW)
```
rousskov, line 803: In my tests, this check fails without `#include <ctime>` on line 795 (i.e. the first line of this `AC_LANG_SOURCE` snippet) because `CLOCK_MONOTONIC_RAW` is not defined without that header in my environment. I warned about the lack of that `#include` [earlier](https://github.com/squid-cache/squid/pull/757#discussion_r557811814), but my warning was probably lost in the noise. Please fix.

lib/profiler/get_tick.h
```diff
@@ -67,6 +67,15 @@ get_tick(void)
     return regs;
 }
 
+#elif defined(HAVE_CLOCK_GETTIME_NSEC_NP) && defined(CLOCK_MONOTONIC_RAW)
+#include <ctime>
```
rousskov, line 71: This `#include` should be at the beginning of this header file. Otherwise, `CLOCK_MONOTONIC_RAW` will probably be undefined (in some environments) in some Profiler.h contexts. Please move it right after the existing `#if USE_XPROF_STATS` check.

rousskov (changes requested):
Please make sure the final version is tested in your environment -- I cannot test on M1, and there are no automated tests that will catch our bugs in this area. I will clear this PR for merging once the last two problems are fixed and you report that the final version works well in your tests. Thank you!

acinclude/os-deps.m4
devnexen: ok did the changes and tested old autoconf settings a bit messed up in my side made me missed this. good point.

devnexen: next time, I ll probably just file a bug report, that will probably be a lot faster, I did not envision this like this.

rousskov: I feel your pain! FWIW, if you just filed a bug report, it would have probably stayed as such for a while because nobody would volunteer to convert it into a pull request. Most pull requests do not require so many rewrites (or those rewrites are done by seasoned Squid developers). This PR is somewhat exceptional because it deals with neglected low-level code on a platform most of us do not have access to.

rousskov (commented):
Thank you very much for getting this PR ready! I am clearing it for merging.

P.S. Long-term, somebody should investigate if we can drop all these special cases and just provide a single implementation based on the standard C++ chrono API.

rousskov (approved):

rousskov:
@yadij, I replaced your `libprofiler` prefix with a more general `Profiling`. I hope that is OK with you. If not, let's remove the prefix to avoid arguing about it. No prefix would be my preference because, as you probably know, I believe that numerous inconsistent prefixes are just noise.

lib/profiler/get_tick.h
rousskov: Yes, most can be. We already auto-format with AStyle, but not every detail like this is covered by the existing formatting rules (unfortunately) because, in part, AStyle is not powerful and careful enough. An attempt to upgrade to clang-based formatting has stalled at #603.

yadij:
This change is isolated to a binary outside the "squid" binary. It is best to label these with the binary name so they can be identified easily in the changesets.

rousskov:
> This change is isolated to a binary outside the "squid" binary

When enabled, is not `libprofiler` a part of the `squid` binary?

```
src/Common.am:LIBPROFILER = $(top_builddir)/lib/profiler/libprofiler.la
src/Common.am:COMPAT_LIB = $(top_builddir)/compat/libcompatsquid.la $(LIBPROFILER)
src/Makefile.am:squid_LDADD = \
	...
	$(COMPAT_LIB) \
```

PR #758, opened 2021-01-17
Support reliable zeroing of sensitive buffers

devnexen:
TODO: Use the new API for more sensible buffers, possibly adding a
wrapper class for sensitive content to automate cleanup.


lib/util.c
```diff
@@ -155,3 +155,21 @@ unsigned int RoundTo(const unsigned int num, const unsigned int what)
...
+/**
+ * to zero a buffer in a more secure manner meant for a handful of purposes.
+ * e.g. for password matters.
+ * The compiler can optimize away a memset call to gain performance here
+ * making sure it does not occur.
+ *
+ * memset_s is designed for such cases otherwise putting the
+ * address in a volatile pointer avoid gcc's likes doing optimizations.
+ */
+void ClearMem(void *dst, size_t len)
```
yadij, line 167: This is not how we do portability in Squid. We use the relevant system call unconditionally in the main code and provide an implementation in libcompat for OS where it is not available.  See the compat/memrchr files for an example.

yadij (changes requested):

compat/memset_s.cc
```diff
@@ -0,0 +1,39 @@
...
+	    return EINVAL;
+    if (dsz > SIZE_MAX)
+	    return E2BIG;
+    if (len > dsz) {
+	    len = dsz;
+	    ret = EOVERFLOW;
+    }
+    void *(*volatile memset_fn)(void *, int, size_t) = &memset;
+    (void)memset_fn(dst, c, len);
+    return ret;
```
rousskov, line 37: This implementation does not properly replace a [standard](https://en.cppreference.com/w/c/string/byte/memset) memset_s() AFAICT. The standard one calls a constraint handler function upon detecting violations, right? Whether we should implement that aspect depends on the (currently unknown to me) PR goals/intent, but we should definitely disclose such a serious implementation deficiency.

N.B. [C11](https://en.cppreference.com/w/c/error/abort_handler_s) says that "the default handler is implementation-defined: it may be abort_handler_s, ignore_handler_s, or some other implementation-defined handler".

compat/memset_s.h
```diff
@@ -0,0 +1,30 @@
...
+#if !HAVE_MEMSET_S
+
+typedef int errno_t;
+typedef size_t rsize_t;
+
+/**
+ * to zero a buffer in a more secure manner meant for a handful of purposes.
+ * e.g. for password matters.
+ * The compiler can optimize away a memset call to gain performance here
+ * making sure it does not occur.
```
rousskov, line 21: Re-documenting a standard function is a bad idea IMO -- our custom documentation will either pointlessly duplicate the official one or miss important caveats. For example, compilers can only remove _some_ memset() calls while the above custom documentation can be misinterpreted as implying that they can remove _any_ memset() call. This compat file is meant to replace a standard memset_s(). We should not document what a standard memset_s() is.

compat/memset_s.cc
```diff
@@ -0,0 +1,39 @@
...
+
+#if !HAVE_MEMSET_S
+
+#include "compat/memset_s.h"
+
+/**
+ * to zero a buffer in a more secure manner meant for a handful of purposes.
+ * e.g. for password matters.
+ * The compiler can optimize away a memset call to gain performance here
+ * making sure it does not occur.
```
rousskov, line 19: Please do not duplicate descriptions. Public APIs should be documented in the headers, and the PR already does that.

compat/memset_s.h
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * memset_s is designed for such cases otherwise putting the
+ * address in a volatile pointer avoid gcc's likes doing optimizations.
```
rousskov, line 24: This is an implementation-specific detail. Please move it to the code implementing it (not the whole function definition where it is currently duplicated, but the actual volatile trick).

configure.ac
```diff
@@ -3229,6 +3229,7 @@ AC_CHECK_FUNCS(\
 	memmove \
 	memrchr \
 	memset \
+	memset_s \
```
rousskov, line 3232: This availability detection (and/or its use cases) would not work "as is" AFAICT because [C11](https://en.cppreference.com/w/c/string/byte/memset) says `As with all bounds-checked functions, memset_s is only guaranteed to be available if __STDC_LIB_EXT1__ is defined by the implementation and if the user defines __STDC_WANT_LIB_EXT1__ to the integer constant 1 before including string.h.`

It feels like Squid should do the above if we want to give the standard memset_s() a chance to be properly detected/used.

rousskov (changes requested):
Please edit the PR description to explain what this PR is trying to accomplish. I am very worried that I am reviewing this PR incorrectly due to the lack of explicit PR goals/intent statements. (That description will automatically become the future official commit message when this PR is merged as documented in [MergeProcedure](https://wiki.squid-cache.org/MergeProcedure))

Some of my change requests may become inapplicable or may need adjustment once the PR goals are clarified. You may ignore my inlined change requests until we settle the PR intent issue.

compat/memset_s.cc
devnexen: here it does the bare minimum, not a real replacement since it is not meant to be generally widely used, protections are mainly to protect against possible bad usage in general but in squid they re properly done, the main point is the placement of memset address into a volatile address. System implementations implement these part in assembly for example but that might be a bit far stretched here :-).

compat/memset_s.h
devnexen: Yes you re right.

configure.ac
devnexen: was able to detect with clang but gcc 9 did not. fair enough

devnexen: After the changes tried on a system undetected (Linux) and another when it was (MacOS).

yadij: Does the following build command detect the functions existence on Linux?
```
 ./configure CXXFLAGS="-D__STDC_WANT_LIB_EXT1__=1"
```

... and this one use the custom compat version on MacOS?
```
 ./configure CXXFLAGS="-D__STDC_WANT_LIB_EXT1__=0"
```

If not the detection may need to use an AC_CHECK_DECLS line with string.h as a dependency include.

devnexen: So FreeBSD, MacOS detects it (rightfully) but not in my Linux box (not present indeed). even the previous mean to detect was ok too same results.

lib/smblib/find_password.c
```diff
@@ -55,7 +55,7 @@ int next_password(char *pw, int pwlen)
     if (pwinit == FALSE) {
 
         pwinit = TRUE;
-        memset(pw, 0, pwlen + 1);
+        memset_s(pw, pwlen + 1, 0, pwlen + 1);
```
yadij, line 58: I don't think we should be doing any changes in this library. It is deprecated and as you can see from the list of protocol names at the top of this file only used for protocols obsolete since mid-1990's which are transferring their password in clear-text around the network.

yadij (changes requested):

lib/smblib/find_password.c
devnexen: true

compat/xstring.h
```diff
@@ -10,6 +10,7 @@
 #define SQUID_COMPAT_XSTRING_H
 
 #if HAVE_STRING_H
+#define __STDC_WANT_LIB_EXT1__ 1
```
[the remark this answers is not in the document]
devnexen: fair enough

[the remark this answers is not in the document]
rousskov: Depending on what this PR tries to accomplish, we may not want to provide a memset_s() replacement when `__STDC_WANT_LIB_EXT1__` is _not_ defined. The longer you advance this PR without an agreement what it should accomplish, the higher is the risk of redoing or rejecting your work. The scope of this PR may seem obvious to you, but it is not obvious to me, and I cannot guide you without knowing what the end goal is.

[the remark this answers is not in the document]
devnexen: It s an additional measure to make sure password data is really cleared regardless of the compiler optimisation. It can be pretty sensitive as highlighted here

https://github.com/squid-cache/squid/blob/master/lib/smblib/smbencrypt.c#L171

that is why some oses have implemented their own features (OpenBSD/explicit_bzero, NetBSD/explicit_memset, ...) to address those concerns. I myself had already implemented such in various softwares successfully (php, libb2, ...).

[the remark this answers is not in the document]
rousskov: Are you implying that the goal of this PR is to make sure (or start making sure) that no-longer-used clear-text password data is wiped clean when it is no longer used?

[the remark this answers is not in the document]
devnexen: I ve updated both comment and PR commit message.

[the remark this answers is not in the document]
yadij: FYI, the PR commit message used by our automation will be the text from your initial submission comment. I have copied the latest commit message to that github comment. Please check that it is correct and edit there if needed.

configure.ac
```diff
@@ -2948,6 +2948,8 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
 dnl Check for special functions
 AC_FUNC_ALLOCA
 
+AC_CACHE_DECL([memset_s], [], [], ([#include <string.h>]])
+
```
yadij, line 2952: Please don't add this extra empty line.

compat/xstring.h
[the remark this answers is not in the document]
yadij: Unless it isn't clear. I am waiting on removal of this #define to approve merge for this PR.

yadij (changes requested):

devnexen:
ah my bad I forgot this one.

yadij:
@rousskov, please re-review. LGTM, so when you are happy it can go in.

yadij (approved):

src/auth/digest/eDirectory/edir_ldapext.cc
```diff
@@ -399,7 +399,7 @@ static int nmasldap_get_simple_pwd(
     }
 
     if (pwdBuf != NULL) {
-        memset(pwdBuf, 0, bufferLen);
+        memset_s(pwdBuf, bufferLen, 0, bufferLen);
```
rousskov, line 402: The combination of this usage and memset_s() replacement implementation contradicts the PR intent AFAICT: 
Calling memset_s() instead of memset() implies that the caller thinks that it is important that the buffer is cleared. Ignoring memset_s() errors implies that the caller relies on the [constraint handler](https://en.cppreference.com/w/c/error/set_constraint_handler_s) to handle clearing errors, but the replacement never calls any constraint handlers, relying on the caller to check for them instead! Moreover, Squid never sets the constraint handler, resulting in essentially undefined behavior when the standard memset_s() is used!

IIRC, this contradiction is closely related to my other, old concern:  Standard memset_s() calls the [constraint handler](https://en.cppreference.com/w/c/error/set_constraint_handler_s). A developer adding a memset_s() call to their code has the right to expect that handler to be called (instead of checking every memset_s() call result). By completely ignoring that critical part of the standard memset_s() API, this PR needlessly increases the uncertainty of the outcome on caller errors:

* When standard memset_s() is available, errors will
    * be always silently ignored in some environments,
    * always crash Squid (without Squid-specific logging/diagnostics) in some environments, and
    * possibly do something else in other environments.
* When standard memset_s() is unavailable, errors will
    * be always silently ignored in all environments.

A PR that focuses on improving Squid security ought to do a lot better than introduce such a variety of bad outcomes when that security is threatened!

There are several ways to fix this problem. I suggest the simplest fix I can think of: Instead of trying to replace a very-difficult-to-correctly-replace memset_s(), provide a high-level equivalent. We can call it Security::ZeroSensitiveMemory() or some such. The function should check for all the caller errors a memset_s() would check, use the volatile pointer trick to call memset(), and return `void`. The function should throw on errors or, if others prefer that, assert() or abort(). 

With the fix proposed above, there is no need to check for memset_s() support or to call memset_s() if it is supported!

rousskov (changes requested):

devnexen:
Is there any remaining concern as there is waiting for author label ? :)

rousskov:
> Is there any remaining concern as there is waiting for author label ? :)

This PR is waiting for my review now (thank you, @yadij, for adjusting the PR state). I hope to review within a day or two.

@devnexen, FYI: Whenever you think you have addressed all reviewer concerns, use GitHub "Reviewers" box in the top right corner of the page to request a new review (like Amos did). Without that explicit request, it is often too time consuming for a reviewer to find which of the dozens of open PRs are waiting for their review, especially since GitHub does not let contributors adjust PR labels.

src/security/ZeroSensitiveMemory.h
```diff
@@ -0,0 +1,49 @@
...
+#ifndef SQUID_SRC_SECURITY_ZEROSENSITIVEMEMORY_H
+#define SQUID_SRC_SECURITY_ZEROSENSITIVEMEMORY_H
+
+#if HAVE_LIMITS_H
+#include <limits.h>
+#endif
+
+#if HAVE_STRING_H
+#include <string.h>
+#endif
```
rousskov, line 18: If possible, please use std::memset() and `<cstring>` instead. Besides removing `#include` guards, using a standard C++ function will reduce the number of declaration variation/problems (e.g., I am not sure that C standard guarantees that C memset is not a macro, and there are [claims](https://stackoverflow.com/questions/18754036/macro-definition-for-memset-function-in-c-on-mac-os-x) that it is, in fact, a macro in some Mac environments).

src/security/ZeroSensitiveMemory.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <stdexcept>
+
+namespace Security {
+
+inline static void
+ZeroSensitiveMemory(void *dst, const size_t len)
+{
+    if (!dst)
```
rousskov, line 27: I suggest accepting and ignoring nil destinations with zero lengths. It can make some of the callers simpler, safer, and/or faster.

```suggestion
    if (!len)
        return; // tolerate zero-length nil dst

    if (!dst)
```

src/security/ZeroSensitiveMemory.h
```diff
@@ -0,0 +1,49 @@
...
+     * e.g. for password clearing matters.
+     * The compiler can optimize away a memset call to gain performance here
+     * making sure it does not occur.
+     *
+     * address in a volatile pointer avoid gcc's likes doing optimizations.
+     * thus it is not mean as memset replacement which would cause a performance
+     * drop.
+     */
+    void *(*volatile memset_fn)(void *, int, size_t) = &memset;
+    (void)memset_fn(dst, 0, len);
```
rousskov, line 43: I suggest renaming `memset_fn` to `setMemory`, `cleaner`, or some such, to avoid violating Squid camelCase style and to avoid calling a function _pointer_ a function.

```suggestion
    (void)setMemory(dst, 0, len);
```

src/security/ZeroSensitiveMemory.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <stdexcept>
+
+namespace Security {
+
+inline static void
+ZeroSensitiveMemory(void *dst, const size_t len)
+{
+    if (!dst)
+        throw std::runtime_error("Cannot clear a null buffer");
```
rousskov, line 28: Squid usually throws TextExceptions. They may give more information to the admin and improve logging. If possible, please throw a TextException instead.

```suggestion
        throw TextException("attempt to clear a non-empty buffer via a nil pointer", Here());
```

rousskov (changes requested):
Thank you for moving this PR forward! I think you are on the right track now. I left a few corrections. I can help polish the PR title, description, and new function documentation once the code is stable. For now, I am ignoring those meta problems.

Unfortunately, the PR is currently not really in a reviewable state -- something went wrong in the (IMO premature, at best) attempts to keep the PR branch in sync with the master branch. You can see a lot of unrelated changes in the "Files changed" tab. Please fix. If you need help with maintaining the PR branch, please do not hesitate to ask!

@yadij, FYI: This PR is now very different from the code you have reviewed AFAICT. While I do not recommend reviewing this PR now due to excessive noise, I recommend dismissing your earlier review as it is no longer applicable (and self-requesting future review if needed).

src/security/ZeroSensitiveMemory.h
```diff
@@ -0,0 +1,49 @@
...
+    if (len > SIZE_MAX)
+        throw std::runtime_error("Cannot clear a buffer of length exceeding SIZE_MAX");
```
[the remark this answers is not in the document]
devnexen: ok then I guess code user will be responsible enough ;-)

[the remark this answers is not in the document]
rousskov: Well, the caller may still make mistakes like feeding `-1` or a huge value to us. My point was that the function itself _cannot_ (as in "does not have the ability to") detect those mistakes (because the compiler will convert them into valid (from the function point of view) values). In many cases, such conversions will happen with a compiler warning that the developer will notice and address, but that is a different story.

Thank you for removing this non-working check. This change request is addressed.

src/security/ZeroSensitiveMemory.h
rousskov: I am marking this change request as addressed because the problem it was describing is gone now. I will submit a new change request about the new problem created in the same place.

src/security/ZeroSensitiveMemory.h
```diff
@@ -0,0 +1,42 @@
...
+namespace Security {
+
+inline void
+ZeroSensitiveMemory(void *dst, const size_t len)
+{
+    if (!len)
+        return;
+
+    if (!dst)
+        return;
```
rousskov, line 23: We must not ignore requests to clear nil buffers with positive sizes, especially in a function meant to deal with "security sensitive" data. Such requests are often an indication of a serious bug in the calling code. We must at least throw or even assert. The old throwing code here was on the right path, but you removed it (not sure why).

Preservation of the old throwing code was one of the reasons I suggested adding a specific "tolerate zero-length nil dst" comment to the new `if (!len)` check. You have added that check (thank you), but not the comment (not sure why). Without that comment and without throwing, there would be no reason to have two `if` statements.

This suggestion should restore the earlier throwing behavior (while keeping the length check):

```suggestion
    if (!dst)
        throw TextException("attempt to clear a non-empty buffer via a nil pointer", Here());
```

src/security/ZeroSensitiveMemory.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+#ifndef SQUID_SRC_SECURITY_ZEROSENSITIVEMEMORY_H
+#define SQUID_SRC_SECURITY_ZEROSENSITIVEMEMORY_H
+
+#include <cstring>
+
+namespace Security {
+
+inline void
```
rousskov, line 16: Is there a good reason to inline this function? I do not see one. AFAICT, the volatile pointer trick should prevent inlining of the internal memset() call, so the whole thing is not going to be fully inlined anyway. And once it is not fully inlined, the negative side effects of exposing implementation internals may outweigh the performance improvements from inlining the part before the memset call. In general, I think we want to declare inline only what is likely to be inlined and has tangible performance benefits from inlining; the default is not to inline.

Please discuss or move the function definition to the corresponding .cc file.

N.B. I have added the same change request [earlier](https://github.com/squid-cache/squid/pull/758#discussion_r575687260), but you seemed to ignore it (I am not sure why).

rousskov (changes requested):
Thank you for fixing the PR branch and moving the PR forward! I only found two important issues to address, one old and one new one.

src/security/ZeroSensitiveMemory.h
devnexen: Sorry I misread your previous change request.

src/security/ZeroSensitiveMemory.h
devnexen: I did not ignore, it got lost in the rest of the noise, sorry.

devnexen: Would it be possible to not put the implementation into .cc ? The security part is not a lib thus each potential part of squid would need to made be aware in their respective Makefile.am, complicating things unnecessarily (IMHO).

yadij: FWIW, this is why I prefer the libcompat approach of using memset_s and implementing the standard behaviour there.

If it is too hard to call the correct policy handler or whatever standard version does that was tricksy. Then using our x prefix and throwing should be safe enough and equivalent to a policy that aborts the processing.

rousskov: > FWIW, this is why I prefer the libcompat approach of using memset_s and implementing the standard behaviour in our replacement.

I think we have already established that it is difficult to correctly implement a standard memset_s replacement while correct use of standard memset_s API requires mainline code changes that are notably more complex than what this PR is currently attempting to do. No such implementation has been offered. I do not understand why all that difficulty and trouble would be _preferred_ to "this" -- a basic question of whether to inline a simple function.

> Then using our x prefix and throwing should be safe enough and equivalent to a policy that aborts the processing.

That is exactly what this PR currently implements, except for the `x` prefix in the name (for obvious reasons). I think it is a much better solution overall, at least for now.

rousskov: > each potential part of squid would need to made be aware in their respective Makefile.am (security/libsecurity.la for instance), complicating things unnecessarily (IMHO).

Yes, every binary that uses libsecurity.la would have to link with it. I agree that more linking is a "complication", of course. However,

1. Inlining alone would not avoid that extra linking -- TextException is defined in `base/` which will have to be linked in if we throw.

2. Adding internal libraries to Squid binaries is not really a "problem". It is inevitability! The only reason we are having this discussion is because helpers like edir_ldapext were relatively recently moved into `src/` (commit 48c5f8e). That move was motivated, in part, by helpers _suffering_ without access to basic Squid-specific functions. We should not start worrying now about adding Squid libraries to Squid helpers (where such addition is warranted). We should enjoy the availability of those libraries.

3. The current inlined solution is not zero-cost either because it unnecessary exposes users to its internal dependencies that some of those users would not have been exposed to otherwise.

4. In its current bare form, ZeroSensitiveMemory() is almost useless -- it provides no protection from the sensitive memory traveling around Squid, being copied, pooled, etc. To properly offer real, tangible protection, we will need to implement non-trivial classes that manage that sensitive memory (and use ZeroSensitiveMemory() internally). Thus, in the long term, some linking will be inevitable even if we restrict ourselves to memory zeroing purposes.

Said that, I suspect that `security/` is actually a wrong place for this new function. That directory/library/namespace cannot possibly contain _everything_ related to Squid security -- that would be too broad and ultimately negate encapsulation benefits. I suspect `security/` should continue to be focused on external communication security while internal security efforts like this new function can be in `base/` (for now).

Going forward, I see two reasonable options to try:

1. Move to `base/`. Do not inline. Link edir_ldapext with `base/`.
2. Move to `base/`. Inline. Assert instead of throwing. Kick the "linking" can down the road without creating a precedent.

@yadij, which option do you prefer?

yadij: You are discarding option 3? - move to compat/xmemset_s.h, use inline and assert.

Of the two you have pre-selected. Option #1 is the closest to what I like.

yadij: There is also option 4 - move to mem/ instead of base/

rousskov (changes requested):
I was not sure this PR was waiting for _me_, but I am trying to push it forward.

devnexen:
I had tried but failed with linkage, so many dependencies.

src/security/ZeroSensitiveMemory.h
rousskov: > I had tried but failed with linkage, so many dependencies.

Thank you. I regret that your attempt confirmed my fears, but its failure resolves this change request by providing a good excuse not to inline this function -- too much out of scope work that nobody is willing to do right now.

src/security/ZeroSensitiveMemory.h
```diff
@@ -0,0 +1,42 @@
...
+ZeroSensitiveMemory(void *dst, const size_t len)
```
rousskov, line 17: I suggest moving the new function to src/mem/Sensitive.h. That move cannot cause any new linking problems AFAICT because the function will continue to be inlined. I do not insist on this move if others prefer the current src/security/ZeroSensitiveMemory.h location.

src/security/ZeroSensitiveMemory.h
rousskov: Just updating this old change request based on the failed attempt to avoid inlining:

```suggestion
    assert(dst);
```

rousskov (changes requested):

rousskov:
I have added the new function documentation in commit e4db81f and adjusted PR title/description (i.e. the future commit message) to match the current PR code and acknowledge that we are far from done here. Please adjust further as needed.

I also [tested](https://godbolt.org/z/xqcqsW5KK) the new code and can confirm that, unlike a bare call to std::memset(), the call to the new function is _not_ optimized away in a basic use case.

@yadij, the ball is on your side AFAICT.

rousskov:
OK to test

src/mem/Sensitive.h
```diff
@@ -0,0 +1,33 @@
...
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID_SRC_MEM_SENSITIVE_H
+#define SQUID_SRC_MEM_SENSITIVE_H
+
+#include <cstring>
+
+namespace mem {
```
yadij, line 14: The namespace name should be CamelCase for Squid code.

```suggestion
namespace Mem {
```

src/auth/digest/eDirectory/edir_ldapext.cc
```diff
@@ -399,7 +400,7 @@ static int nmasldap_get_simple_pwd(
     }
 
     if (pwdBuf != NULL) {
-        memset(pwdBuf, 0, bufferLen);
+        mem::ZeroSensitiveMemory(pwdBuf, bufferLen);
```
yadij, line 403:
```suggestion
        Mem::ZeroSensitiveMemory(pwdBuf, bufferLen);
```

src/auth/digest/eDirectory/edir_ldapext.cc
```diff
@@ -482,7 +483,7 @@ static int nmasldap_get_password(
...
```
yadij, line 486:
```suggestion
        Mem::ZeroSensitiveMemory(pwdBuf, bufferLen);
```

yadij (changes requested):
Just one minor typo.

yadij (approved):

devnexen:
wow exactly one year later :-)

PR #759, opened 2021-01-20
Add tls_key_log to report TLS communication secrets

rousskov:
The new log is meant for triage using traffic inspection tools like
Wireshark. Also known as SSLKEYLOGFILE, it is a common feature across
TLS-capable applications. By default, the new feature is disabled and
has negligible performance impact.

We have rejected the idea of adding an access_log %code instead, as
detailed in squid-dev thread titled "RFC: tls_key_log: report TLS
pre-master secrets, other key material". Such %code can still be added.
http://lists.squid-cache.org/pipermail/squid-dev/2020-July/009605.html

----

Generalized StoreClient::fillChecklist() API into Acl::ChecklistFiller

Code that establishes TLS connections outsources logging of secrets to
Security::KeyLogger. Security::KeyLogger::shouldLog() must consult
tls_key_log ACLs before logging the secrets but it does not know much
about the code/context that delegated this logging to KeyLogger. That
code could fill the checklist in advance, but that is an inferior
solution because it

* prevents outsourcing a rather mundane responsibility to KeyLogger
* creates a potentially stale Checklist, not reflecting TLS progress
* wastes resources on filling a checklist that may not be needed

While adding ConnStateData::fillChecklist() and looking through
ConnStateData code for examples on how to fill a checklist, it became
clear that there is a lot of code duplication in this area, including
some subtle problems with various checklist.foo assignments possibly
missing in some places (it is difficult to know for sure because some
checklist methods may extract the same information from _multiple_
sources). For example,

* on_unsupported_protocol and ssl_bump ignored acl_uses_indirect_client
* ident_lookup_access and delay_pool_access lacked port info and
  other ConnStateData details beyond src/dst addresses

Now, most of that duplicated code is replaced with unified
fillChecklist() calls. This unification is a significant improvement
that provides additional justification for the fillChecklist() design.

Also renamed ACLFilledChecklist::conn() to setConn(). Renaming is not
necessary for the compiler but will help detect no longer necessary
conn() setting calls when this code is ported to other Squid versions.

### DIFF
--- a/src/SquidConfig.cc
+++ b/src/SquidConfig.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 1996-2020 The Squid Software Foundation and contributors
+ * Copyright (C) 1996-2021 The Squid Software Foundation and contributors
  *
  * Squid software is distributed under GPLv2+ license and includes
  * contributions from numerous individuals and organizations.
--- a/src/SquidConfig.cc
+++ b/src/SquidConfig.cc
@@ -9,6 +9,7 @@
 #include "squid.h"
 #include "SquidConfig.h"
 
+// TODO: Who is zeroing this global during startup? C++ does not do it for us!
 class SquidConfig Config;
 
 class SquidConfig2 Config2;
--- a/src/SquidConfig.cc
+++ b/src/SquidConfig.cc
@@ -9,7 +9,6 @@
 #include "squid.h"
 #include "SquidConfig.h"
 
-// TODO: Who is zeroing this global during startup? C++ does not do it for us!
 class SquidConfig Config;
 
 class SquidConfig2 Config2;
--- a/src/security/Session.h
+++ b/src/security/Session.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 1996-2020 The Squid Software Foundation and contributors
+ * Copyright (C) 1996-2021 The Squid Software Foundation and contributors
  *
  * Squid software is distributed under GPLv2+ license and includes
  * contributions from numerous individuals and organizations.
--- a/src/security/Session.h
+++ b/src/security/Session.h
@@ -51,7 +51,8 @@ typedef std::shared_ptr<SSL> SessionPointer;
 typedef std::unique_ptr<SSL_SESSION, HardFun<void, SSL_SESSION*, &SSL_SESSION_free>> SessionStatePointer;
 
 #elif USE_GNUTLS
-using Connection = struct gnutls_session_int;
+// to be finalized when it is actually needed/used
+struct Connection {};
 
 // to be finalized when it is actually needed/used
 struct Session {};
--- a/src/security/Session.h
+++ b/src/security/Session.h
@@ -12,6 +12,7 @@
 #include "base/HardFun.h"
 #include "comm/forward.h"
 #include "security/LockingPointer.h"
+#include "security/forward.h"
 
 #include <memory>
 
@@ -43,11 +44,15 @@ bool CreateServerSession(const Security::ContextPointer &, const Comm::Connectio
 #if USE_OPENSSL
 typedef SSL Connection;
 
+using Session = SSL_SESSION;
+
 typedef std::shared_ptr<SSL> SessionPointer;
 
 typedef std::unique_ptr<SSL_SESSION, HardFun<void, SSL_SESSION*, &SSL_SESSION_free>> SessionStatePointer;
 
 #elif USE_GNUTLS
+using Connection = struct gnutls_session_int;
+
 typedef std::shared_ptr<struct gnutls_session_int> SessionPointer;
 
 // wrapper function to get around gnutls_free being a typedef
--- a/src/security/Session.h
+++ b/src/security/Session.h
@@ -53,6 +53,9 @@ typedef std::unique_ptr<SSL_SESSION, HardFun<void, SSL_SESSION*, &SSL_SESSION_fr
 #elif USE_GNUTLS
 using Connection = struct gnutls_session_int;
 
+// to be finalized when it is actually needed/used
+struct Session {};
+
 typedef std::shared_ptr<struct gnutls_session_int> SessionPointer;
 
 // wrapper function to get around gnutls_free being a typedef
@@ -61,6 +64,8 @@ typedef std::unique_ptr<gnutls_datum_t, HardFun<void, void*, &Security::squid_gn
 
 #else
 typedef std::nullptr_t Connection;
+
+struct Session {};
 
 typedef std::shared_ptr<void> SessionPointer;
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,18 @@
+Changes in squid-5.0.6 (10 May 2021):
+
+	- Bug 5057: Generated response lacks status code
+	- TLS: Handling missing issuer certificates for TLSv1.3
+	- TLS: Detail certificate validation errors during TLS handshake
+	- TLS: Detail client closures of CONNECT tunnels during TLS handshake
+	- TLS: %ssl::<negotiated_version and %ssl::>negotiated_version for TLS/1.3
+	- HTTP: Allow 1xx and 204 responses with Transfer-Encoding headers
+	- Maintenance: Start following Inclusive Naming Initiative advice
+	- Maintenance: Sort source file lists in Makefiles
+	- Maintenance: Support plugin-style scripts for source format enforcement
+	- Cleanup: Deduplicating IPC strand messages
+	- ... and some compile and debugging fixes
+	- ... and all fixes from 4.15
+
 Changes in squid-5.0.5 (02 Feb 2021):
 
 	- HTTP: Do not send Connection:keep-alive/close in HTTP Upgrade requests
@@ -89,6 +104,22 @@ Changes in squid-5.0.1 (14 Jan 2020):
 	- ... and many documentation changes
 	- ... and much code cleanup and polishing
 	- ... and all fixes from 4.10
+
+Changes in squid-4.15 (10 May 2021):
+
+	- Bug 5112: Excessively loud chunked reply parsing error reporting
+	- Bug 5106: Broken cache manager URL parsing
+	- Bug 5104: Memory leak in RFC 2169 response parsing
+	- Bug 3556: "FD ... is not an open socket" for accept() problems
+	- Profiling: CPU timing implemented for MAC non-x86
+	- Fix HttpHeaderStats definition to include hoErrorDetail
+	- Fix Squid-to-client write_timeout triggers client_lifetime timeout
+	- Limit HeaderLookupTable_t::lookup() to BadHdr and specific IDs
+	- Handle more Range requests
+	- Handle more partial responses
+	- Stop processing a response if the Store entry is gone
+	- ... and some portability fixes
+	- ... and some documentation updates
 
 Changes in squid-4.14 (02 Feb 2021):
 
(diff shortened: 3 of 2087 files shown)


rousskov:
I am posting this draft PR to ask the following question: Should we isolate "generalization of StoreClient::fillChecklist() API into Acl::ChecklistFiller" into a separate prerequisite PR? The corresponding changes are discussed in the second half of the PR description (under the horizontal rule). The vast majority of the corresponding code changes is branch commit 8748ac4.

Normally, I would consider such cleanup changes acceptable as a part of the PR that forced/triggered them. In this particular case, I have doubts because reducing duplication of Checklist filling code probably fixed several bugs (see PR description for a list) and at least highlighted old filling dangers (e.g., loops) and inconsistencies (i.e. uncoordinated parts of the filling sequence setting or attempting to set the same Checklist fields). Those old problems are unrelated to the tls_key_log feature, and I did not try to fix most of them here.

@squid-cache/commiters, do you think Acl::ChecklistFiller-related changes should be isolated into a dedicated PR?

src/SquidConfig.cc
```diff
@@ -9,6 +9,7 @@
 #include "squid.h"
 #include "SquidConfig.h"
 
+// TODO: Who is zeroing this global during startup? C++ does not do it for us!
```
yadij, line 12: As I understand it there is an edge case of C++ ... because the variable is global static it is default-initialized as if it had a zero-initializer list. In that situation pointers are implicitly set to nullptr, which happens to prevent crashing during the first parse of Foo::TheConfig and such classes. Basic typed members do not get set until explicitly initialized by the first call to free_all() from configFreeMemory().

yadij:
The checklist filling change is okay here IMO. Although it will hopefully become somewhat unnecessary when the MasterXaction work gets over the current hurdles.

The refactoring of both the logging and config parsing systems should be separated. At least we need to discuss the future of those new designs so I know which way to push future refactoring and replacements to finally remove the outdated mechanisms.

src/SquidConfig.cc
rousskov: I figured it out. `Config` is a non-local variable with static [storage duration](https://en.cppreference.com/w/cpp/language/storage_duration) (and external linkage). Its zero-initialization is guaranteed during static initialization (i.e. the first initialization stage during startup) by the corresponding C++ [rules](https://en.cppreference.com/w/cpp/language/initialization#Non-local_variables). Its dynamic initialization (i.e. the second initialization stage during startup) is currently broken (the generated default SquidConfig constructor will do nothing; it will not zero pointers, etc.), but the first stage is sufficient to mask the lack of proper dynamic initialization for this specific SquidConfig instance.

Thus, our `Config` is zero-intialized as expected. This branch TODO is addressed, and I will remove it.

P.S. FWIW, when I wrote that TODO, I was confused by the "otherwise, nothing is done" statement in the [default initialization](https://en.cppreference.com/w/cpp/language/default_initialization) rules. That misleading statement applies to _dynamic_ initialization only. For [non-locals](https://en.cppreference.com/w/cpp/language/initialization#Non-local_variables) like `Config`, we still need to take into account _static initialization_ during (i.e. the first startup initialization stage), which saves the day.

rousskov:
@yadij, thank you for answering my draft PR [question](https://github.com/squid-cache/squid/pull/759#issuecomment-763719388)!

> [the checklist filling change] will hopefully become somewhat unnecessary when the MasterXaction work gets over the current hurdles.

Yeah. It will take a while for the current transaction state to become fully reachable via a single object that we can give all lower-level code access to (and that single object cannot be the _master_ transaction object, of course). When/if that happens, there will be no need for the filling API indeed. Meanwhile, that API actually helps us reach that ideal by concentrating state sharing into fewer well-known places, reducing minor filling discrepancies that block further unification, and highlighting dependencies among transaction state components.

> The refactoring of both the logging and config parsing systems should be separated. 

To clarify:
* There is some access log refactoring (triggered primarily by your [request](http://lists.squid-cache.org/pipermail/squid-dev/2020-July/009612.html) for a few bells and whistles in this initial implementation of the tls_key_log feature), driven by code duplication concerns. Nothing major.
* There is virtually no refactoring of the old config parsing systems in this PR. The PR adds _new_ APIs that coexist with the legacy ones. It is difficult to separate those new APIs in a useful way because this PR does not convert the old code to use those new APIs. They are introduced for the tls_key_log directive benefit. Other directives can (and should) use the new APIs, of course, but that is a separate issue. My rule of thumb: When a lot of legacy code cannot be easily converted, we should not be forcing (or even encouraging!) new API authors to apply their creations across the Squid code base.

Going forward:

1. I will not separate the checklist filling change per your remark.

2. If you insist, I will create a prerequisite PR with parsing API additions, but they will be essentially unused by that PR code. I am a strong opponent of PRs introducing unused changes, but I can make an exception here because there is an almost-ready draft PR using them already. Please let me know if you insist on that separation.

3. If you insist, I will try to separate access_log-specific changes into a prerequisite PR. They would look a bit pointless -- their primary goal is to avoid _duplicating_ tls_key_log code (that would not exist in that prerequisite PR, of course). Please let me know if you insist on that separation.

> At least we need to discuss the future of those new designs so I know which way to push future refactoring and replacements to finally remove the outdated mechanisms.

Yes, of course. I will propose the use cases appropriate for those new APIs once the dust settles.

rousskov:
Here is a real tls_key_log file excerpt showing metadata added to each record via `#` comments and also how the same connection may get more details in the second record as Squid dives deeper into TLS negotiation (e.g., `conn80`).

````
# 39 conn70
CLIENT_RANDOM bcad8e... -
# 40 conn80
CLIENT_RANDOM 394bd9... -
# 41 conn80
RSA Session-ID:8bef63... Master-Key:1af0f9...
CLIENT_RANDOM 394bd9... 1af0f9...
````

Long hex keys were truncated for presentation clarity. The actual file has complete keys, of course.

**Comment by rousskov** [comment]
I plan to clear this PR for merging in 10 days if there are no objections. I should be able to nail the remaining environment-specific build problems by then.

File: src/security/Session.h [hunk 2]
```diff
@@ -43,11 +44,15 @@ bool CreateServerSession(const Security::ContextPointer &, const Comm::Connectio
...
 typedef SSL Connection;
 
+using Session = SSL_SESSION;
+
 typedef std::shared_ptr<SSL> SessionPointer;
 
 typedef std::unique_ptr<SSL_SESSION, HardFun<void, SSL_SESSION*, &SSL_SESSION_free>> SessionStatePointer;
 
 #elif USE_GNUTLS
+using Connection = struct gnutls_session_int;
```
yadij, line 54: As you can see from the GnuTLS internal name this is a session handle.

src/security/Session.h
```diff
@@ -43,11 +44,15 @@ bool CreateServerSession(const Security::ContextPointer &, const Comm::Connectio
 #if USE_OPENSSL
 typedef SSL Connection;
 
+using Session = SSL_SESSION;
```
yadij, line 47: Adding a generic abstraction here needs to be matched with the equivalent for all other supported libraries. It looks rather odd to have USE_OPENSSL wrappers around code elsewhere that uses only the abstractions, which use of a library-specific re-typename like this does.

yadij (changes requested):
The ConfigParse and logging changes still need breaking out. Yes they are not as large as the other changes here. But they are important changes worth doing as standalone features.

rousskov (commented):
> The ConfigParse and logging changes still need breaking out. Yes they are not as large as the other changes here. But they are important changes worth doing as standalone features.

Sigh. I wish you would not wait to insist on this for three months since my [response](https://github.com/squid-cache/squid/pull/759#issuecomment-766969027) to your original suggestion (which could be reasonably interpreted in many ways), outlining a specific way forward (with rationale)! I am working on extracting those changes into a dedicated PR.

rousskov:
> I am working on extracting those changes into a dedicated PR.

I extracted the API changes into #811. I will re-sync this PR after those changes are merged.

rousskov:
> I am working on extracting those changes into a dedicated PR.

I extracted CustomLog changes into #824. I will re-sync this PR after those changes are merged.

src/security/Session.h
yadij: Hmm, okay for now.

src/security/Session.h
yadij: Yes the CommunicationSecrets members are a good example of such code. The .cc wrapping is okay as-is for now, but when the second library is added will become internal to the methods implemented. The .h lines shold not need or have wrapping.

Is there a reason to care about private members being declared but undefined when they are unused? AFAIK they should just be dropped by the compiler.

yadij:
@rousskov, AFAIK current status is just your reply on our thread about the #if wrappers (which is not a merge blocker). I'd like to do another pass at review before clearing this, is there any more code changes you want to make before I do that?

src/security/Session.h
rousskov: > when the second library is added will become internal to the methods implemented

Yes, provided the second library will benefit from having the same private method(s).

> Is there a reason to care about private members being declared but undefined when they are unused?

I think there is. Without ifdefs, the reader and various automation tools may search for the method definition in vain. It is _usually_ not a big deal, but adding those ifdefs is not a big deal either. Lack of ifdefs will make certain coding bugs more difficult to find during development. And if all other factors were equal, I think being consistent across declarations and definitions is better than the alternative, especially when that consistency costs virtually nothing.

rousskov:
I merged now-official changes that were previously extracted from this PR, resolved conflicts, and synced PR description. The code builds on my machine, and I will work on fixing Semaphore CI "specialization ... in different namespace" build error.

src/security/Session.h
yadij: IMO the reader / dev searching should find the ones inside ifdef and see what is going on. Automation we only need to care about if it has problems or complains. Otherwise it is a matter of style and clean looking code - which is where removing the ifdef from .h helps improve the look of code.

yadij (approved):
@rousskov, I do not think this will need another review after the #824 sync. Please mark for Anubis when you are done.

src/security/Session.h
rousskov: > IMO the reader / dev searching should find the ones inside ifdef and see what is going on.

I see two wrinkles in this logic:

1. Just seeing the ifdefed code in the .cc file would not clarify anything for some. For me, for example, it would only raise questions like "Why were these method declarations left without ifdef protections? Is there another definition of these methods that I have missed?"
2. Modern developers often use automation to quickly "find" things -- they prefer not to grep for names but ask their IDE-like environment where the method is defined. That automation is getting pretty smart these days -- using compiler-produced namespace-aware maps for lookup. Some automation is aware of C++ macros. Unusual patterns may break automation (as discussed elsewhere), and such breakage may affect what developers can easily "find" or "see".

> Automation we only need to care about if it has problems or complains

Many automation failures are silent. For example, we do not know about all the cases where the compiler failed to optimize something away. Thus, we should not misinterpret absence of known problems as absence of problems.

> Otherwise it is a matter of style and clean looking code - which is where removing the ifdef from .h helps improve the look of code.

I agree that there is a "matter of style" component here as well! Unfortunately, we disagree which style is overall better. With all the other factors being equal (and they are not), to me, consistency and lack of surprises/exceptions/explanations is much better than a private class declaration section without ifdefs.

> I do not insist on this change.

Thank you.

rousskov:
@kinkie, just FYI: The Jenkins test for the staged commit [failed](https://build.squid-cache.org/job/5-pr-test/3065/console) for a rather strange reason I have not seen before. That failure did not block merging, but that is probably another problem that I wrote elsewhere about.

PR #760, opened 2021-01-27
cpu_affinity_map support for FreeBSD proposal.

devnexen:
moving the part in separated compat implementation (also couple of other oses support cpu affinity, e.g. NetBSD, with different apis). Most of CPU_* macros are compatible minus CPU_AND.


rousskov:
Just FYI: You do not need to limit your Squid PR branches to one commit. Pushing updates is fine, and individual commits are often useful for reviewers. Our merge bot will squash all your changes anyway. GitHub may get confused by force pushes when managing review requests. That said, it is your call. Sometimes, force-pushing is unavoidable, of course.

compat/sched_affinity.h
```diff
@@ -0,0 +1,85 @@
...
+#if !HAVE_SCHED_SETAFFINITY
+
+#include <stdbool.h>
+#if HAVE_SYS_CPUSET_H
+#include <sys/param.h>
+#include <sys/cpuset.h>
+#undef HAVE_CPU_SET_T
+#define HAVE_CPU_SET_T 1
+#undef HAVE_CPU_AFFINITY
+#define HAVE_CPU_AFFINITY 1
```
yadij, line 21: Why are these macros being force-cleared then force-set ?

compat/sched_affinity.h
```diff
@@ -0,0 +1,85 @@
...
+#define cpu_set_t cpuset_t
+#else
+typedef struct {
+    int bits;
+} cpu_set_t;
+#endif
```
yadij, line 27: The struct definition was previously wrapped in a type-specific check for cpu_set_t existence, which is better portability practice than relying on a header name.

Please separate this from the headers include block and return it to being as-needed. Doing so will fix the issues around HAVE_CPU_SET_T as you can see:
```suggestion

#if !HAVE_CPU_SET_T
// Linux 2.6 and newer renamed it cpuset_t
#if HAVE_CPUSET_T
typedef cpuset_t cpu_set_t;
#else
// No alternative known, define it ourselves
typedef struct {
    int bits;
} cpu_set_t;
#endif
#define HAVE_CPU_SET_T 1
#endif
```

compat/cpu.h
```diff
@@ -14,70 +14,9 @@
 #endif
 #if HAVE_SCHED_H
 #include <sched.h>
+#else
+#include "compat/sched_affinity.h"
```
yadij, line 18: There is no need to keep an empty compat/cpu.h file around. The files using it should include compat/sched.h instead.

FYI, if you did this as a rename-then-edit it may help clarify which code you are actually changing than telling us the entire file contents here are deleted and whole new files added elsewhere (though some git tooling is not that smart AFAIK github is).

yadij (changes requested):
It is not clear to me what you are adding for FreeBSD or other OS. There is a lot of code shuffling going on and few to no comments on changes made. Also, AFAIK the portability code was already building and being used on non-Linux systems.

src/CpuAffinitySet.cc
```diff
@@ -38,7 +38,11 @@ CpuAffinitySet::apply()
     } else {
         cpu_set_t cpuSet;
         memcpy(&cpuSet, &theCpuSet, sizeof(cpuSet));
+        #if !HAVE_CPUSET_T
```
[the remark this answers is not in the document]
devnexen: I weighted that one indeed, should be better now.

configure.ac
```diff
@@ -2934,6 +2935,14 @@ AC_CHECK_TYPE(cpu_set_t,
 #endif
 ])
 
+AC_CHECK_TYPE(cpuset_t,
+   AC_DEFINE(HAVE_CPUSET_T,1,[cpuset_t is defined by the system headers]),,[
+#if HAVE_SYS_CPUSET_H
+#include <sys/param.h>
+#include <sys/cpuset.h>
+#endif
```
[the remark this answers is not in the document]
yadij: Please use the syntax shown above. It ensures that ./configure compiler can provide a clear error message about which includes and/or definitions are the problem when issues occur.

compat/sched_affinity.cc
```diff
@@ -0,0 +1,37 @@
...
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#include "squid.h"
+
+#if !HAVE_SCHED_SETAFFINITY
+
+#include "compat/sched_affinity.h"
+
+int sched_setaffinity(int pid, size_t csize, cpu_set_t *c) {
```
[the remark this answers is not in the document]
yadij: Still missing the line wrap before opening logic brace.

compat/sched.h
```diff
@@ -1,31 +1,37 @@
...
-
-#if !HAVE_CPU_AFFINITY
-/* failing replacements to minimize the number of if-HAVE_CPU_AFFINITYs */
-#if !HAVE_CPU_SET_T
+#if HAVE_SYS_CPUSET_H
+#include <sys/cpuset.h>
+typedef cpuset_t cpu_set_t;
+// CPU_AND needs a little reimplementation due to signature difference and the key difference is it s implemented as a do while loop
+#undef CPU_AND
+#define CPU_AND(destset, srcset1, srcset2) CpuAnd((destset), (srcset1), (srcset2))
```
rousskov, line 25: This PR provides replacements for the Linux-standard sched_getaffinity family of functions using a _partially overlapping_ but different FreeBSD-standard family of functions. I am pretty sure that the overlaps cannot be _safely_ handled by redefining standard macros. Sooner or later, some FreeBSD header will mention the macro will redefined, screwing Squid.

The correct solution here is, IMO, elimination of all CPU_AND() calls from Squid code outside of this compatibility shim. That code should use CpuAnd() and this shim should provide that function in all cases, including Linux. The Linux version will implement it using CPU_AND (where available), of course.

compat/sched.h
```diff
@@ -79,5 +85,9 @@ CpuAnd(cpu_set_t *destset, const cpu_set_t *srcset1, const cpu_set_t *srcset2)
 }
 #endif /* CPU_AND */
 
-#endif /* SQUID_COMPAT_CPU_H */
+#endif
```
rousskov, line 88: Please provide a trailing comment to indicate which `#if` is closed here. I am not sure, but it may be HAVE_SYS_CPUSET_H.

configure.ac
```diff
@@ -2934,6 +2936,16 @@ AC_CHECK_TYPE(cpu_set_t,
 #endif
 ])
 
+AC_CHECK_TYPE(cpuset_t,
+   AC_DEFINE(HAVE_CPUSET_T,1,[cpuset_t is defined by the system headers]),,[
```
rousskov, line 2940: AFAICT, HAVE_CPUSET_T is currently unused in this PR (but it probably should be used!).

rousskov (changes requested):
AFAICT, this PR adds support for CPU affinity-related features to FreeBSD and alike. Such an improvement is very welcomed, thank you!

rousskov:
> also couple of other OS support cpu affinity, e.g. NetBSD, with different APIs

Could you please clarify whether this PR also adds cpu_affinity_map support for NetBSD? I am not sure whether your quoted comment in the PR description implied that. If it did not imply that, why do you want to mention NetBSD?

rousskov:
@devnexen, have you tested this PR on FreeBSD? If not, please do. I do not see how this PR can enable cpu_affinity_map support, but perhaps I am missing something.

configure.ac
devnexen: It does

devnexen:
> > also couple of other OS support cpu affinity, e.g. NetBSD, with different APIs
> 
> Could you please clarify whether this PR also adds cpu_affinity_map support for NetBSD? I am not sure whether your quoted comment in the PR description implied that. If it did not imply that, why do you want to mention NetBSD?

It is more to explain future portability implementations possibility.

devnexen:
> @devnexen, have you tested this PR on FreeBSD? If not, please do. I do not see how this PR can enable cpu_affinity_map support, but perhaps I am missing something.

I did in the first stage then with all the changes it did not work anymore indeed. Should be ok with the changes.

compat/Makefile.am
```diff
@@ -25,7 +25,6 @@ libcompatsquid_la_SOURCES = \
 	compat.h \
 	compat_shared.h \
 	cppunit.h \
-	cpu.h \
```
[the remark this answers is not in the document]
yadij: Since we initially added it the various mixed OS filenames have been renamed and settled on that strange sched.h name.  That one OS file is the include providing all the CPU_* macros, the sched_* functions and the cpuset types. That is why it should be called compat/sched.* per libcompat source layout design.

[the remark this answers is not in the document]
rousskov: >  That one OS file is the include providing all the CPU_* macros, the sched_* functions and the cpuset types

That statement is very inaccurate AFAICT: FreeBSD does not even have `<sched.h>`, and there is [<sys/cpuset.h>](https://github.com/freebsd/freebsd-src/blob/master/sys/sys/cpuset.h) that provides the `CPU_AND` macro that we are wrapping (among other things).

Overall, I see no compelling reason to rename this file. I think the current `cpu.h` name is overall better than `sched.h` and cannot contradict any reasonable design rules. However, if you insist, please ask the author to rename again, and I will not re-raise this issue in this PR. I consider my change request addressed.

devnexen:
Is there any remaining concern as there is waiting for author label ? :)

configure.ac
```diff
@@ -2934,6 +2936,16 @@ AC_CHECK_TYPE(cpu_set_t,
...
+#if HAVE_SYS_PARAM_H
+#include <sys/param.h>
+#endif
+#if HAVE_SYS_CPUSET_H
+#include <sys/cpuset.h>
+#endif
+])
```
rousskov, line 2947: Despite the "cpuset_t is defined by the system headers" description, this code does not seem to check whether `cpuset_t` is defined. For example, it will define HAVE_CPUSET_T even when there are no system headers and, hence, no cpuset_t is defined.

compat/cpu.h
```diff
@@ -9,23 +9,36 @@
...
+typedef cpuset_t cpu_set_t;
+#endif
+#define HAVE_CPU_AFFINITY 1
+inline void
+CpuAnd(cpu_set_t *destset, const cpu_set_t *srcset1, const cpu_set_t *srcset2)
+{
+    cpuset_t desttmp;
+    CPU_COPY(srcset1, &desttmp);
+    CPU_AND(&desttmp, srcset2);
+    CPU_COPY(&desttmp, destset);
```
rousskov, line 36: Do we have to introduce a temporary CPU set here?

```suggestion
    CPU_COPY(srcset1, destset);
    CPU_AND(destset, srcset2);
}
```

compat/cpu.h
```diff
@@ -9,23 +9,36 @@
...
+#endif
+#if HAVE_SYS_PARAM_H
+#include <sys/param.h>
+#endif
+#if HAVE_SYS_CPUSET_H
+#include <sys/cpuset.h>
+#if HAVE_CPUSET_T
+typedef cpuset_t cpu_set_t;
+#endif
+#define HAVE_CPU_AFFINITY 1
```
rousskov, line 29: We should not reset the value defined/controlled by our own ./configure. AFAICT, either ./configure should not set HAVE_CPU_AFFINITY or this code should not set HAVE_CPU_AFFINITY.

```
configure.ac:if test "x$ac_cv_func_sched_getaffinity" = "xyes" -a "x$ac_cv_func_sched_setaffinity" = "xyes" ; then
configure.ac:  AC_DEFINE(HAVE_CPU_AFFINITY,1,[Support setting CPU affinity for workers])
configure.ac:fi
```

compat/cpu.h
```diff
@@ -9,23 +9,36 @@
...
+{
+    cpuset_t desttmp;
+    CPU_COPY(srcset1, &desttmp);
+    CPU_AND(&desttmp, srcset2);
+    CPU_COPY(&desttmp, destset);
+}
+#else
 typedef struct {
     int bits;
 } cpu_set_t;
```
rousskov, line 41: AFAICT, this code is now enabled when HAVE_SCHED_SETAFFINITY is set/true. Does HAVE_SCHED_SETAFFINITY really imply that we should be defining cpu_set_t?! Or did I misparsed the #if/#else combinations above?

compat/cpu.h
```diff
@@ -79,5 +92,15 @@ CpuAnd(cpu_set_t *destset, const cpu_set_t *srcset1, const cpu_set_t *srcset2)
 }
 #endif /* CPU_AND */
 
-#endif /* SQUID_COMPAT_CPU_H */
+#endif /* HAVE_SYS_CPU_SET_H */
```
rousskov, line 95: I do not see HAVE_SYS_CPU_SET_H mentioned earlier in this file.

rousskov (changes requested):

compat/cpu.h
devnexen: No, this else is when !HAVE_SCHED_SETAFFINITY and !HAVE_SYS_CPUSET_H

rousskov: Ah, I see that `#if` now. The code formatting/comments could be improved to make this clearer, but your answer addresses my change request.

compat/cpu.h
```diff
@@ -9,23 +9,34 @@
...
+#if HAVE_STDBOOL_H
+#include <stdbool.h>
+#endif
+#if HAVE_SYS_PARAM_H
+#include <sys/param.h>
+#endif
+
+#if HAVE_SYS_CPUSET_H
+#include <sys/cpuset.h>
+#if HAVE_CPUSET_T
```
rousskov, line 27: The hidden assumptions and dependencies in the above code make it very difficult to understand and maintain IMO. I rewrote this comment several times because it is difficult for me to tease out the exact problems -- they are all intermingled with those implicit assumptions. At the end, I decided to focus on one specific bug (which is not even visible in the diff) for now:

```C++
// glibc prior to 2.6 lacks CPU_COUNT
...
#define CPU_COUNT(set) CpuCount(set)
...

// glibc prior to 2.7 lacks CPU_AND
...
#define CPU_AND(destset, srcset1, srcset2) CpuAnd((destset), (srcset1), (srcset2))
...
```

AFAICT, the above (and other) definitions existed, in part, to provide missing macros in Linux environments where HAVE_SCHED_SETAFFINITY is true. Now, those definitions got secretly moved into a barely visible "!HAVE_SCHED_SETAFFINITY && !HAVE_SYS_CPUSET_H" section which has nothing to do with those Linux environments.

Again, it is difficult for me to give specific advice on how to fix this because I have problems with the overall approach. If you are tired of fixing this PR, please let me know. I will try to find the time to rewrite these sections for testing on your side.

rousskov (changes requested):
Thank you for addressing my previous concerns.

compat/cpu.cc
```diff
@@ -0,0 +1,51 @@
...
+	return cpuset_setaffinity(CPU_LEVEL_WHICH, CPU_WHICH_PID, pid, csize, c);
+#else
+	return ENOTSUP;
+#endif
+}
+
+int
+sched_getaffinity(pid_t pid, size_t csize, cpu_set_t *c)
+{
+#if HAVE_CPUSET_SETAFFINITY
```
yadij, line 42: Need a configure check for cpuset_getaffinity for this.

yadij (changes requested):
Since there seems to be some trouble getting the portability aspects of the new replacements right I have adjusted the cpu.h contents myself in #6d362cf and added the missing check for cpuset_t the polishing highlighted.

Please confirm these changes still work for your OS use-cases.

devnexen:
> Since there seems to be some trouble getting the portability aspects of the new replacements right I have adjusted the cpu.h contents myself in #6d362cf and added the missing check for cpuset_t the polishing highlighted.
> 
> Please confirm these changes still work for your OS use-cases.

Build breaks under FreeBSD, but that small potatoes, just missing type.

PR #763, opened 2021-01-29
Source Format Enforcement

squidadm:


yadij:
OK to test

rousskov (commented):
I suggest removing the years from copyright statements instead of modifying 2000+ files every year. However, I will not block this PR on this ground.

As far as US (and, I bet, most other countries) copyright laws are concerned,

1. these years do not follow the required "first publication year" format of a formal copyright notice,
1. most years _invalidate_ the formal notice by containing a year _past_ the first publication year, and
1. the entire notice itself is [not required](https://copyright.uslegal.com/copyright-notice/omission-of-notice-or-error-in-notice/) for stuff created after 1989 A.D. In the 179 countries that signed the Berne Convention, creations are automatically copyrighted, with or without a valid notice.

scripts/source-maintenance.sh
```diff
@@ -153,6 +153,9 @@ applyPluginsTo ()
 # succeeds if all MakeNamedErrorDetail() names are unique
 checkMakeNamedErrorDetails ()
 {
+    # old git versions do not support --only-matching, skip this check
+    git grep --only-matching . . >/dev/null 2>&1 || return 0
```
rousskov, line 157: We should not skip this check just because it does not work. It is wrong to admit MakeNamedErrorDetail bugs into the repository while pretending to check that there are no such bugs.

If really needed, we can allow the person running source-maintenance.sh to manually disable some required checks via command-line options, but I am not sure the complexity is worth it -- that person can already remove checks by editing the script. The end goal here is to (automatically) format every pull request anyway, removing the vast majority of these problems.

rousskov (dismissed):
This change should not have been sneaked into the supposed-to-be-automated pull request.

scripts/source-maintenance.sh
yadij: It is unfortunate that we did not catch the --only-matching problem when the MakeNamedErrorDetail changes went in. Right now we need to disable this check on master server due to old git version 2.17. Manual runs with newer git supporting that option still run fine and can detect MakeNamedErrorDetail bugs in local tests we do.

Until we get an upgrade of master and/or git on master we need to skip this test. If you want to make it run in a robust way feel free to do a new PR re-writing the scripts test without that option.

rousskov: > Right now we need to disable this check on master server due to old git version 2.17
...
> Until we get an upgrade of master and/or git on master we need to skip this test.

Feel free to do that on the master server (e.g., one can modify the script or alias git). The need for a local workaround does not, IMO, justify violating an important principle in the official script.

> If you want to make it run in a robust way ...

Personally, I do not want that -- I think a git upgrade is an overall better way forward here. However, I would not block quality attempts to compensate for the lack of `git --only-matching` support (in dedicated PRs).

yadij:
> I suggest removing the years from copyright statements instead of modifying 2000+ files every year. However, I will not block this PR on this ground.
> 
> As far as US (and, I bet, most other countries) copyright laws are concerned,

Squid is distributed and used in jurisdictions outside those areas covered by the Berne Convention. The understanding I gained during the copyright update project in 2014 is that to ensure best legal application we do need to explicitly use claim statements. Also that those statements use a syntax containing the historic copyright symbol, word and year indicators. Each has an explicit legal side effect in laws of countries outside the Berne Convention jurisdiction - as do omission and/or errors in those indicators.

> 
>     1. these years do not follow the required "first publication year" format of a formal copyright notice,
> 

The year value follows the format advised by FSF for range of years where _GPL_ copyrights are claimed. Since we release files periodically they are important and mean that any code unique to eg Squid-3.5.1 expires into public domain N years after 2015. Versus code unique to Squid-4.1 expiring into public domain N years after 2020. Where N is the country-specific duration for validity of software copyright claims made by organizations.


>     2. most years _invalidate_ the formal notice by containing a year _past_ the first publication year, and
> 

As discussed during the initial project planning circa 2013-2014;

 * the statement being modified here is a "collective works" claim. It is claiming that our published file*s* were written across all years between 1996 and current year.

 * challenges or third-party violation legalities are left to the specific jurisdiction legal process to determine exact coverage of the lines relevant to the case.

 * being a collection - subsections may exist under dual license. Any individual file may contain additional claims that dual-license that files content in whole or in part at the discretion of the originating author(s).

 * being a collection - each file covered by the claim contains it. There *are* files in the tarball exempt from the GPL coverage. Those are not covered by this claim - either by their nature as non-source with omission of this line (eg icons), or by explicit use of *only* other copyright statements (eg translated texts).


>     3. the entire notice itself is [not required](https://copyright.uslegal.com/copyright-notice/omission-of-notice-or-error-in-notice/) for stuff created after 1989 A.D. In the 179 countries that signed the Berne Convention, creations are automatically copyrighted, with or without a valid notice.

No. The notice is designed and used explicitly to make our claim legally enforceable in maximum possible number of legal jurisdictions. As of this writing there are 20 countries and a number of non-country states and territories who are not Berne signatories.

scripts/source-maintenance.sh
yadij: IMO this does not "admit MakeNamedErrorDetail bugs" because jenkins and semaphore build checks run before each commit use a version of git which copes fine with that option. It is only on the master machine performing release maintenance where it breaks the automation.

I am happy to move the script change to its own PR for clarity if you like. I only left it in the maintenance brance for this PR because I knew this would be getting a QA review before merge, and as expected you spotted it.

PS. the v4 and v5 updates went in smoothly because they were 100% automatic changes. This maintainer commit was not needed for those versions.

eduard-bagdasaryan: FYI: `--only-matching` became available since git [2.19](https://git-scm.com/docs/git-grep/2.19.0#git-grep--o).

rousskov:
> Squid is distributed and used in jurisdictions outside those areas covered by the Berne Convention.

That is a known fact, but you have not demonstrated its relevance. To become relevant, somebody will need to find a country or place where the removal of the year would make a serious negative impact. Ideally, somebody should also show that our handwaving about the year range and year values will _not_ have a negative impact in the rest of those ~5-12 countries and territories that are not covered by Berne or [TRIPS](https://en.wikipedia.org/wiki/TRIPS_Agreement) (the latter has the same provisions making explicit copyright notices obsolete as the Berne Convention).

> The understanding I gained during the copyright update project in 2014 is that to ensure best legal application we do need to explicitly use claim statements.

To justify modifying 2000 files every year, I would like to see a lot more than a statement of understanding. I have substantiated my understanding (that those years are not helping) by references to specific laws and legal opinions. AFAICT, you pretty much have not, and some of your arguments have been invalid, further reducing my confidence in the correctness of unsubstantiated conclusions.

> Alex: 1. these years do not follow the required "first publication year" format of a formal copyright notice,

> Amos: The year value follows the format advised by FSF for range of years where GPL copyrights are claimed

First of all, those two statements do not contradict each other. Whether we are breaking the formal format by following FSF alleged recommendation or on our own volition is irrelevant here.

Second, we probably do _not_ actually follow FSF [suggestion](https://www.gnu.org/licenses/gpl-howto.html) because we do not satisfy its preconditions. See below for the FSF text quote.

Third, it is not a recommendation, but a permission AFAICT: FSF has recognized that listing all years has became impractical so they graciously let folks abbreviate. They do not recommend abbreviation; they just say it is "okay". This permission does not address other problems with (the literal interpretation of) their recommendations, but it is one problem that they could easily address. Their liberal attitude to the format of the formal copyright notice only underscores my argument that once we violate the legal format in one aspect, there is no point in trying to preserve some other aspects of that format regardless of the costs -- I bet FSF knows that the legal notice format is essentially irrelevant these days, and so they feel comfortable playing with it.

I will quote FSF text here to substantiate the above claims:

> FSF: For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage.

BTW, FSF is not an infallible entity -- some of their recommendations are bad for Squid -- but we can still follow their year-range suggestion in some single source file that summarizes Squid licensing policies. We do not need to modify 2000+ files every year to be in sync with that specific FSF suggestion.

> Since we release files periodically they are important and mean that any code unique to eg Squid-3.5.1 expires into public domain N years after 2015. Versus code unique to Squid-4.1 expiring into public domain N years after 2020. Where N is the country-specific duration for validity of software copyright claims made by organizations.

True but irrelevant. The same would happen if we use a single/centralized notice with the range of years. Or no notice at all.

> Alex: 2. most years _invalidate_ the formal notice by containing a year _past_ the first publication year, and

> Amos: As discussed during the initial project planning circa 2013-2014; ...

All those facts are not related to my statement or its implications AFAICT:

* If we believe that the last year in the copyright notice in file X applies to that file X, then the fact I stated stands -- many of those notices are invalidated by our outomated "future dating" of files without fresh copyrightable modifications.
* If we believe that the year range in copyright notice in file X only applies to the whole _collection_ of Squid files, then we are actually not specifying a year for each specific file and, hence, moving the range of years to a centralized location (referred to by the copyright statement in each file) is not going to harm any individual file (any more than it is already harmed by the essentially dateless statement now). In fact, such a move will only improve the clarity of our collection-scope claim!

> Alex: 3. the entire notice itself is [not required](https://copyright.uslegal.com/copyright-notice/omission-of-notice-or-error-in-notice/) for stuff created after 1989 A.D. In the 179 countries that signed the Berne Convention, creations are automatically copyrighted, with or without a valid notice.

> Amos: No. The notice is designed and used explicitly to make our claim legally enforceable in maximum possible number of legal jurisdictions. As of this writing there are 20 countries and a number of non-country states and territories who are not Berne signatories.

I have demonstrated that the notice is not required in all but ~5-12 places. You have not demonstrated that it is required in any single place! Just because there are places that have not signed the Berne Convention or joined TRIPS Agreement does not imply that our notice is required in those places (and certainly does not imply that removing the year range from that notice in individual files will harm Squid in those places!).

You could try to shift that burden of proof to me, but, IMO, the onus of proving the legal necessity of modifying 2000+ files every year is on the person insisting on that modification.

scripts/source-maintenance.sh
rousskov: > IMO this does not "admit MakeNamedErrorDetail bugs" because jenkins and semaphore build checks run before each commit use a version of git which copes fine with that option

1. AFAIK, Semaphore checks do not run `scripts/source-maintenance.sh`. That is still a (complicated) TODO. Not sure about Jenkins, but I suspect the same is true there.
2. Jenkins and Semaphore do not insist that git they use supports `--only-matching` so even if they start running `scripts/source-maintenance.sh`, they might admit MakeNamedErrorDetail bugs in the future if the proposed script change is accepted.

> I am happy to move the script change to its own PR for clarity if you like.

Extracting a PR-unrelated problem into its own PR will not address that problem, of course, but it will help move _this_ PR forward. AFAICT, this change request will apply "as is" to the new PR (containing the same modifications).

> I only left it in the maintenance brance for this PR because I knew this would be getting a QA review before merge, and as expected you spotted it.

FYI: Spotting this problem in a supposed-to-be-automated-by-existing-scripts modification of 2000+ files was an accident.

yadij:
I have re-done this PR as two commits based on todays master branch. It drops the controversial change to scripts/source-maintenance.sh which is the key issue blocking merge (copyright discussions is a board@ policy issue).

* Commit 88a038a contains all code and format related changes.
* Commit df5cca8 should contain only the usual copyright line updates.

rousskov (commented):
Thank you for addressing my blocking concern. I did not review every changed file, but I trust that there are no manual changes in the current PR version.

I can only hope that we will remove the year range instead of modifying 2000+ files again next year.

yadij:
Restored the PR diff to just the copyright and enforcement script changes. Not sure how the bunch of master commits got added.

rousskov:
> PR approval   `Pending — waiting for more votes` 

@yadij, I have not re-checked the PR changes, but please note that this PR has not been approved by any Core developer and, hence, Anubis bugs notwithstanding, will not be merged. If you are confident that the current PR changes are correct, then you may, of course, approve the PR to allow merging.

Somebody will also need to fix and/or retry Jenkins tests, as usual.

yadij (approved):

PR #764, opened 2021-01-30
Bug 5073: Compile error: index was not declared in this scope (#740)

squidadm:
Use strchr(3) instead of a legacy POSIX.1-2001 index(3) API.

Also removed the index() implementation on MS Windows as no longer used.


yadij:
OK to test

PR #770, opened 2021-02-09
Detail certificate validation errors during TLS handshake

chtsanti:
Fix certificate validation error handling in Security::Connect/Accept().
The existing validation details were not propagated/copied to IoResult,
requiring the caller to extract them via ssl_ex_index_ssl_error_detail.
The clunky approach even required a special "ErrorDetail generations"
API to figure out which error detail is "primary": the one received in
IoResult or the just extracted one. That API is removed now.

This change is used by the upcoming improvements that fetch missing TLS
v1.3 server certificates, but it also has an immediate positive effect
on the existing reporting of the client certificate validation errors.
Currently, only a general TLS error is reported for those cases because
Security::Accept() code forgot to check ssl_ex_index_ssl_error_detail.

### DIFF
--- a/src/security/ErrorDetail.cc
+++ b/src/security/ErrorDetail.cc
@@ -438,11 +438,8 @@ Security::ErrorNameFromCode(const ErrorCode err, const bool prefixRawCode)
 
 /* Security::ErrorDetail */
 
-uint64_t Security::ErrorDetail::Generations = 0;
-
 /// helper constructor implementing the logic shared by the two public ones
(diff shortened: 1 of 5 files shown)


rousskov (approved):
LGTM. I especially like how you were able to remove the Security::ErrorDetail::Generations complication, thank you!

If you have a sample of an error message that the Squid admin would see in cache.log and access.log before and after this fix, please post that here to illustrate the difference this fix makes.

yadij (approved):
LGTM

PR #772, opened 2021-02-12
Fix build on Fedora Rawhide

uhliarik:
* add SYSTEMD_LIBS to all binaries using client_side.cc, fixing linking
* add `<limits>` to all sources using std::numeric_limits, fixing gcc-11
  builds

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -2074,6 +2074,7 @@ tests_test_http_range_LDADD = \
 	$(SSLLIB) \
 	$(KRB5LIBS) \
 	$(LIBCPPUNIT_LIBS) \
+	$(SYSTEMD_LIBS) \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 tests_test_http_range_LDFLAGS = $(LIBADD_DL)
@@ -3112,7 +3113,6 @@ nodist_tests_testEventLoop_SOURCES = \
 tests_testEventLoop_LDADD = \
 	base/libbase.la \
 	$(LIBCPPUNIT_LIBS) \
-	$(SYSTEMD_LIBS) \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 tests_testEventLoop_LDFLAGS = $(LIBADD_DL)
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -2465,6 +2465,7 @@ tests_testHttpRequest_LDADD = \
 	$(SSLLIB) \
 	$(KRB5LIBS) \
 	$(LIBCPPUNIT_LIBS) \
+	$(SYSTEMD_LIBS) \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 tests_testHttpRequest_LDFLAGS = $(LIBADD_DL)
@@ -2771,6 +2772,7 @@ tests_testCacheManager_LDADD = \
 	$(SSLLIB) \
 	$(KRB5LIBS) \
 	$(LIBCPPUNIT_LIBS) \
+	$(SYSTEMD_LIBS) \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 tests_testCacheManager_LDFLAGS = $(LIBADD_DL)
@@ -3093,6 +3095,7 @@ tests_testEvent_LDADD = \
 	$(SSLLIB) \
 	$(KRB5LIBS) \
 	$(LIBCPPUNIT_LIBS) \
+	$(SYSTEMD_LIBS) \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 tests_testEvent_LDFLAGS = $(LIBADD_DL)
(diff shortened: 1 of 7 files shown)


yadij (approved):

yadij:
OK to test

src/acl/ConnMark.cc
```diff
@@ -15,6 +15,7 @@
 #include "Debug.h"
 #include "http/Stream.h"
 #include "sbuf/Stream.h"
+#include <limits>
```
rousskov, line 18: Please double check whether this header is needed. I do not see any `std` names in this file.

rousskov (commented):
Thank you for working on fixing build on Fedora Rawhide!

I suspect that these changes were done for Squid v5, but the pull request was then generated against master. In master, some of the PR changes are not needed AFAICT (see one of the inlined change requests), but similar changes might also be missing in other files (see below).

To fix the `#include <limits>` problem in master files, I suggest using the following command to find candidates:

```bash
git grep -l std::numeric_limits | xargs fgrep -L 'include <limits>'
```

AFAICT, the above will add the following files to the list of those that should be changed in a master-based PR: src/ip/QosConfig.cc, src/ipc/mem/PageStack.cc, and src/ssl/helper.cc. In some (possibly all) cases, these files may be getting `<limits>` from another `#include`, but we should not rely on such indirect inclusion in most cases (as the need for this PR illustrates).

One can (and probably should) build master on the target platform to avoid guessing, of course.

src/security/ServerOptions.cc
```diff
@@ -18,6 +18,7 @@
 #if USE_OPENSSL
 #include "compat/openssl.h"
 #include "ssl/support.h"
+#include <limits>
```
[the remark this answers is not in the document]
uhliarik: My bad, somehow missed that I'm addind this include inside guards - fixed. Thanks!

uhliarik:
I will fill another pull request for v5. Thanks for reviewing.

src/Makefile.am
```diff
@@ -2465,6 +2465,7 @@ tests_testHttpRequest_LDADD = \
 	$(SSLLIB) \
 	$(KRB5LIBS) \
 	$(LIBCPPUNIT_LIBS) \
+	$(SYSTEMD_LIBS) \
```
rousskov, line 2468: I am curious how you picked targets that need SYSTEMD_LIBS.

I see that SYSTEMD_LIBS was added in commit 6fa8c6643 along with modification of client_side.cc to call sd_notify(). That library call is still there and was not copied to other source files. I am guessing that you are adding SYSTEMD_LIBS because of that sd_notify() call in client_side.cc. However, this PR:

* does not add SYSTEMD_LIBS to some of the targets that do include client_side.cc (e.g. tests_test_http_range);
* adds SYSTEMD_LIBS to some of the targets that do not include client_side.cc (e.g. tests_testEventLoop).

If current PR additions were done by blind trial and error, then please double check that tests_testEventLoop addition is really needed in your environment (if yes, it would be interesting to understand why it is needed). And you might want to cover tests_test_http_range, just in case and for consistency sake (especially if it turns out that 
tests_testEventLoop is _not_ needed).

----

FWIW, here is a list of all targets that contain client_side.cc AFAICT, obtained via `egrep -h '_SOURCES =|[^_]client_side.cc' src/Makefile.am | fgrep -B1 client_side.cc`.


```
squid_SOURCES = \
        client_side.cc \

tests_test_http_range_SOURCES = \
        client_side.cc \

tests_testHttpRequest_SOURCES = \
        client_side.cc \

tests_testCacheManager_SOURCES = \
        client_side.cc \

tests_testEvent_SOURCES = \
        client_side.cc \
```

rousskov (commented):
Thank you for addressing my previous concerns. I am (still) not blocking this PR but if you can investigate the new issue highlighted in the inlined change request, please do so.

src/Makefile.am
uhliarik: Of course, that's correct :(. In tests_testEventLoop, we dont need to add SYSTEMD_LIBS (in v5, it is needed) and for tests_test_http_range  - SYSTEMD_LIBS are needed as you mentioned above - because sd_notify() call in client_side.cc. 

I somehow removed SYSTEMD_LIBS addition from tests_test_http_range by mistake - such build would fail as well. Adding this change to v5, since client_side.o is also used in tests_test_http_range.

rousskov (approved):
Thank you for addressing my concerns. I polished the PR description to be more precise about the rules behind these changes. Please check that I did not screw anything up. Otherwise, I think this PR is ready to go in.

AFAICT, @kinkie has been aborting Jenkins tests, so I trust he will restart testing this PR when that becomes possible.

uhliarik:
> Thank you for addressing my concerns. I polished the PR description to be more precise about the rules behind these changes. Please check that I did not screw anything up. Otherwise, I think this PR is ready to go in.
> 
> AFAICT, @kinkie has been aborting Jenkins tests, so I trust he will restart testing this PR when that becomes possible.

Thank you Alex for your review! Description is describing exactly what this PR should fix. LGTM.

PR #773, opened 2021-02-14
Fix build on Fedora Rawhide

uhliarik:
* add SYSTEMD_LIBS to all binaries using client_side.cc, fixing linking
* add `<limits>` to all sources using std::numeric_limits, fixing gcc-11
  builds


kinkie:
ok to test

rousskov:
We should finish and merge #772 first.

yadij:
Aye. Waiting on at least the feedback about systemd linkage query in #772.

PR #779, opened 2021-02-22
Delete a few unnecessary checks for standard C++ features

rousskov:
* nullptr ought to be provided by any C++ compiler we support
* HAVE_NULLPTR_T was unused while std::nullptr_t is used w/o it
* HAVE_UNIQUE_PTR was almost unused while std::unique_ptr is used w/o it


yadij (dismissed):
These checks are needed for GCC-4 where C++11 was only conditional. Squid is still being built on CentOS 7 and RHEL 7 with GCC-4.

Unless there is a strong reason to remove we should only do bug fixes on these until at least June 2024 when RHEL 7 becomes deprecated.

rousskov:
> These checks are needed for GCC-4

What makes you think that? Clearly, unused checks are not needed, for any compiler/environment, so your statement is at least partially wrong for sure, but I am trying to figure out whether it is partially right because some of the possibly used checks are actually needed for building master with GCC-4.

kinkie:
I agree with Alex. While in theory these checks are needed, in practice I would expect the build to fail on any environment where these do not work, so in practice, if we have a successful build, I would recommend to remove and simplify.

kinkie (approved):
I support this cleanup

rousskov:
> While in theory these checks are needed

And even that assertion is stretching the truth: The results of one of those checks are never _used_ in the current master code. That unused check runs but has no effect on the compiled code. It may have been needed and used some time ago, of course, but that does not matter today. Another check is essentially in the same category -- its result is mentioned in unit tests, but the same type is used without consulting that result throughout the code.

This analysis leaves only one check: `nullptr`. I bet that check is essentially unused like the rest of them (in any supported environment), but, in this specific case, I cannot prove that speculation by Squid code analysis alone. That is why I asked Amos whether he knows that Squid fails to build without this check somewhere.

FWIW, g++ v4.7.1 does provide `nullptr` according to my [test](https://godbolt.org/z/YzM5MT).

yadij:
Okay, I was wrong about nullptr still being needed.

std::unique_ptr is not just a mention in the cppunit compat code. It was actively preventing cppunit macros being un-defined and replaced in the older compilers where auto_ptr was correct.

I am tentatively marking this as cleared for merge since we have two votes for (@kinkie, @rousskov) and my review is just commentary. We should keep an eye out for any re-appearance of user compile issues with cppunit in the next year or so.

rousskov:
> Amos: I am tentatively marking this as cleared for merge

@yadij, you are still blocking this PR because you have not dismissed your blocking review. Please dismiss or update your review. If you just dismiss that old review, then please also dismiss the review request; otherwise Anubis will probably continue to wait for your review as requested.

yadij (approved):
Apparently I cannot dismiss a review request when being the requestee. Marking as approved simply to let Anubis do its thing for merge.

rousskov:
> I cannot dismiss a review request when being the requestee.

FYI: You can do it by clicking on the gears icon in the Reviewers box and then clicking on your name in the list (it should have a check mark next to it). At the very least, that procedure works for me when I self-request a review.

PR #785, opened 2021-03-02
Replace defective Must2(c, "not c") API

rousskov:
... with Must3(c, "c", Here())

* We should not make a Must*() API different from the standard C++
  static_assert() API. It is just too confusing, especially since the
  two calls are very closely related.

* We should not tell Must*() writers to specify one condition (i.e. what
  must happen) but then describe the opposite condition (i.e. what went
  wrong). When the text describes the opposite condition, it is easy for
  a human writing or thinking about the text to type the wrong
  condition: Must2(got < need, "incomplete message") looks correct!

We should not keep the same macro name when changing the meaning of the
second parameter. Fortunately, adding a third argument to the macro fits
nicely into how modern Squid code should pass the source code location.

Must3() does not support SBuf descriptions. I tried to support them, but
doing so without duplicating non-trivial code is too difficult, and
since the current code does not actually use them, wasteful.

If future (complex) code needs SBuf conditions, then it should probably
use an explicit throw statement instead, especially since Must*() is,
like assert(), supposed to quickly check source code sanity rather than
validate unsafe input. A compile-time condition description and the
source code location is usually enough for sanity checks, while proper
reporting of invalid input usually also requires parsing context info.

### DIFF
--- a/src/sbuf/SBuf.cc
+++ b/src/sbuf/SBuf.cc
@@ -148,7 +148,7 @@ SBuf::rawAppendFinish(const char *start, size_type actualSize)
     debugs(24, 8, id << " finish appending " << actualSize << " bytes");
 
     size_type newSize = length() + actualSize;
-    Must2(newSize <= min(maxSize,store_->capacity-off_), "raw append overflow");
+    Must3(newSize <= min(maxSize, store_->capacity-off_), "raw append fits", Here());
     len_ = newSize;
     store_->size = off_ + newSize;
 }
@@ -258,7 +258,7 @@ SBuf::vappendf(const char *fmt, va_list vargs)
     va_copy(ap, vargs);
     sz = vsnprintf(space, spaceSize(), fmt, ap);
     va_end(ap);
-    Must2(sz >= 0, "vsnprintf() output error");
+    Must3(sz >= 0, "vsnprintf() succeeds", Here());
 
     /* check for possible overflow */
     /* snprintf on Linux returns -1 on output errors, or the size
@@ -270,7 +270,7 @@ SBuf::vappendf(const char *fmt, va_list vargs)
         requiredSpaceEstimate = sz*2; // TODO: tune heuristics
         space = rawSpace(requiredSpaceEstimate);
         sz = vsnprintf(space, spaceSize(), fmt, vargs);
-        Must2(sz >= 0, "vsnprintf() output error despite increased buffer space");
+        Must3(sz >= 0, "vsnprintf() succeeds (with increased buffer space)", Here());
     }
 
     // data was appended, update internal state
--- a/scripts/calc-must-ids.pl
+++ b/scripts/calc-must-ids.pl
@@ -66,7 +66,7 @@ sub ComputeMustIds
         $line =~ s@/[*].*?[*]/@@; # strip simple single-line /* comments */
 
         my($id);
-        if ($line =~ /\bMust2?\s*\(/ || # Must(...) and Must2(...)
+        if ($line =~ /\bMust\s*\(/ || # Must(...)
             $line =~ /\bTexcHere\s*\(/ || # TexcHere(...)
             $line =~ /\bHere\s*\(\s*\)/) { # Here()
             $line =~ s/^\s*//;
--- a/src/base/TextException.cc
+++ b/src/base/TextException.cc
@@ -8,9 +8,7 @@
 
 #include "squid.h"
 #include "base/TextException.h"
-#include "base/Optional.h"
 #include "sbuf/SBuf.h"
-#include "SquidMath.h"
 
 #include <iostream>
 #include <sstream>
@@ -59,49 +57,6 @@ TextException::what() const throw()
 
     return result.what();
 }
-
-SBuf
-MakeFailedCheckMessage(const SBuf &message)
-{
-    static const SBuf prefix("check failed: ");
-    static const SBuf lastResortMessage("check failed: [no details]");
-    static SBuf buf; // optimization to prevent some allocations
-
-    // try hard to handle recursive failures
-    static bool making = false;
-    if (making)
-        return lastResortMessage;
-
-    making = true;
-    try {
-        if (const auto combinedLength = Sum(prefix.length(), message.length())) {
-            buf.clear();
-            buf.reserveSpace(combinedLength.value());
-            buf.append(prefix);
-            buf.append(message);
-            making = false;
-            return buf;
-        }
-        // Handle huge messages by returning lastResortMessage without risking
-        // more troubles by reporting this huge message.
-        // [[fallthrough]]
-    }
-    catch (...) {
-        // Handle this message-forming exception by returning lastResortMessage
-        // without risking more exceptions while reporting this low-level one.
-        // [[fallthrough]]
-    }
-    making = false;
-    return lastResortMessage;
-}
-
-SBuf
-MakeFailedCheckMessage(const char *message)
-{
-    return MakeFailedCheckMessage(SBuf(message));
-}
-
-
 
 std::ostream &
 operator <<(std::ostream &os, const TextException &ex)
--- a/src/base/TextException.cc
+++ b/src/base/TextException.cc
@@ -8,7 +8,9 @@
 
 #include "squid.h"
 #include "base/TextException.h"
+#include "base/Optional.h"
 #include "sbuf/SBuf.h"
+#include "SquidMath.h"
 
 #include <iostream>
 #include <sstream>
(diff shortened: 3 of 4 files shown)


src/sbuf/SBuf.cc
```diff
@@ -148,7 +148,7 @@ SBuf::rawAppendFinish(const char *start, size_type actualSize)
     debugs(24, 8, id << " finish appending " << actualSize << " bytes");
 
     size_type newSize = length() + actualSize;
-    Must2(newSize <= min(maxSize,store_->capacity-off_), "raw append overflow");
+    Must3(newSize <= min(maxSize, store_->capacity-off_), "raw append fits", Here());
     len_ = newSize;
     store_->size = off_ + newSize;
```
rousskov, line 153: FTR, this code needs to be completely rewritten IMO, and that rewrite is likely to remove all the Musts, but I am limiting this PR to fixing/replacing Must2() API, not fixing any existing Must2() users.

src/sbuf/SBuf.cc
```diff
@@ -258,7 +258,7 @@ SBuf::vappendf(const char *fmt, va_list vargs)
     va_copy(ap, vargs);
     sz = vsnprintf(space, spaceSize(), fmt, ap);
     va_end(ap);
-    Must2(sz >= 0, "vsnprintf() output error");
+    Must3(sz >= 0, "vsnprintf() succeeds", Here());
```
rousskov, line 261: A non-negative result is how vsnprintf() manual pages define success. We can say "wrote or intended to write zero or more bytes" instead, but I think the current version is better.

src/sbuf/SBuf.cc
yadij: Can't we just drop this rarely used alternative to Must() ?

rousskov: We could, but that feels like a step backwards to me. The reason we and C++ has the second argument is to convert these typical use cases:

```C++
    Must(a < b);
```

... which forces the reader to figure out what "a < b" really means and forces the developer to figure out which of the three "a < b" assertions has failed in some unknown code (where line numbers do not help much);


```C++
    Must(a < b); // the caller gave as a malformed c value
```
... which solves the first problem but still forces the developer to figure out which of the three "a < b" assertions has failed in some unknown code (where line numbers do not help much);


```C++
    Must((a < b) && "the caller gave as a malformed c value");
```

... which solves the above problems but at the expense of a more complex expression that confuses newbies (Squid does not use this approach, but many projects do);

to 

```C++
    Must(a < b, "the caller gave as a malformed c value");
```
which solves all the problems above. I bet that is exactly why C++11 added the message argument to static_assertion (they could not do it to the C assert() macro for obvious reasons).

My only doubt here is that the current post-C++20 proposal does not add the message argument to C++ asserts: `[[ a < b ]]`. Perhaps they will do that eventually or perhaps they just had to sacrifice the message in order to use the `[[ ... ]]` syntax. If you can find why they did not add the message, please let me know!

Just to make sure we are on the same page regarding the proposed "drop" option consequences: If we drop Must-with-a-message, would you be OK to see `Must(c && "explanation")` replacements?

yadij: > 
> My only doubt here is that the current post-C++20 proposal does not add the message argument to C++ asserts: `[[ a < b ]]`. Perhaps they will do that eventually or perhaps they just had to sacrifice the message in order to use the `[[ ... ]]` syntax. If you can find why they did not add the message, please let me know!
> 


> Just to make sure we are on the same page regarding the proposed "drop" option consequences: If we drop Must-with-a-message, would you be OK to see `Must(c && "explanation")` replacements?

IIRC there were compiler issues with the second argument never being able to be false. Static analysis highlights that as a bug.  I think we should rename the variables used to something that helps self-document when only condition is visible. sed /newSize/newSbufSize/ or such.

yadij: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3928.pdf looks like the C++-17 document making static_assert string becoming optional. Unfortunately it just lists a lot of requested changes with little or no rational behind them.

The C++23 document ("current post-C++-20 proposal" ?)  is scoped as fixing only the issue of compile errors when comma is used in syntax of conditions passed as assert() parameter. It is seeking to be explicitly backward compatible with current syntax usage, and the C language definition of assert() as a macro.

rousskov: > http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3928.pdf looks like the C++-17 document making static_assert string becoming optional. Unfortunately it just lists a lot of requested changes with little or no rational behind them.

I bet the second static_assert argument was made optional to avoid code duplication. I have seen text alluding to that in Sutter slides IIRC. After the two-argument static_assert was added, folks noticed a lot of new code that looked like this:

```C++
static_assert(c, "c");
```

Making the second argument optional was the right solution. To avoid misunderstanding, making the second argument optional does not imply the second argument should not have been added in the first place. There are two different problems here, and, as of C++17, both are solved.

> The C++23 document ("current post-C++-20 proposal" ?) is scoped as fixing only the issue of compile errors when comma is used in syntax of conditions passed as assert() parameter. It is seeking to be explicitly backward compatible with current syntax usage, and the C language definition of assert() as a macro.

The proposed `[[assert]]` language feature I was talking about is mentioned in GotW [97](https://herbsutter.com/2021/01/11/gotw-97-solution-assertions-difficulty-4-10/). It is documented in a contract-based programming support proposal [P0542R5](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0542r5.html). While there is support for contract_violation::comment(), that comment looks compiler-generated rather than developer-specified:

```C++
// Returns: Implementation-defined text describing the predicate of the violated contract.
string_view comment() const noexcept;
```

I do not know what the current `[[assert]]` adoption status is.

rousskov: > > Alex: Just to make sure we are on the same page regarding the proposed "drop" option consequences: If we drop Must-with-a-message, would you be OK to see `Must(c && "explanation")` replacements?

> IIRC there were compiler issues with the second argument never being able to be false. 

You lost me here. If we drop Must-with-a-message, there will be no second argument.

> Static analysis highlights that as a bug.

If you are thinking about `Must(false)` and equivalent, then that is a different/unrelated topic.

If you are thinking that `Must(c && "c")` will cause problems with static analysis, then that is one more reason to keep support for the message argument in Must().

> I think we should rename the variables used to something that helps self-document when only condition is visible. sed /newSize/newSbufSize/ or such.

While self-documenting code is the best outcome we can hope for, most developers will have a negative reaction to refactoring requests motivated by Must() message clarity concerns. Moreover, such refactoring is not always possible or desirable. There are legitimate reasons for using source code comments, of course, and the second Must() argument (or equivalent) is very similar to a source code comment.

rousskov:
@yadij, I will merge this PR if you do not block it.

I know you have labeled this PR as S-waiting-for-author after asking a [question](https://github.com/squid-cache/squid/pull/785#discussion_r592181897), but I do not know what you implied by that label because you did not request any changes (AFAICT). I tried to answer your question. If you want to continue discussing whether we should fix Must-with-a-message or drop it (before this PR is merged), then you should probably block this PR to clarify your position.

src/sbuf/SBuf.cc
yadij: > > > Alex: Just to make sure we are on the same page regarding the proposed "drop" option consequences: If we drop Must-with-a-message, would you be OK to see `Must(c && "explanation")` replacements?
> 
> > IIRC there were compiler issues with the second argument never being able to be false.
> 
> You lost me here. If we drop Must-with-a-message, there will be no second argument.
> 

Sorry, I meant the second boolean argument. The quoted-string part.

We used to have a few (A && "string") conditions in the code. Static analysis tools complained that the "string" part of the condition could never be false and output a bunch of false positives. Apart from that annoyance I am okay with the construct.

> > Static analysis highlights that as a bug.
> 
> If you are thinking about `Must(false)` and equivalent, then that is a different/unrelated topic.
> 
> If you are thinking that `Must(c && "c")` will cause problems with static analysis, then that is one more reason to keep support for the message argument in Must().
> 
> > I think we should rename the variables used to something that helps self-document when only condition is visible. sed /newSize/newSbufSize/ or such.
> 
> While self-documenting code is the best outcome we can hope for, most developers will have a negative reaction to refactoring requests motivated by Must() message clarity concerns. Moreover, such refactoring is not always possible or desirable. There are legitimate reasons for using source code comments, of course, and the second Must() argument (or equivalent) is very similar to a source code comment.

Good points.  I think I am convinced now that we do need a separate Must-with-message - as much of a hack as it is.

Please do seek to prefer new code that does not need it though. Must() should remain our preferred name, one day we may be able to overload it and remove the macro naming hack.

yadij (approved):

src/sbuf/SBuf.cc
rousskov: > Please do seek to prefer new code that does not need it though.

Yes, with all other factors being equal, less code is better (and self-documenting code is better than the code relying on explicit messages or comments for documentation).

PR #787, opened 2021-03-13
Fix HttpHeaderStats definition to include hoErrorDetail

rousskov:
... when Squid is built --with-openssl.

We were "lucky" that the memory area after HttpHeaderStats was not,
apparently, used for anything important enough when HttpHeader::parse(),
indirectly called from errorInitialize() during initial Squid
configuration, was writing to it.

Detected by using AddressSanitizer.

The bug was created in commit 02259ff and cemented by commit 2673511.


yadij (approved):
LGTM

rousskov:
@yadij, I added @MegaManSec to CONTRIBUTORS after you cleared this PR for merging. I hope this is OK with you.

yadij:
Of course. No problem with that.

PR #788, opened 2021-03-16
Bug 5106: Broken cache manager URL parsing

yadij:
Use already parsed request-target URL in cache manager and
update CacheManager to Tokanizer based URL parse

Removing use of sscan() and regex string processing which have
proven to be problematic on many levels. Most particularly with
regards to tolerance of normally harmless garbage syntax in URLs
received.

Support for generic URI schemes is added possibly resolving some
issues reported with ftp:// URL and manager access via ftp_port
sockets.

Truly generic support for /squid-internal-mgr/ path prefix is
added, fixing some user confusion about its use on cache_object:
scheme URLs.

TODO: support for single-name parameters and URL #fragments
are left to future updates. As is refactoring the QueryParams
data storage to avoid SBuf data copying.


src/mgr/QueryParams.cc
```diff
@@ -66,60 +69,59 @@ Mgr::QueryParams::find(const String& name) const
...
-        }
-        if (n < len) {
-            if (!ParseParam(aParamsStr.substr(n, len), param))
-                return false;
-            aParams.params.push_back(param);
-        }
+        if (!ParseParam(foundStr, param))
+            return false;
+        aParams.params.push_back(param);
     }
```
rousskov, line 121: Please try to refactor to reuse tokenizer:

```C++
while (!tok.atEnd())
    aParams.params.push_back(ParseParam(tok));
```

It may be easier to refactor this after fixing other problems.

rousskov (changes requested):
Thank you for working on fixing this bug. This PR will significantly improve cache manager URL parsing code, on several levels.

This is a quick partial review.

src/mgr/QueryParams.cc
```diff
@@ -66,60 +69,59 @@ Mgr::QueryParams::find(const String& name) const
...
+    param.first = SBufToString(name);
+
+    std::vector<int> array;
+    int64_t intVal = 0;
+    while (tok.int64(intVal, 10, false)) {
+        static const CharacterSet comma("comma",",");
+        (void)tok.skipAll(comma);
+        Must(intVal >= std::numeric_limits<int>::min());
+        Must(intVal <= std::numeric_limits<int>::max());
+        array.emplace_back(int(intVal));
```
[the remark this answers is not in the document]
yadij: The code is presenting constructor parameter - which is an int. Thus the int() conversion.
The first suggestion is not correct because the constructor parameter for array<int> is not array<int64_t>.

src/mgr/QueryParams.cc
```diff
@@ -66,60 +69,59 @@ Mgr::QueryParams::find(const String& name) const
...
     }
-    regfree(&stringExpr);
-    regfree(&intExpr);
-    return parsed;
+
+    if (tok.atEnd())
+        param.second = new IntParam(array);
+    else
+        param.second = new StringParam(SBufToString(valueStr));
```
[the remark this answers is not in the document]
yadij: It looks odd, but this is intentional.

Strings that are lists of only integers are put into array. eg "param=999" or "param=1,2,3,4,999999"
 
Strings that are lists containing at least one member that is not an integer are to be treated as just strings. eg "param="foo" or "param=1,2,3,4,5,bar"

src/mgr/QueryParams.cc
```diff
@@ -66,60 +69,59 @@ Mgr::QueryParams::find(const String& name) const
...
+    static const CharacterSet nameChars = CharacterSet("param-name", "_") + CharacterSet::ALPHA + CharacterSet::DIGIT;
+
+    Parser::Tokenizer tok(paramStr);
+
+    SBuf name;
+    if (!tok.prefix(name, nameChars))
+        return false;
+
+    if (!tok.skip('='))
+        return false; // TODO: support name-only parameters
```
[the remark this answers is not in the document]
yadij: Removed when replacing the return with a throw.

src/mgr/QueryParams.cc
yadij: There is a subtle behaviour of tok.token() that is being relied on here to prune away sequences of suffix "&" delimiters without attempting to parse an empty parameter string between each one.

src/mgr/QueryParams.cc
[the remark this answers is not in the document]
rousskov: > The first suggestion is not correct because the constructor parameter for array<int> is not array<int64_t>.

I believe there is a misunderstanding of the problem I am trying to highlight and of what emplace methods are. I will rephrase my change request in hope to clarify the former and explain the latter.

`vector<T>::emplace_back(x)` is essentially `vector<T>::push_back(T(x))` that avoids creation of a temporary T(x) object in certain cases. It is primarily meant to be used in hope to optimize expensive T copies away (via placement new, of all things!), but it can also be used for convenience/readability sake to remove `T`, `(`, and `)` symbols from the function call.

A `vector<T>::emplace_back(T(x))` construct violates the very idea of an emplace API. The reader sees that the code explicitly disables optimizations provided by emplace and purposefully adds symbols made unnecessary by emplace. The reader is left to wonder what is really going on here -- why is this code doing all those things?!

In our specific case, the optimizations are not really important because we are talking about creation of integers -- a temporary on-stack integer costs nearly nothing. If anything, this optimization irrelveance only worsens the reader confusion.

There are many ways to address this problem, including the two snippets I offered in my change request:

```C++
array.emplace_back(intVal);
```

The reader may say: "Ah, OK, the code knows that `array` stores items of a different type than `intVal` and implies that the type conversion (from `intVal` type to the stored type, happening during the hidden stored object construction) is intentional. Good, moving on."

```C++
array.push_back(int(intVal));
```

The reader may say: "Ah, OK, the code knows that `array` stores `int` and tells me that the type conversion (from intVal to `int`, happening right in front of our eyes) is intentional here. Good, moving on."

The first solution is better if you want to write code that is easier to maintain/adjust -- that emplace_back() call will not need to be changed if the type of values stored in the container changes a little (if the type changes in incompatible ways, the compiler will catch the problem).

The second solution is better if you prefer to keep things simple/explicit even if it hurts code maintainability a little.

I am OK with either (or any better) solution in this particular case.

src/mgr/QueryParams.cc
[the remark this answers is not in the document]
rousskov: > It looks odd, but this is intentional.

Ah, I see now. Let's not create this trap for the next reader. If other code adjustments do not clarify the intent here, then please make that well-hidden intent explicit by adding a source code comment. FWIW, the logic you explained to me now was clear in the original if-we-can-parse-it-as-integers-it-is-integers code, but the introduction of a seemingly-independent integer consumption loop muddied the waters in the PR.

src/mgr/QueryParams.cc
rousskov: AFAICT, Tokenizer::token() does not work the way you think it does (I can explain why but I do not want to blow up this discussion).

```C++
    /*
     * \return true if found a non-empty token followed by a delimiter
     */
    bool token(SBuf &returnedToken, const CharacterSet &delimiters);
```

AFAICT, the current code does not handle the last parameters correctly -- it leaves it behind and then returns false. Glad my refactoring request led to that bug discovery. We can return to this refactoring after the bug is fixed; the final code may look a bit different from my original snippet.

BTW, I think it is up to us whether to ignore/skip or reject empty parameters. Safely ignoring them is probably overall better (for URL-generating scripts), but we do not have to be that flexible if being flexible is difficult. I suspect it is not difficult.

Please test before re-requesting review.

src/cache_manager.cc
```diff
@@ -157,73 +171,66 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
...
-    t = sscanf(url, "cache_object://%[^/]/%[^@?]%n@%[^?]?%s", host, request, &pos, password, params);
-    if (t < 3) {
-        t = sscanf(url, "cache_object://%[^/]/%[^?]%n?%s", host, request, &pos, params);
-    }
-    if (t < 1) {
-        t = sscanf(url, "http://%[^/]/squid-internal-mgr/%[^?]%n?%s", host, request, &pos, params);
-    }
-    if (t < 1) {
-        t = sscanf(url, "https://%[^/]/squid-internal-mgr/%[^?]%n?%s", host, request, &pos, params);
-    }
```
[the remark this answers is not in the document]
yadij: Done in 7aaeb80

src/mgr/QueryParams.h
```diff
@@ -13,7 +13,9 @@
 
 #include "ipc/forward.h"
 #include "mgr/QueryParam.h"
+#include "parser/Tokenizer.h"
```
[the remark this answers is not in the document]
yadij: Done in 09c5020

src/mgr/QueryParams.cc
```diff
@@ -66,60 +69,59 @@ Mgr::QueryParams::find(const String& name) const
 }
 
 bool
```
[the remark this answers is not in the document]
yadij: Done in f63dea0

src/mgr/QueryParams.cc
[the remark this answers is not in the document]
yadij: Hopefully this is clearer and resolved by acdf250d2

src/mgr/QueryParams.cc
yadij: Ooh, good catch. Fixed by acdf250d2

src/mgr/QueryParams.cc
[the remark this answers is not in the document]
yadij: Thank you. Not sure how I missed considering that factor.  Fixed in acdf250d2.

src/mgr/QueryParams.cc
```diff
@@ -66,60 +69,59 @@ Mgr::QueryParams::find(const String& name) const
...
+    if (!tok.prefix(name, nameChars))
+        return false;
+
+    if (!tok.skip('='))
+        return false; // TODO: support name-only parameters
+
+    static const CharacterSet badValueChars = CharacterSet("param-value-bad", "&= ");
+    SBuf valueStr = tok.remaining();
+    if (valueStr.findFirstOf(badValueChars) != SBuf::npos)
+        return false;
```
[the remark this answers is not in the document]
yadij: Unnecessary after acdf250d2.

src/cache_manager.cc
```diff
@@ -157,73 +171,66 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
  \retval NULL if the action can't be found or can't be accessed by the user
```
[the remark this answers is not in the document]
yadij: Done as of acdf250d2.

src/mgr/QueryParams.cc
```diff
@@ -66,60 +69,59 @@ Mgr::QueryParams::find(const String& name) const
 }
 
 bool
-Mgr::QueryParams::ParseParam(const String& paramStr, Param& param)
+Mgr::QueryParams::ParseParam(SBuf &paramStr, Param &param)
 {
-    bool parsed = false;
-    regmatch_t pmatch[3];
-    regex_t intExpr;
-    regcomp(&intExpr, "^([a-z][a-z0-9_]*)=([0-9]+((,[0-9]+))*)$", REG_EXTENDED | REG_ICASE);
```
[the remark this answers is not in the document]
yadij: Done in acdf250d2

src/cache_manager.cc
```diff
@@ -157,73 +171,66 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
...
+
+    // TODO: fix when AnyP::Uri::parse() separates path?query#fragment
+    SBuf params;
+    if (tok.skip('?')) {
+        params = tok.remaining();
+        if (!Mgr::QueryParams::Parse(tok, cmd->params.queryParams))
+            return nullptr;
+    }
+
+    debugs(16, 3, "MGR request: host=" << uri.host() << ", action=" << action <<
```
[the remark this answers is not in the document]
yadij: Please note that any trailing garbage in the query-string is detected as a parse error by the QueryParams logic. Before acdf250 refactor a false result would come from QueryParams::Parse() and trigger the error page response. Now then an exception will happen and be handled by the catch statement following this debugs statement.

The f63dea0 change adds support for ignoring fragment trailers explicitly.

src/cache_manager.cc
```diff
@@ -157,73 +171,66 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
...
     }
-#endif
 
-    debugs(16, 3, HERE << "MGR request: t=" << t << ", host='" << host << "', request='" << request << "', pos=" << pos <<
-           ", password='" << password << "', params='" << params << "'");
+    Mgr::Command::Pointer cmd = new Mgr::Command;
+    cmd->params.httpUri = SBufToString(uri.absolute());
+    cmd->params.userName = String();
+
+    const auto mgrDelimiters = MgrPathDelimiters(uri.getScheme());
```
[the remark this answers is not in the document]
yadij: Done in 6eea6921f

src/cache_manager.cc
```diff
@@ -157,73 +171,66 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
...
+    const auto fieldChars = mgrDelimiters.complement("mgr-action");
```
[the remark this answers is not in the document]
yadij: Done in 6eea6921f

src/mgr/QueryParams.cc
```diff
@@ -68,60 +69,61 @@ Mgr::QueryParams::find(const String& name) const
...
+        if (!tok.skip('='))
+            throw TexcHere("missing parameter value");
+
+        SBuf valueStr;
+        if (!tok.prefix(valueStr, valueChars))
+            throw TexcHere("invalid character in parameter value");
+
+        if (tok.skip('=') || tok.skip(' '))
+            throw TexcHere("invalid character in parameter value");
+
```
rousskov, line 122: Unnecessary leftovers?

```suggestion
```

src/mgr/QueryParams.cc
```diff
@@ -68,60 +69,61 @@ Mgr::QueryParams::find(const String& name) const
...
-        param.second = new StringParam(SBufToString(valueStr));
-
-    return true;
+        return new Mgr::StringParam(SBufToString(valueStr));
 }
 
-bool
+/**
+ * Syntax:
+ *   query  = param *( '&' param ) [ '#' fragment ]
```
rousskov, line 97: AFAICT, we do not require any parameters to be there:

```suggestion
 *   query  = [ param *( '&' param ) ] [ '#' fragment ]
```

src/mgr/QueryParams.cc
```diff
@@ -68,60 +69,61 @@ Mgr::QueryParams::find(const String& name) const
...
+        if (!tok.prefix(valueStr, valueChars))
+            throw TexcHere("invalid character in parameter value");
+
+        if (tok.skip('=') || tok.skip(' '))
+            throw TexcHere("invalid character in parameter value");
+
+        const auto name = SBufToString(nameStr);
+        const auto value = ParseParamValue(valueStr);
+        aParams.params.emplace_back(name, value);
+    }
```
rousskov, line 126: AFAICT, the code does not consume `&` between parameters. If I am right, then there is something wrong with the tests used during this PR work -- they should have revealed this basic parsing problem. Roughly speaking:

- If we do not allow empty parameters, then exactly one `&` should be skipped at the end of every the loop iteration (if it is present).
- Otherwise, all `&` should be consumed in the beginning of every loop iteration (if they are present).

src/mgr/QueryParams.cc
rousskov, line 97: This query parameters-parsing function implies that the fragment is a part of the query but does not consume it like the other documented parts of the query. Since we are sharing the tokenizer with the caller, we are responsible for leaving the tokenizer in the right state upon successful function termination.

I see two ways to fix these problems:

1. Keep the fragment in the function description. Adjust the function implementation to consume the fragment. I suspect that this option should be _rejected_ because fragments may follow well-formed parameterless URLs (i.e. URLs without `?`) that this function does not see. Thus, our caller would have to consume fragments as well. That would needlessly duplicate code and will force us to add Tokenizer::skipEverything() or equivalent (see the note below). I see no need for such complications.
2. Remove the fragment section from the current function interface description. Do not consume `#` and the following fragment in this function. Let the caller handle the fragment, with or without the query. I would try implementing this option first.

----

A note for the code that will consume the fragment if it turns our that such code is needed (I hope it is not): AFAICT, Tokenizer does not have the right method for skipping everything. We should either add Tokenizer::skipEverything() OR, if you prefer, overload skipAll() by adding a second skipAll() method (this one without parameters). The skipEverything() option may be slightly better because it avoids overloading, but I am not sure.

rousskov (changes requested):
Another partial review to help you make progress faster. Please test this code!

src/mgr/QueryParams.cc
yadij: Done in 0dc57a18a

src/mgr/QueryParams.cc
yadij: No. These two characters being seen are an error. If we ignore them existing here, they would be treated as a new name by the next loop iteration.

src/mgr/QueryParams.cc
yadij: query should always have a value when received by this method. I was missing an atEnd() check.

yadij: Fixed in 5ed02d1e6

src/mgr/QueryParams.cc
yadij: Dang. Lost it in the re-write. Fixed now I think. Will re-test in a day or so.

src/cache_manager.cc
```diff
@@ -150,42 +151,47 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
...
- \retval NULL if the action can't be found or can't be accessed by the user
+ *
+ * Syntax:
+ *
+ *  scheme "://" authority [ '/squid-internal-mgr' ] path-absolute [ '@' unreserved ] '?' query-string
+ *
+ * see RFC 3986 for definitions of scheme, authority, path-absolute, query-string
+ *
+ \retval Mgr::Command object for the following handling
+ \retval nullptr if the action can't be found or can't be accessed by the user, or errors
```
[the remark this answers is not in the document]
yadij: This PR is scoped at changing the URL parser. I am not going to go as far as changing behaviour of the single caller which already expects and properly handles the nullptr as an error.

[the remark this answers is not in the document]
rousskov: > This PR is scoped at changing the URL parser.

This change request matches that scope perfectly: CacheManager::ParseUrl() API is a part of the "URL parser" that we are fixing here.

> I am not going to go as far as changing behaviour of the single caller

No caller behavior changes have been requested so far. Just moving the misplaced code from the parsing function to its caller, preserving the caller behavior, would be sufficient to address this change request.

> which already expects and properly handles the nullptr as an error.

The caller should use the parser API. The parser API should be fixed. The latter change is perfectly in scope. Trivial caller adjustments to keep in sync with API fixes cannot be refused on scoping grounds.

src/mgr/QueryParams.cc
rousskov: > No. These two characters being seen are an error.

Accepting my suggestion would not change that fact.

> If we ignore them existing here, they would be treated as a new name by the next loop iteration.

And that is exactly what should happen! They will be seen as a start of a new (invalid) parameter name. There is nothing special about these two invalid characters -- many characters cannot be a part of a parameter name (or value). To have a better chance at being correct, a hand-written parser must focus on correctly identifying all valid input. Double checking random invalid cases distracts from that focus and, sooner or later, leads to bugs.

At this point, the code has already parsed the parameter value. Do not second-guess that code. Move on to the next construct to be parsed.

yadij: The "next construct to be parsed" is one of the delimiter characters: 
 * '&' - next param
 * '#' - start fragment
 * '=' - bad baseN encoding terminator
 * SP - bad URI %-encoding
 * NUL - parse complete (aka atEnd())

rousskov: The "bad" items on the list do not exist in the grammar. In theory, one could defend their presence on the grounds of making error messages better or providing other human-oriented enhancements, but I urge you not to change the excuse to that one. With so many bugs, we should be focusing on the essentials instead of poorly arguing about often buggy bells and whistles (which will, in turn, need polishing, causing more arguments, etc.). Writing parsers by code review is painful enough. Let's not make it worse.

The good items on the list are indeed a part of the grammar, but they are a red herring for this change request. They are (or will be, once the code is fixed) handled. This change request is not asking to remove their handling. This change request is specific to an explicit test for two bad items that are not a part of the grammar.

src/cache_manager.cc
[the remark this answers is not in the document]
yadij: Done in b0a755e

src/mgr/QueryParams.cc
yadij: Done in 1d7af98

src/mgr/QueryParams.cc
yadij: Done in ca0de8c

src/mgr/QueryParams.cc
```diff
@@ -68,60 +69,61 @@ Mgr::QueryParams::find(const String& name) const
...
+    static const CharacterSet nameChars = CharacterSet("param-name", "_") + CharacterSet::ALPHA + CharacterSet::DIGIT;
+    static const CharacterSet valueChars = CharacterSet("param-value", "&= #").complement();
+
+    while (!tok.atEnd() && !tok.skip('#')) {
 
-    // success == no garbage in query-string
-    return tok.atEnd();
+        SBuf nameStr;
+        if (!tok.prefix(nameStr, nameChars))
+            throw TexcHere("invalid character in action name");
```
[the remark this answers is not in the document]
yadij: Done in e4602a493

src/mgr/QueryParams.cc
```diff
@@ -65,61 +69,69 @@ Mgr::QueryParams::find(const String& name) const
...
 }
 
-bool
-Mgr::QueryParams::ParseParam(const String& paramStr, Param& param)
+/// Parses the value part of a "param=value" URL section.
+/// Value can be a comma separated list of integers, or an opaque string.
+/// \note opaque string can be a list with any non-integer.
+///       For example; 1,2,3,4,5,z
+Mgr::QueryParam::Pointer
+ParseParamValue(SBuf &valueStr)
```
rousskov, line 77: Please keep the parameter constant and avoid `Str` suffix noise. I suggest:

```suggestion
ParseParamValue(const SBuf &rawValue)
```

I do not think this renaming increases the diff.

src/mgr/QueryParams.cc
```diff
@@ -65,61 +69,69 @@ Mgr::QueryParams::find(const String& name) const
...
+
+    while (!tok.atEnd()) {
+
+        // TODO: remove '# case when AnyP::Uri separates path?query#fragment properly
+        // #fragment handled by caller. Do not throw.
+        if (tok.buf()[0] == '#')
+            return;
+
+        if (tok.skipAll(delims))
+            continue;
```
rousskov, line 119: This code allows empty list items at the beginning of the list but not elsewhere (e.g., `&&c=3` is allowed but `a=1&&` is prohibited). We should be consistent -- either allow empty items everywhere, nowhere, or (with a nod to modern programming languages and URL-generating code) only allow one extra trailing `&`. I cannot quickly tell what the official code does here today, but if it is much easier (from code point of view) to disallow all empty items except for a single trailing one, they perhaps that is what we should do.

src/mgr/QueryParams.cc
```diff
@@ -65,61 +69,69 @@ Mgr::QueryParams::find(const String& name) const
...
-            }
-        }
-        if (n < len) {
-            if (!ParseParam(aParamsStr.substr(n, len), param))
-                return false;
-            aParams.params.push_back(param);
-        }
+    static const CharacterSet nameChars = CharacterSet("param-name", "_") + CharacterSet::ALPHA + CharacterSet::DIGIT;
+    static const CharacterSet valueChars = CharacterSet("param-value", "&= #").complement();
+    static const CharacterSet delims("param-delim", "&");
```
rousskov, line 109: For naming consistency and clarity sake (the variable cannot hold multichar delimiter tokens, only chars):
```suggestion
    static const CharacterSet delimChars("param-delim", "&");
```

src/tests/testCacheManager.cc
```diff
@@ -66,3 +75,86 @@ testCacheManager::testRegister()
...
+
+    unsigned caseNumber = 0;
+    unsigned success = 0;
+    for (const auto &scheme : validSchemes) {
+        mgrUrl.setScheme(scheme);
+
+        for (const auto *action : validPathActions) {
+
+            // all schemes except cache_object require "/squid-internal-mgr" path prefix
+            if (scheme != AnyP::PROTO_CACHE_OBJECT && strlen(action) <= 19)
```
rousskov, line 133: We should not force ourselves to manually add the required magic prefix to nearly every action, decreasing the number of useful/interesting test cases and increasing noise. Please refactor to add the magic prefix to all actions (expecting a failure where it is not allowed) and remove the magic prefix from hard-coded validPathActions values.

----

If this code survives, we should not disassociate this code from its dependencies (and intent) so much! Please add something like a MgrUrMagic or a similar constant and use it here (and elsewhere, as appropriate).

```suggestion
            if (scheme != AnyP::PROTO_CACHE_OBJECT && strncmp(action, MgrUrlMagic, strlen(MgrUrlMagic)) != 0)
```

src/tests/testCacheManager.cc
```diff
@@ -66,3 +75,86 @@ testCacheManager::testRegister()
...
+                        bits.append(action);
+                        bits.append(param);
+                        bits.append(frag);
+                        mgrUrl.path(bits);
+
+                        (void)mgr->ParseUrl(mgrUrl);
+                        ++success;
+                    } catch (...) {
+                        std::cerr << std::endl << "FAIL: " << mgrUrl << std::endl
+                                  << CurrentException << std::endl;
```
rousskov, line 153: Please avoid unqualified/unmarked/unprefixed exceptions. The resulting output can be confusing, and this source code may be very difficult to find when things go seriously wrong:

```suggestion
                                  << "    error: " << CurrentException << std::endl;
```

src/tests/testCacheManager.cc
```diff
@@ -66,3 +75,86 @@ testCacheManager::testRegister()
...
+            // all schemes except cache_object require "/squid-internal-mgr" path prefix
+            if (scheme != AnyP::PROTO_CACHE_OBJECT && strlen(action) <= 19)
+                continue;
+
+            for (const auto *param : validParams) {
+
+                for (const auto *frag : validFragments) {
+                    try {
+                        ++caseNumber;
+                        std::cerr << caseNumber << std::endl;
```
rousskov, line 141: Do we have to be so noisy in test cases? No average human would be able to convert this "case number" to a specific test case anyway. I suggest removing this line.

src/tests/testCacheManager.cc
```diff
@@ -66,3 +75,86 @@ testCacheManager::testRegister()
...
+                        bits.append(frag);
+                        mgrUrl.path(bits);
+
+                        (void)mgr->ParseUrl(mgrUrl);
+                        ++success;
+                    } catch (...) {
+                        std::cerr << std::endl << "FAIL: " << mgrUrl << std::endl
+                                  << CurrentException << std::endl;
+                    }
+                    CPPUNIT_ASSERT_EQUAL(caseNumber, success);
```
rousskov, line 155: I do not understand the intent behind this line. I could understand the desire to fail on the first error or the desire to fail after all possible errors/cases, but this does neither AFAICT. Please adjust the code to either explain the odd choice or, better, to match one of those natural intents.

src/tests/testCacheManager.cc
```diff
@@ -66,3 +75,86 @@ testCacheManager::testRegister()
...
+        "?&foo=bar",
+        "?&&&&foo=bar",
+        "?&foo=bar&",
+        "?&&&&foo=bar&&&&",
+        "?foo=?_weird?~`:[]stuff&bar=okay&&&&&&"
+    };
+
+    const std::vector<const char *> validFragments = {
+        "",
+        "#frag"
```
rousskov, line 122: Please do not skip the "most likely to be mishandled" lonely `#` case! I would also add a few tricky cases.

```suggestion
        "",
        "#",
        "##",
        "#?a=b",
        "#fragment"
```

src/tests/testCacheManager.cc
```diff
@@ -66,3 +75,86 @@ testCacheManager::testRegister()
...
+        "?foo=bar",
+        "?0123456789=bar",
+        "?foo=bar&",
+        "?foo=bar&&&&",
+        "?&foo=bar",
+        "?&&&&foo=bar",
+        "?&foo=bar&",
+        "?&&&&foo=bar&&&&",
+        "?foo=?_weird?~`:[]stuff&bar=okay&&&&&&"
+    };
```
rousskov, line 118: Since the parser attempts to parse parameter value lists, please add a few representative test cases for integer and non-integer lists, especially since that code has proven to be problematic already.

Please also add multi-parameter test cases. Ideally, they should be auto-generated (e.g., from hard-coded parameters) but that would require other improvements.

src/tests/testCacheManager.cc
```diff
@@ -66,3 +75,86 @@ testCacheManager::testRegister()
...
+                    } catch (...) {
+                        std::cerr << std::endl << "FAIL: " << mgrUrl << std::endl
+                                  << CurrentException << std::endl;
+                    }
+                    CPPUNIT_ASSERT_EQUAL(caseNumber, success);
+                }
+            }
+        }
+    }
+}
```
rousskov, line 160: Please add at least a few invalid hard-coded cases where we expect the parser to throw.

src/tests/testCacheManager.cc
```diff
@@ -17,11 +18,19 @@
 
 CPPUNIT_TEST_SUITE_REGISTRATION( testCacheManager );
 
+/// Provides test code access to CacheManager internal symbols
+class CacheManagerInternals : public CacheManager
+{
+public:
+    void ParseUrl(const AnyP::Uri &u) { CacheManager::ParseUrl(u); }
+};
+
```
rousskov, line 27: IMHO, it would be better to provide a CacheManager::TestParseUrl() public method than to add a wrapper class like this that will need to be maintained as CacheManager construction improves. Ideally, the unit test framework should provide a way to access private data members, but I am guessing that CPPUNIT is not that good.

I do not insist on any changes here.

rousskov (changes requested):
Thank you for addressing my previous concerns. I think we are getting close to the final version here.

src/mgr/QueryParams.cc
```diff
@@ -65,61 +69,69 @@ Mgr::QueryParams::find(const String& name) const
     return iter;
 }
 
-bool
-Mgr::QueryParams::ParseParam(const String& paramStr, Param& param)
+/// Parses the value part of a "param=value" URL section.
+/// Value can be a comma separated list of integers, or an opaque string.
+/// \note opaque string can be a list with any non-integer.
+///       For example; 1,2,3,4,5,z
```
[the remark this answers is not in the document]
yadij: I used the English formal form here for the note to explicitly clarify which of the text is the example and which is the "for example". The abbreviated form has a visual ambiguity about whether everything inside the brackets is the example string.

src/mgr/QueryParams.cc
```diff
@@ -65,61 +69,69 @@ Mgr::QueryParams::find(const String& name) const
...
+
+    Parser::Tokenizer tok(valueStr);
+    std::vector<int> array;
+    int64_t intVal = 0;
+    while (tok.int64(intVal, 10, false)) {
+        (void)tok.skipAll(comma);
+        Must(intVal >= std::numeric_limits<int>::min());
+        Must(intVal <= std::numeric_limits<int>::max());
+        array.emplace_back(intVal);
     }
```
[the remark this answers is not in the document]
yadij: Current code regex does not accept any ",," sequences or prefix or suffix ',' characters.

src/tests/testCacheManager.cc
yadij: IMO this wrapper class is better. It keeps the nasty hack business restricted to the test case which requires it and avoids polluting the real class with accessors that might be abused.

src/tests/testCacheManager.cc
yadij: Sure. It was there to assist debug of test writing. It can go now.

src/tests/testCacheManager.cc
yadij: CurrentException prints its own standard prefix. The qualification marking is the FAIL line immediately above and the name of the test it is running. The output to the test-suite log looked good to me, will post an example later so you can see.

src/cache_manager.cc
```diff
@@ -148,82 +150,87 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
...
+    // TODO: fix when AnyP::Uri::parse() separates path?query#fragment
+    SBuf params;
+    if (tok.skip('?')) {
+        params = tok.remaining();
+        Mgr::QueryParams::Parse(tok, cmd->params.queryParams);
+    }
+
+    if (!tok.skip('#') && !tok.atEnd())
+        throw TextException("invalid characters in URL", Here());
+    // else ignore #fragment
```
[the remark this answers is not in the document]
yadij: The word "trailing" implies that the end of the URL is where the problem is. When in fact it may be the first characters in the URL which are the problem. All we know is that *a* character somewhere in the section being parsed is invalid.

e.g.    "cache_object://localhost/? foo0001=bar&foo0002=bar" ... skip ~8KB ... "&foo8000=bar#the-end"

src/tests/testCacheManager.cc
yadij: Will do.

But please be aware that there are multiple magic prefix possible. cache_object:// only allowed "/" prefix. The current code uses "/squid-internal-mgr/". I also expect that we will need to register and move to a "/.well-known/" label since the Squid internal magic prefix are embedded in other peoples domains sometimes.

src/tests/testCacheManager.cc
yadij: It exists for cppunit tracking of tests run. It does halt on the first error - by means of any parser thrown exceptions skipping the ++success code statement.

src/mgr/QueryParams.cc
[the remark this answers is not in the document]
rousskov: > I used the English formal form here

FWIW, I have not seen the form you used (i.e. `For Example; ...`) used in this corner of the world. I could not find any examples of it on the web either, but it is difficult to google punctuation, of course. That form also looks very odd to me because semicolons usually separate independent thoughts/parts from each other rather than delimit labels from the information they label or introductory phrases from the primary ones.

> The abbreviated form has a visual ambiguity about whether everything inside the brackets is the example string.

I do not see any ambiguity -- the suggested parens are a very common format for inlined examples and "that is" clarifications -- but you can surround `1,2,z` with quotes if you think it will help more than it will hurt (i.e. `"1,2,z"`).

src/tests/testCacheManager.cc
rousskov: > It does halt on the first error

Please adjust the code to do that explicitly rather than to rely on the numerical equality between apples (the current case ID) and oranges (the number of cases that did not throw in the past), triggering reader doubts (at best). AFAICT, we should move the CppUnit assert (or equivalent) to the catch clause. If done right, this move would also improve failure diagnostic, possibly significantly.

src/tests/testCacheManager.cc
rousskov: > CurrentException prints its own standard prefix

It does not. And it should not. CurrentException does not know its reporting context so it should not supply any prefix. It is the caller responsibility, and that responsibility is neglected in this code.

> The qualification marking is the FAIL line immediately above

Yes, but the "FAIL" line does not help in filtered results (such as those often pasted in support emails), especially when other test cases produce misleading FAIL lines, and especially since this output does not follow the emerging multiline standard introduced by Debug::Extra -- there will be no visual clue that something important was filtered out. Visually, the reported part of the log will just contain a seemingly random message on its own line. When things go wrong, that line may cause additional headaches. The minimal fix is trivial and is provided. We should not spend time discussing why it should be avoided!

src/mgr/QueryParams.cc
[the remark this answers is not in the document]
yadij: Done in 239abba

src/cache_manager.cc
```diff
@@ -474,4 +485,3 @@ CacheManager::GetInstance()
     }
     return instance;
 }
-
```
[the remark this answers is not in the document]
yadij: Done in fd5507e

src/tests/testCacheManager.cc
yadij: Done (I think) in 351e7ba56.

src/tests/testCacheManager.cc
yadij: Oh you are right. I was thinking of the "exception location:" prefix.

Failure display updated in 351e7ba56. On failures it will now produce something like:
```
...
FAIL: cache_object://localhost/?foo= bar
    error: missing or malformed parameter value
    exception location: src/mgr/QueryParams.cc(129) Parse
F
testCacheManager.cc:157:Assertion
Test name: testCacheManager::testParseUrl
equality assertion failed
- Expected: 21
- Actual  : 20
```

src/tests/testCacheManager.cc
yadij: Removed in 351e7ba.

src/tests/testCacheManager.cc
yadij: Done in 77074c660.

src/cache_manager.cc
```diff
@@ -306,10 +313,14 @@ CacheManager::CheckPassword(const Mgr::Command &cmd)
...
-    if (!cmd) {
+    Mgr::Command::Pointer cmd;
+    try {
+        cmd = ParseUrl(request->url);
+
+    } catch (...) {
+        debugs(16, 2, "MGR request error: " << CurrentException);
         const auto err = new ErrorState(ERR_INVALID_URL, Http::scNotFound, request, ale);
         err->url = xstrdup(entry->url());
         errorAppendEntry(entry, err);
```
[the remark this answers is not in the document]
yadij: Done in 858f2f98d

src/tests/testCacheManager.cc
yadij: Done in cd5118c4d

src/tests/testCacheManager.cc
yadij: Done in 336276e69.

src/cache_manager.cc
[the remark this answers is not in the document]
rousskov: > The word "trailing" implies that the end of the URL is where the problem is.

Not exactly. "Trailing" means we could not parse something after some valid URL portion, which is true even in your `cache_object://localhost/? garbage...` example.

At any rate, the word "trailing" is not critical here. You can use a different word if you find a better one or, if you insist, no word at all.

src/mgr/QueryParams.cc
```diff
@@ -65,61 +69,69 @@ Mgr::QueryParams::find(const String& name) const
...
+    static const CharacterSet comma("comma", ",");
+
+    Parser::Tokenizer tok(rawValue);
+    std::vector<int> array;
+    int64_t intVal = 0;
+    while (tok.int64(intVal, 10, false)) {
+        Must(intVal >= std::numeric_limits<int>::min());
+        Must(intVal <= std::numeric_limits<int>::max());
+        array.emplace_back(intVal);
+        if (tok.remaining().length() > 1)
```
rousskov, line 87: Why do we only want to consume a comma if there are at least two characters left. If this strange condition is not needed, let's remove it. If it is needed, then please add a source code comment to explain what is going on.

```suggestion
```

src/mgr/QueryParams.cc
```diff
@@ -65,61 +69,69 @@ Mgr::QueryParams::find(const String& name) const
     return iter;
 }
 
-bool
-Mgr::QueryParams::ParseParam(const String& paramStr, Param& param)
+/// Parses the value part of a "param=value" URL section.
+/// Value can be a comma-separated list of integers or an opaque string.
+/// \note opaque string may be a list with a non-integer (e.g., "1,2,3,z")
+Mgr::QueryParam::Pointer
+ParseParamValue(SBuf &rawValue)
```
rousskov, line 76: Please keep the parameter constant if possible.

```suggestion
ParseParamValue(const SBuf &rawValue)
```

src/tests/testCacheManager.cc
```diff
@@ -66,3 +75,150 @@ testCacheManager::testRegister()
...
+                            std::cerr << std::endl
+                                      << "FAIL: " << mgrUrl
+                                      << Debug::Extra << "error: should be rejected due to '" << param << "'" << std::endl;
+                            CPPUNIT_FAIL(err);
+                        } catch (const TextException &e) {
+                            CPPUNIT_ASSERT(++success);
+                            if (strcmp(e.what(), err) == 0)
+                                throw;
+                        }
```
rousskov, line 218: I do not understand the logic here because it is not clear where `accepted malformed input` (i.e. `err`) value might come from. What are we trying to detect/test by using `err`? If these are some manual testing leftovers, then we can remove this to simplify:

```
                        } catch (...) {
                            ... report CurrentException here, just in case it is not really what we expect ...
                            CPPUNIT_ASSERT(true); // or perhaps there is CPPUNIT_SUCCESS()?
                        }
```

The above sketch also prints the caught exception. It is never a good idea to hide exceptions, but I am not sure whether CppUnit supports some kind of debugging mode where they can be reported without polluting regular output too much. On the other hand, the current "make check" output is already so polluted that perhaps I should not be worried about additional pollution at this point.

Also, we should not be counting successes if we are not using them anywhere. KISS.

src/tests/testCacheManager.cc
```diff
@@ -66,3 +75,150 @@ testCacheManager::testRegister()
...
+                        try {
+                            SBuf bits;
+                            bits.append(magic);
+                            bits.append(action);
+                            bits.append(param);
+                            bits.append(frag);
+                            mgrUrl.path(bits);
+
+                            (void)mgr->ParseUrl(mgrUrl);
+                            CPPUNIT_ASSERT(++success);
```
rousskov, line 183: It is not clear why this assert inside a loop is needed. And there is another assert like that below. Remove both?

rousskov (changes requested):
Thank you for addressing most of my concerns. I did not do a full review this time, and I hope this is the last iteration. Only one or two potentially important issues remain.

src/tests/testCacheManager.cc
yadij: These are for cppuinit - to mark a check being done.

rousskov: > These are for cppuinit - to mark a check being done.

I do not understand what you mean by "to mark a check being done". The CppUnit [documentation](http://cppunit.sourceforge.net/doc/cvs/group___assertions.html) does not seem to imply that CPPUNIT_ASSERT() has some "marking" side effects (or any side effects), but that documentation is rather ... limited, so perhaps there is more to this assertion than meets the eye.

What would change in test logic and/or output if this CPPUNIT_ASSERT() line (with an never-failing condition) is removed?

src/mgr/QueryParams.cc
yadij: Now that we are forbidding empty list entries this is needed to prevent the list ending on a comma.

src/tests/testCacheManager.cc
yadij: cppunit outputs a '.' character to visually mark successful checks in the logs and keep-alive any shells, network I/O and similar. These are useful to identify that the tested code is actually completing. It can take a long time on slow machines or environments running the processing at very low priority levels.

src/tests/testCacheManager.cc
yadij: CppUnit uses exceptions. Its "ASSERT" macros and such actually are just throwing. Only exceptions which reach the test harness are handled as errors in the code.

So ..
 a) when our code properly throws an exception for the invalid input we need to swallow it to prevent cppunit halting the build. That is what the explicit try-catch is for.

 b) when our code wrongly accepts an invalid input we need to trigger a cppunit exception. That is CPPUNIT_FAIL().
 However our try-catch also swallows these. So we need to identify and re-throw. Use of a static err string allows that efficiently without involving memory management risks.

src/mgr/QueryParams.cc
rousskov: IMHO, it is better to allow a single trailing comma, like many modern languages do, than to complicate code and add source code comments to explain this condition. As an added bonus, allowing a trailing comma will simplify generation of cache manager URLs. Your call though. I see three valid options here:

1. Allow a single trailing comma: Consume the comma unconditionally, perhaps with a trailing `// allow a trailing comma` comment.
2. Ban trailing commas, with an explanation: Keep this condition. Add a source code comment to explain what is going on here.
3. Ban trailing commas, without an explanation: Refactor the code to make it self-documenting. Remove the condition and save skipOne(comma) result in a new sawCommaAtTheEnd boolean variable, to be used together with the atEnd() condition below.

src/tests/testCacheManager.cc
rousskov: > cppunit outputs a '.' character to visually mark successful checks

AFAICT, CPPUNIT_ASSERT() has nothing to do with progress reporting: CppUnit logs `.` at the beginning of each _test_, not after successful assertions. I cannot find any evidence that removing these always-true CPPUNIT_ASSERT()s will harm progress reporting (or anything else). Can you?

```C++
define CPPUNIT_ASSERT(condition)                                                 \
  ( CPPUNIT_NS::Asserter::failIf( !(condition),                                   \
                                 CPPUNIT_NS::Message( "assertion failed",         \
                                                      "Expression: " #condition), \
                                 CPPUNIT_SOURCELINE() ) )
```

```C++
void 
Asserter::failIf( bool shouldFail, 
                  const Message &message, 
                  const SourceLine &sourceLine )
{
  if ( shouldFail )
    fail( message, sourceLine );
}
```

```
$ git grep -F2 '"."' src
src/cppunit/TextTestProgressListener.cpp-TextTestProgressListener::startTest( Test * )
src/cppunit/TextTestProgressListener.cpp-{
src/cppunit/TextTestProgressListener.cpp:  stdCOut() << ".";
src/cppunit/TextTestProgressListener.cpp-  stdCOut().flush();
src/cppunit/TextTestProgressListener.cpp-}
--
src/cppunit/TextTestResult.cpp-{
src/cppunit/TextTestResult.cpp-  TestResultCollector::startTest (test);
src/cppunit/TextTestResult.cpp:  stdCOut() << ".";
src/cppunit/TextTestResult.cpp-}
src/cppunit/TextTestResult.cpp-
```

rousskov (changes requested):
I just noticed that you have requested my review. There are several open change requests from my previous review. I am updating one of them here to formally bounce the ball to your side, but please do not ignore the other ones.

src/cache_manager.cc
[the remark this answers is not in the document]
yadij: Done in 1000961

src/mgr/QueryParams.cc
```diff
@@ -66,60 +69,59 @@ Mgr::QueryParams::find(const String& name) const
...
+
+    param.first = SBufToString(name);
+
+    std::vector<int> array;
+    int64_t intVal = 0;
+    while (tok.int64(intVal, 10, false)) {
+        static const CharacterSet comma("comma",",");
+        (void)tok.skipAll(comma);
+        Must(intVal >= std::numeric_limits<int>::min());
+        Must(intVal <= std::numeric_limits<int>::max());
```
MegaManSec, line 85: Please consider whether this is correct. Without doing additional tests, this causes an exception in some testing, with the following request: https://paste.pr0.tips/Xr?c

rousskov: > this causes an exception in some testing

This code should throw when the raw value cannot be represented as `int`. While I do not recommend using Must() for input validation anymore, I do not see the problem you are alluding to.

MegaManSec: While the patch I am using it outdated (it is from the first, original patch), I believe the issue is there will be a memory leak if this exception happens. At least in my testing which includes the original patch, a memory leak happens.

I cannot test with the current patch because it does not cleanly patch, and I do not have the time to manually do it.

rousskov: > there will be a memory leak if this exception happens

Aha, _that_ is a bug! In summary: The leak (if any) is a bug. The exception is not a bug. Using Must() for input validation is currently still acceptable, but throwing an explicit exception or providing/using an API dedicated to input validation is where we should be heading long-term.

It is trivial to trigger this exception, so Amos would be able to test whether the current PR code still leaks. Thank you for reminding us about that danger!

MegaManSec: Ah, sorry; I was still looking at this bug through the lens of "an exception may occur which will may trigger a memory leak", rather than anything specific to the exception itself. I am glad it is clear now.

yadij: AFAIK none of the data types used for this new parser lead to memory leaks. Our types here have all been designed for exception safe memory handling, others are used as locals to avoid leaks the old allocations had.

src/mgr/QueryParams.cc
yadij: Im going with not changing the API. 
a26458a documents the pattern being parsed. I hope that is clear enough now.

src/tests/testCacheManager.cc
yadij: Since you are pushing this 4a1b8de removes the ++success asserts for now. They can be added back if/when actually needed.

src/mgr/QueryParams.cc
MegaManSec: I will reserve my patches for when this is pushed to master (alternatively, simply send the request to Squid (assuming cachemanager is open) which I posted and check for leaks).

rousskov: The new ParseUrl() sequence _looks_ exception-safe to me, but I agree that this impression should be tested, especially since (minimal) tests are easy to perform in this case.

src/mgr/QueryParams.cc
rousskov: > I hope that is clear enough now.

Unfortunately, it is not clear enough to me. The added comment does not explain the strange condition. In my experience, it is conditions like this that sooner or later break "working" parsing code or make it difficult to maintain.

src/tests/testCacheManager.cc
rousskov: Just to avoid misunderstanding: This change request remains unaddressed. Mys last https://github.com/squid-cache/squid/pull/788#discussion_r608733550 enumerates two simple specific solutions that I see.

yadij: Please note the code as at a24c52a passes semaphore and jenkins builds claiming SUCCESS.

The a24c52a code contains an intentional valid URL in the set intended to prove parser throws on bad cases. On this false-positive the parser will return normally and CPPUNIT_FAIL() throw a char* type of exception. This demonstrates that "catch (const TextException &)" is not sufficient to let cppunit thrown exceptions actually reach the cppunit handlers.

rousskov: > "catch (const TextException &)" is not sufficient to let cppunit thrown exceptions actually reach the cppunit handlers.

I cannot explain that alleged outcome without research. The above conclusion matches neither my C++ understanding nor my quick [test](https://cpp.godbolt.org/z/fhz63orfr), but there are several unknowns here that may lead to misleading conclusions. Even though this mystery is not related to any of the two solutions I have [proposed](https://github.com/squid-cache/squid/pull/788#discussion_r608733550) on this thread, I will study it to ensure I am not missing something for future work.

yadij: Your second proposal has worked AFAICT. So going with that.

src/mgr/QueryParams.cc
yadij: How about the 4451407 texts?

src/mgr/QueryParams.cc
yadij: Basic valgrind testing shows no parser leak for the testCacheManager binary which throws on the bad-input tests.

MegaManSec: FTR, the memory leak I am seeing with the original 3-patches is apparently happening here: https://paste.pr0.tips/3K
with the logs:
http://paste.pr0.tips/qg

I have not tested with the newest code yet.

MegaManSec: The memory leak goes away with the current full patch. Thanks.

src/mgr/QueryParams.cc
rousskov: The primary problem -- strange parsing conditions often lead to bugs -- remains, of course. The secondary problem -- it is difficult to understand what the code is trying to do -- can be considered addressed, albeit in a rather tortured/convoluted (IMO) and sloppy way. I do not have enough time to properly address these non-blocking problems myself, so I will resolve this change request.

src/tests/testCacheManager.cc
```diff
@@ -208,11 +208,10 @@ testCacheManager::testParseUrl()
                             std::cerr << std::endl
                                       << "FAIL: " << mgrUrl
                                       << Debug::Extra << "error: should be rejected due to '" << param << "'" << std::endl;
-                            CPPUNIT_FAIL(err);
                         } catch (const TextException &e) {
-                            if (strcmp(e.what(), err) == 0)
-                                throw;
+                            continue; // success. caught bad input
                         }
+                        CPPUNIT_FAIL("bad result");
```
rousskov, line 214: Let's be more specific to help developers figure out what went wrong if this test case fails:

```suggestion
                        CPPUNIT_FAIL("failed to reject an invalid URL");
```

src/cache_manager.cc
```diff
@@ -148,82 +151,87 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
...
-    cmd->params.actionName = request;
+    // TODO: fix when AnyP::Uri::parse() separates path?query#fragment
+    SBuf params;
+    if (tok.skip('?')) {
+        params = tok.remaining();
+        Mgr::QueryParams::Parse(tok, cmd->params.queryParams);
+    }
+
+    if (!tok.skip('#') && !tok.atEnd())
+        throw TextException("invalid characters  in URL", Here());
```
rousskov, line 229: Minor typo:

```suggestion
        throw TextException("invalid characters in URL", Here());
```

rousskov (commented):
Thank you for addressing my blocking concerns. I found a couple of places that can be improved, but I am not aware of any remaining blocking issues. I lack the time to do a full re-review.

yadij:
Thank you. I think the parser was covered by earlier review and just unit tests recently. So marking for merge.

yadij:
@eduard, @kinkie can you check why Anubis is not completing this? I suspect the test count is being an issue again since Fedora changes in Jenkins.

eduard-bagdasaryan:
>  I suspect the test count is being an issue again since Fedora changes in Jenkins

I fixed the test number in Anubis.

PR #789, opened 2021-03-16
Bug 5112: Excessively loud chunked reply parsing error reporting

rousskov:
Traffic parsing errors should be reported at level 2 (or below) because
Squid admins can usually do nothing about them and a noisy cache.log
hides important problems that they can and should do something about.

TODO: Detail this and similar parsing errors for %err_detail logging.

Also removed an unnecessary used-once macro.

### DIFF
--- a/src/http.cc
+++ b/src/http.cc
@@ -67,15 +67,6 @@
 #include "DelayPools.h"
 #endif
 
-#define SQUID_ENTER_THROWING_CODE() try {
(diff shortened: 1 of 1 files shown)


rousskov:
@MegaManSec, please test this polishing fix if you can.

MegaManSec:
LGTM

yadij (approved):

PR #792, opened 2021-03-31
Check for duplicated http_port entries in config files

hfisiteimprove:
This fixes a random occurring - but very annoying bug. 

We have a squid cluster with high load in our company (8000+ connects). From time to time we noticed a very annoying bug.

ACL roules were not taken, connections didn't bump to SSL. That happens only with very high load and only randomly on very few of the connects.

After a two week debug session, we found out, that it is caused by multiple PortCfg entries. That happened by our configuration files, where we found, that we had multiple entries for http_ports

```
http_port 3128
...
http_port 3128 ssl-bump ..
```

According to our knowledge and to the official squid docs - http://www.squid-cache.org/Doc/config/http_port/ - it makes no sense to give the user this level of freedom.

The desired level of outcome:

- avoid conflicting configuration (e.g. having ssl-bump and no ssl-bump configured on the same port)
- just have one "type" per  port and protocol 

This was a really annoying issue. Hard to find and we developed tools to reproduce high load in a lab environment.

### DIFF
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -921,6 +921,25 @@ configDoConfigure(void)
     }
 
     for (AnyP::PortCfgPointer s = HttpPortList; s != NULL; s = s->next) {
+        for (AnyP::PortCfgPointer s2 = HttpPortList; s2 != NULL; s2 = s2->next) {
+            if( s == s2)
+                continue;
+            if (s->name && s2->name && strcmp(s->name, s2->name) != 0)
+                continue;
+            if (s->defaultsite && s2->defaultsite && strcmp(s->defaultsite, s2->defaultsite) != 0)
+                continue;
+            if (s->transport.protocol == s2->transport.protocol &&
+                s->transport.major == s2->transport.major &&
+                s->transport.minor == s2->transport.minor) {
+                debugs(3, DBG_CRITICAL, "ERROR: multiple definitions for http_port: " <<
+                        s->name << ' ' << s->transport.protocol << ' ' <<  s->transport.major << '.' <<  s->transport.minor );
+                self_destruct();
+                return;
+            }
+        }
+    }
+
+    for (AnyP::PortCfgPointer s = HttpPortList; s != NULL; s = s->next) {
         if (!s->secure.encryptTransport)
             continue;
         debugs(3, DBG_IMPORTANT, "Initializing " << AnyP::UriScheme(s->transport.protocol) << "_port " << s->s << " TLS contexts");
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -920,19 +920,15 @@ configDoConfigure(void)
         }
     }
 
+    // check for duplicated entries, e.g. we won't have duplicated items like this:
+    // http_port 3128
+    // http_port 3128 ssl-bump ...
     for (AnyP::PortCfgPointer s = HttpPortList; s != NULL; s = s->next) {
-        for (AnyP::PortCfgPointer s2 = HttpPortList; s2 != NULL; s2 = s2->next) {
-            if( s == s2)
-                continue;
+        for (AnyP::PortCfgPointer s2 = s->next; s2 != NULL; s2 = s2->next) {
             if (s->name && s2->name && strcmp(s->name, s2->name) != 0)
                 continue;
-            if (s->defaultsite && s2->defaultsite && strcmp(s->defaultsite, s2->defaultsite) != 0)
-                continue;
-            if (s->transport.protocol == s2->transport.protocol &&
-                s->transport.major == s2->transport.major &&
-                s->transport.minor == s2->transport.minor) {
-                debugs(3, DBG_CRITICAL, "ERROR: multiple definitions for http_port: " <<
-                        s->name << ' ' << s->transport.protocol << ' ' <<  s->transport.major << '.' <<  s->transport.minor );
+            if (s->transport.protocol == s2->transport.protocol) {
+                debugs(3, DBG_CRITICAL, "ERROR: multiple definitions for http_port: " << s->name << ' ' << s->transport.protocol);
                 self_destruct();
                 return;
             }


yadij (changes requested):
It would be better to perform this duplication check in parsePortSpecification() along with the other validity checks for the listening port. That function should be adjusted slightly to receive the "head" Pointer with the list of existing ports to check against.

As the internals of parsePortSpecification() should make it obvious the s->name is not relevant. It can be changed by the name= option later to any random string, which in turn would make the debugs() display wrong. Likewise s->transport.protocol is the value of the protocol= option so can also be changed later.

What is important is the IP:port combo. It is only possible to open one listening socket on each unique combination - with some special cases:
 1) special IP 0.0.0.0 conflicts with any IPv4 on the same port number.
 2) special IP :: conflicts with any IPv6 on the same port number.
 3) on TCP split-stack machines the IPv4/IPv6 also matters in addition to IP:port.

rousskov:
@hfisiteimprove, thank you very much for working on this useful improvement! We do want to reject configurations with conflicting port configurations.

I share the _primary_ concern [voiced](https://github.com/squid-cache/squid/pull/792#pullrequestreview-625363563) by Amos -- this PR should check for conflicting listening addresses rather than port names or protocols. Please refactor accordingly.

BTW, rejecting duplicated `*_port name=...` values is also a good idea, but that is not what this PR focus and primary value are.

> Amos: 1. special IP 0.0.0.0 conflicts with any IPv4 on the same port number

I assume the following configuration should accepted as valid in some environments. @yadij, please correct me if I am wrong.

```
# OK: Special http_port for local cache manager queries only
http_port 127.0.0.1:3128 ...
# OK: We listen  for traffic on all other available IP addresses, whatever they are
http_port 0.0.0.0:3128 ...
```

If my assumption is correct, then listening conflicts between specific addresses and wildcards should be out of this PR scope because many of them cannot be reliably detected at configuration time without a lot of complex (and unnecessary IMO) work. They should be detected at runtime, resulting in Squid startup death, but that is a different issue.

----

P.S. @hfisiteimprove, please be extra careful with the Ip::Address APIs. A few of them are badly designed and will confuse uninitiated developers. For example, do not assume that a C++ equality operator tests for Ip::Address object equality. Always check the actual method implementation before relying on its expected or even documented semantics.

rousskov:
I polished PR title/description to fix formatting and to document the primary effect of the change, but they may need more work depending on how we end up defining "conflicting" for this PR purposes.

src/cache_cf.cc
```diff
@@ -920,6 +920,21 @@ configDoConfigure(void)
...
+    // check for duplicated entries, e.g. we won't have duplicated items like this:
+    // http_port 3128
+    // http_port 3128 ssl-bump ...
+    for (AnyP::PortCfgPointer s = HttpPortList; s != NULL; s = s->next) {
+        for (AnyP::PortCfgPointer s2 = s->next; s2 != NULL; s2 = s2->next) {
+            if (s->name && s2->name && strcmp(s->name, s2->name) != 0)
+                continue;
+            if (s->transport.protocol == s2->transport.protocol) {
+                debugs(3, DBG_CRITICAL, "ERROR: multiple definitions for http_port: " << s->name << ' ' << s->transport.protocol);
+                self_destruct();
```
rousskov, line 932: Please throw (the error) on configuration errors instead of calling the deprecated self_destruct(). Bad reconfigurations should be rejected instead of killing Squid, and using exceptions helps achieve that long-term goal.

For error handling examples, see `git grep throw src/cache_cf.cc` but throw a TextException object instead of using the deprecated TexcHere() macro. You can find many examples of thrown TextExceptions in the code as well.

rousskov (changes requested):

yadij:
> ```
> # OK: Special http_port for local cache manager queries only
> http_port 127.0.0.1:3128 ...
> # OK: We listen  for traffic on all other available IP addresses, whatever they are
> http_port 0.0.0.0:3128 ...
> ```

Hmm. Its a long twisted path of indirect flag settings, but it looks like this *should* be working for most Squid builds. However the existence of this bug 4261 and many user complaints indicates that something is not happening like it should.

To open a port like that requires that the explicit-IP:port listener be opened and sets SO_REUSEADDR before the wildcard-IP:port is opened.

Doing a deep code inspection:
 * comm_open_listener() calls comm_openex() with the configured port number.
 * comm_openx() calls commSetReuseAddr() when flag COMM_REUSEADDR set *or* given a port number.
 * commSetReuseAddr() uses opt_reuseaddr to decide whether to set the flag or do nothing.
 * opt_reuseaddr is enabled by default globally, except when "squid -R" is used on the command line.

rousskov:
IMO, it does not matter (for this PR) whether sharing specific and wildcard addresses works today (including "works well", "does not work at all", "only works on Windows", etc.). This PR is about rejecting things that can never ever work and making that decision based on basic configuration analysis alone. In other words, roughly speaking, this is a "configuration validation" PR rather than a "listening code improvements" or "deep analysis of socket flags and bind() APIs" PR.

According to my quick script tests, recollection, and googling:
* specific/wildcard address sharing that I described [works](https://stackoverflow.com/a/33622121) (without any special flags) on modern Windows;
* specific/wildcard address sharing that I described does not work (with or without SO_REUSEADDR) on Unix
* SO_REUSEADDR is [irrelevant](https://blog.cloudflare.com/its-crowded-in-here/#quiz-3) (on Unix) -- it is meant for a different kind of reuse; you may have been thinking about SO_REUSEPORT

However, again, all these low-level details do not matter here IMO -- we do not need to spend time developing consensus regarding them. Comparing IPs:port addresses is what we want AFAICT.

```
# OK: Special http_port for local cache manager queries only
http_port 127.0.0.1:3128 ...
# OK: We listen  for traffic on all other available IP addresses, whatever they are
http_port 0.0.0.0:3128 ...
# Error: two http_port definitions with the same IP:port address (0.0.0.0)
http_port 3128 ...
```

yadij:
> IMO, it does not matter (for this PR) whether sharing specific and wildcard addresses works today (including "works well", "does not work at all", "only works on Windows", etc.). This PR is about rejecting things that can never ever work and making that decision based on basic configuration analysis alone.

IMO it matters because we do get user requests to configure Squid that way and the use-case seems reasonable at face-value. If it was at some point *supposed* to work we should not arbitrarily forbid it here without at least allowing for a future fix.

> In other words, roughly speaking, this is a "configuration validation" PR rather than a "listening code improvements" or "deep analysis of socket flags and bind() APIs" PR.
> 

Agreed on that being the scope of changes. 

> According to my quick script tests, recollection, and googling:
> 
>     * specific/wildcard address sharing that I described [works](https://stackoverflow.com/a/33622121) (without any special flags) on modern Windows;
> 

Then your second config line below should be "OK" on Windows (native?) builds ...


>     * specific/wildcard address sharing that I described does not work (with or without SO_REUSEADDR) on Unix
> 

... and "Error" on all other builds.

>     * SO_REUSEADDR is [irrelevant](https://blog.cloudflare.com/its-crowded-in-here/#quiz-3) (on Unix) -- it is meant for a different kind of reuse; you may have been thinking about SO_REUSEPORT
> 

No the stackexchange discussion and references I found the other day were definitely only about SO_REUSEADDR. I cannot find them again today to check if listen() was part of it.

> 
> However, again, all these low-level details do not matter here IMO -- we do not need to spend time developing consensus regarding them. Comparing IPs:port addresses is what we want AFAICT.
> 
> ```
> # OK: Special http_port for local cache manager queries only
> http_port 127.0.0.1:3128 ...
> # OK: Trying to listen for traffic on all other available IP addresses, whatever they are
> http_port 0.0.0.0:3128 ...
> # Error: Two _port definitions with the same IP:port address (0.0.0.0)
> http_port 3128 ...
> ```

yadij:
What I am thinking the check logic should be is:

```
 for (const auto old ; HttpPortList) {
    if (s->port != old->port)
        continue; // OK: different ports

    if (s->ip == old->ip)
       throw ...; // Error: exactly duplicate IP:port

    if (old->ip.IsAnyAddr())
       throw ...; // Error: explicit IP:port following wildcard line with same port

    if (s->ip.IsAnyAddr() && s->ip == old->ip) {
#if SQUID_WINDOWS
      continue;
#else
      throw ...; // Error: IP + wildcard with same port is not supported
#endif
    }

    // else: OK different IP's and no special-case issues
 }
```

rousskov:
> > * specific/wildcard address sharing that I described [works](https://stackoverflow.com/a/33622121) (without any special flags) on modern Windows;
> > * specific/wildcard address sharing that I described does not work (with or without SO_REUSEADDR) on Unix

> Then your second config line below should be "OK" on Windows (native?) builds ...
> ... and "Error" on all other builds.

It should be OK everywhere, from this PR detection/code point of view. What happens later, at bind() time is out of scope _here_. The low-level details of what a given platform and a given set of flags can or cannot do are out of this PR scope. This PR should not try to detect every bind() conflict on every platform, in every configuration. Detecting all such conflicts _correctly_ would be very difficult, and the end result would be worse than if we keep this PR focused on simple high-level platform-agnostic checks. The checks we should be after here will not _guarantee_ future bind() successes! They will let a few corner cases slip through, and that is OK.

The only slightly tricky part that I foresee in _this_ PR scope is detecting the difference between "any IP", "any IPv4", and "any IPv6" addresses. That high-level detection is in scope.

```
# OK: Special http_port for local cache manager queries only
http_port 127.0.0.1:3128 ...

# OK: Trying to listen for traffic on all other available IPv4 addresses, whatever they are
http_port 0.0.0.0:3128 ...

# OK: Trying to listen for traffic on all IPv6 addresses, whatever they are
http_port :::3128 ...

# Error: Multiple _port definitions with the same IP:port address (0.0.0.0)
# Error: Multiple _port definitions with the same IP:port address (::)
http_port 3128 ...
```

rousskov:
```C++
    if (old->ip.IsAnyAddr())
       throw ...; // Error: explicit IP:port following wildcard line with same port
```

I expect the decision here to be slightly more complex to allow addresses in different IP families. However, I have not checked whether this IsAnyAddr() check covers all those cases -- I just doubt it does.

```C++
#if SQUID_WINDOWS
...
```

Out of scope. Even on Windows, the correct decision depends on many factors. Let bind() take care of all that low-level complexity.

yadij:
The situation is simple enough that _some_ OS permit the explicit+wildcard case. Most do not.

The policy solution is just as simple. Reject the case with "not supported" message (as opposed to "invalid config" or such). Anyone who wants it supported can PR the special-case detection later for their desired OS.

rousskov:
> The situation is simple enough that _some_ OS permit the explicit+wildcard case. Most do not.

The some/most distinction does not simplify anything AFAICT.

Detecting whether the current environment permits the explicit+wildcard case is not simple. It is not just "Windows vs non-Windows" AFAICT. That is why it should not be done in this supposed-to-be-simple high-level code IMO. However, if somebody writes and tests a high-quality implementation that correctly classifies "permitting" and "not permitting" cases, then I would not object to it going in. I expect it will take a lot of effort (on all sides) to do this well, and that energy is better spent elsewhere IMO.

> The policy solution is just as simple. Reject the case with "not supported" message

That sounds like the opposite of what we should do because quickly (but incorrectly) banning a supported use case is much worse than delaying the (correct) ban of unsupported cases.

IMO, since correct detection of those "permitting" environments is difficult _and_ the bind() call will reliably reject unsupported environments later, the configuration checker should simply allow all explicit+wildcard cases instead of working hard to detect whether the current environment is "permitting" those cases.

PR #794, opened 2021-04-02
Removed redundant Http::StatusLine::protocol

rousskov:
Also polished Http::StatusLine initialization, but that part needs a
lot more work to get rid of Http::StatusLine::init().

### DIFF
--- a/src/http.cc
+++ b/src/http.cc
@@ -734,7 +734,6 @@ HttpStateData::processReplyHeader()
 
     // parse headers
     if (!newrep->parseHeader(*hp)) {
-        // XXX: when Http::ProtocolVersion is a function, remove this hack. just set with messageProtocol()
         newrep->sline.set(hp->messageProtocol(), Http::scInvalidHeader);
         debugs(11, 2, "error parsing response headers mime block");
     }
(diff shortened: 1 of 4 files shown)


src/http.cc
```diff
@@ -730,18 +730,12 @@ HttpStateData::processReplyHeader()
...
     //      We do that now to avoid performance regression from using SBuf::c_str()
-    newrep->sline.set(Http::ProtocolVersion(1,1), hp->messageStatus() /* , hp->reasonPhrase() */);
-    newrep->sline.protocol = newrep->sline.version.protocol = hp->messageProtocol().protocol;
-    newrep->sline.version.major = hp->messageProtocol().major;
-    newrep->sline.version.minor = hp->messageProtocol().minor;
+    newrep->sline.set(hp->messageProtocol(), hp->messageStatus() /* , hp->reasonPhrase() */);
 
     // parse headers
     if (!newrep->parseHeader(*hp)) {
         // XXX: when Http::ProtocolVersion is a function, remove this hack. just set with messageProtocol()
```
yadij, line 737: Your change below resolves this XXX. Yay :)
```suggestion
```

yadij (approved):

src/http.cc
rousskov: Done in 4c40ecd.

PR #795, opened 2021-04-02
Bug 4528: ICAP transactions quit on async DNS lookups

rousskov:
When an ICAP OPTIONS transaction needed to look up an IP address of the
ICAP service, and that address was not cached by Squid, the transaction
ended prematurely because Adaptation::Icap::Xaction::doneAll() was
unaware of ipcache_nbgethostbyname()'s async nature.

Adaptation::Icap::ModXact uses the same buggy Xaction code, but hides
the bug because ModXact::startWriting() sets state.writing before
calling openConnection() which schedules the DNS lookup. That "I am
still writing" state makes ModXact::doneAll() false.

However, some REQMOD and RESPMOD transactions are still affected by this
bug because they require an OPTIONS transaction if the service options
have never been fetched before or have expired. For example, the first
few transactions for a given service -- all those started before the DNS
lookup completes and Squid caches its result -- will fail.

Broken since inception (commit fb505fa).

### DIFF
--- a/src/adaptation/icap/Xaction.cc
+++ b/src/adaptation/icap/Xaction.cc
@@ -86,6 +86,7 @@ Adaptation::Icap::Xaction::Xaction(const char *aTypeName, Adaptation::Icap::Serv
     isRetriable(true),
     isRepeatable(true),
     ignoreLastWrite(false),
+    waitingForDns(false),
     stopReason(NULL),
     connector(NULL),
     reader(NULL),
@@ -187,12 +188,17 @@ Adaptation::Icap::Xaction::openConnection()
     debugs(93,3, typeName << " opens connection to " << s.cfg().host.termedBuf() << ":" << s.cfg().port);
 
     // Locate the Service IP(s) to open
+    assert(!waitingForDns);
+    waitingForDns = true; // before the possibly-synchronous ipcache_nbgethostbyname()
     ipcache_nbgethostbyname(s.cfg().host.termedBuf(), icapLookupDnsResults, this);
 }
 
 void
 Adaptation::Icap::Xaction::dnsLookupDone(const ipcache_addrs *ia)
 {
+    assert(waitingForDns);
+    waitingForDns = false;
+
     Adaptation::Icap::ServiceRep &s = service();
 
     if (ia == NULL) {
@@ -418,7 +424,8 @@ void Adaptation::Icap::Xaction::callEnd()
 
 bool Adaptation::Icap::Xaction::doneAll() const
 {
-    return !connector && !securer && !reader && !writer && Adaptation::Initiate::doneAll();
+    return !waitingForDns && !connector && !securer && !reader && !writer &&
+           Adaptation::Initiate::doneAll();
 }
 
 void Adaptation::Icap::Xaction::updateTimeout()
@@ -690,6 +697,9 @@ void Adaptation::Icap::Xaction::fillPendingStatus(MemBuf &buf) const
 
         buf.append(";", 1);
     }
+
+    if (waitingForDns)
+        buf.append("D", 1);
 }
 
 void Adaptation::Icap::Xaction::fillDoneStatus(MemBuf &buf) const
--- a/src/adaptation/icap/Xaction.h
+++ b/src/adaptation/icap/Xaction.h
@@ -137,6 +137,7 @@ protected:
     bool isRetriable;  ///< can retry on persistent connection failures
     bool isRepeatable; ///< can repeat if no or unsatisfactory response
     bool ignoreLastWrite;
+    bool waitingForDns; ///< expecting a ipcache_nbgethostbyname() callback
 
     const char *stopReason;
 


src/adaptation/icap/Xaction.cc
```diff
@@ -187,12 +188,17 @@ Adaptation::Icap::Xaction::openConnection()
     debugs(93,3, typeName << " opens connection to " << s.cfg().host.termedBuf() << ":" << s.cfg().port);
 
     // Locate the Service IP(s) to open
+    assert(!waitingForDns);
```
yadij, line 191: Can we use Must() here instead of assert() ? to only affect the transactions made within the window of brokenness. Other transactions once the data is available should not be prevented by Squid having shutdown.

src/adaptation/icap/Xaction.h
```diff
@@ -137,6 +137,7 @@ class Xaction: public Adaptation::Initiate
     bool isRetriable;  ///< can retry on persistent connection failures
     bool isRepeatable; ///< can repeat if no or unsatisfactory response
     bool ignoreLastWrite;
+    bool waitingForDns; ///< expecting a ipcache_nbgethostbyname() callback
```
yadij, line 140: Please use C++11 initialization for new members whenever possible.

yadij (changes requested):

src/adaptation/icap/Xaction.cc
rousskov: > It would be good to only affect the transactions made within the window of brokenness

Yes, we are on the same page here.

> Can we use Must() instead of assert() for these checks?

Yes, we can, but I do not think we should. I have finally completed the corresponding Error Handling nodes that argue [against](https://github.com/measurement-factory/squid-notes/blob/start/errors.md#invariant-checks-exceptions-vs-assertions) long-term assert() use and [against](https://github.com/measurement-factory/squid-notes/blob/start/errors.md#invariant-checks-missing-api) Must()s in new code like this.

Let's use this change request to decide how to check invariants here and/or in future PRs. I propose adding Assert() and can volunteer to do that (in a separate PR) if we decide that we should go in that direction (this PR code can wait or be merged as is). I cannot volunteer to check and refactor 900 Must()s, unfortunately. See the notes linked above for the details.

src/adaptation/icap/Xaction.h
rousskov: > Please use C++11 initialization for new members whenever possible.

I do not think this PR should use default member initializer _here_. The glaring inconsistency would mislead the reader into suspecting that other class members are left uninitialized (or that their initialization depends on constructor parameters). I agree that we should use default member initializers in most new classes, but I doubt we should create such inconsistencies in old ones, especially where a single explicit class constructor makes it easy to initialize consistently.

Let's wait for [automated](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-default-member-init.html) polishing of such cases. There was already a semi-approved [RFC](http://lists.squid-cache.org/pipermail/squid-dev/2020-April/009569.html) regarding that overall approach, and this modernization should be one of the items added to the candidate list.

src/adaptation/icap/Xaction.cc
yadij: I think the discussion around Must() and any missing exception APIs is an integral part of the design discussion we are going to have about long term try-catch placements. As such better in squid-dev@ than any github PR.

Since we are on the same page about how bad assert is for this and similar. Can we at least throw an exception if you disagree with Must() use as a "best we have" API? this seems to be a rather serious bug to block on designing perfect error handling in Squid.

src/adaptation/icap/Xaction.h
yadij: I disagree with most of your points. The other members *should* be considered uninitialized until constructor initializer list runs. At least until such time as they get converted to earlier initialization.

but do not insist on this change so resolving for now.

src/adaptation/icap/Xaction.cc
rousskov: > Can we at least throw an exception if you disagree with Must() use as a "best we have" API

In this context: 

* I believe assert() is the best short-term solution.
* Must() is worse than assert() because it hides the bug, as now unofficially documented [elsewhere](https://github.com/measurement-factory/squid-notes/blob/start/errors.md#invariant-checks-missing-api).
* Explicitly throwing an exception is worse than Must() because in addition to hiding the bug like Must() would do, it misleads the reader into thinking that this is input validation (rather than a code logic/invariant check). This option is also needlessly verbose, duplicating Must() functionality 1:1.
* Combining manual exception throwing with a custom level-0 debugs("BUG") statement (to expose the bug) is worse than assert() because it adds and then duplicates custom code that will need to be found and replaced with Assert() or similar. If we go down this route, then it is much better to just add and use Assert() instead!
* I am not against adding and using throwing-and-logging Assert(). It the best long-term solution AFAICT. It can be done in a dedicated prerequisite PR.

> I think the discussion around Must() and any missing exception APIs is an integral part of the design discussion we are going to have about long term try-catch placements. As such better in squid-dev@ than any github PR.

Your call.

> this seems to be a rather serious bug to block on designing perfect error handling in Squid.

I agree with the implication that this PR should not be blocked. Please consider unblocking. AFAICT, this PR code follows one of the well-known, wide-spread, valid approaches that modern Squid code uses for invariant checks. That approach is not perfect, but if we cannot agree on a replacement, then I see no valid reasons to block the PR. Needless to say, I am not blocking it.

yadij: Its clear this will need an XXX to highlight the bug then.

rousskov: Which bug?! I have no idea what XXX/problem you are referring to. You are not implying that all suspected transaction-scope asserts() will now need `XXX: Treat as a transaction-scope bug` comments, do you? If that is indeed the implication, then, again, it is much better to add and use Assert() (or similar) instead of sprinkling the code with such comments.

yadij: My current understanding is that this assert() is possible to trigger from a (rare) combination of server and client input timings. Which makes it a bug that they can trigger a full crash and DoS of Squid.

rousskov: > My current understanding is that this assert() is possible to trigger from a (rare) combination of server and client input timings.

What makes you think that? My current understanding is that there is no (known or suspected) way to trigger this assert().

yadij: The third paragraph of the description:
> However, some REQMOD and RESPMOD transactions are still affected by this
> bug because they require an OPTIONS transaction if the service options
> have never been fetched before or have expired. For example, the first
> few transactions for a given service -- all those started before the DNS
> lookup completes and Squid caches its result -- will fail.

Those "will fail" transactions happen between DNS being requested and found. Thus service's waitingForDns is true, making the assert happen.

If no, what am I misreading?

rousskov: I rewrote the PR description in hope to correct this misunderstanding. I am sorry that my original description was difficult to interpret correctly. I hope the current version is clear about what was wrong with REQMOD/RESPMOD transactions and cannot be easily misinterpreted as if they are still suffering from the fixed bug. Please check.

yadij (approved):

src/adaptation/icap/Xaction.cc
yadij: Oh, I see. Confusion was the tense of the word "could" differs between our variants of english. FYI, in my locale it is typically used as present-tense instead of "can" or "might" rather than past-tense.

rousskov:
@kinkie, do you know what happened with the last pending Jenkins test for this PR's staged commit https://github.com/squid-cache/squid/commit/26cf7c3c01c3e58178054eb99b323471e40e9703? GitHub is still waiting for a pass/fail answer, but the corresponding Jenkins console [log](https://build.squid-cache.org/job/5-pr-auto/COMPILER=clang,OS=centos-8,label=docker-build-host/905/console) shows an aborted test AFAICT:

```
21:04:23 .........................................................Scheduled restart of the build (up to 1 times).
02:04:23 Build timed out (after 300 minutes). Marking the build as aborted.
02:04:23 Build was aborted
02:04:23 [Set GitHub commit status (universal)] PENDING on repos [GHRepository@6df847f2[nodeId=MDEwOlJlcG9zaXRvcnkyNDUxOTk4Nw==,description=Squid Web Proxy Cache,homepage=http://www.squid-cache.org/,name=squid,fork=false,archived=false,size=168136,milestones={},language=C++,commits={},source=<null>,parent=<null>,isTemplate=<null>,url=https://api.github.com/repos/squid-cache/squid,id=24519987,nodeId=<null>,createdAt=2014-09-27T00:59:40Z,updatedAt=2021-05-22T04:23:11Z]] (sha:26cf7c3) with context:Jenkins(build test) centos-8,clang
02:04:23 Setting commit status on GitHub for https://github.com/squid-cache/squid/commit/26cf7c3c01c3e58178054eb99b323471e40e9703
02:04:25 Finished: ABORTED
```

I see a timeout and a restart being mentioned, but then Jenkins sends a `PENDING` test status after saying "Build was aborted". That does not make sense to me, but it could be just a Jenkins/etc. bug.

If you can revive this PR testing somehow, please do.

rousskov:
@eduard-bagdasaryan, please check why this PR got merged despite a failed Semaphore CI [test](https://semaphoreci.com/squid/squid/branches/master/builds/520)? The failure itself is benign in this context -- Semaphore CI just did not want to wait long enough, but I would expect Anubis to block merging because of that failure. Perhaps Anubis' expected test count configuration value got stale again?

PR #798, opened 2021-04-04
Prep for 4.15 and 5.0.6

yadij:
Documentation only.


rousskov:
@yadij, do you need this fast-tracked like #765?

yadij:
@rousskov, release is blocked by PR #788 completion and backport right now. Will see about fast-track when that is done.

PR #800, opened 2021-04-04
Fix various Clang-12.0 compiler warnings

MegaManSec:
Fixes -Wrange-loop-construct, -Wmismatched-tags, and -Wshadow.

### DIFF
--- a/src/acl/Asn.cc
+++ b/src/acl/Asn.cc
@@ -595,7 +595,7 @@ ACLDestinationASNStrategy::match (ACLData<MatchType> * &data, ACLFilledChecklist
     const ipcache_addrs *ia = ipcache_gethostbyname(checklist->request->url.host(), IP_LOOKUP_IF_MISS);
 
     if (ia) {
-        for (const auto ip: ia->goodAndBad()) {
+        for (const auto &ip: ia->goodAndBad()) {
             if (data->match(ip))
                 return 1;
         }
--- a/src/http/RegisteredHeadersHash.cci
+++ b/src/http/RegisteredHeadersHash.cci
@@ -1,35 +1,35 @@
 /* C++ code produced by gperf version 3.1 */
-/* Command-line: gperf -m 100000 RegisteredHeadersHash.gperf  */
+/* Command-line: gperf -m 100000 ./RegisteredHeadersHash.gperf  */
 /* Computed positions: -k'3,9,$' */
 
 #if !((' ' == 32) && ('!' == 33) && ('"' == 34) && ('#' == 35) \
-&& ('%' == 37) && ('&' == 38) && ('\'' == 39) && ('(' == 40) \
-&& (')' == 41) && ('*' == 42) && ('+' == 43) && (',' == 44) \
-&& ('-' == 45) && ('.' == 46) && ('/' == 47) && ('0' == 48) \
-&& ('1' == 49) && ('2' == 50) && ('3' == 51) && ('4' == 52) \
-&& ('5' == 53) && ('6' == 54) && ('7' == 55) && ('8' == 56) \
-&& ('9' == 57) && (':' == 58) && (';' == 59) && ('<' == 60) \
-&& ('=' == 61) && ('>' == 62) && ('?' == 63) && ('A' == 65) \
-&& ('B' == 66) && ('C' == 67) && ('D' == 68) && ('E' == 69) \
-&& ('F' == 70) && ('G' == 71) && ('H' == 72) && ('I' == 73) \
-&& ('J' == 74) && ('K' == 75) && ('L' == 76) && ('M' == 77) \
-&& ('N' == 78) && ('O' == 79) && ('P' == 80) && ('Q' == 81) \
-&& ('R' == 82) && ('S' == 83) && ('T' == 84) && ('U' == 85) \
-&& ('V' == 86) && ('W' == 87) && ('X' == 88) && ('Y' == 89) \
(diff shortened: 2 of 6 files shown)


src/acl/Asn.cc
```diff
@@ -595,7 +595,7 @@ ACLDestinationASNStrategy::match (ACLData<MatchType> * &data, ACLFilledChecklist
     const ipcache_addrs *ia = ipcache_gethostbyname(checklist->request->url.host(), IP_LOOKUP_IF_MISS);
 
     if (ia) {
-        for (const auto ip: ia->goodAndBad()) {
+        for (const auto &ip: ia->goodAndBad()) {
```
rousskov, line 598: Could you please let me know what warning did you get for this line. I agree that we should use a reference here, if possible, but I am curious what the compiler was warning us about in this a similar places.

src/http/RegisteredHeadersHash.cci
```diff
@@ -41,7 +41,7 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 #line 24 "RegisteredHeadersHash.gperf"
-struct HeaderTableRecord;
+class HeaderTableRecord;
```
rousskov, line 44: We should not be hand-editing generated files. AFAICT, this needs to be fixed in `src/http/RegisteredHeadersHash.gperf` (which also has a "generated from..." header, but _that_ header is lying IIRC).

You will need a matching gperf version to properly regenerate everything. See `git grep gperf scripts/source-maintenance.sh` for a regeneration command.

rousskov (changes requested):
Thank you for working on fixing these problems. Please add clang version to the PR title and (unique) warnings/codes to the PR description.

BTW, our PR title/description (i.e. the future commit message) format requirements are documented [elsewhere](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message). Following them should remove the M-failed-description label added by Anubis.

src/acl/Asn.cc
rousskov: Never mind. I found the `-Wrange-loop-construct` answer in your branch commit and am resolving this change request. Please note that we normally _squash_ all branch commits, so, if there is useful information in them, we have to copy it to the PR description. This also helps avoid fishing for info in branch commits.

src/http/RegisteredHeadersHash.cci
MegaManSec: Ha! I originally fixed it in the gperf file, noticed it said "generated from", then moved it.

src/http/RegisteredHeadersHash.cci
```diff
@@ -1,35 +1,35 @@
 /* C++ code produced by gperf version 3.1 */
-/* Command-line: gperf -m 100000 RegisteredHeadersHash.gperf  */
+/* Command-line: gperf -m 100000 ./RegisteredHeadersHash.gperf  */
```
MegaManSec, line 2: I don't know why it has generated like this.
I ran `make -C src/http gperf-files` as suggested.

rousskov: I fixed that and reformatted in 42991ff.

src/http/RegisteredHeadersHash.cci
```diff
@@ -363,8 +363,8 @@ static const struct HeaderTableRecord HttpHeaderDefinitionsTable[] =
     {"Last-Modified", Http::HdrType::LAST_MODIFIED, Http::HdrFieldType::ftDate_1123, HdrKind::EntityHeader}
 };
 
-const struct HeaderTableRecord *
-HttpHeaderHashTable::lookup (const char *str, size_t len)
+const class HeaderTableRecord *
+    HttpHeaderHashTable::lookup (const char *str, size_t len)
```
rousskov, line 367: I do not know why my astyle v2.04 indents this line while the official (probably) does not.

rousskov (approved):

src/http/RegisteredHeadersHash.cci
MegaManSec: Thank you.

MegaManSec:
Btw, with the `S-waiting-for-committer` tag, if by 'committer' that means me (rather than someone to comment 'OK to test'), I don't know what i need to do.

rousskov:
OK to test

rousskov:
@MegaManSec, Semaphore CI does not see the latest commit on this branch. I suspect some GitHub notification got lost. It happens from time to time, and there is no Semaphore UI to _force_ a branch pull AFAICT. To wake Semaphore CI up, please feel free to push a dummy or trivial commit (perhaps removing the possibly bogus [indent](https://github.com/squid-cache/squid/pull/800#discussion_r606890647)?).

P.S. In this case, `S-waiting-for-committer` is my signal for other core developers (that have rights to add `M-clreared-for-merge`). I usually give others heads up for impending single-review PR mergers in hope to reduce human-human conflicts.

MegaManSec:
Thank you for the explanation. I had originally thought `for-committer` meant something required on my side (as committer to the pull request), then took a look at some previous reports and understood it to mean "a privileged user needs to do something (as committer to the master branch).

rousskov:
Your commit has [awaken](https://semaphoreci.com/squid/squid/branches/pull-request-800/builds/5) Semaphore CI, thank you!

P.S. We need to revamp and re-[publish](http://lists.squid-cache.org/pipermail/squid-dev/2017-August/009258.html) detailed explanations for these `S-...` labels and perhaps rename to avoid the overloaded term "committer" that means something different/special for Squid insiders than it is for new contributors.

yadij (approved):

PR #802, opened 2021-04-05
Improve comment on snmp_core.cc:snmpCreateOidFromStr

sergiodj:
Fix typo and better reflect what the code does.

### DIFF
--- a/src/snmp_core.cc
+++ b/src/snmp_core.cc
@@ -926,6 +926,10 @@ snmpLookupNodeStr(mib_tree_entry *root, const char *str)
     return e;
 }
 
+/**
+ * Try to convert an OID string representation into binary.
+ * \retval false if the string cannot be parsed entirely.
+ */
 bool
 snmpCreateOidFromStr(const char *str, oid **name, int *nl)
 {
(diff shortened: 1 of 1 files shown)


yadij (changes requested):
This comment is about the API return values. Current style guidelines for documentation of static function APIs is to use doxygen format and place before the function.

For example;
```
/**
 * Try to convert an OID string representation into binary.
 * \retval false if the string cannot be parsed entirely.
 */
bool
snmpCreateOidFromStr(const char *str, oid **name, int *nl)
{
```

sergiodj:
> Try to convert an OID string representation into binary.

OK.

I still think it's useful to have a comment clarifying possible scenarios which will cause the string not to be parsed, so I've adjusted the comment inside the function.

yadij:
It should be obvious to anyone reading that code (and thus seeing the comment) what kind of things reach the free and false result.

Comments have to be manually maintained, like what you are doing now. So trying to list every specific case that can reach that free+return section is difficult (eg your comment still misses cases) and forces work onto future devs. The *true* case(s) are more restricted and thus possibly worth documenting each one (which is already done).

PR #803, opened 2021-04-07
%ssl::<negotiated_version, %ssl::>negotiated_version for TLS/1.3

chtsanti:
Both %codes were logged as dashes when Squid negotiated TLS v1.3.


rousskov (approved):
Thank you for fixing this TLS v1.3 support aspect!

I polished the PR title/description a little to allow for exact %code matches by name, to document the symptoms of the problem, and to remove the low-level details clearly provided by the diff.

yadij (approved):

PR #804, opened 2021-04-16
Fix a few level-2+ debugs(); improve ErrorState debugging

MegaManSec:
When possible, report object contents rather than its memory address.

### DIFF
--- a/src/Debug.h
+++ b/src/Debug.h
@@ -171,7 +171,7 @@ HERE(std::ostream& s)
 #endif
 
 /* some uint8_t do not like streaming control-chars (values 0-31, 127+) */
-inline std::ostream &operator <<(std::ostream &os, const uint8_t d)
+inline std::ostream& operator <<(std::ostream &os, const uint8_t d)
 {
     return (os << (int)d);
 }
--- a/src/Debug.h
+++ b/src/Debug.h
@@ -171,7 +171,7 @@ HERE(std::ostream& s)
 #endif
 
 /* some uint8_t do not like streaming control-chars (values 0-31, 127+) */
-inline std::ostream& operator <<(std::ostream &os, const uint8_t d)
+inline std::ostream &operator <<(std::ostream &os, const uint8_t d)
 {
     return (os << (int)d);
 }
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -762,16 +762,10 @@ FwdState::doneWithRetries()
 }
 
 // called by the server that failed after calling unregister()
-// err may or may not be null here depending on the type of server end.
 void
 FwdState::handleUnregisteredServerEnd()
 {
-    if(err) {
-        const auto errType = errorTypeName(err->type);
-        debugs(17, 2, HERE << "self=" << self << " err=" << errType << ' ' << entry->url());
-    } else {
-        debugs(17, 2, HERE << "self=" << self << " err=" << err << ' ' << entry->url());
-    }
+    debugs(17, 2, HERE << "self=" << self << " err=" << err << ' ' << entry->url());
     assert(!Comm::IsConnOpen(serverConn));
     retryOrBail();
 }
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -762,10 +762,16 @@ FwdState::doneWithRetries()
 }
 
 // called by the server that failed after calling unregister()
+// err may or may not be null here depending on the type of server end.
 void
 FwdState::handleUnregisteredServerEnd()
 {
-    debugs(17, 2, HERE << "self=" << self << " err=" << err << ' ' << entry->url());
+    if(err) {
+        const auto errType = errorTypeName(err->type);
+        debugs(17, 2, HERE << "self=" << self << " err=" << errType << ' ' << entry->url());
+    } else {
+        debugs(17, 2, HERE << "self=" << self << " err=" << err << ' ' << entry->url());
+    }
     assert(!Comm::IsConnOpen(serverConn));
     retryOrBail();
 }
--- a/src/HttpHdrSc.cc
+++ b/src/HttpHdrSc.cc
@@ -108,7 +108,7 @@ HttpHdrSc::parse(const String * str)
         type = scLookupTable.lookup(SBuf(item,ilen));
 
         if (type == SC_ENUM_END) {
-            debugs(90, 2, "hdr sc: unknown control-directive: near '" << item << "' in '" << *str << "'");
+            debugs(90, 2, "unknown control-directive near '" << item << "' in '" << *str << "'");
             type = SC_OTHER;
         }
 
@@ -134,7 +134,7 @@ HttpHdrSc::parse(const String * str)
 
         if (sct->isSet(type)) {
             if (type != SC_OTHER)
-                debugs(90, 2, "hdr sc: ignoring duplicate control-directive: near '" << item << "' in '" << *str << "'");
+                debugs(90, 2, "ignoring duplicate control-directive near '" << item << "' in '" << *str << "'");
 
             ++ scHeaderStats[type].repCount;
 
--- a/src/HttpHdrSc.cc
+++ b/src/HttpHdrSc.cc
@@ -108,7 +108,7 @@ HttpHdrSc::parse(const String * str)
         type = scLookupTable.lookup(SBuf(item,ilen));
 
         if (type == SC_ENUM_END) {
-            debugs(90, 2, "hdr sc: unknown control-directive: near '" << item << "' in '" << str << "'");
+            debugs(90, 2, "hdr sc: unknown control-directive: near '" << item << "' in '" << *str << "'");
             type = SC_OTHER;
         }
 
@@ -134,7 +134,7 @@ HttpHdrSc::parse(const String * str)
 
         if (sct->isSet(type)) {
             if (type != SC_OTHER)
-                debugs(90, 2, "hdr sc: ignoring duplicate control-directive: near '" << item << "' in '" << str << "'");
+                debugs(90, 2, "hdr sc: ignoring duplicate control-directive: near '" << item << "' in '" << *str << "'");
 
             ++ scHeaderStats[type].repCount;
 
--- a/src/errorpage.cc
+++ b/src/errorpage.cc
@@ -1389,7 +1389,7 @@ ErrorState::buildBody()
     if (!Config.errorDirectory)
         err_language = Config.errorDefaultLanguage;
 #endif
-    debugs(4, 2, "No existing error page language negotiated for " << errorPageName(page_id) << ". Using default error file.");
+    debugs(4, 2, "No existing error page language negotiated for " << this << ". Using default error file.");
     return compileBody(error_text[page_id], true);
 }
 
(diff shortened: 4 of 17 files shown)


MegaManSec:
Note: There is possibly one missing fix in this patch. From client_side_request.cc:
https://github.com/squid-cache/squid/blob/c6f5ad1cf45c998f852febaf68113283c63957c2/src/client_side_request.cc#L1274-L1275

It is unclear (to me) whether old_request (and new_request) should be printed here, or old_request->effectiveRequestUri(). I can see arguments for both cases.

src/FwdState.cc
```diff
@@ -762,10 +762,16 @@ FwdState::doneWithRetries()
...
 void
 FwdState::handleUnregisteredServerEnd()
 {
-    debugs(17, 2, HERE << "self=" << self << " err=" << err << ' ' << entry->url());
+    if(err) {
+        const auto errType = errorTypeName(err->type);
+        debugs(17, 2, HERE << "self=" << self << " err=" << errType << ' ' << entry->url());
+    } else {
+        debugs(17, 2, HERE << "self=" << self << " err=" << err << ' ' << entry->url());
+    }
```
rousskov, line 774: Please undo this entire FwdState change because it adds too much always-executed code and noise for debugging sake.

If you really want to improve `err` debugging, overload the corresponding `<<` operator instead (in errorpage.{h,cc} files), keeping the above debugs() line intact. There is a somewhat similar example already (see the first bullet below), but this new _public_ operator would have to print something like `[none]` when there is no error (see the second bullet for inspiration). The operator will improve debugging in several other places (see the third bullet for an incomplete dirty list of candidates -- you can and, ideally, should find (and, if needed, polish) all of them by [deleting](https://godbolt.org/z/4aaWhTK4G) the printing operator first).

* `git grep 'operator <<.*[*]'`
* `git grep '<< "[[]n'`
* `git grep '<< err\b'`

I insist on the removal, but I do not insist on adding the operator. I can volunteer to add it, but it will take some time.

src/HttpHdrSc.cc
```diff
@@ -108,7 +108,7 @@ HttpHdrSc::parse(const String * str)
         type = scLookupTable.lookup(SBuf(item,ilen));
 
         if (type == SC_ENUM_END) {
-            debugs(90, 2, "hdr sc: unknown control-directive: near '" << item << "' in '" << str << "'");
+            debugs(90, 2, "hdr sc: unknown control-directive: near '" << item << "' in '" << *str << "'");
```
rousskov, line 111: Just polishing to improve phrasing and remove duplicated information from this modified debugs() line. The removed information is added automatically by debugs() to all level-2+ messages. This will also make this line a bit more consistent with HttpHdrRange debugging:

```suggestion
            debugs(90, 2, "unknown control-directive near '" << item << "' in '" << *str << "'");
```

src/HttpHdrSc.cc
```diff
@@ -134,7 +134,7 @@ HttpHdrSc::parse(const String * str)
 
         if (sct->isSet(type)) {
             if (type != SC_OTHER)
-                debugs(90, 2, "hdr sc: ignoring duplicate control-directive: near '" << item << "' in '" << str << "'");
+                debugs(90, 2, "hdr sc: ignoring duplicate control-directive: near '" << item << "' in '" << *str << "'");
```
rousskov, line 137: Same reasoning as in the other similar change request:

```suggestion
                debugs(90, 2, "ignoring duplicate control-directive near '" << item << "' in '" << *str << "'");
```

rousskov (changes requested):
Thank you for fixing this debugging bug!

rousskov:
> It is unclear (to me) whether old_request (and new_request) should be printed here, or old_request->effectiveRequestUri().

Actually, given those three options, I believe it is better to continue printing raw pipe pointers there. Those pointers are visible on other debugging lines, and they help correlate pipe activity (better than often-repeated URIs or request addresses do).

Ideally, a BodyPipe ID should be added and printed instead of raw pointers, of course. IIRC, InstanceId did not exist when pipes were added, but it can and should be added/used now. Doing so can be considered outside this PR scope, of course.

MegaManSec:
Thank you for confirming the intent of the logging of the body_pipe objects. I agree that the pointer addresses should be printed in this case, then, because the log is about the 'request' (let's say) itself, not the specific URI of the pipe.

I will push the patches you have suggested.

src/FwdState.cc
MegaManSec: As I am unfamiliar with how operator overloading works in C++, I believe it is best for me to just remove this, as I suspect any changes made by my in this case will just be a (frustrating) game of cat-and-mouse.

rousskov (approved):

rousskov:
OK to test

MegaManSec:
Actually, I decided learning how overloading of operators works is a net positive for me, so I did that. I am now confident enough to be able to know what is required (which is now also clear: not a lot).

However, I wanted to ask. For the case where there is no error, would it be more appropriate to use "ERR_NONE" as the output? Given that the output currently looks like this:
2021/04/17 01:35:45.335| 17,2| FwdState.cc(768) handleUnregisteredServerEnd: self=0x3dd7fa8*2 err=ERR_INVALID_RESP http://website.com/

While I do not think anybody will actually be parsing these debugging lines, I think it is appropriate to keep output in a similar format. What do you think, @rousskov?

rousskov:
> For the case where there is no error, would it be more appropriate to use "ERR_NONE" as the output?

No, it would not be. When the code is printing/reporting a nil pointer, it should not mislead the reader/developer into thinking that a pointer points to an Error object with an `ERR_NONE` error type. Those two cases are usually very different!

N.B. Your simple operator may only print the error type today, but it may be enhanced later to print more error details. The `[nil]` or similar output will always be there to signify a special case of a nil error pointer though.

MegaManSec:
Thanks for the explanation; I am currently reviewing cases of err being printed, and will then push a patch.

MegaManSec:
Alright, I checked all instances, and they seem fine. The only one perhaps worth (is this patch actually really worth anything?) improving is:

```
/// remembers an error to be used if there will be no more connection attempts
void
TunnelStateData::saveError(ErrorState *error)
{
    debugs(26, 4, savedError << " ? " << error);
    assert(error);
    delete savedError; // may be nil
    savedError = error;
}
```

Both savedError and error are obviously an ErrorState. I wonder if a more appropriate message could be used?

Also this patch vastly improves this level-4 debugging in errorAppendEntry:
```
    debugs(4, 4, "Creating an error page for entry " << entry <<
           " with errorstate " << err <<
           " page id " << err->page_id);
```

src/errorpage.h
```diff
@@ -351,5 +351,6 @@ class TemplateFile
  */
 bool strHdrAcptLangGetItem(const String &hdr, char *lang, int langLen, size_t &pos);
 
+std::ostream & operator <<(std::ostream &os, ErrorState *err);
```
yadij, line 354: Since the type names give us nice info about the parameter we do no need to name them here.

src/errorpage.cc
```diff
@@ -128,6 +128,7 @@ class BuildErrorPrinter
     const char *near;
 };
 
+
```
yadij, line 131: Useless empty line
```suggestion
```

yadij (changes requested):
LGTM, just polishing

src/errorpage.h
MegaManSec: I hope that I have understood what you meant here, and will push a patch in a minute. I have also changed other instances of `std::ostream &operator` being defined inconsistently. This is purely a pedantic stylistic change by me: I hope it's OK by you.

src/Debug.h
```diff
@@ -171,7 +171,7 @@ HERE(std::ostream& s)
 #endif
 
 /* some uint8_t do not like streaming control-chars (values 0-31, 127+) */
-inline std::ostream& operator <<(std::ostream &os, const uint8_t d)
+inline std::ostream &operator <<(std::ostream &os, const uint8_t d)
```
rousskov, line 174: > This is purely a pedantic stylistic change by me: I hope it's OK by you.

Please undo those changes. While I prefer this style as well, code style changes (of otherwise unmodified lines) are out of this PR scope. In most cases, such changes do more harm then good overall, especially when applied manually.

rousskov (changes requested):

src/errorpage.h
```diff
@@ -351,5 +351,6 @@ class TemplateFile
  */
 bool strHdrAcptLangGetItem(const String &hdr, char *lang, int langLen, size_t &pos);
 
+std::ostream &operator <<(std::ostream &, ErrorState *);
```
rousskov, line 354: Rule of thumb: If you can make it `const`, make it `const`.

```suggestion
std::ostream &operator <<(std::ostream &, const ErrorState *);
```

src/errorpage.cc
```diff
@@ -1523,3 +1523,12 @@ ErrorPage::ValidateStaticError(const int page_id, const SBuf &inputLocation)
     anErr.validate();
 }
 
+std::ostream &
+operator <<(std::ostream &os, ErrorState *err)
```
rousskov, line 1527:
```suggestion
operator <<(std::ostream &os, const ErrorState *err)
```

Technically, there is one more `const` missing here (for the `err` pointer itself), but Squid code does not usually use it, I myself have not gotten into the habit of adding it, so I am not suggesting it. Your call though.

rousskov (changes requested):

rousskov:
> ```
> /// remembers an error to be used if there will be no more connection attempts
> void
> TunnelStateData::saveError(ErrorState *error)
> {
>     debugs(26, 4, savedError << " ? " << error);
>     assert(error);
>     delete savedError; // may be nil
>     savedError = error;
> }
> ```
> 
> Both savedError and error are obviously an ErrorState. I wonder if a more appropriate message could be used?

Leave as is: It is difficult to write a good message for this particular line, and there is a pending change that rewrites this method IIRC.

yadij (approved):

yadij:
@rousskov, this can be cleared for anubis when your review passes.

src/errorpage.cc
```diff
@@ -1523,3 +1523,12 @@ ErrorPage::ValidateStaticError(const int page_id, const SBuf &inputLocation)
     anErr.validate();
 }
 
+std::ostream &
+operator <<(std::ostream &os, const ErrorState *err)
+{
+    if (err)
+        os << errorTypeName(err->type);
```
rousskov: AFAICT, using errorTypeName() loses previously reported information when ErrorState::type is at or above `ERR_MAX` (i.e. when this is a so called "dynamically loaded" error). Fixed in 419899d.

N.B. ErrorState::type and ErrorState::page_id are pretty much the same, but fixing that bug is out of this PR scope.

rousskov (approved):
I fixed reporting of "dynamic" error pages and polished a little. Will clear for merging if there are no objections for a few days. @yadij, please feel free to clear sooner.

rousskov:
Also synced PR title (i.e. future commit title) with the current PR state.

yadij (approved):

PR #805, opened 2021-04-19
Do not forward malformed warning-values

rousskov:
Since we started policing Warning header field values (commit c679653),
Squid was preserving some malformed warning-values and removing some
other malformed warning-values. I speculate that inconsistency was a
bug. Now we consistently remove all malformed warning-values we detect.

### DIFF
--- a/src/HttpReply.cc
+++ b/src/HttpReply.cc
@@ -600,7 +600,7 @@ void HttpReply::removeStaleWarnings()
 {
     String warning;
     if (header.getList(Http::HdrType::WARNING, &warning)) {
-        const String newWarning = removeStaleWarningValues(warning);
+        const auto newWarning = findFreshWarningValues(warning);
         if (warning.size() && warning.size() == newWarning.size())
             return; // some warnings are there and none changed
         header.delById(Http::HdrType::WARNING);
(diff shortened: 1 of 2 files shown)


rousskov:
Disclaimer: This change has not been tested yet.

MegaManSec:
Thank you for this!

I can, at the very least, say this now makes the Warning header behave the way I would expect it to.

src/HttpReply.cc
```diff
@@ -612,49 +612,45 @@ void HttpReply::removeStaleWarnings()
...
- */
-String HttpReply::removeStaleWarningValues(const String &value)
+/// \returns a list of warning-values with warn-date equal to the reply Date
+String
+HttpReply::findFreshWarningValues(const String &warningFieldValue) const
 {
     String newValue;
     const char *item = 0;
     int len = 0;
     const char *pos = 0;
```
yadij, line 622: Before merge, please correct these to nullptr initialization. This is technically a new function even though it shares a lot of syntax with the old one.

yadij (approved):
LGTM

yadij (changes requested):
Actually. Design breaks HTTP spec.  warn-data is optional. That is what the old keep=true initialization was about, to keep entries without a warn-date.

It looks like this design discards by omission when time<=0. It should consider those "fresh" for v1.1+. We apparently lack the logic detecting v1.0 where warn-date was mandatory.

src/HttpReply.cc
```diff
@@ -612,49 +612,45 @@ void HttpReply::removeStaleWarnings()
...
             while (p >= item && xisspace(*p)) --p; // skip whitespace
 
             if (p >= item && *p == '"' && warnDateBeg - p > 2) {
-                // found warn-text
+                // found warn-text, confirming that warnDateBeg is warn-date
                 String warnDate;
                 warnDate.append(warnDateBeg, warnDateEnd - warnDateBeg);
                 const time_t time = parse_rfc1123(warnDate.termedBuf());
-                keep = (time > 0 && time == date); // keep valid and matching date
+                if (time > 0 && time == date) { // valid and matching
```
yadij, line 647:
```suggestion
                if (time > 0 && time != date)
                    continue; // invalid date field found
                if (time <= 0 && http_ver == Http::ProtocolVersion(1,0))
                    continue; // HTTP/1.0 requires date field on Warning headers
            }

```

src/HttpReply.cc
```diff
@@ -612,49 +612,45 @@ void HttpReply::removeStaleWarnings()
...
+                    if (newValue.size())
+                        newValue.append(", ");
+                    newValue.append(item, len);
+                }
```
yadij, line 651:
```suggestion
```

yadij (changes requested):

rousskov:
> Design breaks HTTP spec. warn-data is optional. That is what the old keep=true initialization was about, to keep entries without a warn-date.

Agreed. I suggest closing this PR then without committing anything. The original code is not as broken if warnings without warn-date must be kept in some messages. And, FWIW, I am not a big fan of making traffic policing decisions based on HTTP protocol minor version (where they can be avoided) -- too many exceptions in that area.

I can also rewrite the PR to just add a source code comment to explain `keep = true`.

@yadij, what is your preference?

yadij:
> @yadij, what is your preference?

Patches reducing brokenness are just as welcome as perfectly complete ones IMO. Up to you how you do that.

rousskov:
Closing because the problem is much smaller (and different) than thought, and I personally do not have the time for the minor (but likely contentions) adjustments that are necessary to improve this code further. This closure has no bearing on future pull requests improving this code; quality improvements are still welcomed, of course.

PR #809, opened 2021-05-01
Replace cbdata::Offset hack with offsetof()

yadij:
Also remove unused OFFSET_OF macro.

### DIFF
--- a/src/cbdata.cc
+++ b/src/cbdata.cc
@@ -46,8 +46,6 @@ public:
 
 #endif
 
-#define OFFSET_OF(TYPE, MEMBER) ((size_t) &(((TYPE) *)0)->(MEMBER))
-
 /**
  * Manage a set of registered callback data pointers.
  * One of the easiest ways to make Squid coredump is to issue a
@@ -116,25 +114,12 @@ public:
 
 #if !WITH_VALGRIND
     size_t dataSize() const { return sizeof(data);}
-    static long MakeOffset();
-    static const long Offset;
 #endif
     /* MUST be the last per-instance member */
     void *data;
 };
 
 const long cbdata::Cookie((long)0xDEADBEEF);
-#if !WITH_VALGRIND
-const long cbdata::Offset(MakeOffset());
-
-long
-cbdata::MakeOffset()
-{
-    cbdata *zero = (cbdata *)0L;
-    void **dataOffset = &zero->data;
-    return (long)dataOffset;
-}
-#endif
 
 static OBJH cbdataDump;
 #if USE_CBDATA_DEBUG
@@ -186,8 +171,7 @@ cbdataInternalInitType(cbdata_type type, const char *name, int size)
     snprintf(label, strlen(name) + 20, "cbdata %s (%d)", name, (int) type);
 
 #if !WITH_VALGRIND
-    assert((size_t)cbdata::Offset == (sizeof(cbdata) - ((cbdata *)NULL)->dataSize()));
-    size += cbdata::Offset;
+    size += offsetof(cbdata, data);
 #endif
 
     cbdata_index[type].pool = memPoolCreate(label, size);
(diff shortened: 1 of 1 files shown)


kinkie (approved):
Good for me to go in as soon as build tests finish. This will also unblock PR#808

src/cbdata.cc
```diff
@@ -373,7 +357,7 @@ cbdataInternalUnlock(const void *p)
 #if WITH_VALGRIND
     c = cbdata_htable.at(p);
 #else
-    c = (cbdata *) (((char *) p) - cbdata::Offset);
+    c = (cbdata *) (((char *) p) - offsetof(cbdata, data));
 #endif
```
rousskov, line 361: Please avoid duplicating this low-level code so many times. I would try to add cbdata::FromUserData(), FromCallerData(), or a similar method, replacing all 5 lines with:

```C++
c = cbdata::FromUserData(p);
```

rousskov (changes requested):
Thank you for polishing this mess!

According to offsetof() [documentation](https://en.cppreference.com/w/cpp/types/offsetof), `If type is not a standard layout type, the behavior is undefined`. Please add a static assert after the cbdata class declaration to make sure we (or future cbdata changes) do not introduce a UB:

```C++
static_assert(std::is_standard_layout<cbdata>::value, "the behavior of offsetof(cbdata) is defined");
```

Please `#include <cstddef>` that provide the offsetof() declaration according to C++ [documentation](https://en.cppreference.com/w/cpp/types/offsetof). It is probably not necessary on currently tested platforms, but that is not a valid excuse to skip it AFAICT.

P.S. I find it interesting that C++17 removes the requirement for the compiler to support offsetof(). I doubt that change will affect Squid though.

rousskov:
I adjusted the PR title to remove `C++11` because [offsetof](https://en.cppreference.com/w/cpp/types/offsetof) was available before C++11. Judging by its `<cstddef>` header, it is also more of a C than C++ thing, but I am only saying that to explain why I did not leave `C++` in the title.

src/cbdata.cc
yadij: Done in 5b44429.

src/cbdata.cc
```diff
@@ -89,6 +90,14 @@ class cbdata
     {}
     ~cbdata();
 
+    static cbdata *FromUserData(const void *p) {
+#if WITH_VALGRIND
+        return cbdata_htable.at(p);
+#else
+        return (cbdata *) (((char *) p) - offsetof(cbdata, data));
```
rousskov, line 97: If you get a chance, please use appropriate C++ casts here.

src/cbdata.cc
```diff
@@ -297,12 +292,7 @@ cbdataRealFree(cbdata *c, const char *file, const int line)
 void *
 cbdataInternalFree(void *p, const char *file, int line)
 {
-    cbdata *c;
-#if WITH_VALGRIND
-    c = cbdata_htable.at(p);
-#else
-    c = (cbdata *) (((char *) p) - cbdata::Offset);
-#endif
+    auto *c = cbdata::FromUserData(p);
```
rousskov, line 295: Some of these new/changed declarations can probably be tightened up because the value of `c` never changes:

```suggestion
    const auto c = cbdata::FromUserData(p);
```

rousskov (commented):
I do not insist on further polishing here.

src/cbdata.cc
yadij: Most of them need to be non-const for addHistory() use when USE_CBDATA_DEBUG is enabled. The one which does not is updated in 207b801.

src/cbdata.cc
yadij: Done in 207b801

yadij:
Clearing for merge when the build checks pass.

src/cbdata.cc
rousskov: The proposed `const auto c = ...` change does not affect addHistory(). Again, the value of `c` never changes. You are probably thinking about the value of `*c` which the proposed expression does not make constant.

PR #811, opened 2021-05-04
API to separate Config.x users/parsers from Config.y details

rousskov:
The new API avoids several kinds of unwanted code dependencies:

* Config.x users do not need to know the type of any Config.y component.
  Forward declarations of Config.y types are sufficient.

* Config.x parsers do not need to know about Config.y parsers. Config.x
  parsers can also be hidden from Config.x users.

* The centralized parsing code does not need to contain Config.x parsing
  code (for any x) and does not maintain "generic" parser registration.
  The compiler/linker do that for us _without_ losing C++ type safety.

Correct API implementation may also help separate the active Config
object from the being-parsed configuration, to eventually support safe
"hot" reconfiguration.

This change does not convert any existing Config fields and does not
imply a policy that existing fields must be converted when touched.

The API was tested on a typical new Config.x component (to be merged
separately). It is likely to evolve, especially when support for
multi-directive components is added.

Activated by using TYPE names with two colons (e.g., TYPE: Security::X).

### DIFF
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -662,7 +662,7 @@ ParseDirective(T &raw, ConfigParser &parser)
 
     // TODO: parser.openDirective(directiveName);
     Must(!raw);
-    raw = ConfigComponent<T>::Parse(parser);
+    raw = Configuration::Component<T>::Parse(parser);
     Must(raw);
     parser.closeDirective();
 }
@@ -678,7 +678,7 @@ DumpDirective(const T &raw, StoreEntry *entry, const char *name)
 
     entry->append(name, strlen(name));
     SBufStream os;
-    ConfigComponent<T>::Print(os, raw);
+    Configuration::Component<T>::Print(os, raw);
     const auto buf = os.buf();
     if (buf.length()) {
         entry->append(" ", 1);
@@ -692,7 +692,7 @@ template <typename T>
 static void
 FreeDirective(T &raw)
 {
-    ConfigComponent<T>::Free(raw);
+    Configuration::Component<T>::Free(raw);
 
     // While the implementation may change, there is no way to avoid zeroing.
     // Even migration to a proper SquidConfig class would not help: While
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -24,6 +24,7 @@
 #include "base/RunnersRegistry.h"
 #include "cache_cf.h"
 #include "CachePeer.h"
+#include "ConfigOption.h"
 #include "ConfigParser.h"
 #include "CpuAffinityMap.h"
 #include "DiskIO/DiskIOModule.h"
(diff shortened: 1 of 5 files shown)


src/cache_cf.cc
```diff
@@ -633,6 +634,75 @@ parseConfigFile(const char *file_name)
...
+    return bool(raw);
+}
+
+/// Sets the given raw SquidConfig data member.
+/// Extracts and interprets parser's configuration tokens.
+template <typename T>
+static void
+ParseDirective(T &raw, ConfigParser &parser)
+{
+    if (SawDirective(raw))
```
yadij, line 660: We have quite a few directives that can be repeated over multiple lines. For this template to replace the legacy API it needs to support those. Perhapse shift this check into Config::Component<T> instead?

src/cache_cf.cc
```diff
@@ -633,6 +634,75 @@ parseConfigFile(const char *file_name)
     }
 }
 
+/*
+ * The templated functions below are essentially ConfigParser methods. They are
+ * not implemented as such because our generated code calling them is the only
+ * code that can instantiate implementations for each T -- we cannot place these
+ * definitions into ConfigParser.cc unless cf_parser.cci is moved there.
+ */
```
yadij, line 642: Since these are static templates they should be added to src/config/gadgets.h and included by cf_parser.cci

yadij (changes requested):
This should be the start of our properly designed config parsing. As such please add src/config/libconfig.la and namespace config with these new pieces inside it.

src/cache_cf.cc
rousskov: I cannot find a logical connection between "these are static templates" and "they should be added to src/config/gadgets.h and included by cf_parser.cci". If anything, the opposite (i.e. "should _not_ be ...") statement would be logically sound. Unfortunately, I do not know what problem you are trying to fix here. Moreover, I see no in-scope problems worth fixing here. Thus, it is difficult for me to suggest specific fixes to this change request that we may be able to agree on.

These templates should not be added to any C++ header file because they are not shared across translation units -- only one translation unit needs them.

Moreover, it is conceptually wrong to create header files just to be included in a _generated_ .cci file (which is itself included into a .cc file). We should keep cf_parser.cci as small/simple as possible, containing code that has to be generated and virtually nothing else. Everything that should not be generated goes into cache_cf.cc which includes cf_parser.cci (or into other manually-written source files). Our cf_parser.cci does not contain `#include` statements, and we should keep it that way (until some `#include` statements must be generated).

src/cache_cf.cc
rousskov: > We have quite a few directives that can be repeated over multiple lines. For this template to replace the legacy API it needs to support those

Yes, we will support more complex duplicate rejection logic. The SawDirective() TODO acknowledges that future work. However, we should not add that support in this PR because such support is not needed by the new API users, because proper support is difficult to design, and because designing APIs without specific users often leads to poor results.

> Perhapse shift this check into Config::Component instead?

We will make that and other complex decisions when we have to. Today, Config::Component embodies `TYPE:`-driven logic. Shifting this check into Config::Component today would imply that `TYPE` determines whether a particular directive can be repeated. That implication may be false. I would not be surprised if future duplicate handling will be driven by code _generated_ from some new cf.data.pre metadata other than `TYPE`!

I am sure ParseDirective() will change from its current simple form as we add support for more configuration cases. There is nothing wrong with that anticipated change as long as the current code addresses our current needs (and no simple changes can make it better). Trying to predict what other cases will need is not going to work well here because that decision space is very complex. Let's keep it simple while we can.

rousskov (commented):
>  please add src/config/libconfig.la and namespace config with these new pieces inside it.

Please do not force me to create a new library with nothing but a single tiny _header_ file in it. That library should be created by migrating the generic configuration code from `src/`.

Please also do not force me to migrate generic configuration code into a new dedicated library. That should be done, but it is a complex undertaking that goes way beyond this PR scope.

Even the name for the new namespace is difficult to pick because we want the name to start with a capital letter (for consistency with all other Squid namespaces), `Config` is already taken by a popular global, `Cfg` or `Conf` abbreviation may not be obvious to some readers, and `Configuration` is too long. I added `Configuration` in branch commit 8db4033, but I can rename to one of those shorter variants if you think one of them will work better then a spelled-out version.

src/cache_cf.cc
yadij: Being static templates they can be added to any .h file.

In context of my request to create src/config/ library for this polished API, they should have a .h there away from the old legacy cruft.

Under our current coding style either each one gets a file named after itself, or being too complex for forward.h they go into gadgets.h. If you can think of a filename descriptive of the group they can use that. I am not fixed on any particular name.

Since only the generated code in .cci uses these templates the best place to include is there. I will accept the cache_cf.cc if you insist though.

rousskov: > Being static templates they can be added to any .h file.

Many things can be added to any .h file. Most such things should not be. This is not an argument that can support or justify placing something into a header file!

> In context of my request to create src/config/ library for this polished API, they should have a .h there away from the old legacy cruft.

These functions should not be in a header file, as I tried to explain in my previous [response](https://github.com/squid-cache/squid/pull/811#discussion_r625812345). I have no new arguments, but if you need me to clarify the existing ones (which your response completely ignored!), please ask questions or point to flaws in my logic.

Yes, it is nice to separate new code from the "old cruft" on a source file basis, but _not_ at the expense of lowering the quality of that new code. FWIW, most of this new code can still be easily detected in this case by searching for Configuration::Component.

> Under our current coding style either each one gets a file named after itself

I see no coding style rules that require these local functions to be in a header file (or even a .cc file named after them). If there are such rules, they should probably adjusted/clarified!

> Since only the generated code in .cci uses these templates the best place to include is there.

Virtually any manually written code that can be placed outside of generated code should be placed outside, for many reasons (please let me know if I should detail them). Whether the generated code is the only user of that manually written code has no bearing on this rule of thumb.

> I will accept the cache_cf.cc if you insist though.

Yes, I do "insist" that placing these functions in a header in this PR would be a mistake. I do so not because I can insist on arbitrary things, but because I have outlined solid reasons to support my position and have been given no reasons that would justify the move (except, perhaps, the "place new code into new files" principle that I worked hard to extract and that cannot be satisfied without violating more important principles here).

yadij (approved):

rousskov:
> It just seems IMO reasonable to do the library creation here, even if it is only headers and have that followup move the existing objects into it.

It can be a reasonable thing to do in one's own PR, and asking for it can be considered reasonable during a superficial review, but a reviewer should not insist on it in this particular PR: This tiny PR was already forced out of a very different project. This PR overheads would become very significant if a proper migration to a convenience library is forced into its scope. The migration will also obscure the true changes introduces by this PR, which was the primary justification for this PR isolation! It is unreasonable to insist on a migration in this PR. Thank you for withdrawing that change request.

> > `Config` is already taken by a popular global

> "Config" is only used within other namespaces, not at the top level where this is defined - so there should be no problem with any properly written code. I am minded to use "namespace Config" and update the SourceLayout details.

FYI:  You may be thinking about `Config` sub-namespaces irrelevant to this part of my argument. I was talking about a popular global called `Config`:

```
src/SquidConfig.h:extern SquidConfig Config;
```

You can see an example of the namespace/global conflict outcome at https://godbolt.org/z/3hGsj6W3M

PR #814, opened 2021-05-06
Simplified Http::Message::parse()

MegaManSec:
size_t hdr_len cannot be negative, and sanityCheckStartLine() already
rejects zero values. We now use (and clearly document) the latter fact.

Also replaced a level-1 warning about "Too large" headers with a level-3
debugging message because huge headers is not a Squid problem, and the
problem should already be visible in access.logs records.

### DIFF
--- a/src/http/Message.cc
+++ b/src/http/Message.cc
@@ -85,8 +85,8 @@ Http::Message::parse(const char *buf, const size_t sz, bool eof, Http::StatusCod
     // find the end of headers
     const size_t hdr_len = headersEnd(buf, sz);
 
-    if(hdr_len > Config.maxReplyHeaderSize || (hdr_len == 0 && sz > Config.maxReplyHeaderSize))
-        debugs(58, 3, "Too large reply header (" << hdr_len << " > " << Config.maxReplyHeaderSize << ")");
+    if (hdr_len > Config.maxReplyHeaderSize || (hdr_len == 0 && sz > Config.maxReplyHeaderSize)) {
+        debugs(58, 3, "input too large: " << hdr_len << " or " << sz << " > " << Config.maxReplyHeaderSize);
         *error = Http::scHeaderTooLarge;
         return false;
     }
--- a/src/http/Message.cc
+++ b/src/http/Message.cc
@@ -85,6 +85,12 @@ Http::Message::parse(const char *buf, const size_t sz, bool eof, Http::StatusCod
     // find the end of headers
     const size_t hdr_len = headersEnd(buf, sz);
 
+    if(hdr_len > Config.maxReplyHeaderSize || (hdr_len == 0 && sz > Config.maxReplyHeaderSize))
+        debugs(58, 3, "Too large reply header (" << hdr_len << " > " << Config.maxReplyHeaderSize << ")");
+        *error = Http::scHeaderTooLarge;
+        return false;
+    }
+
     // sanity check the start line to see if this is in fact an HTTP message
     if (!sanityCheckStartLine(buf, hdr_len, error)) {
         // NP: sanityCheck sets *error and sends debug warnings on syntax errors.
@@ -95,20 +101,7 @@ Http::Message::parse(const char *buf, const size_t sz, bool eof, Http::StatusCod
         return false;
     }
 
-    if (hdr_len > Config.maxReplyHeaderSize || (hdr_len <= 0 && sz > Config.maxReplyHeaderSize)) {
-        debugs(58, DBG_IMPORTANT, "Too large reply header (" << hdr_len << " > " << Config.maxReplyHeaderSize);
-        *error = Http::scHeaderTooLarge;
-        return false;
-    }
-
-    if (hdr_len <= 0) {
-        debugs(58, 3, "failed to find end of headers (eof: " << eof << ") in '" << buf << "'");
-
-        if (eof) // iff we have seen the end, this is an error
-            *error = Http::scInvalidHeader;
-
-        return false;
-    }
+    assert(hdr_len > 0); // sanityCheckStartLine() rejects unparsed headers
 
     const int res = httpMsgParseStep(buf, sz, eof);
 
--- a/src/http/Message.cc
+++ b/src/http/Message.cc
@@ -101,7 +101,7 @@ Http::Message::parse(const char *buf, const size_t sz, bool eof, Http::StatusCod
         return false;
     }
 
-    assert(hdr_len > 0); // sanityCheckStartLine() rejects unparsed headers
+    assert(hdr_len > 0); // sanityCheckStartLine() rejects buffers that cannot be parsed
 
     const int res = httpMsgParseStep(buf, sz, eof);
 
@@ -289,4 +289,3 @@ Http::Message::firstLineBuf(MemBuf &mb)
 {
     packFirstLineInto(&mb, true);
 }
-
--- a/src/http/Message.h
+++ b/src/http/Message.h
@@ -123,7 +123,7 @@ protected:
     /**
      * Validate the message start line is syntactically correct.
      * Set HTTP error status according to problems found.
-     * This method must return false if we do do not have valid headers.
+     * This method must return false if it is not passed parsed headers.
      *
      * \retval true   Status line has no serious problems.
      * \retval false  Status line has a serious problem. Correct response is indicated by error.
--- a/src/http/Message.h
+++ b/src/http/Message.h
@@ -123,7 +123,6 @@ protected:
     /**
      * Validate the message start line is syntactically correct.
      * Set HTTP error status according to problems found.
-     * This method must return false if it is not passed parsed headers.
      *
      * \retval true   Status line has no serious problems.
      * \retval false  Status line has a serious problem. Correct response is indicated by error.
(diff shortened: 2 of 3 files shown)


yadij:
OK to test

MegaManSec:
Upon further invegtiation, sanityCheckStartLine will return false if hdr_len is 0 and thus I have removed another check.

yadij (approved):

src/http/Message.cc
```diff
@@ -95,21 +95,12 @@ Http::Message::parse(const char *buf, const size_t sz, bool eof, Http::StatusCod
         return false;
     }
 
-    if (hdr_len > Config.maxReplyHeaderSize || (hdr_len <= 0 && sz > Config.maxReplyHeaderSize)) {
+    if (hdr_len > Config.maxReplyHeaderSize || sz > Config.maxReplyHeaderSize) {
```
rousskov, line 98: This change introduces a bug. The old code was `(A || B)`. The proposed code is `(A || C)`, which adds a wrong condition C instead of the "correct" (but impossible to reach) condition B.

Legend:

* A: [we have found the end of the header and] the header is larger than the header size limit
* B: we have not found the end of the header but input is already larger than the header size limit
* C: input is larger than the header size limit

```suggestion
    assert(hdr_len > 0); // sanityCheckStartLine() rejects unparsed headers
    if (hdr_len > Config.maxReplyHeaderSize) {
```

N.B. The suggested assertion is essential here and also requires sanityCheckStartLine() API documentation changes in http/Message.h. Nothing in the current sanityCheckStartLine() description says that the method must treat unparsed headers as insane (i.e. return false). While you are fixing that description, please also replace sanityCheckStartLine() description in src/HttpReply.h with an `/* Http::Message API */` comment. This is how we document overloaded methods in Squid.

This PR bug fix raises a second question -- does sanityCheckStartLine() or anybody else reject large unparsable input? Checking (and, if needed, fixing) that should be done before this PR goes in IMO. I did not check.

rousskov (changes requested):

src/http/Message.cc
MegaManSec: Thank you for your suggestions and picking up the incorrect change.

I will incorporate the changes you have suggested.

> does sanityCheckStartLine() or anybody else reject large unparsable input?
The answer to this question is: no. The only check in HttpReply::sanityCheckStartLine is for `hdr_len` being too small(`hdr_len < (size_t)(protoPrefix.psize() + 4)`). However, I am not quite sure what you are suggesting in relation to rejecting large _unparsable_ input – I am confused only about what you consider "unparsable".

If you are suggesting moving the `Config.maxReplyHeaderSize` check into sanityCheckStartLine() and/or headersEnd(), then I support this as it would free up some processing time.

This would make code like Http::One::Parser::grabMimeBlock more 'correct' AFAICT, because it does the following:
```
        if (SBuf::size_type mimeHeaderBytes = headersEnd(buf_, containsObsFold)) {
[...]
        } else { // headersEnd() == 0
            if (buf_.length()+firstLineSize() >= limit) {
                debugs(33, 5, "Too large " << which);
                parseStatusCode = Http::scHeaderTooLarge;
                parsingStage_ = HTTP_PARSE_DONE;

```

and this function is literally called like this (in once instance):
`if (!grabMimeBlock("Response", Config.maxReplyHeaderSize))`

Let me know your thoughts.

rousskov: > I am confused only about what you consider "unparsable".

I should not have used that word, sorry. I should have said something like "endless". Here, all those words essentially mean that headersEnd() returns zero. That is, we cannot find the end of headers. The end of headers may be coming later or perhaps it will never come, we do not know. What we do know is that waiting longer or reading more will not change the ultimate outcome -- the header will be rejected.

As for the right solution, I suspect that sanityCheckStartLine() should be refactored into a method that calls headersEnd() internally and then checks Config.maxReplyHeaderSize (among other things). Perhaps that refactored method would become the Message::parse() itself. Finding the right solution requires more thinking/experimentation than I have the time for right now, sorry.

I would keep the low-level headersEnd() oblivious to various high-level checks/limits, at least in this PR.

yadij: FYI: this parse() method is deprecated. It is being replaced by the Parser class logic. Trying to polish this code up is a waste of time. Please concentrate scope down to major bugs or things compilers complain about directly.

MegaManSec:
>This PR bug fix raises a second question -- does sanityCheckStartLine() or anybody else reject large unparsable input? Checking (and, if needed, fixing) that should be done before this PR goes in IMO. I did not check.

I have deliberately left this part out of this PR. I do not think it should be sanityCheckStartLine()'s job to check for too-large headers, and it should be the caller's job – after all, that function just sanity check's the header.

With respect to the original change in [3c9eabaf2792e51372a0958a79263be49d232745](https://github.com/squid-cache/squid/pull/814/commits/3c9eabaf2792e51372a0958a79263be49d232745), it begs the question whether this (impossible to reach) debugging message is of use. `sanityCheckStartLine` will log `Too small reply header (0 bytes)` in this case, but will not say that header parsing failed.

src/http/Message.h
```diff
@@ -123,6 +123,7 @@ class Message : public RefCountable
     /**
      * Validate the message start line is syntactically correct.
      * Set HTTP error status according to problems found.
+     * This method must return false if it is not passed parsed headers.
```
yadij, line 126: This is wrong.

This method is a legacy code check that the first-line of a message fits HTTP syntax. It is a preparation step before passing the message to legacy HTTP parse. As such if the headers are already parsed successfully by any mechanism this method is irrelevant.

yadij (changes requested):

src/http/Message.h
MegaManSec: OK. Then, I do not know what this meant:
>The suggested assertion is essential here and also requires sanityCheckStartLine() API documentation changes in http/Message.h. Nothing in the current sanityCheckStartLine() description says that the method must treat unparsed headers as insane (i.e. return false

rousskov:
```C++
/**
 * Validate the message start line is syntactically correct.
 * Set HTTP error status according to problems found.
 *
 * \retval true   Status line has no serious problems.
 * \retval false  Status line has a serious problem. Correct response is indicated by error.
 */
virtual bool sanityCheckStartLine(const char *buf, const size_t hdr_len, Http::StatusCode *error) = 0;
```

I think there is a slight misunderstanding here due to vague terminology and my sloppy use of that terminology ("parsed", "headers", "status line", hdr_len, etc.). To address my concern without triggering Amos' (correct) objection, the added comment may say something like

```
* Zero-size hdr_len is treated as a serious problem.
```

The remaining method description already documents what happens when the method detects a serious problem.

rousskov (dismissed):
Thank you for addressing my primary concerns. Please see the [other](https://github.com/squid-cache/squid/pull/814#discussion_r629420256) change request for the comment fix.

src/http/Message.cc
```diff
@@ -85,6 +85,12 @@ Http::Message::parse(const char *buf, const size_t sz, bool eof, Http::StatusCod
     // find the end of headers
     const size_t hdr_len = headersEnd(buf, sz);
 
+    if(hdr_len > Config.maxReplyHeaderSize || (hdr_len == 0 && sz > Config.maxReplyHeaderSize))
+        debugs(58, 3, "Too large reply header (" << hdr_len << " > " << Config.maxReplyHeaderSize << ")");
```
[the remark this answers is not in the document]
MegaManSec: >this may be a request rather than a reply

My understanding is that this is not true (or I am misinterpreting what you have said). If it is, I don't see the point in checking hdr_len against Config.maxReplyHeaderSize here.

Either way, I have incorporated your changes.

src/http/Message.cc
```diff
@@ -95,20 +101,7 @@ Http::Message::parse(const char *buf, const size_t sz, bool eof, Http::StatusCod
...
-
-    if (hdr_len <= 0) {
-        debugs(58, 3, "failed to find end of headers (eof: " << eof << ") in '" << buf << "'");
-
-        if (eof) // iff we have seen the end, this is an error
-            *error = Http::scInvalidHeader;
-
-        return false;
-    }
+    assert(hdr_len > 0); // sanityCheckStartLine() rejects unparsed headers
```
yadij, line 104: To be specific sanityCheckStartLine() rejects buffers that do not start with RFC7230 section 3.1 "Start Line" syntax. That includes HTTP/0.9, HTTP/1.1, ICY and some other protocols with almost identical syntax (eg ICAP, SMTP)


```suggestion
    assert(hdr_len > 0); // sanityCheckStartLine() rejects buffers that cannot be parsed
```

yadij (changes requested):

src/http/Message.cc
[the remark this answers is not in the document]
rousskov: >> this may be a request rather than a reply

> My understanding is that this is not true

You are probably focusing on how this _direction-agnostic_ method is used in the current master code. I do not know whether it is currently used exclusively for responses and, more importantly, I do not think we should care about (or rely upon) the current situation in this context. By design, Http::Message can be a request or a response.

Thank you for incorporating my changes.

yadij:
Cleared for merge under the 10-days no objections criteria.

PR #820, opened 2021-05-10
Do not process empty filenames using CWD command. (#820)

MegaManSec:
Various different functions call ftpSendCwd, but there is no check in this function to ensure filepath is not NULL, for example, via hackShortcut(), which even explicitly deals with the case of filepath being NULL (but does not change it).

### DIFF
--- a/src/clients/FtpGateway.cc
+++ b/src/clients/FtpGateway.cc
@@ -1462,6 +1462,9 @@ ftpSendCwd(Ftp::Gateway * ftpState)
 
     path = ftpState->filepath;
 
+    if(!path)
+        return;
+
     if (!strcmp(path, "..") || !strcmp(path, "/")) {
         ftpState->flags.no_dotdot = 1;
     } else {


MegaManSec:
P.S. This also silences a warning using GCC11.

rousskov (changes requested):
I agree that we must not call strcmp() and snprintf(%s) with a nil parameter.

AFAICT, ftpSendCwd() is supposed to send a CWD command. If the caller wants to send that command, but the path is unknown, then this is a Squid bug. We should not hide that bug by pretending to send the command while doing nothing. We should assert instead.

If there are known code paths that reach this code with a nil path, then we should fix those code paths before adding an assertion.

MegaManSec:
I am happy to change this to an assertion (or simply closing the PR, in fact).


The codepath suggested was via Ftp::Gateway::hackShortcut (AKA: Ftp::Client::SENT_RETR)

ftpSendRetr should already assert if filename is null:
```
    assert(ftpState->filepath != NULL);
    snprintf(cbuf, CTRL_BUFLEN, "RETR %s\r\n", ftpState->filepath);
    ftpState->writeCommand(cbuf);
    ftpState->state = Ftp::Client::SENT_RETR;
```

I assume that means the check in Ftp::Gateway::hackShortcut:
```
        if (pathcomps == NULL && filepath != NULL)
            old_filepath = xstrdup(filepath);
    }
```
is impossible, but I do not know – and I do not have the capability of testing any of this FTP code with any certainty or reproducability, anyway. And I don't really want to play around with or suggest changes to the FTP code.


Other functions which call (or inadvertently call) ftpSendCwd, like ftpReadMkdir, do not have checks/asserts for 'filepath'. ftpReadMkdir is called after ftpSendMkdir, which passes `filepath` to snprintf so would probably cause an issue there before it ever touched ftpSendCwd.


FWIW, there are checks in most other functions like this:
```
    if (ftpState->filepath != NULL) {
        /* Plain file upload */
        snprintf(cbuf, CTRL_BUFLEN, "STOR %s\r\n", ftpState->filepath);
```

but this is no consistent (perhaps for good reason).

I can at least visually see a path to make this bug happen.
By sending a PUT request: https://github.com/squid-cache/squid/blob/master/src/clients/FtpGateway.cc#L361 and then via ftpRestOrList (via https://github.com/squid-cache/squid/blob/master/src/clients/FtpGateway.cc#L1722) with `ftpState->typecode == 'D'`. If we did not send a PUT request, we would instead go to the function ftpSendNlst – which does have a check to ensure filepath is not null – https://github.com/squid-cache/squid/blob/master/src/clients/FtpGateway.cc#L2174)


FYI: the intent of this patch it to mute a warning which has popped up upon updating clang-12 (which seems to imply some static analysis going on but I don't know why, and this along with the bug 5126 are the only instances of 'new' warnings). I do not know whether that warning will ever happen in a production version of clang, though, and there are no other complaints about ftpSendCwd.

If you do not think it is worth exploring further, then I support closing the PR.

rousskov:
> If you do not think it is worth exploring further, then I support closing the PR.

IMO, getting to the bottom of these issues is valuable, but I speculate that there are more valuable things that you and I should be doing instead :-).

Unfortunately, I do not know enough about the FTP code to effortlessly guide you through its state machine. I can only ask basic questions and detect basic flaws in arguments/code. Moreover, I doubt any human can reliably work with the current implementation of the FTP state machine. There are just too many implicit invariants like the `filepath` being set when used. Thus, I cannot encourage anybody to work on FTP code, especially when they are not intimately familiar with it.

If you would rather work on something else, then I suggest two reasonable options for going forward with this PR:

* Add assert()s before dereferencing currently-unchecked filepath or passing it to a library function. Check all filepath uses! Asserting is not ideal, but it is much better than a dangerous UB, on several levels (including the fact that it addresses anticipated compiler warnings). Checking all immediate uses is easy and would result in a reasonable (but still very limited) PR scope.
* Close this PR without further work because it requires too much (from its author point of view) work to address reasonable (from the reviewer point of view) change requests.

Your call. I would support any of the above decisions.

MegaManSec:
Closing for many of the reasons above. If anybody in the future is able to truly trigger this bug, I will buy them a beer[0].

[0] flavour, brand and alcohol content not guaranteed.

PR #822, opened 2021-05-13
Fix --with-valgrind-debug build broken by commit 02f5357

rousskov:
error: cbdata_htable was not declared in this scope


kinkie (approved):
LGTM

PR #823, opened 2021-05-14
Rename MasterXaction to ...

yadij:
* Move Xaction.* files into libbase where they do not conflict with
   any library header files.

* Fix one const correctness bug in NewServer() API


rousskov (changes requested):
I have not review the details, but I see three high-level mostly-independent showstoppers that we should resolve first.

1. We need to agree on what the former MasterXaction class represents before we can decide on the best new name for that class. The documentation adjustments in the PR do not resolve the current disagreement, and the PR description is silent about it.

2. We should not introduce a general purpose `Squid` namespace without a very good reason because, once introduced, virtually everything we write will have to be in that namespace. For the same reason, the introduction of that `Squid` namespace cannot be used to solve any "scoping" or naming problem within Squid. Hence, it is most likely outside this PR scope. The PR should not be silent about the reasons behind that introduction. (The existing use of `Squid` namespace for a few compatibility wrappers was a mistake, but that exceptional use is irrelevant here.)

3. The former MasterXaction class does not belong to `src/base/`. That directory is shaping up to be used for "development" or "code plumbing" classes -- architectural elements that primary Squid code is built with or upon. In a sense, base things are what we would prefer to find in the programming language or its standard library. The former MasterXaction class is a high-level Squid-specific concept unrelated to code architecture in that sense.

yadij:
> I have not review the details, but I see three high-level mostly-independent showstoppers that we should resolve first.

Beyond the NewServer API change and file move, the details are just a few grep and sed for now.

> 
>     1. We need to agree on what the former MasterXaction class represents before we can decide on the best new name for that class. The documentation adjustments in the PR do not resolve the current disagreement, and the PR description is silent about it.
> 

Aye. I think we had a fairly good agreement on what the *concept* you called "master transaction" was and how it worked as a concept and guideline for code design choices.

However, trying to plan out and implement PR #716 showed that the choice of representing that concept with a single class was a bad choice. The "master transaction" is clearly needing to account for sub-transactions in its code representation. It is better kept as a concept where it can be represented by a collection of linked object instances.

IMO this class is still a key part of "master transaction" representation, but not itself the complete state of it as previously envisioned. PR #716 shows the "master" may need to consist of several of this class'es instances linked together.

What we have today in "class MasterXaction" is a scaffold to store the state from a "TCP transaction (aka Comm::Connection tcpClient)" and make it available to the initiators of some child transactions (eg HTTP) but not others (eg TLS), which use that TCP connection as a transport for their own transactions.


>     2. We should not introduce a general purpose `Squid` namespace ...

Agreed. This is temporary with the expectation of the discussion ahead for (1) making namespace unnecessary or picking another.

>     3. The former MasterXaction class does not belong to `src/base/`.

Due to some libraries having files named Xaction.h it is not possible to leave the current named file in src/. For lack of a good place we have base/ for now. This will resolve once the (1) class naming choice is agreed and file given its proper name.

rousskov:
> > 1. We need to agree on what the former MasterXaction class represents before we can decide on the best new name for that class. The documentation adjustments in the PR do not resolve the current disagreement, and the PR description is silent about it.

> Aye. I think we had a fairly good agreement on what the _concept_ you called "master transaction" was and how it worked as a concept and guideline for code design choices.

Glad to hear that.
 
> However, trying to plan out and implement PR #716 showed that the choice of representing that concept with a single class was a bad choice. 

While that PR had problems, I disagree with your conclusion above. The class currently called MasterXaction and the concept behind "master transaction" are not the cause of that PR problems.

> The "master transaction" is clearly needing to account for sub-transactions in its code representation. 

Yes, to some extent. The MasterXaction class already accounts for the from-client connection, but we need to account for more than that, of course. It remains to be seen how each transaction (i.e. what you called sub-transaction) should be represented. We need to be mindful of reference loops and excessive dependencies, but some kind of representation has to be there.

> It is better kept as a concept where it can be represented by a collection of linked object instances.

That "collection of linked object instances" should be represented by a class because all those "linked object instances" do have something in common and that something has to be stored somewhere. That class is MasterXaction today. We should rename it because of the Naming Initiative, but its essence remains the same.

> IMO this class is still a key part of "master transaction" representation, but not itself the complete state of it as previously envisioned. 

I see reasons to represent the concept with a class. I do not see reasons to avoid that representation.

> PR #716 shows the "master" may need to consist of several of this class'es instances linked together.

I do not think that PR showed anything like that.
 
> What we have today in "class MasterXaction" is a scaffold to store the state from a "TCP transaction (aka Comm::Connection tcpClient)" and make it available to the initiators of some child transactions (eg HTTP) but not others (eg TLS), which use that TCP connection as a transport for their own transactions.

I am not sure _why_ you are saying the above. There are many deficiencies in the current code, but addressing them does not require removing MasterXaction. It requires adding more info to that class and creating fewer MasterXaction objects, among other things.

> > 2. We should not introduce a general purpose `Squid` namespace ...
 
> Agreed. 

I am glad to hear that.

> This is temporary with the expectation of the discussion ahead for (1) making namespace unnecessary or picking another.

If this PR contains temporary to-be-revised-before-merging code then please mark it as a Draft PR. If a similar need for a temporary namespace arises in the future, I would recommend naming the temporary namespace into something that shows its temporary nature, like ToBeDecided or Draft. Nothing in this PR indicated the temporary nature of this namespace to me, and that is wrong because I would not have wasted time on it if I knew it is a temporary kludge. Since you have explained that this is a temporary kludge, it is no longer important (for now). We will need to remove or rename that namespace before merging this PR.

> > 3. The former MasterXaction class does not belong to `src/base/`.
 
> Due to some libraries having files named Xaction.h it is not possible to leave the current named file in src/. For lack of a good place we have base/ for now. This will resolve once the (1) class naming choice is agreed and file given its proper name.

Hm... I must be missing something, but I see no reason to rename MasterXaction until we agree what the new name should be. It is especially strange to see that class renamed to something that requires moving its file into another directory (where it does not belong). So much work for no gain (that I can see)! Since you have explained that this is a temporary kludge, it is no longer important (for now), but it does creates noise and distractions. We will most likely need to move the header file into a different directory before merging this PR.

yadij:
Can we agree that the thing initiated by TcpAcceptor is a "client connection" ?

As such, the object created by TcpAcceptor and filled out with pointers about the "client connection" IMO should be renamed as class ClientConnction now.

Objections ...?

rousskov:
> Can we agree that the thing initiated by TcpAcceptor is a "client connection" ?

Yes, I hope we can, for some definition of "client connection". That definition has to include data connections opened by FTP clients, for example. 

This "client connection" thing is a valid concept. Currently, Squid does not have an object that represents that concept. The concept is currently represented by a collection of objects, each a _different_ concept. This cannot be fixed by renaming because there is nothing (as in "no thing" or "no object") to rename.

> As such, the object created by TcpAcceptor and filled out with pointers about the "client connection" IMO should be renamed as class ClientConnction now.
> 
> Objections ...?

Today, TcpAcceptor creates several objects. None of those objects have a type that can be called ClientConnection:

* TcpAcceptor creates a Comm::Connection object. Comm::Connection represents any transport connection (at least), including to-Server connections. We cannot rename this type to ClientConnection.
* TcpAcceptor creates a MasterXaction object. A single client connection may correspond to many master transactions, each needing its own MasterXaction object. We cannot rename this type to ClientConnection. BTW, technically, the MasterXaction object should not even be created by TcpAcceptor (for FTP data connections); that abuse of MasterXaction to carry freshly accepted FTP data connection details is an implementation bug/deficiency.
* For HTTP client connections, there is a client connection manager (ConnStateData+), but those manager objects are created later, via httpAccept() and friends, long after TcpAcceptor completed its work associated with their client connection. Renaming ConnStateData to ClientConnection would be inline with your reasoning, but that renaming has its own hurdles, and, more importantly, that renaming is not what this PR is about.

N.B. I am currently ignoring the problems with the ClientConnection _spelling_ because they are secondary to the above problem.

PR #825, opened 2021-05-15
Bug 5129 pt1: remove Lock use from HttpRequestMethod

yadij:
Removes the need for a custom assignment operator with a questionable
implementation, addressing compiler and static analysis warnings.

### DIFF
--- a/src/http/RequestMethod.h
+++ b/src/http/RequestMethod.h
@@ -23,7 +23,7 @@ class SquidConfig;
  * It has a runtime extension facility to allow it to
  * efficiently support new methods
  */
-class HttpRequestMethod : public RefCountable
+class HttpRequestMethod
 {
 public:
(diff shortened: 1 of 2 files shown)


kinkie (approved):
Nice! You make it seem easy 🤣

yadij:
The main problem is that Lock allows copy in the first place. That is still not resolved and my latest attempt to fix there shows several other classes having the same issue as HttpRequestMethod which may actually need to stay RefCountable.

rousskov (changes requested):
Please remove HttpRequestMethodPointer.

P.S. I did not see this PR when commenting on bug 5129.

rousskov (approved):
Thank you for polishing this PR up.

I fixed the PR description: The assignment operator is still needed/used, we just do not need to customize it. It is that explicit declaration and customization that triggered problems, not the operator itself. Please adjust further if needed.

PR #827, opened 2021-05-19
ICP traffic logging

yadij:
Initially just received packets which pass basic validity
and origin filters.

### DIFF
--- a/src/icp_v2.cc
+++ b/src/icp_v2.cc
@@ -616,24 +616,6 @@ icpHandleIcpV2(int fd, Ip::Address &from, char *buf, int len)
     }
 }
 
-#ifdef ICP_PKT_DUMP
-static void
-icpPktDump(icp_common_t * pkt)
-{
-    Ip::Address a;
-
-    debugs(12, 9, "opcode:     " << std::setw(3) << pkt->opcode  << " " << icp_opcode_str[pkt->opcode]);
-    debugs(12, 9, "version: "<< std::left << std::setw(8) << pkt->version);
-    debugs(12, 9, "length:  "<< std::left << std::setw(8) << ntohs(pkt->length));
-    debugs(12, 9, "reqnum:  "<< std::left << std::setw(8) << ntohl(pkt->reqnum));
-    debugs(12, 9, "flags:   "<< std::left << std::hex << std::setw(8) << ntohl(pkt->flags));
-    a = (struct in_addr)pkt->shostid;
-    debugs(12, 9, "shostid: " << a );
-    debugs(12, 9, "payload: " << (char *) pkt + sizeof(icp_common_t));
-}
-
-#endif
-
 void
 icpHandleUdp(int sock, void *)
 {
@@ -680,15 +662,25 @@ icpHandleUdp(int sock, void *)
         debugs(12, 4, "icpHandleUdp: FD " << sock << ": received " <<
                (unsigned long int)len << " bytes from " << from);
 
-#ifdef ICP_PACKET_DUMP
-
-        icpPktDump(buf);
-#endif
-
         if ((size_t) len < sizeof(icp_common_t)) {
             debugs(12, 4, "icpHandleUdp: Ignoring too-small UDP packet");
             break;
         }
+
+        auto pkt = reinterpret_cast<icp_common_t *>(buf);
+        struct in_addr a;
+        a.s_addr = pkt->shostid;
+        debugs(12, 2, "ICP Client remote=" << from << " FD " << sock);
+        debugs(12, 2, "ICP Client REQUEST:\n---------\n" <<
+               "opcode=(" << std::setw(3) << pkt->opcode << ")" << icp_opcode_str[pkt->opcode] <<
+               ", version=" << pkt->version <<
+               ", length=" << ntohs(pkt->length) <<
+               ", reqnum=" << std::setw(8) << ntohl(pkt->reqnum) <<
+               ", flags=(" << std::hex << std::setw(8) << ntohl(pkt->flags) << ")" <<
+               ", shostid=" << Ip::Address(a) <<
+               Debug::Extra <<
+               Raw(nullptr, (buf+sizeof(icp_common_t)), ntohs(pkt->length)).minLevel(9) <<
+               "\n----------");
 
         icp_version = (int) buf[1]; /* cheat! */
 


src/icp_v2.cc
```diff
@@ -680,16 +662,26 @@ icpHandleUdp(int sock, void *)
...
-
-        icpPktDump(buf);
-#endif
-
         if ((size_t) len < sizeof(icp_common_t)) {
             debugs(12, 4, "icpHandleUdp: Ignoring too-small UDP packet");
             break;
         }
 
+        auto pkt = reinterpret_cast<icp_common_t *>(buf);
```
rousskov, line 670: This cast exposes Squid to unaligned integer access dangers, right? If Squid already had this bug in icpHandleIcpV2() and similar code, then please add an XXX here (at least). Otherwise, we should not add these dangers, especially for the debugging sake.

src/icp_v2.cc
```diff
@@ -680,16 +662,26 @@ icpHandleUdp(int sock, void *)
...
+        struct in_addr a;
+        a.s_addr = pkt->shostid;
+        debugs(12, 2, "ICP Client remote=" << from << " FD " << sock);
+        debugs(12, 2, "ICP Client REQUEST:\n---------\n" <<
+               "opcode=(" << std::setw(3) << pkt->opcode << ")" << icp_opcode_str[pkt->opcode] <<
```
rousskov, line 675: We must not use tainted/unchecked `pkt->opcode` as an array index.

src/icp_v2.cc
```diff
@@ -680,16 +662,26 @@ icpHandleUdp(int sock, void *)
...
+               ", version=" << pkt->version <<
+               ", length=" << ntohs(pkt->length) <<
+               ", reqnum=" << std::setw(8) << ntohl(pkt->reqnum) <<
+               ", flags=(" << std::hex << std::setw(8) << ntohl(pkt->flags) << ")" <<
```
rousskov, line 679: Here and elsewhere, please do not attempt to align/beautify level-2+ debugging output like that. It wastes runtime CPU cycles, log space, code space, compile time, developer time, and often complicates automated log processing. The tiny benefits of using setw() do not outweigh the many cons. Please also avoid excessive decorations like parenthesis around numbers, for similar reasons.

```suggestion
               ", reqnum=" << ntohl(pkt->reqnum) <<
               ", flags=" << std::hex << ntohl(pkt->flags) << std::dec <<
```

The above suggestion also removes "std::hex" long-term side effects.

rousskov, line 679: Please check whether std::hex adds the `0x` prefix. If it does not, we should add it (if possible, when configuring the debugging stream in Debug.cc).

src/icp_v2.cc
```diff
@@ -680,16 +662,26 @@ icpHandleUdp(int sock, void *)
...
+               ", shostid=" << Ip::Address(a) <<
+               Debug::Extra <<
+               Raw(nullptr, (buf+sizeof(icp_common_t)), ntohs(pkt->length)).minLevel(9) <<
```
rousskov, line 682: This raw dump should be in hex so that it is easy to convert to actual raw packets using xxd or similar.

rousskov, line 682: Consider dropping the minLevel() part -- IIRC, it will be added automatically for large pkt-length values.

rousskov, line 682: Please name this field, especially since it will appear on a separate cache.log line! For example:

```suggestion
               Raw("payload", (buf+sizeof(icp_common_t)), ntohs(pkt->length)).minLevel(9) <<
```

rousskov, line 682: AFAICT, `pkt->length` has not been fully validated: There is no guarantee we read (or even allocated) as many bytes as you are trying to print.

src/icp_v2.cc
```diff
@@ -680,16 +662,26 @@ icpHandleUdp(int sock, void *)
...
+        a.s_addr = pkt->shostid;
+        debugs(12, 2, "ICP Client remote=" << from << " FD " << sock);
+        debugs(12, 2, "ICP Client REQUEST:\n---------\n" <<
+               "opcode=(" << std::setw(3) << pkt->opcode << ")" << icp_opcode_str[pkt->opcode] <<
+               ", version=" << pkt->version <<
+               ", length=" << ntohs(pkt->length) <<
+               ", reqnum=" << std::setw(8) << ntohl(pkt->reqnum) <<
+               ", flags=(" << std::hex << std::setw(8) << ntohl(pkt->flags) << ")" <<
+               ", shostid=" << Ip::Address(a) <<
+               Debug::Extra <<
```
rousskov, line 681: Please merge with the next line to emphasize what is "extra" here and help with grepping.

rousskov (changes requested):
I stopped reviewing due to too many problems, but this partial/quick review should allow you to make progress.

src/icp_v2.cc
yadij: I considered it. But to keep the log output consistent and because even short URI which most of these will be may contain credentials I am putting level-9 along with all other full-data dumps.

rousskov: Not sure what inconsistencies you perceive.

If we are concerned about credentials, then we should make that concern explicit IMO. If you are not around, somebody else can easily change or remove that debugging level without realizing that the information is sensitive. Not all level-9 information is sensitive, of course! We can address this by adding and using Raw::sensitive() or a similar method (which can just call minLevel(9) today, I guess, but future implementations could be different).

src/icp_v2.cc
yadij: Maybe. This is a pre-existing if so. Added the XXX with next update.

src/icp_v2.cc
yadij: I am trying to make this output as similar as possible to the level-2 output already used by HTTP and FTP. Those displays are header only, without any payload logged until higher levels (ie DBG_DATA / 9).

Also, I think it a bad idea to create any distinction between small vs large payloads. "KISS" principle, etc. if not to avoid issues from mistakes in admin expectations.

yadij: fixed in 920f275

yadij: fixed in 920f275

src/icp_v2.cc
yadij: fixed in 920f275

src/icp_v2.cc
yadij: fixed in 920f275

yadij:
@rousskov, pushing back to you to close off all the resolved review threads. Hopefully my response about consistency on the payload info resolves or removes the other related comments.

src/icp_v2.cc
```diff
@@ -669,39 +655,56 @@ icpHandleUdp(int sock, void *)
...
-        else if (icp_version == ICP_VERSION_3)
+        else if (pkt->version == ICP_VERSION_3)
             icpHandleIcpV3(sock, from, buf, len);
         else
-            debugs(12, DBG_IMPORTANT, "WARNING: Unused ICP version " << icp_version <<
-                   " received from " << from);
+            throw TextException(ToSBuf("unknown version ", pkt->version), Here());
+
+        } catch (...) {
+            debugs(12, 2, "WARNING: ICP remote=" << from << " FD " << sock << ": " << CurrentException);
```
rousskov, line 705: This hides all errors from the admin, including previously reported "abnormal" or "unexpected" errors from trusted peers. Even ICP access logs will not disclose them AFAICT. Please update the PR description to disclose and justify this significant change.

rousskov (changes requested):
> Hopefully my response about consistency on the payload info resolves or removes the other related comments.

Unfortunately, it does not, but I think I now know where the misunderstanding is coming from.

src/icp_v2.cc
```diff
@@ -669,39 +655,56 @@ icpHandleUdp(int sock, void *)
...
+
+        struct in_addr a;
+        a.s_addr = pkt->shostid;
+        debugs(12, 2, "ICP Client remote=" << from << " FD " << sock);
+        debugs(12, 2, "ICP Client REQUEST:\n---------\n" <<
+               "opcode=(" << std::setw(3) << pkt->opcode << ")" << icp_opcode_str[pkt->opcode] <<
+               ", version=" << pkt->version <<
+               ", length=" << payloadLength <<
+               ", reqnum=" << ntohl(pkt->reqnum) <<
+               ", flags=" << std::hex << ntohl(pkt->flags) << std::dec <<
```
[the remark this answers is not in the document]
bl-ue: It doesn't: https://en.cppreference.com/w/cpp/io/manip/hex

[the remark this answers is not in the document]
rousskov: Thank you. I think I found what we need here:

```suggestion
               ", flags=" << asHex(ntohl(pkt->flags)) <<
```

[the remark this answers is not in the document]
bl-ue: By the way, Squid's really cool!  Keep up the great work! ❤️ 🚀

src/icp_v2.cc
```diff
@@ -669,39 +655,56 @@ icpHandleUdp(int sock, void *)
...
+               ", shostid=" << Ip::Address(a) <<
+               Debug::Extra <<
+               Raw(nullptr, (buf+sizeof(icp_common_t)), payloadLength).minLevel(DBG_DATA).hex() <<
```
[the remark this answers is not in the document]
yadij: This output is displaying ICP protocol syntax received (if valid), not HTTP or whatever protocol meta got sent in the payload. In Squid cache.log payload data is at level-9 / DBG_DATA.

The KISS relevance is that the ICP payload is being added to the same debugs() statement as the ICP header display (at the level-9). The slightly more complex variant would have it as a second debugs() statement with a repeated set of prefix connection details to identify it (in case of overlapping worker outputs).

Also, FTR I thought it was just URLs too initially but when you requested hex I looked at the RFC closely to see if that was necessary (it doubles the log increase per packet). Some of the packets do have non-URL data and binary blobs.

From the security / "safety" aspect, the principle being followed is to reduce the attack surface. In this case the surface area is number of places any potential sensitive info in the URL is available to attackers (CWE "Information Leak"). It is not *necessary* to log the ICP payload to perform basic traffic flow accounting / tracing, so it should not be done at the log level-2 provided for that. The only reason it is logged at all is for advance troubleshooting of URl mangling and normalization between proxies using ICP.

src/icp_v2.cc
```diff
@@ -616,44 +617,29 @@ icpHandleIcpV2(int fd, Ip::Address &from, char *buf, int len)
...
+        try {
+
+        auto len = comm_udp_recvfrom(sock,
                                 buf,
                                 SQUID_UDP_SO_RCVBUF - 1,
                                 0,
                                 from);
 
+        debugs(12, 4, "FD " << sock << ": received " << len << " bytes from " << from);
+
```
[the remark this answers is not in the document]
yadij: There is currently no debugging in comm_udp_recvfrom() indicating what UDP packet size was received. Only what listening socket is being checked for input.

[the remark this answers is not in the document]
yadij: Done in be03d8657

src/icp_v2.cc
```diff
@@ -669,39 +655,56 @@ icpHandleUdp(int sock, void *)
...
-            break;
-        }
-
-        icp_version = (int) buf[1]; /* cheat! */
+        // XXX: possible unaligned integer access
+        auto pkt = reinterpret_cast<icp_common_t *>(buf);
+        auto payloadLength = ntohs(pkt->length);
+
+        if (static_cast<size_t>(len) < sizeof(icp_common_t))
+            throw TextException("too-small UDP packet (ignoring)", Here());
```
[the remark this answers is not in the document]
yadij: Done in 5de1788c0

src/icp_v2.cc
```diff
@@ -669,39 +655,56 @@ icpHandleUdp(int sock, void *)
...
-
-        if ((size_t) len < sizeof(icp_common_t)) {
-            debugs(12, 4, "icpHandleUdp: Ignoring too-small UDP packet");
-            break;
-        }
-
-        icp_version = (int) buf[1]; /* cheat! */
+        // XXX: possible unaligned integer access
+        auto pkt = reinterpret_cast<icp_common_t *>(buf);
+        auto payloadLength = ntohs(pkt->length);
```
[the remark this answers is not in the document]
yadij: Ouch. Fixed in 5de1788c0

src/icp_v2.cc
```diff
@@ -669,39 +655,56 @@ icpHandleUdp(int sock, void *)
             /* or maybe an EHOSTUNREACH "No route to host" message */
             if (xerrno != ECONNREFUSED && xerrno != EHOSTUNREACH)
 #endif
-                debugs(50, DBG_IMPORTANT, "icpHandleUdp: FD " << sock << " recvfrom: " << xstrerr(xerrno));
-
+                throw TextException(xstrerr(xerrno), Here());
```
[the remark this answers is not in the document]
yadij: Done in 5de1788c0

src/icp_v2.cc
```diff
@@ -669,39 +655,56 @@ icpHandleUdp(int sock, void *)
...
+
+        if (payloadLength < len-sizeof(icp_common_t))
+            throw TextException("length value larger than data received (ignoring)", Here());
+
+        // TODO use BinaryTokenizer parse to validate fully
+        if (pkt->opcode < 0 || pkt->opcode >= ICP_END)
+            throw TextException(ToSBuf("unknown opcode ", pkt->opcode), Here());
+
+        struct in_addr a;
+        a.s_addr = pkt->shostid;
```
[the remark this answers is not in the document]
yadij: Removed in c9428fa9b

src/icp_v2.cc
yadij: Reverted the in meaningful debugs error message in 36d79a35f.

The high level message about unknown ICP version being ignored does not need to be logged at level-1.

src/icp_v2.cc
```diff
@@ -669,39 +655,56 @@ icpHandleUdp(int sock, void *)
...
-#endif
-
-        if ((size_t) len < sizeof(icp_common_t)) {
-            debugs(12, 4, "icpHandleUdp: Ignoring too-small UDP packet");
-            break;
-        }
-
-        icp_version = (int) buf[1]; /* cheat! */
+        // XXX: possible unaligned integer access
+        auto pkt = reinterpret_cast<icp_common_t *>(buf);
```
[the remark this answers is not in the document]
yadij: Nice. Totally agreed. Refactoring now.

src/icp_v2.cc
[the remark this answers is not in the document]
rousskov: > There is currently no debugging in comm_udp_recvfrom() indicating what UDP packet size was received

Yes, I know. That fact was irrelevant to my change request. Thank you for improving comm_udp_recvfrom() debugging!

src/icp_v2.cc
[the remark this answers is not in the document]
rousskov: These arguments are pretty much invalidated by the fact that ICP "payload" is not really payload in any true sense of that word when interpreted in relevant/Squid context. It is just a variable-size portion of the header. HTTP has one mechanism to supply such variable-size header information, PROXY protocol has another, ICP has its own way. Yes, the word is spelled "payload" in the RFC, but that is just letters. Sometimes, letters mislead if interpreted too literally. We need to think about the _meaning_ of the ICAP message part we are printing. And that meaning is no different than a part of an HTTP or PROXY protocol header. We should use the same approach to report all those parts.

The fact that some headers carry binary data is not relevant to this change request. That fact only affects how we render the data, and we are in agreement that it should be rendered using hex (unless somebody wants to invest even more time to make that decision based on message type and payload field contents).

But I give up. This problem is just not worth my time. If I failed to convince you, leave minLevel(). No need to explain why I have not convinced you.

----

> with a repeated set of prefix connection details to identify it (in case of overlapping worker outputs).

This is out of topic here, but just in case this resurfaces elsewhere: We should not repeat details. Squid cache.log format already identifies which worker the record is coming from.

src/icp_v2.cc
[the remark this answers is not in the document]
yadij: Done in a0884403e

src/icp_v2.cc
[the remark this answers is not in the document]
yadij: > These arguments are pretty much invalidated by the fact that ICP "payload" is not really payload in any true sense of that word when interpreted in relevant/Squid context. It is just a variable-size portion of the header.

The relevant context is that we are displaying a single message from a message-based protocol. Using that protocols terminology, in ICP specific code.

RFC 2816 section 1 is very explicit how to interpret the octets received:
"
ICP Message Format

   The ICP message format consists of a 20-octet fixed header plus a
   variable sized payload (see Figure 1).
"

The "Squid way" (er, context?) to display messages is level-2 for header, level-9 for payload.

yadij:
@rousskov, current state is that I am waiting for a good conclusion to the thread from https://github.com/squid-cache/squid/pull/827#discussion_r638896499. That will determine whether the other open change requests happen.

rousskov:
>I am waiting for a good conclusion to the thread from [#827 (comment)](https://github.com/squid-cache/squid/pull/827#discussion_r638896499). That will determine whether the other open change requests happen.

If you are waiting for me, my [position](https://github.com/squid-cache/squid/pull/827#discussion_r641190988) has not changed: I have given up. This problem is just not worth my time. If I failed to convince you, leave minLevel(). No need to explain why I have not convinced you. I will do my best not to participate in further discussions on that issue.

hno:
Have to agree with Alex here. In terms of Squid logs HTTP headers In ICP
"payload" is not payload, it is the request headers of the HTTP request
transmitted via ICP. Any decisions regarding verbosity of ICP transmitted
HTTP header data should be aligned with how HTTP header data is handled in
HTTP context. Log levels largely relates to security/sensitivity level and
data meaning, not protocol terms.

For what it is worth the URL is also part of the "payload" in ICP
terminology. The ICP header is pretty much just the ICP opcode with no
details about the queried HTTP request, quite similar to an UDP header.

This said ICP may actually contain the object body, which in such case
should be handled accordingly. But I am not sure if ICP is ever used in
that manner by Squid.

But this is just my opinion. Should not be seen as review of the code, or
an attempt to influence the code. Only an attempt to clarify terminology.

Regards
Henrik

src/icp_v2.cc
yadij: Done in 01eb43d

src/icp_v2.cc
yadij: Done in 01eb43d

yadij:
This PR is blocked by @rouskkov self-requested review and "I will do my best not to participate in further discussions" decision.

rousskov:
> This PR is blocked by @rouskkov self-requested review and "I will do my best not to participate in further discussions" decision.

Both assumptions are false and the misleading quote misrepresents the actual dialog:

* This PR has no active review requests. Not self-requested, not regular ones. None. There was one regular review request long time ago, and it has been [satisfied](https://github.com/squid-cache/squid/pull/827#pullrequestreview-667920447) long time ago.

* I actually [said](https://github.com/squid-cache/squid/pull/827#issuecomment-851608022) "I will do my best not to participate in further discussions **on that issue**" (emphasis added), clearly referring to one specific problem discussed in that comment. I know Henrik has commented on that issue as well, but _my_ plan is still in place exactly as it was stated.

> Amos: current state is that I am waiting for a good conclusion to the thread from https://github.com/squid-cache/squid/pull/827#discussion_r638896499. That will determine whether the other open change requests happen.

AFAICT, everybody who wanted to share their opinion on that thread have done so long time ago. It is your turn to act on that feedback. Once you think that my other change requests are addressed, please re-request my review.

yadij:
>  If I failed to convince you, leave minLevel(). 
...
> It is your turn to act on that feedback.
... and I have.

> Once you think that my other change requests are addressed, please re-request my review.

Going by how all the github conversations are resolved and closed except that one - I think they are.

src/icp_v2.cc
```diff
@@ -670,38 +649,37 @@ icpHandleUdp(int sock, void *)
...
-        else if (icp_version == ICP_VERSION_3)
+        else if (pkt.version == ICP_VERSION_3)
             icpHandleIcpV3(sock, from, buf, len);
         else
-            debugs(12, DBG_IMPORTANT, "WARNING: Unused ICP version " << icp_version <<
-                   " received from " << from);
+            throw TextException(ToSBuf("unknown version ", pkt.version), Here());
+
+        } catch (...) {
+            debugs(12, 2, "WARNING: ICP remote=" << from << " FD " << sock << ": " << CurrentException << " (ignoring)");
```
rousskov, line 680: As discussed [earlier](https://github.com/squid-cache/squid/pull/827#discussion_r638947024), this hides the "unknown ICP version" error messages from the admin. That message was probably a good canary for an attack on the ICP port with random data.

> Amos: The high level message about unknown ICP version being ignored does not need to be logged at level-1.

If the PR hides that error, I am not going to block it on those grounds because I do not have the time to argue and detail a good solution for the overall problem. However, if the PR hides that error, its description must disclose and justify that important reporting change. This is the [second](https://github.com/squid-cache/squid/pull/827#discussion_r638947024) time I am requesting that disclosure.

As always, to save time, please add your justifications to the PR description rather than (just to) this change request thread.

src/icp_v2.cc
```diff
@@ -670,38 +649,37 @@ icpHandleUdp(int sock, void *)
...
-        }
+        icp_common_t pkt(buf, len);
 
-        icp_version = (int) buf[1]; /* cheat! */
+        debugs(12, 2, "ICP Client remote=" << from << " FD " << sock);
+        debugs(12, 2, "ICP Client REQUEST:\n---------\n" <<
+               icp_opcode_str[pkt.opcode] << " #" << pkt.reqnum << " ICP/" << pkt.version << std::endl <<
+               ":length: " << pkt.length << std::endl <<
+               ":flags: " << asHex(pkt.flags) <<
+               Debug::Extra << Raw("payload", (buf+sizeof(icp_common_t)), pkt.length).minLevel(DBG_DATA).hex() <<
```
rousskov, line 666: Given that this is a multiline output that follows very special formatting, the Debug::Extra trick does not work well here -- the payload dump gets misaligned. There is also a problem with Raw() printing an extra space when labeled. I suggest doing something like this to fix those problems:

```C++
":flags: " << asHex(pkt.flags) << '\n' <<
":payload:" << Raw("raw", (buf+sizeof(icp_common_t)), pkt.length).minLevel(DBG_DATA).hex() << '\n' <<
"----------");
```

We can also use "bytes" (or something else) instead of "raw" as the Raw() label. We could use nullptr label instead (and add a space after `:` before Raw()), but that will hide the payload size. Your call. And please test this. I may be missing some formatting nuances.

rousskov (changes requested):
Still buggy AFAICT. I also left a few polishing suggestions.

When you are done with the next iteration, please either update the PR description to show the actual ICP message printout or just remove the misleading (made up?) sample.

rousskov:
I also adjusted PR title to be a bit more specific. Please polish further as needed, of course.

PR #834, opened 2021-05-30
Refactor X-Forwarded-For stats

kinkie:
Use std::unordered_map instead of hash_table
Clean up unused code
Rename a few functions to better reflect name

Test plan:

$ squidclient mgr:forw_headers
HTTP/1.1 200 OK
Server: squid/6.0.0-VCS

       10 foo
       15 baz
        7 bar


src/client_side_request.cc
```diff
@@ -1136,7 +1136,7 @@ clientInterpretRequestHeaders(ClientHttpRequest * http)
 
     if (req_hdr->has(Http::HdrType::X_FORWARDED_FOR)) {
         String s = req_hdr->getList(Http::HdrType::X_FORWARDED_FOR);
-        fvdbCountForw(s.termedBuf());
+        fvdbCountXFF(StringToSBuf(s));
```
yadij, line 1139: Please use the standard header "Forwarded" as basis for naming not the deprecated XFF header.
```suggestion
        fvdbCountForwarded(StringToSBuf(s));
```

src/log/access_log.h
```diff
@@ -14,11 +14,11 @@
 #include "LogTags.h"
 #include "sbuf/forward.h"
 
+/// XXX: these functions preserve all counted values until the next log rotation
 /// count occurrences of the given Via header value
-/// XXX: this function preserves all counted values until the next log rotation
 void fvdbCountVia(const SBuf &);
-
-void fvdbCountForw(const char *key);
+/// count occurrences of the given Via header value
```
yadij, line 20: copy-pasta
```suggestion
/// count occurrences of the given X-Forwarded-For header value
```

src/tests/stub_liblog.cc
```diff
@@ -23,21 +23,21 @@ AccessLogEntry::SslDetails::SslDetails() {STUB}
 #endif
 */
 void accessLogLogTo(CustomLog *, AccessLogEntry::Pointer &, ACLChecklist *) STUB
-void accessLogLog(AccessLogEntry::Pointer &, ACLChecklist *) STUB
-void accessLogRotate(void) STUB
-void accessLogClose(void) STUB
-void accessLogInit(void) STUB
-const char *accessLogTime(time_t) STUB_RETVAL(nullptr)
+    void accessLogLog(AccessLogEntry::Pointer &, ACLChecklist *) STUB
```
yadij, line 26: Changes in this file all look very messy. Were they created by astyle?

Please remove the out of scope changes so only the new ones are visible in the merge.

yadij (changes requested):

src/client_side_request.cc
kinkie: Sure, but then shouldn't we counting Forwarded occurrences instead of XFF? Or even both?

src/log/access_log.h
kinkie: Fixed, thanks

src/tests/stub_liblog.cc
kinkie: Done.

The syntax we use for STUB confuses automatic formatters.
Could we maybe mass-rename it to {STUB;} ? Definitely not in scope for this change tho

src/client_side_request.cc
yadij: I thought about requesting count of both. But the PR for actually using Forwarded is not merged yet, and it makes a bit more sense for those user visible changes to this report be added with or after that PR.

src/tests/stub_liblog.cc
```diff
@@ -48,7 +48,7 @@ bool FormattedLog::usesDaemon() const STUB_RETVAL(false)
 
 #include "log/File.h"
 CBDATA_CLASS_INIT(Logfile);
-Logfile::Logfile(const char *) {STUB}
+Logfile::Logfile(const char *) STUB
```
yadij, line 51: out of scope change

yadij (changes requested):

src/tests/stub_liblog.cc
kinkie: reverted

yadij (approved):
LGTM

src/tests/stub_liblog.cc
rousskov: > Could we maybe mass-rename it

I have not studied the details, but I have seen related problems in the past, and suspect that mass-renaming is probably the right long-term solution despite its pain. I think that dedicated PR should also change the `STUB` and related definitions so that we write `{ Stub(); }` or some such, confusing even fewer automation scripts _and_ allowing the compiler to detect  any forgotten `STUB` uses.

rousskov (commented):
Sorry it took me so long to get to this PR. Clearing for merge.

PR #836, opened 2021-06-03
Remove unused code silencing intercept errors

yadij:
The removed code has not been actively used almost since it was added.

It is now widely accepted that NAT and TPROXY can only be done on the
machine running Squid. The corresponding address lookup errors are an
indication of either a system misconfiguration or an adverse external
event such as flushing of conntrack tables. Since these errors should be
fatal to the affected transactions and the admin usually has the power
to address them, Squid should report them at level 1.

### DIFF
--- a/src/ip/Intercept.cc
+++ b/src/ip/Intercept.cc
@@ -129,7 +129,7 @@ Ip::Intercept::StopInterception(const char *str)
 }
 
 bool
-Ip::Intercept::NetfilterInterception(const Comm::ConnectionPointer &newConn, int silent)
+Ip::Intercept::NetfilterInterception(const Comm::ConnectionPointer &newConn)
 {
 #if LINUX_NETFILTER
     struct sockaddr_storage lookup;
@@ -143,12 +143,8 @@ Ip::Intercept::NetfilterInterception(const Comm::ConnectionPointer &newConn, int
                     newConn->local.isIPv6() ? IP6T_SO_ORIGINAL_DST : SO_ORIGINAL_DST,
                     &lookup,
                     &len) != 0) {
-        if (!silent) {
-            int xerrno = errno;
-            debugs(89, DBG_IMPORTANT, "ERROR: NF getsockopt(ORIGINAL_DST) failed on " << newConn << ": " << xstrerr(xerrno));
-            lastReported_ = squid_curtime;
-        }
-        debugs(89, 9, "address: " << newConn);
+        int xerrno = errno;
+        debugs(89, DBG_IMPORTANT, "ERROR: NF getsockopt(ORIGINAL_DST) failed on " << newConn << ": " << xstrerr(xerrno));
         return false;
     } else {
         newConn->local = lookup;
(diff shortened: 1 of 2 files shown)


rousskov:
> PR description: It is now widely accepted that NAT and TPROXY ... errors that occur should be fixed in the system setup ASAP.

Some NAT and TPROXY errors are not caused by "system setup" errors. For example, flushing system conntrack tables may cause these errors. I adjusted the PR description to relax this assertion a bit while still explaining why these errors should be reported to the admin. Please review.

I have not checked whether we can replace "should be fatal" with "are fatal" in the updated PR description. If Squid already kills affected transactions, then we should replace!

src/ip/Intercept.cc
```diff
@@ -143,12 +143,8 @@ Ip::Intercept::NetfilterInterception(const Comm::ConnectionPointer &newConn, int
                     newConn->local.isIPv6() ? IP6T_SO_ORIGINAL_DST : SO_ORIGINAL_DST,
                     &lookup,
                     &len) != 0) {
-        if (!silent) {
-            int xerrno = errno;
-            debugs(89, DBG_IMPORTANT, "ERROR: NF getsockopt(ORIGINAL_DST) failed on " << newConn << ": " << xstrerr(xerrno));
-            lastReported_ = squid_curtime;
-        }
-        debugs(89, 9, "address: " << newConn);
+        int xerrno = errno;
```
rousskov, line 146: Here and elsewhere in this PR changes, we can and, hence, should easily improve xerrno declaration:

```suggestion
        const auto xerrno = errno;
```

src/ip/Intercept.cc
```diff
@@ -280,11 +273,7 @@ Ip::Intercept::IpfInterception(const Comm::ConnectionPointer &newConn, int silen
     if (x < 0) {
         int xerrno = errno;
         if (xerrno != ESRCH) {
-            if (!silent) {
-                debugs(89, DBG_IMPORTANT, "IPF (IPFilter) NAT lookup failed: ioctl(SIOCGNATL) (v=" << IPFILTER_VERSION << "): " << xstrerr(xerrno));
-                lastReported_ = squid_curtime;
-            }
-
+            debugs(89, DBG_IMPORTANT, "IPF (IPFilter) NAT lookup failed: ioctl(SIOCGNATL) (v=" << IPFILTER_VERSION << "): " << xstrerr(xerrno));
```
rousskov, line 276: Here and elsewhere in changed PR lines, we should add an `ERROR: ` prefix for clarity, consistency, and ease-of-searching sake.

```suggestion
            debugs(89, DBG_IMPORTANT, "ERROR: IPF (IPFilter) NAT lookup failed: ioctl(SIOCGNATL) (v=" << IPFILTER_VERSION << "): " << xstrerr(xerrno));
```

rousskov (approved):
Thank you for removing this dead code. I left a few polishing suggestions, but I do _not_ insist on any of them because we are dealing with moved code here, and the scope of this PR can be reasonably restricted to just the dead code removal.

src/ip/Intercept.cc
yadij: Done in 8b36c4b

src/ip/Intercept.cc
yadij: Done in 8b36c4b

PR #837, opened 2021-06-03
Do not stop listening after "ERROR: NAT/TPROXY lookup failed..."

eduard-bagdasaryan:
_Intermittent_ NAT/TPROXY lookup failures can happen for several reasons
outside Squid control. For example, flushing conntrack tables under load
is likely to trigger such errors. Squid should not stop listening if
there is a good change that the next accept(2) will be fully successful.
We already apply the same logic to several other accept problems today.

Also added an XXX that TcpAcceptor::mustStop() does not work as expected
on error.


rousskov (approved):

src/comm/TcpAcceptor.cc
```diff
@@ -395,9 +396,8 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
     // Perform NAT or TPROXY operations to retrieve the real client/dest IP addresses
     if (conn->flags&(COMM_TRANSPARENT|COMM_INTERCEPTION) && !Ip::Interceptor.Lookup(details, conn)) {
         debugs(50, DBG_IMPORTANT, "ERROR: NAT/TPROXY lookup failed to locate original IPs on " << details);
-        // Failed.
         PROF_stop(comm_accept);
-        return Comm::COMM_ERROR;
+        return Comm::NOMESSAGE;
```
yadij, line 400: Since we did have accept() successful we should now conn->close() to ensure this does not become an orphan.

yadij (dismissed):

src/comm/TcpAcceptor.cc
eduard-bagdasaryan: We should not call conn->close() manually in this case (and in other similar cases as well) because it is done automatically when the passed 'details' pointer becomes abandoned in acceptOne(). This happens if oldAccept() returns NOMESSAGE. There is also a related comment in oldAccept() before fd_open().

rousskov: Please note that a recent commit 4831b46 has already addressed this problem by calling enterOrphanage() for this **known**-to-be-orphan connection. That commit covered virtually all non-OK accept(2) cases with a valid orphan FD, not just this specific one.

yadij: AIUI the definition we have of "orphan FD" is an FD where no AsyncJob with access to the conn has sufficient information to close() it correctly before the Comm::Connection instance destructs. Here we have TcpAcceptor with all the necessary state, and access to the conn instance.

rousskov: > AIUI the definition we have of "orphan FD" is an FD where no AsyncJob with access to the conn has sufficient information to close() it correctly before the Comm::Connection instance destructs.

Technically, an orphan Connection object is a Connection object that gets destructed in an open state (i.e. with a non-negative descriptor). Some _rare_ orphan Connection cases are not covered by this statement, but I am omitting them here for simplicity sake. On a more human level, an orphan Connection object is a Connection object without an owner (where Connection owner is code that explicitly closes the connection or has an associated connection closure handler). 

Owning a Connection correctly is difficult -- there are many known [problems](https://github.com/measurement-factory/squid-notes/blob/start/connection.md#problems) associated with Connection ownership. We are _not_ trying to solve those complex problems in this simple and unrelated PR (but we are working on their long-term solution). This PR does not create new Connection ownership problems and does not make old problems worse.

> Here we have TcpAcceptor with all the necessary state, and access to the conn instance.

While TcpAcceptor could add more code to prevent the `details` Connection object from becoming an orphan, that is not necessary (known/purposeful orphans are not bugs) or desirable (less code is better), at least for now. And even if elimination of known orphans becomes a goal, the [suggested](https://github.com/squid-cache/squid/pull/837/files#r645208377) addition of an explicit close() call will not be sufficient (or even desirable) due to the existence of other orphan-creating outcomes and the possibility of C++ exceptions. Following that suggestion would just add one more "just in case" Connection closing code that misleads readers and complicates refactoring without adding enough real value.

This change request is discussing a problem that has been solved in commit 4831b46. We should (continue to) use the working enter/leaveOrphanage() API until a better API becomes available. Factory is working on a better/comprehensive solution to Connection ownership problems, and we can revisit orphanage API improvements when those solutions are ready for review.

yadij: Even with that definition (I disagree, but anyway) my point is that this is *not* orphaned. The Connection object is a *local variable* and remains one for this entire code path. If the FD is not closed properly before it destructs Squid starts complaining about orphan bugs *wrongly*.

The right place to do that close() is right where FD is detected as broken.

rousskov: > The Connection object is a local variable and remains one for this entire code path.

Yes, after the PR changes, the affected Connection object lifetime is controlled by local variable(s), but that fact is irrelevant for the "orphan" definition used by the official Squid code (and me).

> If the FD is not closed properly before it destructs Squid starts complaining about orphan bugs _wrongly_.

Not in this case. Since commit 1c2b446, Squid does not _complain_ about known (i.e. purposefully created) orphaned Connections being destructed; Squid does not consider such events bugs; Squid simply reports the event itself. And since commit 4831b46, Squid _knows_ that this particular Connection is orphaned. The "wrongful complains" problem you are thinking about (in this specific context) has been solved before this PR (in commit 4831b46).

> The right place to do that close() is right where FD is detected as broken.

There are too many such "right" places for the above error handling strategy to work well in non-trivial C++ code. And we should not be even discussing this strategy here because this PR does not change old or add new places where "FD is detected as broken". This PR simply changes which one of the two _existing_ problem-handling paths is used for a particular case (one of many). None of those paths close the accepted Connection explicitly.

yadij: > This PR simply changes which one of the two existing problem-handling paths is used for a particular case (one of many). None of those paths close the accepted Connection explicitly.

Before this PR the NOMESSAGE code path only existed when accept() output FD -1. This PR is creating a new path where FD >= 0 and orphan flag is set while TcpAcceptor is still in full ownership of the Comm::Connection.

I have opened PR #840 to resolve the issue of orphanage being set too early. If accepted as-is that will resolve this request by no longer needing a close() here to obey the Comm::Connection API design.

rousskov: > Before this PR the NOMESSAGE code path only existed when accept() output FD -1.

That assertion is false AFAICT: There is another `NOMESSAGE` case with an open FD:

    git grep -2 'attempting more than' upstream/master

> I have opened PR #840 to resolve the issue of orphanage being set too early. If accepted as-is that will resolve this request by no longer needing a close() here to obey the Comm::Connection API design.

The `NOMESSAGE with FD` code path this change request is focused on already exists in the official code. I continue to think that this change request is outside this PR scope. We can continue the discussion of what "obeying the APIs" means in _your_ PR. I see no reason to block _this_ unrelated PR on that discussion, regardless of that discussion outcome. This PR may have triggered a re-evaluation of official code on your side, and, as you know, we agree that there are many connection ownership problems still to be addressed, but _this_ PR changes are the right ones, regardless of how the other PR(s) evolve. This PR should not be blocked.

yadij: I agreed to the orphanage mechanism going in as a bandaid mechanism to reduce admin annoyance when we had identified a source of bug3329 events (implicit close) but not yet found the exact code path to place the required explicit close() call which would resolve them.

What this PR has highlighted for me is that the orphanage existence is now being used as an excuse to avoid putting in close() when a specific code path is found that would fix many actually encountered orphans.

The Comm::Connection API was designed with an API that requires *explicit* calls to close() on error conditions being encountered. The goes not just for socket errors (eg system call errors), but also for "meta states" of the Connection being represented (eg pinning closures, or this NAT failure).


All I ask is that the Comm::Connection API be used per its original design. To prevent *adding* orphans with this NAT fix.

rousskov: This change request keeps shifting goal posts. As soon as I demonstrate that the previous concern has been addressed or is invalid, a brand new concern is formulated. It is difficult to make progress that way.

> the orphanage existence is now being used as an excuse to avoid putting in close() when a specific code path is found that would fix many actually encountered orphans.

Even this misinterpretation does not give a reviewer the right to insist on out-of-scope changes: Even if we assume, for the sake of the argument, that adding an explicit close() call above the changed `return` statement is a good idea, a reviewer cannot _insist_ on that out-of-scope improvement. Please drop this change request.

> All I ask is that the Comm::Connection API be used per its original design.

This PR does not change how the code uses that (bad) API. Before this PR, `COMM_ERROR` and `NOMESSAGE` code did not close the (known) orphan. After this PR, `COMM_ERROR` and `NOMESSAGE` code does not close the (known) orphan. All these outcomes are treated the same they were treated before this PR. We should (and eventually will) discuss how error handling cases should be treated, but this PR should not be held hostage to that future discussion.

The original Connection API is broken and needs serious rework. We will improve it. We will have consistent application of Connection error handling procedures. This is not the place to discuss those necessary changes, and this unrelated PR should not be blocked on that future discussion.

> To prevent _adding_ orphans with this NAT fix.

The number of orphan-creating oldAccept() code paths before and after this PR remain unchanged (at 3 code paths). Known orphans are (and continue to be) rare and are (and continue to be) handled correctly. If anything, this PR shortens the (known) orphan lifespan in one of those rare cases, which is obviously a good thing.

It is, of course, possible to expand this PR scope to also include orphan reduction in TcpAcceptor (or even Squid in general), but you have no right to insist on such an expansion. And since _properly_ handling orphans, even in this limited TcpAcceptor area, requires some work (we do not even agree on how to define "orphan"!), I do not want to just give up and apply the requested change, despite the fact that the butchered code will still "work" after that change. Please drop this out-of-scope change request.

rousskov: @yadij, this PR has been waiting for your review for more than 5 months, including more than 3 months in `review-3` state. Please move it forward. FWIW, my June 2021 [recommendation](https://github.com/squid-cache/squid/pull/837#discussion_r654518834) still stands: Please drop this change request; we can address the remaining (difficult, controversial, and not blocking this bug fix) problems in other PRs or discussions.

yadij: Skipping all the (re-)design discussion. I stand by my original change request. The Comm::Connection API *currently* requires explicit use of a call to close() as soon as the NAT error makes the FD stored there unreliable.

Either change this PR to add the requested close() or this bug continues to wait out the indefinite process of redesigning Comm::Connection.

yadij:
Also, under our current release policy the behaviour/debugs message this PR seeks to resolve should have a Bugzilla ID and report outline what the problem is. Please open that for tracking and update the PR title.

src/comm/TcpAcceptor.cc
eduard-bagdasaryan: >  After this PR, COMM_ERROR and NOMESSAGE code does not close the (known) orphan. All these outcomes are treated the same they were treated before this PR.

Agreed.

> Either change this PR to add the requested close() or this bug continues to wait out the indefinite process of redesigning Comm::Connection.

Unfortunately, I cannot address requests that I do not understand. If there are such (waiting) bugs, we should fix all COMM_ERROR and NOMESSAGE cases (in a future PR). Anyway (as this discussion thread tried to explain many times), this PR does not add new bugs, no inconsistency is introduced. On the other hand, adding close() in this particular case (and keeping other similar cases unchanged) makes error handling in oldAccept() inconsistent.

rousskov:
> 

> Amos: Also, under our current release policy the behaviour/debugs message this PR seeks to resolve should have a Bugzilla ID and report outline what the problem is. Please open that for tracking and update the PR title.

> [ReleaseProcess](https://wiki.squid-cache.org/action/info/ReleaseProcess): All backported changes require a corresponding bugzilla report or security advisory. 

Just to explain why I am not rushing off to create that bug report: I could not find anything in [ReleaseProcess](https://wiki.squid-cache.org/action/info/ReleaseProcess) that would require filing a bug report for this master-branch-based PR to be merged. I assume you are referring to the ReleaseProcess "extra note" quoted above. In that note, you require bug reports as a _backporting_ invariant, but this is not a backporting PR, so that requirement does not apply here. Since I think that requirement results in a waste of time[^1], I do not want to create such "shadow" bug reports in advance, especially for PRs that you are currently (incorrectly) preventing from merging.

If the requirement is not improved[^1], then when/if the time comes to backport this fix, somebody will have to create a "shadow" bug report to satisfy the unfortunate requirement.

[^1]: The requirement should have more than just two sufficient preconditions: Bugzilla entry or Advisory. The PR commit message provides (or should provide) the same basic information as the bug report created from that commit message does. We should be using the existing information instead of requiring folks to duplicate it, wasting time and creating a confusing timeline in the process. I am happy to help improve that requirement if you would like me to.

src/comm/TcpAcceptor.cc
rousskov: > Amos: The Comm::Connection API _currently_ requires explicit use of a call to close() as soon as the NAT error makes the FD stored there unreliable.

This statement is not relevant to _this_ PR (so I will not try to fix problems in the statement). You have not demonstrated relevance -- all past attempts were based on clearly false (as in factually wrong) assumptions. Eduard and I have, in several detailed explanations, tried to explain why it is irrelevant. If there are no critical flaws in our arguments, then you should not be blocking or stalling this PR. If there are critical flaws, you should identify them so that either we change our mind or can continue to help you change yours. This iterative process of adjusting one's understanding and reaching consensus is how progress is made in collaborative environments. 

This collaboration can be very painful, but there is currently no reasonable alternative (unfortunately). If you do not want to participate, you do not have to: Just dismiss your negative review and remove yourself as this PR reviewer.

yadij (commented):
This code relies on RAII behaviour from Comm::Connection which is not (yet) a reliable situation. Connections will be closed (resolving this bug) but cache.log info and debugger state will contain some lies about the connection being orphaned. Discussion of that will be scoped for a followup PR once this bug is gone.

rousskov:
@yadij, thank you for unblocking this PR. Please note that we cannot merge this PR until you dismiss the blocking review.

FYI: You do not need to _approve_ this PR, but if you just add a PR comment after a negative review, then GitHub will still honor your old negative vote. To dismiss the earlier negative review, go to the `1 change requested` line at the bottom of the PR page (in the `Changes requested` box), click on the down arrow at the right end of that line, click on the "..." icon at the right end of your negative review line, and then select the "Dismiss review" item.

PR #839, opened 2021-06-06
Remove custom global new/delete operators

kinkie:
These custom operators did not cover all new/delete cases (e.g., array
allocations), were not declared according to C++ standards (triggering
compiler warnings), and were not enabled in clang builds. 

These customizations enabled custom OOM handling (for covered cases),
but it is not clear whether that feature is desirable overall, and C++
has better ways to implement such handling (i.e. set_new_handler()).

These customizations participated in collection of optional statistics
(--enable-xmalloc-statistics), but it is not clear whether that feature
implementation is good enough, and, even if it is, providing these
partial stats does not outweigh recurring customization problems.


rousskov:
Thank you for taking on this cleanup!

I adjusted PR description (and title) to avoid referring to #838 information that will not be easily available to a "git log" reader, especially long-term. Please review and adjust as needed.

@kinkie, please try to write PR titles/descriptions as future official commit messages (because that is what they will become).

rousskov (approved):

yadij (approved):

kinkie:
Thanks. Given the 3 positive votes, this can be merged

rousskov:
> Given the 3 positive votes, this can be merged

Just FYI: A PR can be merged with just one vote (with all other [preconditions](https://github.com/measurement-factory/anubis/blob/master/README.md#which-pull-requests-are-eligible-for-merging) satisfied). Two or more votes just put the PR on the _fast_ track.

PR #840, opened 2021-06-08
Fix FD accept() error handling

yadij:
FD begin their lifecycle belonging to TcpAcceptor which
is responsible for closing per Comm::Connection API.

That means FD which have issues within the TcpAcceptor
logic should be close()'d with a specific ERROR or
WARNING message displayed about the reason for closure.

Do not enter connections for orphanage until their
ownership is passed into the async queue for a Subscriber.

### DIFF
--- a/src/comm/TcpAcceptor.cc
+++ b/src/comm/TcpAcceptor.cc
@@ -287,6 +287,7 @@ Comm::TcpAcceptor::acceptOne()
     if (flag == Comm::NOMESSAGE) {
         /* register interest again */
         debugs(5, 5, "try later: " << conn << " handler Subscription: " << theCallSub);
+        newConnDetails->close();
     } else {
         // TODO: When ALE, MasterXaction merge, use them or ClientConn instead.
         CodeContext::Reset(newConnDetails);
@@ -323,10 +324,14 @@ Comm::TcpAcceptor::notify(const Comm::Flag flag, const Comm::ConnectionPointer &
         params.xaction = new MasterXaction(XactionInitiator::initClient);
         params.xaction->squidPort = listenPort_;
         params.fd = conn->fd;
+        conn->enterOrphanage();
         params.conn = params.xaction->tcpClient = newConnDetails;
         params.flag = flag;
         params.xerrno = errcode;
         ScheduleCallHere(call);
+    } else {
+        // disconnect clients we cannot respond to
+        conn->close();
     }
 }
 
@@ -375,8 +380,6 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
     // so we end up with a uniform "(HTTP|FTP-data|HTTPS|...) remote-ip:remote-port"
     fd_open(sock, FD_SOCKET, "HTTP Request");
     details->fd = sock;
-    details->enterOrphanage();
-
     details->remote = *gai;
 
     // lookup the local-end details of this new connection
--- a/src/comm/TcpAcceptor.cc
+++ b/src/comm/TcpAcceptor.cc
@@ -44,7 +44,7 @@ Comm::TcpAcceptor::TcpAcceptor(const Comm::ConnectionPointer &newConn, const cha
     AsyncJob("Comm::TcpAcceptor"),
     errcode(0),
     theCallSub(aSub),
-    conn(newConn),
+    listenConn(newConn),
     listenPort_()
 {}
 
@@ -52,7 +52,7 @@ Comm::TcpAcceptor::TcpAcceptor(const AnyP::PortCfgPointer &p, const char *, cons
     AsyncJob("Comm::TcpAcceptor"),
     errcode(0),
     theCallSub(aSub),
-    conn(p->listenConn),
+    listenConn(p->listenConn),
     listenPort_(p)
 {}
 
@@ -78,22 +78,22 @@ Comm::TcpAcceptor::start()
         CodeContext::Reset(listenPort_);
     debugs(5, 5, HERE << status() << " AsyncCall Subscription: " << theCallSub);
 
-    Must(IsConnOpen(conn));
+    Must(IsConnOpen(listenConn));
 
     setListen();
 
-    conn->noteStart();
+    listenConn->noteStart();
 
     // if no error so far start accepting connections.
     if (errcode == 0)
-        SetSelect(conn->fd, COMM_SELECT_READ, doAccept, this, 0);
+        SetSelect(listenConn->fd, COMM_SELECT_READ, doAccept, this, 0);
 }
 
 bool
 Comm::TcpAcceptor::doneAll() const
 {
     // stop when FD is closed
-    if (!IsConnOpen(conn)) {
+    if (!IsConnOpen(listenConn)) {
         return AsyncJob::doneAll();
     }
 
@@ -111,13 +111,13 @@ Comm::TcpAcceptor::swanSong()
 {
     debugs(5,5, HERE);
     unsubscribe("swanSong");
-    if (IsConnOpen(conn)) {
+    if (IsConnOpen(listenConn)) {
         if (closer_ != NULL)
-            comm_remove_close_handler(conn->fd, closer_);
-        conn->close();
+            comm_remove_close_handler(listenConn->fd, closer_);
+        listenConn->close();
     }
 
-    conn = NULL;
+    listenConn = nullptr;
     AcceptLimiter::Instance().removeDead(this);
     AsyncJob::swanSong();
 }
@@ -125,16 +125,16 @@ Comm::TcpAcceptor::swanSong()
 const char *
 Comm::TcpAcceptor::status() const
 {
-    if (conn == NULL)
+    if (!listenConn)
         return "[nil connection]";
 
     static char ipbuf[MAX_IPSTRLEN] = {'\0'};
     if (ipbuf[0] == '\0')
-        conn->local.toHostStr(ipbuf, MAX_IPSTRLEN);
+        listenConn->local.toHostStr(ipbuf, MAX_IPSTRLEN);
 
     static MemBuf buf;
     buf.reset();
-    buf.appendf(" FD %d, %s",conn->fd, ipbuf);
+    buf.appendf(" FD %d, %s", listenConn->fd, ipbuf);
 
     const char *jobStatus = AsyncJob::status();
     buf.append(jobStatus, strlen(jobStatus));
@@ -153,7 +153,7 @@ void
 Comm::TcpAcceptor::setListen()
 {
     errcode = errno = 0;
-    if (listen(conn->fd, Squid_MaxFD >> 2) < 0) {
+    if (listen(listenConn->fd, Squid_MaxFD >> 2) < 0) {
         errcode = errno;
         debugs(50, DBG_CRITICAL, "ERROR: listen(..., " << (Squid_MaxFD >> 2) << ") system call failed: " << xstrerr(errcode));
         return;
@@ -163,9 +163,9 @@ Comm::TcpAcceptor::setListen()
 #ifdef SO_ACCEPTFILTER
         struct accept_filter_arg afa;
         bzero(&afa, sizeof(afa));
-        debugs(5, DBG_IMPORTANT, "Installing accept filter '" << Config.accept_filter << "' on " << conn);
+        debugs(5, DBG_IMPORTANT, "Installing accept filter '" << Config.accept_filter << "' on " << listenConn);
         xstrncpy(afa.af_name, Config.accept_filter, sizeof(afa.af_name));
-        if (setsockopt(conn->fd, SOL_SOCKET, SO_ACCEPTFILTER, &afa, sizeof(afa)) < 0) {
+        if (setsockopt(listenConn->fd, SOL_SOCKET, SO_ACCEPTFILTER, &afa, sizeof(afa)) < 0) {
             int xerrno = errno;
             debugs(5, DBG_CRITICAL, "WARNING: SO_ACCEPTFILTER '" << Config.accept_filter << "': '" << xstrerr(xerrno));
         }
@@ -173,7 +173,7 @@ Comm::TcpAcceptor::setListen()
         int seconds = 30;
         if (strncmp(Config.accept_filter, "data=", 5) == 0)
             seconds = atoi(Config.accept_filter + 5);
-        if (setsockopt(conn->fd, IPPROTO_TCP, TCP_DEFER_ACCEPT, &seconds, sizeof(seconds)) < 0) {
+        if (setsockopt(listenConn->fd, IPPROTO_TCP, TCP_DEFER_ACCEPT, &seconds, sizeof(seconds)) < 0) {
             int xerrno = errno;
             debugs(5, DBG_CRITICAL, "WARNING: TCP_DEFER_ACCEPT '" << Config.accept_filter << "': '" << xstrerr(xerrno));
         }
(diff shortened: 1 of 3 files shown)


rousskov (changes requested):
> PR description: Do not enter connections for orphanage until their ownership is passed into the async queue for a Subscriber.

IMO, this PR is a step backwards and should not be merged: The proposed code is less exception safe (exceptions may lead to bogus complains about orphaned connections) and is unnecessary explicit (requiring more thought/maintenance/etc. instead of properly relying on RAII).

The rest of my comments are essentially attempts at clarifications.

----

> FD begin their lifecycle belonging to TcpAcceptor which is responsible for closing per Comm::Connection API.

The responsibility to close does not imply the need to close _explicitly_.

"Belonging" is not a very accurate adjective in this context IMO. A Connection actually does not belong to any job when it is born -- it is always born orphaned. And this particular job never really accepts the ownership of the connection it creates.

> That means FD which have issues within the TcpAcceptor logic should be close()'d with a specific ERROR or WARNING message displayed about the reason for closure.

Yes, errors and warnings are needed in most cases, but no explicit closure is required.

> Do not enter connections for orphanage until their ownership is passed into the async queue for a Subscriber.

The "enter orphanage when ownership is passed from X to Y" reads like an oxymoron to me: A connection is either an orphan or is owned. An "owner" that does not properly close the connection is not really an owner in a true sense of that word. It could be considered to be a "broken owner", I guess, but that complication is unnecessary here IMO. We probably do not have to agree on this terminology right now, but having this simple dichotomy/classification would help with understanding some of the API and code (choices) IMO.

yadij:
> 
> 
> > PR description: Do not enter connections for orphanage until their ownership is passed into the async queue for a Subscriber.
> 
> IMO, this PR is a step backwards and should not be merged: The proposed code is less exception safe (exceptions may lead to bogus complains about orphaned connections) and is unnecessary explicit (requiring more thought/maintenance/etc. instead of properly relying on RAII).
> 

This PR is properly using RAII. When exceptions happen the connection is closed by RAII implicit destruct and claims of bug happening are logged.


> The rest of my comments are essentially attempts at clarifications.
> 
> > FD begin their lifecycle belonging to TcpAcceptor which is responsible for closing per Comm::Connection API.
> 
> The responsibility to close does not imply the need to close _explicitly_.
> 

The collection of seriously nasty issues represented by Bug 3329 are what impose that requirement. To catch such issues we have owners responsibly calling close() and the RAII implicit case logging a bug situation to be identified and fixed.

> "Belonging" is not a very accurate adjective in this context IMO. A Connection actually does not belong to any job when it is born -- it is always born orphaned. And this particular job never really accepts the ownership of the connection it creates.
> 

False. TcpAcceptor is the owner until it calls notify() to begin delivery to some Subscriber, or when it drops its local reference. This PR is about making those two points be the "safe" exits from TcpAcceptor.

> > That means FD which have issues within the TcpAcceptor logic should be close()'d with a specific ERROR or WARNING message displayed about the reason for closure.
> 
> Yes, errors and warnings are needed in most cases, but no explicit closure is required.
> 

see Comm::~Connection() code. Owners either call close() explicitly or bogus log entries about being an orphan are recorded by RAII implicit closure.

I would dearly like to let RAII implicit closure not be a worry. But we still have issues with fd_table and signals causing things like two samples below (yes, until recently server connection orphans were easier to replicate):

"
Bug 3329:
 Christos Tsantilas 2015-04-24 11:05:29 UTC

A description of the problem:

When an idle server-side pinned connection closes then:
 - The ConnStateData::clientPinnedConnectionRead, the server side pinned
   connection read handler,  will be called with the io.flag set to 
   Comm::ERR_CLOSING. The server-side Comm::Connection object will not closed
   inside this method in this case, the close handler expected to do the final
   clean up

 - The ConnStateData::clientPinnedConnectionClosed close handler will be called
   but it does not close the server-side Comm::Connection object too.

 - The related fde object is marked as non open fde::flags.open = false,
   but the Comm::Connection server-side object is still marked as valid.
"

"
 Christos Tsantilas 2015-06-02 11:09:51 UTC

I am also attaching an actions sequence which can cause this bug:

  - Ssl::PeerConnector successfully connects to SSL server.
  - The ConnStateData::httpsPeeked called and pins the connection
  - A pinned squid-to-server connection expired and closes.
  - There is not any timeout Handler so the connection is closed with comm_close(fd).
  - The FD is reused by ICAP or HTTP or other client connection
  - The ConnStateData object of the pinned connection destroyed so the
    ConnStateData::pinning::serverConnection destructor called which closes
    an FD which is uded by other objects
  - Bad thinks happens.
 "


> > Do not enter connections for orphanage until their ownership is passed into the async queue for a Subscriber.
> 
> The "enter orphanage when ownership is passed from X to Y" reads like an oxymoron to me: A connection is either an orphan or is owned.

Between TcpAcceptor and FooServer Jobs there is an indefinite amount of time where the queue of AsyncCall are waiting to be dialed. If things go bad during that period Bug 3329 applies: "Bad Things Happen" and an orphan dies.

The same could be true of the periods spent between other Jobs too.

rousskov:
> > IMO, this PR is a step backwards and should not be merged: The proposed code is less exception safe (exceptions may lead to bogus complains about orphaned connections) and is unnecessary explicit (requiring more thought/maintenance/etc. instead of properly relying on RAII).
 
> When exceptions happen the connection is closed by RAII implicit destruct and claims of bug happening are logged.

Yes, the connection will be closed by RAII, but exceptions are not bugs, so any logged "claims of bug" would be wrong. This problem does not exist in the official code, so introducing such log records is a step backwards.


> > The rest of my comments are essentially attempts at clarifications.
> > > FD begin their lifecycle belonging to TcpAcceptor which is responsible for closing per Comm::Connection API.
> > 
> > 
> > The responsibility to close does not imply the need to close _explicitly_.
 
> The collection of seriously nasty issues represented by Bug 3329 are what impose that requirement. 

I disagree that connection ownership problems (that still do exist) require or force an explicit connection closure (in this PR context). The whole idea of supporting a known orphanage state is to prevent the need for many such closures (and most explicit closures are a _problem_ rather than a desirable requirement or goal because sooner or later they become missing, forgotten, or misplaced).

> To catch such issues we have owners responsibly calling close() and the RAII implicit case logging a bug situation to be identified and fixed.

There are two very different implicit closure cases: A known/expected orphan and an unexpected orphan. The former is not a "bug situation" -- there is nothing to "fix" there. The latter is indeed a bug that should be triaged and fixed. The official code (in this PR scope) creates known orphans. That is not a bug.

> > "Belonging" is not a very accurate adjective in this context IMO. A Connection actually does not belong to any job when it is born -- it is always born orphaned. And this particular job never really accepts the ownership of the connection it creates.
 
> False. TcpAcceptor is the owner until it calls notify() to begin delivery to some Subscriber, or when it drops its local reference. 

It is a matter of definition -- the concept of "ownership" is not well defined in this scope. Our use of the words "own" and "belong" differ. Neither usage is "true" or "false" or, to be more precise, each is either true or false depending on the definition in use by the reader.

> This PR is about making those two points be the "safe" exits from TcpAcceptor.

They are already (equally) safe in the official code.

> see Comm::~Connection() code. Owners either call close() explicitly or bogus log entries about being an orphan are recorded by RAII implicit closure.

Please interpret my comments as if I am well aware of that code. AFAICT, you are missing the fact that some orphans (those marked with `COMM_ORPHANED`) are _expected_, and those orphans are logged as such. Being an orphan is not a bug per se. Only _unexpected_ orphans point to (and logged as) bugs.
 

> Between TcpAcceptor and FooServer Jobs there is an indefinite amount of time where the queue of AsyncCall are waiting to be dialed. If things go bad during that period Bug 3329 applies: "Bad Things Happen" and an orphan dies.

That part of the discussion is outside this PR scope because this PR does not change the orphaned state of the Connection during async queuing time -- the _queued_ connection was orphaned before this PR and remains orphaned after this PR. As you know, we agree that there are many Connection ownership problems in Squid code (including in-scope code). I am working on a comprehensive solution to address them, but all that does not affect _this_ PR.

yadij:
> AFAICT, you are missing the fact that some orphans (those marked with COMM_ORPHANED) are expected, and those orphans are logged as such. Being an orphan is not a bug per se. Only unexpected orphans point to (and logged as) bugs.

Sigh. (A) expected crash/orphan vs (B) unexpected crash/orphan. Which is *better* ? neither. So why call code (C) [explicit close()]  that prevents a crash entirely worse and block its creation in favour of (A).

rousskov:
> (A) expected crash/orphan vs (B) unexpected crash/orphan. Which is better ? neither. So why call code (C) [explicit close()] that prevents a crash entirely worse and block its creation in favour of (A).

The above logic assumes that expected orphans lead to crashes. That assumption is false -- bugs notwithstanding, expected orphans do not lead to crashes. _That_ is  the primary difference between an unexpectedly orphaned Connection (a bug) and a purposefully orphaned one (an intended behavior that is expected to be harmless). Are there bugs? Yes! Does this PR fix bugs? None were identified so far.

Connection lifetime management is a difficult subject related to many misconceptions in the official code and developer knowledge. I have studied the issue, can explain what is actually going on in most cases, can help resolve most misunderstandings, and am working on a comprehensive solution. We can fix specific bugs and should finalize a comprehensive long-term solution. AFAICT, this PR currently does neither and complicates future application of any comprehensive solution.

I suggest closing this spur-of-the-moment PR and revisiting the issue when the comprehensive solution is ready for your review. This will save time and give us a much better chance of success than trying to find bugs and negotiate their fixes in this PR environment, where you are not trying to fix a _specific_ bug (as in "here is the test cases that crashes official Squid and that this PR fixes"), and we do not even have a common vocabulary to talk about these complicated issues!

If you insist, we do not have to actually _close_ this PR; we can just suspend this discussion instead.

If none of the above is acceptable to you, then I see no other option than to request a precise description of a specific "crash" that this PR is allegedly "preventing". Referring to bug reports about already fixed crashes is not sufficient, of course, especially when those bug reports are dealing with a different underlying problem. If a specific crash is identified, then we will be able to discuss whether this PR needs any adjustments to properly address that specific issue. I bet this PR fixes no crashes -- the "crash" you are thinking about either does not exist or is not addressed by this PR -- but, if you insist, you can force me to evaluate bug candidates.

src/comm/TcpAcceptor.cc
```diff
@@ -375,8 +380,6 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
     // so we end up with a uniform "(HTTP|FTP-data|HTTPS|...) remote-ip:remote-port"
     fd_open(sock, FD_SOCKET, "HTTP Request");
     details->fd = sock;
-    details->enterOrphanage();
```
eduard-bagdasaryan, line 378: IMO this (removed) call was placed correctly - the Connection was considered as 'orphan' since the 'details' variable (created and passed from acceptOne()) is not stored in TcpAcceptor itself. In other words, if an exception occurs and passes beyond acceptOne() and oldAccept() the 'details' will be lost.

src/comm/TcpAcceptor.cc
```diff
@@ -323,10 +324,14 @@ Comm::TcpAcceptor::notify(const Comm::Flag flag, const Comm::ConnectionPointer &
         params.xaction = new MasterXaction(XactionInitiator::initClient);
         params.xaction->squidPort = listenPort_;
         params.fd = conn->fd;
+        conn->enterOrphanage();
```
eduard-bagdasaryan, line 327: This line looks wrong because 'conn' is the _listening_ connection; newConnDetails is the connection passed here from acceptOne() should be the subject of 'orphanage'.

src/comm/TcpAcceptor.cc
```diff
@@ -323,10 +324,14 @@ Comm::TcpAcceptor::notify(const Comm::Flag flag, const Comm::ConnectionPointer &
...
         params.conn = params.xaction->tcpClient = newConnDetails;
         params.flag = flag;
         params.xerrno = errcode;
         ScheduleCallHere(call);
+    } else {
+        // disconnect clients we cannot respond to
+        conn->close();
```
eduard-bagdasaryan, line 334: This will close the _listening_ connection.

eduard-bagdasaryan (changes requested):
@yadij , please see my comments.

src/comm/TcpAcceptor.cc
yadij: Fixed in b4e50a8

src/comm/TcpAcceptor.cc
yadij: Fixed in b4e50a8

src/comm/TcpAcceptor.cc
yadij: Going by Alex's definition of orphans as "expected" vs "unexpected" errors the move from here to notify() is correct.

"expected" errors and exceptions are caught and handled before leading to notify() which should perform the closure or orphanage. All others are "unexpected" and should not be silenced by the orphanage flag being set this early.

rousskov: > Eduard: IMO this (removed) call was placed correctly - the Connection was considered as 'orphan' since the 'details' variable (created and passed from acceptOne()) is not stored in TcpAcceptor itself.

Yes, `details` is a correctly handled orphan here and, hence, the enterOrphanage() call should not be moved lower. If anything, the call can be moved _higher_ but that would require enlarging my definition of an orphan Connections to also include not-opened connections (and that definition adjustment is difficult to make correctly because it requires defining Connection ownership).

> Amos: Going by Alex's definition of orphans as "expected" vs "unexpected" errors the move from here to notify() is correct.

I do not know what "orphan" definition you are using, but there is nothing about expected or unexpected errors in my definition from https://github.com/squid-cache/squid/pull/837#discussion_r645998210 (quoted below):

> Technically, an orphan Connection object is a Connection object that gets destructed in an open state (i.e. with a non-negative descriptor). Some _rare_ orphan Connection cases are not covered by this statement, but I am omitting them here for simplicity sake. On a more human level, an orphan Connection object is a Connection object without an owner (where Connection owner is code that explicitly closes the connection or has an associated connection closure handler). 

We should call Connection::enterOrphanage() for x when the following two conditions are both true:

* Connection x is an orphan (see my definition quoted above). For example, placing a "throw 42;" statement immediately after the enterOrphanage() call would result in x destruction while x->fd is not negative.
* We are OK with Connection x being destructed in an open state. Such destruction and the associated fd auto-closure is not a bug (for this x) but a desirable outcome.

The specific enterOrphanage() call in question appears to satisfy all these criteria. I see no reason to (re)move it.

> "expected" errors and exceptions are caught and handled before leading to notify() which should perform the closure or orphanage.

That assertion looks false to me as far as exceptions are concerned: Plenty of function calls right below the existing enterOrphanage() call may throw an exception (now or in the future code). Those exceptions are not caught by any code that can close the open Connection in question (that is why that Connection object is an orphan!). And none of those exceptions would lead to notify() with this Connection object as a parameter.

The existing enterOrphanage() call essentially says: If the code below throws, please automatically close this Connection without complaining -- it is the best thing to do to this Connection under those circumstances. I see no problems with that statement. Do you?

rousskov: > yadij added S-waiting-for-reviewer and removed S-waiting-for-author labels

@yadij, I think resolving this change request (and my related higher level concerns expressed earlier) is still on your side so I am relabeling this with S-waiting-for-author. Please correct me if I misinterpreted the PR state.

yadij: Update: when the Comm::Descriptor proposal is merged I will update this PR to be just a naming fix of s/conn/listenConn/

rousskov: > when the Comm::Descriptor proposal is merged I will update this PR to be just a naming fix of s/conn/listenConn/

Understood. I labeled the PR as `S-waiting-for-PR` while we wait for #1103.

FWIW, I do not think listenConn is a good new name because the object does not represent a connection (we are simply abusing a Connection class for something that does not represent a connection or connection end point). Something like `listeningSocket_`, `'listener_`, or even `listen_` would be better IMO.

yadij:
Closing as superseded by PR #1238

PR #843, opened 2021-06-12
Fix sslcrtd and external_acl helper name lifetimes

rousskov:
helperShutdown: <binary garbage> #Hlpr523 shutting down.
    helperShutdown: See example.net/legal-terms #Hlpr4456 shutting down.

Busy "class helper" objects often outlive configuration that was used to
create them. Ideally, the supplied helper category name should be copied
and maintained by the helper object itself, but that long-term solution
requires a lot more work (TODO) due to out-of-scope bugs.

### DIFF
--- a/src/external_acl.cc
+++ b/src/external_acl.cc
@@ -1115,7 +1115,7 @@ externalAclInit(void)
             p->cache = hash_create((HASHCMP *) strcmp, hashPrime(1024), hash4);
 
         if (!p->theHelper)
-            p->theHelper = new helper(p->name);
+            p->theHelper = new helper("external_acl_type");
 
         p->theHelper->cmdline = p->cmdline;
 
--- a/src/helper.h
+++ b/src/helper.h
@@ -65,6 +65,7 @@ class helper
     CBDATA_CLASS(helper);
 
 public:
+    /// \param name admin-visible helper category (with this process lifetime)
     inline helper(const char *name) :
         cmdline(NULL),
         id_name(name),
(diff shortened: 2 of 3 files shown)


rousskov:
> The existing helper category names are inconsistent. 

```C++
new helper("basicauthenticator");
new helper("digestauthenticator");
new statefulhelper("negotiateauthenticator");
new statefulhelper("ntlmauthenticator");
new helper(p->name); // now "external_acl_type"
new helper("redirector");
new helper("store_id");
new helper(Ssl::TheConfig.ssl_crtd); // now "sslcrtd_program"
new helper("ssl_crt_validator");
```

Let me know if I missed any helpers, especially helpers created with names other than string literals.

yadij:
The other helpers I'm aware of use popen() or ipcCreate().

Removed this text from the commit message since "we should" statements about wishlist work are not relevant to code changes:

> We should also
> make all level-0/1 messages about helpers use the same helper
> identification method (currently a mix of the helper category name and a
> helper program name, at least).
> 
> The existing helper category names are inconsistent. The two new names
> repeat the corresponding squid.conf directive names, to assist the admin
> in classifying the helper in level-0/1 messages and to avoid (or
> postpone) arguments about the best long-term name selection approach.

yadij (approved):

rousskov:
> "we should" statements about wishlist work are not relevant to code changes

They are often very relevant. In this case, the sentence removed from the first paragraph detailed one of the existing problems that prevented this PR from implementing a better fix. The second removed paragraph explains PR choice of helper names (so that the developers that continue this work do not have to start this evaluation from scratch and attempt to guess how these names were chosen given the surrounding inconsistency). Both removed statements help understand the developer state of mind which is often critical in understanding why the changes were implemented the way they were implemented.

PR #844, opened 2021-06-15
Fix ClpMap build error on arm32

kinkie:
static_assert ... "Sum() return type can fit its (unsigned) result"
    in ClpMap member instantiation inside sslCrtvdHandleReplyWrapper()

Honor Sum<T>() caller's explicit request to use a specific accumulation
type T instead of guessing the right type for inner sum iterations
(based on the second, third, etc. arguments) and hitting static
assertions when those guesses were wrong because some of those arguments
used types smaller than T.

This fix also allows Sum() callers to avoid explicit T when the first
argument is already the largest one. Callers should not be forced to be
explicit at all, but computing the largest type is a complex known TODO
outside this fix scope.

### DIFF
--- a/src/base/ClpMap.h
+++ b/src/base/ClpMap.h
@@ -194,12 +194,13 @@ ClpMap<Key, Value, MemoryUsedBy>::MemoryCountedFor(const Key &k, const Value &v)
 
     // approximate calculation (e.g., containers store wrappers not value_types)
     return Sum<uint64_t>(
-               keySz,
                // storage
                sizeof(typename Entries::value_type),
                MemoryUsedBy(v),
                // index
-               sizeof(typename Index::value_type));
+               sizeof(typename Index::value_type),
+               // key
+               keySz);
 }
 
 template <class Key, class Value, uint64_t MemoryUsedBy(const Value &)>
--- a/src/base/ClpMap.h
+++ b/src/base/ClpMap.h
@@ -194,13 +194,12 @@ ClpMap<Key, Value, MemoryUsedBy>::MemoryCountedFor(const Key &k, const Value &v)
 
     // approximate calculation (e.g., containers store wrappers not value_types)
     return Sum<uint64_t>(
+               keySz,
                // storage
                sizeof(typename Entries::value_type),
                MemoryUsedBy(v),
                // index
-               sizeof(typename Index::value_type),
-               // key
-               keySz);
+               sizeof(typename Index::value_type));
 }
 
 template <class Key, class Value, uint64_t MemoryUsedBy(const Value &)>
--- a/src/SquidMath.h
+++ b/src/SquidMath.h
@@ -80,8 +80,8 @@ Sum(const T a, const U b) {
 template <typename T, typename... Args>
 Optional<T>
 Sum(const T first, Args... args) {
-    if (const auto others = Sum(args...)) {
-        return Sum(first, others.value());
+    if (const auto others = Sum<T>(args...)) {
+        return Sum<T>(first, others.value());
     } else {
         return Optional<T>();
     }


kinkie:
Test builds:

https://build.squid-cache.org/job/anybranch-arm32-matrix/17/
https://build.squid-cache.org/job/anybranch-arm64-matrix/4/
https://build.squid-cache.org/job/anybranch-matrix/28/

src/base/ClpMap.h
```diff
@@ -194,12 +194,13 @@ ClpMap<Key, Value, MemoryUsedBy>::MemoryCountedFor(const Key &k, const Value &v)
...
     return Sum<uint64_t>(
-               keySz,
                // storage
                sizeof(typename Entries::value_type),
                MemoryUsedBy(v),
                // index
-               sizeof(typename Index::value_type));
+               sizeof(typename Index::value_type),
+               // key
+               keySz);
```
rousskov, line 203: > Sort arguments supplied to Sum<>() to be in descending order of type size.
> This prevents a static_assert meant to prevent the result of a sum not fitting in the return type

The PR description does not disclose this important detail yet, but I assume you are referring to the first of the following two asserts:

```C++
    static_assert(sizeof(T) >= sizeof(U), "Sum() return type can fit its (unsigned) result");
```

```C++
    static_assert(sizeof(T) >= sizeof(U), "Sum() return type can fit its (signed) result");
```


The return type of this particular invocation of Sum() is uint64_t, set by `Sum<uint64_t>()` caller explicitly. AFAICT, that type can accommodate any unsigned Sum() parameter we can encounter in Squid so a correct Sum() implementation should not assert in this use case.

I hope that to properly fix this bug, the two "typeless" Sum() calls inside the second Sum() definition should be given a specific type -- `Sum<T>` -- instead of letting them to determine the type based on their first argument. The second call is probably OK as is, but I think it is best to be explicit and consistent across both calls:

```C++
Optional<T>
Sum(const T first, Args... args) {
    if (const auto others = Sum<T>(args...)) {
        return Sum<T>(first, others.value());
    } else {
        return Optional<T>();
    }
}
```


That approach would honor the wishes of Sum() callers that specify the type explicitly. However, I do not know whether it would break some other use cases that we currently support. Please try if you do not have better ideas (that do not rely on callers sorting Sum() arguments).

rousskov (changes requested):
Thank you for working on improving arm32 support!

I hope the bug is elsewhere -- we do not want to task Sum() callers with guessing the right parameter order! Please see the inlined change request for details.

Please also add the (abridged if necessary) assertion text to the PR description so that the problem is easy to find in `git log` and so that the reader does not have to guess which static assertion got violated.

src/base/ClpMap.h
kinkie: The error is (on clang 3.8):

`
../../../../src/SquidMath.h:55:5: error: static_assert failed "Sum() return type can fit its (unsigned) result"
../../../../src/SquidMath.h:55:5: error: static_assert failed "Sum() return type
 can fit its (unsigned) result"
    static_assert(sizeof(T) >= sizeof(U), "Sum() return type can fit its (unsign
ed) result");
    ^             ~~~~~~~~~~~~~~~~~~~~~~
../../../../src/SquidMath.h:84:16: note: in instantiation of function template specialization 'Sum<unsigned int, unsigned long long, 0>' requested here
        return Sum(first, others.value());
               ^
../../../../src/SquidMath.h:83:29: note: in instantiation of function template specialization 'Sum<unsigned int, unsigned long long, unsigned int>' requested here
    if (const auto others = Sum(args...)) {
                            ^
../../../../src/base/ClpMap.h:196:12: note: in instantiation of function template specialization 'Sum<unsigned long long, unsigned int, unsigned long long, unsigned int>' requested here
    return Sum<uint64_t>(
           ^
../../../../src/base/ClpMap.h:218:37: note: in instantiation of member function
'ClpMap<SBuf, RefCount<Ssl::CertValidationResponse>, &Ssl::CertValidationResponse::MemoryUsedByResponse>::MemoryCountedFor' requested here
    const auto memoryRequirements = MemoryCountedFor(key, v);
                                    ^
../../../../src/base/ClpMap.h:67:55: note: in instantiation of member function 'ClpMap<SBuf, RefCount<Ssl::CertValidationResponse>, &Ssl::CertValidationResponse::MemoryUsedByResponse>::add' requested here
    bool add(const Key &key, const Value &v) { return add(key, v, defaultTtl_);
}
                                                      ^
../../../../src/ssl/helper.cc:297:55: note: in instantiation of member function
'ClpMap<SBuf, RefCount<Ssl::CertValidationResponse>, &Ssl::CertValidationResponse::MemoryUsedByResponse>::add' requested here
        (void)Ssl::CertValidationHelper::HelperCache->add(crtdvdData->query, validationResponse);
                                                      ^
`

I attacked the problem from this angle because the intent of the author of SquidMath.h seems reasonable and not dodgy, and arm32's type system and this specific code location is the only one where this behaviour manifests adversely

I'll try your proposal and see if it helps

kinkie: Your suggestion seems to work well. Updating the PR accordingly

rousskov: > I attacked the problem from this angle because the intent of the author of SquidMath.h seems reasonable and not dodgy

I agree that there is nothing unreasonable or dodgy in authors intent here (although I am biased because I wrote that code IIRC). However, I do not see the relevance of that "honorable" intent to this PR. The Sum() author missed the fact that when the caller forces the type, that forced type should apply to the whole sequence instead of guessing the type based on the second, third, etc. arguments and triggering static assertions when those guesses go wrong.

Long-term, Sum() should compute the largest type so that there is no need for explicit type specification in most callers. That is a known TODO. Until then, callers dealing with a mixture of types that starts with a smaller type have a workaround -- specify the return type explicitly. Like any workaround, this is not ideal, but I cannot think of a better option for now.

kinkie: Agreed. My fix was even less ambitious :D
But in short, it works and doesn't introduce regressions in testing so far - pending unit testing that is coming

rousskov (approved):
> Your suggestion seems to work well. Updating the PR accordingly

You forgot to update the PR description. I fixed that and adjusted to the PR title to emphasize that this is a build fix rather than a runtime bug fix. I am so glad we added those static assertions! Please adjust PR metadata further as needed.

src/base/ClpMap.h
rousskov: Oops. I resolved this too early -- you forgot to undo this unwanted change.

rousskov (changes requested):

kinkie:
> I am so glad we added those static assertions! Please adjust PR metadata further as needed.

In practice things were working, the numbers involved are tiny and would be unlikely to overflow.
But I agree that covering the bases here was a good call, and I'm happy that multi-platform testing highlighted this corner case that would have been undetected in a 64-bit only world.


I think the PR description is good to go as it it after your changes, so no further changes from me if you agree it's a good merge candidate

src/base/ClpMap.h
kinkie: I actually had thought about that and decided to leave it in, on account that two fixes is better than one, but I can undo that if you think it's better

rousskov: > I can undo that if you think it's better

Yes, please -- I still think this change is unwanted and should be undone. I wish the burden of proof would be on those adding unnecessary changes, but here is my detailed rationale for asking you to undo this change:

* This change is unnecessary, obsoleted by the Sum() definition fix which covers this and other cases.
* This change makes the code slightly less logical -- the official code starts with accounting for the key size and then proceeds to accounting for the value size, which is, I bet, the order that most humans find natural when thinking about key:value maps.
* The change in items order forced you to add a "foo is foo" comment which was not necessary in the official version. That new comment actually goes against the pattern of the other two (necessary) comments; there might be some useful implication in the fact that the key size belongs to neither "storage" not "index" section, but I have not investigated this further.

kinkie: Ok, done

rousskov (approved):

PR #848, opened 2021-06-16
SSL_SESSION::master_key cannot be falsy

kinkie:
SSL_SESSION::master_key data member is an array. It cannot be falsy.

This fixes clang -Wpointer-bool-conversion.

### DIFF
--- a/compat/openssl.h
+++ b/compat/openssl.h
@@ -175,7 +175,7 @@ extern "C" {
     inline size_t
     SSL_SESSION_get_master_key(const SSL_SESSION *session, unsigned char *outStart, size_t outSizeMax)
     {
-        if (!session || !session->master_key || session->master_key_length <= 0)
+        if (!session || session->master_key_length <= 0)
             return 0;
 
         const auto sourceSize = static_cast<size_t>(session->master_key_length);


rousskov (changes requested):
> SSL_SESSION::master_key is an array

Not always. Have you seen the corresponding Coverity triage [note](https://scan5.coverity.com/reports.htm#v46217/p10191/fileInstanceId=194348256&defectInstanceId=51558848&mergedDefectId=1485164) for CID 1485164?

kinkie:
I had not ,thanks for highlighting.
I guess preprocessor magic will be needed to cover both cases, or some way to disable this specific check here. It doesn't add a lot of value anyway

rousskov:
> I guess preprocessor magic will be needed to cover both cases, or some way to disable this specific check here. It doesn't add a lot of value anyway

Perhaps there is still a misunderstanding: This is a runtime tls_key_log bug -- in some (still relatively popular) environments, the code is not checking what it should be checking and then probably not even accessing what it should be accessing after the checks succeeds. The compiler warning is just a side effect of that bug. This bug needs fixing, but the fix is not trivial due to OpenSSL `struct` layout changes. Most likely, the correct fix will be driven by ./configure-time tests of OpenSSL `struct` layouts that Squid is being built with.

kinkie:
Closing the PR and runnign some tests behind the scenes. I'll reopen it once a workable solution is found

rousskov:
> Alex: OpenSSL SSL_SESSION structure layout has changed. ... I suspect that the reinterpret_cast<> assumes the wrong layout

I think I figured it out. I was looking at `ssl_session_asn1_st` structure, which did change. I was looking at the wrong structure though! Instead, we should be looking at `ssl_session_st` structure, which did _not_ change (as far as master_key is concerned). 

In the `ssl_session_st` structure, the master_key is an array in all those older OpenSSL versions. We do not need to test master_key for being nil, and the reinterpret_cast is not terribly wrong (that cast may not be needed at all, but that is a different and very minor problem).

@kinkie, do you want to reopen this PR and remove the unnecessary "master_key is not nil" check (no other changes should be needed compared to master)? Alternatively, I can remove that check in a new branch/PR. Your call.

kinkie:
I'll reopen this, clean it and submit again. The work is already done, after all

rousskov (approved):
Thank you for fixing this! Please check that my PR title/description edits are OK with you and adjust further as needed.

I wonder whether there is a simple static_assert that can flag this code if OpenSSL changes the master_key type to become a pointer, but I do not have the time to investigate that right now.

PR #850, opened 2021-07-02
Relese Prep for 4.16 and 5.0.7

yadij:
Documentation only


rousskov (commented):
Thank you for summarizing recent changes. As usual, I have not compared every reported (and recursively included) change with the commit history for the corresponding branch, but I see no red flags.

rousskov (approved):

PR #855, opened 2021-07-06
Fix build on Ubuntu 21

xypron:
The `-Wunused-result` warning for setuid(0) call was observed on Ubuntu
21.04 (when cross compiling Squid for Ubuntu 21.10).

Signed-off-by: Heinrich Schuchardt <heinrich.schuchardt@canonical.com>

### DIFF
--- a/src/tools.cc
+++ b/src/tools.cc
@@ -582,7 +582,10 @@ enter_suid(void)
     }
 #else
 
-    setuid(0);
+    if (setuid(0) < 0) {
+        const auto xerrno = errno;
+        debugs(21, 3, "enter_suid: setuid failed: " << xstrerr(xerrno));
+    }
 #endif
 #if HAVE_PRCTL && defined(PR_SET_DUMPABLE)
     /* Set Linux DUMPABLE flag */
--- a/src/tools.cc
+++ b/src/tools.cc
@@ -584,7 +584,7 @@ enter_suid(void)
 
     if (setuid(0) < 0) {
         const auto xerrno = errno;
-        debugs(21, 3, "enter_suid: setuid failed: " << xstrerr(xerrno));
+        debugs(21, 3, "setuid(0) failed: " << xstrerr(xerrno));
     }
 #endif
 #if HAVE_PRCTL && defined(PR_SET_DUMPABLE)


src/tools.cc
```diff
@@ -582,7 +582,10 @@ enter_suid(void)
     }
 #else
 
-    setuid(0);
+    if (setuid(0) < 0) {
+        const auto xerrno = errno;
+        debugs(50, DBG_IMPORTANT, "WARNING: enter_suid: setuid(0): " << xstrerr(xerrno));
```
rousskov, line 587: It is a big jump from completely ignoring setuid(0) failure to informing the admin about the failure by default, especially when the code immediately above still ignores (for most practical purposes) setresuid() failures. Do we want to introduce this inconsistency with the setresuid() call?

If you believe that this inconsistency is a good idea, could you please detail why? That explanation may help me suggest a better wording for the WARNING message. The proposed wording has several problems, but the biggest one is that it does not really tell a sysadmin what they should do or what they should actually be worried about. Ideally, level-1 warnings should be more admin-friendly.

If you were just duplicating this code from no_suid() and do not really care about the warning existence and the exact text, just let me know!

rousskov (changes requested):
Thank you for working on fixing this build problem!

I can help with polishing the PR description to clear the [M-failed-description](https://github.com/measurement-factory/anubis/blob/master/README.md#pull-request-labels) label if you tell me what build environment (compiler/OS) triggers this `-Werror=unused-result` error? We have not seen it in any CI environment...

src/tools.cc
xypron: On Ubuntu 21.04 I am cross compiling for RISC-V using 'sbuild --host=riscv64'.

What do you mean by 'informing the admin'? According to inline comments in src/Debug.h debugfs() is output to the debug stream used in unit tests. There is no output without ENABLE_DEBUG_SECTION.

rousskov: > What do you mean by 'informing the admin'? 

Level-0/1 debugs() lines are printed to cache.log by default. `DBG_IMPORTANT` is level 1.

> According to inline comments in src/Debug.h debugs() is output to the debug stream used in unit tests.

You are misinterpreting the word "this" in that (poorly worded) debugs() API description. That word applies to the special case of enabling stderr logging for unit tests, not to the debugs() API itself.

```
 * Unit tests can enable full debugging to stderr for one
 * debug section; to enable this, #define ENABLE_DEBUG_SECTION to the
 * section number before any header
```

rousskov: > On Ubuntu 21.04 I am cross compiling the ubuntu package for RISC-V using 'sbuild --host=riscv64 -d impish'. Impish is Ubuntu 21.10.

Thank you for that information. I updated the PR title/description accordingly, making Anubis happy. That PR metadata will become the official commit message when this PR is merged by [Anubis](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message). If you know what compiler was used for this build, please let me know and I will add that potentially useful detail as well. Please feel free to adjust PR title/description further as needed, of course -- this is your PR, and I am just facilitating progress towards official merge.

xypron: @rousskov:
Thank you for the explanation. So should we use:
```
debugs(50, 7, "enter_suid: setuid(0): " << xstrerr(xerrno));
```

rousskov: If you want a minimal change, then I would suggest:

```suggestion
        debugs(50, 3, "setuid(0) failed: " << xstrerr(xerrno));
```

This duplicates some of the existing code (bad) but given our poor understanding of the true meaning of these failures and the existing official code state, I think that is acceptable (if nobody offers a better solution).

An ideal fix would require a higher level of understanding (which I currently lack) and the corresponding documentation of setuid() caller choices combined with elimination of setuid() code duplication (probably by introducing a single API for all Squid setuid() callers).

----

debugs() at levels 2+ will print the current function name automatically, so new debus() messages should not say `enter_suid` explicitly like some legacy code you are looking at does.

Similar setresuid() failure is reported at level 3 so I would use that level for this new message as well. Level 7 feels way too high for such a rare and potentially security-related issue.

xypron: @rousskov  I have updated the patch. I adjusted both level and section to match the setresuid message.

rousskov: Thank you. I have updated the message further in 5991d57 and now consider this change request addressed.

rousskov (approved):
I think the current PR should be merged because it does not change runtime behavior (apart from additional optional debugging) while improving compilation chances and debugging. If somebody with a better understanding of setuid() failure issues comes up with a better solution to the problem, this change is not going to make things worse for their future work.

rousskov:
OK to test

xypron:
@rousskov: Thank you for reviewing and your valuable feedback.

yadij (approved):

PR #856, opened 2021-07-06
Fix build on RISC-V

xypron:
Compiling on RISC-V (without an explicit -latomic) fails with

    /usr/riscv64-linux-gnu/include/c++/10/ostream:611:
    undefined reference to __atomic_compare_exchange_1

Use std::atomic<uint8_t>::exchange() to detect whether -latomic
implements 1-byte compare-and-exchange API used by Squid.

Signed-off-by: Heinrich Schuchardt <heinrich.schuchardt@canonical.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -451,6 +451,8 @@ AC_DEFUN([LIBATOMIC_CHECKER],[
 #include <cstdint>
       int
       main(int, char **) {
+          std::atomic<unsigned char> foo(0);
+          foo.exchange(1);
           return std::atomic<uint64_t>{}.is_lock_free() ? 0 : 1;
       }
   ]])],[
--- a/configure.ac
+++ b/configure.ac
@@ -450,10 +450,11 @@ AC_DEFUN([LIBATOMIC_CHECKER],[
 #include <atomic>
 #include <cstdint>
       int
-      main(int, char **) {
-          std::atomic<int> foo(0);
-          foo.exchange(1);
-          return std::atomic<uint64_t>{}.is_lock_free() ? 0 : 1;
+      main(int argc, char **) {
+          return (
+              std::atomic<int>(argc).exchange(0) &&
+              std::atomic<uint64_t>{}.is_lock_free()
+              ) ? 0 : 1;
       }
   ]])],[
     AC_MSG_RESULT(yes)
--- a/configure.ac
+++ b/configure.ac
@@ -452,7 +452,7 @@ AC_DEFUN([LIBATOMIC_CHECKER],[
       int
       main(int argc, char **) {
           return (
-              std::atomic<int>(argc).exchange(0) &&
+              std::atomic<uint8_t>(uint8_t(argc)).exchange(0) &&
               std::atomic<uint64_t>{}.is_lock_free()
               ) ? 0 : 1;
       }
--- a/configure.ac
+++ b/configure.ac
@@ -451,7 +451,7 @@ AC_DEFUN([LIBATOMIC_CHECKER],[
 #include <cstdint>
       int
       main(int, char **) {
-          std::atomic<unsigned char> foo(0);
+          std::atomic<int> foo(0);
           foo.exchange(1);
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -451,6 +451,7 @@ AC_DEFUN([LIBATOMIC_CHECKER],[
 #include <cstdint>
       int
       main(int, char **) {
+          int foo = __atomic_exchange_1(&foo, 2, __ATOMIC_ACQ_REL);
```
rousskov, line 454:
```
$ git grep __atomic_exchange_1 | wc -l
0
```

We should not add non-standard function calls to this libatomic checker because they will break the build on platforms that lack those specific calls (not to mention causing reader confusion from seeing seemingly unused by Squid code calls used for testing the necessity of linking with this standard library).

To fix this build problem, we should figure out what standard std::atomics API requires __atomic_exchange_1() in your environment and then use that API in this checker. I have not checked myself, but I bet that standard API is compare_exchange_strong(). Could you please try testing for that instead?

```
$ git grep compare_exchange
src/ipc/StoreMap.cc:    return s.waitingToBeFreed.compare_exchange_strong(expected, true);
src/ipc/mem/PageStack.cc:    } while (!node.compare_exchange_weak(oldValue, newValue.pack()));
src/ipc/mem/PageStack.cc:    } while (!node.compare_exchange_weak(oldValue, newValue));
```

rousskov (changes requested):
Thank you for working on this build fix. I am sure others will hit the errors you are seeing soon.

xypron:
@rousskov: I have update the patch to call std::atomic::exchange().

rousskov:
OK to test

rousskov:
@xypron, please review _and_ test my recent (untested by me) changes. See commit messages for the rationale. Adjust as needed. Once you are happy with the latest PR code _and_ it works in your environment, I will clear this PR for merging. 

P.S. No need to squash any PR branch commits (or to sync with the latest master). Anubis will do that for you, using PR title/description as the commit message.

xypron:
@rousskov: I used an atomic exchange for an 8 bit variable. Why did you change it to 32bit?

rousskov:
> I used an atomic exchange for an 8 bit variable. Why did you change it to 32bit?

I changed `unsigned char` to `int` to avoid `signed int` to `unsigned char` conversion (where you specified `1` as an argument) and to simplify. If the code _must_ use `unsigned char` or equivalent, then my change was wrong, of course, but we should _document_ (probably in the PR description) why we need an (unsigned?) 8-bit type for this check, to avoid revisiting/breaking this in the future as the code changes.

Also, if the code must use an unsigned 8-bit variable, then we should probably use `uint8_t` instead of `unsigned char` to emphasize that this is about the type signness and size, not character manipulation. (And, to avoid misunderstanding, I do realize that all those 8-bit types may result in the same macnine code -- compiled code is not the point here).

rousskov:
> Alex: If the code _must_ use `unsigned char` or equivalent, then my change was wrong

I now realize that `1` in `__atomic_compare_exchange_1` stands for "1 byte". So, yes, my changes were wrong. Please use `int8_t` instead. If you prefer, I can fix my bug myself -- just let me know.

xypron:
@rousskov: When only int is checked the need for libatomic is not detected properly. The std::atomic<T> template class does not prescribe the signedness of T. We can use uint8_t to check for __atomic_compare_exchange_1().

rousskov (approved):
Thank you for fixing my bug. Using `uint8_t` is good because that is what Squid is using itself. I added a corresponding note to the PR description.

I wish the testing code would be simpler, but I do not have good ideas on how to simplify it while still making sure that the compiler does not warn about always-true conditions or optimize away unused code.

I suspect we should add a few similar lines to test other (bigger) types that Squid uses with atomics: uint16_t, uint32_t, and uint64_t. If you want to add those checks now, please do! However, that improvement can be considered outside _this_ PR scope, and I do not want to bother you with a yet another round of changes and tests, so I am approving this PR as is.

yadij (approved):

PR #861, opened 2021-07-12
Fix SslBump reconfiguration leaking public key memory

rousskov:
X509_get_pubkey() increments key reference count.

Probably leaking since commit 2a268a0.

### DIFF
--- a/src/ssl/gadgets.cc
+++ b/src/ssl/gadgets.cc
@@ -376,8 +376,10 @@ mimicExtensions(Security::CertPointer & cert, Security::CertPointer const &mimic
         DecipherOnly
     };
 
-    EVP_PKEY *certKey = X509_get_pubkey(mimicCert.get());
-    const bool rsaPkey = (EVP_PKEY_get0_RSA(certKey) != nullptr);
+    // XXX: Add PublicKeyPointer. In OpenSSL, public and private keys are
+    // internally represented by EVP_PKEY pair, but GnuTLS uses distinct types.
+    const Security::PrivateKeyPointer certKey(X509_get_pubkey(mimicCert.get()));
+    const auto rsaPkey = EVP_PKEY_get0_RSA(certKey.get()) != nullptr;
 
     int added = 0;
     int nid;


src/ssl/gadgets.cc
```diff
@@ -376,8 +376,10 @@ mimicExtensions(Security::CertPointer & cert, Security::CertPointer const &mimic
         DecipherOnly
     };
 
-    EVP_PKEY *certKey = X509_get_pubkey(mimicCert.get());
-    const bool rsaPkey = (EVP_PKEY_get0_RSA(certKey) != nullptr);
+    // XXX: Add PublicKeyPointer. In OpenSSL, public and private keys are
+    // internally represented by EVP_PKEY pair, but GnuTLS uses distinct types.
+    const Security::PrivateKeyPointer certKey(X509_get_pubkey(mimicCert.get()));
```
rousskov, line 381: The `PrivateKeyPointer = get_pubkey()` line looks odd. Squid currently lacks a type dedicated to holding public keys. Adding one may not be trivial because OpenSSL does not have a distinct type for each type of key (it stores both in one EVP_KEY structure), and because of the suspected disparity with GnuTLS that probably has distinct types. I left that future improvement out of this leak-fixing PR scope.

yadij: FTR: yes GnuTLS uses two types to store X.509 public (gnutls_pcert_st) and private (gnutls_privkey_t) keys.

yadij: IMO it is worth adding the typedefs for a PublicKeyPointer to security/forward.h to avoid adding a new XXX here and re-changing these lines in future. The GnuTLS one does not have to be used yet.

yadij (changes requested):

src/ssl/gadgets.cc
rousskov: > IMO it is worth adding the typedefs for a PublicKeyPointer to avoid adding a new XXX here and re-changing these lines in future.

I (obviously) disagree. It is not yet clear (to me) whether having two types that are interchangeable only when OpenSSL is used is actually a good way forward as far as these supposed-to-be-universal types are concerned. Let's cross that bridge when we have to (and have specific use cases for it).

Please withdraw your API improvement request that is not really required for this specific memory leak fix. If you prefer, I can delete this XXX source comment and/or switch to an OpenSSL-only type (in this OpenSSL-only code).

> FTR: yes GnuTLS uses two types to store X.509 public (gnutls_pcert_st) and private (gnutls_privkey_t) keys.

And even that simple assertion does not actually match Squid code. Squid defines PrivateKeyPointer as gnutls_x509_privkey_int, not gnutls_privkey_t. Again, let's postpone resolving this and other non-trivial issues when we have to resolve them, with real needs/examples we can test.

yadij (approved):

PR #863, opened 2021-07-14
Allow sending "squid -k ..." signals to PID 1

jwcnewton:
Inside a container, Squid usually runs with PID=1. Since "squid -k"
feature was added in 1996 (commit 7690e8e), Squid refused to signal a
PID=1 instance. We do not know why, and it could have been an accident.
We cannot find a good reason to keep prohibiting such signalling.

### DIFF
--- a/src/Instance.cc
+++ b/src/Instance.cc
@@ -81,7 +81,7 @@ GetOtherPid(File &pidFile)
 
     debugs(50, 7, "found PID " << rawPid << " in " << TheFile);
 
-    if (rawPid <= 1)
+    if (rawPid < 1)
         throw TexcHere(ToSBuf("Bad ", TheFile, " contains unreasonably small PID value: ", rawPid));
     const auto finalPid = static_cast<pid_t>(rawPid);
     if (static_cast<int64_t>(finalPid) != rawPid)


rousskov:
Please see https://github.com/squid-cache/squid/pull/583#issuecomment-605669239 for the opposition to this change. That PR got closed (by its author) without reaching consensus regarding the right fix/change. AFAICT, there is not even consensus whether the existing ban on PID=1 is worth preserving. Somebody needs to drive that discussion to form consensus. I am happy to review, but I cannot be that driver right now.

Please do not misinterpret anything I said here or in #583 as an indication that the problem you are trying to address is not real or is not worth addressing.

jwcnewton:
Thanks for the response, @rousskov Seems odd that squid would be opinionated. I imagine most people are running squid inside of a container were squid would be bootstrapped to the entrypoint making it always 1

rousskov:
> Seems odd that squid would be opinionated.

I appreciate where you are coming from, but I do not think Squid is opinionated. Opinionated software projects, AFAICT, reject (high-quality) changes simply because they satisfy a (valid) need outside that software project _narrow_ focus/path/model. That is not what is happening here: Your change is not being rejected and, IMHO, your needs are well within the Squid Project scope! Your use case should be supported, one way or another.

We need to find a good way to address your needs, preferably without hurting others (if any) that rely on PID=1 protections today. I see the following options:

1. Simply remove the existing PID=1 ban after reaching consensus that it solves no problems worth solving. This is current PR state.
2. Replace the existing PID=1 ban with a better mechanism that offers similar protection from signalling the init daemon without creating problems for containerized Squids. Amos [suggested](https://github.com/squid-cache/squid/pull/583#issuecomment-605669239) that Squid should check that process with PID=1 is Squid before signalling that PID. Alternatively, we could immediately fork Squid so that it never runs as PID=1 even if started with that PID. There may be other, better alternatives.
3. Making the protection from signalling the init daemon optional/configurable. This is what abandoned #583 was proposing.

I hope we can reach consensus on either option 1 or option 2 so that we do not have to add a yet another configuration directive for option 3, but I do not know whether that is possible. As I said [earlier](https://github.com/squid-cache/squid/pull/583#issuecomment-605691659), I do not fully understand the threat model that Amos has in mind.

After reading #583 discussion, do you think the existing init-signalling protections are worth keeping? If you think they should be removed, please try to substantiate your opinion. If you think they should be provided, can you think of a way to provide them without adding a configuration directive and without inconveniencing containerized Squids?

yadij:
FTR: line-wrapped the description to make Anubis happy.

jwcnewton:
I think either spawn squid to ensure that PID 1 is never used or don't validate that the process isn't 1, the security threat of using PID 1 is new to me and not one I have ever heard before.

rousskov:
> I think either spawn squid to ensure that PID 1 is never used or don't validate that the process isn't 1, the security threat of using PID 1 is new to me and not one I have ever heard before.

@yadij, do you agree that at least one of those two options is viable and that they are mutually exclusive? Since you may be the only person in this discussion who may consider signalling PID=1 a threat worth protecting against (AFAICT), we need your input to make progress.

yadij:
> I think either spawn squid to ensure that PID 1 is never used or don't validate that the process isn't 1, the security threat of using PID 1 is new to me and not one I have ever heard before.

The attack surface for Squid includes a management API which performs SIGTERM and/or SIGHUP on remotely received commands. As such there is a potential attacker-controlled vector to DoS / terminate any process Squid finds in squid.pid. If that process is *not* a Squid we have unknown behaviour and side effects with a wide scope of potential effect - especially in containers where the PID 1 might not be a well-audited init process.

For my opinion:
  option (1) - I have not formed a detailed opinion, open to discussion still. I would probably prefer to drop pidfile_name directive if we do this.
  option (2) - or variants of it - seem okay to me. As you should see from my proposal.
  option (3) - I really dislike is having anything to do with the PID file in squid.conf. The existence of pidfile_name is why the management API adds risk regarding PID in the first place. The other PR was veto'd because it added to squid.conf, not because it wanted to allow PID 1.

rousskov:
> Amos: The attack surface for Squid includes a management API which performs SIGTERM and/or SIGHUP on remotely received commands. 

What management API are you thinking about, exactly? Since you said "remotely received", I would guess you were thinking about the Squid Cache Manager API. However, that API does not use the PID file at all and, hence, is irrelevant to this PR! AFAICT, the only relevant management commands here are the necessarily "local" commands like `squid -k rotate`. Please clarify your definition of the attack we should be thinking about in this PR scope.

> Amos in https://github.com/squid-cache/squid/pull/583#issuecomment-605669239: FTR: the PID check is to prevent services triggering signals directly on the PID 1 init system daemon.

AFAICT, `squid -k ...` sends a signal as user `nobody` (or equivalent) so even if it does signal the init daemon, that signal cannot actually hurt the system because it will never reach a privileged daemon process. Is the attach surface here limited to Squids that run as root?! Beyond warning admins not to do that, we do not need to think about such unsupported use cases, do we?

----

P.S. Answering the "Does squid.pid still point to a Squid process?" question correctly in a portable way sounds like a difficult problem to me. I doubt we will find volunteers to implement the necessary code changes. Thus, I propose to leave that difficult problem for the future, when somebody needs to address it badly enough to invest the necessary resources. That future may never come.

yadij:
> 
> 
> > Amos: The attack surface for Squid includes a management API which performs SIGTERM and/or SIGHUP on remotely received commands.
> 
> What management API are you thinking about, exactly? Since you said "remotely received", I would guess you were thinking about the Squid Cache Manager API.

I was.

> However, that API does not use the PID file at all 

Here you are mistaken. The API commands mgr:shutdown (SIGTERM), mgr:rotate (SIGUSR1 or SIGQUIT), mgr:reconfigure (SIGHUP), mgr:debug all trigger signals sent to squid.pid. Additionally mgr:config allows remote attacker to identify the location of which squid.pid to effect changes on the process whose API they have access to.

That is the current situation. We also have future plans to allow mgr:* to alter the config settings which would include pid_filename and add even more risk to this scenario.

rousskov:
> Alex: Cache manager API does not use the PID file at all, and, hence, is irrelevant to this PR!

> Amos: The API commands mgr:shutdown (SIGTERM), mgr:rotate (SIGUSR1 or SIGQUIT), mgr:reconfigure (SIGHUP), mgr:debug all trigger signals sent to squid.pid.

I do not think so, and it would not really make sense to read PID from squid.pid and send signals in the cache manager context because the recipient of those cache manager commands is already the intended target (i.e. it is the process which PID is written in squid.pid)!

In SMP case, the intended target is the whole Squid rather than just one kid, but broadcasting cache manager commands to all kids is handled by a general mechanism that propagates external cache manager queries to the Coordinator process, letting Coordinator send (now internalized) queries to each kid. All that is done without using signals, of course. Look for Mgr::Forwarder code path in CacheManager::start().

```C++
Mgr::ShutdownAction::dump(StoreEntry *)
{
    debugs(16, DBG_CRITICAL, "Shutdown by Cache Manager command.");
    shut_down(SIGTERM);
}
...
void
shut_down(int sig)
{
    do_shutdown = sig == SIGINT ? -1 : 1;
    ShutdownSignal = sig;
    // ... irrelevant OS-dependent code to restore the signal handler ...
}
```

For example, mgr:shutdown code shown above calls shut_down() which does not send any signals -- it actually triggers the shutdown sequence for the _current_ process.

If you still disagree, please point me to the cache manager code path that you think sends signals to the PID read from squid.pid.

N.B. This PR is about policing values read from the squid.pid file (or equivalent).

jwcnewton:
Just to check the status of this PR, what is the plan going forward? Happy to close this PR if we feel it's better to fix this issue elsewhere

rousskov:
> Just to check the status of this PR, what is the plan going forward?

The ball is on @yadij side right now. No action is required on your part AFAICT.

If Amos confirms that the original (and now clarified) concerns regarding cache manager API are actually unrelated to this PR, then the known reasons not to merge this PR will disappear and, if no new concerns arise, then either Amos or I will be able to approve this PR for merging (after polishing the PR description). Otherwise, there will be another review iteration (and some code changes may be ultimately required).

> Happy to close this PR if we feel it's better to fix this issue elsewhere

IMHO, this PR is very useful. Once the security concerns are resolved (as discussed above), I hope it will be merged (possibly with some code changes but it is too early to tell; see above). Thus, I do not recommend closing this PR and hope you can tolerate the painful review process, but it is naturally _your_ call.

yadij:
Okay. I accept that the shutdown, rotate, and debug are calling themselves without squid.pid read.

I did find the path for mgr:reconfigure sets opt_send_signal which is followed by sendSignal() that reads squid.pid and sends SIGHUP there. The process for reconfigure is not necessarily the coordinator, and the squid.pid is supposed to be the master.
I'm not sure if this is a reasonable vector though, it would require the machine to already be compromised enough to run squid.

TLDR; withdrawing my objection.

yadij (approved):

rousskov:
> I did find the path for mgr:reconfigure sets opt_send_signal 

Where?! AFAICT, mgr::reconfigure is Mgr::ReconfigureAction which calls reconfigure(). That reconfigure() does not set opt_send_signal (and does not send any signals).

rousskov:
I updated PR title/description to better document what this PR actually changes (and why). With so much confusion throughout the PR history, we must document what is actually happening here to make sure everybody is indeed on the same page now.

@yadij, please check that you are OK with the current wording and adjust further if needed. I will not clear this PR for Jenkins tests to give you a chance to check that text before the PR is auto-merged. Please tell Jenkins it can test this PR when you are done.

yadij:
OK to test

rousskov (approved):

PR #865, opened 2021-07-21
Extend MasterXaction lifetime across ::Server existence

yadij:
Give ::Server child classes access to MasterXaction object
for storage of per-protocol transaction details at any time.

### DIFF
--- a/src/servers/Server.h
+++ b/src/servers/Server.h
@@ -94,6 +94,9 @@ public:
     /// grows the available read buffer space (if possible)
     void maybeMakeSpaceAvailable();
 
+    /// transaction details about the client connection
+    MasterXactionPointer xaction;
+
     // Client TCP connection details from comm layer.
     Comm::ConnectionPointer clientConnection;
 
--- a/src/servers/Server.h
+++ b/src/servers/Server.h
@@ -107,9 +107,6 @@ public:
      */
     AnyP::ProtocolVersion transferProtocol;
 
-    /// Squid listening port details where this connection arrived.
-    AnyP::PortCfgPointer port;
-
     /// read I/O buffer for the client connection
     SBuf inBuf;
 
--- a/src/HttpRequest.cc
+++ b/src/HttpRequest.cc
@@ -722,7 +722,7 @@ HttpRequest::manager(const CbcPointer<ConnStateData> &aMgr, const AccessLogEntry
     if (!clientConnectionManager.valid())
         return;
 
-    AnyP::PortCfgPointer port = clientConnectionManager->port;
+    auto &port = clientConnectionManager->xaction->squidPort;
     if (port) {
         myportname = port->name;
         flags.ignoreCc = port->ignore_cc;
--- a/src/acl/FilledChecklist.cc
+++ b/src/acl/FilledChecklist.cc
@@ -88,7 +88,7 @@ ACLFilledChecklist::verifyAle() const
 
     if (!al->cache.port && conn()) {
         showDebugWarning("listening port");
-        al->cache.port = conn()->port;
+        al->cache.port = conn()->xaction->squidPort;
     }
 
     if (request) {
--- a/src/acl/MyPortName.cc
+++ b/src/acl/MyPortName.cc
@@ -18,8 +18,8 @@
 int
 ACLMyPortNameStrategy::match(ACLData<MatchType> * &data, ACLFilledChecklist *checklist)
 {
-    if (checklist->conn() != NULL && checklist->conn()->port != NULL)
-        return data->match(checklist->conn()->port->name);
+    if (checklist->conn() && checklist->conn()->xaction->squidPort)
+        return data->match(checklist->conn()->xaction->squidPort->name);
     if (checklist->request != NULL)
         return data->match(checklist->request->myportname.termedBuf());
     return 0;
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1134,8 +1134,8 @@ findTrailingHTTPVersion(const char *uriAndHTTPVersion, const char *end)
 static char *
 prepareAcceleratedURL(ConnStateData * conn, const Http1::RequestParserPointer &hp)
 {
-    int vhost = conn->port->vhost;
-    int vport = conn->port->vport;
+    int vhost = conn->xaction->squidPort->vhost;
+    int vport = conn->xaction->squidPort->vport;
     static char ipbuf[MAX_IPSTRLEN];
 
     /* BUG: Squid cannot deal with '*' URLs (RFC2616 5.1.2) */
@@ -1151,7 +1151,7 @@ prepareAcceleratedURL(ConnStateData * conn, const Http1::RequestParserPointer &h
         if (tok.skip('/')) // origin-form URL already.
             break;
 
-        if (conn->port->vhost)
+        if (vhost)
             return nullptr; /* already in good shape */
 
         // skip the URI scheme
@@ -1209,18 +1209,18 @@ prepareAcceleratedURL(ConnStateData * conn, const Http1::RequestParserPointer &h
         snprintf(uri, url_sz, SQUIDSBUFPH "://" SQUIDSBUFPH SQUIDSBUFPH, SQUIDSBUFPRINT(scheme), SQUIDSBUFPRINT(host), SQUIDSBUFPRINT(url));
         debugs(33, 5, "ACCEL VHOST REWRITE: " << uri);
         return uri;
-    } else if (conn->port->defaultsite /* && !vhost */) {
-        debugs(33, 5, "ACCEL DEFAULTSITE REWRITE: defaultsite=" << conn->port->defaultsite << " + vport=" << vport);
+    } else if (const auto *defaultHostname = conn->xaction->squidPort->defaultsite) {
+        debugs(33, 5, "ACCEL DEFAULTSITE REWRITE: defaultsite=" << defaultHostname << " + vport=" << vport);
         char vportStr[32];
         vportStr[0] = '\0';
         if (vport > 0) {
             snprintf(vportStr, sizeof(vportStr),":%d",vport);
         }
         const SBuf &scheme = AnyP::UriScheme(conn->transferProtocol.protocol).image();
-        const int url_sz = scheme.length() + strlen(conn->port->defaultsite) + sizeof(vportStr) + url.length() + 32;
+        const int url_sz = scheme.length() + strlen(defaultHostname) + sizeof(vportStr) + url.length() + 32;
         char *uri = static_cast<char *>(xcalloc(url_sz, 1));
         snprintf(uri, url_sz, SQUIDSBUFPH "://%s%s" SQUIDSBUFPH,
-                 SQUIDSBUFPRINT(scheme), conn->port->defaultsite, vportStr, SQUIDSBUFPRINT(url));
+                 SQUIDSBUFPRINT(scheme), defaultHostname, vportStr, SQUIDSBUFPRINT(url));
         debugs(33, 5, "ACCEL DEFAULTSITE REWRITE: " << uri);
         return uri;
     } else if (vport > 0 /* && (!vhost || no Host:) */) {
@@ -1356,8 +1356,8 @@ ConnStateData::parseHttpRequest(const Http1::RequestParserPointer &hp)
            "\n----------");
 
     /* deny CONNECT via accelerated ports */
-    if (hp->method() == Http::METHOD_CONNECT && port != NULL && port->flags.accelSurrogate) {
-        debugs(33, DBG_IMPORTANT, "WARNING: CONNECT method received on " << transferProtocol << " Accelerator port " << port->s.port());
+    if (hp->method() == Http::METHOD_CONNECT && xaction->squidPort && xaction->squidPort->flags.accelSurrogate) {
+        debugs(33, DBG_IMPORTANT, "WARNING: CONNECT method received on " << transferProtocol << " Accelerator port " << xaction->squidPort->s.port());
         debugs(33, DBG_IMPORTANT, "WARNING: for request: " << hp->method() << " " << hp->requestUri() << " " << hp->messageProtocol());
         hp->parseStatusCode = Http::scMethodNotAllowed;
         return abortRequestParsing("error:method-not-allowed");
@@ -1368,7 +1368,7 @@ ConnStateData::parseHttpRequest(const Http1::RequestParserPointer &hp)
      * If seen it signals a broken client or proxy has corrupted the traffic.
      */
     if (hp->method() == Http::METHOD_PRI && hp->messageProtocol() < Http::ProtocolVersion(2,0)) {
-        debugs(33, DBG_IMPORTANT, "WARNING: PRI method received on " << transferProtocol << " port " << port->s.port());
+        debugs(33, DBG_IMPORTANT, "WARNING: PRI method received on " << transferProtocol << " port " << xaction->squidPort->s.port());
         debugs(33, DBG_IMPORTANT, "WARNING: for request: " << hp->method() << " " << hp->requestUri() << " " << hp->messageProtocol());
         hp->parseStatusCode = Http::scMethodNotAllowed;
         return abortRequestParsing("error:method-not-allowed");
@@ -1405,7 +1405,7 @@ ConnStateData::parseHttpRequest(const Http1::RequestParserPointer &hp)
 
     /* set url */
     debugs(33,5, "Prepare absolute URL from " <<
-           (transparent()?"intercept":(port->flags.accelSurrogate ? "accel":"")));
+           (transparent()?"intercept":(xaction->squidPort->flags.accelSurrogate ? "accel":"")));
     /* Rewrite the URL in transparent or accelerator mode */
     /* NP: there are several cases to traverse here:
      *  - standard mode (forward proxy)
@@ -1433,7 +1433,7 @@ ConnStateData::parseHttpRequest(const Http1::RequestParserPointer &hp)
         //  But have not parsed there yet!! flag for local-only handling.
         http->flags.internal = true;
 
-    } else if (port->flags.accelSurrogate) {
+    } else if (xaction->squidPort->flags.accelSurrogate) {
         /* accelerator mode */
         http->uri = prepareAcceleratedURL(this, hp);
         http->flags.accel = true;
@@ -1661,9 +1661,9 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
     request->flags.sslBumped=conn->switchedToHttps();
     // TODO: decouple http->flags.accel from request->flags.sslBumped
     request->flags.noDirect = (request->flags.accelerated && !request->flags.sslBumped) ?
-                              !conn->port->allow_direct : 0;
+                              !conn->xaction->squidPort->allow_direct : 0;
     request->sources |= isFtp ? Http::Message::srcFtp :
-                        ((request->flags.sslBumped || conn->port->transport.protocol == AnyP::PROTO_HTTPS) ? Http::Message::srcHttps : Http::Message::srcHttp);
+                        ((request->flags.sslBumped || conn->xaction->squidPort->transport.protocol == AnyP::PROTO_HTTPS) ? Http::Message::srcHttps : Http::Message::srcHttp);
 #if USE_AUTH
     if (request->flags.sslBumped) {
         if (conn->getAuth() != NULL)
@@ -2228,8 +2228,8 @@ ConnStateData::start()
     BodyProducer::start();
     HttpControlMsgSink::start();
 
-    if (port->disable_pmtu_discovery != DISABLE_PMTU_OFF &&
-            (transparent() || port->disable_pmtu_discovery == DISABLE_PMTU_ALWAYS)) {
+    if (xaction->squidPort->disable_pmtu_discovery != DISABLE_PMTU_OFF &&
+            (transparent() || xaction->squidPort->disable_pmtu_discovery == DISABLE_PMTU_ALWAYS)) {
 #if defined(IP_MTU_DISCOVER) && defined(IP_PMTUDISC_DONT)
         int i = IP_PMTUDISC_DONT;
         if (setsockopt(clientConnection->fd, SOL_IP, IP_MTU_DISCOVER, &i, sizeof(i)) < 0) {
@@ -2250,7 +2250,7 @@ ConnStateData::start()
     AsyncCall::Pointer call = JobCallback(33, 5, Dialer, this, ConnStateData::connStateClosed);
     comm_add_close_handler(clientConnection->fd, call);
 
-    needProxyProtocolHeader_ = port->flags.proxySurrogate;
+    needProxyProtocolHeader_ = xaction->squidPort->flags.proxySurrogate;
     if (needProxyProtocolHeader_) {
         if (!proxyProtocolValidateClient()) // will close the connection on failure
             return;
@@ -2370,7 +2370,7 @@ static bool
 httpsCreate(const ConnStateData *connState, const Security::ContextPointer &ctx)
 {
     const auto conn = connState->clientConnection;
-    if (Security::CreateServerSession(ctx, conn, connState->port->secure, "client https start")) {
+    if (Security::CreateServerSession(ctx, conn, connState->xaction->squidPort->secure, "client https start")) {
         debugs(33, 5, "will negotiate TLS on " << conn);
         return true;
     }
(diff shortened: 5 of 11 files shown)


src/servers/Server.h
```diff
@@ -94,6 +94,9 @@ class Server : virtual public AsyncJob, public BodyProducer
     /// grows the available read buffer space (if possible)
     void maybeMakeSpaceAvailable();
 
+    /// transaction details about the client connection
+    MasterXactionPointer xaction;
```
rousskov, line 98: There can be no "transaction details about the client connection" concept because multiple _concurrent_ master transactions can use the same from-client connection. In HTTP/1, this kind of transaction concurrency requires pipelining. It becomes standard in HTTP/2.

If you disclose a specific problem you are trying to solve, I may be able to suggest a solution that does not involve adding transaction-specific details to a connection manager that may correspond to multiple concurrent transactions.

rousskov (changes requested):

src/servers/Server.h
yadij: > There can be no "transaction details about the client connection" concept because multiple _concurrent_ master transactions can use the same from-client connection. In HTTP/1, this kind of transaction concurrency requires pipelining. It becomes standard in HTTP/2.

Recall that in PR #716 I proposed a model where the major *transfer* protocols protocol-specific transactions were spawning a MasterXaction object to represent their activity (ie. each pipelined HTTP request). You objected (https://github.com/squid-cache/squid/pull/716#discussion_r474948615) on grounds that the idea of sub-master transaction conflicts with the "master transaction" concept.

With the expansion to support non-HTTP transactions initiated directly by clients (PROXY, FTP, TLS splice, HTTP-over-HTTP, unsupported_protocol / WebSockets native / DoH) we should be moving to a model where [TCP, UDP, Squid internal] is the initiator of a "master transaction" and the protocol-specific transactions are some form of child/sub transaction to service that initiator. The MasterXaction class object is more akin to a "browsing session" representative in HTTP terms.

Under that model pipelined "HTTP transaction" are just a few of the secondary transactions that form part of an overall "master transaction" for Squid.  The client is initiating the "master" transaction (in TcpAcceptor). The per-protocol transaction/requests are not "master's", they are transactions required to complete to collect information about the overall "master".

> 
> If you disclose a specific problem you are trying to solve, I may be able to suggest a solution that does not involve adding transaction-specific details to a connection manager that may correspond to multiple concurrent transactions.

The "most specific problem" is that code following the ::Server hierarchy constructors do not have access to the MasterXaction class and data stored in its members - right now they take a copy of every individual member and we *hope* that the copied members are enough for the child objects spawned later by response handling logic. That does not scale (eg when Client-IP can come from 4 different data sources across 3 different protocols objects).

This MasterXaction class is the place where the various protocols attach Pointer to their data, provides an API to access the details from other protocol-specific transactions consistently (eg client IP disambiguation regardless of which protocols provided it)  and (ideally) gets passed around from initiator until logging of the final sub-transaction. None of that purpose can be implemented when the object is dropped (and destructed) after being passed to its first callback.

rousskov: > Recall that in PR #716 I proposed a model where the major _transfer_ protocols protocol-specific transactions were spawning a MasterXaction object to represent their activity (ie. each pipelined HTTP request). You objected ([#716 (comment)](https://github.com/squid-cache/squid/pull/716#discussion_r474948615)) on grounds that the idea of sub-master transaction conflicts with the "master transaction" concept.

The above description of what has happened in that PR does not match my understanding of reality: There is nothing wrong with subtransactions per se. I have [objected](https://github.com/squid-cache/squid/pull/716#discussion_r474948615) to subtransactions of a master transaction being master transactions themselves (i.e. having the type of MasterXaction). That PR was spawning master transaction "layers" and using the same master transaction type for those master transaction layers. That is wrong, as discussed in that PR.

> With the expansion to support non-HTTP transactions initiated directly by clients (PROXY, FTP, TLS splice, HTTP-over-HTTP, unsupported_protocol / WebSockets native / DoH) we should be moving to a model where [TCP, UDP, Squid internal] is the initiator of a "master transaction" and the protocol-specific transactions are some form of child/sub transaction to service that initiator. The MasterXaction class object is more akin to a "browsing session" representative in HTTP terms.

I disagree. A browsing session usually involves multiple MasterXaction objects. For example, roughly speaking, each browser (HTTP) request inside a session should create a MasterXaction object. Similarly, each FTP command received by Squid inside an FTP session should create a MasterXaction object. Still roughly speaking, each access.log record corresponds to a unique master transaction. One browsing session may correspond to numerous access.log records.


> Under that model pipelined "HTTP transaction" are just a few of the secondary transactions that form part of an overall "master transaction" for Squid. The client is initiating the "master" transaction (in TcpAcceptor). The per-protocol transaction/requests are not "master's", they are transactions required to complete to collect information about the overall "master".

I do not think we should define MasterXaction to represent an accepted TCP connection or equivalent. Connections, sessions, and such are ways of delivering individual transactions to Squid. They have their own properties, needs, and actions, quite different from those of a master transaction (i.e. a single received request and associated request-handling activities).


> The "most specific problem" is that code following the ::Server hierarchy constructors do not have access to the MasterXaction class and data stored in its members - right now they take a copy of every individual member and we _hope_ that the copied members are enough for the child objects spawned later by response handling logic. That does not scale (eg when Client-IP can come from 4 different data sources across 3 different protocols objects).

Yes, connection-handler (i.e. ConnStateData and friends) construction code needs refactoring. It has been a persistent source of pain for a long time. For various (mostly historical) reasons, the code poorly reflects the boundary between two very different concepts: the very first master transaction on the accepted connection and that connection metadata itself. Both objects are created (or can be deemed to be created) at the same accept(2) time, but the accepted connection metadata lifetime covers all master transactions on that accepted connection, not just the very first one.

To solve that problem, we should probably:

1. Define AcceptedConnection, a new class that represents accepted connection (or equivalent) metadata. The exact class name is up to discussion, of course, I had to call it something for ease of reference.

2. Adjust MasterXaction to gain a (refcounted) pointer to AcceptedConnection (possibly replacing or moving some existing MasterXaction data members).

3. ConnStateData and similar constructors will still get a MasterXaction object, but instead of copying various details to preserve them across master transactions, they should just copy an AcceptedConnection pointer.
 
The first MasterXaction object triggers the usual transaction processing, gets access-logged, and then disappears. The AcceptedConnection object survives until the corresponding accepted connection is gone (at least).

It could be tempting to avoid creating MasterXaction objects at accept time (and just create AcceptedConnection objects then), but I doubt that is a good idea because we have to access-log accepted connections and we need a MasterXaction for that anyway -- that is what we are logging, essentially. It is probably best to create master transaction objects in as few places as possible to avoid code duplication and to simplify "have we logged that already?" checks/logic.

yadij: > > Recall that in PR #716 I proposed a model where the major _transfer_ protocols protocol-specific transactions were spawning a MasterXaction object to represent their activity (ie. each pipelined HTTP request). You objected ([#716 (comment)](https://github.com/squid-cache/squid/pull/716#discussion_r474948615)) on grounds that the idea of sub-master transaction conflicts with the "master transaction" concept.
> 
> The above description of what has happened in that PR does not match my understanding of reality: There is nothing wrong with subtransactions per se. I have [objected](https://github.com/squid-cache/squid/pull/716#discussion_r474948615) to subtransactions of a master transaction being master transactions themselves (i.e. having the type of MasterXaction). That PR was spawning master transaction "layers" and using the same master transaction type for those master transaction layers. That "using..." part is wrong, as discussed in that PR.
> 

*IF* we assume that each HTTP transaction on an FD is a "master transaction". Then it follows that we treat each other layer of non-HTTP protocol transaction as also a "master transaction" - eg HTTP-over-TLS-over-HTTP (aka HTTPS in CONNECT tunnel). That is what PR 716 implemented as class structure of interlinked "master transaction" all sharing an FD.

The comment I linked to is your statement that "master transaction" (TCP transaction) cannot have sub-master transaction (HTTP transaction) layered inside it. Which makes it forbidden to represent HTTP-over-TCP/80 as two separate "master transaction" (one for each protocol).
 That logical state also forbids *conceptually* calling each HTTP transaction a "master transaction". Derived from the fact that there is always some type of transport protocol initiating the "master transaction" - for which it cannot be a sub-master.



> > With the expansion to support non-HTTP transactions initiated directly by clients (PROXY, FTP, TLS splice, HTTP-over-HTTP, unsupported_protocol / WebSockets native / DoH) we should be moving to a model where [TCP, UDP, Squid internal] is the initiator of a "master transaction" and the protocol-specific transactions are some form of child/sub transaction to service that initiator. The MasterXaction class object is more akin to a "browsing session" representative in HTTP terms.
> 
> I disagree. A browsing session usually involves multiple MasterXaction objects. For example, roughly speaking, each browser (HTTP) request inside a session should create a MasterXaction object. Similarly, each FTP command received by Squid inside an FTP session should create a MasterXaction object. Still roughly speaking, each access.log record corresponds to a unique master transaction. One browsing session may correspond to numerous access.log records.

You have clearly missed my point and are warping the master-worker paradigm. Individual HTTP transactions and FTP commands are represented by HttpRequest object.

"
What is Master-Worker Paradigm
1.
Consists in two entities: a master and several workers. The master decomposes the problem into smaller tasks and distributes them among workers.
"

MasterXaction is the metadata of the master. Each HttpRequest is the metadata of one "worker" in current code.


> 
> > Under that model pipelined "HTTP transaction" are just a few of the secondary transactions that form part of an overall "master transaction" for Squid. The client is initiating the "master" transaction (in TcpAcceptor). The per-protocol transaction/requests are not "master's", they are transactions required to complete to collect information about the overall "master".
> 
> I do not think we should define MasterXaction to represent an accepted TCP connection or equivalent. Connections, sessions, and such are ways of delivering individual transactions to Squid. They have their own properties, needs, and actions, quite different from those of a master transaction (i.e. a single received request and associated request-handling activities).
> 
> > The "most specific problem" is that code following the ::Server hierarchy constructors do not have access to the MasterXaction class and data stored in its members - right now they take a copy of every individual member and we _hope_ that the copied members are enough for the child objects spawned later by response handling logic. That does not scale (eg when Client-IP can come from 4 different data sources across 3 different protocols objects).
> 
> Yes, connection-handler (i.e. ConnStateData and friends) construction code needs refactoring. It has been a persistent source of pain for a long time. For various (mostly historical) reasons, the code poorly reflects the boundary between two very different concepts: the very first master transaction on the accepted connection and that connection metadata itself.

This thread is not about refactoring that code. It is about what we document the xaction member to be.

Regardless of the class name (we already agreed "MasterXaction" needs to change, but that is PR 823 scope, not here) it represents the overall metadata associated with the client connection / FD (so far known across *all* transactions that used it).

> To solve that problem, we should probably:
> 
>     1. Define AcceptedConnection, a new class that represents accepted connection (or equivalent) metadata. The exact class name is up to discussion, of course, I had to call it something for ease of reference.
> 

The v6 code currently calls this type Comm::Connection.
Re-inventing and/or modifying that is out of scope for this PR.

>     2. Adjust MasterXaction to gain a (refcounted) pointer to AcceptedConnection (possibly replacing or moving some existing MasterXaction data members).
> 

The v6 code currently calls this refcounted pointer MasterXaction::tcpClient.
Since the member already exists with self-documenting name, I leave touching it out of scope.

>     3. ConnStateData and similar constructors will still get a MasterXaction object, but instead of copying various details to preserve them across master transactions, they should just copy an AcceptedConnection pointer.
> 

Notice what this PR is doing in commit 892f9a1. Exactly what you describe.


> 
> The first MasterXaction object triggers the usual transaction processing, gets access-logged, and then disappears. The AcceptedConnection object survives until the corresponding accepted connection is gone (at least).
> 
> It could be tempting to avoid creating MasterXaction objects at accept time (and just create AcceptedConnection objects then), but I doubt that is a good idea because we have to access-log accepted connections and we need a MasterXaction for that anyway -- that is what we are logging, essentially. It is probably best to create master transaction objects in as few places as possible to avoid code duplication and to simplify "have we logged that already?" checks/logic.


Notice that what you are calling "AcceptConnection" the v6 code currently calls MasterXaction.

Notice that what you are calling "MasterXaction" the v6 code currently calls HttpRequest, HttpReply, Ftp::Master, IdentStateData, FwdState, Icap::Xaction, etc.

To me the paradigm of "master/worker" is clear - and not at all matching the concept of per-protocol transactions being "master".

rousskov: > _IF_ we assume that each HTTP transaction on an FD is a "master transaction". 

That assumption is incorrect. For example, when Squid forwards a received request to the server(s), the corresponding Squid-server transaction(s) are not master transactions. They are parts of (in an "associated with" or "tied to" sense) the master transaction created when Squid received the request from the client. In fact, even that original client-Squid HTTP transaction itself is just a part of that master transaction.

> The comment I linked to is your statement that "master transaction" (TCP transaction) cannot have sub-master transaction (HTTP transaction) layered inside it.

That is not what I actually [said](https://github.com/squid-cache/squid/pull/716#discussion_r474948615). I did not say the words in parenthesis, and those words completely change the meaning of what I did say. The not-mine statement above is not what I _meant_. Again, there are and always will be transactions that are associated with (or, if you wish, even "belong to") the same master transaction. The problem I was talking about is not the existence of those transactions, but the C++ _type_ #716 was using to represent them: spawnChildLayer() cannot return a MasterXaction. No MasterXactions inside a MasterXaction. A MasterXaction "layer" (a poorly chosen term, but I am just using what that PR was using) cannot be a MasterXaction object.

FWIW, I also find the term "sub-" rather misleading in this context and do not recommend using here; saying "sub-master" not only sounds weird but does not accurately relay the relationship between an HTTP transaction and the master transaction that HTTP transaction is associated with.


> A browsing session usually involves multiple MasterXaction objects. For example, roughly speaking, each browser (HTTP) request inside a session should create a MasterXaction object. Similarly, each FTP command received by Squid inside an FTP session should create a MasterXaction object. Still roughly speaking, each access.log record corresponds to a unique master transaction. One browsing session may correspond to numerous access.log records.

> You ... are warping the master-worker paradigm.

Actually, I am not using (or talking about) that unrelated (as detailed further below) paradigm at all! 

> Individual HTTP transactions and FTP commands are represented by HttpRequest object.

Yes, they are (among other things they are represented by). That fact does not prevent them from being associated with a MasterXaction object.


> What is Master-Worker Paradigm

That paradigm is not closely related to the master transaction concept we are discussing. The word "master" has lots of of different [meanings](https://www.dictionary.com/browse/master). The meaning of "master" in Squid's MasterXaction is that of "chief or principal" rather than "dominant" or "controlling". Its true meaning is much closer to the meaning of the word "master" in phrases like [master service agreement](https://en.wikipedia.org/wiki/Master_service_agreement), [master key](https://en.wikipedia.org/wiki/Master_keying), and, to a certain extent, [master recording](https://en.wikipedia.org/wiki/Mastering_(audio)).

> This thread is ... about what we document the xaction member to be.

I would not phrase it like that. IMO, classes managing an accepted connection (e.g., ConnStateData) should have no such member at all. This change request is not focused on the xaction member _documentation_. This change requests asks not to add that member itself.
 
> Regardless of the class name ... it represents the overall metadata associated with the client connection / FD (so far known across _all_ transactions that used it).

I agree that the spelling of the MasterXaction type is irrelevant to the core problem behind this change request.

I disagree that MasterXaction type represents "the overall metadata associated with the client connection / FD". The type representing _that_ information is currently missing. That is why I suggested to add that type to solve the problem you are trying to solve.

BTW, it is easy to prove that MasterXaction type does not represent "the overall metadata associated with the client connection / FD". Squid has master transactions that have no client connections or clients. Those master transactions are or should be represented by MasterXaction but they have no associated ConnStateData, client Comm::Connection, etc. For example, transactions fetching Cache Digests from cache peers and storing them in a local cache are like that.


> > To solve that problem, we should probably:
> >
> > 1. Define AcceptedConnection, a new class that represents accepted connection (or equivalent) metadata. The exact class name is up to discussion, of course, I had to call it something for ease of reference.
 
> The v6 code currently calls this type Comm::Connection.

Not quite. The (very problematic) Comm::Connection type represents virtually any connection, with a focus on an open connection and associated I/O. We use Comm::Connection for both accepted connections and connections established by Squid. The proposed AcceptedConnection class represents accepted connection metainformation. The two concepts are quite different! For example, I bet that AcceptedConnection should not have a socket descriptor (FD) member.

> > 3. ConnStateData and similar constructors will still get a MasterXaction object, but instead of copying various details to preserve them across master transactions, they should just copy an AcceptedConnection pointer.

> Notice what this PR is doing in commit [892f9a1](https://github.com/squid-cache/squid/commit/892f9a1f84d020307de66160297482de827bffa7). Exactly what you describe.

That commit copies a MasterXaction pointer. I am describing copying an AcceptedConnection pointer. There is a big difference because the former does not belong to ConnStateData (for the already disclosed reasons) while the latter naturally does.


> This PR scope is limited to implementing (3),

One cannot implement (3) correctly without implementing something like (1).

> Notice that what you are calling "AcceptConnection" the v6 code currently calls MasterXaction.

I disagree that AcceptedConnection is MasterXaction. They are completely different classes and concepts. I do not know why you think they are the same.
 
> Notice that what you are calling "MasterXaction" the v6 code currently calls HttpRequest, HttpReply, Ftp::Master, IdentStateData, FwdState, Icap::Xaction, etc.

When I say MasterXaction, I mean the C++ type MasterXaction in Squid source code (and the associated high-level concept). HttpRequest, etc. are other/different types (and concepts).

> To me the paradigm of "master/worker" is clear - and not at all matching the concept of per-protocol transactions being "master".

I agree with these two parts: The master/worker paradigm is clear. Individual HTTP/etc. transactions are not master transactions.

As detailed above, the master/worker paradigm is unrelated to MasterXaction and this change request.

yadij: Context: https://github.com/squid-cache/squid/pull/865#discussion_r674066436

Amos:  IF we assume that each HTTP transaction on an FD is a "master transaction".

Alex:  That assumption is incorrect. For example, when Squid forwards a received request to the server(s), the corresponding Squid-server transaction(s) are not master transactions. They are parts of (in an "associated with" or "tied to" sense) the master transaction created when Squid received the request from the client. In fact, even that original client-Squid HTTP transaction itself is just a part of that master transaction.


Notice how "that original client-Squid HTTP transaction itself is just a part of that master transaction." directly contradicts the description of HTTP/2 and HTTP/1 pipeline's as "multiple concurrent master transactions".

Notice how "forwards a received request to the server(s)" is at times just performing BlindConnector. Eliding TCP connections as part of the "master transaction" the details of those server connections are left without ability to log according to admin policy and similar weird issues.

IMO, either we need to start a "master" at the point a client contacts Squid and treat all protocols in the same "parts of master". Or, we should drop/detatch the concept of "master" entirely from the code syntax and stick with the old style of calling transactions by the feature/action/protocol which initiated it.

rousskov: > Alex: When Squid forwards a received request to the server(s), the corresponding Squid-server transaction(s) are not master transactions. They are parts of (in an "associated with" or "tied to" sense) the master transaction created when Squid received the request from the client. In fact, even that original client-Squid HTTP transaction itself is just a part of that master transaction.

> Amos: Notice how "that original client-Squid HTTP transaction itself is just a part of that master transaction." directly contradicts the description of HTTP/2 and HTTP/1 pipeline's as "multiple concurrent master transactions".

I see no contradiction. In pipelining and HTTP/2 cases, there are concurrent master transactions. Each "original client-Squid HTTP transaction" in those cases is a part of the corresponding master transaction: N master transactions, and N "original client-Squid HTTP transactions", with 1:1 mapping between those two specific sets. Where do you see a contradiction?


> Amos: Notice how "forwards a received request to the server(s)" is at times just performing BlindConnector. Eliding TCP connections as part of the "master transaction" the details of those server connections are left without ability to log according to admin policy and similar weird issues.

Perhaps I am misparsing the above paragraph, but I do not think anybody is trying to elide TCP connection details from master transactions. Naturally, most master transactions will have a non-nil pointer to transport connection details. The above paragraph sounds like an argument against removal of those details from master transactions, but I do not think anybody is advocating _for_ that removal.

 
> IMO, either we need to start a "master" at the point a client contacts Squid

Yes, we should do that (or equivalent). In fact, we already do something like that -- TcpAcceptor::notify() creates a master transaction. Unfortunately, the current code mishandles that master transaction object later on, but those bugs can, should, and will be fixed.

If there is a disagreement (rather than just a misunderstanding), then it might be about when that first master transaction _ends_ or when the next/second master transaction (associated with the same accepted TCP connection) begins (and those two events may happen at different times). In a simple case like a basic HTTP/TCP connection with non-pipelined non-concurrent HTTP/1 transactions, that second master transaction starts with the second HTTP request on that TCP connection. This is not an exact definition because I am worried about us wasting time on unimportant secondary details, but I hope this definition is enough to illustrate the primary point.

>  and treat all protocols in the same "parts of master". 

This statement is too vague for me to agree or disagree. I hope the above two paragraphs contain enough information to clarify.

> Or, we should drop/detatch the concept of "master" entirely from the code syntax and stick with the old style of calling transactions by the feature/action/protocol which initiated it.

I do not know what you mean by "a concept of master transaction associated with the code syntax". The concepts we are discussing have nothing to do with C++ syntax IMO. No mater what we do, the concept of a master transaction is not going to disappear -- it exists in the observable world. IMO, we should (continue to) support that concept in our code. The same can be said about an HTTP transaction or a TCP connection -- all real concepts we should (continue to) support.

yadij: > > Alex: When Squid forwards a received request to the server(s), the corresponding Squid-server transaction(s) are not master transactions. They are parts of (in an "associated with" or "tied to" sense) the master transaction created when Squid received the request from the client. In fact, even that original client-Squid HTTP transaction itself is just a part of that master transaction.
> 
> > Amos: Notice how "that original client-Squid HTTP transaction itself is just a part of that master transaction." directly contradicts the description of HTTP/2 and HTTP/1 pipeline's as "multiple concurrent master transactions".
> 
> I see no contradiction. In pipelining and HTTP/2 cases, there are concurrent master transactions. Each "original client-Squid HTTP transaction" in those cases is a part of the corresponding master transaction: N master transactions and N "original client-Squid HTTP transactions", with 1:1 mapping between those two specific sets. Where do you see a contradiction?
> 

When the pipeline is nested a) inside a CONNECT, or b) a TLS layer inside a CONNECT, or c) inside a PROXY. We end up with nested "master". Which you defined as forbidden.

Moving to a model where a "master" is initiated by the action of client connecting (TCP SYN, UDP read) and everything that occurs within is some non-master type of transaction resolves this easily.

rousskov: > When the pipeline is nested a) inside a CONNECT, or b) a TLS layer inside a CONNECT, or c) inside a PROXY. We end up with nested "master". 

We do not:

1. The master transaction associated with a bumped CONNECT request ends with a CONNECT response; the pipeline and its master transactions start later. And there is no pipeline inside a master transaction associated with a non-bumped CONNECT request, of course.
2. The TLS layer, just like the TCP layer, is not a part of a master transaction. It is a connection layer (which often spans multiple transactions).
3. There is no pipeline _inside_ the PROXY protocol. The PROXY protocol transaction ends with the PROXY protocol message.

> Moving to a model where a "master" is initiated by the action of client connecting (TCP SYN, UDP read) 

That part is true for the master transaction model that Squid already implements (bugs notwithstanding), and that I continue to support.

> and everything that occurs within is some non-master type of transaction

This last part does not make sense to me because it erases the notion of a master transaction that we clearly need (and use) today. This last part describes a connection (or equivalent) that is carrying transactions, not an individual transaction itself. I hope that the fact that a _connection_ (of any kind) carrying transactions (of any kind) and a _transaction_ (of any kind) are distinct concepts is self-evident.

Note that the only problem with your sketch is that the master transaction does not end until the connection (or equivalent) ends -- it encompasses "everything that occurs within". If you fix that by giving that master transaction a lifespan that may be shorter than that of a connection, then you will arrive at the correct model (at least as far as this high-level discussion is concerned).

Imagine a simple case: A regular HTTP connection with a GET and a HEAD request. Nothing special is going on except GET and HEAD get piped through two ICAP services each (e.g., REQMOD and RESPMOD).

* This case involves exactly two master transactions. There are two access.log records, each with a distinct master transaction ID. There are four icap.log records. Among those four icap.log records, there will be two unique master transaction IDs, tying each logged ICAP transaction to one of the two master transactions in access.log.

* If we switch to a model you describe, then this case will involve just one master transaction, which will equate that single master transaction with the connection and, hence, make the concept of a master transaction essentially unused/useless. For example, logging master transaction ID to access.log will provide the same information as logging connection ID. And logging master transaction ID to icap.log will not allow the admin to associate a logged ICAP transaction with the corresponding HTTP request -- all four ICAP transactions in this example will have the same master transaction ID.

PR #828, opened 2021-05-22
Implement RAII wrapper for ctx_* functions

kinkie:
Current uses of ctx_enter and ctx_exit are scope-based.
We can enrich them by implementing a RAII wrapper class
and fully rely on scope for entering and leaving contexts safely

PR #866, opened 2021-07-21
Remove legacy context-based debugging in favor of CodeContext

rousskov:
Added in 1998, context-based debugging has been neglected and

- only covers two relatively small contexts
- unsafe in the presence of exceptions
- produces noise (e.g., "ctx: exit level 0" messages)
- delayed "ctx: exit" messages confuse admins
- uses deprecated urlXXX() API
- difficult to extend to more contexts without performance overheads
- usually provides less info in fewer contexts (than CodeContext)

### DIFF
--- a/src/base/CodeContext.h
+++ b/src/base/CodeContext.h
@@ -28,16 +28,12 @@
  * busy Squid instance constantly switching from one processing context to
  * another.
  *
- * To solve these problems, Squid maintains a CodeContext object tied to
- * (eventually) each processing context. CodeContext objects can report context
- * details that can be tied to access.log records. When Squid switches to
- * another processing context, it switches the current CodeContext object as
- * well. When Squid prints a level-0 or level-1 message to cache.log, it asks
- * the current CodeContext object (if any) to report context details, allowing
- * the admin to correlate the error message with an access.log record. In most
- * cases, the processing context is a master transaction that Squid is working
- * on, which is reported using a master transaction ID and can be recorded in
- * access.log using a %master_xaction logformat code.
+ * To solve these problems, Squid assigns a CodeContext object to a processing
+ * context. When Squid switches to another processing context, it switches the
+ * current CodeContext object as well. When Squid prints a level-0 or level-1
+ * message to cache.log, it asks the current CodeContext object (if any) to
+ * report context details, allowing the admin to correlate the cache.log message
+ * with an access.log record.
  *
  * Squid also reports processing context changes to cache.log when Squid
  * level-5+ debugging is enabled.
--- a/src/base/CodeContext.h
+++ b/src/base/CodeContext.h
@@ -14,6 +14,41 @@
 
 #include <iosfwd>
 
+/** \file
+ *
+ * Most error-reporting code cannot know what transaction or task Squid was
+ * working on when the error occurred. For example, when Squid HTTP request
+ * parser discovers a malformed header field, the parser can report the field
+ * contents, but that information is often useless for the admin without
+ * processing context details like which client sent the request or what the
+ * requested URL was. Moreover, even when the error reporting code does have
+ * access to some context details, it cannot separate important facts from noise
+ * because such classification is usually deployment-specific (i.e. cannot be
+ * hard-coded) and requires human expertise. The situation is aggravated by a
+ * busy Squid instance constantly switching from one processing context to
+ * another.
+ *
+ * To solve these problems, Squid maintains a CodeContext object tied to
+ * (eventually) each processing context. CodeContext objects can report context
+ * details that can be tied to access.log records. When Squid switches to
+ * another processing context, it switches the current CodeContext object as
+ * well. When Squid prints a level-0 or level-1 message to cache.log, it asks
+ * the current CodeContext object (if any) to report context details, allowing
+ * the admin to correlate the error message with an access.log record. In most
+ * cases, the processing context is a master transaction that Squid is working
+ * on, which is reported using a master transaction ID and can be recorded in
+ * access.log using a %master_xaction logformat code.
+ *
+ * Squid also reports processing context changes to cache.log when Squid
+ * level-5+ debugging is enabled.
+ *
+ * CodeContext is being retrofitted into existing code with lots of places that
+ * switch processing context. Identifying and adjusting all those places takes
+ * time. Until then, there will be incorrect and missing context attributions.
+ *
+ * @{
+ **/
+
 /// Interface for reporting what Squid code is working on.
 /// Such reports are usually requested outside creator's call stack.
 /// They are especially useful for attributing low-level errors to transactions.
@@ -116,6 +151,8 @@ CallContextCreator(Fun &&creator)
     creator();
     CodeContext::Reset(savedCodeContext);
 }
+
+/// @}
 
 #endif
 
--- a/src/Debug.h
+++ b/src/Debug.h
@@ -33,11 +33,6 @@
 #define assert(EX)  ((EX)?((void)0):xassert("EX", __FILE__, __LINE__))
 #endif
 
-/* context-based debugging, the actual type is subject to change */
-typedef int Ctx;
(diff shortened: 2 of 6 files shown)


rousskov:
I think it is time to remove this legacy feature, at least from master/v6. The existing CodeContext coverage is usually better and if there are any important CodeContext coverage holes where the legacy feature still works better today, then removing that legacy reporting might actually help expose those holes. As old commit 859f166 and recent PR #828 illustrate, keeping this legacy code adds costs.

yadij (approved):
Thank you for removing this.

I just note that the first paragraph of "rationale" for this features existence should probably be kept as part of the CodeContext object documentation. I do not insist on this.

rousskov:
```C++
/*
 * When you have a long nested processing sequence, it is often impossible
 * for low level routines to know in what larger context they operate. If a
 * routine coredumps, one can restore the context using debugger trace.
 * However, in many case you do not want to coredump, but just want to report
 * a potential problem. A report maybe useless out of problem context.
 * ...
 */
```

> I just note that the first paragraph of "rationale" for this features existence should probably be kept as part of the CodeContext object documentation.

Well, that paragraph (quoted above) does not actually apply well to CodeContext and to modern code. In 1998, we did have deep call stacks, and it was often much easier to find transaction details up that stack -- we just did not want to dump core. Today, async calls make call stacks shallower and shallower (that is a good thing), but their association with master transaction is getting weaker and weaker. That is why CodeContext, unlike the legacy API, is designed to save context info in one call stack and report it in another call stack.

I can offer the following adjustment of CodeContext documentation that borrows a couple of ideas from the legacy paragraph:

```C++
/// Interface for reporting what Squid code is working on.
/// Such reports are usually requested outside creator's call stack.
/// They are especially useful for attributing low-level errors to transactions
/// without asserting and forcing the developers to study the core dump for
/// transaction details that are usually stored outside the current call stack.
class CodeContext: public RefCountable
```

@yadij, do you want me to add the above (or some adjusted version of it)?

yadij:
The structure of the old text was much more clear IMO. Stating what the use-case problem was, then how the class resolved it. It just needs word changes to describe the current situation with Async* and how CodeContext solves that.

rousskov:
> The structure of the old text was much more clear IMO. Stating what the use-case problem was, then how the class resolved it. It just needs word changes to describe the current situation with Async* and how CodeContext solves that.

I agree regarding clarity of the old description structure. A much simpler problem allowed for a simpler explanation of the solution.

Right now, I cannot offer a better description for CodeContext than I already have [offered](https://github.com/squid-cache/squid/pull/866#issuecomment-884937859), but I, of course, welcome improvements. Since you were not excited about it, I will not add the offered text to this PR (unless you tell me to), but please feel free to do so if you think it is better than the current version despite its shortcomings.

yadij:
Suggested update to class CodeContext description:
```
/** Context-based Debugging
 *
 * When you have a long nested callback sequence, or long series of spread out
 * AsyncCall it can be difficult or impossible for low level cod to know in what
 * larger context it operates. If a nested routine coredumps, one can restore the
 * context using debugger trace. However, this does not help with AsyncCall
 * sequences and in many case you do not want to coredump, but just want to
 * report a potential problem. A report maybe useless out of problem context.
 * 
 * To solve these problems, debugs() messages are prepended with the current
 * CodeContext description.
 **/
```

Of course change that "prepended with the current CodeContext" if it is wrong, and follow with details on when CodeContext should be made "current" and how to reset.

src/base/CodeContext.h
```diff
@@ -14,6 +14,41 @@
...
+ * on, which is reported using a master transaction ID and can be recorded in
+ * access.log using a %master_xaction logformat code.
+ *
+ * Squid also reports processing context changes to cache.log when Squid
+ * level-5+ debugging is enabled.
+ *
+ * CodeContext is being retrofitted into existing code with lots of places that
+ * switch processing context. Identifying and adjusting all those places takes
+ * time. Until then, there will be incorrect and missing context attributions.
+ *
```
rousskov, line 48: > Suggested update to class CodeContext description: ...

I had too many reservations about that description to use it, but I hope that this text added in branch commit 880a53d is enough to move this (unrelated to CodeContext documentation!) PR forward.

src/base/CodeContext.h
```diff
@@ -14,6 +14,41 @@
...
+ * To solve these problems, Squid maintains a CodeContext object tied to
+ * (eventually) each processing context. CodeContext objects can report context
+ * details that can be tied to access.log records. When Squid switches to
+ * another processing context, it switches the current CodeContext object as
+ * well. When Squid prints a level-0 or level-1 message to cache.log, it asks
+ * the current CodeContext object (if any) to report context details, allowing
+ * the admin to correlate the error message with an access.log record. In most
+ * cases, the processing context is a master transaction that Squid is working
+ * on, which is reported using a master transaction ID and can be recorded in
+ * access.log using a %master_xaction logformat code.
```
yadij, line 40: Equating things to "master transaction" is currently causing us some discussion friction. This example should be removed until the situation is certain AND agreed on.
```suggestion
 * the admin to correlate the error message with an access.log record.
```

src/base/CodeContext.h
```diff
@@ -14,6 +14,41 @@
...
+ * processing context details like which client sent the request or what the
+ * requested URL was. Moreover, even when the error reporting code does have
+ * access to some context details, it cannot separate important facts from noise
+ * because such classification is usually deployment-specific (i.e. cannot be
+ * hard-coded) and requires human expertise. The situation is aggravated by a
+ * busy Squid instance constantly switching from one processing context to
+ * another.
+ *
+ * To solve these problems, Squid maintains a CodeContext object tied to
+ * (eventually) each processing context. CodeContext objects can report context
```
yadij, line 32: The level of verbosity of this first sentence adds more questions than it resolves.
```suggestion
 * To solve these problems, Squid maintains a CodeContext object. CodeContext objects can report context
```

src/base/CodeContext.h
```diff
@@ -14,6 +14,41 @@
...
+ * details that can be tied to access.log records. When Squid switches to
```
yadij, line 33: This sentence is redundant with later text and out of sequence in the lifecycle of a CodeContext which this paragraph appears to be describing in detail.
```suggestion
 * (eventually) each processing context. When Squid switches to
```

src/base/CodeContext.h
```diff
@@ -14,6 +14,41 @@
...
+ * another processing context, it switches the current CodeContext object as
+ * well. When Squid prints a level-0 or level-1 message to cache.log, it asks
+ * the current CodeContext object (if any) to report context details, allowing
+ * the admin to correlate the error message with an access.log record. In most
```
yadij, line 37: Implication that all level 0 or 1 are errors is false.
```suggestion
 * the admin to correlate the cache.log message with an access.log record. In most
```

src/base/CodeContext.h
yadij: This one could be less verbose, but I can accept different wording.

yadij (changes requested):

src/base/CodeContext.h
rousskov: I rephrased the first sentence in hope to address your concern while preserving essential information lost in the suggested version.

src/base/CodeContext.h
rousskov: Removed as redundant.

src/base/CodeContext.h
rousskov: Fixed.

src/base/CodeContext.h
rousskov: Removed the example to avoid arguing.

rousskov (commented):
I hope commit 5e16c20 addresses your remaining concerns.

yadij (approved):
Thank you for addressing my concerns

PR #869, opened 2021-07-28
Fix Sum() by replacing it with a safer NaturalSum()

rousskov:
While testing future cache entry expiration computation code, we found a
Sum() bug: Sum(0l, -1lu) returns -1 instead of overflowing (long cannot
hold the maximum unsigned long value). This discovery triggered an
investigation that discovered several Sum() flaws, even inside the
supposed-to-be-trivial implementation branch for unsigned operands:

0. AllUnsigned-based optimization path selection was based on the
   detection of unsigned raw S and T types, but the actual s+t sum
   used integral-promoted types that could be signed!

1. "sizeof(T) >= sizeof(U)" assertion missed that same-size types have
   very different maximums when exactly one of the types is signed.

2. "auto sum = a + b" assignment missed that "auto" may be bigger than
   the return type "T" and, hence, the sum may not overflow when it does
   not fit T. The overflow would go undetected and the result will be
   truncated (to fit T) in the return statement. This automatic sum type
   enlargement may be due to signed->unsigned integral conversion
   related to the previous bullet, but it can also happen because of
   integral _promotions_ (e.g., those that convert "char" into "int").

3. Sum() silently truncated its arguments to fit T. Before commit
   1fba9ab, that silent truncation only applied to the first argument
   (if T did not match its actual type). After that commit, all other
   arguments could be silently truncated (except the last one).

4. It is trivial for the caller to do everything right when calling
   Sum() but then assign the result to a variable that cannot hold Sum()
   value, essentially undoing all the overflow detection work.

Fortunately, none of these bugs affected two existing Sum() callers.

Some of these problems were very difficult to fix! Eventually, a new,
simplified concept emerged that was easier to implement and that was a
better match for Squid current and foreseeable needs: NaturalSum().

NaturalSum() is designed for cases where we want an exact sum of
arguments but do not consider negative arguments as valid numbers. This
both simplifies the implementation and protects typical callers from
adding "-1" (e.g., a special "no delay" option setting) to "3600" (e.g.,
configured TTL) and getting a meaningless result of 3599.

NaturalSum() requires the caller to specify the summation type that will
be used to accumulate the sum value as Squid iterates over arguments.
Sum() wanted to automatically use the largest type that can accommodate
(partial) sums, but that complicated task was not implemented and
becomes unnecessary when dealing with only natural numbers -- there is
no need to temporary inflate the partial sum (beyond what the resulting
type can hold) in case some negative operand will decrease it later.

Also added SetToNaturalSumOrMax() to allow the caller to reset a
variable without guessing its type, avoiding silent sum truncation at
assignment time.

Also added unit test cases.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -872,10 +872,11 @@ tests_testBoilerplate_LDFLAGS = $(LIBADD_DL)
 # which uses it.
 
 ## Tests of SquidMath.h
-
 check_PROGRAMS += tests/testMath
 tests_testMath_SOURCES = \
-	tests/testMath.cc
+	 tests/testMath.cc
+nodist_tests_testMath_SOURCES = \
+	 SquidMath.h
 tests_testMath_LDADD = \
 	$(LIBCPPUNIT_LIBS) \
 	$(COMPAT_LIB) \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -871,6 +871,17 @@ tests_testBoilerplate_LDFLAGS = $(LIBADD_DL)
 # dependency-based list so that simpler code is tested before more complex code
 # which uses it.
 
+## Tests of SquidMath.h
+
+check_PROGRAMS += tests/testMath
+tests_testMath_SOURCES = \
+	tests/testMath.cc
+tests_testMath_LDADD = \
+	$(LIBCPPUNIT_LIBS) \
+	$(COMPAT_LIB) \
+	$(XTRA_LIBS)
+tests_testMem_LDFLAGS = $(LIBADD_DL)
+
 ## Tests of mem/*
 
 check_PROGRAMS += tests/testMem
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -881,7 +881,7 @@ tests_testMath_LDADD = \
 	$(LIBCPPUNIT_LIBS) \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
-tests_testMem_LDFLAGS = $(LIBADD_DL)
+tests_testMath_LDFLAGS = $(LIBADD_DL)
 
 ## Tests of mem/*
 
--- a/src/SquidMath.h
+++ b/src/SquidMath.h
@@ -10,6 +10,7 @@
 #define _SQUID_SRC_SQUIDMATH_H
 
 #include "base/forward.h"
+#include "base/Optional.h"
 
 #include <limits>
 #include <type_traits>
@@ -28,16 +29,16 @@ double doubleAverage(const double, const double, int, const int);
 
 } // namespace Math
 
-// If Sum() performance becomes important, consider using GCC and clang
+// If IncreaseSumInternal() speed becomes important, consider using compiler
 // built-ins like __builtin_add_overflow() instead of manual overflow checks.
 
 /// std::enable_if_t replacement until C++14
-/// simplifies Sum() declarations below
+/// simplifies IncreaseSumInternal() declarations below
 template <bool B, class T = void>
 using EnableIfType = typename std::enable_if<B,T>::type;
 
 /// detects a pair of unsigned types
-/// reduces code duplication in Sum() declarations below
+/// reduces code duplication in IncreaseSumInternal() declarations below
 template <typename A, typename B>
 using AllUnsigned = typename std::conditional<
                     std::is_unsigned<A>::value && std::is_unsigned<B>::value,
@@ -45,6 +46,7 @@ using AllUnsigned = typename std::conditional<
                     std::false_type
                     >::type;
 
+// TODO: Replace with std::cmp_less() after migrating to C++20.
 /// whether integer a is less than integer b, with correct overflow handling
 template <typename A, typename B>
 constexpr bool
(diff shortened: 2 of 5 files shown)


src/base/ClpMap.h
```diff
@@ -277,8 +277,9 @@ template <class Key, class Value, uint64_t MemoryUsedBy(const Value &)>
 ClpMap<Key, Value, MemoryUsedBy>::Entry::Entry(const Key &aKey, const Value &v, const Ttl ttl) :
     key(aKey),
     value(v),
-    expires(Sum(squid_curtime, ttl).value_or(std::numeric_limits<time_t>::max()))
+    expires(0) // reset below
 {
+    SetToNaturalSumOrMax(expires, squid_curtime, ttl);
```
rousskov, line 282: I do not like this variable setting syntax because it does not look like an assignment and would look weird (and may cause warnings) if used for data member initialization (where the PR currently uses 0 with a "reset below" comment). 

The problem is difficult to fix because we do not want the developer to guess the variable type. In other words, we do **not** want this:

```C++
expires = NaturalSum<long>(squid_curtime, ttl); // XXX: type guessing
```

We could force developers to use decltype, but that feels a bit cruel, might cause const-related warnings when used in initialization lists, and duplicates code:

```C++
expires = NaturalSum<decltype(expires)>(squid_curtime, ttl); // XXX: code duplication; warnings?
```

I was considering adding a dummy/unused char or c-string parameter to stuff a visual assignment clue:

```C++
SetToNaturalSumOrMax(expires, '=', squid_curtime, ttl); // XXX: Too weird?
```

@squid-cache/commiters, @eduard-bagdasaryan, any better ideas?

yadij:
I assume you are using "Natural" to refer to the mathematical definition "natural numbers". Why not simply say "Unsigned" from the C/C++ vernacular?

rousskov:
> I assume you are using "Natural" to refer to the mathematical definition "natural numbers".

Yes, that math term was one of the inspirations. Better function names are welcomed, as always.

> Why not simply say "Unsigned" from the C/C++ vernacular?

NaturalSum() works for both signed and unsigned arguments. It may return a signed value (and already does in the current code IIRC). Proper support for (dangerous) signed integers is one of the primary motivations for adding this feature.

rousskov:
> yadij self-requested a review 2 months ago

@yadij, there is an increasing amount of code that needs this fix, including your own pull requests.

src/SquidMath.h
```diff
@@ -38,54 +38,90 @@ using EnableIfType = typename std::enable_if<B,T>::type;
...
+IncreaseSum(const S s, const T t) {
+    return
+        // We could support a non-under/overflowing sum of negative numbers, but
+        // our callers use negative values specially (e.g., for do-not-use or
+        // do-not-limit settings) and are not supposed to do math with them.
+        (Less(s, 0) || Less(t, 0)) ? Optional<S>() :
+        // Avoids undefined behavior of signed under/overflows. When A is not T,
+        // a or t undergoes (safe) integral conversion in these expressions.
+        // Sum overflow condition: s + t > maxS or, here, maxS - s < t.
+        // If the sum exceeds maxT, integral conversions will use S, not T.
```
yadij, line 92: This does not make sense to me. There is no use of T here regardless of maxT and sum values, everything is in S type.

src/Makefile.am
```diff
@@ -881,6 +881,17 @@ tests_testBoilerplate_LDFLAGS = $(LIBADD_DL)
 # dependency-based list so that simpler code is tested before more complex code
 # which uses it.
 
+## Tests of SquidMath.h
+
```
yadij, line 885:
```suggestion
```

src/Makefile.am
```diff
@@ -881,6 +881,17 @@ tests_testBoilerplate_LDFLAGS = $(LIBADD_DL)
...
+check_PROGRAMS += tests/testMath
+tests_testMath_SOURCES = \
+	tests/testMath.cc
```
yadij, line 888:
```suggestion
tests_testMath_SOURCES = \
	 tests/testMath.cc
nodist_tests_testMath_SOURCES = \
	 SquidMath.h
```

src/base/ClpMap.h
yadij: How about just replacing the Sum(x,y) with NaturalSum(x,y) instead of trying to re-write the entire line of code?
That removes the need for whole global template function just for this one line of code.

yadij (changes requested):

src/SquidMath.h
rousskov: The type of `t` is T. The sum in question (i.e. `s + t`) adds together a variable of type S and a variable of type T. The resulting sum type (_before_ it is copied to our `Optional<S>` object!) will be either S or T -- we do not know the resulting type; only the compiler does. We need to make sure that the sum does not overflow and that the subsequent copying into `Optional<S>` does not overflow.

* The sum may exceed maxS, and we handle/check for that explicitly.
* The sum may exceed maxT. Why do we _not_ check for that? This comment documents why. If the sum exceeds maxT but not maxS[1], then the sum will be of type S, not T due to C++ integral conversion rules (and we do handle/check for maxS overflows).

[1] Why "but not maxS"? Because this sum is computed _only_ if it does not exceed maxS. Our ternary operator is built that way.

src/base/ClpMap.h
rousskov: > How about just replacing the Sum(x,y) with NaturalSum(x,y)

That replacement must not compile: One cannot just replace Sum(a,b) with NaturalSum(a,b) because, unlike the buggy Sum(), NaturalSum() does not try to guess the type of the stored result by the type of the first function argument. That first argument may have nothing to do with the stored result type and, hence, the subsequent assignment (not the sum) may overflow. This is one of the bugs NaturalSum() fixes and prevents.

There are solutions that would compile, but they have other problems as discussed in my [comment](https://github.com/squid-cache/squid/pull/869/files#r678710650).

yadij: Then we cannot use NaturalSum at all for this initialization. The size of time_t may be smaller type than int/Ttl and its max limit varies depending on environment locale - which can make it much smaller than both types.

IIRC, Sum was designed to rely on type conversion rules in order to avoid this type of limitation.

rousskov: > The size of time_t may be smaller type than int/Ttl and its max limit varies depending on environment locale - which can make it much smaller than both types.

Let's assume that is true.

> Then we cannot use NaturalSum at all for this initialization. 

The proposed initialization code effectively uses NaturalSum<time_t>() (without dangerously hard-coding the current `expires` type). That NaturalSum() specialization is supposed to take time_t limits and all the other factors you mentioned into account. If you find any bugs, please point them out!
 
> IIRC, Sum was designed to rely on type conversion rules in order to avoid this type of limitation.

IIRC, your recollection is wrong. Sum() designers were simply unaware of some of the problems they were supposed to solve and, hence, of the bugs their Sum() implementation was introducing. NaturalSum() is now fixing those bugs.

N.B. C++ type conversions do not protect from overflows.

rousskov:
> Can you please sync this to the latest master so I can test it here?

Sure, we are working on integrating the changes I mentioned [earlier](https://github.com/squid-cache/squid/pull/869#issuecomment-953934595) and will sync as well.

yadij:
removing review-1 label because this is a Draft.

rousskov:
> Amos: Can you please sync this to the latest master so I can test it here?

Done. I am also done modifying the branch and am removing the Draft marker. I believe your change requests have been addressed long time ago, but there were substantial branch changes since then. The PR is ready for your review.

yadij (approved):
LGTM, sorry it took so long to get back to this.

PR #873, opened 2021-08-03
Refactor PROXY protocol support

yadij:
PROXY protocol is designed to be implemented as a handshake
prefix immediately following TCP handshake. When initially added
to Squid the ConnStateData logic caused a lot of complications
integrating it and there was no PROXY support for non-HTTP
protocols. Both of those situations have since changed and we
now have a need to promote PROXY protocol handling to a full
Server protocol between TcpAcceptor any other Server protocol
logic.

Refactor the listening port implementation to pass newly
accept(2)'ed client connections to a Pp2Server which handles
PROXY protocol then passes connection ownership to other
services based on squid.conf port settings.

This results in PROXY protocol support (require-proxy-header)
for https_port and ftp_port traffic. However, actual use of the
PROXY extension details by those protocols is omitted for now.


yadij:
This is a Draft for now because it still needs testing and some polish. I am also intending to add a Job sending PROXY protocol header to cache_peers.

rousskov (changes requested):
I hope we do not need to work on this PR until the partially overlapping #832 and #835 are in. The former adds (partial) support for PROXY protocol on https_port and the latter is just a prerequisite you requested to be extracted from the former. That old work significantly improves how PROXY protocol (and other port-specific) support is integrated with the rest of Squid code. This PR should be using those improvements.

~For now, I am abusing the "requesting changes" mode to avoid adding this early draft PR to the "waiting for your review" list generated by GitHub, but if you do not like that abuse, please let me know, and I will revert to the "review requested" state instead.~ Edit: Abuse ended with a real [review](https://github.com/squid-cache/squid/pull/873#pullrequestreview-721221637).

src/servers/Pp2Server.h
```diff
@@ -0,0 +1,64 @@
...
+#ifndef SQUID__SRC_SERVERS_PP2SERVER_H
+#define SQUID__SRC_SERVERS_PP2SERVER_H
+
+#include "acl/ChecklistFiller.h"
+#include "servers/Server.h"
+
+/**
+ * Manage PROXYv2 client connections
+ */
+class Pp2Server : public ::Server, public Acl::ChecklistFiller
```
rousskov, line 20: The Server class[1] represents the manager of the entire from-client connection, with all the protocols that may be happening on that connection. Introducing "PROXY protocol connection manager" is an odd step backwards:

* From high-level design point of view, if anything, we should continue to refactor the code to separate protocol-specific logic into protocol-specific classes that Server/ConnStateData should use as layers that handle protocol-specific bytes, while Server/ConnStateData itself continues to provide connection-wide management. I am not asking you to do that refactoring, and the decision to do it at all should be discussed separately, but that is the direction where we will have to eventually go to survive in the heavily layered world.

* From medium-level design point of view, the PROXY protocol message might be encrypted with TLS. I do not know whether any product supports that today, but there is nothing fundamentally wrong with that, and there can be very good reasons to encrypt all that information. I would not be surprised if Squid eventually gains such support for securely exchanging traffic forwarding information between distributed cache_peers. If we follow this PR approach, adding encrypted PROXY protocol support would mean introducing a TlsServer class (or similar) and then essentially duplicating Pp2Server code to handle the (decrypted) PROXY protocol message before spawning protocol-specific Server jobs that would have to somehow deal with ongoing decryption of the underlying connection. Those kinds of handovers cannot work well, especially when protocol layers must coexist.

* From low-level details point of view, the PROXY protocol handler should not know or care about what comes after the PROXY protocol message. It should not be responsible for deciding "what to do next". Pp2Server::handleReadData() in this PR does that.

[1] The Server class should really be evaluated together with ConnStateData because the current class split is very problematic.

rousskov (changes requested):
> PR description: we now have a need to promote PROXY protocol handling to a full Server protocol between TcpAcceptor any other Server protocol logic.

I see no such need, and I think that staggering protocol processing using Server classes would be a step backwards as detailed in one of the inlined change requests.

If this PR primary goal is to improve code quality (without changing Squid functionality), then I disagree that the direction of these changes is an improvement. If this PR is trying to solve some specific functionality problem, then please clarify what that problem is, and I will do my best to help you find a good solution for it.

src/servers/Pp2Server.h
yadij: > The Server class[1] represents the manager of the entire from-client connection, with all the protocols that may be happening on that connection.

No. I suspect this use of the word "entire" is where you are having problems with ::Server and ConnStateData separation. That is an old design from when HTTP was the *only* type of client connection possible for Squid. Wrapper protocols like PROXY start and *stop* at specific byte positions in a connection I/O. ConnStateData design is not very compatible with that.

class ::Server is supposed to be the base API supplying shared/generic logic for *any* protocol that may occur on a client connection. Allowing them to pass around ownership between Jobs as you see happening in the Pp2Server::handleReadData() method.

rousskov: > class ::Server is supposed to be the base API supplying shared/generic logic for any protocol that may occur on a client connection. Allowing them to pass around ownership between Jobs as you see happening in the Pp2Server::handleReadData() method

I believe that _replacing_ a single connection manager object (today, a ConnStateData child object) with a set of objects each representing a limited-time but exclusive connection owner is a serious design mistake because that model does not represent reality well. In reality, there is such a concept as a from-client connection. It is a single/distinct thing with a lifetime of the entire connection (at least) and various connection-specific properties. Attempts to remove that thing representation will move code further away from modern traffic realities. We do need a class to represent that thing! Today, that class is ConnStateData (including its Server parent). Yes, ConnStateData needs a lot of work (e.g., it has too much HTTP-specific logic and single-protocol assumptions), but it should not become one of the exclusive client connection owners passing control/ownership to other objects with similar capabilities.

PROXY protocol, TLS, HTTP, FTP, TCP tunnel, etc. they are all layers that Server/ConnStateData should coordinate so that the layers can use/share the same from-client connection correctly. All those future layer classes will have a common base class (e.g. ServerLayer), of course, but each of them will not be a Server/ConnStateData _child_ (direct or indirect)! At some point, we will need to extract Server/ConnStateData from Http::One::Server and Ftp::Server classes, converting those classes into ServerLayer kids, while Server/ConnStateData will continue to fulfill its current role of the client connection owner/layer manager.

> Wrapper protocols like PROXY start and stop at specific byte positions in a connection I/O.  ConnStateData design is not very compatible with that.

Yes, and the reality is even worse: On its own, ConnStateData is not very compatible with any modern high-level protocol. On its own, Server is not very compatible with, well, pretty much anything. Even combined, the pair struggles to fulfill its (evolved) purpose of being a client connection manager that does not assume too much about the high-level protocol(s) using that connection. There are lots of problems in the current code! When we added ftp_port support, we solved some of those problems, but most still remain.

Specifics do and will differ quite a bit, but, at a high level, layer switching/manipulation similar to the Pp2Server::handleReadData() method in this PR is happening in ConnStateData today and will continue to happen there. PROXY protocol is just one of the layers that is and will be manipulated.

----

P.S. I believe my interpretation of the current Server/ConnStateData pair as the connection manager is adequate. The split of those two classes was badly implemented, making it impractical to say exactly what Server is in isolation from ConnStateData. For all practical purposes, that split should be ignored IMO. As you know, I wish you would undo that unilateral (and still essentially unused) change, but I cannot force you to do that. However, this old disagreement is very much secondary to the primary problems in this PR.

yadij: > > class ::Server is supposed to be the base API supplying shared/generic logic for any protocol that may occur on a client connection. Allowing them to pass around ownership between Jobs as you see happening in the Pp2Server::handleReadData() method
> 
> I believe that _replacing_ a single connection manager object (today, a ConnStateData child object) with a set of objects each representing a limited-time but exclusive connection owner is a serious design mistake because that model does not represent reality well.

It represents *exactly* the on-wire bytes. 

> In reality, there is such a concept as a from-client connection. It is a single/distinct thing with a lifetime of the entire connection (at least) and various connection-specific properties.

That statement is only true for connection-oriented protocols. Like TCP, TLS or PROXY. HTTP today is also transferred via UDP and COAP(S) where the closest thing to a "client connection" is the HTTP level request-response exchange. HTTP/3 and QUICK have kind of kept the connection-oriented model, but things are diverging from that at a steady pace. In hindsight this type of future began with ICP and HTCP.

> Attempts to remove that thing representation will move code further away from modern traffic realities.

IME, "modern traffic reality" is HTTP/2 and rapidly becoming HTTP/3 and UDP I am remodelling for. TCP-only traffic and Squid's HTTP/1 model of handling connections should be viewed as legacy code supporting old devices and systems by now.

>We do need a class to represent that thing! 

Since forever we already had the TCP SYN handling logic separated out of ConnStateData. The first two packets of the "client connection" are never seen (nor should be) by any logic in ConnStateData.

The PROXY document is clear that the header supplied MUST be sent in a single I/O cycle or be treated as invalid. This makes it no different from a TCP SYN in regards to the "client connection" concept. There is zero reasonable argument for entangling the PROXY I/O or parsing and handling with any of the protocol handling that inherits from ConnStateData.


>Today, that class is ConnStateData (including its Server parent). Yes, ConnStateData needs a lot of work (e.g., it has too much HTTP-specific logic and single-protocol assumptions), but it should not become one of the exclusive client connection owners passing control/ownership to other objects with similar capabilities.
> 

Hmm. What I see you arguing for here is splitting ConnStateData up into a bunch of per-protocol Job classes. The same end-product *except* that your idea is that ConnStateData sticks around as the arbiter of which of those Jobs is currently "owning" [has I/O exclusivity on] the Comm::Connection object.

Consider, that if we do that approach we end up refactoring ConnStateData until it contains only the logic I moved to ::Server (and bits I still hope to put there later). At which point we can rename ConnStateData to be ::Server and voila - same outcome.

IMO the refactor can be done in both directions. Thus I am pulling the generic operations of ConnStateData into ::Server and you are pulling the protocol-specific bits into per-protocol child classes. The transition is tricky, no doubt there, but that is an artifact of ConnStateData itself being so monolithic and how long the process has taken.
 Making it *more* monolithic with this insistence that PROXY stay inside ConnStateData is just adding pain and trouble to the process. It *can* be implemented as entangles with ConnStateData logic - but SHOULD NOT, so that it ceases to interfere with refactoring the other per-protocol code.


> PROXY protocol, TLS, HTTP, FTP, TCP tunnel, etc. they are all layers that Server/ConnStateData should coordinate so that the layers can use/share the same from-client connection correctly. 

The key word is "coordinate". After all the ConnStateData logic is refactored into per-protocol I expect we will have a ::Server method that performs what I placed in Pp2Server::handleReadData() [for now]. This "hand off" is equivalent to ConnStateData::switchToHttps() without the tight entanglement requiring TLS protocol existence.

The "state data" protocols on a "client connection" share should be shared via what is today a MasterXaction instance.

> All those future layer classes will have a common base class (e.g. ServerLayer), of course, but each of them will not be a Server/ConnStateData _child_ (direct or indirect)!

Er. How does none have a "base class" without inheritence in C++?


 * class ::Server provides the code-sharing for logic which is not specific to each protocol.
 * class Foo::Server implements the protocol-specific parts, making use of ::Server logic when needed.
 * When protocol Foo reaches a state where it specification says "stop Foo. All future I/O is protocol Bar" it is wrong to keep performing Foo::Server. Initiate a Bar::Server and pass it everything it needs to know about the connection (i.e. the MasterXaction instance).

> At some point, we will need to extract Server/ConnStateData from Http::One::Server and Ftp::Server classes, converting those classes into ServerLayer kids, while Server/ConnStateData will continue to fulfill its current role of the client connection owner/layer manager.
> 

It appears that what you are calling "ServerLayer" == what I call "::Server" with a few bits that are still tangled up in ConnStateData.

ConnStateData is the entanglement of TLT, HTTP/1, HTTP/2 (TCP flavour), FTP, PROXY, WebSockets.

We do not need to refactor all of them at the same time. The bits of ServerLayer/::Server that are still tightly tied up in ConnStateData are not needed by PROXY.  So "At some point" == today ... at least for PROXY protocol logic.


> > Wrapper protocols like PROXY start and stop at specific byte positions in a connection I/O.  ConnStateData design is not very compatible with that.
> 
> Yes, and the reality is even worse: On its own, ConnStateData is not very compatible with any modern high-level protocol. On its own, Server is not very compatible with, well, pretty much anything. Even combined, the pair struggles to fulfill its (evolved) purpose of being a client connection manager that does not assume too much about the high-level protocol(s) using that connection.

That last statement is my point. The design where a monolithic "connection manager" (aka ConnStateData) is a dinosaur that cannot cope with, as you said, "modern traffic realities".

IMO we need a connection manager that is very hands-off. Provides the generic logic and a way for protocol-specific children (if children at all) to do their thing for periods of time then switch to other children.

All that *needs* to be shared is the metadata about what is/has been happening on that connection. That is stored in class MasterXaction. ... And we have PR #865 making it possible to put all shared metadata/state inside an xaction member instead of multiple Foo::Server members that need copying around on each switch.

> 
> Specifics do and will differ quite a bit, but, at a high level, layer switching/manipulation similar to the Pp2Server::handleReadData() method in this PR is happening in ConnStateData today and will continue to happen there. PROXY protocol is just one of the layers that is and will be manipulated.
> 

As mentioned, long term I see that being a generic method provided by ::Server for several types of protocol to share. For now the operation is exclusively used by PROXY so that is where I placed it. If you would like me to make it available at that top level please as a review request for that shuffling.


> P.S. I believe my interpretation of the current Server/ConnStateData pair as the connection manager is adequate. The split of those two classes was badly implemented, making it impractical to say exactly what Server is in isolation from ConnStateData. 

Here we differ. I think it was adequate, though far from complete and the refactoring project has stalled. In part this PR is trying to reignite it to reach completion and get rid of the transitional issues we are having with its incompleteness.

> For all practical purposes, that split should be ignored IMO. As you know, I wish you would undo that unilateral (and still essentially unused) change, but I cannot force you to do that. However, this old disagreement is very much secondary to the primary problems in this PR.

No. "Devil is in the details".
 * ConnStateData is the old monolithic connection manager model that we know has issues with modern traffic.
 * class ::Server is the beginning (incomplete) extract the protocol-specific behaviour and logics. It *allows* protocols like PROXY to be implemented as ::Server hierarchy children without having to implement the many ConnStateData APIs designed for other protocols behaviours, or workaround the methods wrongly enforcing other protocols behaviour.

PS. I resent the claim of ::Server being a unilateral change. We agreed on the split, but since been disagreeing on the specific of which code should go protocol-specific and which be kept agnostic (aka move to ::Server). Which is expected disagreement given our opposite views on monolithic vs agile "connection manager" implementations.

rousskov: > > > class ::Server is supposed to be the base API supplying shared/generic logic for any protocol that may occur on a client connection. Allowing them to pass around ownership between Jobs as you see happening in the Pp2Server::handleReadData() method
> > 
> > 
> > I believe that _replacing_ a single connection manager object (today, a ConnStateData child object) with a set of objects each representing a limited-time but exclusive connection owner is a serious design mistake because that model does not represent reality well.
> 
> It represents _exactly_ the on-wire bytes.

The reality is a lot richer, more complex than bytes on the wire. When individual protocol layers take care of their bytes on the (often "virtual") wire, we still need a single owner of the whole from-client connection, managing those protocol layers. Today, that concept is represented by ConnStateData. ConnStateData currently does other things as well, things it should not do, but that connection management role is its primary purpose. (And if that purpose is taken away from ConnStateData, another class would need to be created to fulfill it, so we would be exactly where we are today as far as this discussion is concerned, just with a different class name).

> > In reality, there is such a concept as a from-client connection. It is a single/distinct thing with a lifetime of the entire connection (at least) and various connection-specific properties.
 
> That statement is only true for connection-oriented protocols. Like TCP, TLS or PROXY. HTTP today is also transferred via UDP and COAP(S) where the closest thing to a "client connection" is the HTTP level request-response exchange. HTTP/3 and QUICK have kind of kept the connection-oriented model, but things are diverging from that at a steady pace. In hindsight this type of future began with ICP and HTCP.

My statement applies to TCP, TLS, HTTP/1, HTTP/2, HTTP/3, FTP, and other protocols that are in use today and will be in use tomorrow. It applies very well to connection-oriented protocols that use TCP transport (e.g., HTTP/1) and to connection-oriented protocols that use UDP transport (e.g., HTTP/3). The fundamental idea of higher level multi-message protocols layered on top of low-level transport substrate is here to stay for a long time. This PR moves us a step further away from accurately representing that idea by replacing a single connection manager/owner object with multiple ones.


> > Attempts to remove that thing representation will move code further away from modern traffic realities.
> 
> IME, "modern traffic reality" is HTTP/2 and rapidly becoming HTTP/3 and UDP I am remodelling for. TCP-only traffic and Squid's HTTP/1 model of handling connections should be viewed as legacy code supporting old devices and systems by now.

My statements are not specific to TCP-based protocols. The concepts I am talking about also apply to connection-oriented protocols that use UDP (for example).


> > We do need a class to represent that thing!
> 
> Since forever we already had the TCP SYN handling logic separated out of ConnStateData. The first two packets of the "client connection" are never seen (nor should be) by any logic in ConnStateData.

You lost me here. The PR problems I have identified are way above TCP SYN packets. I do not think we should be talking about TCP SYN packets here at all. They are not relevant AFAICT.



> The PROXY document is clear that the header supplied MUST be sent in a single I/O cycle or be treated as invalid.

The statement above is mixing the (highly questionable) sender requirements with non-existent receiver requirements. The receiver is actually required to "wait" (necessarily implying multiple I/Os if necessary) for the entire PROXY protocol message before interpreting it: "The receiver must wait for the CRLF sequence before starting to decode" and "The receiver must not start to parse an address before the whole address block is received".

Yes, there is a "receiver must also reject incoming connections containing partial protocol headers" requirement, but I believe that statement was supposed to be about truncated headers (that may be received over a single or multiple I/Os), not the number of I/Os at the application level.

Anyway, the PROXY protocol and its documentation was designed/written for simpler devices/applications than it can be used for. It could be fun to talk about MSS-driven protocol requirements, but they are not really related to the from-client connection manager!


> There is zero reasonable argument for entangling the PROXY I/O or parsing and handling with any of the protocol handling that inherits from ConnStateData.

Yes, none of the protocol-specific handlers, not PROXY, not HTTP, should inherit from ConnStateData. Protocol-specific handling should be done by protocol-specific layers, leaving ConnStateData to manage the connection as a whole, including switching parsing control from one protocol layer to another as necessary. Today, that split is only partially implemented.


> > Today, that class is ConnStateData (including its Server parent). Yes, ConnStateData needs a lot of work (e.g., it has too much HTTP-specific logic and single-protocol assumptions), but it should not become one of the exclusive client connection owners passing control/ownership to other objects with similar capabilities.
 
> Hmm. What I see you arguing for here is splitting ConnStateData up into a bunch of per-protocol Job classes. The same end-product _except_ that your idea is that ConnStateData sticks around as the arbiter of which of those Jobs is currently "owning" [has I/O exclusivity on] the Comm::Connection object.

We should extract protocol-specific layers from ConnStateData. And the concept of access to I/O is far more nuanced than an "exclusive" lock (because there are two I/O directions and buffering). The end result is very different from what is proposed in this PR.



> Consider, that if we do that approach we end up refactoring ConnStateData until it contains only the logic I moved to ::Server (and bits I still hope to put there later). At which point we can rename ConnStateData to be ::Server and voila - same outcome.

I do not have enough information to accurately assess what you plan for Server, but the _current_ shape of that (essentially unused) class is so far from what a proper connection manager should be that I very much doubt we are talking about the same outcome. For example, a proper connection manager will do no socket I/O at all (that is a SocketLayer responsibility). We should re-merge Server and ConnStateData to improve our chances of arriving at a reasonable connection manager class in the future; the connection/layer management code that needs to stay is very different from the code that is currently in Server.


 
> IMO the refactor can be done in both directions. Thus I am pulling the generic operations of ConnStateData into ::Server and you are pulling the protocol-specific bits into per-protocol child classes. The transition is tricky, no doubt there, but that is an artifact of ConnStateData itself being so monolithic and how long the process has taken.

I agree that there are many ways to refactor. However, our definitions of "generic operations" evidently differ (e.g., socket I/O does not belong to the connection manager), and this PR violates the basic invariant of having a single connection manager object for a single from-client connection. We need to agree on a lot of design decisions before unilateral refactoring actions have a decent chance of succeeding.


> Making it _more_ monolithic with this insistence that PROXY stay inside ConnStateData is just adding pain and trouble to the process.

Besides general layer activation, I do not want (and naturally would not insist on) PROXY protocol logic staying inside ConnStateData. In fact, we have already moved most of that logic outside of ConnStateData: commit 36c774f reduced 200 lines of low-level PROXY protocol parsing code in ConnStateData::parseProxyProtocolHeader() and ConnStateData::parseProxy2p0() methods to 20 lines redirecting parsing to the ProxyProtocol::Parse(). We need to do more for true layering, but that commit will help a lot with doing that in the future.


 
> > PROXY protocol, TLS, HTTP, FTP, TCP tunnel, etc. they are all layers that Server/ConnStateData should coordinate so that the layers can use/share the same from-client connection correctly.
> 
> The key word is "coordinate". After all the ConnStateData logic is refactored into per-protocol I expect we will have a ::Server method that performs what I placed in Pp2Server::handleReadData() [for now]. This "hand off" is equivalent to ConnStateData::switchToHttps() without the tight entanglement requiring TLS protocol existence.

One of the key differences between the proposed Pp2Server::handleReadData() and proper layer management/coordination is that proper coordination is an _ongoing task_ that requires state -- an object that manages the set of protocol layers. It is not a single switch statement with hard-coded hand off logic hidden in one of the protocol layers. That single coordination object is ConnStateData. The switch statement belongs there.

When it comes to PROXY-related layer switching, the logic is currently hard-coded in the official code and in the PR. I see no _need_ to move that logic from ConnStateData where it belongs. Making things worse "for now" is sometimes a valid strategy, but I see no need for making things worse in this particular case.

And, again, we should not violate the key invariant of having a single from-client connection manager. This PR violates that invariant by having two managers for the same from-client connection: A ConnStateData object and a Pp2Server object.


> The "state data" protocols on a "client connection" share should be shared via what is today a MasterXaction instance.


Layer coordination and data sharing require code, not just state. That code cannot be in MasterXaction for many reasons, including the fact that MasterXaction is also (correctly) used for transactions without from-client connections, and the fact that multiple MasterXaction objects may coexist on the same from-client connection. That coordination code currently belongs to ConnStateData, and it is likely to stay there (possibly under a different/better name) because there will be no reasons to move it.

 
> > All those future layer classes will have a common base class (e.g. ServerLayer), of course, but each of them will not be a Server/ConnStateData _child_ (direct or indirect)!
 
> Er. How does one have a "base class" without inheritence in C++?

Sorry, I do not understand why you think there is no place for C++ inheritance in my statement. TlsServerLayer or HttpServerLayer is a ServerLayer (in general and C++ inheritance sense). On the other hand, ServerLayer is not a ConnStateData; ConnStateData is not a ServerLayer; those two class hierarchies do not intersect. ConnStateData _uses_ ServerLayer objects (no C++ inheritance represents that particular relationship).


> * class ::Server provides the code-sharing for logic which is not specific to each protocol.
> 
> * class Foo::Server implements the protocol-specific parts, making use of ::Server logic when needed.
> 
> * When protocol Foo reaches a state where it specification says "stop Foo. All future I/O is protocol Bar" it is wrong to keep performing Foo::Server. Initiate a Bar::Server and pass it everything it needs to know about the connection (i.e. the MasterXaction instance).


I understand that you want to create a chain of protocol handlers, each one with exclusive but temporary from-client connection ownership. FooServer, then BarServer, then BazServer.

The reality we need to continue to represent is layering, not chaining. Most protocols do not know and do not want to know what comes after, under, or above them. They are just doing their own thing while the layers manager coordinates their activity towards the common end goal. HTTP/2 on top of TLS after HTTP/1 (CONNECT) on top of TLS after PROXY protocol on top of TCP.


> > At some point, we will need to extract Server/ConnStateData from Http::One::Server and Ftp::Server classes, converting those classes into ServerLayer kids, while Server/ConnStateData will continue to fulfill its current role of the client connection owner/layer manager.
 
> It appears that what you are calling "ServerLayer" == what I call "::Server" with a few bits that are still tangled up in ConnStateData.

The official code and this PR implies otherwise. ServerLayer is not a from-client connection manager and does not replace itself with another ServerLayer. ConnStateData does all that. The code that was officially misplaced from ConnStateData into Server does not belong to ServerLayer.

 
> ConnStateData is the entanglement of TLT, HTTP/1, HTTP/2 (TCP flavour), FTP, PROXY, WebSockets.
> 
> We do not need to refactor all of them at the same time. The bits of ServerLayer/::Server that are still tightly tied up in ConnStateData are not needed by PROXY. So "At some point" == today ... at least for PROXY protocol logic.

PROXY protocol logic bits should not be doing I/O. They do not belong to Server/ConnStateData hierarchy. Largely, that is already the case in the official code because we have already moved most of those bits away from ConnStateData.


> The design where a monolithic "connection manager" (aka ConnStateData) is a dinosaur that cannot cope with, as you said, "modern traffic realities".

Yes, we are still in agreement here. We disagree on how to fix it.


> IMO we need a connection manager that is very hands-off. Provides the generic logic and a way for protocol-specific children (if children at all) to do their thing for periods of time then switch to other children.

Yes, we need protocol layers. Layers are not children of the connection manager, of course. They have their own hierarchy.



> All that _needs_ to be shared is the metadata about what is/has been happening on that connection. That is stored in class MasterXaction. ... And we have PR #865 making it possible to put all shared metadata/state inside an xaction member instead of multiple Foo::Server members that need copying around on each switch.

A lot more is needed for, say, TLS layer to interoperate with HTTP/1 layer. Having a data exchange point is not enough (and MasterXaction is a bad place to exchange that kind of data anyway!).


> > Specifics do and will differ quite a bit, but, at a high level, layer switching/manipulation similar to the Pp2Server::handleReadData() method in this PR is happening in ConnStateData today and will continue to happen there. PROXY protocol is just one of the layers that is and will be manipulated.
> 
> As mentioned, long term I see that being a generic method provided by ::Server for several types of protocol to share. For now the operation is exclusively used by PROXY so that is where I placed it. If you would like me to make it available at that top level please as a review request for that shuffling.

I hope it is clear by now, but we should not do protocol switching in a PROXY-protocol specific code. There is no need for that even on a temporary basis. And, on another level, PROXY-protocol specific code does not belong to Server/ConnStateData hierarchy. Server/ConnStateData hierarchy is for from-client connection management code.

We can start moving towards the layered model, but this PR moves us away from that, on several levels.


> * ConnStateData is the old monolithic connection manager model that we know has issues with modern traffic.
> 
> * class ::Server is the beginning (incomplete) extraction of the protocol-specific behaviour and logics.

We should not extract protocol-specific behavior and logic into ConnStateData parent. That code does not belong to the Server/ConnStateData hierarchy at all. It belongs to a separate ServerLayer hierarchy. If you meant that Server could be everything excluding protocol-specific behavior and logic, then I might agree, but the official Server is not that, and not even moving in that direction. For example, it does I/O, which is a layer thing.


> It _allows_ protocols like PROXY to be implemented as ::Server hierarchy children without having to implement the many ConnStateData APIs designed for other protocols behaviours, or workaround the methods wrongly enforcing other protocols behaviour.

In its current state, Server is the wrong foundation for anything. It is wrong for ConnStateData (because it only makes things more complex), and it is wrong for the PROXY protocol layer (because it provides non-layering functionality like ProtocolVersion and Pipeline).

I can provide a sketch for a layering-based hierarchy, but moving Squid towards that model requires a lot of difficult decisions, and we cannot even find a way to agree on the basics. I fear it would just create another mountain of overhead with no positive outcome.


----

 
> PS. I resent the claim of ::Server being a unilateral change.

I cannot find my squid-dev review supporting the patch that split ConnStateData. I do not think I have ever meaningfully supported that change. And I doubt anybody else have. Even if we did discuss the general topic of splitting ConnStateData somewhere somehow (and there are certainly ways to split ConnStateData correctly!), the specific split we are dealing with now was unilateral.

PR #874, opened 2021-08-03
Support "file" syntax for 'squid_error' and 'has' ACL parameters

eduard-bagdasaryan:
The 'squid_error' and 'has' are the only ACLs that do not support
loading their parameters from a file using the "path/filename" syntax.
We see no justification for this exception, and it stands in the way of
(unrelated) configuration improvements we are working on (that will,
among other things, prevent such accidents from happening again).

This change causes no upgrade problems because it does not change
handling of previously accepted configurations. It only expands the set
of acceptable configurations, better matching documentation and admin
expectations with regard to universal "path/filename" syntax support.

### DIFF
--- a/src/acl/HasComponentData.cc
+++ b/src/acl/HasComponentData.cc
@@ -29,7 +29,7 @@ ACLHasComponentData::parse()
         self_destruct();
         return;
     }
-    if (ConfigParser::PeekAtToken()) {
+    if (ConfigParser::strtokFile()) {
         debugs(28, DBG_CRITICAL, "FATAL: multiple components not supported for \"has\" acl");
         self_destruct();
         return;
--- a/src/acl/HasComponentData.cc
+++ b/src/acl/HasComponentData.cc
@@ -23,7 +23,7 @@ ACLHasComponentData::ACLHasComponentData()
 void
 ACLHasComponentData::parse()
 {
-    const char *tok = ConfigParser::NextToken();
+    const auto tok = ConfigParser::strtokFile();
     if (!tok) {
         debugs(28, DBG_CRITICAL, "FATAL: \"has\" acl argument missing");
         self_destruct();
--- a/src/acl/HasComponentData.cc
+++ b/src/acl/HasComponentData.cc
@@ -29,7 +29,9 @@ ACLHasComponentData::parse()
         self_destruct();
         return;
     }
+
     parseComponent(tok);
+
     if (ConfigParser::strtokFile()) {
         debugs(28, DBG_CRITICAL, "FATAL: multiple components not supported for \"has\" acl");
         self_destruct();
--- a/src/acl/HasComponentData.cc
+++ b/src/acl/HasComponentData.cc
@@ -29,12 +29,12 @@ ACLHasComponentData::parse()
         self_destruct();
         return;
     }
+    parseComponent(tok);
     if (ConfigParser::strtokFile()) {
         debugs(28, DBG_CRITICAL, "FATAL: multiple components not supported for \"has\" acl");
         self_destruct();
         return;
     }
-    parseComponent(tok);
 }
 
 bool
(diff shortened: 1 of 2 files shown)


src/acl/HasComponentData.cc
```diff
@@ -23,7 +23,7 @@ ACLHasComponentData::ACLHasComponentData()
 void
 ACLHasComponentData::parse()
 {
-    const char *tok = ConfigParser::NextToken();
+    const auto tok = ConfigParser::strtokFile();
     if (!tok) {
         debugs(28, DBG_CRITICAL, "FATAL: \"has\" acl argument missing");
         self_destruct();
```
rousskov, line 29: The `multiple components not supported...` feature below these lines uses PeekAtToken() which is not compatible with the "file" parameters syntax AFAICT. Fortunately, this ACLHasComponentData code does not actually need to peek at any tokens[1] AFAICT -- it can and, hence, should extract the next token (and fail if it exists)! I suggest replacing the PeekAtToken() call with another strtokFile() call.

Here are untested test cases that may help illustrate my point:

```shell
echo request > /tmp/cfg-request
echo foobar > /tmp/cfg-foobar
echo > /tmp/cfg-empty # or something like that
```

Today, due to other squid.conf parsing bugs, I can only think of one relevant test case:

```
# this should pass (rather than reject "/tmp/cfg-empty"):
acl test has request "/tmp/cfg-empty"
```

Eventually, after fixing out-of-scope squid.conf parsing bugs that prevent reading ACL parameters from multiple parameter files on the same line, there will be more relevant test cases:

```
# this should pass "acl test has request" (rather than reject "/tmp/cfg-empty"):
acl test has "/tmp/cfg-request" "/tmp/cfg-empty"

# this should reject foobar (rather than reject "/tmp/cfg-foobar"):
acl test has "/tmp/cfg-empty" "/tmp/cfg-foobar"

# this should pass as "acl test has request"
acl test has "/tmp/cfg-empty" "/tmp/cfg-request"
```

----

[1] Conceptually, peeking at the next token is only needed when we do not want to _consume_ the next token.

rousskov (changes requested):
Thank you for eliminating these unwanted exceptions.

src/acl/HasComponentData.cc
eduard-bagdasaryan: > I suggest replacing the PeekAtToken() call with another strtokFile() call.

Done.

>  other squid.conf parsing bugs

I agree that the strtokFile()-related code is buggy with its "file" reading mode, and these problems are not limited by the inability of reading multiple files. If an ACL starts using strtokFile(), it should call the method (in a loop) until it returns nil - otherwise, the file remains opened and the next ACL will continue using it (leaving the user confused with unrelated error messages or even unaware of the problem at all).

eduard-bagdasaryan: > Another problem may occur if the "file" contains trailing empty line(s).

The problem with empty lines proved to be a little different: it exposed another problem of why we cannot use the result (a pointer to the static buf) of the first strtokFile() after calling the second strtokFile(). If in this particular case (ACLHasComponentData::parse()), the first call returns a valid component (say, 'request'), the second call will read up to the end of the "file" (skipping empty lines), overriding 'buf' with '\n' and finally returning nil. After that, parseComponent() aborts with an error.
I'll try to figure out how to address this problem.

rousskov: Just use/parse the token right after the first strtokFile() call. Check for leftovers after you parsed the token.

eduard-bagdasaryan: Agreed.

rousskov: Thank you for fixing this.

rousskov:
I cleared this seemingly non-controversial small PR for merging. Last chance to stop this PR from being auto-merged when the slow burner timeout expires in less than two days...

yadij (approved):

PR #878, opened 2021-08-11
fixed bug that was found by static code analyzer

Shchelk:
warn: NULL_AFTER_DEREF
Pointer 'fetch->pd' which was dereferenced at peer_digest.cc:702 is
compared to NULL value at peer_digest.cc:709.

### DIFF
--- a/src/peer_digest.cc
+++ b/src/peer_digest.cc
@@ -699,8 +699,9 @@ peerDigestFetchedEnough(DigestFetchState * fetch, char *buf, ssize_t size, const
     PeerDigest *pd = NULL;
     const char *reason = NULL;  /* reason for completion */
     const char *no_bug = NULL;  /* successful completion if set */
-    const int pdcb_valid = cbdataReferenceValid(fetch->pd);
-    const int pcb_valid = cbdataReferenceValid(fetch->pd->peer);
+    /* if "pdcb_valid" or "pcb_valid" is true then "fetch->pd" is dereferenced in "peerDigestReqFinish" */
+    const int pdcb_valid = fetch->pd ? cbdataReferenceValid(fetch->pd) : 0;
+    const int pcb_valid = fetch->pd ? cbdataReferenceValid(fetch->pd->peer) : 0;
 
     /* test possible exiting conditions (the same for most steps!)
      * cases marked with '?!' should not happen */


rousskov (commented):
Thank you very much for working on this bug!

I fixed one bug in PR changes and also added XXXs marking similar problems in the official code. Would you be able to fix those remaining problems?

Technically, we should not even dereference a _valid_ raw fetch->pd pointer[1], but lots of Squid code violates that design principle so we are not enforcing it. My XXXs are only about dereferencing a possibly cbdata-_invalid_ pointers.

[1] The intended type of a cbdata pointer is `void*` or similar, not `PeerDigest*`. Unfortunately, the code was not written that way when cbdata APIs were added eons ago, creating an endless stream of bugs like the one this PR is fixing. This PR will not fix that API problem, of course. The PR only needs to be careful not to dereference possibly invalid pointers.

rousskov:
@Shchelk, can you disclose the name of the static analysis tool that you used? It is not necessary, of course, but it would be nice to document it in the PR description (which will become a [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message) body when this PR is merged by Anubis).

src/peer_digest.cc
```diff
@@ -699,23 +699,26 @@ peerDigestFetchedEnough(DigestFetchState * fetch, char *buf, ssize_t size, const
     PeerDigest *pd = NULL;
     const char *reason = NULL;  /* reason for completion */
     const char *no_bug = NULL;  /* successful completion if set */
-    const int pdcb_valid = cbdataReferenceValid(fetch->pd);
-    const int pcb_valid = cbdataReferenceValid(fetch->pd->peer);
+    
+    // Here, the meaning of "_valid" differs from cbdata validity:
+    // Cbdata treats nil pointers as valid. This code does not.
+    const auto pdcb_valid = fetch->pd && cbdataReferenceValid(fetch->pd);
+    const auto pcb_valid = pdcb_valid && fetch->pd->peer && cbdataReferenceValid(fetch->pd->peer);
```
yadij, line 706: This doesn't look right to me. It checks the value of peer against nullptr before checking the validity of that value.

FTR: these members should ideally be CbcPointer type and we can just rely on that API get() accessor.

rousskov: > This doesn't look right to me. It checks the value of peer against nullptr before checking the validity of that value.

I do not understand why you think that checking x for nil before calling cbdataReferenceValid(x) is not right. That sequence is (or should be) standard practice when one wants to dereference x. cbdataReferenceValid(nullptr) returns a truthy value so calling that function without checking for nil is not enough to dereference the pointer. This is one of the cbdata API problems that we cannot fix in this PR.

Shchelk:
@rousskov The tool is called 'Svace', it's commercial product of ISP RAN. You can use this telegram channel https://t.me/sdl_static to contact svace developers

rousskov:
> The tool is called 'Svace', it's commercial product of ISP RAN.

Thank you. I have updated the PR description to mention the tool name. The name appears to be unique enough to easily find more information about the tool if anybody needs it.

@Shchelk, just clarifying the current PR state: Please note that the ball is on _your_ side to either address the XXXs I have added or decline the opportunity to do so. If you cannot address those XXXs in this PR, then we will decide whether to merge the PR as it is now (even though it leaves arguably similar bugs behind) or wait for another volunteer to fix those XXXs.

Shchelk:
I apologize for answering for along time. Unfortunately, I haven't resources to work on this and i don't understand the code well. It's would be better if someone else fix this.

rousskov:
@Shchelk, thank you for a clear answer and for identifying the primary bug! The Squid Project can take it from here...

@squid-cache/commiters, anybody willing to take over this PR and fix XXXs added in 752b9bd?

PR #879, opened 2021-08-11
Bug 4922: Improve ftp://... filename extraction

rousskov:
Symptoms include: Unexpected ERR_FTP_NOT_FOUND/FTP_REPLY_CODE=550
outcome for GET ftp://... requests with URLs containing `;type=...`.

Broken since commit 51b5dcf.

### DIFF
--- a/src/clients/FtpGateway.cc
+++ b/src/clients/FtpGateway.cc
@@ -1073,7 +1073,7 @@ Ftp::Gateway::checkUrlpath()
     auto t = request->url.path().rfind(';');
 
     if (t != SBuf::npos) {
-        auto filenameEnd = t-1;
+        const auto filenameEnd = t;
         if (request->url.path().substr(++t).cmp(str_type_eq, str_type_eq.length()) == 0) {
             t += str_type_eq.length();
             typecode = (char)xtoupper(request->url.path()[t]);
--- a/src/clients/FtpGateway.cc
+++ b/src/clients/FtpGateway.cc
@@ -1069,16 +1069,17 @@ Ftp::Gateway::checkAuth(const HttpHeader * req_hdr)
 void
 Ftp::Gateway::checkUrlpath()
 {
-    static SBuf str_type_eq("type=");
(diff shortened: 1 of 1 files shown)


yadij:
Looks okay to me for a minimal fix.

Is there any known reason why we use semi-colon for the delimiter between path and query-string? This seems to be a good opportunity to start generating FTP URLs with "?type=" which is a lot more compatible with other HTTP agents.

yadij (approved):

rousskov:
> Is there any known reason why we use semi-colon for the delimiter between path and query-string?

In this particular use case, Squid does not generate these ftp-scheme URL. It is the FTP client (e.g., lftps) that does, when going through an HTTP proxy. Obviously, we have to support what popular clients use.

FWIW, the use of semicolon-delimited `type=typecode` parameter is standard syntax described in these old specs:

* https://datatracker.ietf.org/doc/html/rfc1738#section-3.2.2
* https://datatracker.ietf.org/doc/html/rfc1808#section-2.1

Even the newer RFC [3896](https://datatracker.ietf.org/doc/html/rfc3986) says:

> RFC [3896](https://datatracker.ietf.org/doc/html/rfc3986): For example, the semicolon (";") and equals ("=") reserved characters are often used to delimit parameters and parameter values applicable to that segment

> Amos: This seems to be a good opportunity to start generating FTP URLs with "?type="

This is not my area of expertise, but I suspect that, conceptually, using `?` would be kind of wrong because `?` is meant to start a _query_. Conceptually, the query in these "give me that FTP file" cases either does not exist at all (i.e. the URL just specifies the file location rather than identifying a query resolver and supplying it with a query) OR should include the path and filename (i.e. treating the FTP server as a query resolver). In other words, it would be odd to treat the file itself as a query resolver...

> will be a lot simpler to use a future Uri::query() API in this FTP code.

AFAICT, formally, the `type=typecode` sequence is a URL segment parameter (with a value). Segments belong to `path`, not  `query` parts of a URI.

PR #880, opened 2021-08-12
Maintenance: code style updates in libcompat

yadij:
The libcompat code is targeted here specifically because it is
rarely touched so is not likely to be soon updated by our
"upgrade when modified" policy for source style changes.

Doing this part of the code semi-manually also allows us
to trivially exclude the directory from automated conversion
attempts and not bother with automating the detection of
whether a .h is included in C files.


rousskov (commented):
You have requested my review, but, AFAICT, the PR approach has not changed since I [reviewed](https://github.com/squid-cache/squid/pull/880#pullrequestreview-728755895) it long time ago. My position remains unchanged. In summary: This PR is a dangerous waste of resources. There are much better solutions available.

If you insist on merging this PR, you have the right to do so -- I am not vetoing this PR because I currently do not have the time to implement a known better solution.

PR #512, opened 2019-11-18
ext_lm_group_acl: Improved username handling

aaron-costello:

PR #884, opened 2021-08-12
ext_lm_group_acl: Improved username handling

Roo4L:
Ensure that NTDomain is terminated even if UserName is long.
It happened to be terminated in the current code, but only by accident.

### DIFF
--- a/src/acl/external/LM_group/ext_lm_group_acl.cc
+++ b/src/acl/external/LM_group/ext_lm_group_acl.cc
@@ -336,7 +336,7 @@ Valid_Global_Groups(char *UserName, const char **Groups)
     DWORD i;
     DWORD dwTotalCount = 0;
 
-    strncpy(NTDomain, UserName, sizeof(NTDomain));
+    xstrncpy(NTDomain, UserName, sizeof(NTDomain));
 
     for (j = 0; j < strlen(NTV_VALID_DOMAIN_SEPARATOR); ++j) {
         if ((domain_qualify = strchr(NTDomain, NTV_VALID_DOMAIN_SEPARATOR[j])) != NULL)


rousskov:
OK to test

src/acl/external/LM_group/ext_lm_group_acl.cc
```diff
@@ -336,7 +336,7 @@ Valid_Global_Groups(char *UserName, const char **Groups)
     DWORD i;
     DWORD dwTotalCount = 0;
 
-    strncpy(NTDomain, UserName, sizeof(NTDomain));
+    xstrncpy(NTDomain, UserName, sizeof(NTDomain));
```
rousskov, line 339: Would it be prudent to return 0/failure when UserName does not fit NTDomain (instead of silently truncating it)?

Same for similar copies elsewhere in this file. All these improvements (if it is a good idea at all) do not need to be done in _this_ pull request, of course.

rousskov (approved):
Thank you for fixing this!

rousskov:
@Roo4L, I adjusted PR title/description (future [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message)) to make it similar to that in #512, fix spelling, and remove excessive detail. Please adjust further as needed.

Roo4L:
LGTM

src/acl/external/LM_group/ext_lm_group_acl.cc
Roo4L: Thanks for your comment. I'm not sure, but I believe that during my work on squid project I faced several examples where decision "truncate or return 0/failure" took place and usually this cases used first solution. So I just followed the general project solution. But I would think over your suggestion.

yadij (approved):

src/acl/external/LM_group/ext_lm_group_acl.cc
rousskov: There cannot be a general, Squid-wide "truncate silently (instead of rejecting)" policy -- the decision has to be context-specific.

Also, a lot of old code was written with a completely different set of assumptions/goals in mind. Consistency does matter, but usually not at the expense of _correctness_. In many cases, especially when dealing with old/neglected code, a change is required for moving forward (at some point).

All this does not have to happen in this PR. Given its current state, I think it is best to merge this PR without those changes. However, if you have the time to work on improving related handling beyond this PR, please consider doing that.

PR #886, opened 2021-08-17
Cleanup dependencies for testEvent unit test

yadij:
Many objects linked to the testEvent unit test were there for
legacy reasons and no longer needed.

Converting Event::dump() to Packable interface additionally
removes all dependency on the Store and StoreEntry API. Vastly
simplifying the complexity of the test setup and clarifying
(lack of) code coverage by this test.


src/tests/testEvent.cc
```diff
@@ -66,20 +60,21 @@ testEvent::testCancel()
...
 testEvent::testDump()
 {
     EventScheduler scheduler;
     CalledEvent event;
     CalledEvent event2;
-    CapturingStoreEntry * anEntry = new CapturingStoreEntry();
-    String expect =  "Last event to run: last event\n"
+    MemBuf expect;
+    expect.init();
+    expect.append("Last event to run: last event\n"
```
rousskov, line 72: We should not force developers to count characters in a long c-string with escape sequences!
* If possible, please MemBuf::vappendf("%s") instead of append().
* Otherwise, feed append() a c-string constant and strlen():
```C++
const auto expectedOutput = "Last event...
expect.append(expectedOutput, strlen(expectedOutput));
```

rousskov (commented):
Thank you for cleaning this up.

src/tests/testEvent.cc
yadij: Done in 06101f8

src/event.cc
```diff
@@ -271,24 +271,22 @@ EventScheduler::clean()
...
+        out->appendf("Last event to run: %s\n\n", last_event_ran);
+
+    out->appendf("%-25s\t%-15s\t%s\t%s\n",
+                 "Operation",
+                 "Next Execution",
+                 "Weight",
+                 "Callback Valid?");
+
+    auto *e = tasks;
+    while (e) {
```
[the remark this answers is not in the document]
yadij: Done in 06101f8

src/tests/testEvent.cc
```diff
@@ -66,20 +60,23 @@ testEvent::testCancel()
...
+    const char *expected = "Last event to run: last event\n"
+                           "\n"
+                           "Operation                \tNext Execution \tWeight\tCallback Valid?\n"
+                           "test event               \t0.000 sec\t    0\t N/A\n"
+                           "test event2              \t0.000 sec\t    0\t N/A\n";
+    MemBuf expect;
+    expect.init();
+    expect.append(expected, strlen(expected));
+    CPPUNIT_ASSERT_EQUAL(strlen(expected), size_t(expect.contentSize()));
+    CPPUNIT_ASSERT(strcmp(expect.content(), expected) == 0);
```
rousskov, line 79: I do not think we should check that MemBuf is working correctly in _this_ unit test. These are basic operations/manipulations that this unit test case can (and, hence, should) take for granted.

```suggestion
```

rousskov (approved):

kinkie:
This PR broke the build on arm32:

See https://build.squid-cache.org/job/trunk-arm32-matrix/237/COMPILER=gcc,OS=debian-stable,label=armv7/console

relevant error:
../../../src/tests/testEvent.cc:94:32: error: comparison of integer expressions of different signedness: 'unsigned int' and 'mb_size_t' {aka 'int'} [-Werror=sign-compare]

rousskov:
> This PR broke the build on arm32

I would not object to reverting this PR, but I hope my untested changes in #891 fix this problem. Please test that PR if you can and adjust as needed.

@kinkie, thank you for staying vigilant regarding arm32 support. I hope we will be able to add its tests to the list of visible-on-GitHub tests or even make it a required test.

kinkie:
No need to revert; your fix works very well and I recommend to fast-track it.
If we want to add arm32 and arm64 to the required list, we can, but I'm not super confident of the reliability of the solution, it's a small number of raspberry pi's in my living room.
With money, we could maybe rent a couple of arm compute modules on aws. Let's bring this discussion elswhere

slaiddominio:
@kinkie @yadij I need to compile a version of Squid for Ubuntu 20.04, which supports the SOCKS protocol, could you help me?

Telegram: @LeandroDeFrança
Email: netfiretecc@gmail.com

Telegram: @LeandroDeFrança
Email: netfiretecc@gmail.com

kinkie:
> @kinkie @yadij I need to compile a version of Squid for Ubuntu 20.04, which supports the SOCKS protocol, could you help me?

The best forum to ask for support is the squid-users mailing list; please include a log of the problems you're seeing to avoid unnecessary round trips.

PR #887, opened 2021-08-19
Expand MAX_PKT{4,6}_SZ to account for "icmpEchoData"

sergiodj:
We found this bug while building squid on Ubuntu Impish (current
development version) using GCC-11.  Here's the error we see:

```
  Icmp4.cc: In member function ‘Icmp4::SendEcho(Ip::Address&, int, char const*, int)’:
  Icmp4.cc:116:11: error: array subscript ‘struct icmpEchoData[0]’ is partly outside array bounds of ‘char[282]’ [-Werror=array-bounds]
    116 |     echo->opcode = (unsigned char) opcode;
        |     ~~~~~~^~~~~~
  In file included from ../../include/squid.h:87,
                   from Icmp4.cc:13:
  Icmp4.cc:87:23: note: while referencing ‘Icmp4::SendEcho(Ip::Address&, int, char const*, int)::pkt’
     87 |     LOCAL_ARRAY(char, pkt, MAX_PKT4_SZ);
        |                       ^~~
  Icmp6.cc: In member function ‘Icmp6::SendEcho(Ip::Address&, int, char const*, int)’:
  Icmp6.cc:151:11: error: array subscript ‘struct icmpEchoData[0]’ is partly outside array bounds of ‘char[282]’ [-Werror=array-bounds]
    151 |     echo->opcode = (unsigned char) opcode;
        |     ~~~~~~^~~~~~
  In file included from ../../include/squid.h:87,
                   from Icmp6.cc:13:
  Icmp6.cc:122:23: note: while referencing ‘Icmp6::SendEcho(Ip::Address&, int, char const*, int)::pkt’
    122 |     LOCAL_ARRAY(char, pkt, MAX_PKT6_SZ);
        |                       ^~~

```

The code is still the same on git HEAD.

This happens because the code creates an `struct icmphdr` whose size
is MAX_PKT4_SZ and then casts it into an `struct icmpEchoData`:

```c
  void
  Icmp4::SendEcho(Ip::Address &to, int opcode, const char *payload, int len)
  {
      int x;
      LOCAL_ARRAY(char, pkt, MAX_PKT4_SZ);

      struct icmphdr *icmp = NULL;
      icmpEchoData *echo;
      size_t icmp_pktsize = sizeof(struct icmphdr);
      struct addrinfo *S = NULL;

      memset(pkt, '\0', MAX_PKT4_SZ);

      icmp = (struct icmphdr *) (void *) pkt;

      /*
       * cevans - beware signed/unsigned issues in untrusted data from
       * the network!!
       */
      if (len < 0) {
          len = 0;
      }

      // Construct ICMP packet header
      icmp->icmp_type = ICMP_ECHO;
      icmp->icmp_code = 0;
      icmp->icmp_cksum = 0;
      icmp->icmp_id = icmp_ident;
      icmp->icmp_seq = (unsigned short) icmp_pkts_sent;
      ++icmp_pkts_sent;

      // Construct ICMP packet data content
      echo = (icmpEchoData *) (icmp + 1);
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      echo->opcode = (unsigned char) opcode;
```

(The code above deals with ICMPv4; the same applies to ICMPv6).

We need to expand the `MAX_PKT{4,6}_SZ` definitions in order to
accommodate `struct icmpEchoData`.  The easiest way to do it was to
adding a `sizeof(struct icmpEchoData)` when calculating the maximum
packet size.


src/icmp/Icmp.h
```diff
@@ -16,8 +16,8 @@
 #define PINGER_PAYLOAD_SZ   8192
 
 #define MAX_PAYLOAD 256 // WAS: SQUIDHOSTNAMELEN
-#define MAX_PKT4_SZ (MAX_PAYLOAD + sizeof(struct timeval) + sizeof (char) + sizeof(struct icmphdr) + 1)
-#define MAX_PKT6_SZ (MAX_PAYLOAD + sizeof(struct timeval) + sizeof (char) + sizeof(struct icmp6_hdr) + 1)
+#define MAX_PKT4_SZ (MAX_PAYLOAD + sizeof(struct timeval) + sizeof (char) + sizeof(struct icmphdr) + sizeof(struct icmpEchoData) + 1)
+#define MAX_PKT6_SZ (MAX_PAYLOAD + sizeof(struct timeval) + sizeof (char) + sizeof(struct icmp6_hdr) + sizeof(struct icmpEchoData) + 1)
```
yadij, line 20: The on-wire size of icmpEchoData is already accounted for in these calculations.
```
struct icmpEchoData {
    struct timeval tv;    // sizeof(struct timeval)
    unsigned char opcode; // sizeof(char)
    char payload[MAX_PAYLOAD];
};
```
What could be going on is compiler padding and/or shuffling the structure members around to "optimal" byte positions that do not match the on-wire bits.

A static_assert should be added after struct icmpEchoData definition to verify that struct actually matches the on-wire bitsize. Then we can clarify the MAX_PKT*_SZ definitions:

```
#define MAX_PKT4_SZ (sizeof(struct icmphdr) + sizeof(struct icmpEchoData) + 1)
#define MAX_PKT6_SZ (sizeof(struct icmp6_hdr) + sizeof(struct icmpEchoData) + 1)
```

rousskov: @yadij, do you know what that `+1` is for? Perhaps we should drop it while we are fixing these constants?

rousskov (changes requested):
Thank you for working on this old bug!

> PR description: the code creates an struct icmphdr whose size is MAX_PKT4_SZ and then casts it into an struct icmpEchoData:

FYI: This is not what this ugly code does. The code creates a buffer of size MAX_PKT4_SZ bytes and then uses that buffer for two structures:

1. At offset 0: struct icmphdr
2. At offset sizeof(icmphdr): struct icmpEchoData

The authors intent was correct, but the implementation disregards alignment/padding issues, as Amos have already [noted](https://github.com/squid-cache/squid/pull/887/files#r692654887) in his review.

Please correct the PR title/description (because they will become the official [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message)).

----

N.B. icmpEchoData format is not stable/portable across systems, but since the other side just treats that payload as opaque bytes when sending it back (i.e., the echo request sender is the only agent interpreting the data in the echo response), it is not a big problem.

src/icmp/Icmp.h
sergiodj: Thanks for the reply and the clarification, @yadij.  Indeed, this is exactly what seems to be happening.  I've added a `__attribute__((packed))` at the definition of `icmpEchoData`, and the build passed.

If I discard the modification to `MAX_PKT{4,6}_SZ` and just add a `static_assert` after the structure definition, then obviously the assert triggers (with the on-wire bitsize being `sizeof(struct timeval) + sizeof(char) + MAX_PAYLOAD`).  It seems to me that using `__attribute__((packed))` would be the less intrusive change in this case.

sergiodj:
> Thank you for working on this old bug!
> 
> > PR description: the code creates an struct icmphdr whose size is MAX_PKT4_SZ and then casts it into an struct icmpEchoData:
> 
> FYI: This is not what this ugly code does. The code creates a buffer of size MAX_PKT4_SZ bytes and then uses that buffer for two structures:
> 
>     1. At offset 0: struct icmphdr
> 
>     2. At offset sizeof(icmphdr): struct icmpEchoData

True; I guess I oversimplified the description here.

> The authors intent was correct, but the implementation disregards alignment/padding issues, as Amos have already [noted](https://github.com/squid-cache/squid/pull/887/files#r692654887) in his review.
> 
> Please correct the PR title/description (because they will become the official [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message)).

OK.

sergiodj:
> > The authors intent was correct, but the implementation disregards alignment/padding issues, as Amos have already [noted](https://github.com/squid-cache/squid/pull/887/files#r692654887) in his review.
> > Please correct the PR title/description (because they will become the official [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message)).
> 
> OK.

FWIW, I've edited the PR description but left it incomplete on purpose until a solution for the bug is approved.  Thanks.

src/icmp/Icmp.h
sergiodj: Of course, if having padding on `icmpEchoData` is not a problem (as it seems to be the case based on what @rousskov said below), then another option would be to relax the `static_assert` (like `sizeof(icmpEchoData) >= sizeof all fields)`) and just leave it like that.  I understand that using `__attribute__((packed))` may be frowned upon because it's GCC-specific.

rousskov: > using `__attribute__((packed))` may be frowned upon because it's GCC-specific.

Yes, we should not use non-standard and/or compiler-specific language extensions when general solutions are within reach.

> `sizeof(icmpEchoData) >= sizeof all fields`

We do not need a static assertion that protects us from insanely broken compilers IMO.

We should have a static assertion that protects us from somebody accidentally making the fixed-by-this-PR `MAX_PKT{4,6}_SZ` value smaller again. To avoid looking silly, that assertion should probably be placed in Icmp4::SendEcho() code and use sizeof(pkt) rather than `MAX_PKT4_SZ`. Something along these lines may work:

```C++
static_assert(sizeof(*icmp) + sizeof(*echo) <= sizeof(pkt), "our custom ICMP Echo payload fits the packet buffer");
```

Technically, we must also rework the `(icmp + 1)` code for the same reasons because it is susceptible to post-cmphdr padding in an array context. Our sizeof+sizeof math and assertion assume no such padding... A safer version would cast `icmp` pointer to `char*` and then add `sizeof(*icmp)` instead, before casting to icmpEchoData. Please use C++ casts.

yadij: > @yadij, do you know what that `+1` is for? Perhaps we should drop it while we are fixing these constants?

I do not recall exactly. Probably a hack to ensure '\0' existence after the echo data, so c-string code does not over-read/write memory when data member is not a terminated.

yadij:
@sergiodj , I think Alex and I are in agreement on changing the MAX_PKT_* macros to using sizeof(struct icmphdr) + sizeof(struct icmpEchoData) and allowing the padding to exist on-wire.

Removal of the +1 should be optional IMO. It would be nice to avoid magic numbers, but will require extra analysis work to prove its safe to drop. As author it is up to you to decide if that extra work is doable for you.

src/icmp/Icmp.h
rousskov: > Probably a hack to ensure '\0' existence after the echo data, so c-string code does not over-read/write memory when data member is not a terminated.

If that is true, there is probably a bug somewhere. No code should interpret raw echo data as a c-string AFAICT. It would be nice to investigate and remove `+1` (fixing any related code if needed), but I do not insist on that. Adding a corresponding TODO would be nice so that we do not need to ask this question again.

sergiodj:
> @sergiodj , I think Alex and I are in agreement on changing the
> MAX_PKT_* macros to using sizeof(struct icmphdr) + sizeof(struct
> icmpEchoData) and allowing the padding to exist on-wire.
> 
> Removal of the +1 should be optional IMO. It would be nice to avoid
> magic numbers, but will require extra analysis work to prove its safe
> to drop. As author it is up to you to decide if that extra work and
> scope extension is doable for you.

Sorry about the delay; I got sidetracked with other things happening
at work.

I have now force-pushed the branch with the proposed modifications; I
think I've addressed all of them.  I'm leaving the `+ 1` as is because
it's somewhat orthogonal to the problem I'm trying to address (namely,
the FTBFS on Ubuntu Impish).

Thanks in advance.

src/icmp/Icmp4.cc
```diff
@@ -111,7 +113,7 @@ Icmp4::SendEcho(Ip::Address &to, int opcode, const char *payload, int len)
     ++icmp_pkts_sent;
 
     // Construct ICMP packet data content
-    echo = (icmpEchoData *) (icmp + 1);
+    echo = (icmpEchoData *) ((char *) icmp + sizeof(*icmp));
```
rousskov, line 116: `pkt` is an array of chars so I hope that something like this works:

```suggestion
    echo = reinterpret_cast<icmpEchoData*>(&pkt[sizeof(*icmp)]);
```

but, if necessary, we can be more explicit:

```suggestion
    echo = reinterpret_cast<icmpEchoData*>(reinterpret_cast<char*>(pkt) + sizeof(*icmp));
```

src/icmp/Icmp6.cc
```diff
@@ -146,7 +148,7 @@ Icmp6::SendEcho(Ip::Address &to, int opcode, const char *payload, int len)
     icmp6_pktsize = sizeof(struct icmp6_hdr);
 
     // Fill Icmp6 ECHO data content
-    echo = (icmpEchoData *) (pkt + sizeof(icmp6_hdr));
+    echo = (icmpEchoData *) ((char *) pkt + sizeof(*icmp));
```
rousskov, line 151: Similarly:

```suggestion
    echo = reinterpret_cast<icmpEchoData*>(&pkt[sizeof(*icmp)]);
```

or, if necessary:

```suggestion
    echo = reinterpret_cast<icmpEchoData*>(reinterpret_cast<char*>(pkt) + sizeof(*icmp));
```

rousskov (commented):
I have polished PR description to comply with commit message formatting requirements and remove unnecessary detail clearly visible in the diff. Please adjust further as needed.

If possible, please use C++ type casts. They are often safer and more readable than C-style casts. I left a few specific (but untested!) suggestions. Please adjust them as needed.

sergiodj:
> I have polished PR title/description, primarily to comply with the official commit message formatting requirements and to remove unnecessary details clearly visible in the diff. Please adjust further as needed.
> 
> If possible, please use C++ type casts. They are often safer and more readable than C-style casts. I left a few specific (but untested!) suggestions. Please adjust them as needed.

I've adopted your suggestions now (I chose the more explicit ones; personal preference).  I've also adjusted the commit message to reflect the changes to the PR description.

This is basically your patch now, so feel free to replace the Signed-off-by line with your name.

Thanks.

rousskov (approved):
> I've also adjusted the commit message to reflect the changes to the PR description.

FYI: Anubis ignores individual PR commit _messages_ when squashing PR commits for [merging](https://github.com/measurement-factory/anubis/blob/master/README.md#pull-request-merging-algorithm). Anubis uses PR title/description instead. Feel free to edit those, of course.

> This is basically your patch now, so feel free to replace the Signed-off-by line with your name.

It is _our_ patch -- we all worked on fixing this bug. Most signed-off-by lines are meaningless (at best) in Squid PRs context, and I do not recommend adding them.

However, your contribution should be acknowledged. Would you mind adding your name and email to the `CONTRIBUTORS` file (in this PR)? Again, no need to squash PR commits.

P.S. I pushed a small polishing [change](https://github.com/squid-cache/squid/pull/887/commits/c05d841fcf6979118c2577cf534ba1c8710113a2) to your branch.

sergiodj:
> However, your contribution should be acknowledged. Would you mind adding your name and email to the `CONTRIBUTORS` file (in this PR)? Again, no need to squash PR commits.

Done.

> P.S. I pushed a small polishing [change](https://github.com/squid-cache/squid/pull/887/commits/c05d841fcf6979118c2577cf534ba1c8710113a2) to your branch.

Thanks.

PR #888, opened 2021-08-20
Reduce dependencies of testConfigParser

yadij:
Move several globals used by the legacy parse logic into cache_cf.cc
and remove unnecessary objects linked to testConfigParser.


rousskov (approved):
Thank you for simplifying this unit test.

PR #889, opened 2021-08-21
support type w in gopher.

zcrayfish:
src/gopher.cc mentions the w gopher type code, but doesn't actually handle it at all...
This enables basic support.

A url that can be used for testing: gopher://gopher.zcrayfish.soy/1/irc/test
contains a link to CNN and a link to gopher://gopher.zcrayfish.soy

### DIFF
--- a/src/gopher.cc
+++ b/src/gopher.cc
@@ -600,7 +600,7 @@ gopherToHTML(GopherStateData * gopherState, char *inbuf, int len)
                         snprintf(tmpbuf, TEMP_BUF_SIZE, "\t%s\n", html_quote(name));
                     } else if (gtype == GOPHER_WWW) {	/* Gopher pointer to W3 */
                         snprintf(tmpbuf, TEMP_BUF_SIZE, "<IMG border=\"0\" SRC=\"%s\"> <A HREF=\"%s\">%s</A>\n",
-                                     selector, html_quote(name));                        
+                                     icon_url, selector, html_quote(name));                        
                     } else {
                         if (strncmp(selector, "GET /", 5) == 0) {
                             /* WWW link */
--- a/src/gopher.cc
+++ b/src/gopher.cc
@@ -598,6 +598,9 @@ gopherToHTML(GopherStateData * gopherState, char *inbuf, int len)
 
                     } else if (gtype == GOPHER_INFO) {
                         snprintf(tmpbuf, TEMP_BUF_SIZE, "\t%s\n", html_quote(name));
+                    } else if (gtype == GOPHER_WWW) {	/* Gopher pointer to W3 */
+                        snprintf(tmpbuf, TEMP_BUF_SIZE, "<IMG border=\"0\" SRC=\"%s\"> <A HREF=\"%s\">%s</A>\n",
+                                     selector, html_quote(name));                        
                     } else {
                         if (strncmp(selector, "GET /", 5) == 0) {
                             /* WWW link */


src/gopher.cc
```diff
@@ -598,6 +598,9 @@ gopherToHTML(GopherStateData * gopherState, char *inbuf, int len)
 
                     } else if (gtype == GOPHER_INFO) {
                         snprintf(tmpbuf, TEMP_BUF_SIZE, "\t%s\n", html_quote(name));
+                    } else if (gtype == GOPHER_WWW) {	/* Gopher pointer to W3 */
+                        snprintf(tmpbuf, TEMP_BUF_SIZE, "<IMG border=\"0\" SRC=\"%s\"> <A HREF=\"%s\">%s</A>\n",
+                                     icon_url, selector, html_quote(name));
```
rousskov, line 603: It looks like similar `if` statements around the new code are using `escaped_selector` or even something more complex like `rfc1738_escape_unescaped(selector...)`. Please either document (via a C++ comment) why no `selector` escaping is necessary for `GOPHER_WWW` to prevent XSS/etc attacks OR adjust the code to use a properly escaped version of `selector`, whatever it is.

rousskov (changes requested):
Thank you for working on this Squid improvement!

Unfortunately, there are no Gopher experts among active Squid developers (AFAIK).  I hope I can help you move this PR forward, but I have to say that Gopher is not my area of expertise and I do not know this code well. Please treat my change requests with that important caveat in mind.

Gopher code is neglected and, in my personal opinion, should be removed from Squid in the foreseeable future if nobody steps up to update and maintain it long-term. If you want to be that person, please us know!

rousskov:
> A url that can be used for testing: gopher://gopher.zcrayfish.soy/1/irc/test
> contains a link to CNN and a link to gopher://gopher.zcrayfish.soy

I have removed the above part from the PR description. It is very useful for the PR, but not appropriate for the official commit. The PR title/description will become the official [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message) when this PR is merged.

src/gopher.cc
```diff
@@ -598,6 +598,9 @@ gopherToHTML(GopherStateData * gopherState, char *inbuf, int len)
 
                     } else if (gtype == GOPHER_INFO) {
                         snprintf(tmpbuf, TEMP_BUF_SIZE, "\t%s\n", html_quote(name));
+                    } else if (gtype == GOPHER_WWW) {	/* Gopher pointer to W3 */
```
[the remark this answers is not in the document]
zcrayfish: > Gopher code is neglected and, in my personal opinion, should be removed from Squid in the foreseeable future if nobody steps up to update and maintain it long-term. If you want to be that person, please us know!

While I will do what I can to help improve the gopher code, my C++ skills are not great... as such, I would tend to agree with you, unless more folks step forward to improve the gopher code in Squid, it should eventually be removed.

> Nitpick: Avoid documentation duplication and inconsistent `if` statement styling. We already document what `GOPHER_WWW` is when we declare that constant (and please feel free to update that documentation if needed)...

Roger that!

src/gopher.cc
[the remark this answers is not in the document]
zcrayfish: I will look into this tonight, thanks for the suggestion!

zcrayfish: Sure thing, the selector field for type w is a URL; it's probably a good idea to escape most special characters after the hostname part, I'll see what can be done.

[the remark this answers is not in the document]
zcrayfish: The internal-link icon fits the bill very well, I have updated to use that icon.
For consistency I applied it to the GET / hack as well.

yadij: To be perfectly safe, the code should be using an AnyP::Uri object to parse() the URL received, and use absolute() accessor to get a normalized+escaped URL for the output only on successful parse.

src/gopher.cc
```diff
@@ -598,9 +604,13 @@ gopherToHTML(GopherStateData * gopherState, char *inbuf, int len)
 
                     } else if (gtype == GOPHER_INFO) {
                         snprintf(tmpbuf, TEMP_BUF_SIZE, "\t%s\n", html_quote(name));
+                    } else if (gtype == GOPHER_WWW) {
+                        snprintf(tmpbuf, TEMP_BUF_SIZE, "<IMG border=\"0\" SRC=\"%s\"> <A HREF=\"%s\">%s</A>\n",
+                                     icon_url, rfc1738_escape_part(selector), html_quote(name));
```
rousskov, line 609: Hm. Should not we be adding the URI scheme and `host` to the `HREF` value? Or is `w` used for _relative_ paths on the same Gopher server (implying a `gopher://` scheme, the current host authority, and the current base path).

Is there a good web page that documents the expected selector (and other details) for the `w` type? My quick search did not yield any usable results for that specific type.

rousskov: AFAICT, your test example at `gopher://gopher.zcrayfish.soy/1/irc/test` appears to imply that the selector contains a regular URL with scheme, authority, and path:

```
curl gopher://gopher.zcrayfish.soy/1/irc/test
...
wCNN HTTPS      http://www.cnn.com/     (NULL)  1
wzdo gopher     gopher://gopher.zcrayfish.soy/  (NULL)  1
```

If these `w` selectors are full URLs, then we cannot encode them with rfc1738_escape_part() because that would encode essential delimiters like `:` and `/` (they are a part of the `RFC1738_ESCAPE_RESERVED` set), right?

zcrayfish: > Is there a good web page that documents the expected selector (and other details) for the `w` type? My quick search did not yield any usable results for that specific type.

Not that I know of... I'm just going by the libwww source (which supported the `w` type as far back as 1992) and every real world example I've ever seen; in which the selector has been a full URL.


> If these `w` selectors are full URLs, then we cannot encode them with rfc1738_escape_part() because that would encode essential delimiters like `:` and `/` (they are a part of the `RFC1738_ESCAPE_RESERVED` set), right?

I think rfc1738_escape_unescaped() would work a lot better there, your thoughts?

yadij: Each part of the URL has different encoding requirements. This is why I think using AnyP::Uri will be safer - detection of non-URLs, incomplete URLs, and RFC 3986 encoding of each part of the URL.

rousskov: > > Alex: Is there a good web page that documents the expected selector (and other details) for the `w` type? My quick search did not yield any usable results for that specific type.

> Zachary: Not that I know of... I'm just going by the libwww source (which supported the `w` type as far back as 1992) and every real world example I've ever seen; in which the selector has been a full URL.

I added that critical information to the PR description.
 
> I think rfc1738_escape_unescaped() would work a lot better there, your thoughts?

I agree. Its documentation matches this code needs AFAICT: "Suitable for safely encoding an absolute URL which may be encoded but is not trusted."

rousskov:
OK to test

src/gopher.cc
```diff
@@ -598,9 +602,13 @@ gopherToHTML(GopherStateData * gopherState, char *inbuf, int len)
 
                     } else if (gtype == GOPHER_INFO) {
                         snprintf(tmpbuf, TEMP_BUF_SIZE, "\t%s\n", html_quote(name));
+                    } else if (gtype == GOPHER_WWW) {
+                        snprintf(tmpbuf, TEMP_BUF_SIZE, "<IMG border=\"0\" SRC=\"%s\"> <A HREF=\"%s\">%s</A>\n",
+                                     icon_url, rfc1738_escape_unescaped(selector), html_quote(name));
                     } else {
                         if (strncmp(selector, "GET /", 5) == 0) {
                             /* WWW link */
+                            icon_url = mimeGetIconURL("internal-link");
```
rousskov, line 611: > From https://github.com/squid-cache/squid/pull/889#discussion_r693644055 :
> The internal-link icon fits the bill very well, I have updated to use that icon.
> For consistency I applied it to the GET / hack as well.

Hm... I am worried about this change because, AFAICT, `GET /` selector might be used to help users download something (e.g., a movie). Unlike, `GOPHER_WWW` selectors, the `gtype` corresponding to the `GET /` selector can be one of the other recognized types that have type-specific icons. We should not overwrite that decision unless we did not recognize `gtype` (resulting in an internal-unknown icon earlier).

Moreover, after this PR, Squid will stop applying the `GET /` trick to `GOPHER_WWW` links while still applying it to all other gtypes that produce non-telnet links. Is that a desirable change? Should this `GET /` removal code (continue to) apply to `GOPHER_WWW` links??

Can libwww sources help us address these concerns?

P.S. Originally, I was under the false impression that `GET /` means that gtype is `G` (i.e. the first letter of `GET` is used for gtype) so I did not raise these concerns earlier, but I know see that it is not how the Squid code is actually written -- gtype is always separate from the selector.

rousskov (commented):
I was about to approve this PR (after a minor polishing change I committed to your branch), but I then realized that `GET /` handling may need more work...

zcrayfish:
> > BTW, if you would like to add your name and email to the `CONTRIBUTORS` file, please do that in this PR.
> 
> @zcrayfish, it looks like this PR has been merged, so you would not be able to update `CONTRIBUTORS` now, but if you want that file to be updated, please let us know the exact name/email spelling, and we will add it in a separate PR. Thanks again for your work on this!

You're welcome! When I have time, I will see if there are any other improvements I can make to the gopher code. I don't desire any attribution in the `CONTRIBUTORS` file at this time, but perhaps on a future PR. Thanks for the collaboration on this, and I look forward to more in the future.

PR #893, opened 2021-09-05
HTTP/2

yadij:
Current state of active HTTP/2 handling in Squid and squidclient

Do not expect this code to work yet. There are many TODO/XXX
which need to be implemented before beta testing can begin.


rousskov (changes requested):
The Draft PR does not disclose its goals and does not ask any questions, so I do not know why it was posted. For this review, I am assuming that the PR title will be something like "Tunnel or reject HTTP/2 automatically". Regardless of whether my guess is wrong, please adjust PR title (at least) to reflect the essence of this PR.

Squid has an on_unsupported_protocol feature designed to handle unsupported protocols. AFAICT, that existing feature has better coverage of relevant scenarios and better configuration abilities that what the approach in this PR can offer. I believe that handling of unsupported HTTP/2 should be done using that feature rather than hard-coded logic.

I would support adding a new ACL that can distinguish HTTP/2 client handshake from other protocols so that admins can target HTTP/2 traffic in their on_unsupported_protocol rules.

There are other problems with the PR, but I think it is best to address the above high-level concern before making other adjustments. Hopefully, some of the problematic code changes will become unnecessary by then.

yadij:
FTR; This is a Draft. To show progress and test out possible design plans. There are many XXX / TODO and untested code present.

rousskov:
> This is a Draft. To show progress and test out possible design plans.

Posting a Draft PR to ask specific questions can be a good idea. This PR does not ask any questions. IMO, we should not be using this Squid Project PR service just to "show progress and test out possible design plans", especially without telling others what they should do with the officially submitted PR. Since I cannot guess which draft PRs their authors want me to ignore, and since I do not control (and cannot even monitor) the "draft" state changes, I see no better option than to ignore none of the PRs.

To save time, I continue to ask that every draft PR explicitly discloses why it was officially submitted and/or what reviewers should do with it. I do not think this is too much to ask.

PR #895, opened 2021-09-07
The adaptation_rep_header ACL is a regex match against the response

Opendium:
headers returned by the last adaptation callout (e.g. ICAP).

i.e. the following would make a case insensitive match for the regex
"bar" in the ICAP response header "X-Foo":
  acl icap_says_bar adaptation_rep_header X-Foo -i bar

### DIFF
--- a/src/AclRegs.cc
+++ b/src/AclRegs.cc
@@ -11,6 +11,7 @@
 #if USE_ADAPTATION
 #include "acl/AdaptationService.h"
 #include "acl/AdaptationServiceData.h"
+#include "acl/AdaptationRepHeader.h"
 #endif
 #include "acl/AllOf.h"
 #include "acl/AnnotateClient.h"
@@ -196,6 +197,7 @@ Acl::Init()
 
 #if USE_ADAPTATION
     RegisterMaker("adaptation_service", [](TypeName name)->ACL* { return new ACLStrategised<const char *>(new ACLAdaptationServiceData, new ACLAdaptationServiceStrategy, name); });
+    RegisterMaker("adaptation_rep_header", [](TypeName name)->ACL* { return new ACLStrategised<HttpHeader*>(new ACLHTTPHeaderData, new ACLAdaptationRepHeaderStrategy, name); });
 #endif
 
 #if SQUID_SNMP
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -1422,6 +1422,10 @@ endif
 	  # adaptation_meta because it starts matching immediately after
 	  # the service has been selected for adaptation.
 
+	acl aclname adaptation_rep_header header-name [-i] any\.regex\.here
+	  # regex match against the adaptation reply headers.  e.g. ICAP
+	  # reply headers. [fast]
+
 	acl aclname transaction_initiator initiator ...
 	  # Matches transaction's initiator [fast]
 	  #
--- a/SPONSORS.list
+++ b/SPONSORS.list
@@ -29,6 +29,11 @@ Treehouse Networks, NZ - http://treenet.co.nz/
 	toward Squid-3+ development and maintenance for their customer
 	gateways and CDN.
 
+Opendium - https://opendium.com
+
+	Opendium has contributed bugfixes, tweaks and new features for
+	their online safety products.
+
 @Squid-5:
 RackSpace - https://www.rackspace.com/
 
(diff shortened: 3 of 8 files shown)


src/cf.data.pre
```diff
@@ -1422,6 +1422,10 @@ endif
 	  # adaptation_meta because it starts matching immediately after
 	  # the service has been selected for adaptation.
 
+	acl aclname adaptation_rep_header header-name [-i] any\.regex\.here
+	  # regex match against the adaptation reply headers.  e.g. ICAP
+	  # reply headers. [fast]
```
rousskov, line 1427: Please document that the ACL does not match if the master transaction has seen no responses from adaptation services. Compared to HTTP headers, this is somewhat unusual, because there are always some HTTP headers (or at least that is how most folks think of HTTP-related ACLs). This new ACL will not match during http_access, for example (unless SslBump is in use), even if REQMOD adaptations are enabled for all HTTP requests.

src/cf.data.pre
```diff
@@ -1422,6 +1422,10 @@ endif
 	  # adaptation_meta because it starts matching immediately after
 	  # the service has been selected for adaptation.
 
+	acl aclname adaptation_rep_header header-name [-i] any\.regex\.here
```
rousskov, line 1425: Does this new ACL match adaptation response headers delivered in

1. An error response from the adaptation service (e.g., ICAP 500)?
2. An ICAP 204 (Leave As Is) response (or equivalent)?

rousskov (changes requested):
Thank you for adding this new ACL. I think this feature is generally useful and should be polished and accepted. Please see inlined change requests for details.

> Old PR description: headers returned by the last adaptation callout (e.g. ICAP).

The above claim does not match the allMeta field documentation:

```C++
/// All REQMOD and RESPMOD meta headers merged. Last field wins conflicts.
HttpHeader allMeta;
```

Please review the adjusted PR title/description to make sure they still reflect your intent. The PR title/description will be used as (squashed) [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message) when your PR is merged.

It would be really nice to add an `adaptation_history` or similar component to the existing `has` ACL, but that addition cannot be a valid precondition for merging this PR.

src/AclRegs.cc
```diff
@@ -196,6 +197,7 @@ Acl::Init()
 
 #if USE_ADAPTATION
     RegisterMaker("adaptation_service", [](TypeName name)->ACL* { return new ACLStrategised<const char *>(new ACLAdaptationServiceData, new ACLAdaptationServiceStrategy, name); });
+    RegisterMaker("adaptation_rep_header", [](TypeName name)->ACL* { return new ACLStrategised<HttpHeader*>(new ACLHTTPHeaderData, new ACLAdaptationRepHeaderStrategy, name); });
```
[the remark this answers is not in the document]
yadij: @rousskov , we have an important use-case of minimum footprint proxies. Adaptation is one feature those builds *do* want to remove. As such this ACL and others like it should remain within the USE_* wrappers.

src/cf.data.pre
yadij: Small modification to the request. Please do not use the term "master". There is only one type of transaction that can be passed to adaptation by Squid - an "HTTP transaction".

src/AclRegs.cc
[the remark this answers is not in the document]
rousskov: This particular `#ifdef` removes very little code and memory use while being inconsistent with the rest of the PR where the bulk of the same feature is added unconditionally.  We should not compile and document a feature unconditionally but then save a few bytes by disabling basic configuration support for that feature, making developer and admin jobs more difficult for virtually nothing.

I believe my change request offers the right logic for deciding whether to make this code line conditional. Please note that the change request does not say that the line has to be conditional or unconditional. The correct decision depends on the (unknown to me) answer to the question posed in the change request.

As for the the so called "minimum footprint" proxies support in general, somebody will need to define that concept and the corresponding optimization principles before we can meaningfully decide whether the Project should support that use case and what that support actually means. Otherwise, it is trivial to use that phrase to argue for `#ifdef`-ing 90% of Squid code!

src/cf.data.pre
rousskov: This ACL does not match any specific HTTP transaction. It matches a master transaction that may include two or more HTTP transactions.

rousskov (commented):
I have pushed a few polishing adjustments. None are controversial (AFAICT), and I tried to explain each of them. I have not tested the result beyond compilation, but I do not expect any material changes.

I also added a paragraph to the PR description. That removal kind of surprised me, and it is not trivial to see/check in a diff due to remaining data members with the same name, so I thought it is worth documenting explicitly.

@Opendium, please review and adjust the code and/or PR description further if needed.

rousskov:
OK to test

PR #896, opened 2021-09-12
Fix "mem_obj->inmem_lo == 0" assertion in StoreEntry::swapOut()

rousskov:
Squid may stop writing to disk well before receiving the entire miss
response (e.g., because of swapout errors). When that happens, Squid
continues to call swapOut() to mem-cache the incoming response body.
Avoid the inapplicable disk-writing part of swapOut() in those cases.


rousskov:
This is one of the several Store related assertions that we have fixed in master/v5 code more than a year ago but could not find enough time to polish and post the fixes. This is the simplest of the bunch. More (complex) PRs to follow as we sort through and port those fixes.

yadij (approved):

PR #897, opened 2021-09-15
Bug 5160: Test suite fails with -flto=auto

rousskov:
FAIL: tests/testStore
    Test name: testStoreHashIndex::testSearch
        - check failed: setHow_ != optUnspecified
        exception location: ../src/base/YesNoNone.h(48) operator bool

Segmentation fault in tests/testStore is another symptom.

testStoreHashIndex calls Store::Root().init() which eventually gets to
MemStore::Requested() which correctly requires Config.memShared to be
configured. Hack the missing configuration replacement, following the
inconsistent pattern in three other test suites.

Similarly, Store::Disks::init() needs Config.store_dir_select_algorithm.

TODO: Reduce configuration code duplication and such bugs by providing
initialization method(s) for all Store::Root()-using tests to (re)use.

What does LTO have to do with any of it? CPPUnit [allegedly] randomizes
execution of test suites using raw C++ pointers, and a test suite in the
testStore collection probably contains the right initialization code.
Unfortunately, Config initialization is evidently shared across test
suites (XXX). LTO evidently changes those raw pointer values enough to
change the order in an unfortunate way.

[allegedly]: https://sourceforge.net/p/cppunit/bugs/182/


rousskov:
This is a Draft PR because I cannot test the fix and because I would very much prefer for somebody to take over and improve the test configuration approach as described in the PR description TODO _instead_ of merging the current hack which only increases inconsistency or problem footprint across test suites.

kinkie:
I can take over testing this one if you want. Does it fail everywhere or on any specific OS/compiler combination?

rousskov:
Thank you. I believe the committed workaround has been [confirmed](https://bugs.squid-cache.org/show_bug.cgi?id=5160#c8) as working, so the testing part is not what holds this PR in a draft state right now. Please see Bug [5160](https://bugs.squid-cache.org/show_bug.cgi?id=5160) for any environment details.

yadij:
If the test-suite randomization claim is true, then I am worried about all the other units relying on a setUp() pseudo-test running first to initialize the environment.

kinkie:
This works for me on ubuntu Jammy on gcc

kinkie:
confirmed working on jammy/clang as well.
I'd say it's good to go in

yadij:
Debian apparently plan to require LTO builds of all software as of some time in the next few months. Since this is already affecting Ubuntu builds and the apparent lack of volunteers to pick this up I think we are best to at least accept this workaround while waiting on the test suite improvements.

Moving out of draft status. Lets start the 10-day countdown for objections or reviews as of this status change rather than original commit date.

rousskov:
FWIW, I still do not object to others merging this PR, especially since we lack volunteers to address the underlying problem. Debian policies can be addressed via Debian port patches, but since this PR is a step forward for Squid, it can be merged regardless of what Debian does or does not do.

kinkie:
Happy to be the one to ship this.
From what I see from the docs, we are using setUp correctly according to cppunit's spec.
The other way to do it would be to have our own setup method and explicitly call it at the beginning of each test case - worst case there will be multiple initializations of some component.
Boring but would work; I'd get there if ever we encounter a problem

rousskov:
> Happy to be the one to ship this.

That is great, but why are you not clearing this PR for merging then? Anubis is configured to wait for the `M-cleared-for-merge` label...

rousskov:
> why are you not clearing this PR for merging then? Anubis is configured to wait for the `M-cleared-for-merge` label...

Perhaps you are waiting for the 10-day countdown since https://github.com/squid-cache/squid/pull/897#issuecomment-1172888647. If so, I apologize for the noise. Anubis does enforce the right countdown automatically, but does not yet restart the countdown when a PR stops being a draft; that feature is on the Anubis to-do list.

Technically, you can merge in _two_ days if you approve the PR because it will give it two core votes (explicit yours and implicit mine).

kinkie:
let's just ship it then, and I'll look into any followups as needed

PR #901, opened 2021-09-22
Fixed a copy-paste typo in HttpHdrCc::hasMinFresh()

eduard-bagdasaryan:
This bug could result in unexpected hits/misses because Squid used the
cached CC:max-stale value (or -1) instead of CC:min-fresh when
calculating entry freshness.

Broken since 810d879.


kinkie (approved):
LGTM, thanks for spotting!

rousskov:
OK to test

rousskov (approved):
Thank you for fixing this bug! Glad Co-Advisor could detect it.

yadij (approved):

PR #911, opened 2021-10-13
Fix reconfiguration leaking tls-cert=... memory

rousskov:
Refactored ReadX509Certificate() API for safe use in more contexts,
including in leaking Security::KeyData::loadX509ChainFromFile().

Abstract direct calls to the dangerous PEM_read_bio_X509() API.

Leaking (under certain conditions) since master/v5 commit 540e296.

### DIFF
--- a/compat/openssl.h
+++ b/compat/openssl.h
@@ -23,6 +23,8 @@
 #error compat/openssl.h depends on USE_OPENSSL
 #endif
 
+#include <algorithm>
+
 #if HAVE_OPENSSL_ASN1_H
 #include <openssl/asn1.h>
 #endif


rousskov:
This PR does _not_ fix memory leaks [reported](https://bugs.squid-cache.org/show_bug.cgi?id=5132) by v5 admins because the fixed leak exists in v4 as well AFAICT (commit a3f6783). And the fixed leak is reconfiguration-specific, while some suffering admins do not reconfigure their Squids often enough to notice.

yadij:
Assuming its been tested, so marking for merge.

rousskov:
> Fails on centos-7, both on clang and g++

I hope commit 5f1b8a1 fixes that.

PR #912, opened 2021-10-15
Avoid reverse DNS lookups when logformat %>A is unused

rousskov:
Initially, the log_fqdn directive decided whether to do reverse DNS
lookups of client IP addresses ASAP, to improve our chances of logging
the resolved-by-then client FQDN.

Since commit 7684c4b, Squid started violating log_fqdn configuration,
enabling early reverse lookups if %>A was used in some logformat. Seven
years later, commit c581e96 completely removed the log_fqdn directive.

Unfortunately, the idea that seeing %>A somewhere means early lookups
are needed is flawed because

1. Some logformats containing %>A are never parsed. For example, the
   icap_squid logformat is still hard-coded as a printf()-based code.
   Using that format should enable early reverse lookups, but does not.

2. Some parsed logformats containing %>A may be unused. This is
   especially true for _default_ logformats that admins cannot control
   at all. Initially, no default logformats were parsed, but four years
   later, commit b11724b converted some hard-coded printf()s into parsed
   default logformats, inadvertently enabling reverse lookups in all
   Squid configurations!

   There is no way an admin could turn those lookups off because our
   DEFAULT: lines are parsed unconditionally; overwriting those defaults
   simply means that the corresponding logformat directive is parsed
   twice, once when parsing defaults (enabling the lookups) and then
   when parsing admin settings (the lookups state remains unchanged).

Transactions do not wait for these DNS lookup queries to be answered
and, due to DNS caching, not every client connection triggers a new DNS
query, but the total volume of these lookups may be significant in
environments with many client IP addresses, especially after a fresh
Squid start (or during a flash crowd arrival), when Squid DNS cache is
empty (or has not been primed with a lot of new addresses yet).

This change replaces the "parsed %>A enables early lookups" logic with
the "used %>A enables early lookups" approach. Logging of FQDNs is still
not guaranteed -- the lookups may not be enabled early enough and may
not complete by the time we use %>A -- but Squid instances not using %>A
are now guaranteed to avoid useless reverse DNS lookups of client IPs.

Also fixed logging of client FQDNs in the default ICAP access log format
(a.k.a. icap_squid): That format has the equivalent of a %>A field.
Thus, the icap_squid logging code must _always_ attempt to log FQDN and
enable early reverse DNS lookups. Neither was happening.

### DIFF
--- a/src/AccessLogEntry.cc
+++ b/src/AccessLogEntry.cc
@@ -8,6 +8,7 @@
 
 #include "squid.h"
 #include "AccessLogEntry.h"
+#include "fqdncache.h"
 #include "HttpReply.h"
 #include "HttpRequest.h"
 #include "MemBuf.h"
@@ -44,6 +45,27 @@ AccessLogEntry::getLogClientIp(char *buf, size_t bufsz) const
     log_ip.applyClientMask(Config.Addrs.client_netmask);
 
     log_ip.toStr(buf, bufsz);
+}
+
+const char *
+AccessLogEntry::getLogClientFqdn(char * const buf, const size_t bufSize) const
+{
+    // TODO: Use indirect client and tcpClient like getLogClientIp() does.
+    const auto &client = cache.caddr;
+
+    // internally generated (and ICAP OPTIONS) requests lack client IP
+    if (client.isAnyAddr())
+        return "-";
+
+    // If we are here, Squid was configured to use %>A or equivalent.
+    // Improve our chances of getting FQDN by resolving client IPs ASAP.
+    Dns::ResolveClientAddressesAsap = true; // may already be true
+
+    // Too late for ours, but FQDN_LOOKUP_IF_MISS might help the next caller.
+    if (const auto fqdn = fqdncache_gethostbyaddr(client, FQDN_LOOKUP_IF_MISS))
+        return fqdn; // TODO: Return a safe SBuf from fqdncache_gethostbyaddr().
+
+    return client.toStr(buf, bufSize);
 }
 
 SBuf
--- a/src/AccessLogEntry.h
+++ b/src/AccessLogEntry.h
@@ -55,6 +55,11 @@ public:
     /// including indirect forwarded-for IP if configured to log that
     void getLogClientIp(char *buf, size_t bufsz) const;
 
+    /// %>A: Compute client FQDN if possible, using the supplied buf if needed.
+    /// \returns result for immediate logging (not necessarily pointing to buf)
+    /// Side effect: Enables reverse DNS lookups of future client addresses.
+    const char *getLogClientFqdn(char *buf, size_t bufSize) const;
+
     /// Fetch the client IDENT string, or nil if none is available.
     const char *getClientIdent() const;
 
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1111,6 +1111,7 @@ tests_testRock_SOURCES = \
 	EventLoop.cc \
 	FadingCounter.cc \
 	FileMap.h \
+	tests/stub_fqdncache.cc \
 	tests/stub_HelperChildConfig.cc \
 	HttpBody.cc \
 	HttpBody.h \
@@ -1285,6 +1286,7 @@ tests_testUfs_SOURCES = \
 	EventLoop.cc \
 	FadingCounter.cc \
 	FileMap.h \
+	tests/stub_fqdncache.cc \
 	tests/stub_HelperChildConfig.cc \
 	HttpBody.cc \
 	HttpBody.h \
@@ -1633,6 +1635,7 @@ tests_testDiskIO_SOURCES = \
 	EventLoop.cc \
 	FadingCounter.cc \
 	FileMap.h \
+	tests/stub_fqdncache.cc \
 	tests/stub_HelperChildConfig.cc \
 	HttpBody.cc \
 	HttpBody.h \
--- a/src/SquidConfig.h
+++ b/src/SquidConfig.h
@@ -289,7 +289,6 @@ public:
         int buffered_logs;
         int common_log;
         int log_mime_hdrs;
-        int log_fqdn;
         int announce;
         int mem_pools;
         int test_reachability;
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -33,6 +33,7 @@
 #include "eui/Config.h"
 #include "ExternalACL.h"
 #include "format/Format.h"
+#include "fqdncache.h"
 #include "ftp/Elements.h"
 #include "globals.h"
 #include "HttpHeaderTools.h"
@@ -3985,6 +3986,7 @@ void
 configFreeMemory(void)
 {
     free_all();
+    Dns::ResolveClientAddressesAsap = false;
     Config.ssl_client.sslContext.reset();
 #if USE_OPENSSL
     Ssl::unloadSquidUntrusted();
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -2228,7 +2228,8 @@ ConnStateData::start()
 void
 ConnStateData::whenClientIpKnown()
 {
-    if (Config.onoff.log_fqdn)
+    debugs(33, 7, clientConnection->remote);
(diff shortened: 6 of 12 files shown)


rousskov:
> This is a killer. Reliability of %>A is not something we can compromise on. Modern Squid can (and do) rely on it for security checks, helpers lookups, or redirect URLs (deny_info).

This PR does not compromise anything.  `%>A` was unreliable before this PR and remains unreliable after this PR.

yadij:
> This PR does not compromise anything.

It compromises all cases where the %>A value is needed immediately without DNS lookup delays. That is all directives using fastCheck ACLs that need it, all helper lookup format generations.

> `%>A` was unreliable before this PR and remains unreliable after this PR.

While there are some situations where DNS lag causes the value to still not be available before http_access is processed. That is a bug which needs to be resolved differently to what this PR is doing.

rousskov:
Just to make sure we are on the same page regarding the official code: The probability of FQDN availability in the Squid DNS cache at the time of %>A expansion is not 100%. %>A is not and has never been a reliable mechanism because the DNS query (in case of a DNS cache miss at the client connection acceptance time) may not be answered by the time %>A is expanded, and because the DNS entry in the cache might expire by the time %>A is expanded (in case of a DNS cache hit at the client connection acceptance time). Thus, the %>A feature is just a best-effort, no-guarantees "nice to have that info if we can get it without delays" or "unreliable" feature.

> It compromises all cases where the %>A value is needed immediately without DNS lookup delays.

I cannot find a way to interpret the above statement so that it becomes true. AFAICT:

1. This PR does not change %>A expansion in cases where FQDN is available in Squid DNS cache at the time of expansion.
2. This PR does not change %>A expansion in cases where FQDN is not available in Squid DNS cache at the time of expansion.
3. This PR does not change the timing of reverse DNS lookups (and, hence, does not change the probability of FQDN availability in the cache at the time of %>A expansion) after the very first %>A expansion attempt (for a given Squid instance configuration). 

In summary, after that very first %>A expansion attempt (for a given Squid instance configuration), there are no changes related to %>A expansion. It is as (un)reliable as it was before this PR.

This PR only changes the timing of the very first (in a given Squid instance configuration scope) reverse DNS lookup, from the client connection acceptance time to the very first %>A expansion. Given the unreliable nature of the existing official %>A feature, the significant amount of extra work/noise accidentally created by current official implementation of that feature (that was never intended!), and serious complexity of alternative means of correctly determining %>A use in a given Squid configuration, I think this is the right change that should be welcomed.

If you insist, we can remove the word "configuration" from the above statements by never clearing the Dns::ResolveClientAddressesAsap flag during Squid reconfiguration. That way, once a Squid instance expands %>A, it will always perform those early reverse DNS lookups. I think keeping the flag stale is wrong because it should match the _current_ configuration, but I do not insist on keeping the flag fresh because the primary value of this PR is in stopping millions of unwanted reverse DNS lookups emitted by Squids today. That goal will be achieved even with a stale Dns::ResolveClientAddressesAsap flag.

> While there are some situations where DNS lag causes the value to still not be available before http_access is processed. That is a bug which needs to be resolved differently to what this PR is doing.

The no-delay property of the existing %>A feature is not a bug. It is a core property of that feature! %>A was intended to work that way. I am sure that folks who wanted/added %>A did not need it for authentication and such. They simply wanted a convenient/cheap way to get a rough picture of who is accessing their public forward or reverse proxy. If somebody uses %>A for authentication and such, they are essentially abusing an unreliable feature.

If somebody wants reliable %>A, they need to add a transaction-blocking wait for the DNS answer. Such an implementation would need serious code changes and a configuration option to turn it on/off (or a new %code) because many (I bet most!) current use cases do not _want_ to block transactions to get client FQDN! Thus, we are talking about a new feature. Adding support for that "reliable/blocking %>A" feature is outside this PR scope. This PR just fixes bugs in the existing "no-delay %>A" feature implementation while preserving the core properties of that existing feature.

rousskov:
@yadij, this PR has been waiting for your review for 3 months. AFAICT, your original review was based on a misunderstanding of how `%>A` works in the official code and/or a misunderstanding of how `%>A` works in PR code. In summary, the PR does not change the reliability of `%>A` -- the feature remains unreliable. And for the vast majority of transactions -- all transactions that happen after the first `%>A` use -- there is absolutely no change whatsoever (not in reliability, not in functionality, not in any side effects, none). I have detailed all that [earlier](https://github.com/squid-cache/squid/pull/912#issuecomment-944941618). If those explanations have clarified enough for you to remove your negative vote, please do that. Otherwise, please point out any mistakes in my reasoning or code. Either way, please move this bug-fixing PR forward.

yadij (approved):

redbaron4:
I came across this PR while trying to work out why squid was making PTR lookups for our internal IP addresses.

Originally our proxies were running Squid < 6 where this PR was not merged. However, even after upgrade to v7 (where this PR has been merged), the behaviour has not changed.

- We use default log format and I am sure no `>A` directive occurs in the config
- No `srcdomain` filters exist in our config

The only thing we do have is using `--enable-icap-client` while building Squid from source. Can this be the cause? Setting `icap_enable` to `on` or `off` in squid config does not appear to have any impact.

I am sorry to hijack an old merged PR thread but as there is no Issues tab, this is the only place I could find to have this discussion.

rousskov:
> The only thing we do have is using `--enable-icap-client` while building Squid from source. Can this be the cause? 

Logging ICAP transactions may trigger/enable reverse DNS lookups as detailed in the last paragraph of this PR description, but that does not imply that your reverse DNS lookups happen (exclusively) because of ICAP transactions, especially given your next observation:

> Setting `icap_enable` to `on` or `off` in squid config does not appear to have any impact.

If you can reproduce the problem with `icap_enable off` while running Squid v7+, then I recommend sharing a pointer to the corresponding compressed debugging log with a Squid developer (privately if needed). For a developer like me, it is often easier to detect the actual trigger by looking at the logs than to guess what triggers a DNS lookup in your environment... Squid wiki has relevant [hints](https://wiki.squid-cache.org/SquidFaq/BugReporting#debugging-a-single-transaction).

> I am sorry to hijack an old merged PR thread but as there is no Issues tab, this is the only place I could find to have this discussion.

Squid Project still uses [Bugzilla](https://bugs.squid-cache.org/) for support, but it looks like that service is down right now. It will be fixed, but that fix may take some time. Meanwhile, please feel free to continue to (ab)use this PR for the discussion. CC: @kinkie .

redbaron4:
Thanks @rousskov for those inputs.

We also had `url_rewrite_program` set. Looking at logs I saw that `>A` is being parsed in a logformat. Turns out default format for `url_rewrite_extras` _does_have `A` in it. So that was turning on reverse lookups for us.

If we set `enable_icap off` and define custom `url_rewrite_extras`, the reverse lookups are no longer made.

Unfortunately, our setup requires `icap_enable on` so looks like we have to live with it. Do you have any idea if we can use a configuration directive to stop reverse lookups even when ICAP is enabled. Looking at code I cannot see such an option.

rousskov:
> Do you have any idea if we can use a configuration directive to stop reverse lookups even when ICAP is enabled. 

Avoid explicit or implicit use of `icap_squid` logformat name, which is logformat name used by `icap_log` _by default_. As an extreme example, saying `icap_log none` should solve the problem, but you ought to be able to specify custom logformats with `icap_log` if needed as well. In your custom logformats, if any, do not use `%>A` fields.

redbaron4:
> As an extreme example, saying `icap_log none` should solve the problem

Thanks. I checked the docs and `icap_log` default is `none` atleast in 7.4. Setting it to `none` explicitly also does not stop the reverse lookups. Looking at debug logs, I can't see it parsing any `A` during startup so I turned my attention to ICAP.

Turns out in our setup we use `squidclamav.conf` and it has `dnslookup 1` in its config meaning it will do reverse lookup of client IPs. We set it to `0` (along with other squid.conf changes as written in previous comment) and reverse DNS lookups are no longer generated. We could have achieved the same by setting `icap_send_client_ip off` but I would rather have the ICAP server receive client IPs.

I am very grateful for the pointers you have provided to help us solve it.

PR #913, opened 2021-10-16
Fix double-free segmentation fault on shutdown

yadij:


src/ssl/ProxyCerts.h
```diff
@@ -11,22 +11,51 @@
...
+#include "ssl/gadgets.h"
 
 class sslproxy_cert_sign
 {
 public:
-    int alg;
-    ACLList *aclList;
-    sslproxy_cert_sign *next;
+    sslproxy_cert_sign() = default;
+    sslproxy_cert_sign(const sslproxy_cert_sign &) = delete; // prohibit all copy/move
```
rousskov, line 21: This does [not](https://godbolt.org/z/8Ws4Mnnax) prohibit some copies and some moves. Please delete the move operator instead, as I keep recommending for all such use cases.

```suggestion
    sslproxy_cert_sign(sslproxy_cert_sign &&) = delete; // prohibit all copy/move
```

src/cache_cf.cc
```diff
@@ -4428,13 +4414,12 @@ static void parse_sslproxy_cert_sign(sslproxy_cert_sign **cert_sign)
...
 
-    *cert_sign = cs;
+    *cert_sign = cs.release();
 }
 
 static void dump_sslproxy_cert_sign(StoreEntry *entry, const char *name, sslproxy_cert_sign *cert_sign)
 {
-    sslproxy_cert_sign *cs;
-    for (cs = cert_sign; cs != NULL; cs = cs->next) {
+    for (auto *cs = cert_sign; cs; cs = cs->next) {
```
rousskov, line 4422: If possible, just like in the other similar loop:

```suggestion
    for (const auto *cs = cert_sign; cs; cs = cs->next) {
```

src/ssl/ProxyCerts.h
```diff
@@ -11,22 +11,51 @@
...
+    ~sslproxy_cert_sign() {
+        if (aclList)
+            aclDestroyAclList(&aclList);
```
rousskov, line 24: This non-trivial destructor that needs external code for its implementation should be moved to a `.cc` file IMO, but I do not insist on this change. Your call.

rousskov (changes requested):
Almost there: One bug and two simple polishing requests left.

Please keep applying these change requests to _both_ classes, where applicable. Thank you.

src/ssl/ProxyCerts.h
yadij: This one I will leave in the .h since there are only one place each is used.

rousskov: > since there are only one place each is used

FWIW, the number of current callers is a poor (primary) justification for creating inlining mistakes because it increases technical debt: Even if incorrect inlining does not create immediate problems, the developer adding another caller is unlikely to notice the problem they triggered, and even if they do notice, fixing that problem would be out of their work scope.

src/cache_cf.cc
```diff
@@ -4320,10 +4320,8 @@ static void free_icap_service_failure_limit(Adaptation::Icap::Config *cfg)
 #if USE_OPENSSL
 static void parse_sslproxy_cert_adapt(sslproxy_cert_adapt **cert_adapt)
 {
-    char *al;
-    sslproxy_cert_adapt *ca = (sslproxy_cert_adapt *) xcalloc(1, sizeof(sslproxy_cert_adapt));
-    if ((al = ConfigParser::NextToken()) == NULL) {
-        xfree(ca);
+    auto *al = ConfigParser::NextToken();
+    if (al) {
```
rousskov, line 4324: The above should be written as

```C++
    if (ConfigParser::NextToken()) {
```

... which exposes a PR bug (and the lack of PR testing) -- a missing negation of the `al` condition:

```suggestion
    const auto al = ConfigParser::NextToken();
    if (!al) {
```

src/ssl/ProxyCerts.h
```diff
@@ -11,22 +11,49 @@
...
+    sslproxy_cert_adapt() = default;
+    sslproxy_cert_adapt(sslproxy_cert_adapt &&) = delete; // prohibit all copy/move
+    ~sslproxy_cert_adapt() {
+        xfree(param);
+        if (aclList)
+            aclDestroyAclList(&aclList);
+        while (const auto first = next) {
+            next = first->next;
+            delete first;
+        }
```
rousskov, line 49: I do not insist on this change because I am worried about creating a precedent that will trigger incorrect change requests in future reviews, but this destructor code will be safer if we delete `next` objects _before_ destroying parts of our object (and, hence, entering a dangerous inconsistent state):

```suggestion
        while (const auto first = next) {
            next = first->next;
            delete first;
        }
        if (aclList)
            aclDestroyAclList(&aclList);
        xfree(param);
```

The above order also matches the default C++ class member destruction order, which is usually unimportant, but is a nice bonus.

Same for sslproxy_cert_sign, of course.

rousskov (approved):
Thank you for addressing most of my earlier concerns. I found another bug, but the fix is trivial, so I am approving this PR in anticipation of that fix, to eliminate one review iteration. Please test your changes.

src/ssl/ProxyCerts.h
yadij: Sure. Done in 98298ed. Also added the missing nullptr assignment

src/cache_cf.cc
yadij: Adding the missing ```!```.

Disagree on the "should". Using the ```if (ConfigParser::NextToken()) {``` style requires nesting the entire function implementation inside that if-statement. That is bad for long/complex functions like this one.

rousskov:
Two wrongs (i.e. a dangerous nil pointer and an excessively long function) 
do not make a right, but the core underlying issue here is the wrong parser 
API that forces us to think about this should-never-be-nil pointer in the 
first place. Once we improve the API, this problem will go away naturally.

The "avoid running around with nil pointers" principle will stay, of course.

PR #921, opened 2021-10-26
NDP EUI64 lookups for IPv6 clients

Opendium:
Adds support for looking up an IPv6 client's MAC address from the
neighbour cache.  The resulting MAC address is converted to an EUI64
address.  Previously, the EUI64 was only available for IPv6 clients
with site-local SLAAC addresses, which almost never occur in the wild.

Communication with the kernel is done through a Netlink socket, which
adds blocking I/O to many Squid transactions.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -1113,9 +1113,8 @@ if USE_SQUID_EUI
 	  # It works on Linux, Solaris, Windows, FreeBSD, and some other
 	  # BSD variants.
 	  #
-	  # The eui_lookup directive is required to be 'on' (the default)
-	  # and Squid built with --enable-eui for MAC/EUI addresses to be
-	  # available for this ACL.
+	  # The eui_lookup directive is required to be 'on' for MAC/EUI
+	  # addresses to be available for this ACL.
 	  #
 	  # Squid can only determine the MAC/EUI address for IPv4
 	  # clients that are on the same subnet. If the client is on a
@@ -10588,7 +10587,7 @@ DOC_END
 NAME: eui_lookup
 TYPE: onoff
 IFDEF: USE_SQUID_EUI
-DEFAULT: on
+DEFAULT: off
 LOC: Eui::TheConfig.euiLookup
 DOC_START
 	Whether to lookup the EUI or MAC address of a connected client.


src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+                mac_ptr = NULL;
+                for (nl_reply_hdr = (struct nlmsghdr *) buf; (! done) && NLMSG_OK(nl_reply_hdr, buff_fill); nl_reply_hdr = NLMSG_NEXT(nl_reply_hdr, buff_fill)) {
+                    if (nl_reply_hdr->nlmsg_type == NLMSG_DONE) done = 1;
+                    else if (nl_reply_hdr->nlmsg_type == RTM_NEWNEIGH) {
+                        nl_reply_msg = (struct nlmsghdr *) NLMSG_DATA(nl_reply_hdr);
+                        rt_reply_len = RTM_PAYLOAD(nl_reply_hdr);
+
+                        for (rt_reply_attr = RTM_RTA(nl_reply_msg); (! done) && RTA_OK(rt_reply_attr, rt_reply_len); rt_reply_attr = RTA_NEXT(rt_reply_attr, rt_reply_len)) {
+                            if (rt_reply_attr->rta_type == NDA_DST) {
+                                if (memcmp(ip.s6_addr, RTA_DATA(rt_reply_attr), 16)) {
```
yadij, line 146: memcmp() returns an integer, not a boolean. So we require the (in-)equality check to be explicit.
```suggestion
                                if (memcmp(ip.s6_addr, RTA_DATA(rt_reply_attr), 16) != 0) {
```

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+
+                            if (found_ip && mac_ptr) {
+                                // We've found the right IP and its MAC address.
+                                // Insert 0xfffe to turn the MAC (EUI48) into an EUI64.
+                                memcpy(eui, mac_ptr, 3);
+                                eui[3] = 0xff;
+                                eui[4] = 0xfe;
+                                memcpy(eui+5, ((char *) mac_ptr) + 3, 3);
+                                done = 1;
+                                success = 1;
```
yadij, line 160: Please just exit the method from here. That removes all need for the "done" and "success" state management conditions threading this code.

```suggestion
                                memcpy(eui+5, ((char *) mac_ptr) + 3, 3);
                                close(rtnetlink_sock);
                                return true;
```

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
 
-    /*
-     * Address was not found on any interface
-     */
-    debugs(28, 3, "id=" << (void*)this << ' ' << c << " NOT found");
-#endif /* 0 */
-
-    clear();
-    return false;
+    if (! success) clear();
```
yadij, line 174: When the "return true" is performed by the code above on IP detection this chunk can be reverted to the original code. Only not-found states will reach the end of the method.

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+                                success = 1;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        close(rtnetlink_sock);
+    }
+    if (! success) debugs(28, 4, "id=" << (void*)this << " NDP Fail on " << c);
```
yadij, line 169:
```suggestion
    debugs(28, 4, "id=" << (void*)this << " NDP Fail on " << c);
```

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+            // Loop through all the responses
+            while ((! done) && ((buff_fill = recv(rtnetlink_sock, buf, sizeof(buf), 0)) >= 0)) {
+                mac_ptr = NULL;
+                for (nl_reply_hdr = (struct nlmsghdr *) buf; (! done) && NLMSG_OK(nl_reply_hdr, buff_fill); nl_reply_hdr = NLMSG_NEXT(nl_reply_hdr, buff_fill)) {
+                    if (nl_reply_hdr->nlmsg_type == NLMSG_DONE) done = 1;
+                    else if (nl_reply_hdr->nlmsg_type == RTM_NEWNEIGH) {
+                        nl_reply_msg = (struct nlmsghdr *) NLMSG_DATA(nl_reply_hdr);
+                        rt_reply_len = RTM_PAYLOAD(nl_reply_hdr);
+
+                        for (rt_reply_attr = RTM_RTA(nl_reply_msg); (! done) && RTA_OK(rt_reply_attr, rt_reply_len); rt_reply_attr = RTA_NEXT(rt_reply_attr, rt_reply_len)) {
```
yadij, line 144: Please auto-type in loop iterator. Also the "done" special casing can go when the "return true" is handled as requested.
```suggestion
                        for (auto rt_reply_attr = RTM_RTA(nl_reply_msg); RTA_OK(rt_reply_attr, rt_reply_len); rt_reply_attr = RTA_NEXT(rt_reply_attr, rt_reply_len)) {
```

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+        sa.nl_family = AF_NETLINK;
+        if (sendmsg(rtnetlink_sock, &msg, 0) >= 0) {
+            // Loop through all the responses
+            while ((! done) && ((buff_fill = recv(rtnetlink_sock, buf, sizeof(buf), 0)) >= 0)) {
+                mac_ptr = NULL;
+                for (nl_reply_hdr = (struct nlmsghdr *) buf; (! done) && NLMSG_OK(nl_reply_hdr, buff_fill); nl_reply_hdr = NLMSG_NEXT(nl_reply_hdr, buff_fill)) {
+                    if (nl_reply_hdr->nlmsg_type == NLMSG_DONE) done = 1;
+                    else if (nl_reply_hdr->nlmsg_type == RTM_NEWNEIGH) {
+                        nl_reply_msg = (struct nlmsghdr *) NLMSG_DATA(nl_reply_hdr);
+                        rt_reply_len = RTM_PAYLOAD(nl_reply_hdr);
```
yadij, line 142: C++ casting please. Also these variables can be reduced to this scope and auto-typed.
```suggestion
                        auto nl_reply_msg = static_cast<struct nlmsghdr *>(NLMSG_DATA(nl_reply_hdr));
                        auto rt_reply_len = RTM_PAYLOAD(nl_reply_hdr);
```

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+    msg.msg_flags = 0;
+
+    // Send rtnetlink request
+    rtnetlink_sock = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+    if (rtnetlink_sock >= 0) {
+        memset(&sa, 0, sizeof(sa));
+        sa.nl_family = AF_NETLINK;
+        if (sendmsg(rtnetlink_sock, &msg, 0) >= 0) {
+            // Loop through all the responses
+            while ((! done) && ((buff_fill = recv(rtnetlink_sock, buf, sizeof(buf), 0)) >= 0)) {
```
yadij, line 136: After the "return true" change is made this condition becomes much simpler

```suggestion
            while ((buff_fill = recv(rtnetlink_sock, buf, sizeof(buf), 0)) >= 0) {
```

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+                mac_ptr = NULL;
+                for (nl_reply_hdr = (struct nlmsghdr *) buf; (! done) && NLMSG_OK(nl_reply_hdr, buff_fill); nl_reply_hdr = NLMSG_NEXT(nl_reply_hdr, buff_fill)) {
```
yadij, line 138:
```suggestion
                for (auto nl_reply_hdr = static_cast<struct nlmsghdr *>(buf); NLMSG_OK(nl_reply_hdr, buff_fill); nl_reply_hdr = NLMSG_NEXT(nl_reply_hdr, buff_fill)) {
```

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+                    else if (nl_reply_hdr->nlmsg_type == RTM_NEWNEIGH) {
+                        nl_reply_msg = (struct nlmsghdr *) NLMSG_DATA(nl_reply_hdr);
+                        rt_reply_len = RTM_PAYLOAD(nl_reply_hdr);
+
+                        for (rt_reply_attr = RTM_RTA(nl_reply_msg); (! done) && RTA_OK(rt_reply_attr, rt_reply_len); rt_reply_attr = RTA_NEXT(rt_reply_attr, rt_reply_len)) {
+                            if (rt_reply_attr->rta_type == NDA_DST) {
+                                if (memcmp(ip.s6_addr, RTA_DATA(rt_reply_attr), 16)) {
+                                    // IP doesn't match - don't loop through the rest of the RTAs
+                                    break;
+                                } else found_ip = 1;
```
yadij, line 149: Example of a one-liner needing brackets added because the rest of the statement uses them. There are others
```suggestion
                                } else {
                                    found_ip = 1;
                                }
```

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+    msg.msg_namelen = sizeof(sa);
+    msg.msg_iov = &iov;
+    msg.msg_iovlen = 1;
+    msg.msg_control = NULL;
+    msg.msg_controllen = 0;
+    msg.msg_flags = 0;
+
+    // Send rtnetlink request
+    rtnetlink_sock = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+    if (rtnetlink_sock >= 0) {
```
yadij, line 131: You can reduce scope nesting of all this code by checking for errors instead of success. The if-statement should perform any appropriate debugs() log entry, cleanup and exit.

src/eui/Eui64.cc
```diff
@@ -75,22 +80,99 @@ Eui::Eui64::lookupSlaac(const Ip::Address &c)
...
+#if _SQUID_LINUX_
+    int rtnetlink_sock;
+    struct {
+        struct nlmsghdr hdr;
+        struct ndmsg ndm;
+    } req;
+    struct sockaddr_nl sa;
+    struct iovec iov;
+    struct msghdr msg;
+    char buf[16384];
```
yadij, line 96: Is there any particular reason for using a 16K buffer here on the stack?
Please consider using SBuf or MemBuf for large-ish buffers that cannot be "static char". That will also allow some later adjustments to allocate based on read() needs rather than magic numbers.

yadij (changes requested):
Style review mostly. Though there are several logic changes that can be made by using C++ style instead of C-style.

I have added notes to several of these but there are more lines needing updating to meet these style issues:
* C++ variables are initialized as late as possible.  Where they exist solely to store a returned value use "auto" if possible. eg  ``` auto rtnetlink_sock = socket(...); ```
* In Squid we require one code statement per line. There are several if-else which have code written as a line suffix. eg ``` if (...) ... code; ```
* When one part of an if-else sequence requires {brackets} they should both be given them for clarity.

PR #923, opened 2021-10-27
Fix HappyConnOpener::checkForNewConnection Must(prime) violation

rousskov:
This change addresses a known problem that triggered unwanted C++
exceptions every time Squid selected a to-server persistent connection
as the primary Happy Eyeballs destination/answer. The bug existed since
HappyConnOpener inception (commit 5562295). It did not seem to affect
the connection requestor directly because the HappyConnOpener job sends
the selected pconn to the requestor _before_ throwing.

Also adjusted Happy Eyeballs state documentation to reflect a successful
termination state. Before and after this change, we may enter that state
in the middle of checkForNewConnection(). The less we think about done()
as an _exceptional_ "only on error" or "only at the end of processing"
state, the fewer similar bugs we will create.

The code also improved after we abandoned the idea of documenting all
primary state changes in checkForNewConnection(). There are too many
nuances/changes to document everything anyway, and moving primary track
handling into a dedicated function significantly improves readability.

### DIFF
--- a/src/HappyConnOpener.cc
+++ b/src/HappyConnOpener.cc
@@ -769,11 +769,10 @@ HappyConnOpener::noteSpareAllowance()
 void
 HappyConnOpener::maybeOpenPrimeConnection()
 {
-    if (prime)
-        return; // nothing to do
+    Must(!prime);
 
     if (destinations->empty())
-        return; // nothing can be done (for now)
+        return;
 
     if (!currentPeer) {
         auto newPrime = destinations->extractFront();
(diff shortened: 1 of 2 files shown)


rousskov:
FWIW, the signs of the problem fixed by this PR are present in bug [5214](https://bugs.squid-cache.org/show_bug.cgi?id=5214) logs for Squid v5. The problematic code itself is definitely present in v5 sources. @yadij, I recommend that this fix is backported to v5. Please let me know if you need Factory help with that.

PR #925, opened 2021-10-29
Alternate origin server selection

Opendium:
When the proxy is making a direct connection to the origin server, it
will now connect to the host contained in the alt-host note, or
X-Alt-Host ICAP response header, if such a note / header exists. This
allows an external ACL or ICAP server to redirect the request to a
different host.

This is useful where services have a specific
"virtual IP" that clients can connect to to receive a modified
service. e.g. this can be used to force the "Safe Search" feature on
for certain search engines.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -909,7 +909,6 @@ FwdState::noteConnection(HappyConnOpener::Answer &answer)
             request->method == Http::METHOD_CONNECT ||
             request->flags.sslPeek ||
             clientFirstBump;
-
         if (originWantsEncryptedTraffic && // the "encrypted traffic" part
                 !peer->options.originserver && // the "through a proxy" part
                 !peer->secure.encryptTransport) // the "exclude HTTPS proxies" part


rousskov (changes requested):
Please do not misinterpret this change request as a vote against this PR in principle. While I see some problems with this PR that can and should be addressed, I am currently only requesting additional high-level information to improve the quality of my future review (if any).


| routing approach || what may change |||
| --- | :---: | :---: | :---: | :---: |
|  | cache key | SNI | protocol | dst IP |
| /etc/hosts | no | no | no | yes |
| Alt-Svc (RFC [7838](https://datatracker.ietf.org/doc/html/rfc7838))[^Alt] | no | no | yes | yes |
| cache_peer originserver | no | yes | yes | yes |
| url_rewrite_program or adaptation | yes | yes | yes | yes |

[^Alt]: The changes in the Alt-Svc row are listed for the _combination_ of the first and second request, assuming Squid is modified to react to Alt-Svc information in origin responses. The listed changes (i.e. the "yes" table cells) happen during the _second_ request (sent to the negotiated alternative service).

Given the above (very rough and approximate!) summary of request rerouting options, could you please clarify whether the _intent_ of this PR is to provide 

1. a "dynamic" equivalent of `cache_peer originserver` peer selection (same basic feature but without the need to add a `cache_peer` directive for every alternative domain),

2. an Alt-Svc substitution (as if the virgin origin server responded with `Alt-Svc` header containing the desirable value, and Squid then automatically used that value as an RFC 7838-aware client would), or

3. something completely different?

yadij (changes requested):
I have two design requests:

 * Please remove the alt-host= option from external_acl_type helper. Authorization is not the right place to be modifying the request message. That is what url_rewrite_helper is intended for  and does better already.

 * IMO it would be better to begin implementing RFC7838 Alt-Svc rather than an X-* header that lacks documentation and possibly good review. There subtle details that Alt-Svc accounts for regarding TLS certificate values and Host obfuscation at the alternative server.

rousskov:
FWIW, I recommend avoiding PR code changes until everybody involved is on the same page with regard to what should be changed.

PR #926, opened 2021-10-30
Remove step2+ stare-and-splice and peek-and-bump support

chtsanti:
Support for these features was successfully disabled (at build time) for
five years. See commit 88a300c for the list of reasons. We did not even
provide an ./configure --enable-... option for it. The corresponding
code does not compile with modern OpenSSL versions, but its mere
presence complicates related code logic and significantly increases
related development efforts. This code is not worth keeping.


yadij (approved):
Excellent news. Thank you.

yadij (changes requested):
There is a use of SQUID_CHECK_OPENSSL_HELLO_OVERWRITE_HACK in configure.ac which also needs to be removed.

chtsanti:
> There is a use of SQUID_CHECK_OPENSSL_HELLO_OVERWRITE_HACK in configure.ac which also needs to be removed.

commit ac5564ae1

yadij (approved):

PR #927, opened 2021-10-31
Bug 5060: Parallel builds are not reliable

ffontaine:
Create tests directory before using it. Needed since commits 44e802f and
9ba9313.

    cp ../../src/tests/stub_debug.cc tests/stub_debug.cc
    cp ../../src/tests/stub_libmem.cc tests/stub_libmem.cc
    cp: cannot create regular file 'tests/stub_debug.cc':
        No such file or directory

### DIFF
--- a/src/icmp/Makefile.am
+++ b/src/icmp/Makefile.am
@@ -95,17 +95,20 @@ time.cc: $(top_srcdir)/src/time.cc
 SquidConfig.cc: $(top_srcdir)/src/SquidConfig.cc
 	cp $(top_srcdir)/src/SquidConfig.cc $@
 
-tests/stub_HelperChildConfig.cc: $(top_srcdir)/src/tests/stub_HelperChildConfig.cc
+tests/stub_HelperChildConfig.cc: $(top_srcdir)/src/tests/stub_HelperChildConfig.cc | tests
 	cp $(top_srcdir)/src/tests/stub_HelperChildConfig.cc $@
 
-tests/stub_fd.cc: $(top_srcdir)/src/tests/stub_fd.cc
+tests/stub_fd.cc: $(top_srcdir)/src/tests/stub_fd.cc | tests
 	cp $(top_srcdir)/src/tests/stub_fd.cc $@
 
-tests/stub_libmem.cc: $(top_srcdir)/src/tests/stub_libmem.cc
+tests/stub_libmem.cc: $(top_srcdir)/src/tests/stub_libmem.cc | tests
 	cp $(top_srcdir)/src/tests/stub_libmem.cc $@
 
-tests/stub_SBuf.cc: $(top_srcdir)/src/tests/stub_SBuf.cc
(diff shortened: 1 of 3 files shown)


rousskov (commented):
I am surprised there is no better way to create a prerequisite directory than to attempt to (remember to) do it for each relevant file, but I cannot offer a better solution, and, hence, will approve this fix if needed.

ffontaine:
An other option, and probably better, is [order-only prerequisite](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html):

```
tests/stub_debug.cc: $(top_srcdir)/src/tests/stub_debug.cc | tests

tests:
        mkdir -p $@
```

I'll update the PR.

rousskov (commented):
Assuming `| tests` is portable, I agree that it is an improvement, thank you!

I still do not like that we have to (remember to) do something for _every_ affected source file (instead of once for a directory containing such files), but perhaps Automake just does not support anything better than that.

rousskov:
I polished PR title/description, primarily to satisfy our formatting requirements for [commit messages](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message). Please adjust further if needed.

rousskov (commented):
Hm.. I noticed that besides tools/ and tools/squidclient/ there are two other source code directories that copy test stubs: src/icmp/ and test-suite/. Why do they _not_ need the same fix?

ffontaine:
An other possibility is to create the tests directory as soon as the Makefile is loaded, for example through:
```
_dummy := $(shell mkdir -p tests)
```
I can update the PR if you prefer this solution.

rousskov:
I am not an Automake expert so I would rather not make judgement calls regarding what looks like a hack to me. I will let Amos decide on the best way forward here when he reviews this PR. Hopefully, we can apply the same high-quality solution to all directories where stubs are copied.

ffontaine:
> Hm.. I noticed that besides tools/ and tools/squidclient/ there are two other source code directories that copy test stubs: src/icmp/ and test-suite/. Why do they not need the same fix?

icmp needs the same fix but I didn't spot it because it is disabled by default.
test-suite doesn't need it because test stubs are not copied to a tests directory.

rousskov:
OK to test

yadij (approved):

yadij:
> I polished PR title/description, primarily to satisfy our formatting requirements for [commit messages](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message). Please adjust further if needed.

I have a possible cleaner alternative. But it requires some major redesign of the testing systems we have. Meanwhile this is as good as we can get with current Squid build system.

sthen:
> Assuming `| tests` is portable, I agree that it is an improvement, thank you!

It seems that `| tests` is a GNU make extension.

rousskov:
Please share build failure details if you have them. Filing a bug report with Squid [Bugzilla](https://bugs.squid-cache.org/query.cgi) may be a good idea.

Hopefully, somebody will come up with a more portable fix or, better, get rid of these file copying hacks!

PR #929, opened 2021-11-07
Fix build on openbsd 7.0

kinkie:
OpenBSD doesn't offer CPU_SET and CPU_ISSET. Implement their stubs as
inline functions to give the compiler proper hints about arguments (non)
use.

We have a const-correctness bug in std::unordered_map when supplying an
allocator that OpenBSD is strict about. Fix it.

Update buildtest.sh to try and use relative paths first. This prevents
autoconf complaining and failing if the directory path includes
characters from an unsafe set.

### DIFF
--- a/acinclude/os-deps.m4
+++ b/acinclude/os-deps.m4
@@ -968,30 +968,3 @@ AC_DEFUN([SQUID_CHECK_BROKEN_SOLARIS_IPFILTER],[
   ])
 
 ])
-
-AH_TEMPLATE(STD_UNORDERED_MAP_ALLOCATOR_WORKS,
-  [OpenBSD brokenness: for some reason it messes up const-ness in std::unordered_map if a allocator is supplied])
-AC_DEFUN([SQUID_CHECK_UNORDERED_MAP_ALLOCATOR_WORKS],[
-  AC_CACHE_CHECK([if std::unordered_map can accept an allocator],
-    squid_cv_std_unordered_map_allocator_works,
-    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
-#include <memory>
-#include <unordered_map>
-template <class T>
-class SomeAllocator : public std::allocator<T>
-{};
-
-using Element = std::pair<int, uint64_t>;
-using AllocatorElement = std::pair<const int, uint64_t>;
-using Map = std::unordered_map<int, uint64_t, std::hash<int>, std::equal_to<int>, SomeAllocator<AllocatorElement> >;
-
-static Map testMap;
-]], [[
-]])],
-    [squid_cv_std_unordered_map_allocator_works=yes],
-    [squid_cv_std_unordered_map_allocator_works=no])
-  )
-  AS_IF([test "x$squid_cv_std_unordered_map_allocator_works" = "xyes"],[
-    AC_DEFINE([STD_UNORDERED_MAP_ALLOCATOR_WORKS],1)
-  ])
-])
--- a/acinclude/os-deps.m4
+++ b/acinclude/os-deps.m4
@@ -968,3 +968,29 @@ AC_DEFUN([SQUID_CHECK_BROKEN_SOLARIS_IPFILTER],[
   ])
 
 ])
+
+AH_TEMPLATE(STD_UNORDERED_MAP_ALLOCATOR_WORKS,
+  [OpenBSD brokenness: for some reason it messes up const-ness in std::unordered_map if a allocator is supplied])
+AC_DEFUN([SQUID_CHECK_UNORDERED_MAP_ALLOCATOR_WORKS],[
+  AC_CACHE_CHECK([if std::unordered_map can accept an allocator],
+    squid_cv_std_unordered_map_allocator_works,
+    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+#include <memory>
+#include <unordered_map>
+template <class T>
+class SomeAllocator : public std::allocator<T>
+{};
+
+using Element = std::pair<int, uint64_t>;
+using Map = std::unordered_map<int, uint64_t, std::hash<int>, std::equal_to<int>, SomeAllocator<Element> >;
+
+static Map testMap;
+]], [[
+]])],
+    [squid_cv_std_unordered_map_allocator_works=yes],
+    [squid_cv_std_unordered_map_allocator_works=no])
+  )
+  if test "x$squid_cv_std_unordered_map_allocator_works" = "xyes" ; then
+    AC_DEFINE([STD_UNORDERED_MAP_ALLOCATOR_WORKS],1)
+  fi
+])
--- a/acinclude/os-deps.m4
+++ b/acinclude/os-deps.m4
@@ -990,7 +990,7 @@ static Map testMap;
     [squid_cv_std_unordered_map_allocator_works=yes],
     [squid_cv_std_unordered_map_allocator_works=no])
   )
-  if test "x$squid_cv_std_unordered_map_allocator_works" = "xyes" ; then
+  AS_IF([test "x$squid_cv_std_unordered_map_allocator_works" = "xyes"],[
     AC_DEFINE([STD_UNORDERED_MAP_ALLOCATOR_WORKS],1)
-  fi
+  ])
 ])
--- a/acinclude/os-deps.m4
+++ b/acinclude/os-deps.m4
@@ -982,7 +982,8 @@ class SomeAllocator : public std::allocator<T>
 {};
 
 using Element = std::pair<int, uint64_t>;
-using Map = std::unordered_map<int, uint64_t, std::hash<int>, std::equal_to<int>, SomeAllocator<Element> >;
+using AllocatorElement = std::pair<const int, uint64_t>;
+using Map = std::unordered_map<int, uint64_t, std::hash<int>, std::equal_to<int>, SomeAllocator<AllocatorElement> >;
 
 static Map testMap;
 ]], [[
--- a/compat/cpu.h
+++ b/compat/cpu.h
@@ -36,7 +36,8 @@ inline int sched_getaffinity(int, size_t, cpu_set_t *) { return ENOTSUP; }
 #endif
 
 #if !defined(CPU_SET)
-inline void CPU_SET(int, cpu_set_t *) {}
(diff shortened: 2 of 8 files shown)


acinclude/os-deps.m4
```diff
@@ -968,3 +968,29 @@ AC_DEFUN([SQUID_CHECK_BROKEN_SOLARIS_IPFILTER],[
...
+
+static Map testMap;
+]], [[
+]])],
+    [squid_cv_std_unordered_map_allocator_works=yes],
+    [squid_cv_std_unordered_map_allocator_works=no])
+  )
+  if test "x$squid_cv_std_unordered_map_allocator_works" = "xyes" ; then
+    AC_DEFINE([STD_UNORDERED_MAP_ALLOCATOR_WORKS],1)
+  fi
```
yadij, line 995: Please use AS_IF() for new autoconf code.

```suggestion
  AS_IF([test "x$squid_cv_std_unordered_map_allocator_works" = "xyes"],[
    AC_DEFINE([STD_UNORDERED_MAP_ALLOCATOR_WORKS],1)
  ])
```

configure.ac
```diff
@@ -383,6 +383,13 @@ if test "x$ac_cv_require_wno_deprecated_register" = "xyes"; then
   SQUID_CXXFLAGS="$SQUID_CXXFLAGS -Wno-deprecated-register"
 fi
 
+dnl openbsd is super strict when dealing with system headers
+SQUID_CC_CHECK_ARGUMENT([ac_cv_support_wno_error_system_headers],
+                         [-Werror -Wno-system-headers])
+if test "x$ac_cv_support_wno_error_system_headers" != "xno"; then
+  SQUID_CXXFLAGS="$SQUID_CXXFLAGS -Wno-system-headers"
+fi
```
yadij, line 391:
```suggestion
AS_IF([test "x$ac_cv_support_wno_error_system_headers" != "xno"],[
  SQUID_CXXFLAGS="$SQUID_CXXFLAGS -Wno-system-headers"
])
```

yadij (changes requested):

kinkie:
openbsd 7.0 comes with autoconf 2.71 . I'm not sure if it's the updated version or an openbsd extension, but it fails with an error if it's invoked with an absolute path; it requires a relative path. The changes are about deferring path discovery and trying with a relative path after bootstrap to satisfy autoconf's requirement

yadij:
That means the problem is in test-builds.sh which should be calling the buildtest.sh with a relative path.

kinkie:
I actually think I know what was going on. It's this line in configure:
`
case $srcdir in
  *[\\\"\#\$\&\'\`$am_lf\ \     ]*)
    as_fn_error $? "unsafe srcdir value: '$srcdir'" "$LINENO" 5;;
esac
`

I can revert the change - it was a case of srcdir containing a $ path. But also, if the path is relative, this is not a problem because there are no dodgy characters in the path

acinclude/os-deps.m4
```diff
@@ -968,3 +968,29 @@ AC_DEFUN([SQUID_CHECK_BROKEN_SOLARIS_IPFILTER],[
...
+#include <memory>
+#include <unordered_map>
+template <class T>
+class SomeAllocator : public std::allocator<T>
+{};
+
+using Element = std::pair<int, uint64_t>;
+using Map = std::unordered_map<int, uint64_t, std::hash<int>, std::equal_to<int>, SomeAllocator<Element> >;
+
+static Map testMap;
```
rousskov, line 987: > PR description: OpenBSD's STL has a const-correctness bug in std::unordered_map when supplying an allocator.

AFAICT, the bug is in Squid code, not OpenBSD or STL. The proposed ./configure test (besides other problems) replicates the same Squid bug and, hence, fails with [many](https://godbolt.org/z/nqso1xqW3) [compilers](https://godbolt.org/z/TbrP4MfjG) outside of OpenBSD environment.

[std::unordered_map](https://en.cppreference.com/w/cpp/container/unordered_map), like other STL containers, requires that the container and the allocator have the same value_type. For std::unordered_map, value_type is `std::pair<const Key, T>`. We are providing an allocator with value_type `std::pair<Key, T>`. Compiler errors[^1] in my tests are actually pretty clear about that requirement -- "unordered container must have the same value_type as its allocator".

Please fix Squid code to use the right allocator value type in all containers. For example, HeaderValueCountsElement should use `std::pair<const SBuf, uint64_t>`. ClpMap::IndexItem should use `std::pair<const Key, EntriesIterator>`.

[^1]: The original error is missing from the PR description. Please add (its abridged version)!

I do not know why some compilers allow the mismatch in some environments, but that is not important.

rousskov (changes requested):
Thank you for improving OpenBSD support and Squid code quality.

acinclude/os-deps.m4
kinkie: Aha! got it; i had tried fiddling with const-ness in the element but I thought it would not need a separate type

```
using Element = std::pair<int, uint64_t>;
using AllocatorElement = std::pair<const int, uint64_t>;
using Map = std::unordered_map<int, uint64_t, std::hash<int>, std::equal_to<int>, SomeAllocator<AllocatorElement> >;
```
works; changing to that instead. Thanks for doing the analysis!

compat/cpu.h
```diff
@@ -36,15 +36,19 @@ inline int sched_getaffinity(int, size_t, cpu_set_t *) { return ENOTSUP; }
...
 #endif
 
 #if !defined(CPU_CLR)
 #define CPU_CLR(cpu, set) (void)0
 #endif
 
 #if !defined(CPU_ISSET)
-#define CPU_ISSET(cpu, set) false
+inline bool
+CPU_ISSET(int, const cpu_set_t *)
```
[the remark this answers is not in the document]
kinkie: All functions emulating macros in that file have the same bug then. Changing all signatures

kinkie:
Successful build on OpenBSD:
https://build.squid-cache.org/job/anybranch-openbsd-matrix/12/

Successful PR-test on Linux:
https://build.squid-cache.org/job/5-pr-test/3626/

compat/cpu.h
```diff
@@ -36,23 +36,27 @@ inline int sched_getaffinity(int, size_t, cpu_set_t *) { return ENOTSUP; }
...
+}
 #endif
 
 // glibc prior to 2.6 lacks CPU_COUNT
 #ifndef CPU_COUNT
 #define CPU_COUNT(set) CpuCount(set)
 /// CPU_COUNT replacement
 inline int
-CpuCount(const cpu_set_t *set)
+CpuCount(cpu_set_t *set)
```
rousskov, line 59: This is a different case: CpuCount() is not a standard macro. This function is our own invention. It should (continue to) correctly declare that the supplied set will not be modified, using `const`. Doing so will not clash with CPU_COUNT() because CPU_COUNT() is a macro.
```suggestion
CpuCount(const cpu_set_t *set)
```

compat/cpu.h
```diff
@@ -68,7 +72,7 @@ CpuCount(const cpu_set_t *set)
 #define CPU_AND(destset, srcset1, srcset2) CpuAnd((destset), (srcset1), (srcset2))
 /// CPU_AND replacement
 inline void
-CpuAnd(cpu_set_t *destset, const cpu_set_t *srcset1, const cpu_set_t *srcset2)
+CpuAnd(cpu_set_t *destset, cpu_set_t *srcset1, cpu_set_t *srcset2)
```
rousskov, line 75: Same here: CpuAnd() is ours.

```suggestion
CpuAnd(cpu_set_t *destset, const cpu_set_t *srcset1, const cpu_set_t *srcset2)
```

src/base/ClpMap.h
```diff
@@ -106,8 +106,9 @@ class ClpMap
     using EntriesIterator = typename Entries::iterator;
 
     using IndexItem = std::pair<Key, EntriesIterator>;
```
rousskov, line 108: This PR makes this type unused. Please either remove it or simplify by renaming IndexItemAllocatorElement back to IndexItem. I recommend the latter. Recall that IndexItem (i.e. map value_type) and IndexAllocatorItem (i.e. allocator value_type) must be the same so we can (continue to) simply use the shorter name for both purposes.

src/log/access_log.cc
```diff
@@ -58,8 +58,9 @@ static void mcast_encode(unsigned int *, size_t, const unsigned int *);
 #if USE_FORW_VIA_DB
 
 using HeaderValueCountsElement = std::pair<SBuf, uint64_t>;
+using HeaderValueCountsAllocatorElement = std::pair<const SBuf, uint64_t>;
```
rousskov, line 61: Same recommendation here:

```suggestion
using HeaderValueCountsElement = std::pair<const SBuf, uint64_t>;
```

... and revert the HeaderValueCounts change below.

rousskov (changes requested):
Thank you for moving the PR forward. Please add abridged error(s) to the PR description as requested in my earlier [review](https://github.com/squid-cache/squid/pull/929#discussion_r744348377). If needed, I can help summarize raw/detailed original messages if you share them.

compat/cpu.h
```diff
@@ -36,23 +36,27 @@ inline int sched_getaffinity(int, size_t, cpu_set_t *) { return ENOTSUP; }
 #endif
 
 #if !defined(CPU_SET)
-#define CPU_SET(cpu, set) (void)0
+inline void CPU_SET(int, cpu_set_t *) {}
```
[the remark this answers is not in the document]
kinkie: I would totally agree replacing a macro with a macro would 100% make sense if this was a library header that someone else needs to reuse, so that they can #undef them. In practice in our use case we don't really need to, but sure, it's a nit so I'll change that to use trampoline macros

src/base/ClpMap.h
kinkie: Right. Removing.

rousskov (commented):
It would be nice to briefly explain buildtest.sh fixes in the PR description. The diff alone does not explain what happened AFAICT.

kinkie:
done

rousskov:
Thank you. I reformatted to satisfy commit message requirements (and added missing periods to two sentences).

kinkie:
@yadij , are you happy and willing to unblock?

yadij (approved):
Approving. I still think we should have the fix in test-builds.sh instead of buildtest,sh, but not enough to block now.

PR #930, opened 2021-11-09
Bug 5090: Must(!request->pinnedConnection()) violation

rousskov:
The bug may be asymptomatic. Visible bug symptoms, if any, may include:

    FATAL: check failed: !request->pinnedConnection()
    exception location: FwdState.cc(1124) connectStart

    FATAL: check failed: transportWait
    exception location: FwdState.cc(675) noteDestinationsEnd

FwdState::usingDestination() should cover 3 post-transportWait periods:

1. peerWait: opening a CONNECT tunnel through the cache_peer
2. encryptionWait: TLS negotiations to secure the connection
3. Comm::IsConnOpen(serverConn): a Squid-peer transaction

The condition for the last period did not account for the time between
FwdState::unregister() and FwdState::complete() (or their equivalents),
when the transport connection is either closed or moved to the pconn
pool, but FwdState is still waiting for complete() and must not attempt
to connect to another destination. The bug is usually hidden because
complete() is usually called immediately after unregister(). However,
RESPMOD adaptation (at least) may delay complete(). If peer selection
news comes during that delay, usingDestination() lies, and various
Must()s may fail, depending on Squid version and HTTP request details.

Now, FwdState does not rely on the to-peer connection state for the
third period condition. Instead, we explicitly track whether the last
dispatch()ed activity has ended. This tracking is tricky because
complete() may be called without dispatching an asynchronous activity,
and because there are at least three ways for an activity to end:
explicit complete(), explicit handleUnregisteredServerEnd(), and
implicit serverClosed() connection closure callback. This tracking will
become easier and more reliable when FwdState no longer co-owns/stores
the to-peer connection. This change simplifies the future removal of
that connection ownership.

Also reordered usingDestination() conditions to match the chronological
order of the corresponding three periods.

Also reordered transportWait-vs-usingDestination() checks to match the
chronological order of those two forwarding stages.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -149,6 +149,7 @@ FwdState::FwdState(const Comm::ConnectionPointer &client, StoreEntry * e, HttpRe
     clientConn(client),
     start_t(squid_curtime),
     n_tries(0),
+    waitingForDispatched(false),
     destinations(new ResolvedPeers()),
     pconnRace(raceImpossible),
     storedWholeReply_(nullptr)
@@ -565,6 +566,9 @@ FwdState::complete()
 
     logReplyStatus(n_tries, replyStatus);
 
+    // will already be false if complete() was called before/without dispatch()
+    waitingForDispatched = false;
+
     if (reforward()) {
         debugs(17, 3, "re-forwarding " << replyStatus << " " << entry->url());
 
@@ -590,10 +594,13 @@ FwdState::complete()
     }
 }
 
+/// Whether a forwarding attempt to some selected destination X is in progress
+/// (after successfully opening/reusing a transport connection to X).
+/// See also: transportWait
 bool
-FwdState::usingDestination() const
+FwdState::transporting() const
 {
-    return encryptionWait || peerWait || Comm::IsConnOpen(serverConn);
+    return peerWait || encryptionWait || waitingForDispatched;
 }
 
 void
@@ -625,16 +632,15 @@ FwdState::noteDestination(Comm::ConnectionPointer path)
 
     destinations->addPath(path);
 
-    if (usingDestination()) {
-        // We are already using a previously opened connection, so we cannot be
-        // waiting for it. We still receive destinations for backup.
-        Must(!transportWait);
-        return;
-    }
-
     if (transportWait) {
+        assert(!transporting());
         notifyConnOpener();
         return; // and continue to wait for FwdState::noteConnection() callback
+    }
+
+    if (transporting()) {
+        Must(!transportWait);
+        return; // and continue to receive destinations for backup
     }
 
     // This is the first path candidate we have seen. Use it.
@@ -665,16 +671,13 @@ FwdState::noteDestinationsEnd(ErrorState *selectionError)
     // if all of them fail, forwarding as whole will fail
     Must(!selectionError); // finding at least one path means selection succeeded
 
-    if (usingDestination()) {
-        // We are already using a previously opened connection, so we cannot be
-        // waiting for it. We were receiving destinations for backup.
-        Must(!transportWait);
-        return;
+    if (transportWait) {
+        assert(!transporting());
+        notifyConnOpener();
+        return; // and continue to wait for FwdState::noteConnection() callback
     }
 
-    Must(transportWait); // or we would be stuck with nothing to do or wait for
-    notifyConnOpener();
-    // and continue to wait for FwdState::noteConnection() callback
+    Must(transporting()); // or we would be stuck with nothing to do or wait for
 }
 
 /// makes sure connection opener knows that the destinations have changed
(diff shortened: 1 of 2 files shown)


rousskov:
@eduard-bagdasaryan, please check whether TunnelStateData suffers from the same underlying flaw. I hope it does not, but I did not check carefully.

We may still want to reorder TunnelStateData conditions and `if` statements just to keep them in sync with FwdState code, but I think we should make that decision with your research results in mind.

yadij (approved):

src/FwdState.cc
```diff
@@ -565,6 +566,9 @@ FwdState::complete()
 
     logReplyStatus(n_tries, replyStatus);
 
+    // will already be false if complete() was called before/without dispatch()
+    waitingForDispatched = false;
```
eduard-bagdasaryan, line 570: I noticed that we reset `waitingForDispatched` before resetting `serverConn` and `destinationReceipt` in two places. Probably we could do likewise in `complete()`, moving the 570 line under the first 'if' because `waitingForDispatched` should not matter in the 'else' terminating with stopAndDestroy(). If so, should not we create a single method for this cleanup?
Also, since dispatch() is always called before syncWithServerConn() (where `serverConn` is assigned), should not we also move there `waitingForDispatched` setting? In other words, if the point since then the destination is considered to be 'used' is after FwdState becomes the owner of `serverConn`, we should probably assign the flag there (and adjust its name, if needed).

rousskov: Thank you for a useful review. I started responding yesterday, but could not finish at that time. I should have removed the M-cleared-for-merge label! We will do a followup PR to reflect your ideas.

eduard-bagdasaryan:
> check whether TunnelStateData suffers from the same underlying flaw

I don't think that TunnelStateData has the same flaw.

The bug affects the '3rd period' noted above ('a Squid-peer transaction') when FwdState tries to start processing a new destination while the old transaction has not been fully finished (i.e., the server connection was closed but complete()/reforward() not called yet). On one hand, TunnelStateData should not suffer these problems because it does not have any equivalents of FwdState's complete()/reforward(). On the other hand, as soon as the 3rd period is started, TunnelStateData::retriable becomes false, and no retryOrBail() (leading to TunnelStateData::startConnecting()) is possible.

rousskov:
> I don't think that TunnelStateData has the same flaw.

Thank you for checking.

src/FwdState.cc
rousskov: > Eduard: I noticed that we reset `waitingForDispatched` before resetting `serverConn` and `destinationReceipt` in two places. Probably we could do likewise in `complete()`, moving the 570 line under the first 'if' because `waitingForDispatched` should not matter in the 'else' terminating with stopAndDestroy(). If so, should not we create a single method for this cleanup?

I think the answer is "yes, with minor modifications[^1]" for the current code, but I think we should delay this cleanup. If we create a single method for the connection/waiting cleanup now, we will go a step back towards tying connection ownership (serverConn) and waitingForDispatched state. That misleading association is what led to the bug this PR fixed.

[^1]: If we adjust complete() along these lines, we should clear both waitingForDispatched and serverConn unconditionally, not inside the first `if` branch. We stop waiting for the dispatched transaction in complete() unconditionally _and_ should get rid of any connection leftovers unconditionally.

> Eduard: Also, since dispatch() is always called before syncWithServerConn() (where `serverConn` is assigned), should not we also move there `waitingForDispatched` setting? In other words, if the point since then the destination is considered to be 'used' is after FwdState becomes the owner of `serverConn`, we should probably assign the flag there (and adjust its name, if needed).

Technically, you are right -- the code always syncs the server connection right before dispatch() -- but that syncing alone does not really place FwdState in a true waiting state. In fact, we currently set waitingForDispatched too _early_ -- our dispatch() method may not actually enter the waiting state either (e.g., in `request->flags.sslPeek` case and in the `default` getScheme() case)! The code "works", AFAICT, but it is misleading/messy. Another problem to fix.

> Alex: We will do a followup PR to reflect your ideas.

I think all this cleanup should happen when or after we change FwdState to stop owning the to-server connection on the way to dispatch(). The connection should probably become the dispatch() parameter, but more (and more complex!) changes are required to fully get rid of the serverConn member and associated problems.

Thank you again for your review -- it exposed more/new problems in the current code and will help us clean it up better. It may take some time for us to clean this up properly, more than I originally expected, but we are on the right path AFAICT.

rousskov:
> Eduard: as soon as the 3rd period is started, TunnelStateData::retriable becomes false, and no retryOrBail() is possible.

You are right, but receiving spare destination paths was still possible during that final 3rd period, which led to the same transportWait violations this PR fixed for FwdState. Please review #985 that is supposed to fix those violations for TunnelStateData.

PR #932, opened 2021-11-11
Remove SBufStream

yadij:
Convert SBuf to provide the Packable API and replace all
SBufStream with PackableStream.

Fix several config dumper streams to output straight to the
StoreEntry instead of a temporary SBuf.


src/HappyConnOpener.cc
```diff
@@ -435,9 +434,9 @@ HappyConnOpener::status() const
     static SBuf buf;
     buf.clear();
 
-    SBufStream os(buf);
+    PackableStream os(buf);
 
-    os.write(" [", 2);
+    os << " [";
```
rousskov, line 439: The PR contains many small changes that go beyond the declared scope. Some of them seem wrong to me. I will go through them one by one eventually, but it will take time.

src/proxyp/Header.cc
```diff
@@ -24,66 +23,69 @@ ProxyProtocol::Header::Header(const SBuf &ver, const Two::Command cmd):
...
+    static char ipBuf[MAX_IPSTRLEN];
+    SBuf result;
+    PackableStream os(result);
+
     switch (headerType) {
 
     case Two::htPseudoVersion:
-        return version_;
+        os << version_;
+        break;
```
rousskov, line 50: Why was this method rewritten so much? It did not use SBufStream except in one special `switch` clause AFAICT. I see problems with some of the changes here, but I wonder if we can avoid evaluating them one by one by undoing virtually all of (i.e. minimizing) these changes instead?

rousskov (changes requested):
> PR description: Convert SBuf to provide the Packable API and replace all SBufStream with PackableStream.

Please document (in the PR description) why we want to add a virtual table (and, hence, a virtual destructor) to the performance-sensitive SBuf class. The change also makes SBuf::append() length type worse by switching back to `int`. There may be other non-disclosed regressions/problems that I have missed. What motivates this change?


> PR description: Fix several config dumper streams to output straight to the StoreEntry instead of a temporary SBuf.

Just to clarify, by "fix" here, do you mean the elimination of temporary SBuf storage on the way from the "config dumper" to the StoreEntry? In other words, this part of the PR does not alter the final dumping _result_, but attempts to optimize dumping, right?

src/HappyConnOpener.cc
yadij: Is this one of them?  If so why the unexpected unique use of write() to send two bytes to a stream?

src/proxyp/Header.cc
yadij: It is the end result of a long series of small code simplification refactors. There are several similar situations where things could improve further with the caller passing a stream in, but I have tried to scope this PR as only altering the SBuf API. Changing those others left to later discussions/PR.

yadij:
> > PR description: Convert SBuf to provide the Packable API and replace all SBufStream with PackableStream.
> 
> Please document (in the PR description) why we want to add a virtual table (and, hence, a virtual destructor) to the performance-sensitive SBuf class. The change also makes SBuf::append() length type worse by switching back to `int`. There may be other non-disclosed regressions/problems that I have missed. What motivates this change?
> 

Er, "this is C++ code" does not seem to be a self-evident situation not worth documenting.


> > PR description: Fix several config dumper streams to output straight to the StoreEntry instead of a temporary SBuf.
> 
> Just to clarify, by "fix" here, do you mean the elimination of temporary SBuf storage on the way from the "config dumper" to the StoreEntry? In other words, this part of the PR does not alter the final dumping _result_, but attempts to optimize dumping, right?

Correct.

rousskov:
> > > PR description: Convert SBuf to provide the Packable API and replace all SBufStream with PackableStream.

> > Alex: Please document (in the PR description) why we want to add a virtual table (and, hence, a virtual destructor) to the performance-sensitive SBuf class. The change also makes SBuf::append() length type worse by switching back to `int`. There may be other non-disclosed regressions/problems that I have missed. What motivates this change?
 
> Amos: Er, "this is C++ code" does not seem to be a self-evident situation not worth documenting.

Indeed; and that fact is not relevant here. I am asking "why we want this", not "why we can do this".

Making performance-sensitive classes more expensive (and any class/API more complex/worse) is not a C++ requirement. It is our decision. We need to justify our decisions, especially those involving non-obvious tradeoffs. This PR is a case where such justification (and consensus on its merits) is required and is currently missing.

src/HappyConnOpener.cc
rousskov: > why the unexpected unique use of write() to send two bytes to a stream

It may eliminate a strlen() call. There are many similar optimizations in Squid. We should be hiding (and securing) them by introducing length-aware constants, and we often do, but this and other status() code needs to be rewritten for other reasons anyway, so it just mimics what several other similar status() methods do.

----

Please note that, ideally, we should not be discussing why we should not be doing out-of-scope changes that require a discussion.

src/proxyp/Header.cc
rousskov: > It is the end result of a long series of small code simplification refactors

Please undo. These out-of-scope changes do not simplify the code in most cases and make it worse in some cases.

yadij:
> > > > PR description: Convert SBuf to provide the Packable API and replace all SBufStream with PackableStream.
> 
> > > Alex: Please document (in the PR description) why we want to add a virtual table (and, hence, a virtual destructor) to the performance-sensitive SBuf class. The change also makes SBuf::append() length type worse by switching back to `int`. There may be other non-disclosed regressions/problems that I have missed. What motivates this change?
> 
> > Amos: Er, "this is C++ code" does not seem to be a self-evident situation not worth documenting.
> 
> Indeed; and that fact is not relevant here. I am asking "why we want this", not "why we can do this".
> 

The "why" is that it removes code duplication.

> Making performance-sensitive classes more expensive (and any class/API more complex/worse) is not a C++ requirement. It is our decision. We need to justify our decisions involving non-obvious tradeoffs. This PR is a case where such justification (and consensus on its merits) is required and is currently missing.

What are you seeing as a cost here from simply having a vtable?
AFAICT it makes SBuf aligned to optimize vtable calls which in turn reduces CPU cache load cycles (few enough not to be measurable or at least balancing any added by virtual call calculations).

src/proxyp/Header.cc
yadij: Undone.

However FTR, all cases of this switch statement except 'htPseudoVersion' perform SBuf allocation to produce the returned object. All the ToSBuf() calls on return statements hide inefficient uses of PackableStream to serialize their parameters into another SBuf.

src/HappyConnOpener.cc
yadij: Undone. Though your logic can be applied to reverse many optimizations in Squid code on grounds that other code does things inefficiently.  ie. leaving this write() call here is actively preserving the collection of technical debt in Squid.

src/proxyp/Header.cc
rousskov: Thank you for undoing these changes.

> all cases of this switch statement except 'htPseudoVersion' perform SBuf allocation

Translation: The PR was adding one more allocation (without disclosure/justification).

> All the ToSBuf() calls on return statements hide inefficient uses of PackableStream to serialize their parameters into another SBuf.

Could you clarify this claim, please? ToSBuf() explicitly tells the code reader that a conversion to SBuf is going on. Where do you see the official method code _hiding_ its inefficiencies?

src/HappyConnOpener.cc
rousskov: > Amos: your logic can be applied to reverse many optimizations in Squid code on grounds that other code does things inefficiently. ie. leaving this write() call here is actively preserving the collection of technical debt in Squid.

I do not follow. Leaving the official write() call "as is" _preserves_ the existing (alleged) optimization. The PR was _removing_ that optimization. Please clarify how leaving this write() "as is" would make anything _worse_.

rousskov (changes requested):
> Alex: Please document (in the PR description) why we want to add a virtual table (and, hence, a virtual destructor) to the performance-sensitive SBuf class. The change also makes SBuf::append() length type worse by switching back to `int`. There may be other non-disclosed regressions/problems that I have missed. What motivates this change?
 
> Amos: The "why" is that it removes code duplication.

Thank you for disclosing what this PR is trying to accomplish. Please add that information to the PR description.

Removing code duplication is a very good goal, of course. I assume that you are talking about SBufStream* classes duplicating PackableStream* classes. I hope we can find a way to remove most of that code duplication without virtualizing SBuf.

@kinkie, do you recall why SBufStream duplicated PackableStream since inception (commit 412da42)? Have you tried avoiding that code duplication but had to duplicate for some specific reason?

@yadij, how about making PackableStreamBuf::buf _type_ a PackableStreamBuf parameter so that PackableStreamBuf can use a creator-supplied SBuf (instead of creator-supplied Packable) for storage?

 
> What are you seeing as a cost here from simply having a vtable?

Virtualizing a C++ class adds

* memory usage overhead: each class instance has to store the pointer to the table;
* performance overhead: virtual function calls (including a destructor) usually incur virtual dispatch/lookup, preventing inlining and increasing CPU cache miss ratio;
* complexity: a class with a base class and a class with virtual functions are more complex than a class without those things

The exact extra cost heavily depends on context, of course. In some cases, it is negligible. In others, it can be noticeable to an outside observer. Since SBuf is a low-level performance-sensitive class that incurs numerous object copies, we can safely assume that some of its current or future use cases will pay significant virtualization costs.

> AFAICT it makes SBuf aligned to optimize vtable calls

I do not know what you mean by "make [class] aligned" and "vtable calls", but if we can improve performance by adjusting SBuf data member alignment, then we can do that _without_ paying the virtualization price.

kinkie:
> @kinkie, do you recall why SBufStream duplicated PackableStream since inception (commit [412da42](https://github.com/squid-cache/squid/commit/412da4279f75d9fa63cf58a7d2eac63e61d35d78))? Have you tried avoiding that code duplication but had to duplicate for some specific reason?

No; part of that is ignorance, I think. But there might be something about intending to optimize by appending lazily as a list of SBufs and then concatenating only once when the Stream is read.
Whether the potential optimization is enough to justify the duplication or, even if it's an optimization at all is up for debate

rousskov:
> Francesco: there might be something about intending to optimize by appending lazily as a list of SBufs and then concatenating only once when the Stream is read. Whether the potential optimization is enough to justify the duplication or, even if it's an optimization at all is up for debate

Thanks for walking down the memory lane for me! Postponing concatenation to learn the total required length size might improve performance, but I am pretty sure that (future hypothetical optimization) can be accomplished without duplicating PackableStream. Thus, there seems to be no (known) showstoppers for removing this code duplication. We just need to remove it without making other things worse, but that is the subject of another [question](https://github.com/squid-cache/squid/pull/932#pullrequestreview-806331455) (for Amos).

src/HappyConnOpener.cc
yadij: C++11 assigns fixed-strings like this a special ```&char[N]``` type which can be handled by streambuf::xsputn(char*, int N) when the compiler is building optimized code. Note that the stdlib operator<<(char*) documentation says N is calculated "_as if_ strlen(s) were used" - not that strlen(s) is actually happening at run-time.

yadij:
> 
> @yadij, how about making PackableStreamBuf::buf _type_ a PackableStreamBuf parameter so that PackableStreamBuf can use a creator-supplied SBuf (instead of creator-supplied Packable) for storage?
> 
> > What are you seeing as a cost here from simply having a vtable?
> 
> Virtualizing a C++ class adds
> 
>     * memory usage overhead: each class instance has to store the pointer to the table;
> 

This is not true for classes with single-line inheritance. For that type of class the vtable offset is purely a compile-time piece of information. AIUI the code produced by modern compilers should inline the offset as a fixed value added when the jump register is added.  So "stored" in the assembler/binary code, not the object memory. That is why we gain data alignment and a change to 32-byte size (minimal CPU cache line size).

>     * performance overhead: virtual function calls (including a destructor) usually incur virtual dispatch/lookup, preventing inlining and increasing CPU cache miss ratio;
> 

These separate despatch only happens for multi-line inheritence where a vtable physically exists and trampolines are used to hold the type vs vtable segment differences.

There is a small extra calculation for that jump register management mentioned above. But this is absorbed on most modern CPUs by the parallel instruction paths so is not a measurable effect.

FTR, I have checked the SBuf object size before (16 bytes) and after (32 bytes) adding the inheritance. This change is insufficient to account for 5x pointer offsets needed by a vtable, but does exactly match exactly the minimum CPU cache line size for the architecture I tested on.


>     * complexity: a class with a base class and a class with virtual functions are more complex than a class without those things
> 

This conceptual issue is balanced out by the reduction in code duplication (and associated mistakes, bugs, etc) and variance over time. Also by our audience of developers being at least aware of how to understand C++ inheritence.

> 
> The exact extra cost heavily depends on context, of course. In some cases, it is negligible. In others, it can be noticeable to an outside observer. Since SBuf is a low-level performance-sensitive class that incurs numerous object copies, we can safely assume that some of its current or future use cases will pay significant virtualization costs.
> 
> > AFAICT it makes SBuf aligned to optimize vtable calls
> 
> I do not know what you mean by "make [class] aligned" and "vtable calls", but if we can improve performance by adjusting SBuf data member alignment, then we can do that _without_ paying the virtualization price.

The alignment related to virtuals is alignment for CPU cache line - smallest of which is 32-*bytes*. Aligning these AFAIK has a potential for large improvement on CPU operations. Though, I have not noticed any difference in speed on my test build.

yadij:
> @yadij, how about making PackableStreamBuf::buf type a PackableStreamBuf parameter so that PackableStreamBuf can use a creator-supplied SBuf (instead of creator-supplied Packable) for storage?

There are multiple objects being used as stores for the stream buffer (StoreEntry, MemBuf, now SBuf). They all have to share a uniform set of identical methods for the stream to call regardless of whether the streambuf itself is templated or not - "Packable" is that API definition.

Yes there are still some things we could to do optimize the Packable and its related stream pieces. But optimizing those is not in scope here. Making SBuf present the current Packable API so we can drop the extra code is the scope.

rousskov:
> Alex: Virtualizing a C++ class adds memory usage overhead: each class instance has to store the pointer to the table;
 
> Amos: This is not true for classes with single-line inheritance.

What I said _is_ true. The costs I am talking about are real and apply to all inheritance patterns. They cannot be optimized away in the contexts relevant to this PR. The need to store a pointer to the vtable regardless of the inheritance pattern is basic C++ knowledge. We should not be arguing about it, but if you insist on doing so, please provide a specific example to disprove my assertion. FWIW, I have wasted time to build a basic [example](https://godbolt.org/z/a65Yf35Yo) that demonstrates the increase in memory footprint (and call costs) due to virtualization.

> Alex: performance overhead: virtual function calls (including a destructor) usually incur virtual dispatch/lookup, preventing inlining and increasing CPU cache miss ratio;
 
> Amos: These separate despatch only happens for multi-line inheritence

Again, what I said is true: The dispatch happens for all inheritance types in all contexts relevant to this discussion. This is basic C++ knowledge we should not be arguing about, but the same basic [example](https://godbolt.org/z/a65Yf35Yo) demonstrates these costs. One can clearly see that optimized testPlainA() compilation results in one trivial assembly instruction -- a jump to the PlainA::bar() function:


```assembly
testPlainA(PlainA*):
        jmp     PlainA::bar()
```

... while testVirtualizedA() compilation results in 10+ instructions with expensive memory accesses to extract vtable and its cells (and other overheads):


```assembly
testVirtualA(VirtualA*):
        mov     rax, QWORD PTR [rdi]
        mov     rdx, QWORD PTR [rax+16]
        cmp     rdx, OFFSET FLAT:VirtualA::foo()
        jne     .L13
        jmp     [QWORD PTR [rax+24]]
.L13:
        sub     rsp, 24
        mov     QWORD PTR [rsp+8], rdi
        call    rdx
        mov     rdi, QWORD PTR [rsp+8]
        mov     rax, QWORD PTR [rdi]
        mov     rax, QWORD PTR [rax+24]
        add     rsp, 24
        jmp     rax
```

The exact results depend on whether known foo() and bar() implementations are inlined -- the optimizing compiler works hard to avoid function calls where it can -- but the essential differences relevant here remain across those variations.
 

> Amos: FTR, I have checked the SBuf object size before (16 bytes) and after (32 bytes) adding the inheritance. This change is insufficient to account for 5x pointer offsets needed by a vtable

I am sorry, but C++ virtualization just does not work like the above FTR statement implies. Please note that I described per-object memory cost (of storing the _pointer_ to vtable). I did _not_ talk about the cost of storing the vtable because that tiny per-type memory cost is immaterial here.

As for the "cache line" part of the discussion, I have serious concerns about some of the related assertions, but would very much prefer to avoid discussing low-level CPU mechanisms in this PR. Additional performance costs of virtual calls in a general-purpose string-pointer object is not something that can be (reliably) compensated by increasing that object size to become closer to typical cache line size.


> Alex: complexity: a class with a base class and a class with virtual functions are more complex than a class without those things

> Amos: This conceptual issue is balanced out by the reduction in code duplication

You might be comparing official (duplicated) code and PR code. I am comparing PR code and an alternative duplication removal solution (that does not use virtualization). In my comparison, there is no duplication in _both_ solutions, and no virtualization complexities in the alternative solution.

In summary, all my statements were based on basic C++ facts. No arguments can disperse the underlying truth -- virtualization adds overheads, on several levels. I just hope we will not have to repeat this discussion again!

----

> Alex: how about making PackableStreamBuf::buf type a PackableStreamBuf parameter so that PackableStreamBuf can use a creator-supplied SBuf (instead of creator-supplied Packable) for storage?

> Amos: There are multiple objects being used as stores for the stream buffer (StoreEntry, MemBuf, now SBuf). 

Yes, PackableStreamBuf should (continue to) be able to accommodate buffers of various types.

> They all have to share a uniform set of identical methods for the stream to call 

The "identical" part of the assertion is false (at least in the scope of this duplication-reducing PR): The method profiles may differ in certain ways. Some differences can be accommodated without caller (i.e. PackableStreamBuf) knowledge (e.g., some methods may be virtual while others with the same name are not). Some differences may require a simple specialization (e.g., different method names or parameter order can often be easily accommodated). In summary, we cannot reject the idea of specifying/templating the type of PackableStreamBuf buffer based on the above false assertion.


> Making SBuf present the current Packable API so we can drop the extra code is the scope.

If there are better duplication-avoiding solutions (that do not require adding a Packable parent to the SBuf class), then we should use one of those solutions instead.

@yadij, do you want to try to implement the idea of specifying the PackableStreamBuf buffer type yourself? If not, I can either find somebody else to do it or do it myself. We can then compare the end result and pick the best solution. I believe it will be the alternative solution (for the reasons already stated), and the basic mistakes in counter-arguments only increase my confidence that an inferior solution was chosen in this PR due to misunderstanding of relevant C++ concepts.

src/HappyConnOpener.cc
rousskov: You are right that an optimizing compiler may use the knowledge of the string literal length to avoid a strlen() call. This fact significantly weakens my simple "eliminates strlen()" example -- strlen() is likely to be avoided in both cases by a good optimizing compiler.

However, there is another (more complex) reason why the official code is an optimization (that cannot be performed by a compiler): The official code uses unformatted output. The PR-attempted operator `<<` replacement uses formatted output. An optimizing compiler cannot (and, in my tests, does not) replace one with the other because that would alter output in cases that a compiler cannot detect. Formatted output code naturally does more before (and after) calling streambuf::sputn() -- the unformatted output function used by formatted and unformatted code paths. Thus, the official code remains an optimization even in compiler-optimized builds.

We can, of course, discuss the merits and drawbacks of this optimization and decide that it should be altered or removed, but all of that is outside this PR scope. This PR is not the right place for such discussions and code changes.

----

Thank you for removing this out of scope change. I am resolving this change request even though I have not checked whether other out-of-scope changes this change request complains about have been removed. If they were not, I will deal with them separately, after we resolve the primary problem with this PR -- SBuf virtualization.

yadij:
> @yadij, do you want to try to implement the idea of specifying the PackableStreamBuf buffer type yourself? If not, I can either find somebody else to do it or do it myself. We can then compare the end result and pick the best solution. I believe it will be the alternative solution (for the reasons already stated), and the basic mistakes in counter-arguments only increase my confidence that an inferior solution was chosen in this PR due to misunderstanding of relevant C++ concepts.

Since I am not understanding how that will resolve the issue you have and it is out of scope for this PR I will leave that to you and wait on furthering this PR until we can compare.

rousskov:
> Since I am not understanding how that will resolve the issue you have 

It will avoid virtualizing SBuf, which is currently the primary showstopper for this PR.

> and it is out of scope for this PR

It is in scope because it will address the same problem this PR is trying to address -- excessive code duplication between PackableStreamBuf and SBufStreamBuf.

> I will leave that to you and wait on furthering this PR until we can compare.

Sounds good to me.

rousskov:
> Alex: do you want to try to implement the idea of specifying the PackableStreamBuf buffer type yourself?
> Amos: I will leave that to you and wait on furthering this PR until we can compare.

Done at #959. I suggest switching to that PR instead of fixing this one because #959 provides a much better solution to the same primary problem.

yadij:
FTR,  PR #959 altered SBufStream internals (removes SBufStreamBuf) without resolving the issue that there are two non-compatible FooStream types for callers to select from. That incompatibility requires full-length memory copy to be performed when translating from temporary SBuf stream storage to StoreEntry contents for traffic delivery or caching.

Abandoning this performance improvement. Anyone working on this in future should start fresh.

PR #933, opened 2021-11-13
Remove recv-announce tool

yadij:
This daemon for receiving send-announce traffic has not
built for many years.


rousskov (approved):
Thank you for removing this dead tool.

We should probably remove the corresponding announcement code from Squid as well, for different reasons, but that can be viewed as out of this PR scope.

yadij:
FYI, I have some ideas for replacing the send part with a log output. But that is out of scope. This tool is long dead.

PR #934, opened 2021-11-13
Remove tcp-banger

yadij:
The test-suite tcp-banger 2 and 3 binaries have not built since Squid-2
and unlikely to be upgraded. There are better ways to bang at Squid
ports than the old Perl script that we do not want to maintain.


rousskov (dismissed):
Thank you for removing these. Less dead, unused code is better, and there are other tools to test TCP port scalability.

We can also remove the corresponding Perl script if you wish, for similar reasons, but I do not insist on that.

rousskov:
Since the Perl script remains, the PR title should be adjusted to avoid the implication that all TCP banging tools were removed.

rousskov (approved):
LGTM. I have updated the PR description to avoid potentially misleading (and unnecessary) details in the description of the Perl script and to avoid recommending a specific test tool, especially the one that has not been designed to test proxies and can easily lead to misleading results. Please adjust further if needed.

yadij:
@rousskov the perl script still works and distributed by default. Thus it may still be in use somewhere. To remove it without trouble we need to either provide an alternative or have release cycles where it is deprecated (and disabled by default).

apachebench is the common tool used for stress testing any HTTP server agent, including proxies.

rousskov:
> the perl script still works and distributed by default. Thus it may still be in use somewhere. 

I agree that the removed script may be in use somewhere. My PR description edits (including their justification) do not reject that possibility.

> To remove it without trouble we need to either provide an alternative or have release cycles where it is deprecated (and disabled by default).

I disagree: Many tools are readily available, including the removed script itself. **We** do not need to provide any [in master/v6].

> apachebench is the common tool used for stress testing any HTTP server agent, including proxies.

My PR description edits (including their justification) were done with that fact in mind. They do not contradict it.

yadij:
Then this PR approach is vetoed by your unwillingness to let me inform admin about alternatives to the tools removed.

rousskov:
> Then this PR approach is vetoed by your unwillingness to let me inform admin about alternatives to the tools removed.

I do not understand why you decided to close this PR. I did not veto this PR approach, whatever that is. I am not against informing admins about alternatives. While I think that we should not advertise a single imperfect alternative in this case, I am sure we could have found a mutually acceptable way to solve that small (description) problem and merge the PR.

PR #941, opened 2021-12-02
Remove unused/broken ACL copying support

yadij:
This code is unused. The clone() methods were the only use of
copy construction and operator. Most ACL related classes lack
implementation of the copy and/or assert so even if used this
code would be quite dangerous.

Explicitly forbid copy/move at the hierarchy base classes
ACL and ACLData. Removing all child specific copy
implementations and prohibitions (now unnecessary).


yadij:
Thanks for the feedback. Done the first two - I missed ACLStrategised the commit for that should push in a few hours when my local re-test completes.

rousskov (approved):

PR #943, opened 2021-12-03
Remove NeXTSTEP support

yadij:
NeXT architecture hardware is no longer available and with that
hardware any support by Apple ceased.

QNX and MacOS descendant systems which still depend on some of
the workarounds already are or can be supported under their own
compatibility section.


kinkie (approved):
good to go

rousskov (commented):
Just like with OSF/1, it is not clear _why_ we are removing this code. Are we implying that any environment-specific Squid code for environments no longer officially supported by their shepherds should be removed from Squid?

yadij:
@rousskov as mentioned in the description it is no longer possible to even obtain hardware this OS and its portability logic are relevant for. AFAICT The hack has no relevance to any of the Squid community.

rousskov:
> AFAICT The hack has no relevance to any of the Squid community.

I suspect you are right, and I am not questioning your opinion regarding this code relevance (or hardware availability). Instead, I am trying to establish criteria or principles behind this and similar code removals. Perhaps we can agree on something like "no known (to PR author and reviewers) current use within the Squid community" as a starting point? This condition does not cover all cases, but it can be a criteria sufficient for justifying many removals.

yadij:
> > AFAICT The hack has no relevance to any of the Squid community.
> 
> I suspect you are right, and I am not questioning your opinion regarding this code relevance (or hardware availability). Instead, I am trying to establish criteria or principles behind this and similar code removals. Perhaps we can agree on something like "no known (to PR author and reviewers) current use within the Squid community" as a starting point? This condition does not cover all cases, but it can be a criteria sufficient for justifying many removals.

There is a small and shrinking set of these cases. Simply discussing new special-case procedures has already taking longer than the existing removal process would have spent to completely eliminate the cases.

rousskov:
> Simply discussing new special-case procedures has already taking longer than the existing removal process would have spent to completely eliminate the cases.

I doubt the above is true, but the danger of that becoming true explain why Francesco's [proposal](http://lists.squid-cache.org/pipermail/squid-dev/2021-December/009730.html) is devoted to a much broader topic of official OS support, and why my [proposal](http://lists.squid-cache.org/pipermail/squid-dev/2021-December/009735.html) for the _code removal_ procedures is not limited to these few special cases. We are trying to create something reusable long-term instead of repeating the same painful steps for each individual PR related to "OS support" or "code removal".

PR #944, opened 2021-12-03
Remove m88k-specific support

yadij:
Modern support for m88k hardware is only available through
OpenBSD and NetBSD which have their own compatibility handling.
We no longer need to detect m88k ourselves.


kinkie (approved):
good to go

rousskov (commented):
Just like with OSF/1, I recommend changing the title to avoid the strong implication that we are removing support for m88k. The PR description implies that we continue to support m88k environments, but are supporting them without m88k-specific code. Suggestion: "Remove m88k-specific code" or "Remove m88k-specific ./configure checks", but I do not insist on any specific wording.

PR #946, opened 2021-12-08
Categorize level-0/1 messages; part 1: "easy" problem messages

rousskov:
Today, it is impractical to reliably identify important cache.log
messages in automated sysadmin notification systems. No regular
expression is guaranteed to match all important messages and none of the
unimportant ones. While message importance is naturally subjective (with
cache_log_message and notification system regexes available to customize
handling of specific messages), even reliable detection of _potentially_
important messages (to bootstrap admin customization efforts) is
impractical today. Any notification configuration is likely to miss an
important message (in today's or future code) and/or spam admins with
unimportant messages.

Fully addressing this problem requires many changes. This change lays a
foundation for that work by standardizing three well-known message
prefixes (`FATAL:`, `ERROR:`, and `WARNING:`) for debugs() messages
reporting various problems. Eventually, all problem-reporting messages
(and only those) will use one of the standard prefixes. This change
converts 300+ debugs(), bringing the total to 900+ messages using a
a standard prefix. There are more than 5'000 debugs() in Squid.

The scope of this change are problem-reporting debugs() statements that
can be detected and converted by a simple script. We also applied a few
easy-to-automate polishing touches to modified debugs(), like replacing
debug level 0 with DBG_CRITICAL, expanding some contractions (e.g.,
"can't"), and fixing grammar (e.g., "is has").

### DIFF
--- a/src/ConfigParser.cc
+++ b/src/ConfigParser.cc
@@ -486,7 +486,7 @@ ConfigParser::NextKvPair(char * &key, char * &value)
     if (!key)
         return false;
     if (!value) {
-        debugs(3, DBG_CRITICAL, "Error while parsing key=value token. Value missing after: " << key);
+        debugs(3, DBG_CRITICAL, "ERROR: Failure while parsing key=value token. Value missing after: " << key);
         return false;
     }
 
--- a/src/DiskIO/DiskThreads/aiops.cc
+++ b/src/DiskIO/DiskThreads/aiops.cc
@@ -512,7 +512,7 @@ squidaio_queue_request(squidaio_request_t * request)
         if (++filter >= filter_limit) {
             filter_limit += filter;
             filter = 0;
-            debugs(43, DBG_IMPORTANT, "squidaio_queue_request: WARNING - Queue congestion (growing to " << filter_limit << ")");
+            debugs(43, DBG_IMPORTANT, "WARNING: squidaio_queue_request: Queue congestion (growing to " << filter_limit << ")");
         }
     }
 
@@ -535,7 +535,7 @@ squidaio_queue_request(squidaio_request_t * request)
 
         if (squid_curtime >= (last_warn + 15) &&
                 squid_curtime >= (high_start + 5)) {
-            debugs(43, DBG_IMPORTANT, "squidaio_queue_request: WARNING - Disk I/O overloading");
+            debugs(43, DBG_IMPORTANT, "WARNING: squidaio_queue_request: Disk I/O overloading");
 
             if (squid_curtime >= (high_start + 15))
                 debugs(43, DBG_IMPORTANT, "squidaio_queue_request: Queue Length: current=" <<
--- a/src/DiskIO/IpcIo/IpcIoFile.cc
+++ b/src/DiskIO/IpcIo/IpcIoFile.cc
@@ -949,7 +949,7 @@ IpcIoFile::DiskerHandleRequest(const int workerId, IpcIoMsg &ipcIo)
     } catch (const Queue::Full &) {
         // The worker pop queue should not overflow because the worker can
         // push only if pendingRequests() is less than QueueCapacity.
-        debugs(47, DBG_IMPORTANT, "BUG: Worker I/O pop queue for " <<
+        debugs(47, DBG_IMPORTANT, "ERROR: Squid BUG: Worker I/O pop queue for " <<
                DbName << " overflow: " <<
                SipcIo(workerId, ipcIo, KidIdentifier)); // TODO: report queue len
 
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -215,7 +215,7 @@ HttpHeader::clean()
         if (e == nullptr)
             continue;
         if (!Http::any_valid_header(e->id)) {
-            debugs(55, DBG_CRITICAL, "BUG: invalid entry (" << e->id << "). Ignored.");
+            debugs(55, DBG_CRITICAL, "ERROR: Squid BUG: invalid entry (" << e->id << "). Ignored.");
         } else {
             if (owner <= hoReply)
                 HttpHeaderStats[owner].fieldTypeDistr.count(e->id);
@@ -1460,7 +1460,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
             return nullptr; // reject if we cannot strip
 
         debugs(55, Config.onoff.relaxed_header_parser <= 0 ? 1 : 2,
-               "NOTICE: Whitespace after header name in '" << getStringPrefix(field_start, field_end-field_start) << "'");
+               "WARNING: Whitespace after header name in '" << getStringPrefix(field_start, field_end-field_start) << "'");
 
         while (name_len > 0 && xisspace(field_start[name_len - 1]))
             --name_len;
--- a/src/acl/FilledChecklist.cc
+++ b/src/acl/FilledChecklist.cc
@@ -76,7 +76,7 @@ showDebugWarning(const char *msg)
         return;
 
     ++count;
-    debugs(28, Important(58), "ALE missing " << msg);
+    debugs(28, Important(58), "ERROR: ALE missing " << msg);
 }
 
 void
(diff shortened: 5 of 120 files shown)


rousskov:
I will not clear this PR for merging without at least one more review because I have seen too many converted messages while working on this PR to reliably detect conversion problems.

When reviewing this PR, please keep in mind that all changes here should be automated -- no manual adjustments! If the script misses a large category of in-scope messages that should be automatically converted, we will enhance the script. On the other hand, if you would like to polish a single old debugs() statement modified by this script, it is usually best to do it elsewhere. There is a large gray area between these two extremes, of course, but keep automation in mind when formulating your change requests.

src/acl/FilledChecklist.cc
```diff
@@ -76,7 +76,7 @@ showDebugWarning(const char *msg)
         return;
 
     ++count;
-    debugs(28, Important(58), "ALE missing " << msg);
+    debugs(28, Important(58), "ERROR: ALE missing " << msg);
```
yadij, line 79: This upgrade is a problem. This function is clearly intended to show *warnings*.

src/icp_v2.cc
```diff
@@ -601,7 +601,7 @@ icpHandleIcpV2(int fd, Ip::Address &from, char *buf, int len)
         break;
 
     default:
-        debugs(12, DBG_CRITICAL, "icpHandleIcpV2: UNKNOWN OPCODE: " << header.opcode << " from " << from);
+        debugs(12, DBG_CRITICAL, "ERROR: icpHandleIcpV2: UNKNOWN OPCODE: " << header.opcode << " from " << from);
```
yadij, line 604: can you lower case the "UNKNOWN OPCODE" text while touching these lines please.

src/log/ModDaemon.cc
```diff
@@ -338,7 +338,7 @@ logfile_mod_daemon_flush(Logfile * lf)
 {
     l_daemon_t *ll = static_cast<l_daemon_t *>(lf->data);
     if (commUnsetNonBlocking(ll->wfd)) {
-        debugs(50, DBG_IMPORTANT, "Logfile Daemon: Couldn't set the pipe blocking for flush! You're now missing some log entries.");
+        debugs(50, DBG_IMPORTANT, "ERROR: Logfile Daemon: Could not set the pipe blocking for flush! You're now missing some log entries.");
```
yadij, line 341: Since we are using formal language now:   s/You're/You are/   or even "There are" if possible.

yadij (approved):
Thank you for doing this.

I have marked a few things we should change or I would like clarification on. But I do not insist on any of them.

src/ConfigParser.cc
```diff
@@ -486,7 +486,7 @@ ConfigParser::NextKvPair(char * &key, char * &value)
     if (!key)
         return false;
     if (!value) {
-        debugs(3, DBG_CRITICAL, "Error while parsing key=value token. Value missing after: " << key);
+        debugs(3, DBG_CRITICAL, "ERROR: Failure while parsing key=value token. Value missing after: " << key);
```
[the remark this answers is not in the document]
rousskov: Left as is. `ERROR: Failure while fooing` is a bit redundant indeed, but it is not too bad, especially when there are two whole sentences describing that `ERROR`.

We cannot just drop `Failure`. The (required) colon at the end of the `ERROR:` label forces us to start a new sentence in this context, and that sentence cannot be "while parsing token.": `ERROR: while parsing token.` just does not look good.

The whole message needs to be refactored to improve this, but that is too much to ask of the script.

src/DiskIO/DiskThreads/aiops.cc
```diff
@@ -512,7 +512,7 @@ squidaio_queue_request(squidaio_request_t * request)
         if (++filter >= filter_limit) {
             filter_limit += filter;
             filter = 0;
-            debugs(43, DBG_IMPORTANT, "squidaio_queue_request: WARNING - Queue congestion (growing to " << filter_limit << ")");
+            debugs(43, DBG_IMPORTANT, "WARNING: squidaio_queue_request: Queue congestion (growing to " << filter_limit << ")");
```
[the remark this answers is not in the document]
rousskov: That would be really nice, but it requires a lot more smarts than the script can muster. Perhaps more importantly, that removal should be done in _all_ legacy debugs() statements. I suspect clang-tidy can be used to do that, but that requires a very serious out-of-scope effort.

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -949,7 +949,7 @@ IpcIoFile::DiskerHandleRequest(const int workerId, IpcIoMsg &ipcIo)
     } catch (const Queue::Full &) {
         // The worker pop queue should not overflow because the worker can
         // push only if pendingRequests() is less than QueueCapacity.
-        debugs(47, DBG_IMPORTANT, "BUG: Worker I/O pop queue for " <<
+        debugs(47, DBG_IMPORTANT, "ERROR: Squid BUG: Worker I/O pop queue for " <<
```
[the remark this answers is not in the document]
rousskov: Hm... I find these BUG labels useful and meaningful. If somebody sees an `ERROR` with `BUG`, they should report them to us. What do you mean by "mess" exactly? Perhaps we should document something better to reduce this mess?

The `ERROR` label conveys the "this is an error" meaning, but the `BUG` label tells the admin what _kind_ of error it is: They (including  their traffic) are not at fault -- their Squid has a bug that needs to be fixed by developers.

src/HttpHeader.cc
```diff
@@ -1460,7 +1460,7 @@ HttpHeaderEntry::parse(const char *field_start, const char *field_end, const htt
             return nullptr; // reject if we cannot strip
 
         debugs(55, Config.onoff.relaxed_header_parser <= 0 ? 1 : 2,
-               "NOTICE: Whitespace after header name in '" << getStringPrefix(field_start, field_end-field_start) << "'");
+               "WARNING: Whitespace after header name in '" << getStringPrefix(field_start, field_end-field_start) << "'");
```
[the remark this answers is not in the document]
rousskov: Sorry, I do not understand the question. There is no "upgrade" here. We are just reducing the number of standard labels to three (`WARNING`, `ERROR`, and `FATAL`) for debugs() messages reporting various problems. In this particular case, the problem is actually an error -- the header is malformed, but the PR script is not smart enough to know that, of course. Also, this change happens to match our own `relaxed_header_parser` documentation:

```
	If set to "warn" then a warning will be emitted in cache.log
	each time such HTTP error is encountered.
```

Please note that for all practical purposes this is a level-2 message so hopefully we do not need to invest a lot of time into it.

src/acl/FilledChecklist.cc
rousskov: There is a problem in this code indeed, but not in the added `ERROR` label. The problem is a bad function name! The reported problem is an error, and often pretty important one. The missing ALE field likely leads to incorrect Squid decisions (e.g., an `http_access` mismatch).

src/log/ModDaemon.cc
rousskov: Good catch. It is too risky for the dumb script to change "You" to "There", but it can expand contractions. Done in 892d596.

src/icp_v2.cc
rousskov: Done in ee36c93.

rousskov (commented):
Thank you for reviewing this PR. I was able to modify the script to address two of your (optional) change requests. I either disagree with or just cannot address (using a dumb script) the other five.

If I hear no objections, I will clear this PR for merging in a few days (to give you a chance to revisit the two changes and my responses).

rousskov:
@yadij, as you know, this PRs reduces Squid monitoring headaches and only changes cache.log messages. Backporting this to v5 will reduce future backporting overheads. On the other hand, somebody might be already deeply invested in the current v5 spelling of some level-0/1 messages. I am happy to backport this change to v5. Will you accept such a backport?

rousskov:
> rousskov on [Jan 27, 2022](https://github.com/squid-cache/squid/pull/946#issuecomment-1023460905): @yadij, as you know, this PRs reduces Squid monitoring headaches and only changes cache.log messages. Backporting this to v5 will reduce future backporting overheads. On the other hand, somebody might be already deeply invested in the current v5 spelling of some level-0/1 messages. I am happy to backport this change to v5. Will you accept such a backport?

@yadij, please make a decision regarding this backport acceptance (in principle). Thank you.

PR #949, opened 2021-12-13
Cleanup doc/ directory

yadij:
Remove unnecessary copies of publicly available RFC documents
and lists of standard information which is available from
external sources.

See https://wiki.squid-cache.org/StandardsCompliance for a
list of RFC documents applicable to Squid.

See https://www.iana.org/assignments/http-status-codes/ for the
official list of HTTP status codes and document references.

Draft documents are kept for now since they may become
unavailable. As several already have.


kinkie:
I'm supportive of the idea to remove rfc text files, would it make sense to still keep an index of the relevant ones? A shortened version of 1-index, maybe

doc/rfc/1-index.txt
```diff
@@ -17,201 +17,3 @@ draft-wilson-wccp-v2-12-oct-2001.txt
 
 draft-vinod-carp-v1-03.txt
 	Microsoft CARP peering algorithm
-
-proxy-protocol.txt
-	The PROXY protocol, Versions 1 & 2
```
rousskov, line 22: Please delete the stale copy of the PROXY protocol specs as well. If I am reading the current GitHub PR rendering correctly, the `doc/rfc/proxy-protocol.txt` file has not been removed.

doc/HTTP-codes.txt
```diff
@@ -1,50 +0,0 @@
...
-
-Notes:
-* HTTP 1.1
-c Cached unless a query response without expiry information
-C Cached
-E Negatively cached if no expiry headers
-t Cached only if expiry information
-- Not cached
-
-Unless other said, the response code is not cached.
```
rousskov, line 50: To (better) cover the removal of this and PROXY protocol files, I suggest simplifying PR description to say something like "Remove copies of publicly available documents and their derivatives".

rousskov (dismissed):
> would it make sense to still keep an index of the relevant ones?

IMO, no, it would not: The current index is already incomplete (at best). I see no good reason to keep a list that is almost impossible to keep in sync with ongoing code and standards changes, especially since the utility of (or gain from) having such a list is very low.

yadij:
> I'm supportive of the idea to remove rfc text files, would it make sense to still keep an index of the relevant ones? A shortened version of 1-index, maybe

That is taken care of by our [wiki page ](https://wiki.squid-cache.org/StandardsCompliance)

doc/HTTP-codes.txt
yadij: Added a reference to the IANA registry in the description for this.

doc/rfc/1-index.txt
yadij: This I kept because that document is still somewhat dynamic with each Haproxy release and the file we have is the one Squid was implemented from. Same goes for the IETF Draft documents retained.

rousskov: > the file we have is the one Squid was implemented from

I am pretty sure the above assertion is false or misleading. While earlier PROXY protocol implementation in Squid may have been based on that specific revision of the specs, significant bug fixes and improvements that followed used (and should continue to use) then-current (i.e. contemporary) revisions.

Moreover, it does not really matter much what document revision existed at the time of a given Squid commit[^1]. What really matters is whether Squid interoperates and, to a lesser extent, follows the latest revision of the specs. Keeping any revision of the specs in the Squid repository does not help with any of that. And keeping a stale revision can only make things worse.

[^1]: If that information is really needed, it can usually be recovered without keeping a copy of the specs in the Squid repository.

PR #950, opened 2021-12-13
Log early level-0/1 debugs() messages to cache_log

rousskov:
Commit d7ca82e dropped cache.log-recording of debugs() messages produced
by `finalizeConfig` runners (e.g., `WARNING: mem-cache size is...`).
This change restores that functionality (by buffering early messages
until the cache.log file is opened) and improves early debugs() handling
as detailed below.

## Squid has three channels for debugs() messages:

* cache.log (`cache_log`): Settles as squid.conf settings take effect.
* stderr (mostly `-d`): Settles when command line options take effect.
* syslog (mostly `-s`): Settles when command line options take effect.

Squid always ignores debugging messages with section/level mismatching
Debug::Levels configuration (driven by a combination of the `-k debug`,
`-X`, and `debug_options` directives). _Beyond_ that ever-present
top-level filter, each debugging channel has its own set of rules that
determine which filtered debugs() messages the channel accepts; the
following approximate summary is based on the changes in this commit:

* cache.log: all messages;
* stderr: either messages satisfying explicit `-d` level restrictions or
  messages that Squid failed to write to cache.log (if no -d);
* syslog: level-0/1 messages and `ForceAllert` messages.

This change encapsulates channel-specific logic in dedicated classes.

## Guiding design principles

* **no repetition**: A channel must record each message at most once.
* **no reordering**: Each channel must preserve same-process debugs()
  call order across all recorded messages.
* **no loss**: Each channel must log all messages matching the channel
  configuration/filters.
* **no delay**: Each channel must record each message ASAP.
* **cache.log primacy**: Admins want messages logged to cache.log if
  possible and to stderr/syslog only if necessary or explicitly
  requested.
* **cache.log locking**: No cache.log updates without a PID lock.

## The fix

Commit d7ca82e dropped level-0/1 messages produced by `finalizeConfig`
runners because, since that commit, the cache.log channel was opened
_after_ the `finalizeConfig` event. Official code also dropped other,
even earlier level-0/1 messages, violating the "cache.log primacy"
principle. We now save early level-0/1 messages into a buffer. This
buffering can be misinterpreted as violating the "no [artificial] delay"
principle, but the messages are actually written to cache.log ASAP;
without the buffer the messages would be missing from the "primary"
cache.log and, in many use cases without stderr capturing, completely.

If Squid fails to open cache.log, early messages saved for the cache.log
channel are given to the stderr channel (following the "no loss"
principle). The stderr channel logs those messages that obey explicit
`-d` restrictions and do not violate the "no reordering" and "no
repetition" principles. Violations are tracked by assigning each message
its debugs() call sequence number.

The early messages buffer is currently dedicated to level-0/1 messages
because we were worried that level-2+ messages (if enabled via `-X`)
would overflow any reasonably-sized buffer[^1]. Correctly handling such
overflows is very difficult (we tried), so we avoid them instead.

## Side effects and surprises

To allow admins to see early level-1 cache.log messages (without adding
an `if early` check to the debugs() macro[^2]), we changed the _initial_
Debug::Levels value from `ALL,0` to `ALL,1`, matching the debug_options
default set later. This fix uncovered a few early level-1 messages that
were previously hidden[^3]:

    09:46:36| Startup: Initializing Authentication Schemes ...
    09:46:36| Startup: Initialized Authentication Scheme 'basic'
    09:46:36| Startup: Initialized Authentication Scheme 'digest'
    09:46:36| Startup: Initialized Authentication Scheme 'negotiate'
    09:46:36| Startup: Initialized Authentication Scheme 'ntlm'
    09:46:36| Startup: Initialized Authentication.
    09:46:36| Processing Configuration File: .../squid.conf ...
    09:46:36| Initializing https:// proxy context
    09:46:36| Set Current Directory to /usr/local/squid/var

The fact that many debugs() messages happen before cache_log can be
opened is fairly obvious, but it is also true that stderr and syslog
channels cannot write messages immediately. Both channels need to wait
for the command line options to be parsed. Even the global `stderr`
variable may not be available during very early debugs() calls! Each
channel now buffers level-0/1 messages until it settles.

Since the early message buffers are limited to level-0/1 messages,
initial cache.log records logged by `squid -X` are level-0/1 messages,
followed by true ALL,9 debugging. The admins can get early ALL,9
messages via stderr, of course.

The `-z` command-line option no longer overrides `-d` settings.

Squid no longer writes to cache_log after removing the PID file.

fatal() text and the `Squid Cache...: Terminated abnormally.` message
are no longer dropped during certain early process terminations.

`squid -k ...` logs some new level-0/1 messages to stderr.

Improved support for assert()/debugs() triggered from within the Debug
module: Besides crashes, such "internal" debugs() could be logged before
earlier "external" messages and some assert() messages could be lost.

[^1]: The restriction to only buffer level-0/1 messages can be easily
removed (after research and discussion) if there is consensus that the
actual memory required to accumulate all typical level-2+ early messages
is worth spending on making `squid -X` write all messages to cache.log.

[^2]: Squid has thousands of debugs() calls (and counting), including
many calls on performance-sensitive paths. Most debugs() calls should do
nothing by default. Thus, the speed at which the debugs() macro can skip
logging is an important common case on the performance sensitive path.
Similarly, disruption to CPU processing pipeline due to top-level
debugs() checks may be important. Actually writing the message to the
cache.log may not be that important -- at that time, the performance
battle can be considered lost -- but the initial rejection is.

[^3]: This change is not about the levels of any specific messages.
Wrong message levels (if any) should be fixed separately. This change
does not imply that the newly discovered messages have wrong levels.

### DIFF
--- a/src/debug.cc
+++ b/src/debug.cc
@@ -963,7 +963,6 @@ SyslogPriority(const DebugMessageHeader &header)
 void
 SyslogChannel::write(const DebugMessageHeader &header, const std::string &body)
 {
-    // TODO: Use SQUIDSBUFPH/SQUIDSBUFPRINT if all syslog()s support %.*s
     syslog(SyslogPriority(header), "%s", body.c_str());
     noteWritten(header);
 }
--- a/src/debug.cc
+++ b/src/debug.cc
@@ -112,17 +112,30 @@ public:
     bool forceAlert; ///< debugs() forceAlert flag
 };
 
+// Avoid SBuf for CompiledDebugMessageBody:
+// * SBuf's own debugging may create a lot of reentrant debugging noise.
+// * Debug::Context::buf is an std::string-based STL ostream. Converting its
+//   buf() result to a different kind of string may increase complexity/cost.
+// TODO: Consider switching to a simple fixed-size buffer and a matching stream!
+/// The processed "content" (i.e. the last parameter) part of a debugs() call.
+using CompiledDebugMessageBody = std::string;
+
 /// a fully processed debugs(), ready to be logged
 class CompiledDebugMessage
 {
 public:
     using Header = DebugMessageHeader;
-    CompiledDebugMessage(const Header &aHeader, const std::string &aBody);
+    using Body = CompiledDebugMessageBody;
+
+    CompiledDebugMessage(const Header &aHeader, const Body &aBody);
 
     Header header; ///< debugs() meta-information; reflected in log line prefix
-    std::string body; ///< the log line after the prefix (without the newline)
+    Body body; ///< the log line after the prefix (without the newline)
 };
 
+// We avoid PoolingAllocator for CompiledDebugMessages to minimize reentrant
+// debugging noise. This noise reduction has negligible performance overhead
+// because it only applied to early messages, and there are few of them.
 /// debugs() messages captured in LogMessage() call order
 using CompiledDebugMessages = std::deque<CompiledDebugMessage>;
 
@@ -150,7 +163,7 @@ public:
 
     /// Write the message to the channel if the channel accepts (such) messages.
     /// This writing may be delayed until the channel configuration is settled.
-    void log(const DebugMessageHeader &, const std::string &body);
+    void log(const DebugMessageHeader &, const CompiledDebugMessageBody &);
 
 protected:
     /// output iterator for writing CompiledDebugMessages to a given channel
@@ -187,16 +200,16 @@ protected:
     virtual bool shouldWrite(const DebugMessageHeader &) const = 0;
 
     /// write the corresponding debugs() message into the channel
-    virtual void write(const DebugMessageHeader &, const std::string &body) = 0;
+    virtual void write(const DebugMessageHeader &, const CompiledDebugMessageBody &) = 0;
 
     /// stores the given early message (if possible) or forgets it (otherwise)
-    void saveMessage(const DebugMessageHeader &, const std::string &body);
+    void saveMessage(const DebugMessageHeader &, const CompiledDebugMessageBody &);
 
     /// stop saving and log() any "early" messages, in recordNumber order
     static void StopSavingAndLog(DebugChannel &, DebugChannel * = nullptr);
 
     /// Formats a validated debugs() record and writes it to the given FILE.
-    void writeToStream(FILE &, const DebugMessageHeader &, const std::string &body);
+    void writeToStream(FILE &, const DebugMessageHeader &, const CompiledDebugMessageBody &);
 
     /// reacts to a written a debugs() message
     void noteWritten(const DebugMessageHeader &);
@@ -225,7 +238,7 @@ public:
 protected:
     /* DebugChannel API */
     virtual bool shouldWrite(const DebugMessageHeader &) const final;
-    virtual void write(const DebugMessageHeader &, const std::string &body) final;
+    virtual void write(const DebugMessageHeader &, const CompiledDebugMessageBody &) final;
 };
 
 /// DebugChannel managing messages destined for "standard error stream" (stderr)
@@ -246,7 +259,7 @@ public:
 protected:
     /* DebugChannel API */
     virtual bool shouldWrite(const DebugMessageHeader &) const final;
-    virtual void write(const DebugMessageHeader &, const std::string &body) final;
+    virtual void write(const DebugMessageHeader &, const CompiledDebugMessageBody &) final;
 
 private:
     /// whether we are the last resort for logging debugs() messages
@@ -264,7 +277,7 @@ public:
 protected:
     /* DebugChannel API */
     virtual bool shouldWrite(const DebugMessageHeader &) const final;
-    virtual void write(const DebugMessageHeader &, const std::string &body) final;
+    virtual void write(const DebugMessageHeader &, const CompiledDebugMessageBody &) final;
 
 private:
     bool opened = false; ///< whether openlog() was called
@@ -287,7 +300,7 @@ public:
 
     /// Log the given debugs() message to appropriate channel(s) (eventually).
     /// Assumes the message has passed the global section/level filter.
-    void log(const DebugMessageHeader &, const std::string &body);
+    void log(const DebugMessageHeader &, const CompiledDebugMessageBody &);
 
     /// Start using an open cache_log file as the primary debugs() destination.
     /// Stop using stderr as a cache_log replacement (if we were doing that).
@@ -375,7 +388,7 @@ DebugModule::DebugModule()
 }
 
 void
-DebugModule::log(const DebugMessageHeader &header, const std::string &body)
+DebugModule::log(const DebugMessageHeader &header, const CompiledDebugMessageBody &body)
 {
     cacheLogChannel.log(header, body);
     stderrChannel.log(header, body);
@@ -458,7 +471,7 @@ DebugChannel::stopEarlyMessageCollection()
 }
 
 void
-DebugChannel::log(const DebugMessageHeader &header, const std::string &body)
+DebugChannel::log(const DebugMessageHeader &header, const CompiledDebugMessageBody &body)
 {
     if (header.recordNumber <= lastWrittenRecordNumber)
         return;
@@ -515,7 +528,7 @@ DebugChannel::StopSavingAndLog(DebugChannel &channelA, DebugChannel *channelBOrN
 }
 
 void
-DebugChannel::saveMessage(const DebugMessageHeader &header, const std::string &body)
+DebugChannel::saveMessage(const DebugMessageHeader &header, const CompiledDebugMessageBody &body)
 {
     if (!earlyMessages)
         return; // we have stopped saving early messages
@@ -534,7 +547,7 @@ DebugChannel::saveMessage(const DebugMessageHeader &header, const std::string &b
 }
 
 void
-DebugChannel::writeToStream(FILE &destination, const DebugMessageHeader &header, const std::string &body)
+DebugChannel::writeToStream(FILE &destination, const DebugMessageHeader &header, const CompiledDebugMessageBody &body)
 {
     fprintf(&destination, "%s%s| %s\n",
             debugLogTime(header.timestamp),
@@ -559,7 +572,7 @@ CacheLogChannel::shouldWrite(const DebugMessageHeader &) const
 }
 
 void
-CacheLogChannel::write(const DebugMessageHeader &header, const std::string &body)
+CacheLogChannel::write(const DebugMessageHeader &header, const CompiledDebugMessageBody &body)
 {
     writeToStream(*TheLog.file(), header, body);
     fflush(TheLog.file());
@@ -588,7 +601,7 @@ StderrChannel::shouldWrite(const DebugMessageHeader &header) const
 }
 
 void
-StderrChannel::write(const DebugMessageHeader &header, const std::string &body)
+StderrChannel::write(const DebugMessageHeader &header, const CompiledDebugMessageBody &body)
 {
     writeToStream(*stderr, header, body);
 }
@@ -653,7 +666,7 @@ DebugMessageHeader::DebugMessageHeader(const DebugRecordCount aRecordNumber, con
 
 /* CompiledDebugMessage */
 
-CompiledDebugMessage::CompiledDebugMessage(const Header &aHeader, const std::string &aBody):
+CompiledDebugMessage::CompiledDebugMessage(const Header &aHeader, const Body &aBody):
     header(aHeader),
     body(aBody)
 {
@@ -961,7 +974,7 @@ SyslogPriority(const DebugMessageHeader &header)
 }
 
 void
-SyslogChannel::write(const DebugMessageHeader &header, const std::string &body)
+SyslogChannel::write(const DebugMessageHeader &header, const CompiledDebugMessageBody &body)
 {
     syslog(SyslogPriority(header), "%s", body.c_str());
     noteWritten(header);
@@ -970,7 +983,7 @@ SyslogChannel::write(const DebugMessageHeader &header, const std::string &body)
 #else
 
 void
-SyslogChannel::write(const DebugMessageHeader &, const std::string &)
+SyslogChannel::write(const DebugMessageHeader &, const CompiledDebugMessageBody &)
 {
     assert(!"unreachable code because opened, shouldWrite() are always false");
 }
@@ -1219,7 +1232,7 @@ Debug::Context::rewind(const int aSection, const int aLevel)
     assert(upper == Current);
     assert(!waitingForIdle);
 
-    buf.str(std::string());
+    buf.str(CompiledDebugMessageBody());
     buf.clear();
     // debugs() users are supposed to preserve format, but
     // some do not, so we have to waste cycles resetting it for all.
--- a/src/debug.cc
+++ b/src/debug.cc
@@ -18,35 +18,68 @@
 #include "util.h"
 
 #include <algorithm>
+#include <deque>
+#include <functional>
+#include <memory>
 
 /* for shutting_down flag in xassert() */
 #include "globals.h"
 
 char *Debug::debugOptions = NULL;
 int Debug::override_X = 0;
-int Debug::log_stderr = -1;
 bool Debug::log_syslog = false;
 int Debug::Levels[MAX_DEBUG_SECTIONS];
 char *Debug::cache_log = NULL;
 int Debug::rotateNumber = -1;
 DebugMessages TheDebugMessages;
-static const char *debugLogTime(void);
+
+/// a counter related to the number of debugs() calls
+using DebugRecordCount = uint64_t;
+
+class DebugModule;
+
+/// Debugging module singleton.
+static DebugModule *Module_ = nullptr;
+
+/// Explicitly configured maximum level for debugs() messages written to stderr.
+/// debugs() messages with this (or lower) level will be written to stderr (and
+/// possibly other channels).
+static int ExplicitStderrLevel = -1;
+// TODO: Use static Optional<int> ExplicitStderrLevel;
+static bool ExplicitStderrLevelSet = false;
+
+/// ExplicitStderrLevel preference or default: Just like with
+/// ExplicitStderrLevel, debugs() messages with this (or lower) level will be
+/// written to stderr (and possibly other channels), but this setting is ignored
+/// when ExplicitStderrLevel is set. This setting is also ignored after major
+/// problems that prevent logging of important debugs() messages (e.g., failing
+/// to open cache_log or assertions).
+static int DefaultStderrLevel = -1;
+
+/// early debugs() with higher level are not buffered and, hence, may be lost
+static constexpr int EarlyMessagesLevel = DBG_IMPORTANT;
+
+static const char *debugLogTime(time_t t);
 static const char *debugLogKid(void);
 #if HAVE_SYSLOG
 #ifdef LOG_LOCAL4
 static int syslog_facility = 0;
 #endif
-static void _db_print_syslog(const bool forceAlert, const char *format, va_list args);
 #endif
-static void _db_print_stderr(const char *format, va_list args);
-static void _db_print_file(const char *format, va_list args);
 
 #if _SQUID_WINDOWS_
 extern LPCRITICAL_SECTION dbg_mutex;
 typedef BOOL (WINAPI * PFInitializeCriticalSectionAndSpinCount) (LPCRITICAL_SECTION, DWORD);
 #endif
 
-/// a (FILE*, file name) pair that uses stderr FILE as the last resort
+static void ResetSections(const int level = DBG_IMPORTANT);
+
+/// a cached calculateMyLeadingRole() result
+extern const char *XXX_Role;
+const char *XXX_Role = nullptr;
+static const char *myLeadingRole_XXX();
+
+/// a named FILE with very-early/late usage safety mechanisms
 class DebugFile
 {
 public:
@@ -60,8 +93,8 @@ public:
     /// go back to the initial state
     void clear() { reset(nullptr, nullptr); }
 
-    /// logging stream; the only method that uses stderr as the last resort
-    FILE *file() { return file_ ? file_ : stderr; }
+    /// an opened cache_log stream or nil
+    FILE *file() { return file_; }
 
     char *name = nullptr;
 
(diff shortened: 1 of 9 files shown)


rousskov:
The following GitHub gist uses a representative sample of debugs() statements to show how debugs() message visibility changed from before #474 to after #474 to after this PR changes: https://gist.github.com/rousskov/c13f3850ff86ac2a5566aca231cd194c

yadij:
I have yet to review the code or design, but that gist is already showing me a few problems (assuming it is accurate).

* The ```-X``` command line argument is supposed to force *full* debugging (ie. level-9) yet there are messages not being logged after this PR.
* Also, *force* meaning the ```-X``` is supposed to override ```-d``` and squid.conf settings whenever used.
* The (non-)existence of ```-N``` or ```-z``` should have no effect on whether a message is logged. They are about what role the process is performing - not debug levels.

rousskov:
> * The `-X` command line argument is supposed to force _full_ debugging (ie. level-9) yet there are messages not being logged after this PR.

I do not know which channel you are talking about. In the future reviews, please be specific because each channel has its own admission control policy and settling moment. To speed things up, I will cover all three channel in this response:

* The syslog channel does not and, IMO, should not get level-2+ messages (before and after these PRs).
* The cache.log channel is covered by "The early messages buffer is currently dedicated..." paragraph in the PR description: This channel does not and should not write messages until the PID file "lock" is acquired. We could save all early level-2+ messages and log them after obtaining that lock, but that would be a change to how Squid has worked before all these PRs, and it is debatable whether that change is a good idea, so I think it is best to leave that discussion/change outside this PR scope. This PR lays the necessary foundation for saving even more early cache.log messages if we decide that saving even more early cache.log messages is a good idea.
* The stderr channel is covered by your next bullet, so I will skip it here and discuss it below.

> * Also, _force_ meaning the `-X` is supposed to override `-d` and squid.conf settings whenever used.

"Force full debugging" in `-X` documentation implies overriding or ignoring "less than full" debugging (as configured by `debug_options` or defaults). `-X` is about debugging verbosity being "full", not about _where_ that debugging info is sent.

Ignoring `-d` when seeing `-X` does not enable any otherwise inaccessible feature or functionality because the admin adding `-X` can obviously also add `-d9` if that is what they are after. And there is harm in ignoring `-d` because doing so deprives the admin from the ability to control stderr channel verbosity.
 
> * The (non-)existence of `-N` or `-z` should have no effect on whether a message is logged. They are about what role the process is performing - not debug levels.

I am not sure what specific problem you are thinking about here. Please clarify after noting that `-N` and `-z` naturally affect which debugs() statements are reached/executed, and that `-z` used to override `-d` (that bug has been fixed in this PR).

yadij:
To clarify what I mean:

*  ```squid -N -d0``` - all level 0 messages should be logged to stderr. Use debugs_options for cache.log debug level.
* ```squid -N -X``` - all messages should be logged to cache.log. Command line ```-X``` forces all selected channels to level 9, ignore debug_options.
* ```squid -N -X -d9``` - all messages should be logged to both cache.log and stderr channels. Command line ```-X``` forces debug to level 9. ```-d``` selects channel stderr. Value on ``-d 9`` is irrelevant / overridden by ```-X``` forcing full debug.
* There is no output set for ```-X -d0``` etc to show that the value of the ```-d``` is ignored when ```-X``` is used.

src/debug.cc
```diff
@@ -71,19 +98,359 @@ class DebugFile
...
+using CompiledDebugMessageBody = std::string;
+
+/// a fully processed debugs(), ready to be logged
+class CompiledDebugMessage
+{
+public:
+    using Header = DebugMessageHeader;
+    using Body = CompiledDebugMessageBody;
+
+    CompiledDebugMessage(const Header &aHeader, const Body &aBody);
```
yadij, line 129: aHeader and aBody parameter names are redundant. please remove
```suggestion
    CompiledDebugMessage(const Header &, const Body &);
```

src/debug.cc
```diff
@@ -71,19 +98,359 @@ class DebugFile
...
+/// The processed "content" (i.e. the last parameter) part of a debugs() call.
+using CompiledDebugMessageBody = std::string;
+
+/// a fully processed debugs(), ready to be logged
+class CompiledDebugMessage
+{
+public:
+    using Header = DebugMessageHeader;
+    using Body = CompiledDebugMessageBody;
+
```
yadij, line 128: This class is quite simple. No need to complicate it with these typedef equivalents.

src/debug.cc
```diff
@@ -71,19 +98,359 @@ class DebugFile
...
+/// debugs() messages captured in LogMessage() call order
+using CompiledDebugMessages = std::deque<CompiledDebugMessage>;
+
+/// a receiver of debugs() messages (e.g., stderr or cache.log)
+class DebugChannel
+{
+public:
+    using EarlyMessages = std::unique_ptr<CompiledDebugMessages>;
+
+    explicit DebugChannel(const char * const aName): name(aName), earlyMessages(new CompiledDebugMessages()) {}
```
yadij, line 147: Please wrap
```suggestion
    explicit DebugChannel(const char * const aName) :
       name(aName),
       earlyMessages(new CompiledDebugMessages())
   {}
```

src/debug.cc
```diff
@@ -71,19 +98,359 @@ class DebugFile
...
+    public:
+        using difference_type = void;
+        using value_type = void;
+        using pointer = void;
+        using reference = void;
+        using iterator_category = std::output_iterator_tag;
+
+        explicit Logger(DebugChannel &ch): channel(ch) {}
+
+        Logger &operator=(const CompiledDebugMessage &message) {
```
yadij, line 180:
```suggestion
        Logger &operator=(const CompiledDebugMessage &message)
        {
```

src/debug.cc
```diff
@@ -71,19 +98,359 @@ class DebugFile
...
+    bool coveringForCacheLog = false;
+};
+
+/// syslog DebugChannel
+class SyslogChannel: public DebugChannel
+{
+public:
+    SyslogChannel(): DebugChannel("syslog") {}
+
+    void markOpened() { opened = true; }
```
yadij, line 274: Missing documentation

src/debug.cc
```diff
@@ -71,19 +98,359 @@ class DebugFile
...
+LoggingSectionGuard::LoggingSectionGuard()
+{
+    ++LoggingConcurrencyLevel;
+}
+
+LoggingSectionGuard::~LoggingSectionGuard()
+{
+    if (--LoggingConcurrencyLevel == 0)
+        Debug::LogWaitingForIdle();
+}
```
yadij, line 371: Please inline these to the local class definition.

src/debug.cc
```diff
@@ -92,6 +459,219 @@ ResyncDebugLog(FILE *newFile)
...
+    lastWrittenRecordNumber = header.recordNumber;
+}
+
+/* CacheLogChannel */
+
+bool
+CacheLogChannel::shouldWrite(const DebugMessageHeader &) const
+{
+    return TheLog.file();
+}
```
yadij, line 571: Please inline this and similar one-liner methods especially if they are virtuals.

src/debug.cc
```diff
@@ -454,20 +962,57 @@ _db_set_syslog(const char *facility)
...
+    noteWritten(header);
+}
+
+#else
 
 void
-_db_set_syslog(const char *)
+SyslogChannel::write(const DebugMessageHeader &, const CompiledDebugMessageBody &)
+{
+    assert(!"unreachable code because opened, shouldWrite() are always false");
```
yadij, line 986: This trick will be reported as an error by static analysis. Please just make this a ```{}``` method definition if it has to be defined at all.

src/debug.cc
```diff
@@ -618,22 +1183,39 @@ debugLogKid(void)
     return "";
 }
 
+/// The number of xassert() calls in the call stack. Treat as private to
```
yadij, line 1186: bool is not a number.

src/main.cc
```diff
@@ -521,8 +521,9 @@ mainHandleCommandLineOption(const int optId, const char *optValue)
 
     case 'd':
         /** \par d
-         * Set global option Debug::log_stderr to the number given following the option */
-        Debug::log_stderr = xatoi(optValue);
+         * debugs() messages with the given debugging level (and the more
+         * important ones) should be written to stderr */
+        Debug::ResetStderrLevel(xatoi(optValue));
```
yadij, line 526: Missing check that -X is not forcing the value to -d9.

 And -X handling missing check for channels already activated that need a reset to level-9.

src/debug.cc
```diff
@@ -689,13 +1291,31 @@ Debug::Start(const int section, const int level)
...
     if (Current->level <= DBG_IMPORTANT)
         Current->buf << CurrentCodeContextDetail;
 
-    // TODO: Optimize to remove at least one extra copy.
-    _db_print(Current->forceAlert, "%s\n", Current->buf.str().c_str());
+    if (Current->waitingForIdle) {
+        const auto past = Current;
+        Current = past->upper;
+        past->upper = nullptr;
+        // do not delete `past` because we store it in WaitingForIdle below
```
yadij, line 1305: Please use double-quotes instead of shell-code quotation.
```suggestion
        // do not delete "past" because we store it in WaitingForIdle below
```

yadij (changes requested):
Done a basic check. Mostly these are style issues.

In regards to saving debug levels 2-9 prior to cache.log opening. I think we should save to level-9 at startup, then prune that saved set as soon as the command line X/d/l arguments are known.

rousskov:
> * `squid -N -d0` - all level 0 messages should be logged to stderr. Use debugs_options for cache.log debug level.

Yes, but, ideally, we should not phrase it that way because `squid -d` controls what gets logged to stderr _after_ the primary debug_options filter (or equivalent). The `-d` command line option does not override that primary filter. It did not overwrite it before these PRs and should not overwrite it after these PRs. No changes here.

For example, if we assume, for illustration purposes, that `debug_options ALL,none` or equivalent configuration is supported and disables logging of level-0+ (i.e. all) messages, then `squid -N -d0` will not log any messages to stderr once that debug_options setting takes effect. This is how debugs() have worked for a very very long time, and it is the right approach.

> * `squid -N -X` - all messages should be logged to cache.log. Command line `-X` forces all selected channels to level 9, ignore debug_options.

This is correct except for the special case of early level-2+ debug messages that are emitted before cache.log is opened, as discussed in "The early messages buffer is currently dedicated..." paragraph in the PR description and again in my [comment](https://github.com/squid-cache/squid/pull/950#issuecomment-993610831).

> * `squid -N -X -d9` - all messages should be logged to both cache.log and stderr channels.

For stderr, that is correct except for the special case of the early messages emitted before command line options are parsed.

The caveats of the cache.log logging are covered by the previous bullet discussion.

> Command line `-X` forces debug to level 9.

Correct. `-X` essentially removes the primary/top-level filter which applies to all debugs() and all channels.

> `-d` selects channel stderr.

One can say that, but `-d` is pretty much meaningless alone, without the number. The "selection" is done for a given subset of debugs() messages, specified by the intersection of the debug_options settings (or equivalent -- the primary filter) and the `-d` value.

> Value on `-d 9` is irrelevant / overridden by `-X` forcing full debug.

As I tried to [explain](https://github.com/squid-cache/squid/pull/950#issuecomment-993610831), there is no reason for (and there is harm in) overriding `-d` with `-X`, but, it this specific case of `-d9`, the effect is the same as if overriding was going on because `-d9` admits all messages (that were already admitted by `-X`) to stderr.

> * There is no output set for `-X -d0` etc to show that the value of the `-d` is ignored when `-X` is used.

Again, with `-X`, `-d` was not really ignored before these PRs, is not ignored in the current master, will not be ignored after this PR, and should not be ignored in the foreseeable future.

I have now added `-N -X -d0` and `-N -X -d1` sets to the [sample](https://gist.github.com/rousskov/c13f3850ff86ac2a5566aca231cd194c/edit), and both show expected/correct results IMO. You might wonder why, before these PRs, some level-1+ messages were logged to stderr despite the `-d0` restriction. The answer is not "because `-X` overrode `-d`". Those early messages were logged to stderr because, before these PRs, Squid blindly used stderr until it could write to cache.log, even when the admin explicitly asked Squid not to use stderr for level-1+ messages.[^1]. That violation of admin direct orders is fixed now.

[^1]: Before these fixes, the `debug_log` macro simply returned stderr when cache_log `FILE` was nil, and _db_print_file() then blindly wrote messages to that debug_log as if it were cache_log. This temporary effect disappears after Squid opens cache_log, illustrating what the code already says -- `-X` does not override `-d`.

rousskov:
> Amos: In regards to saving debug levels 2-9 prior to cache.log opening. I think we should save to level-9 at startup, then prune that saved set as soon as the command line X/d/l arguments are known.

Let's discuss that change after this PR is merged, for the reasons I have already given in the "The early messages buffer is currently dedicated..." paragraph in the PR description and detailed in another [comment](https://github.com/squid-cache/squid/pull/950#issuecomment-993610831).

I will go through your polishing change requests in the background, but we need to resolve the big misunderstanding of "`-X` overrides `-d`". I hope my even more detailed [response](https://github.com/squid-cache/squid/pull/950#issuecomment-994289167) and added [samples](https://gist.github.com/rousskov/c13f3850ff86ac2a5566aca231cd194c) will do that, but the ball is on your side.

src/debug.cc
rousskov: This is obviously not a shell-related code but a source code comment describing a piece of C++ code. These `backticks` are well-known (in modern markdown and alike) escape sequences that are commonly used to distinguish variable names from words (where the two can be easily confused). Double "quotes" are worse (for this purpose) because, in prose like this, they are commonly used to mean "in indirect sense". I am not forcing others to use backticks, but I hope I will not be forced to use double quotes either.

src/debug.cc
rousskov: > This trick will be reported as an error by static analysis.

What makes you think that? What error? FWIW, we are already using the `assert(!"explanation")` trick in `src/acl/Options.h`. It is a well-known trick.

> Please just make this a {} method definition

I think we should assert if a supposed-to-be-unreachable code has been reached, especially when the code is somewhat sensitive and detecting a violation is easy.

src/debug.cc
rousskov: Why? I think they would distract from the code comprehension if inlined, especially since one of the paired methods is a multi-line statement with non-trivial logic.

src/debug.cc
rousskov: Why? I think this method should _not_ be inlined especially because it is virtual:

* all similar methods are defined consistently this way (because most are not one-liners);
* inlining distracts the reader from comprehending the class declaration
* minor code changes (e.g., adding a source code comment) will not require moving the method;
* inlining of virtual methods is unlikely, but the optimizing compiler will inline (if it can) regardless of where we put the definition;
* method definitions (and their history of changes) are easier to find when they are _not_ inlined

yadij: > Why?

Primarily as an optimization to avoid any method jump or virtual overheads in the logic which can be inlined. A bunch of the "thousand cuts" against Squid performance is coming from simple methods like this which could be inlined but are not.

> I think this method should _not_ be inlined especially because it is virtual:
> 
>     * all similar methods are defined consistently this way (because most are not one-liners);

Only if you isolate your view to this class. The wider Squid code style regularly inlines simple methods like these.

>     * inlining distracts the reader from comprehending the class declaration

Not important. Verbosity of the mandatory comments does more harm there than one-liner implementations. In fact the implementation being present can help avoid some of the comment.

>     * minor code changes (e.g., adding a source code comment) will not require moving the method;
 
Moving is not difficult, and with our emphasis on design perfection should be a rarity for methods this simple.

FYI, I am okay with inline methods having a few lines even. I just dont request it for 2+ and leave it to the authors judgment on whether such a move just for a comment is worth it.

>     * inlining of virtual methods is unlikely, but the optimizing compiler will inline (if it can) regardless of where we put the definition;

Only advanced optimization levels are capable of that when the definition is in the .cc file. The compileler may not even have access to the on-line definition at simpler optimization levels. By placing it in the header we guarantee that even un-optimized builds can inline and gain the performance benefits.

> 
>     * method definitions (and their history of changes) are easier to find when they are _not_ inlined

IME this is not true, history is easily available with git regardless of where they are placed.

src/debug.cc
yadij: It was that already existing usage that brought it to my attention. Coverity added an issue. I'm not finding it there right now, but IIRC it was complaining about constant boolean test causing dead code existence.

src/debug.cc
yadij: To use yes that is true. But the scripts at times place lines onto stdin/stdout or variables where the shell can get its fingers on the syntax.
If you want to avoid double-quotes it is fine to use singles or brackets etc. But please not the backticks.

src/debug.cc
rousskov: > > Alex: Why?
> Amos: Primarily as an optimization to avoid any method jump or virtual overheads in the logic which can be inlined.

In this case, a modern optimizing compiler will (not) inline regardless of where the method is defined. Also, the compiler will either know the exact address of the virtual function in the caller context or not, regardless of where the virtual method is defined. The definition _location_ has no affect on that knowledge (and, hence, on the compiler ability to create code that avoids runtime vtable lookups).

Your primary motivation is based on false assumptions. The secondary arguments below do not invalidate my points in defense of the current method definition location. Please withdraw your change request.


> > Alex: I think this method should _not_ be inlined especially because it is virtual:
> > * all similar methods are defined consistently this way (because most are not one-liners);
 
> Amos: Only if you isolate your view to this class. The wider Squid code style regularly inlines simple methods like these.

The wider Squid code that inlines virtual methods in similar circumstances is wrong. There is no one simple "inline all one-liners" rule. The decision should take more factors into account, and that is exactly what this PR is doing.

> > Alex: * inlining distracts the reader from comprehending the class declaration

> Amos: Not important.

It is important enough to be a factor IMO.

> Amos: Verbosity of the mandatory comments does more harm there than one-liner implementations. In fact the implementation being present can help avoid some of the comment.

Let's not discuss your misinterpretation of description rules and your opinion on this PR comments in _this_ change request. If you strongly believe some PR-added comments should be removed, please open dedicated change requests.

> > Alex: * minor code changes (e.g., adding a source code comment) will not require moving the method;

> Amos: Moving is not difficult, and with our emphasis on design perfection should be a rarity for methods this simple.

Your arguments do not invalidate my point, so I will not detail their serious flaws.

> > Alex: * inlining of virtual methods is unlikely, but the optimizing compiler will inline (if it can) regardless of where we put the definition;

> Amos: Only advanced optimization levels are capable of that when the definition is in the .cc file. The compileler may not even have access to the on-line definition at simpler optimization levels. By placing it in the header we guarantee that even un-optimized builds can inline and gain the performance benefits.

In this particular case, the definition and the declaration are both in the same `.cc` file so the compiler can see everything and safely assume a lot. I do not know what you call "advanced optimization levels", but the basic inlining we are talking about here starts with `-O2`. I would not call that level "advanced". In fact, that is usually the minimum level optimized builds should use. Finally, unoptimized builds do not inline even if the definition is provided at the declaration time! In summary, the arguments against my point do not apply to the use case in question or are invalid.
 
> > Alex: * method definitions (and their history of changes) are easier to find when they are _not_ inlined
 
> Amos: IME this is not true, history is easily available with git regardless of where they are placed.

I am not talking about availability. I am talking about the ease of finding/locating method definitions.

If you do not believe me, try writing a `git grep` search expression that locates the NullaryMemFunT::doDial() method definition and only that method definition, without relying on a particular implementation of that method, its formatting, and any other details that the searching developer (or script!) may not know or that may change from one git commit to another. If that method were not inlined, the task would be trivial, of course: `git grep NullaryMemFunT::doDial\b`.

Increased search difficulty goes beyond `git grep`, of course. I am just using that command as the simplest example.

src/debug.cc
rousskov: Perhaps Coverity has improved to handle these cases correctly, just like it obviously handles dozens of `assert(false)` cases. I can volunteer to address associated Coverity issues if they (re)surface. Meanwhile, we should make our code worse just because a future Coverity version might mishandle correct code. If that mishandling happens, we will deal with it. Please withdraw your change request.

src/debug.cc
rousskov: > the scripts at times place lines onto stdin/stdout or variables where the shell can get its fingers on the syntax

I do not think we should make our code worse because of a possibility that a _buggy_ script will mishandle it. We should be allowed (and are already using) lots of shell-unsafe, HTML-unsafe, whatever-unsafe characters in C++ comments! This is normal/expected -- each syntax has its own set of safe characters.

We should not invent new controversial restrictions like that without strong consensus behind them!

Please withdraw your change request.

>  If you want to avoid double-quotes it is fine to use singles or brackets etc. But please not the backticks.

I do not want to avoid double-quotes. I want to (continue to) use a well-known, modern notation for quoting `code`.

* Single quotes do not emphasize enough that this is a `code` quotation rather than something else. That is especially important for dictionary words like "past".
* Square and angle brackets are commonly used to mean "optional" or "required" in similar contexts.

Please do not misinterpret the bullets above as an invitation to make me evaluate all possible escape sequences. The PR is using the best sequence I know (in this context). If you could not come up with a clearly better one in two attempts, we should stop wasting time on this search.

src/debug.cc
yadij:
```suggestion
/// Whether there are any xassert() calls in the call stack. Treat as private to
```

src/debug.cc
yadij: For now, ignoring this. I can patch it later if necessary.

src/debug.cc
yadij: or, if you have a good reason for adding them. Please add a comment to the class to say why so they do not get polished away in future.

src/main.cc
yadij: This one is important please. I recall a number of times users have provided incorrect or useless debug traces because of these options behaviour colliding invisibly.

If you disagree on the exact interaction behaviour I outlined then at least add a clear debugs() to tell users what level is actually going to be logged when these values differ.

yadij:
STATUS: waiting on the unresolved requests from [previous review](https://github.com/squid-cache/squid/pull/950#pullrequestreview-832252376) to be implemented.

src/main.cc
rousskov: > I recall a number of times users have provided incorrect or useless debug traces because of these options behaviour colliding invisibly.

I doubt we can make this complex issue (three individually configurable channels, 9 levels, numerous debugging sections) disappear completely, but I hope that our changes will reduce the number of cases where users get not what they should expect (and cannot figure it out on their own). If needed, I can analyze any specific examples you have in mind (i.e. which options you asked users to use, which options they actually used, and what was wrong with the collected output).

> If you disagree on the exact interaction behaviour I outlined then at least add a clear debugs() to tell users what level is actually going to be logged when these values differ.

Yes, I disagree (as discussed [earlier](https://github.com/squid-cache/squid/pull/950#issuecomment-994289167
)). I added a debugs() for admins using `-X`. Please see 50d363b. I hope it delivers what you are looking for despite my incomplete understanding of what specific problem you are trying to address.

----

And, in case you were also waiting for explicit responses to the original change request, here they are:

> Missing check that -X is not forcing the value to -d9.
> And -X handling missing check for channels already activated that need a reset to level-9.

`-X` unconditionally resets all _section_ levels to 9 (see the code quoted below). As for the channels:

* stderr channel: `-X` should not force `-d9` as already discussed and illustrated at https://github.com/squid-cache/squid/pull/950#issuecomment-994289167
* syslog channel: `-X` should not force "level-9" for the syslog channel because that channel never logs anything above level 1 (except for forced alerts, but those are never above level 1 IIRC).
* cache.log channel: This channel does not have a level that can be forced or even configured! Or, if you prefer, the cache.log channel level is essentially hard-coded to infinity -- that channel logs _everything_ that the global section filter passes. That global section filter is applied before _any_ channel gets control. The corner cases of early and late debugging messages aside (they are discussed [elsewhere](https://github.com/squid-cache/squid/pull/950#issuecomment-994292460)), this means that a successfully opened cache.log will receive all debug messages from its opening to its closure.


```C++
case 'X':
    /** \par X
     * Force full debugging */
    Debug::parseOptions("rotate=0 ALL,9");
    Debug::override_X = 1;
    sigusr2_handle(SIGUSR2);
    break;
```

src/debug.cc
rousskov: Left as is. If you insist, I will remove these `using` declarations, but I believe they make the code a bit more readable/clear and do not increase complexity. They will also minimize the noise in the diff when/if we refactor to get rid of the Debug "class" add a namespace for the debugging-related APIs.

I doubt anybody accidentally removes these declarations. If they do get removed, then somebody would probably have a good reason for that removal, so no problem there.

src/debug.cc
rousskov: Fixed in 28393b0.

src/debug.cc
rousskov: De-inlined instead in 8213def.

src/debug.cc
rousskov: Fixed in 614c518. Similar problems fixed in c2c643e. We need to find a way to automate this.

src/debug.cc
rousskov: Fixed in commit f0d27aa. Thank you.

rousskov (commented):
I hope I have addressed all change requests.

src/debug.cc
yadij: IIRC the current astyle should be able to do it. I do not think we have enabled and tested that ability yet though.

src/main.cc
yadij: Thank you. I consider this resolved.

yadij (approved):

PR #951, opened 2021-12-14
Fix GCC v10.3 build after commit 8b082ed

rousskov:
basic_smb_auth.cc:57:1: warning: no previous declaration for
    `void print_esc(FILE*, char*)` [-Wmissing-declarations]

### DIFF
--- a/src/auth/basic/SMB/basic_smb_auth.cc
+++ b/src/auth/basic/SMB/basic_smb_auth.cc
@@ -53,7 +53,7 @@ struct SMBDOMAIN *lastdom = NULL;
  * to the read command of the bourne shell.
  */
 
-void
+static void
 print_esc(FILE * p, char *s)
 {
     char buf[HELPER_INPUT_BUFFER];


rousskov:
@kinkie, I suspect commit 8b082ed triggered this warning, but I did not check carefully. AFAICT, we had `-Wmissing-declarations` for GCC _before_ that commit, and that warning is not a part of the `-Wextra` set. I am guessing your commit secretly fixed something that properly enabled that and possibly other warnings, but I did not research that speculation. Please remove or change commit reference from/in the PR title if you think another master commit may be responsible.

yadij:
I suspect a change of compiler version somewhere making it visible rather than any code changes we have made. The more recent compilers are naturally getting more pedantic about things like this absent "static" keyword.

rousskov:
I think I figured it out.

Before commit 8b082ed, `-Wmissing-declarations` was added to `CFLAGS` but not `CXXFLAGS` (or equivalent):

```
configure:40040: BUILD EXTRA C FLAGS: -Wall -Wpointer-arith -Wwrite-strings -Wmissing-prototypes -Wmissing-declarations -Wcomments -Wshadow -Werror -pipe -D_REENTRANT
configure:40044: BUILD EXTRA C++ FLAGS: -Wall -Wpointer-arith -Wwrite-strings -Wcomments -Wshadow -Woverloaded-virtual -Werror -pipe -D_REENTRANT -I/usr/include/libxml2
```

Since commit 8b082ed, `-Wmissing-declarations` is added to `CXXFLAGS` but not `CFLAGS` (or equivalent):

```
Makefile:SQUID_CFLAGS =  -Werror -pipe -D_REENTRANT
Makefile:SQUID_CXXFLAGS =  -Werror -Wextra -Wno-unused-private-field -Wimplicit-fallthrough=2 -Wpointer-arith -Wwrite-strings -Wcomments -Wshadow -Wmissing-declarations -Werror -pipe -D_REENTRANT -I/usr/include/libxml2
```

Given the above, I am comfortable with leaving commit 8b082ed as a reference point for this fix (but if you want to remove it from the PR title, for any reason, please do!).

----

Why are you not seeing this warning and I do? There is another Squid bug that causes basic_smb_auth to be built _without_  `SQUID_CXXFLAGS`, masking the problem:

```
make -C src/auth/basic/SMB
g++ ... -DHELPERSCRIPT=... -g -O2 -march=native -c -o basic_smb_auth-basic_smb_auth.o ...
```

As you can see (if you trust my editing), the above build command uses basic compilation flags without any `SQUID_CXXFLAGS` entries.

Why? I speculate that we are incorrectly _overwriting_ the flags automatically assembled by Automake (instead of _adding_ our custom `DHELPERSCRIPT` flag to them):

```
Makefile.am:basic_smb_auth_CXXFLAGS = -DHELPERSCRIPT=...
```

Should not that customization also add `$(AM_CXXFLAGS)` or equivalent?

In my build environment, I am effectively bypassing that bug by specifying `-Wmissing-declarations` (and other extra flags) directly in well-known `CXXFLAGS` which, unlike custom `SQUID_CXXFLAGS`, is (evidently) still picked up and added to the basic_smb_auth build command...

If my speculation is correct, then there may be a couple of similar flag-suppressing bugs in other build targets. See `git grep '^[a-z].*_CXXFLAGS ='` for a starting point (a dirty and possibly incomplete list of _candidates_). Please fix that bug if you can.

yadij (approved):
Confirmed and replicated @rousskov suspicion about AM_CXXFLAGS missing from at least this helper build.

PR #953, opened 2021-12-20
Bug 5179: Correction of CVE 2021-28116 fix

amk1969:
The CVE changes did not handle received WCCP packets correctly.

This update allows to operate again, to the level squid
implemented WCCPv2 before. Tested with one cache and one router,
in both hash and mask assignment modes.

### DIFF
--- a/src/wccp2.cc
+++ b/src/wccp2.cc
@@ -297,7 +297,7 @@ static struct wccp2_capability_info_header_t wccp2_capability_info_header;
 struct wccp2_capability_element_t {
     uint16_t capability_type;
     uint16_t capability_length;
-    uint32_t capability_value;
+    uint32_t capability_value; // All capabilities have length of 4
 };
 static struct wccp2_capability_element_t wccp2_capability_element;
 
@@ -305,8 +305,8 @@ static struct wccp2_capability_element_t wccp2_capability_element;
 #define WCCP2_CAPABILITY_FORWARDING_METHOD  0x01
 #define WCCP2_CAPABILITY_ASSIGNMENT_METHOD  0x02
 #define WCCP2_CAPABILITY_RETURN_METHOD      0x03
-// 0x04 ?? - advertised by a 4507 (ios v15.1) Cisco switch
-// 0x05 ?? - advertised by a 4507 (ios v15.1) Cisco switch
+#define WCCP2_CAPABILITY_TRANSMIT_T         0x04
+#define WCCP2_CAPABILITY_TIMER_SCALE        0x05
 
 /* capability values */
 #define WCCP2_METHOD_GRE        0x00000001
@@ -372,17 +372,22 @@ struct router_identity_info_t {
     /* dynamic list of cache IP addresses */
 };
 
-/* The received packet for a mask assignment is unusual */
+/* The received packet for a mask assignment contains no mask elements */
 
 /** \interface WCCPv2_Protocol
  * Sect 5.7.7 Mask Element  ???
- * see code below. apparently the supposed IP address at position num1 can be equal to 3.
+ * Only applicable in case mask_element_count == 0
  */
 struct cache_mask_info_t {
     struct in_addr addr;
-    uint32_t num1;
-    uint32_t num2;
-    uint32_t num3;
+    uint16_t hash_revision;
+    uint16_t bits;
+//#define WCCP2_ASSIGNMENT_CURRENT    0x1
+//#define WCCP2_MASK_ASSIGNMENT_DATA  0x2
+    uint32_t mask_element_count;
+
+    uint16_t weight;
+    uint16_t status;
 };
 
 /** \interface WCCPv2_Protocol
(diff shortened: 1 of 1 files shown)


src/wccp2.cc
```diff
@@ -1476,6 +1486,9 @@ wccp2HandleUdp(int sock, void *)
 
                 SetField(cache_identity, ptr, router_view_header, router_view_size,
                          "malformed packet (truncated router view info cache w/o assignment hash)");
+                if (cache_identity->bits & 6 >> 1 != 0) {
+                    fatalf("Assignment type mismatch in router view");
```
yadij, line 1490: This needs to throw instead. The core problem behind CVE 2021-28116 was letting values supplied by routers abort Squid.

src/wccp2.cc
```diff
@@ -1488,25 +1501,33 @@ wccp2HandleUdp(int sock, void *)
 
                 SetField(cache_mask_info, ptr, router_view_header, router_view_size,
                          "malformed packet (truncated router view info cache w/o assignment mask)");
+                if (cache_mask_info->bits & 6 >> 1 != 1) {
+                    fatalf("Assignment type mismatch in router view");
```
yadij, line 1505: throw instead

src/wccp2.cc
```diff
@@ -1488,25 +1501,33 @@ wccp2HandleUdp(int sock, void *)
...
+                    if (ntohl(cache_mask_identity->mask.number_values) == 0) {
+                        // Need to set entire field
+                        SetField(cache_mask_identity, ptr, router_view_header, router_view_size,
+                             "malformed packet (truncated router view info cache mask)");
 
-                    ptr += sizeof(struct wccp2_cache_mask_identity_info_t);
+                        ptr += sizeof(struct wccp2_cache_mask_identity_info_t);
+                        cache_list_ptr->weight = ntohs(cache_mask_identity->weight);
+                    } else {
+                        fatalf("Unexpected cache assignment info number of mask values received\n");
```
yadij, line 1523: throw instead

src/wccp2.cc
```diff
@@ -1488,25 +1501,33 @@ wccp2HandleUdp(int sock, void *)
...
+                    }
 
-                    memcpy(&cache_address, &cache_mask_identity->addr, sizeof(struct in_addr));
                 } else {
-
-                    ptr += sizeof(struct cache_mask_info_t);
-
-                    memcpy(&cache_address, &cache_mask_info->addr, sizeof(struct in_addr));
+                    fatalf("Unexpected mask cache assignment info element count received\n");
```
yadij, line 1527: throw instead

yadij (changes requested):

yadij:
OK to test

src/wccp2.cc
amk1969: My understanding of the problem was that a broken/fake packet could access memory beyond its size.
Initially I used Must3 like at other places, but the outcome was not good (throw after some struct updated caused list of caches corrupted leading to segfault during processing of next packet) or infinite loop of negotiation attempt and ignored message, like the current situation.

I preferred to let the cache stop entirely over continuing with wccp disabled. It should happen during initial setup only. If that is not a good approach, I would also reinitialize wccp completely and start negotiation from scratch.

rousskov: The code parsing received traffic should not reach fatal(). In the vast majority of cases, when throwing breaks things, it is not throwing that should be blamed -- the bug is elsewhere (and should be fixed elsewhere). Moreover, IIRC, this code already throws, even before this PR changes. Thus, even if this PR avoids throwing, the post-throwing problems you have observed will probably continue to exist; you are not really eliminating them.

> It should happen during initial setup only

If not today, then tomorrow that "initial setup" event might happen ten days after the Squid instance has started (e.g., as a result of a reconfiguration).

> I would also reinitialize wccp prior exception and force negotiation from scratch.

Please handle exceptions when/after catching them, not when/prior to throwing them. Adding a `try/catch` may be necessary -- I have not checked.

yadij: Ah. The design of the logic is that a thrown exception when handling a packet aborts handling only that packet. Any information taken out of it should only be saved after validation of that piece of information.

Since WCCP sends a bunch of settings Squid is able to accept one piece (eg return method) while rejecting others (eg router list) provided the one used is valid regardless of the rejected piece. When there is any dependency situation between packet values they should be kept in local variables until it is all validated. Not added to the operational config as separate updates.

yadij:
@amk1969 please indicate whether you are able to do the requested changes. If not I will pick this up in the next few days and push towards merge. We will still need you to assist with testing the result either way. Thank you.

amk1969:
> If not I will pick this up in the next few days and push towards merge.
I am unable to allocate time for coding. It would be nice if you could implement the change needed, testing will be possible.

Thanks in advance

yadij:
@amk1969, I have finally done this. Since I am reluctant to touch your repository "master" branch I have opened PR #970 with the combo of our parts.

Please test #970 when you can.
 * If you encounter any segmentation faults our [wiki](https://wiki.squid-cache.org/SquidFaq/BugReporting) has info on how to obtain stack traces even if your proxy needs to be running in production.
 * If you encounter WCCP handshake/announce loops I am interested in a packet dump of what those looping packets contain. There may be another part of the parser needing this same treatment to preserve sanity of the info in Squid.

amk1969:
Comments from testing added you #970.

PR #957, opened 2022-01-05
Bug 5132: Close the tunnel if to-server conn closes after client

rousskov:
Since commit 25d2603, blind CONNECT tunnel "jobs" (and equivalent) were
not destroyed upon a "lonely" to-server connection closure, leading to
memory leaks. And when a from-client connection was still present at the
time of the to-server connection closure, we did not try to reforward,
violating the spirit of commit 25d2603 changes. Calling retryOrBail() is
sufficient to handle both cases.

### DIFF
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -318,6 +318,8 @@ void
 TunnelStateData::serverClosed()
 {
     server.noteClosure();
+
+    retryOrBail(__FUNCTION__);
 }
 
 /// TunnelStateData::clientClosed() wrapper
@@ -417,9 +419,6 @@ TunnelStateData::checkRetry()
 void
 TunnelStateData::retryOrBail(const char *context)
 {
-    // Since no TCP payload has been passed to client or server, we may
-    // TCP-connect to other destinations (including alternate IPs).
-
     assert(!server.conn);
 
     const auto *bailDescription = checkRetry();


rousskov:
Disclaimer: This change has not been tested yet. I am posting this Draft PR for those who are following Bug 5132 to test.

src/tunnel.cc
```diff
@@ -417,9 +419,6 @@ TunnelStateData::checkRetry()
 void
 TunnelStateData::retryOrBail(const char *context)
 {
-    // Since no TCP payload has been passed to client or server, we may
-    // TCP-connect to other destinations (including alternate IPs).
-
```
rousskov, line 422: The above comment should be removed because retryOrBail() preconditions are not the ones listed in the comment. The checkRetry() call below determines when we can retry and when we have to bail. We should not try to re-document that method in its caller.

rousskov:
Removing "Draft" annotation from the PR after [successful](https://bugs.squid-cache.org/show_bug.cgi?id=5132#c29) initial deployment tests.

yadij (approved):

src/tunnel.cc
```diff
@@ -318,6 +318,8 @@ void
 TunnelStateData::serverClosed()
 {
     server.noteClosure();
+
+    retryOrBail(__FUNCTION__);
```
eduard-bagdasaryan, line 322: I agree that this fixes memory leaks. I think retries should not possible due to TunnelStateData::retriable [flag](https://github.com/squid-cache/squid/pull/489#discussion_r780835913) protection.

thomasjfox:
This pull request came right on time: We upgraded from squid 3.5.28 (..) to squid 5.3 mid-December 2021 and faced a regression with "hanging" RDP-over-HTTPS tunnels upon initial connection. I didn't even know one can tunnel RDP over HTTPS.

The root problem was that the RDP server is pretty busy and rejected connections often. squid 3.5 retried the connection while squid 5.3 results in "hanging connections" on the Windows 10 client side.

This PR fixed the regression we debugged for a few hours already. We tested again without the patch and the issue was back instantly.

Thanks a lot Alex! This unbroke inventory handling of a  public library.

PR #958, opened 2022-01-07
Bug 5134: assertion failed: Transients.cc:221: "old == e"

rousskov:
Make sure the StoreMap anchor we open for reading has our key (and not
just happens to be at our hash position). Prior to this change, two
openForReadingAt() calls were missing a sameKey() post-call check due to
a buggy backport (v5 commit ec50061; mis-attributed to me).

Now the sameKey() check is integrated into the openForReadingAt() method
itself, as was already done in master/v6 since commit b2aca62.

### DIFF
--- a/src/ipc/StoreMap.cc
+++ b/src/ipc/StoreMap.cc
@@ -102,7 +102,7 @@ Ipc::StoreMap::openOrCreateForReading(const cache_key *const key, sfileno &filen
 
     // start with reading so that we do not overwrite an existing unlocked entry
     auto idx = fileNoByKey(key);
-    if (const auto anchor = openForReadingAt(idx)) {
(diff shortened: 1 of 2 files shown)


rousskov:
Disclaimer: This fix has not been tested yet. I have shared it with bug [5134](https://bugs.squid-cache.org/show_bug.cgi?id=5134) followers that can reproduce the bug and, I hope, will test this fix.

rousskov:
> Disclaimer: This fix has not been tested yet. I have shared it with bug [5134](https://bugs.squid-cache.org/show_bug.cgi?id=5134) followers that can reproduce the bug and, I hope, will test this fix.

Initial tests were [successful](https://bugs.squid-cache.org/show_bug.cgi?id=5134#c38).

yadij (approved):

PR #959, opened 2022-01-09
Maintenance: De-duplicate PackableStreamBuf and SBufStreamBuf

rousskov:
SBufStreamBuf was almost identical to PackableStreamBuf. Both classes
are pass-through write-only streambufs that lack their own put area. Now
their common code lives in AppendingStreamBuf.

No functionality changes intended.

The removed std::streambuf method descriptions were imprecise and some
were misleading. STL documentation describes this standard/parent API.

While also similar, PackableStream and SBufStream have important
differences in API and implementation. For example, unlike
PackableStream, SBufStream owns the sink buffer and does not modify the
buffer it was created with, providing a safe content extraction method
instead. The two classes cannot be merged without changing their users,
and it may be impossible to justify SBufStream users exposure to the
dangers of forgetting to sync() the streambuf before accessing the sink.

### DIFF
--- a/src/base/PackableStream.h
+++ b/src/base/PackableStream.h
@@ -13,20 +13,18 @@
 
 #include <ostream>
 
-/**
- * Provides a streambuf interface for writing to Packable objects.
- * Typical use is via a PackableStream rather than direct manipulation
- */
-class PackableStreamBuf : public std::streambuf
+// TODO: Move to src/base/AppendingStreamBuf.h
+/// write-only std::streambuf that append()s all writes to a given Buffer
+template <class Buffer>
+class AppendingStreamBuf : public std::streambuf
 {
 public:
-    explicit PackableStreamBuf(Packable &p) : buf_(p) { buf_.buffer(); }
-    virtual ~PackableStreamBuf() = default;
+    explicit AppendingStreamBuf(Buffer &p): buf_(p) { postInit(); }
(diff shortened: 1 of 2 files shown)


src/base/PackableStream.h
```diff
@@ -13,20 +13,18 @@
 
 #include <ostream>
 
-/**
- * Provides a streambuf interface for writing to Packable objects.
- * Typical use is via a PackableStream rather than direct manipulation
- */
-class PackableStreamBuf : public std::streambuf
+// TODO: Move to src/base/AppendingStreamBuf.h
```
rousskov, line 16: This TODO can be in this PR scope. I did not create src/base/AppendingStreamBuf.h because it would hide/obscure code changes. I suggest doing that last step after the PR is otherwise ready for merging.

rousskov:
> GitHub: yadij added the `S-waiting-for-author` label [4 days ago](https://github.com/squid-cache/squid/pull/959#event-5864969367)

@yadij, I do not know why you added that label. Please clarify what you want me to do here!

Please note that GitHub does not notify folks working on the PR when PR labels are added. I would have asked earlier, but I did not know that you added that label.

yadij:
It was to resolve the Semaphoreci issues on 8051e7a. But they appear to have been fixed by the updates to sync with master. So this is good to merge.

PR #962, opened 2022-01-11
Fix GCC v5.5.0 build after 82fe21f

eduard-bagdasaryan:
Old problems revealed by recent commits:

    XactionRep.cc:174:20: error: declaration of 'String name' shadows
    a parameter [-Werror=shadow]

    XactionRep.cc:486:33: error: declaration of 'services' shadows
    a previous local [-Werror=shadow]


rousskov (approved):
Thank you for fixing these eCAP warnings.

yadij (approved):
Yay, less String.

PR #964, opened 2022-01-18
V5 next backports

squidadm:


yadij:
OK to test

src/security/ServerOptions.cc
```diff
@@ -410,6 +410,7 @@ Security::ServerOptions::updateContextConfig(Security::ContextPointer &ctx)
     updateContextClientCa(ctx);
 
 #if USE_OPENSSL
+    SSL_CTX_set_mode(ctx.get(), SSL_MODE_NO_AUTO_CHAIN);
```
rousskov, line 413: This change (commit 94ff35b) will break v5 deployments that rely on the current Squid v5 behavior (and some did that without knowing -- we discovered this bug by accident!). At the very least, this breaking change should be prominently disclosed.

I am not going to veto this v5 change because the number of deployments it will break ought to be small (possibly none because we will warn those folks that we know are relying on v5 behavior), and I want to continue to give v5 maintainer as much v5 decision freedom as possible, but please make sure you fully understand its consequences.

yadij: Thank you for that info. I will drop this from the backports until someone actively requests it.

PR #965, opened 2022-01-23
Use RegisteredRunners to initialize/clean the ESI module

yadij:


kinkie:
I support the change, just.. is there any reason why the singletons are in the top level and not static in the registry class itself?

yadij:
Because placing the static as a class member with nullptr default initialize gives:
```
error: ‘constexpr’ needed for in-class initialization of static data member ‘ESIParser::Register* EsiExpatRr::prExpat’ of non-integral type
```

src/esi/ExpatParser.cc
```diff
@@ -12,8 +12,28 @@
...
+    void finalizeConfig()
+    {
+        prExpat = new ESIParser::Register("expat", &ESIExpatParser::NewParser);
+    }
+
+    void finishShutdown()
+    {
+      delete prExpat;
+      prExpat = nullptr;
+    }
```
rousskov, line 32: Same problem as in https://github.com/squid-cache/squid/pull/954#discussion_r772620417. Please remove this override (and adjust the remaining code accordingly).

Assuming ESIParser::Register destructor itself is safe/working, the (adjusted) cleanup code should go into the EsiExpatRr destructor AFAICT.

src/esi/Makefile.am
```diff
@@ -12,6 +12,11 @@ noinst_LTLIBRARIES = libesi.la
 
 ESI_PARSER_SOURCES =
 
+##
+## List in reverse order of preference.
+## The last parser to initialize will be used as default.
```
rousskov, line 17: We cannot guarantee/control static initialization order across translation units. AFAICT, in the current PR code, the static initialization order[^1] determines the order of the following two calls:

1. ESIParser::Register("expat", &ESIExpatParser::NewParser)
2. ESIParser::Register("libxml2", &ESILibxml2Parser::NewParser)

The fact that (some specific) initialization order is usually _preserved_ across builds on a given platform is irrelevant.

[^1]: The order of items in (presumably) the `ESI_PARSER_SOURCES` "list" in this Makefile is not necessarily the (static) initialization order.

src/esi/ExpatParser.cc
```diff
@@ -12,8 +12,28 @@
 
 #if USE_SQUID_ESI && HAVE_LIBEXPAT
 
+#include "base/RunnersRegistry.h"
 #include "esi/ExpatParser.h"
 
+static ESIParser::Register *prExpat = nullptr;
```
rousskov, line 18: Please make prExpat a (non-static) EsiExpatRr data member. We do not need to add this static and this global.

src/esi/ExpatParser.cc
```diff
@@ -12,8 +12,28 @@
...
+
+class EsiExpatRr : public RegisteredRunner
```
rousskov, line 20: Please place this new class inside the Esi namespace, adjusting its name accordingly. I hope that moving the corresponding RunnerRegistrationEntry() declaration/macro into the same Esi namespace will make that namespace-unaware macro work correctly.

rousskov (changes requested):
Thank you for migrating ESI modile initialization/cleanup to the RegisteredRunners API. The PR implementation bug and other problems (detailed in the inlined change requests) should be easy to address.

I also adjusted the PR title/description to better match the changes. Please adjust further as needed, keeping in mind that we should not re-document why we should be using RegisteredRunners for module initialization/cleanup.

yadij:
> 
> I also adjusted the PR title/description to better match the changes. Please adjust further as needed, keeping in mind that we should not re-document why we should be using RegisteredRunners for module initialization/cleanup.

So you have removed the documentation about scope and "why" question you keep insisting I publish. Then place on my shoulders a whole pile of extra behaviour changes to how ESI initialization and shutdown operate which were explicitly out of scope.

src/esi/Makefile.am
yadij: FTR; This comment is retained more as an advisory about implicit behaviour than a guarantee of which parser will be used.

src/esi/ExpatParser.cc
yadij: Done in d28e46a

src/esi/ExpatParser.cc
yadij: Done in d28e46a

src/esi/ExpatParser.cc
yadij: Done in d28e46a. And updated description to highlight that this is a significant change to shutdown behaviour.

rousskov:
> So you have removed the documentation about scope 

I do not think so, but please feel free to adjust further as needed.

> and "why" question you keep insisting I publish.

IMO, we do not need to document the obvious, especially when documenting it correctly is not trivial. IMO, my version of the PR title/description document this PR (including the "why"s) better than the old PR description did. As I said, there is no need to (attempt to) redocument why RegisteredRunners are used to initialize/clean Squid modules -- any capable developer would either already know that information (from her past Squid experience) or can easily figure it out (from the readily-available RegisteredRunners documentation and the diff).

> Then place on my shoulders a whole pile of extra behaviour changes to how ESI initialization and shutdown operate

No behavior changes (compared to the official Squid) have been requested. The PR was buggy and unpolished. I asked you to fix the bug and polish. If you think that I am asking you to change behavior (compared to the official Squid)), you have misunderstood the change requests.

src/esi/Makefile.am
rousskov: > FTR; This comment is retained

FTR: This comment is not "retained". The comment _text_ was _moved_ from the context where it could have been actually correct (at the very least, it made sense) to the context where its most natural interpretation would mislead the reader that the listed order controls which parser is going to be used by default.

> more as an advisory about implicit behaviour than a guarantee of which parser will be used.

If this is a comment about Squid behavior (i.e. Squid would use the last parser listed as the default parser when given a choice), then it is lying (i.e. when given a choice, Squid may use a different default parser instead). This Makefile list does not control implicit Squid behavior.

> I mean, this is _our_ preference of library. Vendors will link against their own choices.

If this comment is about Squid Project preference (whatever that means), then it should be rephrased to detail what that "preference" means (to the point where a reader can actually understand what we are talking about) and, more importantly, to remove the implication that the listed order actually controls which parser will be used as the default parser (when more then two are available).

rousskov:
> PR description v3: Cleanup has been shifted from shutdown to process exit.

What makes you think that?

src/esi/Makefile.am
rousskov: However, once the initialization order dependency bug is fixed, the comment should probably go back to the initialization code, where the old text will make sense again -- no pressing need to change the original comment text if its "original habitat" is effectively restored:

```C++
    // register in reverse order of preference.
    // The latest registered parser will be used as default.
```

yadij:
> > PR description v3: Cleanup has been shifted from shutdown to process exit.
> 
> What makes you think that?

The code. You requested that the cleanup current code does on shutdown be moved to runner deconstructor (program exit).

What makes you think I'm wrong?

```
            RunRegisteredHere(RegisteredRunner::finishShutdown);
            enter_suid();
            masterExit();
```
and
```
    RunRegisteredHere(RegisteredRunner::finishShutdown);
    memClean();
    exit(shutdown_status);
```

rousskov:
>>> PR description v3: Cleanup has been shifted from shutdown to process exit.

>> Alex: What makes you think that?

> Amos: The code. You requested that the cleanup current code does on shutdown be moved to runner deconstructor

Yes, there is a way to interpret the above as a true (albeit misleading) statement. The destructor _is_ where regular runners process the shutdown event, so I was only asking you to move the code from the method that probably should not exist in its current form (as I have mentioned earlier) to the destructor, but both functions are executed during the same shutdown event processing (as far as regular runners are concerned).

> (program exit).

No, this part is your (incorrect) assumption. Regular runners are destroyed before exit() and, more importantly for this discussion, before the memClean() call (if any).

 
> What makes you think I'm wrong?

There are many ways to reach that conclusion, and I do not remember exactly how I did it, but here is one possibility: The runners are dynamically-created objects -- they do not get automagically destroyed at program exit. Yet, their destructors are obviously used/called. Thus, we have code that deletes them _explicitly_, at some well-defined point. It would be rather foolish to place that destruction point _after_ cleaning memory pools that the modules may use, of course. A quick documentation check then finds this in the RegisteredRunner "Shutdown events" section, near the destructor I suggested using:

```C++
/// Called after stopping the main loop and before releasing memory.
/// Meant for quick/basic cleanup that does not require any other modules.
virtual ~RegisteredRunner() {}
```

The actual `delete` statement is fairly easy to find in RunnersRegistry.cc. At that point, all is left is to confirm where the RegisteredRunner::finishShutdown event is triggered in main.cc.

yadij:
Oh, you are right.

Since this is a repeating confusion. Perhapse it would be best to just use finishShutdown() for consistency with the other registry hooks, rather than forcing actions into the destructors.

src/esi/Makefile.am
yadij: Please clarify what you mean by "the initialization order dependency bug".

The closest thing to a bug I see with ESI initiailization order, is that they do not report clearly which one will be used in absence of admin configuration. IIRC, this "default order" was an arbitrary maintainer decision based on which library had/has the most popularity and least security issues last time it was reviewed.

rousskov: > Please clarify what you mean by "the initialization order dependency bug".

Assuming both libraries are enabled/configured and `esi_parser` is set to `auto`, The ESIParser::NewParser() result in this PR is determined by static initialization order across two different translation units (i.e. by compiler, linker, and/or similar external and possibly random factors) rather than by Squid source code. More specifically,

* In the official code, the "auto" parser is ESIExpatParser[^1].
* In the PR code, it is impossible to reliably predict what the "auto" parser is based on the Squid source code tarball alone. That decision is no longer made by Squid code. It is made by static initialization order across translation units which is outside of our control. This is a bug.

[^1]: Yes, I know that cf.data.pre and convoluted Modules.cc documentation say it should be ESILibxml2Parser instead. If I am right, then this is a bug in the official code, introduced by 57b6788. If I am wrong, it will be ESILibxml2Parser instead. However, as far as _this_ PR is concerned, the bug is _not_ in the specific value returned by ESIParser::NewParser() when `esi_parser` is set to `auto`. I am not asking you to fix any old bugs here! The new bug introduced by this PR is that the return value became unpredictable, dependent on static initialization order across translation units, random, not controlled by Squid sources, whatever you want to call it.

rousskov: And see the [initial](https://github.com/squid-cache/squid/pull/965#discussion_r790299237) change request text for the calls which order was randomized by this PR.

yadij: Hmm, I see. You are right there is a bug in the official code. I am have opened bug 5192 to fix that, when that is gone this change will be unnecessary.

rousskov: > when [Bug 5192] is gone this change will be unnecessary

That depends on how Bug 5192 is fixed. Minimal/surgical fixes of Big 5192 (e.g., replacing front() with back() in ESIParser::NewParser()) would not address _this_ change request because this change request is about a different bug introduced in this PR:

* Bug [5192](http://bugs.squid-cache.org/show_bug.cgi?id=5192): Documented order is the reverse of the actual order.
* This PR [bug](https://github.com/squid-cache/squid/pull/965#discussion_r790299237): The actual order is random/unpredictable/global-initialization-dependent/whatever-you-want-to-call-it.

rousskov: > Minimal/surgical fixes of Big 5192 would not address this change request

In fact, I see no practical way to leave this PR code "as is" after _any_ Bug 5192 fix, no matter how advanced. Some changes will be needed here. What kind of changes depends on how Bug 5192 is fixed.

yadij: see PR #968.

rousskov: I reviewed PR 968. As I [expected](https://github.com/squid-cache/squid/pull/965#discussion_r791172205), that PR does not address this change request. After PR 968 lands, addressing this change request would require a complete removal of the PR-added comment this change request is attached to.

src/esi/ExpatParser.cc
```diff
@@ -12,8 +12,30 @@
...
+class ExpatRr : public RegisteredRunner
+{
+public:
+    void finalizeConfig()
+    {
+        prExpat.reset(new ESIParser::Register("expat", &ESIExpatParser::NewParser));
+    }
+
+private:
+    std::unique_ptr<ESIParser::Register> prExpat;
```
rousskov, line 32: We should not use unique_ptr for objects _shared_ with others. This Register object (i.e. the ESI parser registration object) is shared with the ESIParser registry (i.e. the std::list container accessible via ESIParser::GetRegistry()).

If you would like to clean this up, I can detail how to do that. However, this cleanup can be legitimately viewed as being outside this PR scope -- the official code has essentially the same flaw (unrelated to the module initialization/cleanup sequence itself) -- so I cannot insist on it.

src/esi/Libxml2Parser.cc
```diff
@@ -16,8 +16,30 @@
...
 
+#include "base/RunnersRegistry.h"
 #include "esi/Libxml2Parser.h"
 
+#include <memory>
+
+namespace Esi
+{
+
+class Xml2Rr : public RegisteredRunner
```
rousskov, line 27: AFAICT, this class should be called Libxml2Rr for consistency sake.

src/esi/Libxml2Parser.cc
```diff
@@ -16,8 +16,30 @@
...
+{
+public:
+    void finalizeConfig()
+    {
+        prLibxml.reset(new ESIParser::Register("libxml2", &ESILibxml2Parser::NewParser));
+    }
+
+private:
+    std::unique_ptr<ESIParser::Register> prLibxml;
```
rousskov, line 36: This data member should be called prLibxml2 for consistency sake. However, to avoid "foo inside Foo" noise it is better to rename this member and the matching ExpatRr::prExpat member to `registration` or a similar library-neutral purpose-focused name.

rousskov (commented):
I only have optional polishing and can-be-viewed-as-out-of-scope suggestions left.

> Since this is a repeating confusion. Perhapse it would be best to just use finishShutdown() for consistency with the other registry hooks, rather than forcing actions into the destructors.

Removing finishShutdown() completely (as discussed earlier) will remove the repeated confusion and simplify the code overall, without requiring us to move current destructor bodies into finishShutdown(). Removing finishShutdown() would also avoid questions like "Do I put X into the destructor or finishShutdown()?". Thus, my knee jerk reaction continues to be "let's remove finishShutdown() if we can", but I need more time to answer your out-of-scope question _fully_. Please do not hesitate to remind me about it if I forget -- we do not have a good way of tracking the state of such architectural issues.

src/esi/ExpatParser.cc
yadij: Left out of scope.

src/esi/Libxml2Parser.cc
yadij: Done in cbc3ab2

src/esi/Libxml2Parser.cc
yadij: Done in cbc3ab2

PR #969, opened 2022-01-26
Bug 5194: Remove all unused debug sections

rousskov:
... that we can find quickly.

Naming/documenting debugging sections is a good idea, but that should
not be done in every source code file that relies on that section. We
cannot remove all such DEBUG: declarations without developing a proper
way to name/document sections, but we can (and, given Bug 5194
existence, probably should) remove the unused ones -- their removal does
not remove any immediately usable info.

Bug 5194 report was specific to Section 56: Folks misinterpret section
56 "HTTP Message Body" title as promising to dump message bodies to
cache.log, which is not a functionality that should be driven by a
debugging section. Currently, Squid lacks such functionality.

### DIFF
--- a/doc/debug-sections.txt
+++ b/doc/debug-sections.txt
@@ -109,14 +109,10 @@ section 54    Windows Interprocess Communication
 section 55    HTTP Header
 section 57    HTTP Status-line
 section 58    HTTP Reply (Response)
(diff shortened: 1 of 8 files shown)


hno (approved):
Trivial. Looks good.

PR #970, opened 2022-01-28
Bug 5179 part 2

yadij:
Followup to PR 953, using exceptions instead of fatal().

The safe use of exceptions required some refactoring to reduce scope of temporary variables and std::unique_ptr to destruct temporary memory in a clean manner.

Refactoring to STL types and even better list management is left out of scope. Initial review implies that would touch a lot of other WCCP logic. This is (attempting, approximately) minimal change for the bug fix.

Todo:
 * test and find runtime issues
 * audit the WCCP2_I_SEE_YOU router list parse for similar issues


src/wccp2.cc
```diff
@@ -1136,15 +1140,17 @@ CheckSectionLength(const void *sectionStart, const size_t sectionLength, const v
...
-CheckFieldDataLength(const FieldHeader *header, const size_t dataLength, const void *areaStart, const size_t areaSize, const char *error)
+CheckFieldDataLength(const FieldHeader *header, const size_t headerLength, const size_t dataLength, const void *areaStart, const size_t areaSize, const char *error)
 {
     assert(header);
-    const auto dataStart = reinterpret_cast<const char*>(header) + sizeof(header);
+    const auto dataStart = reinterpret_cast<const char*>(header) + headerLength;
     CheckSectionLength(dataStart, dataLength, areaStart, areaSize, error);
-    return sizeof(header) + dataLength; // no overflow after CheckSectionLength()
+    return headerLength + dataLength; // no overflow after CheckSectionLength()
 }
```
rousskov, line 1154: This code may work, but this change needs refactoring because it is a significant step backwards in the level of safety offered by this safety-focused method. It is better to break WCCP completely than to switch back to "working" WCCP with less safe code IMO. The current PR code may be 100% correct, but there is no good reason to to significantly increase the probability of future safety mistakes, similar to those we recently fixed.

This buggy function was designed to check (header, data) protocol fields or elements. In terms of the protocol, both "router capability" and "router capability element" are such elements. The official code _correctly_ (and naturally) assumes that FieldHeader is a field _header_. Converting that function parameter into a "this is a field header in some cases and something else in other cases so please try to remember to supply the correct field header length in all cases" is not a good solution.

AFAICT, the bug this PR needs to fix is in router_capability_element declaration. Here are the existing declarations that, once you strip decorations, show the problem:

```C++
struct wccp2_item_header_t { // A TLV header field
    uint16_t type;
    uint16_t length;
};

struct wccp2_capability_info_header_t { // A TLV field header
    uint16_t capability_info_type;
    uint16_t capability_info_length;
    /* dynamic length capabilities list */ // No, the list is not inside the header, but we all know what the author meant.
};

struct wccp2_capability_element_t { // A field itself!
    uint16_t capability_type;
    uint16_t capability_length;
    uint32_t capability_value;
};
```

I think we can fix the bug we are trying to fix without introducing a lot of cascading changes by adjusting the types along these lines:

```C++
struct wccp2_capability_element_header_t {
    uint16_t capability_type;
    uint16_t capability_length;
};

struct wccp2_capability_element_t: public wccp2_capability_element_header_t {
    /// convenience wrapper to access the whole header
    wccp2_capability_element_header_t *header() { return this; }
    uint32_t capability_value;
};

...

const auto elementSize = CheckFieldDataLength(
    router_capability_element.header(), ntohs(router_capability_element->capability_length),
    ...);
```

There are much better fixes, of course, but they will require a lot more code changes.

After the above is done, we should also add a static assert to the original CheckFieldDataLength(), to catch similar bugs in the future:

```C++
// A weak check that FieldType is a header rather than the whole field (or something else).
// TODO: Extract the notion of a field header into a dedicated type!
static_assert(sizeof(*header) == 4, "FieldHeader is a typical WCCPv2 TLV header");
```

After the above is done, we should also fix the sizeof(header) expressions inside the original CheckFieldDataLength(). They should read sizeof(*header) instead AFAICT. Another well-hidden bug in this untested official code!..

After the above is done, there might be at least one more bug to fix here, but I would prefer to postpone that analysis until the above changes are enacted.

I can enact these changes myself if you prefer. Just let me know.

rousskov (changes requested):
This is a partial review to help you make progress.

src/wccp2.cc
yadij: > I can enact these changes myself if you prefer. Just let me know.

Please do. That would be faster than explaining the details.

rousskov: Done in 542b81a.

src/wccp2.cc
```diff
@@ -1398,6 +1402,14 @@ wccp2HandleUdp(int sock, void *)
...
 
+                // XXX: Here and elsewhere, we do not check that
+                // capability_value actually fits inside the area limited by
+                // router_capability_element->capability_length! For example, we
+                // might receive a bogus capability_length value of 1, which is
+                // smaller than sizeof(capability_value). We do check that
+                // router_capability_element fits into the area limited by
+                // router_capability_data_length a whole, but that check is too
+                // weak because there may be other elements in that area.
                 if (!(ntohl(router_capability_element->capability_value) & Config.Wccp2.return_method)) {
```
rousskov, line 1413: This is another problem we need to address.

src/wccp2.cc
```diff
@@ -1072,16 +1083,9 @@ wccp2ConnectionClose(void)
...
 
-                if (cache_list_ptr != &router_list_ptr->cache_list_head) {
-                    xfree(cache_list_ptr);
-                } else {
-
-                    memset(cache_list_ptr, '\0', sizeof(struct wccp2_cache_list_t));
-                }
-            }
+            delete router_list_ptr->cache_list_head.next;
+            memset(&router_list_ptr->cache_list_head, '\0', sizeof(struct wccp2_cache_list_t));
```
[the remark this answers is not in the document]
rousskov: BTW, I get lots of compiler warnings related (AFAICT) to this change request. I forgot to copy the build log, but I believe GCC did not like us memset()-ing non-trivial objects either.

rousskov (changes requested):

amk1969:
I testing this code against a Cisco router. There are some troubles... With hash assignment, one cache is ignored and a broken assignment packet is sent to router. This change fixes it, not sure how to propose with github...

`--- a/src/wccp2.cc
+++ b/src/wccp2.cc
@@ -1849,7 +1849,7 @@ wccp2AssignBuckets(void *)
                 if (num_caches) {
                     int cache;
 
-                    for (cache = 0, cache_list_ptr = &router_list_ptr->cache_list_head; cache_list_ptr->next; cache_list_ptr = cache_list_ptr->next, ++cache) {
+                    for (cache = 0, cache_list_ptr = &router_list_ptr->cache_list_head; cache_list_ptr; cache_list_ptr = cache_list_ptr->next, ++cache) {
                         /* add caches */
`

For mask assignment, one cache/router works, but I fear that in case more get added one gets ignored as well. Need to run more instances, hopefully tomorrow.

BTW, would it be possible to make mask configurable and not hardcoded as 0x1741? Less bits set in the mask saves resources at the router side and makes smaller packets.

amk1969:
The diff from yesterday is not sufficient. Running two caches and two routers again results in malformed packet:

```
    Assignment Info
        Type: Assignment Info (6)
        Length: 300
        Assignment Key IP Address: 192.168.1.4
        Assignment Key Change Number: 4
        Number of Routers: 2
        Router IP: 192.168.2.1
        Router IP: 192.168.1.1
        Number of WC: 2
            WC IP: 192.168.1.4 id: 0
            WC IP: 0.0.0.0 id: 1
[Malformed Packet: WCCP]
```
The IP of second cache is not output, first 4 buckets are interpreted as IP address and there are not enough buckets. Also the buckets themselves are all 0, while half should be going to the other cache.

Further troubles noticed:
There is trouble with address comparison in hash mode. cache with IP 192.168.1.3 logged:
`2022/02/13 11:33:16.788 kid1| 80,5| wccp2.cc(1575) wccp2HandleUdp: I am not the lowest ip cache - not assigning buckets`
The other created bad assignment above.

In the mask assignment mode, the lower IP created it, but it sends different assignment to each router - one router is instructed to redirect everything to one cache, other to the other. This is not how it should work.

I also notice that the wccp assignment packets are created for each router separately. This can be optimized - actually all routers/caches should get the same wccp messages, the protocol supports also multicast.

In all HIA messages the web cache view contains Number of Web Caches=0. This does not lead to problem with operation, but it is confusing when checking the packets.

rousskov:
@amk1969, thank you for testing this and suggesting fixes! Clearly, a lot more needs to be done before this code becomes production-worthy. I do not plan to resume working on WCCP myself until summoned, but GitHub should notify you when there are new PR commits.

src/wccp2.cc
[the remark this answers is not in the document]
yadij: Fixed in aa85c0a547. Although there are more of the same still to resolve in later code.

yadij:
To make some more progress I have now rebased this PR on recent master branch and solved several of the compile issues being identified by GCC-12.

One GCC warning revealed bad bitmask handling of flags that would have resulted in attempted mask assignments being rejected as invalid or interpreted as hash assignment. Hash may have worked due to a fluke overlap of flags meanings. I have verified that the correct flag bits are being used now - but not able to test how this alters the protocol negotiation. This may be related to the strange router 0.0.0.0 appearance and WC=0 mentioned by https://github.com/squid-cache/squid/pull/970#issuecomment-1038020144 (eg mask assignment bucket(s) being interpreted as hash assignment buckets)

Three GCC errors remain all of type:
   "writing to an object of type ‘struct wccp2_cache_list_t’ with no trivial copy-assignment"
I hope to have more time to look into implementing copy assignment properly in a few days.

callumapplsys:
Updating the PR here as I noticed there was some activity on the thread I started on the squid-users mailing list. I was able to modify this patch to get it to build without the three GCC errors referring to the wccp2_cache_list_t struct, and talk successfully to the Cisco router, at least initially.

However I had to comment out lines 2584 to 2596 in wccp2.cc which is the "Swap if we need to" section of the wccp2SortCacheList function. I'm trying to understand why the cache list needs to be sorted to attempt to fix it. From section 3.9 of this WCCP spec, it suggests that the lowest IP cache is selected as the designated cache for a given service group. Is that what this code is for? Is it possible that given my single router setup I would not see the consequences of this change? See [https://datatracker.ietf.org/doc/html/draft-param-wccp-v2rev1-01#section-3.9](https://datatracker.ietf.org/doc/html/draft-param-wccp-v2rev1-01#section-3.9)

There are also two other issues I have noted so far in testing:
 * In doing a packet capture I can see the HERE_I_AM, and I_SEE_YOU messages as expected, but the REDIRECT_ASSIGNMENT message is being flagged as corrupted in Wireshark. It says "Malformed packet (exception occurred)". 
 * When exiting Squid there is an invalid pointer which causes it to abort:
```
2023/08/23 11:36:51| FD 12 Closing WCCPv2 socket
free(): invalid pointer
Aborted
```
Running on Ubuntu 20.04.6 LTS with a GRE tunnel to a Cisco 2821

PR #972, opened 2022-02-01
Moved regcomp(3)-specific RegexPattern code inside RegexPattern     Also moved non-ACL regex configuration parsing code inside ConfigParser. It is possible to move ACL (data) regex configuration parsing code as well, but that move is a lot more complex due to regex-pattern-joining optimizations, and there are no immediate plans to support non-regcomp(3) regexes in ACL data. We may do that move later as we get more experience with non-regcomp(3) regexes and decide to join them too.     These moves clean up existing regex-using code and allow adding support for non-regcomp(3) regexes (e.g., regexes based on C++11 <regex>) without adjusting ConfigParser::regex() callers. Such support would also require "virtualizing" RegexPattern. To avoid increasing complexity and hurting performance, that (simpler) step should be done only if we decide to actually add support for non-regcomp(3) regexes.

rousskov:
The above changes allowed us to improve RegexData error reporting: Squid
no longer reports REs _optimization_ failure when it is an individual RE
that is broken (and reported as such). Squid still ignores the fact that
broken REs can be "optimized" into a completely "different" valid
combined RE: We do not compile individual REs unless optimization fails.

Also simplified and polished affected code.

### DIFF
--- a/src/ConfigParser.cc
+++ b/src/ConfigParser.cc
@@ -9,6 +9,7 @@
 #include "squid.h"
 #include "acl/Gadgets.h"
 #include "base/Here.h"
+#include "base/RegexPattern.h"
 #include "cache_cf.h"
 #include "ConfigParser.h"
 #include "Debug.h"
@@ -477,17 +478,29 @@ ConfigParser::RegexStrtokFile()
     return token;
 }
 
-char *
-ConfigParser::RegexPattern()
+RegexPattern *
+ConfigParser::regex(const char *expectedRegexDescription)
 {
-    if (ConfigParser::RecognizeQuotedValues) {
-        debugs(3, DBG_CRITICAL, "FATAL: Can not read regex expression while configuration_includes_quoted_values is enabled");
-        self_destruct();
-    }
+    if (RecognizeQuotedValues)
+        throw TextException("Cannot read regex expression while configuration_includes_quoted_values is enabled", Here());
+
+    SBuf pattern;
+    int flags = REG_EXTENDED | REG_NOSUB;
+
     ConfigParser::RecognizeQuotedPair_ = true;
-    char * token = NextToken();
+    const auto flagOrPattern = token(expectedRegexDescription);
+    if (flagOrPattern.cmp("-i") == 0) {
+        flags |= REG_ICASE;
+        pattern = token(expectedRegexDescription);
+    } else if (flagOrPattern.cmp("+i") == 0) {
+        flags &= ~REG_ICASE;
+        pattern = token(expectedRegexDescription);
+    } else {
+        pattern = flagOrPattern;
+    }
     ConfigParser::RecognizeQuotedPair_ = false;
-    return token;
+
+    return new RegexPattern(pattern.c_str(), flags);
 }
 
 char *
--- a/src/ConfigParser.cc
+++ b/src/ConfigParser.cc
@@ -500,7 +500,7 @@ ConfigParser::regex(const char *expectedRegexDescription)
     }
     ConfigParser::RecognizeQuotedPair_ = false;
 
-    return std::unique_ptr<RegexPattern>(new RegexPattern(pattern.c_str(), flags));
+    return std::unique_ptr<RegexPattern>(new RegexPattern(pattern, flags));
 }
 
 char *
--- a/src/ConfigParser.cc
+++ b/src/ConfigParser.cc
@@ -478,7 +478,7 @@ ConfigParser::RegexStrtokFile()
     return token;
 }
 
-RegexPattern *
+std::unique_ptr<RegexPattern>
 ConfigParser::regex(const char *expectedRegexDescription)
 {
     if (RecognizeQuotedValues)
@@ -500,7 +500,7 @@ ConfigParser::regex(const char *expectedRegexDescription)
     }
     ConfigParser::RecognizeQuotedPair_ = false;
 
-    return new RegexPattern(pattern.c_str(), flags);
+    return std::unique_ptr<RegexPattern>(new RegexPattern(pattern.c_str(), flags));
 }
 
 char *
--- a/src/RefreshPattern.h
+++ b/src/RefreshPattern.h
@@ -93,9 +93,9 @@ private:
 };
 
 inline std::ostream &
-operator <<(std::ostream &os, const RefreshPattern &rp)
+operator <<(std::ostream &os, const RefreshPattern &r)
 {
-    rp.printHead(os);
+    r.printHead(os);
     return os;
 }
 
--- a/src/RefreshPattern.h
+++ b/src/RefreshPattern.h
@@ -11,6 +11,8 @@
 
 #include "base/RegexPattern.h"
 
+#include <memory>
+
 /// a representation of a refresh pattern.
 class RefreshPattern
 {
@@ -26,11 +28,15 @@ public:
      */
 #define REFRESH_DEFAULT_MAX static_cast<time_t>(259200)
 
-    RefreshPattern(const char *aPattern, const decltype(RegexPattern::flags) &reFlags) :
-        pattern(reFlags, aPattern),
+    using RegexPointer = std::unique_ptr<RegexPattern>;
+
+    // If given a regex, becomes its owner, creating an explicit refresh_pattern
+    // rule. Otherwise, creates an implicit/default refresh_pattern rule.
+    explicit RefreshPattern(RegexPointer regex):
         min(0), pct(0.20), max(REFRESH_DEFAULT_MAX),
         next(NULL),
-        max_stale(0)
+        max_stale(0),
+        regex_(std::move(regex))
     {
         memset(&flags, 0, sizeof(flags));
     }
@@ -42,9 +48,7 @@ public:
             delete t;
         }
     }
-    // ~RefreshPattern() default destructor is fine
 
-    RegexPattern pattern;
     time_t min;
     double pct;
     time_t max;
@@ -72,7 +76,28 @@ public:
         uint64_t matchCount;
         // TODO: some stats to indicate how useful/less the flags are would be nice.
     } stats;
+
+    /// reports configuration excluding trailing options
+    void printHead(std::ostream &) const;
+
+    /// reports the configured pattern or a fake pattern of the implicit rule
+    void printPattern(std::ostream &os) const;
+
+    // TODO: Refactor external refresh_pattern rules iterators to make private.
+    /// configured regex; do not use except when iterating configured rules
+    const RegexPattern &regex() const;
+
+private:
+    /// configured regex or, for the implicit refresh_pattern rule, nil
+    RegexPointer regex_;
 };
+
+inline std::ostream &
+operator <<(std::ostream &os, const RefreshPattern &rp)
+{
+    rp.printHead(os);
+    return os;
+}
 
 #endif /* SQUID_REFRESHPATTERN_H_ */
 
--- a/src/acl/RegexData.cc
+++ b/src/acl/RegexData.cc
@@ -243,13 +243,18 @@ ACLRegexData::parse()
     }
 
     try {
+        // ignore the danger of merging invalid REs into a valid "optimized" RE
         compileOptimisedREs(data, sl, flagsAtLineStart);
     } catch (...) {
-        debugs(28, DBG_IMPORTANT, "WARNING: optimisation of regular expressions failed; using fallback method without optimisation" <<
-               Debug::Extra << "configuration: " << cfg_filename << " line " << config_lineno << ": " << config_input_line <<
-               Debug::Extra << "optimisation failure: " << CurrentException);
-
         compileUnoptimisedREs(data, sl, flagsAtLineStart);
+        // Delay compileOptimisedREs() failure reporting until we know that
+        // compileUnoptimisedREs() above have succeeded. If
+        // compileUnoptimisedREs() also fails, then the compileOptimisedREs()
+        // exception caught earlier was probably not related to _optimization_
+        // (and we do not want to report the same RE compilation problem twice).
+        debugs(28, DBG_IMPORTANT, "WARNING: Failed to optimize a set of regular expressions; will use them as-is instead;" <<
+               Debug::Extra << "configuration: " << cfg_filename << " line " << config_lineno << ": " << config_input_line <<
+               Debug::Extra << "optimization error: " << CurrentException);
     }
 }
 
--- a/src/acl/RegexData.cc
+++ b/src/acl/RegexData.cc
@@ -125,7 +125,7 @@ compileREs(std::list<RegexPattern> &curlist, const SBufList &RE, int flags)
  * The ultimate goal is to have only one RE per ACL so that match() is
  * called only once per ACL.
  */
-static int
+static void
 compileOptimisedREs(std::list<RegexPattern> &curlist, const SBufList &sl, const int flagsAtLineStart)
 {
     std::list<RegexPattern> newlist;
@@ -195,8 +195,6 @@ compileOptimisedREs(std::list<RegexPattern> &curlist, const SBufList &sl, const 
         debugs(28, (opt_parse_cfg_only?DBG_IMPORTANT:2), "WARNING: there are more than 100 regular expressions. " <<
                "Consider using less REs or use rules without expressions like 'dstdomain'.");
     }
-
-    return 1;
 }
 
 static void
@@ -216,7 +214,8 @@ compileUnoptimisedREs(std::list<RegexPattern> &curlist, const SBufList &sl, cons
             } catch (...) {
                 // TODO: Make these configuration failures fatal (by default).
                 debugs(28, DBG_CRITICAL, "ERROR: Skipping regular expression: '" << configurationLineWord << "'" <<
-                       Debug::Extra << "compilation failure: " << CurrentException);
+                       Debug::Extra << "configuration: " << cfg_filename << " line " << config_lineno << ": " << config_input_line <<
+                       Debug::Extra << "regex compilation failure: " << CurrentException);
             }
         }
     }
@@ -247,7 +246,9 @@ ACLRegexData::parse()
         compileOptimisedREs(data, sl, flagsAtLineStart);
     } catch (...) {
         debugs(28, DBG_IMPORTANT, "WARNING: optimisation of regular expressions failed; using fallback method without optimisation" <<
+               Debug::Extra << "configuration: " << cfg_filename << " line " << config_lineno << ": " << config_input_line <<
                Debug::Extra << "optimisation failure: " << CurrentException);
+
         compileUnoptimisedREs(data, sl, flagsAtLineStart);
     }
 }
--- a/src/acl/RegexData.cc
+++ b/src/acl/RegexData.cc
@@ -104,28 +104,21 @@ removeUnnecessaryWildcards(char * t)
     return t;
 }
 
-/// XXX: This was returning false on regcomp() failures, but it now throws.
-static bool
+static void
 compileRE(std::list<RegexPattern> &curlist, const SBuf &RE, int flags)
 {
-    if (RE.isEmpty())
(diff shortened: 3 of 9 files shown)


src/acl/RegexData.cc
```diff
@@ -264,9 +242,19 @@ ACLRegexData::parse()
         }
     }
 
-    if (!compileOptimisedREs(data, sl, flagsAtLineStart)) {
-        debugs(28, DBG_IMPORTANT, "WARNING: optimisation of regular expressions failed; using fallback method without optimisation");
+    try {
+        // ignore the danger of merging invalid REs into a valid "optimized" RE
```
rousskov, line 246: This source code comment applies to the old/official code as well -- this PR does not affect the existing tradeoff between configuration speed and correctness: We (still) choose to ignore the fact that broken REs can be "optimized" into a completely different[^1] valid combined RE:

    compileOptimisedREs: adding RE 're)?((1'
    compileOptimisedREs: adding RE '2))+(re'
    RegexPattern: (re)?((1)|(2))+(re)
    compileOptimisedREs: 2 REs are optimised into one RE.

[^1]: It is impossible to meaningfully compare with broken REs, but we can assume that the combined version was _not_ the REs author intent!

src/acl/RegexData.cc
```diff
@@ -113,45 +104,28 @@ removeUnnecessaryWildcards(char * t)
     return t;
 }
 
-static bool
-compileRE(std::list<RegexPattern> &curlist, const char * RE, int flags)
+static void
+compileRE(std::list<RegexPattern> &curlist, const SBuf &RE, int flags)
 {
-    if (RE == NULL || *RE == '\0')
-        return curlist.empty(); // XXX: old code did this. It looks wrong.
```
rousskov, line 120: I removed "XXX: old code did this. It looks wrong" code/comments during this refactoring. The comment (added in commit e2b7452) can be considered correct: Out of the 3-4 situations where aggregated REs can be empty, the "old" odd-looking code handled only cases 1 and 4.

1. "-i/+i" at the beginning of the ACL data: No accumulated REs yet.
2. "-i/+i" after a buffering/joining overflow emptied accumulated REs.
3. `regex1 +i -i regex2` and similar sequences: Accumulated REs (regex1) are emptied upon seeing "+i". Squid then tries to compile still-empty accumulated REs upon seeing the "-i".
4. An empty RE token. This may not be possible with recent RE parsers.

PR code does not call compileREs() with an empty list of REs, handling situations 1-3. I let the regex library handle the empty (zero-length) RE token (case 4) instead of treating it as a special case. Again, I doubt that use case is actually real.

src/base/RegexPattern.h
```diff
@@ -22,26 +23,42 @@ class RegexPattern
...
 public:
     RegexPattern() = delete;
-    RegexPattern(int aFlags, const char *aPattern);
+    RegexPattern(const SBuf &aPattern, int aFlags);
     ~RegexPattern();
 
-    // regex type varies by library, usually not safe to copy
-    RegexPattern(const RegexPattern &) = delete;
-    RegexPattern &operator =(const RegexPattern &) = delete;
+    RegexPattern(RegexPattern &&) = delete; // no copying of any kind
```
rousskov, line 29: I removed these unused (and slightly problematic) RegexPattern move optimizations to simplify. Technically, calling POSIX regfree() with a zeroed regex was a UB because regfree() is only defined for compiled regexes. Since these move optimizations were actually unused (and are not expected to be used), we should just delete this (slightly) problematic code path.

rousskov (commented):
@yadij, this PR implements many of the "changes to the internal regex support in Squid" that we discussed on [squid-dev](http://lists.squid-cache.org/pipermail/squid-dev/2022-January/009760.html) in the context of adding support for regexes based on STL `<regex>`. The remaining (still unimplemented) parts (including RegexPattern "virtualization") should wait for the decision to actually add that support.

The ball is on your side for that squid-dev thread (there are two specific questions at the end of the last [email](http://lists.squid-cache.org/pipermail/squid-dev/2022-January/009760.html)), but this PR is an improvement even if we never add support for STL regexes. It does not need to wait for that STL regex decision.

src/acl/RegexData.cc
```diff
@@ -197,19 +173,18 @@ compileOptimisedREs(std::list<RegexPattern> &curlist, const SBufList &sl, const
...
-    if (!compileRE(newlist, accumulatedRE, flags))
-        return 0;
-
-    accumulatedRE.clear();
-    reSize = 0;
+    if (!accumulatedRE.empty()) {
+        compileREs(newlist, accumulatedRE, flags);
+        accumulatedRE.clear();
+        reSize = 0;
+    }
```
rousskov, line 187: After the cleanup in this PR, this code duplication became even more obvious. It is possible to refactor the code further to avoid the code duplication surrounding compileREs() calls, but that duplication existed before this PR, and such refactoring would require a lot of changes unrelated to this PR, so it is best done in a dedicated PR. We are not making things worse here.

src/acl/RegexData.cc
yadij: Case 4 can occur when the parser state implies that a 0-length string is the token for pattern. Real-world situations usually involve multi-file configs with the values and ordering spread out, so not as obviously silly as this test case:
```
configuration_includes_quoted_values on
acl regex dstdomain_regex ""
refresh_pattern "" 0 0 0
```

A unit test to check the library behaviour seems warranted if you don't want to add logic to detect and throw a config error on such madness.

src/acl/RegexData.cc
```diff
@@ -113,45 +104,28 @@ removeUnnecessaryWildcards(char * t)
     return t;
 }
 
-static bool
-compileRE(std::list<RegexPattern> &curlist, const char * RE, int flags)
+static void
+compileRE(std::list<RegexPattern> &curlist, const SBuf &RE, int flags)
```
yadij, line 108: Now that you have moved the internals away and added compileREs() for all the callers there is no need for this static function. Please inline it to the (one?) caller.

src/acl/RegexData.cc
yadij: Question is *which* code duplication. I recall some of it being intentional, though not that exact reasons. If you take on the de-dupe work be careful.

src/cache_cf.cc
```diff
@@ -2713,13 +2713,11 @@ static void
...
 {
     while (head != NULL) {
-        storeAppendPrintf(entry, "%s%s %s %d %d%% %d",
-                          name,
-                          head->pattern.flags&REG_ICASE ? " -i" : null_string,
-                          head->pattern.c_str(),
-                          (int) head->min / 60,
-                          (int) (100.0 * head->pct + 0.5),
-                          (int) head->max / 60);
+        SBufStream os;
```
yadij, line 2716: "entry" is a Packable object. Please use PackableStream instead of adding temporary SBuf allocations.

src/base/RegexPattern.h
```diff
@@ -22,26 +23,42 @@ class RegexPattern
...
 
-    RegexPattern(RegexPattern &&);
-    RegexPattern &operator =(RegexPattern &&);
+    /// whether the regex differentiates letter case
+    bool caseSensitive() const { return !(flags & REG_ICASE); }
+
+    /// Whether this is an "any single character" regex ("."). In some contexts,
+    /// that regex is (ab)used as a special "should match anything" default.
+    bool isDot() const { return pattern.length() == 1 && pattern[0] == '.'; }
```
yadij, line 36: FYI; In GNU regex there are implicit prefix/suffix wildcards in the absence of an anchor. The pattern (.) actually means "any 1-or-more characters". To match a specific character count requires anchors. i.e. the pattern (^.$) is what means "any single character"

Are there specific cases of "abuse" you can point at?

src/refresh.cc
```diff
@@ -693,12 +684,13 @@ refreshStats(StoreEntry * sentry)
     storeAppendPrintf(sentry, "\nRefresh pattern usage:\n\n");
     storeAppendPrintf(sentry, "  Used      \tChecks    \t%% Matches\tPattern\n");
     for (const RefreshPattern *R = Config.Refresh; R; R = R->next) {
-        storeAppendPrintf(sentry, "  %10" PRIu64 "\t%10" PRIu64 "\t%6.2f\t%s%s\n",
+        SBufStream os;
+        R->printPattern(os);
+        storeAppendPrintf(sentry, "  %10" PRIu64 "\t%10" PRIu64 "\t%6.2f\t" SQUIDSBUFPH "\n",
```
yadij, line 689: sentry is a Packable. You should be able to avoid all the SBuf and SBufStream allocations by using a PackableStream instead of storeAppendPrintf().

src/refresh.cc
```diff
@@ -727,6 +719,33 @@ refreshStats(StoreEntry * sentry)
...
+    return *regex_;
+}
+
+void
+RefreshPattern::printPattern(std::ostream &os) const
+{
+    if (regex_)
+        regex_->print(os, nullptr); // refresh lines do not inherit line flags
+    else
+        os << "<none>";
```
yadij, line 735: This function is used to dump the squid.conf settings. But I do not see anything in the new parser handling "<none>" as a special case pattern. 

I think this should be an assert(regex_) since refresh_pattern cannot be configured without a pattern and this function is for mgr:config dumping the squid.conf settings.

yadij (changes requested):

src/acl/RegexData.cc
rousskov: I do not want to work on handling empty regexes _specially_ at least until there is some evidence that it is possible to create an empty regex. I knew that empty configuration tokens are possible in certain contexts, but the example you sketched does not work because `configuration_includes_quoted_values on` cases are officially excluded from handling regexes:

```C++
if (RecognizeQuotedValues)
    throw TextException("Cannot read regex expression while configuration_includes_quoted_values is enabled", Here());
```

src/cache_cf.cc
```diff
@@ -2870,22 +2846,12 @@ parse_refreshpattern(RefreshPattern ** head)
...
-    }
-
-    if ((errcode = regcomp(&comp, pattern, flags)) != 0) {
-        char errbuf[256];
-        regerror(errcode, &comp, errbuf, sizeof errbuf);
-        debugs(22, DBG_CRITICAL, "" << cfg_filename << " line " << config_lineno << ": " << config_input_line);
-        debugs(22, DBG_CRITICAL, "ERROR: refreshAddToList: Invalid regular expression '" << pattern << "': " << errbuf);
-        xfree(pattern);
-        return;
+            debugs(22, DBG_CRITICAL, "ERROR: Unknown refresh_pattern option: " << token);
```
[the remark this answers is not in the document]
rousskov: > We are going with the non-sentence casing now right?

Not for level-0/1 messages, especially those that start with an all-caps label followed by a colon. Should I try to summarize the few existing rules so that we can polish them together (outside of this PR)?

I believe the casing in PR is better because it matches many American English [styles](https://www.tckpublishing.com/capitalization-after-colons/), and we try to follow American English variants rather than the British ones.

src/acl/RegexData.cc
rousskov: > there is no need for this static function

There is a need: If we inline it, we would duplicate its logic. The logic is simple, but there is no good reason to duplicate it. The compiler will probably inline it into both callers (in optimized builds), and that is fine.

P.S. Just to avoid misunderstanding: I did not add compileREs(). I renamed one of the two old compileRE() into compileREs(). I can undo that renaming if you insist, but the code will become more difficult to read/navigate than the current PR code because the context does not always make it clear which of the two same-name functions is being called.

P.P.S. Neither of the two static functions should probably exist as such, but refactoring all this code to properly use class methods is not trivial and is outside this PR scope.

src/RefreshPattern.h
```diff
@@ -72,7 +76,28 @@ class RefreshPattern
...
     } stats;
+
+    /// reports configuration excluding trailing options
+    void printHead(std::ostream &) const;
+
+    /// reports the configured pattern or a fake pattern of the implicit rule
+    void printPattern(std::ostream &os) const;
+
+    // TODO: Refactor external refresh_pattern rules iterators to make private.
+    /// configured regex; do not use except when iterating configured rules
```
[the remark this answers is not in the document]
rousskov: The correct refactoring here would probably _not_ result in any friend functions, but that refactoring is quite complex and is outside this PR scope. That iterating code should be changed quite a bit; it is currently rather convoluted and fragile. This PR does not change that, and such changes would be out of this PR scope: This PR is about hiding regcomp()-related RegexPattern internals inside RegexPattern. It is not about improving RefreshPattern code quality. I only touched RefreshPattern code that I had to touch for in-scoppe reasons, and I stayed away from any serious refactoring of out-of-scope code.

src/acl/RegexData.cc
```diff
@@ -230,15 +203,20 @@ compileUnoptimisedREs(std::list<RegexPattern> &curlist, const SBufList &sl, cons
...
         } else if (configurationLineWord == plus_i) {
             flags &= ~REG_ICASE;
         } else {
-            if (!compileRE(curlist, configurationLineWord.c_str(), flags))
-                debugs(28, DBG_CRITICAL, "ERROR: Skipping regular expression. "
-                       "Compile failed: '" << configurationLineWord << "'");
+            try {
+                compileRE(curlist, configurationLineWord, flags);
+            } catch (...) {
+                // TODO: Make these configuration failures fatal (by default).
```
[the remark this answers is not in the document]
rousskov: Changing Squid behavior in a way that affects (probably many!) existing deployments is outside this PR scope. I am very glad you agree with this `TODO`, but it deserves a dedicated PR.

rousskov (commented):
I believe I addressed all open change requests. Please re-review.

src/RefreshPattern.h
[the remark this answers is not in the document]
yadij: Sorry if I wasn't clear. I meant just using "friend" to keep this addition out of the RefreshPattern public API, reducing the comment warnings and notes. eg RefreshPattern should be touched *less* in this PR scope.

Just a nit. I do not insist on this.

src/acl/RegexData.cc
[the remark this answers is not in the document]
yadij: IMO these types of TODO should be implemented ASAP to start bringing them to peoples attention as early as possible. It reduces the remaining lifespan of broken configs.

Your call though. I do not insist on this.

src/acl/RegexData.cc
yadij: I disagree on the utility of de-duplicating one-liners. The argument could be used to de-duplicate the lines calling the wrapper function, and that wrapper function, and that wrapper function...

Your call though.

src/base/RegexPattern.h
yadij: We seem to be in agreement on the technical details. That clears up my worry about the technical side, thank you. The below I do not insist on:

My issue is documenting it as "abuse" to use it as a default. IMO the existing DefaultRule is the broken part - it matches non-existent URLs and ACL values.

IMO that second sentence should be dropped and (if anything) a TODO added to DefaultRule.

src/refresh.cc
yadij: Not quote what I asked for. But the rest can be out of scope.

src/refresh.cc
yadij: My worry remains. Though I will trust you on this one and see what happens.

yadij (commented):
Reviewed the changes. Code is good enough. Just three of the existing conversations to conclude.

src/RefreshPattern.h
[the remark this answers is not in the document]
rousskov: Your change request was clear. I was talking about proper refactoring. As for a quick "friends" hack, I think that adding two friends and changing one static function into a global one (to make it a friend without triggering [problems](https://stackoverflow.com/a/40937193)) is about the same as making this method public because this method can only be used (publicly) by those iterating/friend functions -- there is just no other way to access configured RefreshPattern objects.

Thank you for not insisting on this change.

src/acl/RegexData.cc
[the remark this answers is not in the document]
rousskov: > IMO these types of TODO should be implemented ASAP

I very much agree! Are you OK with regex compilation errors leading to a hard-coded configuration rejection (i.e. this code will throw)? If yes, I will be happy to volunteer a followup PR to do that. If you think we should deprecate first and/or add a configuration option to bypass this check, then I may have to wait for others to do that extra (and, IMO, unnecessary in the master branch) work.

src/base/RegexPattern.h
rousskov: I dropped the second sentence in afba89c00083223961305981e7fa2d0f28a5886e

I am pretty sure it is abuse (as I tried to prove but apparently failed), but the sentence is not critical. My primary motivation for documenting this abuse was to reduce the temptation to reuse this method or this approach in new code.

I agree that the DefaultRule is also problematic. However, it is not really being abused IMO -- we just lack a feature to make those default refresh parameters configurable explicitly (i.e. without using a `.` regex that Squid then need to find and treat specially).

rousskov:
> Reviewed the changes. Code is good enough. Just three of the existing conversations to conclude.

I believe I made the requested change for [one](https://github.com/squid-cache/squid/pull/972#discussion_r799143394) of the remaining three change requests. The [other](https://github.com/squid-cache/squid/pull/972#discussion_r799139804) [two](https://github.com/squid-cache/squid/pull/972#discussion_r799138211) are not blocking (AFAICT), and I prefer to leave them as is for now. Details in the corresponding change request threads.

Please re-review.

src/acl/RegexData.cc
rousskov: It is not really the number of lines that counts. It's what is _inside_ those line(s) that does. De-duplicating names is usually pointless (and would lead to the endless de-duplication loop you are thinking about). Virtually anything else is fair game! Every case should be evaluated on its own merits, of course, but, when in doubt, we should not duplicate. I have seen numerous times how seemingly trivial de-duplication improvements exposed or fixed serious bugs.

And, as far as this specific duplication is concerned: Adding something to an _ordered_ collection is a known source of bugs -- you and me had to deal with two of emplace_back() bugs in the past few weeks alone! We do not want to duplicate that kind of code!

src/acl/RegexData.cc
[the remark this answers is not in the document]
yadij: > > IMO these types of TODO should be implemented ASAP
> 
> I very much agree! Are you OK with regex compilation errors leading to a hard-coded configuration rejection (i.e. this code will throw)? 

yes.

> If yes, I will be happy to volunteer a followup PR to do that.

I can go with that.

yadij (approved):

src/ConfigParser.cc
```diff
@@ -477,17 +478,29 @@ ConfigParser::RegexStrtokFile()
...
+    int flags = REG_EXTENDED | REG_NOSUB;
+
     ConfigParser::RecognizeQuotedPair_ = true;
-    char * token = NextToken();
+    const auto flagOrPattern = token(expectedRegexDescription);
+    if (flagOrPattern.cmp("-i") == 0) {
+        flags |= REG_ICASE;
+        pattern = token(expectedRegexDescription);
+    } else if (flagOrPattern.cmp("+i") == 0) {
+        flags &= ~REG_ICASE;
```
eduard-bagdasaryan, line 496: We could drop this line since REG_ICASE is unset by default.

rousskov: You are right, of course, but let's keep it. I think the nearby code becomes a tiny bit clearer that way and, perhaps more importantly, less error-prone to seemingly unrelated modifications (e.g., passing initial `flags` value from the outside). Eventually, this code may become a loop (so that multiple options are detected/supported) and having this explicit flags manipulation may help with that conversion.

src/acl/RegexData.cc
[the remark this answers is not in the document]
rousskov: Done in #979.

PR #974, opened 2022-02-05
Improve coredump_dir on FreeBSD and Solaris based OS

devnexen:
Disclose that coredump_dir may also make the process "traceable",
including enabling core dumps and ptrace(2) attachments in certain
environments.

Add support for making the process traceable in FreeBSD- and
Solaris-like environments.

When Squid does not know how to make the process traceable in a given
environment, it does nothing under the assumption that the process is
already traceable enough (by default) to dump cores. Alternatively, we
could warn the admin about the lack of tracing support for that
environment. It is not clear which option is better, but we are betting
on the processes being traceable by default in most not explicitly
covered environments.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -5639,6 +5639,12 @@ DOC_START
 	that exists, Squid will chdir() to that directory at startup
 	and coredump files will be left there.
 
+	Linux, FreeBSD, Solaris/Illumos:
+
+	Note on supported platforms, the processes permissions are updated
+	to be able to be traced in order to be able to generate a core
+	file.
+
 CONFIG_START
 
 # Leave coredumps in the first cache dir
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -5642,9 +5642,7 @@ DOC_START
 	In addition to changing the directory, the process permissions are updated
 	to enable process tracing and/or coredump file generation. The details are
 	OS-specific, but look for prctl(2) PR_SET_DUMPABLE and procctl(2)
-	PROC_TRACE_CTL documentation as guiding examples. If Squid does not know
-	how to update relevant process permissions in your build environment, it
-	logs an ENOSYS-based ERROR message.
+	PROC_TRACE_CTL documentation as guiding examples.
 
 CONFIG_START
 
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -5639,11 +5639,12 @@ DOC_START
 	that exists, Squid will chdir() to that directory at startup
 	and coredump files will be left there.
 
-	Linux, FreeBSD, Solaris/Illumos:
-
-	Note on supported platforms, the processes permissions are updated
-	to be able to be traced in order to be able to generate a core
-	file.
+	In addition to changing the directory, the process permissions are updated
+	to enable process tracing and/or coredump file generation. The details are
+	OS-specific, but look for prctl(2) PR_SET_DUMPABLE and procctl(2)
+	PROC_TRACE_CTL documentation as guiding examples. If Squid does not know
+	how to update relevant process permissions in your build environment, it
+	logs an ENOSYS-based ERROR message.
 
 CONFIG_START
 
--- a/src/tools.cc
+++ b/src/tools.cc
@@ -279,23 +279,46 @@ rusage_pagefaults(struct rusage *r)
 #endif
 }
 
+/// Make the process traceable if possible. Call setTraceability() instead!
 /// Traceable processes may support attachment via ptrace(2) or ktrace(2),
 /// debugging sysctls, hwpmc(4), dtrace(1) and core dumping.
-static int
+static void
 makeTraceable(void)
 {
+    const auto handleError = [](const char * const syscall, const int savedErrno) {
+        throw TextException(ToSBuf(syscall, " failure: ", xstrerr(savedErrno)), Here());
+    };
 #if HAVE_PRCTL && defined(PR_SET_DUMPABLE)
-	return prctl(PR_SET_DUMPABLE, 1);
+    if (prctl(PR_SET_DUMPABLE, 1) != 0)
+        handleError("prctl(PR_SET_DUMPABLE)", errno);
 #elif HAVE_PROCCTL && defined(PROC_TRACE_CTL)
 	// TODO: when FreeBSD 14 becomes the lowest version, we can
 	// possibly save one getpid syscall, for now still necessary.
 	int traceable = PROC_TRACE_CTL_ENABLE;
-	return procctl(P_PID, getpid(), PROC_TRACE_CTL, &traceable);
+    if (procctl(P_PID, getpid(), PROC_TRACE_CTL, &traceable) != 0)
+        handleError("procctl(PROC_TRACE_CTL_ENABLE)", errno);
 #elif HAVE_SETPFLAGS
-	return setpflags(__PROC_PROTECT, 0);
+    if (setpflags(__PROC_PROTECT, 0) != 0)
+        handleError("setpflags(__PROC_PROTECT)", errno);
 #else
-	return ENOSYS;
+    debugs(50, 2, "WARNING: Assuming this process is traceable");
 #endif
+}
+
+/// Make the process traceable if necessary. \sa makeTraceable()
+static void
+setTraceability()
+{
+    // for now, setting coredump_dir is required to make the process traceable
+    if (!Config.coredump_dir)
+        return;
+
+    try {
+        makeTraceable();
+    } catch (...) {
+        debugs(50, DBG_IMPORTANT, "ERROR: Cannot make the process traceable:" <<
+               Debug::Extra << "exception: " << CurrentException);
+    }
 }
 
 void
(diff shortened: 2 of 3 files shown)


src/comm/Tcp.cc
```diff
@@ -15,6 +15,9 @@
 #if HAVE_NETINET_TCP_H
 #include <netinet/tcp.h>
 #endif
+#if HAVE_NETINET_IN_H
+#include <netinet/in.h>
+#endif
```
yadij, line 20: This change also does not appear to be related to the rest of this patch. If this is a second issue being fixed please submit it separately.

compat/proc.h
```diff
@@ -0,0 +1,39 @@
...
+#endif
+#if HAVE_SYS_PROCCTL_H
+#include <sys/procctl.h>
+#endif
+#if HAVE_PRIV_H
+#include <priv.h>
+#endif
+
+inline int
+ProcSetTraceable() {
```
yadij, line 23: This function does not match up with the intended use of libcompat as a substitute source of missing API symbols. eg PR_SET_DUMPABLE and prctl() themselves in this case.

I like your implementation better than what would be needed to fit here in libcompat. So my request is that you move this to a static function in src/tools.cc instead of extending libcompat.

yadij (changes requested):

src/CpuAffinitySet.cc
```diff
@@ -38,7 +38,7 @@ CpuAffinitySet::apply()
     } else {
         cpu_set_t cpuSet;
         memcpy(&cpuSet, &theCpuSet, sizeof(cpuSet));
-        (void) CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
+        CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
```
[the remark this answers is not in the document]
devnexen: I know but it does not build otherwise and not sure how fast it would be merged if I do a separate PR.

src/comm/Tcp.cc
devnexen: ditto.

compat/proc.h
devnexen: ok.

src/CpuAffinitySet.cc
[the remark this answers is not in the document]
yadij: That is not a valid reason to combine changes.

Yes, the QA work to figure out why that ```(void)``` exists and test on OS other than yours will take some amount of time. That has to be done to accept this change regardless of which PR it is part of. Merging several changes just slows the whole bunch to the speed of the most controversial part.

yadij:
Marking this PR as a feature addition since it adds previously unsupported/missing core dumping. It will need a doc/release/release-6.sgml note about the squid.conf coredup_dir behaviour change.

yadij (approved):
LGTM.

yadij:
OK to test

rousskov (changes requested):
Thank you for improving support on these platforms. Just a few polishing and documentation requests from me. Most are optional. Please let me know if you prefer that I enact them myself.

> Enable coredump_dir on FreeBSD and Solaris

The documented, primary functionality of `coredump_dir` is already "enabled" on those OSes -- it is not OS-specific. Please adjust the title to something like "Improve coredump_dir support on ..."

yadij:
Actually, kicking this back to @devnexen since the [requested](https://github.com/squid-cache/squid/pull/974#discussion_r800069544) src/cf.data.pre text is still missing even though GH says that request is outdated

src/cf.data.pre
```diff
@@ -5639,6 +5639,13 @@ DOC_START
 	that exists, Squid will chdir() to that directory at startup
 	and coredump files will be left there.
 
+	In addition to changing the directory, the process permissions are updated
+	to enable process tracing and/or coredump file generation. The details are
+	OS-specific, but look for prctl(2) PR_SET_DUMPABLE and procctl(2)
+	PROC_TRACE_CTL documentation as guiding examples. If Squid does not know
+	how to update relevant process permissions in your build environment, it
+	logs an ENOSYS-based ERROR message.
```
rousskov, line 5647: It feels wrong to log an ERROR message when Squid did not fail to update permissions. In the current `return ENOSYS` case, Squid does not even know whether such an update is necessary in the current environment! Should not that be a WARNING (or just a "for your information" status/progress message) instead?

rousskov (changes requested):
PR code just needs minor fixes/polishing, but I would like to resolve one question _before_ changing the code. Please see the inlined change request for details.

I also committed two documentation updates. Please make sure you are OK with them (or adjust the PR further as needed to make you happy).

src/cf.data.pre
devnexen: I find a ERROR a bit "overreacting" indeed, this is more a icing on a cake.

rousskov: Glad we agree! I will wait for @yadij opinion before proceeding with the final(?) adjustments.

yadij: In earlier review I assumed there was some importance you knew of to bump the message up to level DBG_IMPORTANT and label as ERROR.

If this is just a nice-to-have behaviour leaving it as level-2 with a WARNING should be sufficiently high. I do not have a preference between WARNING or NOTICE.

rousskov: > In earlier review I assumed there was some importance you knew of to bump the message up to level DBG_IMPORTANT and label as ERROR.

Your assumption was correct. In fact, I explicitly [stated](https://github.com/squid-cache/squid/pull/974#discussion_r800069683) the exact reason -- we should let the admin know (that there was a failure in applying their configuration directive). I do not consider level-2+ messages as "letting admin know". Level-2+ messages can be essential in triage, but they do not work well as admin notifications. Level-2+ is "pull", level-0/1 can be considered "push". Configuration failures deserve push notifications.

> If this is just a nice-to-have behaviour leaving it as level-2 with a WARNING should be sufficiently high.

Leaving failure reporting at level-2 does not address the [earlier](https://github.com/squid-cache/squid/pull/974#discussion_r800069683) concern, but this thread has enough information for me to be able to polish this. I will be back with specific changes shortly.

>  I do not have a preference between WARNING or NOTICE.

Whenever possible, for problem messages, we should be using one of the three standardized labels: `FATAL:`, `ERROR:`, `WARNING:`. See d816f28 for more information. That commit was about level-0/1 messages, but it is easy to adjust its rationale to become applicable to level-2 problem messages.

rousskov: > this thread has enough information for me to be able to polish this. I will be back with specific changes shortly.

Done. The primary changes are in 44c7c4d. I consider this change request addressed.

src/tools.cc
```diff
@@ -273,6 +279,49 @@ rusage_pagefaults(struct rusage *r)
...
+setTraceability()
+{
+    // for now, setting coredump_dir is required to make the process traceable
+    if (!Config.coredump_dir)
+        return;
+
+    try {
+        makeTraceable();
+    } catch (...) {
+        debugs(50, DBG_IMPORTANT, "ERROR: Cannot make the process traceable:" <<
```
devnexen, line 320: There can be some mixing messages, having WARNING and ERROR in same time ?

rousskov (approved):
I have adjusted PR code and description. See commit 44c7c4d for primary details.  Here is a sample cache.log error message:

    2022/02/09 12:14:29| ERROR: Cannot make the process traceable:
        exception: prctl(PR_SET_DUMPABLE) failure: (1) Operation not permitted
        exception location: tools.cc(289) operator()

Please review and adjust further if needed, but I am now happy to see this PR merged.

@devnexen, thank you for improving this code and exposing an important coredump_dir side effect!

src/tools.cc
rousskov: No, the level-2 `WARNING` code does not throw and, hence, is not going to trigger this level-1 `ERROR` message.

devnexen: ok

src/tools.cc
```diff
@@ -273,6 +279,50 @@ rusage_pagefaults(struct rusage *r)
...
+    // possibly save one getpid syscall, for now still necessary.
+    int traceable = PROC_TRACE_CTL_ENABLE;
+    if (procctl(P_PID, getpid(), PROC_TRACE_CTL, &traceable) != 0)
+        handleError("procctl(PROC_TRACE_CTL_ENABLE)", errno);
+#elif HAVE_SETPFLAGS
+    if (setpflags(__PROC_PROTECT, 0) != 0)
+        handleError("setpflags(__PROC_PROTECT)", errno);
+#else
+    debugs(50, 2, "WARNING: Assuming this process is traceable");
+    (void)handleError; // just "use" the variable; there is no error here
```
devnexen, line 305: [[maybe_unused]] would not work ?

rousskov: It [will](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused), but only after we upgrade to C++17. For now, Squid uses C++11 and `(void)`.

rousskov: And normally we do not add comments for such `(void)x` statements -- they are considered self-documenting. This one is a special case, as I tried to explain in 07b45b0.

devnexen: ah right I forgot it s only C++11...

rousskov:
@yadij, this PR is now on your side: There were significant changes after your earlier approval, so I do not want to merge it without giving you a chance to look over the current PR code/description...

yadij (approved):

PR #975, opened 2022-02-05
Fix FreeBSD 14 build

devnexen:
FreeBSD 14 defines 3-parameter CPU_AND() macro as a `do {} while` loop.
Our (void) in front of that loop creates a syntax error.

    CpuAffinitySet.cc:41:16: error: expected expression 
    (void) CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);

That (void) was added in commit 7ec6d51 to "remove GNU-specific syntax",
but we cannot tell what specific problem that 10-year old change solved.
Known 3-parameter CPU_AND(3) documentation says the call returns void.

Also included a missing header providing IPPROTO_TCP definition.

### DIFF
--- a/src/CpuAffinitySet.cc
+++ b/src/CpuAffinitySet.cc
@@ -38,7 +38,7 @@ CpuAffinitySet::apply()
     } else {
         cpu_set_t cpuSet;
         memcpy(&cpuSet, &theCpuSet, sizeof(cpuSet));
-        (void) CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
+        CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
         if (CPU_COUNT(&cpuSet) <= 0) {
             debugs(54, DBG_IMPORTANT, "ERROR: invalid CPU affinity for process "
                    "PID " << getpid() << ", may be caused by an invalid core in "
--- a/src/comm/Tcp.cc
+++ b/src/comm/Tcp.cc
@@ -15,6 +15,9 @@
 #if HAVE_NETINET_TCP_H
 #include <netinet/tcp.h>
 #endif
+#if HAVE_NETINET_IN_H
+#include <netinet/in.h>
+#endif
 #if HAVE_SYS_SOCKET_H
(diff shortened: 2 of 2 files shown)


yadij:
OK to test

yadij:
Please mention in the description the exact compiler errors happening with that CPU_AND.

yadij:
Oh, and which OS versions you are aware of having those build issues.

devnexen:
`CpuAffinitySet.cc:41:16: error: expected expression
        (void) CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
               ^
/usr/include/sys/cpuset.h:61:29: note: expanded from macro 'CPU_AND'
#define CPU_AND(d, s1, s2)              __BIT_AND2(CPU_SETSIZE, d, s1, s2)`

FreeBSD 14.

rousskov:
> Please mention in the description the exact compiler error(s) happening with that CPU_AND.

50% done, using @devnexen [answer](https://github.com/squid-cache/squid/pull/975#issuecomment-1030631122) as a starting point. @yadij, please see whether you agree with my PR description change -- it assumes that we do not know exactly why we added that `(void)`. Please adjust as needed, especially if you know that reason. Hopefully, that old build failure will not resurface -- the code will compile where it used to compile before this change. If it does resurface, we will be able to figure it out at that time.

I did not mention the Tcp.cc fix because @devnexen has not described the corresponding error. IMO, fixing both build errors in one PR is fine.

devnexen:
it s just needed for IPPROTO_TCP definition.

rousskov:
> it s just needed for IPPROTO_TCP definition.

I have updated PR description with that info. Thank you.

yadij:
From the description of https://github.com/squid-cache/squid/commit/7ec6d5177d3e128ee4a084e9e489a596a4016401 it looks to like one of the changes when we were adding clang support officially. Since Jenkins is currently reporting no clang issues with this PR I think we are good to accept.

yadij (approved):

rousskov (approved):

PR #978, opened 2022-02-08
Prep for 5.4.1

yadij:


rousskov (commented):
@yadij, I have no objections to this going in, but I am reluctant to approve this PR because 

* I do not think we need a special 5.4.1 release (i.e., we can just do a v5.5)
* I have not verified the change log details
* I believe some of the change log entries are not very admin-friendly/usable (and admins are or should be the primary change log audience)
* I think we should remove `HERE` from v5, and I do not see that going in
* I think we should fix recently botched v5 `CONTRIBUTORS`, and I do not see that going in 
* As you probably [know](https://github.com/squid-cache/squid/pull/946#issuecomment-1023460905), I think we should improve level-0/1 message labeling in v5 (#946), and I do not see that going in
* I think we should fix early message logging in v5 (#950), and I do not see that going in

I am not asking you to respond to any of the above concerns here -- most of these are just opinions or suggestions -- I am just documenting why I am reluctant to hit the Approve button.

Said that, if you need this merged on the fast track, please let me know and I will approve, disregarding the above reservations. As you know, I do not want to interfere with your v5 maintenance.

yadij:
>     * I have not verified the change log details
> 
>     * I believe some of the change log entries are not very admin-friendly/usable (and admins are or should be the primary change log audience)
> 

I would like your feedback on those entries. Particularly if I have missed anything. These were generated mostly from the bug 5055 commit message "also we fixed", "that meant we had to also" statements.

FYI: these are targeted both at admin (to be aware of which features in Squid are behaviourally changed), and at sysops developers (to be aware of their patches are affected).


>     * I think we should fix recently botched v5 `CONTRIBUTORS`, and I do not see that going in
> 

I am already working on the PR fro master where we can workout what "fixed" actually means. For now the v5 CONTRIBUTORS is an accurate record of what each contributor called themselves at the time of that contribution up to v5.

> 
> Said that, if you need this merged on the fast track, please let me know and I will approve, disregarding the above reservations. As you know, I do not want to interfere with your v5 maintenance.

I would like to fast-track. But only if someone is happy to signoff on an early release.

kinkie:
To me, it’s just a release name. Doesn’t sound like anything worth discussing much over, so I’m happy signing off

kinkie (approved):

yadij:
Thanks kinkie.

@rousskov, I will give it another day for your response. After that schedule it for Anubis unless you do so earlier.

rousskov:
> I am clearing this PR for merging.

... but I am not comfortable removing your request for @hno review, so the PR will not be auto-merged until Henrik reviews or you remove your review request.

PR #979, opened 2022-02-10
Do not skip problematic regexes in ACLs

rousskov:
This change has two partially overlapping parts:

* Reject configurations with ACLs containing failed-to-compile regexes.
* Do not skip ACL regexes with BUFSIZ or more characters.

Invalid or long ACL regexes were skipped (with an ERROR message),
resulting in a dangerous mismatch between admin (mis)interpretation of
their squid.conf and the actual configuration of the running instance.
Given the volume of ERRORs busy Squids are often reporting, including
transaction errors that admins may consider irrelevant, admins are not
paying enough attention to non-fatal configuration ERRORs, especially
reconfiguration ERRORs. Squid should reject erroneous configs instead.

And until we find (and this time document!) a specific reason to
artificially restrict individual regex length, let the Squid
configuration tokenizer and the regex library limit RE lengths. KISS.

Also deliver what commit 0fa036e promised but failed to do: "Squid no
longer reports REs _optimization_ failure when it is an individual RE
that is broken (and reported as such)". In that commit, I overlooked the
fact that Squid still ignored broken REs at the time, fooling the
higher-level code into thinking that they were OK (and triggering
unnecessary reporting of misleading optimization errors).

Besides rejection of invalid REs, there are two known side effects:

1. Squid may silently start using previously skipped long REs[^1].
2. Squid may fail/succeed/change regex optimization because the
   previously skipped long REs now participate in optimization attempts.
   In nearly all cases, this does not affect the ACL matching outcome.

[^1]: The regex library is unlikely to reject regexes based on their
length alone because libraries ought to accept regexes much longer than
longest Squid configuration tokens. For example, glibc accepts 4MB-long
regexes on 32-bit systems AFAICT based on a quick scan of regcomp code.

### DIFF
--- a/src/acl/RegexData.cc
+++ b/src/acl/RegexData.cc
@@ -209,14 +209,7 @@ compileUnoptimisedREs(std::list<RegexPattern> &curlist, const SBufList &sl, cons
         } else if (configurationLineWord == plus_i) {
             flags &= ~REG_ICASE;
         } else {
-            try {
-                compileRE(curlist, configurationLineWord, flags);
-            } catch (...) {
-                // TODO: Make these configuration failures fatal (by default).
-                debugs(28, DBG_CRITICAL, "ERROR: Skipping regular expression: '" << configurationLineWord << "'" <<
-                       Debug::Extra << "configuration: " << cfg_filename << " line " << config_lineno << ": " << config_input_line <<
-                       Debug::Extra << "regex compilation failure: " << CurrentException);
-            }
+            compileRE(curlist, configurationLineWord, flags);
         }
     }
 }
--- a/src/acl/RegexData.cc
+++ b/src/acl/RegexData.cc
@@ -233,13 +233,8 @@ ACLRegexData::parse()
     SBufList sl;
     while (char *t = ConfigParser::RegexStrtokFile()) {
         const char *clean = removeUnnecessaryWildcards(t);
-        if (strlen(clean) > BUFSIZ-1) {
-            debugs(28, DBG_CRITICAL, cfg_filename << " line " << config_lineno << ": " << config_input_line);
-            debugs(28, DBG_CRITICAL, "ERROR: Skipping regular expression. Larger than " << BUFSIZ-1 << " characters: '" << clean << "'");
-        } else {
-            debugs(28, 3, "buffering RE '" << clean << "'");
-            sl.emplace_back(clean);
-        }
+        debugs(28, 3, "buffering RE '" << clean << "'");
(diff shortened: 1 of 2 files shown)


rousskov:
FWIW, here is rough Squid changes history as far as treatment of long REs is concerned:

* Before 6564dae: RegexData accepts REs as long as the longest strtokFile() token. No RE-joining optimization yet.

* 6564dae: Introduces BUFSIZ restriction on individual REs, most likely because an on-stack array (called largeRE) that was added to join individual REs needed an aggressive size restriction. That array was added with a BUFSIZ capacity. BUFSIZ-long and larger REs are now rejected, even though compileUnoptimisedREs() still does not use that array/limit. An accident?

* d80c844: The largeRE array gets replaced with SBufList, allowing for a much larger buffer, but the commit _reduces_ REs accumulation limit from BUFSIZ (i.e. 4096) to 1024. Individual REs are still allowed to reach BUFSIZ-1 in length, so reducing accumulation limit below BUFSIZ looks odd. The change is not discussed (or even disclosed) in the commit message.

```diff
-        } else if (RElen + largeREindex + 3 < BUFSIZ-1) {
+        } else if (reSize < 1024) {
```

@kinkie, do you recall why d80c844 reduced RE accumulation limit from BUFSIZ (i.e. 4096) to 1024?

yadij (approved):
LGTM

rousskov:
> > PR description: Do not skip ACL regexes longer than 4095 characters

> Since BUFSIZ may be different, I am not sure we should make this statement. For example, I got it 8092 in RegexData.cc (defined stdio.h).

Good point! I did not know that `BUFSIZ` is usually defined _outside_ of Squid. A quick web search shows that it can even be as small as 512 bytes in some environments! 512 characters is still very long for a regex, but some folks could be using real-world URLs for regexes... This clarification makes the changes in this PR even more useful AFAICT.

I changed the PR description to use `BUFSIZ` instead of using the Squid-default value of `4096` (minus one).

rousskov:
... and since the BUFSIZ constant is actually [standard](https://en.cppreference.com/w/cpp/io/c) in C++, we should not be defining it like we do today, but that is a problem for another PR to fix. This PR just reduces the surface of that problem a bit.

kinkie:
> @kinkie, do you recall why [d80c844](https://github.com/squid-cache/squid/commit/d80c844681549b513c63b723041d453481821041) reduced RE accumulation limit from BUFSIZ ~(i.e. 4096)~ to 1024?

I don't remember; I think it was arbitrary, out of an abundance of caution

PR #982, opened 2022-02-11
mmappedio using the MAP_NOSYNC flag for FreeBSD/DragonFlyBSD

devnexen:
Allows to optimise the data being flushed to disk by not syncing in a regular basis
which can decrease the performance noticeably overtime.

### DIFF
--- a/src/DiskIO/Mmapped/MmappedFile.cc
+++ b/src/DiskIO/Mmapped/MmappedFile.cc
@@ -20,6 +20,9 @@
 #include <cerrno>
 #if HAVE_SYS_MMAN_H
 #include <sys/mman.h>
+#ifndef MAP_NOSYNC
+#define MAP_NOSYNC 0
+#endif
 #endif
 #if HAVE_SYS_STAT_H
 #include <sys/stat.h>
@@ -236,7 +239,9 @@ Mmapping::map()
     static const int pageSize = getpagesize();
     delta = offset % pageSize;
 
-    buf = mmap(NULL, length + delta, prot, flags, fd, offset - delta);
+    // on FreeBSD/DragonFlyBSD MAP_NOSYNC allows to optimize the data being
+    // flushed to the disk allowing to gain performance over time.
+    buf = mmap(NULL, length + delta, prot, flags | MAP_NOSYNC, fd, offset - delta);
 
     if (buf == MAP_FAILED) {
         const int errNo = errno;


rousskov:
@devnexen, in your environment, do you use this new mmap() flag with `cache_dir ufs` disk caches? What kind of performance improvement do you see from adding this option in your environment? Does your Squid deal primarily with large files or perhaps some other special/unusual circumstances?

yadij:
OK to test

devnexen:
> @devnexen, in your environment, do you use this new mmap() flag with `cache_dir ufs` disk caches? What kind of performance improvement do you see from adding this option in your environment? Does your Squid deal primarily with large files or perhaps some other special/unusual circumstances?

The benefits only shows with large files upon accesses where it could freeze due to syncing, with small files it s less obvious.

devnexen:
The key differences with the changes, with a file ~16 MB :
- more consistent access to the cache due to less frequent syncing to the disk, more apparent upon multiple clients. Without the changes I observed a decrease of 5/10% of requests per seconds.
- it is a bit of a trade-off, there is slightly more chances to get cache misses (I did not observe in my personal case but that is a possibility).

rousskov:
> more consistent access to the cache due to less frequent syncing to the disk

I assume that by "cache" you mean "Squid disk cache" rather than some OS cache.

What do you mean by "consistent access" in this context? Something like fewer sudden jumps in transaction response times, when Squid freeezes to dump dirty RAM pages to disk?

> Without the changes I observed a decrease of 5/10% of requests per seconds.

Noted, thank you. If the offered test load depends on the proxy, it is a red flag, but we probably do not get into that for now.


> it is a bit of a trade-off, there is slightly more chances to get cache misses

Can you explain why this mmap option affects cache hit ratio in your test? I do expect `MAP_NOSYNC` to represent certain trade-offs, but not the "better response time for the price of worse cache hit ratio" trade-off!

In your test, what happens to the large file after it is written to the Squid cache? Do clients usually revisit/access it? Or does it usually get deleted before it is revisited?

My current primary concern here is that this PR adds a `MAP_NOSYNC` option that appears to be designed for a _completely different_ mmap usage pattern than what most Squid ufs cache_dirs experience _and_ that, if my understanding of its side effects is correct, might make things worse in certain realistic use cases. I was hoping that your performance results would either correct my understanding or prove that the option does not have a measurable effect. Neither has happened so far.

> PR description: Allows to optimise the data being flushed to disk by not syncing in a regular basis which can decrease the performance noticeably overtime.

I do understand that  `MAP_NOSYNC` disables "regular" syncs. I also know that disk syncs can block/freeze a busy Squid when they are not frequent enough. You do not need to convince me that bad syncs are bad.

What I do not yet understand is why disabling regular syncs (and presumably getting highly irregular ones?!) would be a good thing in Squid `cache_dir ufs` context. There are certainly contexts where this option does wonders (e.g., a database that constantly updates some mmapped file using random accesses), but that does not seem to apply to how `cache_dir ufs` works -- Squid writes each `cache_dir ufs` file sequentially, from start to finish. If we are not talking about some very special Squid use case, that ufs file (i.e. cache entry) data has to be written to disk sooner or later -- not syncing it at all is not an option (unlike the database use case where the same in-memory block may be written and overwritten many times), right?

Can you explain why `MAP_NOSYNC` option would be unconditionally good for Squid, or are you approaching this more from a "I am not sure why or how this works, but it seems to help in my tests" position? Please note that I am not trying to criticize your approach (whatever it is) at this point! I am trying to understand what effect this option has in your tests, why it has that effect, and what effect we can expect in other realistic workloads.

devnexen:
Thanks for your reply.
> > more consistent access to the cache due to less frequent syncing to the disk
> 
> I assume that by "cache" you mean "Squid disk cache" rather than some OS cache.
> 
> What do you mean by "consistent access" in this context? Something like fewer sudden jumps in transaction response times, when Squid freeezes to dump dirty RAM pages to disk?
> 

Yes.
> > Without the changes I observed a decrease of 5/10% of requests per seconds.
> 
> Noted, thank you. If the offered test load depends on the proxy, it is a red flag, but we probably do not get into that for now.
> 
> > it is a bit of a trade-off, there is slightly more chances to get cache misses
> 
> Can you explain why this mmap option affects cache hit ratio in your test? I do expect `MAP_NOSYNC` to represent certain trade-offs, but not the "better response time for the price of worse cache hit ratio" trade-off!
> 
> In your test, what happens to the large file after it is written to the Squid cache? Do clients usually revisit/access it? Or does it usually get deleted before it is revisited?
> 
> My current primary concern here is that this PR adds a `MAP_NOSYNC` option that appears to be designed for a _completely different_ mmap usage pattern than what most Squid ufs cache_dirs experience _and_ that, if my understanding of its side effects is correct, might make things worse in certain realistic use cases. I was hoping that your performance results would either correct my understanding or prove that the option does not have a measurable effect. Neither has happened so far.
> 
> > PR description: Allows to optimise the data being flushed to disk by not syncing in a regular basis which can decrease the performance noticeably overtime.
> 
> I do understand that `MAP_NOSYNC` disables "regular" syncs. I also know that disk syncs can block/freeze a busy Squid when they are not frequent enough. You do not need to convince me that bad syncs are bad.
> 
> What I do not yet understand is why disabling regular syncs (and presumably getting highly irregular ones?!) would be a good thing in Squid `cache_dir ufs` context. There are certainly contexts where this option does wonders (e.g., a database that constantly updates some mmapped file using random accesses), but that does not seem to apply to how `cache_dir ufs` works -- Squid writes each `cache_dir ufs` file sequentially, from start to finish. If we are not talking about some very special Squid use case, that ufs file (i.e. cache entry) data has to be written to disk sooner or later -- not syncing it at all is not an option (unlike the database use case where the same in-memory block may be written and overwritten many times), right?

I have not considered it really, even tough despite its name MAP_NOSYNC does not mean never sync, however looking further at the freebsd code it seems more efficient with shared regions than files.
> 
> Can you explain why `MAP_NOSYNC` option would be unconditionally good for Squid, or are you approaching this more from a "I am not sure why or how this works, but it seems to help in my tests" position? Please note that I am not trying to criticize your approach (whatever it is) at this point! I am trying to understand what effect this option has in your tests, why it has that effect, and what effect we can expect in other realistic workloads.

PR #983, opened 2022-02-13
Fix build on Illumos

devnexen:


src/auth/basic/LDAP/basic_ldap_auth.cc
```diff
@@ -134,6 +134,10 @@ PFldap_start_tls_s Win32_ldap_start_tls_s;
 #include <lber.h>
 #include <ldap.h>
 
+#if _SQUID_SOLARIS_
```
rousskov, line 137: Should we generalize this to other "broken" platforms?

```suggestion
#ifndef LDAP_SECURITY_ERROR
```

rousskov (commented):
It would be nice to cite the actual build errors in the PR description so that others facing these errors can easily find this fix. If you have those compiler errors saved, please paste them there.

We can help with Markdown/commit [formatting](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message) if you would rather not deal with 72-character-lines ASCII-only restrictions and four-leading-spaces Markdown quoting.

src/auth/basic/LDAP/basic_ldap_auth.cc
devnexen: this is the only unix I stumbled across this but not against the change.

devnexen:
errors are long gone + I forgot if I compiled with clang or gcc, but it s for sure not previous declared function and undefined reference kind of errors.

rousskov:
I thought you have access to the build environment where you can easily reproduce the old errors and retest the latest fixes. If not, that is OK -- we do not _have_ to provide those errors in the commit message.

rousskov:
OK to test

rousskov (approved):

yadij (approved):

PR #984, opened 2022-02-15
SourceLayout: Move debugs()-related code into src/debug/

rousskov:
The new library is needed to facilitate debugs() reuse and, perhaps more
importantly, reuse of other Squid code that already uses debugs() -- all
without build-breaking hacks of copying debug.cc source files and very
dangerous stub use in production running code.

The Raw class, asHex(), and other I/O manipulators are not specific to
debugs() streams. They are now in src/base/.

Also removed test-suite/debug, addressing an XXX. That test was
misplaced and not really doing anything useful -- the compiler already
tests what test-suite/debug was testing (and much more).

Also removed the no longer necessary xassert() stub from test-suite/.

TODO: Adjust the moved code to use Debug namespace. Those adjustments
are likely to trigger debug/Stream.h reorganization, simplifying exposed
debugging API. They deserve a dedicated PR.

TODO: Convert more helpers to use the new debugging library. This change
only adjusts pinger because that adjustment did not require other
significant changes.


rousskov:
To simplify review, I split this PR into two commits. The first commit (b95c023) contains the primary changes. The second commit (1eff2bb) is just the necessary but boring and voluminous `#include` adjustments.

src/tests/stub_debug.cc
```diff
@@ -14,9 +14,9 @@
  * not abort the unit test.
  */
 #include "squid.h"
-#include "Debug.h"
+#include "debug/Stream.h"
```
yadij, line 17: Now that the object being stub'ed is a library there are a few changes relevant to this file to bring it inline with the other library stubs.

 * rename to stub_libdebug.cc and update related references to the filename, STUBS.list primarily.
 * all .h files in the library have an include statement below the STUB_API definition (not in the top #include list),
 * each library .h include is followed by the block of stubs for *all* its public API symbols.

test-suite/test_tools.cc
```diff
@@ -11,14 +11,6 @@
 
 #include "squid.h"
 #include "dlink.h"
-#include <iostream>
```
yadij, line 14: These iostream and xassert() removals do not seem relevant.  Please split into a separate PR or document in the description how its related.

src/LoadableModules.cc
```diff
@@ -7,8 +7,8 @@
  */
 
 #include "squid.h"
-#include "Debug.h"
-#include "DebugMessages.h"
+#include "debug/Stream.h"
+#include "debug/Messages.h"
```
yadij, line 11: Please run ./scripts/source-maintenance.sh on this branch. There are more than a few of these includes which should change their ordering with the new names.

yadij (changes requested):
Looks good. Just some polish. The biggest is updating the stubs file with the change to a library.

src/tests/stub_debug.cc
rousskov: We should not do that in this PR because the API exposed by libdebug may change quite a bit in the followup PR(s). If you insist on doing these changes now, I will probably waste time on stubbing APIs that will be changed or removed very soon. Let's revisit after the API-related changes are done in followup PRs. I will volunteer to update tests/stub_debug.cc at that time.

There is even a possibility that we will not need to stub libdebug at all after this PR series is complete.

src/LoadableModules.cc
rousskov: Commit f580a4c.

As you probably know, ./scripts/source-maintenance.sh is not stable enough (yet) for developers to run against their PR branches -- it produces too many unwanted changes, forcing the developer to waste time undoing them. We should not ask to run it.

src/Makefile.am
```diff
@@ -31,18 +31,36 @@ LOADABLE_MODULES_SOURCES = \
 	LoadableModules.cc \
 	LoadableModules.h
 
-SUBDIRS = mem base anyp helper dns ftp parser comm error eui acl format clients sbuf servers fs repl store DiskIO proxyp
-DIST_SUBDIRS = mem base anyp helper dns ftp parser comm error eui acl format clients sbuf servers fs repl store DiskIO proxyp
```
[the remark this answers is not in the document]
rousskov: Removed in commit c43da7f.

src/base/Raw.h
```diff
@@ -0,0 +1,121 @@
...
+    os << pd.label << '=';
+    if (pd.ptr)
+        return os << *pd.ptr;
+    else
+        return os << "[nil]";
+}
+
+/// std::ostream manipulator to print integers as hex numbers prefixed by 0x
+template <class Integer>
+class AsHex
```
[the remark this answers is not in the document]
rousskov: Move to base/IoManip.* in commit 543c949.

rousskov (commented):
I have addressed all your change requests except one that should be addressed later. Please re-review.

src/tests/stub_debug.cc
yadij: Okay. I can accept it as a followup.

yadij:
Updated the title label since code shuffling into convenience libraries is the ongoing https://wiki.squid-cache.org/Features/SourceLayout project.

src/LoadableModules.cc
yadij: That period of many unrelated changes should be over now as of this Jan enforcement cycle initiating the HERE purge.

IMO it is reasonable for a reviewer to ask any dev to run it if the PR submission contains a large amount of tiny style issues that it can erase without needing to review-comment specifics. Keeping the official code compliant is one of the things that falls on maintainer shoulders if these things get past review.

yadij (approved):

src/LoadableModules.cc
rousskov: > That period of many unrelated changes should be over now as of this Jan enforcement cycle initiating the HERE purge.

I was not talking about changes to ./scripts/source-maintenance.sh. I was talking about the lack of stability of that script _results_. They are unstable -- two executions of the script may (and do) edit two identical source files differently, resulting in unwanted formatting changes.

There are also problems with gperf and seasonal problems with copyright years. We are slowly fixing the script problems, but it is not yet ready for production use.

There is also a scope problem: Until we start rejecting malformed PRs, it is wrong to ask one developer to format code already deformed by recently merged PRs. The test enforcing PR formatting may be ready (IIRC, it has been running on Sempahore CI for more than a year, but it overwrites the failure result with a success). However, we cannot fully enable this enforcement until the quality of ./scripts/source-maintenance.sh is improved.

> IMO it is reasonable for a reviewer to ask any dev to run it

Not until the script is reliable _and_ we start enforcing formatting before each PR is merged IMO.

rousskov:
> Updated the title label since code shuffling into convenience libraries is the ongoing https://wiki.squid-cache.org/Features/SourceLayout project.

I am not going to undo that title change to avoid delaying this PR merge, but I consider the change wrong. We have discussed this before: The more title prefixes we have, the less useful they become. It also becomes harder and harder to pick the right _single_ prefix among the partially overlapping ones. These things should be annotations, not title prefixes.

The `Source Maintenance` prefix can be considered legitimate because it tells the reader that this commit is focused on source code polishing rather than bug fixes or new features. An ideal `Source Maintenance` commit has no functionality changes beyond minor debugging lines adjustment and such. The biggest prefix-specific problem with that prefix is its length.

The `SourceLayout` label is just an unimportant sub-case of `Source Maintenance`. The SourceLayout wiki page is not peer-reviewed and is not really much more than a collection of informal ideas and notes. Compared to `Source Maintenance`, referring to that wiki page in the commit title adds little value, if any.

If you agree, please undo the title change. Otherwise, the current version will be merged.

yadij:
> > Updated the title label since code shuffling into convenience libraries is the ongoing https://wiki.squid-cache.org/Features/SourceLayout project.
> 
...
> The `SourceLayout` label is just an unimportant sub-case of `Source Maintenance`. The SourceLayout wiki page is not peer-reviewed 

You may not remember, but it was indeed reviewed shortly after **we** began the project. The "Problems" section and following comments are directly from that review discussion. There have only been minor adjustments to update the status as Squid official code changed - which were usually discussed separately before the official code was altered.

rousskov:
> You may not remember, but it was indeed reviewed shortly after **we** began the project. The "Problems" section and following comments are directly from that review discussion. There have only been minor adjustments to update the status as Squid official code changed - which were usually discussed separately before the official code was altered.

I remember that, at some point, we collaborated on that page. However, there is a huge difference between that collaboration and a peer-reviewed resource:

* I cannot quickly tell whether every aspect of SourceLayout was really agreed upon at some point in time.
* I do not know whether any important (to me) edits were done since that currently-unknown-to-me moment.
* I cannot know how that wiki page will be changed after I check it (if I do).

Perhaps more importantly, all these problems and complications are not relevant to the arguments I was making against using SourceLayout as a PR title prefix! Even if somebody considers SourceLayout peer-reviewed, it still should not be used as a PR title prefix based on those arguments! I see no value in arguing about any of this.

PR #985, opened 2022-02-16
Bug 5186: noteDestinationsEnd check failed: transportWait

rousskov:
When the "no more destinations to try" notification comes after the last
forwarding/tunneling attempt has failed, the !destinationsFound block
does not run (because destinations were found), the usingDestination()
block does not run (because we are done with that last/failed
destination), but transportWait is false (for the same reason).

Also applied Bug 5090 (master/v6 commit 15bde30) FwdState protections to
tunnel.cc code. Tunnels may continue writing to the client while the
to-server connection is closing or closed, so TunnelStateData can be
potentially exposed to bug 5090 "no server connection but still
transporting" concerns. TunnelStateData never _retries_ successfully
established tunnels and, hence, can (and now does) stop receiving spare
destinations after committedToServer becomes true, but future tunnels
may start reforwarding in many cases, and most of the code does not need
to know about this (temporary?) simplification.

Also re-unified and polished related FwdState and TunnelStateData code,
including fixing lying source code comments and debug messages.

### DIFF
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -98,6 +98,9 @@ public:
         return (server.conn != NULL && server.conn->getPeer() ? server.conn->getPeer()->host : request->url.host());
     };
 
+    /// start owning the given to-server connection; do not look for any other
+    void commitToServer(const Comm::ConnectionPointer &);
+
     /// Whether the client sent a CONNECT request to us.
     bool clientExpectsConnectResponse() const {
         // If we are forcing a tunnel after receiving a client CONNECT, then we
@@ -186,6 +189,10 @@ public:
     /// whether another destination may be still attempted if the TCP connection
     /// was unexpectedly closed
     bool retriable;
+
+    /// whether we will not consider using any other server
+    bool committedToServer;
+
     // TODO: remove after fixing deferred reads in TunnelStateData::copyRead()
     CodeContext::Pointer codeContext; ///< our creator context
 
@@ -263,9 +270,8 @@ private:
 
     /// \returns whether the request should be retried (nil) or the description why it should not
     const char *checkRetry();
-    /// whether the successfully selected path destination or the established
-    /// server connection is still in use
-    bool usingDestination() const;
+
+    bool transporting() const;
 
     /// details of the "last tunneling attempt" failure (if it failed)
     ErrorState *savedError = nullptr;
@@ -362,6 +368,7 @@ TunnelStateData::TunnelStateData(ClientHttpRequest *clientRequest) :
     destinations(new ResolvedPeers()),
     destinationsFound(false),
     retriable(true),
+    committedToServer(false),
     codeContext(CodeContext::Current())
 {
     debugs(26, 3, "TunnelStateData constructed this=" << this);
@@ -1005,8 +1012,7 @@ void
 TunnelStateData::notePeerReadyToShovel(const Comm::ConnectionPointer &conn)
 {
     assert(!client.dirty);
-    retriable = false;
-    server.initConnection(conn, tunnelServerClosed, "tunnelServerClosed", this);
+    commitToServer(conn);
 
     if (!clientExpectsConnectResponse())
         tunnelStartShoveling(this); // ssl-bumped connection, be quiet
@@ -1019,6 +1025,15 @@ TunnelStateData::notePeerReadyToShovel(const Comm::ConnectionPointer &conn)
         client.write(mb->content(), mb->contentSize(), call, mb->freeFunc());
         delete mb;
     }
+}
+
+void
+TunnelStateData::commitToServer(const Comm::ConnectionPointer &conn)
+{
+    committedToServer = true;
+    retriable = false; // may already be false
+    PeerSelectionInitiator::subscribed = false; // may already be false
+    server.initConnection(conn, tunnelServerClosed, "tunnelServerClosed", this);
 }
 
 static void
@@ -1248,17 +1263,16 @@ TunnelStateData::noteDestination(Comm::ConnectionPointer path)
 
     destinations->addPath(path);
 
-    if (usingDestination()) {
-        // We are already using a previously opened connection but also
-        // receiving destinations in case we need to re-forward.
-        Must(!transportWait);
-        return;
-    }
-
     if (transportWait) {
+        assert(!transporting());
         notifyConnOpener();
         return; // and continue to wait for tunnelConnectDone() callback
     }
+
+    // the current commitToServer()-based code cannot receive new destinations
+    // while transporting(), but future enhancements may enable this code path
+    if (transporting())
+        return; // and continue to receive destinations for backup
 
     startConnecting();
 }
@@ -1287,12 +1301,12 @@ TunnelStateData::noteDestinationsEnd(ErrorState *selectionError)
     Must(!selectionError); // finding at least one path means selection succeeded
 
     if (transportWait) {
-        assert(!usingDestination());
+        assert(!transporting());
         notifyConnOpener();
         return; // and continue to wait for the noteConnection() callback
     }
 
-    if (usingDestination()) {
+    if (transporting()) {
         // We are already using a previously opened connection (but were also
         // receiving more destinations in case we need to re-forward).
         debugs(17, 7, "keep transporting");
@@ -1305,12 +1319,13 @@ TunnelStateData::noteDestinationsEnd(ErrorState *selectionError)
     sendError(savedError, "all found paths have failed");
 }
 
+/// Whether a tunneling attempt to some selected destination X is in progress
+/// (after successfully opening/reusing a transport connection to X).
+/// \sa transportWait
 bool
-TunnelStateData::usingDestination() const
+TunnelStateData::transporting() const
 {
-    // server.conn->close() does not end usingDestination(); we may receive new
-    // paths while server.conn is closing (XXX: and client is still writing)
-    return encryptionWait || peerWait || server.conn;
+    return encryptionWait || peerWait || committedToServer;
 }
 
 /// remembers an error to be used if there will be no more connection attempts
@@ -1369,7 +1384,7 @@ TunnelStateData::startConnecting()
         request->hier.startPeerClock();
 
     assert(!destinations->empty());
-    assert(!usingDestination());
+    assert(!transporting());
     AsyncCall::Pointer callback = asyncCall(17, 5, "TunnelStateData::noteConnection", HappyConnOpener::CbDialer<TunnelStateData>(&TunnelStateData::noteConnection, this));
     const auto cs = new HappyConnOpener(destinations, callback, request, startTime, 0, al);
     cs->setHost(request->url.host());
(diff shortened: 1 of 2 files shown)


rousskov:
**Disclaimer**: This PR has not been adequately tested yet so I am marking it as a Draft. I do not have an automated test for this bug and relying on several active users on Bugzilla to help with testing this fix while it is being reviewed.

src/tunnel.cc
```diff
@@ -1285,21 +1300,32 @@ TunnelStateData::noteDestinationsEnd(ErrorState *selectionError)
...
+        // We are already using a previously opened connection (but were also
+        // receiving more destinations in case we need to re-forward).
+        debugs(17, 7, "keep transporting");
         return;
     }
 
-    Must(transportWait); // or we would be stuck with nothing to do or wait for
-    notifyConnOpener();
+    // destinationsFound, but none of them worked, and we were waiting for more
+    assert(savedError);
```
eduard-bagdasaryan, line 1317: We could probably rely on the similar assertion in sendError() below.

eduard-bagdasaryan (commented):
Please see my comments.

src/tunnel.cc
```diff
@@ -1021,6 +1027,15 @@ TunnelStateData::notePeerReadyToShovel(const Comm::ConnectionPointer &conn)
     }
 }
 
+void
+TunnelStateData::commitToServer(const Comm::ConnectionPointer &conn)
+{
+    committedToServer = true;
+    retriable = false; // may already be false
```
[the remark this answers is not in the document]
rousskov: Yes, but that can be viewed as a temporary coincidence, so I would rather not merge the two fields with rather different meanings, especially because it will make merging with FwdState more difficult. In the future, tunnels may retry more Squid-server errors, making `retriable` scope larger (crossing into the true `committedToServer` area) and also adding code that makes true committedToServer false again. Just like FwdState already does!

src/tunnel.cc
rousskov: You are right. I am a bit worried that addressing the `XXX` below will make sendError() calls _conditional_ while this assertion (about the job state as it pertains to _this_ method) remains valid even when we cannot sendError(). I will remove this assert() if somebody insists on the removal.

src/tunnel.cc
```diff
@@ -186,6 +189,10 @@ class TunnelStateData: public PeerSelectionInitiator
     /// whether another destination may be still attempted if the TCP connection
     /// was unexpectedly closed
     bool retriable;
+
+    /// whether we will not consider using any other server
```
yadij, line 193: Please avoid negative phrasing of a positive-style name.
```suggestion
    /// whether the choice of server is absolute. Indeterminate data/state has likely been shared between it and client.
```

src/tunnel.cc
```diff
@@ -263,9 +270,8 @@ class TunnelStateData: public PeerSelectionInitiator
 
     /// \returns whether the request should be retried (nil) or the description why it should not
     const char *checkRetry();
-    /// whether the successfully selected path destination or the established
-    /// server connection is still in use
-    bool usingDestination() const;
+
+    bool transporting() const;
```
yadij, line 274: Unexplained method with an odd name.

If this is done to bring the code better in sync with FwdState API it needs a reference to those docs:
```suggestion
    /// \see FwdState::transporting()
    bool transporting() const;
```

NP: long-term maybe we should start some sections of ``` /* Foo API (mimic) */``` for AsyncJob/FwdState where these things can go without need to document.

src/tunnel.cc
```diff
@@ -1248,18 +1263,17 @@ TunnelStateData::noteDestination(Comm::ConnectionPointer path)
...
-    }
-
     if (transportWait) {
+        assert(!transporting());
         notifyConnOpener();
         return; // and continue to wait for tunnelConnectDone() callback
     }
 
+    // the current commitToServer()-based code cannot receive new destinations
+    // while transporting(), but future enhancements may enable this code path
```
yadij, line 1273: IMO, it is reasonable even today for the first server to be used while alternative server details keep coming in. No need to pick out this implementation detail for comment unless you want to make it a TODO.
```suggestion
    // performance optimization; avoid connecting to alternative servers once transport begins.
```
or
```suggestion
    // the current commitToServer()-based code cannot receive new destinations
    // while transporting(). TODO open and pool alternative server connections for other clients or future use.
```

src/tunnel.cc
```diff
@@ -1275,8 +1289,9 @@ TunnelStateData::noteDestinationsEnd(ErrorState *selectionError)
         if (selectionError)
             return sendError(selectionError, "path selection has failed");
 
+        // TODO: Merge with FwdState and remove this likely unnecessary check.
         if (savedError)
-            return sendError(savedError, "all found paths have failed");
+            return sendError(savedError, "path selection found no paths (with an impossible early error)");
```
yadij, line 1294: What does "impossible early error" mean here ?

If it is truly impossible for savedError to be non-nul this should be an assert instead of an if-statement

src/tunnel.cc
```diff
@@ -1285,21 +1300,32 @@ TunnelStateData::noteDestinationsEnd(ErrorState *selectionError)
...
-        // We are already using a previously opened connection but also
-        // receiving destinations in case we need to re-forward.
-        Must(!transportWait);
+    if (transportWait) {
+        assert(!transporting());
+        notifyConnOpener();
+        return; // and continue to wait for the noteConnection() callback
+    }
+
+    if (transporting()) {
```
yadij, line 1309: The assert here makes me wonder if we should reverse the order of these if-statements. I do not insist on any change here.

src/tunnel.cc
yadij: I agree with @eduard-bagdasaryan on this one.  One of the possible implementations of that XXX may may be to avoid allocating savedError objects entirely when they cannot be delivered. Whatever that ends up being this assert can be added by that change if-necessary.

yadij (approved):
The logic looks sound to me, though I have not looked at whether it fixed the referenced bug.

Since I only have documentation polish requests I'm marking this as approved. It should not need another review from me unless there is logic change after testing.

src/tunnel.cc
rousskov: I hate negative definitions of booleans, but I could not come up with a better one.

I could not use your suggestion "as is" (for various technical reasons that I can discuss if you insist), but it helped to find a better way to describe this boolean data member (commit 410e396). Thank you! The current code reads: `whether the decision to use a particular destination was final`.

src/tunnel.cc
```diff
@@ -98,6 +98,9 @@ class TunnelStateData: public PeerSelectionInitiator
         return (server.conn != NULL && server.conn->getPeer() ? server.conn->getPeer()->host : request->url.host());
     };
 
+    /// start owning the given to-server connection; do not look for any other
```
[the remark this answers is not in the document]
rousskov: > We have differences of definition for "ownership" Ownership which make this description unclear about scope/timing.

Yeah. Soon we will be running out of standard, technical words we can use without arguments!

> I would prefer this were phrased in clearer terms. Something like this maybe:

```diff
- start owning the given to-server connection; do not look for any other
+ Finalize the given to-server connection for use. Retry or use of others is now prohibited.
```

The "finalize for use" part is too vague to be useful IMO. I have replaced "start owning" with "store" in hope to avoid the banned "ownership" word, but it makes the description less useful. Ownership is a valid technical term that is very useful in many contexts. We should not ban it, especially in places where it has very simple/reasonable/basic definition like "the object owner determines that object lifetime", even though the abuse of shared pointers makes such ownership claims less precise (they effectively talk about co-ownership rather than exclusive ownership).

Here is the current version:

```C++
/// store the given to-server connection; prohibit retries and do not look
/// for any other destinations
void commitToServer(const Comm::ConnectionPointer &);
```

rousskov (commented):
Thank you for reviewing this PR and not insisting on any changes. I have tried to address your concerns and applied a few corrections inspired by your change requests.

I am not clearing this Draft PR for merging yet because I still hope that somebody will confirm that the code works in their environment (or expose PR flaws).

src/tunnel.cc
yadij: Sigh. You are right of course.

src/tunnel.cc
yadij: >     * impossible: I believe that such early errors do not exist today, but I am not sure and consider figuring the correct answer to that existence question difficult out-of-scope (and possibly even unnecessary due to other XXXs) work.
> 

Given that, I suggest "very early" to avoid others like me confusing this with a certainty.

> 
> > If it is truly impossible for savedError to be non-nul this should be an assert instead of an if-statement
> 
> Agreed. If we had Assure(), I might have had the courage to put it in here in this PR, but we still do not have Assure(), and convincing myself that an assert() is safe here is difficult out-of-scope (and possibly unnecessary) work.

That would be why Must() currently gets used in places like this.

src/tunnel.cc
yadij: Ah. IMO exclusivity is better represented by if-else structure. This sequential style can look nice, but does imply some order relations to the statements - which is prone to devs "improving" the sequence for performance goals.

Anyway, closing the thread.

src/tunnel.cc
yadij: looks good. thank you.

src/tunnel.cc
rousskov:
```
"path selection found no paths (with an impossible early error)"
```

> > * early: Here, roughly speaking, it is an error that occurred before TunnelStateData received either the first destination or a destination search failure notification.
> > * impossible: I believe that such early errors do not exist today, but I am not sure and consider figuring the correct answer to that existence question difficult out-of-scope (and possibly even unnecessary due to other XXXs) work.
 
> Given that, I suggest "very early" to avoid others like me confusing this with a certainty.

I do not understand how replacing "early" with "very early" reduces _certainty_ for others like you. For me, it would only introduce a third class of errors -- the "very early" ones -- something I would rather not do. Would you prefer "a likely impossible early error" phrasing instead? Adding "likely" would indeed decrease certainty (for any reader) IMO.
 
> > > If it is truly impossible for savedError to be non-nul this should be an assert instead of an if-statement
 
> > Agreed. If we had Assure(), I might have had the courage to put it in here in this PR, but we still do not have Assure(), and convincing myself that an assert() is safe here is difficult out-of-scope (and possibly unnecessary) work.
 
> That would be why Must() currently gets used in places like this. 

Exactly.

> Would you dare join the rabble in using it?

Not in this PR. As I said, investigating whether those errors are indeed impossible requires serious out of scope (and likely wasteful) work. On the other hand, if this PR starts killing these transactions (because those early errors happen to be possible after all), that would be an out of scope change (and a bug).

src/tunnel.cc
rousskov: > Ah. IMO exclusivity is better represented by if-else structure.

The `if-else` structure is indeed the best for representing an exclusive `c` and `not c` relationship (when handling of each condition requires very little code). Unfortunately, this code has nothing to do with that kind of relationship.

Here, instead of two mutually exclusive conditions, we have three: `c1`, `c2`, and `neither c1 nor c2`. C++ does not have syntax for expressing such mutual exclusivity well. And no, `if-else-if-else` does not express such mutual exclusivity (at all).

> This sequential style can look nice, but does imply some order relations to the statements - which is prone to devs "improving" the sequence for performance goals.

I do not think your assumption is correct. The clauses in _both_ if-else-if-else and if-return-if-return styles may imply some order relation _and_ may be reordered for performance (or other) reasons.

C++ just does not give us a good way to express mutual exclusivity of three conditions. The current code is the best we can do with the tools we have AFAICT.

rousskov:
> I am not clearing this Draft PR for merging yet because I still hope that somebody will confirm that the code works in their environment (or expose PR flaws).

I have lost that hope: We have not received any explicit confirmations that this code works. We cannot force folks to test this, and these or similar changes are necessary to move forward. I am removing the Draft status and plan to clear this PR for merge into master on Monday or so. Merging this PR will expose this code to more potential testers and let us build on top of it.

rousskov:
> I have lost that hope: We have not received any explicit confirmations that this code works. We cannot force folks to test this, and these or similar changes are necessary to move forward. I am removing the Draft status and plan to clear this PR for merge into master on Monday or so. Merging this PR will expose this code to more potential testers and let us build on top of it.

Cleared for merge.

PR #986, opened 2022-02-18
Fix Ubuntu Xenial clang build broken by commit b2188e4

kinkie:
CpuAffinitySet.cc:41: expression result unused [-Wunused-value]

On Ubuntu Xenial, clang expands CPU_AND() into a loop containing an unused
expression. Commit 7ec6d51 added (void) to quiet the corresponding warning.
Commit b2188e4 removed that decoration to fix FreeBSD 14 build because there
CPU_AND() expands into a do-while loop which cannot be prefixed by (void).

### DIFF
--- a/src/CpuAffinitySet.cc
+++ b/src/CpuAffinitySet.cc
@@ -37,8 +37,7 @@ CpuAffinitySet::apply()
                "this process: " << xstrerr(xerrno));
     } else {
         cpu_set_t cpuSet;
-        memcpy(&cpuSet, &theCpuSet, sizeof(cpuSet));
-        CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
+        CPU_AND(&cpuSet, &theCpuSet, &theOrigCpuSet);
         if (CPU_COUNT(&cpuSet) <= 0) {
             debugs(54, DBG_IMPORTANT, "ERROR: invalid CPU affinity for process "
                    "PID " << getpid() << ", may be caused by an invalid core in "


rousskov:
Thank you for working on this fix!

> The fix for FreeBSD breaks Ubuntu Xenial / Clang.

Let's not repeat commit 7ec6d51 mistake that probably resulted in this _repeated_ build breakage! Please adjust the PR description to document what breakage you observed. An abridged compiler error message is usually all it takes. I can help with formatting if needed. I adjusted the PR title assuming it is a build error, but please change it if it is not.

I also adjusted PR title to use commit SHA instead of PR number as a reference point. If you [recall](https://github.com/squid-cache/squid/pull/549#issuecomment-579795152), commit SHAs are arguably more unique/Squid-specific references than GitHub-specific PR numbers. Commit SHAs also make it easier to find the relevant change when working on the command line (while still providing reasonably convenient access to the PR for those who browse GitHub pages). Most official Squid commits use SHAs instead of PR numbers to refer to other commits.

I also adjusted the PR description to simplify and prepare it for the official merge. Please adjust further as needed, keeping in mind that the PR title/description will become the official commit message when the PR is merged.

I know this is a Draft PR, but bad descriptions may force the reviewer to block the PR because the reviewer does not control when the Draft flag is going to be removed (and the unblocked PR merged with the bad description). IMO, we should write ready-for-merge PR descriptions (to the extent possible) and then use GitHub comments for any Draft-specific discussions.

rousskov (commented):
I fixed the PR description. However, now that I actually know what is going on, I am pretty sure this PR does not fix the Ubuntu Xenial clang build -- the `__dest` expression result in the macro is still unused.

```C++
(__extension__ ({
       cpu_set_t *__dest = (&s1);
       const __cpu_mask *__arr1 = (&s2)->__bits;
       const __cpu_mask *__arr2 = (&s3)->__bits;
       size_t __imax = (sizeof (cpu_set_t)) / sizeof (__cpu_mask);
       size_t __i;
       for (__i = 0; __i < __imax; ++__i) (
           (__cpu_mask *) __dest->__bits)[__i] = __arr1[__i] & __arr2[__i];
           __dest; // XXX: unused expression result
}));
```

rousskov:
I cannot fully reproduce this (see comment in the test source code), but if you want to play with the actual warning in real-time, you can use https://godbolt.org/z/1oGefGEaK

rousskov:
> The bizarre fact here is that CPU_AND is documented as returning void

Yes, we have discussed the limitations of that documentation before. Since CPU_AND is a macro, the details of that function-based documentation should be taken with a grain of salt. It is "as if CPU_AND was declared like this" documentation rather than "this is the actual function profile" documentation. Clearly, whoever implemented CPU_AND() in the environment in question did not realize that they need to cast the result to (void).

kinkie:
Yeah, that's the clean route. Xenial will be around for another five years, so the investment might be worthwhile

rousskov:
> if the stars align poorly, parts of it might not be [noop], because the testing logic is a mishmash

I hope that since sched_setaffinity() and sched_getaffinity() return errors where they are not supported, the stars cannot align poorly -- we will always end up in an error situation even if some manipulations are accidentally successful on the way.

> it would be better to be intentional and really noop it out

One cannot "noop out" a function that returns a result (e.g., `CPU_COUNT` or `CPU_ISSET`) -- the compiler would not allow us to do nothing. The choices are:

* lie about the result (i.e. `return false`),
* never return (i.e. throw or abort()),
* prohibit calls/use (i.e. make sure the function/macro is not declared).

If we want a better guarantee then we could stop lying and start throwing on unsupported platforms. That may require some additional work to make sure the user gets a reasonable error message when they try to use the corresponding feature on an unsupported platform, but it should not be difficult to get it right.

> I'll see if the effort to properly support affinity on freebsd is not too high as an alternate solution to what we have now

Adding proper CPU affinity support to FreeBSD would be nice, of course. We already have custom classes that insulate OS-specific differences from the rest of the code, but those classes will need to be adjusted to add FreeBSD support, of course. FWIW, I believe the original design applies to future FreeBSD support well and should be preserved:

```
Squid builds on systems without Linux CPU affinity calls, but setting affinity
only works if there are sched_getaffinity(2) and sched_setaffinity(2)
available. If there is no OS support but cpu_affinity options are configured,
Squid quits with an error. If there is OS support but calls fail, Squid prints
an error but does not quit.
```

Not sure any of the above will help with the problem this PR is trying to fix though :-).

kinkie:
Abasndoning as it was found to be an issue with ccache

PR #987, opened 2022-02-20
Use xmlCtxtUseOptions instead of xmlSetFeature

nwellnhof:
xmlSetFeature has been deprecated for 10+ years and will eventually be
removed from libxml2. Use xmlCtxtUseOptions instead. Despite the
misleading name, XML_PARSE_NOENT enables substitution of entities
(no entity nodes in the result tree).


nwellnhof:
Sorry, I made a mistake and will create a new PR.

PR #988, opened 2022-02-20
ESI: Drop incorrect and unnecessary xmlSetFeature() call

nwellnhof:
xmlSetFeature() has been deprecated for 10+ years and will eventually be
removed from libxml2. Squid calls xmlSetFeature() with the wrong
argument: a nil `value` pointer instead of a pointer to a zero value.
When called with a nil `value`, the function does nothing but returning
an error. Squid does not check whether xmlSetFeature() call is
successful, and the bug went unnoticed since libxml2 support was added
in commit 964b44c.

Since libxml2 does not substitute entities by default, the call can be
removed to achieve the intended effect.

### DIFF
--- a/src/esi/Libxml2Parser.cc
+++ b/src/esi/Libxml2Parser.cc
@@ -118,7 +118,6 @@ ESILibxml2Parser::ESILibxml2Parser(ESIParserClient *aClient) : theClient (aClien
 
     /* TODO: grab the document encoding from the headers */
     parser = xmlCreatePushParserCtxt(&sax, static_cast<void *>(this), NULL, 0, NULL);
-    xmlSetFeature(parser, "substitute entities", 0);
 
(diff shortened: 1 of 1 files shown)


rousskov (approved):
@nwellnhof, thank you for fixing this code!

I adjusted the PR title/description (i.e. the future [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message)) to fix formatting and add a bit more info. Please check whether the current PR title/description are OK with you.

rousskov:
OK to test

rousskov:
@yadij, I have cleared this fix for merging. The changes seem non-controversial and simple enough to merge with just one review, and you have ~8 days to stop this from going in.

yadij (approved):
LGTM, thank you

PR #989, opened 2022-02-24
Drop redundant LIBADD_DL overwrite

yadij:
The removed code questions whether it is needed.

The answer is no. LIBADD_DL is already set by one of the LT_INIT
macros to the same value we are overwriting it to.

Except, in the case where the user has supplied a custom libdl
location this code broke their build.


kinkie (approved):
let's ship this

PR #991, opened 2022-02-28
Improve modularity of IDENT protocol code

yadij:
Simple changes to further isolate the IDENT code
in ident/libident.la and redux the mainInitialize() sequence.

Ident::Start() is already checking for and triggering ident_hash
initialization. No need to do so in main.cc if there is never
any use of IDENT. Removing one unnecessary memory allocation.

### DIFF
--- a/src/acl/RegexData.h
+++ b/src/acl/RegexData.h
@@ -10,10 +10,9 @@
 #define SQUID_ACLREGEXDATA_H
 
 #include "acl/Data.h"
+#include "base/RegexPattern.h"
 
 #include <list>
-
-class RegexPattern;
 
 class ACLRegexData : public ACLData<char const *>
 {
--- a/src/acl/RegexData.h
+++ b/src/acl/RegexData.h
@@ -10,9 +10,10 @@
 #define SQUID_ACLREGEXDATA_H
 
 #include "acl/Data.h"
-#include "base/RegexPattern.h"
 
 #include <list>
+
+class RegexPattern;
 
 class ACLRegexData : public ACLData<char const *>
 {
--- a/src/ident/Ident.cc
+++ b/src/ident/Ident.cc
@@ -11,7 +11,10 @@
 #include "squid.h"
 
 #if USE_IDENT
+#include "acl/RegexData.h"
+#include "acl/UserData.h"
 #include "base/JobWait.h"
+#include "base/RunnersRegistry.h"
 #include "comm.h"
 #include "comm/Connection.h"
 #include "comm/ConnOpener.h"
@@ -19,6 +22,7 @@
 #include "comm/Write.h"
 #include "CommCalls.h"
 #include "globals.h"
+#include "ident/AclIdent.h"
 #include "ident/Config.h"
 #include "ident/Ident.h"
 #include "MemBuf.h"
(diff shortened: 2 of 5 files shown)


src/acl/RegexData.h
```diff
@@ -10,11 +10,10 @@
 #define SQUID_ACLREGEXDATA_H
 
 #include "acl/Data.h"
+#include "base/RegexPattern.h"
 
 #include <list>
 
-class RegexPattern;
-
```
rousskov, line 17: This change exposes all src/acl/RegexData.h users to a RegexPattern declaration. Some of those (current or future) users may not need to know how RegexPattern is declared to compile, so we should not create such an unnecessary exposure. Please undo and `#include` RegexPattern.h where it is necessary instead.

rousskov (changes requested):

src/ident/Ident.cc
```diff
@@ -78,6 +82,17 @@ static void ClientAdd(IdentStateData * state, IDCB * callback, void *callback_da
...
+class IdentRr : public RegisteredRunner
+{
+public:
+    IdentRr() {
+        Acl::RegisterMaker("ident", [](Acl::TypeName name)->ACL* { return new ACLIdent(new ACLUserData, name); });
+        Acl::RegisterMaker("ident_regex", [](Acl::TypeName name)->ACL* { return new ACLIdent(new ACLRegexData, name); });
+    }
+};
+
+RunnerRegistrationEntry(IdentRr);
```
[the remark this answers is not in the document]
yadij: There is also:

 5) use the constructor as an event handler for Squid startup/initialization (pre-configuration).

That is symmetric with how destructor is required to be used as post-finishShutdown() event handler and can be used by many of the components currently doing their own things.

src/acl/RegexData.h
yadij: Done in [485057f](https://github.com/squid-cache/squid/pull/991/commits/485057f53116f54fbc76a6a27f4d41ff11858e29)

src/ident/Ident.h
```diff
@@ -28,12 +28,6 @@ namespace Ident
  */
```
rousskov, line 28:
```
 * will call Ident::Init() itself if the manager has not been initialized already.
```

I cannot attach a change request to it, but please delete the stale line 27 above. Besides referring to the function removed by this PR, that line has become an irrelevant implementation detail.

Please do not forget to replace `,` on line 26 with `.` afterwords.

rousskov:
FYI: I replaced Runner with RegisteredRunner in the PR description. There are IndependentRunner classes already, and future code may add more classes with similar names. It does not hurt to be more precise here.

src/ident/Ident.cc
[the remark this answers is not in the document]
yadij: Okay moving this runner change to a followup PR and going with your proposal.

src/ident/Ident.h
yadij: Done in [974d953](https://github.com/squid-cache/squid/pull/991/commits/974d953b779f5ba1148d63484fd7ef442ea46f60)

src/ident/Ident.cc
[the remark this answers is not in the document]
yadij: Added PR #992 for the RunnersRegistry API change

rousskov (approved):

There is a hidden dependency between IdentStateData object and ident_hash lifetimes. I see no problems now, but to improve safety during future Ident.cc code refactoring, we may want to assert that the hash exists inside hash_remove_link(). I do not insist on that arguably out-of-scope change.

PR #992, opened 2022-02-28
Add RegisteredRunners::bootstrapConfig event hook

yadij:


src/base/RunnersRegistry.h
```diff
@@ -39,6 +39,12 @@ class RegisteredRunner
 public:
     /* Related methods below are declared in their calling order */
 
+    /* Startup events */
+
+    /// Called right before parsing squid.conf.
+    /// Meant for initializing/preparing configuration parsing facilities.
+    virtual void bootstrapConfig() {}
+
     /* Configuration events */
```
rousskov, line 48: Please move the new configuration-related event to the `Configuration events` group. We may need a set of startup events unrelated to configuration as well, of course, but this PR is not about that.

src/main.cc
```diff
@@ -1174,6 +1174,8 @@ mainInitialize(void)
 
 #endif
 
+    // TODO make these init() calls happen as RegisteredRunner::finalizeConfig
```
rousskov, line 1177: The exact scope of the "these init() calls" phrase is unclear, and not all of the initialization calls below should be converted to finalizeConfig. I suggest dropping this out-of-scope change to avoid arguing about it.

src/main.cc
```diff
@@ -1568,6 +1570,9 @@ SquidMain(int argc, char **argv)
 
         Mem::Init();
 
+        RunRegisteredHere(RegisteredRunner::bootstrapConfig);
+        // TODO: convert the below init() into RegisteredRunners::bootstrapConfig
```
rousskov, line 1574: The comment seems misplaced (or oddly placed), and I doubt the proposed conversion is the best solution. I suggest dropping this out-of-scope change to avoid arguing about it.

src/main.cc
```diff
@@ -1587,7 +1592,7 @@ SquidMain(int argc, char **argv)
         Auth::Init();      /* required for config parsing. NOP if !USE_AUTH */
         Ip::ProbeTransport(); // determine IPv4 or IPv6 capabilities before parsing.
 
-        Format::Token::Init(); // XXX: temporary. Use a runners registry of pre-parse runners instead.
+        Format::Token::Init();
```
rousskov, line 1595: FWIW, I am not against this change because I cannot understand what the removed XXX means, but if you want to keep this PR focused (after removing other comment changes), you may want to remove this change as well. Your call.

rousskov (changes requested):

src/main.cc
yadij: Done in c6b6dbd

src/main.cc
yadij: Done in c6b6dbd

src/main.cc
yadij: Done in c6b6dbd

src/base/RunnersRegistry.h
yadij: Done in c6b6dbd

src/main.cc
```diff
@@ -1568,6 +1568,8 @@ SquidMain(int argc, char **argv)
 
         Mem::Init();
 
+        RunRegisteredHere(RegisteredRunner::bootstrapConfig);
```
rousskov, line 1571: We should trigger this event later, as we promise in the event description -- just before parsing the configuration file. This line should be moved below the Format::Token::Init() call, just above the try/catch block that parses the configuration file.

We probably need a different bootstrapping event that happens somewhere here as well, but that configuration-unrelated event is outside this PR scope (unless you want to add two largely independent events in one PR, of course).

rousskov (changes requested):

src/main.cc
yadij: I placed it here with the comment you requested removed, because all the init() calls between here and the parsing are supposed to be independent and should be runners implementing this hook.

I want to actively prohibit any of those runners accidentally working with a dependency on one of the not yet converted init() calls. Only memory allocation should be assumed working.

rousskov (approved):
Thank you for addressing my concern.

I trust there will be a followup PR using bootstrapConfig soon so that we are not adding unused code for long.

PR #993, opened 2022-03-02
Ensure initClient MasterXactions have listening ports

eduard-bagdasaryan:
MasterXaction creation code that forgets to supply a port for an
initClient transaction will now be rejected by the compiler.

This safety improvement hit two problems described below.

The TcpAcceptor class was incorrectly creating a MasterXaction object
for FTP DATA connections. We moved MasterXaction creation to TcpAcceptor
callbacks because only they know whether accepting a TCP connection
implies a new master transaction start (it does not for FTP DATA cases).

The Downloader class was implemented to spawn its own MasterXaction
using XactionInitiator info supplied by the download requestor. That
design is incompatible with the new static assertion because each
MasterXaction creator must now hard-code XactionInitiator value for the
assertion to work at _compile_ time. All other contexts naturally
hard-code that value.

We saw two ways to resolve this conflict:

a) Let the download requestor create the MasterXaction object for the
   Downloader. The primary advantage of this (implemented) approach is
   that it allows (future) client request processing code to satisfy
   client requests using Downloader. Such request satisfaction requires
   sharing the master transaction between the client transaction and the
   downloader transaction, and this change enables such sharing. This
   change moves the "Which master transaction does this download belong
   to?" question from the Downloader into download requestors.

b) Move initClient-has-port enforcement (and squidPort) from
   MasterXaction into XactionInitiator. The primary advantage of this
   (not implemented) approach is that it places that enforcement inside
   the type it is meant to police (essentially) -- XactionInitiator.

The two changes are complementary. We did not implement (b) because it
requires significant XactionInitiator redesign, moving _all_ originating
client information from MasterXaction to XactionInitiator (currently
squidPort and tcpClient).


src/client_side_reply.cc
```diff
@@ -2140,7 +2140,8 @@ clientReplyContext::createStoreEntry(const HttpRequestMethod& m, RequestFlags re
      */
 
     if (http->request == NULL) {
-        const MasterXaction::Pointer mx = new MasterXaction(XactionInitiator::initClient);
+        const auto connManager = http->getConn();
+        const auto mx = MasterXaction::MakePortful(connManager ? connManager->port : nullptr);
```
yadij, line 2144: That a "Portful" xaction can be passed nullptr is a bug.

This should be:
```suggestion
        MasterXaction::Pointer mx;
        if (const auto connManager = http->getConn())        
            mx = MasterXaction::MakePortful(connManager->port);
        else
            mx = MasterXaction::MakePortless(XactionInitiator::initClient);
```

or
```suggestion
        const auto mx = MasterXaction::MakePortful(XactionInitiator::initClient);
```

yadij (changes requested):

eduard-bagdasaryan:
> update that https://github.com/squid-cache/squid/pull/835 branch-based text

I removed the non-existing method.

src/client_side_reply.cc
yadij: The meaning of the words "Portful" and "Portless" are opposites and mutually exclusive. It is a semantic and logical flaw for **both** to mean "no port exists". Otherwise there is no point in having the "Portless" variant at all.

There were two suggestions, I have now corrected the typo in second. Though I think the first is what we all prefer.

eduard-bagdasaryan: Note that both your suggestions have _MakePortless(XactionInitiator::initClient)_, leading to a static assertion.

rousskov: > The meaning of the words "Portful" and "Portless" are opposites and mutually exclusive.

Correct.

> It is a semantic and logical flaw for both to mean "no port exists" (the case where nullptr is accepted by Portful).

You are criticizing a meaning that does not exist. Portful means "the port should exist (but, due to problems beyond caller control, information about the port may have been lost)". The existing method descriptions attempt to document this, but we should improve that documentation if it is not clear enough. Please suggest such improvements.

```diff
-        const auto mx = MasterXaction::MakePortful(connManager ? connManager->port : nullptr);
+        const auto mx = MasterXaction::MakePortless(XactionInitiator::initClient);
```

The whole point of these methods is to prevent `MakePortless(initClient)` or similar calls (at _compile_ time) because humans will place them where they should not exist -- bugs notwithstanding, all initClient contexts are associated with a port.

Squid code does have bugs, and we cannot fix them all in this PR. Thus, we want to emphasize that something went wrong in a particular context by explicitly supplying a nil port to MakePortful. The contradiction is fairly obvious and that is the whole point of this design. There is no way to guarantee the absence of bugs, of course, but the proposed way attracts attention to code that requires special attention. Like the code this change request is attached to.

That special attention does not imply that the caller code itself is necessarily buggy. It only means that the author and reviewer must make sure the port information may have been lost indeed, for reasons (i.e. bugs) outside the caller control.

yadij: You are both nitpicking the specific use of ```initClient``` in a **suggestion**.

You missed that my point was about using an if-statement so that ```MakePortful()``` is **only** called when a port is known. ```MakePortless()``` whenever it is unknown or lost.


> 
> The whole point of these methods is to prevent `MakePortless(initClient)` or similar calls (at _compile_ time) because humans will place them where they should not exist -- bugs notwithstanding, all initClient contexts are associated with a port.

... except when they are associated with a nullptr. Like this PR line does.

To me accepting nulltpr into MakePortful() is one of those very bugs this PR change is trying to catch.

i.e.
This MakePortful() accepting nullptr is a clear demonstration that devs can/should use ```MakePortful(foo, nullptr)``` any time they want.
 -> In other words; your own code demonstrates that MakePortless() is pointless fluff which can be dropped.
 -> Or to take it one step further; demonstrating that the original code we have right now is good enough and no benefit is added by this PR.

I think we can agree that those implications are bad and we should not to accept code with the pattern ```MakePortful(nullptr)```

rousskov: I am not nitpicking at all. I am trying to explain what this PR is about because the suggestions in this change request have missed the very essence of this PR.  AFAICT, Eduard is doing the same, but he is using compiler output -- a static assertion -- to illustrate essentially the same point.

To make progress, we should first try to reach an understanding of what problem this PR is trying to address. Only after we develop such a shared understanding it makes sense to discuss specific solutions to that problem. Currently, it looks like this precondition has not been met yet.

The problem this PR is trying address is folks creating a MasterXaction object with an initClient initiator while _forgetting_ to supply the port information to that object, even though that port information _is_ available. We cannot guarantee that such bugs will never happen, but we can provide an API that decreases the probability of such bugs.

@yadij, does my explanation clarify what risk this PR tries to reduce? Do you agree that the risk of forgetting to supply the port (when port is available) is worth decreasing? If yes, then we can discuss the best way to reduce that risk.

> To me accepting nulltpr into MakePortful() is one of those very bugs this PR change is trying to catch.

Exactly! Every MakePortful(nullptr) is a bug. This PR claims that all similar expressions in this PR are old bugs that we cannot easily fix (and should not fix) in this PR. The PR essentially marks those old bugs. There is no harm in such markings and there is some obvious benefit, but those old bugs are not what this PR is focusing on.

The key idea behind this PR is to decrease the probability of new/future bugs of the same kind -- forgetting to supply an available port. Every time an author or reviewer sees a new MakePortful(nullptr) call being added to Squid, they should pause -- "What is going on here? Did I/they forget to supply the required port information?". Hopefully, we will answer such questions correctly in most cases and reduce the number of _new_ bugs that way.

Eventually, all old bugs will be fixed as well, and we will be able to prohibit nil ports for initClient MasterXactions. It may take a while though because weak pointers that lose port information are widespread in Squid.

If we allow MakePortful(initClient) or MakePortful() expressions, then the idea behind this PR stops working. Authors and reviewers are a lot less likely to think twice about those expressions because they look "normal" or "ordinary". It is the explicit `nullptr` in `MakePortful(nullptr)` that makes (some of) us pause and think.

yadij: > @yadij, does my explanation clarify what risk this PR tries to reduce?

Yes, and I do understand that.

What I am trying to get across is that ...

> > To me accepting nulltpr into MakePortful() is one of those very bugs this PR change is trying to catch.
> 
> Exactly! Every MakePortful(nullptr) is a bug. This PR claims that all similar expressions in this PR are old bugs that we cannot easily fix (and should not fix) in this PR. The PR essentially marks those old bugs. There is no harm in such markings and there is some obvious benefit, but those old bugs are not what this PR is focusing on.
> 

IMO **for this PRs case** simply replacing the API and leaving this bug is not good enough ...


> If we allow MakePortful(initClient) or MakePortful() expressions, then the idea behind this PR stops working. Authors and reviewers are a lot less likely to think twice about those expressions because they look "normal" or "ordinary". It is the explicit `nullptr` in `MakePortful(nullptr)` that makes (some of) us pause and think.

I am okay with ```MakePortless(initClient)``` being prohibited.

What I am against is ```MakePortful(nulltpr)``` being left as allowed. Specifically because this Downloader usage is essentially **the** bug with current API. This PR having a seamless change of all other code demonstrates that.

yadij: Maybe an alternative suggestion would help clarify that I don't really care **how** this bug is removed, just that it is ...

```suggestion
        const auto connManager = http->getConn();
        assert(connManager); // XXX: Downloader without any client connection not yet supported
        const auto mx = MasterXaction::MakePortful(connManager->port);
```

rousskov: @eduard-bagdasaryan, can we fix this particular out of scope bug if we require all clientReplyContext creators to supply their MasterXaction object (that the clientReplyContext constructor will store via a shared pointer)? In other words, refactor clientReplyContext to never create a new MasterXaction object?

@eduard-bagdasaryan, I believe the official code creating initClient MasterXaction here is another out-of-scope bug. Some of these transactions are not necessarily initClient AFAICT. Do you agree?

eduard-bagdasaryan: > Some of these transactions are not necessarily initClient AFAICT

I think you are right. For example, clientReplyContext created from clientBeginRequest() initiated by ESIInclude::Start() should have initEsi instead.

rousskov: Sounds good. Please try implementing that. If that fix works, I hope we will be able to require port information for all initClient master transactions.

yadij (approved):
LGTM now.

@rousskov, if these latest changes look good to you it can go in.

src/comm/TcpAcceptor.cc
```diff
@@ -323,8 +323,7 @@ Comm::TcpAcceptor::notify(const Comm::Flag flag, const Comm::ConnectionPointer &
     if (theCallSub != NULL) {
         AsyncCall::Pointer call = theCallSub->callback();
         CommAcceptCbParams &params = GetCommParams<CommAcceptCbParams>(call);
-        params.xaction = new MasterXaction(XactionInitiator::initClient);
-        params.xaction->squidPort = listenPort_;
+        params.xaction = MasterXaction::MakePortful(listenPort_);
```
rousskov, line 326: This call will assert on FTP data transactions that have no (and should have no) PortCfg.

Please move MasterXaction creation back to the TcpAcceptor notification _recipient_. This move will delay MasterXaction creation, but we have the right to delay it by stating that Comm::TcpAcceptor is just "low-level" Comm code that should not deal with such complicated transaction-level issues as MasterXaction creation.

I think this is a fairly simple change but, unless @yadij  stops you, I would do it in a yet another separate PR to avoid splitting this one later (which wastes even more time).

rousskov (changes requested):

src/comm/TcpAcceptor.cc
yadij: @rousskov, I think we have here another situation demonstrating that class MasterXaction is not actually supposed to be representing your "master transaction" concept.

A "class MasterXaction" object is supposed to be just the root typename of a structure storing multiple transactions details. In the XactionInitiator::initClient code paths it starts with the client connecting and filled out by TcpAcceptor with that clients TCP transaction/connection related state (XactionInitiator, PortCfg, Comm::Connection).

The state for what you are insiting on calling a "master transaction" is not added to the class until the Http1::Server Job receives the MasterXaction object. Which does not (and should not) always happen.

yadij: FWIW; the FTP DATA transactions TcpAcceptor should be receiving an XactionInitiator value indicating that it is a FTP client. The TcpAcceptors subscriber should link the new MasterXactino object as a child transaction of the MasterXaction object of the FTP CTRL connection.

I am okay with this major change/refactor being deferred to a later PR. Should be enough to make TcpAcceptor aware of the initiator enum value.

rousskov: > Amos: @rousskov, I think we have here another situation demonstrating that class MasterXaction is not actually supposed to be representing your "master transaction" concept.

We do not AFAICT.

> A "class MasterXaction" object is supposed to be just the root typename of a structure storing multiple transactions details. In the XactionInitiator::initClient code paths it starts with the client connecting and filled out by TcpAcceptor with that clients TCP transaction/connection related state (XactionInitiator, PortCfg, Comm::Connection).

I do not know what the "root typename of a structure" is exactly, but the rest of the above paragraph can be interpreted, with some effort, as being in line with what MasterXaction class is for.

> The state for what you are insiting on calling a "master transaction" is not added to the class until the Http1::Server Job receives the MasterXaction object. Which does not (and should not) always happen.

AFAICT, you do not know what I am insisting on because your statements do not match my understanding of what MasterXaction is for. Specifically, master transaction details are collected at various times, in various places. For example, TcpAcceptor currently collects contemporary PortCfg and accepted client connection detail for master transactions initiated by HTTP user agents.

yadij: > > Amos: @rousskov, I think we have here another situation demonstrating that class MasterXaction is not actually supposed to be representing your "master transaction" concept.
> 
> We do not AFAICT.
> 

That is half of the problem.

> > A "class MasterXaction" object is supposed to be just the root typename of a structure storing multiple transactions details. In the XactionInitiator::initClient code paths it starts with the client connecting and filled out by TcpAcceptor with that clients TCP transaction/connection related state (XactionInitiator, PortCfg, Comm::Connection).
> 
> I do not know what the "root typename of a structure" is exactly, 
> 

There is the barrier.

> > The state for what you are insisting on calling a "master transaction" is not added to the class until the Http1::Server Job receives the MasterXaction object. Which does not (and should not) always happen.
> 
> AFAICT, you do not know what I am insisting on because your statements either do not match or do not contradict my understanding of what MasterXaction is for. Specifically, master transaction details are collected at various times, in various places. For example, TcpAcceptor currently collects contemporary PortCfg and accepted client connection detail for master transactions initiated by HTTP user agents. Later, the Server job may add more details, of course.

I know that I created class MasterXaction **specifically** as an object to be created by *_port listening TcpAcceptor (and any future matching UdpAcceptor, etc) and passed down through the related Job objects accumulating data each step of the way. This PR current implementation is a good clean demonstration of that use-case passing the initiator and port details down to sub-transactions.

The fact that you keep making requests like this thread OP and (in other PRs like #865) insisting that class MasterXaction "should not be used" for its own original use-case clearly demonstrates that your "master transaction" concept is **different** from the class's intended purpose/meaning.

I suggest that we leave the redefinition and rename of "class MasterXaction" to PR #865. This PR current use of the class is correct per its creator (me). We only need to fix that assertion bug you spotted in FTP sub-transactions use of TcpAcceptor.

rousskov: > the FTP DATA transactions TcpAcceptor should be receiving an XactionInitiator value indicating that it is a FTP client.

This is not a good idea because when dealing with an FTP data connection, TcpAcceptor should not create a new MasterXaction object (that object already exists) and, hence, TcpAcceptor does not need to "receive XactionInitiator value indicating that it is a FTP client" -- it would have no use for that value.

> The TcpAcceptors subscriber should link the new MasterXactino object as a child transaction of the MasterXaction object of the FTP CTRL connection.

This is not a good idea, for at least one and may be more reasons:

1. MasterXactions do not have and should not have child MasterXactions. The whole idea behind the MasterXaction class is that it encompasses all parts/activities related to the originating "control", "primary", or "master" transaction. In FTP case, that originating transaction is the FTP client transaction on the FTP control connection. The DATA transfer is one of those parts/activities the FTP MasterXaction object may detail (probably via another object, but certainly _not_ via another MasterXaction object).
2. FTP data transfer is not really a transaction. It is a _part_ of a transaction. In theory, we could change that, but I doubt it is a good idea to view that data transfer as a transaction (of any sort) because of the unidirectional nature of the FTP information exchange that happens during that transfer. There is no request-response or similar exchange that folks normally associate with a term "transaction".

> I am okay with this major change/refactor being deferred to a later PR.

Even though we obviously disagree on whether the ideas in your [comment](https://github.com/squid-cache/squid/pull/993#discussion_r824255689) should be implemented (in any PR), I hope we can make progress with _this_ sub-sub-PR by implementing the changes I [requested](https://github.com/squid-cache/squid/pull/993#discussion_r823958841). Implementing my request does not really block those ideas from being implemented in the future[^1] -- the request changes are fairly isolated and can be nearly reverted if really needed.  If you see some serious flaws with that request, please propose a better plan to move this (currently buggy) PR forward.

[^1]:  after/if you manage to change my assessment of their correctness

rousskov: > I know that I created class MasterXaction **specifically** as an object to be created by *_port listening TcpAcceptor (and any future matching UdpAcceptor, etc) and passed down through the related Job objects accumulating data each step of the way.

You are not the only MasterXaction class creator, and our 2013 (mis)conceptions do not really matter much now. What really matters is how we move forward from where we are now.

> This PR current implementation is a good clean demonstration of that use-case passing the initiator and port details down to sub-transactions.

"This PR current implementation" is buggy so it is not a "good clean demonstration" of anything. Moreover, some TcpAcceptor objects in the current official code do not start any transactions at all. They simply accept TCP data connections. If all TcpAcceptor objects were starting transactions, this bug and this change request would not exist. However, I certainly agree that creating and passing MasterXaction objects from some transaction-creating contexts does make a lot of sense. There is nothing fundamentally wrong with that idea. It just does not work well in this low-level context.


> The fact that you keep making requests like this thread OP and (in other PRs like #865) insisting that class MasterXaction "should not be used" for its own original use-case clearly demonstrates that your "master transaction" concept is **different** from the class's intended purpose/meaning.

Those facts only demonstrate that our understanding of the intended purpose/meaning differ.

> I suggest that we leave the redefinition and rename of "class MasterXaction" to PR #865. This PR current use of the class is correct per its creator (me). We only need to fix that assertion bug you spotted in FTP sub-transactions use of TcpAcceptor.

I have proposed the changes that will fix that bug. IMO, they do not redefine/rename anything and represent a clean long-term solution. However, if you have a better idea for the fix, please propose it. I see other (dirty) ways to fix this bug, but I do not want to waste our time proposing them if we can simply implement your proposal.

eduard-bagdasaryan: > Please move MasterXaction creation back to the TcpAcceptor notification recipient.

Just to clarify your intention: do you mean that we need moving MasterXaction creation to _all_ existing TcpAcceptor recipients passed via aSub parameter to TcpAcceptor constructors? These are Ftp::Server::acceptDataConnection(), Ftp::Gateway::ftpAcceptDataConnection() and others.

rousskov: > Alex: Please move MasterXaction creation back to the TcpAcceptor notification recipient.
 
> Eduard: Just to clarify your intention: do you mean that we need moving MasterXaction creation to _all_ existing TcpAcceptor recipients passed via aSub parameter to TcpAcceptor constructors? These are Ftp::Server::acceptDataConnection(), Ftp::Gateway::ftpAcceptDataConnection() and others. IIUC, we need this moving because only recipients know the port they are listening on (i.e., distinguish comm_local_port(conn->fd) from TcpAcceptor::_listenPort).

Just to avoid misunderstanding: Besides the two FTP data connection subscribers you have mentioned, there is a third category of notification recipients here -- the ones subscribing via clientListenerConnectionOpened().

I am not sure I can predict/describe all the low-level side effects exactly, but my suggestion should lead to the following situation:

* TcpAcceptor will know nothing about MasterXaction. This class is too low level to deal with MasterXaction (especially with master transaction _creation_) correctly.
* TcpAcceptor subscribers that receive HTTP and FTP requests on *_ports (via clientListenerConnectionOpened()) will create initClient MasterXaction for each new connection they are notified about. For these subscribers, the corresponding MasterXaction creation code currently residing in TcpAcceptor will be moved and adjusted, but the overall outcome will be essentially the same as what happens in the official code today. We are just moving the MasterXaction object creation to the other side of connection notification, where we know it is needed.
* TcpAcceptor subscribers that await FTP data connections (i.e. the two methods you have mentioned) will not create any MasterXaction objects upon receiving that data connection. For these subscribers, the corresponding MasterXaction creation code will be deleted, but the overall outcome will be essentially the same as what happens in the official code today. We are just removing the creation of a MasterXaction object that was ignored by these subscribers anyway. AFAICT, the corresponding two FTP data methods may remain intact -- they are ignoring the received MasterXaction now, and they will continue to do that (i.e. to do nothing) after these changes.

Does this clarify?

yadij: Yes, it clarifies that you want the lifetime of MasterXaction to start on the code line before a ::Server constructor is run (the TcpAcceptor subscriber).

Combining that info with your requirement stated in PR #865 that the MastrerXaction object is **not** stored by ::Server class'es it will thus effectively destruct when the ::Server constructor completes and TcpAcceptor subscriber scope ends.

As a result you have turned MasterXaction into a weird duplicate of CommAcceptParams. Which is useless.

eduard-bagdasaryan: > ::Server class'es it will thus effectively destruct when the ::Server constructor completes and TcpAcceptor subscriber scope ends.

Agreed, but note that this happens already without any changes.
So, regarding MasterXaction created by TcpAcceptor, I see two cases (in the official code):
 * TcpAcceptor subscribers that await FTP data connections. As Alex noted above, these subscribers ignore MasterXaction. We can safely delete MasterXaction creation code.
 * TcpAcceptor subscribers that receive HTTP and FTP requests. These subscribers use MasterXaction as a temporary container object (storing such fields as PortCfgPointer, Connection, etc), which is passed to ::Server hierarchy during initialization and is destructed afterwards. The bona fide MasterXaction is created later for HttpRequest object. We probably do not need to abuse MasterXaction for just initialization purposes - we can create another helper class for that.

After these changes, we can completely remove MasterXaction from TcpAcceptor (and its notification recipients).

rousskov: > Yes, it clarifies that you want the lifetime of MasterXaction to start on the code line before a ::Server constructor is run (the TcpAcceptor subscriber).

My sketch is not detailed enough to specify the exact line, but, yes, my suggestion implies that the MasterXaction object should be created shortly before or while the client connection handling object is created/initialized. The best exact location is to be determined.


> Combining that info with your requirement stated in PR #865 that the MastrerXaction object is **not** stored by ::Server class'es it will thus effectively destruct when the ::Server constructor completes and TcpAcceptor subscriber scope ends.

Correct. Where/when to improve that buggy official code behavior is something we can certainly discuss, but we should keep in mind that this PR did not create that problem, and that the problem does not invalidate the ConnStateData-manages-concurrent-transactions principle.

Server and ConnStateData should not store a MastrerXaction object (directly) because those classes are meant to handle many transactions, including many _concurrent_ transactions. We have ClientHttpRequest and other classes that are dedicated to handling a single master transaction. Those single-transaction classes should store the corresponding MasterXaction object. This is the design we should be targeting when refactoring the corresponding code. The current official code does not match that design.

We need to define when/where an HTTP/FTP[^1] client transaction starts (and then create a MasterXaction object close to that moment/event handler). It is a difficult and mostly out of scope question, with several reasonable answers, including these answers related to the first transaction on an HTTP/FTP connection:

[^1]: Where the differences are unimportant, I am going to use HTTP instead of HTTP/FTP for simplicity sake. Most of these discussions apply to master transactions started by http_port, https_port, and ftp_port requests/queries.

The first master transaction on connection C starts around the time Squid:

1. accepts C
2. receives the first TCP byte on C
3. receives the first HTTP byte on C
4. receives the first HTTP request header on C

The key differences between these (and several other possible) designs boil down to two test situations/questions like these:

* The client closes the connection just before sending HTTP bytes. Was there a master transaction in this case?
* The client waits a minute before proceeding with a regular HTTP request. Will that first minute be included in the master transaction statistics like the total lifetime?

Since we want to access-log connections that do not carry full transactions, I think it is best to consider the first master transaction on C to start around the time Squid accepts C -- the fewer "no access to master transaction" special cases we have when evaluating various ACLs, logformat %codes, and `if` statements, the better.

Currently, ClientHttpRequest and other transaction-carrying classes are created way too late. There is also PROXY protocol that does not deserve a _dedicated_ master transaction and, hence, should be associated with the first HTTP transaction that comes after the PROXY protocol header (represented by, again, ClientHttpRequest and other transaction-carrying classes).

We can avoid large refactoring while following the ConnStateData-manages-concurrent-transactions and accept-starts-master-transaction design principles by creating a new MasterXaction object as suggested in this change request. I (still) recommend doing that.

If we want to go a step further, we can also "temporary" store that MasterXaction object in the ConnStateData::firstXaction or similar field and "move" that first master transaction object to the very first parsed or faked request (e.g., in buildHttpRequest()). I would not do these out-of-scope changes in this PR, but if others insist on them being implemented to unblock this PR, we would have to accommodate such a request.

> As a result you have turned MasterXaction into a weird duplicate of CommAcceptParams. Which is useless.

It looks like the official code deficiencies has misled you into misinterpreting the effects of my suggestions: While the current official code abuses MasterXaction for temporary storage of accept(2) details, the MasterXaction class itself is still genuinely useful in numerous other contexts and will be useful in this context once the bugs in the official code are eliminated.

eduard-bagdasaryan: To move this PR forward, I addressed the initial request at f46e9ed.

src/client_side.cc
```diff
@@ -2303,22 +2303,24 @@ ConnStateData::acceptTls()
 void
 httpAccept(const CommAcceptCbParams &params)
 {
-    MasterXaction::Pointer xact = params.xaction;
-    AnyP::PortCfgPointer s = xact->squidPort;
+    assert(params.port);
```
rousskov, line 2306:
```suggestion
    Assure(params.port);
```

src/client_side.cc
```diff
@@ -2503,22 +2505,25 @@ httpsSslBumpAccessCheckDone(Acl::Answer answer, void *data)
 static void
 httpsAccept(const CommAcceptCbParams &params)
 {
-    MasterXaction::Pointer xact = params.xaction;
-    const AnyP::PortCfgPointer s = xact->squidPort;
+    assert(params.port);
```
rousskov, line 2508:
```suggestion
    Assure(params.port);
```

src/servers/FtpServer.cc
```diff
@@ -241,21 +241,24 @@ Ftp::Server::noteBodyConsumerAborted(BodyPipe::Pointer ptr)
 void
 Ftp::Server::AcceptCtrlConnection(const CommAcceptCbParams &params)
 {
-    MasterXaction::Pointer xact = params.xaction;
-    AnyP::PortCfgPointer s = xact->squidPort;
+    assert(params.port);
```
rousskov, line 244:
```suggestion
    Assure(params.port);
```

rousskov (changes requested):
Thank you for addressing my earlier concern!

I adjusted PR title/description to better match the current PR code. Please check and adjust further as needed. I also found a few minor problems and opportunities to use the recently added Assure() macro.

eduard-bagdasaryan (commented):
Done (88e1b60).

src/servers/FtpServer.cc
```diff
@@ -241,21 +241,24 @@ Ftp::Server::noteBodyConsumerAborted(BodyPipe::Pointer ptr)
...
         // Its possible the call was still queued when the client disconnected
-        debugs(33, 2, s->listenConn << ": FTP accept failure: " << xstrerr(params.xerrno));
+        debugs(33, 2, params.port->listenConn << ": FTP accept failure: " << xstrerr(params.xerrno));
         return;
     }
 
     debugs(33, 4, params.conn << ": accepted");
     fd_note(params.conn->fd, "client ftp connect");
 
+    auto xact = MasterXaction::MakePortful(params.port);
```
rousskov, line 257: I may be missing something, but it looks like this and several very similar places can use `const` because the `xact` pointer is not modified and the Server class constructor accepts a constant pointer. Please check and, if possible, add `const` here and in similar such places.

rousskov (dismissed):
Thank you for addressing all of my previous concerns.

@yadij, there were some code changes since you approved this PR (we had to fix an assertion), so you may want to look through the current code again. I plan to clear it for merging once the last cosmetic `const` issue is resolved.

src/servers/FtpServer.cc
eduard-bagdasaryan: I agree that we can use const in this context. Making the similar fix in other contexts required making NewServer() argument constant. Done at e6ae845.

yadij (approved):

src/client_side.cc
```diff
@@ -2299,22 +2299,24 @@ ConnStateData::acceptTls()
...
     }
 
     debugs(33, 4, params.conn << ": accepted");
     fd_note(params.conn->fd, "client http connect");
+    const auto xact = MasterXaction::MakePortful(params.port);
+    xact->tcpClient = params.conn;
 
     // Socket is ready, setup the connection manager to start using it
     auto *srv = Http::NewServer(xact);
+    // XXX: do not abandon the MasterXaction object
```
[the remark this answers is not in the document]
rousskov: FTR, PR #865 problems do not invalidate this XXX. There are, of course, ways to address this XXX correctly.

PR #994, opened 2022-03-04
Fix ignore-cc/act-as-origin in wildcard split-stack ports

eduard-bagdasaryan:
The PortCfg::clone() hack (and clone_http_port_list() before it) forgot
to copy those two flags to the IPv4 port variant.

Compilers will now be able to warn us if copying misses future members.

Also prohibited other forms of copying, nearly restricting copying to
the parsing code with special needs.

### DIFF
--- a/src/anyp/PortCfg.cc
+++ b/src/anyp/PortCfg.cc
@@ -57,7 +57,7 @@ AnyP::PortCfg::~PortCfg()
 }
 
 AnyP::PortCfg::PortCfg(const PortCfg &other):
-    next(),
+    next(), // special case; see assert() below
     s(other.s),
     transport(other.transport),
     name(other.name ? xstrdup(other.name) : nullptr),
@@ -73,7 +73,7 @@ AnyP::PortCfg::PortCfg(const PortCfg &other):
     disable_pmtu_discovery(other.disable_pmtu_discovery),
     workerQueues(other.workerQueues),
     tcp_keepalive(other.tcp_keepalive),
-    listenConn(),
+    listenConn(), // special case; see assert() below
     secure(other.secure)
 {
     // to simplify, we only support port copying during parsing
--- a/src/anyp/PortCfg.cc
+++ b/src/anyp/PortCfg.cc
@@ -8,6 +8,7 @@
 
 #include "squid.h"
 #include "anyp/PortCfg.h"
+#include "anyp/UriScheme.h"
 #include "comm.h"
 #include "fatal.h"
 #include "security/PeerOptions.h"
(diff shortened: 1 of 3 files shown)


yadij (commented):
FTR: This violates the precepts laid down by Alex that no copy-construct or copy-assign operator shall **ever** do less than a full copy of the original object.

This is the sole reason clone() exists here and ~10 other classes. Apparently it is/was more important to follow that rule than having the compiler catch dev mistakes. If Alex has relaxed his insistence on that we have to review and fix this same issue in them.

rousskov:
> This violates the precepts that no copy-construct or copy-assign operator shall ever do less than a full copy of the original object.

@yadij, can you clarify what you mean by "this"? In other words, which PR-related copying method does "less than a full copy"? AFAICT, the copy constructor modified in this PR copies _everything_.

src/anyp/PortCfg.cc
```diff
@@ -55,29 +56,39 @@ AnyP::PortCfg::~PortCfg()
...
 
-    return b;
+AnyP::PortCfg *
+AnyP::PortCfg::ipV4clone() const
+{
+    const auto clone = new PortCfg(*this);
+    clone->s.setIPv4();
+    debugs(3, 3, AnyP::UriScheme(transport.protocol).image() << "_port: " <<
+           "cloned wildcard address for split-stack: " << s << " and " << clone->s);
+    return clone;
```
rousskov, line 91: > This is the sole reason clone() exists here and ~10 other classes. Apparently it is/was more important to follow that rule than having the compiler catch dev mistakes.

FWIW, all of the above statements make no sense to me. A clone() method should copy everything as well and, where possible, should be implemented using a copy constructor (for many reasons, including so that the compiler can catch a developer forgetting to copy a data member).

According to the PR, the official clone() was buggy. The PR code removes that clone() method and protects the new special ipv4clone() method from accidentally forgetting to copy something. If you see a problem with this PR, please try to be more specific when describing it.

FWIW, since ipv4clone() is not named clone(), the new method has the right to do special new object modifications/adjustments -- the caller should not be surprised by them. If you can suggest a better name for that new method, please do so.

yadij:
> > This violates the precepts that no copy-construct or copy-assign operator shall ever do less than a full copy of the original object.
> 
> @yadij, can you clarify what you mean by "this"? In other words, which PR-related copying method does "less than a full copy"? AFAICT, the copy constructor added in this PR copies _everything_.

Except that next() and listenConn() are not copied. They are default-initialized.

src/anyp/PortCfg.cc
```diff
@@ -55,29 +56,39 @@ AnyP::PortCfg::~PortCfg()
...
-    b->connection_auth_disabled = connection_auth_disabled;
-    b->workerQueues = workerQueues;
-    b->ftp_track_dirs = ftp_track_dirs;
-    b->disable_pmtu_discovery = disable_pmtu_discovery;
-    b->tcp_keepalive = tcp_keepalive;
-    b->secure = secure;
+    // to simplify, we only support port copying during parsing
+    assert(!other.next);
+    assert(!other.listenConn);
+}
```
rousskov, line 82: > Alex: which PR-related copying method does "less than a full copy"? AFAICT, the copy constructor added in this PR copies everything.

> Amos: Except that next() and listenConn() are not copied. They are default-initialized.

They are effectively copied -- their default initialization values are asserted to be the same as the values of those fields in the object being copied. Thus, the effect of this copy constructor is "everything is copied", as it should be.

If you insist, we can explicitly copy those nil values as well. IMO, the current code is better because its intent is clearer: We cannot meaningfully copy those two fields (and that is OK if and only if there is nothing to copy). The source code comment helps relay that idea.

We can also add assertions that the default-initialized `next` and `listenConn` values (in `this`) are also nil.

I considered also asserting that `next` and `other.next` are the same, but I think that would send the wrong message -- the code should avoid comparing raw port and connection pointers.

yadij: > > Alex: which PR-related copying method does "less than a full copy"? AFAICT, the copy constructor added in this PR copies everything.
> 
> > Amos: Except that next() and listenConn() are not copied. They are default-initialized.
> 
> They are effectively copied -- their default initialization values are asserted to be the same as the values of those fields in the object being copied. Thus, the effect of this copy constructor is "everything is copied", as it should be.
> 
> If you insist, we can explicitly copy those nil values as well. IMO, the current code is better because its intent is clearer: We cannot meaningfully copy those two fields (and that is OK if and only if there is nothing to copy). The source code comment helps relay that idea.
> 

If you are sure that nothing goes wrong with the case that they are not nil, I will trust you on that and accept.


> We can also add assertions that the default-initialized `next` and `listenConn` values (in `this`) are also nil.
> 

Okay/good.

> I considered also asserting that `next` and `other.next` are the same, but I think that would send the wrong message -- the code should avoid comparing raw port and connection pointers.

Um, it should not matter whether the pointer is raw or not for such a test. Even a safe-pointer comparison is just checking the raw value is identical. So caring enough to impose a "should" is not reasonable IMO.

rousskov: > If you are sure that nothing goes wrong with the case that they are not nil, I will trust you on that and accept.

I am reasonably sure that this code only runs during (re)configuration -- parsePortCfg() -- when the being-copied port is freshly created with those two fields set to nil. The PR code asserts their nil state (and hides port copying from the rest of the code). This is the best we can do for now AFAICT.

Long-term, those fields should not be in the PortCfg class, of course.

> it should not matter whether the pointer is raw or not for such a test

I agree. The word "raw" in my explanation was just misleading noise.

rousskov (approved):
LGTM

rousskov:
@yadij, I have cleared this PR for merging, but you have a few days to stop it from going in if I misinterpreted your current opinion.

yadij (approved):

PR #996, opened 2022-03-16
Move generatingConnect flag to HttpRequest

yadij:
It is a flag to control ACL matching against the HttpRequest data
and should not have been directly part of MasterXaction at all.

### DIFF
--- a/src/HttpRequest.h
+++ b/src/HttpRequest.h
@@ -193,6 +193,9 @@ public:
     /// whether we have responded with HTTP 100 or FTP 150 already
     bool forcedBodyContinuation;
 
+    /// whether we are currently creating a CONNECT header (to be sent to peer)
+    bool generatingConnect = false;
+
 public:
     bool multipartRangeRequest() const;
 
--- a/src/acl/AtStep.cc
+++ b/src/acl/AtStep.cc
@@ -42,12 +42,7 @@ ACLAtStepStrategy::match(ACLData<XactionStep> * &data, ACLFilledChecklist *check
         if (!checklist->request)
             return 0; // we have warned about the missing request earlier
 
-        if (!checklist->request->masterXaction) {
-            debugs(28, DBG_IMPORTANT, "ERROR: Squid BUG: at_step GeneratingCONNECT ACL is missing master transaction info. Assuming mismatch.");
-            return 0;
-        }
-
-        return checklist->request->masterXaction->generatingConnect ? 1 : 0;
+        return checklist->request->generatingConnect ? 1 : 0;
     }
 
     return 0;
--- a/src/clients/HttpTunneler.cc
+++ b/src/clients/HttpTunneler.cc
@@ -154,7 +154,7 @@ Http::Tunneler::writeRequest()
     MemBuf mb;
 
     try {
-        request->masterXaction->generatingConnect = true;
+        request->generatingConnect = true;
 
         mb.init();
         mb.appendf("CONNECT %s HTTP/1.1\r\n", url.c_str());
@@ -168,10 +168,10 @@ Http::Tunneler::writeRequest()
         hdr_out.clean();
         mb.append("\r\n", 2);
 
-        request->masterXaction->generatingConnect = false;
+        request->generatingConnect = false;
     } catch (...) {
         // TODO: Add scope_guard; do not wait until it is in the C++ standard.
-        request->masterXaction->generatingConnect = false;
+        request->generatingConnect = false;
         throw;
     }
 
--- a/src/HttpRequest.cc
+++ b/src/HttpRequest.cc
@@ -247,6 +247,7 @@ HttpRequest::inheritProperties(const Http::Message *aMsg)
     myportname = aReq->myportname;
 
     forcedBodyContinuation = aReq->forcedBodyContinuation;
+    generatingConnect = aReq->generatingConnect;
 
     // main property is which connection the request was received on (if any)
     clientConnectionManager = aReq->clientConnectionManager;
--- a/src/MasterXaction.h
+++ b/src/MasterXaction.h
@@ -55,9 +55,6 @@ public:
     /// the initiator of this transaction
     XactionInitiator initiator;
 
-    /// whether we are currently creating a CONNECT header (to be sent to peer)
-    bool generatingConnect = false;
-
     // TODO: add state from other Jobs in the transaction
 };
 


src/acl/AtStep.cc
```diff
@@ -42,12 +42,7 @@ ACLAtStepStrategy::match(ACLData<XactionStep> * &data, ACLFilledChecklist *check
         if (!checklist->request)
             return 0; // we have warned about the missing request earlier
 
-        if (!checklist->request->masterXaction) {
-            debugs(28, DBG_IMPORTANT, "ERROR: Squid BUG: at_step GeneratingCONNECT ACL is missing master transaction info. Assuming mismatch.");
-            return 0;
-        }
-
-        return checklist->request->masterXaction->generatingConnect ? 1 : 0;
+        return checklist->request->generatingConnect ? 1 : 0;
```
eduard-bagdasaryan, line 45:
```suggestion
        return checklist->request->generatingConnect;
```

eduard-bagdasaryan: Interesting. I thought an inplicit integral [promotion](https://en.cppreference.com/w/cpp/language/implicit_conversion) should take place here, where true becomes 1.

rousskov (changes requested):
> It is a flag to control ACL matching against the HttpRequest data

What makes you think that? AFACIT, commit 090f1d3 reached (and painstakingly documented) essentially the opposite conclusion.

> and should not have been directly part of MasterXaction at all.

That may be true long-term, but I am not sure there is a better class to place this flag right now.

How were this PR changes tested?

src/acl/AtStep.cc
rousskov: > boolean true does not always implicitly convert to `1`.

It does in C++ [AFAICT](https://stackoverflow.com/a/2725063).

However, for consistency sake, we may return 1s and 0s from match() that returns 0s and 1s in other places.

yadij: @eduard, you are right. I was thinking of the bool representation which can be the ```-1``` literal.

src/clients/HttpTunneler.cc
```diff
@@ -154,7 +154,7 @@ Http::Tunneler::writeRequest()
     MemBuf mb;
 
     try {
-        request->masterXaction->generatingConnect = true;
+        request->generatingConnect = true;
```
rousskov, line 157: AFAICT, the `request` here is the request _received_ from the client or equivalent. Commit 090f1d3 message agrees with that conclusion.

We should not be modifying the received request when creating a new request to be sent. To the extent possible, the received request should be treated as (and will eventually become) constant in this context.

There are many places in the current code where HttpRequest is abused to store unrelated information, but I see no reason to add one more such place. The existing MasterXaction-based code does not have this problem.

src/HttpRequest.h
```diff
@@ -193,6 +193,9 @@ class HttpRequest: public Http::Message
     /// whether we have responded with HTTP 100 or FTP 150 already
     bool forcedBodyContinuation;
 
+    /// whether we are currently creating a CONNECT header (to be sent to peer)
+    bool generatingConnect = false;
```
rousskov, line 197: > PR description: It is a flag to control ACL matching against the HttpRequest data

> Alex: What makes you think that? AFACIT, commit [090f1d3](https://github.com/squid-cache/squid/commit/090f1d3c3c4e1aa9ab96b84787907967313ccfcf) reached (and painstakingly documented) essentially the opposite conclusion.

> Amos: The actual current code usage of this flag.

That answer does not compute for me because the "actual current code usage of this flag" does not "control ACL matching against the HttpRequest data". The current code uses this flag to determine whether the current code context is the `CONNECT` request generation context. That generation uses many objects, including HttpRequest and MasterXaction. More specifically, the `at_step` ACL matches the current master transaction processing step, not "HttpRequest data". The transaction processing step concept lives on a level higher than HttpRequest.

> It should really be a flag of HttpStateData object, but that object is not passed to the access control code.

That would not be a good idea because `CONNECT` generation is not always related to or uses HttpStateData objects.

> The HttpRequest object used by HttpStateData::httpBuildRequestHeader() is where those access controls get details from HttpStateData.

The primary ACL  relevant here -- `at_step` -- gets the current step information from ConnStateData and MasterXaction objects. The ConnStateData source is a temporary flaw -- that source should be MasterXaction or equivalent instead. The MasterXaction source is not perfect either, but it is the best we can do for now AFAICT.

In summary, the key assumption behind this PR remains unsubstantiated and contradicts commit 090f1d3 conclusions. I believe it is also false, but there is another change request dedicated to that problem.

rousskov (changes requested):

src/HttpRequest.h
yadij: > > PR description: It is a flag to control ACL matching against the HttpRequest data
> 
> > Alex: What makes you think that? AFACIT, commit [090f1d3](https://github.com/squid-cache/squid/commit/090f1d3c3c4e1aa9ab96b84787907967313ccfcf) reached (and painstakingly documented) essentially the opposite conclusion.
> 
> > Amos: The actual current code usage of this flag.
> 
> Alex:  That answer does not compute for me because the "actual current code usage of this flag" does not "control ACL matching against the HttpRequest data".
> The current code uses this flag to determine whether the current code context is the `CONNECT` request generation context.


> That generation uses many objects, including HttpRequest and MasterXaction

Specifically it uses HttpRequest as the primary object outside ACL namespace to locate this flag.
 ```
-    return checklist->request->masterXaction->generatingConnect ? 1 : 0;
+    return checklist->request->generatingConnect ? 1 : 0;
```

> . More specifically, the `at_step` ACL matches the current master transaction processing step, not "HttpRequest data".

You are comparing different things. Abstract description of **entire** ACL purpose naturally is not the same as where the implementation gets data to check.

Also, as I am sure you know one piece of "HttpRequest data" is the up-reference to which "master transaction" the HttpRequest is a part of.

> The transaction processing step concept lives on a level higher than HttpRequest.
> 

Indeed, abstract concept does exist at a higher level than any one of the components implementing that concept.

Please be aware that there are at least 2 layers of transaction nested between the HttpStateData context which is performing the "current transaction step" and the "master transaction" context by any reasonable definition of "master" and "transaction". The "master transaction" several layers earlier may have nothing to do with this need to generate a CONNECT tunnel.

By the time this code occurs there are potentially 6 HttpRequest objects in play representing the different layers of transaction nesting and adaptation sub-transactions (all of which are "master transaction", hmm). This flag indicates the one being used by ACL as data source (checklist->request) represents a transaction generating a new HttpRequest with CONNECT method.

It is wrong to set in MasterXaction because ACL checks are not necessarily synchronous and this flag should not affect any of the other transaction HttpRequest objects linked to the MasterXaction.

AFAICT this code only works today because the ACLs and all HttpStateData methods using the flag are completely synchronous for the entire period it is set to true.


> > It should really be a flag of HttpStateData object, but that object is not passed to the access control code.
> 
> That would not be a good idea because `CONNECT` generation is not always related to or uses HttpStateData objects.

This "CONNECT request generation context" is part of the HttpStateData internal logics. So yes it is always **directly** related to and **requires** an HttpStateData object.

> 
> > The HttpRequest object used by HttpStateData::httpBuildRequestHeader() is where those access controls get details from HttpStateData.
> 
> The primary ACL relevant here -- `at_step` -- gets the current step information from ConnStateData and MasterXaction objects. The ConnStateData source is a temporary flaw -- that source should be MasterXaction or equivalent instead. The MasterXaction source is not perfect either, but it is the best we can do for now AFAICT.
> 

No. ```return checklist->request->....generatingConnect ? 1 : 0;```

HttpRequest is better than MasterXaction as demonstrated by this PR code simplification and reducing side effects of the flag being set.

> In summary, the key assumption behind this PR

Which unstated assumption are you refering to?

src/clients/HttpTunneler.cc
yadij: > AFAICT, the `request` here is the request _received_ from the client or equivalent.

The "or equivalent" is important. That means;
 a) there may not be a client, or
 b) the request may be completely different from the one the client sent, or
 c) the client sent it wrapped in nested protocols that we secretly unwrapped without its consent, or
 d) this HttpRequest is just our (HttpStateData's) temporary representation of one of the above.

yadij: Regarding your const'ness statement. The HttpRequest is not const today, it cannot be const so long as the HTTP request is being processed. This code is part of the HTTP processing.

Please stop referring to "commit https://github.com/squid-cache/squid/commit/090f1d3c3c4e1aa9ab96b84787907967313ccfcf" - the deep technical reason documented there for rejecting this (case 1.2) approach is "It feels strange to consult such an old/"disconnected" message". Time between the parent HttpRequest starting and this new tunnel request being generated is not a valid consideration.

PR #997, opened 2022-03-16
Logformat %lp expands to "-" in wildcard listening port configs

rousskov:
FindListeningPortAddress() and its helpers look for "local" address of
an accepted connection. When FindListeningPortAddress() is called in %la
context, we must skip Ip::Address objects with INADDR_ANY IP addresses
because we are looking for a specific IP address, not a wildcard.
However, when called in %lp context, skipping those Ip::Address objects
may, in some cases, result in skipping the only object that actually
contains the port information, resulting in %lp expanding to "-".

Similarly, zero-port Ip::Address objects (with non-any IPs) could be, in
theory, returned instead of continuing the search for an object with a
non-zero port number, although this case was not observed in tests.

Now we configure the address searching helpers with a "good Ip::Address
object" filter so that each %code code path can customize its search.

The problem was introduced in commit ea35939 that expanded INADDR_ANY
check scope from %la to %lp.


src/HttpRequest.h
```diff
@@ -285,5 +285,10 @@ void UpdateRequestNotes(ConnStateData *csd, HttpRequest &request, NotePairs cons
 /// nil parameter(s) indicate missing caller information and are handled safely
 const Ip::Address *FindListeningPortAddress(const HttpRequest *, const AccessLogEntry *);
 
+/// \returns listening/*_port port number used by the client connection (or zero)
+/// nil parameter(s) indicate missing caller information and are handled safely
+unsigned short FindListeningPortNumber(const HttpRequest *, const AccessLogEntry *);
```
rousskov, line 290: We can rename the old/existing FindListeningPortAddress() function to FindListeningPortIp() to improve the distinction between the two similar functions. I did not rename, keeping the diff small and to-the-point, but I am happy to rename if others think it is a good idea.

yadij: Ah, the multiple meanings of "port".

"PortIp" looks like it is missing the word "And" which would be wrong. "Port Address" is correct for the detail being located.

src/HttpRequest.h
```diff
@@ -285,5 +285,10 @@ void UpdateRequestNotes(ConnStateData *csd, HttpRequest &request, NotePairs cons
...
+
+
```
yadij, line 292: double empty line.

yadij (changes requested):

src/HttpRequest.h
rousskov: I am OK with leaving FindListeningPortAddress() name as is, but if you prefer FindListeningPortIpAddress() or some other name, please let me know.

src/HttpRequest.h
rousskov: I am surprised you are blocking a PR because of an extra double line, but it is now fixed at [ef62eba](https://github.com/squid-cache/squid/pull/997/commits/ef62ebab2f55687658d94e4eaa3e40247fd2edda).

yadij (approved):

rousskov:
> I have restarted Jenkins build test of the staged commit because the last attempt [1067](https://build.squid-cache.org/job/5-pr-auto/COMPILER=gcc,OS=fedora-33,label=docker-build-host/1067) has failed for what looks like a PR-unrelated reason to me:
> 
> ```
> src/MemObject.h:34: undefined reference to `_ZZN9MemObjecttPoolEvE7thePool'
> ```

@kinkie, could you place check why attempt [1068](https://build.squid-cache.org/job/5-pr-auto/COMPILER=gcc,OS=fedora-33,label=docker-build-host/1068/) has failed as well? Perhaps you have seen similar failures earlier? This PR only adds a function to an unrelated piece of code; it feels like these build/linking failures are not related to this PR changes, but I may be missing something here.

src/HttpRequest.h
yadij: I am surprised you are not aware of standard review procedures. Code style/polish noticed during review always required fixing before merge. Making sure these types of things do not get into the code is one of the maintainer duties you should be performing.

rousskov: IMO, there is no need to block a PR to ask for a new line to be deleted.

PR #999, opened 2022-03-17
WIP: Add RegisteredRunner for CpuAffinity module

yadij:


yadij:
NP: this is not yet fully tested. Specifically the use of RegisteredRunner::finalizeConfig moves from enter_suid() to leave_suid() capabilities. I am not sure if that has any relevance to setting affinity on any other OS. The reconfigure actions do not appear to require enter_suid() so it may be unnecessary for the startup actions too.

src/CpuAffinity.cc
```diff
@@ -64,3 +67,21 @@ CpuAffinityCheck()
     }
 }
 
+class CpuAffinityRr : public RegisteredRunner
+{
+public:
+    virtual void finalizeConfig() override {
```
rousskov, line 73: The finalizeConfig() event has a specific and documented purpose that calling CpuAffinityInit() does not match. We should use a different event for CpuAffinityInit(). I will sketch two mutually exclusive solutions:

1. If CpuAffinityInit() does not require enter_suid() privileges, then it should be called from useConfig():

```suggestion
    virtual void useConfig() override {
```

2. If CpuAffinityInit() requires privileges that useConfig() does not provide, then we need to add a RegisteredRunner event dedicated to that special need and use that here. For example:

```suggestion
    virtual void useConfigWithPriviliges() override {
```

The useConfigWithPriviliges() event can be triggered where CpuAffinityInit() is called now AFAICT.

----

AFAICT, the action in question here is sched_setaffinity(). On Linux, that action requires:

* an effective user ID equal to the real user ID;
* an effective user ID equal to the effective user ID of the thread identified by pid; or
* possession of  the CAP_SYS_NICE capability in the user namespace of the thread pid.

AFAICT, the official code always runs sched_setaffinity() with enter_suid() privileges. Most likely we must continue to do so. In that case, my second suggestion applies.

src/CpuAffinity.cc
```diff
@@ -22,7 +22,8 @@
 
 static CpuAffinitySet *TheCpuAffinitySet = NULL;
 
-void
+/// set CPU affinity for this process on startup
```
rousskov, line 25: This function documentation does not match this function usage -- the function is called during _reconfiguration_ as well, not just on startup.

```suggestion
/// Makes CPU affinity of this process match configured CPU affinity.
/// Assumes that we have never set the CPU affinity or have cleared it.
```

rousskov (changes requested):
Thank you for improving this code.

> The reconfigure actions do not appear to require enter_suid()

AFAICT, _reconfiguration_ is not a problem for this PR -- CpuAffinityReconfigure() in the official code runs with the same privileges as syncConfig() used by the PR code. Thus, this PR does not change affinity reconfiguration privileges.

BTW, in the official code, CpuAffinityReconfigure() runs under enter_suid(), right?

src/CpuAffinity.cc
yadij: > The finalizeConfig() event has a specific and documented purpose that calling CpuAffinityInit() does not match.

IMO, affinity is currently lacking checks that the CPUs match process numbers. That depends on configuration for workers and cache_dir etc.

Also, (performance benefit) affinity assigned before claimMemoryNeeds() allows the OS to ring-fence the memory allocations to the CPU assigned that process.


> AFAICT, the action in question here is sched_setaffinity(). On Linux, that action requires:
> 
>     * an effective user ID equal to the real user ID;
> 
>     * an effective user ID equal to the effective user ID of the thread identified by pid; or
> 
>     * possession of  the CAP_SYS_NICE capability in the user namespace of the thread pid.
> 
> 
> AFAICT, the official code always runs sched_setaffinity() with enter_suid() privileges. Most likely we must continue to do so. In that case, my second suggestion applies.

Thank you for that info. The code running with high privileges should be minimized, which means each runner actually needing it should do the calls itself.

rousskov: > > Alex: The finalizeConfig() event has a specific and documented purpose that calling CpuAffinityInit() does not match.

> IMO, affinity is currently lacking checks that the CPUs match process numbers. That depends on configuration for workers and cache_dir etc.

> Also, (performance benefit) affinity assigned before claimMemoryNeeds() allows the OS to ring-fence the memory allocations to the CPU assigned that process.

The mismatch I was talking about does not depend on the things you are talking about. If you cannot find a matching event, let's add a new (matching) event, but please do not put this code into a mismatching event handler. The finalizeConfig() is a mismatching event.

Assuming you can justify adding a new event, please do not add that new event _before_ claimMemoryNeeds() because doing so may prevent future code handling the new event from allocating certain memory blocks. Please note that claimMemoryNeeds() does _not_ allocate memory; large memory allocations should start _after_ claimMemoryNeeds() completes.

> The code running with high privileges should be minimized, 

Agreed.

> which means each runner actually needing it should do the calls itself.

I have not studied this issue; I am not against this "enter/leave_suid() as needed" approach if you are sure it is the best way forward despite how much each enter_suid() and leave_suid() calls do already.

Please note that both enter_suid() and leave_suid() depend on squid.conf and, hence, they should not be called until _after_ the finalizeConfig() event completes (at least).

src/CpuAffinitySet.cc
```diff
@@ -52,19 +56,27 @@ CpuAffinitySet::apply()
...
+            enter_suid();
         if (sched_setaffinity(0, sizeof(theOrigCpuSet), &theOrigCpuSet)) {
             int xerrno = errno;
             debugs(54, DBG_IMPORTANT, "ERROR: failed to restore original CPU "
                    "affinity for process PID " << getpid() << ": " <<
                    xstrerr(xerrno));
         }
         CPU_ZERO(&theOrigCpuSet);
+        if (!reconfiguring)
+            leave_suid();
```
rousskov, line 79: > PR commit 7ec84c528813c745c9896070dd2764d986036cbe: Restrict SUID changes to startup
> reconfigure runner already operates with enter_suid() context

I am not against CpuAffinitySet requiring an suid context, either explicitly (i.e. by calling enter/exit_suid()) or implicitly (i.e. by requiring the higher level code to provide that context for CpuAffinitySet), but CpuAffinitySet should not know anything about "runners" and their relationship with suid contexts, especially when that relationship is undocumented.

* The correct long-term solution for this part of the problem is to make nested enter/exit_suid() calls a no-op (among other suid scope control improvements). That is a difficult project that deserves its own PR. I assume you do not want to block this PR on that future project, but it is your call.
* AFAICT, the correct medium-term solution for this part of the problem is to move these explicit enter/exit_suid() calls (without a `reconfiguring` guard) to useConfig() if (and only if) useConfig() is not executed inside an suid context already. Otherwise, that solution is to remove these explicit calls as unnecessary.

P.S. There is no "reconfigure runner". There are two event handlers dedicated to two different reconfiguration stages/events.

src/CpuAffinity.cc
yadij: CpuAffinityCheck() which is called from here depends on all workers and cache_dir in order to validate the cpu_affinity_map directive value. IMO that meets the criteria for finalizeConfig().

src/CpuAffinitySet.cc
yadij: > I am not against CpuAffinitySet requiring an suid context, either explicitly (i.e. by calling enter/exit_suid()) or implicitly (i.e. by requiring the higher level code to provide that context for CpuAffinitySet), but CpuAffinitySet should not know anything about "runners" and their relationship with suid contexts, especially when that relationship is undocumented.
> 

This is not adding any knowledge of runners. These ```if (!reconfiguring)``` checks are preserving current behaviour of this CpuAffinitySet code being run with low-privileges during reconfigure signal handling. I assume that behaviour was intentional. Otherwise the if-conditions should be dropped.


>     * The correct long-term solution for this part of the problem is to make nested enter/exit_suid() calls a no-op (among other suid scope control improvements). That is a difficult project that deserves its own PR. I assume you do not want to block this PR on that future project, but it is your call.

IMO the correct solution is to follow security best practice when dealing with SUID. That is: to minimize the amount of time spent with high privileges.

For purposes of this PR we are moving this code from a codepath with high privileges to one providing low privileges. Thus we need to add these calls around (only) the syscall which requires high privilege.

> 
>     * AFAICT, the correct medium-term solution for this part of the problem is to move these explicit enter/exit_suid() calls (without a `reconfiguring` guard) to useConfig()

Absolutely not.

The use of Must() exceptions for validation in the CpuAffinitySet code and indirectly the memory allocators it uses, would open a window for the leave_suid() code to be bypassed. Squid would return to normal operations with high SUID privileges after the exception handlers.


> if (and only if) useConfig() is not executed inside an suid context already. Otherwise, that solution is to remove these explicit calls as unnecessary.
> 

I have audited the Runners events and they all (except bootstrapConfig) run in the leave_suid() privilege level. So according to your information about ```sched_set*()```requiring root these are necessary (at least on startup/shutdown).

src/CpuAffinity.cc
```diff
@@ -64,3 +68,23 @@ CpuAffinityCheck()
...
+        if (IamPrimaryProcess())
+            CpuAffinityCheck();
+    }
+
+    virtual void useConfig() override {
+        CpuAffinityInit();
+    }
+
+    virtual void syncConfig() override {
+        finalizeConfig();
```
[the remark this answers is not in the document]
yadij: That is a subjective code style opinion. I'm not going to fight against the bloat though, so done in [443c819](https://github.com/squid-cache/squid/pull/999/commits/443c81917c92a341689fd2c1cd5ea13905a5403e)

src/CpuAffinitySet.cc
rousskov: > > I am not against CpuAffinitySet requiring an suid context, but CpuAffinitySet should not know anything about "runners" and their relationship with suid contexts, especially when that relationship is undocumented.

> This is not adding any knowledge of runners.

The proposed code relies on the assumption that the useConfig()/initialization event currently runs outside an suid context and the syncConfig()/reconfiguration event currently runs inside an suid context.

> These `if (!reconfiguring)` checks are preserving current behaviour of this CpuAffinitySet code being run with low-privileges during reconfigure signal handling. I assume that behaviour was intentional. Otherwise the if-conditions should be dropped.

My change request is not about whether this code preserves the current behavior and is not about whether that current behavior is correct. Those are separate issues.

> > * The correct long-term solution for this part of the problem is to make nested enter/exit_suid() calls a no-op (among other suid scope control improvements). That is a difficult project that deserves its own PR. I assume you do not want to block this PR on that future project, but it is your call.
 
> IMO the correct solution is to follow security best practice when dealing with SUID. That is: to minimize the amount of time spent with high privileges.

That is not a solution but a principle. I agree with that principle, and the long-term solution I outlined does not contradict it. In fact, it makes it a lot easier and safer to follow that principle!

Unfortunately, Squid code is not ready to fully follow that principle yet. PR changes also do not fully follow the "minimize the amount of time spent with high privileges" principle, but that is OK -- no reasonable in-scope solution would be able to do that now. PR changes make things worse in other areas, even though they _are_ inspired by the correct principle.  There is an overall better way to handle this situation given the current code limitations.

> For purposes of this PR we are moving this code from a codepath with high privileges to one providing low privileges. Thus we need to add these calls around (only) the syscall which requires high privilege.

This PR does not move this code to a low-privileged section. If it did, these `if` guards would be unnecessary! What actually happens in the current PR code is that the same code runs in two different sections (depending on the event): one that has high privileges and one that has low. That dual placement requires special handing such as these `if` guards. However, these guards create problems for future code improvements.

The medium-term solution I am proposing does not attempt to solve all the problems, but it does not make things worse. It is a uncontroversial step forward. Maybe not as big of a step as we want, but the suid management code is just not ready for bigger steps.

> > * AFAICT, the correct medium-term solution for this part of the problem is to move these explicit enter/exit_suid() calls (without a `reconfiguring` guard) to useConfig()
 
> Absolutely not.
>
> The use of Must() exceptions for validation in the CpuAffinitySet code and indirectly the memory allocators it uses, would open a window for the leave_suid() code to be bypassed. Squid would return to normal operations with high SUID privileges after the exception handlers.

That danger is already present in the official code, and even the current PR code is not completely immune to it! We do not have to address these old/existing problems in _this_ PR. However, if you do want to address them here, you can still follow my medium-term suggestion _and_ add a try/catch block inside the new enter/exit_suid() calls pair.

src/CpuAffinity.cc
yadij: Oh. I see what you mean.

That means we have a documentation issue with finalizeConfig(). It clearly states that it is for ```options that depend on other``` options - where it should say ```are depended on```.

src/CpuAffinitySet.cc
yadij: You are making this situation far more complicated than it needs to be. None of the extra refactoring or handling is necessary if we just accept that the PR current code is better than current official code.

The most likely future changes to this code are someone implementing CAP_SYS_NICE and removing these SUID changes entirely. OR, properly testing the affinity reconfigure behaviour to determine whether the if-guards are necessary.


> It is a uncontroversial step forward.

Disagree. It is controversial as evidenced by our disagreements on security aspects.

> Maybe not as big of a step as we want, but the suid management code is just not ready for bigger steps.

IMO the proposed half-step is worse than the half-step this PR already implements.

src/CpuAffinity.cc
yadij: Request done in [0d6b6bb](https://github.com/squid-cache/squid/pull/999/commits/0d6b6bb4bd2601fdc13a3c28cd9be2834aba3714)

rousskov: > we have a documentation issue with finalizeConfig(). It clearly states that it is for `options that depend on other` options - where it should say `are depended on`.

FWIW, I do not know what documentation issue you are thinking about. Both "options that depend on X" and "options are depended on X" phrases mean the same thing to me, but the second phrase is grammatically [incorrect](https://forum.wordreference.com/threads/is-depended-on-vs-depends-on.3372884/post-17098859).

src/CpuAffinitySet.cc
rousskov: > None of the extra refactoring or handling is necessary if we just accept that the PR current code is better than current official code.

I disagree that the current PR code is better than the official code. PR changes make the code (subject to this change request) worse, not better. Yes, some work is necessary to address this problem (for the time being). I believe that work is not very complex -- no SUID refactoring is needed to find a good-enough solution (while we wait for SUID refactoring which is a big project indeed). If the amount of work is what stops you from moving this PR forward, I can volunteer to do this work.

src/CpuAffinity.cc
```diff
@@ -64,3 +42,38 @@ CpuAffinityCheck()
...
+        CpuAffinityInit();
+    }
+
+    virtual void syncConfig() override {
+        if (TheCpuAffinitySet) {
+            TheCpuAffinitySet->undo();
+            delete TheCpuAffinitySet;
+            TheCpuAffinitySet = nullptr;
+        }
+        CpuAffinityInit();
```
rousskov, line 74: Please do not clear the current affinity _before_ checking whether the new affinity is valid. Eventually, CpuAffinityCheck() will reject invalid configurations. When we reject an invalid configuration during reconfiguration, the old set affinity should stay intact.

It is hard to see what is going on in the diff because you moved the essentially unchanged code, but the old code did not have that flaw. It is easy to fix though AFAICT: Just move the `if` statement calling CpuAffinityCheck() into the beginning of useConfig() and syncConfig() methods. Eventually, there will be a checkConfig() event, and that duplicated code will go into that single event handler.

rousskov (changes requested):

src/CpuAffinitySet.cc
rousskov: > Alex: CpuAffinityReconfigure() in the official code runs with the same privileges as syncConfig() used by the PR code. 

> Amos: I have audited the Runners events and they all (except bootstrapConfig) run in the leave_suid() privilege level. So according to your information about sched_set*()requiring root these are necessary (at least on startup/shutdown).

AFAICT, the combination of the above statements implies that the PR code should call enter/exit_suid() _unconditionally_, addressing this change request. Do you see a problem with that solution?

yadij:
Making a draft for now. I have essentially abandoned this approach, but am leaving the PR open in case someone wants to try and resolve the issues @rousskov identified with the current design.

PR #1001, opened 2022-03-23
SourceLayout: Move time related tools to time/libtime.la

yadij:
Just code moves and documentation.

Solves the occasional build issues of copying time.cc for linking to
pinger and tools/ binaries. Which now link to the library.


src/time/operators.h
```diff
@@ -0,0 +1,53 @@
...
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID__SRC_TIME_OPERATORS_H
+#define SQUID__SRC_TIME_OPERATORS_H
+
+// TODO: Remove direct timercmp() calls in legacy code.
+
+inline bool
+operator <(const timeval &a, const timeval &b)
```
rousskov, line 15: The idea of moving operators into a dedicated header file sounds odd to me: Why treat operators differently than other functions?

src/time/forward.h
```diff
@@ -13,6 +13,7 @@
```
rousskov, line 13: Please update the header guard.

rousskov (changes requested):
Thank you for working on this cleanup. This is a partial review to help you make progress.

src/time/forward.h
```diff
@@ -25,8 +26,21 @@ extern struct timeval current_time;
 extern double current_dtime;
 extern time_t squid_curtime;
 
-time_t getCurrentTime(void);
-int tvSubMsec(struct timeval, struct timeval);
+/// Update squid_curtime, current_time, and current_dtime
+/// \returns new value of squid_curtime
+time_t getCurrentTime();
```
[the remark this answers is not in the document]
yadij: Alex, 21 Dec 2021:
> We (reluctantly) use "gadgets" when we cannot find a common word for the collection of assorted APIs.

The API in this case is ```namespace Time```. Looking towards replacing all these definitions with ```<chrono>``` API calls soon. IMO it is not worth adding a gadgets file, just to remove it in the followups to libtime.

If you insist on not using "forward.h" I will rename this to "old_api.h"

src/time/operators.h
yadij: To restrict the amount of global namespace pollution by non-forward declarations in forward.h. These are a distinct group of operators for timeval which are only used by two callers.

If you have a better filename suggestion, I am open to ideas.

rousskov: > Alex: Why treat operators differently than other functions?
> Amos: To restrict the amount of global namespace pollution by non-forward declarations in forward.h.

These declarations do not belong to `forward.h`, so the pollution you are talking about will not exist in the finished PR and, hence, cannot be used to justify this special treatment of operators.

> Amos: If you have a better filename suggestion, I am open to ideas.

It is not a naming problem -- the group itself should not exist (and, hence, does not need a name). Please move these declarations to `time/gadgets.h` (together with all other declarations that should be there) unless you can propose a better place for them (based on their semantics rather than the fact that they are declared using an `operator` keyword).

rousskov (commented):
Just updating existing change requests...

rousskov (commented):
@yadij, this PR is one of the missing piece required to start using libbase and libsbuf in helpers. AFAICT, all other pieces either have been merged or have a pending (and hopefully not controversial) PRs. I would be happy to apply the requested changes to this PR to reach that coveted goal sooner. Please let me know if you want me to do the next round of modifications. You will be able to undo them if you do not like them, of course.

src/time/forward.h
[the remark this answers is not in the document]
yadij: I picked "old_api.h" for the same reasons that name is used in src/mem/. It is the literal name/term for what these symbols are. These bits are _not_ gadgets of Squids latest Time accessing API (that is namespace Time and Time::Engine).

The symbols here are a mix of definitions from **several** outdated APIs that should not be used going forward. But fixing/replacing them would bloat this PR so much I left them out of scope.

yadij:
> @yadij, this PR is one of the missing piece required to start using libbase and libsbuf in helpers. AFAICT, all other pieces either have been merged or have a pending (and hopefully not controversial) PRs. I would be happy to apply the requested changes to this PR to reach that coveted goal sooner. Please let me know if you want me to do the next round of modifications. 

Your requests are not as small as you may think. I am having to redo a lot of this PR's work to resolve all the include dependency side effects.

rousskov:
> Your requests are not as small as you may think. I am having to redo a lot of this PR's work to resolve all the include dependency side effects.

My offer does not depend on the amount of work.

src/time/forward.h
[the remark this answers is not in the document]
rousskov: > I picked "old_api.h" for the same reasons that name is used in src/mem/. It is the literal name/term for what these symbols are.

The word "old" is not a good term to describe an API. It relays no useful information to the reader -- the API in question is as "new" as it gets to newcomers and as old as most other Squid APIs to experienced Squid developers. And those `<chrono>` APIs can be reasonably called "old" as well -- C++11 is more than 10 years old! "Old" does not imply deprecated, but many of these functions are not going away in the foreseeable future either -- we cannot deprecate the APIs used by system calls, for example. Anyway you look at it, "old" is just not a good choice here.

The (unfortunate) existence of the `src/mem/old_api.cc` file does not invalidate this reasoning. There is no need to elevate that naming mistake to the header file level.

> The symbols here are a mix of definitions from **several** outdated APIs that should not be used going forward. But fixing/replacing them would bloat this PR so much I left them out of scope.

I agree that this PR should not fix/replace those APIs. And it does not matter whether or how many of those APIs are "outdated" because, after this PR, we can and will replace or improve any outdated APIs as needed (and, ideally, without renaming the corresponding header files, but we will make those decisions when we work on those replacements).

> These bits are not gadgets of Squids latest Time accessing API (that is namespace Time and Time::Engine).

They are gadgets (for the lack of a better word). They can be declared together with future gadgets that may or may not be placed inside the Time namespace (that namespace placement decision should be made for each specific new function/type). I am not even sure we need a Time namespace, but this change request is not about that.

src/time/forward.h
yadij: Done in [d53ffe3](https://github.com/squid-cache/squid/pull/1001/commits/d53ffe3f66a8db05d3cc6327960a50594d1a80be)

src/time/operators.h
yadij: Undone in [d53ffe3](https://github.com/squid-cache/squid/pull/1001/commits/d53ffe3f66a8db05d3cc6327960a50594d1a80be)

src/time/forward.h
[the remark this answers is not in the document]
yadij: Done in [d53ffe3](https://github.com/squid-cache/squid/pull/1001/commits/d53ffe3f66a8db05d3cc6327960a50594d1a80be)

yadij:
Please re-review, this has some major differences from before. Least of which is the rfc1123 time representation code overlooked earlier.

src/time/gadgets.h
```diff
@@ -98,29 +147,4 @@ operator ==(const timeval &a, const timeval &b)
...
-
-/** Display time as a formatted human-readable string.
- * Time string syntax used is that of Apache httpd.
- * "DD/MMM/YYYY:hh:mm:ss zzzz"
- *
- * Output is only valid until next call to this function.
- */
-const char *FormatHttpd(time_t t);
-
-} // namespace Time
```
rousskov, line 123: If possible, please return the deleted/moved lines above to their original place to expose the actual changes (if any).

src/time/gadgets.h
```diff
@@ -6,27 +6,83 @@
...
+ * Output is only valid until next call to this function.
+*/
+const char *FormatHttpd(time_t);
+
+} // namespace Time
+
+/* old C struct timeval API */
+
+#include <iosfwd>
+// NP: sys/time.h is included by libcompat
```
rousskov, line 68: I recommend removing this comment. This header should not know what inside `squid.h` provides `sys/time.h`, and anybody curious about it can figure it out.

If you keep it, please either keep the original comment (so I do not need to review/adjust it) or drop the `NP: ` prefix. As we have discussed, it is not a standard/common-enough abbreviation that many readers are likely to recognize, FWIW, I still do not know what it is despite trying to find out!

rousskov (dismissed):
Thank you for addressing my earlier concerns. This PR is almost done, I think!

src/time/gadgets.h
```diff
@@ -6,27 +6,83 @@
...
+ * Time string syntax used is that of Apache httpd.
+ * "DD/MMM/YYYY:hh:mm:ss zzzz"
+ *
+ * Output is only valid until next call to this function.
+*/
+const char *FormatHttpd(time_t);
+
+} // namespace Time
+
+/* old C struct timeval API */
```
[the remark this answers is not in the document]
yadij: I do not recommend adding any code that mixes types from these different APIs. As I have added comments about, they are not compatible with each other even when supplying the same information using the same object types.

Luckily, all these have STL equivalents provided through ```<chrono>``` which should be used instead. I am hoping to do conversion followups, this is scoped at non-logic change.

src/time/gadgets.h
yadij: The change *is* the move itself to group the parts of this squid_curtime API together.

yadij: Done in [82347a0](https://github.com/squid-cache/squid/pull/1001/commits/82347a06be8f3cd9f24ab99432d7713d3fd8c87e)

src/time/gadgets.h
yadij: > I recommend removing this comment. This header should not know what inside `squid.h` provides `sys/time.h`, and anybody curious about it can figure it out.
> 

Copy-n-paste from the old includes list. But fine, done in [82347a0](https://github.com/squid-cache/squid/pull/1001/commits/82347a06be8f3cd9f24ab99432d7713d3fd8c87e).

> If you keep it, please either keep the original comment (so I do not need to review/adjust it) or drop the `NP: ` prefix. As we have discussed, it is not a standard/common-enough abbreviation that many readers are likely to recognize, FWIW, I still do not know what it is despite trying to find out!

Interesting, it is so common over here I had not considered it as being completely unknown anywhere else. With a colon it is "Note Please:", Australian English form of the Latin phrase "Nota Bene".

src/time/gadgets.h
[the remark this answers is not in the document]
yadij: This was to clarify where the boundaries are between the multiple APIs. Fine though, removed in [82347a0](https://github.com/squid-cache/squid/pull/1001/commits/82347a06be8f3cd9f24ab99432d7713d3fd8c87e).

src/time/gadgets.h
rousskov: We have discussed these `NP` problems before. In most Squid contexts (including this one), any such prefix is essentially noise, but if you are sure it is needed in any specific context, please use its more common, better understood, and less ambiguous `N.B.` variant.

src/time/gadgets.h
```diff
@@ -6,27 +6,83 @@
...
+/// the current UNIX time in timeval {seconds, microseconds} format
+extern struct timeval current_time;
+
+/// timeval subtraction operation.
+/// \returns (A-B) in whole microseconds
+/// XXX: result is not compatible with time_msec_t millisecond storage
+int tvSubUsec(struct timeval A, struct timeval B);
+
+/// timeval subtraction operation.
+/// \returns (A-B) in seconds (with microsecond decimal)
```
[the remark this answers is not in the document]
rousskov: > Please fix other occurrences as well.

Done in 7f55800.

rousskov:
I also formatted changed sources in [b5a46c8](https://github.com/squid-cache/squid/pull/1001/commits/b5a46c851b8f141a44dc80acc34f7cafe47fd2f1), but my script does not enforce all the formatting rules yet, so you may want to double check.

rousskov:
```
src/time/old_api.cc:23: multiple definition of `getCurrentTime()';
src/security/cert_generators/file/security_file_certgen.cc:83: first defined here
```

I tried to use the result of this PR, but got the linking error shown above. Not sure why this error was not present in CI tests, but removing getCurrentTime() implementation (done in PR branch commit [e121f0b](https://github.com/squid-cache/squid/pull/1001/commits/e121f0be01ac24f43ff24f8f6744568d30fc2ce7)) feels like the right/necessary fix anyway because the new libtime provides that definition. Please let me know if I missed something.

yadij:
Your test build seems to still contain the file time/old_api.cc that is not present in the tree. I'm good with removing the redundant helper definition of that function, even though it does mean a logic change happens in that helper.

rousskov:
> I'm good with removing the redundant helper definition of that function, even though it does mean a logic change happens in that helper.

What logic changes do you expect from that removal?

FWIW, the helper code does not call getCurrentTime() directly AFAICT. That call was removed in a0b971d.

PR #1003, opened 2022-03-25
Detach libsbuf from StatHist to facilitate SBuf reuse

rousskov:
The stated purpose of sbuf/DetailedStats was to "avoid adding external
dependencies to the SBuf code". DetailedStats failed to accomplish that
because it introduced a dependency on StatHist (which depends on Store).

To break the unwanted dependency, we outsource at-destruction-time size
statistics collection to external-to-SBuf code, configurable via the new
SBufStats fields. This also allows to avoid non-trivial SBuf statistics
collection in programs that do not need that statistics.

Since the new SBufStats fields are set at cache manager configuration
time in mainInitialize(), earlier SBuf and MemBlob destructions are not
accounted for. With more code changes/complications, we could initialize
the fields much earlier, but this delay may be considered a _positive_
change because it removes unusual SBuf stats from long-term histograms.

Also do not use the SBuf STUB in the deployed/non-test pinger program.
TODO: Convert more helpers to use SBuf. This change only adjusts pinger
because that adjustment did not require other significant changes.


src/sbuf/SBuf.cc
```diff
@@ -70,7 +69,7 @@ SBuf::~SBuf()
 {
     debugs(24, 8, id << " destructed");
     --stats.live;
-    recordSBufSizeAtDestruct(len_);
+    SBufStats::RecordSBufSizeAtDestruct(len_);
```
eduard-bagdasaryan, line 72: Since this method may throw ( StatHist is created dynamically), should not we protect our destructor? I just wonder whether such problems are considered minor (and we just turn a blind eye to it).

kinkie: > but that (old and widespread) bug is not for _this_ PR to fix IMO. 

I agree. Non-regression is a good enough criteria for accepting for me

kinkie (approved):
Looks legitimate

yadij (approved):
LGTM, Thank you for closing off this annoying issue.

PR #1005, opened 2022-03-28
Remove SCO 3.2 support

yadij:
This OS is now obsolete. Users wishing to build for this OS
can use build time options:
 ./configure --disable-poll CFLAGS="-lintl"

also add to lib/util.c:
  #define rint(X) floor((X) + 0.5)


kinkie (approved):
Enthusiastic YES!

PR #1008, opened 2022-03-29
Maintenance: Removed a few known unused globals from src/tests

rousskov:


kinkie (approved):
Thanks for cleaning things up!

PR #1009, opened 2022-03-29
comm/libminimal.la to facilitate helper use of convenience libs

rousskov:
Helpers may not (want to) use fd_open() and fd_close() directly, but
they (want to) use libdebug which does use fd_open() and fd_close().

Currently, the Comm "importing" fd_open() and Comm "delisting"
fd_close() APIs are not implemented inside src/comm/ but they do belong
to the Comm module and use its internals. Moving fd.h to src/comm/ will
require a lot of noisy out-of-scope changes deserving a dedicated PR.

Despite their names, the minimally-implemented functions do not open(2)
and close(2) file descriptors in their full implementations either: That
full implementation just updates fd-associated Squid-specific metadata
that Squid helpers do not need/use.

Do not use stub_fd.cc in the deployed/non-test pinger program.

Avoid build-breaking copying of stub_fd.cc source file in test-suite/.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -614,7 +614,6 @@ ufsdump_SOURCES = \
 	cbdata.h \
 	dlink.cc \
 	dlink.h \
-	tests/stub_fd.cc \
 	fd.h \
 	helper/ChildConfig.h \
 	int.cc \
@@ -637,6 +636,7 @@ ufsdump_LDADD = \
 	ipc/libipc.la \
 	mgr/libmgr.la \
 	debug/libdebug.la \
+	comm/libminimal.la \
 	$(XTRA_OBJS) \
 	$(REPL_OBJS) \
 	$(NETTLELIB) \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -614,6 +614,7 @@ ufsdump_SOURCES = \
 	cbdata.h \
 	dlink.cc \
 	dlink.h \
+	tests/stub_fd.cc \
 	fd.h \
 	helper/ChildConfig.h \
 	int.cc \
@@ -636,7 +637,6 @@ ufsdump_LDADD = \
 	ipc/libipc.la \
 	mgr/libmgr.la \
 	debug/libdebug.la \
-	comm/libminimal.la \
 	$(XTRA_OBJS) \
 	$(REPL_OBJS) \
 	$(NETTLELIB) \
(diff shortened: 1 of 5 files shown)


src/Makefile.am
```diff
@@ -637,6 +636,7 @@ ufsdump_LDADD = \
 	ipc/libipc.la \
 	mgr/libmgr.la \
 	debug/libdebug.la \
+	comm/libminimal.la \
```
rousskov, line 639: Disclaimer: I cannot test `ufsdump` changes because that tool does not currently build (`-Wunused-parameter`, `-Wmissing-declarations`, and other problems unrelated to _this_ PR). I think the ufsdump changes in this PR are conceptually correct, but I will remove them if others think that this PR should not touch that rotten tool.

yadij: FWIW; the reason that old code was being kept around for is now unlikely to occur. So I am okay with leaving it untouched here and dropping entirely later.

yadij (commented):
FYI; The initial plan for src/helper/ was to have two libraries. One for Squid main binary and one for generic things like this linked to helper binary. That second does not exist yet because the protocol_defines.h does not need any .cc code.

Please consider completing that plan by moving this stub library to src/helper/.

src/Makefile.am
rousskov: I removed those ufsdump changes in fc9d761 to avoid arguing about them, under the assumption that ufsdump will be removed in #1013.

rousskov:
> The initial plan for src/helper/ was to have two libraries. One for Squid main binary and one for generic things like this linked to helper binary.
>
> Please consider completing that plan by moving this stub library to src/helper/.

I have actually started the series of recent helper-related PRs (to which this PR belongs) with a single library that would "prime" or "bootstrap" helper code with working but minimal versions of various APIs. I eventually rejected that approach because different helpers need a different set of priming implementations, because I was able to make some of the existing libraries reusable without priming tricks (reducing the problem surface), and because a focused library is better than a hodgepodge of different-API implementations.

Another problem with the `src/helper/libprimer.la` approach specifically is that not all non-sbin/squid programs are helpers and some such programs do need to use Squid convenience libraries. My current series of PRs is driven by helper needs, but as some of these PRs already illustrate, this work benefits non-helper programs like squidclient and test-suite cases as well.

We may come back to this idea later as helper needs evolve and require a dedicated helper-specific library. This PR series does not stand in the way of creation of such a library.

yadij (approved):

PR #1011, opened 2022-03-31
Improve debugs() handling in helpers

rousskov:
This change also reduces libdebug dependency on globals.cc, improving
libdebug reusability.

Also do not default-reset debug sections after they were explicitly set:

* In most sbin/squid contexts, DebugModule constructor is called before
  Debug::parseOptions(). That call order results in the Levels array
  being reset to default values before it is reset to configured values.
  No problem.

* In sbin/squid -X context, DebugModule constructor is called after
  Debug::parseOptions(), but the override_X flag protected the Levels
  array from being reset to default values in this case. No problem.

* In helper contexts, DebugModule constructor may be called after
  Debug::parseOptions(), and the override_X flag stays false. This order
  results in the parseOptions() effects erased by ResetSections() called
  from the constructor.

This bug was detected while trying to understand why pinger's hard-coded
(and wrong) ALL,10 default has no effect on pinger's debugging. The two
bugs cancelled each other.

The following changes affect pinger (now) and other helpers that will
eventually use libdebug and its new NameThisHelper() API:

* Label helper debugs() lines with the helper name (e.g., "pinger"),
  similar to how we already label SMP debugs() lines with "kidN". This
  change improves cache.log readability, distinguishes output from
  different helpers, and distinguishes helper output from sbin/squid
  output in non-SMP logs.

* Make sure level-1 debugs() messages are logged.

* Stop pointless accumulation of cache.log and syslog channel messages.

* Automatically honor SQUID_DEBUG environment variable set by the parent
  Squid process. The pinger helper was already honoring it.

The following changes are specific to pinger:

* Removed the now-duplicated references to "pinger" in pinger debugs().

* Do not default pinger debugging levels to ALL,10. The default pinger
  debugging levels should be the same as the default Squid debugging
  levels (i.e. "ALL,1"). Bugs in the debugging module prevented the
  hard-coded "ALL,10" (or any other elevated setting) from having an
  effect, but we now fixed the last of those bugs. AFAICT, the change of
  default from ALL,1 to ALL,10 in commit cc192b5 was accidental.

### DIFF
--- a/src/debug/Stream.h
+++ b/src/debug/Stream.h
@@ -88,6 +88,7 @@ public:
     static int override_X;
     static bool log_syslog;
 
+    // TODO: Convert all helpers to use debugs() and NameThisHelper() APIs.
     /// Use the given name for debugs() messages from this helper process.
     /// Side effect: Commits to using helper-appropriate debug levels/channels.
     /// \sa NameThisKid()
--- a/src/debug/Stream.h
+++ b/src/debug/Stream.h
@@ -88,6 +88,15 @@ public:
     static int override_X;
     static bool log_syslog;
 
+    /// Use the given name for debugs() messages from this helper process.
+    /// Side effect: Commits to using helper-appropriate debug levels/channels.
+    /// \sa NameThisKid()
+    static void NameThisHelper(const char *name);
+
+    /// Use the given ID for debugs() messages from this SMP kid process.
+    /// \sa NameThisHelper()
+    static void NameThisKid(int kidIdentifier);
+
     static void parseOptions(char const *);
 
     /// minimum level required by the current debugs() call
--- a/src/debug/debug.cc
+++ b/src/debug/debug.cc
@@ -56,8 +56,11 @@ static int DefaultStderrLevel = -1;
 /// early debugs() with higher level are not buffered and, hence, may be lost
 static constexpr int EarlyMessagesLevel = DBG_IMPORTANT;
 
+/// pre-formatted name of the current process for debugs() messages (or empty)
+static std::string ProcessLabel;
+
 static const char *debugLogTime(time_t t);
-static const char *debugLogKid(void);
+
 #if HAVE_SYSLOG
 #ifdef LOG_LOCAL4
 static int syslog_facility = 0;
@@ -356,6 +359,45 @@ ResetSections(const int level)
         sectionLevel = level;
 }
 
+/// optimization: formats ProcessLabel once for frequent debugs() reuse
+static void
+LabelThisProcess(const char * const name, const Optional<int> id = Optional<int>())
+{
+    assert(name);
+    assert(strlen(name));
+    std::stringstream os;
+    os << ' ' << name;
+    if (id.has_value()) {
+        assert(id.value() >= 0);
+        os << id.value();
+    }
+    ProcessLabel = os.str();
+}
+
+void
+Debug::NameThisHelper(const char * const name)
+{
+    LabelThisProcess(name);
+
+    // do not restrict helper (i.e. stderr) logging beyond debug_options
+    EnsureDefaultStderrLevel(DBG_DATA);
+
+    BanCacheLogUse();
+    SettleStderr();
+    SettleSyslog();
+}
+
+void
+Debug::NameThisKid(const int kidIdentifier)
+{
+    // to reduce noise and for backward compatibility, do not label kid messages
+    // in non-SMP mode
+    if (kidIdentifier)
+        LabelThisProcess("kid", Optional<int>(kidIdentifier));
+    else
+        ProcessLabel.clear(); // probably already empty
+}
+
 /* LoggingSectionGuard */
 
 LoggingSectionGuard::LoggingSectionGuard()
@@ -556,7 +598,7 @@ DebugChannel::writeToStream(FILE &destination, const DebugMessageHeader &header,
 {
     fprintf(&destination, "%s%s| %s\n",
             debugLogTime(header.timestamp),
-            debugLogKid(),
+            ProcessLabel.c_str(),
             body.c_str());
     noteWritten(header);
 }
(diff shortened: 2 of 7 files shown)


rousskov:
Here is a cache.log sample, showing both first worker and pinger messages of various debugging levels:

```
13:10:50.610 kid1| 51,3| fd.cc(169) fd_open: fd_open() FD 44 Pinger Socket
13:10:50.610 kid1| 51,3| fd.cc(169) fd_open: fd_open() FD 45 Pinger Socket
13:10:50.611 kid1| Pinger socket opened on FD 45
13:10:50.611 kid1| 37,2| IcmpSquid.cc(89) SendEcho: to [::1], opcode 1, len 13
13:10:50.611 kid1| 37,2| IcmpSquid.cc(89) SendEcho: to 127.0.0.1, opcode 1, len 9
13:10:50.612 kid1| 51,3| fd.cc(94) fd_close: fd_close FD 44 Pinger Socket
13:10:50.615 pinger| 3,2| tools.cc(47) ProbeTransport: Detected IPv6 hybrid or v4-mapping stack...
13:10:50.616 pinger| 3,2| tools.cc(70) ProbeTransport: Detected functional IPv6 loopback ...
13:10:50.616 pinger| 3,2| tools.cc(77) ProbeTransport: IPv6 transport Enabled
13:10:50.616 pinger| Initialising ICMP pinger ...
13:10:50.616 pinger| ICMP socket opened.
13:10:50.616 pinger| ICMPv6 socket opened
13:10:50.616 pinger| 42,2| IcmpPinger.cc(199) Recv:  Pass [::1] off to ICMPv6 module.
13:10:50.616 pinger| 42,5| Icmp6.cc(173) SendEcho: Send Icmp6 packet to [::1].
13:10:50.616 pinger| 42,9| Icmp6.cc(186) SendEcho: x=38
```

src/debug/debug.cc
```diff
@@ -352,10 +361,55 @@ DebugStream() {
...
+
+    if (const auto parentProcessDebugOptions = getenv("SQUID_DEBUG")) {
+        assert(!debugOptions);
+        debugOptions = xstrdup(parentProcessDebugOptions);
+    }
+
+    // do not restrict helper (i.e. stderr) logging beyond debug_options
+    EnsureDefaultStderrLevel(DBG_DATA);
+
+    BanCacheLogUse();
```
yadij, line 397: Please add a comment documenting why this is banned.
```suggestion
    BanCacheLogUse(); // our parent process diverts stderr to the cache.log instead
```

src/debug/debug.cc
```diff
@@ -352,10 +361,55 @@ DebugStream() {
...
+    SettleStderr();
+    SettleSyslog();
```
yadij, line 399: FYI; The helpers we bundle should not be using syslog. They should only be using stderr currently.

Please ensure that non-use for consistency until the other helpers are updated to this new debugs API. Then we can do the switch to enable it for all of them at some easily documented vNN release.

src/debug/Stream.h
```diff
@@ -88,6 +88,15 @@ class Debug
     static int override_X;
     static bool log_syslog;
 
+    /// Use the given name for debugs() messages from this helper process.
+    /// Side effect: Commits to using helper-appropriate debug levels/channels.
+    /// \sa NameThisKid()
+    static void NameThisHelper(const char *name);
```
yadij, line 94: I do not see many of the C++ helpers being converted to use this new API. Yet all of them do debug logging (usually via the compat/debug.h API).

Please add a TODO about getting the rest converted so we do not loose track of this.

yadij (changes requested):
Thank you for this cleaning up and highlighting that pinger is not following the (old/existing) helper UI design properly.

There are a few maintenance/release related nits to resolve before this API change is ready for release.

Also, any plans to remove/switch the existing helper debug() API ?

src/debug/debug.cc
rousskov: > Please ensure that non-use for consistency until the other helpers are updated to this new debugs API.

I believe the proposed code does not enable syslog use by helpers: SettleSyslog() does not enable syslog use by default, and existing helpers do not call Debug::ConfigureSyslog() to overwrite that default. I see no pinger messages in syslog when testing this PR.

src/debug/Stream.h
rousskov: > I do not see many of the C++ helpers being converted to use this new API.

I am not aware of any helpers other than pinger that should be converted in _this_ PR. Most helpers should not be converted until the remaining helper-related PRs are merged, and the helpers themselves start using debugs().

> Please add a TODO about getting the rest converted so we do not loose track of this.

Added `Convert all helpers to use debugs() and NameThisHelper() APIs` TODO to the PR description. I am already working on converting some of the helpers, but we need to merge the prerequisite PRs (at least) before it makes sense to post the corresponding helper conversion PRs.

src/debug/debug.cc
rousskov: I added a similar comment in 741bd10.

rousskov (commented):
I believe I have addressed all your concerns as of branch commit 741bd10. Please re-review.

rousskov:
> Also, any plans to remove/switch the existing helper debug() API ?

Nothing specific on my side. Eventually, I think that macro should be either removed (when its last user is gone) or adjusted to become a debugs() wrapper. However, I think this kind of improvements should be driven by specific use cases, and I am not working on any such case at the moment. All the helper-related pending PRs will make those or similar future changes easier, but that is the extent of my _current_ involvement with debug().

src/debug/Stream.h
yadij: IMO we should add important TODOs to the code rather than commit message. I am not aware of anyone or tools that scan the git history for TODO comments, whereas we do often grep the code for them.

rousskov: I moved TODO from the PR description to the header file to avoid further arguments.

yadij (approved):

PR #1012, opened 2022-04-01
Kid restart leads to persistent queue overflows, delays/timeouts

eduard-bagdasaryan:
WARNING: communication with ... may be too slow or disrupted...
    WARNING: abandoning ... I/Os
    ERROR: worker I/O push queue for ... overflow...
    ERROR: Collapsed forwarding queue overflow...

SMP queues rely on the shared memory QueueReader::popSignal flag to
reduce the number of UDS messages that queue readers and writers need to
send each other. If the flag is true but there is no corresponding "wake
up, you have new queued items to read" UDS message for the reader, the
reader may stall.  This happens when the reader restarts (e.g., after
hitting an assertion) while the flag is true. A stalled queue reader
leads to delays and queue overflows:

* When the problem affects worker-disker queues, disk I/O delays under
  the hard-coded 7-second timeout are not reported to the admin but may
  affect user experience. Larger delays trigger level-1 WARNINGs. Push
  queue overflows trigger level-1 ERRORs.

* Transient worker-worker queue problems may stall concurrent
  transactions that are reading from the cache entry being written by
  another process. Overflows trigger level-1 ERRORs.

The restarted worker usually starts working just fine because it does
not expect any messages. A busy restarted worker may also appear to
continue working fine because workers always pop queued items before
pushing new ones -- as long as the worker queues new items, it will see
and pop responses to earlier requests, masking the problem. However, the
"stuck popSignal" problem never goes away: Squid only clears the flag
when receiving a notification, but sending new notifications is blocked
by that stuck flag.

Upon kid start, we now clear popSignal (to reflect the actual
communication state) and empty the queue (to reduce overflows). Since
commit 4c21861, any corresponding in-flight UDS queue notification is
ignored because it was sent to the previous process playing the same kid
role. The queue writer will see the false popSignal flag and send a new
notification when queuing a new item, preventing queue stalls.

Also properly ignore stale disker responses about cache_dirs we have not
opened yet, especially since we are now trying to empty the queues ASAP
after a restart, before Coordinator has a chance to inform us about
available diskers, populating the IpcIoFiles container. We already have
similar protection from stale UDS messages and from stale disker queue
messages about _opened_ cache_dirs ("LATE disker response to...").

----

Cherry-picked master/v6 commit 5faec1a without BaseMultiQueue::stat()
changes related to mgr:store_queues statistics reporting. That cache
manager report was not backported to v5.

### DIFF
--- a/src/CollapsedForwarding.cc
+++ b/src/CollapsedForwarding.cc
@@ -9,6 +9,7 @@
 /* DEBUG: section 17    Request Forwarding */
 
 #include "squid.h"
+#include "base/AsyncFunCalls.h"
 #include "CollapsedForwarding.h"
 #include "globals.h"
 #include "ipc/mem/Segment.h"
@@ -48,8 +49,12 @@ void
 CollapsedForwarding::Init()
 {
     Must(!queue.get());
-    if (UsingSmp() && IamWorkerProcess())
+    if (UsingSmp() && IamWorkerProcess()) {
         queue.reset(new Queue(ShmLabel, KidIdentifier));
+        AsyncCall::Pointer callback = asyncCall(17, 4, "CollapsedForwarding::HandleNewDataAtStart",
+                                                NullaryFunDialer(&CollapsedForwarding::HandleNewDataAtStart));
+        ScheduleCallHere(callback);
+    }
 }
 
 void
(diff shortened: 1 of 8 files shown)


rousskov:
@eduard-bagdasaryan, I have adjusted the PR description footer to explain that this backport was not a straight cherry-pick. Please adjust that footer further if needed.

rousskov:
> I have adjusted the PR description footer to explain that this backport was not a straight cherry-pick. Please adjust that footer further if needed.

... and re-adjusted again after reading your non-GitHub comment about these changes. Please check.

eduard-bagdasaryan:
> re-adjusted again 

Looks correct.

rousskov:
@yadij, please note that GitHub pull requests are supposed to be merged using PR title and PR description as the commit message. That is why we are polishing that PR metadata to get it right! Anubis uses that metadata when forming the official commit. GitHub will not do it for us on its own. If you merge v5 PRs without using Anubis, please do whatever is necessary to achieve the same correct result.  Otherwise, we get wrong official commit messages (e.g., f6ee682 and 1332f8d).

PR #1014, opened 2022-04-02
Prep for 5.5

yadij:


doc/release-notes/release-5.sgml
```diff
@@ -261,6 +267,9 @@ This section gives a thorough account of those changes in three categories:
 	   listening queue for each worker in SMP.
 	<p>New <em>CONDITIONAL_AUTH</em> flag for <em>sslflags=</em> option to
 	   request client certificate(s) but not reject clients without any.
+	<p>Squid-5.5 introduces the OpenSSL <em>SSL_MODE_NO_AUTO_CHAIN</em> when
+	   delivering server CA certificates. The <em>tls-cert=</em> file must
+	   now contain both the server certificate and CA chain to send.
```
rousskov, line 272: IMHO, the above description is unclear and misleading:

* Nobody except perhaps a handful of people with very intimate Squid code knowledge would know how OpenSSL's `SSL_MODE_NO_AUTO_CHAIN` constant is related to the bug [5177](https://bugs.squid-cache.org/show_bug.cgi?id=5177) fix and the associated changes in Squid behavior. The connection there is very subtle!
* The `tls-cert` file does not _have to_ contain the "CA chain" for things to work correctly (in some cases) -- the "must now contain" phrasing is inaccurate.
* The "CA chain" phrase may be misinterpreted as containing the root CA certificate.

I do not insist on any specific wording here, but can suggest the following replacement:

```suggestion
	<p>Squid-5.5 no longer sends intermediate CA certificates that form a chain
	   with the server certificate found in the <em>tls-cert=</em> file unless
	   those intermediate CA certificates are actually present in that
	   <em>tls-cert=</em> file. Before this change, Squid could also send
	   matching intermediate CA certificates mentioned in other configuration
	   options (see Squid Bug #5177 for details). If you want your Squid to
	   send intermediate CA certificates to TLS clients, then make sure the
	   corresponding <em>tls-cert=</em> file contains those certificates.
```

Please adjust as needed and note that the above suggestion may not have correct leading whitespace (`TAB SP SP SP`) due to browser and GitHub conversions.

yadij: Referencing the bug fixed is a detail for ChangeLog file. And verbose expansion of the change is text for the release announcement email.

This text is just a brief statement of what has changed in the Squid UI. Which is two things:

1) Squid now uses that OpenSSL option. It is up to admin interested to investigate what it that means for their specific OpenSSL version + Squid build. That does vary by libssl version, build options, and config - making your suggested statements untrue for many installations.

2) As of this release ensuring the "CA chain to send", exists in the tls-cert= file is the only way to guarantee that it gets delivered.

rousskov: > Referencing the bug fixed is a detail for ChangeLog file

Why not both?! It is needlessly difficult to connect the current PR version of the release note text to any specific ChangeLog entry: The ChangeLog entry in question says "Bug 5177: clientca certificates sent to https_port clients". The PR version of the release note talks about server certificates, not clientca certificates and does not mention any bugs at all. Instead, it (incorrectly) implies that this upgrade introduces some new Squid configuration option...

> And verbose expansion of the change is text for the release announcement email.

Why?! Why not place (directly or via references) all the information the admin needs in one place instead of expect folks to be hunting down release announcement emails and searching for relevant information there, especially when there is no clear need for them to do so -- based on the PR version of the release notes, many admins would reasonably expect a seamless upgrade from 5.4 to 5.5.

> This text is just a brief statement of what has changed in the Squid UI. 

The bug fix in question has _not_ changed Squid UI. Squid UI (including squid.conf) remains unchanged.

> Squid now uses that OpenSSL option. It is up to admin interested to investigate what it that means for their specific OpenSSL version + Squid build

Even if an admin is lucky enough to guess that they should be interested, virtually no admin would be able to understand the effects of that option on their Squid based on OpenSSL documentation. Even seasoned developers cannot do that without intimate knowledge of Squid code. The issue is just too complex/low-level, and the removed side effects are far from intuitive.

In addition to that complexity, we have to keep in mind that some Squids that were working fine will stop working after this upgrade. This change does not enable some optional functionality that admins can safely ignore; it actually removes some functionality that (in some arguably misconfigured deployments) was required for the essential certificates to be sent. It is our duty to warn admins about this potentially critical upgrade surprise, and the PR version of the text does not do that properly IMO.

> As of this release ensuring the "CA chain to send", exists in the tls-cert= file is the only way to guarantee that it gets delivered.

This statement can be interpreted as correct (I would s/delivered/sent/). Unfortunately, the PR version of the text can be easily misinterpreted to mean something else (as I have detailed in my change request).

yadij: > > Referencing the bug fixed is a detail for ChangeLog file
> 
> Why not both?!

Different audiences with different information needs.

The release notes file is typically used by "sysops" admin performing a major version upgrade. They need to know specifically what parts of the UI have changed so as to ensure the systems they have integrated Squid into will continue to work, or plan alterations. Adding mention of changes on a point release like this is an exceptional situation, which is why v5.5 is specifically being called out.

ChangeLog is typically used by "devops" admin whose environments include custom builds of Squid and need references into things like bug IDs, assert/exception text, and Squid internal feature nomenclature that let them plan upgrades involving code changes. One may think the github commits page is sufficient, but we have distributors who need to re-publish this list with entries about their own customization. Every point release gets additions.

> 
> > And verbose expansion of the change is text for the release announcement email.
> 
> Why?! Why not place (directly or via references) all the information the admin needs in one place

Because "admins" is not a uniform group. It is a broad categorization of many different roles.  Each of these documents has different audiences within that broad "admin" collective. Writing all variants of the information together in one place becomes "word salad" to those with time constraints and often critical details get overlooked by those at most need of it.

The announcement email is for admin wanting push-notifications about important things being changed. Highlighting what behaviour they might have noticed happening and why it is a problem they should go through the bother of an upgrade to fix. This is part of the Squid Project marketing more than technical records.


> > This text is just a brief statement of what has changed in the Squid UI.
> 
> The bug fix in question has _not_ changed Squid UI. Squid UI (including squid.conf) remains unchanged.
> 

The bug fix makes changes to the behaviour of several popular UI settings without providing any startup or run-time warnings, nor log entries to bring the change to attention. If those setups were not a misconfiguration to begin with I would not have allowed backport.


> > Squid now uses that OpenSSL option. It is up to admin interested to investigate what it that means for their specific OpenSSL version + Squid build
> 
> Even if an admin is lucky enough to guess that they should be interested,

There is no guessing when the option is named. They can look it up. Hiding that name is what creates problems.

> 
> > As of this release ensuring the "CA chain to send", exists in the tls-cert= file is the only way to guarantee that it gets delivered.
> 
> This statement can be interpreted as correct (I would s/delivered/sent/). Unfortunately, the PR version of the text can be easily misinterpreted to mean something else (as I have detailed in my change request).

Ack. I have re-written it in 40d183a.

rousskov: > > > Amos: Referencing the bug fixed is a detail for ChangeLog file

> > Alex: Why not both?!

> Amos: Different audiences with different information needs.

Your answer, including (snipped) detailed descriptions of the "sysops" and "devops" audiences, does not explain why we should avoid mentioning a bug ID in the release notes. Clearly, there are lots of different documentation use cases that go well beyond the described "sysops" audience, and even within that audience, some folks will find a bug ID reference very useful in this particular case.

> > > Amos: And verbose expansion of the change is text for the release announcement email.

> > Alex: Why?! Why not place (directly or via references) all the information the admin needs in one place

> Because "admins" is not a uniform group. It is a broad categorization of many different roles. Each of these documents has different audiences within that broad "admin" collective. Writing all variants of the information together in one place becomes "word salad" to those with time constraints and often critical details get overlooked by those at most need of it.

I disagree that placing (directly or via references) all the information the admin needs in one place must result in a "word salad", especially since nobody is proposing to remove succinct ChangeLog lines and other summaries/indexes for those who do not need the details.


> > > Amos: This text is just a brief statement of what has changed in the Squid UI.

> > Alex: The bug fix in question has _not_ changed Squid UI. Squid UI (including squid.conf) remains unchanged.

> The bug fix makes changes to the behaviour of several popular UI settings

If "change in behavior of a setting" is a "UI change", then most Squid changes are "UI changes". I do not think this is a reasonable definition of what constitutes "changes in UI". On the other hand, I do not think this note should be limited to UI changes so perhaps this disagreement is not important here.

> > > Squid now uses that OpenSSL option. It is up to admin interested to investigate what it that means for their specific OpenSSL version + Squid build

> > Even if an admin is lucky enough to guess that they should be interested,

> There is no guessing when the option is named. They can look it up. Hiding that name is what creates problems.

Knowing the name of an option is very different from knowing that they should be interested in it. A lot of guessing is required to connect the original text with concepts a typical admin can operate with.
 
> I have re-written it in [40d183a](https://github.com/squid-cache/squid/commit/40d183a78eb88a61f8a78b84a8a8522bc728ddaf).

> `when OpenSSL library supports <em>SSL_MODE_NO_AUTO_CHAIN</em> mode.`

There is no "when" here AFAICT: All OpenSSL versions that Squid can be compiled with support SSL_MODE_NO_AUTO_CHAIN. Squid sources assume that all OpenSSL versions have this constant. The constant was added to OpenSSL before v0.9.8.

> `Squid-5.5 will no longer use <em>tls-clientca=</em> certificates`

I am also not sure whether `tls-clientca` was the only possible "surprising" source of intermediate CA certificates. @chtsanti probably knows/remembers more about this than I do.

As for the rest of the updated note text, it looks better, thank you, but I still think that many (perhaps most) admins will not be able to understand what it really means.

FWIW, I still do not insist on any changes, including the adoption of the version I proposed [earlier](https://github.com/squid-cache/squid/pull/1014#discussion_r841145545).

PR #1016, opened 2022-04-05
Fix typo in pkg.m4

eltociear:
occurence -> occurrence


yadij:
This file is imported from pkg-config. Changes to it, especially typos like this should be submitted to https://www.freedesktop.org/wiki/Software/pkg-config/

PR #1017, opened 2022-04-11
Work around broken CPU_AND API on Ubuntu Xenial and clang

kinkie:
At very high levels of compiler strictness checks, the CPU_AND system interface exposes a broken API.
Hack around the breakage, by explicitly instructing the compiler to disregard the non-standard return value

Fully tested in https://build.squid-cache.org/job/anybranch-amd64-matrix/48/

### DIFF
--- a/src/CpuAffinitySet.cc
+++ b/src/CpuAffinitySet.cc
@@ -38,7 +38,10 @@ CpuAffinitySet::apply()
     } else {
         cpu_set_t cpuSet;
         memcpy(&cpuSet, &theCpuSet, sizeof(cpuSet));
-        CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
+        // hack for clang on ubuntu-xenial, where CPU_AND returns a value
+        // which under very strict compiler checks fails the build
+        auto s = CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
+        (void) s;
         if (CPU_COUNT(&cpuSet) <= 0) {
             debugs(54, DBG_IMPORTANT, "ERROR: invalid CPU affinity for process "
                    "PID " << getpid() << ", may be caused by an invalid core in "


src/CpuAffinitySet.cc
```diff
@@ -38,7 +38,10 @@ CpuAffinitySet::apply()
     } else {
         cpu_set_t cpuSet;
         memcpy(&cpuSet, &theCpuSet, sizeof(cpuSet));
-        CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
+        // hack for clang on ubuntu-xenial, where CPU_AND returns a value
+        // which under very strict compiler checks fails the build
+        auto s = CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
+        (void) s;
```
eduard-bagdasaryan, line 44: Can we do this in a single line?

```suggestion
        (void) CPU_AND(&cpuSet, &cpuSet, &theOrigCpuSet);
```

kinkie: No. That was the state before commit b2188e4d6bce80f2f93e3f99e11f3d054f5e8b60 and it wasn't working. That change broke xenial/clang instead

rousskov (changes requested):
If this approach works, I think you found a nice solution! However, I do not see how this approach can work on FreeBSD where CPU_AND() is a do-while loop, and my suspicions are supported by failing (primitive and possibly inapplicable) [tests](https://cpp.godbolt.org/z/5nMT89P6a) emulating that CPU_AND() implementation flavor.

```
<source>:16:10: error: 'void s0' has incomplete type
<source>:1:20: error: expected primary-expression before 'do'
<source>:18:10: error: variable or field 's2' declared void
```

If my suspicions are unfounded, and tests are wrong, I left a polishing suggestion. Otherwise, I think we will need a ./configure-time test to properly detect CPU_AND() implementations that effectively prohibit a `(void)` prefix.

----

In any event, when the dust settles, please adjust PR description to details the errors that triggered this PR. See b2188e4d6bce80f2f93e3f99e11f3d054f5e8b60 for an example. Compared to the current PR metadata, the series of such notes has a better chance of allowing the next developer to avoid repeating previous mistakes.

Please also move a temporary Jenkins link from the PR description (the future commit message) to the regular PR comment.

kinkie:
Abandoning in favour of PR #1050

PR #1021, opened 2022-04-15
sslcrtd_program: Use (more) Squid-wide APIs

rousskov:
This switch allows us to use Squid-native debugging and error handling
code in ssl/libsslutil.la. Such use is necessary to fix and enhance
certificate handling code currently located in that library. For now, we
only did straightforward adjustments like fixing parseBytesOptionValue()
and Ssl::CrtdMessage::parseRequest() error detection/handling. More
serious changes deserve dedicated PRs.

This change also converts the remaining bare std::runtime_error uses in
runtime Squid code and almost all std::cerr uses in sslcrtd_program:

* std::runtime_error: TextException provides the source code location of
  the thrower and will be enhanced along with Squid improvements.

* std::cerr: debugs() provides much better runtime logging, on many
  levels. The usage() dump still uses std::cerr because debugs()
  decorations are not useful in that special case and because, IMO, that
  case should be using std::cout instead (it is not reporting an error).

Removing STL APIs (that have Squid-native alternatives) from the old
helper code also reduces the temptation to use wrong APIs in new code,
especially when authors are not familiar with Squid conventions/plans.

Also replaced sslcrtd_program's `Here` hack with SourceLocation.

Also report all started helpers (at debug level 2) because a successful
helper start is a significant event worth reporting (when level-2
debugging is enabled) across all helpers.

### DIFF
--- a/src/debug/debug.cc
+++ b/src/debug/debug.cc
@@ -397,6 +397,8 @@ Debug::NameThisHelper(const char * const name)
 
     SettleStderr();
     SettleSyslog();
+
+    debugs(84, 2, "starting " << name << " with PID " << getpid());
 }
 
 void
--- a/src/sbuf/Makefile.am
+++ b/src/sbuf/Makefile.am
@@ -23,7 +23,3 @@ libsbuf_la_SOURCES = \
 	Stats.h \
 	Stream.h \
 	forward.h
-
-libsbuf_la_LIBADD = \
-	$(top_builddir)/src/base/libbase.la \
-	$(top_builddir)/src/debug/libdebug.la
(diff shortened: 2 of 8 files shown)


rousskov:
This is the first PR that uses recently added convenience libraries to improve helpers. This PR focuses on straightforward sslcrtd_program improvements, clearing the way for more difficult/sensitive fixes that deal with TLS error handling.

A lot more helper code should be converted to use Squid-wide APIs now, of course. For example, helper code that was previously forced to use std::string may be a good target.

kinkie (approved):
looks okay to me

PR #1023, opened 2022-04-19
Refactor Half-Closed FD monitoring logic

yadij:
The complex event handler to determine when half-closed FDs have
lost their read handler can be replaced with a fd_table flag set
when the mode is enabled and unset on disable.

This flag can be tested by the read handler to re-schedule
monitoring correctly even when the mode is disabled while that
Call is pending in the Async dialer queue.

It also provides a far simpler method for other comm related
logic to test for write-only (half-closed) status without the
current complex registry lookup.

The DescriptorSet registry already duplicated the fd_table
halfClosedReader member. Now with the write_only flag it is
fully redundant and can be removed.

### DIFF
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -957,8 +957,8 @@ ConnStateData::kick()
      * half-closed _AND_ then, sometimes, spending "Timeout" time in
      * the keepalive "Waiting for next request" state.
      */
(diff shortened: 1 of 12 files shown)


rousskov (changes requested):
Please ping me when this is ready for review or you would like to discuss any changes.

BTW, are there signs that we should keep support for half-closed sockets?

PR #1024, opened 2022-04-20
libcomm pt2: Convert Comm::iocb_table to a std::map

yadij:
Implementing an old TODO and removing the need to
initialize and destruct the map on startup/shutdown.

### DIFF
--- a/src/comm.cc
+++ b/src/comm.cc
@@ -1141,9 +1141,6 @@ comm_init(void)
     /* make sure the accept() socket FIFO delay queue exists */
     Comm::AcceptLimiter::Instance();
 
-    // make sure the IO pending callback table exists
-    Comm::CallbackTableInit();
-
     /* XXX account fd_table */
     /* Keep a few file descriptors free so that we don't run out of FD's
      * after accepting a client but before it opens a socket or a file.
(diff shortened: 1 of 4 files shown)


rousskov (changes requested):
There are problems with the proposed refactoring. I have added specific change requests to start addressing them, but there is a high-level alternative to consider _first_: Should we remove Comm::iocb_table completely (by merging its (used) data members into `fde` (i.e. fd_table items)? Both tables should have the same size and lifetime AFAICT. What justifies iocb_table existence in the refactored code?

rousskov:
> > What justifies iocb_table existence in the refactored code?
 
> Only that I have run out of time for more refactoring of this table. It is just enough to redux the startup/shutdown logic.

Since the current PR direction requires non-trivial work and code complications to achieve usable results and the alternative approach is likely to result in simpler code, we should try to remove this table instead. I have not checked all the details, but I suspect that the amount of changes would be about the same as in the current PR _if_ one removes this table _without_ changing how the remaining fd_table is initialized/destroyed. That minimal change would still accomplish the stated goals of this PR -- startup/shutdown logic improvements related to the iocb_table.

PR #1025, opened 2022-04-21
Preserve caller context across (and improve) deferred reads

eduard-bagdasaryan:
The transaction context was not saved/restored when dealing with
deferred reads initiated by events like the DelayPools::Update() event.
To fix this, we refactored MemObject::delayRead() and its descendants to
use an AsyncCall, which automatically stores/restores code context.

Using explicit async callbacks highlighted the danger of passing
Connection object via CommRead that does not maintain a closure
callback. There was also a related "stuck transaction" suspicion
documented in DeferredReadManager::kickARead(). Fortunately, all these
problems could now be solved by removing DeferredRead and CommRead
classes! The delayed readers already store the Connection object,
maintain closure callbacks, and have to check stored Connection validity
before reading anyway. The general/centralized delayed reading logic is
not really about reading and Connections (those parts are handled by
transaction-specific code) but about triggering reading attempts.
Asynchronous calls are perfect (and sufficient) for doing that.

Also fixed Delay Pools for Gopher: delayAwareRead() was initiated only
once from gopherSendComplete() and the subsequent read calls were
delay-unaware (i.e. immediate) reads.

Also fixed a Delay Pools problem with active transactions: A transaction
started with Delay Pools on becomes stuck if a reconfiguration turns
Delay Pools off.

Also refactored the existing AsyncCall FIFO intrusive storage, making
its reuse possible (and marked one candidate with a TODO).


src/clients/FtpClient.cc
```diff
@@ -930,9 +952,16 @@ Ftp::Client::maybeReadVirginBody()
...
+    const auto amountToRead = entry->bytesWanted(Range<size_t>(0, read_sz));
+
+    if (amountToRead <= 0) {
+        delayRead();
+        return;
+    }
+
+    using ReadDialer = CommCbMemFunT<Client, CommIoCbParams>;
+    AsyncCall::Pointer readCallback = JobCallback(9, 5, ReadDialer, this, Client::dataRead);
+    comm_read(data.conn, data.readBuf->space(), amountToRead, readCallback);
```
yadij, line 964: Please use the AsyncCall-aware Comm::Read / Comm::ReadNow API instead of the deprecated comm_read().

Usage of the new API is essentially what you are doing here ... except that this block of code should **only** pass the ```readCallback``` object to ```Comm::Read()```. The buffer calculations go in ```Client::dataRead()``` logic to setup for a ```Comm::ReadNow()``` fetching the data (no longer passed to the handler as a callback parameter).

Same change request for all the other AsyncJob methods equivalent to this delayAwareRead(). The gopher change and other old non-AsyncJob callbacks are fine doing this old way.

yadij (changes requested):
Thanks for this much needed cleanup. Just a few things.

src/clients/FtpClient.cc
eduard-bagdasaryan: I  started refactoring at first, addressing your request, but then realized that we cannot simply substitute the old comm_read(), requiring MemBuf with the new Comm::Read, requiring SBuf as an input buffer. Probably we could go further, converting DataChannel::readBuf to SBuf, but I am afraid this is a little out of the scope. There are other(minor) difficulties with this refactoring as well. For example, should we remove the deprecated replyBodySpace()? What to do with the strange 'read_sz<2' condition? Overall, I would do this refactoring in a separate PR, adding a TODO here instead.

> Same change request for all the other AsyncJob methods

I see only two comm_read() occurrences in this PR: this one and the Gopher's (which is out of the question).

eduard-bagdasaryan (commented):
@yadij , I partially addressed your concerns.

src/clients/FtpClient.cc
rousskov: FWIW, I agree with Eduard that the requested change can and should be considered out of this PR scope. Properly upgrading FtpClient to use Comm::ReadNow() would require converting _both_ control and data channels to use SBuf storage and modern Comm reading APIs, triggering quite a bit of side effects that deserve careful analysis and review. All that work is useful but is best done in a dedicated PR, especially since this PR does not increase deprecated comm_read() usage but essentially "inlines" and "localizes" the old one (i.e. we can say that the PR is not making things worse in this regard).

yadij: Okay, request withdrawn.

yadij (approved):

rousskov (changes requested):
I love the simplifications provided by this PR!

I did not check low-level details, but there is one high-level problem that I think we should fix before merging this.

src/DeferredReadManager.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef SQUID_DEFERREDREADMANAGER_H
+#define SQUID_DEFERREDREADMANAGER_H
+
+#include "base/forward.h"
+
+#include <vector>
+
+// TODO: create dedicated header/source files
+/// maintains a list of async calls and schedules them at once
+class DeferredReadManager
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done (up to 008dfc9f).

rousskov:
I pushed a few polishing changes and a couple of changes that simplified the result quite a bit (and probably addressed a subtle problem with the earlier PR code). @eduard-bagdasaryan, please check that the current branch meets your expectations and still passes your tests.

src/tests/stub_comm.cc
```diff
@@ -21,14 +21,10 @@
...
 
-/* should be in stub_CommRead */
-#include "CommRead.h"
-CommRead::CommRead(const Comm::ConnectionPointer &, char *, int, AsyncCall::Pointer &) STUB
-CommRead::CommRead() STUB
-DeferredReadManager::~DeferredReadManager() STUB
-DeferredRead::DeferredRead(DeferrableRead *, void *, CommRead const &) STUB
-void DeferredReadManager::delayRead(DeferredRead const &) STUB
-void DeferredReadManager::kickReads(int const) STUB
+/* should be in stub_DelayedAsyncCalls */
```
yadij, line 24: No, files in libraries should be in the library stub, not their own.
```suggestion
/* should be in stub_libbase */
```

yadij (changes requested):

src/base/AsyncCallQueue.cc
```diff
@@ -15,52 +15,27 @@
 
 AsyncCallQueue *AsyncCallQueue::TheInstance = 0;
 
-AsyncCallQueue::AsyncCallQueue(): theHead(NULL), theTail(NULL)
+AsyncCallQueue::AsyncCallQueue()
```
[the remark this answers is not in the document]
yadij: These lines disappeared while I was reviewing. Now there is a undefined default-constructor in the header. That should either; go away (restoring the default construction ability), or now be declared "= deleted;" in the normal top-of-class public area.

src/CompositePoolNode.h
```diff
@@ -13,7 +13,8 @@
 
 #if USE_DELAY_POOLS
 #include "auth/UserRequest.h"
-#include "CommRead.h"
+#include "base/DelayedAsyncCalls.h"
+#include "base/forward.h"
```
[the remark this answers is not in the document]
rousskov: In general, the list of headers included by file A is essentially A's implementation detail. It should not be known and relied upon by human developers writing `#include "A"` in their code. Until generation of `#include` lines is automated, we should accept a logical or reasonable set of `#include` lines rather than attempt to maintain a _minimal_ set. When a header file needs a name from some `forward.h` file, it should include that `forward.h` file without fear and without spending time on investigating whether some of the other included files currently include that `forward.h`.

FWIW, I am OK with removing some technically excessive `#include` lines in this PR to save time, but I hope we do not need to revisit this aspect in the future.

src/base/AsyncCallQueue.cc
```diff
@@ -15,52 +15,25 @@
...
-    assert(!call->theNext);
-    if (theHead != NULL) { // append
-        assert(!theTail->theNext);
-        theTail->theNext = call;
-        theTail = call;
-    } else { // create queue from cratch
-        theHead = theTail = call;
-    }
-}
+AsyncCallQueue::AsyncCallQueue() = default;
```
eduard-bagdasaryan, line 18: I would move this to the header.

src/base/AsyncCallList.h
```diff
@@ -0,0 +1,39 @@
...
+
+#ifndef SQUID_BASE_ASYNCCALLLIST_H
+#define SQUID_BASE_ASYNCCALLLIST_H
+
+#include "base/forward.h"
+#include "base/RefCount.h"
+
+/// An efficient (but intrusive) AsyncCall storage preserving FIFO order.
+/// A given AsyncCall object may reside in at most one such storage.
+class AsyncCallList
```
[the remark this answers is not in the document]
rousskov: > The documentation you are reading now applies to the (duplicated) code you have approved during the previous iteration

To clarify, the documentation applies to old/official AsyncCallQueue code from where this implementation was moved into a dedicated class (AsyncCallList) and to old/official Comm uses of the same invasive list for close handlers. The delay readers code simplified in this PR used an std::vector before ff0ac21 and then was converted to use the same AsyncCallList class to avoid reimplementing that code functionality and to gain AsyncCallList efficiencies.

src/base/AsyncCallQueue.cc
rousskov: Yes, please. Such a move should avoid wasting more time on this issue. The point of a `= default` declaration in a .cc file is to hide the (generated) implementation from the header users. In some cases, such hiding solves dependency problems and/or reduces code size.

In this particular case, we are hiding how the `scheduled` data member is initialized, but the gain is probably very small (if any) -- the current code including AsyncCallQueue.h may not change when we stop hiding because that code "knows too much" already. Moreover, for proper hiding, we may need to hide the generated destructor implementation as well, and we do not do that.

src/CompositePoolNode.h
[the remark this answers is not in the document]
eduard-bagdasaryan: Removed.

src/MemObject.h
```diff
@@ -9,7 +9,8 @@
 #ifndef SQUID_MEMOBJECT_H
 #define SQUID_MEMOBJECT_H
 
-#include "CommRead.h"
+#include "base/DelayedAsyncCalls.h"
+#include "base/forward.h"
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Ok.

src/base/AsyncCallList.h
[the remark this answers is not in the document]
eduard-bagdasaryan: Ok. let's keep the old intrusive-list implementation and revisit this area in a future improvement.

src/base/DelayedAsyncCalls.cc
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (C) 1996-2022 The Squid Software Foundation and contributors
+ *
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#include "squid.h"
+
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

src/base/DelayedAsyncCalls.cc
```diff
@@ -0,0 +1,29 @@
...
+DelayedAsyncCalls::delay(const AsyncCall::Pointer &call)
+{
+    debugs(5, 3, call << " after " << deferredReads.size());
+    deferredReads.add(call);
+}
+
+void
+DelayedAsyncCalls::kick()
+{
+    // XXX: For fairness this SHOULD randomize the order
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Removed.

src/base/DelayedAsyncCalls.h
```diff
@@ -0,0 +1,32 @@
...
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID_BASE_DELAYEDASYNCCALLS_H
+#define SQUID_BASE_DELAYEDASYNCCALLS_H
+
+#include "base/AsyncCallList.h"
+#include "base/forward.h"
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

src/tests/stub_comm.cc
eduard-bagdasaryan: Ok.

eduard-bagdasaryan (commented):
Done (72a897e).

src/base/AsyncCallQueue.cc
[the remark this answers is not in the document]
eduard-bagdasaryan: Removed.

src/base/AsyncCallQueue.cc
eduard-bagdasaryan: Removed.

src/base/AsyncCallQueue.h
```diff
@@ -22,18 +21,13 @@ class AsyncCallQueue
...
 
     // make this async call when we get a chance
-    void schedule(AsyncCall::Pointer &call);
+    void schedule(const AsyncCallPointer &call) { scheduled.add(call); }
 
     // fire all scheduled calls; returns true if at least one was fired
     bool fire();
 
 private:
-    AsyncCallQueue();
```
rousskov, line 31: This private declaration should stay (in some shape or form) because it is meaningful -- we are telling the reader that they should not be creating AsyncCallQueue objects (because AsyncCallQueue is designed as a singleton). We can make that idea more explicit if you think it was not explicit enough:

```C++
    // private placement prohibits creation of competing queues by the public
    AsyncCallQueue() = default;
```

eduard-bagdasaryan: > AsyncCallQueue is designed as a singleton

No, it was not a singleton because we could still create a copy (via implicit copy constructor). To enforce this pattern I would suggest, in addition, to delete both copy constructor and operator=.

rousskov: >> AsyncCallQueue is designed as a singleton

> No

AsyncCallQueue is designed as a singleton. Whether that design was implemented correctly is a different question ;-).

> we could still create a copy (via implicit copy constructor). To enforce this pattern I would suggest, in addition, to delete both copy constructor and operator=.

Yes, and I would try to go a step further, actually -- since a single AsyncCall object should not be present in two AsyncCallLists, let's try to prohibit duplicating of AsyncCallList objects (moving is OK). If we can do that, then it will automatically prevent duplication of AsyncCallQueues as well (and no code can accidentally _move_ a singleton anyway). I am not sure we can do that because some copying (of empty lists) might be going somewhere already, and we probably do not want to improve that code in this PR.

Long-term, we might even replace the whole AsyncCallQueue class with a pair of global Schedule()/FireScheduled() functions, but assessing that change is out of scope here, of course, and prohibiting AsyncCallList copying is still valuable even if AsyncCallQueue is gone.

While you are at it please also delete an extra empty line at the very end of AsyncCallList class declaration, before the closing curly brace.

src/base/AsyncCallQueue.h
```diff
@@ -27,6 +27,10 @@ class AsyncCallQueue
     bool fire();
 
 private:
+    AsyncCallQueue() = default;
+    AsyncCallQueue(const AsyncCallQueue &) = delete;
+    AsyncCallQueue &operator=(const AsyncCallQueue &) = delete;
```
rousskov, line 32: We normally `delete` prohibited methods publicly, but I am OK with making an exception in this special case because we have to keep the default constructor private (for other reasons).

src/base/AsyncCallQueue.h
eduard-bagdasaryan: > I was talking about AsyncCallList objects.

Done (at the second attempt) - I misinterpretted your initial suggestion.

src/base/AsyncCallQueue.h
eduard-bagdasaryan: >  if we can prohibit copying of AsyncCallLists 

We can - so I removed these two lines.

yadij (approved):

PR #1026, opened 2022-04-24
Simplify gopherToHTML()

yadij:
No logic change.

### DIFF
--- a/src/gopher.cc
+++ b/src/gopher.cc
@@ -423,7 +423,6 @@ gopherToHTML(GopherStateData * gopherState, char *inbuf, int len)
     char *host = NULL;
     char *port = NULL;
     char *escaped_selector = NULL;
-    const char *icon_url = NULL;
     char gtype;
     StoreEntry *entry = NULL;
 
@@ -568,7 +567,7 @@ gopherToHTML(GopherStateData * gopherState, char *inbuf, int len)
                     /* escape a selector here */
                     escaped_selector = xstrdup(rfc1738_escape_part(selector));
 
-                    icon_url = gopherState->iconUrl(gtype);
+                    const auto icon_url = gopherState->iconUrl(gtype);
 
                     if ((gtype == GOPHER_TELNET) || (gtype == GOPHER_3270)) {
                         if (strlen(escaped_selector) != 0)
--- a/src/gopher.cc
+++ b/src/gopher.cc
@@ -102,7 +102,8 @@ public:
 
     ~GopherStateData();
 
-    /// URL for icon to display, given the item-type code
+    /// URL for icon to display (or nil), given the Gopher item-type code.
+    /// The returned c-string is invalidated by the next call to this function.
     const char *iconUrl(char);
 
 public:
--- a/src/gopher.cc
+++ b/src/gopher.cc
@@ -102,6 +102,9 @@ public:
 
     ~GopherStateData();
 
+    /// URL for icon to display, given the item-type code
+    const char *iconUrl(const char);
+
 public:
     StoreEntry *entry;
     enum {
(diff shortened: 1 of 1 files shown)


src/gopher.cc
```diff
@@ -102,6 +102,9 @@ class GopherStateData
 
     ~GopherStateData();
 
+    /// URL for icon to display, given the item-type code
+    const char *iconUrl(const char);
```
rousskov, line 106: This particular `const` is not a part of the API. It is a part of that API _implementation_. IIRC, some compilers will even complain about such extraneous `const`s.

```suggestion
    const char *iconUrl(char);
```

src/gopher.cc
```diff
@@ -102,6 +102,9 @@ class GopherStateData
 
     ~GopherStateData();
 
+    /// URL for icon to display, given the item-type code
```
rousskov, line 105: We should document the lifetime of the returned-by-pointer object. We should not document how the URL will be _used_ -- that is up to the caller:

```suggestion
    /// MIME type icon URL for the given menu item type (or nil). The returned
    /// c-string is invalidated by the next call to this or similar functions.
```

or, if you prefer a more specific/accurate but less convenient variant (that requires the reader to start chasing a long list of poorly documented functions that all share the same lifetime validity flaws):

```suggestion
    /// MIME type icon URL for the given menu item type (or nil). See
    /// mimeGetIconURL() for the lifetime of the returned value.
```

I do not insist on any specific text here, of course.

src/gopher.cc
```diff
@@ -515,70 +568,7 @@ gopherToHTML(GopherStateData * gopherState, char *inbuf, int len)
...
-
-                    case GOPHER_WWW:
-                        icon_url = mimeGetIconURL("internal-link");
-                        break;
-
-                    default:
-                        icon_url = mimeGetIconURL("internal-unknown");
-                        break;
-                    }
+                    icon_url = gopherState->iconUrl(gtype);
```
rousskov, line 571: If such a move became possible, please move the declaration here, to avoid stale or meaningless `icong_url` pointer living outside its intended context.

```suggestion
                    const auto icon_url = gopherState->iconUrl(gtype);
```

The above GitHub suggestion requires also removing the current icon_url declaration, of course.

yadij: Done in [41c8941](https://github.com/squid-cache/squid/pull/1026/commits/41c894181d7ed8a11ad519326c10db2c090c1afc)

src/gopher.cc
yadij: Added the lifetime mention in [89be89e](https://github.com/squid-cache/squid/pull/1026/commits/89be89e11f0dc0fbf6f93776fd2c2c17b40e720c).

I am not mentioning "mime" or "menu" as those concepts are irrelevant to this method API or caller(s). This is simply an icon URL (not necessarily mime related), and the input is a Gopher "item-type" code (not restricted to menus).

src/gopher.cc
```diff
@@ -102,6 +102,10 @@ class GopherStateData
 
     ~GopherStateData();
 
+    /// URL for icon to display (or nil), given the Gopher item-type code.
```
rousskov, line 105: To clarify the second part of my old [change request](https://github.com/squid-cache/squid/pull/1026#discussion_r857135338): We should not document how the URL will be used (e.g. "to display") -- that usage is up to the caller:

```suggestion
    /// Icon URL for the given Gopher item-type code (or nil).
```

This documentation suffers from the "foo is foo" syndrome a bit, but that is the best I can come up with because the method name is already self-documenting in this regard. The other details in the description compensate for this (small) flaw.

rousskov (commented):
Thank you for addressing most of my concerns.

yadij:
PR passed 10-day no-objections criteria.

PR #1027, opened 2022-04-24
Fix CID 1461131 Invalid type in argument to printf

kinkie:
in ext_edirecory_userip, persist_timeout is defined
as a time_t, wihich doesn't fly well with printf.
Cast it to int for printing; since it is set using atoi
it is guaranteed not to overflow anyway

### DIFF
--- a/src/acl/external/eDirectory_userip/ext_edirectory_userip_acl.cc
+++ b/src/acl/external/eDirectory_userip/ext_edirectory_userip_acl.cc
@@ -444,7 +444,7 @@ DisplayConf()
     if (edui_conf.mode & EDUI_MODE_PERSIST) {
         local_printfx("	Persistent mode: ON\n");
         if (edui_conf.persist_timeout > 0)
-            local_printfx("	Persistent mode idle timeout: %d\n", static_cast<int>(edui_conf.persist_timeout));
+            local_printfx("	Persistent mode idle timeout: %ld\n", static_cast<long int>(edui_conf.persist_timeout));
         else
             local_printfx("	Persistent mode idle timeout: OFF\n");
     } else
--- a/src/acl/external/eDirectory_userip/ext_edirectory_userip_acl.cc
+++ b/src/acl/external/eDirectory_userip/ext_edirectory_userip_acl.cc
@@ -444,7 +444,7 @@ DisplayConf()
     if (edui_conf.mode & EDUI_MODE_PERSIST) {
         local_printfx("	Persistent mode: ON\n");
         if (edui_conf.persist_timeout > 0)
-            local_printfx("	Persistent mode idle timeout: %d\n", edui_conf.persist_timeout);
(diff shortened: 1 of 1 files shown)


src/acl/external/eDirectory_userip/ext_edirectory_userip_acl.cc
```diff
@@ -444,7 +444,7 @@ DisplayConf()
     if (edui_conf.mode & EDUI_MODE_PERSIST) {
         local_printfx("	Persistent mode: ON\n");
         if (edui_conf.persist_timeout > 0)
-            local_printfx("	Persistent mode idle timeout: %d\n", edui_conf.persist_timeout);
+            local_printfx("	Persistent mode idle timeout: %d\n", static_cast<int>(edui_conf.persist_timeout));
```
rousskov, line 447: If commit 5ef3cbb is to be believed, we should use `long int` for consistency sake:

```suggestion
            local_printfx("	Persistent mode idle timeout: %ld\n", static_cast<long int>(edui_conf.persist_timeout));
```

That cast would still work for the atoi()-parsed persist_timeout, of course.

yadij: This is not a situation where we need to place trust in a vague old opinion of yours.

Fact is casting to int may lead to 32-bit issues when time_t and int have different type sizes. Up-casting to long int does not cause problems on 32-bit OS regardless of whether it defines time_t as 32-bit or 64-bit.

yadij (changes requested):
Please implement using the long int cast instead.

src/acl/external/eDirectory_userip/ext_edirectory_userip_acl.cc
kinkie: Accepting the suggestion, it makes sense. TBH I still maintain that in practice we don't have a problem with 32-bitness, because the only place this variable is set is via atoi() which returns int, but upcasting will not hurt so let's just do it

rousskov: > I still maintain that in practice we don't have a problem with 32-bitness, because ... this variable is set is via atoi()

Just to avoid misunderstanding: The original problem in the official code is not with the variable value, but with the size of that value: printf(%d) expects/extracts a sizeof(int) parameter, but the call stack will actually contain a sizeof(time_t) parameter. When those sizes are different, printf() mishandles the stack. You are right that your original fix was also correct in that sense.

Any casting here is not ideal, but casting to long is the lesser of all known evils, for value truncation sake (if the code is refactored away from atoi()) and for consistency sake (if commit commit https://github.com/squid-cache/squid/commit/5ef3cbbdf51f78b3f9dbe96b428a64ada0da8324 is to be believed).

Thank you for accepting my suggestion without throwing insults.

Please do not forget to update the PR description accordingly.

yadij (approved):

PR #1028, opened 2022-04-24
Convert GopherStateData to AsyncJob

yadij:
Base callback handlers on HttpStateData equivalents.

Fixes sharing of server connection. Access server
connection via 'fwd' member and remove local copy.

Fixes traffic accounting on error conditions.

Also,  Cleanup GopherStateData initialization:
 - convert to C++11 initialization
 - document overlap with ::Client API

### DIFF
--- a/src/gopher.cc
+++ b/src/gopher.cc
@@ -85,14 +85,6 @@ class GopherStateData
 public:
     GopherStateData(FwdState *aFwd) :
         entry(aFwd->entry),
-        conversion(NORMAL),
-        HTML_header_added(0),
-        HTML_pre(0),
-        type_id(GOPHER_FILE /* '0' */),
-        overflowed(false),
-        cso_recno(0),
-        len(0),
-        buf(NULL),
         fwd(aFwd)
     {
         *request = 0;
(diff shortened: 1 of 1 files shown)


src/gopher.cc
```diff
@@ -77,21 +78,14 @@
  * Gopher is somewhat complex and gross because it must convert from
  * the Gopher protocol to HTTP.
  */
-class GopherStateData
+class GopherStateData : public AsyncJob
```
rousskov, line 81: Proper conversion to AsyncJob is not an easy task. I did not check all the details, but I can already see that this PR still needs significant work to complete that conversion. AFAICT, if we are to upgrade Gopher client (instead of deleting that feature), then GopherStateData should become a child of Client. Otherwise, we will be spending a lot of effort on replacing a broken but relatively simple "special" code with more complex but still very "special" code (that is partially duplicating `Client` code base).

FWIW, I think we should just drop Gopher support until we have more resources to do it right.

rousskov (changes requested):

src/gopher.cc
yadij: This is the first step towards ::Client inheritence. As mentioned in the description all these callbacks are renamed following HttpStateData which is already a child of ::Client API. Further conversion to ::Client API requires handling AsyncCall member functions - which this PR refactoring prepares GopherStateData to receive.

rousskov: > This is the first step towards ::Client inheritence

It is up to you how many steps to make towards the Client inheritance goal, of course. In this case, the total work required to merge two PRs is a lot more than merging one cumulative PR (as discussed in my change request). If you have already done the first step, you may do the second step to address this change request. Doing so is likely to decrease your work as well (as far as reaching the Client inheritance goal is concerned).

Alternatively, let's drop Gopher support.

PR #1029, opened 2022-04-24
Initialize all HttpStateData data members

kinkie:
If fwd->serverConnection is null, _peer is read before being
initialized. We now make sure all HttpStateData data members are
initialized (using in-class member initialization).

Detected by Coverity. CID 1494356: Uninitialized pointer read (UNINIT).

Also removed HttpStateData::read_sz as unused.

### DIFF
--- a/src/http.cc
+++ b/src/http.cc
@@ -75,14 +75,8 @@ static void copyOneHeaderFromClientsideRequestToUpstreamRequest(const HttpHeader
         HttpHeader * hdr_out, const int we_do_ranges, const Http::StateFlags &);
 
 HttpStateData::HttpStateData(FwdState *theFwdState) :
-    _peer(nullptr),
     AsyncJob("HttpStateData"),
-    Client(theFwdState),
-    httpChunkDecoder(NULL),
-    lastChunk(0),
-    payloadSeen(0),
-    payloadTruncated(0),
-    sawDateGoBack(false)
+    Client(theFwdState)
 {
     debugs(11,5, "HttpStateData " << this << " created");
     ignoreCacheControl = false;
--- a/src/http.cc
+++ b/src/http.cc
@@ -79,11 +79,9 @@ HttpStateData::HttpStateData(FwdState *theFwdState) :
     Client(theFwdState)
 {
     debugs(11,5, "HttpStateData " << this << " created");
-    ignoreCacheControl = false;
-    surrogateNoStore = false;
     serverConnection = fwd->serverConnection();
 
-    if (fwd->serverConnection())
+    if (fwd->serverConnection() != NULL)
         _peer = cbdataReference(fwd->serverConnection()->getPeer());         /* might be NULL */
 
     flags.peering =  _peer;
--- a/src/http.cc
+++ b/src/http.cc
@@ -75,10 +75,11 @@ static void copyOneHeaderFromClientsideRequestToUpstreamRequest(const HttpHeader
         HttpHeader * hdr_out, const int we_do_ranges, const Http::StateFlags &);
 
 HttpStateData::HttpStateData(FwdState *theFwdState) :
+    _peer(nullptr),
     AsyncJob("HttpStateData"),
     Client(theFwdState),
-    lastChunk(0),
     httpChunkDecoder(NULL),
+    lastChunk(0),
     payloadSeen(0),
     payloadTruncated(0),
     sawDateGoBack(false)
@@ -88,7 +89,7 @@ HttpStateData::HttpStateData(FwdState *theFwdState) :
     surrogateNoStore = false;
     serverConnection = fwd->serverConnection();
 
-    if (fwd->serverConnection() != NULL)
+    if (fwd->serverConnection())
         _peer = cbdataReference(fwd->serverConnection()->getPeer());         /* might be NULL */
 
     flags.peering =  _peer;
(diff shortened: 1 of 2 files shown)


src/http.cc
```diff
@@ -75,10 +75,11 @@ static void copyOneHeaderFromClientsideRequestToUpstreamRequest(const HttpHeader
         HttpHeader * hdr_out, const int we_do_ranges, const Http::StateFlags &);
 
 HttpStateData::HttpStateData(FwdState *theFwdState) :
+    _peer(nullptr),
     AsyncJob("HttpStateData"),
     Client(theFwdState),
```
rousskov, line 80: We should initialize parent classes first:

```suggestion
    AsyncJob("HttpStateData"),
    Client(theFwdState),
    _peer(nullptr),
```

You can also  initialize `_peer`  in the class declaration instead so that you do not have to worry about the proper initialization order. We already initialize `upgradeHeaderOut` that way AFAICT.

Making all HttpStateData initializations consistent (e.g., moving `payloadTruncated` and `surrogateNoStore` initialization to the header file) may be declared in-scope as well, but that is optional. Your call.

src/http.cc
```diff
@@ -75,10 +75,11 @@ static void copyOneHeaderFromClientsideRequestToUpstreamRequest(const HttpHeader
...
-    lastChunk(0),
     httpChunkDecoder(NULL),
+    lastChunk(0),
```
rousskov, line 82: Please undo this out of scope and incorrect (AFAICT) change: lastChunk should be initialized before httpChunkDecoder because it is declared before httpChunkDecoder.

src/http.cc
```diff
@@ -88,7 +89,7 @@ HttpStateData::HttpStateData(FwdState *theFwdState) :
     surrogateNoStore = false;
     serverConnection = fwd->serverConnection();
 
-    if (fwd->serverConnection() != NULL)
+    if (fwd->serverConnection())
```
rousskov, line 92: Please avoid out of scope changes.

rousskov (changes requested):
Thank you for working on this cleanup!

If at all possible, please initialize all other HttpStateData data members that need initialization. I did not check all of them, but I see that at least `eof` is not explicitly initialized.

src/http.cc
yadij: FYI: This complexity of initialization order is one of the reasons C++11 initialization is different, and should be used.

yadij (changes requested):
Please fix by converting the HttpStateData class to C++11 initialization. The parent Client class is already converted.

eduard-bagdasaryan:
Please also consider removing an unused HttpStateData::read_sz field as a part if this cleanup.

src/http.cc
kinkie: I'm moving all to the header. It allows for more clarity IMO and might enable optimizations by the compiler

src/http.cc
kinkie: Right. Moving it all to the header

src/http.cc
kinkie: I recall we had agreed to modernize as we go over close changes, do I mis rembember?

kinkie:
> Please also consider removing an unused HttpStateData::read_sz field as a part if this cleanup.

done

kinkie:
> Please fix by converting the HttpStateData class to C++11 initialization. The parent Client class is already converted.

done

src/http.cc
rousskov: > I recall we had agreed to modernize as we go over close changes, do I mis rembember?

We agreed to modernize code that has to be modified for other reasons: Imagine that you postpone all modernization decisions until after everything else in the PR is settled. At that time, you should (in most cases) modernize the lines that the diff shows as added or modified. This rule of thumb is meant to reduce arguments about modernization scope, simplify associated decisions, and reduce diff noise.

In this particular case, this line will not be in the everything-but-modernization diff described above and, hence, should not be modernized.

src/http.cc
```diff
@@ -76,19 +76,14 @@ static void copyOneHeaderFromClientsideRequestToUpstreamRequest(const HttpHeader
...
-    lastChunk(0),
-    httpChunkDecoder(NULL),
-    payloadSeen(0),
-    payloadTruncated(0),
-    sawDateGoBack(false)
+    Client(theFwdState)
 {
     debugs(11,5, "HttpStateData " << this << " created");
     ignoreCacheControl = false;
     surrogateNoStore = false;
```
rousskov, line 83: These fields are now initialized in the header file.
```suggestion
```

rousskov (commented):
@kinkie, I have updated the PR title/description to reveal the primary change in the title, fix a misspelling, record recent changes, and make the future commit message look more consistent with similar recent messages. Please review and adjust further as needed.

rousskov (approved):
Thank you for addressing all of my concerns.

rousskov:
I have replaced "might be read" with "is read" in the PR description. There is no uncertainty in the "then" part of that assertion. The corresponding `if` statement is never true in the current code, but that is another story.

yadij (approved):
LGTM. Though you may want to consider converting the nested ReuseDecision class as well.

rousskov:
@kinkie, I am keeping an eye on this PR as well. I am not restarting the tests here right now because only one PR can be staged at any given time, and I already started PR 1030 tests. I suspect I should have restarted this PR tests first (because its staged commit is fresher), but it does not matter much -- I will make sure both PRs merge despite these false test failures.

PR #1030, opened 2022-04-24
Improve Ipc::Mem page limits accounting

kinkie:
Undocumented Ipc::Mem::PageId::maxPurpose is used to mark freed pages.
Freed pages do not participate in estimating future memory needs, so the
Limits array does not need to store their needs. However, some of the
needs estimation code is written without that assumption in mind because
maxPurpose is a legitimate/used purpose value (rather than an enum end
marker). This change improves this code in case our assumption change.

Detected by Coverity:
* CID 1504262: Out-of-bounds write (OVERRUN)
* CID 1504263: Out-of-bounds read (OVERRUN)

### DIFF
--- a/src/ipc/mem/Pages.cc
+++ b/src/ipc/mem/Pages.cc
@@ -21,7 +21,7 @@
 // TODO: make pool id more unique so it does not conflict with other Squids?
 static const char *PagePoolId = "squid-page-pool";
 static Ipc::Mem::PagePool *ThePagePool = 0;
-static int TheLimits[Ipc::Mem::PageId::maxPurpose];
+static int TheLimits[Ipc::Mem::PageId::maxPurpose+1];
 
 // TODO: make configurable to avoid waste when mem-cached objects are small/big
 size_t
--- a/src/ipc/mem/Pages.cc
+++ b/src/ipc/mem/Pages.cc
@@ -55,7 +55,7 @@ size_t
 Ipc::Mem::PageLimit()
 {
     size_t limit = 0;
-    for (int i = 0; i < PageId::maxPurpose; ++i)
+    for (int i = 0; i <= PageId::maxPurpose; ++i)
         limit += PageLimit(i);
     return limit;
 }


rousskov (approved):
Thank you for addressing this problem! I rewrote the PR description to explain what is actually going on in this code and added one more (untested) code change to remove the inconsistency that probably contributed to the wrong assumptions in the original PR description.

I also formatted Coverity references for consistency sake and put the primary change in the PR title.

Please review and adjust further as needed.

We can refactor this code to make it less surprising by adding `endPurpose` or a similar enum-end marker, but I hope that your minimal changes are sufficient for now. We have bigger fish to fry.

This is another Coverity false positive AFAICT.

kinkie:
Hm.. what's happening with Anubis and failing staging checks? Any idea @rousskov ?

yadij:
> 
> This is another Coverity false positive AFAICT.

Coverity is trying to tell us that several Must(position <= N) rather than Must(position < N) imply that position N is accessible on an N-1 length array.

rousskov:
> > This is another Coverity false positive AFAICT.
 
> Coverity is trying to tell us that several Must(position <= N) rather than Must(position < N) imply that position N is accessible on an N-1 length array.

It is a false positive because there are actually no accesses at (an invalid) position N. However, we do not need to agree on that assertion because this PR commit will not claim that it is a false positive. Feel free to adjust the corresponding Coverity defect metadata anyway you want.

rousskov:
> Hm.. what's happening with Anubis and failing staging checks?

The last one failed due to Semaphore CI or network instability AFAICT -- the test ran fine until either the testing vm or the connection to the remote testing box suddenly died (or something like that). I have seen that many times before. I have restarted the test and will keep an eye on it.

N.B. The best way to overcome false test failures on Semaphore CI (at least) is to restart the failed test(s) rather than to merge from master. This is especially true for failed _staging_ tests. Merging modifies the PR branch and, hence, forces a restart of _all_ tests, including tests that were previously successful (and including PR tests that must succeed before the PR is re-staged). Sorry for sounding like a broken record. Please just ignore this comment if you already know all this and are merging for other reasons!

rousskov:
> @kinkie removed the S-waiting-for-committer label 20 hours ago

BTW, when you, as an author, remove the `S-waiting-for-committer` label, it is best to either add `M-cleared-for-merge` to allow Anubis to merge the PR (usually the intent behind adding the `S-waiting-for-committer` label) or say something to clarify what the current PR state is -- why you have not added that `M-cleared-for-merge` label. Please note that neither label depends on the current test status -- CI tests is an orthogonal issue.

PR #1031, opened 2022-04-27
Ensure null-termination of string in negotiate_wrapper

kinkie:
Coverity identified a theoretical chance that a buffer may not be
null-terminated in negotiate_wrapper. The code flow is clean, adding a
forced null termination to apply defensive programming practices.

### DIFF
--- a/src/auth/negotiate/wrapper/negotiate_wrapper.cc
+++ b/src/auth/negotiate/wrapper/negotiate_wrapper.cc
@@ -257,6 +257,7 @@ processingLoop(FILE *FDKIN, FILE *FDKOUT, FILE *FDNIN, FILE *FDNOUT)
                 return 0;
             }
         }
+        buff[sizeof(buff)-1] = '\0';
         fprintf(stdout,"%s",buff);
         if (debug_enabled)
             fprintf(stderr, "%s| %s: Return '%s'\n",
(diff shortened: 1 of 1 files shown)


src/auth/negotiate/wrapper/negotiate_wrapper.cc
```diff
@@ -257,6 +257,7 @@ processingLoop(FILE *FDKIN, FILE *FDKOUT, FILE *FDNIN, FILE *FDNOUT)
                 return 0;
             }
         }
+        buff[sizeof(buff)-1] = '\0';
```
rousskov, line 260: Like you said, the official code probably does terminate `buff` correctly so this proposed code causes no harm, but if we ignore the well-hidden "already terminated correctly" fact, then this proposed code looks as if it silently truncates input! We can avoid that (false) implication by doing something like this:

```suggestion
        buff[sizeof(buff)-1] = '\0';
        assert(strlen(buff) < sizeof(buf)-1); // the previous assignment did not truncate
```

or this


```suggestion
        buff[sizeof(buff)-1] = '\0'; // paranoid; already terminated correctly
```

rousskov (commented):
If at all possible, please cite the exact Coverity defect ID(s) (with the corresponding defect gists) in the PR description like we did in commits 5d84beb, 9fa45f8, 699ade2. I would not claim "resolves" like we did in commits like ca0d7ed unless you have actually _tested_ that this change resolved that Coverity defect.

src/auth/negotiate/wrapper/negotiate_wrapper.cc
rousskov: Thank you for addressing this concern.

yadij:
Passed 10 days with no objections.

PR #1032, opened 2022-04-27
Initialise default_keytab in negotiate_kerberos_auth

kinkie:
Address a Coverity-identified issue, where default_keytab might be read
when uninitialised in negotiate_kerberos_auth.
Ensure it is initialised at declaration.

Detected by Coverity, CID 1503291 (Uninitialized scalar variable)

### DIFF
--- a/src/auth/negotiate/kerberos/negotiate_kerberos_auth.cc
+++ b/src/auth/negotiate/kerberos/negotiate_kerberos_auth.cc
@@ -370,6 +370,7 @@ main(int argc, char *const argv[])
     const unsigned char *spnegoToken = NULL;
     size_t spnegoTokenLength = 0;
 
+    default_keytab[0] = '\0';
     setbuf(stdout, NULL);
     setbuf(stdin, NULL);
 
--- a/src/auth/negotiate/kerberos/negotiate_kerberos_auth.cc
+++ b/src/auth/negotiate/kerberos/negotiate_kerberos_auth.cc
@@ -349,7 +349,7 @@ main(int argc, char *const argv[])
     char *service_principal = NULL;
     char *keytab_name = NULL;
     char *keytab_name_env = NULL;
-    char default_keytab[MAXPATHLEN];
+    char default_keytab[MAXPATHLEN] = {};
 #if HAVE_KRB5_MEMORY_KEYTAB
     char *memory_keytab_name = NULL;
     char *memory_keytab_name_env = NULL;
@@ -370,7 +370,6 @@ main(int argc, char *const argv[])
     const unsigned char *spnegoToken = NULL;
     size_t spnegoTokenLength = 0;
 
-    default_keytab[0] = '\0';
     setbuf(stdout, NULL);
     setbuf(stdin, NULL);
 


src/auth/negotiate/kerberos/negotiate_kerberos_auth.cc
```diff
@@ -370,6 +370,7 @@ main(int argc, char *const argv[])
     const unsigned char *spnegoToken = NULL;
     size_t spnegoTokenLength = 0;
 
+    default_keytab[0] = '\0';
```
rousskov, line 373: If we really want defensive programming, then we should use C++ initialization to zero the buffer (where the variable is declared):

```C++
char default_keytab[MAXPATHLEN] = "";
```

```C++
char default_keytab[MAXPATHLEN] = {};
```

The above is slightly more expensive than setting just one byte, but, in this case, all these variables are initialized just once at the program start time so we can afford that additional expense IMO. Same for other similar variables.

rousskov (changes requested):
> PR description: Address a Coverity false-positive

This does not look like a false positive to me: AFAICT, `default_keytab` is used uninitialized when check_k5_err() returns a truthy value (with both keytab_name and keytab_name_env being nil). Roughly speaking, `default_keytab` is used uninitialized when the corresponding krb5_init_context() call fails. Am I missing something?

If at all possible, please cite the exact Coverity defect ID(s) (with the corresponding defect gists) in the PR description.

src/auth/negotiate/kerberos/negotiate_kerberos_auth.cc
kinkie: Sounds like a good idea, and today I learn :)

rousskov (changes requested):
> PR description: Address a Coverity false-positive

This does not look like a false positive to me: AFAICT, `default_keytab` is used uninitialized when check_k5_err() returns a truthy value (with both keytab_name and keytab_name_env being nil). Roughly speaking, `default_keytab` is used uninitialized when the corresponding krb5_init_context() call fails. Am I missing something?

Thank you for addressing my other concern.

kinkie:
Added mention of Coverity ID and gist

rousskov (changes requested):
> PR description: Address a Coverity false-positive

This does not look like a false positive to me: AFAICT, `default_keytab` is used uninitialized when check_k5_err() returns a truthy value (with both keytab_name and keytab_name_env being nil). Roughly speaking, `default_keytab` is used uninitialized when the corresponding krb5_init_context() call fails. Am I missing something?

P.S. The above paragraph repeats my previous review because you have requested a review without addressing the above concern. If you checked that this is indeed a false positive, what did I miss in my analysis above?

kinkie:
> > PR description: Address a Coverity false-positive
> 
> This does not look like a false positive to me: AFAICT, `default_keytab` is used uninitialized when check_k5_err() returns a truthy value (with both keytab_name and keytab_name_env being nil). Roughly speaking, `default_keytab` is used uninitialized when the corresponding krb5_init_context() call fails. Am I missing something?
> 
> P.S. The above paragraph repeats my previous review because you have requested a review without addressing the above concern. If you checked that this is indeed a false positive, what did I miss in my analysis above?

Doh! I thought the other concern was mentioning the CID in the commit message, sorry. 
I'll look that up and signal you

kinkie:
> > PR description: Address a Coverity false-positive
> 
> This does not look like a false positive to me: AFAICT, `default_keytab` is used uninitialized when check_k5_err() returns a truthy value (with both keytab_name and keytab_name_env being nil). Roughly speaking, `default_keytab` is used uninitialized when the corresponding krb5_init_context() call fails. Am I missing something?

You are correct, except that I have added the initializer where default_keytab is declared as you have suggested, which means that the outcome might be somewhat unexpected but at least not a crash.

The whole logic of how identifying a keytab is harder to follow than it needs to be

> P.S. The above paragraph repeats my previous review because you have requested a review without addressing the above concern. If you checked that this is indeed a false positive, what did I miss in my analysis above?

sorry for the repetition. It is not a false positive, indeed. There are scenarios where this can cause a crash.
I'm adjusting the description

rousskov (approved):
Thank you for addressing my concern.

PR #1033, opened 2022-04-27
Defensive coding for ESI stackmember

kinkie:
Add default initializer for POD struct stackmember in ESI,
to ensure all its fields are initialised in all instantiations

Coverity issue 1494364; likely a false positive

### DIFF
--- a/src/esi/Expression.cc
+++ b/src/esi/Expression.cc
@@ -63,15 +63,15 @@ typedef enum {
 } literalhint;
 
 struct _stackmember {
-    evaluate *eval;
+    evaluate *eval = nullptr;
     union {
-        char *string;
+        char *string = nullptr;
         double floating;
         int integral;
     } value;
-    literalhint valuestored;
-    evaltype valuetype;
-    int precedence;
+    literalhint valuestored = ESI_LITERAL_INVALID;
+    evaltype valuetype = ESI_EXPR_INVALID;
+    int precedence = 0;
 };
 
 static void cleanmember(stackmember *);


src/esi/Expression.cc
```diff
@@ -63,15 +63,15 @@ typedef enum {
...
 
 struct _stackmember {
-    evaluate *eval;
+    evaluate *eval = nullptr;
     union {
-        char *string;
+        char *string = nullptr;
         double floating;
         int integral;
     } value;
```
rousskov, line 71: This implies that sizeof(double) <= sizeof(char*) and also opens up a possibility that a future bigger union member will be half-initialized by mistake. From a defensive coding point of view, we should do this instead AFAICT:

```suggestion
        char *string;
        double floating;
        int integral;
    } value = {};
```

rousskov (changes requested):
If at all possible, please add the corresponding defect gists to the Coverity defect ID in the PR description.

Disclaimer: I did not check whether the proposed defaults are appropriate for ESI.

src/esi/Expression.cc
yadij: We should be able to verify the situation with static_assert as well if there is any doubt.

rousskov: I found the following [cppreference](https://en.cppreference.com/w/cpp/language/aggregate_initialization) statement that invalidates my specific suggestion: "When a union is initialized by aggregate initialization, only its first non-static data member is initialized". That is not what GCC did in my test[^1], but that does not matter -- we should go by the standard, of course. My suggested ` = {}` change is worse than the current PR code because the latter is _explicitly_ flawed.

[^1]: I did that test before making my bad suggestion because I could not find the right standard promise at the time. GCC is probably buggy in this area as discussed on [SA](https://stackoverflow.com/questions/68881411/aggregate-initialization-of-a-union-in-c-with). The adjusted test (currently set to use clang) is [available](https://cpp.godbolt.org/z/ocqaz7s6T) if you want to play with it.

AFAICT, fixing this needs an explicit union constructor that memsets the whole union[^2]. There is no other way to do it in C++.

[^2]: That was actually my first idea, but then I saw other folks using aggregate initialization for unions and, after validating with an evidently buggy GCC, succumbed to the temptation :-(.

> Amos: We should be able to verify the situation with static_assert as well if there is any doubt.

I do not see how it is possible to reliably verify runtime zero-initialization with a static assert but would be glad to be proven wrong. With an explicit memset() solution, we would not have to double check.

yadij: > 
> > Amos: We should be able to verify the situation with static_assert as well if there is any doubt.
> 
> I do not see how it is possible to reliably verify runtime zero-initialization with a static assert 

I was thinking the below, but it does require that GCC initializer-list behaviour:
```
static_assert(Foo().value.string == nullptr);
static_assert(Foo().value.floating == 0.0);
static_assert(Foo().value.integral == 0);
```
clang complains about "active member" being different from the "accessed member" on the latter two.

rousskov: The static assert you were thinking about is not reliable enough to be worth using IMO: That assert may succeed[^1] in test environments where we want it to fail.

[^1]: It succeed when the memory behind the temporary Foo() object is zeroed for other reasons, which is often the case. If you have seen my [test](https://cpp.godbolt.org/), you know what kind of hoops we need to jump through to test this initialization reliably, but that testing approach is probably too cumbersome or impossible for a static assert. Fortunately, there are no reasons to double check a basic memset() result!

> clang complains about "active member" being different from the "accessed member" on the latter two.

That is a side effect of broken initialization. That specific complaint will go away when the initialization is fixed (or if the assert expression is refactored), but that complaint is not the reason why we should not use that static assert.

rousskov: @kinkie, please let me know if you want me to take a stab at fixing this using memset(). I do not know whether my verbal description was good enough, and the changes require touching more lines than a GitHub suggestion for this change request can express...

kinkie: I've taken a stab at it

src/esi/Expression.cc
```diff
@@ -62,16 +62,23 @@ typedef enum {
...
+    evaltype valuetype = ESI_EXPR_INVALID;
+    int precedence = 0;
+
+    stackmember() :
+        eval(nullptr), valuestored(ESI_LITERAL_INVALID),
+        valuetype(ESI_EXPR_INVALID), precedence(0)
+    {
+        memset(&value, 0, sizeof(value));
+    }
 };
```
rousskov, line 82: We should initialize the union by adding a union constructor (rather than declaring an encompassing class constructor, _duplicating_ all that class initialization code, and renaming that class). Here is a [sketch](https://cpp.godbolt.org/z/3crPqj8Gj).

I know that the union does not have a type name, but you can trivially name it (e.g., Value). Adding that name ought to require fewer changes than changing the existing name of the encompassing class!

rousskov, line 82: > PR description: Add default initializer for POD struct stackmember in ESI, to ensure all its fields are initialised in all instantiations

The current PR description does not quite match the current code.

rousskov (dismissed):

src/esi/Expression.cc
kinkie: Changed. Note that at least one c++ linter complains about memsetting a union that contains a double. I don't think we need to be concerned by this

Feel free to change the PR description if you feel it is not adequate

rousskov: > Feel free to change the PR description if you feel it is not adequate

Thank you for adjusting it. The current version matches the code so I left it as is after adding a Coverity defect type ID -- `UNINIT` -- something stable that is easier to search for than the textual description of that type.

rousskov: > [kinkie](https://github.com/kinkie) added S-waiting-for-reviewer and removed S-waiting-for-author labels [22 minutes ago](https://github.com/squid-cache/squid/pull/1033#event-6543285471)

This PR is still on your side because this change request has not been addressed.

rousskov: Perhaps you have not pushed your changes?

kinkie: Should be okay as of commit 42ae241 ; is it not?

rousskov: > Should be okay as of commit [42ae241](https://github.com/squid-cache/squid/commit/42ae241e05bf47cb4ae2ba84a6b2d36b05a6bb58) ; is it not?

Yes, it is OK (as far as this specific change request is concerned), thank you. I am resolving this change request.

FWIW, that commit did not exist on GitHub when I wrote my [comment](https://github.com/squid-cache/squid/pull/1033/files#r864133523).

rousskov (commented):
Our CI will make sure that the build is fixed, so I do not need to insist on any of these changes.

src/esi/Expression.cc
```diff
@@ -32,7 +32,7 @@
  * ) has 0
  */
 
-typedef struct _stackmember stackmember;
+struct stackmember;
```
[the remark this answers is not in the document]
kinkie: Is it a showstopper? current trunk feels old-style, and it feels silly to have a PR for that only.

src/esi/Expression.cc
```diff
@@ -62,16 +62,17 @@ typedef enum {
...
-    evaluate *eval;
-    union {
+struct stackmember {
+    evaluate *eval = nullptr;
+    union value_ {
         char *string;
         double floating;
         int integral;
+        _value() { memset(this, 0, sizeof(*this));  }
     } value;
```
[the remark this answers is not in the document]
kinkie: Yes. Adopting, thanks for reminding me

src/esi/Expression.cc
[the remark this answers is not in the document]
rousskov: > Is it a showstopper?

IMO, no. I am not blocking this PR.

> current trunk feels old-style, and it feels silly to have a PR for that only

You are correct on both counts, but neither observation justifies merging these out-of-scope changes:

* PR changes in question do not follow the current Squid code styling guidelines. They attempt to update the old-style code, but they are themselves problematic.

* There are hundreds of similar lines that need a style update. Such an update can and should be automated. We encourage (and some of us even require!) style updates for lines _modified_ due to in-scope changes (because doing so often does not complicate diff comprehension and often does not increase merge conflicts -- the diff already contains the changed lines), but these lines do qualify for that special exception.

* A PR should not have out of scope changes, for several obvious/basic reasons. A reviewer should not be put in the position where they have to defend the request to remove out of scope changes. If there are no special circumstances (none have been provided so far), out of scope changes should be removed.

rousskov (approved):

PR #1034, opened 2022-04-27
Refactor HttpStateData::ReuseDecision

kinkie:
Add in-class initializers for HttpStateData::ReuseDecision
to ensure all data members are initialised on each instantiation

### DIFF
--- a/src/http.h
+++ b/src/http.h
@@ -32,13 +32,14 @@ public:
         enum Answers { reuseNot = 0, cachePositively, cacheNegatively, doNotCacheButShare };
 
         ReuseDecision(const StoreEntry *e, const Http::StatusCode code);
+        ReuseDecision() = delete;
         /// stores the corresponding decision
         Answers make(const Answers ans, const char *why);
 
-        Answers answer; ///< the decision id
-        const char *reason; ///< the decision reason
-        const StoreEntry *entry; ///< entry for debugging
-        const Http::StatusCode statusCode; ///< HTTP status for debugging
+        Answers answer = reuseNot; ///< the decision id
+        const char *reason = nullptr; ///< the decision reason
+        const StoreEntry *entry = nullptr;
+        const Http::StatusCode statusCode = Http::StatusCode::scNone;
     };
 
     HttpStateData(FwdState *);
--- a/src/http.h
+++ b/src/http.h
@@ -32,7 +32,6 @@ public:
         enum Answers { reuseNot = 0, cachePositively, cacheNegatively, doNotCacheButShare };
 
         ReuseDecision(const StoreEntry *e, const Http::StatusCode code);
-        ReuseDecision() = delete;
         /// stores the corresponding decision
         Answers make(const Answers ans, const char *why);
 
--- a/src/http.cc
+++ b/src/http.cc
@@ -2620,7 +2620,7 @@ HttpStateData::abortAll(const char *reason)
 }
 
 HttpStateData::ReuseDecision::ReuseDecision(const StoreEntry *e, const Http::StatusCode code)
-    : answer(HttpStateData::ReuseDecision::reuseNot), reason(nullptr), entry(e), statusCode(code) {}
+    : entry(e), statusCode(code) {}
 
 HttpStateData::ReuseDecision::Answers
 HttpStateData::ReuseDecision::make(const HttpStateData::ReuseDecision::Answers ans, const char *why)


src/http.h
```diff
@@ -35,10 +35,12 @@ class HttpStateData : public Client
...
-        Answers answer; ///< the decision id
-        const char *reason; ///< the decision reason
-        const StoreEntry *entry; ///< entry for debugging
-        const Http::StatusCode statusCode; ///< HTTP status for debugging
+        Answers answer = reuseNot; ///< the decision id
+        const char *reason = nullptr; ///< the decision reason
+        const StoreEntry *entry = nullptr; ///< entry for debugging
+        const Http::StatusCode statusCode = scNone; ///< HTTP status for debugging
+    private:
+        ReuseDecision();
```
eduard-bagdasaryan, line 43: Can we delete the unused constructor instead of hiding it?

rousskov (dismissed):
> Add c++11 initializers

I cannot find a standard short term for "non-static data member [initialization](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2756.htm)", but it looks like many folks call these "in-class" initializers. I have also seen an NSDMI acronym being used. C++11 has added many new initializers, but you are talking about specific ones here. Please consider being more specific using one of those commonly used terms. I do not insist on that, of course.

> That class is always instantiated ... or the make() factory method

That last part is a lie. The class is instantiated using the constructor. The make() method is not really a factory method. It does not even return a new object. It only adjusts the fields of an existing (i.e. instantiated) decision object.

P.S. If I were to improve that class, I would probably rename the make() method into change() and set the default `reason` member value to "default" rather than nullptr.

src/http.h
kinkie: done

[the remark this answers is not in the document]
kinkie: aha, okay, deleting it

src/http.h
```diff
@@ -35,10 +35,10 @@ class HttpStateData : public Client
...
         Answers make(const Answers ans, const char *why);
 
-        Answers answer; ///< the decision id
-        const char *reason; ///< the decision reason
-        const StoreEntry *entry; ///< entry for debugging
-        const Http::StatusCode statusCode; ///< HTTP status for debugging
+        Answers answer = reuseNot; ///< the decision id
+        const char *reason = nullptr; ///< the decision reason
+        const StoreEntry *entry = nullptr;
+        const Http::StatusCode statusCode = Http::StatusCode::scNone;
```
rousskov, line 41: This _looks_ rather odd: We are declaring a constant and set it to scNone. That is not what is actually going on, but that is how this code _looks_ because the statusCode data member is (correctly) declared constant, but its default in-class initialization to scNone is ignored by the compiler in favor of the constructor code that actually sets this field to the right value (which remains constant).

The same could have been said about `entry` but that field was not declared constant.

```suggestion
        const StoreEntry * const entry;
        const Http::StatusCode statusCode;
```

src/http.h
```diff
@@ -35,10 +35,12 @@ class HttpStateData : public Client
         /// stores the corresponding decision
         Answers make(const Answers ans, const char *why);
 
-        Answers answer; ///< the decision id
-        const char *reason; ///< the decision reason
-        const StoreEntry *entry; ///< entry for debugging
-        const Http::StatusCode statusCode; ///< HTTP status for debugging
+        Answers answer = reuseNot; ///< the decision id
+        const char *reason = nullptr; ///< the decision reason
```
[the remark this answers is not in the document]
yadij: done in [5d1b192](https://github.com/squid-cache/squid/pull/1034/commits/5d1b192809000b22fdd9cb99439e5cd4c04ffff1)

src/http.h
yadij: Not sure why you think this looks odd. Perhapse you have forgotten that C/C++ const'ness always "looks odd" due to types reading right-to-left.

yadij (approved):

yadij:
FYI, This PR has passed the criteria of 2-votes (my review and implicitly @kinkie) for fast-track merge. I have restrained from clearing it to Anubis to give the open conversation with @rousskov a chance to be closed first.

src/http.h
rousskov: > Not sure why you think this looks odd

It looks odd because, normally, `const foo = scNone` means that `foo` is `scNone`; many readers would find it odd that a class needs a constant field like that -- the code could just use `scNone` directly... Here, the constant `foo` member actually gets a _different_ value elsewhere, so there is no problem other than an odd look of things.

> Perhapse you have forgotten that C/C++ const'ness always "looks odd" due to types reading right-to-left.

I have not, and this particular case reads left-to-right just fine. The problem is elsewhere (as I have detailed).

kinkie:
> > Add c++11 initializers
> 
> I cannot find a standard short term for "non-static data member [initialization](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2756.htm)", but it looks like many folks call these "in-class" initializers. I have also seen an NSDMI acronym being used. C++11 has added many new initializers, but you are talking about specific ones here. Please consider being more specific using one of those commonly used terms. I do not insist on that, of course.

"in-class" sounds reasonable. reworked.

> > That class is always instantiated ... or the make() factory method
> 
> That last part is a lie. The class is instantiated using the constructor. The make() method is not really a factory method. It does not even return a new object. It only adjusts the fields of an existing (i.e. instantiated) decision object.

removed the text


> P.S. If I were to improve that class, I would probably rename the make() method into change() and set the default `reason` member value to "default" rather than nullptr.

It's a POD class, so I mainly wanted to ensure that default values clearly show it has not been initialised.
Elsewhere we use an "invalid" value and scNone is the closest I could find in StatusCode

rousskov:
> PR description: Add in-class initializers for HttpStateData::ReuseDecision to ensure all data members are on each instantiation

Missing a word after "are"?

BTW, all members were already initialized "on each instantiation" before this PR, so the stated justification feels inapplicable to me.

> It's a POD class

FWIW, technically, ReuseDecision is [not](https://stackoverflow.com/a/5442736) a [POD](https://en.cppreference.com/w/cpp/named_req/PODType) type because it has a user-defined constructor. It is not a [trivial](https://en.cppreference.com/w/cpp/named_req/TrivialType) type either, for the same reason.

> I mainly wanted to ensure that default values clearly show it has not been initialised.

I understand this motivation, but when these special in-class "default" values are actually never applied (or never "used", never "seen" by any code, whatever you want to call it), it is difficult for me to see this as an overall improvement. Looks more like a distraction to me.

In-class initialization is a great tool, but, as any tool, it has its intended application areas. I am pretty sure this PR goes beyond those areas (i.e. abuses a great tool) and makes things slightly worse. However, since the damage is small, and we lack a set of good rules that can prove one of us right or that we can follow for consistency sake, I am not fighting it.

kinkie:
> > PR description: Add in-class initializers for HttpStateData::ReuseDecision to ensure all data members are on each instantiation
> 
> Missing a word after "are"?

right. added.

> BTW, all members were already initialized "on each instantiation" before this PR, so the stated justification feels inapplicable to me.

> > It's a POD class
> 
> FWIW, technically, ReuseDecision is [not](https://stackoverflow.com/a/5442736) a [POD](https://en.cppreference.com/w/cpp/named_req/PODType) type because it has a user-defined constructor. It is not a [trivial](https://en.cppreference.com/w/cpp/named_req/TrivialType) type either, for the same reason.

Aha. I was using my intuition of "a class that doesn't enforce conditions on its data". Fair enough

> > I mainly wanted to ensure that default values clearly show it has not been initialised.
> 
> I understand this motivation, but when these special in-class "default" values are actually never applied (or never "used", never "seen" by any code, whatever you want to call it), and when this in-class initialization can easily hide initialization bugs in future code, it is difficult for me to see this as an overall improvement. Looks more like a dangerous distraction to me.
> 
> In-class initialization is a great tool, but, as any tool, it has its intended application areas. I am pretty sure this PR goes beyond those areas (i.e. abuses a great tool) and makes things slightly worse. However, since the damage is small, and we lack a set of good rules that can prove one of us right or that we can follow for consistency sake, I am not fighting it.

You have a point. The existing constructor intialises all data members, and given there's no other way to build an uninitialised object, there's no reason to change. I'm abandining this PR

rousskov:
> I was using my intuition of "a class that doesn't enforce conditions on its data".

Construction aside, ReuseDecision does enforce some conditions on its data -- its statusCode data member cannot be changed.

> The existing constructor intialises all data members, and given there's no other way to build an uninitialised object, there's no reason to change. I'm abandining this PR

I am sorry that you have wasted your time on this. I wish I knew how to minimize such waste. On the bright side, we have discovered how ReuseDecision can be [improved](https://github.com/squid-cache/squid/pull/1034#pullrequestreview-955080990) (and I personally learned a formal POD definition) so perhaps this was not a total waste.

kinkie:
> I am sorry that you have wasted your time on this. I wish I knew how to minimize such waste. On the bright side, we have discovered how ReuseDecision can be [improved](https://github.com/squid-cache/squid/pull/1034#pullrequestreview-955080990) (and I personally learned a formal POD definition) so perhaps this was not a total waste.

I don't perceive the time spent writing this as waste; if anything it's dwarfed by the time we spent discussing it.
Thanks for doing that

rousskov:
> if anything it's dwarfed by the time we spent discussing it.

FWIW, I do include discussion time in my estimate of the wasted time.

PR #1035, opened 2022-04-27
Bug 5208: Part 1: Restart kids killed by SIGKILL

rousskov:
OOM killer uses SIGKILL. Squid did not restart kids killed by SIGKILL.
Kids are essential Squid components. Essential components should be
revived (by default) because providing a service without an essential
component would violate Squid functionality requirements.

Squid did not revive a kid killed by SIGKILL because we thought that
doing so will interfere with the "squid -k kill" feature that uses that
signal to kill the whole Squid instance. However, that feature does not
actually work[^1] -- the signal is sent to (and kills) the master
process only, the process which PID is in squid.pid file. This change is
orthogonal to fixing "squid -k kill" (a difficult out-of-scope project).

[^1]: Except in the special case of the no-daemon (squid -N) mode.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -564,6 +564,7 @@ mainHandleCommandLineOption(const int optId, const char *optValue)
             /** \li On interrupt send SIGINT. */
             opt_send_signal = SIGINT;
         else if (!strncmp(optValue, "kill", strlen(optValue)))
+            // XXX: In SMP mode, uncatchable SIGKILL only kills the master process
             /** \li On kill send SIGKILL. */
             opt_send_signal = SIGKILL;
 
--- a/src/ipc/Kid.cc
+++ b/src/ipc/Kid.cc
@@ -100,7 +100,6 @@ bool Kid::shouldRestart() const
              exitedHappy() ||
              hopeless() ||
              shutting_down ||
-             signaled(SIGKILL) || // squid -k kill
              signaled(SIGINT) || // unexpected forced shutdown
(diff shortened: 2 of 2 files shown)


src/main.cc
```diff
@@ -564,6 +564,7 @@ mainHandleCommandLineOption(const int optId, const char *optValue)
             /** \li On interrupt send SIGINT. */
             opt_send_signal = SIGINT;
         else if (!strncmp(optValue, "kill", strlen(optValue)))
+            // XXX: In SMP mode, uncatchable SIGKILL only kills the master process
```
rousskov, line 567: I would be happy to remove this comment if it is controversial.

yadij: What does the comment mean though?

yadij (approved):
Approving, though the comment could be better.

rousskov:
@eduard-bagdasaryan, please check on Anubis. It feels odd that it has not updated any PRs in 13+ hours...

eduard-bagdasaryan:
> check on Anubis

I do not see any problems with Anubis. What missing updates do you expect? There are two other PRs cleared for merge, but they are in the 'waiting for objections' state.

rousskov:
It looks like Anubis awoke, staged a merge commit, and pushed this PR forward three hours after I posted that comment. I was expecting Anubis to do all that before I posted that comment, but perhaps I missed some precondition. Thank you for checking!

PR #1036, opened 2022-04-28
Fix CID 1503308: Logically dead code (DEADCODE)

yadij:
The rsa variable has already been checked for nil and
cannot be unset before this extra check.

Detected by Coverity Scan


kinkie (approved):
go go go!

rousskov:
I do not understand why it is so hard to cite Coverity defects in a consistent way. IMHO, it is obvious that consistent citations save time in the long term and improve the overall impression of Squid code management. And even if one does not really care, it does not take much to satisfy another developer request in this case! However, every fresh PR does it differently (if at all) despite my repeated pleas. I have given up fighting these windmills and will no longer beg for Coverity citation consistency.

PR #1037, opened 2022-04-28
Remove unused libbsd

guijan:
Squid links against libbsd, but doesn't actually use any of its
functions.

I've build tested this on Alpine Linux.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2931,7 +2931,6 @@ AC_SEARCH_LIBS([res_init],[resolv])
 AC_SEARCH_LIBS([__res_search],[resolv])
 AC_SEARCH_LIBS([bind],[socket])
 AC_SEARCH_LIBS([opcom_stack_trace],[opcom_stack])
-AC_SEARCH_LIBS([strlcpy], [bsd])
 AC_SEARCH_LIBS([yp_match], [nsl nss_nis nss_nisplus])
 dnl Check for Winsock only on MinGW, on Cygwin we must use emulated BSD socket API
 if test "x$squid_host_os" = "xmingw" ; then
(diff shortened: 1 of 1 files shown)


kinkie:
Most of the changes are formatting related, it might be better to split the PR in two, one with the formatting changes and one with the libbsd related change

Also, strlcpy is a strong ask on openbsd due to security concerns. I'd encourage to try and build on that platform first - the build farm openbsd node is broken right now however, I plan to revive it in a few days

yadij:
strlcpy() itself has not been used for a long while. Though there are other functions provided by the library which would have been affecting ./configure symbol detection (eg the MD5 library API) once ```-lbsd``` is in LIBS.

@kinkie, this library is for non-BSD environments to access API which is provided by BSD natively. So testing for its removal should be most useful on all non-BSD nodes.

yadij:
OK to test

rousskov (changes requested):
Thank you for detecting and removing this unused code/library! If you can, please remove whitespace-only changes unrelated to libsd removal. Otherwise, please let me know and I will add the corresponding fixup myself.

rousskov:
> I've also changed the commit title to match the edited PR title.

Just FYI: PR branch commit message(s) are not used in the (effectively squashed by GitHub) official commit that will be added to master. We use PR title/description instead. There is no need to adjust PR branch commit messages or manually squash PR branch commits. See [Anubis](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message) documentation for details.

rousskov (approved):
Thank you for addressing the whitespace concern. I am not clearing this PR for merging (yet) because this is not my area of expertise, but I hope that others will do so soon.

yadij (approved):

guijan:
Sorry to notify you guys, I'm just mentioning another PR here to add a mention there.
https://github.com/openwrt/openwrt/pull/9714

PR #1040, opened 2022-04-29
Update status code 413 compliance

yadij:
The latest HTTP specification requires the data
transmitted in HTTP messages be called 'content'.

The official reason text for status code 413 has
been altered to comply.


yadij:
Passed 10-days with no objections, marking for merge.

PR #1043, opened 2022-05-05
Fix uninitalised var in eDirectory digest helper

kinkie:
In certain conditions, some string pointers in the eDirecotry digest
helper might be read before they are initialized. Initialize them.

Detected by Coverity, CID 1494358 (Uninitialized scalar variable).

### DIFF
--- a/src/auth/digest/eDirectory/ldap_backend.cc
+++ b/src/auth/digest/eDirectory/ldap_backend.cc
@@ -203,8 +203,10 @@ getpassword(char *login, char *realm)
     char **value = NULL;
     char *password = NULL;
     int retry = 0;
-    char filter[8192] = "";
-    char searchbase[8192] = "";
+    char filter[8192];
+    *filter='\0';
+    char searchbase[8192]
+    *searchbase='\0';
     char *universal_password = NULL;
     size_t universal_password_len = 256;
     int nmas_res = 0;
--- a/src/auth/digest/eDirectory/ldap_backend.cc
+++ b/src/auth/digest/eDirectory/ldap_backend.cc
@@ -203,8 +203,8 @@ getpassword(char *login, char *realm)
     char **value = NULL;
     char *password = NULL;
     int retry = 0;
-    char filter[8192] = {};
-    char searchbase[8192] = {};
+    char filter[8192] = "";
+    char searchbase[8192] = "";
     char *universal_password = NULL;
     size_t universal_password_len = 256;
     int nmas_res = 0;
@@ -213,8 +213,10 @@ getpassword(char *login, char *realm)
         if (usersearchfilter) {
             char escaped_login[1024];
             snprintf(searchbase, sizeof(searchbase), "%s", userbasedn);
+            searchbase[sizeof(searchbase)] = '\0';
             ldap_escape_value(escaped_login, sizeof(escaped_login), login);
             snprintf(filter, sizeof(filter), usersearchfilter, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login);
+            filter[sizeof(filter)] = '\0';
 
 retrysrch:
             debug("user filter '%s', searchbase '%s'\n", filter, searchbase);
--- a/src/auth/digest/eDirectory/ldap_backend.cc
+++ b/src/auth/digest/eDirectory/ldap_backend.cc
@@ -216,10 +216,8 @@ getpassword(char *login, char *realm)
         if (usersearchfilter) {
             char escaped_login[1024];
             std::snprintf(searchbase, sizeof(searchbase), "%s", userbasedn);
-            searchbase[sizeof(searchbase)-1] = '\0';
             ldap_escape_value(escaped_login, sizeof(escaped_login), login);
             std::snprintf(filter, sizeof(filter), usersearchfilter, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login);
-            filter[sizeof(filter)-1] = '\0';
 
 retrysrch:
             debug("user filter '%s', searchbase '%s'\n", filter, searchbase);
--- a/src/auth/digest/eDirectory/ldap_backend.cc
+++ b/src/auth/digest/eDirectory/ldap_backend.cc
@@ -11,13 +11,14 @@
  */
 #include "squid.h"
 
+#include <cstdio>
+
 #define LDAP_DEPRECATED 1
 
 #include "auth/digest/eDirectory/ldap_backend.h"
 
 #if _SQUID_WINDOWS_ && !_SQUID_CYGWIN_
 
-#define snprintf _snprintf
 #include <windows.h>
 #include <winldap.h>
 #ifndef LDAPAPI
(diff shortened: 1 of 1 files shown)


src/auth/digest/eDirectory/ldap_backend.cc
```diff
@@ -203,8 +203,8 @@ getpassword(char *login, char *realm)
     char **value = NULL;
     char *password = NULL;
     int retry = 0;
-    char filter[8192];
-    char searchbase[8192];
+    char filter[8192] = {};
+    char searchbase[8192] = {};
```
rousskov, line 207: Technically, this is a performance regression because we are now zeroing 16KB worth of stack on every getpassword() call. However, I assume that this function is called only at helper startup or we just do not care enough about sample helper performance to worry about such minor things.

rousskov (commented):
Thank you for fixing this. I polished the PR description, primarily to fix a misspelling and to reformat to save one line. Please adjust further as needed.

I have no objection for this going in.

src/auth/digest/eDirectory/ldap_backend.cc
yadij: FYI this function is called on every lookup

yadij: Does this initialization trick resolve the issue?
```suggestion
    char searchbase[8192];
    *searchbase = 0;
```

rousskov: The answer depends on what you mean by "issue":

* If you are asking about the performance regression, then, yes, zeroing just the first character would address that concern.

* If you are asking about the initialization problem being fixed by this PR, then I do not know the answer -- it depends on how these buffers are used, and I have not checked that. I bet @kinkie knows the answer.

@kinkie, please do not forget that there are _two_ buffers in this PR scope so if we decide to initialize just the first character it has to be done for both of them despite the above suggestions focusing on just one of the buffers.

Needless to say, there is a way to avoid all these low-level initialization tricks and risks, but that would require code refactoring that is probably beyond the current scope of this PR.

kinkie: Q: from a performance point of view, if our only aim is to ensure the first char is \0, which is more efficient among:

1. char s[1024] = {};
2. char string[1024] = { 0 };
3. char string[1024] = "";
4. char string[1024]; string[0] = 0;

?

rousskov: The last item (number four) is faster from C++11 rules point of view. That item initializes nothing and then sets the first array item. AFAICT, that is the absolute minimum of the (required by the question) initialization work for a stack-allocated array.

AFAICT, the first three items are identical in terms of what a non-optimizing compiler should produce per C++11 rules -- they all initialize the _whole_ `string` array with 1024 zeros. The cppreference site does not have all these rules in one place AFAICT, but you can find the appropriate rules in various places talking about initialization: Items one and three are covered at https://en.cppreference.com/w/cpp/language/zero_initialization and item two (after the type conversions) is covered at https://en.cppreference.com/w/cpp/language/aggregate_initialization#Character_arrays

FWIW, you can see whole-array memset()-calling initialization for your first three items in my basic [tests](https://godbolt.org/z/1774nqceG).

If an optimizing compiler knows that it does not need to initialize the whole thing, then, I am _guessing_, it does not have to do that.

src/auth/digest/eDirectory/ldap_backend.cc
```diff
@@ -213,8 +215,10 @@ getpassword(char *login, char *realm)
         if (usersearchfilter) {
             char escaped_login[1024];
             snprintf(searchbase, sizeof(searchbase), "%s", userbasedn);
+            searchbase[sizeof(searchbase)] = '\0';
             ldap_escape_value(escaped_login, sizeof(escaped_login), login);
             snprintf(filter, sizeof(filter), usersearchfilter, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login);
+            filter[sizeof(filter)] = '\0';
```
rousskov, line 221: Both changes above represent an out-of-bounds write AFAICT. Please fix.

How about using [std::snprintf](https://en.cppreference.com/w/cpp/io/c/fprintf) that the C++ standard guarantees to terminate?

```suggestion
            std::snprintf(searchbase, sizeof(searchbase), "%s", userbasedn);
            ldap_escape_value(escaped_login, sizeof(escaped_login), login);
            std::snprintf(filter, sizeof(filter), usersearchfilter, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login);
```

... and `#include <cstdio>`.

src/auth/digest/eDirectory/ldap_backend.cc
```diff
@@ -204,7 +204,9 @@ getpassword(char *login, char *realm)
     char *password = NULL;
     int retry = 0;
     char filter[8192];
-    char searchbase[8192];
+    *filter='\0';
+    char searchbase[8192]
+    *searchbase='\0';
```
rousskov, line 209: Just formatting (or consistency) improvements if you get a chance:

```suggestion
    *filter = '\0';
    char searchbase[8192]
    *searchbase = '\0';
```

rousskov (changes requested):

src/auth/digest/eDirectory/ldap_backend.cc
```diff
@@ -204,17 +205,21 @@ getpassword(char *login, char *realm)
...
     char *universal_password = NULL;
     size_t universal_password_len = 256;
     int nmas_res = 0;
     int rc = -1;
     if (ld) {
         if (usersearchfilter) {
             char escaped_login[1024];
-            snprintf(searchbase, sizeof(searchbase), "%s", userbasedn);
+            std::snprintf(searchbase, sizeof(searchbase), "%s", userbasedn);
+            searchbase[sizeof(searchbase)-1] = '\0';
```
rousskov, line 219: Here and elsewhere in the PR, I think we should not add termination when using std::snprintf() that is guaranteed to terminate and, I presume, has no known cases where that guarantee is broken. Avoiding noisy and error-prone explicit termination is one of the biggest selling points of using std::snprintf() instead of its less standardized and known-to-be-broken-in-some-environments variants!

```suggestion
```

Removing PR-added termination lines will also make the modified snprintf() callers more consistent with each other because other callers modified in this PR do not terminate.

You are the first author switching to std::snprintf(), so there may be surprises we do not know about, but I think we should assume standard-compliant STL implementations until proven otherwise (at least). Otherwise, we are likely to increase code inconsistencies while getting lost in second-guessing standard functions and ourselves.

src/auth/digest/eDirectory/ldap_backend.cc
```diff
@@ -204,17 +205,21 @@ getpassword(char *login, char *realm)
...
             ldap_escape_value(escaped_login, sizeof(escaped_login), login);
-            snprintf(filter, sizeof(filter), usersearchfilter, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login);
+            std::snprintf(filter, sizeof(filter), usersearchfilter, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login, escaped_login);
+            filter[sizeof(filter)-1] = '\0';
```
rousskov, line 222:
```suggestion
```

rousskov (commented):
LGTM with two minor changes.

PR #1044, opened 2022-05-07
Bug 5211: support.cc:355: "!filledCheck->sslErrors" assertion

rousskov:
One master transaction may encounter several certificate
validation-related errors because Squid may need to validate several
server certificates for one connection and may even need to open several
server connections. Since 2012 commit 4fb72cb, ssl_verify_cb()
incorrectly assumed that no past errors were possible. The callback
started to assert when recent code changes (commit e227da8?) improved
delivery of past errors to ACL checks.

While fixing this bug, we discovered that ServerBump::serverSession
ignored new TLS connection when Squid connected to another destination
after a failed TLS handshake. That stale value was then used to extract
stale information about past TLS errors. In some cases, that would
prevent Squid from hitting the now-fixed assertion. In other cases, it
could result in wrong sslproxy_cert_sign and sslproxy_cert_adapt
decisions. Now also fixed; see Ssl::ServerBump::attachServerSession().

A nil PeerConnector::check->sslErrors value may get stale because
PeerConnector::check is filled just once: sslError will remain nil
during the second validation pass (after PeerConnector fetches missing
intermediate certificates and revalidates), despite validation errors
discovered during the first pass. Today, ssl_verify_cb() does not really
use FilledChecklist::sslErrors preset by its PeerConnector, so a stale
value currently has no effect. Fixing this waiting-to-happen problem is
not trivial (we tried). A proper fix deserves a dedicated PR. Added XXX.

While working on these changes, we discovered that Squid implements two
mutually exclusive ssl_error (FilledChecklist::sslErrors) definitions:

* In sslproxy_cert_error context, ssl_error matches the last or current
  error being examined by the sslproxy_cert_error code.

* In sslproxy_cert_sign and sslproxy_cert_adapt context, ssl_error
  matches any previously observed validation error.

This change updates Squid documentation to reflect the above findings.
Unfortunately, it is not feasible to remove this unwanted directive
"sensitivity" without deprecating/replacing sslproxy_cert_error and
ssl_error: Selecting any one of the two ssl_error definitions above may
seriously (and silently) break existing configurations:

    # If all-errors semantics is adopted for sslproxy_cert_error, then
    # this configuration will start allowing certificates with _any_
    # error as long as that certificate also has a minorError:
    acl minorError ssl_error ...
    sslproxy_cert_error allow minorError

    # If last-error semantics is adopted for sslproxy_cert_sign, then
    # this configuration will stop properly signing certificates that
    # mimic self-signed real certificates if the very last error during
    # that real certificate validation was _not_ certSelfSigned:
    sslproxy_cert_sign signSelf ssl::certSelfSigned


src/cf.data.pre
```diff
@@ -1486,6 +1486,13 @@ IF USE_OPENSSL
 	acl aclname ssl_error errorname
 	  # match against SSL certificate validation error [fast]
 	  #
+	  # When used with sslproxy_cert_error, this ACL tests a single
+	  # certificate validation error currently being evaluated by that
+	  # directive. When used with slproxy_cert_sign or sslproxy_cert_adapt,
+	  # the ACL tests all past certificate validation errors associated with
+	  # the current Squid-to-server connection (attempt). This ACL is not yet
+	  # supported for use with other directives.
+	  #
```
rousskov, line 1495: > PR description: This change updates Squid documentation to reflect the above findings. Unfortunately, it is not feasible to remove this unwanted directive "sensitivity" without deprecating ssl_error and adding two new ACL types: Selecting any one of the two definitions above may seriously (and silently) break existing configurations: ...

I think it is best for this PR to focus on the bug fixes, but I cannot think of a better long-term approach than to deprecate and split `ssl_error` ACL, so that each of the two use cases (last/current-error and all-seen-errors) is covered by a dedicated ACL. Please let me know if you see a better solution.

src/ssl/support.cc
```diff
@@ -352,7 +352,7 @@ ssl_verify_cb(int ok, X509_STORE_CTX * ctx)
         if (error_no != SQUID_X509_V_ERR_INFINITE_VALIDATION) {
             if (check) {
                 ACLFilledChecklist *filledCheck = Filled(check);
-                assert(!filledCheck->sslErrors);
+                const auto savedErrors = filledCheck->sslErrors;
```
rousskov, line 355: AFAICT, today, nobody but this code uses the `check` object. And this code does not use the old sslErrors value (because it implements the last/current-error semantics). Thus, we do not need to save and restore old sslErrors. We can just remove the official assert(), overwrite the old value (nil or not), and then reset sslErrors to nil.

I think the current code is slightly better than blind overwrite because it relays the idea that sslErrors might become useful for other/future `check` object users. However, I will remove `savedErrors` if others think the code is better without them.

rousskov: > AFAICT, today, nobody but this code uses the `check` object. And this code does not use the old sslErrors value (because it implements the last/current-error semantics). Thus, we do not need to save and restore old sslErrors.

My logic above is flawed because the `check` object _itself_ "uses" the sslErrors value -- ACLFilledChecklist destructor cbdata-unlocks it. Thus, we need to either properly get rid of the old locked value or restore it; we cannot just forget it. The current restoring solution is better long-term (as detailed earlier), so I am closing this thread. The corresponding code in the current PR should stay as is.

src/ssl/support.cc
```diff
@@ -352,7 +352,7 @@ ssl_verify_cb(int ok, X509_STORE_CTX * ctx)
         if (error_no != SQUID_X509_V_ERR_INFINITE_VALIDATION) {
             if (check) {
                 ACLFilledChecklist *filledCheck = Filled(check);
-                assert(!filledCheck->sslErrors);
```
rousskov, line 355: I believe the removed assertion was conceptually wrong because this certificate validation code cannot know what has happened before it and should not assume that there were no errors earlier.

However, instead of just removing this assertion we could add a cache.log `WARNING` because if there were earlier errors then there is an increased chance that admin's `ssl_error` ACLs will work incorrectly here. Those ACLs could have been written under the assumption that those earlier errors would be detected here, and they will not be detected because this `sslproxy_cert_error` context must focus on detecting the _current_ validation error as we now document in squid.conf.documented.

After the fix in this PR, there are no known (to me) use cases where `sslErrors` would be set in this context, but there is no fundamental reason they cannot exist. Before this PR, we could rely on the assertion to alert us of their existence -- as byg 5211 did -- but if we just remove the assertion, we may not know whether this conflict between old errors and the new validation error does happen in the real world.

Please let me know if you think adding a `WARNING` is a good idea. Here is an incomplete sketch:

```C++
if (filledCheck->sslErrors)
    debugs(83, Important(65), "WARNING: Previously seen TLS errors present but will not be detected by sslproxy_cert_error");
```

I am worried that the above will not be enough to explain what is going on. On the other hand, if we never actually see that `WARNING` in the real world, then we will not have to deal with the "explanation" part of the problem; we can delay solving that (by providing a detailed explanation somewhere and referencing it from the error message) until we confirm that the conflict is still real.

P.S. This problem is one of the reasons we probably need to split `ssl_errors` into two as discussed [elsewhere](https://github.com/squid-cache/squid/pull/1044/files#r867287851).

rousskov, line 355: I believe the removed assertion was conceptually wrong because this certificate validation code cannot know what has happened before it and should not assume that there were no errors earlier.

However, instead of just removing this assertion we could add a cache.log `WARNING` because if there were earlier errors then there is an increased chance that admin's `ssl_error` ACLs will work incorrectly here. Those ACLs could have been written under the assumption that those earlier errors would be detected here, and they will not be detected because this `sslproxy_cert_error` context must focus on detecting the _current_ validation error as we now document in squid.conf.documented.

After the fix in this PR, there are no known (to me) use cases where `sslErrors` would be set in this context, but there is no fundamental reason they cannot exist. Before this PR, we could rely on the assertion to alert us of their existence -- as byg 5211 did -- but if we just remove the assertion, we may not know whether this conflict between old errors and the new validation error does happen in the real world.

Please let me know if you think adding a `WARNING` is a good idea. Here is an incomplete sketch:

```C++
if (filledCheck->sslErrors)
    debugs(83, Important(65), "WARNING: Previously seen TLS errors present but will not be detected by sslproxy_cert_error");
```

I am worried that the above will not be enough to explain what is going on. On the other hand, if we never actually see that `WARNING` in the real world, then we will not have to deal with the "explanation" part of the problem; we can delay solving that (by providing a detailed explanation somewhere and referencing it from the error message) until we confirm that the conflict is still real.

P.S. This problem is one of the reasons we probably need to split `ssl_errors` into two as discussed [elsewhere](https://github.com/squid-cache/squid/pull/1044/files#r867287851).

src/cf.data.pre
rousskov: @chtsanti came up with a better long-term solution: Instead of splitting `ssl_error` into two ACLs as described in my comment above, we should provide a new _directive_ that tests whether the last/current error should be honored:

1. Deprecate/rename the `ssl_errors` ACL to something like `tls_errors` or `tls::errors` (to make sure that admins currently relying on its "sometimes this matches just the current validation error" semantics stop doing that) but keep it a single ACL that can be used (and has the same meaning) in any already supported context.

2. Deprecate `sslproxy_cert_error` in favor of the new `sslproxy_certificate_validation_error` or a similar directive that, by its very structure/syntax is focuses on handling just one error at a time:

```
sslproxy_certificate_validation_error <error-name> allow|deny acl...
```

Again, these plans do not affect the proposed bug fix (i.e. this PR code changes). I am just sharing ideas on how to solve the configuration problems identified while working on this PR.

PR #1045, opened 2022-05-07
Cleanup client http request

yadij:
- fix spelling and grammar of comments
 - remove unused clientConnection member
 - remove redundant parameter names
 - fix whitespace within line syntax
 - C++11 initialize all members and sub-classes
 - shuffle member and method declarations to comply
  with coding guidelines

### DIFF
--- a/src/client_side_request.h
+++ b/src/client_side_request.h
@@ -82,11 +82,6 @@ public:
     /// the processing tags associated with this request transaction.
     const LogTags &loggingTags() const { return al->cache.code; }
 
-    /** Details of the client socket which produced us.
-     * Treat as read-only for the lifetime of this HTTP request.
-     */
-    Comm::ConnectionPointer clientConnection;
-
     /// Request currently being handled by ClientHttpRequest.
     /// Usually remains nil until the virgin request header is parsed or faked.
     /// Starts as a virgin request; see initRequest().
--- a/src/client_side_request.h
+++ b/src/client_side_request.h
@@ -39,11 +39,9 @@ class ClientHttpRequest
     CBDATA_CLASS(ClientHttpRequest);
 
 public:
-    ClientHttpRequest(ConnStateData *csd);
+    ClientHttpRequest(ConnStateData *);
+    ClientHttpRequest(ClientHttpRequest &&) = delete;
     ~ClientHttpRequest();
-    /* Not implemented - present to prevent synthetic operations */
-    ClientHttpRequest(ClientHttpRequest const &);
-    ClientHttpRequest& operator=(ClientHttpRequest const &);
 
     String rangeBoundaryStr() const;
     void freeResources();
@@ -62,7 +60,7 @@ public:
     StoreEntry *loggingEntry() const { return loggingEntry_; }
     void loggingEntry(StoreEntry *);
 
-    ConnStateData * getConn() const {
+    ConnStateData *getConn() const {
         return (cbdataReferenceValid(conn_) ? conn_ : nullptr);
     }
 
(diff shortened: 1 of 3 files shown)


yadij:
FTR: the removal of empty line at the end of [src/client_side_request.h](https://github.com/squid-cache/squid/pull/1045/files#diff-fbb2c9b970d883b7ff7507a3b228ad786be8e2eff7a784994fed4ae978ebcefe) is intentional - currently there are two where only one is needed.

src/client_side_request.h
```diff
@@ -82,155 +80,149 @@ class ClientHttpRequest
...
 public:
-    void startAdaptation(const Adaptation::ServiceGroupPointer &g);
+    void startAdaptation(const Adaptation::ServiceGroupPointer &);
     bool requestSatisfactionMode() const { return request_satisfaction_mode; }
 
+    /* AsyncJob API */
+    virtual bool doneAll() const override {
+        return Initiator::doneAll() &&
+               BodyConsumer::doneAll() &&
+               false; // XXX: why?
```
eduard-bagdasaryan, line 206: There is a similar 'false' in Adaptation::AccessCheck::doneAll() with "not done until mustStop" comment. Probably the same condition is relevant here, but I am not sure.

rousskov, line 206: This class has not been converted into a proper AsyncJob yet. Among many other (far more important) things, the class does not express the job termination condition correctly. Full refactoring is a difficult task because it requires, among other things, removing synchronous access to this class public methods.

```suggestion
               false; // TODO: Refactor into a proper AsyncJob
```

rousskov: This class has not been converted into a proper AsyncJob yet. Among many other (far more important) things, the class does not express the job termination condition correctly. Full refactoring is a difficult task because it requires, among other things, removing synchronous access to this class public methods.

```suggestion
               false; // TODO: Refactor into a proper AsyncJob
```

src/client_side_request.h
```diff
@@ -39,11 +39,9 @@ class ClientHttpRequest
     CBDATA_CLASS(ClientHttpRequest);
```
rousskov, line 39: [Jenkins](https://build.squid-cache.org/job/5-pr-test/4217/COMPILER=clang,OS=debian-stable,label=docker-build-host/console):

    src/client_side_request.h:39:5: error: 'toCbdata' overrides a member function
        but is not marked 'override' [-Winconsistent-missing-override]

To fix this particular error, you can probably use this approach:

```suggestion
#if USE_ADAPTATION
    CBDATA_CHILD(ClientHttpRequest);
#else
    CBDATA_CLASS(ClientHttpRequest);
#endif
```

... but I suspect it would not be sufficient to fix other similar problems, _and_ I think the best correct long-term fix is actually different: Adding an unconditional (virtual) AsyncJob base to the ClientHttpRequest class (and adjusting the code accordingly).

However, reviewing that proper fix while surrounded by all the noise in this PR would be very difficult (for me). Thus, I think the best solution here is one of the following two:

1. Undo "override" changes for ClientHttpRequest in this PR. Do not change ClientHttpRequest parents. Merge this PR. Followup with a dedicated PR that makes ClientHttpRequest an AsyncJob child even when `USE_ADAPTATION` is false. Add "override" specs in that second PR.
2. Pause the work on this PR. Do a dedicated PR that makes ClientHttpRequest an AsyncJob child even when `USE_ADAPTATION` is false. Resume the work on this PR after merging the first one and rebasing this one.

Neither solutions imply proper conversion of ClientHttpRequest into an AsyncJob. That should be done separately. It is probably best to do that _after_ the two PRs above. I do not recommend doing that third "proper conversion" PR at this time because I believe it requires developing more AsyncJob-related expertise.

rousskov (changes requested):

src/client_side_request.h
yadij: Done in [6364001](https://github.com/squid-cache/squid/pull/1045/commits/6364001fc669630be0b94754acc730e0af75b3a1)

rousskov (commented):
I have no objection to this PR going in.

FWIW, I did not check every code change, but I would do these polishing touches:

1. Add `Maintenance: ` prefix to the PR title (assuming this PR does not change Squid behavior).
2. Replace the itemization of edits in the PR description with a far more useful "no behavior changes" or another statement summarizing the _effects_ of this cleanup. If the `Maintenance: ` prefix is added to the title (item 1 above), then even this summary would be redundant -- the PR description can and probably should be empty.
3. Replace "client http request" with "ClientHttpRequest-related code" in the PR title.

kinkie (approved):
no objection either. go go go!

PR #1046, opened 2022-05-09
Source Format Enforcement

squidadm:


src/debug/Makefile.am
```diff
@@ -11,7 +11,7 @@ include $(top_srcdir)/src/TestHeaders.am
 noinst_LTLIBRARIES = libdebug.la
 
 libdebug_la_SOURCES = \
-	debug.cc \
 	Messages.h \
-	Stream.h
+	Stream.h \
+	debug.cc
```
rousskov, line 16: We should change the ordering script so that the order of these entries is case-insensitive because that would make it easier for humans to find the files they are looking for and to insert new files correctly. This script improvement does not have to block this PR, but if you agree and want to implement it, doing it before/for this PR will reduce the total number of Makefile.am changes.

yadij:
OK to test

yadij (approved):

PR #1048, opened 2022-05-09
Fixed (and renamed) Ssl::ReadX509Certificate() API

rousskov:
This pointer-returning non-throwing function was used to load an unknown
number of certificates from a file. A nil result normally indicated the
end of the certificate sequence but also a certificate loading failure.
The caller could not distinguish the two nil-result outcomes without
OpenSSL error stack analysis, which is difficult to do correctly, and
which none of the callers bothered to do, resulting in some certificate
configuration errors silently ignored by Squid.

The adjusted API uses exceptions to report errors. A nil result is still
used to report the end of the certificate sequence. To help callers that
must have at least one certificate, one function is now dedicated to
loading required certificate(s) while a companion function is added to
load optional ones.

Avoid exponential growth in the number of changed functions by wrapping
callers inside the not-ready-to-throw code chains (that cannot be easily
converted) with try/catch-return-false blocks.

Also removed the need for Ssl::readCertFromMemory() by adding
Ssl::ReadOnlyBioTiedTo(). Functions reading things (from various input
sources) should be kept separately from functions creating input sources
(for various reading functions). Otherwise, we will end up creating a
lot more functions (at least readers*creators) with virtually no gain in
performance or convenience. OpenSSL BIO API exists specifically to
separate I/O mechanisms from I/O users. Use it to our advantage.

We now also clear old/stale OpenSSL error stack before performing the
certificate loading operation that may trigger one or more errors.

Also forget stale errno when forgetting OpenSSL errors. This reset
avoids misleading error reports based on stale errno values.

Also detail certificate loading errors, now that they are not ignored.

ReportSysError was added to avoid SysErrorDetail::NewIfAny() memory
allocations when reporting basic system call errors. Perhaps there is a
way to reuse SysErrorDetail here via (enhanced) Optional, but the
resulting complexity and the current SBuf-based reporting overheads do
not justify the reuse of "if errno then strerror(errno)" logic. We may
revisit this when Optional supports non-trivial classes.

Also moved most SysErrorDetail implementation details to the newly added
.cc file (to reduce user exposure), replacing strerror() with xstrerr().

ForgetErrors guts were moved from Security to Ssl because we cannot link
helpers with src/security (yet). We may have to refactor Security to
make such reuse possible in the future.


src/ssl/cert_validate_message.cc
```diff
@@ -90,6 +91,22 @@ get_error_id(const char *label, size_t len)
...
 Ssl::CertValidationMsg::parseResponse(CertValidationResponse &resp)
+{
+    try {
+        tryParsingResponse(resp);
+        return true;
+    }
+    catch (...) {
+        debugs(83, DBG_IMPORTANT, "ERROR: Cannot parse sslcrtvalidator_program response:" <<
+               Debug::Extra << "problem: " << CurrentException);
+        return false;
```
rousskov, line 102: Here and in similar PR places, the try/catch blocks are stopping the propagation of code changes required for accommodating the now-throwing certificate reading API. As we improve callers, these blocks should be moved higher up and eventually be removed completely, but that usually requires non-trivial caller changes that are best done in dedicated PRs. Otherwise, the ensuing noise is likely to cause memory leaks and other problems.

rousskov:
Sample error output:

```
2022/05/09 15:32:39.818| ERROR: 'HTTP_port [::]:3128' cannot load intermediate certificates from '...pem':
    problem: cannot read a PEM-encoded certificate: 
    OpenSSL-saved error #1: 0x9091064
    exception location: gadgets.cc(762) ReadOptionalCertificate
```

```
$ openssl errstr 0x9091064
error:09091064:PEM routines:PEM_read_bio_ex:bad base64 decode
```

The ugly `'HTTP_port [::]:3128'` prefix is inherited from the official code and preserved in the PR for code consistency reasons. It should be fixed in a separate PR affecting several code areas untouched here.

chtsanti (approved):
Looks good and operational.

PR #1049, opened 2022-05-10
Shuffle TCP_NODELAY handling to libcomm API

yadij:
no behaviour changes

### DIFF
--- a/src/comm.cc
+++ b/src/comm.cc
@@ -77,9 +77,6 @@ static void commPlanHalfClosedCheck();
 static Comm::Flag commBind(int s, struct addrinfo &);
 static void commSetReuseAddr(int);
 static void commSetNoLinger(int);
-#ifdef TCP_NODELAY
-static void commSetTcpNoDelay(int);
-#endif
 static void commSetTcpRcvbuf(int, int);
 
 bool
@@ -495,11 +492,8 @@ comm_apply_flags(int new_socket,
             return -1;
         }
 
-#ifdef TCP_NODELAY
     if (sock_type == SOCK_STREAM)
-        commSetTcpNoDelay(new_socket);
-
-#endif
+        Comm::ApplyTcpNoDelay(new_socket);
 
     if (Config.tcpRcvBufsz > 0 && sock_type == SOCK_STREAM)
         commSetTcpRcvbuf(new_socket, Config.tcpRcvBufsz);
@@ -1116,22 +1110,6 @@ commSetCloseOnExec(int fd)
 
 #endif
 }
-
-#ifdef TCP_NODELAY
-static void
-commSetTcpNoDelay(int fd)
-{
-    int on = 1;
-
-    if (setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, (char *) &on, sizeof(on)) < 0) {
-        int xerrno = errno;
-        debugs(50, DBG_IMPORTANT, MYNAME << "FD " << fd << ": " << xstrerr(xerrno));
-    }
-
-    fd_table[fd].flags.nodelay = true;
-}
-
-#endif
 
 void
 comm_init(void)
@@ -1891,11 +1869,8 @@ comm_open_uds(int sock_type,
         }
     }
 
-#ifdef TCP_NODELAY
     if (sock_type == SOCK_STREAM)
-        commSetTcpNoDelay(new_socket);
-
-#endif
+        Comm::ApplyTcpNoDelay(new_socket);
 
     if (Config.tcpRcvBufsz > 0 && sock_type == SOCK_STREAM)
         commSetTcpRcvbuf(new_socket, Config.tcpRcvBufsz);
--- a/src/comm/Tcp.cc
+++ b/src/comm/Tcp.cc
@@ -11,6 +11,7 @@
 #include "squid.h"
 #include "comm/Tcp.h"
 #include "debug/Stream.h"
+#include "fde.h"
 
 #if HAVE_NETINET_TCP_H
 #include <netinet/tcp.h>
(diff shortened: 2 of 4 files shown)


src/comm.cc
```diff
@@ -1891,11 +1869,8 @@ comm_open_uds(int sock_type,
...
 
-#ifdef TCP_NODELAY
     if (sock_type == SOCK_STREAM)
-        commSetTcpNoDelay(new_socket);
-
-#endif
+        Comm::ApplyTcpNoDelay(new_socket);
 
     if (Config.tcpRcvBufsz > 0 && sock_type == SOCK_STREAM)
         commSetTcpRcvbuf(new_socket, Config.tcpRcvBufsz);
```
rousskov, line 1876: The generic comm_init() should not know about TCP-specific details such as application of the TCP_NODELAY socket option. I would generalize this code to look like this:

```suggestion
        Comm::ApplyTcpAdjustments(new_socket);
```

... and call ApplyTcpNoDelay() and commSetTcpRcvbuf() from ApplyTcpAdjustments() as needed. The ApplyTcpNoDelay() call may be inlined there if it is not used elsewhere (but does not have to be -- your call).

rousskov (changes requested):
I would remove the word "API" from the PR title -- we are creating a new public interface (arguably the "API" part) and moving its implementation (not "API"). I do not insist on this title change.

src/comm.cc
yadij: Out of scope for his PR which is only touching the TCP_NODELAY option setter.

rousskov: It is possible to address my change request while only touching the TCP_NODELAY option setter: The _essence_ of the problem here is an architectural mistake of leaving TCP_NODELAY knowledge in comm.cc while trying to move TCP_NODELAY handling to libcomm. That problem can be solved even if you do not want to move the lines related to Config.tcpRcvBufsz along with lines related to TCP_NODELAY. Here is the corresponding sketch:

```suggestion
        Comm::ApplyTcpAdjustments(new_socket);

    // TODO: Handle this in Comm::ApplyTcpAdjustments().
    if (Config.tcpRcvBufsz > 0 && sock_type == SOCK_STREAM)
        commSetTcpRcvbuf(new_socket, Config.tcpRcvBufsz);
```

src/comm.cc
```diff
@@ -1117,22 +1111,6 @@ commSetCloseOnExec(int fd)
...
-        int xerrno = errno;
-        debugs(50, DBG_IMPORTANT, MYNAME << "FD " << fd << ": " << xstrerr(xerrno));
-    }
-
-    fd_table[fd].flags.nodelay = true;
-}
-
-#endif
-
 void
```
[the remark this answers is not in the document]
yadij: The flag is set but never used, so removed in [7cab59a](https://github.com/squid-cache/squid/pull/1049/commits/7cab59aaae2d8f746b2601cc8b0ed549e28f38e8)

src/comm.cc
yadij: I understand what you are trying to achieve, but 
a) this shuffled setter is not for generic "TCP adjustments". It is performing one specific flag set.
b) a generic setter already exists as comm_apply_flags()  - refactoring comm_open_uds() to use it is also out of scope.

rousskov: > this shuffled setter is not for generic "TCP adjustments". It is performing one specific flag set.

Yes, but my change request does not contradict those facts. Code "performing one specific flag set" can and should be called from the "perform TCP adjustments" code. The code in comm.cc should only know about the latter. It would be nice to move all TCP adjustments, but if you insist on moving just one, I can support such limitation.

> a generic setter already exists as comm_apply_flags()

I am not asking to add another generic setter. I am asking to add ApplyTcpAdjustments() -- a function that handles TCP flags and such. That function should be in comm/Tcp.cc. This PR is moving some of the TCP adjustment code into Tcp.cc. All I am asking is that the moved code is accessed through the right public function so that comm_apply_flags() does not need to know about the TCP-specific concept of `TCP_NODELAY`. It is the right thing to do and it is trivial to do it.

This PR does not have to complete the work for implementing ApplyTcpAdjustments(), but it should start that work (instead of adding the wrong public API that will then need to be removed in favor of ApplyTcpAdjustments()).

> - refactoring comm_open_uds() to use it is also out of scope.

I am not asking for that refactoring.

rousskov: > yadij deleted a comment from rousskov [2 hours ago](https://github.com/squid-cache/squid/pull/1049#event-6592126494)

Not sure why you removed my comment, but I think that comment is valuable enough, so here it is again:

----

> I am not asking for that refactoring

... but TCP code in comm_apply_flags() and TCP code in comm_open_uds() should be kept in sync, of course.

BTW, the TCP (SOCK_STREAM) code in comm_open_uds() is currently unused, but we cannot assume that it will never be used without refactoring comm_open_uds(). Again, I do not suggest such refactoring in this PR.

rousskov: > Amos: b) a generic setter already exists as comm_apply_flags() - refactoring comm_open_uds() to use it is also out of scope.

> Alex: I am not asking for that refactoring

> Amos: The first step of that refactor is to create a function in libcomm with a generic name

Naming a new public function foo() instead of bar() has nothing to do with refactoring comm_open_uds() to use comm_apply_flags()!

No matter how you look at it, comm_open_uds() and comm_apply_flags() will need to be modified in this PR. They are already modified. Without significant changes that nobody is asking for, they should call the same function. They already call the same function. All that is left is to name that function correctly. This is not "refactoring comm_open_uds() to use comm_apply_flags()" by any standard. It is just copying comm_apply_flags() changes to comm_open_uds(), to keep the two functions in sync.

PR #1051, opened 2022-05-13
Add runner to initialize NTLM auth

yadij:


src/AuthReg.cc
```diff
@@ -45,10 +42,6 @@ Auth::Init()
 #if HAVE_AUTH_MODULE_NEGOTIATE
     static const char *negotiate_type = Auth::Negotiate::Scheme::GetInstance()->type();
     debugs(29,DBG_IMPORTANT,"Startup: Initialized Authentication Scheme '" << negotiate_type << "'");
-#endif
-#if HAVE_AUTH_MODULE_NTLM
-    static const char *ntlm_type = Auth::Ntlm::Scheme::GetInstance()->type();
-    debugs(29,DBG_IMPORTANT,"Startup: Initialized Authentication Scheme '" << ntlm_type << "'");
 #endif
     debugs(29,DBG_IMPORTANT,"Startup: Initialized Authentication.");
```
rousskov, line 46: Since this debugs() is now lying -- some authentication initialization happens later after PR changes -- and since some of us think it has wrong level, I suggest removing it completely:

```suggestion
```

src/auth/ntlm/Scheme.cc
```diff
@@ -9,16 +9,27 @@
...
+#include "base/RunnersRegistry.h"
 #include "debug/Messages.h"
 #include "debug/Stream.h"
 #include "helper.h"
 
-Auth::Scheme::Pointer Auth::Ntlm::Scheme::_instance = NULL;
+class NtlmAuthRr : public RegisteredRunner
+{
+public:
+    void bootstrapConfig() {
```
rousskov, line 20: Please be explicit about this method being a parent's method override.

```suggestion
    /* RegisteredRunner API */
    virtual void bootstrapConfig() override {
```

rousskov (commented):
I (still) have no objections to this PR going in (and left another minor polishing suggestion).

I hope Semaphore CI will resume seeing this branch after the next branch commit. I am guessing a GitHub notification was lost or something like that.

PR #1052, opened 2022-05-13
Load modules immediately

yadij:


src/base/RunnersRegistry.h
```diff
@@ -45,6 +45,11 @@ class RegisteredRunner
     /// Meant for initializing/preparing configuration parsing facilities.
     virtual void bootstrapConfig() {}
 
+    /// Called after parsing loadable_modules directive
+    /// Meant for initializing/preparing configuration parsing facilities
+    /// located in dynamic libraries [modules].
+    virtual void bootstrapModule() {}
```
rousskov, line 51: We should be moving towards configurations that can be fully validated before they are applied. AFAICT, the proposed event goes in the opposite direction by effectively applying a partial configuration (loading a configured module) before the entire configuration can be loaded/validated.

Squid already has an example of module-driven configuration interface -- eCAP adapters/modules are loaded after the entire configuration is parsed and are given their configuration information at that time. What prevents us from following that pattern for non-eCAP modules?

rousskov (changes requested):

src/base/RunnersRegistry.h
yadij: Time travel is not possible yet. As the comment says this is for initializing ConfigParser with details located inside the module (eg which flag values to accept, what parameter names are valid and so on.

PR #1053 is provided as an example of a module where the auth_param scheme name "ntlm" and and which options are valid vs invalid are unknown to ConfigParser before the module is loaded.

yadij: > > Time travel is not possible yet. As the comment says this is for initializing ConfigParser with details located inside the module (eg which flag values to accept, what parameter names are valid and so on.
> 
> Yes, my change request is based on those facts. While it is certainly _possible_ to start applying the `loadable_modules` directive before the entire configuration has been parsed, we should not do it because it goes against an important principle of configuration application [atomicity](https://en.wikipedia.org/wiki/Atomicity_(database_systems)) 

This principle comes from the realm(s) of data validity/consistency in threaded systems. Squid is neither of those - the principle is not currently part of Squid design and is actively violated by most of the configuration directives.

>and because it is not _necessary_ for configurable modules support (as has been proven by eCAP modules that are loaded and configured after the entire configuration has been parsed).

The style used by ecap requires strictly unordered options, prohibits parameters and module-specific directives existence. It is **not** a good example for configuring the modules this PR is supporting.

rousskov: > > Alex: While it is certainly possible to start applying the loadable_modules directive before the entire configuration has been parsed, we should not do it because it goes against an important principle of configuration application [atomicity](https://en.wikipedia.org/wiki/Atomicity_(database_systems))

> Amos: This principle comes from the realm(s) of data validity/consistency in threaded systems. 

The fundamental principle of operation atomicity is not tied to threaded systems. I bet it existed (perhaps under a different name) way before computers were invented!

> the principle is not currently part of Squid design

It is a part of Squid design when it comes to new code that I help develop. It should be a part of Squid design when it comes to any new code IMO. It is a good principle; following it results in better user experience than violating it does.

> and is actively violated by most of the configuration directives.

Yes, that is probably true, but is not relevant to my arguments. And, in most cases, removing those violations is trivial once the minority of difficult cases are dealt with. That is why we should avoid creating more of those difficult cases. They are the ones that matter, even though they are in the minority.

> The style used by ecap requires strictly unordered options,

It does not.

> prohibits parameters 

It does not.

> and [prohibits] module-specific directives existence.

Not on any fundamental level. I really doubt we should add module-specific _directives_, but it is not impossible to support even those if really necessary.

PR #1053, opened 2022-05-13
Make libntlm a dynamic module

yadij:
Demonstration dynamic module for Squid.
Requires PR #1051 and #1052

Moves the NTLM auth scheme logic to /usr/lib/squid/libntlm.so
which can be loaded using:
   loadable_modules /usr/lib/squid/libntlm.so

### DIFF
--- a/src/auth/Gadgets.cc
+++ b/src/auth/Gadgets.cc
@@ -130,10 +130,6 @@ authenticateCachedUsersList()
     if (Auth::SchemeConfig::Find("negotiate"))
         u1 = Auth::Negotiate::User::Cache()->sortedUsersList();
 #endif
-#if HAVE_AUTH_MODULE_NTLM
-    if (Auth::SchemeConfig::Find("ntlm"))
-        u2 = Auth::Ntlm::User::Cache()->sortedUsersList();
-#endif
     if (u1.size() > 0 || u2.size() > 0) {
         v2.reserve(u1.size()+u2.size());
         std::merge(u1.begin(), u1.end(),u2.begin(), u2.end(),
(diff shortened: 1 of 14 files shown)


src/auth/Gadgets.cc
```diff
@@ -129,10 +129,6 @@ authenticateCachedUsersList()
 #if HAVE_AUTH_MODULE_NEGOTIATE
     if (Auth::SchemeConfig::Find("negotiate"))
         u1 = Auth::Negotiate::User::Cache()->sortedUsersList();
-#endif
-#if HAVE_AUTH_MODULE_NTLM
-    if (Auth::SchemeConfig::Find("ntlm"))
-        u2 = Auth::Ntlm::User::Cache()->sortedUsersList();
 #endif
```
rousskov, line 132: Why are we no longer reporting NTLM users? The PR description is silent about this seemingly out of scope change.

src/auth/Gadgets.cc
```diff
@@ -129,10 +129,6 @@ authenticateCachedUsersList()
...
     if (u1.size() > 0 || u2.size() > 0) {
```
rousskov, line 133: After PR changes, `u2` is guaranteed to be empty here. The whole "paired" merging code in this function looks odd to me -- why not merge each container right after it is computed? -- but I do not insist on rewriting it. However, if you do not want to rewrite this code, please at least add a comment to note that `u2` is always empty here. For example,

```suggestion
    // TODO: Rewrite to avoid an always empty u2.
    if (u1.size() > 0 || u2.size() > 0) {
```

rousskov (changes requested):
Please ping me when this PR is no longer a draft (or if you have questions about it). Meanwhile, I left two change requests in hope to speed up future review.

yadij:
Abandoned. PR #1052 dependency not being added to Squid.

PR #1055, opened 2022-05-16
Fix PeerConnector::handleNegotiationResult() error debugging

rousskov:
ERROR: failure while establishing TLS connection on FD: 70x123456*1

* FD value was malformed. Now report all Connection details instead.
* Error details were not reported (a pointer value was reported).

Also, for error message prefix, use a single/constant phrase instead of
error-dependent phrases. This principle is for level-0/1 messages, but
it is OK to apply it here as well, especially given the related TODO.

Also improved RawPointer reporting so that we can use it for this fix.
Some of these changes are not necessary (for the final code state), but
all of them except the last one were necessary at some point during the
work on this fix, and they all may become handy in the future:

* Print nothing if ptr is nil. Both existing callers benefit from this.
* Allow the caller to set the label/object delimiter. Also faster.
* Support reporting object values on a dedicated Debug::Extra line.
* Support (and do not attempt to print) nil RawPointer() labels.

### DIFF
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -9,6 +9,8 @@
 #ifndef SQUID_SRC_BASE_IO_MANIP_H
 #define SQUID_SRC_BASE_IO_MANIP_H
 
+#include "debug/Stream.h"
+
 #include <iostream>
 #include <iomanip>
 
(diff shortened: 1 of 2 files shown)


rousskov:
Sample error message after the fixes:

```
2022/05/16 14:24:53.355| 83,2| PeerConnector.cc(276) handleNegotiationResult: ERROR: Cannot establish a TLS connection to conn27 local=127.0.0.1:56354 remote=127.0.0.1:443 HIER_DIRECT FD 24 flags=1:
    problem: failure
    detail: SQUID_TLS_ERR_CONNECT+TLS_LIB_ERR=1408F10B+TLS_IO_ERR=1
```

Let me know if I should move the Connection details reporting to its own Debug::Extra line.

yadij (approved):

PR #1057, opened 2022-05-18
TLS library-agnostic X509 certificate interrogation functions

rousskov:
Use added X509_check_issued() replacements. The only case left is in
src/ssl/gadgets.cc which is used by certificate helpers that cannot be
linked with libsecurity yet.

Use added X509_NAME_oneline() replacements, where feasible. This change
speeds up ssl_verify_cb() and other functions that used to extract and
copy certificate name into a buffer even when that name was unused
because debugging levels were not elevated enough, including by default.

Also fixes memory leak when debugging section 83 at level 3+ of an
OpenSSL-using Squid (missing name cleanup in clientNegotiateSSL()).

Also fixes a (usually symptom-free) sslcrtd bug: C strings allocated by
OpenSSL were freed by xfree() instead of OPENSSL_free().

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -522,8 +522,8 @@ squid_LDADD = \
 	DiskIO/libdiskio.la \
 	comm/libcomm.la \
 	anyp/libanyp.la \
-	$(SSL_LIBS) \
 	security/libsecurity.la \
+	$(SSL_LIBS) \
 	error/liberror.la \
 	ipc/libipc.la \
 	mgr/libmgr.la \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -522,8 +522,8 @@ squid_LDADD = \
 	DiskIO/libdiskio.la \
 	comm/libcomm.la \
 	anyp/libanyp.la \
-	security/libsecurity.la \
 	$(SSL_LIBS) \
+	security/libsecurity.la \
 	error/liberror.la \
 	ipc/libipc.la \
 	mgr/libmgr.la \
--- a/src/security/Certificate.cc
+++ b/src/security/Certificate.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 1996-2021 The Squid Software Foundation and contributors
+ * Copyright (C) 1996-2022 The Squid Software Foundation and contributors
  *
  * Squid software is distributed under GPLv2+ license and includes
  * contributions from numerous individuals and organizations.
--- a/src/security/Certificate.cc
+++ b/src/security/Certificate.cc
@@ -11,6 +11,10 @@
 #include "sbuf/SBuf.h"
 #include "security/Certificate.h"
 
+#if USE_OPENSSL
+#include "ssl/gadgets.h"
+#endif
+
 #include <iostream>
 #if USE_OPENSSL
 #if HAVE_OPENSSL_X509V3_H
@@ -30,10 +34,10 @@ Security::IssuerName(Certificate &cert)
 {
     SBuf out;
 #if USE_OPENSSL
+    Ssl::ForgetErrors();
     const auto s = X509_NAME_oneline(X509_get_issuer_name(&cert), nullptr, 0);
     if (!s) {
-        const auto x = ERR_get_error();
-        debugs(83, DBG_PARSE_NOTE(2), "WARNING: cannot get certificate Issuer: " << Security::ErrorString(x));
+        debugs(83, DBG_PARSE_NOTE(2), "WARNING: cannot get certificate Issuer:" << Ssl::ReportAndForgetErrors);
         return out;
     }
     out.append(s);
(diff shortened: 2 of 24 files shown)


rousskov:
This PR is based on @yadij  work on #831, but all bugs are my responsibility. I have extracted certificate manipulation functions from that PR branch (as of commit b96c3c1 dated 2022-01-18) so that they can be polished (here) and used (here and in other projects) sooner. Technically, adding those functions was outside #831 primary scope, so this PR also provides a more focused/smaller change.

src/security/Certificate.h
```diff
@@ -0,0 +1,43 @@
...
+SBuf SubjectName(Certificate &);
+
+/// The Issuer field of the given certificate (if found) or an empty SBuf.
+SBuf IssuerName(Certificate &);
+
+/// \returns whether cert was (correctly) issued by the given issuer
+/// Due to complexity of the underlying checks, it is impossible to clearly
+/// distinguish pure negative answers (e.g., two independent certificates)
+/// from errors (e.g., the issuer certificate lacks the right CA extension).
+bool IsIssuedBy(Certificate &cert, Certificate &issuer);
```
rousskov, line 30: Please let me know if you prefer other spellings for this function name. Here are a few reasonable variants:

1. IssuedBy(cert, issuer)
2. IssuerOf(issuer, cert) -- matches OpenSSL X509_check_issued() argument order
3. Issued(issuer, cert) -- matches OpenSSL X509_check_issued() argument order
4. AissuedB(issuer, cert) -- matches OpenSSL X509_check_issued() argument order; violates CamelCase for readability sake
5. AissuedByB(cert, issuer) -- violates CamelCase for readability sake

I am actually leaning towards IssuedBy() (and SelfSigned() below). IssuerOf() also looks good. There is no reason to use an "Is" prefix in this context -- the spelling cannot be mistaken for a command or request to modify the certificate; it is clearly a test or check.

I am not a fan of the last two, but will play along if others insist on one of them.

src/security/Certificate.cc
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    gnutls_datum_t name;
+    x = gnutls_x509_dn_get_str(subject, &name);
+    if (x != GNUTLS_E_SUCCESS) {
+        debugs(83, DBG_PARSE_NOTE(2), "WARNING: cannot describe certificate SubjectName: " << ErrorString(x));
+        return out;
+    }
+    out.append(reinterpret_cast<const char *>(name.data), name.size);
+    gnutls_free(name.data);
```
rousskov, line 98: The above and similar PR code is not exception-safe -- if out.append() throws, we are going to leak `name` memory. I think this should be fixed, but I am not sure whether this should be done here or in a dedicated PR: There are similar leaks in the official code, in the PR from where this code was copied, and properly addressing these leaks will require non-trivial out-of-scope changes (to make the solutions easier to reuse). I lean towards ignoring this (very unlikely!) leak in this PR and waiting for a proper reusable fix.

src/security/Certificate.h
yadij: > I am actually leaning towards IssuedBy() (and SelfSigned() below). 

likewise.

src/security/PeerConnector.cc
```diff
@@ -677,13 +677,13 @@ Security::PeerConnector::certDownloadingDone(SBuf &obj, int downloadStatus)
         if (!Ssl::findIssuerCertificate(cert, certsList, ctx)) {
             if (const auto issuerUri = Ssl::findIssuerUri(cert)) {
                 debugs(81, 5, "certificate " <<
-                       X509_NAME_oneline(X509_get_subject_name(cert), buffer, sizeof(buffer)) <<
+                       *cert <<
                        " points to its missing issuer certificate at " << issuerUri);
```
yadij, line 681: line wrapping here is no longer necessary for at least the first two lines of this debugs(). similar for the other touched one below.

yadij (changes requested):

src/security/Certificate.h
rousskov: Renamed in 0dbbf52. I also polished their descriptions a little in 9b02b6b.

src/security/PeerConnector.cc
rousskov: Adjusted to avoid arguing. Commit 4c16b6c.

src/Makefile.am
```diff
@@ -522,8 +522,8 @@ squid_LDADD = \
 	DiskIO/libdiskio.la \
 	comm/libcomm.la \
 	anyp/libanyp.la \
-	security/libsecurity.la \
 	$(SSL_LIBS) \
+	security/libsecurity.la \
```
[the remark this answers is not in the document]
rousskov: Undone in 2a3cac8. The official/natural order was giving me linking problems at some point, but I can no longer reproduce them with the current PR code, so there is, hopefully, no need to fiddle with this order (in this PR).

rousskov (commented):
I believe I have addressed all of your concerns. One of the change request was only partially addressed; if you insist on fixing those GnuTLS memory leaks in this PR, I will do that.

I also removed one branch-added comment in 9846104.

src/security/Certificate.cc
yadij: I still think this is far easier than you make out. I do not insist on this.

yadij (approved):

PR #187, opened 2018-04-15
Bug 4831: filter chain certificates for validity when loading

yadij:
51e09c08a5e6c582e7d93af99a8f2cfcb14ea9e6 adding
GnuTLS support required splitting the way
certificate chains were loaded. This resulted in the
leaf certificate being added twice at the prefix of a
chain in the serverHello.

It turns out that some recipients validate strictly that the
chain delivered by a serverHello does not contain extra
certificates and reject the handshake if they do.

This patch implements the XXX about filtering certificates
for chain sequence order and self-sign properties, added
in the initial PR. Resolving the bug 4831 regression and also
reporting failures at startup/reconfigure for admins.

Also, add debug display of certificate names for simpler
detection and administrative fix when loaded files fail
these tests.

PR #1058, opened 2022-05-22
Do not send more than one self-signed certificate

rousskov:
Configured chain L-M-R-R was sent as L-M-R-R instead of just L-M-R.

Configured chain R was sent as R-R instead of just R!

... where R is a self-signed (i.e. "Root CA") certificate that signed
(possibly indirectly, via an intermediate certificate M) the traffic
signing (i.e. "Leaf" or "end-entity") certificate L (if any).

Security::KeyData::loadX509ChainFromFile() has other problems, but
properly solving those problems needs significant code refactoring. It
is best done in a dedicated no-functionality-changes PR. The same
refactoring is likely to simplify the existing code logic, handling all
root certificates in one place.

Also removed disabled code so that we do not have to keep maintaining
it. Incorrectly disabling this code at the very end of commit 540e296
work created the bugs we are fixing in this branch. For some, it is
easier to comprehend active code without disabled code misdirection.
This disabled code removal does _not_ mean that Squid should keep
sending Root CA certificates (when they are not the signing/end-entity
certificates). However, stopping that behavior deserves a dedicated PR.

### DIFF
--- a/src/security/KeyData.cc
+++ b/src/security/KeyData.cc
@@ -112,6 +112,14 @@ Security::KeyData::loadX509ChainFromFile()
         while (const auto ca = Ssl::ReadOptionalCertificate(bio)) {
             // checks that the chained certs are actually part of a chain for validating cert
             if (IssuedBy(*latestCert, *ca)) {
+
+                if (SelfSigned(*latestCert)) { // TODO: Rename to lastChained
+                    Assure(SelfSigned(*ca));
+                    Assure(!SelfSigned(*cert)); // TODO: Rename to leafCert or signingCert
+                    debugs(83, DBG_PARSE_NOTE(2), "WARNING: Ignoring repeated Root CA: " << *ca);
+                    continue;
+                }
+
                 debugs(83, DBG_PARSE_NOTE(3), "Adding issuer CA: " << *ca);
                 // OpenSSL API requires that we order certificates such that the
                 // chain can be appended directly into the on-wire traffic.
--- a/src/security/KeyData.cc
+++ b/src/security/KeyData.cc
@@ -98,8 +98,10 @@ Security::KeyData::loadX509ChainFromFile()
         return;
     }
 
-    // TODO: Remove this diff minimization
-    {
+    if (SelfSigned(*cert)) {
+        debugs(83, DBG_PARSE_NOTE(2), "Signing certificate is self-signed: " << *cert);
+        // TODO: Warn if there are other (unusable) certificates present.
+    } else {
         debugs(83, DBG_PARSE_NOTE(3), "Using certificate chain in " << certFile);
         // and add to the chain any other certificate exist in the file
         CertPointer latestCert = cert;
@@ -108,10 +110,6 @@ Security::KeyData::loadX509ChainFromFile()
         // loop from the very first certificate in certFilename, and that
         // certificate is a copy of the already loaded _signing_ this->cert.
         while (const auto ca = Ssl::ReadOptionalCertificate(bio)) {
-
-            // XXX: When this->cert is self-signed, we chain its copy here,
-            // resulting in two copies of that signing certificate sent.
-
             // checks that the chained certs are actually part of a chain for validating cert
             if (IssuedBy(*latestCert, *ca)) {
                 debugs(83, DBG_PARSE_NOTE(3), "Adding issuer CA: " << *ca);
--- a/src/security/KeyData.cc
+++ b/src/security/KeyData.cc
@@ -89,7 +89,6 @@ void
 Security::KeyData::loadX509ChainFromFile()
 {
 #if USE_OPENSSL
-    // XXX: loadX509CertFromFile() has already opened and read this file.
     const char *certFilename = certFile.c_str();
     Ssl::BIO_Pointer bio(BIO_new(BIO_s_file()));
     if (!bio || !BIO_read_filename(bio.get(), certFilename)) {
@@ -106,9 +105,6 @@ Security::KeyData::loadX509ChainFromFile()
         // and add to the chain any other certificate exist in the file
         CertPointer latestCert = cert;
 
-        // XXX: The first ca value is usually not a CA certificate because we
-        // loop from the very first certificate in certFilename, and that
-        // certificate is a copy of the already loaded _signing_ this->cert.
         while (const auto ca = Ssl::ReadOptionalCertificate(bio)) {
             // checks that the chained certs are actually part of a chain for validating cert
             if (IssuedBy(*latestCert, *ca)) {
@@ -126,8 +122,6 @@ Security::KeyData::loadX509ChainFromFile()
                 latestCert = CertPointer(ca);
                 chain.emplace_back(latestCert);
             } else {
-                // XXX: This usually logs a misleading "ignoring CA" message for
-                // the non-CA signing certificate which is not actually ignored!
                 debugs(83, DBG_PARSE_NOTE(2), certFile << ": Ignoring non-issuer CA " << *ca);
             }
         }
--- a/src/security/KeyData.cc
+++ b/src/security/KeyData.cc
@@ -89,6 +89,7 @@ void
 Security::KeyData::loadX509ChainFromFile()
 {
 #if USE_OPENSSL
+    // XXX: loadX509CertFromFile() has already opened and read this file.
     const char *certFilename = certFile.c_str();
     Ssl::BIO_Pointer bio(BIO_new(BIO_s_file()));
     if (!bio || !BIO_read_filename(bio.get(), certFilename)) {
@@ -103,7 +104,14 @@ Security::KeyData::loadX509ChainFromFile()
         // and add to the chain any other certificate exist in the file
         CertPointer latestCert = cert;
 
+        // XXX: The first ca value is usually not a CA certificate because we
+        // loop from the very first certificate in certFilename, and that
+        // certificate is a copy of the already loaded _signing_ this->cert.
         while (const auto ca = Ssl::ReadOptionalCertificate(bio)) {
+
+            // XXX: When this->cert is self-signed, we chain its copy here,
+            // resulting in two copies of that signing certificate sent.
+
             // checks that the chained certs are actually part of a chain for validating cert
             if (IssuedBy(*latestCert, *ca)) {
                 debugs(83, DBG_PARSE_NOTE(3), "Adding issuer CA: " << *ca);
@@ -112,6 +120,8 @@ Security::KeyData::loadX509ChainFromFile()
                 latestCert = CertPointer(ca);
                 chain.emplace_back(latestCert);
             } else {
+                // XXX: This usually logs a misleading "ignoring CA" message for
+                // the non-CA signing certificate which is not actually ignored!
                 debugs(83, DBG_PARSE_NOTE(2), certFile << ": Ignoring non-issuer CA " << *ca);
             }
         }
--- a/src/security/KeyData.cc
+++ b/src/security/KeyData.cc
@@ -97,25 +97,13 @@ Security::KeyData::loadX509ChainFromFile()
         return;
     }
 
-#if TLS_CHAIN_NO_SELFSIGNED // ignore self-signed certs in the chain
-    if (SelfSigned(*cert)) {
-        debugs(83, DBG_PARSE_NOTE(2), "Certificate is self-signed, will not be chained: " << *cert);
-    } else
-#endif
+    // TODO: Remove this diff minimization
     {
         debugs(83, DBG_PARSE_NOTE(3), "Using certificate chain in " << certFile);
         // and add to the chain any other certificate exist in the file
         CertPointer latestCert = cert;
 
         while (const auto ca = Ssl::ReadOptionalCertificate(bio)) {
-
-#if TLS_CHAIN_NO_SELFSIGNED // ignore self-signed certs in the chain
-            // self-signed certificates are not valid in a sent chain
-            if (SelfSigned(*ca)) {
-                debugs(83, DBG_PARSE_NOTE(2), "CA certificate is self-signed, will not be chained: " << *ca);
-                continue;
-            }
-#endif
             // checks that the chained certs are actually part of a chain for validating cert
             if (IssuedBy(*latestCert, *ca)) {
                 debugs(83, DBG_PARSE_NOTE(3), "Adding issuer CA: " << *ca);


src/security/KeyData.cc
```diff
@@ -97,33 +98,36 @@ Security::KeyData::loadX509ChainFromFile()
         return;
     }
 
-#if TLS_CHAIN_NO_SELFSIGNED // ignore self-signed certs in the chain
     if (SelfSigned(*cert)) {
-        debugs(83, DBG_PARSE_NOTE(2), "Certificate is self-signed, will not be chained: " << *cert);
-    } else
-#endif
-    {
+        debugs(83, DBG_PARSE_NOTE(2), "Signing certificate is self-signed: " << *cert);
```
rousskov, line 102: The original (and previously disabled) "Certificate is self-signed, will not be chained: <signing certificate>" debugs() message contents sounded odd because the signing certificates should never be added to the chain object (signing certificates are sent because they are added to the TLS context object), regardless of whether they are self-signed. What we probably meant to say is that Squid will not form/sent a chain of CA certificates after this signing certificate because/when the signing certificate is self-signed. When the TODO below is implemented, we will warn about those "extra" certificates (if any).

src/security/KeyData.cc
```diff
@@ -97,33 +98,36 @@ Security::KeyData::loadX509ChainFromFile()
...
+        // TODO: Warn if there are other (unusable) certificates present.
+    } else {
         debugs(83, DBG_PARSE_NOTE(3), "Using certificate chain in " << certFile);
         // and add to the chain any other certificate exist in the file
         CertPointer latestCert = cert;
 
+        // XXX: The first ca value is usually not a CA certificate because we
+        // loop from the very first certificate in certFilename, and that
+        // certificate is a copy of the already loaded _signing_ this->cert.
         while (const auto ca = Ssl::ReadOptionalCertificate(bio)) {
```
rousskov, line 112: @chtsanti has already written the code to address this and most other (possibly all) XXXs and TODOs added in this PR.  I did not add those improvements to this bug-fixing PR because the associated refactoring noise completely hides the bug fix. I think those refactoring changes are best reviewed and merged in a dedicated PR, separate from this "surgical" bug fix.

If you insist on this or any other PR-added TODOs and XXXs to be removed/excluded, just say so. No need to justify your change request -- to save time, I will simply remove each controversial comment (including all of them) from this PR without arguing.

src/security/KeyData.cc
yadij: FTR; when seen in the context of cache.log trace with the surrounding certificate load sequence, this looks much more correct. It will only look "odd" if the file contains a misconfiguration - in which case the oddity makes it clearer to admin where/which misconfiguration occured.

src/security/KeyData.cc
yadij: This XXX implies that we are going to begin the practice of lying to admin about what is going on. Squid is currently very clear when it loads that first certificate that it is skipped due to not being the issuer of itself - except when it is.

src/security/KeyData.cc
```diff
@@ -89,6 +89,7 @@ void
 Security::KeyData::loadX509ChainFromFile()
 {
 #if USE_OPENSSL
+    // XXX: loadX509CertFromFile() has already opened and read this file.
```
yadij, line 92: FTR: previous iteration of this comment said:
```
    // XXX: This BIO loads the public cert as first chain cert,
    //      so the code appending chains sends it twice in handshakes.
```
before @rousskov requested its removal as not being true.

Ironic that the bug claimed by this PR is:  "Configured chain R was sent as R-R instead of just R!"

yadij (changes requested):
I object to this change in its entirety.

* It is a direct regression re-opening bug 4831.

* The code wrapped in TLS_CHAIN_NO_SELFSIGNED was intentionally disabled per discussion in https://github.com/squid-cache/squid/pull/187#discussion_r191411920.
  AFAIK there has been no evidence that it was needing to be enabled by anyone yet. As such the expectation is still that the code block be removed instead of enabled.

rousskov (commented):
> I object to this change in its entirety.

Your objection is based on several false assumptions:
 
> * It is a direct regression re-opening bug 4831.

What makes you think that? PR #187 fixing bug 4831 summarizes that bug as "the leaf certificate being added twice at the prefix of a chain in the serverHello". The changes in this PR add no certificates at all! Thus, they cannot reopen a bug about adding an extra certificate. If PR #187 summary was also incorrect, then please specify what regression you think this PR introduces.

> * The code wrapped in TLS_CHAIN_NO_SELFSIGNED was intentionally disabled per discussion in [#187 (comment)](https://github.com/squid-cache/squid/pull/187#discussion_r191411920).

While that code was intentionally disabled indeed, disabling it did not have (just) the effect that the author thought it had. Disabling it also created two bugs fixed in this PR. Unfortunately, I did not review disabling changes when they happened back then (AFAICT, I was only given 12 hours, from 2AM to 2PM to do that before those changes were merged). I am fixing those bugs now, in this PR.

> AFAIK there has been no evidence that it was needing to be enabled by anyone yet.

The above statement appears to assume that this PR re-enabled disabled code. That is not an accurate assumption. `TLS_CHAIN_NO_SELFSIGNED` disables two `if` statements in the official code. Only one of those two statements is resurrected in this PR. The other `if` statement is different from the disabled one. The result of this PR changes is _not_ equivalent to enabling that previously disabled code.

As for the evidence of need, please see this PR description for the summary of two bugs that this PR is fixing -- official Squid is sending two copies of the same certificate in certain cases. In one of those cases, Squid is not even (mis)configured to send two copies but sends two copies anyway. This is a bug-fixing PR.

> As such the expectation is still that the code block be removed instead of enabled.

Agreed, and that is how I recommend one views this PR -- the disabled code was removed. Now forget about it. It was never running and now it is gone. End of that story. What was added after that removal is new code that should be evaluated from scratch rather than based on (false) assumptions about what the disabled code did or did not do.

Please dismiss your negative review or re-review with the above corrections in mind.

P.S. As promised, I have removed all PR-added comments that you did not like and will not waste time on defending myself from personal attacks in reviews of those comments.

src/security/KeyData.cc
```diff
@@ -97,33 +98,36 @@ Security::KeyData::loadX509ChainFromFile()
...
+                }
+
                 debugs(83, DBG_PARSE_NOTE(3), "Adding issuer CA: " << *ca);
                 // OpenSSL API requires that we order certificates such that the
                 // chain can be appended directly into the on-wire traffic.
                 latestCert = CertPointer(ca);
                 chain.emplace_back(latestCert);
             } else {
+                // XXX: This usually logs a misleading "ignoring CA" message for
+                // the non-CA signing certificate which is not actually ignored!
```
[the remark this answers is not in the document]
yadij: Your response implies that you disagree with something I wrote. Please answer the question posed:
> Did you mean this to say "non-CA server certificate" ?

yadij:
> > I object to this change in its entirety.
> 
> Your objection is based on several false assumptions:
> 
> > * It is a direct regression re-opening bug 4831.
> 
> What makes you think that?

The client-side behaviour of this PR will be the same TLS verify failures. Duplicate leaf is just one of the cases that can lead to that rejection. more below.

> > * The code wrapped in TLS_CHAIN_NO_SELFSIGNED was intentionally disabled per discussion in [#187 (comment)](https://github.com/squid-cache/squid/pull/187#discussion_r191411920).
> 
> While that code was intentionally disabled indeed, disabling it did not have (just) the effect that the author thought it had.

Correction, the effect was exactly what the author *knew* it would do.

> Disabling it also created two bugs fixed in this PR.

Intentionally so, for bug compatibility with existing TLS/SSL clients that reject traffic where the root CA is omitted.

> > AFAIK there has been no evidence that it was needing to be enabled by anyone yet.
> 
> The above statement appears to assume that this PR re-enabled disabled code. 
>That is not an accurate assumption.

The PR diff shows it clearly doing so.

> `TLS_CHAIN_NO_SELFSIGNED` disables two `if` statements in the official code. Only one of those two statements is resurrected in this PR.

No. Both are converted from dead code to always-occuring.

The first one at line 100 simply has its #if wrappers removed.
 * This has the effect of erasing the chain portion of server handshake when Squid is configured with a self-signed tls-cert= option. This is a problem when the self-signed cert was cross-signed by an alternative CA and the client only trusts that alternative CA's root.


The second one at line 112 has its wrappers removed and moved down two lines into a sub-scope.
 * This has the effect of truncating a multi-root chain when the first root CA is encountered.
 *-> expected behaviour is that both branches of the multi-root chain are delivered. Optionally sans the root CA's (only).


> 
> > As such the expectation is still that the code block be removed instead of enabled.
> 
> Agreed, and that is how I recommend one views this PR -- the disabled code was removed. Now forget about it. It was never running and now it is gone. End of that story. What was added after that removal is new code that should be evaluated from scratch rather than based on (false) assumptions about what the disabled code did or did not do.
> 

The PR diff contradicts the validity of this approach. What was "added after that removal" is "new code" **identical** to the so-called "removed" code.

> Please dismiss your negative review or re-review with the above corrections in mind.
> 

If you really insist on continuing with this, please consider and test your changes with the following certificate chains in the tls-cert= file. At the very least case 1 & 2 have active users we should not break things for.

Case 1) L-A-R (aka. CA cross-signing)

 where L is self-signed, L is signed by A, and A signed by R.
  -> client verify should always work when client only trusts one of (L, A, R)

Case 2) L-A-B-R (aka. CA key transitions)

 where L is signed by A, L is signed by B, B signed by A, A signed by R, and B signed by R
  -> client verify should always work when client only trusts one of (A, B, R)

Case 3) L-A-R1-B-R2 (aka. composite chains)

 where L is signed by A, A signed by R1, L is signed by B, and B signed by R2
  -> client verify should always work when client only trusts one of (A, B, R1, R2)

** this case (3) is not supported by the current code. So I do not insist on it, but if you can get it going we will have some more happy users.

src/security/KeyData.cc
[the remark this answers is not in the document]
rousskov: > > Removed [a comment marking an alleged out-of-scope problem] to avoid arguing.

> Your response implies that you disagree with something I wrote.

My response does not imply a disagreement (or agreement). There may be a way to find an interpretation of what you wrote that I can agree with, but that takes time. Making sure that you agree with that interpretation takes even more time. Adjusting the comment to reflect that consensus takes time. I just want to avoid spending more time on an out-of-scope comment, especially since the problem the comment describes is very likely to be resolved in future PRs (where spending time on getting to an agreement regarding the problem and the best solution would be appropriate).

In other words, I was hoping for a quick win -- an informal marker of a problem I knew about -- but was content with settling with the status quo -- official code with no markings. The suggested simple drop-it-or-leave-it procedure would have accomplished a satisfactory outcome without time waste. What we ended up having instead is the same no-markings outcome, but with a 20+ minutes waste (so far).

> Please answer the question posed: Did you mean this to say "non-CA server certificate" ?

The answer is "No, I did not", but that answer does not imply that your wording is wrong (i.e. has no interpretation that makes it correct). The (obviously imperfect) comment said what I meant it to say. The comment is gone now. The alleged problem the gone comment was marking is out of this PR scope. Can we cut our losses here and move on, please?

rousskov:
> Amos: The client-side behaviour of this PR will be the same TLS verify failures [as in bug 4831].

Thank you for providing specific examples of the alleged "same failures" (later in your response). Please see below for their analysis. In summary, I am still unaware of any use case that we want to support, that was working before this PR, and that will stop working if this PR is merged. In other words, the PR fixes a few problems without breaking the already supported use cases.

> > > Amos: The code wrapped in TLS_CHAIN_NO_SELFSIGNED was intentionally disabled

> > Alex: Disabling it also created two bugs fixed in this PR.

> Amos: Intentionally so, for bug compatibility with existing TLS/SSL clients that reject traffic where the root CA is omitted.

The fixed bugs could not help with any legitimate use cases -- sending two copies of the same certificate is never useful. The cases you are thinking about are not the cases affected by this PR AFAICT.

Please note that this PR changes do not create situations with "traffic where the root CA is omitted": If the configured port chain ends in a root CA certificate R, then this PR does not affect R presence in the sent TLS ServerHello. In other words, PR code still sends R in cases where the official code sends R (e.g., R, L-R, L-M-R, and similar sent certificate sequences; see below for details and more cases). This PR only removes _extra_ R entries, sending the first R copy just as the official code sends it.

> Amos: The first one at line 100 simply has its #if wrappers removed.
> 
> This has the effect of erasing the chain portion of server handshake when Squid is configured with a self-signed tls-cert= option.

I am not sure what you mean by the "chain portion" in this context. To reduce the number of review iterations, I will cover several configuration cases where Squid is given a self-signed certificate (R) in `tls-cert`. Perhaps one of them will match the case you are thinking about (see further below for the legend). If not, please give a specific example.

1. `tls-cert=R`. Prior to this PR, Squid was sending R-R (i.e. two identical certificates -- a bug). After this PR, Squid sends just one R, as it should.

2. `tls-cert=R-R...`. Prior to this PR, Squid was sending R-R-R... (i.e. at least three identical certificates R). After this PR, Squid sends just one R, as it should.

3. `tls-cert=R-X...` (where X is not R). Prior to this PR, Squid was sending R-R (i.e. two identical certificates). After this PR, Squid sends just one R, as it should.

4. `tls-cert=L-M-R`. No changes here. Squid still sends L-M-R. This is a typical case.

5. `tls-cert=L-M-R-R...`. Prior to this PR, Squid was sending L-M-R-R... (i.e. two or more identical certificates R at the end of the chain). After this PR, Squid sends just L-M-R, as it should.

6. `tls-cert=L-M-R-X...` (where X is not R). No changes here. Squid still sends L-M-R, dropping X.

In the above, the left-most certificate in all chains is the "signing" certificate used to encrypt Squid port traffic. Each chain represents the certificates in the `tls-cert` bundle, in the specified order. Here is the legend for the certificate letters used above (just like in the PR description plus `X`):

* L: Leaf end-entity certificate. Cannot issue/sign any certificates.
* M: InterMediate CA certificate (that issued/signed L).
* R: Root CA (a.k.a. "self-signed") certificate. Issued/signed M (where present). Issued/signed R.
* X: Any certificate. The relationship (or lack thereof) with other certificates including other X certificates, is unimportant (except as explicitly stated in specific use cases).

I see no PR-relevant problems with the above use cases.

> Amos: The second one at line 112 has its wrappers removed and moved down two lines into a sub-scope.
> This has the effect of truncating a multi-root chain when the first root CA is encountered.

Please define "multi-root chain".

FWIW, the second added `if` only avoids adding extra copies of the already chained root CA certificate (i.e. case 5 above: `L-M-R-R...`). It does not change how Squid handles certificate bundles with multiple different roots -- Squid still sends only one chain to the client -- no changes there.

> *-> expected behaviour is that both branches of the multi-root chain are delivered. Optionally sans the root CA's (only).

AFAICT, the behavior you expect is not supported by official Squid (and that support is not added by this bug-fixing PR). Official Squid sends certificates that form a single chain. That chain starts with the first certificate in the bundle. Certificates are considered in the order they are stored, and the first-discovered path/chain wins. All other certificates are dropped. The `IssuedBy()` check takes care of all that (R duplication bugs notwithstanding).

> If you really insist on continuing with this, please consider and test your changes with the following certificate chains in the tls-cert= file. At the very least case 1 & 2 have active users we should not break things for.

Again, thank you for providing specific cases you are concerned about. Here is their analysis:
 
> Case 1) L-A-R (aka. CA cross-signing)
> 
> where L is self-signed, L is signed by A, and A signed by R. -> client verify should always work when client only trusts one of (L, A, R)

Official Squid sends L-L.
PR code sends L, fixing that bug.

Warning: Here, I am using L as _you_ define it for this case. (In my own examples/cases and in the PR description, L is a leaf end-entity certificate, never a self-signed CA certificate).

BTW, this is not a valid cross-signing case: If L is self-signed, then the client would not follow the chain from L to A. This is true even if the client does not trust L but has and trusts A. The chain always stops at the self-signed certificate. Before trust verification happens, clients restore certificate chains based on issued-by pairings (among other conditions). In case of a self-signed certificate, the issuer is the certificate subject itself so the client would not add any other certificate to the candidate chain -- that chain is done/terminated.

In a valid cross-signing case, L is not self-signed (it is a leaf end-entity or an intermediate CA certificate).

FWIW, the following Stack Exchange answer has nice illustrations for various cross-signing cases: https://security.stackexchange.com/a/15583


> Case 2) L-A-B-R (aka. CA key transitions)
> 
> where L is signed by A, L is signed by B, B signed by A, A signed by R, and B signed by R -> client verify should always work when client only trusts one of (A, B, R)

AFAICT, this case is not supported by the official code: Squid just sends L-A-R (because B did not issue A and, hence, Squid skips B after adding A to the chain). This PR does not change this behavior. Again, the official IssuedBy() check takes care of this.


> Case 3) L-A-R1-B-R2 (aka. composite chains)
> 
> where L is signed by A, A signed by R1, L is signed by B, and B signed by R2 -> client verify should always work when client only trusts one of (A, B, R1, R2)
> 
> this case (3) is not supported by the current code. So I do not insist on it, but if you can get it going we will have some more happy users.

Since this case is not yet supported, let's not discuss it in this bug-fixing PR. One this PR is merged, I am happy to discuss this case elsewhere (e.g., Zulip?). Just ping me then.

---

FWIW, we should really stop mentioning removed disabled code in any shape or form in this change request context IMO -- it only delays progress. Let's focus on the changes to the non-disabled code instead! Only they should matter in this discussion. However, if you insist on continuing that part of the conversation, here are the relevant corrections/clarifications:

> Amos: The second one at line 112 has its wrappers removed and moved down two lines into a sub-scope.

... and its `if` condition was changed! It is still an `if-condition-then-continue` code but with a new condition, and a new pre-condition. Any similarities with the previously disabled code are completely unimportant. There is no value (and clearly real harm!) in trying to compare this new code to the officially-disabled old code IMO.

> > Alex: `TLS_CHAIN_NO_SELFSIGNED` disables two `if` statements in the official code. Only one of those two statements is resurrected in this PR.

> Amos: No. Both are converted from dead code to always-occuring.

The first block of disabled code was effectively resurrected -- we are in agreement there.

The second disabled block was removed and was _not_ resurrected. Yes, there is a _similar_ block of new code, but it is different code in a different location, not a resurrection! That similar but new `if` statement has a different condition (and, hence, a different effect) compared to the second (disabled by `TLS_CHAIN_NO_SELFSIGNED`) `if` statement. There is an abridged diff below that pinpoints the difference between the removed and new code.

> > > Amos: As such the expectation is still that the code block be removed instead of enabled.

> > Alex: Agreed, and that is how I recommend one views this PR -- the disabled code was removed. Now forget about it. It was never running and now it is gone. End of that story. What was added after that removal is new code that should be evaluated from scratch rather than based on (false) assumptions about what the disabled code did or did not do.
 
> Amos: The PR diff contradicts the validity of this approach. What was "added after that removal" is "new code" **identical** to the so-called "removed" code.

... except it is not identical! Here is an abridged diff that might help you see the difference:

```diff
 while (...) {

-#if TLS_CHAIN_NO_SELFSIGNED
-    if (conditionB)
-       continue;
-#endif

    if (conditionX) {
+       if (conditionC)
+           continue;
        ...
```

See? ConditionB is gone. The added `continue` code runs if conditionX _and_ conditionC are true. Those are different from conditionB! Or, simplifying further, the PR change is equivalent to this:

```diff
 while (...) {

-#if TLS_CHAIN_NO_SELFSIGNED
-    if (conditionB)
-       continue;
-#endif

+   if (conditionXC)
+       continue;

    if (conditionX)
        ...
```

Where conditionXC is (conditionX and conditionC). ConditionXC is not the previously disabled and now removed conditionB. No resurrection here...

src/security/KeyData.cc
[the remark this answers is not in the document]
yadij: Okay.

yadij:
> Please define "multi-root chain".

see [commet ](https://github.com/squid-cache/squid/pull/1058#issuecomment-1146622862) case (2) and (3).


> See? ConditionB is gone. The added continue code runs if conditionX and conditionC are true. Those are different from conditionB! Or, simplifying further, the PR change is equivalent to this:

What you are calling conditionC(ca) is actually conditionB(++ca). You have simply swapped the order of two **"&&"** criteria being evaluated and obfuscated them across loop cycles.

Unwrapping the loops makes this clearer:
Existing code:
```
// while:
if (conditionB( n=0 ) ) continue;
if (conditionX( n=1 ) )
 ...

if (conditionB( n=1 ) ) continue;
if (conditionX( n=2 ) )
 ...

if (conditionB( n=2 ) ) continue;
if (conditionX( n=3 ) ) ...
```

This PR:
```
if (conditionB( n=0 )) return;

// while:
if (conditionX( n=1 )
if (conditionB( n=1) ) continue;
 ...

if (conditionX( n=2 )
if (conditionB( n=2) ) continue;
  ...

if (conditionX( n=3 )
if (conditionB( n=3) ) continue;
 ...
```

yadij:
I am giving up arguing about this. I will let the users bug reports demonstrate issues in future.

rousskov:
Here is another simple way to prove that this PR effect is not resurrection of commented out code. Consider an ordinary certificate bundle containing L-M-R.

* Official code built with `TLS_CHAIN_NO_SELFSIGNED=1`: Sends L-M (because conditionB skips ca=R).
* PR code: Sends L-M-R (because conditionC(M) is still false during ca=R iteration).

I hope that is sufficient proof that, for certificate bundles that do not start with a self-signed certificate, this PR does not resurrect the previously commented out code but implements different logic.

> I will let the users bug reports demonstrate issues in future.

No bug reports are needed -- we already know this code is buggy! This PR does not address all known bugs. It only eliminates duplicate root certificate entries in sent chains.

----

> > Alex: ConditionB is gone. The added continue code runs if conditionX and conditionC are true. Those are different from conditionB! Or, simplifying further, the PR change is equivalent to this:

> Amos: What you are calling conditionC(ca) is actually conditionB(++ca). 

Even this simple statement is false: conditionC is not evaluated for `ca`. It is evaluated for `latestCa` that may _skip_ some `ca`s.

More importantly, we must not ignore the presence of conditionX in the PR code (which is ANDed with conditionC). To claim resurrection, the entire condition XC would have to be equivalent to condition B.

> Amos: You have simply swapped the order of two **"&&"** criteria being evaluated and obfuscated them across loop cycles.

No, the PR code is substantially different from the official code with built with `TLS_CHAIN_NO_SELFSIGNED=1`. The code manipulations you are describing do not preserve the (incorrect) outcome of the official code built with `TLS_CHAIN_NO_SELFSIGNED=1`, as demonstrated by the L-M-R example above.

PR #1060, opened 2022-05-26
Maintenance: Reduce HTTP/FTP code duplication

eduard-bagdasaryan:
Removed code duplication  probably caused bug fixed in commit 049eeeb.

### DIFF
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -3336,38 +3336,13 @@ clientHttpConnectionsOpen(void)
             continue;
         }
 
-        // Fill out a Comm::Connection which IPC will open as a listener for us
-        //  then pass back when active so we can start a TcpAcceptor subscription.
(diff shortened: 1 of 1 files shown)


rousskov (approved):
Nice, thank you!

Our `ftp_port` documentation already implies that `worker-queues` should be honored. I assume that the option is already accepted by the official code when parsing `ftp_port` configuration. We just did not pass the option to Comm when creating listening sockets (in `ftp_port` cases). Please correct me if my assumptions are wrong.

yadij (changes requested):
Please use separate PR for code cleanup (large clientHttpConnectionsOpen changes) and feature addition (small clientStartListeningOn changes).

eduard-bagdasaryan:
Separated into  [PR1061](https://github.com/squid-cache/squid/pull/1061).

yadij (approved):

PR #1063, opened 2022-06-05
Use SBuf for Fs::Ufs::RebuildState file paths

kinkie:
In theory, a combination of a non-terminating-on-overflows snprintf()
and an incorrectly guessed MAXPATHLEN (or a cache_dir root path being at
the edge of a correctly guessed MAXPATHLEN) may result in non-terminated
fullpath. Use SBuf to avoid these particular headaches.

Detected by Coverity. CID 1461166: Copy of overlapping memory
(OVERLAPPING_COPY).

### DIFF
--- a/src/fs/ufs/RebuildState.cc
+++ b/src/fs/ufs/RebuildState.cc
@@ -43,8 +43,6 @@ Fs::Ufs::RebuildState::RebuildState(RefCount<UFSSwapDir> aSwapDir) :
     _done(false),
     cbdata(NULL)
 {
-    *fullpath = 0;
-    *fullfilename = 0;
 
     /*
      * If the swap.state file exists in the cache_dir, then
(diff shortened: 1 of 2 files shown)


rousskov (approved):
Thank you for addressing this Coverity-reported defect and improving code.

I fixed PR description formatting and spelling. While doing that, I noticed that the bug discovered by Coverity is theoretically real and adjusted that part of the description as well.

I also adjusted PR title to point explicitly to UFS RebuildState. Rock as a similar class called Rebuild, and I wanted to avoid any confusion, especially since we might eventually remove the "State" noise from the UFS class name.

Please check PR title/description and adjust further as needed. I do not insist on any specific wording here (but Anubis will insist on the correct format, and we should avoid misspellings).

kinkie:
Cleared from me, removing label and waiting for fast track

PR #1065, opened 2022-06-05
Add default initializers in RandomUuid

kinkie:
Coverity defect 1504546 - Uninitialized scalar field

RandomUuid only initialises some data fields on serialization.
Assign a default value on constrution

### DIFF
--- a/src/base/RandomUuid.h
+++ b/src/base/RandomUuid.h
@@ -62,12 +62,12 @@ private:
      * accurately represent the actual UUID version 4 structure which, the six
      * version/variant bits aside, contains just random bits.
      */
-    uint32_t timeLow;
-    uint16_t timeMid;
-    uint16_t timeHiAndVersion;
-    uint8_t clockSeqHiAndReserved;
-    uint8_t clockSeqLow;
-    uint8_t node[6];
+    uint32_t timeLow = 0;
+    uint16_t timeMid = 0;
+    uint16_t timeHiAndVersion = 0;
+    uint8_t clockSeqHiAndReserved = 0;
+    uint8_t clockSeqLow = 0;
+    uint8_t node[6] = {};
 };
 
 std::ostream &operator<<(std::ostream &os, const RandomUuid &uuid);


rousskov (changes requested):
> PR description: RandomUuid only initialises some data fields on serialization.

There are two problems with this assertion:

* The (default) constructor (that this Coverity defect talks about) is not related to serialization. The constructor just creates a UUID object. There is a deserialization constructor, but that is a different Coverity defect 1504548 (that you have already correctly classified as a false positive on Coverity side).
* The (default) constructor does initialize all data fields AFAICT. See two memcpy() calls into raw() and the corresponding static_assert(). Please double check your assessment.

I am against using in-class initialization as proposed in this PR because it will hide true initialization errors (if we do make them in the future). It also uses essentially bogus values for initialization (zero is not a valid value for some of these data members).

@eduard-bagdasaryan, do you know of a good way to modify Squid code or Coverity model to help Coverity see that we are initializing all fields when copying into raw()? I bet it is not possible, especially when we have to do two memcpy() operations, but I would be glad to be proven wrong.

If there is no good way to adjust our code to convince Coverity that all fields are initialized (without in-class initialization), then I think we will be better off with keeping the code as is so that Coverity flags new (true) bugs if we happen to add them.

kinkie:
> > PR description: RandomUuid only initialises some data fields on serialization.
> 
> There are two problems with this assertion:
> 
> * The (default) constructor (that this Coverity defect talks about) is not related to serialization. The constructor just creates a UUID object. There is a deserialization constructor, but that is a different Coverity defect 1504548 (that you have already correctly classified as a false positive on Coverity side).
> * The (default) constructor does initialize all data fields AFAICT. See two memcpy() calls into raw() and the corresponding static_assert(). Please double check your assessment.

You are right. Coverity can't undersatnd if the object's backing memory is being written into

> I am against using in-class initialization as proposed in this PR because it will hide true initialization errors (if we do make them in the future). It also uses essentially bogus values for initialization (zero is not a valid value for some of these data members).

Right. I think a similar argument can be made about writing into the object's memory; while I can understand it is more efficient than assigning data members, it feels riskier and makes assumptions about the object's representation. e.g. what would happen if the object were to be made virtual in the future?

While I can't claim I hold a strong opinion on this, it wouldn't hurt if we were intentional in using these patterns

> @eduard-bagdasaryan, do you know of a good way to modify Squid code or Coverity model to help Coverity see that we are initializing all fields when copying into raw()? I bet it is not possible, especially when we have to do two memcpy() operations, but I would be glad to be proven wrong.
> 
> If there is no good way to adjust our code to convince Coverity that all fields are initialized (without in-class initialization), then I think we will be better off with keeping the code as is so that Coverity flags new (true) bugs if we happen to add them.

Sure, the PR makes no sense as is, unless we rework how we initialise objects; I'll close it as soon as this conversation is finished

yadij:
> I am against using in-class initialization as proposed in this PR because

I disagree with your reasons stated for that opinion...

> it will hide true initialization errors (if we do make them in the future).

That is inverted. This style makes the default values visible in-class. We can actively *see* when they are erroneous. Relying on implicit initialization risks leaving values uninitialized or partially initialized - either of which has unpredictable buggy behaviour when used.

> It also uses essentially bogus values for initialization (zero is not a valid value for some of these data members).

Case in point. The values used by this PR mirror the constructor implicit defaults.  The fact that some of them are bogus/invalid has not been detected until you saw them being used as initializers.

eduard-bagdasaryan:
> a good way to modify Squid code or Coverity model 

There is a standard way to suppress such false positives:
```
As with UNINIT, the best way to suppress an UNINIT_CTOR
false positive is to use a code-line annotation to suppress an event.
```
I guess we need to add a code annotation something like `coverity[uninit_member : FALSE]`.

rousskov:
> Eduard: There is a standard way to suppress such false positives:
> 
> ```
> As with UNINIT, the best way to suppress an UNINIT_CTOR
> false positive is to use a code-line annotation to suppress an event.
> ```
> 
> I guess we need to add a code annotation something like `coverity[uninit_member : FALSE]`.

Understood, thank you. I do not think we should add those annotations here: Marking Coverity false positive as such (in Coverity) is a better way, especially when dealing with local (i.e. not spreading to callers) Coverity false positives.

rousskov:
> > Alex: I am against using in-class initialization as proposed in this PR because it will hide true initialization errors (if we do make them in the future). It also uses essentially bogus values for initialization (zero is not a valid value for some of these data members).
 
> Francesco: Right. 

Glad we agree.

> I think a similar argument can be made about writing into the object's memory;

I am not sure how a similar argument would look like, so I cannot comment on that until I see one, but I am happy to discuss improvements to (initializing) UUIDs. Please keep in mind the comment describing RandomUuid fields: They are not what they seem to be. More on that below.

> while I can understand it is more efficient than assigning data members, 

Efficiency is not the reason that code is using memcpy() in constructors. The primary reason is that the object itself represents a container of 122 random bits (128 bits with the exception of six bits in odd places demanded by the RFC). Those memcpy() statements and the surrounding code are initializing those random bits (from sources of randomness or from their serialized representation).

The only sane alternative I know is to completely move away from the RFC-prescribed structure. An earlier (and incomplete) [implementation](https://github.com/measurement-factory/squid/blob/12882c1736bbf8510087232fffc64e2047d6f499/src/base/RandomUuid.cc) of RandomUuid actually did that, but we [reverted](https://github.com/measurement-factory/squid/commit/ae2f9b5) that change because it makes the code even worse -- instead of dealing with fairly simple memcpy() during initialization, we have to deal with even uglier low-level manipulations whenever we need to _access_ those specific fields. You may wonder why we need to access those specific fields if they are, essentially, just random bits. Unfortunately, we need that access to _render_ UUIDs in the RFC-prescribed/admin-expected way _and_ to correctly serialize UUIDs for exchanges between machines with different endianness. We discovered those caveats the hard way and had to rework the code (multiple times).

> what would happen if the object were to be made virtual in the future?

The static assertions protecting these memcpy() operations will fail in that case (because the object size will grow to accommodate the virtual table pointer).

> While I can't claim I hold a strong opinion on this, it wouldn't hurt if we were intentional in using these patterns

Agreed. The current design is intentional. More importantly, we tried alternatives and found them worse. This risky low-level pattern is not something we should blindly propagate to other use cases, of course! Each such memcpy() case requires a context-specific justification.

> > If there is no good way to adjust our code to convince Coverity that all fields are initialized (without in-class initialization), then I think we will be better off with keeping the code as is so that Coverity flags new (true) bugs if we happen to add them.
 
> Sure, the PR makes no sense as is, unless we rework how we initialise objects; I'll close it as soon as this conversation is finished

Thank you. Again, I am happy to discuss improvements. We may have missed a better way to implement this, of course.

rousskov:
> > Alex: I am against using in-class initialization as proposed in this PR because
> 
> Amos: I disagree with your reasons stated for that opinion...

Those reasons were facts. My opinion may still be wrong, of course, but it is based on facts.

> > it will hide true initialization errors (if we do make them in the future).
 
> That is inverted. This style makes the default values visible in-class. We can actively _see_ when they are erroneous. Relying on implicit initialization risks leaving values uninitialized or partially initialized - either of which has unpredictable buggy behaviour when used.

Here, "initialization error" is "field initialized with the wrong value or field not initialized at all". Using in-class initialization will relace all "not initialized" errors with "initialized with the wrong value" errors that are not visible to Coverity. In other words, it will effectively hide initialization errors from Coverity! Coverity will think that the field is properly initialized while it would be initialized improperly (i.e. initialized using the wrong value). This is a fact, not an opinion!

For example, imagine adding another RandomUuid constructor that is identical to the current default constructor except it does not have memcpy() calls (a bug). Today, Coverity will correctly flag that new constructor as buggy.  If we start using in-class initialization, then Coverity will be happy, but the new constructor will still be buggy.

Needless to say, Coverity can still miss some initialization bugs without in-class initializers -- no approach is fool-proof -- but that fact does not invalidate my reasoning against in-class initialization in this case. If somebody can suggest a better approach, we should use that instead.

In-class initialization is a good tool, but, as any tool, it has limitations. This particular class is not suitable for in-class initialization.

> > It also uses essentially bogus values for initialization (zero is not a valid value for some of these data members).

> Case in point. The values used by this PR mirror the constructor implicit defaults. The fact that some of them are bogus/invalid has not been detected until you saw them being used as initializers.

There are no "constructor implicit defaults" in the official code. The official code does not use any "bogus/invalid" values to initialize RandomUuid data members. My complaint was about zero values used exclusively in the PR code. That complaint is not related to any values used in the official code. This PR has not resulted in detecting any problems in the official code (so far).

kinkie:
Let me clarify what I mean.

Let say we have:
class foo {
  public:
  foo();
  int bar;
  float gazonk
  char baz[5];
};

we have several cases where:
foo::foo() {
  memset(this,0,sizeof(*this));
}
and similar patterns for copy-constructors

I wonder whether it would make sense as a matter of policy to mandate that we construct data members explicitly one by one. We risk sacrificing a bit of runtime efficiency in exchange of readability and stronger correctness guarantees

e.g.
foo::foo() : bar(0), gazonk(0.0), baz("") {}
// or even better with in-class initializers

rousskov:
> Francesco: I wonder whether it would make sense as a matter of policy to mandate that we construct data members explicitly one by one.

We should use the best initialization method available for the given context. In most cases, "one by one" is better than memset(*this). There are exceptional situations (like this PR), but a common winner should be individual member initialization (where initialization is needed) rather than memset(). I bet that most memset()s you see come from legacy/C code where there was no other choice...

You may interpret what I said above as supporting your policy suggestion, but I am very worried about words like "matter of policy" and "mandate" because those things have a history of being applied too literally/rigidly in the Squid Project context, resulting in hours (sometimes days) wasted on bad code and/or needless PR discussions. We should not need a special Project policy to know that, in C++ code, the best initialization approach is usually not memset(*this)! It is basic/common knowledge that should not need to codify in a Project policy IMHO.

> We risk sacrificing a bit of runtime efficiency 

I do not think efficiency is a primary concern here (again, by default): I bet that, in most cases, the optimizing compilers are smart enough to optimize all-zero member-by-member initialization so that it is as fast as memset().

> This might even be the key behind the issue in PR#897?

No, that PR is about a different problem -- some of our test initialization code is not compatible with the random test suite execution order. See the last paragraph of that PR description for details.

PR #1068, opened 2022-06-06
Remove reference to deprecated std::binary_function

kinkie:
C++11 deprecates std::binary_function. C++17 removes it.
It only provides typedefs that are unused since C++11.


yadij (approved):

rousskov (approved):
Thank you for removing this deprecated and unused feature. I fixed (and polished) PR description -- according to [cppreference](https://en.cppreference.com/w/cpp/utility/functional/binary_function), C++17 does not deprecate std::binary_function. It _removes_ it. The struct has been deprecated since C++11.

PR #1069, opened 2022-06-06
Remove use of deprecated std::iterator

kinkie:
c++17 deprecates std::iterator.
Explicitly declare traits in our iterator classes instead of
using std::iterator


yadij (approved):
Do these need to be public,  or should they go above in the initially private area?

kinkie:
The examples I've seen make them public, which intuitively makes sense as they are traits. I haven't tried making these private to be honest

PR #1070, opened 2022-06-08
RFC 9218 Priority header registration

yadij:
This allows the Priority header to more efficiently be added and
removed by reverse-proxy administrators using the
reply_header_add, reply_header_access, reply_header_replace
directives.


kinkie (approved):
I'm fine with the change

PR #1073, opened 2022-06-14
Avoid reading tls-cert=bundle twice when loading certificates

rousskov:
After opening and reading the certificate bundle to load the required
traffic-signing certificate, Squid was opening and reading the same file
again to load the optional intermediate certificates. Loading the
traffic-signing certificate twice triggered bugs and further
complicating naturally tricky code.

No functionality changes expected except minor cache.log message changes
and certificate reporting improvements in "squid -k parse" mode.

### DIFF
--- a/src/security/KeyData.cc
+++ b/src/security/KeyData.cc
@@ -15,31 +15,51 @@
 #include "ssl/bio.h"
 #include "ssl/gadgets.h"
 
-/// load a signing certificate from certFile
+/// load the traffic-signing certificate and the chain from certFile
 bool
-Security::KeyData::loadX509CertFromFile()
+Security::KeyData::loadCertificates()
 {
     debugs(83, DBG_IMPORTANT, "Using certificate in " << certFile);
     cert.reset(); // paranoid: ensure cert is unset
 
 #if USE_OPENSSL
-    const char *certFilename = certFile.c_str();
-    Ssl::BIO_Pointer bio(BIO_new(BIO_s_file()));
-    if (!bio || !BIO_read_filename(bio.get(), certFilename)) {
-        const auto x = ERR_get_error();
-        debugs(83, DBG_IMPORTANT, "ERROR: unable to load certificate file '" << certFile << "': " << ErrorString(x));
-        return false;
-    }
+    const auto bundledCerts = Ssl::LoadCertificates(certFile.c_str());
 
-    try {
-        cert = Ssl::ReadCertificate(bio);
-        return true;
-    }
-    catch (...) {
-        // TODO: Convert the rest of this method to throw on errors instead.
-        debugs(83, DBG_IMPORTANT, "ERROR: unable to load certificate file '" << certFile << "':" <<
-               Debug::Extra << "problem: " << CurrentException);
-        return false;
+    // selected bundled certificates in sending order: wireCerts = cert + chain
+    CertList wireCerts;
+    for (const auto &bundledCert: bundledCerts) {
+        assert(bundledCert);
+
+        // the very first bundled certificate is the required traffic-signing cert
+        if (!cert) {
+            cert = bundledCert;
+            assert(wireCerts.empty());
+            wireCerts.emplace_back(bundledCert);
+            debugs(83, DBG_PARSE_NOTE(2), "Using traffic-signing certificate: " << *cert);
+            continue;
+        }
+
+        assert(!wireCerts.empty());
+
+        // We cannot chain any certificate after a self-signed certificate. This
+        // check also protects the IssuedBy() check below from adding duplicated
+        // (i.e. listed multiple times in the bundle) self-signed certificates.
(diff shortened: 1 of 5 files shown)


src/security/KeyData.cc
```diff
@@ -33,7 +34,7 @@ Security::KeyData::loadX509CertFromFile()
 
     try {
         cert = Ssl::ReadCertificate(bio);
-        return true;
+        debugs(83, DBG_PARSE_NOTE(2), "Loaded signing certificate: " << *cert);
```
rousskov, line 37: The `return true` above is gone (to load intermediate certificate) but the function will return true later anyway (where GnuTLS code already returns true).

src/security/KeyData.cc
```diff
@@ -42,6 +43,40 @@ Security::KeyData::loadX509CertFromFile()
...
+            }
+
+            debugs(83, DBG_PARSE_NOTE(2), "WARNING: Ignoring certificate that does not extend the chain: " << *bundledCert);
+        }
+    }
+    catch (...) {
+        // TODO: Reject configs with malformed intermediate certs instead.
+        debugs(83, DBG_IMPORTANT, "ERROR: Failure while loading intermediate certificate(s) from '" << certFile << "':" <<
+               Debug::Extra << "problem: " << CurrentException);
+    }
```
rousskov, line 78: All these try/catch blocks, old and moved, are ugly. They are needed for a "minimal" PR that avoids changing how various certificate errors are handled. They will be gone in subsequent PRs that focus on error handling. Meanwhile, I recommend ignoring them when interpreting PR code.

src/security/KeyData.cc
```diff
@@ -73,6 +108,9 @@ Security::KeyData::loadX509CertFromFile()
         });
     }
 
+    // XXX: implement chain loading
+    debugs(83, 2, "Loading certificate chain from PEM files not implemented in this Squid.");
```
rousskov, line 112: This XXX was moved "as is". There is another PR addressing it IIRC.

src/security/KeyData.cc
```diff
@@ -84,58 +122,6 @@ Security::KeyData::loadX509CertFromFile()
     return bool(cert);
 }
 
-/// load any intermediate certs that form the chain with the loaded signing cert
-void
-Security::KeyData::loadX509ChainFromFile()
```
rousskov, line 89: I could preserve this function (instead of moving/inlining it). Doing so would make the PR diff even smaller, but I would have to give this function a BIO reference as a parameter, and that would make its profile OpenSSL-specific. Also, it would, IMO, make comprehending this code _more_ difficult: The temporary try/catch noise aside, the code became easier to read once all certificate loading code was merged together in side loadCertificates().

Said that, if others insist on restoring this function (with a BIO parameter), I will do that. Just let me know.

src/security/KeyData.cc
```diff
@@ -42,6 +43,40 @@ Security::KeyData::loadX509CertFromFile()
...
+        // selected bundled certificates in sending order: wireCerts = cert + chain
+        CertList wireCerts(1, cert);
+
+        while (const auto bundledCert = Ssl::ReadOptionalCertificate(bio)) {
+            assert(!wireCerts.empty()); // this->cert is there (at least)
+
+            // We cannot chain any certificate after a self-signed certificate. This
+            // check also protects the IssuedBy() check below from adding duplicated
+            // (i.e. listed multiple times in the bundle) self-signed certificates.
+            if (SelfSigned(*wireCerts.back())) {
```
rousskov, line 56: Note that we are still sending self-signed certificates. No changes in that area (yet). We are also still avoiding sending the same self-signed certificate twice. The code got simpler because we no longer have to deal with loading the first certificate twice.

rousskov (commented):
I hope the attached comments help reviewing this functionality-preserving PR. The PR does not address some of the known certificate-loading problems we have discussed recently, but this "minimal" PR is useful on its own, and it paves the way for those improvements.

src/security/KeyData.cc
yadij: IIRC the original idea behind this separation was to allow for loading files defined by other port settings (eg tls-cafile=,  clientca=, or capath= ) without having it locked to the tls-cert= file. But it ended up getting locked anyway with internal use of the certFile member.

yadij (approved):

src/security/KeyData.cc
rousskov: We can always extract this code into a method again if it becomes useful to do so. The current PR code is best for the current use case, and it does not create any serious lock-in AFAICT.

src/security/KeyData.cc
rousskov: > Note that we are still sending self-signed certificates. No changes in that area (yet).

That specific change is now proposed in a dedicated PR: #1173.

PR #1079, opened 2022-07-03
Maintenance: Better xgetaddrinfo() local variable declarations

kinkie:


compat/getaddrinfo.cc
```diff
@@ -105,18 +105,15 @@ xgetaddrinfo (const char *nodename, const char *servname,
     struct servent *servent;
     const char *socktype;
     int port;
-    struct addrinfo hint, result;
-    struct addrinfo *ai, *sai, *eai;
+    struct addrinfo hint = {} , result = {};
+    struct addrinfo *ai = nullptr, *sai = nullptr, *eai = nullptr;
     char **addrs;
```
rousskov, line 110: I do not object to these improvements, but I think that if we modernize this code, we should do more:

* one complete type/name/value declaration statement per line (no cost here and often reduces future diffs)
* move the variable declaration as close to its use as possible/reasonable instead of preserving legacy hoisting to the top of the function (often clarifies code a lot and sometimes prevents bugs)
* cover all variables in the function (initializing `sai` while leaving, say, `servent` pointer uninitialized makes the reader waste time on wondering whether the inconsistency is meaningful)

rousskov (commented):
I polished PR title (to cover all improvements) and emptied PR description. We do not need to (and usually should not) document the low-level details that are clearly visible in the diff... However, I do not insist on any of these metadata changes.

compat/getaddrinfo.cc
rousskov: To avoid misunderstanding: You have the right to limit the scope of this PR to memset() elimination, of course. However, in that case, you should not touch variable declarations at all -- just replace memset(x) with `x = {}` assignments. I do _not_ recommend doing that, but it would at least present a consistent, minimal scope unlike the current strange mix of assorted and seemingly unfinished improvements.

kinkie: Fair enough. Let me modernize all the things; back soon with an update

kinkie:
updated; I've moved all declarations at the innermost useable scope and ensured proper initialization

compat/getaddrinfo.cc
```diff
@@ -136,7 +130,7 @@ xgetaddrinfo (const char *nodename, const char *servname,
         /* Note: maintain port in host byte order to make debugging easier */
         if (isdigit (*servname))
             port = strtol (servname, nullptr, 10);
-        else if ((servent = getservbyname (servname, socktype)))
+        else if ((const struct servent = getservbyname (servname, socktype)))
```
rousskov, line 133: Wrong type and you can actually simplify this further:

```suggestion
        else if (const auto servent = getservbyname(servname, socktype))
```

compat/getaddrinfo.cc
```diff
@@ -190,7 +184,7 @@ xgetaddrinfo (const char *nodename, const char *servname,
     h_errno = 0;
 #endif
     errno = 0;
-    hp = gethostbyname(nodename);
+    const struct hostent hp = gethostbyname(nodename);
```
rousskov, line 187: Wrong type and you can actually simplify this further:

```suggestion
    const auto hp = gethostbyname(nodename);
```

compat/getaddrinfo.cc
```diff
@@ -233,8 +227,10 @@ xgetaddrinfo (const char *nodename, const char *servname,
 
     /* For each element pointed to by hp, create an element in the
        result linked list. */
-    sai = eai = nullptr;
-    for (addrs = hp->h_addr_list; *addrs; addrs++) {
+    struct addrinfo *sai = nullptr;
+    struct addrinfo *eai = nullptr;
+    for (char **addrs = hp->h_addr_list; *addrs; addrs++)
+    {
```
rousskov, line 233: Please simplify if possible (and avoid unusual/new `{` position):

```suggestion
    for (auto addrs = hp->h_addr_list; *addrs; ++addrs) {
```

compat/getaddrinfo.cc
```diff
@@ -262,7 +258,7 @@ xgetaddrinfo (const char *nodename, const char *servname,
         }
 
         result.ai_family = hp->h_addrtype;
-        ai = dup_addrinfo (&result, &sa, addrlen);
+        struct addrinfo *ai = dup_addrinfo(&result, &sa, addrlen);
```
rousskov, line 261: Please simplify and improve const-correctness if possible :

```suggestion
        const auto ai = dup_addrinfo(&result, &sa, addrlen);
```

rousskov (changes requested):
Thank you, but it looks like you are modifying sources that are _not_ compiled in your test environment. Please check that the code compiles. I do not think the modified code can compile in its current state (in environments that need it). The rest is polishing.

compat/getaddrinfo.cc
yadij: Small correction. We should always include the ```*``` qualifier to ensure robust type deduction when the variable is a raw pointer (same for references).

rousskov: > Small correction. We should always include the `*` qualifier 

We do not have a policy about it. IMHO, it is best to omit `*` in code that does not care whether the type is a raw pointer or a smart pointer (or a pointer at all). This change request is attached to an (imperfect) example of such code -- code that does not and should not care about such details. I do not recommend adding `*` to my suggested correction.

In his Almost Always Auto (AAA) [post](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/), Herb Sutter considers asterisk-free auto declarations to be "modern C++ style" and gives many arguments in support of that style in general (while acknowledging that it requires some learning/adjustment/practice). Many disagree, of course. In my experience, most anti-auto (and for-asterisk) arguments are rooted in unrelated problems. Auto just amplifies those problems but gets blamed for them.

When used correctly, an explicit `*` does not change the resulting type (i.e. there is no type-based advantage of using it), but an explicit `*` is not compatible with smart pointers (and it obviously uses more ink). In a sense, it is not an explicit `*` that makes a variable a pointer -- it is how that variable is used (or the set of operations the variable supports).

It is also not trivial to use `*` correctly -- there are many [ways](https://godbolt.org/z/aTP5GKG1e) to add an explicit `*`:

```C++
const auto y1 = x; // a constant pointer; whether it is a pointer to a constant depends on whether x is a pointer to a constant
auto * const y2 = x; // the same thing said using more ink

const auto *y3 = x; // a non-constant pointer to a constant, even if x is not a pointer to a constant

const auto * const y4 = getservbyname(); // a constant pointer to a constant, even if x is not a pointer to a constant
```


> to ensure robust type deduction when the variable is a raw pointer

An explicit `*` does not make type deduction "robust" (for any definition of "robust" that I can think of). In fact, it is kind of the opposite -- an explicit `*` makes type deduction more susceptible to source type changes (i.e. makes the declaration _less_ "robust" in the presence of code changes).

Using an explicit `*` means that the author either needs to add memory constness (y3 and y4 above) or needs to insist that the pointer is a raw pointer. In most use cases, there is no such need.

With all other factors being equal, we should aim for code that does not need to know whether the variable is a _raw_ pointer. That aspect may change and most correctly written code should not be affected by such changes.

> (same for references).

References are a completely different use case! With references, an explicit `&` is _required_ by C++. One simply cannot declare an auto-typed _reference_ without an explicit `&`.

kinkie:
Okay, I'd like to change course on this one: this code hasn't been built in a long time, and I would like to just remove it.

If we look at getaddrinfo.h:69 , where xgetaddrinfo is declared, it contains in its parameters a sctruct addrinfo *; which is never declared in a visible scope for that declaration.

Forcing the precompiler guards to ensure that code is built, results in:
```

In file included from ../compat/compat.h:87,
                 from ../include/squid.h:32,
                 from GnuRegex.c:39:
../compat/getaddrinfo.h:71:45: error: 'struct addrinfo' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
                                const struct addrinfo *hints, struct addrinfo **res);
                                             ^~~~~~~~
../compat/getaddrinfo.h:75:41: error: 'struct addrinfo' declared inside parameter list will not be visible outside of this definition or declaration [-Werror]
 SQUIDCEXTERN void xfreeaddrinfo (struct addrinfo *ai);
                                         ^~~~~~~~
cc1: all warnings being treated as errors
```

So my proposal is to just remove this compat file, it's probably a relic from a long time ago when a proper Internet stack was not present in all OSes.

yadij:
FWIW this code is/was for IPv6 support on native MinGW Windows builds.

rousskov (changes requested):

compat/getaddrinfo.cc
kinkie: I'd follow Alex' idea and let the compiler deal with it.
With the recent move to c++17, we might want to start using unique_ptr and shared_ptr more unless there's any reason why our refcount implementation is better - is there one?

rousskov: To avoid misunderstanding, the discussion below is unrelated to the primary/original concern of this change request: `auto` is a better default than `auto *` _regardless_ of specific smart pointer types in use.

----

> we might want to start using shared_ptr more unless there's any reason why our refcount implementation is better - is there one?

RefCount is an _invasive_ shared pointer. Compared to non-invasive std::shared_ptr, RefCount advantages include:

* The ability to obtain the right shared pointer from a raw object pointer. A lot of current Squid code relies on this feature IIRC.
* No extra memory allocation (to hold an internal "control block" object containing reference counting metadata that is shared by all users of a shared pointer instance) in all use cases. std::shared_ptr prevents that extra memory allocation only when created by calling [std::make_shared](https://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared) or [std::allocate_shared](https://en.cppreference.com/w/cpp/memory/shared_ptr/allocate_shared), and that kind of creation is kind of going against our current Mem::Allocator architecture by allocating objects of different sizes/types than the (often pooled) object returned by dereferencing std::shared_ptr.

In summary, there should be no mechanical conversion to std::shared_ptr. Each conversion case must be carefully evaluated. In some cases, the correct conversion target should not be std::shared_ptr but std::unique_ptr or no pointer at all! If we start using std::shared_ptr, we may need to add proper tools to integrate it with memory pools.

None of the above concerns apply to std::unique_ptr (which has a very different purpose compared to any shared pointer). That pointer should be used as needed.

compat/Makefile.am
```diff
@@ -31,8 +31,6 @@ libcompatsquid_la_SOURCES = \
 	eui64_aton.c \
 	eui64_aton.h \
 	fdsetsize.h \
-	getaddrinfo.cc \
-	getaddrinfo.h \
```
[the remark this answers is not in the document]
rousskov: This change request is still valid AFAICT.

configure.ac
```diff
@@ -3181,7 +3181,8 @@ AC_REPLACE_FUNCS(\
 	tempnam \
 )
 
-AC_CHECK_DECLS([getaddrinfo,getnameinfo,inet_ntop,inet_pton,InetNtopA,InetPtonA],,,[
+# TODO: recheck all these again
+AC_CHECK_DECLS([getnameinfo,inet_ntop,inet_pton,InetNtopA,InetPtonA],,,[
 /*
```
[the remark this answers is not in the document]
rousskov: This change request is still valid AFAICT.

rousskov (changes requested):
@kinkie, you have requested my re-review, but I see no changes or comments related to two of my active change requests. I have touched those change requests in this review to bring them to your attention.

compat/getaddrinfo.h
```diff
@@ -1,85 +0,0 @@
...
-/* These functions are provided by the OS */
-#if !HAVE_DECL_GETADDRINFO
-
-#ifndef EAI_SYSTEM
-/* Not defined on mingw32. */
-#define EAI_SYSTEM     11  /* System error returned in `errno'.  */
-#endif
-#ifndef EAI_OVERFLOW
-/* Not defined on mingw32. */
-#define EAI_OVERFLOW   12 /* Argument buffer overflow.  */
```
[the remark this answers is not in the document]
rousskov: This change request is still valid AFAICT. I missed it when counting change requests in my [review](https://github.com/squid-cache/squid/pull/1079#pullrequestreview-1231683468) a minute ago. There are at least three of them, not two.

yadij:
@kinkie, you requested my review. But with the latest iteration of this I am waiting on confirmation that the mingw builds no longer need this compat code.

IIRC squidclient.exe used to build despite issues in the other code - so it should be usable as a test binary for the necessary proof-of-need.

kinkie:
note to self: test build triggered at https://build.squid-cache.org/job/anybranch-mingw-cross/1/console

kinkie:
Merging a very stale branch messed it up. Closing it, and opening a new one about removing:

getaddrinfo, getnameinfo.h, inet_ntop.h, inet_pton.h, getnameinfo.c, inet_ntop.c, inet_pton.c

rousskov:
> Closing it, and opening a new one

Before you open a new PR, please check that the [outstanding change requests](https://github.com/squid-cache/squid/pull/1079#pullrequestreview-1231683468) for this PR are addressed in the new PR branch.

BTW, you can always start from scratch in your workspace and then force-push your branch to the public PR branch. Force pushing should be avoided, but when the choice is between "closing one PR (with non-trivial review history) and opening a new one" and "force push", force pushing may be the lesser evil.

PR #1080, opened 2022-07-03
Maintenance: Better inet_ntop6() local variable declarations

kinkie:


rousskov (commented):
Pretty much the same metadata polishing and no-objection-but-looks-unfinished/inconsistent review as in https://github.com/squid-cache/squid/pull/1079#pullrequestreview-1026856242

kinkie:
Fair enough. My initial scope was to just get rid of memset as I don't find it fitting to the OO model, but I'm happy to exted scope to modernizing the functions I touch

rousskov (commented):
Thank you for improving this code. To finish the improvements discussed earlier, I would move `tmp` declaration way lower, to the "Format the result" code that uses that buffer, but I do not insist on any changes.

----

> My initial scope was to just get rid of memset

FWIW, the initial PR changes in [5654306](https://github.com/squid-cache/squid/pull/1080/commits/56543067e088070417a9abc07654f5266f538905) went _beyond_ the above scope. My request was based on the scope of those initial changes. I am only saying this because I do not want my request to be perceived as enlarging the perfectly tight scope of the original change...

compat/inet_ntop.cc
```diff
@@ -163,23 +163,22 @@ inet_ntop6(const u_char *src, char *dst, size_t size)
...
      */
-    memset(words, '\0', sizeof words);
-    for (i = 0; i < NS_IN6ADDRSZ; i++)
+    for (int i = 0; i < NS_IN6ADDRSZ; ++i)
         words[i / 2] |= (src[i] << ((1 - (i % 2)) << 3));
-    best.base = -1;
-    best.len = 0;
-    cur.base = -1;
-    cur.len = 0;
     for (i = 0; i < (NS_IN6ADDRSZ / NS_INT16SZ); i++) {
```
yadij, line 182: After moving the earlier function-scope declaration into previous for-loop scope this ```i``` is undefined.

Same for other loop below.

yadij (changes requested):

compat/inet_ntop.cc
rousskov: To be more precise, `i` is undeclared (not undefined). More `int`s are needed for those loops.

More importantly, we need to at least compile this code before merging this PR -- same problem as in https://github.com/squid-cache/squid/pull/1079#pullrequestreview-1026897488 :-(

kinkie: > More importantly, we need to at least compile this code before merging this PR -- same problem as in [#1079 (review)](https://github.com/squid-cache/squid/pull/1079#pullrequestreview-1026897488) :-(

Right :( I suppose that systems lacking this function are quite rare, especially among the ones we support.
Might be worthwhile to actually run a test to see what we could drop from compat/

Rechecking both PRs and ensuring they can build. Apologies for the noise.

kinkie: tested with CPPFLAGS=-D_ARPA_INET_H_ to prevent the system header from being included. Now passes build. updating pr

kinkie: Confirmed tested and working. resolving

compat/inet_ntop.cc
```diff
@@ -204,8 +203,8 @@ inet_ntop6(const u_char *src, char *dst, size_t size)
     /*
      * Format the result.
      */
-    tp = tmp;
```
rousskov, line 207: You have (correctly) moved `tp` declaration to this formatting code where `tp` is used. Why not move the `tmp` declaration as well? The two are closely related and used in the same code.

compat/inet_ntop.cc
```diff
@@ -163,24 +163,23 @@ inet_ntop6(const u_char *src, char *dst, size_t size)
      * Keep this in mind if you think this function should have been coded
      * to use pointer overlays.  All the world's not a VAX.
      */
-    char tmp[sizeof "ffff:ffff:ffff:ffff:ffff:ffff:255.255.255.255"], *tp;
```
rousskov, line 166: The same string literal is used below in the code (near line 228), and the two uses are dangerously tied to each other. Consider replacing that second use with `tmp`:

```diff
-        tp += snprintf(tp, (tmp + sizeof("ffff:ffff:ffff:ffff:ffff:ffff:255.255.255.255") - tp), "%x", words[i]);
+        tp += snprintf(tp, (tmp + sizeof(tmp) - tp), "%x", words[i]);
```

rousskov (commented):
Just a couple of comments in hope to help you make progress. I still do not insist on any changes here.

compat/inet_ntop.cc
kinkie: Sure; it won't change scope end as it's used at the very end of the function but this won't hurt.
I'm also renaming it as "buf" because in the end it's a buffer.

compat/inet_ntop.cc
kinkie: I've factored out to a constexpr size variable. It may not be minimal but I think it better expresses intent

compat/inet_ntop.cc
```diff
@@ -204,8 +200,11 @@ inet_ntop6(const u_char *src, char *dst, size_t size)
     /*
      * Format the result.
      */
-    tp = tmp;
-    for (i = 0; i < (NS_IN6ADDRSZ / NS_INT16SZ); i++) {
+    constexpr int maxbuflen =
```
rousskov, line 203: Overall, I am against the unnecessary (here) and dangerous (always) disconnect between `buf` size and expressions that need to know `buf` size. In other words, here, `sizeof(buf)` is safer than a dedicated named constant that is supposed to contain `buf` size. See below for specific suggestions if you decide to keep this constant.

* We should be using `size_t` for object sizes.
* We should be using auto when possible.
* We should be using camelCase.
* This is not a _maximum_ buffer size. It is (constant) buffer size.
* If you add a `buf` size constant, then please use it _everywhere_ instead of `sizeof(buf)`, for consistency sake. Using two ways to refer to the same buffer size increases dangers where parts of this code go out of sync with each other.
* Consexpr is a distracting overkill here IMO. This function cannot be used at compile time, of course.  We should only be using `constexpr` in special cases where we focus on or at least foresee compile-time usage, not everywhere it is usable. FWIW, some folks [share](https://stackoverflow.com/a/6785367) this opinion.

```suggestion
    const auto bufSize =
```

You can use the following test to play with different variations of this declaration. For example, I find the effect of removing `const` interesting: https://godbolt.org/z/Kqjj1xssa

I do not insist on any changes.

kinkie: I'm very okay with using const auto in place of constexpr int and renaming the variable; changing.

I think in this case using temporary variable conveys intent better than using sizeof() twice; in the other case (inet_ntop) the code is much simpler and cleaner, I'd leave as-is there.

I'm not sure I get the philosophical implications constexpr versus const and when to use it, but they are out of scope for this change, where the compiler will very likely optimize all away anyway.

Thanks for your feedback

rousskov: > I'm not sure I get the philosophical implications constexpr versus const and when to use it

I am writing this comment in hope to avoid similar discussions in other PRs: I recommend not using `constexpr` until you know when to use it. If you want to discuss and polish more specific and more general recommendations on `constexpr` use (for future use), I am happy to participate (here or elsewhere)! FWIW, my suggestion for the starting point is "use `constexpr` only where its use is already necessary in the current code or where it is likely to become necessary in future code".  I expect this rule of thumb to result in `constexpr` use in special cases where we focus on or at least foresee compile-time usage of a function or variable with compiler optimizations turned _off_. I suspect `constexpr` is not really focused on performance optimization. It is more about writing clearer code (that also happens to work at compile time).

> philosophical implications constexpr versus const and when to use it ... are out of scope for this change

I agree that they are out of scope as long as the code does not use (and does not have to use) `constexpr`. Thank you for removing it from this PR!

kinkie:
@yadij kind reminder about the pending re-review

compat/inet_ntop.cc
```diff
@@ -150,37 +150,33 @@ inet_ntop4(const u_char *src, char *dst, size_t size)
...
  */
 static const char *
 inet_ntop6(const u_char *src, char *dst, size_t size)
 {
     /*
      * Note that int32_t and int16_t need only be "at least" large enough
-     * to contain a value of the specified size.  On some systems, like
-     * Crays, there is no such thing as an integer variable with 16 bits.
+     * to contain a value of the specified size.  On some systems
+     * there is no such thing as an integer variable with 16 bits.
```
yadij, line 162: IMO, drop this change. The text is still correct for this code's implementation design choices even if not relevant to Squid in general.

compat/inet_ntop.cc
```diff
@@ -150,37 +150,33 @@ inet_ntop4(const u_char *src, char *dst, size_t size)
...
      */
-    char tmp[sizeof "ffff:ffff:ffff:ffff:ffff:ffff:255.255.255.255"], *tp;
-    struct { int base, len; } best, cur;
-    u_int words[NS_IN6ADDRSZ / NS_INT16SZ];
-    int i;
+    struct  {
+        int base = -1;
+        int len = 0;
+    } best, cur;
+    u_int words[NS_IN6ADDRSZ / NS_INT16SZ] = {};
```
yadij, line 170: Since you have scoped this as better definitions. This should now be:
```suggestion
    unsigned int words[NS_IN6ADDRSZ / NS_INT16SZ] = {};
```

yadij (changes requested):

compat/inet_ntop.cc
rousskov: The comment is wrong or extremely misleading (in its current context), but I agree that just removing Crays does more harm than good and support the recommendation to keep the official comment "as is" in this PR.

kinkie: reverting

kinkie:
@yadij if you are satisied with the current state, feel free to clear for merge, no need to wait for me

yadij (approved):

PR #1081, opened 2022-07-04
Maintenance: Improve stability of generated automake Makefiles

rousskov:
Our source-maintenance.sh script generates five automake Makefiles using
"git ls-files <pattern>" output as the source of filename entries. These
entries were piped through "sort -u". That problematic step is now gone:

* "sort" may not produce stable results across different locales.

* "sort" does not use the order enforced by our format-makefile-am.pl.
  That Perl script itself did not fix the order because it only ordered
  entries in *_SOURCES groups that these generated files do not use.

* "sort -u" could silently drop duplicate Makefile entries (that should
  not exist) instead of letting that Perl script warn us about them.

The first two bullets made automake Makefile generation unstable (e.g.,
502bcc7 reversed the ASCII order of en.lang and en_AU.lang entries, but
current runs in unofficial environments could reorder those entries).

format-makefile-am.pl now recognizes the generated group name patterns,
properly formatting entries in those five generated automake Makefiles.

That fix alone created persistent misleading "File ... changed: by ...
format-makefile-am.pl" NOTICEs because the Perl script was (always)
changing the _generated_ file. We now pipeline the generation and
formatting steps, applying NOTICE-generation check to the final result.

(Positive) side effect: The generated automake Makefiles now include a
notice that they are generated.


scripts/format-makefile-am.pl
```diff
@@ -19,7 +19,9 @@
     }
     # accumulate files and prep for sorting
     my %files = ();
-    if (/^(\S+_SOURCES)\s*(\+?=)\s*(.*[^\\])$/ ) {
+    # TODO: Handle or rename /\S+SOURCE=/ and /\S*[^_]SOURCES=/
```
rousskov, line 22: This TODO deserves a dedicated PR. It is not directly related to the problem solved in this PR, it requires some non-trivial work to identify all applicable group names, and it will trigger a lot of Makefile.am changes.

doc/manuals/language.am
```diff
@@ -6,16 +6,18 @@
...
-    oc.lang \
-    pt.lang \
-    ro.lang \
-    ru.lang
+	ar.lang \
+	cs.lang \
+	de.lang \
+	en.lang \
+	en_AU.lang \
+	es.lang \
```
rousskov, line 17: Lots of whitespace changes in Makefiles because generated files were not using tabs. To see essential changes on GitHub, tell GitHub to hide whitespace changes.
Screenshot from 2022-07-03 23-43-37

rousskov (commented):
There are other sources of unstable source-maintenance.sh output. YMMV, but I am wasting a little bit of time nearly every time I run the script for a PR because my results differ from the official ones in various places due to those stability problems. This situation also makes it more difficult for me to encourage others to run the script for their PRs. If I get a chance, I will post more fixes.

scripts/format-makefile-am.pl
yadij: FTR; The *_SOURCES tag is a reserved keyword by automake. It has logic side effects during build.

yadij (approved):
Looks reasonable for QoL fixes.

scripts/source-maintenance.sh
```diff
@@ -108,6 +108,9 @@ if ! git diff --quiet; then
 	exit 1
 fi
 
+made="generated" # a hack: prevents $GeneratedByMe searches matching this file
+GeneratedByMe="This file is $made by scripts/source-maintenance.sh."
```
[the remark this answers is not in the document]
rousskov: Many other such disclaimers do not mention what will happen to edits and, more importantly, I failed to find a good way to describe what will happen to edits. All my attempts resulted in statements that were either useless because they were obvious or technically inaccurate because this file is not regenerated in most common circumstances (so editing it in some special cases will "work" and might even be useful).

rousskov:
> Looks reasonable for QoL fixes.

What do you mean be "QoL" in this context? The only matching abbreviation I could find -- "Quality of Life" -- does not seem to be particularly relevant to fixes in this PR...

PR #1083, opened 2022-07-09
Make Acl::Answer constructor explicit

kinkie:


src/acl/Acl.h
```diff
@@ -121,8 +121,7 @@ namespace Acl {
 class Answer
 {
 public:
-    // not explicit: allow "aclMatchCode to Acl::Answer" conversions (for now)
-    Answer(const aclMatchCode aCode, int aKind = 0): code(aCode), kind(aKind) {}
+    explicit Answer(const aclMatchCode aCode, int aKind = 0): code(aCode), kind(aKind) {}
```
rousskov, line 124: The need to make caller code look worse in so many files is not obvious to me. What is the motivation behind this PR?

If the goal is improving overall code quality (by removing this implicit conversion), then I recommend closing this PR because the proposed changes is not how we should get rid of that implicit conversion, and the correct solution can wait until we solve more important/pressing problems.

FWIW, the "(for now)" part of the official comment does not imply that explicitly surrounding existing `ACCESS_*` constants with this constructor name is the right step forward. Such a mechanical change is _not_ what stopped us from making the constructor explicit. There are other ways to get rid of this implicit conversion, but they require non-mechanical API changes.

rousskov (changes requested):

kinkie:
I'll do that, but could you please change that comment to clarify intent?
Thanks

PR #1085, opened 2022-07-10
Fix mingw-cross build

kinkie:
cast from 'void*' HANDLE to long int loses precision [-fpermissive]


kinkie:
note: mswindows.h changes are from the mingw-cross-1 branch . I'm not entirely sure
how to stack PR on top of each other so that github only shows the current PR changes

compat/mswindows.cc
```diff
@@ -189,9 +189,10 @@ getgrnam(char *unused) {
 int
 _free_osfhnd(int filehandle)
 {
-    if (((unsigned) filehandle < SQUID_MAXFD) &&
-            (_osfile(filehandle) & FOPEN) &&
-            (_osfhnd(filehandle) != (long) INVALID_HANDLE_VALUE)) {
+    if (((unsigned)filehandle < SQUID_MAXFD) &&
```
yadij, line 192: Nit: please take advantage of the change to remove the initial c-style cast here.

yadij (dismissed):

compat/mswindows.cc
```diff
@@ -203,7 +204,7 @@ _free_osfhnd(int filehandle)
             SetStdHandle(STD_ERROR_HANDLE, nullptr);
             break;
         }
-        _osfhnd(filehandle) = (long) INVALID_HANDLE_VALUE;
+        _osfhnd(filehandle) = reinterpret_cast<intptr_t>(INVALID_HANDLE_VALUE);
```
rousskov, line 207: This cast does not make sense to me (in Squid context) because Squid defines _osfhnd() to return a `long` data member reference:

```C++
/* internal to Microsoft CRTLIB */
typedef struct {
    long osfhnd;    /* underlying OS file HANDLE */
...
}   ioinfo;
```

The above struct is used by the _osfhnd() macro used by the PR-changed code:

```C++
#define _osfhnd(i)  ( _pioinfo(i)->osfhnd )
#define _pioinfo(i) ( __pioinfo[...] + ... )
SQUIDCEXTERN _CRTIMP ioinfo * __pioinfo[];
```

AFAICT, [some](https://doxygen.reactos.org/d7/d45/crtdefs_8h.html#a0fbe4a4f8dd857ee04923a901f27465f) code defines that `osfhnd` member as `intptr_t` so I am guessing it is the modern way to define that data member, but the above PR change is incompatible with our own `ioinfo` definition...

I do not know whether our own definition should be removed (e.g., as no longer unused) or this comparison code should be made smarter about the actual type of the `osfhnd` data member (to accommodate cases where our own definition _is_ used).

rousskov (commented):
I do not insist on any changes because some of my change requests are not worth fighting for, and I lack the expertise to be certain about the others (but I can point out the surface-level problems that you may need to be aware of).

rousskov:
> note: mswindows.h changes are from the mingw-cross-1 branch . I'm not entirely sure how to stack PR on top of each other so that github only shows the current PR changes

AFAIK, GitHub does not support such "stacked" PRs. GitHub PRs are based on git branches. GitHub computes PR branch forking point against PR target branch (i.e. the master branch in this case). The existence of other PRs/branches/etc. is absolutely irrelevant in this computation.

In general, reasonable choices include:

1. Create a dedicated branch for PR B, branching off the future target branch (i.e. the master branch in this case).
2. Wait for the PR A to merge before (creating the next PR B branch and) posting PR B.
3. Post a stacked PR B immediately but mark it as a Draft PR and explain that it contains PR A changes. The `S-waiting-for-PR` PR label may be useful in such cases! Be ready to forgive reviewers if they cannot review the resulting mess (correctly). After PR A merges, rebase the PR B branch, force push, and remove the Draft designation from PR B. It may also be possible to merge the target branch into PR B to avoid force-pushing.

All this requires a bit of extra work/pain. The extra cost increases PR minimization price, of course. In most cases, smaller dedicated PRs are still the lesser evil.

compat/mswindows.cc
```diff
@@ -189,9 +189,10 @@ getgrnam(char *unused) {
 int
 _free_osfhnd(int filehandle)
 {
-    if (((unsigned) filehandle < SQUID_MAXFD) &&
-            (_osfile(filehandle) & FOPEN) &&
-            (_osfhnd(filehandle) != (long) INVALID_HANDLE_VALUE)) {
```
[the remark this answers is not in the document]
kinkie: The error message is

`
In file included from /usr/share/mingw-w64/include/winbase.h:19,
                 from /usr/share/mingw-w64/include/windows.h:70,
                 from /usr/share/mingw-w64/include/winsock2.h:23,
                 from ../compat/os/mswindows.h:283,
                 from ../compat/compat.h:71,
                 from ../include/squid.h:32,
                 from mswindows.cc:12:
mswindows.cc: In function 'int _free_osfhnd(int)':
mswindows.cc:207:38: error: cast from 'HANDLE' {aka 'void*'} to 'long int' loses precision [-fpermissive]
  207 |         _osfhnd(filehandle) = (long) INVALID_HANDLE_VALUE;
      |                                      ^~~~~~~~~~~~~~~~~~~~
`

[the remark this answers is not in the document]
kinkie: note: INVALID_HANDLE_VALUE is -1 ; we need a signed comparison here.
I think the main reason is that on mswindows, sizeof(long) != sizeof(void *)

compat/mswindows.cc
kinkie: We don't really have a consistent policy when it comes to out-of-scope trivial changes. I propose that we leave that to the discretion of the PR author. It might be inconsistent, but it allows everyone to focus on the things that matter the most

compat/mswindows.cc
kinkie: What I think Squid is doing in that context is to define a copy of the Windows internals to keep the compiler happy in some contexts.

Here is what the preprocessed code looks like:

```
    if (((unsigned)filehandle < 2048) &&
        (( ( __pioinfo[(filehandle) >> 5] + ((filehandle) & ((1 << 5) - 1)) )->osfile ) & 0x01) &&
        (( ( __pioinfo[(filehandle) >> 5] + ((filehandle) & ((1 << 5) - 1)) )->osfhnd ) != reinterpret_cast<intptr_t>( ((HANDLE) (LONG_PTR)-1) )))
```
http://www.shrutisagarashram.org/ashram/VC/VC98/CRT/SRC/OSFINFO.C looks a lot like the code we are using; maybe it served as inspiration

I suspect we will want to revisit Windows IO, but we need to get to compile first

compat/mswindows.cc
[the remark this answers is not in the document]
rousskov: > INVALID_HANDLE_VALUE is -1

What makes you think that? At least in some variants, `INVALID_HANDLE_VALUE` is a `HANDLE`, and Windows [`HANDLE`](https://en.wikibooks.org/wiki/Windows_Programming/Handles_and_Data_Types#HANDLE) is a pointer, not an integer:

```C
typedef void *PVOID;
typedef PVOID HANDLE;
#define INVALID_HANDLE_VALUE ((HANDLE)(ULONG_PTR)-1)
```

AFAICT, the reviewed PR code is casting `INVALID_HANDLE_VALUE` to a pointer, not to an integer (intptr_t).

Yes, on a 32-bit system, the long integer bits may look the same as pointer bits, but that does not make `INVALID_HANDLE_VALUE` -1.

> we need a signed comparison here.

Given the above `INVALID_HANDLE_VALUE` definition and reasoning, we do not.

compat/mswindows.cc
rousskov: > We don't really have a consistent policy when it comes to out-of-scope trivial changes. 

I agree, but I am not sure there is consensus about that.

> I propose that we leave that to the discretion of the PR author.

That is essentially my current approach, with the following caveats:

* Any modified code may still be reviewed/discussed. Reviewers have their discretion as well! In most cases, the reviewer should either fix trivial problems inherited by the modified code from its official version (see the next bullet) or leave them be without creating a change request, but it is impossible to hard-code the best decision logic for every possible situation.
* Once the author decides to touch a line of code, the reviewer may (but does not have to) polish that line of code further if such polishing makes sense (because the line is already a part of the diff). Most of such changes should be limited to formatting and other "mechanical" issues, but it is impossible to hard-code what makes sense for every possible situation.

For example, in this particular case, if Amos did not ask you to start modifying these out-of-scope lines (in ways going beyond trivial formatting changes), then I probably would not suggest undoing these changes (to remove the temptation to modify these lines), but I could have reverted these formatting changes myself (to minimize the diff and to clarify what this PR is actually fixing).

compat/mswindows.cc
rousskov: Your comment pretty much matches my understanding and the reasons behind my change request. Yes, we are defining what some Windows flavors are probably using internally, but this PR changes are not compatible with our (old) definitions. Secondary problems aside, on the surface, it looks like this PR fixes one use case by breaking another. We should not do that, especially when that "other use case" is based on our own code (that defines `ioinfo`).

> Here is what the preprocessed code looks like: ...

Is that based on Squid-provided `ioinfo` definition where `osfhnd` is `long`? Or is it based on Windows-provided `ioinfo` definition? If it is the latter, what is `osfhnd` type? As I have mentioned in the change request, I have seen that type being a pointer (rather than `long`) in some environments, which could explain why your changes fix the build (in those environments), but there are other explanations as well.

Your should be able to answer the first question by, for example, commenting out our `ioinfo` definition or renaming its  `osfhnd` field (and adjusting the corresponding accessor macro). If the preprocessed code does not change, then you are not using our definition and, hence, is not reproducing the PR problem that will surface in environments that do use that definition.

compat/mswindows.cc
yadij: > > We don't really have a consistent policy when it comes to out-of-scope trivial changes.
> 
> I agree, but I am not sure there is consensus about that.
> 

We do have a set of style changes that have been agreed to be done in all PRs on touched lines. Updating cast style is one of those.

*If* the lines containing C-style casts remain touched by this PR they should be updated to C++ casts. There should not be any problem/bugs **added* by a correct cast upgrade.


> > I propose that we leave that to the discretion of the PR author.
> 

Notice that is precisely what my OP request did. This style detail was marked as a "nit" (to be fixed at authors discretion) and PR approved regardless.

> That is essentially my current approach, with the following caveats:
> 
>     * Any modified code may still be reviewed/discussed. Reviewers have their discretion as well! In most cases, the reviewer should either fix trivial problems inherited by the modified code from its official version (see the next bullet) or leave them be without creating a change request, but it is impossible to hard-code the best decision logic for every possible situation.
> 

Why do you place this on the reviewers shoulders to perform the fix?
The whole point of style change requests is to make submissions easy for reviewers to a) read and b) identify subtle bugs hiding in the C/C++ syntax layout. Accepting arbitrarily styled code just because the author likes their own coding style better is a bad idea.

>     * Once the author decides to touch a line of code, the reviewer may (but does not have to) polish that line of code further if such polishing makes sense (because the line is already a part of the diff).

Again with placing work on reviewer shoulders. We already have agreement that some changes (eg these cast upgrades) be done in each PR touched line.

> Most of such changes should be limited to formatting and other "mechanical" issues, but it is impossible to hard-code what makes sense for every possible situation.
> 

That "but" being the reason things like this are pointed out as review items, not fixed automagically by the maintenance scripts. It is not a reason to avoid mentioning them in reviews.

> 
> For example, in this particular case, if Amos did not ask you to start modifying these out-of-scope lines (in ways going beyond trivial formatting changes), then 

... then I would be inconsistently applying the agreed policy of touched lines being updated to meet current coding/style guidelines.

rousskov: > > > > Amos: please take advantage of the change to remove the initial c-style cast here.

> > > > Alex: I disagree that this line should be changed (at all).

> > > Francesco: We don't really have a consistent policy when it comes to out-of-scope trivial changes.

> > Alex: I agree, but I am not sure there is consensus about that.
 
> Amos: We do have a set of style changes that have been agreed to be done in all PRs on touched lines. 

I was discussing which lines should (not) be touched in the first place.  I was not discussing style changes on touched lines, but, apparently, we do not have a clear agreement on that either! This exchange appears to reconfirm Francesco's and my opinions regarding "consistent policy".

> Updating cast style is one of those.

Is it? Is there a list of those somewhere? Neither [MergeProcedure](https://wiki.squid-cache.org/MergeProcedure) nor [SquidCodingGuidelines](https://wiki.squid-cache.org/SquidCodingGuidelines) mention the word "cast" AFAICT.

As this very PR illustrates, in many cases, correctly changing casting is far from trivial, may have runtime consequences, and, hence, should not be treated the same way as "mechanical" whitespace formatting, HERE removal, or NULL changes.


> _If_ the lines containing C-style casts remain touched by this PR they should be updated to C++ casts. There should not be any problem/bugs *_added_ by a correct cast upgrade.

@kinkie, that "add no problems" condition is often difficult to satisfy -- there is no single C++ cast that can correctly replace a C-style cast in all cases; one has to pick carefully. Moreover, upgrading a wrong C cast to a wrong C++ cast is a step backwards (i.e. it adds a problem). Thus, I continue to recommend closing this Pandora box by undoing these whitespace changes instead of polishing those lines further.

> > > I propose that we leave that to the discretion of the PR author.
 
> Notice that is precisely what my OP request did. This style detail was marked as a "nit" (to be fixed at authors discretion) and PR approved regardless.

Just to avoid misunderstanding: All of the change requests in this PR are optional. Nobody has insisted on anything yet.

> > That is essentially my current approach, with the following caveats:
> > * Any modified code may still be reviewed/discussed. Reviewers have their discretion as well! In most cases, the reviewer should either fix trivial problems inherited by the modified code from its official version (see the next bullet) or leave them be without creating a change request, but it is impossible to hard-code the best decision logic for every possible situation.
 
> Why do you place this on the reviewers shoulders to perform the fix?

I would not phrase it that way -- the reviewer has an option to do nothing at all. However, if the reviewer wants a trivial fix, then it is often better (lower cost, fewer tensions, fewer out-of-scope arguments, etc.) for the qualified reviewer to apply that trivial fix without a change request rather than filing a change request, waiting for the author to address it (or ask questions about or, or argue with it!), and then validate that the author understood and applied the change request correctly and without triggering more problems. All this overhead is usually completely unnecessary. It is often much better to accomplish the same improvement (i.e. the same code state) by implementing it.

> The whole point of style change requests is to make submissions easy for reviewers to a) read and b) identify subtle bugs hiding in the C/C++ syntax layout.

I doubt that is the whole or even primary point, but we do not need to agree on that. The strategy I recommend does not contradict the statement quoted above.

> Accepting arbitrarily styled code just because the author likes their own coding style better is a bad idea.

This feels like an unrelated point/discussion, but whether that is a bad idea depends on specifics. Until we auto-format or regulate everything, there will always be gray areas where we have to either accept author's choice or argue using such usually-vague concepts as consistency, symmetry, tradition, etc.


> > * Once the author decides to touch a line of code, the reviewer may (but does not have to) polish that line of code further if such polishing makes sense (because the line is already a part of the diff).
 
> Again with placing work on reviewer shoulders. 

Again, I would not phrase it that way -- nobody is forcing the reviewer to do anything here. BTW, writing a high-quality change request is also "work". YMMV, but in my experience, that kind of work is often much harder (and costly) than applying trivial code changes, especially in Squid Project context.
 
> > Most of such changes should be limited to formatting and other "mechanical" issues, but it is impossible to hard-code what makes sense for every possible situation.

> That "but" being the reason things like this are pointed out as review items, not fixed automagically by the maintenance scripts. It is not a reason to avoid mentioning them in reviews.

Agreed -- it is not the reason. There are other/different reasons (as detailed above).
 
> > For example, in this particular case, if Amos did not ask you to start modifying these out-of-scope lines (in ways going beyond trivial formatting changes), then
 
> ... then I would be inconsistently applying the agreed policy of touched lines being updated to meet current coding/style guidelines.

Not if you undid the formatting-only changes instead (for example). However, as we have established, there is no consensus (or sufficient knowledge) about the details of that allegedly "agreed policy", so we should not be wasting time discussing what that policy does or does not imply in this case. We should (re)establish consensus regarding that policy first.

compat/os/mswindows.h
```diff
@@ -404,7 +404,7 @@ typedef char * caddr_t;
 
 /* internal to Microsoft CRTLIB */
 typedef struct {
-    long osfhnd;    /* underlying OS file HANDLE */
+    HANDLE osfhnd;    /* underlying OS file HANDLE */
```
rousskov, line 407: For the code to work, our `osfhnd` type must match whatever type the actual `__pioinfo` array is using for that data member. FWIW, I cannot find authoritative information on that internal-to-Windows array and have no idea whether all Windows versions use `HANDLE` for the first data member of that array elements.

We should test this (which means not just compiling but _running_ this code and making sure osfhnd values in Windows memory make sense). If we cannot test this, then I doubt we should be fixing this problem because we would be just guessing a rather sensitive/critical type. We can easily make things worse. Converting this PR into a Draft PR and waiting for somebody who can help may be an overall better strategy if we cannot test.

rousskov (commented):
I committed a few adjustments under the EasierDone experiment.

I see no internal inconsistencies in the current PR diff, and I do not object to merging this PR if others think it is a good idea to do so. However, I cannot validate whether the changed code matches what Windows wants/expects in all (or any) of the Windows environments this code is targeting. I am concerned about the ioinfo::osfhnd type change as detailed in the change request.

compat/mswindows.cc
rousskov: @yadij, I believe the current code addresses this change request.

kinkie:
My current plan with this stack:
1. get it to compile using mingw cross
2. run it, fix any issues
3. try porting to a native windows stack. I have a VM almost ready to do it and keep it sustainable

kinkie:
> > note: mswindows.h changes are from the mingw-cross-1 branch . I'm not entirely sure how to stack PR on top of each other so that github only shows the current PR changes
> 
> AFAIK, GitHub does not support such "stacked" PRs. 

[...]
Pity; for $dayjob we use phabricator and it supports PR stacks, it's quite convenient.

> All this requires a bit of extra work/pain. The extra cost increases PR minimization price, of course. In most cases, smaller dedicated PRs are still the lesser evil.

I agree.
So, this current PR is targeted to fix compat/mswindows.*  and it should be almost ready. Squid doesn't build yet due to problems in sspwin32.cc,  squidclient.cc, Transport.cc, cachemgr.cc, ssl .

I'll tackle each of those with a dedicated PR, waiting for the previous one to be merged before moving on.

rousskov:
> we use phabricator and it supports PR stacks, it's quite convenient.

To avoid misunderstanding: GitHub supports a flavor of stacked PRs for _same-repository_ branches. We use that feature internally quite often! However, here, we are dealing with a public Project repository, and GitHub (somewhat naturally) does not let you create PRs against branches that are _not_ in the repository.

> I'll tackle each of those with a dedicated PR, waiting for the previous one to be merged before moving on.

Good luck with making progress in this messy code!

Please note that it is not clear what the current `S-waiting-for-reviewer` label meant to indicate because there are two active reviewers and no reviews have been requested. To request review(s), use the Reviewers panel in the top right corner.

kinkie:
This work is interim but it doesn't regress the general status and is a step forward to the intended goal.
I'd like to merge it to checkpoint and start anew.
The XXX @rousskov  marked in the last commit will require us to have an inline wrapper for ::bind to use throughout the code; the current preprocessor-based approach breaks std::bind used in splay.cc:135 
But I can't imagine a way to redefine ::bind that doesn't involve the preprocessor

kinkie:
Re: merging: @yadij  already approved so in theory it could be merged now; since there have been changed since I'd like to have @rousskov 's opinion

rousskov:
> since there have been changed since I'd like to have @rousskov 's opinion

My [opinion](https://github.com/squid-cache/squid/pull/1085#pullrequestreview-1048568272) have not changed.

rousskov:
> This work ... doesn't regress the general status and is a step forward to the intended goal.

I hope that is true, but I cannot independently validate that claim.

>  The XXX ... marked in the last commit will require us to have a ... wrapper for ::bind to use throughout the code; the current preprocessor-based approach breaks std::bind used in splay.cc:135.  But I can't imagine a way to redefine ::bind that doesn't involve the preprocessor

What exactly prevents a Windows-specific wrapper from exporting a ::bind() function (rather than a macro)? The comments say "Each of these functions is defined in the Squid namespace so as not to clash with the winsock.h and winsock2.h definitions". AFAICT, those definitions use `HANDLE` as the type of the first argument. The definitions we add use `int`. Perhaps we can limit support to Windows flavors where `HANDLE` is not an integer and use polymorphism to provide our own versions of those functions where the first argument is an `int`?

kinkie:
> > This work ... doesn't regress the general status and is a step forward to the intended goal.
> 
> I hope that is true, but I cannot independently validate that claim.
> 
> > The XXX ... marked in the last commit will require us to have a ... wrapper for ::bind to use throughout the code; the current preprocessor-based approach breaks std::bind used in splay.cc:135.  But I can't imagine a way to redefine ::bind that doesn't involve the preprocessor
> 
> What exactly prevents our Windows-specific wrapper from exporting a ::bind() function (rather than a macro)? The comments say "Each of these functions is defined in the Squid namespace so as not to clash with the winsock.h and winsock2.h definitions". AFAICT, those definitions use `HANDLE` as the type of the first argument. The definitions we add use `int`. Perhaps we can limit support to Windows flavors where `HANDLE` is not an integer and use polymorphism to provide our own versions of those functions where the first argument is an `int`?

the callsites use bind(int..). The macro in compat/os/mswindows.h is for them to maintain the illusion that this is what the OS provides. That macro then calls Squid::bind(int..) which does the HANDLE lookup, calls ::bind(HANDLE), and adapts the error back.

Since the macro bind breaks std::bind, I am afraid that the next step is to define a Squid::bind which in the non-mswindows case just forwards the call to ::bind, and adapt callsites to use it.
It might make sense to wrap all the network API calls in that case for cleanliness, and stop using the preprocessor for this compatibility shim

rousskov:
> the callsites use bind(int..). The macro in compat/os/mswindows.h is for them to maintain the illusion that this is what the OS provides. That macro then calls Squid::bind(int..) which does the HANDLE lookup, calls ::bind(HANDLE), and adapts the error back.

Please (re)interpret what I said, including my question, with the assumption that I know all of the above facts.

rousskov:
In other words, what will break if we remove the macros _and_ remove the Squid namespace?

kinkie:
> In other words, what will break if we remove our macros _and_ remove our Squid namespace?

it builds. It might be that mingw is doing the emulation job for us? I can try to remove most of mswindows.h if that is the case and see what happens

rousskov:
> It might be that mingw is doing the emulation job for us? I can try to remove most of mswindows.h if that is the case and see what happens

I do not know the answer to the above question. I can only keep whining that we should _understand_ what happens (instead of using trial-and-error approach to find the first magical combination that appears to work in a given black-box test). When it comes to Windows code, I can spot high-level problems and make suggestions based on basic principles, but I cannot do the legwork necessary to find, validate, and defend the exact Windows-specific solution. These problems only reinforce my doubts in the Project current abilities when it comes to providing Windows support, for any reasonable definition of "support"...

P.S. To avoid misunderstanding, by "remove Squid namespace", I meant removing `namespace Squid {` and the matching `}` characters while still keeping our custom wrappers defined inside those curly braces.

compat/mswindows.cc
yadij: Indeed it does.

kinkie:
Scope creep is too big, parts of this have been reworked in more constrained PRs. Abandoning

PR #1086, opened 2022-07-10
Maintenance: Eliminate a (wrong) solution in a TODO comment

rousskov:


src/acl/Acl.h
```diff
@@ -121,7 +121,8 @@ namespace Acl {
 class Answer
 {
 public:
-    // not explicit: allow "aclMatchCode to Acl::Answer" conversions (for now)
+    // TODO: Find a good way to avoid implicit conversion (without explicitly
+    // casting every ACCESS_ argument in implicit constructor calls).
     Answer(const aclMatchCode aCode, int aKind = 0): code(aCode), kind(aKind) {}
```
rousskov, line 126: > @kinkie  in https://github.com/squid-cache/squid/pull/1083#issuecomment-1179662619: could you please change that comment to clarify intent?

The existing comment already reflects the intent correctly -- we should add `explicit`. AFAICT, you opened #1083 not because you misunderstood the intent of that comment, but because you wanted to address that to-do. The problem with this official comment is not an unclear (expression of) intent but the complete lack of documentation regarding  _how_ that to-do item should be addressed. That problem is typical/natural for to-do comments.

Unfortunately, I cannot offer a blueprint for the correct solution without spending unwarranted (IMO) amounts of time on it -- I do not know all the necessary details myself and would need time to research/evaluate the correct solution.

In this PR I quickly documented one known wrong solution in hope to address your request.

kinkie (approved):
Good for me, thanks!

PR #1094, opened 2022-07-19
Replaced most custom high-level callbacks with a unified API

rousskov:
Regular asynchronous calls remember call arguments at call creation
time, when the type of the call is well known. Service callbacks do not
know call arguments until it is time for the service to (asynchronously)
call the previously supplied callback. Thus, the service has to get
access to call parameters inside the stored generic (from service
p.o.v.) async call pointer. Services got that access by declaring a
custom callback dialer class and expecting the service user code to use
that dialer when creating an async call object. That was problematic:

1. Each service class had to create nearly identical dialer classes, at
   least one for each unique answer type. Code duplication was rampant.
2. Each service dialer class was specific to one service user type
   (e.g., HappyConnOpener answer dialer did not support AsyncJob users).
3. A compiler could not check whether the service got the right
   callback: A dynamic_cast operation converted a generic dialer to the
   service-declared dialer class, asserting the right type at runtime.

This change replaces all but one custom service callback with three
general service-independent dialers, one for each callback style: a job
call, a call to a cbdata-protected class method, and a stand-alone
function call. This unified API solves problem 1 mentioned above. A new
Callback template supports all three dialers (addressing problem 2) and
checks the callback answer type at compile time (addressing problem 3).

Replaced custom HappyConnOpener, Security::PeerConnector, Downloader,
CertValidationHelper, Http::Tunneler, Ipc::Inquirer, and most
Ipc::StartListening() callbacks (and more such callbacks would have been
added by upcoming PRs!). Kept ListeningStartedDialer as detailed below.

Replaced Ipc::Inquirer callback with a regular call: The class does not
need a callback mechanism because it can (and should and now does)
create an async call from scratch when the call arguments become known.

Kept ListeningStartedDialer because client_side.cc relies on delivering
additional, non-answer parameters with the callback. Normally, this
additional information is stored inside the job object. In this case, no
job awaits the port opening, so we use a custom dialer hack to keep that
information during the wait. Improving that code is out of scope. This
may be the only custom high-level callback left in the code!

The asyncCall() return type improvement preserves the exact AsyncCall
type for the Callback constructor to pick up. It will also remove
explicit/imprecise AsyncCall::Pointer uses throughout the code.

Also fixed reporting of small responses received by Downloader.

Unfortunately, we cannot easily unify the old JobCallback() with the new
asyncCallback() because we do not yet know how to detect that we need to
use one of the Comm-specific dialers (e.g., CommCbMemFunT) that feed the
job pointer to the CommCommonCbParams constructor. CommCommonCbParams do
not have a default constructor because they require, probably
incorrectly, that the recipient (e.g., job) pointer is stored together
with the future callback argument(s). This is why JobCallback() requires
a Dialer as a parameter -- it does not know how to guess it either. Comm
callback improvements will need a dedicated PR.


src/clients/HttpTunneler.cc
```diff
@@ -59,14 +61,12 @@ Http::Tunneler::doneAll() const
     return !callback || (requestWritten && tunnelEstablished);
 }
 
+// TODO: Remove this diff-reducer before merging this PR.
```
rousskov, line 64: This and other "before merging" TODOs should and will be addressed before merging the PR. Please do not block this PR on their account. Once these diff reducers are removed, the diff noise will increase, complicating comprehension of the important/key PR changes.

src/security/PeerConnector.cc
```diff
@@ -640,27 +619,29 @@ Security::PeerConnector::certDownloadNestingLevel() const
...
+    const auto &obj = downloaderAnswer.resource;
+    const auto downloadStatus = downloaderAnswer.outcome;
+
     ++certsDownloads;
     debugs(81, 5, "Certificate downloading status: " << downloadStatus << " certificate size: " << obj.length());
 
     Must(Comm::IsConnOpen(serverConnection()));
     const auto &sconn = *fd_table[serverConnection()->fd].ssl;
 
+    // XXX: Do not parse the response when the download has failed.
```
rousskov, line 644: Addressing this and other XXX/TODO PR-added comments (without "before merging" marks) are out of scope. If you do not think these comments should be added, please just say so, and I will remove them without arguing.

src/ipc/StartListening.cc
```diff
@@ -18,26 +19,22 @@
...
+    if (answer.errNo)
+        os << ", err=" << answer.errNo;
+    return os;
 }
 
 void
 Ipc::StartListening(int sock_type, int proto, const Comm::ConnectionPointer &listenConn,
-                    FdNoteId fdNote, AsyncCall::Pointer &callback)
+                    FdNoteId fdNote,
+                    StartListeningCallback &callback)
```
yadij, line 34:
```suggestion
                    FdNoteId fdNote, StartListeningCallback &callback)
```

src/security/PeerConnector.cc
```diff
@@ -35,7 +36,11 @@
 
 CBDATA_NAMESPACED_CLASS_INIT(Security, PeerConnector);
 
-Security::PeerConnector::PeerConnector(const Comm::ConnectionPointer &aServerConn, AsyncCall::Pointer &aCallback, const AccessLogEntryPointer &alp, const time_t timeout) :
+Security::PeerConnector::PeerConnector(
+    const Comm::ConnectionPointer &aServerConn,
+    const AsyncCallback<EncryptorAnswer> &aCallback,
+    const AccessLogEntryPointer &alp,
+    const time_t timeout):
```
yadij, line 43:
```suggestion
Security::PeerConnector::PeerConnector(const Comm::ConnectionPointer &aServerConn, const AsyncCallback<EncryptorAnswer> &aCallback,
    const AccessLogEntryPointer &alp, const time_t timeout):
```

src/security/PeerConnector.h
```diff
@@ -125,7 +116,7 @@ class PeerConnector: virtual public AsyncJob, public Acl::ChecklistFiller
     void startCertDownloading(SBuf &url);
 
     /// Called by Downloader after a certificate object downloaded.
-    void certDownloadingDone(SBuf &object, int status);
+    void certDownloadingDone(DownloaderAnswer &answer);
```
yadij, line 119:
```suggestion
    void certDownloadingDone(DownloaderAnswer &);
```

src/security/PeerConnector.h
```diff
@@ -52,18 +53,8 @@ class PeerConnector: virtual public AsyncJob, public Acl::ChecklistFiller
...
-    public:
-        virtual ~CbDialer() {}
-        /// gives PeerConnector access to the in-dialer answer
-        virtual Security::EncryptorAnswer &answer() = 0;
-    };
-
-public:
     PeerConnector(const Comm::ConnectionPointer &aServerConn,
-                  AsyncCall::Pointer &aCallback,
+                  const AsyncCallback<EncryptorAnswer> &aCallback,
```
yadij, line 57:
```suggestion
                  const AsyncCallback<EncryptorAnswer> &,
```

src/tests/stub_libsecurity.cc
```diff
@@ -77,7 +76,11 @@ class TlsNegotiationDetails: public RefCountable {};
 CBDATA_NAMESPACED_CLASS_INIT(Security, PeerConnector);
 namespace Security
 {
-PeerConnector::PeerConnector(const Comm::ConnectionPointer &, AsyncCall::Pointer &, const AccessLogEntryPointer &, const time_t) :
+PeerConnector::PeerConnector(
+    const Comm::ConnectionPointer &,
+    const AsyncCallback<EncryptorAnswer> &,
+    const AccessLogEntryPointer &,
+    const time_t):
```
yadij, line 83:
```suggestion
PeerConnector::PeerConnector(const Comm::ConnectionPointer &, const AsyncCallback<EncryptorAnswer> &, const AccessLogEntryPointer &, const time_t):
```

yadij (changes requested):

rousskov (commented):
I hope all change requests are addressed as of PR commit 91e4128.

@yadij, please let me know when it is time to complete the pending diff-reduction TODOs.

rousskov:
@yadij, please get back to this `review-1` PR. I believe all your change requests were addressed four weeks ago, and you have resolved all but two of them already. None of the remaining two threads warrant blocking this PR AFAICT:
* [one](https://github.com/squid-cache/squid/pull/1094#discussion_r927024255) is a discussion about adding more documentation to the (_unchanged_ by this PR) Downloader Answer field;
* the [other](https://github.com/squid-cache/squid/pull/1094#discussion_r927037901) one is a discussion about the optimization effects (or lack of thereof) of stripping a local variable declaration from its `const` decoration.

src/Downloader.h
```diff
@@ -23,6 +24,20 @@ typedef RefCount<DownloaderContext> DownloaderContextPointer;
...
 
+/// download result
+class DownloaderAnswer {
+public:
+    // The body of a successfully received HTTP 200 OK reply to our GET request.
+    // Unused unless outcome is Http::scOkay.
+    SBuf resource;
+
+    /// Download result summary.
+    /// May differ from the status code of the downloaded HTTP reply.
```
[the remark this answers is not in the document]
yadij: I am okay with that not being in the public description, but FTR believe it should be recorded somewhere in the code to avoid future confusion.

src/base/AsyncCallbacks.h
```diff
@@ -0,0 +1,206 @@
...
+        assert(answer_);
+        return *answer_;
+    }
+
+    /// make this smart pointer nil
+    /// \return the AsyncCall pointer we used to manage before this call
+    AsyncCall::Pointer release()
+    {
+        answer_ = nullptr;
+        const auto call = call_;
```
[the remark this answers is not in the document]
yadij: Looking at the ```-O0``` build shows that there are still 2 complete construct+copy+destruct cycles being performed here to generate the ```call` object, then to generate the returned object.
It is just that the optimizer detects that ```call``` object is implicitly const in your release2() test case. Then RefCount move optimization can be used on both. That decomposes into a raw-pointer move and nil assignment + a bunch of dead code.

yadij (approved):

yadij:
Sorry for not getting back to this sooner. I have no issues with the code going in despite the discussions. I just needs those polish changes you marked for after-review then it can be marked for merge.

src/base/AsyncCallbacks.h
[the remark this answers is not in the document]
rousskov: > Looking at the `-O0` build shows that ...

AFAICT, we were discussing how to improve this code performance. I see no connection between unoptimized build details and the previous performance-focused discussion. A lot of wasteful things happen (and should happen) in unoptimized builds. They are usually irrelevant in performance discussions, especially when talking about _optional_ optimizations. Moreover, removing `const` does not change unoptimized code!

> It is just that the optimizer detects that `call` object is implicitly const in your release2() test case. Then RefCount move optimization can be used on both.

As I detailed [earlier](https://github.com/squid-cache/squid/pull/1094/files#r927158143), the optimizing compiler does _not_ apply a move optimization to the `call` object, regardless of `const` presence. In both cases, the compiler uses a _better_ optimization -- copy elision.

The move optimization does happen in this code, but in a place unrelated to this discussion thread -- in the `call_ = nullptr` assignment.

If you want to see what is actually happening, you can use the updated [test](https://cpp.godbolt.org/z/a9KsYG7e8) that marks some RefCount methods and places in release1() and release2() functions with a markUse() call. Tracking string literals used as markUse() parameters lets us see the initial a copy constructor (LC3 is used before LC6), followed by the nullptr assignment move optimization (LC7 between LC6 and LC8) and no moves or copies after LC8 (due to copy elision).

```asm
.LC3:
        .string "copy ctor"
.LC6:
        .string "call = call_ done"
.LC7:
        .string "move assignment"
.LC8:
        .string "call_ = nullptr done"
```

In summary, removing `const` does not change this code performance. Thus, `const` should not be removed on "better performance" grounds.

src/clients/HttpTunneler.cc
rousskov: Done in 146c808.

src/security/PeerConnector.cc
```diff
@@ -640,27 +615,29 @@ Security::PeerConnector::certDownloadNestingLevel() const
...
 void
-Security::PeerConnector::certDownloadingDone(SBuf &obj, int downloadStatus)
+Security::PeerConnector::certDownloadingDone(DownloaderAnswer &downloaderAnswer)
 {
     certDownloadWait.finish();
 
+    // TODO: Remove diff reduction before merging
+    const auto &obj = downloaderAnswer.resource;
+    const auto downloadStatus = downloaderAnswer.outcome;
+
```
rousskov: Done in aa80a3b.

rousskov (commented):
Removed diff reducers. Clearing for merge.

yadij:
GCC 12 is coming up with use-after-free build errors after this merge. see bug 5232.

PR #1096, opened 2022-07-20
Silence GnuRegex warnings on gcc 12.1

kinkie:
GnuRegex does pointer magic that gcc 12.1 doesn't like.

    compat/GnuRegex.c: error: array subscript -1 is outside array bounds
    of 'const char[]' [-Werror=array-bounds]

    compat/GnuRegex.c: error: pointer may be used after 'realloc'
    [-Werror=use-after-free]

A long term fix is to unship GnuRegex which is an old
snapshot and is not maintained.

### DIFF
--- a/compat/GnuRegex.c
+++ b/compat/GnuRegex.c
@@ -40,6 +40,12 @@
 
 #if USE_GNUREGEX /* only if squid needs it. Usually not */
 
+/* recent GCC 12.1 does not like array magic done here */
+#if __GNUC__ >= 12 && __GNUC_MINOR__ >= 1
+#pragma GCC diagnostic ignored "-Warray-bounds"
+#pragma GCC diagnostic ignored "-Wuse-after-free"
+#endif
+
 #if !HAVE_ALLOCA
 #define REGEX_MALLOC 1
 #endif
--- a/compat/GnuRegex.c
+++ b/compat/GnuRegex.c
@@ -40,8 +40,9 @@
 
 #if USE_GNUREGEX /* only if squid needs it. Usually not */
 
-/* recent GCC 12.1 does not like array magic done here */
-#if __GNUC__ >= 12 && __GNUC_MINOR__ >= 1
+/* Starting with v12.1, GCC warns of various problems with this ancient code. */
+/* GCC versions prior to v12.1 do not support these pragmas. */
+#if (__GNUC__ == 12 && __GNUC_MINOR__ >= 1) || (__GNUC__ > 12)
 #pragma GCC diagnostic ignored "-Warray-bounds"
 #pragma GCC diagnostic ignored "-Wuse-after-free"
 #endif


kinkie:
This is triggered by the failures at https://build.squid-cache.org/job/anybranch-amd64-matrix/COMPILER=gcc,OS=fedora-36,label=amd64-linux/73/consoleFull

I've build-tested it with 
$ do-build --do-build --in=fedora-36 --layer=02 --parallel=10

rousskov:
> A long term fix is to unship GnuRegex which is an old snapshot and is not maintained,

I agree.

> and rely instead on gnulib

I probably disagree. Switching to [std::regex](https://en.cppreference.com/w/cpp/regex/basic_regex) is probably a better long-term solution. I [asked](http://lists.squid-cache.org/pipermail/squid-dev/2022-January/009760.html) "Do you think we should bite the bullet and add STL regex support to Squid master branch?" on squid-dev, but did not get any responses. Commit 0fa036e was a step towards that goal.

I removed the corresponding part from the PR description so that it does not block merging this PR.

kinkie:
> > and rely instead on gnulib
> 
> I probably disagree. Switching to [std::regex](https://en.cppreference.com/w/cpp/regex/basic_regex) is probably a better long-term solution. I [asked](http://lists.squid-cache.org/pipermail/squid-dev/2022-January/009760.html) "Do you think we should bite the bullet and add STL regex support to Squid master branch?" on squid-dev, but did not get any responses. Commit [0fa036e](https://github.com/squid-cache/squid/commit/0fa036e32cfa348ed90d849699ee0a2918b415f1) was a step towards that goal.

It is a better long term fix, I agree. It requires using an entirely different API so it's probably more work, but in principle I agree.

> I removed the corresponding part from the PR description so that it does not block merging this PR.

Thanks

compat/GnuRegex.c
```diff
@@ -40,6 +40,12 @@
 
 #if USE_GNUREGEX /* only if squid needs it. Usually not */
 
+/* recent GCC 12.1 does not like array magic done here */
+#if __GNUC__ >= 12 && __GNUC_MINOR__ >= 1
```
rousskov, line 44: We should not skip GCC v13.0, v14.0, etc.. Also, we should explain why we need to be fairly precise here.

Finally, let's not assert that this is only about array magic.

```suggestion
/* Starting with v12.1, GCC warns of various problems with this ancient code. */
/* GCC versions prior to v12.1 do not support these pragmas. */
#if (__GNUC__ == 12 && __GNUC_MINOR__ >= 1) || (__GNUC__ > 12)
```

rousskov (commented):
I am not excited about adding pragmas, but I think we can make an exception for `compat/` code that is difficult (AFAICT) to fix and that is going to be removed in the foreseeable future.

yadij:
IIRC we only held on to this library fro MinGW. But the current compilers used for cross-build support should provide C++11 regex capabilities there. So I propose we try to drop the library again and migrate to C++11 regex before or with the MinGW port is resumed.

yadij:
Correction. I dug into my mail archive and found another rationale also related to CentOS 7 which is still under support until June 2024.

So @rousskov do we need to support the latest shiny Squid version on CentOS 7 or are you going to bump the Squid-6 minimum version of CentOS 8 supported?

kinkie:
I think we all agree that std::regex is the future, and that moving to it is a nontrivial endeavour.
This warning is triggered by:

1. - configure --enable-gnuregex
2. - mingw - I'm currently seeing if I can fix the issues with crosscompiling to that target
3. - very old OS with very new compiler


The first one is a blocker to testing on modern distros; the other two are very very corner cases.
This hack unblocks 1. I don't think we should worry about 3, and 2 is a WIP

IMO this hack is good enough until someone finds the time to port to std::regex (obviously).
Another option would be to remove the --enable-gnuregex configure switch, it makes little sense except for testing the gnuregex library which is not commonly used but then we would carry around untested code

rousskov:
> I think we all agree that std::regex is the future

Looks like it. I will resume working on that support.

> Francesco: IMO this hack is good enough

I (still) agree, provided the [suggested](https://github.com/squid-cache/squid/pull/1096#discussion_r926055072) corrections are committed.

rousskov:
> I dug into my mail archive and found another rationale [for keeping built-in GnuRegex] also related to CentOS 7 which is still under support until June 2024.

Can you clarify/detail why Squid on CentOS 7 requires Squid keeping built-in GnuRegex? FWIW, googling shows many references to regcomp() and libglib on CentOS 7, so I am guessing that there are libraries providing POSIX regular expressions support on CentOS 7?

> So @rousskov do we need to support the latest shiny Squid version on CentOS 7 or are you going to bump the Squid-6 minimum version of CentOS 8 supported?

It is not my decision; it is a Project decision. So far, we have failed to agree on what "support" means, so it is not possible to answer such "support" questions in a meaningful way.

FWIW, I support removing GnuRegex under the (unverified) assumption that such removal will not prevent folks from using Squid (perhaps with some additional effort) on CentOS 7. My question above may help verify that assumption.

yadij:
> > I dug into my mail archive and found another rationale [for keeping built-in GnuRegex] also related to CentOS 7 which is still under support until June 2024.
> 
> Can you clarify/detail why Squid on CentOS 7 requires Squid keeping built-in GnuRegex? 

It doesn't require gnuregex. Its version of GCC (4.8) has broken std::regex. For CentOS we can drop GnuRegex and require -lregex.

rousskov:
> > > I dug into my mail archive and found another rationale [for keeping built-in GnuRegex] also related to CentOS 7 which is still under support until June 2024.

> > Can you clarify/detail why Squid on CentOS 7 requires Squid keeping built-in GnuRegex?
 
> It doesn't require gnuregex. Its version of GCC (4.8) has broken std::regex.

Ah, OK. I assume that it is not very difficult to install a modern GCC version if somebody really wants to build Squid from source on CentOS 7.

> For CentOS we can drop GnuRegex and require -lregex.

Great (until we decide to remove support for that as well). However, libregex will not cover everything std::regex can do so support on default CentOS 7 install will lack some features. IMHO, that is OK, especially if we can warn the admin about that problem.

kinkie:
> > Francesco: IMO this hack is good enough
> 
> I (still) agree, provided the [suggested](https://github.com/squid-cache/squid/pull/1096#discussion_r926055072) corrections are committed.

I accepted them, thanks for noticing my code wasn't future proof

rousskov (approved):

PR #1097, opened 2022-07-22
Fix handling of bad SNMP variable type when parsing IP_* vars

rousskov:
AFAICT, we never associate snmp_netIpFn() with the wrong variable type,
so we may be dealing with a bug in an effectively unreachable code here.

1999 commit 736eb6a correctly removed snmp_var_free() call from most
refactored SNMP functions but missed snmp_netIpFn() and
snmp_meshCtblFn() cases. The latter was fixed in 2000 commit d20b1cd.

Detected by static analyzer Svace v3.2.1.


rousskov:
Disclaimer: I am not sure what this code is doing exactly; terminology in the PR description may be wrong.

yadij (approved):
Looks good and correct to me.

FTR; This is indeed unreachable code right now. IIRC it is there simply for paranoid handling in case a future dev links the function to a new OID without adjusting it appropriately.

PR #1099, opened 2022-07-25
Maintenance: Improve stability of gperf-generated sources

rousskov:
Control gperf from source-maintenance.sh: Existing Makefile rule did not
use (and should not use) primary "make" functionality -- we do not want
to automatically trigger RegisteredHeadersHash.cci regeneration based on
file timestamps alone because most users will not have (the right
version of) gperf and (the right version of) astyle available. The rule
did use Makefile variables, but that resulted in unstable gperf output
because variables like $src change depending on the developer build
environment. This change finishes the move started by commit 3d50bea.

Simplify formater.pl, making its interface similar to other formatting
tools used by source-maintenance.sh: The script now sends its results to
stdout. Renamed the script to reduce surprises, fix the old naming typo,
and to improve consistency with format-makefile-am.pl. The script needs
more polishing; we avoided otherwise unmodified lines except removing an
unused buggy $pfx line that triggered warnings.

Move "astyle did not corrupt code" checks from source-maintenance.sh to
formater.pl, where they belong and are easier to implement. This move
simplifies reuse of the Perl script inside source-maintenance.sh. It
also removes dependency on md5sum/etc. $CppFormatter now contains the
exact command to (re)use.

Fix "Generated by.." marker. Its old location in .cci was hard to
notice. The marker was also present in hand-edited .gperf file!

Report RegisteredHeadersHash.cci changes.

Also stopped stripping "register" from gperf output. Gperf v3.1 stopped
adding "register" to C++ output. We (should) use that gperf version.
Removing "register" complicated gperf error handling because a pipe to
sed ate gperf exit code.


scripts/format-cpp.pl
```diff
@@ -91,63 +66,71 @@
...
-sub safeRename
-{
-    my ($from, $to) = @_;
-    die() if -e $to;
-    rename($from, $to) or die("Failed to rename $from to $to: $!, stopped");
-}
+        if (!length $formattedCode) {
+            warn("ERROR: Running astyle produced no output while formatting $out\n".
+                 "    astyle command: $ASTYLE_BIN\n");
+            print $originalCode;
```
rousskov, line 94: I doubt we should restore the original (and possibly unformatted) output. Instead, we should just report the error and exit.

The current PR code kind of preserves the official behavior that was probably motivated by the combination of buggy astyle versions and lack of content preservation guarantees. These days, modern astyle does not fail very often, and git checks guarantee that nothing is lost. Thus, we can and, hence, should treat such failures like most source-maintenance.sh failures -- report and exit, leaving any artifacts as they were.

If you agree _and_ think that this improvement/simplification belongs to _this_ PR, then please let me know, and I will implement it.

This applies to other `print $originalCode` places, of course.

scripts/source-maintenance.sh
```diff
@@ -12,15 +12,6 @@
 # on Squid
 #
 
-## Source Code Format Enforcement
-#
-# A checker to recursively reformat all source files: .h .c .cc .cci
-# using a custom astyle formatter and to use MD5 to validate that
-# the formatter has not altered the code syntax.
```
rousskov, line 19: I removed this stale and partially duplicated comment because we no longer depend on MD5. I can leave it intact if you prefer. IIRC, another pending PR will replace it anyway.

scripts/source-maintenance.sh
```diff
@@ -351,21 +343,12 @@ for FILENAME in `git ls-files`; do
...
-			else
-				rm -f $FILENAME.astylebak
-			fi
-        	fi
+	if test "$CppFormatter"; then
+		if $CppFormatter $FILENAME > $FILENAME.new; then
+			updateIfChanged $FILENAME $FILENAME.new 'by astyle'
+		else
+			rm $FILENAME.new
+		fi
```
rousskov, line 351: This is another place where we are automatically "recovering" from formatting errors. As discussed in another commit, I do not think we should be doing that, and I can polish this code further to fail on formatting problems. The developer running this script can always use `--keep-going` to get past early failures.

PR #1103, opened 2022-07-26
Do not "open" Comm::Connection on oldAccept() errors

rousskov:
When Comm::TcpAcceptor::oldAccept() discovers an error, it should close
the socket it just accepted. We should use RAII to avoid leaking the
open socket, especially in the presence of multiple error detection
areas and C++ exceptions. Using the available Connection object for
controlling socket lifetime does not work well because that Connection
object often lingers well past oldAccept() -- the object delivers
various low-level error details (e.g., the remote HTTP client address of
the failed attempt) to TcpAcceptor users.

Instead of "opening" the Connection object ASAP to avoid FD leaks and
then struggling to find the right time to close it, we now delay that
opening until oldAccept() succeeds. Meanwhile, the socket lifetime and
Comm registration are controlled by a simple RAII Descriptor object.

Eventually, we will reuse Comm::Descriptor in Connection and other
descriptor-owning code. One known prerequisite for those improvements is
Optional<> support for non-trivial types (a work in progress).

### DIFF
--- a/src/comm/TcpAcceptor.cc
+++ b/src/comm/TcpAcceptor.cc
@@ -342,12 +342,12 @@ Comm::Flag
 Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
 {
     ++statCounter.syscalls.sock.accepts;
-    int sock;
     struct addrinfo *gai = nullptr;
     Ip::Address::InitAddr(gai);
 
     errcode = 0; // reset local errno copy.
-    if ((sock = accept(conn->fd, gai->ai_addr, &gai->ai_addrlen)) < 0) {
+    const auto rawSock = accept(conn->fd, gai->ai_addr, &gai->ai_addrlen);
+    if (rawSock < 0) {
         errcode = errno; // store last accept errno locally.
 
         Ip::Address::FreeAddr(gai);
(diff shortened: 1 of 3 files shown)


yadij:
Squid deals with multiple types of descriptors. Some of which need to be closed with (eg TCP), some of which must not be closed (eg UDP).

IMO this new class should deal with that distinction.

a) Rename the class to clarify that these are descriptors that require fd_close() ?

b) Adjust the API somehow to choose between RAII close or ignore behaviour selectable by the creator object?

c) other solutions?

rousskov:
> Squid deals with multiple types of descriptors. Some of which need to be closed with (eg TCP), some of which must not be closed (eg UDP).

Please give an example of a Comm-managed descriptor that must not be closed. All Comm-created descriptors that I can think of should be closed, including UDP descriptors (e.g., DNS code calls comm_close() for IPPROTO_UDP descriptors that it creates). The OS closes all descriptors on process exit, seemingly violating the stated "must not be closed" invariant that defines this special type of a descriptor.

> a) Rename the class to clarify that these are descriptors that require fd_close()?

We could, but should not the need to close be the implied default (i.e. something that does not need explicit naming)? If some very special Comm descriptor needs some very special handling, then that descriptor can be handled by a specially named class/code. No need to bother everybody else with that exceptional situation AFAICT.

> b) Adjust the API somehow to choose between RAII close or ignore behaviour selectable by the creator object?

Sure, we can add Descriptor methods to mark special descriptors (that still want to use Comm::Descriptor) specially when/if we discover such a need. It is a change that does not need to be planned in advance AFAICT.

rousskov:
> It is not possible to use a proper Client/Server child class that receives a Comm::Descriptor with UDP descriptor for any of the UDP protocols Squid currently services.

What makes it impossible? What prevents some future job code from passing a Comm::Descriptor object to a "proper Client/Server" job? And why is the problem specific to UDP? And where does "must not close" come into play? I must be missing something important that blocks me from solving this mystery.

In fact, I hope that Comm::Connection (which is passed from one job to another in the existing code) will use the proposed Descriptor as a data member (instead of a raw integer) in the foreseeable future. I see nothing in the Descriptor design that would prevent that existing exchanges from continuing or would prevent adding new exchanges, including those that pass a UDP Descriptor.

PR #1105, opened 2022-07-28
Cleanup cache.log output of protocol flow

yadij:
In debug sections dedicated to displaying protocols, level-2 is
reserved for logging a human readable trace of messages sent and
received.

Provide a DBG_PROTOCOL constant easily used for these log
outputs. This also provides an easier way to search for examples
of the syntax use by these traces.

Also, update log output syntax for protocols which already
contain message flow debugs().

Future work:
* display traces for more protocols not yet traced
* add options to debug_options directive for displaying just
  these traces
* add traces to clarify protocol switching on a connection


src/clients/FtpClient.cc
```diff
@@ -822,10 +822,12 @@ Ftp::Client::dataClosed(const CommCloseCbParams &)
 void
 Ftp::Client::writeCommand(const char *buf)
 {
-    char *ebuf;
-    /* trace FTP protocol communications at level 2 */
-    debugs(9, 2, "ftp<< " << buf);
+    debugs(9, DBG_PROTOCOL, "FTP Server REQUEST: " << ctrl.conn <<
+           "\n----------\n" <<
+           buf <<
+           "\n----------");
```
rousskov, line 829: If we are going to touch all these lines and add so many new ones, let's stop duplicating all this code. Please add and reuse a function. I suggest calling that new function debugProtoHeader(), debugHeader(), debugProtoMetadata().

src/Downloader.cc
```diff
@@ -127,6 +127,15 @@ downloaderDetach(clientStreamNode * node, ClientHttpRequest * http)
...
 }
 
+static const char *
+debugPrintHeaders(const HttpHeader &h)
+{
+    static MemBuf tmp;
+    tmp.init();
+    h.packInto(&tmp, true);
+    return tmp.buf;
+}
```
rousskov, line 137: Please add an std::ostream `<<` operator to print HttpHeader instead. While implementing HttpHeader::print() or similar, please avoid legacy MemBuf and returning local/static buffers. Ideally, we should provide a zero-copy ostream-to-Packable converter (let me know if you are willing to wait for me to post a corresponding PR), but I will not insist on such an out-of-scope detour.

src/debug/Stream.h
```diff
@@ -37,10 +37,11 @@
...
 /* defined names for Debug Levels */
-#define DBG_CRITICAL    0   /**< critical messages always shown when they occur */
-#define DBG_IMPORTANT   1   /**< important messages always shown when their section is being checked */
-/* levels 2-8 are still being discussed amongst the developers */
-#define DBG_DATA    9   /**< output is a large data dump only necessary for advanced debugging */
+#define DBG_CRITICAL   0 /**< critical messages always shown when they occur */
+#define DBG_IMPORTANT  1 /**< important messages always shown when their section is being checked */
+#define DBG_PROTOCOL   2 /**< show flow of protocol messages. Excluding message content/payload which is in DBG_DATA. */
+/* levels 3-8 are still being discussed amongst the developers */
+#define DBG_DATA       9 /**< output is a large data dump only necessary for advanced debugging */
```
rousskov, line 45: > PR description: add options to debug_options directive for displaying just these traces

Based on the proposed `DBG_PROTOCOL` definition, I would expect that to be already supported by `debug_options ALL,2`. How would those future options differ from that existing support?

rousskov, line 45: Please avoid modifying otherwise unchanged lines above just to align their number "column". We should avoid such decorations because they force us to modify code that should not be touched without adding sufficient value in return.

rousskov, line 45: If we describe/define `DBG_PROTOCOL` like this, we will need to change 400+ level-2 debugs() messages (that currently do not show protocol messages) to use debug level 3 or higher. Will you do that at the end of this PR work, once all other issues are resolved?

Otherwise, we need to adjust `DBG_PROTOCOL` description/definition to include those level-2 messages.

src/debug/Stream.h
```diff
@@ -37,10 +37,11 @@
 #define MAX_DEBUG_SECTIONS 100
 
 /* defined names for Debug Levels */
-#define DBG_CRITICAL    0   /**< critical messages always shown when they occur */
-#define DBG_IMPORTANT   1   /**< important messages always shown when their section is being checked */
-/* levels 2-8 are still being discussed amongst the developers */
-#define DBG_DATA    9   /**< output is a large data dump only necessary for advanced debugging */
+#define DBG_CRITICAL   0 /**< critical messages always shown when they occur */
+#define DBG_IMPORTANT  1 /**< important messages always shown when their section is being checked */
+#define DBG_PROTOCOL   2 /**< show flow of protocol messages. Excluding message content/payload which is in DBG_DATA. */
```
rousskov, line 42: For clarity and consistency sake:

```suggestion
#define DBG_PROTOCOL   2 /**< protocol messages excluding content/payload which is DBG_DATA */
```

or

```suggestion
#define DBG_PROTOCOL   2 /**< protocol message header/metadata (message content/payload is DBG_DATA) */
```

rousskov (changes requested):
Thank you for unifying protocol message reporting code.

src/debug/Stream.h
yadij: ```ALL,2``` includes a large number of sections that are not about a protocol and print output at level-2. To get a clearer trace right now, admin have to know the individual section numbers which are about protocols and set them to level-2 individually.

The goal for that future work item would be to let admin specify the protocols by name instead of section number. I have a few half-baked ideas so far. Not ready for being in-scope here.

yadij: > Will you do that at the end of this PR work, once all other issues are resolved?

I do not intend to do that myself, though I would not object to anyone else doing so.

> 
> Otherwise, we need to adjust `DBG_PROTOCOL` description/definition to include those level-2 messages.

IMO we could define a second label for section 2 of non-protocol sections. This label is not about a severity level like CRITICAL/IMPORTANT were.

src/clients/FtpClient.cc
yadij: Done in[02047b6](https://github.com/squid-cache/squid/pull/1105/commits/02047b63724d1dbb06c16cebabe84f553b2f086b).  I made it a macro following the debugs() definition to simplify the variance of parameter types.

src/debug/Stream.h
yadij: Define "sufficient" ?

src/Downloader.cc
yadij: I considered that approach at first. However, like you have realized HttpHeader really needs a different packInto() API and that is significant scope creep.

For now I made this static function to match all other header printing code. This Downloader is the oddity among HTTP message senders as all other callers generate the MemBuf which can be passed to debugs().

src/debug/Stream.h
rousskov: To address this change request, we need to make a choice:

1. Reserve level-2 debugging for protocol header dumps and such. Add `DBG_PROTOCOL`. Deprecate level-2 use for other purposes. Get rid of legacy uses.
2. Invent a new model/meaning for the debugs() level and/or section parameters to accommodate the proposed "protocol message debugging" use case and especially its [planned](https://github.com/squid-cache/squid/pull/1105#discussion_r931999860) "protocol X message debugging" enhancement. Adjust the PR code accordingly.
3. Do not introduce the new `DBG_PROTOCOL` level name at all. Continue to use anonymous level 2 for message header dumps and other existing use cases.
4. Broaden `DBG_PROTOCOL` level name/meaning to include most current level-2 uses.

You are not excited about item 1 prospects AFAICT, and I share that assessment. I can help with item 2 if that is the direction you favor; it will require non-trivial amounts of work (but hopefully not a lot of additional code changes). I doubt item 4 will work well.

I strongly recommend to start with item 3 because it does not require a lot of work and is fully compatible with future protocol-specific tracing enhancements. With item 3, we can always upgrade to item 2 if/when that becomes necessary without undoing a lot of work.

rousskov: To address this change request, I suggest adopting the following wording for the second bullet in the PR TODO list:

* Enhance debug_options to support tracing named protocols

or, a more verbose variant:

* Enhance debug_options to support tracing just the protocol(s) selected by their names.

rousskov: Whether a given value is sufficient to overcome known problems cannot be defined -- it is a judgement call. The author is responsible for convincing the reviewer that reformatting three out of three old definitions to gain alignment is worth causing all those known problems, including the time spent on this change request. So far, I am unconvinced. My recommendation is to cut our losses by keeping the official definition lines unchanged (and to avoid creating similar problems in the future by avoiding touching the lines to align something that was not aligned and does not need to be aligned).

src/servers/Http1Server.cc
```diff
@@ -394,12 +392,14 @@ Http::One::Server::noteTakeServerConnectionControl(ServerConnectionContext serve
...
 Http::NewServer(const MasterXaction::Pointer &xact)
 {
+    debugs(11, DBG_PROTOCOL, "switch to HTTP on: " << xact->tcpClient);
     return new Http1::Server(xact, false);
 }
 
 ConnStateData *
 Https::NewServer(const MasterXaction::Pointer &xact)
 {
+    debugs(11, DBG_PROTOCOL, "switch to HTTP on: " << xact->tcpClient);
```
rousskov, line 402: Please remove this line. This change is highly controversial (e.g., I do not think we are switching to HTTP here) and lies outside the declared PR scope.

src/servers/Http1Server.cc
```diff
@@ -394,12 +392,14 @@ Http::One::Server::noteTakeServerConnectionControl(ServerConnectionContext serve
 ConnStateData *
 Http::NewServer(const MasterXaction::Pointer &xact)
 {
+    debugs(11, DBG_PROTOCOL, "switch to HTTP on: " << xact->tcpClient);
```
rousskov, line 395: Please remove this line. This change is controversial (e.g., I am not sure we are switching to HTTP _here_ and would need to develop a clear notion of protocol scope to be certain) and lies outside the declared PR scope.

src/debug/Stream.h
```diff
@@ -207,6 +208,14 @@ void ResyncDebugLog(FILE *newDestination);
         } \
    } while (/*CONSTCOND*/ 0)
 
+/// Convenience wrapper for displaying protocol message flow in cache.log
+#define protocolTrace(SECTION, LABEL, ID, MESSAGE) \
```
rousskov, line 212: I suggest calling the new debugs() wrapper debugProtoHeader(), debugHeader(), debugProtoMetadata(). Keeping the same prefix for both macros will help with code understanding and maintenance.

If you are violently against that suggestion, then at least call the new function using a function-like name -- traceProtocol() -- instead of the proposed class/object-like name -- protocolTrace().

src/clients/HttpTunneler.cc
```diff
@@ -341,6 +338,7 @@ Http::Tunneler::handleResponse(const bool eof)
     // preserve any bytes sent by the server after the CONNECT response
     futureAnswer.leftovers = readBuf;
 
+    debugs(11, DBG_PROTOCOL, "switch to TCP on:" << connection);
```
rousskov, line 341: Please remove this line. This change is controversial (e.g., I am not sure we are switching to TCP here and would need to develop a clear notion of protocol scope to be certain) and lies outside the declared PR scope.

rousskov (changes requested):
Thank you for removing so much duplicated code!

I have updated unaddressed change requests and opened a few new ones.

src/debug/Stream.h
yadij: done

yadij: Ergo, you are imposing on all contributors a requirement that every change no matter how small be justified and defended extensively to some undefined arbitrary standard of "quality". Any contributor attempting to improve code readability that does not go to those unreasonable length in advance is forced to expend extra work and time either reverting trivial formatting _improvements_ of justifying some ROV retrospectively.

IMO that is unreasonable expectation. Contributors should have the right to choose how they contribute without having to justify self-evident code polish.

src/clients/HttpTunneler.cc
yadij: The "protocol scope" you are looking for is defined in RFC 9111 section 9.3.6:
 "if successful, thereafter restrict its behavior to blind forwarding of data, in both directions, until the tunnel is closed."

Squid currently only supports (and HttpTunneler implements) HTTP/1.x tunnels over TCP transport layer. So from the perspective of HTTP protocol trace this is a switch to TCP.

FTR, this line does **not** relate to whether the opaque "TCP" immediately switches to some other Foo protocol (eg TLS for SSL-Bump). That would be shown in the trace of TCP protocol as a "switch to Foo" log entry.

src/servers/Http1Server.cc
yadij: An HTTP server AsyncJob is being created and assigned ownership of the connection (per your definition of Comm::Connection lifecycle vs ownership discussions). This should not be controversial.

src/servers/Http1Server.cc
yadij: An HTTP server AsyncJob is being created and assigned ownership of the connection (per your definition of Comm::Connection lifecycle vs ownership discussions). This should not be controversial.

src/debug/Stream.h
rousskov: > you are imposing on all contributors a requirement that every change no matter how small be justified and defended extensively to some undefined arbitrary standard of "quality".

I do not. Process improvement suggestions are welcome.

> Contributors should have the right to choose how they contribute without having to justify self-evident code polish.

I agree. In this specific case, we probably disagree on a few _other_ things, including whether introducing or fixing an unwelcomed alignment is "self-evident code polish" or an easily fixable mistake.

src/debug/Stream.h
yadij: Done in [65b9b1c](https://github.com/squid-cache/squid/pull/1105/commits/65b9b1c6cfaaafead9611aaa647d767f9734a4f1)

src/debug/Stream.h
yadij: Dropped in [e3a7b48](https://github.com/squid-cache/squid/pull/1105/commits/e3a7b48eb0e00683d6b7284126ccbf260c6ec64b)

yadij: Dropped with DBG_PROTOCOL removal in [e3a7b48](https://github.com/squid-cache/squid/pull/1105/commits/e3a7b48eb0e00683d6b7284126ccbf260c6ec64b)

src/debug/Stream.h
```diff
@@ -207,6 +207,11 @@ void ResyncDebugLog(FILE *newDestination);
         } \
    } while (/*CONSTCOND*/ 0)
 
+/// Convenience wrapper for displaying protocol message flow in cache.log
+#define traceProtocol(SECTION, LABEL, ID, MESSAGE) \
+    debugs(SECTION, 2, LABEL << ": " << ID << \
```
rousskov, line 212: Please protect macro arguments (where possible):

```suggestion
    debugs((SECTION), 2, (LABEL) << ": " << ID << \
```

src/debug/Stream.h
```diff
@@ -207,6 +207,11 @@ void ResyncDebugLog(FILE *newDestination);
...
+           "\n----------\n" << MESSAGE << "\n----------")
```
rousskov, line 213: When dealing with boundary separators, it would be really nice to switch from very inconvenient dash counting/typing to words that are easy to type and search for. A few leading dashes is more than enough to make these strings unique (because Squid virtually never prints dashes at the beginning of a line) and we can add a few trailing dashes to decorate (they will not be needed when searching for things). For example:

```suggestion
           "\n--begin--\n" << MESSAGE << "\n--end----")
```

or 

```suggestion
           "\n--start--\n" << MESSAGE << "\n--stop---")
```

src/comm/TcpAcceptor.cc
```diff
@@ -419,6 +419,8 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
         }
     }
 
+    traceProtocol(5, "TCP Client REQUEST", conn, details);
```
rousskov, line 422: Please do not abuse the message-printing traceProtocols() to print accepted TCP connection details. If we were working at a TCP packet level, we could print SYN packets and such, of course, but this is not message dumping that traceProtocols() is dedicated to. It will also produce quite a bit of unwelcomed multiline noise where it is totally unnecessary. The line is also misplaced IMO, and polishing all this can be viewed as an out of scope change.


```suggestion
```

src/servers/FtpServer.cc
```diff
@@ -700,10 +700,11 @@ Ftp::Server::parseOneRequest()
     Must(parsed && cmd.length());
     consumeInput(tok.parsedSize()); // TODO: Would delaying optimize copying?
 
-    debugs(33, 2, ">>ftp " << cmd << (params.isEmpty() ? "" : " ") << params);
-
     cmd.toUpper(); // this should speed up and simplify future comparisons
 
+    traceProtocol(9, "FTP Client REQUEST", clientConnection,
```
rousskov, line 705: Please report the received `cmd` before we change it. These message dumps should use as much _raw/actual_ data as practical to improve triage experience.

src/Downloader.cc
rousskov: Do you mind if I try to address my change request without introducing a lot of out-of-scope changes? If I succeed, this PR will not add dangerous code (that also uses legacy APIs that we want to get rid of). I think these benefits warrant such an attempt.

src/servers/Http1Server.cc
```diff
@@ -394,12 +392,14 @@ Http::One::Server::noteTakeServerConnectionControl(ServerConnectionContext serve
...
 Http::NewServer(const MasterXaction::Pointer &xact)
 {
+    debugs(11, 2, "switch to HTTP on: " << xact->tcpClient);
     return new Http1::Server(xact, false);
 }
 
 ConnStateData *
 Https::NewServer(const MasterXaction::Pointer &xact)
 {
+    debugs(11, 2, "switch to HTTP on: " << xact->tcpClient);
```
rousskov, line 402: Moving outdated https://github.com/squid-cache/squid/pull/1105#discussion_r932224018 to new code:

> Alex: Please remove this line. This change is highly controversial (e.g., I do not think we are switching to HTTP here) and lies outside the declared PR scope.

> Amos: An HTTP server AsyncJob is being created and assigned ownership of the connection. This should not be controversial.

IMO, this code does not switch to HTTP and arguably does not (or should not) even create an HTTP server job. You may disagree, of course, but that disagreement is what makes this out-of-scope change controversial.

I support tracing protocol switches, but doing that right requires more work, and that work is outside _this_ PR scope. We want minimal, tightly-focused PRs, right? Removing this and the remaining "switch" tracing lines will move this PR closer to that ideal and avoid lengthy unnecessary discussions!

rousskov (changes requested):
Thank you for moving this PR forward. It looks much better. I updated unresolved old change requests and added a few new ones, but we are making progress.

PR #1106, opened 2022-07-28
Support non-trivial Optional values

rousskov:
Commit 7224761 skipped this std::optional feature when adding initial
Optional implementation because that feature is difficult to support.
However, several in-progress projects now need to make non-trivial
objects Optional. This implementation maintains key Optional invariants,
matching those of std::optional:

* A valueless Optional object does not construct or destruct the value.
* A valued Optional object supports move/copy ops supported by value.

Maintaining the first invariant is tricky:

* Union prevents value construction/destruction in a valueless Optional.
* Explicit destructor call destructs the value in a _valued_ Optional.
* A dummy union member works around a C++ requirement for constexpr
  unions to have at least one active (i.e. initialized) member. Since a
  valueless Optional cannot initialize the value, we need another union
  member that we can initialize.
* We use an _anonymous_ union because C++ places more requirements on
  named unions, triggering a more complex implementation with
  placement-new and to-be-deprecated std::aligned_storage_t tricks!

XXX: This simplified implementation violates the following std::optional
invariant. We believe that this violation does not hurt the current and
foreseeable Squid code. In our tests, optimizing compilers still
optimize Optional<Trivial> destructors away.

* std::optional<Value> is trivially destructible if Value is.

### DIFF
--- a/src/base/Optional.h
+++ b/src/base/Optional.h
@@ -58,7 +58,11 @@ public:
         return *this;
     }
 
-    Optional(Optional<Value> &&other) { *this = std::move(other); }
+    Optional(Optional<Value> &&other): Optional()
+    {
+        if (other.hasValue_)
+            *this = std::move(other);
+    }
 
     Optional &operator =(Optional<Value> &&other)
     {
--- a/src/base/Optional.h
+++ b/src/base/Optional.h
@@ -49,7 +49,7 @@ public:
     Optional &operator =(const Optional &other)
     {
         if (this != &other) {
-            if (other.has_value()) {
+            if (other.hasValue_) {
                 *this = other.value_;
             } else {
                 reset();
@@ -68,7 +68,7 @@ public:
     Optional &operator =(Optional<Value> &&other)
     {
         if (this != &other) {
-            if (other.has_value()) {
+            if (other.hasValue_) {
                 *this = std::move(other.value_);
                 // no other.reset() per std::optional move semantics
             } else {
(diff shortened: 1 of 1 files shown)


rousskov:
FWIW, we already have working (but unpolished) code that uses non-trivial values with Optional for at least two different purposes and will submit the corresponding PRs after this enhancement lands. With time, I expect such non-trivial Optionals to be used in many places where the underlying value is not allocated dynamically, making std::unique_ptr unacceptable for indicating the lack of a value.

yadij:
FWIW, std::optional is another feature which is only needed by CentOS 7 and similar (unsupported) old OS. This base/Optional.h code should be provided by libcompat under the name std::optional when the STL feature is missing.

yadij (approved):

rousskov:
> std::optional is another feature which is only needed by CentOS 7 and similar (unsupported) old OS.

I wish we could define "supported" so that we (know when we) can, for example, upgrade the required C++ standard instead of wasting many hours on working around C++11 deficiencies. I plan to resume Francesco's effort in that direction shortly.

yadij:
> > std::optional is another feature which is only needed by CentOS 7 and similar (unsupported) old OS.
> 
> I wish we could define "supported" so that we (know when we) can, for example, upgrade the required C++ standard instead of wasting many hours on working around C++11 deficiencies. I plan to resume Francesco's effort in that direction shortly.

For now I am working with supported being a) we have a build farm node for the OS checking PRs and/or merge, and b) distrowatch indicates the OS version has a future EOL date.

PR #1107, opened 2022-07-29
Use std::optional where available

yadij:
Our class Optional is a portability replacement for std::optional on
older compilers.

### DIFF
--- a/compat/compat.h
+++ b/compat/compat.h
@@ -55,6 +55,7 @@
 #define __STDC_FORMAT_MACROS
 
 #include "compat/types.h"
+#include "compat/optional.h"
 
 /*****************************************************/
 /* per-OS hacks. One file per OS.                    */
--- a/compat/optional.h
+++ b/compat/optional.h
@@ -9,13 +9,12 @@
 #ifndef SQUID__COMPAT_OPTIONAL_H
 #define SQUID__COMPAT_OPTIONAL_H
 
-#if __cplusplus
-#if __has_include(<optional>)
+#if __cplusplus > 201701L
 
 #include <optional>
 template<class T> using Optional = std::optional<T>;
 
-#else
+#elif __cplusplus
 
 #include <exception>
 #include <type_traits>
@@ -129,5 +128,4 @@ private:
     bool hasValue_ = false;
 };
 #endif
-#endif /* __cplusplus */
 #endif /* SQUID__COMPAT_OPTIONAL_H */
--- a/compat/optional.h
+++ b/compat/optional.h
@@ -6,8 +6,16 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-#ifndef SQUID__SRC_BASE_OPTIONAL_H
-#define SQUID__SRC_BASE_OPTIONAL_H
+#ifndef SQUID__COMPAT_OPTIONAL_H
+#define SQUID__COMPAT_OPTIONAL_H
+
+#if __cplusplus
+#if __has_include(<optional>)
+
+#include <optional>
+template<class T> using Optional = std::optional<T>;
+
+#else
 
 #include <exception>
 #include <type_traits>
@@ -120,6 +128,6 @@ private:
 
     bool hasValue_ = false;
 };
-
-#endif /* SQUID__SRC_BASE_OPTIONAL_H */
-
+#endif
+#endif /* __cplusplus */
+#endif /* SQUID__COMPAT_OPTIONAL_H */
--- a/compat/Makefile.am
+++ b/compat/Makefile.am
@@ -44,6 +44,7 @@ libcompatsquid_la_SOURCES = \
 	memrchr.h \
 	mswindows.cc \
 	openssl.h \
+	optional.h \
 	os/aix.h \
 	os/android.h \
 	os/dragonfly.h \
--- a/src/SquidMath.h
+++ b/src/SquidMath.h
@@ -10,7 +10,6 @@
 #define _SQUID_SRC_SQUIDMATH_H
 
 #include "base/forward.h"
-#include "base/Optional.h"
 
 #include <limits>
 #include <type_traits>
--- a/src/base/ClpMap.h
+++ b/src/base/ClpMap.h
@@ -9,7 +9,6 @@
 #ifndef SQUID__SRC_BASE_CLPMAP_H
 #define SQUID__SRC_BASE_CLPMAP_H
 
-#include "base/Optional.h"
 #include "mem/PoolingAllocator.h"
 #include "SquidMath.h"
 #include "time/gadgets.h"
--- a/src/base/Makefile.am
+++ b/src/base/Makefile.am
@@ -48,7 +48,6 @@ libbase_la_SOURCES = \
 	JobWait.h \
 	Lock.h \
 	LookupTable.h \
-	Optional.h \
 	Packable.h \
 	PackableStream.h \
 	RandomUuid.cc \
--- a/src/base/forward.h
+++ b/src/base/forward.h
@@ -20,8 +20,6 @@ class BadOptionalAccess;
 class Raw;
 class RegexPattern;
 
-template <typename Value> class Optional;
-
 template<class Cbc> class CbcPointer;
 template<class RefCountableKid> class RefCount;
 template<class Job> class JobWait;
--- a/src/debug/debug.cc
+++ b/src/debug/debug.cc
@@ -9,7 +9,6 @@
 /* DEBUG: section 00    Debug Routines */
 
 #include "squid.h"
-#include "base/Optional.h"
 #include "base/TextException.h"
 #include "debug/Stream.h"
 #include "fd.h"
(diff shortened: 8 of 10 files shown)


compat/optional.h
```diff
@@ -6,8 +6,16 @@
...
-#ifndef SQUID__SRC_BASE_OPTIONAL_H
-#define SQUID__SRC_BASE_OPTIONAL_H
+#ifndef SQUID__COMPAT_OPTIONAL_H
+#define SQUID__COMPAT_OPTIONAL_H
+
+#if __cplusplus
+#if __has_include(<optional>)
+
+#include <optional>
+template<class T> using Optional = std::optional<T>;
```
rousskov, line 16: :question: I am concerned that doing this will introduce difficult to triage runtime bugs where our buggy[^1] Optional code in the reporter environment is substituted by correct std::optional implementation in the developer environment, hiding the problem. It will definitely cause PR-time (or, worse, later!) headaches when a developer writes code that is supported by std::optional but not by our Optional and (if we are lucky!) leads to a compile-time error in some CI environment.

Since we have to write Squid code with the "weakest point" (i.e. our custom/incomplete Optional implementation) in mind anyway, what do we have to _gain_ by doing this substitution? A tiny bit of performance improvement in some std::optional implementations? Even that is not guaranteed and has not been demonstrated. I cannot think of enough reasons to justify the associated risks...

[^1]: I am not implying that the current or [proposed](https://github.com/squid-cache/squid/pull/1106) Optional replacement is buggy, but I know first-hand that it is easy to make difficult-to-discover bugs in that low-level code. We are operating at the edge of our abilities here and, hence, should be extra careful. The cost/benefit tradeoff here is different than in most other/legitimate compat/ cases!

rousskov (changes requested):
Please start with a change request marked with :?:. It may be a good idea to ignore the other change requests until that one is resolved.

compat/optional.h
rousskov: I have updated PR title/description to emphasize the important part of this PR changes. The move to libcompat is a boring technicality compared to the ramifications of using std::optional (instead of Optional) where the former is available. Please adjust further as needed.

yadij: What we gain with this model of portability is that bugs in our custom code is restricted to affecting only outdated OS.

While this is not great for users of those OS, outdated OS only decrease in usage over time. Reducing the impact of any bugs present in the custom implementation(s). Also counteracting technical debt incurred by delays bug fixing.

compat/optional.h
```diff
@@ -6,8 +6,16 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-#ifndef SQUID__SRC_BASE_OPTIONAL_H
-#define SQUID__SRC_BASE_OPTIONAL_H
+#ifndef SQUID__COMPAT_OPTIONAL_H
+#define SQUID__COMPAT_OPTIONAL_H
+
+#if __cplusplus
+#if __has_include(<optional>)
```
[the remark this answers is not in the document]
yadij: You are right. I mistook what the error was supposed to be. This should be resolved by reverting to the old version detection in [57f992b](https://github.com/squid-cache/squid/pull/1107/commits/57f992bedfb853a88d2acef10bad36208ac139eb)

compat/optional.h
```diff
@@ -6,8 +6,16 @@
...
+
```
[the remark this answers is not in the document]
yadij: removed in [57f992b](https://github.com/squid-cache/squid/pull/1107/commits/57f992bedfb853a88d2acef10bad36208ac139eb)

compat/optional.h
```diff
@@ -6,8 +6,15 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-#ifndef SQUID__SRC_BASE_OPTIONAL_H
-#define SQUID__SRC_BASE_OPTIONAL_H
+#ifndef SQUID__COMPAT_OPTIONAL_H
+#define SQUID__COMPAT_OPTIONAL_H
+
+#if __cplusplus > 201701L
+
```
rousskov, line 13:
```C++
#if __cplusplus >= 201703L
```
Based on the linked godbolt tests and compiler documentation, this or similar conditions will only become true in Squid builds starting with [GCC](https://gcc.godbolt.org/z/7E8eneGee) v11. And it will not become true for [clang](https://gcc.godbolt.org/z/3YYbfG7jv) until we actually switch to C++17 (when this code becomes useless and should be removed).

I think that behavior does not match the [stated](https://github.com/squid-cache/squid/pull/1107/files#r933767132) goal of detecting "only outdated OS".

compat/optional.h
```diff
@@ -6,8 +6,15 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-#ifndef SQUID__SRC_BASE_OPTIONAL_H
-#define SQUID__SRC_BASE_OPTIONAL_H
+#ifndef SQUID__COMPAT_OPTIONAL_H
+#define SQUID__COMPAT_OPTIONAL_H
+
+#if __cplusplus > 201701L
```
rousskov, line 12: If this test survives despite its other, higher-level problems (detailed in other change requests), please use a [standard](https://en.cppreference.com/w/cpp/preprocessor/replace) constant value.

```suggestion
#if __cplusplus >= 201703L
```

[SO](https://stackoverflow.com/a/38456243) has more information about the evolution of these constants. The answer there appears to support the change suggested above.

However, if I missed something, and we have to use a non-standard constant value, then we should document why (via a source code comment).

rousskov (changes requested):

compat/optional.h
yadij: Okay. Description change acceptable.

compat/optional.h
yadij: I started with the obvious ```>= 201703L``` test. However testing which OS chose to build the custom code and which used libstd showed some precompilers having trouble with the ```>=``` (operator not supported) resulting in false for the condition.

Also, some libstdc++ libraries defining it to values related to their latest update rather than the standard document release.

compat/optional.h
rousskov: > Okay. Description change acceptable.

I am not sure what description change you are talking about, but this change request is about the mismatch between the effects of the proposed _code_ and its stated (in the [comment](https://github.com/squid-cache/squid/pull/1107/files#r933767132) quoted below) goal of detecting "only outdated OS". This code does not detect outdated OSes. For example, this code may not enable std::optional use on the very latest OS release running clang (please see the change request for the details of what this code detects). If the intent is correct, then this code is wrong. If the intent is wrong, then I need to know what the correct intent is in order to review this code. If the tests this change request is based on are wrong, then the change request itself needs a fix. None of that is related to any "description change" though.

> https://github.com/squid-cache/squid/pull/1107/files#r933767132: What we gain with this model of portability is that bugs in our custom code is restricted to affecting only outdated OS.

compat/optional.h
rousskov: > However testing which OS chose to build the custom code and which used libstd showed some precompilers having trouble with the >= (operator not supported) resulting in false for the condition.

I fear that those test results were the side effects of other problems, discussed in other change requests. Will wait until those change requests are resolved before asking you to detail your "operator not supported" test results here. For now, please note that Squid uses `>=` in preprocessor code quite a bit already:

```C++
compat/GnuRegex.c:#if (__GNUC__ == 12 && __GNUC_MINOR__ >= 1) || (__GNUC__ > 12)
compat/fdsetsize.h:#if (__GLIBC__ > 2) || (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 2)
compat/shm.cc:#if _SQUID_FREEBSD_ && (__FreeBSD__ >= 7)
compat/tempnam.c:#if (L_tmpnam >= L_tmpmin + LONG_BIT / 4)
src/client_side.cc:#if defined(OPENSSL_VERSION_NUMBER) && OPENSSL_VERSION_NUMBER >= 0x00908000L
src/ip/Intercept.cc:#if defined(IPFILTER_VERSION) && (IPFILTER_VERSION >= 4000027)
src/security/ServerOptions.cc:#if USE_OPENSSL && OPENSSL_VERSION_NUMBER >= 0x0090800fL && !defined(OPENSSL_NO_ECDH)
src/ssl/support.cc:#if OPENSSL_VERSION_MAJOR >= 3
```

rousskov (commented):
Just updating old change requests...

compat/compat.h
```diff
@@ -55,6 +55,7 @@
 #define __STDC_FORMAT_MACROS
 
 #include "compat/types.h"
+#include "compat/optional.h"
```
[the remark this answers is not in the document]
yadij: FTR; The requirement is to *prefer* including libcompat headers by default (via squid.h) in order to make the while compat mechanism as invisible as possible to the remainder of the source code. We do not have to follow that strictly, but should consider whether the special extra #include "compat/..." overheads are worth littering the code.

STL optional is rather light in terms of STL includes. The sub-includes are already pulled into Squid ``.cc`` everywhere from other headers. The main difference from the home-brew logic is ``<new>`` which is one of the ones already pulled in widely. So there is almost no _additional_ weight added by this being in compat.

yadij:
Closing this PR since we are on the verge of updating to C++17 support already, making this irrelevant.

PR #1108, opened 2022-07-31
Fix build on arm32

kinkie:
On arm32, size_t and long are 32 bits in size, exposing these bugs.

### DIFF
--- a/src/security/PeerOptions.cc
+++ b/src/security/PeerOptions.cc
@@ -334,7 +334,7 @@ static struct ssl_option {
 #endif
 #if defined(SSL_OP_ALL)
     {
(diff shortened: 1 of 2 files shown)


rousskov (approved):
Thank you for fixing these bugs. If the `(long)` one did not result in a build failure, then my PR title adjustment is wrong and should be reverted.

I also polished the description to remove the `M-failed-description` label.

kinkie:
> Thank you for fixing these bugs. If the `(long)` one did not result in a build failure, then my PR title adjustment is wrong and should be reverted.



It did, that’s how I noticed it. Thanks for improving the description

yadij (approved):

PR #1109, opened 2022-08-01
Unify r.n.g. seed generation

rousskov:
Using time-based expressions for seeding pseudo random number generators
often results in multiple SMP workers getting the same seed. That, in
turn, may result in unwanted correlation of supposed-to-be-independent
worker activities: making artificially similar decisions (that are
supposed to be "random") and/or concentrating rather than
spreading/sharing load.

The recently added RandomUuid code did not use time-based seeds in
anticipation of this unification.

This change converts all std::mt19937 users in src/ except the random
ACL code (that probably deserves a dedicated change).

C++ standard does not guarantee that std::random_device does not force
users to wait while more entropy is accumulated. Based on public
comments and source code analysis of popular STL implementations, we
should be getting the desired no-wait behavior from popular STLs, in
popular environments. If Squid encounters a special environment where
the default std::random_device waits for entropy, more code and/or
configuration options will be needed to accommodate that special case.
The biggest drawback with this approach is that it may be difficult for
the admins to discover that std::random_device is waiting in their
environment. This is mitigated by the low probability of that happening.


src/auth/digest/Config.cc
```diff
@@ -22,6 +22,7 @@
 #include "auth/State.h"
 #include "auth/toUtf.h"
 #include "base/LookupTable.h"
+#include "base/Random.h"
```
rousskov, line 25: This and probably all other files that are using the new seeding API now `#include` both base/Random.h and `<random>` (not visible in the diff). I think that is the best approach because the fact that base/Random.h also includes `<random>` is not something the seeding API is meant to guarantee. It was just easier to implement those functions using `<random>` types. However, if others insist, we will remove `<random>` from this and similar files.

src/base/Random.cc
```diff
@@ -0,0 +1,39 @@
...
+    //
+    // * "/dev/urandom": Blocks GCC from picking the best entropy source (e.g.,
+    //   arc4random(3)) and leads to GCC/clang exceptions in some environments.
+    //
+    // If a special supported environment needs a non-default device name, we
+    // will add a random_device_name configuration directive. We cannot detect
+    // such needs in general code and choose to write simpler code until then.
+    static std::random_device dev;
+    return dev();
+}
```
rousskov, line 31: It would be easy to provide a time-based backup implementation. The difficulty here is not in providing that implementation but in knowing _when_ to use it -- we cannot reliably detect that random_device is waiting to accumulate more entropy... I think the current PR code represents the best (and simplest) implementation, at least until we discover a real use case that needs more.

src/fs/ufs/UFSSwapDir.cc
```diff
@@ -1078,7 +1079,7 @@ Fs::Ufs::UFSSwapDir::HandleCleanEvent()
          * value.  j equals the total number of UFS level 2
          * swap directories
          */
-        std::mt19937 mt(static_cast<uint32_t>(getCurrentTime() & 0xFFFFFFFF));
+        static std::mt19937 mt(Seed32());
```
rousskov, line 1082: This change avoids creation of large std::mt19937 objects every time we run this UFS cleaning code, but I think this "optimization" is not significant enough to warrant a PR description mention.

src/auth/digest/Config.cc
yadij: I do object. Please minimize the dependency include tree wherever possible.

src/base/Random.cc
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#include "base/Random.h"
+
+std::mt19937::result_type
+Seed32()
```
yadij, line 12: I am a little dubious about the name of this and the 64--bit variant. IMO it would be better called "RandomSeed*" to clarify that it is not a fixed value seed or such.

yadij (changes requested):

src/auth/digest/Config.cc
rousskov: > I do object. Please minimize the dependency include tree wherever possible.

Will do.

P.S. Please note that it is impractical to "minimize the dependency include tree wherever possible" -- such minimization is always "possible", but the amount of resources required to do that usually outweighs the benefits (by orders of magnitude!), and the changes required to achieve that minimum would often go out of the PR scope. "Minimum possible" cannot be our standard (until we have the necessary tools to achieve it without manual labor).

src/base/Random.cc
eduard-bagdasaryan: Done (1bc3f7a).

src/auth/digest/Config.cc
eduard-bagdasaryan: Done (30e4b89)

yadij (approved):

PR #1110, opened 2022-08-03
Remove send-announce

yadij:
This is another feature which is very much outdated and unused.


kinkie:
I agree to remove this feature.
I wish we could have some kind of telemetry, sending us a summary of the features used in production systems, it would help us make more informed choices about what matters most to our users :\

kinkie (approved):

src/Makefile.am
```diff
@@ -412,8 +412,6 @@ squid_SOURCES = \
 	refresh.h \
 	repl_modules.h \
 	sbuf/StringConvert.h \
-	send-announce.cc \
-	send-announce.h \
 	stat.cc \
```
rousskov, line 415: If you are removing the sender of these announcements, please remove the _recipient_ as well: src/recv-announce.cc.

src/Makefile.am
```diff
@@ -412,8 +412,6 @@ squid_SOURCES = \
 	refresh.h \
```
rousskov, line 412: > PR description: This is another feature which is very much outdated and unused.

Unless we are absolutely certain that nobody uses this feature, I would relax the above statement to say something like "This feature is outdated. It is probably unused. Modern replacements are available or can be easily built."

src/Makefile.am
yadij: You mean that tool which was removed last year?

rousskov: Yes, my bad -- looked at the wrong branch and missed [90086c5](https://github.com/squid-cache/squid/commit/90086c5a850470d7ce5b815c2f8612c3e59d1697). Sorry about the noise.

BTW, we can mention an apparent complaints-free removal of recv-announce as an additional circumstantial evidence of the announcement feature being unused.

PR #1114, opened 2022-08-08
Bug 3193 pt2: NTLM decoder truncating strings

yadij:
The initial bug fix overlooked large 'offset' causing integer
wrap to extract a too-short length string.

Improve debugs and checks sequence to clarify cases and ensure
that all are handled correctly.

### DIFF
--- a/lib/ntlmauth/ntlmauth.cc
+++ b/lib/ntlmauth/ntlmauth.cc
@@ -108,10 +108,19 @@ ntlm_fetch_string(const ntlmhdr *packet, const int32_t packet_size, const strhdr
     int32_t o = le32toh(str->offset);
     // debug("ntlm_fetch_string(plength=%d,l=%d,o=%d)\n",packet_size,l,o);
 
-    if (l < 0 || l > NTLM_MAX_FIELD_LENGTH || o + l > packet_size || o == 0) {
-        debug("ntlm_fetch_string: insane data (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
+    if (l < 0 || NTLM_MAX_FIELD_LENGTH < l) {
+        debug("ntlm_fetch_string: insane string length (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
         return rv;
     }
+    else if (o <= 0 || packet_size < o) {
+        debug("ntlm_fetch_string: insane string offset (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
+        return rv;
+    }
+    else if (l > packet_size - o) {
+        debug("ntlm_fetch_string: truncated string data (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
+        return rv;
+    }
+
     rv.str = (char *)packet + o;
     rv.l = 0;
     if ((flags & NTLM_NEGOTIATE_ASCII) == 0) {
--- a/lib/ntlmauth/ntlmauth.cc
+++ b/lib/ntlmauth/ntlmauth.cc
@@ -108,11 +108,11 @@ ntlm_fetch_string(const ntlmhdr *packet, const int32_t packet_size, const strhdr
     int32_t o = le32toh(str->offset);
     // debug("ntlm_fetch_string(plength=%d,l=%d,o=%d)\n",packet_size,l,o);
 
-    if (l < 0 || NTLM_MAX_FIELD_LENGTH < l) {
+    if (l < 0 || l > NTLM_MAX_FIELD_LENGTH) {
         debug("ntlm_fetch_string: insane string length (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
         return rv;
     }
-    else if (o <= 0 || packet_size < o) {
+    else if (o <= 0 || o > packet_size) {
         debug("ntlm_fetch_string: insane string offset (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
         return rv;
     }


lib/ntlmauth/ntlmauth.cc
```diff
@@ -108,10 +108,19 @@ ntlm_fetch_string(const ntlmhdr *packet, const int32_t packet_size, const strhdr
     int32_t o = le32toh(str->offset);
```
rousskov, line 108: Bad cast here and on the line above it. We should use `auto` to avoid blindly casting unsigned integer to a signed integer. And `const` if possible. Fixing this will require changing code that uses `l` and `o`, of course.

There may be a similar problem with the argument of the function -- the function parameter is unsigned but strhdr::offset is signed. I did _not_ check whether we can change `strhdr` declaration to use unsigned types and whether we at least check `strhdr` field values for not being negative before using them here.

lib/ntlmauth/ntlmauth.cc
```diff
@@ -108,10 +108,19 @@ ntlm_fetch_string(const ntlmhdr *packet, const int32_t packet_size, const strhdr
...
     // debug("ntlm_fetch_string(plength=%d,l=%d,o=%d)\n",packet_size,l,o);
 
-    if (l < 0 || l > NTLM_MAX_FIELD_LENGTH || o + l > packet_size || o == 0) {
-        debug("ntlm_fetch_string: insane data (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
+    if (l < 0 || NTLM_MAX_FIELD_LENGTH < l) {
```
rousskov, line 111: IMHO, the original order feels more natural, easier to read/grok:

```suggestion
    if (l < 0 || l > NTLM_MAX_FIELD_LENGTH) {
```

The above "item > limit" order is also more consistent with the last l check below.

lib/ntlmauth/ntlmauth.cc
```diff
@@ -108,10 +108,19 @@ ntlm_fetch_string(const ntlmhdr *packet, const int32_t packet_size, const strhdr
...
+        debug("ntlm_fetch_string: insane string length (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
         return rv;
     }
+    else if (o <= 0 || packet_size < o) {
```
rousskov, line 115: IMHO, the following order is more natural, easier to read/grok:

```suggestion
    else if (o <= 0 || o > packet_size) {
```

The above "item > limit" order is also more consistent with the last `l` check below.

I also find it odd that offset `o` cannot be zero, but length `l` can. If anything, I would expect the opposite, but I know the official code rejects zero offsets, and I do not even pretend to understand this code.

lib/ntlmauth/ntlmauth.cc
```diff
@@ -108,10 +108,19 @@ ntlm_fetch_string(const ntlmhdr *packet, const int32_t packet_size, const strhdr
     int32_t o = le32toh(str->offset);
     // debug("ntlm_fetch_string(plength=%d,l=%d,o=%d)\n",packet_size,l,o);
 
-    if (l < 0 || l > NTLM_MAX_FIELD_LENGTH || o + l > packet_size || o == 0) {
-        debug("ntlm_fetch_string: insane data (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
+    if (l < 0 || NTLM_MAX_FIELD_LENGTH < l) {
+        debug("ntlm_fetch_string: insane string length (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
```
rousskov, line 112: Here and on other new/changed debug() lines, `%d` codes do not match actual argument types (but this bug is present in the official code as well).

lib/ntlmauth/ntlmauth.cc
```diff
@@ -108,10 +108,19 @@ ntlm_fetch_string(const ntlmhdr *packet, const int32_t packet_size, const strhdr
...
+        return rv;
+    }
+    else if (l > packet_size - o) {
+        debug("ntlm_fetch_string: truncated string data (pkt-sz: %d, fetch len: %d, offset: %d)\n", packet_size,l,o);
+        return rv;
+    }
+
     rv.str = (char *)packet + o;
     rv.l = 0;
     if ((flags & NTLM_NEGOTIATE_ASCII) == 0) {
```
rousskov, line 126: The loop below appears to assume that `l` is even when dealing with `NTLM_NEGOTIATE_ASCII` but we do not actually check for that.

rousskov (approved):
AFAICT, the PR fixes the bug it claims to fix. The other problems can be viewed as being outside this PR scope.

lib/ntlmauth/ntlmauth.cc
yadij: Conflicting with previous requirements that range validation conditions always use the form   ```A < B < C```

lib/ntlmauth/ntlmauth.cc
yadij: There is a prefix octet at offset 0. It cannot be part of any contained string. Strings can be 'empty' though since some are mandatory fields.

lib/ntlmauth/ntlmauth.cc
yadij: Huh?  the loop iterator ```len``` is set to ```l>>1``` which erases the final bit. This is looping the number of UTF-16 characters, not number of octets.

lib/ntlmauth/ntlmauth.cc
yadij: ```%d``` is correct for printing 16/32-bit signed variables. It does not become a problem unless the types are converted to unsigned per the earlier request (which is not the case in current code).

lib/ntlmauth/ntlmauth.cc
yadij: You may have been fooled by the check that ASCII flag is set to **false** (```== 0```).

lib/ntlmauth/ntlmauth.cc
yadij: The casting effects are intentional and happen prior to input validation so that insane results can be handled per the logic this PR touches.

lib/ntlmauth/ntlmauth.cc
rousskov: > `%d` is correct for printing 16/32-bit signed variables.

I disagree: Lying about small fixed-size argument types "works" in Squid environments due to [default argument promotion](https://en.cppreference.com/w/cpp/language/variadic_arguments), but it is not "correct" because those promotions are not guaranteed to have the right effect for fixed-size types (e.g., `int32_t` is not guaranteed to be promoted to `int` because `int` is not guaranteed to even have 32 bits).

The best solution here is to get rid of printf() style itself, of course, but that requires a lot of work.

lib/ntlmauth/ntlmauth.cc
rousskov: > the loop iterator `len` is set to `l>>1` which erases the final bit. This is looping the number of whole UTF-16 characters, not number of octets.

Exactly. When `l` is 3 bytes, and we expect to see 2-byte characters, then something went wrong, and we should probably reject (or at least warn about) input instead of ignoring the fact that the total size of parsed 2-byte characters does not match the specified total length. In other words, we should reject or warn about input containing half-characters.

lib/ntlmauth/ntlmauth.cc
rousskov: > The casting effects are intentional

AFAICT, until C++20, these effects are implementation defined. In other words, this code relies on an undefined behavior. For example, the compiler has the right to ignore the assignment completely, put a random number into `o`, or call abort(). Here is the corresponding quote from [cppreference](https://en.cppreference.com/w/cpp/language/implicit_conversion):

> If the destination type is signed, the value does not change if the source integer can be represented in the destination type. Otherwise the result is implementation-defined until C++20

Needless to say, this code usually "works" in Squid environment, but it is still a bad cast.

PR #1115, opened 2022-08-08
Optimize ephemeral port reuse with IP_BIND_ADDRESS_NO_PORT

pponakan:
commBind: Cannot bind socket ... : (98) Address already in use

When opening a TCP connection from a specific IP address (e.g., set by
tcp_outgoing_address), Squid usually lets the OS select the source port
by performing the traditional bind(2)+connect(2) sequence with zero
source port. This sequence consumes ephemeral ports at the connection
opening rate even if many destinations are unique (and, hence, could
reuse the same source port) because the OS must pick a source port at
bind(2) time, before connect(2) supplies the destination address.

Starting with v4.2, Linux supports IP_BIND_ADDRESS_NO_PORT socket option
designed to decrease ephemeral port consumption by delaying port binding
until connect(2) time. If available, Squid now uses that option when
opening any outgoing TCP connection bound to a source IP address.


rousskov (changes requested):
Thank you very much for this contribution! While the current implementation cannot be accepted (for the reasons identified in specific change requests), it sounds like a useful feature that we should help you fix and accept. Would you be willing to address the blocking problems?

I am ignoring minor code imperfections for now. We can easily polish this after the bigger problems are solved.

src/comm.cc
```diff
@@ -200,11 +201,32 @@ comm_local_port(int fd)
     return F->local_addr.port();
 }
 
+/* Sets the IP_BIND_ADDRESS_NO_PORT sock option to enable reuse of Linux ephemeral ports */
+static inline void
+CommSetBindNoPort(int s)
+{
+    int bind_noport_flag = 1;
+#if _SQUID_LINUX_
+    if (setsockopt(s, IPPROTO_IP, IP_BIND_ADDRESS_NO_PORT, (char *) &bind_noport_flag, sizeof(int)) < 0) {
```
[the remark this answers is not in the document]
pponakan: Thanks for pointing this out. Updated with a new check that should work.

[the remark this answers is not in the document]
pponakan: As long as the socket is not being used as a listen port, this option should not have any drawbacks. None that were observed while testing this fix, which does not include ftp traffic.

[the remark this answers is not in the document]
pponakan: BTW, let me know if you think this function is unnecessary. If it makes more sense, the code block can be moved to the new check in comm_apply_flags().

src/comm.cc
```diff
@@ -200,11 +201,32 @@ comm_local_port(int fd)
...
+}
+
 static Comm::Flag
 commBind(int s, struct addrinfo &inaddr)
 {
     ++ statCounter.syscalls.sock.binds;
 
+    /* If this is a socket for an outbound TCP session, set the flag for ephemeral source port reuse */
+    sockaddr_in* addr = (sockaddr_in*)inaddr.ai_addr;
+    if (inaddr.ai_socktype == SOCK_STREAM && addr->sin_family == AF_INET && addr->sin_port == 0) {
```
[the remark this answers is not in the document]
pponakan: Thanks for pointing the FTP server use case. Hope the new code matches what you were thinking.

src/comm.cc
```diff
@@ -200,11 +201,32 @@ comm_local_port(int fd)
...
+CommSetBindNoPort(int s)
+{
+    int bind_noport_flag = 1;
+#if _SQUID_LINUX_
+    if (setsockopt(s, IPPROTO_IP, IP_BIND_ADDRESS_NO_PORT, (char *) &bind_noport_flag, sizeof(int)) < 0) {
+        int xerrno = errno;
+        debugs(50, DBG_IMPORTANT, MYNAME << "setsockopt(IP_BIND_ADDRESS_NO_PORT) " << (bind_noport_flag?"ON":"OFF") << " for FD " << s << ": " << xstrerr(xerrno));
+    }
+#else
+    debugs(50, DBG_CRITICAL, MYNAME << "WARNING: setsockopt(IP_BIND_ADDRESS_NO_PORT) not supported on this platform");
```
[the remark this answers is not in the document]
pponakan: Removed the debug print and the #if defined() check should suffice.

pponakan:
> Thank you very much for this contribution! While the current implementation cannot be accepted (for the reasons identified in specific change requests), it sounds like a useful feature that we should help you fix and accept. Would you be willing to address the blocking problems?
> 
> I am ignoring minor code and PR formatting imperfections for now. We can easily polish this after the bigger problems are solved.

Sure, glad to help out and please do point out any code paths that might break due to this fix.

src/comm.cc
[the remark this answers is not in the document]
rousskov: Thank you for addressing this (and other) problems. I am resolving this change request because the problem I flagged was addressed. I will open a new change request to discuss (and possibly adjust) _how_ that problem was addressed.

src/servers/FtpServer.cc
```diff
@@ -376,7 +376,7 @@ Ftp::Server::listenForDataConnection()
     conn->local = transparent() ? port->s : clientConnection->local;
     conn->local.port(0);
     const char *const note = uri.c_str();
-    comm_open_listener(SOCK_STREAM, IPPROTO_TCP, conn, note);
+    OpenBoundedListener(SOCK_STREAM, IPPROTO_TCP, conn, note);
```
rousskov, line 379: Hm... Is there any use case where a comm_open_listener() caller would want to use `IP_BIND_ADDRESS_NO_PORT`? I suspect there are no such use cases in the current code and there is no reason to believe that such cases will appear in the future: If I am opening a _listening_ port, then I do not _need_ (i.e. cannot benefit from) `IP_BIND_ADDRESS_NO_PORT`, right? If that is the case, then we do not need OpenBoundedListener() AFAICT -- its functionality can be moved into comm_open_listener() itself. If you agree, please adjust the code accordingly.

P.S. I should have realized the mutually exclusive nature of the new feature and comm_open_listener() in the original review. Sorry for this "extra" iteration.

rousskov (changes requested):
Thank you for advancing this PR.

I am still ignoring secondary problems until the OpenBoundedListener() situation is resolved. After that, unless you stop me, I will address them myself -- they are easier to fix than to describe in a series of change requests. You will be able to review and undo any unwanted commits/changes, of course.

rousskov:
> P.S. If the code diff looks good, I would like do some soak testing in our env before merging.

You can retest at any time, of course. I will let you know when I think the code is ready for another round of tests. We are not there yet.

src/servers/FtpServer.cc
pponakan: Ah .. I hadn't realized that the comm_open_listener() is only called for listen sockets which will never need the IP_BIND_ADDRESS_NO_PORT sockopt. Removed the extra wrapper function.

pponakan:
> 

Apologies if I still have any formatting issues or am not following some specific coding guidelines. I was not able to get the scripts/format-cpp.pl script to test my code diff for formatting. If there is anything else I should change, I would be glad to clean up. BTW, I wasnt sure whether I should use CamelCase or C-style naming for the private method that was added to comm.cc, as the file has a mix of both formats.

rousskov:
Excellent progress, @pponakan, thank you! Please give me a day to do a polishing round. Do not worry about formatting and various gray areas (in this inconsistent official code) -- we will take care of that. The ball is on my side now.

pponakan:
> Excellent progress, @pponakan, thank you! Please give me a day to do a polishing round. Do not worry about formatting and various gray areas (in this inconsistent official code) -- we will take care of that. The ball is on my side now.

Thanks @rousskov! I will run a canary test with this diff in our env to verify it still works as the initial patch.

src/comm/Connection.h
```diff
@@ -52,6 +52,7 @@ namespace Comm
 #define COMM_REUSEPORT          0x40 //< needs SO_REUSEPORT
 /// not registered with Comm and not owned by any connection-closing code
 #define COMM_ORPHANED           0x40
+#define COMM_BIND_NOW           0x80 // Skips IP_BIND_ADDRESS_NO_PORT sockopt to ensure local port is assigned
```
pponakan, line 55: @rousskov BTW, had noticed multiple comm flags using 0x40 in Connection.h. If you think it makes things simpler for you, I can push another change to have the new flag COMM_BIND_NOW introduced here to use the next bit 0x100.

rousskov: Your call. If I do not see that commit, I will do it myself. It was one of the changes on my to-do list. I still hope to work on this later today.

pponakan: Just pushed a commit to change it to 0x100.

rousskov:
OK to test

rousskov:
@pponakan, I am done with code changes for now. I tried to document most changes in commit messages. Please review and test. The changes took a somewhat unexpected turn after I cleaned up the code a bit. I hope I did not screw something up and you like the end result. Please adjust as needed, including reversal of anything and everything!

I will polish the PR title/description later -- out of time for now. Do not worry about the M-failed-description label until then.

pponakan:
Thanks @rousskov, the diff looks fine IMO. I will use this to build a new image for our canary test. Will keep you posted on how the test is doing after letting it soak for about 24 hrs.

src/comm.h
```diff
@@ -39,7 +39,6 @@ void comm_import_opened(const Comm::ConnectionPointer &, const char *note, struc
 
 /**
  * Open a port specially bound for listening or sending through a specific port.
- * This is a wrapper providing IPv4/IPv6 failover around comm_openex().
```
rousskov, line 42: I removed this comment because it mentioned comm_openex() which became private in this PR. It also read as if comm_open_listener() provides IPv4/IPv6 failover while relying on comm_openex() for the rest of the documented functionality. In reality, it is comm_openex() that provides failover; comm_open_listener() still provides failover indirectly, of course -- no changes there! The failover detail is still mentioned further below in the description blob so we are not losing any critical information here AFAICT.

If others disagree, here is a replacement line that would also work:

```suggestion
 * Provides IPv4/IPv6 failover as detailed below.
```

The above suggestion would need to be applied manually due to GitHub limitations.

rousskov (approved):
> Alex: I will polish the PR title/description later -- out of time for now.

Done. @pponakan, please review PR title/description. They will become the official commit message title/body when this PR is merged into the master branch by [Anubis](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message). Adjust as needed, keeping line lengths at or under 72 character limit enforced by Anubis.

I also looked through all comm_local_port() and getsockname() callers, trying to spot any code that would try to obtain the local port address right after calling comm_open() with zero port. Any such caller would be broken by this PR. FWIW, I did not find any. Unfortunately, I do not see a way to enforce that invariant by the compiler without drastic source code changes.

Please update this PR with your test results when you get them. No rush.

pponakan:
> 

Thanks. BTW when I pull this branch and attempt to run autoconf, I get the following error on an ubuntu system. I am trying to build squid from the latest commit with openssl 1.1.1n

configure.ac:13: error: possibly undefined macro: AM_INIT_AUTOMAKE
       If this token and others are legitimate, please use m4_pattern_allow.
       See the Autoconf documentation.
configure.ac:16: error: possibly undefined macro: AM_MAINTAINER_MODE
configure.ac:42: error: possibly undefined macro: AM_PROG_CC_C_O
configure.ac:129: error: possibly undefined macro: AM_CONDITIONAL

yadij:
You may need to update or install some missing build dependencies. On Ubuntu that can be done with:
```
  apt-get update
  apt-get build-dep squid
```
  (you may need to add the ```deb-src``` line for your Ubuntu version to APT sources.list)

Then re-bootstrap the toolchain:
```
  ./bootstrap.sh
  ./configure ......
  make
```

If that does not fix the build, what version of Ubuntu, automake and autoconf are you building with?

pponakan:
Thanks @yadij, I had forgotten to run ./bootstrap.sh, that fixed the build. I am on ubuntu 20.04.
However while running tests with a small local_port range I couldnt find any evidence of source port reuse on outbound TCP connections. I might have to recheck whether I did something wrong with the build and ended up not picking up the full diff. Will continue testing tomorrow.

src/comm.cc
```diff
@@ -226,6 +244,10 @@ comm_open(int sock_type,
           int flags,
           const char *note)
 {
+    // assume zero-port callers do not need to know the assigned port right away
+    if ((flags & COMM_DOBIND) && sock_type == SOCK_STREAM && addr.port() == 0)
```
rousskov, line 248: The current PR condition does not work as expected because some popular callers do _not_ set COMM_DOBIND even though they do want Comm to bind() to the address. The following condition does work in my quick tests:

```suggestion
    if (sock_type == SOCK_STREAM && addr.port() == 0 && !addr.isAnyAddr())
```

... but please review whether further adjustments are needed to cover all the cases we want to cover (and only them).

rousskov (changes requested):

src/comm.cc
rousskov: If we want to apply IP_BIND_ADDRESS_NO_PORT whenever Comm calls commBind() with zero TCP port, then this condition should probably become

```suggestion
    if (sock_type == SOCK_STREAM && addr.port() == 0 && ((flags & COMM_DOBIND) || !addr.isAnyAddr()))
```

Compared to the previous [suggestion](https://github.com/squid-cache/squid/pull/1115#discussion_r943464151), the above condition also covers use cases where the caller did not give a _specific_ IP address to bind to but did request Comm to bind (presumably to some address of OS choosing). I do not know whether such use cases exist now, but I suspect then may appear in the future even if they do not exist today.

The official condition for calling commBind() is

```C++
    if ( (flags & COMM_DOBIND) || addr.port() > 0 || !addr.isAnyAddr() ) {
```

pponakan:
Update on what I have been able to test. I have confirmed my build steps are pulling in the specific commits into the build image.

I went back to the very first commit on this PR and was able to observe the intended behavior. For example with local_port range reduced to 32000-32010 the following TCP sessions appear in netstats, where A.B.C.D is the tcp_outgoing_address and X.X.X.X, Y.Y.Y.Y are 2 dest IPs being accessed via the squid proxy.

tcp        0      0 A.B.C.D:32008   X.X.X.X:80      ESTABLISHED
tcp        0      0 A.B.C.D:32004   Y.Y.Y.Y:80       ESTABLISHED
tcp        0      0 A.B.C.D:32010   X.X.X.X:80      ESTABLISHED
tcp        0      0 A.B.C.D:32004   X.X.X.X:80      ESTABLISHED
tcp        0      0 A.B.C.D:32007   X.X.X.X:80      ESTABLISHED
tcp        0      0 A.B.C.D:32006   Y.Y.Y.Y:80       ESTABLISHED
tcp        0      0 A.B.C.D:32002   X.X.X.X:80      ESTABLISHED
tcp        0      0 A.B.C.D:32000   Y.Y.Y.Y:80       ESTABLISHED
tcp        0      0 A.B.C.D:32003   X.X.X.X:80      ESTABLISHED
tcp        0      0 A.B.C.D:32008   Y.Y.Y.Y:80       ESTABLISHED
tcp        0      0 A.B.C.D:32006   X.X.X.X:80      ESTABLISHED
tcp        0      0 A.B.C.D:32000   X.X.X.X:80      ESTABLISHED
tcp        0      0 A.B.C.D:32002   Y.Y.Y.Y:80       ESTABLISHED
tcp        0      0 A.B.C.D:32009   X.X.X.X:80      ESTABLISHED

I picked a few commits midway (my last commit 64d195b552d07e27bc81868032372fe5d46a43ed and 6a093ae6e67e323165e3c7c08cb21b280e09092f). Both seem to have issues as I only see 4 kid processes (we have 30 in the config) and source port was not very evident. I will switch back to the latest commit and try again.

pponakan:
> Can ignore the errors mentioned on my previous message. They donot appear if the local_port range is reduced after squid with 30 workers is brought up. Confirmed latest commit does not show reuse of outbound TCP source ports and a local health checking service on the same host as squid also runs out of ports that causes incorrect health status of squid to be reported (which was observed with the official 5.x versions that lead us to this PR).

rousskov:
> Confirmed latest commit does not show reuse of outbound TCP source ports

Please see my earlier [review](https://github.com/squid-cache/squid/pull/1115#discussion_r943464151) for that bug triage and suggested fixes.

pponakan:
FYI, I just retested commit 6a093ae6e67e323165e3c7c08cb21b280e09092f by reducing the local_port range after squid fully up. It shows source port reuse as well as the health checking service continues to function while all ports within the ephemeral port range are in use (with different 4-tuples).

src/comm.cc
rousskov: Just to avoid misunderstanding: I am not saying the code is buggy just because the code "looks" wrong. I did test the latest branch code (6cfbf23), and it does _not_ run IP_BIND_ADDRESS_NO_PORT-setting code (at all) when Squid is configured with a tcp_outgoing_address. That code does not run because Squid does _not_ set `COMM_DOBIND` in that case.

pponakan:
> > Confirmed latest commit does not show reuse of outbound TCP source ports
> 
> Please see my earlier [review](https://github.com/squid-cache/squid/pull/1115#discussion_r943464151) for that bug triage and suggested fixes.

@rousskov  I have built squid with the fix suggested [here](https://github.com/squid-cache/squid/pull/1115#discussion_r943479584) and run a spot test with a small local_port range. The image shows clear evidence of source port reuse in our env. I will wait till this potential solution is fully reviewed before starting a canary test with it next week.

src/comm.cc
yadij: FTR these are the cases (feel free to edit this table as needed):
<table>
<tr><th>case</th><th>address</th><th>port</th><th>DOBIND</th><th>REUSEADDR</th><th>IP_BIND_ADDRESS_NO_PORT</th></tr>

<tr><th>listen-port</th><td>NO</td><td>non-0</td><td>YES</td><td>NO</td><td>NO</td></tr>
<tr><th>listen-host:port</th><td>YES</th><td>non-0</td><td>YES</td><td>NO</td><td>NO</td></tr>

<tr><th>tcp_outgoing_address</th><td>YES</td><td>0</td><td>YES</td><td>NO</td><td>YES</td></tr>

<tr><th>FTP gateway DATA</th><td>YES</td><td>0</td><td>NO</td><td>NO</td><td>YES</td></tr>
<tr><th>FTP gateway PORT fallback</th><td>YES</td><td>non-0</td><td>NO</td><td>NO</td><td>NO</td></tr>
<tr><th>FTP gateway PASV</th><td>YES</td><td>non-0</td><td>NO</td><td>NO</td><td>NO</td></tr>
<tr><th>FTP relay DATA</th><td>YES</td><td>0</td><td>YES</td><td>NO</td><td>YES</td></tr>

<tr><th>DNS TCP</th><td>YES</td><td>0</td><td>YES</td><td>NO</td><td>YES</td></tr>

</table>

rousskov: I committed the second proposed suggestion (branch commit [d96d055](https://github.com/squid-cache/squid/pull/1115/commits/d96d055e52a5b91900540941a2d172c830105b23)). This commit addresses my change request, so I am resolving this thread.

> FTR these are the cases (feel free to edit this table as needed):

FWIW, I have not audited the proposed table, but it does not seem to contradict the committed fix AFAICT. Going forward, I suspect that commBind() triggers should be revised, but that work is outside _this_ PR scope.

rousskov (approved):

pponakan:
Thanks for the last commit d96d055e52a5b91900540941a2d172c830105b23 to the PR. If we need to wait for further approvals, can you please let me know so that I can plan for a canary test with the latest diff.

BTW, will this fix be made available with squid version 5.x as well (if 6.0 is the intended release)? Just to have a fallback in case we run into problems with version 6.0.

rousskov:
> Thanks for the last commit [d96d055](https://github.com/squid-cache/squid/commit/d96d055e52a5b91900540941a2d172c830105b23) to the PR. If we need to wait for further approvals, can you please let me know so that I can plan for a canary test with the latest diff.

Please keep in mind that Squid PR reviews are not carefully coordinated by some central authority -- no one can predict exactly what will happen to an open PR. FWIW, my _plan_ is to clear this PR for merging in a few days if you do not report test failures and there are no new change requests. Clearing a PR for merging does not guarantee anything either -- a last minute blocking review can stop the [automated](https://github.com/measurement-factory/anubis/blob/master/README.md#which-pull-requests-are-eligible-for-merging) merge sequence...

* If you would like to minimize the number of tests, then you should wait until the PR has no negative reviews _and_ is either at least ten days old or has at least two positive reviews. The "PR approval" status check at the bottom of the PR GitHub page will turn green when this condition is reached.
* If you want to test every (possibly temporary) cleared candidate, then please test now.

I would also recommend asking us not to merge the PR until your tests are over. We do not want to forget about this PR, but we are not in a rush to land it either and would be happy to wait for your green light. BTW, thank you for testing changes!

> BTW, will this fix be made available with squid version 5.x as well?

I cannot speak for the v5 maintainer -- @yadij will make that call. Please do not misinterpret my answer as agreeing (or disagreeing) with you that the proposed change is a "fix" ;-).

pponakan:
Thanks for explaining the merge process.

> * If you would like to minimize the number of tests, then you should wait until the PR has no negative reviews _and_ is either at least ten days old or has at least two positive reviews. The "PR approval" status check at the bottom of the PR GitHub page will turn green when this condition is reached.

I can actually start off 2 canaries, one that receives a smaller amount traffic and later another canary that usually exhausts all ephemeral ports during peak traffic w/o this socket option applied. I'll start off the 1st canary tomorrow and observe it for a few days. Based on how that goes and whether new changes are required, I will then plan on starting the more extensive 2nd canary early next week. 

> I cannot speak for the v5 maintainer -- @yadij will make that call. 
Thanks.

> Please do not misinterpret my answer as agreeing (or disagreeing) with you that the proposed change is a "fix" ;-).
Hopefully this PR enhances the scaling aspect w/o breaking other existing functionality. :)

rousskov:
> Hopefully this PR enhances the scaling aspect w/o breaking other existing functionality. :)

Yes, that is my hope or understanding as well.

pponakan:
Update on testing. Canary-1 has been running fine with the latest diff and shows a small number of source ports being reused even though the number of concurrent TCP sessions is only about 2-4% of the system's ip_local_port_range. Will continue with the plan of running the latest commit with canary-2 next week unless anything changes.

pponakan:
The 2nd canary has been running fine with the same image over the last few days. The peak concurrent outbound session count on this canary went beyond 2x the ip_local_port_range w/o any problems binding to ephemeral ports.

PR #1118, opened 2022-08-10
Fix FTBFS due to `-Werror=alloc-size-larger-than` on GCC 12

sergiodj:
GCC 12 emits an error when compiling squid due to
`-Werror=alloc-size-larger-than`.  The error looks like this:

```
store/Disks.cc:690:64: error: argument 1 value '18446744073709551615' exceeds maximum object size 9223372036854775807 [-Werror=alloc-size-larger-than=]
  690 |         const auto tmp = new SwapDir::Pointer[swap->n_allocated];
      |                                                                ^

```

Arguably, this is likely a compiler bug and not squid's fault.  Either
way, I believe it's worth having more type safety in this code.

### DIFF
--- a/src/SquidConfig.h
+++ b/src/SquidConfig.h
@@ -64,7 +64,7 @@ public:
     ~DiskConfig() { delete[] swapDirs; }
 
     RefCount<SwapDir> *swapDirs = nullptr;
-    int n_allocated = 0;
+    unsigned int n_allocated = 0;
     int n_configured = 0;
     /// number of disk processes required to support all cache_dirs
     int n_strands = 0;
--- a/src/pconn.h
+++ b/src/pconn.h
@@ -80,7 +80,7 @@ private:
     Comm::ConnectionPointer *theList_;
 
     /// Number of entries theList can currently hold without re-allocating (capacity).
-    int capacity_;
+    unsigned int capacity_;
     ///< Number of in-use entries in theList
     int size_;
 
--- a/src/pconn.cc
+++ b/src/pconn.cc
@@ -167,7 +167,7 @@ IdleConnList::clearHandlers(const Comm::ConnectionPointer &conn)
 void
 IdleConnList::push(const Comm::ConnectionPointer &conn)
 {
-    if (size_ == capacity_) {
+    if ((unsigned int) size_ == capacity_) {
         debugs(48, 3, "growing idle Connection array");
         capacity_ <<= 1;
         const Comm::ConnectionPointer *oldList = theList_;
--- a/src/store/Disks.cc
+++ b/src/store/Disks.cc
@@ -787,7 +787,7 @@ allocate_new_swapdir(Store::DiskConfig &swap)
         swap.swapDirs = new SwapDir::Pointer[swap.n_allocated];
     }
 
-    if (swap.n_allocated == swap.n_configured) {
+    if (swap.n_allocated == (unsigned int) swap.n_configured) {
         swap.n_allocated <<= 1;
         const auto tmp = new SwapDir::Pointer[swap.n_allocated];
         for (int i = 0; i < swap.n_configured; ++i) {


src/SquidConfig.h
```diff
@@ -64,7 +64,7 @@ class DiskConfig {
     ~DiskConfig() { delete[] swapDirs; }
 
     RefCount<SwapDir> *swapDirs = nullptr;
-    int n_allocated = 0;
+    unsigned int n_allocated = 0;
     int n_configured = 0;
     /// number of disk processes required to support all cache_dirs
     int n_strands = 0;
```
rousskov, line 70: I agree that `n_allocated` type should be unsigned, but we should not make types even less consistent and force type casts to achieve that "unsigned n_allocated" goal because those things make the code even worse overall.

Fixing such problems is often difficult because they tend to snowball (and this is one of those places). No recipe can work in all cases. Often, the following approach helps:

1. Focus on one class at a time (e.g., DiskConfig)
2. Change all the related fields at once (e.g., all `DiskConfig::n_*` data members). Repeated types usually need names, with a few exceptions (e.g., use `size_t` for single-object capacity/counting, like STL does and use `uint64_t` for response content lengths and other things that may exceed a single memory buffer in size)
3. Adjust the rest of the code as needed (e.g., to make it compile) while avoiding C type casts and other "regressions". Sometimes, more variables or data members need to be changed. Sometimes, an occasional static_cast or similar is a good idea to stop the snowball from growing.

If you find changing a lot of similar code (e.g., all the Config.cacheSwap.n_configured iteration loops), then it may be time to pause and refactor `Config.cacheSwap.swapDirs` storage to use an STL container and C++ for-range loops. That is an even bigger change, but it is long overdue and it will change a lot of the same code.

If the changes get too difficult or likely to get too large, please stop and discuss. It is much easier, on many levels, to move forward _before_ a lot of problematic changes have been committed!

src/pconn.h
```diff
@@ -80,7 +80,7 @@ class IdleConnList: public hash_link, private IndependentRunner
     Comm::ConnectionPointer *theList_;
 
     /// Number of entries theList can currently hold without re-allocating (capacity).
-    int capacity_;
+    unsigned int capacity_;
     ///< Number of in-use entries in theList
     int size_;
```
rousskov, line 85: Similar to the other change request, do not let the two related data members to grow apart (and use `size_t` for counting items in a single container):

```suggestion
    size_t capacity_;
    ///< Number of in-use entries in theList
    size_t size_;
```

rousskov (changes requested):
Interesting error! Thank you for trying to improve Squid code while working around alleged GCC bugs.

src/SquidConfig.h
sergiodj: Thank you for the prompt and considerate reply.

I confess I didn't have much time to come up with a better fix, so I apologize.  I understand your rationale and agree with it; my first attempt was to change `n_allocated` et al to `size_t`, but that requires a lot more changes in other places and my main goal was to fix the FTBFS.

Anyway, I will go back to the drawing board and work on a better solution.  This might take some time because I'm very busy with other stuff (we're entering the final stages before the next Ubuntu is released).  Thanks again.

rousskov: Glad we are on the same page! I will mark this PR as a Draft to reflect that you are looking for a better solution (and to slightly reduce CI costs).

There is no rush, of course. And if a quick workaround is needed, we can probably cast/hack something where the allocation is happening without changing the underlying types around the code. Again, please let me know if you need my help with that.

juju4:
is this still under work/consideration?

I am getting this error when compiling squid for tls inspection on ubuntu-24.04 (debian package rebuild) as per https://github.com/juju4/ansible-squid/actions/runs/10429779309/job/28887616336#step:8:1
No issue on 22.04.

Thanks

rousskov:
> is this still under work/consideration?

We still want to improve warning-related code, but it looks like nobody is working on that right now. If you can post an alternative PR (that takes my [review concerns](https://github.com/squid-cache/squid/pull/1118#pullrequestreview-1067599056) into consideration), please do so.

yadij:
FYI; Please check whether you are building with LTO support enabled by default. This `n_allocated` size issue shows up reliably with LTO, you may be able to build fine with LTO disabled.

juju4:
> FYI; Please check whether you are building with LTO support enabled by default. This `n_allocated` size issue shows up reliably with LTO, you may be able to build fine with LTO disabled.

On https://github.com/juju4/ansible-squid/actions/runs/11621187815/job/32364459556, I can see compilation arguments "-flto=auto -ffat-lto-objects" but no other clear logs match. Outside of enabling openssl, the rest of the options are part of default debian package.

yadij:
> arguments "-flto=auto -ffat-lto-objects" 

Okay, that is the known LTO bug. Thank you for working on it. I will adjust the title appropriately.

juju4:
What's the impact of disabling LTO? how to do it? no such option in `./configure --help` and adding `--without-lto` does not change anything and result still in non compilable sources.

src/SquidConfig.h
yadij: This is resolved by [2d232fa](https://github.com/squid-cache/squid/pull/1118/commits/2d232fadbbcc6c65961b6313770cfee113da94ab)

src/pconn.h
yadij: Done in [2d232fa](https://github.com/squid-cache/squid/pull/1118/commits/2d232fadbbcc6c65961b6313770cfee113da94ab)

yadij:
OK to test

src/pconn.h
```diff
@@ -81,9 +81,9 @@ class IdleConnList: public hash_link, private IndependentRunner
     Comm::ConnectionPointer *theList_;
 
     /// Number of entries theList can currently hold without re-allocating (capacity).
-    int capacity_;
+    size_t capacity_ = 0;
     ///< Number of in-use entries in theList
-    int size_;
+    size_t size_ = 0;
```
rousskov, line 86: > yadij requested a review from rousskov

@yadij, you made changes to this PR and requested my review. A few days earlier, a new Squid contributor created #1929 with similar changes that I (partially) reviewed. We were making good progress with #1929 and revealed hidden [problems](https://github.com/squid-cache/squid/pull/1929#discussion_r1825048381) with some of the changes. AFAICT, those problems are now present in this PR as well. To avoid further duplication of effort (at least), please explicitly say which of the two PRs you want us to advance and/or close the PR that you do not want us to advance.

rousskov (changes requested):

juju4:
Confirming that building without LTO is successful on Ubuntu 24.04 as per https://github.com/juju4/ansible-squid/actions/runs/11669115650/job/32490348166#step:8:532 but fails for 22.04 or earlier
I did as gentoo prescribed on https://wiki.gentoo.org/wiki/LTO#Disable_LTO_per_Package with CFLAGS and others.

On LTO, mostly from gcc docs and stackoverflow and matching the "introduced subtle bugs"
https://gcc.gnu.org/onlinedocs/gccint/LTO-Overview.html
https://stackoverflow.com/questions/23736507/is-there-a-reason-why-not-to-use-link-time-optimization-lto

rousskov:
> Confirming that building without LTO is successful on Ubuntu 24.04 as per https://github.com/juju4/ansible-squid/actions/runs/11669115650/job/32490348166#step:8:532

FWIW, that build log shows `-flto=auto` in LDFLAGS. I know that Gentoo [instructions](https://wiki.gentoo.org/wiki/LTO#Disable_LTO_per_Package) you found do not include LDFLAGS, and I do not know whether LDFLAGS alone are enough to enable some LTO aspects despite `-fno-lto` in CXXFLAGS, but it does not look like a _clean_ "without LTO" build to me.

rousskov:
I adjusted PR description to fix formatting, to add the [second known error](https://github.com/squid-cache/squid/pull/1929#issuecomment-2453032166) to the "errors" list, and to restore the original "likely a compiler bug" speculation.

squid-anubis:
Cannot create a git commit message from PR title and description.

Error while parsing future commit message title: `the line is too long 75>72`

Problematic parser input:

    Bug 5160: Compile fails with -flto=auto [-Walloc-size-larger-than=] (#1118)
Please see PR [title](https://github.com/measurement-factory/anubis#pr-title) and [description](https://github.com/measurement-factory/anubis#pr-description) formatting requirements for more details.

This message was added by Anubis bot. Anubis will add a new message if the error text changes. Anubis will remove M-failed-description label when there are no corresponding failures to report.

PR #1119, opened 2022-08-12
Do not tell admins that we are replacing their .* with our .*

rousskov:
acl hasHeaderFoo req_header Foo .*

    2022/08/12 11:43:27| WARNING: regular expression '.*' has only
    wildcards and matches all strings. Using '.*' instead.

Broken since RE optimization inception (commit 6564dae), but the
excessive WARNING was "invisible" to many admins until 61be1d8 started
saving early level-0/1 messages to cache.log.

### DIFF
--- a/src/acl/RegexData.cc
+++ b/src/acl/RegexData.cc
@@ -79,6 +79,9 @@ ACLRegexData::dump() const
 static const char *
 removeUnnecessaryWildcards(char * t)
 {
+    if (strcmp(t, ".*") == 0) // we cannot simplify that further
+        return t; // avoid "WARNING: ... Using '.*' instead" below
+
     char * orig = t;
 
     if (strncmp(t, "^.*", 3) == 0)


kinkie (approved):
Maybe we could review this code to be a bit more generic, but this covers the need so let's ship it

rousskov:
> Maybe we could review this code to be a bit more generic, but this covers the need so let's ship it

Not sure what you mean by generic. FWIW, the official code only does half of what should be done to fulfill its undocumented purpose AFAICT, but those improvements deserve a dedicated PR (and possibly a study of what modern regex compilers can optimize without our clumsy help).

This is just a simple bug fix. I tried, but still do not see how this bug fix can be improved without rewriting the whole function. Thank you for approving!

kinkie:
> > Maybe we could review this code to be a bit more generic, but this covers the need so let's ship it
> 
> Not sure what you mean by generic. FWIW, the official code only does half of what should be done to fulfill its undocumented purpose AFAICT, but those improvements deserve a dedicated PR (and possibly a study of what modern regex compilers can optimize without our clumsy help).

This is exactly what I meant

yadij:
FTR; the purpose of that logic is to prune the more commonly found unnecessary pattern prefixes. Allowing our aggregation to accumulate more patterns together in the ```A|B|C|D``` form before the string capacity is reached. Optimization effects are still left entirely to how the regex compiler handles the aggregate.

PR #1121, opened 2022-08-17
Honor httpd_suppress_version_string in more contexts

rousskov:
* HTML "signature" part of generated HTTP responses to URN requests
* Server header in generated HTTP redirect responses
* boundary strings in generated HTTP 206 multipart/byteranges responses

Also document known contexts outside the configuration directive reach.
We may assert that all of them are outside the current "httpd" scope
(implied by the directive name), but admins may not know how we define
that "httpd" scope; being more explicit may reduce surprises.


yadij (approved):

rousskov:
@kinkie, any idea why Jenkins is not testing this PR staged commit [c81de62](https://github.com/squid-cache/squid/commit/c81de627914ae4a943fbedc10e880555b508a059)? Can you kick Jenkins somehow? It has been more than 24 hours since that commit was staged... Thank you.

PR #1124, opened 2022-08-20
Remove temporary MasterXaction objects

yadij:
The ::Server hierarchy and Foo::NewServer() construction
parameters taking MasterXactionPointer existed only as
preparation for the vetoed PR #865 change.

With that implementation no longer happening this code now
results in unnecessary memory allocation of MasterXaction
objects which will be dropped immediately. Revert to original
(circa Squid-3) construction parameters and remove TODOs which
are no longer be relevant without the rejected design.

### DIFF
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -2162,15 +2162,15 @@ ConnStateData::lifetimeTimeout(const CommTimeoutCbParams &io)
     terminateAll(ERR_LIFETIME_EXP, lte);
 }
 
-ConnStateData::ConnStateData(const MasterXaction::Pointer &xact) :
+ConnStateData::ConnStateData(const AnyP::PortCfgPointer &listener, const Comm::ConnectionPointer &client) :
     AsyncJob("ConnStateData"), // kids overwrite
-    Server(xact)
+    Server(listener, client)
 #if USE_OPENSSL
     , tlsParser(Security::HandshakeParser::fromClient)
 #endif
 {
     // store the details required for creating more MasterXaction objects as new requests come in
-    log_addr = xact->tcpClient->remote;
+    log_addr = client->remote;
     log_addr.applyClientMask(Config.Addrs.client_netmask);
 
     // register to receive notice of Squid signal events
@@ -2312,11 +2312,9 @@ httpAccept(const CommAcceptCbParams &params)
 
     debugs(33, 4, params.conn << ": accepted");
     fd_note(params.conn->fd, "client http connect");
-    const auto xact = MasterXaction::MakePortful(params.port);
-    xact->tcpClient = params.conn;
 
     // Socket is ready, setup the connection manager to start using it
-    auto *srv = Http::NewServer(xact);
+    auto *srv = Http::NewServer(params.port, params.conn);
     // XXX: do not abandon the MasterXaction object
     AsyncJob::Start(srv); // usually async-calls readSomeData()
 }
@@ -2512,14 +2510,11 @@ httpsAccept(const CommAcceptCbParams &params)
         return;
     }
 
-    const auto xact = MasterXaction::MakePortful(params.port);
-    xact->tcpClient = params.conn;
-
     debugs(33, 4, params.conn << " accepted, starting SSL negotiation.");
     fd_note(params.conn->fd, "client https connect");
 
     // Socket is ready, setup the connection manager to start using it
-    auto *srv = Https::NewServer(xact);
+    auto *srv = Https::NewServer(params.port, params.conn);
     // XXX: do not abandon the MasterXaction object
     AsyncJob::Start(srv); // usually async-calls postHttpsAccept()
 }
--- a/src/client_side.h
+++ b/src/client_side.h
@@ -84,7 +84,7 @@ class ConnStateData:
 {
 
 public:
-    explicit ConnStateData(const MasterXactionPointer &xact);
+    explicit ConnStateData(const AnyP::PortCfgPointer &, const Comm::ConnectionPointer &);
     virtual ~ConnStateData();
 
     /* ::Server API */
--- a/src/clients/FtpRelay.cc
+++ b/src/clients/FtpRelay.cc
@@ -238,7 +238,6 @@ Ftp::Relay::updateMaster()
         if (Ftp::Server *srv = dynamic_cast<Ftp::Server*>(mgr.get()))
             return *srv->master;
     }
-    // this code will not be necessary once the master is inside MasterXaction
     debugs(9, 3, "our server side is gone: " << mgr);
     static Ftp::MasterState Master;
     Master = Ftp::MasterState();
--- a/src/comm/TcpAcceptor.cc
+++ b/src/comm/TcpAcceptor.cc
@@ -27,7 +27,6 @@
 #include "ip/Intercept.h"
 #include "ip/QosConfig.h"
 #include "log/access_log.h"
-#include "MasterXaction.h"
 #include "SquidConfig.h"
 #include "StatCounters.h"
 
@@ -288,7 +287,6 @@ Comm::TcpAcceptor::acceptOne()
         /* register interest again */
         debugs(5, 5, "try later: " << conn << " handler Subscription: " << theCallSub);
     } else {
-        // TODO: When ALE, MasterXaction merge, use them or ClientConn instead.
         CallBack(newConnDetails, [&] {
             debugs(5, 5, "Listener: " << conn <<
                    " accepted new connection " << newConnDetails <<
--- a/src/servers/FtpServer.cc
+++ b/src/servers/FtpServer.cc
@@ -50,9 +50,9 @@ static bool SupportedCommand(const SBuf &name);
 static bool CommandHasPathParameter(const SBuf &cmd);
 };
 
-Ftp::Server::Server(const MasterXaction::Pointer &xact):
+Ftp::Server::Server(const AnyP::PortCfgPointer &listenPort, const Comm::ConnectionPointer &client):
     AsyncJob("Ftp::Server"),
-    ConnStateData(xact),
+    ConnStateData(listenPort, client),
     master(new MasterState),
     uri(),
     host(),
@@ -254,11 +254,7 @@ Ftp::Server::AcceptCtrlConnection(const CommAcceptCbParams &params)
     debugs(33, 4, params.conn << ": accepted");
     fd_note(params.conn->fd, "client ftp connect");
 
-    const auto xact = MasterXaction::MakePortful(params.port);
-    xact->tcpClient = params.conn;
-
-    AsyncJob::Start(new Server(xact));
-    // XXX: do not abandon the MasterXaction object
+    AsyncJob::Start(new Server(params.port, params.conn));
 }
 
 void
--- a/src/servers/FtpServer.h
+++ b/src/servers/FtpServer.h
@@ -35,8 +35,6 @@ typedef enum {
     fssError
 } ServerState;
 
-// TODO: This should become a part of MasterXaction when we start sending
-// master transactions to the clients/ code.
 /// Transaction information shared among our FTP client and server jobs.
 class MasterState: public RefCountable
 {
@@ -59,7 +57,7 @@ class Server: public ConnStateData
     CBDATA_CHILD(Server);
 
 public:
-    explicit Server(const MasterXaction::Pointer &xact);
+    explicit Server(const AnyP::PortCfgPointer &, const Comm::ConnectionPointer &);
     virtual ~Server() override;
 
     /* AsyncJob API */
@@ -74,8 +72,8 @@ public:
     /// responding to the FTP client and closing the data connection.
     void stopWaitingForOrigin(int status);
 
-    // This is a pointer in hope to minimize future changes when MasterState
-    // becomes a part of MasterXaction. Guaranteed not to be nil.
+    // This is a pointer in hope to minimize future changes.
+    // Guaranteed not to be nil.
     MasterState::Pointer master; ///< info shared among our FTP client and server jobs
 
 protected:
--- a/src/servers/Http1Server.cc
+++ b/src/servers/Http1Server.cc
@@ -24,9 +24,9 @@
 
 CBDATA_NAMESPACED_CLASS_INIT(Http1, Server);
 
-Http::One::Server::Server(const MasterXaction::Pointer &xact, bool beHttpsServer):
+Http::One::Server::Server(const AnyP::PortCfgPointer &listener, const Comm::ConnectionPointer &client, bool beHttpsServer):
     AsyncJob("Http1::Server"),
-    ConnStateData(xact),
+    ConnStateData(listener, client),
     isHttpsServer(beHttpsServer)
 {
 }
@@ -392,14 +392,14 @@ Http::One::Server::noteTakeServerConnectionControl(ServerConnectionContext serve
 }
 
 ConnStateData *
-Http::NewServer(const MasterXaction::Pointer &xact)
+Http::NewServer(const AnyP::PortCfgPointer &listener, const Comm::ConnectionPointer &client)
 {
-    return new Http1::Server(xact, false);
+    return new Http1::Server(listener, client, false);
 }
 
 ConnStateData *
-Https::NewServer(const MasterXaction::Pointer &xact)
+Https::NewServer(const AnyP::PortCfgPointer &listener, const Comm::ConnectionPointer &client)
 {
-    return new Http1::Server(xact, true);
+    return new Http1::Server(listener, client, true);
 }
 
--- a/src/servers/Http1Server.h
+++ b/src/servers/Http1Server.h
@@ -22,7 +22,7 @@ class Server: public ConnStateData
     CBDATA_CLASS(Server);
 
 public:
-    Server(const MasterXaction::Pointer &xact, const bool beHttpsServer);
+    Server(const AnyP::PortCfgPointer &, const Comm::ConnectionPointer &, const bool beHttpsServer);
     virtual ~Server() {}
 
     void readSomeHttpData();
--- a/src/servers/Server.cc
+++ b/src/servers/Server.cc
@@ -23,11 +23,11 @@
 #include "StatCounters.h"
 #include "tools.h"
 
-Server::Server(const MasterXaction::Pointer &xact) :
+Server::Server(const AnyP::PortCfgPointer &listener, const Comm::ConnectionPointer &client) :
     AsyncJob("::Server"), // kids overwrite
-    clientConnection(xact->tcpClient),
-    transferProtocol(xact->squidPort->transport),
-    port(xact->squidPort),
+    clientConnection(client),
+    transferProtocol(listener->transport),
+    port(listener),
     receivedFirstByte_(false)
 {
     clientConnection->leaveOrphanage();
--- a/src/servers/Server.h
+++ b/src/servers/Server.h
@@ -29,7 +29,7 @@
 class Server : virtual public AsyncJob, public BodyProducer
 {
 public:
-    Server(const MasterXactionPointer &xact);
+    Server(const AnyP::PortCfgPointer &, const Comm::ConnectionPointer &);
     virtual ~Server() {}
 
     /* AsyncJob API */
--- a/src/servers/forward.h
+++ b/src/servers/forward.h
@@ -9,9 +9,8 @@
 #ifndef SQUID_SERVERS_FORWARD_H
 #define SQUID_SERVERS_FORWARD_H
 
-class MasterXaction;
-template <class C> class RefCount;
-typedef RefCount<MasterXaction> MasterXactionPointer;
+#include "anyp/forward.h"
+#include "comm/forward.h"
 
 class ConnStateData;
 
@@ -24,7 +23,7 @@ class Server;
 } // namespace One
 
 /// create a new HTTP connection handler; never returns NULL
-ConnStateData *NewServer(const MasterXaction::Pointer &xact);
+ConnStateData *NewServer(const AnyP::PortCfgPointer &, const Comm::ConnectionPointer &);
 
 } // namespace Http
 
@@ -32,7 +31,7 @@ namespace Https
 {
 
 /// create a new HTTPS connection handler; never returns NULL
-ConnStateData *NewServer(const MasterXaction::Pointer &xact);
+ConnStateData *NewServer(const AnyP::PortCfgPointer &, const Comm::ConnectionPointer &);
 
 } // namespace Https
 


src/client_side.cc
```diff
@@ -2162,15 +2162,15 @@ ConnStateData::lifetimeTimeout(const CommTimeoutCbParams &io)
     terminateAll(ERR_LIFETIME_EXP, lte);
 }
 
-ConnStateData::ConnStateData(const MasterXaction::Pointer &xact) :
+ConnStateData::ConnStateData(const AnyP::PortCfgPointer &listener, const Comm::ConnectionPointer &client) :
     AsyncJob("ConnStateData"), // kids overwrite
```
rousskov, line 2166: ConnStateData should continue to be created with a MasterXaction object. The correct solution to the "create and immediately drop MasterXaction" problem is not removing the _creation_ of that object but eliminating its premature _destruction_: The MasterXaction object should be created ASAP[^1] and should be used for the first master transaction on that freshly accepted connection (instead of being discarded).

[^1]: It should probably be created even earlier than it is created now, but that it a separate issue. Conceptually, the first master transaction on a connection starts with the acceptance of that connection. The timing of the MasterXaction creation affects debugging, response time and other performance statistics, ACL checks, etc.

rousskov (changes requested):
This review focuses on the proposed code changes. It does not indicate an agreement with the statements in the current PR description.

src/client_side.cc
yadij: > ConnStateData should continue to be created with a MasterXaction object.

No it should not. As explained in the description this API **only** existed as preparation for PR #865 which was vetoed.

> The correct solution to the "create and immediately drop MasterXaction" problem is not removing the _creation_ of that object but eliminating its premature _destruction_:

Use of this class is deeply buried in controversy. All purpose this logic was intended to have is gone now. Whatever design you may be planning for future can use another object and/or name. Best to drop this logic and the controversy along with it.


> The MasterXaction object should be created ASAP and should be used for the first master transaction on that freshly accepted connection (instead of being discarded).
> 
>     1. It should probably be created even earlier than it is created now,

It **was** created on connection acceptance. Until ad05b958fa522d73023eebf37dbfbf57829a9507
> > "We moved MasterXaction creation to TcpAcceptor
callbacks because only they know whether accepting a TCP connection
implies a new master transaction start (it does not for FTP DATA cases)."

> 
> but that it a separate issue. Conceptually, the first master transaction on a connection starts with the acceptance of that connection.

Sigh. I refer back to PR #865 discussions:

> @rousskov 23 Jul 2021:
>> "each browser (HTTP) request inside a session should create a MasterXaction object. Similarly, each FTP command received by Squid inside an FTP session should create a MasterXaction object. Still roughly speaking, each access.log record corresponds to a unique master transaction. One browsing session may correspond to numerous access.log records."

By that definition MasterXaction **cannot** exist until an HTTP (or FTP) protocol message is received. One TCP connection may contain one or more "browsing session"s therefore multiple MasterXaction - all of which begins long **after** the logic being touched in this PR and is represented by other MasterXaction instances allocated elsewhere.

rousskov: > > Alex: ConnStateData should continue to be created with a MasterXaction object.
 
> Amos: No it should not. As explained in the description this API **only** existed as preparation for PR 865

I disagree on both counts, and I have provided specific (and unrelated to PR 865) reasons for that object early creation.

> > The correct solution to the "create and immediately drop MasterXaction" problem is not removing the _creation_ of that object but eliminating its premature _destruction_:

> Use of this class is deeply buried in controversy.

Yes, unfortunately. There are no technical reasons for that controversy IMO, but our inability to agree on (or even productively discuss) such fundamentals impedes Squid improvement.

> All purpose this logic was intended to have is gone now.

I disagree. And just to avoid misunderstanding: If you are talking about _your_ intentions for this logic, then you ought to be correct, of course, but this class and this code represents a lot more than your intentions. Again, I have provided specific reasons for early MasterXaction creation (which the current official code more-or-less implements).

> Whatever design you may be planning for future can use another object and/or name. Best to drop this logic and the controversy along with it.

If I understand you correctly, you are proposing something like the following sequence of steps:

1. Amos: Removes official code that creates a MasterXaction early. This PR proposes such a removal.
2. Somebody: Proposes a replacement for the MasterXaction name/object. The new class will be called X.
3. Alex: Proposes code that creates X early.

I see no benefit in doing steps 1 and 3. We can just do step 2. If step 2 succeeds, step 1 and step 3 become unnecessary. If step 2 fails, leaving bare step 1 (without a compensating step 3) would makes things worse, not better. In summary, I do not see why you think this sequence is "best", but perhaps you have a different sequence of steps in mind.

> > The MasterXaction object should be created ASAP and should be used for the first master transaction on that freshly accepted connection (instead of being discarded).
> > Footnote: It should probably be created even earlier than it is created now,
 
> It **was** created on connection acceptance. Until [ad05b95](https://github.com/squid-cache/squid/commit/ad05b958fa522d73023eebf37dbfbf57829a9507)

Yes, it was. I do not see any contradiction between what my ASAP goal, the footnote, and ad05b95. As you know, that commit explains why the master transaction object creation had to be moved. There are ways to move that creation earlier, but they require significant API changes that were outside ad05b95 scope. At any rate, the move in ad05b95 was minor compared to the significant difference between ConnStateData creation and the second MasterXaction object creation in Http::One::Server::buildHttpRequest() (that should not happen for the first request -- the MasterXaction object created earlier should be reused instead). We should be working on eliminating that second creation instead.


> > Alex: Conceptually, the first master transaction on a connection starts with the acceptance of that connection.

> Amos: Sigh. I refer back to PR #865 discussions:
 
> > > Alex at https://github.com/squid-cache/squid/pull/865#discussion_r675117801 : A browsing session usually involves multiple MasterXaction objects. For example, each browser (HTTP) request inside a session should create a MasterXaction object. Similarly, each FTP command received by Squid inside an FTP session should create a MasterXaction object. Still roughly speaking, each access.log record corresponds to a unique master transaction. One browsing session may correspond to numerous access.log records.

> Amos: By that definition MasterXaction **cannot** exist until an HTTP (or FTP) protocol message is received. 

No, there is no "until message is (fully) received" implication. In the above quote, I was rejecting the "master transaction is a browsing session" idea by providing a very rough outline for what should be happening inside a browsing session. In that context, I was focusing on the existence of many "client requests", "master transactions", and "access.log records" inside one session. I was _not_ defining or even describing the exact moment objects representing each "client request", "master transaction", or "access.log record" are created!

If we start talking about the exact moment when the very first transaction on an HTTP connection starts, then that moment comes way before we receive the HTTP protocol message! For example:

* If a client sends a truncated HTTP request, there are no HTTP protocol messages at all (from the high-level code point of view), but there is (or should be) a master transaction, an HTTP transaction, and an access.log record.
* If a client takes 1 minute to complete an HTTP request header, there is no HTTP protocol message _during_ that minute, but there is (or should be) a master transaction, an HTTP transaction, and a later access.log record should reflect that minute.

> One TCP connection may contain one or more "browsing session"s therefore multiple MasterXaction - all of which begins long **after** the logic being touched in this PR and is represented by other MasterXaction instances allocated elsewhere.

I agree that multiple master transactions inside a single TCP connection are possible, of course, but this specific discussion is about the very _first_ among those transactions. That first transaction starts (or should start) ASAP after the TCP connection acceptance, not "long after". Once the connection is accepted, we ought to start various response time clocks and be ready to check ACLs, provide transaction-specific timeout debugging, and add an access.log record. All that and more with or without additional traffic on that TCP connection! To do all of that well, we need a master transaction object. We have it today. We just do not propagate it to the request parsing/handling code correctly.

yadij: > > > Alex: ConnStateData should continue to be created with a MasterXaction object.
> 
> > Amos: No it should not. As explained in the description this API **only** existed as preparation for PR 865
> 
> I disagree on both counts, and I have provided specific (and unrelated to PR 865) reasons for that object early creation.
> 
> > > The correct solution to the "create and immediately drop MasterXaction" problem is not removing the _creation_ of that object but eliminating its premature _destruction_:
> 
> > Use of this class is deeply buried in controversy.
> 
> Yes, unfortunately. There are no technical reasons for that controversy IMO, but our inability to agree on (or even productively discuss) such fundamentals impedes Squid improvement.
> 
> > All purpose this logic was intended to have is gone now.
> 
> I disagree. And just to avoid misunderstanding: If you are talking about _your_ intentions for this logic, then you ought to be correct, of course, but this class and this code represents a lot more than your intentions. Again, I have provided specific reasons for early MasterXaction creation (which the current official code more-or-less implements).
> 
> > Whatever design you may be planning for future can use another object and/or name. Best to drop this logic and the controversy along with it.
> 
> If I understand you correctly, you are proposing something like the following sequence of steps:
> 
>     1. Amos: Removes official code that creates a MasterXaction early. This PR proposes such a removal.
>

That is all I am proposing.

 
>     2. Somebody: Proposes a replacement for the MasterXaction name/object. The new class will be called X.
> 

I did that some time ago in PR #823. You objected as usual.


>     3. Alex: Proposes code that creates X early.
> 

You have already done step 2+3 yourself. PR #1103 essentially makes X = Comm::Connection and adds Comm::Descriptor for the TCP transaction state.


> > > The MasterXaction object should be created ASAP and should be used for the first master transaction on that freshly accepted connection (instead of being discarded).
> > > Footnote: It should probably be created even earlier than it is created now,
> 
> > It **was** created on connection acceptance. Until [ad05b95](https://github.com/squid-cache/squid/commit/ad05b958fa522d73023eebf37dbfbf57829a9507)
> 
> Yes, it was. I do not see any contradiction

Of course you don't.  **contradiction** is not relevant.

> 
> > > Alex: Conceptually, the first master transaction on a connection starts with the acceptance of that connection.
> 
> > Amos: Sigh. I refer back to PR #865 discussions:
> 
> > > > Alex at [#865 (comment)](https://github.com/squid-cache/squid/pull/865#discussion_r675117801) : A browsing session usually involves multiple MasterXaction objects. For example, each browser (HTTP) request inside a session should create a MasterXaction object. Similarly, each FTP command received by Squid inside an FTP session should create a MasterXaction object. Still roughly speaking, each access.log record corresponds to a unique master transaction. One browsing session may correspond to numerous access.log records.
> 
> > Amos: By that definition MasterXaction **cannot** exist until an HTTP (or FTP) protocol message is received.
> 
> No, there is no "until message is (fully) received" implication.

I did not write "fully". To be clear I mean simply **received*, somewhere in/after the read(2) callback.

Before read(2) Squid has zero information that your definition of "master transaction" exists.  The TCP connection may be closed unused (eg Happy Eyeballs backup connection) and is represented by Comm::Connection being orphaned **not** allocating a "master transaction" object.

> I was _not_ defining or even describing the exact moment objects representing each "client request", "master transaction", or "access.log record" are created!

The above statement tells me you are still misunderstanding the concept "class MasterXaction" was designed to implement. You are still talking in some completely different paradigm where those objects are different things.

> 
> If we start talking about the exact moment when the very first transaction on an HTTP connection starts, then that moment comes way before we receive the HTTP protocol message! For example:

Conflating paradigms. That is an "HTTP transaction" which can be a sub-transaction, therefore not "the master transaction".

> 
>     * If a client sends a truncated HTTP request, there are no HTTP protocol messages at all (from the high-level code point of view), but there is (or should be) a master transaction, an HTTP transaction, and an access.log record.

FTR; That is an implementation state resulting from your rejection of the design from PR #865 and its followup work.
As stated in the description this PR is a rollback of the incomplete now vetoed design where "TCP transaction" and "HTTP Transaction" are both sharing a MasterXaction instance.

> 
>     * If a client takes 1 minute to complete an HTTP request header, there is no HTTP protocol message _during_ that minute, but there is (or should be) a master transaction, an HTTP transaction, and a later access.log record should reflect that minute.
> 

The conclusion reached does not follow from those facts stated.

Per your definition of "master transaction" imposed on PR #865: The "master transaction" should begin when the first read(2) receives bytes that are expected to be HTTP.

> 
> > One TCP connection may contain one or more "browsing session"s therefore multiple MasterXaction - all of which begins long **after** the logic being touched in this PR and is represented by other MasterXaction instances allocated elsewhere.
> 
> I agree that multiple master transactions inside a single TCP connection are possible, of course, but this specific discussion is about the very _first_ among those transactions. 

Indeed. The code being touched by this PR has exactly two cases:

 1) first "master transaction" does not exist yet.
  * Per your definition of "master transaction" being relative to HTTP requests (used to oppose PR #865 and #716)  this case has two forms:
    i) MasterXaction will be generated by the Server class (eg ConnStateData) once a "first HTTP message" is read.
    ii) there will never be a "master transaction" for non-HTTP protocols that do not get mapped to an HttpRequest or AccessLogEntry by Squid..

 2) there is some other "master transaction" for which this TCP connection **is a child**.
  * Per definition of master-of-master being invalid "master transaction" type/relation (used to oppose PR #716) allocating a MasterXaction for this callback is invalid.


If you object to this logic you must have changed your opposition to one or more of the previous PR designs. Please clarify which and we can take those PR(s) forward.

rousskov: TLDR: This discussion has failed to advance this PR and consumes a lot of valuable time. I suggest discussing these issues live, in a Zoom call or on Zulip. Please ping me with time suggestions if you agree. I can also offer to submit a PR that documents MasterXaction design (where many of these high-level issues can be settled). Details are below.

> > Amos: Whatever design you may be planning for future can use another object and/or name. Best to drop this logic and the controversy along with it.

> > 1. Amos: Removes official code that creates a MasterXaction early. This PR proposes such a removal.
 
> That is all I am proposing.

AFAICT, you are _also_ proposing how to address my current concerns (about this PR changes) in future PRs (see the first paragraph quoted above). Sometimes, introducing temporary problems and then solving introduced and bigger problems later is the best way forward. In this particular case, that plan does not work as I have detailed in my previous response. Again, I am not implying that is _your_ plan -- I am just _guessing_ because you have not detailed how to address my concerns besides saying "best drop now" (and merge this PR which, by itself, obviously does not address my concerns).

FWIW, with this three-step list I was just trying to flesh out your "drop now (and use another object in the future)" proposal. I still do not know for sure (whether that outline matched) what you had in mind for moving forward here. I am briefly responding to your step-specific comments, but please keep that bigger goal in mind when deciding how to proceed from here.


> > 2. Somebody: Proposes a replacement for the MasterXaction name/object. The new class will be called X.
 
> I did that some time ago in PR #823. You objected as usual.

While that specific proposal could not be accepted, I am sure it is possible to find an acceptable step 2 implementation if really needed. I doubt any of these steps are _needed_, but that is another question.


> > 3. Alex: Proposes code that creates X early.

> You have already done step 2+3 yourself. PR #1103 essentially makes X = Comm::Connection and adds Comm::Descriptor for the TCP transaction state.

No, Comm::Connection is not a master transaction replacement and will never be. The two concepts are very different.


> > > > Alex: Conceptually, the first master transaction on a connection starts with the acceptance of that connection.

> > > Amos: Sigh. I refer back to PR #865 discussions:

> > > > > Alex at [#865 comment](https://github.com/squid-cache/squid/pull/865#discussion_r675117801): A browsing session usually involves multiple MasterXaction objects. For example, each browser (HTTP) request inside a session should create a MasterXaction object. Similarly, each FTP command received by Squid inside an FTP session should create a MasterXaction object. Still roughly speaking, each access.log record corresponds to a unique master transaction. One browsing session may correspond to numerous access.log records.

> > > Amos: By that definition MasterXaction **cannot** exist until an HTTP (or FTP) protocol message is received.

> > No, there is no "until message is (fully) received" implication.
 
> I did not write "fully".

There is no "until message is received" implication. This statement covers _both_ "fully received" and "partially received" cases. Sorry for being too cryptic when expressing this idea in my previous attempt.

>  To be clear I mean simply *_received_, somewhere in/after the read(2) callback.

That is still not clear to me, but it sounds like you mean something like "... until at least the first byte of the ... protocol message is read or the read(2) failed, whichever comes first". I disagree that receiving the first byte (or read(2) failure) is the right event for the master transaction creation. We need a master transaction to start receiving that first byte (i.e. start waiting for that first byte to be available for reading).


> Before read(2) Squid has zero information that your definition of "master transaction" exists.

Not sure what you mean by "has zero information that a ... definition exists", but Squid can (and does and should) create a master transaction object "before read(2)". At the very least, Squid knows which TCP connection that master transaction is going to use, how long Squid has been waiting for the request, etc. All that information is (or should be) associated with a master transaction object that Squid can and should create ASAP after successfully accepting a TCP connection on an http_port (or equivalent).


> The TCP connection may be closed unused and is represented by Comm::Connection being orphaned **not** allocating a "master transaction" object.

I disagree with the "represented by..." part. Unused connections need a master transaction object to be handled correctly (as detailed earlier). While a master transaction object does not really "represent" any single connection (master transaction is a higher-level abstraction), the master transaction object is still required (along with a Comm::Connection object and other details) for that handling.

> > I was _not_ defining or even describing the exact moment objects representing each "client request", "master transaction", or "access.log record" are created!

> The above statement tells me you are still misunderstanding the concept "class MasterXaction" was designed to implement. You are still talking in some completely different paradigm where those objects are different things.

Bugs notwithstanding, I believe MasterXaction design matches my statements about it. You are stating that it does not match. To make progress on that level, we need to agree on what MasterXaction should represent -- that is why I am (repeatedly) documenting MasterXaction properties/invariants (that I believe in) and providing examples to illustrate them. That technical information can be used to make progress. On the contrary, asserting that I do not understand the MasterXaction _concept_ does _not_ help us make progress (regardless of whether that assertion is correct). It only introduces additional non-technical barriers to cooperation! Two people can assert that the other person does not understand some concept while making zero (or negative) progress, forever. Let's not be those people.

The only way to convince me that we should proceed with _your_ understanding of MasterXaction design is to define what MasterXaction is (in your opinion) and then address my concerns about that definition so that we arrive at a definition we all agree with. That has not happened yet. You do not have to do that, of course, and I am not asking you to do that. Instead, we can do other things, including:

* Continue making low-level changes (like this PR is attempting to do). However, you cannot address my concerns about those low-level changes simply by referring to the MasterXaction design. That argument is inadmissible because our MasterXaction design understandings differ. With this option, I will naturally make/review changes based on my understanding of that design, and you will naturally make/review changes based on your understanding of that design. Progress will be painful but still possible in _some_ low-level cases (this PR is probably not one of those).

* Ask me to document my understanding of MasterXaction and review that (based on the merits of the documented ideas/approaches rather than what you think has been said or implied about MasterXaction in the past). I have pretty much doing bits of this option in various PRs, but I can open a _dedicated_ documentation-only PR instead and try to come up with a comprehensive MasterXaction description that we both can (eventually, after discussions and changes) agree on. Reaching consensus may be painful but, if we succeed, we can move much faster with related low-level issues (like this PR).

If you want to work on code modifying MasterXaction and its uses in significant ways and am willing to cooperate on documenting MasterXaction-related design, then I recommend the option described in the last bullet above. Please let me know if I should start working on that documentation PR.


> > If we start talking about the exact moment when the very first transaction on an HTTP connection starts, then that moment comes way before we receive the HTTP protocol message! For example:
 
> Conflating paradigms. That is an "HTTP transaction" which can be a sub-transaction, therefore not "the master transaction".

I do not know what a sub-transaction is, but there is no conflation in what I said: There are no (or should be no) HTTP transactions without a master transaction. Every HTTP transaction is (or should be) a part of some master transaction. Thus, a master transaction object must already exist when (or should be created just before) an HTTP transaction starts.

The relationship between master transaction and the very first HTTP transaction on an accepted TCP connection is very similar to the relationship between a human and a human head -- even before an obstetrician sees the baby head crowning for the first time, there is already a record of that human in the hospital database, in government documents, in parents minds, etc. The MasterXaction class represents that record (and, to some extent, that human as a whole).


> > * If a client sends a truncated HTTP request, there are no HTTP protocol messages at all (from the high-level code point of view), but there is (or should be) a master transaction, an HTTP transaction, and an access.log record.
 
> FTR; That is an implementation state resulting from your rejection of the design from PR #865 and its followup work. As stated in the description this PR is a rollback of the incomplete now vetoed design where "TCP transaction" and "HTTP Transaction" are both sharing a MasterXaction instance.

FTR: The idea that a MasterXaction instance can refer (directly or indirectly) to a TCP connection and an HTTP transaction has never been vetoed by me.

We are unlikely to make progress by appealing to something as vast/complex/vague/undefined as "design from PR 865" or its "rejection": This PR description says a lot of things; I disagree with some of them. Past proposals contained a lot of things; I disagreed with some of them. Past vetoes were based on many things, and you probably disagree with some of them. I believe my statements in this discussion are accurate. If you disagree and want to change my mind, please point out specific flaws without referring to such vast/complex/vague/undefined concepts as a design from a rejected PR and its followup work.

> > * If a client takes 1 minute to complete an HTTP request header, there is no HTTP protocol message _during_ that minute, but there is (or should be) a master transaction, an HTTP transaction, and a later access.log record should reflect that minute.

> The conclusion reached does not follow from those facts stated.

It is just an _example_ illustrating the design I am describing. It is not a proof or conclusion but an illustration.

> Per your definition of "master transaction" imposed on PR #865: The "master transaction" should begin when the first read(2) receives bytes that are expected to be HTTP.

There is no such definition (and, hence, it was not imposed). The example from PR 865 [discussion](https://github.com/squid-cache/squid/pull/865#discussion_r675117801) that you have cited earlier does not provide such a definition (as I have explained already). If you think you found some other specific sources of such a definition, feel free to cite and discuss them, but it may be more efficient to just assume that you have misinterpreted them (like you have misinterpreted the already discussed example) and move on.


> > > One TCP connection may contain one or more "browsing session"s therefore multiple MasterXaction - all of which begins long **after** the logic being touched in this PR and is represented by other MasterXaction instances allocated elsewhere.

> > I agree that multiple master transactions inside a single TCP connection are possible, of course, but this specific discussion is about the very _first_ among those transactions.

> Indeed. The code being touched by this PR has exactly two cases:
> 
> 1. first "master transaction" does not exist yet.
> 
> * Per your definition of "master transaction" being relative to HTTP requests (used to oppose PR [Extend MasterXaction lifetime across ::Server existence #865](https://github.com/squid-cache/squid/pull/865) and [WIP: Support for layered transport state #716](https://github.com/squid-cache/squid/pull/716))  this case has two forms:
> i) MasterXaction will be generated by the Server class (eg ConnStateData) once a "first HTTP message" is read.

There is no such definition and/or implication.

> ii) there will never be a "master transaction" for non-HTTP protocols that do not get mapped to an HttpRequest or AccessLogEntry by Squid..

There is no such implication. Depending on what you mean by "mapped", this statement might be true for the current Squid code, but it does not relay MasterXaction design well. MasterXaction design does not preclude non-HTTP(-mapped) protocol transactions from being associated with a MasterXaction. I do not think we need to detail/decide this now, but I see nothing prohibiting such a design in principle.

> 2. there is some other "master transaction" for which this TCP connection **is a child**.

There is no such implication AFAICT, but I am not sure what you mean by "TCP connection is a child of a master transaction" -- the two concepts are not connected by an "X is a child of Y" relationship, regardless of which of them is X and Y.

> * Per definition of master-of-master being invalid "master transaction" type/relation (used to oppose PR [WIP: Support for layered transport state #716](https://github.com/squid-cache/squid/pull/716)) allocating a MasterXaction for this callback is invalid.

There is no such implication.


> If you object to this logic you must have changed your opposition to one or more of the previous PR designs. Please clarify which and we can take those PR(s) forward.

I am not aware of any changes on my side. AFAICT, you are seeing implications that I do not see, but I cannot pinpoint the failure because you are referring to vast/complex/vague/undefined concepts as the source of those implications. I have tried to suggest specific ways forward above.

yadij: > FWIW, with this three-step list I was just trying to flesh out your "drop now (and use another object in the future)" proposal. I still do not know for sure (whether that outline matched) what you had in mind for moving forward here

To be clear ... I do not have any plans past "drop now" for this touched code.

[ elided my re-response to the rest of your response since it is very clear we are again talking about different concepts ]

rousskov: > I do not have any plans past "drop now" for this touched code.

Understood. The "drop now" part alone is obviously insufficient to address this change request, and (as detailed earlier in this thread) I do not see a good way forward if the first step is "drop now".

PR #1125, opened 2022-08-22
Fixed "make check" in a --disable-auth-basic build

rousskov:
FATAL: Bungled squidconf/bug4832 line 1: auth_param basic /dev/null
    FAIL: squid.conf test: squidconf/bug4832

Also added initial support for test-suite/squidconf/ tests that depend
on ./configure options/results. This framework will be extended to
support testing with invalid configuration files that Squid must reject.

Existing Squid configuration files in test-suite/squidconf/ were given a
.conf filename extension for ease of auto-distinguishing them from
.instructions files that contain testing "rules". The customary .conf
extension also improves handling of those files by humans.


test-suite/Makefile.am
```diff
@@ -133,10 +134,10 @@ VirtualDeleteOperator_SOURCES = \
 	VirtualDeleteOperator.cc \
 	stub_libmem.cc
 
-squid-conf-tests: $(top_builddir)/src/squid.conf.default $(srcdir)/squidconf/*
+squid-conf-tests: $(top_builddir)/src/squid.conf.default $(srcdir)/squidconf/*.conf
```
rousskov, line 137: This dependency change is wrong: The test should also be redone if any of the `.instructions` files change. I want to  restore/fix this dependency to include all files (and also include the new test-squid-conf.sh script), but that will require more changes like skipping non-config files in the loop below.

However, before I do all that and make the diff look more complicated, I wanted to get the general feedback about this PR overall direction _and_ to ask whether I should move the `echo FAIL/PASS` functionality below to the new test-squid-conf.sh script as well -- it is much easier to write that code in shell than in shell-inside-Makefile-rule, of course, but it will increase the diff further.

test-suite/test-squid-conf.sh
```diff
@@ -0,0 +1,74 @@
...
+                echo "$here: ERROR: Cannot determine status of $defineName macro";
+                exit 1;
+            fi
+        else
+            echo "$here: ERROR: Unknown test-squid-conf.sh instruction name: $instructionName";
+            exit 1;
+        fi
+    done < $instructionsFile
+
+    # TODO: Add support for the "require-failure" instruction.
```
rousskov, line 71: I can add this support in _this_ PR if others think it is a good idea. I tried to minimize this PR scope.

test-suite/squidconf/bug4832.conf.instructions
```diff
@@ -0,0 +1 @@
+skip-unless-autoconf-defines HAVE_AUTH_MODULE_BASIC
```
rousskov, line 1: I am not aware of other dependencies like this in the _existing_ test-suite/squidconf/ configuration files. If you know of any, please share them, and I will add them to this PR (if it is not going to blow up its scope).

rousskov (commented):
I am posting this draft PR to ask the questions in the change requests below.

test-suite/test-squid-conf.sh
```diff
@@ -0,0 +1,74 @@
...
+##
+
+# Orchestrates a "squid -k parse ..." test of a single Squid configuration
+# file (with an optional .instructions file containing testing directions).
+# Usage: test-squid-conf.sh <top_builddir> <squid.conf>
+
+top_builddir=$1
+configFile=$2
+
+instructionsFile="$configFile.instructions"
```
yadij, line 17: Why not add instructions as a comment to the top of the config file itself? save on all the special case handling for .instructions files.

yadij (changes requested):

test-suite/test-squid-conf.sh
rousskov: > Why not add instructions as a comment to the top of the config file itself?

I started with that design but eventually rejected it after realizing that combining test metadata (i.e. configuration testing instructions) and data (i.e. the configuration itself) creates more problems than it solves:

* The combination requires special syntax that would allow the testing script to separate metadata from data. There are many ways to do that, with various pros and cons, but even the best segregation approaches require _additional_ parsing/handling in the testing script (that we want to keep simple but which language is not designed for parsing), additional documentation/training, and are still likely to trigger debates about their specific merits (e.g., "why did not you use a special comment symbol/prefix instead of a special section separator?"). It is overall easier/cheaper to separate using files.
* The combination should result in creation of temporary configuration files (the ones without metadata). Those files require _additional_ cleanup code and complicates failure triage/reproduction. [WYSIWYG](https://en.wikipedia.org/wiki/WYSIWYG) is an overall better approach when it comes to such basic test data.
* Combining different syntax, different formatting, different spelling rules, etc. in one file often complicates human editing and automated formatting/validation.
* Restricting test cases to `*.conf` files is arguably an _improvement_ in itself -- the existing `*` approach is arguably too broad/rigid/risky.
* The additional work required to add proper Makefile handling for `.instructions` files is essentially a one-time investment while the extra work caused by mixing metadata and data is likely to be ongoing.

> save on all the special case handling for .instructions files.

The overall net difference would be a loss IMO.

yadij: Okay, thank you for the details.

yadij (approved):

test-suite/Makefile.am
rousskov: > @rousskov If there are no more big changes from you I think this can be made a non-Draft and marked for Anubis to merge.

Thank you but the dependency problem described in this change request should be fixed first. There is no reason to introduce this bug in this PR. Since you like the overall PR approach, I can now focus on this fix. I will remove Draft designation once I push that fix.

rousskov: > the dependency problem described in this change request should be fixed first.

Done in 8bf0505. Will merge in a few days unless there are any last-minute objections. Please feel free to clear for merge sooner.

PR #1127, opened 2022-08-24
Stop (ab)using Transient entry flags for CF requirement mngmt

rousskov:
The ENTRY_REQUIRES_COLLAPSING flag was used for relaying "hitting this
entry carries Collapsed Forwarding risks right now" info to other
workers. That flag is not necessary to relay that info because remote
workers already know whether the entry is attached to one of the shared
cache stores. Moreover, relaying that flag correctly is difficult and
would require a different implementation:

* The flag is changed in relatively high-level Store code, possibly
  _after_ we stopped writing, when the transients entry is already
  read-only. This bug leads to an isWriter() assertion in
  Transients::clearCollapsingRequirement() called via a quick or aborted
  StoreEntry::startWriting().

* Basics.flags is not an atomic field. It lacks an "[app]" mark -- its
  readers will not get a reliable value due to race conditions when
  Squid will need to update the Transient entry flags in "append" mode.

* The flag is changed assuming that simply sending stuff to disk is
  sufficient for other workers to read it. Since 18102f7, this is
  inaccurate because the store index is updated only after the slot is
  written to disk.

Remote workers now use Store attachment info to set the local StoreEntry
object flag, ignoring the no-longer-used Transients entry flags.

The ENTRY_REQUIRES_COLLAPSING flag is still used to tell transactions
which StoreEntry objects are subject to CF risks. It would be nice to
use the presence of MemObject::reply_ to relay the same info for
store-unattached entries (and attachment info for others), but we could
not find a way to do that (during commit d2a6dcb work and then again
during this project) without a lot of essentially out of scope rewrites.

Also removed misleading Ipc::StoreMapAnchor::exportInto() debugging.

### DIFF
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -164,15 +164,6 @@ Transients::get(const cache_key *key)
         return nullptr;
     }
 
-    // store hadWriter before checking ENTRY_REQUIRES_COLLAPSING to avoid racing
-    // the writer that clears that flag and then leaves
-    const auto hadWriter = map->peekAtWriter(index);
-    if (!hadWriter && EBIT_TEST(anchor->basics.flags, ENTRY_REQUIRES_COLLAPSING)) {
-        debugs(20, 3, "not joining abandoned entry " << index);
-        map->closeForReadingAndFreeIdle(index);
-        return nullptr;
-    }
-
     StoreEntry *e = new StoreEntry();
     e->createMemObject();
     anchorEntry(*e, index, *anchor);
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -304,7 +304,6 @@ Transients::status(const StoreEntry &entry, Transients::EntryStatus &entryStatus
                          map->writeableEntry(idx) : map->readableEntry(idx);
     entryStatus.hasWriter = anchor.writing();
     entryStatus.waitingToBeFreed = anchor.waitingToBeFreed;
-    entryStatus.collapsed = EBIT_TEST(anchor.basics.flags, ENTRY_REQUIRES_COLLAPSING);
 }
 
 void
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -198,21 +198,6 @@ Transients::findCollapsed(const sfileno index)
 }
 
 void
-Transients::clearCollapsingRequirement(const StoreEntry &e)
-{
-    assert(map);
-    assert(e.hasTransients());
-    assert(isWriter(e));
-    const auto idx = e.mem_obj->xitTable.index;
-    auto &anchor = map->writeableEntry(idx);
-    if (EBIT_TEST(anchor.basics.flags, ENTRY_REQUIRES_COLLAPSING)) {
-        // XXX: Non-atomic, [app]-unmarked flags are off limits in append mode.
-        EBIT_CLR(anchor.basics.flags, ENTRY_REQUIRES_COLLAPSING);
-        CollapsedForwarding::Broadcast(e);
-    }
-}
-
-void
 Transients::monitorIo(StoreEntry *e, const cache_key *key, const Store::IoStatus direction)
 {
     if (!e->hasTransients()) {
@@ -292,11 +277,7 @@ Transients::anchorEntry(StoreEntry &e, const sfileno index, const Ipc::StoreMapA
     xitTable.index = index;
     xitTable.io = Store::ioReading;
 
-    const auto hadWriter = hasWriter(e); // before computing collapsingRequired
     anchor.exportInto(e);
-    const bool collapsingRequired = EBIT_TEST(anchor.basics.flags, ENTRY_REQUIRES_COLLAPSING);
-    assert(!collapsingRequired || hadWriter);
-    e.setCollapsingRequirement(collapsingRequired);
 }
 
 bool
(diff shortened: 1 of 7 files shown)


src/Transients.cc
```diff
@@ -197,21 +197,6 @@ Transients::findCollapsed(const sfileno index)
...
-{
-    assert(map);
-    assert(e.hasTransients());
-    assert(isWriter(e));
-    const auto idx = e.mem_obj->xitTable.index;
-    auto &anchor = map->writeableEntry(idx);
-    if (EBIT_TEST(anchor.basics.flags, ENTRY_REQUIRES_COLLAPSING)) {
-        // XXX: Non-atomic, [app]-unmarked flags are off limits in append mode.
-        EBIT_CLR(anchor.basics.flags, ENTRY_REQUIRES_COLLAPSING);
-        CollapsedForwarding::Broadcast(e);
```
rousskov, line 211: #1117 identified and promised to address this (out-of-scope for that PR) XXX.

src/Transients.cc
```diff
@@ -292,11 +277,7 @@ Transients::anchorEntry(StoreEntry &e, const sfileno index, const Ipc::StoreMapA
     xitTable.index = index;
     xitTable.io = Store::ioReading;
 
-    const auto hadWriter = hasWriter(e); // before computing collapsingRequired
     anchor.exportInto(e);
-    const bool collapsingRequired = EBIT_TEST(anchor.basics.flags, ENTRY_REQUIRES_COLLAPSING);
-    assert(!collapsingRequired || hadWriter);
-    e.setCollapsingRequirement(collapsingRequired);
 }
```
rousskov, line 281: FWIW, this whole method will be removed in the followup PR (addressing another bug).

rousskov (commented):
This PR continues where the recently merged #1117 left off. There is at least one more small bug fix on this line. Hopefully, when taken together, they will fix many of the master/v6 problems our performance tests are currently triggering. We will retest and post a summary when this series is over.

src/store.cc
```diff
@@ -1714,15 +1714,14 @@ StoreEntry::startWriting()
     rep->packHeadersUsingSlowPacker(*this);
     mem_obj->markEndOfReplyHeaders();
 
+    // Same-worker collapsing risks end with the receipt of the headers.
+    // SMP collapsing risks remain until the headers are actually cached, but
+    // that event is announced via CF-agnostic disk I/O broadcasts.
```
eduard-bagdasaryan, line 1719: This implies that there is some code that clears ENTRY_REQUIRES_COLLAPSING in `anchor.basics`, but I could not find it. Perhaps we do not need it anymore? Also what about SMP shm-only configurations?

src/store.cc
```diff
@@ -1714,15 +1714,14 @@ StoreEntry::startWriting()
...
+    if (hittingRequiresCollapsing())
+        setCollapsingRequirement(false);
```
eduard-bagdasaryan, line 1721: We can skip hittingRequiresCollapsing() check here - setCollapsingRequirement() already cares about this.

src/store.cc
rousskov: > This implies that there is some code that clears ENTRY_REQUIRES_COLLAPSING in `anchor.basics`

I do not see such an implication. To me, the comment just explains why we are clearing collapsing risks for this worker (i.e. this StoreEntry object) but not notifying other workers yet. Please detail how you get from this comment to anchor.basics clearing, and I will try to polish the comment to remove that implication (that I do not yet see).

AFAICT, there should be no code that explicitly updates ENTRY_REQUIRES_COLLAPSING in anchor.basics because we are not supposed to use anchor.basics to signal these requirements -- that is one of the simplifications/improvements contained in this PR.


> Perhaps we do not need it anymore?

I think this `if` statement is still needed to end same-worker collapsing risks. Or did you mean something else by `it`?

> Also what about SMP shm-only configurations?

I would expect MemStore to broadcast any changes as the headers get cached there. The comment should not be disk-specific though. I suggest `s/disk I.O/Store writing/` to fix that.

src/store.cc
rousskov: Good point. Done in ae49f1c5d1d3686cf9f2e25032f5561415fa62ba.

src/store/Controller.cc
```diff
@@ -803,11 +796,6 @@ Store::Controller::syncCollapsed(const sfileno xitIndex)
     Transients::EntryStatus entryStatus;
     transients->status(*collapsed, entryStatus);
 
-    if (!entryStatus.collapsed) {
```
eduard-bagdasaryan, line 806: It looks like `entryStatus.collapsed` was used only here. Should we remove it?

src/store/Controller.cc
```diff
@@ -885,33 +870,37 @@ Store::Controller::anchorToCache(StoreEntry &entry, bool &inSync)
...
-        else
-            debugs(20, 5, "failed to anchor " << entry);
+        debugs(20, 5, "failed to anchor " << entry);
         return true;
     }
 
     if (entryStatus.waitingToBeFreed) {
         debugs(20, 5, "failed on marked unattached " << entry);
-        inSync = false;
         return true;
```
rousskov, line 894: The pair of `found` and `inSync` flags is difficult to interpret correctly because the two are not independent. This PR does not change that -- it only improves symmetry in flags handling across anchorToCache() and syncCollapsed() methods. We are working on a followup PR that fixes another bug and eliminates the `inSync` flag from this method.

eduard-bagdasaryan (approved):

PR #1128, opened 2022-08-26
Fewer "busy shutdown" crashes; simpler shutdown cleanup

rousskov:
Preparing for shutdown after 186724 requests
    Waiting 1 seconds for active connections to finish
    Closing HTTP(S) port [::]:3128
    FATAL: assertion failed: Transients.cc:192:
        "oldE->mem_obj && oldE->mem_obj->xitTable.index == index"

This change removes two steps at the end of the manual shutdown sequence
(FreeAllFs() and Store::FreeMemory()) and associated deadly Store hacks,
adjusting a few Store checks as needed. Also, StoreFileSystem hierarchy
declarations were slightly updated when half of its virtual methods were
removed. Key changes are detailed below.

### Do not disable Store [by dropping Store::Root()] during shutdown

Disabling Store (i.e. calling Store::FreeMemory()) resulted in "do not
use Root() during shutdown" assertion-avoiding hacks that bypassed code
maintaining key invariants. Those hacks were probably accommodating
Store-using destruction sequences triggered by SquidShutdown() or later
code, _but_ the hacks also ran during the graceful shutdown period!
Violating those invariants in MemObject destructor led to the quoted
assertions when CollapsedForwarding::HandleNotification() ran during
graceful shutdown.

We could add more state to narrow down hacks scope, or we could disable
IPC notifications that triggered the known deadly shutdown sequence, but

* We should not disable assertions during shutdown because many are
  about essential (e.g., preventing data corruption) invariants and
  because it is not practical to disable all assertions that should be
  disabled (and keep disabling the new ones correctly). The idea leads
  to unsafe code that is also more difficult to write and maintain.

* We should not disable kid-to-kid notifications because some are
  required for finishing transactions (that can still be successfully
  finished) during graceful shutdown. Increasing successful completion
  chances is the whole idea behind that graceful shutdown feature!

Instead, Store should provide (essential) services during shutdown.

TODO: Update those unit test cases that use Store::FreeMemory() but
should not. That requires careful analysis because the next test cases
may rely on the previous case removing its Store object/statistics/etc.
The same dedicated project might (help) address commit 27685ef TODO
about unwanted Store unit test order dependencies.

### Do not disable StoreFileSystem service during shutdown

Similar to the Store service, manually killing FS service is a bad idea.
The underlying code was also essentially unused and buggy:

* Nobody called RegisterAllFsWithCacheManager(). A caller would assert.
* Nobody called Fs::Clean(). A call could crash Squid because deleted
  objects could still be in use by others (they are not refcounted).
* SetupAllFs() was called but did essentially nothing.
* FreeAllFs() was called but had only harmful effects: It leaked memory
  and removed usable FS registrations, increasing shutdown dangers.

See also: Commit 3a85184 documents why we should avoid explicit "delete
essentialService" calls (and why some crashes may still occur if we do).


rousskov (commented):
If others insist, I can split this PR into three, but I hope to avoid spending time on that because all of these changes solve the same basic problem, and most of these changes simply remove code.

src/MemStore.cc
```diff
@@ -587,9 +587,8 @@ MemStore::shouldCache(StoreEntry &e) const
         return false;
     }
 
-    // Store::Root() in the next check below is FATALly missing during shutdown
```
yadij, line 590: why is this comment no longer relevant?

yadij:
I would like the LEAK_CHECK_MODE removal as s separate PR. It is one of those feature-like changes that need highlighting for those who may be trying to use it. I do not have a strong opinion on the other two.

src/main.cc
```diff
@@ -2112,24 +2108,7 @@ SquidShutdown()
...
-    storeFreeMemory();
-    /*stmemFreeMemory(); */
-    netdbFreeMemory();
-    ipcacheFreeMemory();
-    fqdncacheFreeMemory();
-    asnFreeMemory();
-    clientdbFreeMemory();
-    statFreeMemory();
-    eventFreeMemory();
-    mimeFreeMemory();
```
rousskov, line 2129: > I would like the LEAK_CHECK_MODE removal as s separate PR. It is one of those feature-like changes that need highlighting for those who may be trying to use it.

Will extract. Please note that the LEAK_CHECK_MODE feature has been disabled since 2006 commit afec404. It cannot be enabled without changing Squid source code, including dealing with the no longer globally available mimeFreeMemory() that was hidden since 2012 commit b65ce00.

rousskov: > besides resolving trivial code conflicts, this PR description will need to be updated to refer to the rationale partially documented in PR 1130

Done (after PR 1130 got merged).

yadij (approved):
Thank you for removing all this dead code

PR #1129, opened 2022-08-29
Fix printing Security::ErrorDetail

rousskov:
Squid dumps RefCount pointer details instead of the intended TLS error
details like "SQUID_TLS_ERR_ACCEPT+TLS_LIB_ERR=14094418":

    ERROR: failure while accepting a TLS connection...: 0x564017165ef0*1

We overload operator "<<" for ErrorDetail::Pointer to report errors, but
the compiler picks our generic operator "<<" overload for RefCount<C>
instead. I believe this happens because the actual type of the
being-printed handshakeResult.errorDetail object (i.e.
Security::IoResult::ErrorDetailPointer) is not ErrorDetail::Pointer but
Security::ErrorDetail::Pointer; there is no overload for the latter.

This simple solution "works" but it does not solve the underlying
problem: Other ErrorDetail kids (existing and future) and other similar
class hierarchies using RefCount pointers (e.g., Http::Message) are or
will be mishandled because of missing kid-specific overloads.

To properly address this XXX, we probably should add traits/interfaces
to all RefCountable classes that want their Pointers to be printed
specially (e.g., ErrorDetail) _and_ teach our generic operator "<<"
overload for RefCount<C> to honor those. Also, an InstanceId data member
might be recognized/printed by default instead of the memory address; it
may even be added to RefCountable.

### DIFF
--- a/src/error/Detail.h
+++ b/src/error/Detail.h
@@ -40,7 +40,10 @@ ErrorDetail::Pointer MakeNamedErrorDetail(const char *name);
 /// dump the given ErrorDetail (for debugging)
 std::ostream &operator <<(std::ostream &os, const ErrorDetail &);
 
-/// dump the given ErrorDetail pointer which may be nil (for debugging)
+// XXX: Every ErrorDetail child, especially those declaring their own Pointer
+// types should overload this operator. The compiler will not find this overload
+// for child pointers. See Security::ErrorDetail overload for an example.
+/// dump the given ErrorDetail via a possibly nil pointer (for debugging)
 std::ostream &operator <<(std::ostream &os, const ErrorDetail::Pointer &);
 
 #endif /* _SQUID_SRC_ERROR_DETAIL_H */
--- a/src/error/Detail.h
+++ b/src/error/Detail.h
@@ -40,5 +40,8 @@ ErrorDetail::Pointer MakeNamedErrorDetail(const char *name);
 /// dump the given ErrorDetail (for debugging)
 std::ostream &operator <<(std::ostream &os, const ErrorDetail &);
 
+/// dump the given ErrorDetail pointer which may be nil (for debugging)
+std::ostream &operator <<(std::ostream &os, const ErrorDetail::Pointer &);
+
 #endif /* _SQUID_SRC_ERROR_DETAIL_H */
 
--- a/src/error/Detail.h
+++ b/src/error/Detail.h
@@ -40,8 +40,5 @@ ErrorDetail::Pointer MakeNamedErrorDetail(const char *name);
 /// dump the given ErrorDetail (for debugging)
 std::ostream &operator <<(std::ostream &os, const ErrorDetail &);
 
-/// dump the given ErrorDetail pointer which may be nil (for debugging)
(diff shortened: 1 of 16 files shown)


src/error/Detail.h
```diff
@@ -40,7 +40,10 @@ ErrorDetail::Pointer MakeNamedErrorDetail(const char *name);
 /// dump the given ErrorDetail (for debugging)
 std::ostream &operator <<(std::ostream &os, const ErrorDetail &);
 
-/// dump the given ErrorDetail pointer which may be nil (for debugging)
+// XXX: Every ErrorDetail child, especially those declaring their own Pointer
+// types should overload this operator. The compiler will not find this overload
+// for child pointers. See Security::ErrorDetail overload for an example.
+/// dump the given ErrorDetail via a possibly nil pointer (for debugging)
 std::ostream &operator <<(std::ostream &os, const ErrorDetail::Pointer &);
```
rousskov, line 47: > PR description: To properly address this XXX, we probably should add traits/interfaces to all RefCountable classes that want their Pointers to be printed specially (e.g., ErrorDetail) and teach our generic operator "<<" overload for RefCount to honor those.

I can implement this long-term fix in this PR or in a separate PR if others would support it. I did not want to introduce such changes here without explicit support. I did try to implement a few other, arguably simpler long-term solutions, but they all failed. Please let me know if you think this PR should be used to properly address this XXX or if you would support a PR dedicated to that change (after this PR fixes the immediate/known problem).

yadij: Better design would be to treat smart pointers the same as raw-pointers and display the pointed-to object by default. Dropping the unusual RefCount operator<<() overload which is causing this issue.

rousskov: > Better design would be to treat smart pointers the same as raw-pointers and display the pointed-to object by default. 

The primary design question is not about the _default_ (that would be trivial to change) but about handling two (or more) kinds of use cases (e.g., printing the address in some cases and printing the object in others). Accommodating several use cases precludes us from treating smart pointers any _one_ way, including "the same way as raw pointers". Several kinds of use cases exist in the official code today.

> Dropping the unusual RefCount operator<<() overload which is causing this issue.

If you are talking about dropping the overload in RefCount.h, then it is already treating smart pointers as raw pointers (essentially) -- printing the memory address. That overload does not cause the bug this PR is working around: If we remove that overload from the official code, the compiler will issue an error instead of picking the ErrorDetail overload we want it to pick.

Also, we cannot drop that overload in favor of simply dereferencing the pointer when it needs to be printed. That simple design does not work well, neither for raw pointers nor for smart pointers because

* many of the pointers might by nil: overloads are needed to handle nil pointers correctly
* many of the existing objects printed today via pointers do not have an object printing operator (i.e. cannot be printed)

We have been solving all these problems using custom overloads, one overload per class that requires special class-specific handling. The bug this PR is working around shows the limits of that approach -- it does not work well for class _hierarchies_ like ErrorDetail and Http::Message. There are other problems as well -- a human code reader cannot easily tell what will be printed in each particular case (without knowing and remembering a lot of low-level details).

yadij (approved):

PR #1134, opened 2022-09-04
Do not lock an SMP hit twice for the same transaction

rousskov:
Before commit 4310f8b, anchorToCache() (known then as anchorCollapsed())
was only called for not-yet-anchored entries. For anchored entries, we
called updateCollapsed(). In that commit, a new anchorToCache() caller
(i.e. allowSharing() after peek()) started calling anchorToCache() for
all entries, but anchorToCache() code was not updated to skip anchoring
of anchored entries, leading to an extra read lock placed on a peek()ed
hit entry. The extra lock prevented the entry slot from being reused for
other entries when the entry was later purged (i.e. marked for removal),
drastically decreasing hit ratio in some environments.

Fixed anchorToCache() no longer anchors anchored entries. To make sure
we covered all callers/cases, especially in patches, we changed the call
parameters, addressing an old code simplification TODO.

### DIFF
--- a/src/fs/rock/RockSwapDir.cc
+++ b/src/fs/rock/RockSwapDir.cc
@@ -72,8 +72,10 @@ Rock::SwapDir::get(const cache_key *key)
 }
 
 bool
-Rock::SwapDir::anchorToCache(StoreEntry &entry, bool &inSync)
+Rock::SwapDir::anchorToCache(StoreEntry &entry)
 {
+    Assure(!entry.hasDisk());
+
     if (!map || !theFile || !theFile->canRead())
         return false;
 
@@ -84,8 +86,7 @@ Rock::SwapDir::anchorToCache(StoreEntry &entry, bool &inSync)
         return false;
 
     anchorEntry(entry, filen, *slot);
-    inSync = updateAnchoredWith(entry, *slot);
-    return true; // even if inSync is false
+    return true;
 }
 
 bool
@@ -96,13 +97,7 @@ Rock::SwapDir::updateAnchored(StoreEntry &entry)
 
     assert(entry.hasDisk(index));
 
-    const Ipc::StoreMapAnchor &s = map->readableEntry(entry.swap_filen);
-    return updateAnchoredWith(entry, s);
-}
-
-bool
-Rock::SwapDir::updateAnchoredWith(StoreEntry &entry, const Ipc::StoreMapAnchor &anchor)
-{
+    const auto &anchor = map->readableEntry(entry.swap_filen);
     entry.swap_file_sz = anchor.basics.swap_file_sz;
     return true;
 }
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -369,18 +369,13 @@ Store::Controller::find(const cache_key *key)
 void
 Store::Controller::allowSharing(StoreEntry &entry, const cache_key *key)
 {
-    // TODO: refactor to throw on anchorToCache() inSync errors!
-
     // anchorToCache() below and many find() callers expect a registered entry
     addReading(&entry, key);
 
     if (entry.hasTransients()) {
         // store hadWriter before computing `found`; \see Transients::get()
         const auto hadWriter = transients->hasWriter(entry);
-        bool inSync = false;
-        const bool found = anchorToCache(entry, inSync);
-        if (found && !inSync)
-            throw TexcHere("cannot sync");
+        const auto found = anchorToCache(entry);
         if (!found) {
             // !found should imply hittingRequiresCollapsing() regardless of writer presence
             if (!entry.hittingRequiresCollapsing()) {
@@ -820,7 +815,15 @@ Store::Controller::syncCollapsed(const sfileno xitIndex)
         found = true;
         inSync = swapDir->updateAnchored(*collapsed);
     } else {
-        found = anchorToCache(*collapsed, inSync);
+        try {
+            found = anchorToCache(*collapsed);
+            inSync = found;
+        } catch (...) {
+            // TODO: Write an exception handler for the entire method.
+            debugs(20, 3, "failed to anchor XXX or to sync?" << *collapsed << ": " << CurrentException);
+            inSync = false;
+            found = true; // syncing failure implies there was something to sync with
+        }
     }
 
     if (entryStatus.waitingToBeFreed && !found) {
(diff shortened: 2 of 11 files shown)


src/store/Controller.cc
```diff
@@ -855,49 +858,51 @@ Store::Controller::syncCollapsed(const sfileno xitIndex)
...
 /// \returns true for other entries after synchronizing them with their store
-/// and setting inSync to reflect that synchronization outcome.
 bool
-Store::Controller::anchorToCache(StoreEntry &entry, bool &inSync)
+Store::Controller::anchorToCache(StoreEntry &entry)
 {
     assert(entry.hasTransients());
     assert(transientsReader(entry));
 
+    // TODO: Attach entries to both memory and disk
```
rousskov, line 870: This TODO is out of this bug-fixing PR scope. It will probably take us quite a while to improve the surrounding code enough to make that complex improvement possible.

src/store/Controller.cc
```diff
@@ -855,49 +858,51 @@ Store::Controller::syncCollapsed(const sfileno xitIndex)
...
+
+    // TODO: Reduce code duplication with syncCollapsed()
```
rousskov, line 872: This TODO can probably be addressed without a lot of changes leaking outside Controller.cc, but doing so will require refactoring syncCollapsed() quite a bit, including how it (and its callers) handle exceptions. That refactoring is best done in a dedicated PR or, at least, when we have to fix some syncCollapsed() bug. This PR fixes an anchorCollapsed() bug. It is best to keep this PR as focused on that fix/bug as possible.

src/store/Controller.cc
```diff
@@ -811,7 +806,7 @@ Store::Controller::syncCollapsed(const sfileno xitIndex)
     if (sharedMemStore && collapsed->mem_obj->memCache.io == MemObject::ioDone) {
         found = true;
         inSync = true;
-        debugs(20, 7, "fully mem-loaded " << *collapsed);
+        debugs(20, 7, "already handled by memory store: " << *collapsed);
```
rousskov, line 809: The branch-added anchorToCache() debugging got it right: "memCache.io == MemObject::ioDone" does not mean that the entire object is currently in the local transient memory. It only means that it was fully loaded there at some point in the past. It may be gone now. It may not even be in the shared memory cache anymore. The difference is not important for this code -- we do not want to anchor the same entry multiple times anyway (at least for now, for simplicity sake) -- but debugs() should not mislead the reader.

I think we should keep the two very similar debugs() in similar code in sync even if doing so triggers this out-of-scope polishing change.

rousskov (commented):
This PR continues the bug-fixing series started by #1117 and #1127. There may be one or two additional small PRs after this; we are running a new set of tests that may expose more bugs.

----

The following abridged cache.log shows two `opened entry 2382 for reading rock` lines, one during the peek() sequence and the other during the allowSharing() sequence, both starting from the same Store::Controller::find() call that returns a HIT. Each line represents a reading lock increment, resulting in double/excessive locking.

     Controller.cc(430) peek: 66CB47D2ABB8F68D3A4BCD2D6647A50C
     StoreMap.cc(492) opened entry 2382 for reading rock
     store.cc(330) StoreEntry constructed
     store.cc(1953) attachToDisk: attaching entry
     Controller.cc(459) peek: got disk-cached entry: e:d2382@0=sw2V...*0
 
     store.cc(402) hashInsert: Inserting Entry e:tr5792d2382@0=sw2V...*0
     Controller.cc(861) anchorToCache: anchoring e:tr5792d2382@0=sw2V...
     StoreMap.cc(492) opened entry 2382 for reading rock
     store.cc(1953) attachToDisk: attaching entry
     Disks.cc(627) anchorToCache: ... anchors e:tr5792d2382@0=sw2V...*0
     Controller.cc(874) anchorToCache: anchored e:tr5792d2382@0=sw2V...*0
 
     client_side_reply.cc(1746) identifyFoundObject: default HIT

src/store/Controller.cc
```diff
@@ -855,49 +858,51 @@ Store::Controller::syncCollapsed(const sfileno xitIndex)
...
     assert(entry.hasTransients());
     assert(transientsReader(entry));
 
+    // TODO: Attach entries to both memory and disk
+
+    // TODO: Reduce code duplication with syncCollapsed()
+    if (sharedMemStore && entry.mem().memCache.io == MemObject::ioDone) {
+        debugs(20, 5, "already handled by memory store: " << entry);
+        return true;
+    } else if (sharedMemStore && entry.hasMemStore()) {
```
eduard-bagdasaryan, line 876: Is it possible that `entry.mem().memCache.io == MemObject::ioDone` but  entry.hasMemStore() is true - i.e., can we join these two conditions?

rousskov: > Is it possible that `entry.mem().memCache.io == MemObject::ioDone` but `entry.hasMemStore()` is true

I do not think it is possible today, but it might become common as we continue to separate shared memory cache from the worker-specific memory storage. AFAICT, there is no code ensuring this invariant today -- MemStore just marks entries with ioDone when it refuses to cache them OR is done caching them, releasing the lock.

> can we join these two conditions?

We can just check for the `!(false and false)` condition, but I doubt it will improve this code; it will just make the same information a bit more cryptic for the cache.log (and code) reader.

| io | hasMemStore() | meaning |
| --- | -------------------- | ----------- |
| 0 | 0 | has not attempted to mem-cache yet |
| 0 | 1 | is attempting to mem-cache now |
| 1 | 0 | attempted to mem-cache and not holding the lock |
| 1 | 1 | attempted to mem-cache and still holding the lock |

Also, any changes here would increase the difference with syncCollapsed() unless we modify that method too. Perhaps that can be done later, when syncCollapsed() is polished to handle exceptions and the code duplication between the two methods is removed/reduced.

src/fs/rock/RockSwapDir.cc
```diff
@@ -84,8 +86,7 @@ Rock::SwapDir::anchorToCache(StoreEntry &entry, bool &inSync)
         return false;
 
     anchorEntry(entry, filen, *slot);
-    inSync = updateAnchoredWith(entry, *slot);
-    return true; // even if inSync is false
+    return true;
```
eduard-bagdasaryan, line 89: Don't we need to set `entry.swap_file_sz` as the old `updateAnchoredWith()` did?

src/store/Controller.cc
eduard-bagdasaryan: Ok.
Don't we need a comment explaining why we have the "already handled by memory store..." condition for Memory store but don't have such condition for Disk store?

rousskov: We could try to explain the current shared memory store deficiencies here (and in a few other places, I am sure!), but I would rather not do that given the complexity of that explanation and the fact that we are actively working on _removing_ those deficiencies. When we are done, the (shared) memory store condition will be the same as the disk store condition. The shared memory store will no longer use the current copyFromShm() hack, eliminating the need for the current ioDone check.

src/fs/rock/RockSwapDir.cc
eduard-bagdasaryan: Agreed, the existing `anchorEntry()` does(did) this already.

PR #1135, opened 2022-09-07
Replaced FDECB callback with a basic async call

eduard-bagdasaryan:
The "FD passing callback" was misleading because it was not used as a
callback. We just need an async call with a known-to-the-caller
parameter instead. Simplifying CommCalls (by removing this "special
callback") helps improve that API.

### DIFF
--- a/src/comm.cc
+++ b/src/comm.cc
@@ -10,6 +10,7 @@
 
 #include "squid.h"
 #include "ClientInfo.h"
+#include "base/AsyncFunCalls.h"
 #include "comm/AcceptLimiter.h"
 #include "comm/comm_internal.h"
 #include "comm/Connection.h"
@@ -805,19 +806,19 @@ old_comm_reset_close(int fd)
 }
 
 static void
-commStartTlsClose(const FdeCbParams &params)
+commStartTlsClose(const int fd)
 {
-    Security::SessionSendGoodbye(fd_table[params.fd].ssl);
+    Security::SessionSendGoodbye(fd_table[fd].ssl);
 }
 
 static void
-comm_close_complete(const FdeCbParams &params)
+comm_close_complete(const int fd)
 {
-    fde *F = &fd_table[params.fd];
+    auto F = &fd_table[fd];
     F->ssl.reset();
     F->dynamicTlsContext.reset();
-    fd_close(params.fd);        /* update fdstat */
-    close(params.fd);
+    fd_close(fd); /* update fdstat */
+    close(fd);
 
     ++ statCounter.syscalls.sock.closes;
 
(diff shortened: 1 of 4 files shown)


src/comm.cc
```diff
@@ -805,19 +806,19 @@ old_comm_reset_close(int fd)
...
+comm_close_complete(const int fd)
 {
-    fde *F = &fd_table[params.fd];
+    auto F = &fd_table[fd];
     F->ssl.reset();
     F->dynamicTlsContext.reset();
-    fd_close(params.fd);        /* update fdstat */
-    close(params.fd);
+    fd_close(fd); /* update fdstat */
+    close(fd);
```
yadij, line 821: This callback should now receive the FD value as a Comm::Descriptor object. That will make these two function calls unnecessary as it will implicitly call them when the Call scope ends.

yadij (changes requested):
Thank you for this cleanup.

src/comm.cc
rousskov: I bet it is not possible to use Comm::Descriptor here and with most relevant async calls (in this PR) for a few reasons, including:

1. Comm::Descriptor constructor registers the descriptor with Comm (via fd_open()), but most (former FDECB) callers deal with the previously registered descriptors.
2. Many AsyncCall Dialers copy call arguments, but Comm::Descriptor objects cannot be copied.
3. commCallCloseHandlers() needs to async-notify many callbacks about the _same_ descriptor, but Comm::Descriptor objects cannot be copied.

After/if we upgrade Connection and other descriptor owners to Comm::Descriptor, we should be able to enhance Comm::Descriptor and possibly dialers to support _moving_ the descriptor from one context to another, but even that will not help with item 3 problems. To solve those, we would probably have to wait until the higher-level code is refactored enough to stop sharing descriptors (e.g., via Connection pointers) -- so that there is only one exclusive owner of any descriptor. All that requires a lot of complex work way outside this simple PR scope.

rousskov: > Alex: 3. commCallCloseHandlers() needs to async-notify many callbacks about the same descriptor, but Comm::Descriptor objects cannot be copied.

Correction: This exact problem does not apply here because Comm close handlers called by commCallCloseHandlers() are not FDECB.

However, the same general principle/problem applies to commStartTlsClose caller, which _is_ the former FDECB caller. That async call cannot assume ownership of the descriptor (even if it could do that from C++ point of view) because Comm still needs to notify _others_ about the same descriptor. Thus, the essence of problem 3 is still applicable.

rousskov (approved):

yadij (approved):

PR #1136, opened 2022-09-08
Bug 5232: Fix GCC v12 build [-Wuse-after-free]

eduard-bagdasaryan:
This warning is a false positive. Introduced in GCC v12, use-after-free
checks have attracted dozens of bug reports, with several reports still
open as of GCC v12.2, including one that looks very similar to our case:
https://gcc.gnu.org/bugzilla//show_bug.cgi?id=106776

Removing (excessive) "inline" works around this GCC bug. We do not have
a better workaround right now. If we see more false positives like this,
we may disable -Wuse-after-free until its support matures.

### DIFF
--- a/src/base/AsyncCall.h
+++ b/src/base/AsyncCall.h
@@ -151,7 +151,7 @@ private:
 };
 
 template <class Dialer>
-inline RefCount< AsyncCallT<Dialer> >
+RefCount< AsyncCallT<Dialer> >
 asyncCall(int aDebugSection, int aDebugLevel, const char *aName,
           const Dialer &aDialer)
 {


eduard-bagdasaryan:
This patch should fix the gcc-12 build problem, but it is not still clear _why_ it made gcc-12 happy. We need some more time to figure out what the underlying problem is and whether this solution is the correct one (or come up with another solution).

yadij (approved):
Looks good and confirmed it fixes the build.

rousskov:
Since there is no emergency here (AFAICT), I support spending a bit more time on trying to figure out what is really going on, now that we have a much easier-to-reproduce case than the Coverity ones.

* If this is a false positive, then Coverity and GCC are making the same kind of mistake and are likely to keep making similar mistakes. It would be good to develop a better understanding of the underlying issue instead of just kicking the can down the road.
* If this is a Squid bug, then fixing it is definitely more important than merging this PR on the fast track.

rousskov:
@eduard-bagdasaryan, thanks a lot for researching this problem! I have updated PR description to reflect your findings. Please adjust as needed. Once the PR description is agreed upon, we should merge this PR.

rousskov (approved):

eduard-bagdasaryan:
The PR description looks correct now.

rousskov:
Thank you. I will give Amos a few days to review the updated PR description before clearing this PR for merging. @yadij , please feel free to clear at any time.

PR #1137, opened 2022-09-09
Improve SQUID_YESNO error messages

yadij:
This macro performs a very specific validation check but was
passed a mix of different error messages to display on failure.
Many of which were unclear or wrong.

Move the text of the error inside the macro to enforce
consistency and accuracy about what failed. It actually only
needs to be passed the name of the ./configure options which
was configured badly.

Also, the options configured can only take the positive form
when they fail '--with-foo=X' or 'enable-foo=X' so update the
second argment to always show that form regardless of whether
the configure.ac logic has it default enabled or disabled.

Also, fixed a bug in --with-nat-devpf validation which was
testing the wrong variable (enableval). This bug only affects
the errors displayed so no builds were significantly affected.

### DIFF
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -211,7 +211,7 @@ unset squid_tmp_define
 dnl aborts with an error specified as the second argument if the first argument doesn't
 dnl contain either "yes" or "no"
 AC_DEFUN([SQUID_YESNO],[
-  AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([$2])])
+  AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([Bad argument for $2, which does not accept a value.])])
 ])
 
 dnl check the build parameters for a library to auto-enable
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -211,7 +211,7 @@ unset squid_tmp_define
 dnl aborts with an error specified as the second argument if the first argument doesn't
 dnl contain either "yes" or "no"
 AC_DEFUN([SQUID_YESNO],[
-  AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([Bad argument for $2, which does not accept a value.])])
+  AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([Bad argument for $2: "$1". Expecting "yes", "no", or no argument.])])
 ])
 
 dnl check the build parameters for a library to auto-enable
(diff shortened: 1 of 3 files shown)


acinclude/squid-util.m4
```diff
@@ -211,7 +211,7 @@ unset squid_tmp_define
 dnl aborts with an error specified as the second argument if the first argument doesn't
 dnl contain either "yes" or "no"
 AC_DEFUN([SQUID_YESNO],[
-  AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([$2])])
+  AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([Bad argument for $2, which does not accept a value.])])
```
rousskov, line 214:
```
Bad argument for --enable-ecap, which does not accept a value.
```

The new error text looks wrong to me because `--enable-foo` options do accept at least two values: "yes" and "no". I realize that some of the existing callers had similar problems, but we should not hard-code those selected caller bugs into the macro implementation, of course.

Here is an untested correction sketch:

```suggestion
  AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([Bad argument for $2: "$1". Expecting "yes", "no", or no argument.])])
```

As always, please adjust as needed to address the _essence_ of this change request. No need to argue about any specific suggestion flaws.

acinclude/squid-util.m4
```diff
@@ -211,7 +211,7 @@ unset squid_tmp_define
...
 ])
```
rousskov, line 215: > PR title: Polish use of SQUID_YESNO macro

Please remove "use of" from the PR title. The PR changes the macro itself (in addition to changing how it is used).

FWIW, I would title this PR "Improve SQUID_YESNO error messages". The PR description and the diff make it clear where the changes are IMO.

configure.ac
```diff
@@ -357,8 +355,7 @@ AC_ARG_ENABLE(optimizations,
 AC_ARG_ENABLE(debug-cbdata,
   AS_HELP_STRING([--enable-debug-cbdata],
       [Provide some debug information in cbdata]), [
-SQUID_YESNO([$enableval],
-    [unrecognized argument to --enable-debug-cbdata: $enableval])
+  SQUID_YESNO([$enableval],[--enable-debug-cbdata])
```
rousskov, line 358: There is a lot of code duplication in all these blobs. I wonder whether SQUID_YESNO could automatically extract or reconstruct the name of the option from some variable set by AC_ARG_ENABLE and AC_ARG_WITH (e.g., the variable that holds their parameter). This is just a comment. I do not insist on any changes here because it may not be feasible (or may not be a good idea) to remove this duplication.

rousskov (changes requested):
Good idea, thank you!

configure.ac
yadij: Aye, part of the aim of these cleanups is to see how much of the unnecessary uniqueness can be removed. In hopes that someone can eventually write a macro and hide the M4 verbosity/duplication.

rousskov (approved):
Thank you for addressing my primary concerns.

PR #1138, opened 2022-09-09
Rename ./configure option --with-libxml2 to --with-xml2

yadij:
The 'lib' prefix is supposed to be omitted from
--with/--without names but libxml2 was not
correctly implemented from inception.


rousskov (approved):
There is also `SQUID_AUTO_LIB(libcap...)` with the same TODO comment, but if you prefer to fix just one case in this PR, that is fine too, of course.

yadij:
Ah, thanks I missed that one. Will do it as separate PR.

PR #1139, opened 2022-09-09
Fix 'AC_LINK_IFELSE was called before AC_USE_SYSTEM_EXTENSIONS'

yadij:
... produced during bootstrap.sh by autoconf 2.71

AC_USE_SYSTEM_EXTENSIONS needs to be placed after detecting
which OS and compiler are being used. But before any of the
checks detecting capabilities of those OS and compilers.

AC_CANONICAL_HOST is a prerequisite of the simple_host_os logic.
Attach it to that code block instead of the compiler detection
block.

Replace deprecated '$GCC' variable with autoconf cached
'$ac_cv_c_compiler_gnu' variable used as the real source for the
yes/no value of GCC usability after AC_PROG_CC.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,8 +52,7 @@ AC_PROG_CC
 AM_PROG_CC_C_O
 AC_PROG_CXX
 
-# this needs to be before any test is run, to have more standard
-# functions available on some Unix sysems (e.g. Solaris)
+# AC_USE_SYSTEM_EXTENSIONS should be called before any macros that run the C compiler.
 AS_IF([test "x$squid_host_os" = "solaris" -a "x$GCC" = "xyes"],[
   AC_USE_SYSTEM_EXTENSIONS
 ])
--- a/configure.ac
+++ b/configure.ac
@@ -35,12 +35,30 @@ PRESET_LDFLAGS="$LDFLAGS"
 dnl Set default LDFLAGS
 AS_IF([test "x$LDFLAGS" = "x"],[LDFLAGS="-g"])
 
-# Check for GNU cc
+# check for host OS detail
+AC_CANONICAL_HOST
+AC_MSG_CHECKING([simplified host os])
+simple_host_os=`echo $host_os|sed 's/[0-9].*//g;s/-.*//g'`
+squid_host_os_version=`echo $host_os|tr -d "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-"`
+AS_IF([test -n "$squid_host_os_version"],[
+  squid_host_os="`echo $simple_host_os| sed s/$squid_host_os_version//g`"
+],[
+  squid_host_os="$simple_host_os"
+])
+AC_MSG_RESULT($squid_host_os (version $squid_host_os_version))
+# on windows squid_host_os is either mingw or cygwin, version is 32
+
+# Find available compilers
 AC_PROG_CC
 AM_PROG_CC_C_O
 AC_PROG_CXX
 AC_LANG([C++])
-AC_CANONICAL_HOST
+
+# this needs to be before any test is run, to have more standard
+# functions available on some Unix sysems (e.g. Solaris)
+AS_IF([test "x$squid_host_os" = "solaris" -a "x$ac_cv_c_compiler_gnu" = "xyes"],[
+  AC_USE_SYSTEM_EXTENSIONS
+])
 
 # Clang 3.2 on some CPUs requires -march-native to detect correctly.
 # GCC 4.3+ can also produce faster executables when its used.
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -573,7 +574,7 @@ for module in $squid_disk_module_candidates none; do
             ],[
               SQUID_CFLAGS="$SQUID_CFLAGS -D_REENTRANT"
               SQUID_CXXFLAGS="$SQUID_CXXFLAGS -D_REENTRANT"
-              AS_IF([test "x$GCC" = "xyes" -a "x$PRESET_LDFLAGS" = "x"],[LDFLAGS="$LDFLAGS -pthread"])
+              AS_IF([test "x$ac_cv_c_compiler_gnu" = "xyes" -a "x$PRESET_LDFLAGS" = "x"],[LDFLAGS="$LDFLAGS -pthread"])
```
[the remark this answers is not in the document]
yadij: I am not sure why the AC_PROG_C is absent from that documentation but AC_PROG_C_C_0 is present, though it is not alone quite a few things are missing.
It is one of the set where Fortran and F77 are highlighted - ("ac_cv_" [LANGUAGE] "_compiler_" [TOOLCHAIN])

configure.ac
```diff
@@ -35,12 +35,30 @@ PRESET_LDFLAGS="$LDFLAGS"
...
+  squid_host_os="$simple_host_os"
+])
+AC_MSG_RESULT($squid_host_os (version $squid_host_os_version))
+# on windows squid_host_os is either mingw or cygwin, version is 32
+
+# Find available compilers
 AC_PROG_CC
 AM_PROG_CC_C_O
 AC_PROG_CXX
 AC_LANG([C++])
```
rousskov, line 55: > AC_LANG (language):  Do compilation tests using the compiler, preprocessor, and file extensions for the specified language. 

AFAICT, the position of the AC_LANG macro contradicts AC_USE_SYSTEM_EXTENSIONS documentation: "[AC_USE_SYSTEM_EXTENSIONS] should be called _before_ any macros that run the C compiler".

Please consider moving AC_LANG after AC_USE_SYSTEM_EXTENSIONS.

configure.ac
```diff
@@ -35,12 +35,30 @@ PRESET_LDFLAGS="$LDFLAGS"
...
+squid_host_os_version=`echo $host_os|tr -d "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-"`
+AS_IF([test -n "$squid_host_os_version"],[
+  squid_host_os="`echo $simple_host_os| sed s/$squid_host_os_version//g`"
+],[
+  squid_host_os="$simple_host_os"
+])
+AC_MSG_RESULT($squid_host_os (version $squid_host_os_version))
+# on windows squid_host_os is either mingw or cygwin, version is 32
+
+# Find available compilers
```
rousskov, line 51: Let's not re-document what these macros do, especially when these macros do a lot more than "finding available compilers" (and some of them do not find compilers at all!).

```suggestion
```

rousskov (changes requested):

configure.ac
yadij: Done in [192671e](https://github.com/squid-cache/squid/pull/1139/commits/192671eae9d9f9ce0240b53a95b005743632f1c3)

configure.ac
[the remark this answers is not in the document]
yadij: Oh well, can be done later. Reverted in [192671e](https://github.com/squid-cache/squid/pull/1139/commits/192671eae9d9f9ce0240b53a95b005743632f1c3)

configure.ac
```diff
@@ -35,12 +35,30 @@ PRESET_LDFLAGS="$LDFLAGS"
...
+])
+AC_MSG_RESULT($squid_host_os (version $squid_host_os_version))
+# on windows squid_host_os is either mingw or cygwin, version is 32
+
 AC_PROG_CC
 AM_PROG_CC_C_O
 AC_PROG_CXX
+
+# this needs to be before any test is run, to have more standard
+# functions available on some Unix sysems (e.g. Solaris)
```
rousskov, line 56: The "any test" text is too general to be meaningful in this context -- clearly we have already performed some tests above. Let's use AC_USE_SYSTEM_EXTENSIONS documentation instead. I also do not recommend re-documenting What AC_USE_SYSTEM_EXTENSIONS does.

```suggestion
# AC_USE_SYSTEM_EXTENSIONS should be called before any macros that run the C compiler.
```

I realize that you are just moving an old comment, but we should take in to account that the old comment tried and _failed_ to prevent the bugs this PR is fixing. Improving it is in scope IMO, but I do not insist on this change.

P.S. I did write a very similar change request in the first review but it looks like I somehow deleted it back then. Sorry for one more iteration.

rousskov (approved):

PR #1140, opened 2022-09-11
Rename ./configure option --with-libcap to --with-cap

yadij:
The 'lib' prefix is supposed to be omitted from --with/--without
names but libcap was not correctly named.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2670,7 +2670,8 @@ AS_IF([test "x$with_cap" != "xno"],[
   AS_IF([test "x$LIBCAP_LIBS" != "x"],[
     with_cap=yes
     CPPFLAGS="$LIBCAP_CFLAGS $CPPFLAGS"
-    LDFLAGS="$LIBCAP_PATH $LDFLAGS"
+    LIBCAP_LIBS="$LIBCAP_PATH $LIBCAP_LIBS"
+    AC_DEFINE(USE_LIBCAP,1,[Linux capabilities library support])
   ],[test "x$with_cap" = "xyes"],[
     AC_MSG_ERROR([Required library libcap not found])
   ],[
@@ -2678,8 +2679,7 @@ AS_IF([test "x$with_cap" != "xno"],[
     with_cap=no
   ])
 ])
-SQUID_DEFINE_BOOL(USE_LIBCAP,${with_cap:=yes},[Linux capabilities support])
-AC_MSG_NOTICE([Linux capabilities support enabled: ${with_cap} ${LIBCAP_PATH} ${LIBCAP_LIBS}])
+AC_MSG_NOTICE([Linux capabilities support enabled: ${with_cap} ${LIBCAP_LIBS}])
 
 dnl Check for needed libraries
 AC_SEARCH_LIBS([gethostbyname],[nsl resolv bind])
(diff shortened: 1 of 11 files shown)


rousskov (approved):

rousskov:
Please do not merge this PR until you have a chance to check ./configure output for errors _after_ #1151 and #1152 land.

configure.ac
```diff
@@ -2713,39 +2713,34 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
...
-      # or it can't be supported: disable
-      with_libcap=no
-    ])
+  AC_MSG_NOTICE([libcap headers are ok: $squid_cv_sys_capability_works])
+  AS_IF([test "x$squid_cv_sys_capability_works" = "xno"],[LIBCAP_LIBS=""])
+  SQUID_STATE_ROLLBACK(squid_libcap_state)
+
+  AS_IF([test "x$LIBCAP_LIBS" != "x"],[
+    with_cap=yes
+    CXXFLAGS="$LIBCAP_CFLAGS $CXXFLAGS"
```
rousskov, line 2732: See #1153 and related PKG_CHECK_MODULES() [documentation](https://autotools.info/pkgconfig/pkg_check_modules.html).

```suggestion
    CPPFLAGS="$LIBCAP_CFLAGS $CPPFLAGS"
```

There is a similar problem with LIBCAP_CFLAGS assignment in an earlier code blob (unchanged by this PR).

configure.ac
```diff
@@ -2713,39 +2713,34 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
...
+    LDFLAGS="$LIBCAP_PATH $LDFLAGS"
```
rousskov, line 2733: This line looks like a copied typo to me; PKG_CHECK_MODULES() sets and we are then checking LIBCAP_LIBS rather than LIBCAP_PATH.

```suggestion
    LDFLAGS="$LIBCAP_LIBS $LDFLAGS"
```

There is a similar problem with a LIBCAP_PATH assignment in an earlier code blob (unchanged by this PR).

rousskov (dismissed):

configure.ac
yadij: This particular AS_IF block (and equivalents for each --with-foo library option) is setting up the final/global build environment based on the test results. This line is informing the linker about the admins custom PATH location. Makefile.am use the *_LIBS only when needed to identify which bits to pull from there.

rousskov: I now understand that we are adding LIBCAP_PATH because SQUID_AUTO_LIB sets it.

I still find it odd that there is a context with LIBCAP_PATH but without LIBCAP_LIBS; both variables may contain `-L` options. I understand that LIBCAP_LIBS may contain `-l` options, and that it may be wrong to add `-l` options to LDFLAGS, but those considerations do not make this look less odd. Perhaps we need a variable that contains just the `-L` options from `pkg-config`, or perhaps we should not update LDFLAGS at all, or perhaps there is a completely different solution. I did not study this problem enough to recommend a specific fix. I can only point out the inconsistency/oddity.

yadij:
> Please do not merge this PR until you have a chance to check ./configure output for errors _after_ #1151 and #1152 land.

Done. Checked and besides the discovery of LIBCAP_BROKEN being broken it appears fine.

configure.ac
yadij: Most other libraries are pre-pending foo_PATH into foo_LIBS instead of global LDFLAGS. Testing that now.

rousskov (approved):

Disclaimer: I did not check low-level PR details.

PR #1141, opened 2022-09-11
Fix 'warning: stray \ before -'

yadij:
grep 3.8 started detecting overly-escaped regex patterns

### DIFF
--- a/errors/Makefile.am
+++ b/errors/Makefile.am
@@ -53,7 +53,7 @@ translate-warn:
 
 $(TRANSLATE_LANGUAGES): $(ERROR_TEMPLATES)
 
-NOTIDY=`$(PO2HTML) --help | grep -o "\--notidy"`
+NOTIDY=`$(PO2HTML) --help | grep -o "[-]-notidy"`
 
 .po.lang:
 	if test "$(PO2HTML)" != "" && test "$(PO2HTML)" != "no" && test "$(PO2HTML)" != "off" && test -f $(top_srcdir)/errors/en.po; then \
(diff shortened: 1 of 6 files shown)


errors/Makefile.am
```diff
@@ -53,7 +53,7 @@ translate-warn:
 
 $(TRANSLATE_LANGUAGES): $(ERROR_TEMPLATES)
 
-NOTIDY=`$(PO2HTML) --help | grep -o "\-\-notidy"`
+NOTIDY=`$(PO2HTML) --help | grep -o "\--notidy"`
```
rousskov, line 56: While grep has [added](https://lists.nongnu.org/archive/html/bug-gnulib/2022-06/txtL0fZB_K0xd.txt)  special code to suppress the stray backslash warning in the beginning of the regex, I think we should stop escaping what should not be escaped instead of relying on that special handling:

```suggestion
NOTIDY=`$(PO2HTML) --help | grep -o "[-]-notidy"`
```

or

```suggestion
NOTIDY=`$(PO2HTML) --help | grep -o -- "--notidy"`
```

or

```suggestion
NOTIDY=`$(PO2HTML) --help | grep -o -e "--notidy"`
```

rousskov (commented):
There are other places where stray backslashes should be removed from grep patterns. I recommend fixing all of them, even though they may not trigger warnings right now, but I do not insist on any changes.

errors/Makefile.am
yadij: Going with the "[-]" form since the ```-e``` and ```--``` are not universally accepted by non-GNU grep implementations.

rousskov (approved):
Thank you.

PR #1143, opened 2022-09-11
Cleanup libexpat variables

yadij:
... to use pkg-config compatible naming and take advantage of
squid provided macros better.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -815,25 +815,23 @@ AC_ARG_ENABLE(esi,
                   Enabling ESI will cause squid reverse proxies to be capable
                   of the Edge Acceleration Specification (www.esi.org).]),
                  [squid_opt_use_esi=$enableval],[])
-HAVE_LIBEXPAT=0
-EXPATLIB=
 
 # ESI support libraries: expat
+AH_TEMPLATE(HAVE_LIBEXPAT,[Define to 1 if you have the expat library])
 SQUID_AUTO_LIB(expat,[ESI expat library],[LIBEXPAT])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -818,27 +818,25 @@ AC_ARG_ENABLE(esi,
...
+    LIBEXPAT_LIBS="-lexpat"
+    AC_CHECK_HEADERS([expat.h])
+    AC_DEFINE(HAVE_LIBEXPAT,1,[Define to 1 if you have the expat library])
+  ])
+  AS_IF([test "x$LIBEXPAT_LIBS" != "x"],[squid_opt_use_esi=yes],
+    [test "x$with_expat" = "xyes"],[AC_MSG_ERROR([Required library expat not found.])],
+    [AC_MSG_NOTICE([Library expat not found.])]
+  )
 ])
+AM_CONDITIONAL(ENABLE_LIBEXPAT,true)
```
rousskov, line 838: The always-true condition for enabling libexpat does not look right. Please test where libexpat should _not_ be enabled.

rousskov (changes requested):

configure.ac
yadij: Thanks. Fixed in [1728cd6](https://github.com/squid-cache/squid/pull/1143/commits/1728cd6dd4dc655ba0a0c11eb36ffc09670b9fcc)

rousskov (approved):
Since this PR now adds a pkg-config check as well, I would adjust the PR title as you did for the libnettle PR.

yadij:
FTR, just waiting until the overlapping libxml changes are merged. Will clear for Anubis after resolving the resulting conflicts.

PR #1144, opened 2022-09-11
Cleanup libnettle detection

yadij:
... to use pkg-config compatible variable naming and break out
the code test to a dedicated macro per the style used by other
libraries.

Also, our libcompat provides local replacement for Base64 on
builds with older libnettle. Fix the library link order to
ensure our functions get linked properly.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -514,7 +514,6 @@ squid_LDADD = \
 	debug/libdebug.la \
 	$(XTRA_OBJS) \
 	$(REPL_OBJS) \
-	$(NETTLELIB) \
 	$(CRYPTLIB) \
 	$(REGEXLIB) \
 	$(ADAPTATION_LIBS) \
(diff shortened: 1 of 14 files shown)


src/Makefile.am
```diff
@@ -2109,10 +2109,10 @@ tests_testHttpReply_LDADD=\
 	$(top_builddir)/lib/libmisccontainers.la \
 	$(top_builddir)/lib/libmiscencoding.la \
 	$(top_builddir)/lib/libmiscutil.la \
-	$(NETTLELIB) \
 	$(SSLLIB) \
 	$(LIBCPPUNIT_LIBS) \
 	$(COMPAT_LIB) \
+	$(LIBNETTLE_LIBS) \
```
rousskov, line 2115: Do we need to change the order of LDADD entries? If yes, perhaps the PR description should explain why? If this and similar change are accidental, then perhaps we should keep the official order in this PR? This is just a comment. I do not know enough to insist on any changes.

yadij: Sure. Updated description.

rousskov (approved):
Thank you for addressing my concern. I edited PR description to fix a typo.

PR #1145, opened 2022-09-16
Cleanup libxml2 variables

yadij:
... to use pkg-config compatible naming and take advantage of
squid provided macros better.

Remove unnecessary path detection of /usr vs /usr/local
locations. Admin can use --with-foo=PATH to provide any needed
customization of those from whichever is standard for the build
environment.


rousskov (approved):

PR #1146, opened 2022-09-17
Do not check unused RECV_ARG_TYPE

kinkie:
The check and corresponding defined preprocessor value
are not used anywhere in the code; remove the check


yadij (approved):

yadij:
If you can find any info on why this check existed at all it would be worth mentioning in the description before merge happens. Please feel free to mark as cleared for merge if you don't want to research that.

rousskov (approved):

kinkie:
I suspect it was added as a preliminary step to support windows: in that part of the code, the second argument to recv is cast to char*, where everywhere else it is a void*.
We'll see when we get serious about windows support, if needed.
There's probably cleaner ways to support this anyway when we will get serious about windows support

PR #1147, opened 2022-09-17
Cleanup libsystemd variables

yadij:
Rename SYSTEMD_LIBS and related `*_CFLAGS` / `*_PATH` variables to
`LIBSYSTEMD_*` prefix in line with other libraries naming style.

Polish detection log outputs in line with other libraries.

Fix pollution of global LDFLAGS variable.

Also, expand possible -lsystemd builds beyond Linux. Some OS provide a
library for compatibility. Cross-builds can also need to link against
systemd.


rousskov (approved):
I adjusted PR description to make all `_` characters visible in GitHub rendering and to fix "in line" spelling.

PR #1150, opened 2022-09-19
Fix 'warning: stray \ before white space'

yadij:
grep 3.8 started detecting overly-escaped regex patterns


yadij:
Missed this warning in PR #1141 as it showed up in the test logs, not the build output itself.

PR #1153, opened 2022-09-19
Fix ./configure detection of libxml2 headers

rousskov:
checking for LIBXML2... yes
    checking libxml/parser.h usability... yes
    checking libxml/parser.h presence... no
    configure: WARNING: libxml/parser.h: accepted by the compiler,
               rejected by the preprocessor!
    configure: WARNING: libxml/parser.h: proceeding with the compiler's
               result

PKG_CHECK_MODULES() documentation warns that it uses a "misleading" name
for the foo_CFLAGS variable it sets: "the variable provides the flags to
pass to the preprocessor, rather than the compiler". Lots of Squid code
has been mislead by that macro, but this fix is specific to a recent
regression introduced in commit 866a092. TODO: Fix all others.


yadij:
FYI, I have a change upcoming in the autoconf macro cleanup that will fix this and most others

yadij (approved):

PR #1155, opened 2022-09-22
Report (problems with) DEAD-at-start cache_peers

rousskov:
When a cache_peer is (re)configured, Squid probes its address using a
throw-away TCP connection. If and only if that TCP probe succeeds, peer
selection algorithms may pick the peer. Until then, the cache_peer is
treated as if it was DEAD.

This change preserves the logic summarized above but starts _reporting_
the initial probe failure and the fact that it marks the cache_peer
DEAD. Without these reports, admins often naturally assume that the
cache_peer is alive, especially if traffic can be served without it.

### DIFF
--- a/src/CachePeer.h
+++ b/src/CachePeer.h
@@ -143,7 +143,11 @@ public:
     char *digest_url = nullptr;
 #endif
 
-    int tcp_up = 0;         /* 0 if a connect() fails */
+    /// The number of failures sufficient to stop selecting this cache_peer. All
+    /// peer selection algorithms skip peers with zero tcp_up values. The
+    /// initial zero value prevents unprobed cache_peers from being selected.
+    int tcp_up = 0;
+
     /// whether to do another TCP probe after current TCP probes
     bool reprobe = false;
 
--- a/src/CachePeer.h
+++ b/src/CachePeer.h
@@ -144,8 +144,8 @@ public:
 #endif
 
     /// The number of failures sufficient to stop selecting this cache_peer. All
-    /// peer selection algorithms skip peers with zero tcp_up values. The
-    /// initial zero value prevents unprobed cache_peers from being selected.
+    /// cache_peer selection algorithms skip cache_peers with 0 tcp_up values.
+    /// The initial 0 value prevents unprobed cache_peers from being selected.
     int tcp_up = 0;
 
     /// whether to do another TCP probe after current TCP probes
--- a/src/neighbors.cc
+++ b/src/neighbors.cc
@@ -1275,30 +1275,33 @@ peerRefreshDNS(void *data)
     eventAddIsh("peerRefreshDNS", peerRefreshDNS, nullptr, 3600.0, 1);
 }
 
-static void
-peerConnectFailedSilent(CachePeer * p)
+// TODO: Move to CachePeer::noteFailure() or similar.
+// TODO: Require callers to detail failures instead of using one (and often
+// misleading!) "TCP" label for all of them.
+void
+peerConnectFailed(CachePeer * const p)
 {
     p->stats.last_connect_failure = squid_curtime;
+    if (p->tcp_up > 0)
+        --p->tcp_up;
 
-    if (!p->tcp_up) {
-        debugs(15, 2, "TCP connection to " << p->host << "/" << p->http_port <<
-               " dead");
-        return;
+    // TODO: Report peer name. Same-addresses peers often have different names.
+
+    const auto consideredAliveByAdmin = p->stats.logged_state == PEER_ALIVE;
+    const auto level = consideredAliveByAdmin ? DBG_IMPORTANT : 2;
+    debugs(15, level, "ERROR: TCP connection to " << p->host << "/" << p->http_port << " failed");
+
+    if (consideredAliveByAdmin) {
+        if (!p->tcp_up) {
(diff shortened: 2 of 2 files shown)


src/neighbors.cc
```diff
@@ -1275,32 +1275,35 @@ peerRefreshDNS(void *data)
     eventAddIsh("peerRefreshDNS", peerRefreshDNS, nullptr, 3600.0, 1);
 }
 
-static void
-peerConnectFailedSilent(CachePeer * p)
+// TODO: Move to CachePeer::noteFailure() or similar.
+// TODO: Require callers to detail failures instead of using one (and often
+// misleading!) "TCP" label for all of them.
```
rousskov, line 1280: We will convert this function into a method when improving reporting of the "TCP connection ... failed" problems. That work deserves a dedicated PR.

src/neighbors.cc
```diff
@@ -1275,32 +1275,35 @@ peerRefreshDNS(void *data)
...
     p->stats.last_connect_failure = squid_curtime;
+    if (p->tcp_up > 0)
+        --p->tcp_up;
 
-    if (!p->tcp_up) {
-        debugs(15, 2, "TCP connection to " << p->host << "/" << p->http_port <<
-               " dead");
-        return;
-    }
+    // TODO: Report peer name. Same-addresses peers often have different names.
```
rousskov, line 1288: We actually wasted some time during triage because the admin thought that Squid reports unexpected multiple failures for the same cache_peer. Turns out they were expected single/isolated failures associated with different same-address cache_peers (which is a perfectly valid configuration).

I can address this TODO in this PR if anybody insists on that, but doing so will change level-1 messages that will be changed again when we start detailing these failures (see the other TODOs above) and will probably trigger changes beyond this function. I think it is best not to address this problem in this PR.

src/neighbors.cc
```diff
@@ -1275,32 +1275,35 @@ peerRefreshDNS(void *data)
...
 
-    -- p->tcp_up;
+    const auto consideredAliveByAdmin = p->stats.logged_state == PEER_ALIVE;
```
rousskov, line 1290: When consideredAliveByAdmin is true and tcp_up is zero, Squid itself treats the cache_peer as "dead" or "down" -- the peer selection algorithms will not pick this peer. However, the admin has not been informed about the problem yet. These code changes do not change how Squid tracks/marks dead and "unprobed" peers. They only fix problem _reporting_. There is probably a better way to track peer state, but improving that is outside this (reporting-)bug-fix PR.

src/CachePeer.h
```diff
@@ -143,7 +143,11 @@ class CachePeer
     char *digest_url = nullptr;
 #endif
 
-    int tcp_up = 0;         /* 0 if a connect() fails */
+    /// The number of failures sufficient to stop selecting this cache_peer. All
+    /// peer selection algorithms skip peers with zero tcp_up values. The
```
eduard-bagdasaryan, line 147: Just for consistency:
```suggestion
    /// peer selection algorithms skip cache_peers with zero tcp_up values. The
```
or use 'peer' in all three lines.

src/neighbors.cc
yadij: What do you mean by this exactly? the p->name is already reported on L#1296:

>            debugs(15, DBG_IMPORTANT, "Detected DEAD " << neighborTypeStr(p) << ": " << p->name);


AFAICS, altering the "is still DEAD" debugs output to include the p->name is all that is necessary to meet this TODO.

src/CachePeer.h
yadij: Seems to me all three should be "cache_peer". While the latter two could be interpreted as plural they are used as one piece of a compound noun.

rousskov: Commit f8de58d changed all "peer(s)" to "cache_peer(s)". I hope we do not need to argue about pieces of compound nouns in this context!

src/neighbors.cc
rousskov: > What do you mean by this exactly?

Sorry, I do not know which "this" you are asking about.

> the p->name is already reported on L#1296:

... but not in some other contexts. Here are all the ones I could find quickly (citing here their PR variants, but these or equivalent debugs() are present in the official code as well):

```C++
debugs(38, 3, "netdbExchangeHandleReply: for '" << ex->p->host << ":" << ex->p->http_port << "'");
debugs(15, Important(29), "Configuring " << neighborTypeStr(p) << " " << p->host << "/" << p->http_port << "/" << p->icp.port);
debugs(15, level, "ERROR: TCP connection to " << p->host << "/" << p->http_port << " failed");
debugs(15, 2, "cache_peer " << p->host << "/" << p->http_port << " is still DEAD");
debugs(15, 2, "TCP connection to " << p->host << "/" << p->http_port << " succeeded");
debugs(72, 3, "peerDigestCheck: peer " <<  pd->peer->host << ":" << pd->peer->http_port);
debugs(15, DBG_IMPORTANT, "WARNING: Peer looks like this host." <<
       Debug::Extra << "Ignoring " <<
       neighborTypeStr(thisPeer) << " " << thisPeer->host <<
       "/" << thisPeer->http_port << "/" <<
       thisPeer->icp.port);
```

Ideally, a dedicated PR should make CachePeer reporting consistent and convenient while avoid code duplication and other problems. I volunteer to post such a PR. However, again, if you insist that we do this out-of-scope polishing here, I will do it here.

PR #1156, opened 2022-09-26
Redux broken krb5.h handling

yadij:
De-duplicate the #include sequence into a libcompat header to
ensure all uses of the hack(s) including autoconf test macros
are using the same logic. This fixes some macros which were
using inconsistent logic for Solaris.

Simplify the if-elif use of __cplusplus detection. All the
extern C hacks apply only when C++ is used.

Wrapping the entire file peer_proxy_negotiate_auth.cc with
'extern C {...}' is unnecessary (it is a .cc) and breaks the
Solaris workaround.

### DIFF
--- a/compat/krb5.h
+++ b/compat/krb5.h
@@ -6,8 +6,8 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-#ifndef _SQUID__COMPAT_KRB5_H_
-#define _SQUID__COMPAT_KRB5_H_
+#ifndef SQUID_COMPAT_KRB5_H
+#define SQUID_COMPAT_KRB5_H
 
 /*
  * The Kerberos library krb5.h header file has various major
--- a/compat/krb5.h
+++ b/compat/krb5.h
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 1996-2022 The Squid Software Foundation and contributors
+ *
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef _SQUID__COMPAT_KRB5_H_
+#define _SQUID__COMPAT_KRB5_H_
+
+/*
+ * The Kerberos library krb5.h header file has various major
+ * bugs in some implementations:
+ *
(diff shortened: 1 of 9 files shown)


compat/krb5.h
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (C) 1996-2022 The Squid Software Foundation and contributors
+ *
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef _SQUID__COMPAT_KRB5_H_
+#define _SQUID__COMPAT_KRB5_H_
```
rousskov, line 10: We are not consistent across the code base, but the PR spelling is less frequently used and, IMO, is inferior to its simpler/safer equivalents that do not start and end with underscores (that resemble various "reserved" names we should not be using):

```suggestion
#ifndef SQUID_COMPAT_KRB5_H
#define SQUID_COMPAT_KRB5_H
```

There is one more macro name at the end of the file.

compat/krb5.h
```diff
@@ -0,0 +1,49 @@
...
+ * - MacOS marks as deprecated the POSIX compatible APIs.
+ *   Apparently to encourage code to use their internal APIs.
+ *
+ * - Heimdal may lack the extern "C" syntax for building in C++
+ *
+ * - Solaris incorrectly implements its own the extern "C" replacment
+ *   macros and as a result krb5.h contains a trailing '}'.
+ *   see http://bugs.opensolaris.org/bugdatabase/view_bug.do?bug_id=6837512
+ *
+ * This file exists to fix those issues as best we can and to
```
rousskov, line 25: Grammar nitpick:

```suggestion
 * This file exists to fix those issues the best we can and to
```

compat/krb5.h
```diff
@@ -0,0 +1,49 @@
...
+ * ensure the logic is identical in Squid code, Squid helpers,
+ * and autoconf tests. see acinclude/krb5.h for the later.
```
rousskov, line 27: Grammar(?) and spelling nitpicks:

```suggestion
 * and autoconf tests. See acinclude/krb5.h for the latter.
```

rousskov (approved):
I welcome the reduction of krn5.h-related code duplication but did not check the low-level details.

PR #1157, opened 2022-09-29
WIP: Polish checks for working Kerberos library

yadij:
These checks test whether the library chosen is able to be used
without any specific features Squid may have workarounds for.

Take advantage of AC_CACHE better to enable reentrant macros
instead of requiring order-specific calls and update to use the
simpler AC_LANG_PROGRAM macro.

Now that they are reentrant shuffle library-specific brokenness
checks inside SQUID_CHECK_WORKING_KRB5 macro and set the
relevant AC_DEFINE as-needed instead of in callers.

### DIFF
--- a/acinclude/krb5.m4
+++ b/acinclude/krb5.m4
@@ -8,60 +8,74 @@
 dnl these checks must be performed in the same order as here defined,
 dnl and have mostly been lifted out of an inlined configure.ac.
 
-dnl checks for a broken solaris header file, and sets squid_cv_broken_krb5_h
-dnl to yes if that's the case
-AC_DEFUN([SQUID_CHECK_KRB5_SOLARIS_BROKEN_KRB5_H], [
+dnl checks for a broken solaris header file
+AC_DEFUN([SQUID_CHECK_KRB5_SOLARIS_BROKEN_KRB5_H],[
   AC_CACHE_CHECK([for broken Solaris krb5.h],squid_cv_broken_krb5_h, [
     SQUID_STATE_SAVE(squid_krb5_solaris_test)
     CPPFLAGS="-I${srcdir:-.} $CPPFLAGS"
-    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
-#include <krb5.h>
-int i;
-]])], [ squid_cv_broken_krb5_h=no ], [
-    AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+    AC_COMPILE_IFELSE([
+      AC_LANG_PROGRAM([[#include <krb5.h>]],[[int i;]])
+    ],[squid_cv_broken_krb5_h=no],[
+      AC_COMPILE_IFELSE([
+        AC_LANG_PROGRAM([[
 #define HAVE_BROKEN_SOLARIS_KRB5_H  1
 #include "compat/krb5.h"
-int i;
-]])], [ squid_cv_broken_krb5_h=yes ], [ squid_cv_broken_krb5_h=no ])
+        ]],[[int i;]])
+      ],[
+        squid_cv_broken_krb5_h=yes
+        AC_MSG_WARN([You have a broken Solaris <krb5.h> system include.])
+        AC_MSG_WARN([Please see http://bugs.opensolaris.org/bugdatabase/view_bug.do?bug_id=6837512])
+        AC_MSG_WARN([If you need Kerberos support you will have to patch])
+        AC_MSG_WARN([your system. See contrib/solaris/solaris-krb5-include.patch])
+      ],[squid_cv_broken_krb5_h=no])
     ])
     SQUID_STATE_ROLLBACK(squid_krb5_solaris_test)
   ])
-]) dnl SQUID_CHECK_KRB5_SOLARIS_BROKEN_KRB5_H
+  AS_IF([test "x$squid_cv_broken_krb5_h" = "xyes"],[
+    AC_DEFINE(HAVE_BROKEN_SOLARIS_KRB5_H,1,[Define to 1 if Solaris krb5.h is broken for C++])
+  ])
+])
 
-
-AC_DEFUN([SQUID_CHECK_KRB5_HEIMDAL_BROKEN_KRB5_H], [
+AC_DEFUN([SQUID_CHECK_KRB5_HEIMDAL_BROKEN_KRB5_H],[
   AC_CACHE_CHECK([for broken Heimdal krb5.h],squid_cv_broken_heimdal_krb5_h, [
     SQUID_STATE_SAVE(squid_krb5_heimdal_test)
     CPPFLAGS="-I${srcdir:-.} $CPPFLAGS"
-    AC_RUN_IFELSE([AC_LANG_SOURCE([[
-#include <krb5.h>
-int
-main(void)
-{
-        krb5_context context;
-
-        krb5_init_context(&context);
-
-        return 0;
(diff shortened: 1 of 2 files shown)


acinclude/krb5.m4
```diff
@@ -8,125 +8,156 @@
...
+dnl check how to access the max skew in the krb5 context
+AC_DEFUN([SQUID_CHECK_KRB5_GET_MAX_TIME_SKEW],[
+  AC_CACHE_CHECK([for max_skew in struct krb5_context],squid_cv_max_skew_context,[
+    SQUID_STATE_SAVE(squid_krb5_test)
+    CPPFLAGS="-I${srcdir:-.} $CPPFLAGS"
+    AC_CHECK_LIB(krb5,[krb5_get_max_time_skew],[
+      squid_cv_max_skew_context="yes"],[
+      AC_COMPILE_IFELSE([
+        AC_LANG_PROGRAM([[#include "compat/krb5.h"]],[[krb5_context kc; kc->max_skew = 1;]])
+      ],[squid_cv_max_skew_context=replace],[squid_cv_max_skew_context=no],[squid_cv_max_skew_context=no])
```
rousskov, line 89: IMO, it is not clear _why_ the existence of krb5_context::max_skew member implies that we should risk _replacing_ the standard krb5_get_max_time_skew() API function with our own same-name macro. Should not we use krb5_get_max_time_skew() if it is available? I suspect we should not replace standard API functions, but if we should, then we probably need to explain why (with a source code comment).

rousskov, line 89: Maybe I am counting brackets incorrectly, but it looks like this AC_COMPILE_IFELSE() call has a total of four actual parameters instead of the three allowed by [autoconf](https://www.gnu.org/software/autoconf/manual/autoconf-2.71/html_node/Running-the-Compiler.html):

1. input (`AC_LANG_PROGRAM`)
2. action-if-true (`squid_cv_max_skew_context=replace`)
3. action-if-false (the first `squid_cv_max_skew_context=no`)
4. ??? (the second `squid_cv_max_skew_context=no`)

rousskov (changes requested):

acinclude/krb5.m4
rousskov: After noticing the surrounding AC_CHECK_LIB(krb, X) call, I now suspect that the intent here was not "replace their X with ours" but "provide our X because they are missing it". If this is true, then I suggest:

* renaming HAVE_MAX_SKEW_IN_KRB5_CONTEXT with PROVIDE_KRB5_GET_MAX_TIME_SKEW because we set HAVE_MAX_SKEW_IN_KRB5_CONTEXT _not_ when krb5_context::max_skew exists but when that member exists _and_ we need to use that member to provide krb5_get_max_time_skew()
* replacing `squid_cv_max_skew_context` and its "replace" value with `squid_cv_krb5_get_max_time_skew` and "provide" value, to avoid the implication that we are _replacing_ something (and that we are replacing something like `krb5_context::max_skew`).

... where X is krb5_get_max_time_skew(), of course.

I do _not_ insist on any _specific_ words/changes for this change request, but I do think the current code is confusing and can be easily improved along the lines I outlined above.

acinclude/krb5.m4
```diff
@@ -8,125 +8,156 @@
...
-    [ squid_cv_max_skew_context=no ])
+      ]],[[krb5_context context;
+        krb5_get_init_creds_opt *options = nullptr;
+        krb5_get_init_creds_opt_free(context, options)]])],
+    [squid_cv_krb5_get_init_creds_free_context=yes],
+    [squid_cv_krb5_get_init_creds_free_context=no],[:])
     SQUID_STATE_ROLLBACK(squid_krb5_test)
   ])
+  AS_IF([test "x$squid_cv_krb5_get_init_creds_free_context" = "xyes"],[
+    AC_DEFINE(HAVE_KRB5_GET_INIT_CREDS_FREE_CONTEXT,1,[Define to 1 if you krb5_get_init_creds_free requires krb5_context])
```
rousskov, line 117:
```suggestion
    AC_DEFINE(HAVE_KRB5_GET_INIT_CREDS_FREE_CONTEXT,1,[Define to 1 if krb5_get_init_creds_opt_free() requires krb5_context])
```

* "you" is a typo; probably meant to be "your", but the word is not really needed there
* There is no krb5_get_init_creds_free(). 

This is old moved code, so I do not insist on this fix. Your call.

acinclude/krb5.m4
yadij: Ah yes. mixup with AC_RUN_IFELSE. Fixed in [arc-autoconf-ng-13](https://github.com/squid-cache/squid/pull/1157/commits/5799bdfcd48d4f448769ee25af39e40529ef7b93) along with one other.

yadij: I have refactored so the code should be clearer that this is about *where* to get the skew value from. Please review the changes in https://github.com/squid-cache/squid/pull/1157/commits/5799bdfcd48d4f448769ee25af39e40529ef7b93 to see if they make more sense to you.

compat/krb5.h
```diff
@@ -46,4 +46,9 @@
 #  endif
 #endif /* HAVE_KRB5_H */
 
+#if !HAVE_KRB5_GET_MAX_TIME_SKEW && HAVE_MAX_SKEW_IN_KRB5_CONTEXT
+#define krb5_get_max_time_skew(x) (x)->max_skew
+#define HAVE_KRB5_GET_MAX_TIME_SKEW 1
+#endif
```
rousskov, line 52: I think this should be refactored to avoid redefining `HAVE_KRB5_GET_MAX_TIME_SKEW` and to avoid defining a macro with the standard library name. Here is an untested sketch:

```suggestion
/// krb5_get_max_time_skew() portability wrapper
/// \retval DEFAULT_SKEW if we do not know how to get the maximum
krb5_deltat
GetMaxTimeSkew(const krb5_context &context)
{
#if HAVE_KRB5_GET_MAX_TIME_SKEW
    return krb5_get_max_time_skew(&context);
#elif HAVE_MAX_SKEW_IN_KRB5_CONTEXT
    return context.max_skew;
#else
    (void)context;
    return DEFAULT_SKEW;
#endif
}
```

Please adjust as needed. The caller will need to be trivially adjusted as well, of course.

Alternatively, this simple logic should be moved into the caller itself (until it needs to be _reused_ elsewhere). 

The above GetMaxTimeSkew() suggestion makes the logic reusable but may require more work. I am OK with either approach.

acinclude/krb5.m4
rousskov: Very nice rewrite, thank you.

compat/krb5.h
yadij: Absolutely Not.
`krb5_get_max_time_skew(x)` is the official library API we are using in the main Squid code. This libcompat portability code is **only** providing a replacement hack for that API call as-needed when the distro build environment library does not properly present it (eg outdated library version).

rousskov: AFAICT, the facts in your response do not contradict and do not address my change request.

yadij:
This PR has taken so long to merge that the underlying Krb5 logic has had some significant changes and alternative bug fixes applied. I am breaking this down into refactor steps and submitting new PRs to do those.

compat/krb5.h
yadij: Skew hack has been dropped from recent Squid code.

PR #1158, opened 2022-09-29
Avoid negative test-builds.sh results due to lack of LDAP

rousskov:
TESTING: layer-02-maximus
    BUILD: test-suite/buildtests/layer-02-maximus.opts
    configure: error: Required library ldap not found

Broken by commit 0d57ed2.

### DIFF
--- a/test-suite/buildtests/layer-02-maximus.opts
+++ b/test-suite/buildtests/layer-02-maximus.opts
@@ -48,6 +48,7 @@ MAKETEST="distcheck"
 #   --with-gnugss \
 #   --with-heimdal-krb5 \
 #   --with-mit-krb5 \
+#   --with-ldap \
 #
 #   --enable-cpu-profiling \  Requires CPU support.
 #
@@ -111,7 +112,6 @@ DISTCHECK_CONFIGURE_FLAGS=" \
 	--with-gnu-ld \
 	--with-ipv6-split-stack \
 	--with-large-files \
-	--with-ldap \
 	--with-pic \
 	--with-pthreads \
 	--enable-build-info=squid\ test\ build \
--- a/test-suite/buildtests/layer-04-noauth-everything.opts
+++ b/test-suite/buildtests/layer-04-noauth-everything.opts
@@ -37,6 +37,7 @@ MAKETEST="distcheck"
 #   --with-gnugss \
 #   --with-heimdal-krb5 \
 #   --with-mit-krb5 \
+#   --with-ldap \
 #
 #	Following features require special support from other optional packages.
 #	We can't test them automatically everywhere without detecting those
@@ -113,7 +114,6 @@ DISTCHECK_CONFIGURE_FLAGS=" \
 	--with-gnu-ld \
 	--with-ipv6-split-stack \
 	--with-large-files \
-	--with-ldap \
 	--with-pic \
 	--with-pthreads \
 	--enable-build-info=squid\ test\ build \
--- a/test-suite/buildtests/layer-04-noauth-everything.opts
+++ b/test-suite/buildtests/layer-04-noauth-everything.opts
@@ -37,7 +37,6 @@ MAKETEST="distcheck"
 #   --with-gnugss \
 #   --with-heimdal-krb5 \
 #   --with-mit-krb5 \
-#   --with-ldap \
 #
 #	Following features require special support from other optional packages.
 #	We can't test them automatically everywhere without detecting those
@@ -48,6 +47,7 @@ MAKETEST="distcheck"
 #   --enable-kqueue \
 #   --enable-win32-service \
 #   --with-valgrind-debug \
+#   --with-ldap \
 #
 #   --enable-cpu-profiling \  Requires CPU support.
 #
--- a/.github/workflows/default.yaml
+++ b/.github/workflows/default.yaml
@@ -105,7 +105,6 @@ jobs:
           sudo apt-get --quiet=2 update
           sudo apt-get --quiet=2 install libtool-bin
           sudo apt-get --quiet=2 install libcppunit-dev
-          sudo apt-get --quiet=2 install libldap2-dev
           # TODO: Add some deb-src URI to sources.list and do build-dep
           # instead of the above installs?
           # sudo apt-get --quiet=2 build-dep squid


yadij:
Not specifically objecting. But do we really have to revert this part?
LDAP should be widely available and easy to install.

rousskov:
> But do we really have to revert this part? 

Yes, I believe we should: The guiding principle should remain to be something like "test-builds.sh should succeed in a supported environment [unless its failure exposes a Squid bug]". If we decide to downgrade that to something like "test-builds.sh should succeed in some supported environments [but may fail in other supported environments] unless...", then the script would lose its utility as the cornerstone of automated CI build testing -- we would not be able to simply run it in CI scripts to automatically detect build bugs. I see no good reason to sacrifice that key feature.

> LDAP should be widely available and easy to install.

That assertion should be irrelevant to this change/PR IMO. It may be relevant to our future configure.ac improvements like `--with-ldap=auto` and corresponding CI environment updates (we should install LDAP libraries on CI platforms where doing so is easy).

kinkie:
I agree with Alex on this one. The test builds should check what is the minimum viable system environment Squid supports. It's fine to declare that LDAP is mandatory to build Squid, but we would need to be explicit about it.

yadij:
> > But do we really have to revert this part?
> 
> Yes, I believe we should: The guiding principle should remain to be something like "test-builds.sh should succeed in a supported environment [unless its failure exposes a Squid bug]".

IIRC that has never been the guiding principle for test-builds.sh. I understand how you could interpret what it does as meeting that though.
...

> If we decide to downgrade that to something like "test-builds.sh should succeed in some supported environments [but may fail in other supported environments] unless...",

Ironically that downgrade is exactly the situation this PR would cause.

By removing these "--with-ldap" any supported environment lacking LDAP that our CI tests will succeed and be presented to admin as working when their builds in the same environment *with* LDAP could be badly broken.

> 
> > LDAP should be widely available and easy to install.
> 
> That assertion should be irrelevant to this change/PR IMO. It may be relevant to our future configure.ac improvements like `--with-ldap=auto` and corresponding CI environment updates (we should install LDAP libraries on CI platforms where doing so is easy).

It is relevant to making CI have LDAP as a requirement so code using it can be consistently tested "in all supported environments".


In reply to kinkie:
> I agree with Alex on this one. The test builds should check what is the minimum viable system environment Squid supports. It's fine to declare that LDAP is mandatory to build Squid, but we would need to be explicit about it.

"Minimum viable system" is what layer-01-minimal tests. These layer-02-maximus and layer-04-noauth-everything are testing that Squid features are implemented properly and interact ok with each other when used.

Only features which are turned on in layer-02-maximus can be said to be tested "in all supported environments". Otherwise we have only tested that it can be (auto-)disabled at build.

rousskov:
TLDR: Let's restore the status quo[^1] quickly and then discuss any serious changes.

----

> > Alex: The guiding principle should remain to be something like "test-builds.sh should succeed in a supported environment [unless its failure exposes a Squid bug]".

> Amos: IIRC that has never been the guiding principle for test-builds.sh.

FWIW, it was my guiding principle IIRC; it still is. The existing comments in the official option files match it IMO. Constructive proposals for establishing or improving principles are always welcome, of course. Let's not waste time on arguing whether an undocumented principle existed -- it is not an argument anyone can win on merit.

> > Alex: If we decide to downgrade that to something like "test-builds.sh should succeed in some supported environments [but may fail in other supported environments] unless...",
 
> Amos: that downgrade is exactly the situation this PR would cause.

I disagree (as detailed below).

> By removing these "--with-ldap" any supported environment lacking LDAP that our CI tests will succeed and be presented to admin as working when their builds in the same environment _with_ LDAP could be badly broken.

That outcome does not contradict the guiding principle in question. Script success does _not_ imply the lack of bugs. Nothing can guarantee the lack of bugs, of course.

The improvements I [outlined](https://github.com/squid-cache/squid/pull/1158#issuecomment-1263030003) would prevent this sad outcome (while still following the same guiding principle).

> > > Amos: LDAP should be widely available and easy to install.
 
> > Alex: That assertion should be irrelevant to this change/PR IMO.
 
> Amos: It is relevant to making CI have LDAP as a requirement

The script was and (after this PR) will remain useful beyond our CI. Our CI requirements are (or should be) outside _this_ PR scope. I am not violently against installing LDAP in all our CIs, after a proper assessment, but that change in CI policy is a red herring in _this_ PR context.

> so code using it can be consistently tested "in all supported environments".

That "all" ideal cannot be reached, with or without the proposed change: "All supported environments" include environments without LDAP (that we will not fully test if all our CI environments have LDAP!). We decide what gets into our CI environments, and we decide how smart the script is about LDAP, but there will always be cases that we do not cover. To reduce mess and surprises, we need to follow good guiding principles when picking which cases we should be testing. If you can improve those principles, let's discuss those improvements (hopefully without delaying this quick trivial fix for a recent change[^1]).

[^1]: We may disagree whether that recent change is a bug/regression or a controversial improvement, but its exact classification should not really matter in this particular case. It is a surprise and a controversial change. It was fast-tracked without an in-depth review. If we want to reduce overheads, we should both fast track uncontroversial no-known-problems PRs _and_ fast track their quick fixes like this one. Delaying the latter would preclude the former by increasing the risk of spending a lot of time on undoing unwanted/unexpected (by some) consequences of fast-tracked PRs.

> Francesco: I agree with Alex on this one. The test builds should check what is the minimum viable system environment Squid supports. It's fine to declare that LDAP is mandatory to build Squid, but we would need to be explicit about it.

> Amos: "Minimum viable system" is what layer-01-minimal tests. These layer-02-maximus and layer-04-noauth-everything are testing that Squid features are implemented properly and interact ok with each other when used.

Amos' correction is accurate but irrelevant to this PR. Francesco's wording was too narrow, but its overall premise is sound -- if we want to make LDAP mandatory, let's discuss, but let's not delay this quick trivial fix of a recent change[^1].

> Only features which are turned on in layer-02-maximus can be said to be tested "in all supported environments". 

I disagree. Those features can only be said to be "tested" in _our_ CI environments. Our CI environments are not "all supported environments".

> Otherwise we have only tested that it can be (auto-)disabled at build.

We have tested more than that, but all those details and decisions are irrelevant to _this_ PR.

test-suite/buildtests/layer-02-maximus.opts
```diff
@@ -111,7 +112,6 @@ DISTCHECK_CONFIGURE_FLAGS=" \
 	--with-gnu-ld \
 	--with-ipv6-split-stack \
 	--with-large-files \
-	--with-ldap \
```
yadij, line 114: Status quo would be to change this to "--without-ldap"

test-suite/buildtests/layer-04-noauth-everything.opts
```diff
@@ -113,7 +114,6 @@ DISTCHECK_CONFIGURE_FLAGS=" \
 	--with-gnu-ld \
 	--with-ipv6-split-stack \
 	--with-large-files \
-	--with-ldap \
```
yadij, line 116: Status quo would be to change this to "--without-ldap"

yadij:
> TLDR: Let's restore the status quo[^1] quickly and then discuss any serious changes.

Status quo is to build without LDAP and not test any of the code inside LDAP related #if/#endif


----

> > Alex: The guiding principle should remain to be something like "test-builds.sh should succeed in a supported environment [unless its failure exposes a Squid bug]".

> Amos: IIRC that has never been the guiding principle for test-builds.sh.

> Alex: FWIW, it was my guiding principle IIRC; it still is. The existing comments in the official option files match it IMO. Constructive proposals for establishing or improving principles are always welcome, of course. Let's not waste time on arguing whether an undocumented principle existed -- it is not an argument anyone can win on merit.

That is your opinion sure. As the original author of that script and its design - it had *other* guiding principles. Lets agree to disagree on that.
 FTR: documentation is at https://wiki.squid-cache.org/Features/CodeTestBed, and no we were not talking in terms of principles back then. You may have missed it ```Defines everything which may be enabled to ON. ```


> > Alex: If we decide to downgrade that to something like "test-builds.sh should succeed in some supported environments [but may fail in other supported environments] unless...",
> Amos: that downgrade is exactly the situation this PR would cause.
> Alex: I disagree (as detailed below).

After merging this PR test-builds.sh will fail to report bugs in LDAP code for some supported environments. That is a fact, not opinion.


> Amos: By removing these "--with-ldap" any supported environment lacking LDAP that our CI tests will succeed and be presented to admin as working when their builds in the same environment _with_ LDAP could be badly broken.

> Alex: That outcome does not contradict the guiding principle in question. Script success does _not_ imply the lack of bugs. Nothing can guarantee the lack of bugs, of course.

Reminder: the guarantee that we / Squid Project / Squid Foundation offers is that the published /released code *will build* on supported environments with whatever ./configure options admin choose to use. This PR violates that by making the CI system ignore build errors in supported environments _using_ LDAP.

1) As I mentioned what you have called *the* guiding principle - is not.
  - the actual guiding principle is to test *every* permutation of Squid features. Two new ./configure options are added in 0d57ed2e8fd8a9721e7577e4deeb2a486972db9a and their use needs testing by the CI.

2) that response was to your statement that accepting 0d57ed2e8fd8a9721e7577e4deeb2a486972db9a behaviour requires a "downgrade" in guiding principle.
  - status-quo was that script pre-0d57ed2e8fd8a9721e7577e4deeb2a486972db9a contained a bug which CI silently skips testing of all LDAP related code.
  -  this PR is a proposal to reinstate that bug as official testing policy
  - therefore; truth as demonstrated, this PR is the change requiring an alteration in principles.


> > > Amos: LDAP should be widely available and easy to install.
...
> I am not violently against installing LDAP in all our CIs, after a proper assessment, but that change in CI policy is a red herring in _this_ PR context.

I do not think so. This comes back to the scripts design. It was initially created to *be* the CI system. It still forms the very core and bulk of what Jenkins does for testing.

Which is why I question the need for this PR change and its implied change of what the script can ignore. It will cripple the CI ability to accurately provide guarantees of buildable code on the group of "some supported environments" lacking LDAP.
Installing LDAP to our official build nodes and relying on layer-00-default.opts result is not enough because that layer is the one which intentionally silences broken optional libraries.


> Amos: so code using it can be consistently tested "in all supported environments".

> Alex: That "all" ideal cannot be reached, with or without the proposed change: "All supported environments" include environments without LDAP (that we will not fully test if all our CI environments have LDAP!).

 * correct build (and some operations) **without** LDAP is tested by layer-01-minimal for all supported environments. Regardless of what libraries they have installed.
 * correct build with auto-detection logic and default is tested by layer-00-default
 * the CI system does not yet support testing any layer-03 permutations (required library not available).

rousskov (commented):
TLDR: The current set of possible test-builds.sh principles under discussion is:

1. Bugs notwithstanding, test-builds.sh should succeed in any supported environment. Consequently, the presence or absence of any _optional_ library should have no effect on test-builds.sh success; the layers should not blindly enable any features that cannot be built in some supported environments.
2. The goal of test-builds.sh maximus layer is to test builds with as many features enabled as feasible without violating the above principle. Consequently, the maximus layer should not disable any feature supported by the environment[^1]. Here, "maximus" stands for the very useful "maximum the current build environment can handle" rather than practically useless "everything".
3. Ideally, test-builds.sh should test every permutation of Squid 60+ features. Testing 2<sup>60+</sup> permutations is not practical, so we want to test as many permutations as possible in reasonable time, with a focus on the more important/likely combinations.

[^1]: One possible exception is when disabling feature X is required to enable a more valuable feature Y -- if there are multiple "local maximums", then we would have to pick winners and losers.

There is currently no consensus regarding the above principles.

----


> FTR: documentation is at https://wiki.squid-cache.org/Features/CodeTestBed, and no we were not talking in terms of principles back then. You may have missed it `Defines everything which may be enabled to ON. `

That phrase does not contradict the principles I have stated -- it can and should be interpreted without violating those principles.

 
> > > Alex: If we decide to downgrade that to something like "test-builds.sh should succeed in some supported environments [but may fail in other supported environments] unless...",
> > > Amos: that downgrade is exactly the situation this PR would cause.
> > > Alex: I disagree (as detailed below).
 
> After merging this PR test-builds.sh will fail to report bugs in LDAP code for some supported environments. That is a fact, not opinion.

Agreed. However, that fact does not downgrade any principle -- which is what I have stated. As far as LDAP is concerned, PR test-builds.sh follows all of the three principles restated above while the current official code violates principle 1.

> > Amos: By removing these "--with-ldap" any supported environment lacking LDAP that our CI tests will succeed and be presented to admin as working when their builds in the same environment _with_ LDAP could be badly broken.
> 
> > Alex: That outcome does not contradict the guiding principle in question. Script success does _not_ imply the lack of bugs. Nothing can guarantee the lack of bugs, of course.
> 
> Reminder: the guarantee that we / Squid Project / Squid Foundation offers is that the published /released code _will build_ on supported environments with whatever ./configure options admin choose to use. 

There is no such guarantee, and it is not possible to guarantee that: In any given supported environment, using some ./configure options will result in a failed build because that environment does not support those options. No environment supports all ./configure options!


> This PR violates that by making the CI system ignore build errors in supported environments _using_ LDAP.

Even if we assume, for the sake of the argument, the existence of that guarantee, this PR does not violate it. Moreover, if our CI system does absolutely nothing, then we could still give that guarantee! In that hypothetical extreme case we would simply be relying exclusively on admins to report bugs.


> the actual guiding principle is to test _every_ permutation of Squid features.

I agree that it is desirable, in principle, to test every permutation, but that goal is so vast and reaching it is so impractical, that I doubt it offers a lot of guidance. Said that, that goal does not contradict the principles I have stated. They are all complimentary, not mutually exclusive. I have added a variation of this goal as principle 3.


> Two new ./configure options are added in [0d57ed2](https://github.com/squid-cache/squid/commit/0d57ed2e8fd8a9721e7577e4deeb2a486972db9a) and their use needs testing by the CI.

Agreed. We already test `--without-ldap`. Currently, we are unable to properly test `--with-ldap` in the CI environment. That's OK -- you have tested it manually, and our CI is currently unable to properly test most `--with-foo` options. Overall, we are not making things worse here and, eventually, we will improve our CI configurations and CI scripts to test those options.

 
> * status-quo was that script pre-0d57ed2e8fd8a9721e7577e4deeb2a486972db9a contained a bug which CI silently skips testing of all LDAP related code.

I disagree that the pre-0d57ed2e8fd8a9721e7577e4deeb2a486972db9a script was buggy. Not testing LDAP code is not a bug. It is a limitation of a given CI setup and scripts. It does not violate any principles or guarantees. We do not test everything we want to test. On the other hand, failing when LDAP is not available is a bug because it violates principle 1.


> * this PR is a proposal to reinstate that bug as official testing policy

I disagree that the pre-0d57ed2e8fd8a9721e7577e4deeb2a486972db9a script was buggy (as discussed immediately above). 

As for policy, let's simply agree _not_ treat this PR as policy setting or reinstating. This PR obviously exposed a disagreement regarding test-builds.sh design principles and led to the explicit formulation of two or three principles that, once finalized, may be helpful for future development. I am happy to continue that discussion to finalize those statements. I do not think we need to held this simple and useful change hostage to those discussions, and I appreciate you not doing so. Thank you.


> * therefore; truth as demonstrated, this PR is the change requiring an alteration in principles.

AFAICT, so far, no specific principle has been proposed that this PR alters or violates.


> I question the need for this PR change and its implied change of what the script can ignore.

The immediate practical need for this PR change is simple: Without it, test-builds.sh fails in supported environments where it previously ran just fine. We can and should separate any policy changes/discussions from that simple change/fix. Let's agree that this PR does not set any precedent, just like 0d57ed2e8fd8a9721e7577e4deeb2a486972db9a does not set any.


> It will cripple the CI ability to accurately provide guarantees of buildable code on the group of "some supported environments" lacking LDAP.

Since official test-builds.sh fails in environments lacking LDAP, I do not see how this PR (which replaces that failure with tests) cripples any related guarantees. Compared to pre-0d57ed2e8fd8a9721e7577e4deeb2a486972db9a code, this PR will not cripple anything. Commit 0d57ed2e8fd8a9721e7577e4deeb2a486972db9a breaks tests in some environments while improving tests in others; it is not a step forward in that regard; it is a step forward and a step backward. Once this PR is merged, we will be moving forward, just not as fast (in some directions) as we may wish, but strictly forward nonetheless. Moreover, the exposed design disagreements may help us design better CI configurations and tests! This is good progress.


> Installing LDAP to our official build nodes and relying on layer-00-default.opts result is not enough because that layer is the one which intentionally silences broken optional libraries.

Installing LDAP is not enough to test code that enables LDAP support by default. However, I think that (small) problem should be solved outside of test-builds.sh. We should add an _environment-specific_ script that checks whether test-builds.sh or ./configure enabled support for the libraries we expect to be auto-enabled in _that_ environment.

Please note that the current official code is "not enough" either: It does not test whether --with-ldap actually resulted in enabling LDAP support. "Not enough" is not a very good criteria because virtually no change is "enough" -- we can virtually always test more.

> > > Amos: so code using it can be consistently tested "in all supported environments".

> > Alex: That "all" ideal cannot be reached, with or without the proposed change: "All supported environments" include environments without LDAP (that we will not fully test if all our CI environments have LDAP!).

> * correct build (and some operations) **without** LDAP is tested by layer-01-minimal for all supported environments.

Not really. The minimal layer does not test whether LDAP support is properly disabled by default when LDAP is not available. And if all our CI environments have LDAP, then no other layer will test that either! The "all" ideal cannot be reached in practice -- there are just too many permutations and not enough testing code. However, I doubt we should spend time arguing about it. I was just illustrating that there is value in some CI environments not having LDAP. I do not insist on us having such environments.

kinkie:
> 2\. The goal of test-builds.sh maximus layer is to test builds with as many features enabled as feasible without violating the above principle. Consequently, the maximus layer should not disable any feature supported by the environment[1](#user-content-fn-1-26403732debf31920b68826e081aa515). Here, "maximus" stands for the very useful "maximum the current build environment can handle" rather than practically useless "everything".

Actually, this is not my understanding of the maximus layer. That one, I understand as "test everything. It is up to the test environment to provide what's needed". Since test-builds.sh allows specifying test layers, it shouldn't be a showstopper for anyone who doesn't have a fully-loaded environment and isn't willing to set one up

rousskov:
> > 2. The goal of test-builds.sh maximus layer is to test builds with as many features enabled as feasible without violating the above principle. Consequently, the maximus layer should not disable any feature supported by the environment. Here, "maximus" stands for the very useful "maximum the current build environment can handle" rather than practically useless "everything".

> Actually, this is not my understanding of the maximus layer. That one, I understand as "test everything. It is up to the test environment to provide what's needed".

AFAICT, a "test everything" test will fail in every environment because no environment can provide _everything_ -- some features are mutually exclusive and some are OS-specific.

> Since test-builds.sh allows specifying test layers, it shouldn't be a showstopper for anyone who doesn't have a fully-loaded environment and isn't willing to set one up

I see two problems with this statement, but I will continue to focus on the second one because it should be easier/simpler to address: There is no such thing as a "fully-loaded environment" AFAICT. For example, no environment probably supports Windows-specific `--enable-win32-service`, Linux-specific `--with_netfilter_conntrack`, Solaris-specific `--enable-devpoll`, and OpenBSD-specific `--with-nat-devpf`. I am just quickly building a counter example here to illustrate the point; please ignore any minor discrepancies/problems.

What am I missing?

rousskov:
The discussions appeared to fizzle out seven days ago. I merged from master to sync the recently added workflow file (commit [c3ef6e1](https://github.com/squid-cache/squid/pull/1158/commits/c3ef6e198e1329d91a71bd2948baac7ea6b43329)) and plan to clear this PR for merging in 24 hours unless there are any last-minute objections.

We can (and should!) continue to discuss whether test-builds.sh may fail in supported environments and how the "maximus" layer should be configured -- those topics are not really specific to this simple PR.

yadij:
FTR, based on the logic presented we now need a followup PR enacting:
  "Fix ./test-builds.h in environments to always succeed when admin does not have the Squid source code"

yadij:
We should then also "fix" ./bootstrap.sh, ./configure, and Makefiles to always produce success when the build fails due to any reason.

rousskov:
> based on the logic presented we now need a followup PR enacting: "Fix ./test-builds.h in environments to always succeed when admin does not have the Squid source code"

> We should then also "fix" ./bootstrap.sh, ./configure, and Makefiles to always produce success when the build fails due to any reason.

I see no logical connection between the principles proposed earlier and the above suggestions. If the above conclusions were based on the first proposed principle, then please note that it explicitly does not cover test-builds.sh behavior in the presence of bugs:

> 1. Bugs notwithstanding, test-builds.sh should succeed in any supported environment.

The script should fail when there are bugs, of course. We can explicitly add such an obvious and obviously-unreachable goal to the list if doing so would reduce the number of wrong conclusions. However, I suspect that a more productive reaction would be to polish  principle 1 wording:

* Perhaps we should add text that explains that the principle does not imply that test-builds.sh should succeed when it detects Squid bugs? I was thinking that saying "bugs notwithstanding" is enough, but perhaps I was too optimistic.
* Perhaps we should add text that explains that the principle does not cover misconfigured environments? I was thinking that saying "supported environment" is enough, but perhaps I was too optimistic.
* I see no practical way to define "supported environment" exactly but perhaps it exists. To me, it is obvious that an environment without Squid sources is not "supported" in this context, but perhaps I am too optimistic. I also realize that there are several ways to provide Squid sources (e.g., git repository and bootstrapped tarballs) but I see no practical way to define exactly what kind of sources are required for the test-builds.sh execution environment to keep its "supported" designation. Would giving an _example_ of a supported environment help? I am worried that there will [always](https://github.com/measurement-factory/squid-notes/tree/start#doc1-it-is-possible-to-misapply-any-non-trivial-principle) be a way to say "you did not cover X or Y"...

rousskov:
> yadij changed the title [12 hours ago](https://github.com/squid-cache/squid/pull/1158#event-7567626597)
```diff
-Fix test-builds.sh in environments without LDAP
+Revert fix of test-builds.sh in environments without LDAP
```

I accidentally noticed the change in PR title and changed the title again, this time to a neutral variant that should not conflict with the opposite opinions regarding the meaning of this change (i.e. "bug fix" vs. "reversal of a bug fix").

```diff
-Revert fix of test-builds.sh in environments without LDAP
+Avoid test-builds.sh failures due to lack of LDAP
```

rousskov:
> Producing negative result is _not_ a "test-builds.sh failures". 

That is a matter of definition. Our definitions apparently differ. In this specific context, I see "failure" as a "negative result" synonym, but I do not insist on that specific interpretation and can use whatever terminology we can agree on.

> It is successful detection of a broken build environment.

I do not consider a build environment without LDAP "broken". I consider it one of the supported environment variations. IMO, bugs notwithstanding, test-builds.sh should succeed in supported build environments.

Moreover, I do not think test-builds.sh should be tasked with detecting broken build environments. It is not an environment test. It is a Squid test. IMO, test-builds.sh outcome in a broken build environment is, in general, undefined: Some specific breakages may result in test-builds.sh exiting with non-zero status code. Others will result in test-builds.sh exiting with zero status code. Neither outcome is guaranteed to be preserved across test-builds.sh executions and updates.

In hope to find an even more neutral wording, I have renamed the PR again:

```diff
-Avoid test-builds.sh failures due to lack of LDAP
+Avoid negative test-builds.sh results due to lack of LDAP
```
 
> As kinkie mentioned environments that lack test dependencies should be limited to only run the layers they are capable of. 

As I [mentioned](https://github.com/squid-cache/squid/pull/1158#issuecomment-1267580305), I disagree with Francesco's assertion. Moreover, I believe I have proven that all environments lack test dependencies, so such a precondition is impossible to satisfy and, hence, should not be used.

PR #1159, opened 2022-10-03
Fix GCC build on macos-11: -Wunused-parameter

rousskov:


kinkie (approved):
I wish we could be less spaghetti-ish.. :(
No reason to defer this going in

rousskov:
> I wish we could be less spaghetti-ish.. :( No reason to defer this going in

While I would not call this particular code spaghetti-like, I do agree that it can be refactored for clarity and quality. I cannot volunteer those improvements because I have to focus on far more important problems right now, but we will probably get there eventually. I also agree that such refactoring is outside this trivial build-fixing PR scope. Thank you for approving.

PR #1160, opened 2022-10-04
Start using Github Actions for CI tests

rousskov:
This change imports existing Sempaphore CI functionality test harness
and migrates it to Github Actions. Using a stand-alone harness offered
many advantages, but that approach goes against the current custom of
importing CI configurations into being-tested repositories. Semaphore CI
itself switched to such invasive configurations since v2. We have to
start using invasive CI configurations to be able to evolve Squid CI and
solve known CI problems.

While importing the test harness, we made basic CI functionality tests
available from the command line so that developers can run them even
before submitting a PR and while addressing reviewer concerns. They do
require internet access because the tests are downloaded from GitHub,
but they avoid permanent service installations in developer environment.
The tests do create/overwrite logs and start services like Squid. A lot
more work is still required to make these tests user friendly.

The new test configuration/harness consists of two primary parts:

* Github Actions "workflow" configuration, including embedded scripts.
* Various test scripts used by the "workflow" configuration above.

There are many ways to split the test harness among those two parts.
This particular split is based on these split-specific principles/ideas:

1. When running tests locally, developers do not want to risk
   disruptive, permanent, potentially dangerous actions such as globally
   installing new packages or overwriting checked out Squid sources.
   These actions belong to the workflow configuration file that is only
   interpreted by Github Actions, on Github virtual machines. The test
   scripts have only one sudo command (to start a Squid manipulating
   agent as nobody) which is bypassed of the agent is already running.

2. Developers should be able to repeat basic tests from the command
   line, against their version of Squid sources, their Squid builds, and
   without learning the ever-changing testing details. If their PR has
   failed a CI test, they may simple want to run the corresponding
   script locally to reproduce the issue. This means that nearly all
   testing logic belongs to the test scripts, not the Actions
   configuration file (that developers cannot easily interpret/apply).

We also thought that some developers would want to separate
functionality from code quality tests (because those are often applied
at different development stages) and to repeat (many times) just the
failed test case (because that often helps when triaging and fixing
bugs). These considerations resulted in the creation of two scripts,
each accepting individual test case names as optional parameters:

* test-suite/test-functionality.sh
* test-suite/test-sources.sh

### DIFF
--- a/.github/workflows/default.yaml
+++ b/.github/workflows/default.yaml
@@ -3,7 +3,7 @@ name: GitHub CI
 
 on:
   push:
-    branches: [ "master", "v5" ]
+    branches: [ "master", "v5", "auto" ]
 
   pull_request:
     branches: [ "master", "v5" ]


test-suite/test-sources.sh
```diff
@@ -0,0 +1,262 @@
...
+    # astyle's version.
+
+    # Avoid distracting $checker warnings; TODO: Fix $checker instead.
+    touch boilerplate_fix.sed
+
+    if ! run $checker $copyrightOption no
+    then
+        echo_error "Running $checker modified sources"
+        CHECK_OUTCOME_PHRASE="Ignored $checker failure" # maybe overwritten below
+        # TODO: Require source-maintenance.sh application instead of ignoring this error.
```
rousskov, line 162: This TODO and its sister TODO at the very end of this function deserve a dedicated policy-changing PR (which can also be used for discussing whether we are ready for requiring source-maintenance.sh application from contributors).

.github/workflows/default.yaml
```diff
@@ -0,0 +1,128 @@
...
+        if: runner.os == 'macOS'
+        run: |
+           brew install automake
+           brew install cppunit
+
+      - name: Checkout sources
+        uses: actions/checkout@v3
+
+      - run: ./test-builds.sh
+        # XXX: Fix distclean cleanup on macOS instead
```
rousskov, line 120: I posted #1159 that fix other _known_ macOS build errors, but I did not know enough how to fix the last remaining one:

```
ERROR: files left in build directory after distclean:
./src/cf_gen.dSYM/Contents/Resources/DWARF/cf_gen
./src/cf_gen.dSYM/Contents/Info.plist
make[1]: *** [distcleancheck] Error 1
make: *** [distcheck] Error 1
buildtest.sh result is 2
Log end: btlayer-00-default.log
Error: Process completed with exit code 1.
```

If somebody can address it, please do so in a separate PR. FWIW, other projects had to fix similar-looking [bugs](https://code.monotone.ca/p/monotone/issues/view/attachment/39/Makefile.am.diff).

test-suite/test-functionality.sh
```diff
@@ -0,0 +1,264 @@
...
+
+check_pconn() {
+    run_confirmed_test pconn
+}
+
+check_busy_restart() {
+    if ! run_confirmed_test busy-restart
+    then
+        # XXX: Make the test stable instead!
+        echo_warning "Ignoring unstable test failure: busy-restart"
```
rousskov, line 176: This PR was motivated, in part, by Amos request to make Semaphore CI tests more stable. AFAICT, most of Sempaphore CI failures could be attributed to two things:

* Problems/complications with the overall Semaphore CI setup. We are _not_ inheriting those problems here.
* An unstable proxy-collapsed-forwarding test. That test is stable enough in Factory lab, but not in Sempahore CI environment. It appears to be stable enough in Github Actions environment (which is much closer to what we use in the lab), which is the outcome I was hoping for.

This busy-restart test was stable in the Semaphore CI environment. It is stable in our lab as well. I am surprised it usually fails in Github Actions environment. Factory will work on making it stable in Github Actions environment. This test is not as complex as the proxy-collapsed-forwarding, and I am more optimistic that we can find a way to make its caching less susceptible to various natural race conditions/problems.

For now, I think it is best to keep it running but ignore its failures. The PR does that. If others insist, I will remove it instead.

test-suite/test-sources.sh
```diff
@@ -0,0 +1,262 @@
...
+    STARTING_POINT="refs/pull/${PULL_REQUEST_NUMBER}/merge^1"
+elif test -n "${FORK_POINT}"
+then
+    STARTING_POINT="${FORK_POINT}"
+fi
+
+# Customizable check outcome description logged by run_one_check()
+CHECK_OUTCOME_PHRASE=""
+
+# XXX: echo_*() and run() functions are duplicated in test-functionality.sh.
```
rousskov, line 28: We can remove this code duplication if we move duplicated functions into a third script that the two posted scripts will include by "sourcing" it. I am not sure whether the increased complexity is worth it, but am ready to implement that fix if others think it is a good idea. The code duplication will only become worse with time, of course.

test-suite/test-functionality.sh
```diff
@@ -0,0 +1,264 @@
...
+    sudo -n --background -u nobody perl < $SQUID_OVERLORD_DIR/overlord.pl > $log 2>&1 || return
+    echo "Started squid-overlord service at $url"
+}
+
+setup_test_tools() {
+    echo "::group::Setup test tools"
+
+    clone_repo https://github.com/measurement-factory/daft $DAFT_DIR || return
+    clone_repo https://github.com/measurement-factory/squid-dafts $SQUID_DAFTS_DIR || return
+    clone_repo https://github.com/measurement-factory/squid-overlord $SQUID_OVERLORD_DIR || return
```
rousskov, line 98: The above three repositories contain Factory test cases and the tools they are based on. I do not like the fact that we are hard-coding such a dependency into Squid repository, but I see no better way. This is one of the negative side effects of using invasive CI configurations that dominate CI tools these days...

One possible alternative is to move these scripts into our squid-ci repository and download them here. However, that feels like swimming against the current again: If we are going to use invasive configurations, then we should not essentially move big parts of those configurations into a different repository, recreating problems that invasive configurations are meant to solve/avoid!

test-suite/test-functionality.sh
```diff
@@ -0,0 +1,264 @@
...
+    local log=$TMPDIR/squid-overlord.log
+    if test -e $log && curl -H 'Pop-Version: 4' --no-progress-meter $url/check > /dev/null
+    then
+        echo "Will reuse squid-overlord service running at $url"
+        return 0;
+    fi
+
+    # Do not be tempted to simply run `sudo ... overlord.pl`: User nobody will
+    # lack read permissions, and sudo will ask for a password.
+    sudo -n --background -u nobody perl < $SQUID_OVERLORD_DIR/overlord.pl > $log 2>&1 || return
```
rousskov, line 89: This is the only `sudo` left in the test scripts. I do not see a good way of removing it because the overlord.pl process should run as the Squid user (i.e. `nobody`) and these scripts usually run as a Github Actions user or, when executed manually, as a developer.

This `sudo` is not a problem for Actions (it does not ask for a password there).

If this `sudo` step bothers a developer that runs this script manually and has a password-asking sudo setup, then they do have an option of running this command before starting the manual test -- the script will then detect the overlord.pl process running and skip this sudo step. The same trick is useful if one wants to run a custom overlord.pl version (which is useful during test case development). This trick is one of the reasons why the script does not try to kill the overlord.pl process at the end of the test.

.github/workflows/default.yaml
```diff
@@ -0,0 +1,128 @@
...
+
+    steps:
+
+      - name: Install prerequisite Linux packages
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get --quiet=2 update
+          sudo apt-get --quiet=2 install libtool-bin
+          sudo apt-get --quiet=2 install libcppunit-dev
+          sudo apt-get --quiet=2 install libldap2-dev
```
rousskov, line 108: This line is required until the test-builds.sh fix in #1158 lands. After that, we should either remove `libldap2-dev` or add every used-by-Squid library that we can easily add to this CI environment (and that is not already preinstalled by GitHub). That addition (if any) should be done in a separate PR, of course.

test-suite/test-sources.sh
```diff
@@ -0,0 +1,262 @@
...
+        return 0;
+    fi
+
+    if ! git diff --quiet
+    then
+        CHECK_OUTCOME_PHRASE="Skipped due to a dirty working directory"
+        return 0;
+    fi
+
+    # TODO: Remove this year-2020 workaround? Fresh master PRs should have Bug 5021 fix.
```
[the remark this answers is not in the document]
yadij: IMO drop it. Old pending PRs that might be affected either need author updates or a manual re-sync to master, so issues can be fixed at that time.

.github/workflows/default.yaml
yadij: MacOS is not one of the OS the Squid Foundation board has agreed on being considered "officially supported". It should not be required by SemaphoreCI at all until its build becomes bug-free and stable enough to join that set.

test-suite/test-functionality.sh
yadij: IMO moving those script to the squid-ci repository leaves the system in a better overall state. Especially since we (the full community) can then review and add to the test set. But agree that is not a good change for this PR. I recommend doing it as a followup though.

test-suite/test-functionality.sh
yadij: FYI, there are some tricks I have picked up from Debian people regarding solutions to safely kill the script on closure. We likely may even be able to use the Debian/Ubuntu developer tools instead of sudo. That is a change for later though.

.github/workflows/default.yaml
yadij: FTR Squid build environment OS providing apt-get is ```apt-get build-dep squid```. The ```apt-get install ...``` is only needed for unusual package dependencies such as those default-disabled by Squid or if the PR is adding a new one to the dependency set.

.github/workflows/default.yaml
yadij: see http://build.squid-cache.org/job/5-pr-test/ - CentOS, Debian, Fedora, OpenSuse, Ubuntu - latest three versions.

.github/workflows/default.yaml
rousskov: Thank you.

test-suite/test-functionality.sh
yadij: Thank you. Things are indeed clearer. I am in agreement with the current implementation going in as-is.

yadij (approved):
Approving as a good starting point.

The inclusion of MacOS as "officially supported" still worries me as it does not have any other testing performed and until very recently been a quite unstable build. For now it appears to be working though.

.github/workflows/default.yaml
yadij: That matrix list is the result of the recurring/regular discussions and analysis by the board members about which OS we are able to call "officially supported" as of the Squid-5 stable cycle.

Failure to build and test successfully on any of those listed OS blocks merge. Whereas issues seen on other environments **may** be ignored at reviewer discretion.

rousskov:
> The inclusion of MacOS as "officially supported" still worries me

This PR does not make MacOS officially supported. We should make that decision separately. MacOS is included in the workflow configuration because it illustrates GitHub Actions matrix testing capabilities. We may decide to officially support it, to completely remove it from the matrix, or even something in-between.

> Approving as a good starting point.

Thank you for approving. I will work with @eduard-bagdasaryan to make sure Anubis can merge this somewhat unusual PR.

.github/workflows/default.yaml
rousskov: > That matrix list is the result of the recurring/regular discussions and analysis by the board members
> 
> Failure to build and test successfully on any of those listed OS blocks merge. Whereas issues seen on other environments **may** be ignored at reviewer discretion.

I agree with the quoted parts. However, I do not think any of the above equates that Jenkins matrix with the "officially supported" designation (unless we have actually agreed to make that matrix equivalent to that designation, which would require, among other things, agreeing on never modifying that matrix without strong consensus). AFAICT, there is currently no agreement regarding the list of officially supported environments -- we still need to work that out. The lack of that agreement does not prevent Jenkins improvements and merging this PR, of course.

yadij: That matrix is the unified sum of a series discussions and decisions about what OS can (or not) be officially supported for the Squid Software Foundation testing and promises of stability for Squid release series. Each release series has its own N-pr-test matrix relevant to the time(s) it was released.
```
  Grass does not need a field to grow - Nevyn
```

.github/workflows/default.yaml
```diff
@@ -0,0 +1,131 @@
...
+        run: |
+           brew install automake
+           brew install cppunit
+
+      - name: Checkout sources
+        uses: actions/checkout@v3
+
+      - run: ./test-builds.sh
+        # XXX: Fix distclean cleanup on macOS instead
+        continue-on-error: ${{ runner.os == 'macOS' }}
```
rousskov, line 124: > > Alex: This PR does not make MacOS officially supported.

> My understanding is that the checks performed by this PR will be required to be passed before any PR following it is allowed to merge.

The reality is different: The MacOS test in this PR currently fails (due to bugs already present in the current official sources and preserved in this PR sources). Yet, that failure does not preclude this PR or any other PR from passing all the tests because that failure is explicitly ignored (the poorly named GitHub Actions `continue-on-error` directive has that effect).

> Having MacOS as part of the mandatory tests implicitly adds it to the "officially supported" OS set.

That does not happen in this PR because MacOS build failures are ignored.

> > Alex: We should make that decision separately. MacOS is included in the workflow configuration because it illustrates GitHub Actions matrix testing capabilities. We may decide to officially support it, to completely remove it from the matrix, or even something in-between.
 
> Amos: Then please drop MacOS from this PR before it merges. As a demonstration/test that we can check multiple OS it has served that purpose already. It can be re-added when a formal policy decision is made to support that OS.

Will do. Doing so will re-hide MacOS build failures from any developer that wants to fix them. I fixed a couple while this test was included, but I did not fix all of them, and do not plan on fixing any more, especially since they will become invisible again.

.github/workflows/default.yaml
rousskov: > That matrix is the unified sum of a series discussions and decisions about what OS can (or not) be officially supported for the Squid Software Foundation testing and promises of stability for Squid release series.

I disagree that the Jenkins matrix is equivalent to the "officially supported" designation. Thus, there is currently no consensus that it is equivalent. It may become equivalent if we explicitly decide that it is or if the explicit past decision making it equivalent is discovered. Until then, there is no official definition of "officially supported".

.github/workflows/default.yaml
rousskov: > please drop MacOS from this PR before it merges

Done at eb34c13.

rousskov:
> I will work with @eduard-bagdasaryan to make sure Anubis can merge this somewhat unusual PR.

_Status update_: GitHub is having problems with fast-forwarding this PR's staged commit. The commit exists and looks OK, but when Anubis tells GitHub to fast-forwarded master to that commit, GitHub API returns a 404 error. That is why Anubis labels this PR with `M-failed-staging-other`. We have not seen that kind of failure before and cannot explain it. My current plan is to squash all PR branch commits. If that does not help, I will try to fast-forward the staged commit manually.

rousskov:
> GitHub is having problems with fast-forwarding this PR's staged commit

FTR: Anubis was able to merge the PR after we found a way to manually bypass a small but necessary staging step (that was giving a still-unexplained and undocumented GitHub error) -- re-pointing the `auto` branch to the staged commit already created by Anubis. Hopefully, that new GitHub problem does not resurface any time soon.

PR #1161, opened 2022-10-05
Bug 5241: Block to-localhost, to-link-local requests by default

rousskov:
Squid suggested blocking to-localhost access since 2001 commit 4cc6eb1.
At that time, the default was not adjusted because some use cases were
known to require to-localhost access. However, the existence of special
cases should not affect defaults! The _default_ configuration should
either block all traffic or only allow traffic that is unlikely to
introduce new attack vectors into the network.

Also block to-link-local traffic (by default), for very similar reasons:
Popular cloud services use well-known IPv4 link-local (i.e. RFC 3927)
addresses (a.k.a. APIPA), to provide sensitive instance metadata
information, via HTTP, to instance users and scripts. Given cloud
popularity, those special addresses become nearly as ubiquitous as
127.0.0.1. Cloud provider networks shield metadata services from
external accesses, but proxies like Squid that forward external HTTP
requests may circumvent that protection.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -1071,9 +1071,9 @@ DEFAULT: all src all
 DEFAULT: manager url_regex -i ^cache_object:// +i ^[^:]+://[^/]+/squid-internal-mgr/
 DEFAULT: localhost src 127.0.0.1/32 ::1
 DEFAULT: to_localhost dst 127.0.0.0/8 0.0.0.0/32 ::1/128 ::/128
-DEFAULT: to_APIPA dst 169.254.0.0/16
+DEFAULT: to_linklocal dst 169.254.0.0/16
 DEFAULT: CONNECT method CONNECT
-DEFAULT_DOC: ACLs all, manager, localhost, to_localhost, to_APIPA, and CONNECT are predefined.
+DEFAULT_DOC: ACLs all, manager, localhost, to_localhost, to_linklocal, and CONNECT are predefined.
 DOC_START
 	Defining an Access List
 
@@ -1897,9 +1897,9 @@ http_access deny manager
 # assume that only local users can access them at "localhost" ports.
 http_access deny to_localhost
 
-# Protect cloud servers that often provide local users with sensitive info
-# about their server via certain well-known APIPA link-local addresses.
-http_access deny to_APIPA
+# Protect cloud servers that provide local users with sensitive info about
+# their server via certain well-known link-local (a.k.a. APIPA) addresses.
+http_access deny to_linklocal
 
 #
 # INSERT YOUR OWN RULE(S) HERE TO ALLOW ACCESS FROM YOUR CLIENTS
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -1071,8 +1071,9 @@ DEFAULT: all src all
 DEFAULT: manager url_regex -i ^cache_object:// +i ^[^:]+://[^/]+/squid-internal-mgr/
 DEFAULT: localhost src 127.0.0.1/32 ::1
 DEFAULT: to_localhost dst 127.0.0.0/8 0.0.0.0/32 ::1/128 ::/128
+DEFAULT: to_APIPA dst 169.254.0.0/16
 DEFAULT: CONNECT method CONNECT
-DEFAULT_DOC: ACLs all, manager, localhost, to_localhost, and CONNECT are predefined.
+DEFAULT_DOC: ACLs all, manager, localhost, to_localhost, to_APIPA, and CONNECT are predefined.
 DOC_START
 	Defining an Access List
 
@@ -1892,10 +1893,13 @@ http_access deny CONNECT !SSL_ports
 http_access allow localhost manager
 http_access deny manager
 
-# We strongly recommend the following be uncommented to protect innocent
-# web applications running on the proxy server who think the only
-# one who can access services on "localhost" is a local user
-#http_access deny to_localhost
+# Protect web applications running on the same server as Squid. They often
+# assume that only local users can access them at "localhost" ports.
+http_access deny to_localhost
+
+# Protect cloud servers that often provide local users with sensitive info
+# about their server via certain well-known APIPA link-local addresses.
+http_access deny to_APIPA
 
 #
 # INSERT YOUR OWN RULE(S) HERE TO ALLOW ACCESS FROM YOUR CLIENTS
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -1071,7 +1071,7 @@ DEFAULT: all src all
 DEFAULT: manager url_regex -i ^cache_object:// +i ^[^:]+://[^/]+/squid-internal-mgr/
 DEFAULT: localhost src 127.0.0.1/32 ::1
 DEFAULT: to_localhost dst 127.0.0.0/8 0.0.0.0/32 ::1/128 ::/128
-DEFAULT: to_linklocal dst 169.254.0.0/16
+DEFAULT: to_linklocal dst 169.254.0.0/16 fe80::/10
 DEFAULT: CONNECT method CONNECT
 DEFAULT_DOC: ACLs all, manager, localhost, to_localhost, to_linklocal, and CONNECT are predefined.
 DOC_START


src/cf.data.pre
```diff
@@ -1071,8 +1071,9 @@ DEFAULT: all src all
 DEFAULT: manager url_regex -i ^cache_object:// +i ^[^:]+://[^/]+/squid-internal-mgr/
 DEFAULT: localhost src 127.0.0.1/32 ::1
 DEFAULT: to_localhost dst 127.0.0.0/8 0.0.0.0/32 ::1/128 ::/128
+DEFAULT: to_APIPA dst 169.254.0.0/16
```
rousskov, line 1074: We should be matching link-local IPv6 addresses as well, right? I think they are covered by RFC 2462 which mentions `fe80::0` prefix for those, but AWS [uses](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) a ULA `fd00:ec2::254` address instead. Should we add something like the following?

```suggestion
DEFAULT: to_APIPA dst 169.254.0.0/16 fe80::/10
```

rousskov (commented):
Disclaimer: This is not my area of expertise. I will not merge this PR without an explicit approval from another core developer. Please feel free to adjust anything as needed.

Here is a relevant excerpt from an AWS EC2 [description](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) of the corresponding instance metadata service:

> Although you can only access instance metadata and user data from within the instance itself, the data is not protected by authentication or cryptographic methods. Anyone who has direct access to the instance, and potentially any software running on the instance, can view its metadata.

P.S. I am also not attached to the term "APIPA" in any way. Please use whatever term you think is best or let me know what to rename.

plygrnd:
@rousskov Dan from AWS here. Some comments below:

> I am also not attached to the term "APIPA"

Neither am I :) "link-local" would probably be more appropriate. You could call the rule `DEFAULT: to_linklocal` as well.

> (from Bugzilla) What about IPv6? AFAICT, EC2 uses fd00:ec2::254 for instance metadata service,
but that is not a link-local IPv6 address, is it?

It's not a link-local address; it's a [ULA](https://en.wikipedia.org/wiki/Unique_local_address) and is specific to EC2. If you block `fd00:ec2::254/128`, that'll work for EC2.

I can't speak for Microsoft or Google, but I know that Azure doesn't offer their IMDS via IPv6. Google Cloud CNAMEs theirs so you'd need to add `metadata.google.internal` in a deny rule.

src/cf.data.pre
plygrnd: Suggest adding `fd00:ec2::254/128` for EC2.

plygrnd: Also, `DEFAULT: to_linklocal` would work better IMO. I've learned that "APIPA" is outmoded terminology recently.

src/cf.data.pre
```diff
@@ -1892,10 +1893,13 @@ http_access deny CONNECT !SSL_ports
...
-# web applications running on the proxy server who think the only
-# one who can access services on "localhost" is a local user
-#http_access deny to_localhost
+# Protect web applications running on the same server as Squid. They often
+# assume that only local users can access them at "localhost" ports.
+http_access deny to_localhost
+
+# Protect cloud servers that often provide local users with sensitive info
+# about their server via certain well-known APIPA link-local addresses.
+http_access deny to_APIPA
```
plygrnd, line 1902: Suggest `http_access deny to_linklocal`

rousskov: Done in 6afb67e.

src/cf.data.pre
```diff
@@ -1071,8 +1071,9 @@ DEFAULT: all src all
 DEFAULT: manager url_regex -i ^cache_object:// +i ^[^:]+://[^/]+/squid-internal-mgr/
 DEFAULT: localhost src 127.0.0.1/32 ::1
 DEFAULT: to_localhost dst 127.0.0.0/8 0.0.0.0/32 ::1/128 ::/128
+DEFAULT: to_linklocal dst 169.254.0.0/16 fe80::/10
```
rousskov, line 1074: > @plygrnd: Suggest adding `fd00:ec2::254/128` for EC2.

That addition feels wrong to me because it is too specific to EC2 and because that address is not a link-local address.

We could add another ACL like `to_cloud_metaservice` that would match things like that, but I hope there is a more _general_ solution for the _default_ configuration than an ever-growing collection of company-specific addresses. Perhaps we can also deny access to all ULA addresses by default?

yadij:
IMO the same rationale should be used for site-local connections. That is the range FEC0::/10.

yadij:
Also, do you really want to return Squid to the default of performing an extra DNS lookup on every request?

You just went to all the trouble of removing the logformat ```%>A``` performance optimization to remove these early DNS costs.

src/cf.data.pre
yadij: An ACL definition of ```to_localnet``` using **dst** seems reasonable.

yadij (approved):
As of commit a72f4b28672ea21a8a0d939b5819c60720cfb0e6 looks good.

This review approval does not cover other suggestions or topics being discussed.

src/cf.data.pre
rousskov, line 1074: > Amos at https://github.com/squid-cache/squid/pull/1161#issuecomment-1269266477: IMO the same rationale should be used for site-local connections. That is the range FEC0::/10.

I disagree that [site-local](https://4sysops.com/archives/ipv6-tutorial-part-6-site-local-addresses-and-link-local-addresses/) addresses fall into the same category as the "localhost and equivalent" addresses this PR proposes to block by default. AFAICT, site-local are used primarily for communicating with other hosts (within the same organization). The proposed addition is not about that kind of communication. In other words, the proposed addition is not about "local to the organization" in general, but about "same-host" specifically.

Once could argue that denied link-local addresses are not _necessarily_ same-host either. That is a valid argument, but we are explicitly making an _exception_ there because of the widespread use of those addresses for a specific _same-host_ use case: sensitive instance metaservices in the cloud[^1]. Site-local addresses do not fall into this specific exception. They are also [deprecated](https://www.ietf.org/rfc/rfc3879.txt) since 2004. Overall, adding site-local would create more confusion about the intent of these rules without covering many actual use cases.

[^1]: If we had a more precise way to cover that common use case without creating a list of popular cloud services, we would have used that instead of to_linklocal.

src/cf.data.pre
```diff
@@ -1892,10 +1893,13 @@ http_access deny CONNECT !SSL_ports
 http_access allow localhost manager
 http_access deny manager
 
-# We strongly recommend the following be uncommented to protect innocent
-# web applications running on the proxy server who think the only
-# one who can access services on "localhost" is a local user
-#http_access deny to_localhost
+# Protect web applications running on the same server as Squid. They often
+# assume that only local users can access them at "localhost" ports.
+http_access deny to_localhost
```
rousskov, line 1898: > Amos at https://github.com/squid-cache/squid/pull/1161#issuecomment-1269273970: Also, do you really want to return Squid to the default of performing an extra DNS lookup on every request?

I do not _want_ that, but I think it is the right tradeoff for the _default_ configuration. As you know, I strive to make such decisions based on principles rather than whims of the moment. The specific principle here is [documented](https://github.com/squid-cache/squid/pull/1161#issue-1398153736) as "The default configuration should either block all traffic or only allow traffic that is unlikely to introduce new attack vectors into the network". In this context, the need for DNS queries and their performance impact is a secondary concern.

Eventually, Squid may gain a feature that would allow to achieve the same safety level without any DNS lookups. At that time, this default configuration could be adjusted to eliminate those lookups.

> You just went to all the trouble of removing the logformat %>A performance optimization to remove these early DNS costs.

It sounds like you are misinterpreting what happened in commit a8c7a11. That commit did not remove any performance optimizations. It gave the admin control over whether DNS lookups are performed while preserving the existing "early lookup" feature for virtually all use cases.

This new default configuration does not take the control restored in a8c7a11 away from the admin -- the admin should adjust the default to suit their needs, of course, including properly resolving various performance/safety tradeoffs.

src/cf.data.pre
rousskov: > An ACL definition of `to_localnet` using **dst** seems reasonable.

I agree, but I am not going to add it in this PR for the following (combined) reasons:

1. Adding `to_localnet` is outside this PR scope. If we add `to_localnet`, then we should probably add `linklocal` as well. It can easily snowball into lots of unrelated, distracting changes.
2. Adding `to_localnet` will duplicate a lot of non-trivial RFC-number-sensitive `localnet` configuration lines while hiding that duplication from the casual observer because it happens in different places many lines apart.
3. I am strongly against defining such "standard" ACLs in squid.conf and allowing (and even encouraging!) the admin to "adapt" them (as we do today with the `localnet` ACL). Creating two ACLs with an identical set of addresses, each independently "adaptable" by the admin only exacerbates related problems!

rousskov (commented):
> Amos: As of commit https://github.com/squid-cache/squid/commit/a72f4b28672ea21a8a0d939b5819c60720cfb0e6 looks good. This review approval does not cover other suggestions or topics being discussed.

I am interpreting the combination of the approval, the `S-waiting-for-committer` label, and the last sentence above as "please feel free to merge but let's continue ongoing discussions". If you do not want me not merge, please dismiss your PR approval. Otherwise, I will merge in a few days. Thank you.

I hope I did not miss an open discussions. I created change requests for several of them, to ease tracking.

yadij:
> > Amos: As of commit [a72f4b2](https://github.com/squid-cache/squid/commit/a72f4b28672ea21a8a0d939b5819c60720cfb0e6) looks good. This review approval does not cover other suggestions or topics being discussed.
> 
> I am interpreting the combination of the approval, the `S-waiting-for-committer` label, and the last sentence above as "please feel free to merge but let's continue ongoing discussions". If you do not want me not merge, please dismiss your PR approval. Otherwise, I will merge in a few days. Thank you.
> 

Aye, that is what was intended. Modulo the discussion causing any more changes being added in this looks good as-is.

src/cf.data.pre
yadij: I meant the same reasons apply to justify considering a ```to_sitelocal``` or ```to_localnet``` ACL separate to this ```to_linklocal```. **IF** we have any reason to add such a ```dst``` ACL.

The central consideration for all these is the security restriction preventing Squid routing WAN traffic into sensitive scopes. In this regard Squid should comply with and enforce (by default) the same behaviour/limits the RFCs require routers to use.

rousskov: > I meant the same reasons apply to justify considering a `to_sitelocal` or `to_localnet` ACL separate to this `to_linklocal`. **IF** we have any reason to add such a `dst` ACL.

> The central consideration for all these is the security restriction preventing Squid routing WAN traffic into sensitive scopes. In this regard Squid should comply with and enforce (by default) the same behaviour/limits the RFCs require routers to use.

AFAICT, you want to enlarge the default protection scope from PR's "localhost and equivalent" to something like "local net" (comprising all non-routable addresses?)[^1]. In principle, I am _not_ against that or any enlargement -- we just need to clearly define the new scope, provide a rationale for it, and match it with a new set of ACLs (without creating more problems that the new default would solve).

I do not know what RFCs say about default router configurations, but I believe we must balance purely technical security needs (prescribing a simple `http_access deny all` default) with human tendencies to remove obstacles instead of fine-tuning them (making a simple `http_access deny all` default a terrible idea). Improving that balance requires more work than we have done so far in this PR. Most likely, we need to do more than adding a couple of configuration lines and quickly writing a paragraph of text.

To make steady progress, I plan to merge this PR and welcome new/followup PR(s) that properly enlarge protection scope further, including PRs that make (relatively narrow rules added in) this PR unnecessary.

[^1]: Or something like that -- unfortunately, I do not know what addresses routers are supposed to protect by default according to RFCs.

PR #1163, opened 2022-10-11
Fix ./configure detection of various LDAP functions

rousskov:
Commit 0d57ed2 started using $LIBLDAP_LIBS as AC_CHECK_LIB() argument,
but that does not work because AC_CHECK_LIB() expects a library name (X)
rather than a linker option (-lX). The macro adds a "-l" prefix itself:

    checking for ldap_url_desc2str in -l-lldap -llber... no
    checking for ldap_url_parse in -l-lldap -llber... no
    checking for ldap_start_tls_s in -l-lldap -llber... no

### DIFF
--- a/acinclude/ldap.m4
+++ b/acinclude/ldap.m4
@@ -85,13 +85,13 @@ AC_DEFUN([SQUID_CHECK_LDAP_API],[
   AC_CHECK_LIB(ldap,[ldapssl_client_init],[
     AC_DEFINE(HAVE_LDAPSSL_CLIENT_INIT,1,[Define to 1 if you have ldapssl_client_init])
   ])
-  AC_CHECK_LIB($LIBLDAP_LIBS,[ldap_url_desc2str],[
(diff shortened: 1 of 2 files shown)


rousskov:
After 0d57ed2e8fd8a9721e7577e4deeb2a486972db9a, I was getting:

```
checking for ldap_url_desc2str in -l-lldap -llber... no
checking for ldap_url_parse in -l-lldap -llber... no
checking for ldap_start_tls_s in -l-lldap -llber... no
```

With these PR fixes, I am getting:

```
checking for library containing ldap_url_desc2str... -lldap 
checking for library containing ldap_url_parse... none required
checking for library containing ldap_start_tls_s... none required
```

The "none required" part is misleading. I added a source code comment to explain that when AC_SEARCH_LIBS() says
* "checking for library containing X" it really means
* "checking for _additional_ library containing X" or, in other words,
* "checking for a library containing X that should be added to the list of libraries created by previous checks".

yadij (approved):

PR #1166, opened 2022-10-18
Do not blame cache_peer for 4xx CONNECT responses

rousskov:
Authored-by: Eduard Bagdasaryan <eduard.bagdasaryan@measurement-factory.com>

To avoid using a problematic cache_peer, Squid penalizes cache_peers for
a variety of problems. However, a problem like an HTTP 403 Forbidden
response may be caused by the client or other factors unrelated to the
cache_peer that sent the response to Squid. In those cases, the
cache_peer is not at fault and should not be marked dead, even after
many such responses.

This change stops blaming cache_peers for HTTP 4xx outcomes of Squid
CONNECT requests. Currently, such outcomes only happen when a cache_peer
responds with a 4xx reply, but the new code also treats Squid-generated
4xx error responses (while trying to establish a connection to the
cache_peer) the same way.

This hard-coded logic covers known use cases. If different use cases
surface, we can make Squid behavior configurable.


src/acl/FilledChecklist.cc
```diff
@@ -247,9 +249,36 @@ ACLFilledChecklist::ACLFilledChecklist(const acl_access *A, HttpRequest *http_re
...
+void
+ACLFilledChecklist::setOutgoingConnection(const Comm::ConnectionPointer &outgoingConn)
+{
+    // TODO: Replace code that still does this manually with setOutgoingConnection().
+    if (!outgoingConn)
+        return;
+
+    setPeer(outgoingConn->getPeer());
+
+    // TODO: ACLIP::match(ip) does not check whether ip has been set. Should it?
```
rousskov, line 269: It feels wrong that ACLIP::match() was (and still is) treating default-constructed (i.e. setEmpty()) IP addresses as regular IP addresses. Most likely, those checks never actually match any IP an admin can configure via squid.conf, so I doubt there is a big problem with that, but it is still conceptually wrong to silently use an unknown/unset address IMO. I will remove this TODO without arguing if it turns out to be controversial.

src/neighbors.cc
```diff
@@ -468,6 +465,10 @@ peerAlive(CachePeer *p)
 
     p->stats.last_reply = squid_curtime;
     p->stats.probe_start = 0;
+
+    // TODO: Remove or explain how we could detect an alive peer without IP addresses
+    if (!p->n_addresses)
+        ipcache_nbgethostbyname(p->host, peerDNSConfigure, p);
```
rousskov, line 471: This `if` statement is old code from peerConnectSucceded(). AFAICT, it either never runs at all (and should be removed) or runs only in some very rare, special cases (which we should document), and this move does not change any of that. If others insist on preserving the location of this code, we would have to add a new dedicated global function or make peerDNSConfigure() public/global, just to call it from CachePeer::noteSuccess(). Otherwise, peerDNSConfigure() would bring too many dependencies into CachePeer.cc that this PR does not care about.

src/cf.data.pre
```diff
@@ -4016,6 +4016,47 @@ DOC_START
...
+	sequential problems is marked as "dead". Squid does not forward requests
+	to dead cache_peers.
+
+		cache_peer_fault allow|deny [!]aclname ...
+
+	Connection-cache_peer association starts when Squid attempts to open a TCP
+	connection to the cache_peer and ends when that connection establishment
+	attempt fails or the open connection is closed. Currently, a failure to
+	resolve a cache_peer domain name is not associated with that cache_peer,
+	but that should change.
```
rousskov, line 4038: AFAICT, when Squid fails to resolve a cache_peer domain name, the associated code simply does not go through the usual peerConnectFailed() steps. Hence, there is no warning about the peer becoming dead; just this:

```
2022/10/18 22:18:09.660| WARNING: DNS lookup for 'nxdomain.test' failed!
```

This PR does not change this existing logic.

* If we agree that the logic itself is flawed, then whoever fixes that code will be able to add the usual "Detected DEAD ..." handling to those DNS resolution failures. That author would then decide whether it is possible to apply this directive to that failure case as well, somehow triggering new background DNS lookups on the next access attempt (or some such) _instead_ of announcing the peer as dead (and waiting for the next DNS refresh to resolve its name).
* If we do not agree, then I will rephrase to avoid the implication that the current behavior is flawed.

src/acl/FilledChecklist.cc
yadij: FTR, setEmpty() initializes to IPv6 ANY_ADDR which is valid for display and matches as non-IPv4 (eg ```http_access deny !ipv4```) while being invalid for routing. It can be configured in any IP address type ACL with ```::``` (or equivalent longer strings).

rousskov: There is a ton of opportunities for a serious misunderstanding here! I will highlight a few relevant facts in hope to reduce the number of such opportunities.

Ip::Address is more than just an IN6/sin6_addr address. It also manages address family, address port, and other address info:

```C++
// Ip::Address manages the whole sockaddr_in6 rather than just sin6_addr:
struct sockaddr_in6 {
    sa_family_t     sin6_family;
    in_port_t       sin6_port;
    uint32_t        sin6_flowinfo;
    struct in6_addr sin6_addr;
    uint32_t        sin6_scope_id;
};
```

Ip::Address::setEmpty() zeros the _whole_ Ip::Address object. There is currently no single Ip::Address method that identifies such objects except for the compareWhole() method with another setEmpty() address as a parameter. If we create such a method and call it isEmpty() then, for some Ip::Address objects, isEmpty() will not return the same value as IsAny(). Those are two _different_ concepts/states. For convenience, I will use isEmpty() as a short version of "compareWhole() method with another setEmpty() address as a parameter".

Calling setEmpty() does make Ip::Address::isAny() true (because IN6_IS_ADDR_UNSPECIFIED() returns true for an all-zero IN6 address), but that is _not_ the right way to make that method true. The right way to make isAny() true is to call setAnyAddr().

Configuring an IP ACL with a `::` address creates an acl_ip_data object rather than (just) an Ip::Address object. The first Ip::Address part of that configured acl_ip_data object (i.e. acl_ip_data::addr1) is _not_ all-zeros. The address family of add1 is set to 10 when the configuration says `::`. Thus, that addr1 isAny() but not isEmpty(). There might be a way to configure an IP ACL to have an all-zeros acl_ip_data::addr1, but I do not know it.

The TODO in question is _not_ about addresses used to configure IP ACLs but about traffic IP addresses being fed to the already configured IP ACLs. The current ACLIP::match() code does not ignore all-zeros (i.e. unknown/unset/isEmpty()) traffic addresses, and those addresses will match some of the configured ACL addresses, including `::`. This out-of-scope TODO should decide whether such ACL matches are a bug.

yadij:
A new directive is overkill and incorrect fix for the stated issue.

The peer should always be handled along these lines:

a) probe unable to be performed (ignore marking),
b) probe failure (mark peer bad),
c) probe success (mark peer good),

d) HTTP in-band 4xx from peer (ignore marking, then retry or pass to client),
e) HTTP in-band 5xx from peer (mark peer bad, then retry or pass to client),
f) HTTP in-band 1xx/2xx/3xx from peer (mark peer good).

The issue stated in PR description indicates that case (d) is currently buggy. The correct solution is to make squid ignore marking the peer on any in-band 4xx status received.

src/neighbors.cc
yadij: IMO should be out of scope for this PR. The whole probing and neighbors.cc logic needs a refactor, this if-statement is no different from the many other things which should (re)move at that time.

src/cf.data.pre
yadij: Agreed that is yet another bug in the probing. Does seem to be outside the scope of the issue in PR description, so no need to fix here.

src/acl/FilledChecklist.cc
```diff
@@ -247,9 +249,36 @@ ACLFilledChecklist::ACLFilledChecklist(const acl_access *A, HttpRequest *http_re
...
+{
+    // TODO: Replace code that still does this manually with setPeer().
+    if (dst_peer_name.isEmpty() && peer)
+        dst_peer_name = peer->name;
+}
+
+void
+ACLFilledChecklist::setOutgoingConnection(const Comm::ConnectionPointer &outgoingConn)
+{
+    // TODO: Replace code that still does this manually with setOutgoingConnection().
```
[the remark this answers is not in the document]
yadij: I would be nice to have this TODO associated with the code that should be updated. But I do not insist on that.

yadij (changes requested):
Please reduce this to actually fixing the bug described (HTTP 4xx response from the peer should *not* be marked DEAD).

eduard-bagdasaryan:
> A new directive is overkill and incorrect fix for the stated issue

Simplified, following your suggestion (up to cfb1d02), please have a look.

src/neighbors.cc
eduard-bagdasaryan: Note that this `if` in the old code and new one has the same effect (p->n_addresses>0 on other peerAlive() callers). This 'out of scope' change is simply intended to minimize more of scope changes...

src/CachePeer.cc
```diff
@@ -47,6 +48,47 @@ CachePeer::~CachePeer()
...
+// misleading!) "TCP" label for all of them.
+/// noteFailure() helper for handling failures attributed to this peer
+void
+CachePeer::countFailure()
+{
+    stats.last_connect_failure = squid_curtime;
+    if (tcp_up > 0)
+        --tcp_up;
+
+    // TODO: Report peer name. Same-addresses peers often have different names.
```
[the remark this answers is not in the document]
rousskov: Done at #1174.

src/CachePeer.cc
```diff
@@ -47,6 +48,47 @@ CachePeer::~CachePeer()
...
+CachePeer::noteSuccess()
+{
+    if (!tcp_up) {
+        debugs(15, 2, "TCP connection to " << host << "/" << http_port << " succeeded");
+        tcp_up = connect_fail_limit; // NP: so peerAlive() works properly.
+        peerAlive(this);
+    } else {
+        tcp_up = connect_fail_limit;
+    }
+}
```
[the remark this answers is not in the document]
yadij: I am okay with that choice.

src/neighbors.cc
yadij: Sorry for not being clear. I was agreeing with the choice to copy the broken if-statement and add TODO for later fix.

src/security/BlindPeerConnector.cc
```diff
@@ -70,13 +70,13 @@ Security::BlindPeerConnector::noteNegotiationDone(ErrorState *error)
 
     if (error) {
         debugs(83, 5, "error=" << (void*)error);
-        // XXX: forward.cc calls peerConnectSucceeded() after an OK TCP connect but
-        // we call peerConnectFailed() if SSL failed afterwards. Is that OK?
-        // It is not clear whether we should call peerConnectSucceeded/Failed()
+        // XXX: FwdState calls NoteOutgoingConnectionSuccess() after an OK TCP connect, but
+        // we call noteFailure() if SSL failed afterwards. Is that OK?
```
yadij, line 74: To answer the question: this is not okay. noteSuccess() is our callers responsibility if/when their transaction completes (i.e. the client gets what they want via this peer).

I am not sure whether to request just fixing this XXX message, or the bug behind it. You have fixed similar bugs in this PR and it should be easy to just drop noteSuccess() from this class.

yadij (changes requested):
Thank you for attending to my request. This looks good, just some polishing touches now.

src/CachePeer.cc
```diff
@@ -47,6 +48,56 @@ CachePeer::~CachePeer()
...
+{
+    stats.last_connect_failure = squid_curtime;
+    if (tcp_up > 0)
+        --tcp_up;
+
+    // TODO: Report peer name. Same-addresses peers often have different names.
+
+    const auto consideredAliveByAdmin = stats.logged_state == PEER_ALIVE;
+    const auto level = consideredAliveByAdmin ? DBG_IMPORTANT : 2;
+    debugs(15, level, "ERROR: TCP connection to " << host << "/" << http_port << " failed");
```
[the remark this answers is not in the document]
rousskov: Let's not change this TCP-based wording in _this_ PR. We have another PR that polishes cache_peer reporting (already in the official review) that will modify this line and another in-progress project that will actually properly detail these failures. This PR is not a good opportunity to fix the misleading TCP label.

Said that, if you insist, I will commit your suggestion to avoid blocking this PR.

src/CachePeer.h
```diff
@@ -32,6 +33,13 @@ class CachePeer
     CachePeer() = default;
     ~CachePeer();
 
+    /// reacts to a successful establishment of a TCP connection to this cache_peer
```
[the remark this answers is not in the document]
rousskov: I do not think the suggested wording reflects how the official and PR code actually work. The exact condition is difficult to define precisely, but these successes and failures are not about cache_peer _use_. They are about _establishment_ of a connection (or _preparation_ for use). As we polish how cache_peer failures are reported/detailed (another in-progress project), we should drop the TCP part. For now, let's keep it as an imperfect marker of a code that needs further work that is out of scope here.

Said that, if you insist, I will drop TCP from the description in hope to avoid blocking this PR.

src/CachePeer.h
```diff
@@ -32,6 +33,13 @@ class CachePeer
...
+    void noteSuccess();
+
+    /// reacts to a failure on a TCP connection to this cache_peer
+    /// \param code a received response status code, if any
+    void noteFailure(Http::StatusCode code);
```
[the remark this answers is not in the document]
rousskov: We should not remove the parameter name because we are _documenting_ it by name, right?

src/security/BlindPeerConnector.cc
rousskov: This is no easy fix here, some fixes will have non-trivial effects on deployed Squids, and there is currently no consensus whether your answer is correct. The current official/PR code does not work the way you describe. We have an in-progress PR that will address this XXX. Let's postpone this discussion until then. This old XXX and all those related problems are out of this PR scope.

rousskov (commented):
I have applied one of your suggestions but would prefer to keep the rest of the proposed changes outside this PR scope. Please re-review.

src/CachePeer.h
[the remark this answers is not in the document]
yadij: As I see it success should be marked if and only when either a) a probe completes whatever transaction is tried, or b) an HTTP transaction completed without 4xx/5xx status. Despite the great variance of technical fine-details hidden behind those statements - the term "use" is vague enough to cover them all without ambiguity. Less wide terms I can think of all seem to conflict or imply wrong things about some of the fine-detail. Better suggestions welcome.

I do insist on at least removing the tight association with TCP connection setup. That is a common case, but not the only one.

src/CachePeer.h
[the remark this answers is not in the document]
yadij: Er yes. You are right.

src/security/BlindPeerConnector.cc
yadij: Ack, FTR I described the ideal (many bugs fail to meet up to it).

Since you have a followup already underway I will not insist.

src/CachePeer.cc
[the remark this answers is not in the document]
yadij: Huh? the line is being touched by this PR, there is no better time to fix a bug like this.

yadij (changes requested):
Just waiting on the existing discussions to resolve.

src/CachePeer.h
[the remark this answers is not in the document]
rousskov: > As I see it success should be marked if and only when either a) a probe completes whatever transaction is tried, or b) an HTTP transaction completed without 4xx/5xx status.

That may be the right _future_ principle, but it is not how the official code works (and is not how the official code was _designed_ to work -- so this is not just about bugs). In other words, the official code is using a different principle (which is a lot closer to TCP than to HTTP). This PR does not change any of that and should not change any of that.

> I do insist on at least removing the tight association with TCP connection setup.

Removed in 7aa16d6.

src/CachePeer.cc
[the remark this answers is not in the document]
rousskov: > Huh? the line is being touched by this PR, there is no better time to fix a bug like this.

You are thinking that removing TCP fixes a bug. I believe it just obscures the real problem, and the change itself is out-of-scope.

Removed in 7aa16d6 in hope to unblock this PR.

rousskov (commented):
I believe all the remaining concerns are addressed. Please re-review.

yadij (approved):

PR #1167, opened 2022-10-19
Fix milliseconds in certain cache.log messages

eduard-bagdasaryan:
This change fixes two bugs:

1. The milliseconds part of an early message time was wrong because
   debugLogTime() got it from the current_time global instead of the
   saved message timestamp. This problem probably affected only early
   saved messages in rare situations where current_time was updated
   between message creation and message writing. This fix adjusts
   DebugMessageHeader::timestamp to store message time with microsecond
   precision.

2. Level-0/1 messages that followed same-second level-2+ messages were
   logged with an unwanted (and stale) milliseconds value. This fix
   forces a reset of the message line format buffer in relevant cases.

### DIFF
--- a/src/debug/debug.cc
+++ b/src/debug/debug.cc
@@ -56,7 +56,7 @@ static constexpr int EarlyMessagesLevel = DBG_IMPORTANT;
 /// pre-formatted name of the current process for debugs() messages (or empty)
 static std::string ProcessLabel;
 
-static const char *debugLogTime(time_t t);
+static const char *debugLogTime(const timeval &);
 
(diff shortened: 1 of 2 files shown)


rousskov (approved):
Thank you for fixing these bugs! I have adjusted PR title/description to remove the focus on the _rare_ bug; the other bug only needs a one-line fix, but I observe it very often. Please adjust further as needed.

eduard-bagdasaryan:
The adjusted description looks correct.

yadij (approved):

PR #1168, opened 2022-10-19
Mimic GET reforwarding decisions when our CONNECT fails

rousskov:
Authored-by: Eduard Bagdasaryan <eduard.bagdasaryan@measurement-factory.com>

Use FwdState::reforwardableStatus() logic when deciding whether to
reforward our CONNECT request after a failed tunneling attempt.

Also honor forward_max_tries limit when retrying tunneling attempts.

These TunnelStateData fixes deal with ordinary CONNECT traffic (no
SslBump). FwdState also handles CONNECT requests (with SslBump). We make
that CONNECT handling more consistent across these classes (in addition
to making it more consistent across CONNECT and GET/etc. methods).

Co-authored-by: Alex Rousskov <rousskov@measurement-factory.com>

### DIFF
--- a/src/http/StatusCode.cc
+++ b/src/http/StatusCode.cc
@@ -9,6 +9,7 @@
 #include "squid.h"
 #include "debug/Stream.h"
 #include "http/StatusCode.h"
+#include "SquidConfig.h"
 
 const char *
 Http::StatusCodeString(const Http::StatusCode status)
@@ -276,3 +277,28 @@ Http::StatusCodeString(const Http::StatusCode status)
     return "Unassigned";
 }
 
+bool
+Http::IsReforwardableStatus(const Http::StatusCode s)
+{
+    switch (s) {
+
+    case Http::scBadGateway:
+
+    case Http::scGatewayTimeout:
+        return true;
+
+    case Http::scForbidden:
+
+    case Http::scInternalServerError:
+
+    case Http::scNotImplemented:
+
+    case Http::scServiceUnavailable:
+        return Config.retry.onerror;
+
+    default:
+        return false;
+    }
+
+    /* NOTREACHED */
+}
--- a/src/http/StatusCode.cc
+++ b/src/http/StatusCode.cc
@@ -278,22 +278,18 @@ Http::StatusCodeString(const Http::StatusCode status)
 }
 
 bool
-Http::IsReforwardableStatus(const Http::StatusCode s)
+Http::IsReforwardableStatus(const StatusCode s)
 {
     switch (s) {
 
-    case Http::scBadGateway:
-
-    case Http::scGatewayTimeout:
+    case scBadGateway:
+    case scGatewayTimeout:
         return true;
 
-    case Http::scForbidden:
-
-    case Http::scInternalServerError:
-
-    case Http::scNotImplemented:
-
-    case Http::scServiceUnavailable:
+    case scForbidden:
+    case scInternalServerError:
+    case scNotImplemented:
+    case scServiceUnavailable:
         return Config.retry.onerror;
 
     default:
@@ -302,3 +298,4 @@ Http::IsReforwardableStatus(const Http::StatusCode s)
 
     /* NOTREACHED */
 }
+
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -1407,6 +1407,8 @@ TunnelStateData::usePinned()
         const auto serverConn = ConnStateData::BorrowPinnedConnection(request.getRaw(), al);
         debugs(26, 7, "pinned peer connection: " << serverConn);
 
+        ++n_tries;
+
         // Set HttpRequest pinned related flags for consistency even if
         // they are not really used by tunnel.cc code.
         request->flags.pinned = true;
@@ -1423,7 +1425,7 @@ TunnelStateData::usePinned()
         sendError(error, "pinned path failure");
         return;
     }
-    ++n_tries;
+
 }
 
 CBDATA_CLASS_INIT(TunnelStateData);
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -1044,7 +1044,7 @@ TunnelStateData::noteConnection(HappyConnOpener::Answer &answer)
 {
     transportWait.finish();
 
-    Must(n_tries <= answer.n_tries); // n_tries cannot decrease
+    Assure(n_tries <= answer.n_tries); // n_tries cannot decrease
     n_tries = answer.n_tries;
 
     ErrorState *error = nullptr;
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -190,9 +190,9 @@ public:
     ResolvedPeersPointer destinations; ///< paths for forwarding the request
     int n_tries; ///< the number of forwarding attempts so far
     bool destinationsFound; ///< At least one candidate path found
-    /// whether another destination may be still attempted if the TCP connection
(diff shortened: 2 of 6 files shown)


src/tunnel.cc
```diff
@@ -188,13 +189,15 @@ class TunnelStateData: public PeerSelectionInitiator
     time_t startTime; ///< object creation time, before any peer selection/connection attempts
     ResolvedPeersPointer destinations; ///< paths for forwarding the request
     bool destinationsFound; ///< At least one candidate path found
-    /// whether another destination may be still attempted if the TCP connection
-    /// was unexpectedly closed
-    bool retriable;
```
rousskov, line 193: I renamed this to banRetries because we have to modify most of its code. This change improves symmetry with FwdState (where a very similar field is also "negative"), provides better debugging, and avoids creating a somewhat unusual true/set-at-construction flag. Said that, it is slightly out of scope and can be undone if others insist on keeping this field  the same in this PR.

src/tunnel.cc
```diff
@@ -391,12 +398,20 @@ TunnelStateData::checkRetry()
...
     if (!FwdState::EnoughTimeToReForward(startTime))
         return "forwarding timeout";
-    if (!retriable)
-        return "not retriable";
+    if (banRetries)
+        return banRetries;
     if (noConnections())
         return "no connections";
+
+    // TODO: Use Optional for peer_reply_status to avoid treating zero value specially.
```
rousskov, line 410: Completing this TODO would require changes going way beyond this tunneling code, but we can add them to this PR if others insist.

src/tunnel.cc
```diff
@@ -391,12 +398,20 @@ TunnelStateData::checkRetry()
...
+    if (request->hier.peer_reply_status != Http::scNone && !Http::IsReforwardableStatus(request->hier.peer_reply_status))
+        return "received HTTP status code is not reforwardable";
+
+    // TODO: check pinned connections; see FwdState::pinnedCanRetry()
```
rousskov, line 414: Completing this TODO may change Squid behavior in splice-at-step2 deployments (at least). I think it is best done in a dedicated PR, especially since this PR is not really about SslBump.

src/tunnel.cc
```diff
@@ -951,6 +966,9 @@ TunnelStateData::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
 
     al->cache.code.update(LOG_TCP_TUNNEL);
 
+    // XXX: al->hier.peer_reply_status (i.e. *status_ptr) should not be (re)set
+    // until we actually start responding to the client. Right here/now, we only
+    // know how this cache_peer has responded to us.
     if (answer.peerResponseStatus != Http::scNone)
         *status_ptr = answer.peerResponseStatus;
```
rousskov, line 973: Addressing this TODO requires quite a few of out-of-scope changes with difficult-to-predict outcomes, including access logging changes. It deserves a dedicated PR.

src/tunnel.cc
```diff
@@ -1062,6 +1084,8 @@ TunnelStateData::connectDone(const Comm::ConnectionPointer &conn, const char *or
         ResetMarkingsToServer(request.getRaw(), *conn);
     // else Comm::ConnOpener already applied proper/current markings
 
+    // TODO: add pconn race state tracking
```
rousskov, line 1087: Another large out-of-scope change. FwdState has quite a bit of complex code dedicated to properly handle persistent connection races!

src/tunnel.cc
```diff
@@ -1357,8 +1388,13 @@ TunnelStateData::startConnecting()
 
     assert(!destinations->empty());
     assert(!transporting());
+
+    delete savedError; // may still be nil
+    savedError = nullptr;
+    request->hier.peer_reply_status = Http::scNone; // TODO: Move to startPeerClock()?
```
rousskov, line 1394: We can address this TODO here, but it will affect FwdState code path (at least) as well. Perhaps it is best left for a dedicated PR.

rousskov (commented):
The mess in the PR-related code is complicated by our desire to keep/improve the symmetry between FwdState and TunnelStateData classes (that should eventually be merged/refactored to remove code duplication). FwdState has two distinct recovery modes: 

* retrying (i.e. checkRetry(); before the StoreEntry is "completed") and
* reforwarding (i.e. reforward(); when we have a peer response or equivalent).

TunnelStateData cannot easily mimic that approach because it does not use Store and StoreEntry. Current TunnelStateData code is structured to kind of combine both code paths around TunnelStateData::checkRetry(), and that is the method we fix.

src/tunnel.cc
```diff
@@ -951,6 +966,9 @@ TunnelStateData::tunnelEstablishmentDone(Http::TunnelerAnswer &answer)
 
     al->cache.code.update(LOG_TCP_TUNNEL);
 
+    // XXX: al->hier.peer_reply_status (i.e. *status_ptr) should not be (re)set
```
eduard-bagdasaryan, line 969: I guess you meant  al->http.code:
```suggestion
    // XXX: al->http.code (i.e. *status_ptr) should not be (re)set
```

src/http/StatusCode.cc
```diff
@@ -276,3 +277,25 @@ Http::StatusCodeString(const Http::StatusCode status)
     return "Unassigned";
 }
 
+bool
+Http::IsReforwardableStatus(const StatusCode s)
```
yadij, line 281: For symmetry with the equivalent method semantic checks this function should be a member of the StatusLine class.

This implies fixing the bug where ```answer.peerResponseStatus``` only passes the status code value, dropping the protocol, version and phrase parts of the server status line on custom/unknown codes.

src/tunnel.cc
yadij: It looks incorrect to ```return nullptr``` in this case.

IMO this should forbid retry until that later PR implements the bump edge case(s) correctly.

yadij (approved):
Just two minor design nits. I do not insist on either.

src/tunnel.cc
rousskov: > It looks incorrect to `return nullptr` in this case.

I agree.

> IMO this should forbid retry until that later PR implements the bump edge case(s) correctly.

IMO this _change_ in the official Squid behavior (and a bug fix) deserve a dedicated PR for several reasons, including this one: IIRC, the last time we adjusted similar FwdState code, there was no consensus regarding (of even sufficient knowledge about!) how certain esoteric pinned cases should be handled. We should proceed with caution and isolate those changes in a dedicated PR.

Thank you for not insisting on doing these changes in this PR.

src/http/StatusCode.cc
rousskov: > For symmetry with the equivalent method semantic checks this function should be a member of the StatusLine class.

I disagree on both counts:

* There is not enough symmetry here IMO: HttpRequestMethod represents HTTP method, but Http::StatusLine does not represent HTTP status code.

* Http::StatusLine represents HTTP status line (that has a lot more information than just the status code) while the function in question is specific to the status code. If we start making these reforwarding decisions based on other status line parts, _then_ we would move this function into that class. For now, the function is best placed into the more specific/narrow StatusCode scope.

src/tunnel.cc
rousskov: > We should proceed with caution and isolate those changes in a dedicated PR.

FWIW, I have created the corresponding internal project and hope to see this aspect improved (and this TODO addressed) in the foreseeable future.

PR #1169, opened 2022-10-20
Maintenance: Drop an unused, misleading condition

rousskov:
According to Autoconf NEWS, AC_USE_SYSTEM_EXTENSIONS is defined with
AC_DEFUN_ONCE since Autoconf v2.63b: "This means a subtle change in
semantics; previously, an AC_DEFUN macro could expand [this macro]
multiple times or surround the macro inside shell conditional text to
bypass [its] effects, but now the macro will expand exactly once, and
prior to the start of any enclosing AC_DEFUN macro".

In the generated ./configure script, the if statement no longer had any
code inside it, and AC_USE_SYSTEM_EXTENSIONS was unconditionally
expanded above the if statement.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,9 +55,7 @@ AM_PROG_CC_C_O
 AC_PROG_CXX
 
 # AC_USE_SYSTEM_EXTENSIONS should be called before any macros that run the C compiler.
-AS_IF([test "x$squid_host_os" = "solaris" -a "x$GCC" = "xyes"],[
-  AC_USE_SYSTEM_EXTENSIONS
-])
+AC_USE_SYSTEM_EXTENSIONS
 
 AC_LANG([C++])
 


rousskov:
```diff
--- configure-before 2022-10-19 17:56:51.669071764 -0400
+++ configure-after  2022-10-20 11:48:58.056396807 -0400
@@ -6924,11 +6924,6 @@
   $as_echo "#define _TANDEM_SOURCE 1" >>confdefs.h
 
 
-if test "x$squid_host_os" = "solaris" -a "x$GCC" = "xyes"; then :
-
-
-
-fi
 
 ac_ext=cpp
 ac_cpp='$CXXCPP $CPPFLAGS'
```

yadij:
The if-statement being removed reflects the fact that the macro is necessary only on Solaris GCC builds.
All other systems will automatically define it elsewhere **or not at all** as you identified.

yadij:
Oops, miss-click on the close.

rousskov:
> The if-statement being removed reflects the fact that the macro is necessary only on Solaris GCC builds.

The `if` statement poorly reflects that alleged fact. It misleads the reader into thinking that the macro is only _used_ on Solaris GCC builds. And since this `if` statement condition was ignored for many years, we can no longer be sure that the macro is not needed on some other builds -- we are building official Squid with AC_USE_SYSTEM_EXTENSIONS unconditionally expanded/used on all platforms.

If you insist, I will add a source code comment to reflect the intent of AC_USE_SYSTEM_EXTENSIONS inclusion. IMO, the intent will be clear enough in git history (without a source code comment), and the comment itself will not help ./configure.ac writers.

> All other systems will automatically define it elsewhere or not at all

All other systems will expand AC_USE_SYSTEM_EXTENSIONS just like Solaris GCC builds. The ./configure diff proves that the `if` _condition_ is completely ignored with no other changes to ./configure logic. The ./configure script is the same on all systems so the _condition_ (and only the condition) is ignored on all systems. I have also checked that the macro itself is expanded and not just stripped/ignored (more on that below).

> Was your before/after check from a configure created in a recent Solaris with GCC environment ?

No, but bootstrapping environment is irrelevant here: Bootstrapping artifacts, including ./configure are supposed to be platform-agnostic, right? The `if` statement condition is not checked/evaluated during bootstrapping. It is checked/evaluated during ./configure _execution_. At that time, the `if` statement _body_ is already empty, as shown by the earlier diff (and as predicted by the cited Autoconf documentation): Autoconf expands our AC_USE_SYSTEM_EXTENSIONS outside its `if` statement.

If I completely remove the AC_USE_SYSTEM_EXTENSIONS macro, then the generated ./configure becomes very different. The macro is not ignored on my build platform -- it is expanded and used. I bet the same happens on all platforms.

----

The above section discussed the effects on generated ./configure. This section discusses the effects on generated include/autoconf.h.in and include/autoconf.h which might be easier to comprehend because they do not involve autoconf magic. These examples illustrate the same point: It does not matter whether AC_USE_SYSTEM_EXTENSIONS is included inside that `if` statement designed to detect Solaris GCC builds; official and PR code produce identical include/autoconf.h.in files. The only thing that matters is whether the macro is mentioned in configure.ac.

For example, on my platform, _both_ the official and PR code result in the following lines added to the include/autoconf.h.in template (i.e. _before_ ./configure runs) as the result of AC_USE_SYSTEM_EXTENSIONS being mentioned in configure.ac:

```diff
--- include/autoconf.h.in-without-macro       2022-10-21 09:08:32.783150196 -0400
+++ include/autoconf.h.in-with-macro          2022-10-19 17:56:46.220919609 -0400
@@ -1568,6 +1568,28 @@
 /* systemd support is available */
 #undef USE_SYSTEMD
 
+/* Enable extensions on AIX 3, Interix.  */
+#ifndef _ALL_SOURCE
+# undef _ALL_SOURCE
+#endif
+/* Enable GNU extensions on systems that have them.  */
+#ifndef _GNU_SOURCE
+# undef _GNU_SOURCE
+#endif
+/* Enable threading extensions on Solaris.  */
+#ifndef _POSIX_PTHREAD_SEMANTICS
+# undef _POSIX_PTHREAD_SEMANTICS
+#endif
+/* Enable extensions on HP NonStop.  */
+#ifndef _TANDEM_SOURCE
+# undef _TANDEM_SOURCE
+#endif
+/* Enable general extensions on Solaris.  */
+#ifndef __EXTENSIONS__
+# undef __EXTENSIONS__
+#endif
+
+
 /* Samba TrivialDB support is available */
 #undef USE_TRIVIALDB
 
@@ -1607,6 +1629,16 @@
 /* Enable support for the X-Accelerator-Vary HTTP header */
 #undef X_ACCELERATOR_VARY
 
+/* Define to 1 if on MINIX. */
+#undef _MINIX
+
+/* Define to 2 if the system does not provide POSIX.1 features except with
+   this defined. */
+#undef _POSIX_1_SOURCE
+
+/* Define to 1 if you need to in order for `stat' and other things to work. */
+#undef _POSIX_SOURCE
+
 /* Nameserver Counter for IPv6 _res */
 #undef _SQUID_RES_NSADDR6_COUNT
```

The above platform-agnostic template is then populated with platform-specific settings by running ./configure. After my ./configure runs, I get:

```diff
--- include/autoconf.h-without-macro  2022-10-21 09:35:43.668667406 -0400
+++ include/autoconf.h-with-macro     2022-10-20 11:49:23.873117341 -0400
@@ -1447,7 +1447,7 @@
 /* #undef STACK_DIRECTION */
 
 /* Define to 1 if you have the ANSI C header files. */
-/* #undef STDC_HEADERS */
+#define STDC_HEADERS 1
 
 /* common adaptation support */
 #define USE_ADAPTATION 1
@@ -1569,6 +1569,28 @@
 /* systemd support is available */
 /* #undef USE_SYSTEMD */
 
+/* Enable extensions on AIX 3, Interix.  */
+#ifndef _ALL_SOURCE
+# define _ALL_SOURCE 1
+#endif
+/* Enable GNU extensions on systems that have them.  */
+#ifndef _GNU_SOURCE
+# define _GNU_SOURCE 1
+#endif
+/* Enable threading extensions on Solaris.  */
+#ifndef _POSIX_PTHREAD_SEMANTICS
+# define _POSIX_PTHREAD_SEMANTICS 1
+#endif
+/* Enable extensions on HP NonStop.  */
+#ifndef _TANDEM_SOURCE
+# define _TANDEM_SOURCE 1
+#endif
+/* Enable general extensions on Solaris.  */
+#ifndef __EXTENSIONS__
+# define __EXTENSIONS__ 1
+#endif
+
+
 /* Samba TrivialDB support is available */
 /* #undef USE_TRIVIALDB */
 
@@ -1608,6 +1630,16 @@
 /* Enable support for the X-Accelerator-Vary HTTP header */
 #define X_ACCELERATOR_VARY 0
 
+/* Define to 1 if on MINIX. */
+/* #undef _MINIX */
+
+/* Define to 2 if the system does not provide POSIX.1 features except with
+   this defined. */
+/* #undef _POSIX_1_SOURCE */
+
+/* Define to 1 if you need to in order for `stat' and other things to work. */
+/* #undef _POSIX_SOURCE */
+
 /* Nameserver Counter for IPv6 _res */
 /* #undef _SQUID_RES_NSADDR6_COUNT */
```

yadij (approved):

PR #1171, opened 2022-10-24
CI: Fix support for running a subset of GitHub Actions tests

rousskov:
The scripts were always running all/default tests instead of those named
on the command line (if any). Broken since inception in commit 2ed8cc4.

### DIFF
--- a/test-suite/test-functionality.sh
+++ b/test-suite/test-functionality.sh
@@ -260,5 +260,5 @@ main() {
     run_tests $tests
 }
 
-main
+main "$@"
 exit $?
--- a/test-suite/test-sources.sh
+++ b/test-suite/test-sources.sh
@@ -229,7 +229,8 @@ main() {
     fi
     echo "Starting point: $STARTING_POINT (`git rev-parse $STARTING_POINT`)"
 
-    checks="$@"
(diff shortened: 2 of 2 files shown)


test-suite/test-sources.sh
```diff
@@ -229,7 +229,8 @@ main() {
     fi
     echo "Starting point: $STARTING_POINT (`git rev-parse $STARTING_POINT`)"
 
-    checks="$@"
+    local checks="$@"
```
rousskov, line 232: This out of scope polishing change reduces unwanted differences between the two test-suite scripts. There is more about this code duplication in the PR that introduced it at https://github.com/squid-cache/squid/pull/1160/files#r986398359

test-suite/test-functionality.sh
```diff
@@ -260,5 +260,5 @@ main() {
     run_tests $tests
 }
 
-main
+main "$@"
```
rousskov, line 263: FWIW, I forgot to add these parameters when I moved the code above into main(), a function created while polishing the original PR [1160](https://github.com/squid-cache/squid/pull/1160/files#r986398359) code.

yadij (approved):

PR #1172, opened 2022-10-24
CI: Assume recent CI fixes make busy-restart test stable

rousskov:


rousskov:
When (re)starting Squid, newer Daft code waits for all SMP kids to register with their Coordinator process, narrowing the period when the Squid instance is already listening on the configured ports but is not yet ready to cache responses on disk. That improvement is not a full fix (the latter may require Squid modifications), but, in our tests, including a three recent tests in the official repository (listed below), the changes are sufficient to make the busy-restart test stable in GitHub Actions environment. It has already been stable in Semaphore CI environment.

* https://github.com/squid-cache/squid/actions/runs/3309522729/jobs/5462791220
* https://github.com/squid-cache/squid/actions/runs/3304483969/jobs/5462868999
* https://github.com/squid-cache/squid/actions/runs/3309590453/jobs/5462926220

rousskov:
FYI: This and possibly other all-green PRs are not being staged because Anubis is not receiving "GitHub CI" check statuses from GitHub (while receiving statuses of the other three checks). Most likely, this was happening since those GitHub Actions checks were added, but it was not a problem until those checks became _required_ (per [plan](http://lists.squid-cache.org/pipermail/squid-dev/2022-October/009826.html)) and, hence, tracked by Anubis. We are aware of this strange problem and @eduard-bagdasaryan is working on explaining/solving it.

rousskov:
_Status update_: Eduard has identified the reason behind the lack of check statuses -- even though all statuses look the same to humans reading PR pages, GitHub is using a new/special API to report GitHub Actions runs. Eduard is working on adding Anubis code to fetch those statuses.

PR #1173, opened 2022-10-28
Do not send self-signed certificates from tls-cert=bundle

rousskov:
... except the very first/signing certificate itself (if it is
self-signed), of course.

Sending such certificates is pointless: Recipients cannot trust a
received self-signed certificate unless they have already obtained it
from other (trusted) sources. Thus, this change is not expected to
affect any reasonable TLS client implementations (besides reducing noise
on the wire and associated overheads).

Co-authored-by: Christos Tsantilas <christos@chtsanti.net>

### DIFF
--- a/src/security/KeyData.cc
+++ b/src/security/KeyData.cc
@@ -50,11 +50,13 @@ Security::KeyData::loadCertificates()
         while (const auto bundledCert = Ssl::ReadOptionalCertificate(bio)) {
             assert(!wireCerts.empty()); // this->cert is there (at least)
 
-            // We cannot chain any certificate after a self-signed certificate. This
-            // check also protects the IssuedBy() check below from adding duplicated
-            // (i.e. listed multiple times in the bundle) self-signed certificates.
-            if (SelfSigned(*wireCerts.back())) {
-                debugs(83, DBG_PARSE_NOTE(2), "WARNING: Ignoring certificate after a self-signed one: " << *bundledCert);
+            // We ignore a self-signed certificate because it should not be
+            // sent: The recipients that do not already have it should not trust
+            // it. This check also protects the IssuedBy() check below from
+            // adding duplicated (i.e. listed multiple times in the bundle)
+            // self-signed signing certificates.
+            if (SelfSigned(*bundledCert)) {
+                debugs(83, DBG_PARSE_NOTE(2), "WARNING: Ignoring self-signed certificate: " << *bundledCert);
                 continue; // ... but keep going to report all ignored certificates
             }
 
--- a/src/ssl/support.cc
+++ b/src/ssl/support.cc
@@ -970,8 +970,15 @@ void
 Ssl::chainCertificatesToSSLContext(Security::ContextPointer &ctx, Security::ServerOptions &options)
 {
     assert(ctx);
+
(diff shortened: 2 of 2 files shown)


yadij:
FTR, this change breaks the following TLS features which Squid may encounter in important LAN use cases:

* TOFU - where the self-signed certificate can be authorized by the client (or manually by a user) on first sighting and added to the clients trusted root set. The root CA is (almost) always provided in-band as part of the CA chain.

* DANE - where the signature key(s) (only) for the root CA is obtained via DNS records. The content of root CA may be provided in-band as part of the CA chain.

* null cipher - where the root CA is dynamically generated and only trusted for the connection it was received over.

yadij:
I propose adding a ```tls-tofu``` setting to Security::ServerOptions to enable sending self-signed root CA on https_port server handshakes. The DANE and null cipher cases need extra logic that Squid currently lacks so are not reliable today even if they may be encountered on cache_peer connections.

src/security/KeyData.cc
```diff
@@ -50,11 +50,13 @@ Security::KeyData::loadCertificates()
         while (const auto bundledCert = Ssl::ReadOptionalCertificate(bio)) {
             assert(!wireCerts.empty()); // this->cert is there (at least)
 
-            // We cannot chain any certificate after a self-signed certificate. This
-            // check also protects the IssuedBy() check below from adding duplicated
-            // (i.e. listed multiple times in the bundle) self-signed certificates.
-            if (SelfSigned(*wireCerts.back())) {
-                debugs(83, DBG_PARSE_NOTE(2), "WARNING: Ignoring certificate after a self-signed one: " << *bundledCert);
+            // We ignore a self-signed certificate because it should not be
+            // sent: The recipients that do not already have it should not trust
```
yadij, line 54: This is a more significant behavioural change than it seems. When I attempted this a few years back there was immediate feedback about some environments failing to validate the TLS connections to Squid.

I am not against another attempt to fix this, but please do update the squid.conf documentation for all directives using tls-cert= options and also Squid-6 release notes for their behaviour change.

yadij (changes requested):

rousskov:
> I propose adding a tls-tofu setting ... to enable sending self-signed root CA on https_port server handshakes.

We could add a new http_port and https_port setting[^1] but, AFAICT, it would only be useful in some SslBump environments (where the configured certificate is not the traffic-signing certificate). In all other cases, the admin can already configure Squid to send self-signed certificates when needed (by adding those self-signed certificates to the bundle). There is no need for a new configuration option AFAICT. I am not against adding such an option but do not plan on volunteering the necessary code.

[^1]: FWIW, I would name that new setting to be behavior-specific (e.g., send-self-signed-certificate or send-complete-certificate-chain) rather than use-case-specific (e.g., tls-tofu). The setting can also enable Squid validation code that would reject configurations that do _not_ contain a self-signed certificate (that Squid is supposed to send).

> this change breaks the following TLS features which Squid may encounter in important LAN use cases: ...

I am closing this PR based purely on your assertion that some relevant use cases require sending a self-signed certificate (in addition to the signing certificate). Unfortunately, I do not know enough to immediately confirm the existence of those cases.

PR #1174, opened 2022-10-31
Improve cache_peer reporting and NetDB cache_peer search

rousskov:
Squid used several cache_peer configuration components and CachePeer
fields to identify cache peers in code and cache.log messages: (the
lowercase version of) cache_peer hostname, cache_peer (hostname,
http_port) tuple, and cache_peer name=value (which uses cache_peer
hostname by default).

Inconsistent identification led to admin confusion (e.g., a message
about cache_peer A was misinterpreted as being about cache_peer B) and
bugs (e.g., indexing storage by component X, but searching using
component Y). This change cements a single component as a unique
cache_peer identifier: CachePeer::name.

Also reject configurations with cache_peer naming errors in
cache_peer_access and neighbor_type_domain directives (instead of
reporting but otherwise ignoring the problematic configuration lines).

The following CachePeer reporting principles were used:

* To identify a specific cache_peer p (among all the configured ones) in
  a cache.log message, print that peer (i.e. use `os << *p` or
  equivalent). CachePeer printing code will print that peer name, which
  is guaranteed to be unique across all configured cache_peers.

* After identifying a cache_peer, do not dump its various details like
  ICP port numbers, especially when they are not relevant to the message
  (i.e. provide no useful context-related information going beyond that
  cache_peer identification).

The following bugs were fixed:

* NetDB code is storing cache_peer hostname[1] but was searching for a
  cache_peer with a matching name[2] and, hence, usually failing to find
  explicitly named cache_peers. Probably broken since 2003 commit
  be75332 that added name=value support and updated peerFindByName() to
  use CachePeer::name but left storage code[1] unchanged.
    
    [1] netdbPeerAdd(): p->peername = netdbPeerName(e->host);
    [2] netdbClosestParent(): p = peerFindByName(h->peername);
    
* netdbClosestParent() could not find the closest cache_peer in certain
  cases involving multiple cache_peers with the same hostname because
  the code incorrectly stopped looking for eligible same-hostname
  cache_peers after correctly discarding the first one.

  The above NetDB problems may imply that NetDB code needs to be
  refactored to store CachePeer pointers instead of CachePeer hostnames,
  but that serious change deserves more analysis (and a dedicated PR).

* cache_peer name=value configurations with empty/missing name value
  could dereference a null name pointer while misleading developers into
  thinking that a previously set name could be erased/forgotten. Squid
  now rejects such configurations.

* neighborIgnoreNonPeer() was allocating CachePeer::host using new() but
  ~CachePeer is freeing its host data member using xfree().
    
Also removed peerFindByNameAndPort() unused since inception at db1cd23.


src/CachePeer.cc
```diff
@@ -14,9 +14,17 @@
 #include "pconn.h"
 #include "PeerPoolMgr.h"
 #include "SquidConfig.h"
+#include "util.h"
 
 CBDATA_CLASS_INIT(CachePeer);
 
+CachePeer::CachePeer(const char * const hostname):
+    name(xstrdup(hostname)),
+    host(xstrdup(hostname))
```
rousskov, line 23: FWIW, I started by changing `name` type to be SBuf instead of raw c-string, but had to revert all those improvements because they triggered too many source code changes, most of which were out of scope and some could not even be completed in this PR. That conversion and more CachePeer polishing changes should be done, but we should not do it now/here.

src/neighbors.h
```diff
@@ -42,8 +42,10 @@ void neighbors_init(void);
 #if USE_HTCP
 void neighborsHtcpClear(StoreEntry *, HttpRequest *, const HttpRequestMethod &, htcp_clr_reason);
 #endif
-CachePeer *peerFindByName(const char *);
-CachePeer *peerFindByNameAndPort(const char *, unsigned short);
+
+/// cache_peer with a given name (or nil)
+CachePeer *findCachePeerByName(const char *);
```
rousskov, line 47: We do not have to rename peerFindByName() but I think it is a good idea because there were bugs in the official code where the old function was called with the wrong c-string parameter (a host name instead of cache_peer name). It is probably best to trigger build failures when this change is merged into branches where peerFindByName() is used incorrectly in contexts unknown/unchecked in this PR.

src/neighbors.cc
```diff
@@ -1172,22 +1153,22 @@ neighborUp(const CachePeer * p)
...
 
     if (p->stats.probe_start != 0 &&
             squid_curtime - p->stats.probe_start > Config.Timeout.deadPeer) {
-        debugs(15, 8, "neighborUp: DOWN (dead): " << p->host << " (" << p->in_addr << ")");
+        debugs(15, 8, "DOWN (dead): " << *p);
         return 0;
     }
 
-    debugs(15, 8, "neighborUp: UP: " << p->host << " (" << p->in_addr << ")");
+    debugs(15, 8, "UP: " << *p);
```
rousskov, line 1171: To avoid creating inconsistent output and/or excessive diff noise, I tried to leave excessive MYNAME prefixes in old debugs() statements unless we modified all such debugs() statements in a function OR the statements already mixed MYNAME and MYNAME-free wording. I am sure I got a couple of cases wrong and will undo or change them upon request. IMO, we should remove (most) such prefixes in a single PR covering all Squid sources.

src/neighbors.cc
```diff
@@ -1289,26 +1270,26 @@ peerConnectFailed(CachePeer * const p)
...
-            debugs(15, DBG_IMPORTANT, "Detected DEAD " << neighborTypeStr(p) << ": " << p->name);
+            debugs(15, DBG_IMPORTANT, "Detected DEAD " << neighborTypeStr(p) << ": " << *p);
             p->stats.logged_state = PEER_DEAD;
         } else {
             debugs(15, 2, "additional failures needed to mark this cache_peer DEAD: " << p->tcp_up);
         }
     } else {
         assert(!p->tcp_up);
-        debugs(15, 2, "cache_peer " << p->host << "/" << p->http_port << " is still DEAD");
+        debugs(15, 2, "cache_peer is still DEAD");
```
eduard-bagdasaryan, line 1284:
```suggestion
        debugs(15, 2, *p << " is still DEAD");
```

src/cf.data.pre
```diff
@@ -3959,13 +3959,22 @@ DOC_START
...
-			Required if you have multiple peers on the same host
-			but different ports.
-			This name can be used in cache_peer_access and similar
-			directives to identify the peer.
+			Required if you have multiple cache_peers with the same hostname.
+			Defaults to cache_peer hostname when not explicitly specified.
+
+			Other directives (e.g., cache_peer_access), cache manager reports,
+			and cache.log messages referring to a specific cache_peer use this
+			name to uniquely identify a cache_peer.
```
eduard-bagdasaryan, line 3967:
```suggestion
			name to uniquely identify the cache_peer.
```

src/error/forward.h
```diff
@@ -90,6 +90,7 @@ typedef enum {
 
 class Error;
 class ErrorDetail;
+class ErrorState;
```
eduard-bagdasaryan, line 93: Do we need this in this PR?

src/neighbors.cc
yadij: You do not have to change every debugs() in a function. But please do remove the MYNAME portion for debugs() that are touched by this PR already.

src/peer_userhash.cc
```diff
@@ -185,7 +185,7 @@ peerUserHashSelectParent(PeerSelector *ps)
         combined_hash += combined_hash * 0x62531965;
         combined_hash = ROTATE_LEFT(combined_hash, 21);
         score = combined_hash * tp->userhash.load_multiplier;
-        debugs(39, 3, "peerUserHashSelectParent: " << tp->name << " combined_hash " << combined_hash  <<
+        debugs(39, 3, *tp << " combined_hash " << combined_hash  <<
```
yadij, line 188:
```suggestion
        debugs(39, 3, *tp << " combined_hash " << combined_hash <<
```

src/tests/stub_CachePeer.cc
```diff
@@ -13,5 +13,7 @@
 
 #include "CachePeer.h"
 
+void CachePeer::rename(const char *) STUB
```
yadij, line 16: Please remove the empty line above this to match the stubs implementation style.

yadij (changes requested):

src/cf.data.pre
yadij: FTR, speaking only for myself I do not insist. But to me "a X" implies vagueness about which X - whereas the subject of the sentence is about unique identification of one specific X.

eduard-bagdasaryan: It's not a big deal for me either. However, what does the second 'cache_peer' stand for in this sentence then? Arbitrary (another) cache_peer or the same cache_peer that 'cache.log messages referring to'?

src/tests/stub_CachePeer.cc
rousskov: Done in 076f60f.

src/peer_userhash.cc
rousskov: Done in c970498.

src/neighbors.cc
rousskov: Done in fbfaaf4 to avoid arguing. Please let me know if I missed any applicable cases.

src/neighbors.cc
```diff
@@ -1113,8 +1112,8 @@ neighborsUdpAck(const cache_key * key, icp_common_t * header, const Ip::Address
             neighborIgnoreNonPeer(from, opcode);
         } else if (p->stats.pings_acked > 100) {
             if (100 * p->icp.counts[ICP_DENIED] / p->stats.pings_acked > 95) {
-                debugs(15, DBG_CRITICAL, "95%% of replies from '" << p->host << "' are UDP_DENIED");
-                debugs(15, DBG_CRITICAL, "Disabling '" << p->host << "', please check your configuration.");
+                debugs(15, DBG_CRITICAL, "95% of replies from " << *p << " are UDP_DENIED");
```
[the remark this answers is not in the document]
rousskov: > We could also improve logs by combining the two debugs() into one line.

I combined and polished these lines further in commit f9547bf.

> The if-statement can be improved a bit too. No need to waste CPU multiplying by 100, just check against 0.95.

There is such a need if we want to continue save CPU cycles by using fast integer math rather than switching to doubles. However, that is not the reason I left that (problematic for other reasons!) statement alone. I simply do not want to imply that this PR changes that (possibly buggy) condition logic.

src/cf.data.pre
rousskov: This is not my area of expertise, but, AFAICT, the "logic" here is that we do not know what that "specific cache_peer" is in each of the described contexts (that peer depends on each of the described context details that are unknown to us), so we should not imply that we have somehow identified it or that the reader knows which specific cache_peer we are talking about. We did not, and they do not.

However, I may have found a different/better(?) way to phrase this: Instead of trying to describe how the `name` parameter is used in general, we can talk about a specific name value usage. See commit 6901173.

src/Makefile.am
```diff
@@ -1691,6 +1692,7 @@ tests_testACLMaxUserIP_SOURCES = \
 	tests/testACLMaxUserIP.cc \
 	tests/testACLMaxUserIP.h
 nodist_tests_testACLMaxUserIP_SOURCES = \
+	tests/stub_CachePeer.cc \
```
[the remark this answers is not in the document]
rousskov: Done in 847ea49.

src/Makefile.am
```diff
@@ -2017,6 +2020,7 @@ tests_testHttp1Parser_LDFLAGS = $(LIBADD_DL)
 
 check_PROGRAMS += tests/testHttpReply
 tests_testHttpReply_SOURCES = \
+	tests/stub_CachePeer.cc \
```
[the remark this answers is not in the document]
rousskov: Done in 847ea49.

src/base/IoManip.h
```diff
@@ -38,13 +45,24 @@ RawPointer(const char *label, const Pointer &ptr)
     return RawPointerT<Pointer>(label, ptr);
 }
 
+/// convenience wrapper for creating  RawPointerT<> objects without a label
```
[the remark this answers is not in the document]
rousskov: Done in 28fc3aa. I fixed both the original comment and this new variation.

src/cache_cf.cc
```diff
@@ -2515,7 +2511,7 @@ parse_peer_access(void)
         return;
     }
 
-    CachePeer *p = peerFindByName(host);
+    const auto p = findCachePeerByName(host);
```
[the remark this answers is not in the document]
rousskov: Done in 52d451b.

FWIW, I do not see how upgrading fatal() to throw implies that this PR is the right place to reject configurations with incorrect cache_peer references, but I certainly agree that we should reject such configurations. Now we do.

src/cache_cf.cc
```diff
@@ -2543,7 +2539,7 @@ parse_hostdomaintype(void)
 
     char *domain = nullptr;
     while ((domain = ConfigParser::NextToken())) {
-        CachePeer *p = peerFindByName(host);
+        const auto p = findCachePeerByName(host);
```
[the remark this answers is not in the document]
rousskov: Done in 52d451b.

rousskov (commented):
@yadij, I hope that I addressed all change requests. Please re-review.

src/cache_cf.cc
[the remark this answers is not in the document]
rousskov: Please note that peername and peername_regex ACLs still ignore peer-name typos. I think they should be fixed as well, but perhaps that is best done in a dedicated PR? It will require quite a few out-of-scope changes, especially if we want to at least warn the admin if peername_regex cannot match _any_ of the configured cache_peers.

src/cf.data.pre
yadij: That reads better to me, and its shorter :-)

src/CachePeer.cc
```diff
@@ -55,3 +73,9 @@ CachePeer::connectTimeout() const
     return Config.Timeout.peer_connect;
 }
 
+std::ostream &
+operator <<(std::ostream &os, const CachePeer &p)
```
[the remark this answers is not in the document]
yadij: Its an optimization for logging, so I do not insist.

src/base/IoManip.h
```diff
@@ -25,7 +25,14 @@ class RawPointerT {
     /// Report the pointed-to-object on a dedicated Debug::Extra line.
     RawPointerT<Pointer> &asExtra() { onExtraLine = true; return *this; }
 
+    /// enable and, optionally, customize reporting of nil pointers
+    RawPointerT<Pointer> &orNil(const char *nilTextToUse = "[nil]") { nilText = nilTextToUse; return *this; }
+
     const char *label; /// the name or description of the being-debugged object
+
+    /// whether and how to report a nil pointer; use orNil() to enable
+    const char *nilText = nullptr;
```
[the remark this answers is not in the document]
yadij: Okay. It does call into question why we have a special setter though.

src/cache_cf.cc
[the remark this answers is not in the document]
yadij: Up to you, I took the debugs() as being in-scope due to the changes with detecting and reporting other details on cache-peer and cache_peer_access specifically. If you want to call the ACL value validation out of scope, thats fine.

yadij (approved):

src/base/IoManip.h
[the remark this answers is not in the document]
rousskov: > It does call into question why we have a special setter though.

The `*this`-returning setters are required for the streaming syntax to work: `... << RawPointer(p).foo().bar() << ...`.

src/Makefile.am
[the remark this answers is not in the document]
rousskov: > These lists should really be sorted by their full string value.

@yadij, please note that your definition of correct `_SOURCES` entry order differs from the one currently used by `scripts/source-maintenance.sh` (or by `scripts/format-makefile-am.pl`, to be precise). The current automated order is based on the `$filename.$prefix` expression, where `$prefix` is `tests/stub_` or `tests/test`.

If you think the automated order is wrong, then please post a PR to change `scripts/format-makefile-am.pl`.

FWIW, I would support simplifying the currently overly complex `_SOURCES` sorting rules, migrating to the "full string value" algorithm you have mentioned. Such migration would allow sorting these fields in text editors. It would also make the algorithm much simpler and robust. The resulting loss of spacial proximity of related stubbed/test/full files is an acceptable sacrifice in this case IMHO.

PR #1176, opened 2022-11-01
Avoid level-3 "failed to find or read error text file" WARNINGs

eduard-bagdasaryan:
... when (re)configuring ERR_REQUEST_PARSE_TIMEOUT and ERR_RELAY_REMOTE
error templates. These errors cannot be customized externally and should
have been listed as such in commits 9ce4a1e and f5e1794, respectively.

Also polished hard-coded error storage code.


src/errorpage.cc
```diff
@@ -203,9 +213,18 @@ static int error_page_count = 0;
...
 static IOCB errorSendComplete;
 
+template <class Messages>
+static const char *
+FindHardCodedTemplate(const err_type type, const Messages &messages)
+{
+    const auto foundDefault = std::find_if(messages.cbegin(), messages.cend(), [&](const HardCodedMessage &m) {
+        return m.type == type;
+    });
+    return foundDefault == messages.cend() ? nullptr : foundDefault->text;
```
yadij, line 225: This is better implemented as a ranged-for loop. For both readability and performance.
```suggestion
    for (const auto &m : messages) {
        if (m.type == type)
            return m.text;
    }
    return nullptr;
```

yadij (changes requested):
Thank you for fixing this regression.  I just request simplification of the core search loop. I have checked and ranged-for has 15% shorter and faster code when built.

yadij (approved):

src/errorpage.cc
```diff
@@ -180,20 +180,30 @@ error_hard_text[] = {
...
+    }
+}};
+
+/// error messages that may be configured/customized by the admin,
+/// like the regular ones, but that also have a hard-coded last-resort default
+static std::array<HardCodedMessage, 1> ErrorsWithDefaultTemplates = {{
     {
         MGR_INDEX,
         "mgr_index"
     }
```
rousskov, line 199: > commit b073fc4: Reserve the template name MGR_INDEX for use by cachemgr scripts. But do not distribute any preset template. This allows manager apps to provide their own static template with linked scripts and objects.

@yadij, I am curious, why are we not adding `MGR_INDEX` to the errors/templates/? Why make this a special case? I read the related commit message bullet quoted above, but it did not answer these questions for me. Clearly, "manager apps" (whatever those are) can still "provide their own static template" even if we distribute one.

Please do not misinterpret this question as an objection to this bug fix! I am just trying to _understand_ why we have to treat `MGR_INDEX` specially in this context (especially since that need for special treatment created the bug we are fixing in this PR).

yadij: MGR_INDEX is intended to be provided by external tools. It's (non-)existence (externally visible 404 status on /index URL) is the factor cachemanager CGI tool uses to decide about displaying its own report selection logic or redirecting/showing that MGR_INDEX page content.

rousskov: > MGR_INDEX ... (non-)existence (externally visible 404 status on /index URL) is ...

Wait, do you mean that _Squid_ should return a 404 when the `MGR_INDEX` file template is missing?! If I understand this PR description correctly, Squid did not do that after buggy commit 7e6eabbc6976f939079e2abbbc52c2a98d65bd89 and will still not do it if this PR is merged "as is". Or are you talking about some other, out-of-band test for file existence that results in a 404 response?

yadij: Yes, Squid should return 404 if there is no MGR_INDEX file.
The relevant test it needs to **fail** is the javascript found here: https://github.com/squid-cache/squid/blob/master/tools/cachemgr.cc#L278

yadij: Ah, Doh. I took on faith that the bug ID was one of the bugs this PR fixes. Looking at it now, it should (currently) not be in the title.

Using the hard-coded string instead of the MGR_INDEX file is a major bug in itself, and there are those log warnings fixed. So based on that I consider this change worth adding even if it does not fix the bug 5152 issue. Your call whether you want to include bug 5152 fix or leave it for later PR.

rousskov: > I took on faith that the bug ID was one of the bugs this PR fixes.

AFAICT, everybody acted in good faith: PR authors thought that this PR fixes that bug. You thought the same. However, PR authors understanding of that bug did not quite match yours.

> Your call whether you want to include bug 5152 fix or leave it for later PR.

Now that I know what you want Squid to do, I do not think we should merge these code changes right now. They introduce errors classification that may not exist in a fully fixed Squid. That would be a step backwards (along with a step forward in other directions).

I will mark this PR as a Draft and ask @eduard-bagdasaryan to post a new PR that teaches Squid to send 404 responses when MGR_INDEX template file is not found. That new PR branch can reuse as much (or as little) code from this PR branch, of course. Once that second PR merges, we will either close this PR or merge the remaining/still-useful bits, depending if there is anything left to merge here at that time.

rousskov: > Alex: I will mark this PR as a Draft and ask @eduard-bagdasaryan to post a new PR that teaches Squid to send 404 responses when MGR_INDEX template file is not found. That new PR branch can reuse as much (or as little) code from this PR branch as needed, of course. 

That new PR is #1191.

> Once that second PR merges, we will either close this PR or merge the remaining/still-useful bits, depending if there is anything left to merge here at that time.

That is still the plan.

src/errorpage.cc
```diff
@@ -180,20 +180,30 @@ error_hard_text[] = {
...
     },
+    {
+        ERR_REQUEST_PARSE_TIMEOUT,
+        "request parse timedout"
+    },
+    {
+        ERR_RELAY_REMOTE,
+        "relay server response"
+    }
+}};
```
rousskov, line 191: AFAICT, adding ERR_REQUEST_PARSE_TIMEOUT/ERR_RELAY_REMOTE to this container and the conversion of this container to std::array are the only improvements worth keeping in this PR after master/v6 commit 29304c7.  We should keep errorFindHardText() name/interface (while updating its implementation) IMO.

@eduard-bagdasaryan, please update this PR branch and PR title/description accordingly, removing the current Draft mark from the PR.

src/errorpage.cc
rousskov: Resolving this thread after PR 1191 got merged (master/v6 commit 29304c72e91ad2a3c3b2fd6a9e1a5223e4ef4bf0). The next steps are now detailed in a dedicated change request.

rousskov (changes requested):

src/errorpage.cc
eduard-bagdasaryan: Done.

src/errorpage.cc
```diff
@@ -142,44 +142,49 @@ static void ValidateStaticError(const int page_id, const SBuf &inputLocation);
...
-        "<div id=\"footer\">\n"
-        "Generated %T by %h (%s)\n"
-        "</div>\n"
-        "</body></html>\n"
-    },
-    {
-        TCP_RESET,
-        "reset"
-    },
+static const std::array<HardCodedError, 7> HardCodedErrors = {
```
rousskov, line 153: We can convert this into an std::map or similar (and get rid of HardCodedError completely), but since there are just seven items, the current linear search is probably fast enough, especially since these errors are loaded/checked only once per (re)configuration.

src/errorpage.cc
```diff
@@ -142,44 +142,49 @@ static void ValidateStaticError(const int page_id, const SBuf &inputLocation);
 
 /* local constant and vars */
 
-/**
- \ingroup ErrorPageInternal
- *
- \note  hard coded error messages are not appended with %S
- *      automagically to give you more control on the format
```
rousskov, line 149: This stale description was rather misleading because, in modern code, stored hard-coded error messages are never actually sent (so "control on the format" is useless) with the exception of ERR_SQUID_SIGNATURE, which is not a true error but the signature (i.e. `%S` substitution) itself (so "appending %S" would not make sense for it).

src/errorpage.cc
```diff
@@ -142,44 +142,49 @@ static void ValidateStaticError(const int page_id, const SBuf &inputLocation);
...
-        "unexpected client disconnect"
-    },
-    {
-        ERR_SECURE_ACCEPT_FAIL,
-        "secure accept fail"
-    },
-    {
-        ERR_REQUEST_START_TIMEOUT,
-        "request start timedout"
+        {
```
rousskov, line 155: These formatting changes are the result of the required double-curly braces for our std::array initialization. Unfortunately, they obscure the true changes in this PR -- adding two trailing entries to this array.

rousskov (approved):
LGTM, thank you. I pushed a couple of minor improvements and polished PR title/description. Please check.

Amos has approved an earlier version of this PR, but I am re-requesting his review because the PR changed a lot since then.

yadij (approved):

PR #1177, opened 2022-11-01
Sort CA certificates in tls-cert=bundle

rousskov:
... as opposed to requiring the bundle to contain CA certificates in the
correct on-the-wire/issuing order and warning about (and then ignoring)
out-of-order bundled certificates.

This enhancement makes it easier to configure Squid to send the right
intermediate certificates, especially during certificate upgrades when
intermediate certificates (which may be obtained from a source not
tightly coordinated with the signing certificate source) is likely to
contain a mix of old and new intermediate certificates.

Squid has to (and did) check the certificate issuing order anyway. The
new code uses very similar checks to sort intermediate certificates.

No on-the-wire changes expected for Squid configurations that have
correctly ordered certificate bundles. Other configurations may start
sending the previously missing intermediate certificates.

### DIFF
--- a/src/security/KeyData.cc
+++ b/src/security/KeyData.cc
@@ -47,25 +47,23 @@ Security::KeyData::loadCertificates()
         // Squid sends `cert` (loaded above) followed by certificates in `chain`
         // (formed below by loading and sorting the remaining certificates).
 
-        // scan all the remaining configured certificates and keep intermediates
+        // load all the remaining configured certificates
         CertList intermediates;
-        while (const auto ca = Ssl::ReadOptionalCertificate(bio)) {
-            // We ignore a self-signed certificate because it should not be sent:
-            // The recipients that do not already have it should not trust it.
-            // Ignoring self-signed also simplifies the intermediates loop below.
-            if (SelfSigned(*ca)) {
-                debugs(83, DBG_IMPORTANT, "WARNING: Ignoring a self-signed CA " << *ca);
-                continue;
-            }
-
+        while (const auto ca = Ssl::ReadOptionalCertificate(bio))
             intermediates.emplace_back(ca);
-        }
 
         // Push certificates into `chain` in on-the-wire order, as defined by
         // RFC 8446 Section 4.4.2: "Each following certificate SHOULD directly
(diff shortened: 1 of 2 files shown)


src/security/KeyData.cc
```diff
@@ -44,32 +44,40 @@ Security::KeyData::loadCertificates()
...
+            // We cannot chain any certificate after a self-signed certificate.
+            // This check also protects the IssuedBy() search below from adding
+            // duplicated (i.e. listed multiple times) self-signed certificates.
+            if (SelfSigned(*precedingCert))
+                break;
+
+            const auto issuerPos = std::find_if(candidates.begin(), candidates.end(), [&](const CertPointer &i) {
+                return IssuedBy(*precedingCert, *i);
+            });
+            if (issuerPos == candidates.end())
```
yadij, line 70: Please prefer ranged-for loop instead of std::find_if(). When the search covers the entire container and can be short-circuited by an internal condition the range scan is faster.

yadij (approved):
I do not see anything particularly problematic. However, I do wonder how being this lenient of admin misconfiguration will fare in real-world usage.

src/security/KeyData.cc
rousskov: > Please prefer ranged-for loop instead of std::find_if(). 

I already do. However, here, the range-based `for` loop does not work because we need to know the _position_ of the found certificate (to delete it from the `candidates` container).

> When the search covers the entire container and can be short-circuited by an internal condition the range scan is faster.

FWIW, I do not think the above statement is accurate. All std::find_if() implementations short-circuit just like a range-based `for` loop would, of course. In optimized builds, the (small) performance difference, if any, may come from the fact that range-based `for` loop does not return the iterator to the caller (which the caller usually has to dereference again).

rousskov (commented):
> I do wonder how being this lenient of admin misconfiguration will fare in real-world usage.

I share your overall concern about unknown side effects (even though I do not consider out-of-wire-order bundles to be a sign of a misconfiguration). I think it is the risk worth taking due to the added convenience of assembling certificate bundles from various sources.

Thank you for putting this PR on the fast track.

PR #1178, opened 2022-11-06
Remove --enable-kill-parent-hack

yadij:
This feature has been deprecated for some time.

The useful functionality can more safely be implemented in the parent
script by using the --foreground command line option and waiting for
the Squid process to exit.

It has always been labeled dangerous as the parent process can be
PID 1 or some other vital system process than the intended
RunCache script. Currently it breaks SMP support for manually
killing a single worker or disker process.


kinkie (approved):
yes. please!

rousskov (commented):
Thank you for removing this hack. I cannot review the details carefully due to travel, but I support this removal in principle.

PR #1179, opened 2022-11-07
Refactor SignalEngine

yadij:
Add LegacySignalRr to perform actions which should be
in module specific runners, but not yet extracted.


src/main.cc
```diff
@@ -940,116 +892,166 @@ mainReconfigureFinish(void *)
...
-#if ICAP_CLIENT
-    icapLogRotate();               /*icap.log*/
-#endif
-    icmpEngine.Open();
-    redirectInit();
+    virtual void endingShutdown() {
 #if USE_AUTH
-    authenticateInit(&Auth::TheConfig.schemes);
+        /* detach the auth components (only do this on full shutdown) */
+        Auth::Scheme::FreeAll();
```
rousskov, line 1050: AFAICT, we should not destroy authentication state while the event loop is still running.

If possible, we should not destroy it (manually) at all.

If necessary, we should destroy at or after RR destruction time. The current official code effectively tries to do the latter -- calling this function after the endingShutdown event. This PR can probably mimic that by moving this call into the RR destructor.

rousskov (changes requested):
Just a partial review. Hopefully, this will help us make steady progress.

src/main.cc
```diff
@@ -940,116 +892,166 @@ mainReconfigureFinish(void *)
...
 
     RunRegisteredHere(RegisteredRunner::syncConfig);
+    reconfiguring = 0;
+}
 
-    if (IamPrimaryProcess())
-        CpuAffinityCheck();
-    CpuAffinityReconfigure();
+void
+SignalEngine::doReconfigure()
```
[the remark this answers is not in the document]
yadij: Part of this PR is trying to avoid/reduce C-style pre-declaration of static function. The mainReconfigureFinish() needs to be defined before it is used.

src/main.cc
yadij: Hmm. The current official code fails to meet that intention. There is at least one more cycle of events scheduled after this auth free operation completes.

Done in [55738e4](https://github.com/squid-cache/squid/pull/1179/commits/55738e49c3879fcf01e8f63c9b671c921f9a343b)

src/main.cc
```diff
@@ -1027,6 +916,119 @@ mainRotate(void)
     externalAclInit();
 }
 
+/// perform legacy actions for process signal handling
+/// TODO: move actions into module runners
+class LegacySignalRr : public RegisteredRunner
```
rousskov, line 921: These actions are not necessarily "legacy"; they are simply misplaced. That is why the TODO comment says we want to _move_ them (rather than deprecate/replace -- things that usually happen to _legacy_ code). I suggest calling this class MainRr, MyRr, or something "local" like that.

rousskov (changes requested):

src/main.cc
```diff
@@ -916,86 +868,23 @@ mainReconfigureFinish(void *)
                ") requires a full restart. It has been ignored by reconfigure.");
         Config.workers = oldWorkers;
     }
+    leave_suid();
 
     RunRegisteredHere(RegisteredRunner::syncConfig);
+    reconfiguring = 0;
+}
 
-    if (IamPrimaryProcess())
```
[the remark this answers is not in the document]
rousskov: I just noticed that the current changes result in assertions in most GitHub CI functionality [tests](https://github.com/squid-cache/squid/actions/runs/3446978512/jobs/5752495938#step:9:130):

```
2022/11/11 18:04:16| FATAL: assertion failed: Scheme.cc:62: "shutting_down"
  Aborted (core dumped)
```

I do not know whether the PR bug causing those assertions is related to my overall concerns about actions order dependencies and uncertainty.

yadij:
Abandoning this approach.

PR #1181, opened 2022-11-16
Shuffle MemAllocator to Mem namespace

yadij:
Update and add missing documentation for class members.

Update syntax to latest standard compliance and Squid
coding guideline style.

### DIFF
--- a/src/cbdata.cc
+++ b/src/cbdata.cc
@@ -476,7 +476,7 @@ cbdataDump(StoreEntry * sentry)
     storeAppendPrintf(sentry, "types\tsize\tallocated\ttotal\n");
 
     for (int i = 1; i < cbdata_types; ++i) {
-        if (auto *pool = cbdata_index[i].pool) {
+        if (const auto pool = cbdata_index[i].pool) {
 #if WITH_VALGRIND
             int obj_size = pool->objectSize();
 #else
--- a/src/cbdata.cc
+++ b/src/cbdata.cc
@@ -132,7 +132,7 @@ static OBJH cbdataDumpHistory;
 #endif
 
 struct CBDataIndex {
-    MemAllocator *pool;
+    Mem::Allocator *pool;
 }
 *cbdata_index = nullptr;
 
@@ -476,9 +476,7 @@ cbdataDump(StoreEntry * sentry)
     storeAppendPrintf(sentry, "types\tsize\tallocated\ttotal\n");
 
     for (int i = 1; i < cbdata_types; ++i) {
-        MemAllocator *pool = cbdata_index[i].pool;
-
-        if (pool) {
+        if (auto *pool = cbdata_index[i].pool) {
 #if WITH_VALGRIND
             int obj_size = pool->objectSize();
 #else
--- a/src/mem/Allocator.h
+++ b/src/mem/Allocator.h
@@ -58,7 +58,7 @@ public:
      * \param minSize Minimum size needed to be allocated.
      * \retval n Smallest size divisible by sizeof(void*)
      */
-    static size_t RoundedSize(size_t minSize) {return ((minSize + sizeof(void*) - 1) / sizeof(void*)) * sizeof(void*);}
+    static size_t RoundedSize(const size_t minSize) { return ((minSize + sizeof(void*) - 1) / sizeof(void*)) * sizeof(void*); }
 
 protected:
     /**
--- a/src/mem/Allocator.h
+++ b/src/mem/Allocator.h
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 1996-2022 The Squid Software Foundation and contributors
+ *
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID_SRC_MEM_ALLOCATOR_H
+#define SQUID_SRC_MEM_ALLOCATOR_H
+
+#include "base/TypeTraits.h"
+#include "mem/forward.h"
+
+namespace Mem
+{
+
+/// An interface for memory allocators that deal with fixed-size objects.
+/// Allocators may optimize repeated de/allocations using memory pools.
+class Allocator : public Interface
+{
+public:
+    explicit Allocator(const char * const aLabel): label(aLabel) {}
+
+    // TODO make this method const
+    /**
+     * fill the given object with statistical data about pool
+     * \returns Number of objects in use, ie. allocated.
+     */
+    virtual int getStats(MemPoolStats *, int accumulate = 0) = 0;
+
+    virtual PoolMeter const &getMeter() const = 0;
+
+    /// provide (and reserve) memory suitable for storing one object
+    virtual void *alloc() = 0;
+
+    /// return memory reserved by alloc()
+    virtual void freeOne(void *) = 0;
+
+    /// brief description of objects returned by alloc()
+    virtual char const *objectType() const { return label; }
+
+    /// the size (in bytes) of objects managed by this allocator
+    virtual size_t objectSize() const = 0;
+
+    /// the difference between the number of alloc() and freeOne() calls
+    virtual int getInUseCount() = 0;
+
+    /// \see doZero
+    void zeroBlocks(const bool doIt) { doZero = doIt; }
+
+    int inUseCount() { return getInUseCount(); } // XXX: drop redundant?
+
+    /// XXX: Misplaced -- not all allocators have a notion of a "chunk". See MemPoolChunked.
+    virtual void setChunkSize(size_t) {}
+
+    /**
+     * \param minSize Minimum size needed to be allocated.
+     * \retval n Smallest size divisible by sizeof(void*)
+     */
+    static size_t RoundedSize(size_t minSize) {return ((minSize + sizeof(void*) - 1) / sizeof(void*)) * sizeof(void*);}
+
+protected:
+    /**
+     * Whether to zero memory on initial allocation and on return to the pool.
+     *
+     * We do this on some pools because many object constructors are/were incomplete
+     * and we are afraid some code may use the object after free.
+     * When possible, set this to false to avoid zeroing overheads.
+     */
+    bool doZero = true;
+
+private:
+    const char *label = nullptr;
+};
+
+} // namespace Mem
+
+#endif /* SQUID_SRC_MEM_ALLOCATOR_H */
--- a/src/mem/AllocatorProxy.h
+++ b/src/mem/AllocatorProxy.h
@@ -9,7 +9,10 @@
 #ifndef _SQUID_SRC_MEM_ALLOCATORPROXY_H
 #define _SQUID_SRC_MEM_ALLOCATORPROXY_H
 
-class MemAllocator;
+// XXX: remove AllocatorProxy.h include from mem/forward.h
+namespace Mem {
+class Allocator;
+}
 class MemPoolStats;
 
 /**
@@ -80,11 +83,11 @@ public:
     void zeroBlocks(bool doIt);
 
 private:
-    MemAllocator *getAllocator() const;
+    Allocator *getAllocator() const;
 
     const char *label;
     size_t size;
-    mutable MemAllocator *theAllocator;
+    mutable Allocator *theAllocator;
     bool doZero;
 };
 
--- a/src/mem/Pool.cc
+++ b/src/mem/Pool.cc
@@ -247,7 +247,7 @@ memPoolsTotalAllocated(void)
     return stats.TheMeter->alloc.currentLevel();
 }
 
-MemImplementingAllocator::MemImplementingAllocator(char const *aLabel, size_t aSize) :
+MemImplementingAllocator::MemImplementingAllocator(char const * const aLabel, const size_t aSize):
     Mem::Allocator(aLabel),
     next(nullptr),
     alloc_calls(0),
--- a/src/mem/Pool.cc
+++ b/src/mem/Pool.cc
@@ -90,18 +90,6 @@ MemPools::setDefaultPoolChunking(bool const &aBool)
     defaultIsChunked = aBool;
 }
 
-char const *
-MemAllocator::objectType() const
-{
-    return label;
-}
-
-int
-MemAllocator::inUseCount()
-{
-    return getInUseCount();
-}
-
 void
 MemImplementingAllocator::flushMeters()
 {
@@ -216,7 +204,6 @@ int
 memPoolGetGlobalStats(MemPoolGlobalStats * stats)
 {
     int pools_inuse = 0;
-    MemAllocator *pool;
     MemPoolIterator *iter;
 
     memset(stats, 0, sizeof(MemPoolGlobalStats));
@@ -226,7 +213,7 @@ memPoolGetGlobalStats(MemPoolGlobalStats * stats)
 
     /* gather all stats for Totals */
     iter = memPoolIterate();
-    while ((pool = memPoolIterateNext(iter))) {
+    while (const auto pool = memPoolIterateNext(iter)) {
         if (pool->getStats(&pp_stats, 1) > 0)
             ++pools_inuse;
     }
@@ -246,19 +233,10 @@ memPoolGetGlobalStats(MemPoolGlobalStats * stats)
     stats->tot_items_inuse = pp_stats.items_inuse;
     stats->tot_items_idle = pp_stats.items_idle;
 
-    stats->tot_overhead += pp_stats.overhead + MemPools::GetInstance().poolCount * sizeof(MemAllocator *);
+    stats->tot_overhead += pp_stats.overhead + MemPools::GetInstance().poolCount * sizeof(Mem::Allocator *);
     stats->mem_idle_limit = MemPools::GetInstance().idleLimit();
 
     return pools_inuse;
-}
-
-MemAllocator::MemAllocator(char const *aLabel) : doZero(true), label(aLabel)
-{
-}
-
-size_t MemAllocator::RoundedSize(size_t s)
-{
-    return ((s + sizeof(void*) - 1) / sizeof(void*)) * sizeof(void*);
 }
 
 int
@@ -269,7 +247,8 @@ memPoolsTotalAllocated(void)
     return stats.TheMeter->alloc.currentLevel();
 }
 
-MemImplementingAllocator::MemImplementingAllocator(char const *aLabel, size_t aSize) : MemAllocator(aLabel),
+MemImplementingAllocator::MemImplementingAllocator(char const *aLabel, size_t aSize) :
+    Mem::Allocator(aLabel),
     next(nullptr),
     alloc_calls(0),
     free_calls(0),
--- a/src/mem/Pool.h
+++ b/src/mem/Pool.h
@@ -28,6 +28,7 @@
  *     might be the way to go.
  */
 
+#include "mem/Allocator.h"
 #include "mem/Meter.h"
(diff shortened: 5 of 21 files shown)


src/mem/AllocatorProxy.h
```diff
@@ -9,7 +9,10 @@
 #ifndef _SQUID_SRC_MEM_ALLOCATORPROXY_H
 #define _SQUID_SRC_MEM_ALLOCATORPROXY_H
 
-class MemAllocator;
+// XXX: remove AllocatorProxy.h include from mem/forward.h
```
rousskov, line 12: Any reason this XXX should not be addressed in this PR? It seems to be very much in scope, but I am probably missing some caveats.

src/mem/Pool.cc
```diff
@@ -259,7 +246,7 @@ memPoolGetGlobalStats(MemPoolGlobalStats * stats)
 
     /* gather all stats for Totals */
     iter = memPoolIterate();
-    while ((pool = memPoolIterateNext(iter))) {
+    while (auto *pool = memPoolIterateNext(iter)) {
```
rousskov, line 249: If we are going to polish this code, let's improve its const-correctness:

```suggestion
    while (const auto pool = memPoolIterateNext(iter)) {
```

I do not insist on this change. There are many other PR-modified places where const-correctness can be improved. I am not going to mark any of them, but, if you do not mind, please improve them. Also, if you prefer that I do these changes, please let me know, and I will.

src/mem/Pool.h
```diff
@@ -171,73 +172,8 @@ class MemPools
...
-private:
-    const char *label;
-};
-
 /// \ingroup MemPoolsAPI
-class MemImplementingAllocator : public MemAllocator
+class MemImplementingAllocator : public Mem::Allocator
 {
 public:
     MemImplementingAllocator(char const *aLabel, size_t aSize);
```
rousskov, line 179: > PR description: Update and add missing documentation for class members. Including some child classes.

Please remove overriden method _descriptions_ below since we are going to be polishing their original/true descriptions in Mem::Allocator. Please check grandkids as well.

src/tests/stub_libmem.cc
```diff
@@ -11,6 +11,12 @@
 #define STUB_API "mem/libmem.la"
 #include "tests/STUB.h"
 
+#include "mem/Allocator.h"
+namespace Mem {
+//Allocator::Allocator(char const *aLabel);
```
rousskov, line 16: I do not think it is a good idea to comment out inlined methods. These comments often become stale and create noise in code searches. The PR already correctly drops a similar method -- inUseCount().

```suggestion
```

src/mem/Pool.h
```diff
@@ -171,73 +172,8 @@ class MemPools
...
-     */
-    bool doZero;
-
-private:
-    const char *label;
-};
-
 /// \ingroup MemPoolsAPI
-class MemImplementingAllocator : public MemAllocator
+class MemImplementingAllocator : public Mem::Allocator
```
rousskov, line 176: > PR description: Update virtual/override syntax to latest standard compliance and squid coding guideline style.

This PR does not add `override`. Should we drop that claim? Mark the PR as a Draft and that claim as a TODO?

FWIW, I certainly do not insist on (and do not even recommend!) adding override manually to the otherwise unchanged lines. Most likely, it should be done automatically, using a clang-tidy script.

rousskov (dismissed):
Thank you for improving MemAllocator.

src/mem/Allocator.h
```diff
@@ -0,0 +1,90 @@
...
+#include "mem/forward.h"
+
+namespace Mem
+{
+
+/**
+ * Basic API definition for memory pooling allocators.
+ *
+ * A pool is a [growing] space for objects of the same size
+ */
```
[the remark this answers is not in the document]
yadij: IMO the pooling mention should stay. The distinction of this API from, eg the STL allocators, is that it explicitly plugs the configurable Squid "Memory Pools" into the allocator logic. Without the pooling ability (even when disabled/unused) there would be no point in Squid providing its own allocator(s).

The part that I think we should discard is the "same size" / "fixed-size" wording from the old definition. Since the child PoolMalloc already deals with non-same size objects and the "Memory Pools" documentation should be where the definition of a pool is.

src/mem/Pool.h
yadij: Sorry. Missed dropping it on the scope re-scope.

src/mem/AllocatorProxy.h
yadij: I have scoped this PR at only the MemAllocator class. Mem::AllocatorProxy is a very widely used class and would drag in a lot of include sequence changes that bloat the PR.

src/mem/Pool.cc
yadij: Unfortunately the getStats() logic alters the pool. For now I would like to to leave those deep logic changes out of scope.
Adding a TODO now.

src/mem/Pool.h
yadij: Dropped the claim.

src/mem/Pool.h
yadij: Done in[b48d53b](https://github.com/squid-cache/squid/pull/1181/commits/b48d53b23db6f2b01f52b8b94f7a63e7e6a31ea9)

src/mem/Pool.cc
yadij: TODO added in [b48d53b](https://github.com/squid-cache/squid/pull/1181/commits/b48d53b23db6f2b01f52b8b94f7a63e7e6a31ea9)

src/mem/Allocator.h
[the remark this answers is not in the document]
rousskov: > The distinction of this API from, eg the STL allocators, is that it explicitly plugs the configurable Squid "Memory Pools" into the allocator logic.

I see nothing about pools in this API. There are a couple of words "pool" sprinkled around for historical reasons, but the things they name represent pooling-agnostic concepts (or what should be pooling-agnostic concepts).

> Without the pooling ability there would be no point in Squid providing its own allocator(s).

I disagree. The statistics maintained by Allocators is very valuable without any pooling. And there is also memory zero-ing functionality.


> The part that I think we should discard is the "same size" / "fixed-size" wording from the old definition. Since the child PoolMalloc already deals with non-same size objects

MemPoolMalloc does not "deal with non-same size objects" AFAICT. It allocates `obj_size`-byte objects. Am I missing something?

> and the "Memory Pools" documentation should be where the definition of a pool is.

Agreed, but "fixed-size objects" is not about pooling.

src/mem/Pool.cc
rousskov: You are misinterpreting what the suggestion does. It is unrelated to the TODO you have added. Try applying the change. If it works, consider improving other, similar changed places as well.

src/mem/Allocator.h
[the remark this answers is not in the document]
yadij: The hierarchy of classes for which this is the base/root class _is_ "memory pools". Each class of this hierarchy is providing various ways to access a pool instance. To be clear what I am proposing is to keep the line I added in this PR
, with maybe some slight modifications like so:
 * "Basic API definition for memory pooling allocators."
 * "Basic API definition for Memory Pools allocators."
 
What you are missing about PoolMalloc is that the obj_size it allocates can be updated. 




> > The distinction of this API from, eg the STL allocators, is that it explicitly plugs the configurable Squid "Memory Pools" into the allocator logic.
> 
> I see nothing about pools in this API. There are a couple of words "pool" sprinkled around for historical reasons, but the things they name represent pooling-agnostic concepts (or what should be pooling-agnostic concepts).
> 

Exactly why mention of memory pools should stay in the description. To clarify that this API is how pool allocators are accessed by code not caring about how the pool operates.


> > Without the pooling ability there would be no point in Squid providing its own allocator(s).
> 
> I disagree. The statistics maintained by Allocators is very valuable without any pooling. And there is also memory zero-ing functionality.
> 

In general that might be true, but the statistics implemented by Squid are all about pools. Memory zero'ing logic is also all about pools. Non-pooling would (and does) just use calloc().

> > The part that I think we should discard is the "same size" / "fixed-size" wording from the old definition. Since the child PoolMalloc already deals with non-same size objects
> 
> MemPoolMalloc does not "deal with non-same size objects" AFAICT. It allocates `obj_size`-byte objects. Am I missing something?
> 

Sorry, I was crossing PoolMalloc with the xmalloc pools. To remove the old C-style buffer/strings/rigid we will need a pool class that accepts multiple sizes of object and manages them. I am thinking that would be PoolMalloc growing ability to provide re-alloc.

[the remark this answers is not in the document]
rousskov: > The hierarchy of classes for which this is the base/root class _is_ "memory pools".

I disagree. It is a hierarchy of memory allocators. Pooling is just one of the optional features some allocators are capable of.

> Each class of this hierarchy is providing various ways to access a pool instance. 

Each leaf class implements a memory allocator interface. Implementation details and extras that vary from class to class are not important when we define that interface.

> To be clear what I am proposing is to keep the line I added in this PR , with maybe some slight modifications like so:
> * "Basic API definition for memory pooling allocators."
> * "Basic API definition for Memory Pools allocators."

We should document the interface declared by the Allocator class. That interface has nothing to do with pooling. That interface clients/users do not care about pooling. Even if every single leaf class implements some form of pooling, it would be still wrong to document this base interface as a pooling allocator interface. There is no reason to add things to the interface description that the interface itself and its users/clients do not care about, especially when those things may be disabled.


> > > The distinction of this API from, eg the STL allocators, is that it explicitly plugs the configurable Squid "Memory Pools" into the allocator logic.

> > I see nothing about pools in this API. There are a couple of words "pool" sprinkled around for historical reasons, but the things they name represent pooling-agnostic concepts (or what should be pooling-agnostic concepts).

> Exactly why mention of memory pools should stay in the description. To clarify that this API is how pool allocators are accessed by code not caring about how the pool operates.

The suggested wordings are not _clarifications_ IMO. They mislead the reader and may be in the way of adding special allocators that do not pool at all. However, I support mentioning pooling in the following (or similar) non-interface-describing way. It may help with code navigation and such.

```suggestion
/// An interface for memory allocators that deal with fixed-size objects.
/// Allocators may optimize repeated de/allocations using memory pools.
```

> > > Without the pooling ability there would be no point in Squid providing its own allocator(s).
> > 
> > 
> > I disagree. The statistics maintained by Allocators is very valuable without any pooling. And there is also memory zero-ing functionality.
> 
> In general that might be true, but the statistics implemented by Squid are all about pools.

The statistics implemented by Squid has very valuable parts that are not related to pooling in any way. One may disable pooling and still get valuable memory allocation stats.

> Memory zero'ing logic is also all about pools.

Memory zeroing is about preparation of allocated (or freed) object memory. It is not related to pooling in any way.

> Non-pooling would (and does) just use calloc().

The existence of multiple ways to allocate, zero, and pool object memory does not imply that there is no point in Squid providing its own allocators without pooling.

> To remove the old C-style buffer/strings/rigid we will need a pool class that accepts multiple sizes of object and manages them.

I am not sure that is the right path forward, but we do not need to agree on that right now. If it turns out to be the right path, we can adjust the Allocator API accordingly, of course.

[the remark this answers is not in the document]
yadij: Taking your updated suggestion as "good enough". The remainder of the points we disagree on are perhapse best resolved later in PR relevant to those points of discussion. That last point in particular I am moving to the PR about AllocatorProxy.

src/mem/Allocator.h
```diff
@@ -0,0 +1,88 @@
...
+    /// free a element allocated by Allocator::alloc()
+    virtual void freeOne(void *) = 0;
+
+    /// get a display label for objects in this pool
+    virtual char const *objectType() const { return label; }
+
+    /// the size (in bytes) of objects managed by this allocator
+    virtual size_t objectSize() const = 0;
+
+    /// the number of objects currently allocated
```
rousskov, line 46: The undefined term "allocated object" can be easily misinterpreted. We do not allocate objects. We provide memory for objects, but this method does not return the number of currently allocated memory blocks.

```suggestion
    /// the difference between the number of alloc() and freeOne() calls
```

src/mem/Allocator.cc
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+#include "squid.h"
+#include "mem/Allocator.h"
+
+size_t
+Mem::Allocator::RoundedSize(size_t s)
+{
+    return ((s + sizeof(void*) - 1) / sizeof(void*)) * sizeof(void*);
+}
```
rousskov, line 16: Please consider inlining this trivial no-dependencies method, especially if doing so eliminates the need for Allocator.cc.

src/mem/Allocator.h
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Mem
+{
+
+/// An interface for memory allocators that deal with fixed-size objects.
+/// Allocators may optimize repeated de/allocations using memory pools.
+class Allocator
+{
+public:
+    Allocator(char const *aLabel) : label(aLabel) {}
```
rousskov, line 22: If possible, please make this single-parameter constructor explicit:

```suggestion
    explicit Allocator(const char * const aLabel): label(aLabel) {}
```

src/mem/Allocator.h
```diff
@@ -0,0 +1,88 @@
...
+#define SQUID_SRC_MEM_ALLOCATOR_H
+
+#include "mem/forward.h"
+
+namespace Mem
+{
+
+/// An interface for memory allocators that deal with fixed-size objects.
+/// Allocators may optimize repeated de/allocations using memory pools.
+class Allocator
```
rousskov, line 19: Please consider inheriting this interface class from Interface mixin, to fix several problems with the current Allocator declaration. If you do so, the explicit destructor declaration can be removed.

src/mem/Pool.cc
rousskov: I unresolved this old change request because it was not resolved.

rousskov (commented):
There is a lot we can still polish in the moved descriptions and moved/adjusted code, but I do not insist on further changes here because I do not have the time to provide them and they seem to be of minor importance.

src/mem/Pool.cc
yadij:
```
../../../src/mem/Pool.cc: In function 'int memPoolGetGlobalStats(MemPoolGlobalStats*)':
../../../src/mem/Pool.cc:238:27: error: passing 'const MemImplementingAllocator' as 'this' argument discards qualifiers [-fpermissive]
  238 |         if (pool->getStats(&pp_stats, 1) > 0)
      |             ~~~~~~~~~~~~~~^~~~~~~~~~~~~~
In file included from ../../../src/mem/Pool.h:31,
                 from ../../../src/mem/PoolChunked.h:12,
                 from ../../../src/mem/Pool.cc:14:
../../../src/mem/Allocator.h:30:17: note:   in call to 'virtual int Mem::Allocator::getStats(MemPoolStats*, int)'
   30 |     virtual int getStats(MemPoolStats *, int accumulate = 0) = 0;
      |                 ^~~~~~~~

```

src/mem/Allocator.h
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Allows you tune chunk size of pooling. Objects are allocated in chunks
+     * instead of individually. This conserves memory, reduces fragmentation.
+     * Because of that memory can be freed also only in chunks. Therefore
+     * there is tradeoff between memory conservation due to chunking and free
+     * memory fragmentation.
+     *
+     * \note  As a general guideline, increase chunk size only for pools that keep
+     *        very many items for relatively long time.
+     */
```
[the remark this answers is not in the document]
yadij: Huh? this makes me think you do not understand the purpose of a virtual is. Whether the pool has an internal concept of chunk or not,  the interface and the caller do and they need to communicate about chunks.

src/mem/Pool.cc
rousskov: Please try my suggestion (instead of trying something I am _not_ suggesting). My suggestion does not make MemImplementingAllocator constant. Moreover, I just tried it (using official Squid code), and it does work as expected.

src/mem/Allocator.h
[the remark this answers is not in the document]
rousskov: > Whether the pool has an internal concept of chunk or not, the interface and the caller do and they need to communicate about chunks.

This interface and the callers using it are currently broken (because some of the allocators do not support a notion of a "chunk"). This method does not belong to the root hierarchy class. That base class is supposed to define what is _common_ among derived classes. It is not a collection of all the things some child classes can do.

src/mem/Pool.cc
yadij: Oh, its the asterisk. sorry.

src/mem/Allocator.h
```diff
@@ -0,0 +1,79 @@
...
+    int inUseCount() { return getInUseCount(); } // XXX: drop redundant?
+
+    /// XXX: Misplaced -- not all allocators have a notion of a "chunk". See MemPoolChunked.
+    virtual void setChunkSize(size_t) {}
+
+    /**
+     * \param minSize Minimum size needed to be allocated.
+     * \retval n Smallest size divisible by sizeof(void*)
+     */
+    static size_t RoundedSize(size_t minSize) {return ((minSize + sizeof(void*) - 1) / sizeof(void*)) * sizeof(void*);}
```
rousskov, line 61:
```suggestion
    static size_t RoundedSize(const size_t minSize) { return ((minSize + sizeof(void*) - 1) / sizeof(void*)) * sizeof(void*); }
```

src/mem/Pool.cc
```diff
@@ -269,7 +247,8 @@ memPoolsTotalAllocated(void)
     return stats.TheMeter->alloc.currentLevel();
 }
 
-MemImplementingAllocator::MemImplementingAllocator(char const *aLabel, size_t aSize) : MemAllocator(aLabel),
+MemImplementingAllocator::MemImplementingAllocator(char const *aLabel, size_t aSize) :
```
rousskov, line 250:
```suggestion
MemImplementingAllocator::MemImplementingAllocator(char const * const aLabel, const size_t aSize):
```

src/cbdata.cc
```diff
@@ -476,9 +476,7 @@ cbdataDump(StoreEntry * sentry)
     storeAppendPrintf(sentry, "types\tsize\tallocated\ttotal\n");
 
     for (int i = 1; i < cbdata_types; ++i) {
-        MemAllocator *pool = cbdata_index[i].pool;
-
-        if (pool) {
+        if (auto *pool = cbdata_index[i].pool) {
```
rousskov, line 479:
```suggestion
        if (const auto pool = cbdata_index[i].pool) {
```

rousskov (approved):
Thank you for polishing PR code. I found a few places where the code can be trivially polished a bit more, but I still do not insist on any changes here.

yadij:
> Thank you for polishing PR code. I found a few places where the code can be trivially polished a bit more, but I still do not insist on any changes here.

Please don't forget C/C++ are not a functional languages. A "const X" parameter is not constant when X is a trivial type.

rousskov:
> Amos: A "const X" parameter is not constant when X is a trivial type.

A parameter `foo` declared as `const X foo` is a "const object" (a.k.a. a constant) regardless of whether `X` is trivial:

> [const object](https://en.cppreference.com/w/cpp/language/cv) - an object whose type is const-qualified, or a non-mutable subobject of a const object. Such object cannot be modified: attempt to do so directly is a compile-time error, and attempt to do so indirectly (e.g., by modifying the const object through a reference or pointer to non-const type) results in undefined behavior.

PR #1182, opened 2022-11-18
Integrate Mem::AllocatorProxy with Mem::Allocator

yadij:
AllocatorProxy is supposed to provide MEMPROXY_CLASS
objects with access to Mem::Allocator API but with
delayed instantiation of the allocator.

But the APIs were implemented slightly different
and over time have further grown apart.

Resolve this situation for the long-term by inheriting
Mem::AllocatorProxy directly from Mem::Allocator
interface class and drop code duplication.


yadij:
Draft since this depends on PR #1181. The changes from that PR have been included to allow build and testing.
Significant changes of this PR are in mem/AllocatorProxy.(h,cc), mem/forward.h and mem/minimal.cc files.

src/mem/minimal.cc
```diff
@@ -25,19 +24,36 @@ Mem::AllocatorProxy::alloc()
...
     return Alive;
 }
 
+void
+Mem::AllocatorProxy::zeroBlocks(bool doIt)
+{
+    this->Allocator::zeroBlocks(doIt);
+}
+
+// TODO: extract this class from mem/Pool.h
```
yadij, line 45: This TODO is implemented in a followup PR.

src/mem/AllocatorProxy.h
```diff
@@ -48,43 +46,29 @@ namespace Mem
 /**
  * Support late binding of pool type for allocator agnostic classes
  */
-class AllocatorProxy
```
rousskov, line 51: I do not yet understand why we need this class in the current code. The class description does not help me because it is not clear what "late" really means in this context. The "pool type" is, I presume, objectType(), that is AllocatorProxy::label. That label is known at AllocatorProxy creation time. At that AllocatorProxy creation time, AFAICT, we usually[^1] call `Pool().alloc()` and alloc() indirectly creates/finalizes theAllocator via getAllocator() which calls `MemPools::GetInstance().create(objectType(), size)`, creating a very specific pool. There is no "late" or delayed actions in this sequence AFAICT. Why do we need all this indirection?

In other words, what would break if we remove the AllocatorProxy class completely and simply let MEMPROXY_CLASS's Pool() method call `MemPools::GetInstance().create(#CLASS, sizeof(CLASS), false)` and save/return the resulting Allocator?

[^1]: I see only one minor exception to that "usually" rule: Pool().inUseCount() does not call getAllocator() for a pool that never created any objects. That exception is not important, AFAICT, because if a pool created no objects, it would return correct statistics without the AllocatorProxy class existence anyway.

src/mem/AllocatorProxy.h
```diff
@@ -48,43 +46,29 @@ namespace Mem
...
-    MemPoolMeter const &getMeter() const;
-
-    /**
-     * \param stats Object to be filled with statistical data about pool.
-     * \retval      Number of objects in use, ie. allocated.
-     */
-    int getStats(MemPoolStats * stats);
-
-    void zeroBlocks(bool doIt);
+    /* Mem::Allocator API */
```
rousskov, line 58: If this class survives, please consider adding either `final override` or `override` to these "new" API overrides.

rousskov (changes requested):

src/mem/AllocatorProxy.h
yadij: > I do not yet understand why we need this class in the current code. The class description does not help me because it is not clear what "late" really means in this context. The "pool type" is, I presume, objectType(), that is AllocatorProxy::label.

FWIW, I am working with the understanding that in this context:
 * "pool" is essentially an alias for "allocator".
 * "pool type" is the specific allocator implementation Squid provides (eg chunked vs malloc vs xmalloc).
 * "late" means that the pool is not created until the first instance of a MEMPROXY class needs to be allocated by that pool.

As such Mem::AllocatorProxy is a first-class allocator alongside PoolChunked, PoolMalloc and xmalloc/xcalloc allocators.

AIUI the reason for this AllocatorProxy is to remove overheads from every class using MEMPROXY_CLASS macro. It provides a per-class singleton with all the meta data necessary for later pool creation. Allowing for the real allocator to be switched at run-time.
 That said, I'm not aware of any current logic using the run-time switch capability. There are several features on my wish-list that would use that ability, so I am not keen to remove it just yet.

> In other words, what would break if we remove the AllocatorProxy class completely and simply let MEMPROXY_CLASS's Pool() method call `MemPools::GetInstance().create(#CLASS, sizeof(CLASS), false)` and save/return the resulting Allocator?

Removing AllocatorProxy would mean adding its member variables to every class type. Including the ones this PR outsources with Mem::Allocator inheritance.
 - At which point it makes more sense to keep AllocatorProxy (may under some other name) and inherit MEMPROXY classes from it as well as using the macro to get the right operators. 


>     1. I see only one minor exception to that "usually" rule: Pool().inUseCount() does not call getAllocator() for a pool that never created any objects. That exception is not important, AFAICT, because if a pool created no objects, it would return correct statistics without the AllocatorProxy class existence anyway.

The exception has some importance ... it prevents the pool being created just by fetching stats.  What I am uncertain about myself is whether stats can actually be queried without the pool existing first.  Hopefully further cleanup will reveal the specifics so it can be documented properly or removed later.

src/mem/AllocatorProxy.h
yadij: Will do after the prerequisite PRs are merged.

src/mem/AllocatorProxy.h
rousskov: TLDR: Please remove this class unless you can find some other/new reasons to keep it. I still do not see any valid ones.

----

> AIUI the reason for this AllocatorProxy is to remove overheads from every class using MEMPROXY_CLASS macro.

AFAICT, this class does not remove any overheads. It clearly adds overheads.

> Allowing for the real allocator to be switched at run-time.

Existing AllocatorProxy does not support switching allocators at runtime. Moreover, _if_ we decide to add such support in the future (and that is a big "if"!), it is likely require a different base allocator API. Those changes may result in a very different AllocatorProxy class or no such class at all. In other words, that future/hypothetical switch is a red herring in this context.

> Removing AllocatorProxy would mean adding its member variables to every class type.

AFAICT, removing AllocatorProxy class will not add any data members to any other classes. The effectively unused class will be just gone. Here are its data members:

```C++
    const char *label; // the base class, MemAllocator has it
    size_t size; // MemImplementingAllocator (the only MemAllocator child) has it; others might even hard-code this
    mutable MemAllocator *theAllocator; // this is a "this" pointer in all other allocators
    bool doZero; // the base class, MemAllocator has it
```

> it prevents the pool being created just by fetching stats

It does not prevent creation of the AllocatorProxy object. If we make AllocatorProxy an Allocator child, like this PR attempts to do, then there will be essentially no difference (and there will be an added confusion of two kinds of pools, "real" and "reverse proxied" ones). It would be a step in the wrong direction, rather than an improvement.

yadij: > AFAICT, removing AllocatorProxy class will not add any data members to any other classes. The effectively unused class will be just gone. Here are its data members:
> 
> ```c++
>     const char *label; // the base class, MemAllocator has it
>     bool doZero; // the base class, MemAllocator has it

Remember that MemAllocator does not exist until the "late binding" call to Pool() occurs. These values need to be stored between compile time and runtime, with doZero being able to alter at runtime base don configuration.

>     size_t size; // MemImplementingAllocator (the only MemAllocator child) has it; others might even hard-code this

This is currently called obj_size in other classes. The MemImplementingAllocator cleanup left as followup to at least this PR should merge it up to MemAllocator and sync the two.

>     mutable MemAllocator *theAllocator; // this is a "this" pointer in all other allocators

... and for MEMPROXY classes would need to be stored in the allocated class type since it can change at reconfigure.

Only label and (obj_)size members can be passed to the pool create statement as constexpr. The others need to be stored as classes members of the MEMPROXY class. Currently they are all isolated into a single AllocatorProxy instance per-type. That is what I am referring to as extra overheads.

> 
> > it prevents the pool being created just by fetching stats
> 
> It does not prevent creation of the AllocatorProxy object. If we make AllocatorProxy an Allocator child, like this PR attempts to do, then there will be essentially no difference (and there will be an added confusion of two kinds of pools, "real" and "reverse proxied" ones). It would be a step in the wrong direction, rather than an improvement.

No, there is no step in *any* direction. That is the current status-quo situation. This PR just clarifies / reveals what the situation really is.

The design of memory pools (classified API anyway) should not be difficult to understand. FWIW, Here is the plan from that architecture you didn't want to be told about, (converted to the terms you have used above):

 * PoolX  are "real" allocators that maintain a local pool of free'd space to reuse.
  - currently Malloc and Chunked implemented.
  - TODO add pool for /dev/shm memory blocks to pass around with SMP.

 * AllocatorX are interfaces providing some optional functionality to the "real" allocator.
  - AllocatorProxy adds ability to change which "real" allocator / pool is used at runtime.
  - TODO add ability to switch between array of allocators for buffer (re)-alloc.

 * MemImplementingAllocator adds statistics gathering onto all child allocators.
  - misnamed, it should have been called MemAllocatorStats or similar
   - TODO merge memory accounting up to MemAllocator (always needed) or rename to AllocatorX convention

rousskov: TLDR: The official code supports my claims very well (and contradicts yours) AFAICT. If you would like, I can commit the first step towards AllocatorProxy removal in this PR branch or post a dedicated alternative PR.

----

> > Alex: AFAICT, removing AllocatorProxy class will not add any data members to any other classes. The effectively unused class will be just gone. Here are its data members:
```c++
    const char *label; // the base class, MemAllocator has it
    bool doZero; // the base class, MemAllocator has it
```
 
> Amos: Remember that MemAllocator does not exist until the "late binding" call to Pool() occurs. These values need to be stored between compile time and runtime, with doZero being able to alter at runtime based on configuration.

Official Squid code does not work that way. In AllocatorProxy context, all the MemAllocator construction parameter values, including doZero, are hard-coded. Here is the corresponding code snippets (comments are mine):

```C++
static inline Mem::AllocatorProxy &Pool() { \
    static Mem::AllocatorProxy thePool(#CLASS, sizeof(CLASS), false); \ // future parameters hard-coded here
    return thePool; \
} \
...
MemAllocator *
Mem::AllocatorProxy::getAllocator() const
{
    if (!theAllocator) {
        theAllocator = MemPools::GetInstance().create(objectType(), size); // previously-hard-coded parameters used here
        theAllocator->zeroBlocks(doZero);                                  // ... and here
    }
    return theAllocator;
}
```

No "late binding" delay can change hard-coded values, of course. That delay is a red herring in this part of the discussion.

```C++
size_t size; // MemImplementingAllocator (the only MemAllocator child) has it; others might even hard-code this
```
 
> Amos: This is currently called obj_size in other classes. The MemImplementingAllocator cleanup left as followup to at least this PR should merge it up to MemAllocator and sync the two sides of the MemAllocator hierarchy.

Regardless of what you plan to cleanup, the object size parameter is still going to be hard-coded to `sizeof(CLASS)` value in any AllocatorProxy context. In other words, that cleanup, whatever it is, will not affect the fact that AllocatorProxy does nothing useful but adds runtime and development-time overheads.

```C++
mutable MemAllocator *theAllocator; // this is a "this" pointer in all other allocators
```
 
> ... and for MEMPROXY classes would need to be stored in the allocated class type since it can change at reconfigure.

In the current official code, there is no need to store theAllocator (beyond MEMPROXY's Pool() where the allocator will be created instead of AllocatorProxy, of course)  because existing AllocatorProxy does not support switching allocators at runtime. Moreover, if we decide to add such support in the future (and that is a big "if"!), it is likely require a different base allocator API. Those changes may result in a very different AllocatorProxy class or no such class at all. In other words, that future/hypothetical switch is a red herring in this context.
 
> Only label and (obj_)size members can be passed to the pool create statement as constexpr. The others need to be stored as classes members of the MEMPROXY class. 

Please see above for specific code snippets that prove that all the variables necessary for allocator creation have hard-coded values in the official code.

> Currently they are all isolated into a single AllocatorProxy instance per-type.

That one-allocator-instance-per-type principle would still be true if we remove AllocatorProxy and let Pool() do what getAllocator() does. No additional storage is needed. We would be removing one extra pointer/indirection without adding any new members or storage.

> This PR just clarifies / reveals what the situation really is.

I disagree. This PR moves the existing problematic AllocatorProxy class into the Allocator hierarchy. We should completely remove AllocatorProxy instead (because we can, because that class does essentially nothing and adds no value). That removal would "properly clarify / reveal what the situation really is"!
 
> FWIW, Here is the plan from that architecture you didn't want to be told about
> * PoolX are "real" allocators that maintain a local pool of free'd space to reuse.
> * currently Malloc and Chunked implemented.
> * TODO add pool for /dev/shm memory blocks to pass around with SMP.
> * AllocatorX are interfaces providing some optional functionality to the "real" allocator.
> * AllocatorProxy adds ability to change which "real" allocator / pool is used at runtime.
> * TODO add ability to switch between array of allocators for buffer (re)-alloc.
> * MemImplementingAllocator adds statistics gathering onto all child allocators.
> * misnamed, it should have been called MemAllocatorStats or similar
> * TODO merge memory accounting up to MemAllocator (always needed) or rename to AllocatorX convention

FWIW, I disagree with many aspects of the above plan. I do not recommend implementing it. We can discuss the long-term future of allocation-related code, but this PR is probably not the best venue for that discussion. Hopefully, by the time simple cleanup steps like removal of AllocatorProxy are completed, it would be easier to reach consensus regarding that long-term future and ways to get there.

yadij:
In https://github.com/squid-cache/squid/pull/1181#discussion_r1028335044 you indicate indecision about the planned allocator for conversion of old memStrings/Buf/Rigid allocation API:

> > To remove the old C-style buffer/strings/rigid we will need a pool class that accepts multiple sizes of object and manages them.
> 
> I am not sure that is the right path forward, but we do not need to agree on that right now. If it turns out to be the right path, we can adjust the Allocator API accordingly, of course.

Obviously if we drop the AllocatorProxy class now it can no longer be extended and used for selection between buffer pools.

If you have an alternative design for buffer allocator that should be implemented instead, I am open to ideas. The primary consideration for this approach is/was that all key pieces of logic have existed for a long time already and been well tested separately in the existing APIs. Restricting the amount of risky new logic.

In absence of alternative implementation designs. This PR should be treated as incremental progress reducing divergence of the current allocator APIs.

rousskov:
> > > Amos at https://github.com/squid-cache/squid/pull/1181#discussion_r1027178937: To remove the old C-style buffer/strings/rigid we will need a pool class that accepts multiple sizes of object and manages them.
 
> > Alex at https://github.com/squid-cache/squid/pull/1181#discussion_r1028335044: I am not sure that is the right path forward, but we do not need to agree on that right now. If it turns out to be the right path, we can adjust the Allocator API accordingly, of course.
 
> Amos: Obviously if we drop the AllocatorProxy class now it can no longer be extended and used for selection between buffer pools.

Sounds good -- we should not extend a problematic class. Obviously, other PRs can add any class they need, including a class that helps managing raw buffers.
 
> If you have an alternative design for buffer allocator that should be implemented instead, I am open to ideas. 

It is impossible to offer an _alternative_ to the design one has not seen. If you insist on working on raw buffer management problems now, then I can contribute. Let's start with defining/detailing which raw buffers we want to manage better. That is, what does "old C-style buffer/strings/rigid" really mean? Are we looking for a modern alternative to memAllocString()? Something else? All this future work is outside this PR scope, but we can continue it somewhere else (e.g., on Zulip).

> The primary consideration for this approach is/was that all key pieces of logic have existed for a long time already and been well tested separately in the existing APIs. Restricting the amount of risky new logic.

We should reuse good code, of course, but AllocatorProxy is not that kind of code.

> In absence of alternative implementation designs. This PR should be treated as incremental progress reducing divergence of the current allocator APIs.

The correct design alternative, in this PR scope, is AllocatorProxy removal. Moving a problematic and effectively unused AllocatorProxy class into the being-polished Allocator hierarchy is the opposite of progress. My offer to make (at least) the first step towards AllocatorProxy removal still stands.

kinkie:
@yadij , should we abandon this effort for now? It's grown stale by now, maybe there's a different approach to it?

PR #1184, opened 2022-11-20
Remove Mem::Report() and polish memory_pools_limit storage

yadij:
The official code should not call Mem::Report(). There is no
justification for main.cc treating memory module in such a
special way. The reporting code itself has problems.
Removing all this code is better than leaving or polishing it.

### DIFF
--- a/src/mem/Pool.h
+++ b/src/mem/Pool.h
@@ -131,9 +131,9 @@ public:
      * not strict upper limit, but a hint. When MemPools are over this limit,
      * deallocate attempts to release memory to the system instead of pooling.
      */
-    void setIdleLimit(ssize_t newLimit) {poolIdleLimit = newLimit;}
-    /// \see void setIdleLimit(ssize_t)
-    ssize_t idleLimit() const {return poolIdleLimit;}
+    void setIdleLimit(const ssize_t newLimit) { idleLimit_ = newLimit; }
+    /// \copydoc idleLimit_
+    ssize_t idleLimit() const { return idleLimit_; }
 
     /**
      \par
@@ -170,10 +170,10 @@ public:
     bool defaultIsChunked = false;
 
 private:
-    /// Limit for allocated (but unused) memory in memory pools.
+    /// Limits the cumulative size of allocated (but unused) memory in all pools.
     /// Initial value is 2MB until first configuration,
-    /// \see squid.conf memory_pools_limit directive.
-    ssize_t poolIdleLimit = (2 << 20);
+    /// See squid.conf memory_pools_limit directive.
+    ssize_t idleLimit_ = (2 << 20);
 };
 
 /**
--- a/src/mem/Pool.h
+++ b/src/mem/Pool.h
@@ -129,11 +129,11 @@ public:
     /**
      * Sets upper limit in bytes to amount of free ram kept in pools. This is
      * not strict upper limit, but a hint. When MemPools are over this limit,
-     * totally free chunks are immediately considered for release. Otherwise
-     * only chunks that have not been referenced for a long time are checked.
+     * deallocate attempts to release memory to the system instead of pooling.
      */
-    void setIdleLimit(ssize_t new_idle_limit);
-    ssize_t idleLimit() const;
+    void setIdleLimit(ssize_t newLimit) {poolIdleLimit = newLimit;}
+    /// \see void setIdleLimit(ssize_t)
+    ssize_t idleLimit() const {return poolIdleLimit;}
 
     /**
      \par
@@ -166,9 +166,14 @@ public:
     void setDefaultPoolChunking(bool const &);
 
     MemImplementingAllocator *pools = nullptr;
-    ssize_t mem_idle_limit = (2 << 20) /* 2MB */;
     int poolCount = 0;
     bool defaultIsChunked = false;
+
+private:
+    /// Limit for allocated (but unused) memory in memory pools.
+    /// Initial value is 2MB until first configuration,
+    /// \see squid.conf memory_pools_limit directive.
+    ssize_t poolIdleLimit = (2 << 20);
 };
 
 /**
--- a/src/mem/old_api.cc
+++ b/src/mem/old_api.cc
@@ -10,7 +10,6 @@
 
 #include "squid.h"
 #include "base/PackableStream.h"
-#include "base/RunnersRegistry.h"
 #include "ClientInfo.h"
 #include "dlink.h"
 #include "event.h"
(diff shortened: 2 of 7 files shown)


kinkie:
@yadij , could you run a test of the change?

yadij:
Done before submission. Are you seeing issues?

src/mem/old_api.cc
```diff
@@ -456,15 +457,29 @@ Mem::Init(void)
...
+{
+
+static void
+LogPoolSummary()
 {
     debugs(13, 3, "Memory pools are '" <<
            (Config.onoff.mem_pools ? "on" : "off")  << "'; limit: " <<
            std::setprecision(3) << toMB(MemPools::GetInstance().idleLimit()) <<
            " MB");
 }
```
rousskov, line 471: Please completely remove Mem::Report() instead. 

The official code should not call Mem::Report(). There is no justification for main.cc treating memory module in such a special way. The PR code abuses finalizeConfig() and syncConfig() APIs for reporting configuration details. The reporting code itself has problems. Removing all this code is better than leaving or polishing it.

These level-3 details are meant for developers. Developers can use the existing configuration debugging to extract very similar configuration details. If somebody actually needs more debugging, let's add it to MemPools::setIdleLimit() and/or other configuration methods, where it belongs.

rousskov (changes requested):

src/mem/old_api.cc
yadij: FTR; a lot of non-developers use levels 3-9 for behaviour analysis, flow monitoring and other use-cases.

Anyhow, testing complete removal and the associated cleanup that allows now. Will re-focus this PR when that passes.

yadij: Done as of [d8b4186](https://github.com/squid-cache/squid/pull/1184/commits/d8b4186f3fcd8f033ea9ceadbf5fcd6115f9e7cf)

src/mem/Pool.h
```diff
@@ -166,9 +166,14 @@ class MemPools
...
 
     MemImplementingAllocator *pools = nullptr;
-    ssize_t mem_idle_limit = (2 << 20) /* 2MB */;
     int poolCount = 0;
     bool defaultIsChunked = false;
+
+private:
+    /// Limit for allocated (but unused) memory in memory pools.
+    /// Initial value is 2MB until first configuration,
+    /// \see squid.conf memory_pools_limit directive.
```
rousskov, line 175: I hope this `\see` reference emits Doxygen errors or warnings. Let's avoid Doxygen commands with parameters that Doxygen cannot grok:

```suggestion
    /// See squid.conf memory_pools_limit directive.
```

src/mem/Pool.h
```diff
@@ -166,9 +166,14 @@ class MemPools
...
+    ssize_t poolIdleLimit = (2 << 20);
```
rousskov, line 176: Please avoid fooX inside Foo:

```suggestion
    ssize_t idleLimit_ = (2 << 20);
```

src/mem/Pool.h
```diff
@@ -166,9 +166,14 @@ class MemPools
     void setDefaultPoolChunking(bool const &);
 
     MemImplementingAllocator *pools = nullptr;
-    ssize_t mem_idle_limit = (2 << 20) /* 2MB */;
     int poolCount = 0;
     bool defaultIsChunked = false;
+
+private:
+    /// Limit for allocated (but unused) memory in memory pools.
```
rousskov, line 173: If we are going to polish all this, let's clarify that this limit is about bytes rather than object count:

```suggestion
    /// Limits the cumulative size of allocated (but unused) memory in all pools.
```

src/mem/Pool.h
```diff
@@ -129,11 +129,11 @@ class MemPools
...
      * not strict upper limit, but a hint. When MemPools are over this limit,
-     * totally free chunks are immediately considered for release. Otherwise
-     * only chunks that have not been referenced for a long time are checked.
+     * deallocate attempts to release memory to the system instead of pooling.
      */
-    void setIdleLimit(ssize_t new_idle_limit);
-    ssize_t idleLimit() const;
+    void setIdleLimit(ssize_t newLimit) {poolIdleLimit = newLimit;}
+    /// \see void setIdleLimit(ssize_t)
+    ssize_t idleLimit() const {return poolIdleLimit;}
```
rousskov, line 136: Just minor polishing suggestions (that already account for the suggested renaming of the data member):

```suggestion
    void setIdleLimit(const ssize_t newLimit) { idleLimit_ = newLimit; }
    /// \copydoc idleLimit_
    ssize_t idleLimit() const { return idleLimit_; }
```

rousskov (commented):
Thank you for removing Mem::Report(). The memory_pools_limit-related polishing _looks_ rather out of scope, but I suspect it was actually done to remove a libmem dependency. If my suspicion is right, then perhaps we should rename the PR accordingly: "Reduce dependence on libmem" or something like that.

I adjusted the PR title/description. I hope those adjustments were not controversial.

FWIW, I removed my name from the PR description primarily because the (future) commit message ought to represent _consensus_. It does not matter who came up with the justification for the change. If we agree that removing this code is the right thing to do, then we should remove it, for the reasons we can all agree on.

yadij:
No objections to your changes. I left your name there for the credits with the design change this is now as much yours as mine.

The privatization of idleLimit_ was not for a specific issue, just coding "best practice" by me. If you insist on some specific Mem::Report scope definition the extra polish can be unwound. Or we can call the scope "incremental cleanup" and move on to other PRs.

rousskov (approved):
> I left your name there for the credits with the design change this is now as much yours as mine.

FWIW, in similar situations, I recommend using semi-standard `Co-authored-by` footer for such crediting. It does not have the problems associated with prefixing partial change justification with a name of a person.

To avoid misunderstanding, I do not want any credits for this trivial change!
 
> If you insist on some specific Mem::Report scope definition the extra polish can be unwound. Or we can call the scope "incremental cleanup" and move on to other PRs.

After the latest changes, I am OK with this PR going in as it is.

rousskov:
> Francesco: @yadij, could you run a test of the change?

> Amos: Done before submission. Are you seeing issues?

@kinkie, last chance to stop this from going in! If you know of any problems with the current PR state, please request your review (at least) now to stop this PR from merging as is.

src/mem/old_api.cc
rousskov: > FTR; a lot of non-developers use levels 3-9 for behaviour analysis, flow monitoring and other use-cases.

Just like developers, those folks can use the existing configuration debugging to extract very similar configuration details. If somebody actually needs more debugging, let's add it to MemPools::setIdleLimit() and/or other configuration methods, where it belongs.

Said that, I think we should not go the extra mile to accommodate "non-developer" needs in level-3+ debugging logs. To improve debugging interface and its performance, we should focus on satisfying "developer" needs instead. Otherwise, nobody will be happy, neither "developers" nor "non-developers" that read level-3+ debugging logs.

PR #1185, opened 2022-11-21
Do not cache (and do not serve cached) cache manager responses

rousskov:
The fixed bug affected cache manager transactions that were using
/squid-internal-mgr URL path prefix with http(s) URL scheme. It did not
affect transactions that were using legacy cache_object URL scheme.

Stale cache manager responses had their Age response header set to the
number of seconds since Unix epoch. If disk and memory caches were
disabled, then cache manager requests just triggered "found KEY_PRIVATE"
WARNINGs in cache.log (for reasons that remain unclear).

Probably broken since 2011 commit e37bd29 that did not expand
HttpRequest::maybeCacheable() (called cacheable() back then)
PROTO_CACHE_OBJECT check to include /squid-internal-mgr requests.

Also added missing Access-Control-* response headers to cache manager
responses in SMP mode and reduced code duplication related to sending
those headers (which led to them missing in SMP Squids).


yadij (dismissed):
The bug cited as justification for this would be far better resolved by simply making the cache manager generate the response with appropriate Cache-Control:no-store header. That will allow us in future to fix the cases where it should be some other vale - some only need a must-revalidate, some need private, and some should in future be resumable from cache via ETag.

The secondary issue of receiving traffic from outdated Squid with the above bug should only need a much more limited hack on the server-facing side to make manager responses without Cache-Control implicitly have "no-store" set. Or maybe even just document a suggestion to admin for store_miss ACL rules to workaround it.

src/mgr/Inquirer.cc
```diff
@@ -73,17 +74,22 @@ Mgr::Inquirer::start()
...
         const auto mx = MasterXaction::MakePortless<XactionInitiator::initIpc>();
         auto *req = HttpRequest::FromUrlXXX(url, mx);
         ErrorState err(ERR_INVALID_URL, Http::scNotFound, req, nullptr);
         std::unique_ptr<HttpReply> reply(err.BuildHttpReply());
+        CacheManager::PutCommonResponseHeaders(*reply, originOrNil);
         replyBuf.reset(reply->pack());
     } else {
         std::unique_ptr<HttpReply> reply(new HttpReply);
         reply->setHeaders(Http::scOkay, nullptr, "text/plain", -1, squid_curtime, squid_curtime);
+        CacheManager::PutCommonResponseHeaders(*reply, originOrNil);
```
rousskov, line 92: Note that these Inquerer.cc changes added missing Access-Control-* response headers to cache manager responses in SMP mode.

rousskov:
> @[yadij](https://github.com/yadij) yadij dismissed their [stale review](https://github.com/squid-cache/squid/pull/1185#pullrequestreview-1187630611) 3 days ago

Just to avoid misunderstanding, please note that this PR is still blocked on your review -- dismissing the old review does not (and should not) automatically dismiss the outstanding review _request_. You can do the latter manually in the Reviewers box on the PR page (and adjust the PR labels accordingly).

src/CacheManager.h
```diff
@@ -49,6 +50,11 @@ class CacheManager
     static CacheManager* GetInstance();
     const char *ActionProtection(const Mgr::ActionProfilePointer &profile);
 
+    /// Add HTTP response headers specific/common to all cache manager replies,
+    /// including cache manager errors and Action reports.
+    /// \param httpOrigin the value of Origin header in the trigger HTTP request (or nil)
+    static void PutCommonResponseHeaders(HttpReply &, const char *httpOrigin);
+
```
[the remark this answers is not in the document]
yadij: > IMO, the request-time changes in the PR are valuable on their own and should stay because
> 
>     * The sooner Squid knows that the response is going to be uncachable, the better.

mgr responses being non-cacheable is not guaranteed. As I mentioned earlier there are multiple cases where the response SHOULD be cached. All these request-time changes make it much harder to correct those responses caching in future.

> This knowledge helps reduce cache manager request impact on Store state, which may be very valuable in triage and tests. 

I disagree on the utility. It is certain to add one more case to the existing difficulty troubleshooting why apparently cacheable mgr responses from upstream servers are not being stored. Whereas it is easy to tell from upstream headers when a response is incorrectly made cacheable by them.

> Most clients sending cache manager requests do not tell Squid that the responses are not cachable,

Irrelevant. Cacheability in HTTP is determined by the server. At most a client can request a cache be bypassed and/or refreshed.

>     * Those request-time changes fix a known problem with flags.cachable getting overwritten by maybeCacheable(), reduce code duplication of tricky code, improve debugging, improve code documentation, and make Optional trivially copyable. They are a step forward on many levels, and they are not complex.
> 

Maybe, BUT, none of that extra change is actually fixing the cache manager response bug. It would make review and change documentation a lot better if each of those issue fixes were isolated in their own PRs, or at least PRs scoped more tightly to each particular problematic object.

> > Or maybe even just document a suggestion to admin for store_miss ACL rules to workaround it.
> 
> I agree that warning admins that use affected cache manager requests and sharing this workaround is a good idea.
> Once this PR lands, I can volunteer to post a message to squid-users about it if you think that is the best way to
> document that suggestion.

A post to squid-announce is probably better. That would go to squid-users plus a bunch of others.

For backport I would like the change of generated headers, but the rest is unclear at present.

src/RequestFlags.h
```diff
@@ -115,6 +125,11 @@ class RequestFlags
      */
     RequestFlags cloneAdaptationImmune() const;
 
+    /// Convenience wrapper for checking whether the response may be cached. By
+    /// default (i.e. unless we explicitly allow caching), the response is not
+    /// cachable. \sa noCache
+    bool cachable() const { return missCachingDecision.value_or(false); }
```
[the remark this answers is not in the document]
yadij: FTR; RequestFlags::noCache should be the value of Cache-Control:no-cache received in request headers. Please remember that in any renaming considerations.

src/client_side_reply.cc
```diff
@@ -1528,6 +1528,9 @@ clientReplyContext::identifyStoreObject()
 
     // client sent CC:no-cache or some other condition has been
     // encountered which prevents delivering a public/cached object.
+    // XXX: The above text does not match the condition below. It might describe
```
yadij, line 1531: IMO this can and should be fixed by this PR. Reversing the if-else like below should be correct.
```
@@ -1528,6 +1528,9 @@ clientReplyContext::identifyStoreObject()

    if (r->flags.noCache) {
        // client sent CC:no-cache which prevents delivering a public/cached object.
        identifyFoundObject(nullptr, "no-cache");
    } else {
        const auto e = storeGetPublicByRequest(r);
        identifyFoundObject(e, storeLookupString(bool(e)));
    }
```

src/client_side_request.cc
```diff
@@ -1095,15 +1098,19 @@ clientInterpretRequestHeaders(ClientHttpRequest * http)
 
 #endif
 
-    request->flags.cachable = http->request->maybeCacheable();
+    // Make (or update a previously made positive) flags.missCachingDecision.
+    // TODO: Integrate this and most of this function logic with request parsing
+    // code, to avoid direct manipulation of low-level flags in high-level code.
+    if (request->flags.missCachingDecision.value_or(true))
+        request->flags.missCachingDecision = http->request->maybeCacheable();
```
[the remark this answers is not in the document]
yadij: This seems to be a bug fix all of its own, slightly tangled up with the issue of what correct default value for "flags.cacheable" should actually be. I suggest that this part of that pile of issues also should not be fixed here.

src/RequestFlags.cc
```diff
@@ -23,3 +26,11 @@ RequestFlags::cloneAdaptationImmune() const
     return *this;
 }
 
+void
+RequestFlags::disableCacheUse(const char * const reason)
+{
+    debugs(16, 3, "for " << reason);
+    missCachingDecision = false; // may already be set to false
+    noCache = true; // may already be true
```
yadij, line 34: Setting these flags to these values is always valid and previous state does not matter.
Only code setting the inverse values away from these needs the attention a comment brings.
```suggestion
    missCachingDecision = false;
    noCache = true;
```

src/internal.cc
```diff
@@ -32,12 +33,21 @@ void
...
+    Assure(request);
     const SBuf upath = request->url.path();
     debugs(76, 3, clientConn << " requesting '" << upath << "'");
 
+    Assure(request->flags.internal);
+
+    if (ForSomeCacheManager(upath)) {
+        debugs(17, 2, "calling CacheManager");
+        CacheManager::GetInstance()->start(clientConn, request, entry, ale);
+        return;
```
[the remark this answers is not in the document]
yadij: Please return. These lines were ordered by frequency of request type to optimize performance despite the string comparisons. mgr URLs are quite rare in reality compared to the others.

src/internal.h
```diff
@@ -29,5 +29,9 @@ char *internalRemoteUri(bool, const char *, unsigned short, const char *, const
 const char *internalHostname(void);
 int internalHostnameIs(const char *);
 
+/// whether the given request URL path points to a cache manager (not
+/// necessarily running on this Squid instance)
+bool ForSomeCacheManager(const SBuf &);
```
yadij, line 34: This might be better placed in the class CacheManager API.  I do not insist on this.

src/peer_digest.cc
```diff
@@ -332,7 +332,7 @@ peerDigestRequest(PeerDigest * pd)
     /* update timestamps */
     pd->times.requested = squid_curtime;
     pd_last_req_time = squid_curtime;
-    req->flags.cachable = true;
+    req->flags.missCachingDecision = true;
```
yadij, line 335: Setting this flag true is always a risky operation worth documenting.
```suggestion
    // XXX flag currently defaults to false. It should be true to allow collapsing and re-sharing of digests while they are fresh.
    req->flags.missCachingDecision = true;
```

src/store_digest.cc
```diff
@@ -421,7 +421,7 @@ storeDigestRewriteStart(void *)
     auto req = HttpRequest::FromUrlXXX(url, mx);
 
     RequestFlags flags;
-    flags.cachable = true;
+    flags.missCachingDecision = true;
```
yadij, line 424: Setting this flag to true is a risky operation worth documenting why it needs to be true
```suggestion
    // XXX flag currently defaults to false. It should be true for ....
    flags.missCachingDecision = true; 
```

yadij (changes requested):

src/base/Optional.h
```diff
@@ -122,6 +122,53 @@ class Optional
     bool hasValue_ = false;
 };
 
+/// Specialization to make Optional<bool> trivially-copyable. XXX: Keep this
+/// temporary (until we switch to C++17 std::optional) hack in sync with the
+/// generic Optional above, copying generic methods where possible.
+template <>
+class Optional<bool>
```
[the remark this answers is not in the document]
rousskov: Done in #1188.

src/RequestFlags.h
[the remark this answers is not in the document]
rousskov: > IMO this type change should be a separate PR

Done in #1189.

src/internal.cc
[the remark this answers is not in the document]
rousskov: Done in de9d0cf to avoid arguing.

src/CacheManager.h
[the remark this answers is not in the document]
rousskov: > Alex: Once this PR lands, I can volunteer to post a message to squid-users about it if you think that is the best way to document that suggestion.

> Amos: A post to squid-announce is probably better. That would go to squid-users plus a bunch of others.

Sounds good, will do.

src/internal.h
rousskov: Yeah, it is difficult to say for sure. I think this is the best location _for now_ because the function is used, in part, by the code that does not need to know any of the cache manager APIs. This is also where many other similar functions are already located. This function _implementation_ also benefits from being close to its sister functions -- it is easier to keep that code in sync that way.

We can and perhaps should refactor this code to allow individual test functions to be separated into their own modules, but it is best done in a dedicated PR, affecting all of these functions. I would not be surprised if proper refactoring will completely remove this pair of files, but all that is out of scope here.

rousskov (commented):
I believe I have addressed all the change requests. The branch code now relies on the recently merged PRs extracted from this PR earlier. No changes to the primary fixes. @yadij, please re-review.

rousskov:
@yadij, just a gentle reminder: This `review-1` PR has been waiting for your review for two weeks.

yadij (commented):
Just the ongoing discussion about additional code to add/change. The rest of it looks okay.

src/client_side_reply.cc
rousskov: > the titular scope/purpose of this PR is to make Squid obey CC:no-cache on manager requests.

I see no such scope/purpose or even a hint at such scope/purpose! This PR is about making new cache manager responses uncachable and preventing previously cached responses from being served from the cache. That is all. This PR changes do not (and should not) care about Cache-Control headers in cache manager requests! What can I change in PR description to clarify that?

rousskov: @yadij, please move this bug-fixing PR forward. The current basic disagreement related to this added `XXX` comment should not take weeks to resolve, but it has been more than two weeks since your last response. Again, if it helps, I will remove that `XXX` comment.

rousskov: @yadij, please move this bug-fixing `review-1` PR forward. It has been waiting for your review for a month already.

rousskov: @yadij, please move this bug-fixing `review-1` PR forward. It has been waiting for your review for two months already.

yadij: Issue worked around by [c380732](https://github.com/squid-cache/squid/pull/1185/commits/c38073237e29e0bafb3520c418da57625ad9c5ac).

yadij (approved):

PR #1186, opened 2022-11-22
Refactor memory pools statistics gathering

yadij:


src/mem/Pool.h
```diff
@@ -190,7 +186,7 @@ class MemAllocator
      \param stats   Object to be filled with statistical data about pool.
      \retval        Number of objects in use, ie. allocated.
      */
-    virtual int getStats(MemPoolStats * stats, int accumulate = 0) = 0;
+    virtual int getStats(MemPoolStats * stats) = 0;
```
rousskov, line 189: Since we are going to change every method, please also fix the type of the first parameter. If it is not going to trigger lots of changes, please fix the return type as well:

```suggestion
    virtual size_t getStats(MemPoolStats &stats) = 0;
```

If you want to avoid adding many s/pointer/reference/ changes to the method implementations while this PR is getting reviewed, add temporary diff reducers to those implementations. For example:

```C++
    const auto stats = &statsReference; // TODO: Remove this diff reducer before merging
```

src/mem/PoolChunked.h
```diff
@@ -31,7 +31,7 @@ class MemPoolChunked : public MemImplementingAllocator
      \param stats   Object to be filled with statistical data about pool.
      \retval        Number of objects in use, ie. allocated.
      */
-    virtual int getStats(MemPoolStats * stats, int accumulate);
```
rousskov, line 34: Here and in similar places, please drop the misplaced/duplicated/incomplete method description or replace it with a `/* Allocator API */` comment.

src/mem/Pool.h
```diff
@@ -352,15 +325,13 @@ extern void memPoolIterateDone(MemPoolIterator ** iter);
...
+ * Fills a MemPoolStats with statistical data about overall
  * usage for all pools.
  *
- * \param stats   Object to be filled with statistical data.
- *
  * \return Number of pools that have at least one object in use.
  *        Ie. number of dirty pools.
  */
-extern int memPoolGetGlobalStats(MemPoolGlobalStats * stats);
+extern size_t memPoolGetGlobalStats(MemPoolStats &);
```
rousskov, line 334: Please move this function into the Mem namespace.

src/mem/Pool.h
```diff
@@ -284,45 +279,23 @@ class MemImplementingAllocator : public MemAllocator
 class MemPoolStats
```
rousskov, line 279: The PR probably touches most of the lines that mention this class by name already. If so, please move this class into the Mem namespace.

rousskov (commented):
Thank you for working on this cleanup.

src/mem/PoolChunked.h
yadij: To keep this refactor small I have deferred these cleanups to PRs updating the other MemPoolX classes.

src/mem/Pool.h
yadij: Done in [b06b4d9](https://github.com/squid-cache/squid/pull/1186/commits/b06b4d9064ecd1d9988818ada2e4fff970161701)

src/mem/Pool.h
yadij: Done in [8250bc6](https://github.com/squid-cache/squid/pull/1186/commits/8250bc6eeb9ee0263caa9e5937dea7f02b5d5363)

src/mem/Pool.h
yadij: Done in [fdf49db](https://github.com/squid-cache/squid/pull/1186/commits/fdf49db060d1b6529ac0a48a4925768662c339d3) and [1d1b2aa](https://github.com/squid-cache/squid/pull/1186/commits/1d1b2aa7f65c0bf400f4c44278b23b31bbcb0ac1)

src/mem/Stats.cc
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+#include "squid.h"
+#include "mem/Pool.h"
+#include "mem/Stats.h"
+
+size_t
+Mem::GlobalStats(PoolStats &stats)
+{
+    MemPools::GetInstance().flushMeters(); /* recreate TheMeter */
```
rousskov, line 16: This copied comment incorrectly (and unnecessary) documents what flushMeters() does.

```suggestion
    MemPools::GetInstance().flushMeters();
```

src/stat.cc
```diff
@@ -724,8 +724,8 @@ DumpInfo(Mgr::InfoActionData& stats, StoreEntry* sentry)
     storeAppendPrintf(sentry, "\tTotal accounted:       %6.0f KB\n",
                       stats.total_accounted / 1024);
     {
-        MemPoolGlobalStats mp_stats;
-        memPoolGetGlobalStats(&mp_stats);
+        Mem::PoolStats mp_stats;
+        Mem::GlobalStats(mp_stats); // XXX: pointless lookup ?
```
rousskov, line 728:
```suggestion
        Mem::GlobalStats(mp_stats); // XXX: pointless lookup
```

src/mem/Pool.h
```diff
@@ -85,10 +74,10 @@ class MemPools
     void flushMeters();
 
     /**
-     \param label   Name for the pool. Displayed in stats.
-     \param obj_size    Size of elements in MemPool.
+     * Create an allocator with given name to allocate objects
+     * of the specified fixed-size.
```
rousskov, line 78:
```suggestion
     * Create an allocator with given name to allocate fixed-size objects
     * of the specified size.
```

or simply

```suggestion
     * Create an allocator with given name to allocate objects
     * of the specified size.
```

src/stat.cc
```diff
@@ -1917,10 +1917,3 @@ statGraphDump(StoreEntry * e)
 }
 
 #endif /* STAT_GRAPHS */
```
rousskov, line 1919: PR removes a customary empty line from the end of this file. If possible, please format changed sources before merging.

```suggestion
#endif /* STAT_GRAPHS */

```

rousskov (approved):
I left a few polishing suggestions, but they are all optional, of course.

rousskov:
Disclaimer: I have not verified most of the numerous changes in this PR.

src/mem/Pool.cc
```diff
@@ -77,11 +47,17 @@ MemPools::MemPools()
...
+    // For now leak these Allocator descendants on shutdown.
+
+    MemImplementingAllocator *newPool;
     if (defaultIsChunked)
-        return new MemPoolChunked (label, obj_size);
+        newPool = new MemPoolChunked(label, obj_size);
     else
-        return new MemPoolMalloc (label, obj_size);
+        newPool = new MemPoolMalloc(label, obj_size);
+    pools.emplace_back(newPool);
```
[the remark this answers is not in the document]
yadij: Several things here.

* With pools being "late binding" this **is** (indirectly) performance sensitive code. Any call to new for a previously unused object type may initiate pool creation during performance sensitive logic.

* raw-pointers do not have destructors called on scope exit - they "leak". So the implicit vs explicit destructor dangers do not exist here. These pool pointers are about to become ref-counted Pointer (followup PR) which implement move semantics to remain emplace-safe[1]. Also strict move safety is also why pool.back() gets returned instead of these newPool pointers - which may be nullptr (or any undefined state) after the emplace.

[1] FYI, the type of side effects emplace have on move is part of why move semantics require (MUST) that the moved-from object be left in a valid null state by explicit move operators/constructors.

src/mem/Stats.cc
yadij: FWIW, I do not think that was supposed to be documenting what flushMeters does, but what the GlobalStats() activity needed doing.

src/stat.cc
yadij: Are you sure of that being pointless (eg no side effects of rebuilding TheMeter or such)?

If yes, then I will drop the lookup instead of leaving the XXX question.

rousskov: * The mp_stats lookup itself (i.e. filling mp_stats with values) is pointless because mp_stats is not used.
* Mem::GlobalStats() has side effects; those side effects might be important in this DumpInfo() context. They should be removed, but I have not studied all the details to suggest a specific fix.
* DumpInfo() should not have side effects like those injected by Mem::GlobalStats(), but something else may need to "flush" statistics before or after DumpInfo() is called (in some cases). I did not study all the necessary details to suggest a specific fix.

Due to complications in the second and third bullets, I do not recommend removing this call in this PR. Please adjust the XXX text if you think the current version needs adjustment (or leave it as is if you think it does not).

src/mem/Pool.cc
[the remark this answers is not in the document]
rousskov: > With pools being "late binding" this is (indirectly) performance sensitive code. Any call to new for a previously unused object type may initiate pool creation during performance sensitive logic.

Code that quickly executes just a few times per Squid instance lifetime is not performance sensitive code. MemPools::create() is no exception to that rule.

> implicit vs explicit destructor dangers do not exist here

The dangers are related to the construction, not destruction. Those dangers are not bugs in the current code. They are bugs that may happen when this code is improved (e.g., when raw pointers become smart). Such bugs have happened to good developers improving very similar code (as detailed in my change request).

We should not be spending time on discussing complicated performance optimizations in some hypothetical future code in the context of a once-in-a-lifetime pool creation code. We should just use the safer push_back() call. In the unlikely event the actual future code needs a different solution, that different solution will be evaluated then, when all the variables are known.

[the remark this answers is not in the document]
rousskov: > Amos: move semantics require (MUST) that the moved-from object be left in a valid null state by explicit move operators/constructors.

There is no "valid null state" requirement for explicit move methods. If you disagree, please provide a reference.

There is a similar-sounding, but different [design choice](https://en.cppreference.com/w/cpp/utility/move) for STL objects: 

> STL: Unless otherwise specified, all standard library objects that have been moved from are placed in a "valid but unspecified state", meaning the object's class invariants hold (so functions without preconditions, such as the assignment operator, can be safely used on the object after it was moved from).

There is also a similar-sounding description of a [typical implementation](https://en.cppreference.com/w/cpp/language/move_constructor) of a move constructor:

> C++: Move constructors typically "steal" the resources held by the argument rather than make copies of them, and leave the argument in some valid but otherwise indeterminate state.

Same for move assignment operators.

P.S. This part of the discussion has nothing to do with the code in question and actual emplacement dangers. I am only trying to correct this particular misunderstanding because it is likely to bite future PRs that attempt to implement move optimizations in other contexts. Statistically speaking, our current expected probability of a move optimization-related proposal being bug-free is close to zero.

src/snmp_agent.cc
```diff
@@ -339,11 +341,14 @@ snmp_prfSysFn(variable_list * Var, snint * ErrP)
...
+    case PERF_SYS_MEMUSAGE: {
+        Mem::PoolStats stats;
+        Mem::GlobalStats(stats);
         Answer = snmp_var_new_integer(Var->name, Var->name_length,
-                                      (snint) statMemoryAccounted() >> 10,
+                                      (snint) stats.meter->alloc.currentLevel() >> 10,
                                       ASN_INTEGER);
-        break;
+    }
+    break;
```
rousskov, line 351: The original location/indentation of the `break` statement looks more logical/natural to me:

```suggestion
        break;
    }
```

yadij: The weird indentation is an astyle thing. I am reluctant to move the break statement into sub-scope as that implies situations where the switch-case falls through.

rousskov: > [moving] the break statement into sub-scope ... implies situations where the switch-case falls through.

There is no such implication IMO: The curly braces factually and visually clearly apply to the entire `case` statement (as they should).

src/mem/Pool.cc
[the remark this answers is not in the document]
yadij: > > With pools being "late binding" this is (indirectly) performance sensitive code. Any call to new for a previously unused object type may initiate pool creation during performance sensitive logic.
> 
> Code that quickly executes just a few times per Squid instance lifetime is not performance sensitive code. MemPools::create() is no exception to that rule.
> 

FWIW, that "rule" seems to be vague and does not have reliable scope for what performance impact "quickly" and "just a few times" are having.

FYI, The definition I use for performance sensitive (code operating during action X, ie. from point A to point B in code) can be deterministically defined in terms of which specific lines of code are relevant. Allowing policy defined groups of actions with different amount of acceptable impact or size/speed optimization targets.


> > implicit vs explicit destructor dangers do not exist here
> 
> The dangers are related to the construction, not destruction.

Then I have completely missed what your point was. The references you provided as examples are all about bugs caused by code calling destructors twice (on moved and moved-from objects) after emplace() calls a badly designed move-constructor.
 These are expected "undefined behaviour" effects of buggy class X being passed around. Not dangers of using emplace general. eg. the same bugs would occur if instances of such class X are returned by-value from any custom code.


> > Amos: move semantics require (MUST) that the moved-from object be left in a valid null state by explicit move operators/constructors.
> 
> There is no "valid null state" requirement for explicit move methods. If you disagree, please provide a reference.
> 

C++11 document N337 section 12.2.1 on xvalue (temporary object) states.
"
Even when the creation of the temporary object is unevaluated (Clause 5) or otherwise avoided (12.8 [copy/move construction]), all the semantic restrictions shall be respected as if the temporary object had been created and later destroyed.
"

The relevant "semantic restrictions" for emplace_back() being the MoveConstructible trait defined in section 17.6.3.1 table 20:
"
 Note:rv must still meet the requirements of the library component that is using it. The operations listed in those requirements must work as specified whether rv has been moved from or not.
"
 [ ```rv``` being the moved-from object]

Thus the value "shall be" left valid as per default-initialization (ergo: nullptr).

src/snmp_agent.cc
yadij:
```c++
case W: {...}
case X:
 {...}
case Y:
  break;
```

Which **one** falls through and which **two** do not?

rousskov:
```c++
case W: {...}
case X:
 {...}
case Y:
  break;
```
 
> Which **one** falls through and which **two** do not?

In a correctly structured code, like the one this change request is about, none fall through. Such code, including code this change request is about, marks fall through cases explicitly and does not position curly braces poorly (and inconsistently).

In some hypothetical poorly written code, like the one sketched in your example, it does not really matter which ones fall through. The code needs to be polished before it makes sense to discuss such style issues. We should not even discuss code like that in this change request.

rousskov (approved):
Thank you for addressing my primary concern.

src/mem/Pool.cc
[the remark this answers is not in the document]
rousskov: > > > Amos: With pools being "late binding" this is (indirectly) performance sensitive code. Any call to new for a previously unused object type may initiate pool creation during performance sensitive logic.
 
> > Alex: Code that quickly executes just a few times per Squid instance lifetime is not performance sensitive code. MemPools::create() is no exception to that rule.
 
> Amos: FWIW, that "rule" seems to be vague and does not have reliable scope for what performance impact "quickly" and "just a few times" are having.

Agreed. Fortunately, any reasonable definition of "quickly" and "few" would exclude MemPools::create() from being classified as performance-sensitive in this Squid-specific context, and that is all that matters as far as this change request is concerned. In other words, the proposed rule of thumb is good enough to easily make the correct decision here (and in the vast majority of real cases). That is what rules of thumb are best for -- quickly and correctly handling the vast majority of use cases.

 
> FYI, The definition I use for performance sensitive (code operating during action X, ie. from point A to point B in code) can be deterministically defined in terms of which specific lines of code are relevant. Allowing policy defined groups of actions with different amount of acceptable impact or size/speed optimization targets.

I do not know what definition of "performance sensitive" you are using. If you would like to establish some formal terminology for future shared use, please make a specific proposal. FWIW, I am sure that any successful outcome of that activity will _not_ classify MemPools::create() (in the current or any foreseeable Squid code context) as "performance sensitive", but I am prepared to be proven wrong.

> > > implicit vs explicit destructor dangers do not exist here

> > The dangers are related to the construction, not destruction.

> The references you provided as examples are all about bugs caused by code calling destructors twice (on moved and moved-from objects)

Not all of them. Yes, one of the provided problems results in double-destruction. There are also problems that have nothing to do with destructors (e.g., look for `1<<20` in https://abseil.io/tips/112). And even if we focus just on that double-destruction problem on [SO](https://stackoverflow.com/a/36919571), its _cause_  was in _construction_ (and that is where the emplace() danger is):

> after emplace() calls a badly designed move-constructor.

In the SO example, emplace() does not call a move constructor. Moreover, I would not call std::unique_ptr move constructor "badly designed". In fact, no move constructor can be badly designed -- they all have pretty much the same design (dictated by C++ language rules). Only their implementations differ.

In the SO example, emplace() calls an explicit `std::unique_ptr<T>::unique_ptr(T*)` constructor that converts a raw T pointer into a unique pointer. There is nothing wrong with that constructor either -- it is essential and well-designed.

The problems here are not triggered by any specific constructor. They are triggered by emplace_back() secret use of a  type conversion that is supposed to be explicit (via the use of an explicit constructor in the method implementation invisible to the code writer). Implicit type conversions can be very dangerous, of course!

In other words, emplace_back() essentially makes an explicit constructor implicit, effectively removing protections that explicit constructors are supposed to provide, such as exposing dangerous type conversions to the developer (who normally has to type the constructor name explicitly to get that conversion).

> These are expected "undefined behaviour" effects of buggy class X being passed around. 

There are no buggy classes in that SO problem. All classes are OK! Emplace() screws up code where there are no class-level bugs: `emplace_back(a)` compiles where the code author would really want it to fail (and where `push_back(a)` does fail to compile, exposing the bug in the caller).

----

> > > Amos: move semantics require (MUST) that the moved-from object be left in a valid null state by explicit move operators/constructors.

> > There is no "valid null state" requirement for explicit move methods. If you disagree, please provide a reference.
 
> C++11 document N337 section 12.2.1 on xvalue (temporary object) states. " Even when the creation of the temporary object is unevaluated (Clause 5) or otherwise avoided (12.8 [copy/move construction]), all the semantic restrictions shall be respected as if the temporary object had been created and later destroyed. "

That text places no restrictions on the move constructor! It places restrictions on what happens when the move constructor call is _avoided_. In simple in-scope terms, it tells the compiler, "If you were supposed to create a temporary object X (e.g., to pass to a move constructor), but you decided not to create X (and not call that move constructor), then you shall still respect all the semantic restrictions that would have existed if X had been created and later destroyed". 

For example, that temporary object may be avoided by applying move elision optimization. In fact, the C++11 standard gives that example right below the 12.2.1 text you have sited (at least it does in N3242 copy that I could [find](https://stackoverflow.com/a/7747129)).

> The relevant "semantic restrictions" for emplace_back() being the MoveConstructible trait defined in section 17.6.3.1 table 20: " Note:rv must still meet the requirements of the library component that is using it. The operations listed in those requirements must work as specified whether rv has been moved from or not. " [ `rv` being the moved-from object]

std::list::emplace_back() does not place MoveConstructible restrictions (on anything). It places [EmplaceConstructible](https://en.cppreference.com/w/cpp/named_req/EmplaceConstructible) restrictions on std::list value_type. In simpler terms, we can use emplace_back(x) when the list allocator can create value_type object in-place, using supplied parameter `x`. This has nothing to do with semantic restrictions on the move constructor.
 
> Thus the value "shall be" left valid as per default-initialization (ergo: nullptr).

The MoveConstructible requirement does not mention default-initialization. Furthermore, MoveConstructible explicitly places no restrictions on the `rv` value after construction. Here are citations from two sources:

* N3242: `rv` remains a valid object. Its state is unspecified.
* [cppreference](https://en.cppreference.com/w/cpp/named_req/MoveConstructible): The new value of `rv` is unspecified.

The only restriction is that `rv` must be a valid object (e.g., we can assign a new value to it without fear). I do not understand how one can get from MoveConstructible to "default-initialization".

In summary, neither the inapplicable specs nor the applicable ones place a "valid null state" or even "valid default-initializated state" requirement on explicit move methods. The actual post-move requirements are very simple and allow for various performance optimizations -- any valid state would do!

PR #1189, opened 2022-11-28
Do not overwrite caching bans

rousskov:
To ban caching, Squid made RequestFlags::cachable false in several
places (e.g., when a "cache deny" rule matched). To permit caching,
Squid also made the flag true in several places (e.g., when
maybeCacheable() was true). That combination worked as intended only
when the cachable=false veto always came after all the cachable=true
support votes. The code did not (and should not) enforce such a
complicated/fragile timing invariant.

Squid now correctly honors caching bans regardless of the updates order.

This change addresses an old XXX, but we are not aware of any specific
bugs fixed by this change. The primary purpose of this change is to make
the existing baseline ban-honoring functionality easy to maintain.

Also reduced code duplication across cachable=false,noCache=true code.

### DIFF
--- a/src/RequestFlags.h
+++ b/src/RequestFlags.h
@@ -11,7 +11,7 @@
 #ifndef SQUID_REQUESTFLAGS_H_
 #define SQUID_REQUESTFLAGS_H_
 
-#include "base/Optional.h"
+#include "base/SupportOrVeto.h"
 
 /** request-related flags
  *
@@ -31,8 +31,8 @@ public:
     /** do not use keytabs for peer Kerberos authentication */
     bool auth_no_keytab = false;
 
-    /// explicit decision to allow storing the response in the cache
-    Optional<bool> missCachingDecision;
+    /// whether the response may be stored in the cache
+    SupportOrVeto cachable;
 
     /** the request can be forwarded through the hierarchy */
     bool hierarchical = false;
@@ -118,11 +118,6 @@ public:
      *  a related (e.g. ICAP-adapted) request.
      */
     RequestFlags cloneAdaptationImmune() const;
-
-    /// Convenience wrapper for checking whether the response may be cached. By
-    /// default (i.e. unless we explicitly allow caching), the response is not
-    /// cachable. \sa noCache
-    bool cachable() const { return missCachingDecision.value_or(false); }
 
     // if FOLLOW_X_FORWARDED_FOR is not set, we always return "done".
     bool doneFollowXff() const {
--- a/src/RequestFlags.h
+++ b/src/RequestFlags.h
@@ -11,6 +11,8 @@
 #ifndef SQUID_REQUESTFLAGS_H_
 #define SQUID_REQUESTFLAGS_H_
 
+#include "base/Optional.h"
+
 /** request-related flags
  *
  * Contains both flags marking a request's current state,
@@ -28,8 +30,10 @@ public:
     bool auth = false;
     /** do not use keytabs for peer Kerberos authentication */
     bool auth_no_keytab = false;
-    /** he response to the request may be stored in the cache */
-    bool cachable = false;
+
+    /// explicit decision to allow storing the response in the cache
+    Optional<bool> missCachingDecision;
+
     /** the request can be forwarded through the hierarchy */
     bool hierarchical = false;
     /** a loop was detected on this request */
@@ -115,6 +119,11 @@ public:
      */
     RequestFlags cloneAdaptationImmune() const;
 
+    /// Convenience wrapper for checking whether the response may be cached. By
+    /// default (i.e. unless we explicitly allow caching), the response is not
+    /// cachable. \sa noCache
+    bool cachable() const { return missCachingDecision.value_or(false); }
+
     // if FOLLOW_X_FORWARDED_FOR is not set, we always return "done".
     bool doneFollowXff() const {
         return done_follow_x_forwarded_for || !FOLLOW_X_FORWARDED_FOR;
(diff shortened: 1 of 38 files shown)


rousskov:
This PR should not build until the low-level changes in #1188 are merged. I have verified that this PR build will succeed by temporary merging that PR changes into this PR branch, but our CI setup will enforce that anyway.

src/RequestFlags.h
```diff
@@ -28,8 +30,10 @@ class RequestFlags
     bool auth = false;
     /** do not use keytabs for peer Kerberos authentication */
     bool auth_no_keytab = false;
-    /** he response to the request may be stored in the cache */
-    bool cachable = false;
+
+    /// whether the response may be stored in the cache
+    SupportOrVeto cachable;
+
```
rousskov, line 36: > Amos at https://github.com/squid-cache/squid/pull/1185#discussion_r1032642857: Correctly implemented this request flag should mirror the HTTP semantics with default-true for easier validation of the resulting Squid behaviour. 

I strongly disagree. Due to the security concerns related to placing content in a _shared_ cache, we must continue to default this flag to false. While going against HTTP's "reuse is the default" design principle might feel counter-intuitive, it is the best implementation approach (among the flag-based ones) because not all flagged traffic is HTTP, and because there are no "MUST cache" protocol requirements (while there are plenty of "MUST NOT cache" rules that we must check if we decide to cache).

Caching is and will remain an _optional_ security-sensitive optimization. As long as we limit ourselves to a simple boolean decision flag, the only safe implementation default is "false". Thankfully, it is also the current default.

----

It would be even better to guarantee that the code has reached all the required caching decision points for the involved protocol(s) before allowing caching. In other words, the ideal implementation would upgrade from the boolean support-or-veto flag (that may miss a required decision making point and not know it) to an **all**-the-required-decision-makers-support-or-veto state. Reaching that ideal is too complex for this PR, but this PR changes will make such future adjustments (if any) easier by correctly enforcing vetos (and preserving the right default).

src/client_side_request.cc
```diff
@@ -1095,7 +1098,10 @@ clientInterpretRequestHeaders(ClientHttpRequest * http)
 
 #endif
 
-    request->flags.cachable = http->request->maybeCacheable();
+    if (http->request->maybeCacheable())
+        request->flags.cachable.support();
+    else
+        request->flags.cachable.veto();
```
rousskov, line 1104: We can remove/simplify the above `if` statement as illustrated below, but I am worried that folks will start adding `update(false)` and `update(true)` calls elsewhere if we add such an update() method. Compared to explicitly calling support() and veto(), calling update() decreases code readability, especially since the difference between true and false goes beyond pure boolean algebra here (due to veto powers of negative votes). Thus, I do not think we should add update() (or a similar method).

```C++
         request->flags.cachable.update(http->request->maybeCacheable());
```

src/RequestFlags.h
yadij: You are mixing concepts here. Caching as a whole is optional, that is not related to whether an individual response default is to *be* cacheable.

From a security standpoint the risks are greater whenever the standard behaviour is *not* followed. HTTP/1.1 has clear mechanisms which Squid makes use of now to prevent casual risks from cached security sensitive content (eg private responses, corrupt responses) affecting the network. This ```cacheable``` flag is not part of those protections, in part because it does not follow the standards and thus cannot be relied on.
  It would be nice to make it properly compliant with HTTP so we *could* rely on seeing a ```false``` value for security checks to forbid caching. Right now we cannot trust false to be true or just the HTTP-violation default.

rousskov: > Caching as a whole is optional, that is not related to whether an individual response default is to *be* cacheable.

Agreed. Nothing in my earlier response or this PR contradicts the above statement AFAICT.

> From a security standpoint the risks are greater whenever the standard behaviour is not followed.

In the scope of this PR, Squid implementation continues to follow the standard by refusing to cache objects that have not been through any standard cachability verification rules. The standard requires such verification (because there are MUST NOT cache requirements that a compliant implementation ought to check/obey). The official and PR code enforces that basic principle.

> HTTP/1.1 has clear mechanisms which Squid makes use of now to prevent casual risks from cached security sensitive content (eg private responses, corrupt responses) affecting the network.

Yes, and when Squid code implementing those mechanisms is reached, and request cachability is verified (i.e. the risks are mitigated), the cachable flag should become true. Until those mechanisms are reached, the flag should remain false. The official and PR code use that approach. It is not only very reasonable, it is probably the only safe approach based on a single boolean flag (as I detailed earlier).

> This cacheable flag is not part of those protections,

I disagree: We set this flag to true when HttpRequest::maybeCacheable() returns true. We veto caching when HttpRequest::maybeCacheable() returns false. That HttpRequest method is a part of those protections and, hence, this flag is too.

> It would be nice to make it properly compliant with HTTP so we _could_ rely on seeing a `false` value for security checks to forbid caching. 

We can (and do) already "rely on seeing a `false` value for security checks to forbid caching". No changes to the flag default are needed to achieve that goal! A false value means that we have not checked certain MUSTs yet or we have checked them and they prohibited caching. In both cases, we should forbid caching.

This flag does have a problem, but it is not its default value. It is its lack of awareness of whether we have checked all MUSTs that we are supposed to check. That problem cannot be solved within the confines of a boolean flag (as I detailed earlier).

> Right now we cannot trust false to be reality or just the HTTP-violation default.

Agreed (although it does not violate any HTTP MUSTs). And if we change the default from false to true, we would not be able to trust `true` to be reality either! To be able to fully trust this field, we need to convert it from a boolean flag to a list of the required protocol-specific checkers (that are activated later in the code). Only then we can really trust this field to fully reflect reality.

rousskov:
@yadij I would like to merge this sub-PR ASAP because it is blocking its parent PR and has no (known to me) flaws. If you do not intend to veto this PR, please consider approving it to speed things up. Thank you.

yadij:
Cleared for merge under our 10-days no objections policy.

PR #1190, opened 2022-11-29
Maintenance: Missing squid.h include

rousskov:
Broken since inception (i.e. master/v6 commit 4a28fc5).
Detected by scripts/source-maintenance.sh.


yadij (approved):
ouch

PR #1192, opened 2022-11-30
kerberos_ldap_group: fix unused variable warnings

yadij:
When built with disabled LDAP support.


src/acl/external/kerberos_ldap_group/kerberos_ldap_group.cc
```diff
@@ -489,7 +489,7 @@ strup(char *s)
 #else
 #include <cstdlib>
 int
-main(int argc, char *const argv[])
+main(int , char *const [])
```
rousskov, line 492:
```suggestion
main()
```

See (1) at https://en.cppreference.com/w/cpp/language/main_function

rousskov (approved):

PR #1194, opened 2022-11-30
Fix helper module auto-detection algorithm

yadij:
Two issues here causing no helpers to be detected as buildable when
./configure parameters are missing entirely (eg "./configure ").
These were not detected by our build farm because missing helpers
is not an error, they are optional, and we lack tests validating
optional helpers.

1) Over-escaping of "$" variable prefix (M4 "$$" expands to "$")
   prevented the path string being expanded to correct value for
   the helper directory test. Causing it to always fail.

2) The $ac_top_srcdir variable internal to autoconf may not be set
   at the time AC_DEFUN macros are expanded. Making it unreliable.
   The $top_srcdir variable which should be used instead has the
   same problem with "top" not yet being determined. As such we
   use $srcdir instead and hope it expands correctly at ./configure
   run time.


yadij:
Waiting for PR #1193 and #1194 to be merged. Without them this exposes broken helpers to build testing on Debian 12.

kinkie (approved):
I'm not fully confident in the "hope that" part of the PR description, but I trust that you run-tested this and it's working for you

yadij:
Yes this is run-tested and "works for me".

FWIW, The "hope" derives from the previous iteration also having similar testing and working well at that time, on those OS. So I am less confident about this one being a permanent solution.

kinkie:
It's not going to make things worse, so let's land it and iterate if neeeded.

rousskov (approved):
I cannot validate the details of this change, but I do support this going in, assuming Amos saw that this change _fixes_ the specific known bug (i.e. that the helpers are now built in a use case where they were not built before).

> PR description: These were not detected by our build farm because missing helpers is not an error, they are optional.

I would document the root cause instead: These bugs were not detected by our CI testing because we lack tests validating whether optional helpers are built in environments where they should be built.

I do not insist on any PR description changes.

I also do not insist on any corresponding CI improvements either, although I suspect it would not be difficult to improve our GitHub Actions scripts to validate that a few optional helpers are built on the known/current GitHub Actions runners (where we _know_ which optional helpers should be built).

wweric:
Just wanted to say I had same issue of helpers not being detected/built as well and this seems to have corrected it, so far...

PR #1195, opened 2022-12-02
Refactor HTTP Digest authentication Nonce handling

yadij:
Convert the Nonce object to class with MEMPROXY memory
pooling, smart pointer reference counting, and most of the
old management functions converted to members.

While logic changes are implicit from refactoring, behaviour
change is avoided as much as possible.

However, one bug was identified which would prevent a new
nonce generated to replace a stale existing nonce being
associated to the username needing it. The if-condition
causing this ban has been dropped.

### DIFF
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -342,28 +342,21 @@ digest_nonce_h::stale()
     return false;
 }
 
-/**
- * \retval  0    the digest is not stale yet
- * \retval -1    the digest will be stale on the next request
- */
-int
-authDigestNonceLastRequest(digest_nonce_h * nonce)
+bool
+digest_nonce_h::lastRequest() const
 {
-    if (!nonce)
-        return -1;
-
-    if (nonce->nc == 99999997) {
+    if (nc == 99999997) {
         debugs(29, 4, "Nonce count about to overflow");
-        return -1;
+        return true;
     }
 
-    if (nonce->nc >= static_cast<Auth::Digest::Config*>(Auth::SchemeConfig::Find("digest"))->noncemaxuses - 1) {
+    if (nc >= static_cast<Auth::Digest::Config*>(Auth::SchemeConfig::Find("digest"))->noncemaxuses - 1) {
         debugs(29, 4, "Nonce count about to hit user limit");
-        return -1;
+        return true;
     }
 
     /* and other tests are possible. */
-    return 0;
+    return false;
 }
 
 void
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -257,15 +257,6 @@ authDigestNonceUnlink(digest_nonce_h * nonce)
         delete nonce;
 }
 
-const char *
-authenticateDigestNonceNonceHex(const digest_nonce_h * nonce)
-{
-    if (!nonce)
-        return nullptr;
-
-    return (char const *) nonce->key;
-}
-
 static digest_nonce_h *
 authenticateDigestNonceFindNonce(const char *noncehex)
 {
@@ -278,7 +269,7 @@ authenticateDigestNonceFindNonce(const char *noncehex)
 
     nonce = static_cast < digest_nonce_h * >(hash_lookup(digest_nonce_cache, noncehex));
 
-    if ((nonce == nullptr) || (strcmp(authenticateDigestNonceNonceHex(nonce), noncehex)))
+    if (!nonce || strcmp(nonce->hex(), noncehex) != 0)
         return nullptr;
 
     debugs(29, 9, "Found nonce '" << nonce << "'");
@@ -472,12 +463,12 @@ Auth::Digest::Config::fixHeader(Auth::UserRequest::Pointer auth_user_request, Ht
 
     debugs(29, 9, "Sending type:" << hdrType <<
            " header: 'Digest realm=\"" << realm << "\", nonce=\"" <<
-           authenticateDigestNonceNonceHex(nonce) << "\", qop=\"" << QOP_AUTH <<
+           nonce->hex() << "\", qop=\"" << QOP_AUTH <<
            "\", stale=" << (stale ? "true" : "false"));
 
     /* in the future, for WWW auth we may want to support the domain entry */
     httpHeaderPutStrf(&rep->header, hdrType, "Digest realm=\"" SQUIDSBUFPH "\", nonce=\"%s\", qop=\"%s\", stale=%s",
-                      SQUIDSBUFPRINT(realm), authenticateDigestNonceNonceHex(nonce), QOP_AUTH, stale ? "true" : "false");
+                      SQUIDSBUFPRINT(realm), nonce->hex(), QOP_AUTH, stale ? "true" : "false");
 }
 
 /* Initialize helpers and the like for this auth scheme. Called AFTER parsing the
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -38,11 +38,6 @@
 #include "StrList.h"
 #include "wordlist.h"
 
-/* digest_nonce_h still uses explicit alloc()/freeOne() MemPool calls.
- * XXX: convert to MEMPROXY_CLASS() API
- */
-#include "mem/Pool.h"
-
 static AUTHSSTATS authenticateDigestStats;
 
 helper *digestauthenticators = nullptr;
@@ -50,7 +45,6 @@ helper *digestauthenticators = nullptr;
 static hash_table *digest_nonce_cache;
 
 static int authdigest_initialised = 0;
-static Mem::Allocator *digest_nonce_pool = nullptr;
 
 enum http_digest_attr_type {
     DIGEST_USERNAME,
@@ -90,7 +84,6 @@ DigestFieldsLookupTable(DIGEST_INVALID_ATTR, DigestAttrs);
 
 static void authenticateDigestNonceCacheCleanup(void *data);
 static digest_nonce_h *authenticateDigestNonceFindNonce(const char *noncehex);
-static void authenticateDigestNonceDelete(digest_nonce_h * nonce);
 static void authenticateDigestNonceSetup(void);
 static void authDigestNonceEncode(digest_nonce_h * nonce);
 static void authDigestNonceLink(digest_nonce_h * nonce);
@@ -118,8 +111,6 @@ authDigestNonceEncode(digest_nonce_h * nonce)
 digest_nonce_h *
 authenticateDigestNonceNew(void)
 {
-    digest_nonce_h *newnonce = static_cast < digest_nonce_h * >(digest_nonce_pool->alloc());
-
     /* NONCE CREATION - NOTES AND REASONING. RBC 20010108
      * === EXCERPT FROM RFC 2617 ===
      * The contents of the nonce are implementation dependent. The quality
@@ -161,8 +152,7 @@ authenticateDigestNonceNew(void)
     static xuniform_int_distribution<uint32_t> newRandomData;
 
     /* create a new nonce */
-    newnonce->nc = 0;
-    newnonce->flags.valid = true;
+    digest_nonce_h *newnonce = new digest_nonce_h;
     newnonce->noncedata.creationtime = current_time.tv_sec;
     newnonce->noncedata.randomdata = newRandomData(mt);
 
@@ -184,24 +174,8 @@ authenticateDigestNonceNew(void)
 }
 
 static void
-authenticateDigestNonceDelete(digest_nonce_h * nonce)
-{
-    if (nonce) {
-        assert(nonce->references == 0);
-        assert(!nonce->flags.incache);
-
-        safe_free(nonce->key);
-
-        digest_nonce_pool->freeOne(nonce);
-    }
-}
-
-static void
 authenticateDigestNonceSetup(void)
 {
-    if (!digest_nonce_pool)
-        digest_nonce_pool = memPoolCreate("Digest Scheme nonce's", sizeof(digest_nonce_h));
-
     if (!digest_nonce_cache) {
         digest_nonce_cache = hash_create((HASHCMP *) strcmp, 7921, hash_string);
         assert(digest_nonce_cache);
@@ -288,7 +262,7 @@ authDigestNonceUnlink(digest_nonce_h * nonce)
     debugs(29, 9, "nonce '" << nonce << "' now at '" << nonce->references << "'.");
 
     if (nonce->references == 0)
-        authenticateDigestNonceDelete(nonce);
+        delete nonce;
 }
 
 const char *
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -87,7 +87,6 @@ static digest_nonce_h *authenticateDigestNonceFindNonce(const char *noncehex);
 static void authenticateDigestNonceSetup(void);
 static void authDigestNonceEncode(digest_nonce_h * nonce);
 static void authDigestNonceLink(digest_nonce_h * nonce);
-static void authDigestNonceUserUnlink(digest_nonce_h * nonce);
 
 static void
 authDigestNonceEncode(digest_nonce_h * nonce)
@@ -228,7 +227,8 @@ authenticateDigestNonceCacheCleanup(void *)
             /* invalidate nonce so future requests fail */
             nonce->flags.valid = false;
             /* if it is tied to a auth_user, remove the tie */
-            authDigestNonceUserUnlink(nonce);
+            if (nonce->user)
+                nonce->user->unlink(nonce);
             authDigestNoncePurge(nonce);
         }
     }
@@ -425,7 +425,7 @@ Auth::Digest::Config::fixHeader(Auth::UserRequest::Pointer auth_user_request, Ht
         return;
 
     bool stale = false;
-    digest_nonce_h *nonce = nullptr;
+    digest_nonce_h::Pointer nonce;
 
     /* on a 407 or 401 we always use a new nonce */
     if (auth_user_request != nullptr) {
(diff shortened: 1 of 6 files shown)


src/auth/digest/Config.cc
```diff
@@ -606,79 +544,6 @@ authenticateDigestStats(StoreEntry * sentry)
...
-    nonce->user = nullptr;
-}
-
-/* authDigesteserLinkNonce: add a nonce to a given user's struct */
-void
-authDigestUserLinkNonce(Auth::Digest::User * user, digest_nonce_h * nonce)
-{
-    dlink_node *node;
-
-    if (!user || !nonce || !nonce->user)
```
yadij, line 654: FTR; the moved code (now in User.cc) drops the ```|| !nonce->user``` clause. This function was often being called for freshly generated nonce objects whose ```user``` member is a nullptr ... apparently intending this function to set that ```user``` member. But this condition prevents the remainder of this function ever working as intended and setting the ```user``` member.

yadij:
FYI: to simplify review I have left the nonce class name and most of the code as-is. Once these logic related changes pass review I intend to rename it ```Auth::Digest::Nonce``` and move out of Config.{h,cc} to more appropriate Nonce.{h,cc} files.

For now I have left the digest_nonce_cache unchanged due to lack of time and it being more intrusive that the dlink removal. If requested it may also be included for the final merge.

src/auth/digest/Config.cc
```diff
@@ -38,19 +38,13 @@
 #include "StrList.h"
```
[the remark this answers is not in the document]
yadij: I mean that keeping this a pure "no logic changes" is unreasonable for a refactor. Do you want to state all the logic changes visible in the patch in the description?

src/auth/digest/Config.h
```diff
@@ -25,42 +26,61 @@ class User;
...
+
+    /// The HEX encoded unique identifier for this nonce
+    const char *hex() const { return static_cast<const char *>(key); }
+
+    /// Check the nonce and invalidate if any tests fail.
+    /// \retval true if the nonce is valid.
+    bool valid(char clientCount[9]);
+
+    /// Check the freshness of this nonce and invalidate if stale.
+    /// \retval true if the nonce is stale.
```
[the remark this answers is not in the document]
yadij: Not adding this change because stale is a subset of invalid for nonce. Hopefully the update to clarify the preceeding comments will resolve this misunderstanding.

src/auth/digest/Config.cc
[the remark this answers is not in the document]
rousskov: > I mean that keeping this a pure "no logic changes" is unreasonable for a refactor.

Please rephrase to clarify what you mean. For example, we could say something like "This refactoring changes Squid behavior in many ways, including these major changes: ..." (if that is accurate/true). This is just an example; please do not attack it -- I do not know what the logic changes are, you do. Please rephrase the best you can to address the concern.

> Do you want to state all the logic changes visible in the patch in the description?

The answer depends on what logic changes are clearly visible in the patch to you. I do not see _any_ obvious logic changes in the patch, but this is not my area of expertise, and I am not the author of this PR. Perhaps you can give two or three examples of the biggest changes in Squid behavior that this PR introduces?

src/auth/digest/Config.h
```diff
@@ -25,42 +26,61 @@ class User;
...
+    digest_nonce_h() = default;
+    digest_nonce_h(const digest_nonce_h &) = delete; // non-copyable
+    ~digest_nonce_h() { xfree(key); }
+
+    /// The HEX encoded unique identifier for this nonce
+    const char *hex() const { return static_cast<const char *>(key); }
+
+    /// Check the nonce and invalidate if any tests fail.
+    /// \retval true if the nonce is valid.
+    bool valid(char clientCount[9]);
```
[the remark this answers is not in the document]
yadij: This is a setter method. Done the latter.

src/auth/digest/User.h
```diff
@@ -31,17 +31,21 @@ class User : public Auth::User
...
 
+public:
     HASH HA1;
     int HA1created;
 
     /* what nonces have been allocated to this user */
-    dlink_list nonces;
-
-    digest_nonce_h * currentNonce();
+    std::list<digest_nonce_h::Pointer> nonces;
```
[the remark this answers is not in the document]
yadij: Done in d973d636134da894fbdcbf08c415340cd2876663

src/auth/digest/User.h
```diff
@@ -31,17 +31,21 @@ class User : public Auth::User
...
+
     /* Auth::User API */
     static CbcPointer<Auth::CredentialsCache> Cache();
     virtual void addToNameCache() override;
 
+public:
     HASH HA1;
     int HA1created;
 
     /* what nonces have been allocated to this user */
```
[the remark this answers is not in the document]
yadij: Done in 2c9e1286e50568a4022772edcb230c63bbcb38ef

src/auth/digest/Config.cc
```diff
@@ -312,113 +269,94 @@ authenticateDigestNonceFindNonce(const char *noncehex)
...
 
     debugs(29, 9, "Found nonce '" << nonce << "'");
 
     return nonce;
 }
 
-int
-authDigestNonceIsValid(digest_nonce_h * nonce, char nc[9])
+bool
+digest_nonce_h::valid(char aCount[9])
```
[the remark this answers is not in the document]
yadij: Done in [67d6269](https://github.com/squid-cache/squid/pull/1195/commits/67d6269bc28caa2ffb39a5a9c6e518e978de87b7)

src/auth/digest/Config.h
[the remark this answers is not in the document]
yadij: Done in f9cf75b97444e0256c3c1d69520920a1a145fd3a

src/auth/digest/Config.h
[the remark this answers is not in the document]
yadij: Done in f9cf75b97444e0256c3c1d69520920a1a145fd3a

src/auth/digest/Config.h
```diff
@@ -25,42 +26,61 @@ class User;
...
+
+    /// Check the freshness of this nonce and invalidate if stale.
+    /// \retval true if the nonce is stale.
+    bool stale();
```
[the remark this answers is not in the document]
yadij: Done in f9cf75b97444e0256c3c1d69520920a1a145fd3a

src/auth/digest/Config.cc
```diff
@@ -265,39 +239,22 @@ authenticateDigestNonceCacheCleanup(void *)
...
+// XXX: replace raw-pointers using this with Pointer
 static void
 authDigestNonceLink(digest_nonce_h * nonce)
 {
     assert(nonce != nullptr);
-    ++nonce->references;
-    assert(nonce->references != 0); // no overflows
-    debugs(29, 9, "nonce '" << nonce << "' now at '" << nonce->references << "'.");
+    nonce->lock();
 }
```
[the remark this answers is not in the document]
yadij: Done in 7673dd7783b8661165816f74eb1f0f294fce9fe9

src/auth/digest/Config.h
```diff
@@ -25,42 +26,61 @@ class User;
...
-struct _digest_nonce_data {
-    time_t creationtime;
-    uint32_t randomdata;
-};
+public:
+    typedef RefCount<digest_nonce_h> Pointer;
 
-/* the nonce structure we'll pass around */
+    digest_nonce_h() = default;
+    digest_nonce_h(const digest_nonce_h &) = delete; // non-copyable
```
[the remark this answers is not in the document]
yadij: Done in 7673dd7783b8661165816f74eb1f0f294fce9fe9

src/auth/digest/Config.cc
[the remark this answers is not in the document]
yadij: Uhm, https://github.com/squid-cache/squid/pull/1195#pullrequestreview-1202353848 (nonce renewal behaviour) and https://github.com/squid-cache/squid/pull/1195#issuecomment-1335111479 (dlink removal).

src/auth/digest/Config.cc
```diff
@@ -161,8 +151,7 @@ authenticateDigestNonceNew(void)
     static xuniform_int_distribution<uint32_t> newRandomData;
 
     /* create a new nonce */
-    newnonce->nc = 0;
-    newnonce->flags.valid = true;
+    digest_nonce_h *newnonce = new digest_nonce_h;
```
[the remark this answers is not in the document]
yadij: Since you have requested including the hash removal. this will become a Pointer as part of that.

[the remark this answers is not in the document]
yadij: Done in [d550605](https://github.com/squid-cache/squid/pull/1195/commits/d550605b93db91b4b370ea3b0901558ce4b96256)

src/auth/digest/Config.cc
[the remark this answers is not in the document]
yadij: Done in [4fdefa6](https://github.com/squid-cache/squid/pull/1195/commits/4fdefa691fcbde1d8eee7888074909da5d6c05fa)

src/auth/digest/Config.cc
```diff
@@ -151,85 +156,61 @@ authenticateDigestNonceNew(void)
...
-        hash_first(digest_nonce_cache);
-
-        while ((nonce = ((digest_nonce_h *) hash_next(digest_nonce_cache)))) {
-            assert(nonce->flags.incache);
-            authDigestNoncePurge(nonce);
-        }
+    // empty the cache of any nonces left in there
+    for (const auto &entry: NonceCache()) {
+        entry.second->purge();
     }
```
rousskov, line 185: Modifying the container we are iterating using a for-range loop is a UB -- the loop does not compute and store the next iterator value before we invalidate the current iterator.

Another, more subtle problem here (or in the purge() method) is that purge() is very forgiving -- it does nothing if the entry is not cached. It _should_ be cached, but if that implicit invariant is violated, we will loop forever here.

The flags.incache hack complicates things quite a bit, but I suspect the best solution here is either 

* a while-not-empty loop that purges the begin() entry and asserts that purge() was successful; that assertion can be implemented without relying on purge() internals by comparing the cache size() before and after purge()
* a for-range loop that clears flags.incache, followed by a single NonceCache.clear() call; this somewhat invasive option can be viewed as a performance optimization for large caches

If these caches may be large, I would go for the second option, especially since flags.incache itself is essentially an optimization (AFAICT).

src/auth/digest/Config.cc
```diff
@@ -151,85 +156,61 @@ authenticateDigestNonceNew(void)
...
-    digest_nonce_h *nonce;
     debugs(29, 3, "Cleaning the nonce cache now");
     debugs(29, 3, "Current time: " << current_time.tv_sec);
-    hash_first(digest_nonce_cache);
 
-    while ((nonce = ((digest_nonce_h *) hash_next(digest_nonce_cache)))) {
-        debugs(29, 3, "nonce entry  : " << nonce << " '" << (char *) nonce->key << "'");
+    for (const auto &entry: NonceCache()) {
+        auto nonce = entry.second;
+        if (!nonce)
```
rousskov, line 202: How is a nil cache entry possible? The only code that adds to the cache is the digest_nonce_h::cache() method, right? That method adds "this" so it cannot add a nil pointer. If I am not missing something, we should assert() or assume that the entry is there.

src/auth/digest/Config.cc
```diff
@@ -151,85 +156,61 @@ authenticateDigestNonceNew(void)
...
-            assert(nonce->flags.incache);
-            /* invalidate nonce so future requests fail */
-            nonce->flags.valid = false;
+        if (!nonce->validate()) {
+            debugs(29, 4, "Removing nonce " << nonce->hex() << " from cache due to timeout.");
             /* if it is tied to a auth_user, remove the tie */
             if (nonce->user)
                 nonce->user->unlink(nonce);
-            authDigestNoncePurge(nonce);
+            nonce->purge();
```
rousskov, line 213: Purging in a range-based for loop is a UB. See another change request about the same problem.

src/auth/digest/Config.h
```diff
@@ -73,15 +91,12 @@ class digest_nonce_h : public hash_link, public RefCountable
 
     /* has this nonce been invalidated ? */
     struct {
-        bool valid = true;
+        mutable bool valid = true;
```
rousskov, line 94: Can we undo this change? Is there some code that modifies this flag in a constant object?

I did suggest making this flag mutable in a change request, but only in conjunction with making valid() constant. In the current PR code, there are only non-constant validate() methods. They do not need this flag to be mutable.

src/auth/digest/Config.h
```diff
@@ -73,15 +91,12 @@ class digest_nonce_h : public hash_link, public RefCountable
...
         bool incache = false;
     } flags;
```
rousskov, line 96: If possible, let's make these dangerous/sensitive flags private.

src/auth/digest/Config.cc
rousskov: Alternatively, consider doing nothing in this shutdown method (other than asserting that we are shutting down, perhaps). What is the point of cleaning up a _temporary cache_ in a to-be-terminated program? Such a cache is essentially an optional optimization that we should be able to ignore on shutdown, right?

rousskov (changes requested):
Marked some [UBs](https://en.wikipedia.org/wiki/Undefined_behavior) to help you make progress.

src/auth/digest/Config.cc
rousskov: BTW, these repeated bugs are related to the same API problem -- an object should not control its own membership in an externally-managed container. Completely solving this problem is not trivial because it requires protecting the cache from invasive external actions, but moving away from nonce::purge() to a Purge(iterator) function may reduce similar problems in future code. I recommend trying to make a few steps in that direction, but I do not insist on those difficult improvements.

src/auth/digest/Config.cc
yadij: These are std::list - "Adding, removing and moving the elements within the list or across several lists does not invalidate the iterators or references. An iterator is invalidated only when the corresponding element is deleted. "
Which specifically means that ranged for are safely defined behaviour.

rousskov: > An iterator is invalidated only when the corresponding element is deleted

... which is exactly what this code is doing.

>  Which specifically means that ranged for are safely defined behaviour.

... unless that loop deletes the current element.

rousskov: > These are std::list

This loop iterates std::unordered_map. However, in this digest_nonce_h::purge()/NonceCache().erase() context, std::unordered_map has the same iterator invalidation problem as std::list would have had -- the erase() method invalidates the current iterator.

yadij: https://en.cppreference.com/w/cpp/container/unordered_map/erase
"The order of the elements that are not erased is preserved. (This makes it possible to erase individual elements while iterating through the container.) "

rousskov: > The order of the elements that are not erased is preserved. (This makes it possible to erase individual elements while iterating through the container.)

It is, indeed, possible to erase individual elements while iterating through std::unordered_map. Nothing in this thread contradicts that fact! However, we still have to avoid using invalidated iterators when erasing-while-iterating. There is no guarantee that every range-based `for` loop implementation encountered by Squid will avoid that UB. Moreover, a well-known _natural_ [implementation](https://en.cppreference.com/w/cpp/language/range-for) triggers that UB when it increments `__begin` iterator invalidated by the `loop-statement`:

```C++
for (auto __begin = begin-expr, __end = end-expr; __begin != __end; ++__begin) {
    range-declaration = *__begin;
    loop-statement
}
```

In my [test](https://godbolt.org/z/zz6G5GcoT), GCC increments the iterator invalidated by the `loop-statement` in a range-based `for` loop, matching the above "natural implementation" of that loop (AFAICT):

```asm
    call std::__detail::_Node_iterator<...>::operator*() const
    ...
    call std::unordered_map<...>::erase(int const&)
    ...
    call std::__detail::_Node_iterator<...>::operator++()
```

yadij:
To make some progress I am re-winding to a smaller scope. This branch/PR now just converts the memory pool from the old Squid-2 style to using `MEMPROXY_CLASS`.
That requires promoting the `struct` to a full `class`, but otherwise no logic outside the allocator is touched.

src/auth/digest/Config.cc
```diff
@@ -121,7 +120,7 @@ authDigestNonceEncode(digest_nonce_h * nonce)
 digest_nonce_h *
 authenticateDigestNonceNew(void)
 {
-    digest_nonce_h *newnonce = static_cast < digest_nonce_h * >(digest_nonce_pool->alloc());
+    auto *newnonce = new Auth::Digest::Nonce;
```
rousskov, line 123: If possible, please mark that variable as `const` (and avoid unnecessary decorations that will require more changes later). AFAICT, this function does not change `newnonce` value.

```suggestion
    const auto newnonce = new Auth::Digest::Nonce;
```

src/auth/digest/Config.h
```diff
@@ -11,48 +11,13 @@
 
 #if HAVE_AUTH_MODULE_DIGEST
 
+#include "auth/digest/Nonce.h"
```
rousskov, line 14: If possible, please use a forward declaration of digest_nonce_h (and any other types) instead -- the header code below may not need to know digest_nonce_h class details... For example:

```suggestion
#include "auth/digest/forward.h"
```

Addressing this change request probably requires moving (temporary) digest_nonce_h declaration hack to forward.h as discussed in another change request: If that move is not feasible, then we cannot use forward.h here.

src/auth/digest/Nonce.h
```diff
@@ -0,0 +1,53 @@
...
+    /// data to be encoded into the nonce's hex representation
+    struct _Data {
+        time_t creationtime = 0;
+        uint32_t randomdata = 0;
+    } noncedata;
+
+    /// number of uses we've seen of this nonce
+    unsigned long nc = 0;
+
+    // TODO: replace with RefCountable
```
rousskov, line 34: I suggest removing this TODO because it may be easily misinterpreted as if Nonce::references should be replaced with RefCountable::LockCount() as opposed to making Nonce class RefCountable.

```suggestion
```

src/auth/digest/Nonce.h
```diff
@@ -0,0 +1,53 @@
...
+        bool valid = false; ///< whether this nonce has been invalidated
+        bool incache = false; ///< whether this nonce is linked to the digest_nonce_cache
+    } flags;
+};
+
+} // namespace Digest
+} // namespace Auth
+
+// TODO: remove diff-reduction alias
+typedef Auth::Digest::Nonce digest_nonce_h;
```
rousskov, line 51: If possible, please use `using` in new code:

```suggestion
using digest_nonce_h = Auth::Digest::Nonce;
```

If this alias declaration can be moved into the new forward.h file, let's do that.

src/auth/digest/Nonce.h
```diff
@@ -0,0 +1,53 @@
...
+{
+namespace Digest
+{
+
+class Nonce : public hash_link
+{
+    MEMPROXY_CLASS(Auth::Digest::Nonce);
+public:
+    /// data to be encoded into the nonce's hex representation
+    struct _Data {
```
rousskov, line 26: If we do not have to start this name with an underscore, let's not do that:

```suggestion
    struct Data {
```

or even completely remove that name:

```suggestion
    struct {
```

src/auth/digest/Nonce.h
```diff
@@ -0,0 +1,53 @@
...
+
+    /// number of uses we've seen of this nonce
+    unsigned long nc = 0;
+
+    // TODO: replace with RefCountable
+    /// reference counting
+    uint64_t references = 0;
+
+    /// the auth_user this nonce has been tied to
+    Auth::Digest::User *user = nullptr;
```
rousskov, line 39: Now that the class has been moved into a proper namespace, I hope that we no longer have to repeat that namespace info:

```suggestion
    User *user = nullptr;
```

src/auth/digest/Config.cc
```diff
@@ -51,7 +51,6 @@ Helper::ClientPointer digestauthenticators;
 static hash_table *digest_nonce_cache;
```
rousskov, line 51: A few lines above, there is an XXX comment quoted below. Please remove or update quoted code because this PR appears to address its "convert to..."  to-do and definitely makes parts of that comment stale:

```C++
/* digest_nonce_h still uses explicit alloc()/freeOne() MemPool calls.
 * XXX: convert to MEMPROXY_CLASS() API
 */
#include "mem/Allocator.h"
#include "mem/Pool.h"
```

rousskov (changes requested):
Thank you for reducing this PR complexity and risk. Please keep 2022 discussions in mind when resurrecting the other changes in future PRs.

I adjusted PR description, primarily to mention a full new class name that may be useful in feature commit log searches. Please adjust further as needed.

PR #1196, opened 2022-12-02
add refresh_pattern option 'ignore-must-revalidate' back in

compholio:
This is a partial revert of commit 064679ea374d2f58ae4660d9a2af213b9be24bba, which removed the "ignore-must-revalidate" option.  The option was not actually broken, but it is not about the storage of objects - it is about whether they need to be retrieved again for new requests.

For example, I encountered a university download service that has set both "Expires: 0" and "Cache-Control: must-revalidate" for unchanging content.  The refresh_pattern "override-expires" option will allow you to store a cache of this content, but due to the "Cache-Control: must-revalidate" header the proxy server will always re-fetch the content because the server has effectively indicated (incorrectly) that it is a live data feed.  By re-enabling the "ignore-must-revalidate" refresh_pattern option it is possible to override the behavior for the download service's specific URL, such that the proxy can properly cache the content and not have to constantly re-fetch a bunch of multi-gigabyte files.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -6445,6 +6445,7 @@ DOC_START
 		 reload-into-ims
 		 ignore-reload
 		 ignore-no-store
+		 ignore-must-revalidate
 		 ignore-private
 		 max-stale=NN
 		 refresh-ims
@@ -6477,6 +6478,11 @@ DOC_START
 		it causes.
 
 		ignore-no-store ignores any ``Cache-control: no-store''
+		headers received from a server. Doing this VIOLATES
+		the HTTP standard. Enabling this feature could make you
+		liable for problems which it causes.
+
+		ignore-must-revalidate ignores any ``Cache-Control: must-revalidate``
 		headers received from a server. Doing this VIOLATES
 		the HTTP standard. Enabling this feature could make you
 		liable for problems which it causes.
--- a/src/http.cc
+++ b/src/http.cc
@@ -395,7 +395,7 @@ HttpStateData::reusableReply(HttpStateData::ReuseDecision &decision)
             mayStore = true;
 
             // HTTPbis pt6 section 3.2: a response CC:must-revalidate is present
-        } else if (rep->cache_control->hasMustRevalidate()) {
+        } else if (rep->cache_control->hasMustRevalidate() && !REFRESH_OVERRIDE(ignore_must_revalidate)) {
             debugs(22, 3, "Authenticated but server reply Cache-Control:must-revalidate");
             mayStore = true;
 
@@ -404,7 +404,7 @@ HttpStateData::reusableReply(HttpStateData::ReuseDecision &decision)
             // HTTPbis WG verdict on this is that it is omitted from the spec due to being 'unexpected' by
             // some. The caching+revalidate is not exactly unsafe though with Squids interpretation of no-cache
             // (without parameters) as equivalent to must-revalidate in the reply.
-        } else if (rep->cache_control->hasNoCacheWithoutParameters()) {
+        } else if (rep->cache_control->hasNoCacheWithoutParameters() && !REFRESH_OVERRIDE(ignore_must_revalidate)) {
             debugs(22, 3, "Authenticated but server reply Cache-Control:no-cache (equivalent to must-revalidate)");
             mayStore = true;
 #endif
--- a/src/refresh.cc
+++ b/src/refresh.cc
@@ -336,6 +336,9 @@ refreshCheck(const StoreEntry * entry, HttpRequest * request, time_t delta)
      */
     const bool revalidateAlways = EBIT_TEST(entry->flags, ENTRY_REVALIDATE_ALWAYS);
     if (revalidateAlways || (staleness > -1 &&
+#if USE_HTTP_VIOLATIONS
+                             !R->flags.ignore_must_revalidate &&
+#endif
                              EBIT_TEST(entry->flags, ENTRY_REVALIDATE_STALE))) {
         debugs(22, 3, "YES: Must revalidate stale object (origin set " <<
                (revalidateAlways ? "no-cache or private" :
--- a/src/RefreshPattern.h
+++ b/src/RefreshPattern.h
@@ -63,6 +63,7 @@ public:
         bool reload_into_ims;
         bool ignore_reload;
         bool ignore_no_store;
+        bool ignore_must_revalidate;
         bool ignore_private;
 #endif
     } flags;
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -877,6 +877,13 @@ configDoConfigure(void)
         }
 
         for (R = Config.Refresh; R; R = R->next) {
+            if (!R->flags.ignore_must_revalidate)
+                continue;
+            debugs(22, DBG_IMPORTANT, "WARNING: use of 'ignore-must-revalidate' in 'refresh_pattern' violates HTTP");
+            break;
+        }
+
+        for (R = Config.Refresh; R; R = R->next) {
             if (!R->flags.ignore_private)
                 continue;
 
@@ -2711,6 +2718,9 @@ dump_refreshpattern(StoreEntry * entry, const char *name, RefreshPattern * head)
         if (head->flags.ignore_no_store)
             storeAppendPrintf(entry, " ignore-no-store");
 
+        if (head->flags.ignore_must_revalidate)
+            storeAppendPrintf(entry, " ignore-must-revalidate");
+
         if (head->flags.ignore_private)
             storeAppendPrintf(entry, " ignore-private");
 #endif
@@ -2739,6 +2749,7 @@ parse_refreshpattern(RefreshPattern ** head)
     int reload_into_ims = 0;
     int ignore_reload = 0;
     int ignore_no_store = 0;
+    int ignore_must_revalidate = 0;
     int ignore_private = 0;
 #endif
 
@@ -2785,7 +2796,6 @@ parse_refreshpattern(RefreshPattern ** head)
             store_stale = 1;
         } else if (!strncmp(token, "max-stale=", 10)) {
             max_stale = xatoi(token + 10);
-
 #if USE_HTTP_VIOLATIONS
 
         } else if (!strcmp(token, "override-expire"))
@@ -2794,6 +2804,8 @@ parse_refreshpattern(RefreshPattern ** head)
             override_lastmod = 1;
         else if (!strcmp(token, "ignore-no-store"))
             ignore_no_store = 1;
+        else if (!strcmp(token, "ignore-must-revalidate"))
+            ignore_must_revalidate = 1;
         else if (!strcmp(token, "ignore-private"))
             ignore_private = 1;
         else if (!strcmp(token, "reload-into-ims")) {
@@ -2846,6 +2858,9 @@ parse_refreshpattern(RefreshPattern ** head)
 
     if (ignore_no_store)
         t->flags.ignore_no_store = true;
+
+    if (ignore_must_revalidate)
+        t->flags.ignore_must_revalidate = true;
 
     if (ignore_private)
         t->flags.ignore_private = true;


kinkie:
ok to test

compholio:
Fixed that the previous push was missing the `#if USE_HTTP_VIOLATIONS` protection in `src/refresh.cc`.  I'm not sure what the M-failed-description tag means, presumably some sort of formatting issue?

src/cf.data.pre
```diff
@@ -6481,6 +6482,11 @@ DOC_START
 		the HTTP standard. Enabling this feature could make you
 		liable for problems which it causes.
 
+		ignore-must-revalidate ignores any ``Cache-Control: must-revalidate``
+		headers received from a server. Doing this VIOLATES
+		the HTTP standard. Enabling this feature could make you
+		liable for problems which it causes.
+
```
rousskov, line 6489: Please add text to warn the admin that enabling this feature prevents Squid from caching authenticated responses containing must-revalidate and no-cache directives.

Roughly speaking, folks enable ignore-must-revalidate to "cache more" rather than to "cache less". The opposite side effect (for authenticated responses) is likely to be unexpected by most admins. We should document that side effect.

src/refresh.cc
```diff
@@ -336,6 +336,9 @@ refreshCheck(const StoreEntry * entry, HttpRequest * request, time_t delta)
      */
     const bool revalidateAlways = EBIT_TEST(entry->flags, ENTRY_REVALIDATE_ALWAYS);
     if (revalidateAlways || (staleness > -1 &&
+#if USE_HTTP_VIOLATIONS
+                             !R->flags.ignore_must_revalidate &&
+#endif
                              EBIT_TEST(entry->flags, ENTRY_REVALIDATE_STALE))) {
```
rousskov, line 342: This change has two problems AFAICT:

* It applies ignore-must-revalidate configuration option to responses with `Cache-Control: max-age`. That is wrong.
* It applies ignore-must-revalidate configuration option to responses cached _before_ the option was enabled. The proposed documentation does not make it clear that the option affects already cached responses. We should be explicit about this unexpected (for some) "backdating" effect or unexpected (for some) lack of thereof.

The fix(es) would depend on whether we want ignore-must-revalidate to apply to previously cached responses. AFAICT, if we want the simplest solution, then we should answer "no" and

* "move" this check, in its REFRESH_OVERRIDE() form, to where hasMustRevalidate() is called by HttpStateData::haveParsedReplyHeaders(), eventually resulting in setting ENTRY_REVALIDATE_STALE;
* adjust proposed documentation to explicitly exclude already cached responses from the configuration option scope/control.

src/cf.data.pre
rousskov, line 6489: _After_ addressing other change requests, please explicitly document whether the option applies to `Cache-Control: proxy-revalidate` responses.

src/cf.data.pre
```diff
@@ -6445,6 +6445,7 @@ DOC_START
 		 reload-into-ims
 		 ignore-reload
 		 ignore-no-store
+		 ignore-must-revalidate
```
rousskov, line 6448: What happens to folks that had ignore-must-revalidate in their old Squid configurations when they update to this new Squid code _without_ going through any Squid version that does not have ignore-must-revalidate? Will there be any changes in relevant Squid behavior (while their configuration remains unchanged)? If yes, we should disclose those changes, either in the PR description or Squid release notes. Personally, I prefer the latter.

rousskov (changes requested):
Thank you for offering this enhancement!

FYI: PR description (i.e. the future official commit message) will need to be updated, but I think we should come back to that after code changes are settled. I interpret the overall scope of this PR as "adding ignore-must-revalidate or a similar option" and ignoring secondary PR description problems (for now).

rousskov:
> I'm not sure what the M-failed-description tag means, presumably some sort of formatting issue?

Yes, probably. Please see [Anubis](https://github.com/measurement-factory/anubis/blob/master/README.md#pull-request-labels) docs about pull request labels (linked from each Anubis label but GitHub makes it difficult to see/use those links). I would not worry about that formatting for now, but that is your call. FWIW, the PR title/description will become the official commit message when your PR merges; any decent editor will let you quickly reformat PR description text to follow typical Git commit message guidelines ([documented](https://github.com/measurement-factory/anubis/blob/master/README.md#pr-description) and enforced by Anubis).

src/refresh.cc
compholio: Thanks for the review!

Maybe I'm missing something, but doesn't `Cache-Control: max-age` get processed [after this](https://github.com/squid-cache/squid/blob/master/src/refresh.cc#L406) (checked independently)?

It seems to me like the easiest thing would be to:
1. Document that this option does not affect `Cache-Control: proxy-revalidate`, presumably someone who has turned on that option knows what they are doing (and if they don't, a separate option probably makes sense)
3. Allow the option to apply to cached information from before the option was enabled, to match the original behavior (to minimize unexpected changes, as mentioned in your other comment)
4. Document this behavior so that it is clear that the option is about _revalidation_ of cached objects, not really the original caching

Proposed documentation:
> ignore-must-revalidate ignores any ``Cache-Control: must-revalidate`` headers received from a server. This means that previously cached content (potentially cached before the option was enabled) will be reused instead of checking whether the server has updated content. Doing this VIOLATES the HTTP standard. Enabling this feature could make you liable for problems which it causes. Please note this option does not change the behavior of ``Cache-Control: proxy-revalidate``, but does take affect for authenticated transactions where ``Cache-Control: must-revalidate`` or ``Cache-Control: no-cache`` is set (without parameters) for backwards-compatibility with the pre-v4 behavior.

Note: The authenticated transaction part is not relevant in my use case, but I figured in re-introducing this option it would be best to have it behave the same way it did before.

src/cf.data.pre
yadij: Better yet, please restrict the options behaviour to only ignoring ```must-revalidate```. Leaving other controls being obeyed correctly until we have an explicit case appear needing them to be violated.

src/refresh.cc
yadij: FYI: this is a generic check that the object is stale and therefor needs revalidating. If it gets removed the objects will likely go through a full re-fetch which is counter to intentions.
 Its relationship to ```must-revalidate``` is through how the revalidate requirement implies immediate staleness. If you can identify the code converting response ```Cache-Control:must-revalidate``` into the ```staleness=0``` value that would likely be a better place to put the override check.

I would prefer splitting this complicated if-condition up so the effects of the relevant cache-control values can be clearly seen and this ignore/override can be done to them individually. Though that may require expanding out the ENTRY_REVALIDATE_STALE flag meanings.

yadij (changes requested):
Alex has pointed out most of the major issues already. I've just added more to the pile.

> The "ignore-must-revalidate" option. The option was not actually broken, but it is not about
> the storage of objects - it is about whether they need to be retrieved again for new requests.

I think you have misunderstood the original removal. Unless the ```Cache-Control:no-store``` is present an object can always be stored. The option was removed because the useful part of its previous implementation are now the default behaviour of HTTP/1.1 Squid and all that remained was the security bugs and broken servers.

> For example, I encountered a university download service that has set both "Expires: 0" and "Cache-Control: must-revalidate" for unchanging content.

[FWIW; have you attempted to contact that one broken service and inform them of the problem? it is often more useful to get the few broken servers fixed than to open the whole Internet up to security vulnerabilities just to let them stay broken.]

> The refresh_pattern "override-expires" option will allow you to store a cache of this content, 

No. The mere existence of HTTP/1.1 ```Cache-Control``` header makes the HTTP/1.0 ```Expires``` header be ignored by default.

 So the ```Cache-Control:must-revalidate``` alone is determining what happens. Not your refresh_pattern.

> but due to the "Cache-Control: must-revalidate" header the proxy server will always re-fetch

What ```must-revalidate``` does is tell Squid the object should be considered stale when it arrived and that a short/fast revalidate is all that is needed to re-use it. Proxy will **always** fetch something for stale content, period.

The real problem is that example server responding to revalidate requests by delivering the whole object in full even when it has not changed.

compholio:
> Alex has pointed out most of the major issues already. I've just added more to the pile.
> 
> > The "ignore-must-revalidate" option. The option was not actually broken, but it is not about
> > the storage of objects - it is about whether they need to be retrieved again for new requests.
> 
> I think you have misunderstood the original removal. Unless the `Cache-Control:no-store` is present an object can always be stored. The option was removed because the useful part of its previous implementation are now the default behaviour of HTTP/1.1 Squid and all that remained was the security bugs and broken servers.

I was attempting to have the feature behave the same way it did before, but it sounds like it would do so even without the `!REFRESH_OVERRIDE(ignore_must_revalidate)` bit.

> > For example, I encountered a university download service that has set both "Expires: 0" and "Cache-Control: must-revalidate" for unchanging content.
> 
> [FWIW; have you attempted to contact that one broken service and inform them of the problem? it is often more useful to get the few broken servers fixed than to open the whole Internet up to security vulnerabilities just to let them stay broken.]

The "contact webmaster" link is broken, so we'll have to do a little digging to figure out who to talk to.  We meet with the science group associated with the site on occasion, so they may be able to point us in the right direction.  Is there a way that broken sites can be addressed that doesn't cause you concern about security vulnerabilities?  Would moving the location of the fix to wherever `Cache-Control:must-revalidate` sets `staleness=0` address your concern (your other comment)?

> > The refresh_pattern "override-expires" option will allow you to store a cache of this content,
> 
> No. The mere existence of HTTP/1.1 `Cache-Control` header makes the HTTP/1.0 `Expires` header be ignored by default.
> 
> So the `Cache-Control:must-revalidate` alone is determining what happens. Not your refresh_pattern.

I apologize, the `store-stale` option was also set and I misunderstood which option was responsible for caching of the file.  With just the `store-stale` option set the server caches the file, but the proxy server always returns `TCP_REFRESH_MODIFIED` for additional requests without changing the behavior of the revalidation.

> > but due to the "Cache-Control: must-revalidate" header the proxy server will always re-fetch
> 
> What `must-revalidate` does is tell Squid the object should be considered stale when it arrived and that a short/fast revalidate is all that is needed to re-use it. Proxy will **always** fetch something for stale content, period.
> 
> The real problem is that example server responding to revalidate requests by delivering the whole object in full even when it has not changed.

Yes, it would be great if the server did "the right thing" such that this was not necessary.

src/cf.data.pre
rousskov: I support Amos suggestion about restricting this option to `must-revalidate` (_if_ that can be correctly implemented without heroic efforts!), but I think this particular change request does not depend on that decision -- we should explicitly document whether this option applies to `proxy-revalidate` regardless of whether it applies, especially because it _did_ apply before it was removed in 064679e.

src/http.cc
```diff
@@ -404,7 +404,7 @@ HttpStateData::reusableReply(HttpStateData::ReuseDecision &decision)
             // HTTPbis WG verdict on this is that it is omitted from the spec due to being 'unexpected' by
             // some. The caching+revalidate is not exactly unsafe though with Squids interpretation of no-cache
             // (without parameters) as equivalent to must-revalidate in the reply.
-        } else if (rep->cache_control->hasNoCacheWithoutParameters()) {
+        } else if (rep->cache_control->hasNoCacheWithoutParameters() && !REFRESH_OVERRIDE(ignore_must_revalidate)) {
             debugs(22, 3, "Authenticated but server reply Cache-Control:no-cache (equivalent to must-revalidate)");
```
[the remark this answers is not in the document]
rousskov: I agree with Amos. Please note that, unlike old 064679ea374d2f58ae4660d9a2af213b9be24bba code, in modern code, `Cache-Control:no-cache` sets ENTRY_REVALIDATE_ALWAYS (rather than ENTRY_REVALIDATE_STALE that this PR is working on).

Before ignore-must-revalidate was removed in 064679ea374d2f58ae4660d9a2af213b9be24bba, the two flags were the same -- ENTRY_REVALIDATE_ALWAYS did not exist back then -- so this `if` statement had no choice but treat ignore-must-revalidate specially. We split ENTRY_REVALIDATE into two flags in fa83b76. Now, this `if` statement does not need to worry about ignore-must-revalidate.

```C++
if (ccNoCacheNoParams || ccPrivate)
    EBIT_SET(entry->flags, ENTRY_REVALIDATE_ALWAYS);
else if (ccMustRevalidate || ccSMaxAge)
    EBIT_SET(entry->flags, ENTRY_REVALIDATE_STALE);
```

src/refresh.cc
rousskov: > Alex: Sorry, I should have said `s-maxage` not `max-age`...

Hm... I wrote that [comment](https://github.com/squid-cache/squid/pull/1196#discussion_r1038519601) a few days ago, but only received an email notification about it now, after adding a couple of fresh comments to other change requests. It looks like my comment was stuck for all those days (in a pending state?). GitHub still shows the correct (from my point of view) comment order in this change request, but my comment has a much later timestamp than the subsequent Amos' comment. Anyway, I think that stuck/resurfaced comment is still valid. Sorry about the mess!

yadij (changes requested):
This PR is predicated on a misunderstanding about how `Expires` operates.

The HTTP Specification requires that when a `Cache-Control` header is provide the `Expires` header is to be ignored. It does not matter whether your configuration says that value X is to be used instead of `Expires: Y` when that header is being ignored.

Please be aware that what `refresh_pattern ... override-expires ignore-must-revaliate` means when applied to the semantics of
```
 Expires: 0
 Cache-Control: must-revalidate
```
is that Squid should operate as if it had received an **empty** `Cache-Control` header. It **does not** mean use the `Expires` value.

yadij:
@compholio are you intending to continue with this PR? if not I think we should close the PR to help clear up the review queue.

PR #1197, opened 2022-12-03
Update Host, Via, and other headers in-place when possible

wweric:
This change optimizes Host header synchronization code that usually does
not need to change the Host header field but was always deleting the old
field and generating/appending a new one. Appending Host also violated
an RFC 9110 section 7.2 rule (on behalf of the user agent whose Host we
were rewriting): "A user agent that sends Host SHOULD send it as the
first field in the header section".

Also, removing old header fields and appending new ones introduced HTTP
header changes that broke otherwise working peers. For example, some
origins denied Squid-proxied messages exclusively due to an unusual
(from those origins point of view) Host header position. This change
reduces (but does not eliminate) such unnecessary and risky changes.

### DIFF
--- a/src/servers/Http1Server.cc
+++ b/src/servers/Http1Server.cc
@@ -189,10 +189,13 @@ Http::One::Server::buildHttpRequest(Http::StreamPointer &context)
     // some code still uses Host directly so normalize it using the previously
     // sanitized URL authority value.
     // For now preserve the case where Host is completely absent. That matters.
-    if (const auto x = request->header.delById(Http::HOST)) {
-        debugs(33, 5, "normalize " << x << " Host header using " << request->url.authority());
-        SBuf tmp(request->url.authority());
-        request->header.putStr(Http::HOST, tmp.c_str());
+    if(request->header.findEntry(Http::HOST)){
+        SBuf domain(request->url.authority());
+        HttpHeaderEntry* host = new HttpHeaderEntry(Http::HOST, NULL, domain.c_str());
+        if(int x = request->header.delById(Http::HOST)){
+            debugs(33, 5, "normalize " << x << " Host header using " << request->url.authority());
+            request->header.insertEntry(host);
+        }
     }
 
     // TODO: We fill request notes here until we find a way to verify whether
--- a/src/servers/Http1Server.cc
+++ b/src/servers/Http1Server.cc
@@ -189,12 +189,14 @@ Http::One::Server::buildHttpRequest(Http::StreamPointer &context)
     // some code still uses Host directly so normalize it using the previously
     // sanitized URL authority value.
     // For now preserve the case where Host is completely absent. That matters.
-    if(request->header.findEntry(Http::HOST)){
-        SBuf domain(request->url.authority());
-        HttpHeaderEntry* host = new HttpHeaderEntry(Http::HOST, NULL, domain.c_str());
-        if(int x = request->header.delById(Http::HOST)){
-            debugs(33, 5, "normalize " << x << " Host header using " << request->url.authority());
-            request->header.insertEntry(host);
+    if(request->header.has(Http::HOST)){
+        String host = request->header.getById(Http::HOST);
+        if(host.cmp(request->url.authority().c_str()) != 0){
+            if (const auto x = request->header.delById(Http::HOST)) {
+                debugs(33, 5, "normalize " << x << " Host header using " << request->url.authority());
+                SBuf tmp(request->url.authority());
+                request->header.putStr(Http::HOST, tmp.c_str());
+            }
         }
     }
 
--- a/src/servers/Http1Server.cc
+++ b/src/servers/Http1Server.cc
@@ -189,16 +189,8 @@ Http::One::Server::buildHttpRequest(Http::StreamPointer &context)
     // some code still uses Host directly so normalize it using the previously
     // sanitized URL authority value.
     // For now preserve the case where Host is completely absent. That matters.
-    if(request->header.has(Http::HOST)){
-        String host = request->header.getById(Http::HOST);
-        if(host.cmp(request->url.authority().c_str()) != 0){
-            if (const auto x = request->header.delById(Http::HOST)) {
-                debugs(33, 5, "normalize " << x << " Host header using " << request->url.authority());
-                SBuf tmp(request->url.authority());
-                request->header.putStr(Http::HOST, tmp.c_str());
-            }
-        }
-    }
+    if (request->header.has(Http::HdrType::HOST))
+        request->header.updateOrAddStr(Http::HdrType::HOST, request->url.authority());
 
     // TODO: We fill request notes here until we find a way to verify whether
     // no ACL checking is performed before ClientHttpRequest::doCallouts().
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1029,10 +1029,7 @@ HttpHeader::addVia(const AnyP::ProtocolVersion &ver, const HttpHeader *from)
         if (!strVia.isEmpty())
             strVia.append(", ", 2);
         strVia.append(buf);
-        // XXX: putStr() still suffers from String size limits
-        Must(strVia.length() < String::SizeMaxXXX());
-        delById(Http::HdrType::VIA);
-        putStr(Http::HdrType::VIA, strVia.c_str());
+        updateOrAddStr(Http::HdrType::VIA, strVia);
     }
 }
 
@@ -1165,6 +1162,9 @@ HttpHeader::updateOrAddStr(const Http::HdrType id, const SBuf &newValue)
 {
     assert(any_registered_header(id));
     assert(Http::HeaderLookupTable.lookup(id).type == Http::HdrFieldType::ftStr);
+
+    // XXX: HttpHeaderEntry::value suffers from String size limits
+    Must(newValue.length() < String::SizeMaxXXX()); // TODO: Assure() in master/v6
 
     if (!CBIT_TEST(mask, id)) {
         auto newValueCopy = newValue; // until HttpHeaderEntry::value becomes SBuf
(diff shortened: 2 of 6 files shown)


rousskov (changes requested):
Thank you for trying to reduce unwanted header mangling introduced by Squid. It is an old known problem.

> The purpose of this update is to insert the sanitized Host: header back on top the header list

The original Host header may not be "on top of the list". Yes, it is on top in many (perhaps even in the vast majority of) use cases, but it is located elsewhere in some cases. The changes will fix many use cases but might also break some. At the very minimum, we should be upfront about it.

There are also implementation problems, but we can (and, IMO, should) put them aside _for now_. We will help you resolve them later. Let's finalize the overall solution scope/plan first.

Assuming you have a specific use case where this or a similar code change fixes the problem that you can reproduce, I would like to propose the following plan:

1. If the original/received and Squid-adjusted Host header values in your use case are identical, then we should focus on _not_ updating the Host header when such an update is unnecessary. Such a change can be sold as a useful performance optimization that carries virtually no tradeoffs baggage and does not venture into complex/controversial areas. I can help you with these code changes.
2. Otherwise, let's work on adjusting your code so that the original Host header (if any) is updated in-place (if possible) rather than removed and then added. Again, I can help you with the actual code changes here, but they are likely to be more complex/larger than those in item 1.

If you do not have a specific use case where putting the new Host header on top fixes the problem, then we could still do item 1 (because it is a generally useful change!), but it may be best to leave this improvement to the time when the primary developer has access to such a use case (because a comprehensive solution may look very different than header-specific hacks).

yadij:
FTR; RFC 9110 section 7.2:   "agent that sends Host SHOULD send it as the first field in the header section of a request"

While this is technically a requirement on agents generating requests. We do expect Squid to fixup small issues like this and emit fully compliant messages.

src/servers/Http1Server.cc
```diff
@@ -189,10 +189,13 @@ Http::One::Server::buildHttpRequest(Http::StreamPointer &context)
...
-    if (const auto x = request->header.delById(Http::HOST)) {
-        debugs(33, 5, "normalize " << x << " Host header using " << request->url.authority());
-        SBuf tmp(request->url.authority());
-        request->header.putStr(Http::HOST, tmp.c_str());
+    if(request->header.findEntry(Http::HOST)){
+        SBuf domain(request->url.authority());
+        HttpHeaderEntry* host = new HttpHeaderEntry(Http::HOST, NULL, domain.c_str());
+        if(int x = request->header.delById(Http::HOST)){
+            debugs(33, 5, "normalize " << x << " Host header using " << request->url.authority());
+            request->header.insertEntry(host);
```
yadij, line 197: Please add a reference to the RFC requirement about these.
```suggestion
            // RFC 9110 section 7.2: agent that sends Host SHOULD send it as the first field in the header section
            request->header.insertEntry(host);
```

yadij (changes requested):

src/servers/Http1Server.cc
wweric: In regards to a specific case, 'linuxconfig.org' for example is blocking the request coming from Squid that is being used as a cache server in my situation. 

I became curious and inspected which the only noted difference between raw browser request and going through Squid was the Host header position in the header list, made the submitted adjustment, and it seems to have done the trick. 

Another thing keeping the Host header 'on top' seems to overcome is getting CAPTCHA by sites such as 'linuxquestions.org' for example as well as a handful of others.

Hard to believe that simply the Host header position is triggering such things, but so far, it seems to be exactly the issue unless I'm tripping.

Regarding not updating the Host header if not needed, agreed, leave it be, but in regards to updating in place the thing that crossed my mind is along the lines of memory addressing/allocation becoming jacked. Not sure as need to learn more about Squid code, just came up with this easy fix which seems to overcome the above stated issue which can be a real nuisance sometimes aside from the fact that Squid seems to do an awesome job as a cache server otherwise. 

I will admit that don't totally understand the reason for the Host header being mangled in the first place?

wweric: If it is being mangled here simply for sanitation purposes, then seems like it could be done way before this point which would ultimately eliminate the need for the code section here in the first place.

Personally need to dig further into Squid code as only took a few hours to track this down after getting tired of being blocked/checked by some sites which seems an issue that could deter others from wanting use it if coming across the same issue. 

Again, so far, other than this issue, Squid seems to do a rocking job as a cache server.

yadij: There are two reasons for mangling the Host header.

1) to ensure that it is valid when sent. If the client presented an invalid header, or if Squid internally changed the destination URL. The latter may occur at any point up to and including peer selection (cache_peer forceddomain option).

2) to avoid old and buggy squid code. This is what the existing comment refers to I am not sure if those bugs still exist.

It is mangled here because this function is where the outbound HTTP headers are initially created. Any previous code would be altering Squid's view of the clients request, which is semantically supposed to be a different message and needs preserving for logging later.

rousskov:
> Eric: the only noted difference between raw browser request and going through Squid was the Host header position in the header list

AFAICT, the discussion so far makes [item 1](https://github.com/squid-cache/squid/pull/1197#pullrequestreview-1203585060) the best next step: Focus on _not_ updating the Host header when such an update is unnecessary.

@wweric, I can help implementing the corresponding code changes (in your git branch), but do you want to take the first stab at them?

> Amos: We do expect Squid to fixup small issues like this

I disagree that we expect Squid to move the Host header to the top when forwarding requests, especially since popular servers/CDNs use the Host position for fingerprinting clients. Moreover, if we do start "fixing" this, we should not do it in Http::One::Server -- a class (partially) responsible for _receiving_ HTTP requests that _does not even see_ some of the HTTP requests received by Squid. These difficult decisions/discussions is one of the reasons we should focus on the above simple/non-controversial "item 1" solution instead.

wweric:
@rousskov, do you mean to instead check for a match between the Host header and request->url.authority() which if true then simply leave it alone, if not, then use the original code that deletes and appends Vs the update I've suggested of delete and insert?

rousskov:
> @rousskov, do you mean to instead check for a match between the Host header and request->url.authority() which if true then simply leave it alone, if not, then use the original code that deletes and appends Vs the update I've suggested of delete and insert?

That would be one possible first step. If you make that step, I can help with a followup optimization that does not delete when a delete is not necessary (and also inserts instead of appends). I have not checked all the details, but I believe that the _final_ PR code will be a new HttpHeader method that achieves the correct result with no unnecessary updates, no unnecessary deletions/insertions, and a single scan of all stored header fields.

Another possible first step is to keep the matching Host header field intact if possible (as you said) and delete-and-insert otherwise (similar to the current PR code but with fixes that, among other things, make it compile). This step will need the same kind of followup that I described above (which I can help with).

It is up to you how far you want to take this. If you stop before the finish line, I am happy to finish your work. Low-level HttpHeader manipulation is tricky, and I do not want you to suffer trying to navigate all the mines in that code (unless you want to ;-).

wweric:
@rousskov I will take a stab at it as thinking that the following steps are along the lines of what you are saying:

1. Check if Host header exists
2. If so, compare against request->url.authority()
3. If they match, leave it alone
4. If no match, then use original code which deletes and appends value of #2

src/servers/Http1Server.cc
rousskov: @yadij, I added that (important) RFC quote to the PR description (for now). I do not think we should add it to the code itself for several reasons, in no particular order:

1. Even if this code enforces (or preserves) the correct Host header position, Squid often inserts another header before Host. We have to do more to actually enforce that RFC SHOULD.
2. I doubt it is a good idea to move the Host header to the correct position. If some bot is attacking a CDN while hiding behind Squid, and that CDN can fingerprint that bot by that bad Host header position, let them block bot+Squid combo. The opposite "let's fix what we can" strategy will only result in them fingerprinting and blocking _Squid_ instead (in all environments) -- a far worse outcome! As you know, these little "fixes" sometimes misfire for unexpected peer compatibility reasons, requiring configuration options or additional compensating hacks. I think preserving the user agent Host header position, good or bad, is an overall better strategy, especially since that position is correct in the vast majority of cases.
3. insertEntry() is very expensive performance-wise. Its push_front() behavior is a surprise to many developers. It is used incorrectly by some code. We should work towards removing it completely.

If you insist, we could add a source code comment with that RFC quote and augment it with similar caveats/explanations.

rousskov (approved):
I am done with this PR adjustments. @wweric, please review PR title/description/code and make any adjustments you feel are necessary (including, but not limited to, reverting my changes). Please also test that the changes still address your specific use case.

While working on this code, I noticed that this PR is against Squid v5! It will be up to @yadij whether to admit this change into v5. Normally, changes that apply to both Squid master and v5 branches are posted as pull requests against the master branch _first_. Once merged, they are cherry-picked into (or backported to) v5 if they are allowed to go into that "stable" branch.

@wweric, if you want these changes to be in master/v6, please either post the corresponding PR or rebase this PR branch and adjust this PR metadata to target master/v6. I recommend the former (and this v5 PR/branch can then stay in case a v5 backport is needed). If you cannot do either, then please let us know, and we will find the time to open a new master/v6 pull request (with you as the author).

yadij (approved):
Looks okay to me.

yadij:
I'm holding this as waiting for the master branch PR to go in first. We also need a bugzilla record with a clear precise of the issue being fixed to reference in the title for the v5 behaviour changelog and release announcement.

wweric:
@rousskov, I'm having build problems with master branch for some reason as not all the /libexec folder code is building, but the /sbin/squid exec compiles with no warnings/errors. It will try to run, but exits after processing config file due to /libexec/log_file_daemon not being present as a starting error.

Have tested the exact same changes with v5.7 and seems to work good maintaining the goal of not being blocked/checked.

Regarding the build issue, not sure if it's issue on my end or something in code as I'm not very good with complicated makefiles. Going to try and figure it out unless you already may have idea?

New pull #1203

PR #1200, opened 2022-12-05
Optimize RefCount comparison with raw pointers

sameer-here:
GCC v12 and Clang v12 tests show that compilers cannot avoid creating
temporaries when comparing RefCount objects with arbitrary pointers.

Also fixes (not yet supported) C++20 build. C++20 automatically reverses
equality comparisons, triggering an ambiguity:

    src/esi/Esi.cc:1920:26: error: ambiguous overload for 'operator=='
        (operand types are 'ESISegment::Pointer' {aka
        'RefCount<ESISegment>'} and std::nullptr_t)
        assert (output->next == nullptr);

    src/esi/Esi.cc:77:6: note: candidate: 'bool operator==(const
        ESIElement*, const Pointer&)' (reversed)
 
    src/base/RefCount.h:82:10: note: candidate: 'bool
        RefCount<C>::operator==(const RefCount<C>&) const [with C =
        ESISegment]'

### DIFF
--- a/src/base/RefCount.h
+++ b/src/base/RefCount.h
@@ -87,12 +87,16 @@ public:
         return p.p_ != p_;
     }
 
-    bool operator == (C const * p) const {
+    template <class Other>
+    bool operator ==(const Other * const p) const
+    {
         return p == p_;
     }
 
-    bool operator != (C const * p) const {
-        return p!= p_;
+    template <class Other>
+    bool operator !=(const Other * const p) const
+    {
+        return p != p_;
     }
 
 private:
--- a/src/base/RefCount.h
+++ b/src/base/RefCount.h
@@ -87,6 +87,14 @@ public:
         return p.p_ != p_;
     }
 
+    bool operator == (C const * p) const {
+        return p == p_;
+    }
+
+    bool operator != (C const * p) const {
+        return p!= p_;
+    }
+
 private:
     void dereference(C const *newP = nullptr) {
         /* Setting p_ first is important:


src/base/RefCount.h
```diff
@@ -87,6 +87,14 @@ class RefCount
         return p.p_ != p_;
     }
 
+    bool operator == (C const * p) const {
+        return p == p_;
+    }
+
+    bool operator != (C const * p) const {
+        return p!= p_;
+    }
```
rousskov, line 96: There are many ways to fix ESI build. To find the overall best solution, let's ignore ESI build problems for a second, especially since the proposed fix applies to the _entire_ Squid code.

> There's some comparisons that implicitly construct a RefCounted object from a pointer

This PR manually adds "simpler" code that an optimizing compiler, in theory, may be able to generate itself. The question is: Are modern compilers smart enough to optimize away creation of a temporary RefCount object?

* If yes, then we are writing (trickier than it may seem!) code that the compiler can generate for us. We should not do that.
* If no, then we should try to cover more raw pointer equality cases, similar to what we did for RefCount<C> creation from a raw pointer -- `C` in these new equality test operators should probably be a templated Other parameter.

Do you know (or can you find out experimentally) the answer to the above question?

rousskov (commented):
Thank you for making Squid more compatible with C++20!

src/base/RefCount.h
sameer-here: > The question is: Are modern compilers smart enough to optimize away creation of a temporary RefCount object?

I don't think the issue here is whether an optimizing compiler could optimize away the creation of a temporary RefCount object, it's that there's ambiguity about which of two candidates to use.  I believe this is due to how comparison operators [can be reversed/rewritten in C++20](https://stackoverflow.com/a/65947030).  We could also modify ESI.cc, as you mentioned.

rousskov: > > Are modern compilers smart enough to optimize away creation of a temporary RefCount object?
 
> the issue here is ... that there's ambiguity about which of two candidates to use.

AFAICT, you are thinking about a specific ESI-related build failure. I am not disagreeing with you, but let's ignore that specific build failure for a moment. Assume, for the sake of the argument, that ESI code does not exist.

> I believe this is due to how comparison operators [can be reversed/rewritten in C++20](https://stackoverflow.com/a/65947030).

Yes, we share that understanding. I am just trying to put that [ESI-introduced/specific] ambiguity aside (for a moment) when deciding whether/how we should improve RefCount. After all, all the correct prose aside, this PR is actually changing RefCount, not any ESI-specific code.

sameer-here: I see your point.  It seems the less invasive change is to just fix the couple places in Esi.cc where this issue occurs by using `.getRaw()` in the comparison, if you think that is a reasonable approach.

rousskov: > Alex: Are modern compilers smart enough to optimize away creation of a temporary RefCount object?

I built a [test](https://godbolt.org/z/rovMhfW1x) to answer my own question: The short answer is "no". Inlining RefCount shows that optimizing GCC v10.3:

* Avoids temporaries when comparing Pointer with nullptr (function test1() at the bottom of the test code), but 
* Cannot avoid temporaries when comparing Pointer with an arbitrary raw pointer (function test2() at the bottom of the test code). Search for "XXX: Added for testing" to remove your RefCount methods and see test2() assembly on the right exploding from a trivial 2-line pointer comparison to countless lines creating complex temporary objects.

Same result for GCC v12 and clang.

Thus, the proposed RefCount changes are an optimization that should be welcomed (regardless of ESI and C++20 issues). I am resolving this thread, adjusting PR title/description, and will post specific improvement suggestions.

rousskov, line 96: If possible, let's support other compatible raw pointer types, as we do elsewhere in RefCount:

```suggestion
    template <class Other>
    bool operator ==(const Other * const p) const
    {
        return p == p_;
    }

    template <class Other>
    bool operator !=(const Other * const p) const
    {
        return p != p_;
    }
```

The above suggestion avoids future/hypothetical build problems like the one illustrated by the following test: https://godbolt.org/z/To4zKjv59

The above suggestion also polishes formatting and const-correctness a little.

sameer-here: Thanks Alex, I appreciate it!

rousskov (approved):
I have updated PR title/description. Please check that you agree with the updated wording -- it will become the official commit message when this PR is merged. I also have one improvement request.

rousskov:
OK to test

yadij (approved):
FWIW; code checking against nullptr or NULL should be updated to use the ```operator !()```.

rousskov:
> code checking against nullptr or NULL should be updated to use the operator !().

Yes, in most cases it should, but those updates should not be done manually (unless the same code line has to be changed for other reasons). Thus, they are out of _this_ PR scope.

The problems addressed by this PR go beyond comparison with nil pointers.

rousskov:
@sameer-here, I am working on updating Squid [CONTRIBUTORS](https://github.com/squid-cache/squid/blob/50562b7/CONTRIBUTORS) file. Right now, our [script](https://github.com/squid-cache/squid/blob/50562b7/scripts/update-contributors.pl) adds a record for your contribution (i.e. this PR) using your GitHub profile name and email address: `sameer-here <76567789+sameer-here@users.noreply.github.com>`. If you would like that record to use a different name spelling or a different/no email, please let me know. If you prefer that your contribution is not explicitly acknowledged in the CONTRIBUTORS file, please let me know as well! Thank you.

sameer-here:
Hi @rousskov could you put me down as "Sameer Manchanda <smanchanda13@bloomberg.net>"?

rousskov:
> Hi @rousskov could you put me down as "Sameer Manchanda [smanchanda13@bloomberg.net](mailto:smanchanda13@bloomberg.net)"?

Done at #1225. Thank you.

PR #1208, opened 2022-12-09
Fix ACL type typo in req_header, rep_header key-changing ERRORs

rousskov:


yadij (commented):
Typo fix is okay. But we have a bug in that ignoring admin configured values in an ACL affects security and access control behaviour. Such critical level configuration flaws should be FATAL and halt Squid instead of allowing the broken security to continue operating.

rousskov:
> Typo fix is okay.

Please consider putting this PR on the fast track then, especially if you want this fix in v5.

> But we have a bug in that ignoring admin configured values in an ACL affects security and access control behaviour. Such critical level configuration flaws should be FATAL and halt Squid instead of allowing the broken security to continue operating.

Glad we agree! I [still](https://github.com/squid-cache/squid/pull/1209#discussion_r1044837971) plan to post a PR that will reject such configurations.

rousskov:
> I [still](https://github.com/squid-cache/squid/pull/1209#discussion_r1044837971) plan to post a PR that will reject such configurations.

Done in #1210.

yadij (approved):

PR #1209, opened 2022-12-09
Ignore changes in "note" ACL key; inform admins

rousskov:
acl banned note color green   # never matches before this change
    acl banned note weight heavy  # never matches after this change
    http_access deny banned

Admins expect the above configuration to deny both green transactions
and heavy transactions, but that is not what actually happens.

Before this change, Squid was denying only heavy transactions,
effectively (and silently!) ignoring the first "acl banned..." line.
Now, Squid denies only green transactions (with a configuration-time
ERROR) and otherwise ignores the second and any subsequent "acl
banned..." lines.

We now honor the first instead of the last "acl" line (for the same-name
note ACL) because:

* req_header/rep_header ACLs are doing that since 2016 commit a0b240c;
* it is easier to implement while still reporting the ignored location
  (which we do not need to remember or pass around in this case);
* we should just reject these configurations as invalid instead.

### DIFF
--- a/src/acl/NoteData.cc
+++ b/src/acl/NoteData.cc
@@ -49,6 +49,18 @@ ACLNoteData::parse()
 {
     char* t = ConfigParser::strtokFile();
     assert (t != nullptr);
+
+    if (!name.isEmpty() && name.cmp(t) != 0) {
+        debugs(28, DBG_CRITICAL, "ERROR: Ignoring conflicting 'note' ACL configuration:" <<
+               Debug::Extra << "honored annotation name: " << name <<
+               Debug::Extra << "ignored annotation name: " << t <<
+               Debug::Extra << "configuration location: " << ConfigParser::CurrentLocation() <<
+               Debug::Extra << "advice: To match annotations with different names, " <<
+               "use note ACLs with different names " <<
+               "(that may be ORed using an 'any-of' ACL.");
+        return;
+    }
+
     name = t;
     values->parse();
 }
--- a/src/acl/NoteData.cc
+++ b/src/acl/NoteData.cc
@@ -57,7 +57,7 @@ ACLNoteData::parse()
                Debug::Extra << "configuration location: " << ConfigParser::CurrentLocation() <<
                Debug::Extra << "advice: To match annotations with different names, " <<
                "use note ACLs with different names " <<
-               "(that may be ORed using an 'any-of' ACL.");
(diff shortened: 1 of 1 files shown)


rousskov:
Sample error message:

```
2022/12/09 15:49:49| ERROR: Ignoring conflicting 'note' ACL configuration:
    honored annotation name: color
    ignored annotation name: weight
    configuration location: /usr/local/squid/etc/squid.conf(8) acl
    advice: To match annotations with different names, use note ACLs with different names (that may be ORed using an 'any-of' ACL).
```

src/acl/NoteData.cc
```diff
@@ -49,6 +49,18 @@ ACLNoteData::parse()
...
+               Debug::Extra << "honored annotation name: " << name <<
+               Debug::Extra << "ignored annotation name: " << t <<
+               Debug::Extra << "configuration location: " << ConfigParser::CurrentLocation() <<
+               Debug::Extra << "advice: To match annotations with different names, " <<
+               "use note ACLs with different names " <<
+               "(that may be ORed using an 'any-of' ACL).");
+        return;
+    }
+
     name = t;
```
rousskov, line 64: IMO, configurations with misconfigured key-changing note, req_header, or rep_header ACLs should be rejected, but I think that change should be done (for all three ACLs at once) in a dedicated PR. This PR is still useful as a step that unifies the relevant note, req_header, or rep_header ACL behavior without changing req_header and rep_header.

However, if others insist, I can close this PR and post a PR that rejects the corresponding malformed configurations (for all three ACLs).

yadij: I agree with this being a CRITICAL level error, it has security implications on many installations.
By that same security reason critical configuration errors should not continue to let Squid run. I do insist on that.

This issue should be handled the same way as if the admin changed the type of an already named ACL.
Which is currently the old destruct mechanism (https://github.com/squid-cache/squid/blob/master/src/acl/Acl.cc#L235).

yadij (changes requested):

src/acl/NoteData.cc
rousskov: > Alex: if others insist, I can close this PR and post a PR that rejects the corresponding malformed configurations (for all three ACLs).

> Amos: I do insist on that.

Closing. Will post a configuration-rejecting PR.

PR #1210, opened 2022-12-11
Ban acl key changes in req_header, rep_header, and note ACLs

rousskov:
* req_header and rep_header ACLs; the key is the header-name argument:

    acl barred req_header X-1 bar
    acl barred req_header X-2 bar # never matches since commit a0b240c
    http_access deny barred

* the "note" ACL; the key is the annotation name argument:

    acl banned note color green  # never matches since commit 39baccc
    acl banned note weight heavy
    http_access deny banned

Squid did write a cache.log ERROR for req_header/rep_header key changes
but was silent about the preceding "note" ACL rules being ineffective
after a key change (e.g., Squid was not denying "green" requests above).

Now, Squid rejects configurations containing such ACL key changes.

Also significantly reduced the corresponding parsing code duplication.

### DIFF
--- a/src/ConfigParser.cc
+++ b/src/ConfigParser.cc
@@ -504,6 +504,38 @@ ConfigParser::regex(const char *expectedRegexDescription)
     return std::unique_ptr<RegexPattern>(new RegexPattern(pattern, flags));
 }
 
+void
+ConfigParser::SetAclKey(SBuf &keyStorage, const char *keyParameterName)
+{
+    // TODO: Refactor (distant) callers to store ACL name as CodeContext
+    // information, so that high-level parsing code can then report that context
+    // when reporting configuration parsing exceptions.
+    extern const char *AclMatchedName;
+
+    const auto newKey = strtokFile();
+    if (!newKey) {
+        throw TextException(ToSBuf("An acl declaration is missing a ", keyParameterName,
+                                   Debug::Extra, "ACL name: ", AclMatchedName),
+                            Here());
+    }
+
+    if (keyStorage.isEmpty()) {
+        keyStorage = newKey;
+        return;
+    }
+
+    if (keyStorage.caseCmp(newKey) == 0)
+        return; // no change
+
+    throw TextException(ToSBuf("Attempt to change the value of the ", keyParameterName, " argument in a subsequent acl declaration:",
+                               Debug::Extra, "previously seen value: ", keyStorage,
+                               Debug::Extra, "new/conflicting value: ", newKey,
+                               Debug::Extra, "ACL name: ", AclMatchedName,
+                               Debug::Extra, "advice: Use a dedicated ACL name for each distinct ", keyParameterName, " value ",
+                               " (and group those ACLs together using an 'any-of' ACL)."),
+                        Here());
+}
+
 CachePeer &
 ConfigParser::cachePeer(const char *peerNameTokenDescription)
 {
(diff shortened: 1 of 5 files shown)


rousskov:
Sample cache.log output:

```
2022/12/10 23:07:03| ERROR: configuration failure: Attempt to change the value of the header-name argument in a subsequent acl declaration:
    previously seen value: Cache-Control
    new/conflicting value: Surrogate-Control
    ACL name: controlled
    advice: Use a dedicated ACL name for each distinct header-name value  (and group those ACLs together using an 'any-of' ACL).
    exception location: ConfigParser.cc(536) SetAclKey
2022/12/10 23:07:03| FATAL: Bungled squid.conf line 14: acl controlled rep_header Surrogate-Control no-store
```

src/tests/stub_acl.cc
```diff
@@ -9,10 +9,12 @@
...
-#include "acl/Gadgets.h"
 
 #define STUB_API "acl/"
 #include "tests/STUB.h"
 
+#include "acl/Gadgets.h"
 size_t aclParseAclList(ConfigParser &, Acl::Tree **, const char *) STUB_RETVAL(0)
 
+#include "acl/Acl.h"
+const char *AclMatchedName = nullptr;
```
yadij, line 20: Please add this in alphabetical sorting order for the included filepath (eg "acl/A..." above "acl/G...").

src/ConfigParser.cc
```diff
@@ -504,6 +504,38 @@ ConfigParser::regex(const char *expectedRegexDescription)
...
+
+    if (keyStorage.caseCmp(newKey) == 0)
+        return; // no change
+
+    throw TextException(ToSBuf("Attempt to change the value of the ", keyParameterName, " argument in a subsequent acl declaration:",
+                               Debug::Extra, "previously seen value: ", keyStorage,
+                               Debug::Extra, "new/conflicting value: ", newKey,
+                               Debug::Extra, "ACL name: ", AclMatchedName,
+                               Debug::Extra, "advice: Use a dedicated ACL name for each distinct ", keyParameterName, " value ",
+                               " (and group those ACLs together using an 'any-of' ACL)."),
```
yadij, line 535: Extra whitespace in the display string due to incorrect string breaks.

Also we have possible confusion between this message using the word sequence "name value" to mean the config file characters for the "key" name, and documentation which uses the word "value" to represent the datum being checked against.

IMO we can just remove the word "value" without loosing any meaning.
```suggestion
                               Debug::Extra, "advice: Use a dedicated ACL name for each distinct ", keyParameterName,
                               " (and group those ACLs together using an 'any-of' ACL)."),
```

src/ConfigParser.cc
```diff
@@ -504,6 +504,38 @@ ConfigParser::regex(const char *expectedRegexDescription)
     return std::unique_ptr<RegexPattern>(new RegexPattern(pattern, flags));
 }
 
+void
+ConfigParser::SetAclKey(SBuf &keyStorage, const char *keyParameterName)
+{
+    // TODO: Refactor (distant) callers to store ACL name as CodeContext
+    // information, so that high-level parsing code can then report that context
+    // when reporting configuration parsing exceptions.
```
yadij, line 512: This TODO advises perpetuation of a bad design, please drop.

yadij (changes requested):
Since you have chosen to make this change for multiple ACL types please also include the external, user_cert, and ca_cert ACL types which have the same type of "name type key" syntax.

yadij (approved):
Thank you for addressing my requests. The reasons for omitting other ACL types from this update is fine.

rousskov:
> In summary, that ACLCertificateData conversion should happen in a dedicated followup PR

That PR is #1242.

PR #1211, opened 2022-12-13
Merge MemImplementingAllocator and Mem::Allocator

yadij:


src/mem/Allocator.h
```diff
@@ -54,13 +72,62 @@ class Allocator : public Interface
...
+     *
+     * Specific Optimizations:
+     * \li Objects have fixed-size,
+     *     so all calculations about bytes can be deferred.
+     * \li Objects pooled (if any) can be inferred,
+     *     so saved_free counting is unnecessary.
+     * \li Objects actively in use (if any) can be inferred,
+     *     so no inuse_count management on alloc() or freeOne().
+     */
+    struct FastCounter {
```
rousskov, line 119: The above justifications for having FastCounter do not seem to support its existence:

* "Objects have fixed-size, so all calculations about bytes can be deferred." True, but that deference can happen without FastCounter existence AFAICT. In fact, it does happen that way already, at least in one context: `getMeter().gb_allocated.bytes = getMeter().gb_allocated.count * obj_size`. We cannot defer the calculation of the _number_ of allocations (`count.allocs` in the current PR code), but this bullet/justification is not about that; it is about deferring _byte_ calculations.

* "Objects pooled (if any) can be inferred, so saved_free counting is unnecessary." Please rephrase to clarify whether this is about inferring the number of "currently pooled objects" or the number of "saved free() calls (or equivalent)". AFAICT, there is no way to compute saved_free based on allocs, saved_allocs, and freed: If allocs is 2, saved_allocs is 0, and freed is 2, then saved_free could be 2 (because deallocate() added both allocated objects to the pool, after they were allocated,  used, and returned) OR saved_free could could be 0 (because deallocate() freed both returned objects).

* "Objects actively in use (if any) can be inferred, so no inuse_count management on alloc() or freeOne()." True, but we can get the number of objects in use from meter.inuse which is already maintained by deallocate() methods. Thus, we are not saving any CPU cycles for this specific reason.

Please note that I did not check whether we need FastCounter. I would not be very surprised if we do need it. At this point, I am just saying that I do not see how the comment explaining _why_ we need matches code/actual needs.

src/mem/Allocator.h
```diff
@@ -54,13 +72,62 @@ class Allocator : public Interface
     /// XXX: Misplaced -- not all allocators have a notion of a "chunk". See MemPoolChunked.
     virtual void setChunkSize(size_t) {}
 
+    virtual bool idleTrigger(int shift) const = 0;
+
+    virtual void clean(time_t maxage) = 0;
+
+    /**
+     * Flush temporary counter values into the statistics held in 'meter'.
+     * Performs some CPU intensive calculations which have been deferred for performance.
```
rousskov, line 81:
```C++
void update(size_t items, size_t itemSize) {
    count += items;
    bytes += (items * itemSize);
}
```

The two increments an one multiplication above are hardly "CPU intensive calculations" worth "deferring for performance", especially since the bytes update is not really needed until the meters are aggregated and/or reported. Are you sure all this complexity is added for performance reasons?

src/mem/Allocator.h
```diff
@@ -29,22 +36,33 @@ class Allocator : public Interface
...
+        deallocate(obj);
+        ++count.freed;
+    }
 
     /// brief description of objects returned by alloc()
     virtual char const *objectType() const { return label; }
 
     /// the size (in bytes) of objects managed by this allocator
-    virtual size_t objectSize() const = 0;
+    size_t objectSize() const { return objectSize_; }
```
rousskov, line 62: Why not expose a constant obj_size data member? AFAICT, doing so will simplify/clarify the code (a long-term benefit) and significantly reduce the number of trivial code changes in this PR (a short term benefit).

I do not recommend doing that in _this_ PR (because it will add noise to an already noisy and complex changes), but we can also remove this method (after this PR adds a constant data member).

src/mem/Allocator.h
```diff
@@ -29,22 +36,33 @@ class Allocator : public Interface
      */
     virtual size_t getStats(PoolStats &) = 0;
 
-    virtual PoolMeter const &getMeter() const = 0;
+    /// get current statistics tracked for this allocator
+    virtual PoolMeter const &getMeter() const { return meter; }
+    virtual PoolMeter &getMeter() { return meter; }
```
rousskov, line 41: I think we should just make `meter` public because nobody overrides this virtual method, and I doubt we will need to override it in the foreseeable future. The proposed design is also dangerous -- it would be trivial for some future Allocator code to access `meter` instead of a very different meter accessible via getMeter()! Keeping the `meter` data member public will also reduce the footprint of this PR and improve performance of metering code.

If you are sure getMeter() might become overwritten in the future, then just comment out the `virtual` keyword for now.

src/mem/Allocator.h
```diff
@@ -72,6 +139,8 @@ class Allocator : public Interface
 
 private:
     const char *label = nullptr;
+    const size_t objectSize_ = 0;
```
rousskov, line 142: We may have discussed this before, but I find constants with two values odd and constants with bogus values confusing/dangerous. I suggest making both data members constant and rely on the constructor for their proper initialization:

```suggestion
    const char * const label;
    const size_t objectSize_;
```

rousskov (changes requested):
Thank you for moving MemImplementingAllocator into Mem::Allocator! This is a partial review in hope to help you make steady progress.

src/mem/Allocator.h
yadij: Er, these are not about FastCounter existence. They are justifications for **omitting** the counters that are obviously useful but missing from it.

src/mem/Allocator.h
yadij: AFAICT these counters elide 4*999 additions of objectSize(), 999 virtual table lookups of getMeter(), 2*999 pointer dereference calculations, and lag time loading the separate 'meter' object into CPU cache 999 times. The "fast" counters are members of the Allocator which puts them already in CPU memory as side effects of loading it for the unavoidable part of the allocation. At cost of 3 extra multiplications every 1000 allocation events.

So yeah, I'm pretty sure it is there for performance. If you think it is *not enough* gain, we can add a few more zeros to FlushLimit. I would rather not touch it until the membanger tests are refreshed.

src/mem/Allocator.h
yadij: Simply that this PR focus is for merging of the two classes. Polishing that is not directly necessary for that merge is out of scope.

src/mem/Allocator.h
yadij: That is polish which I am avoiding for this PR.

rousskov: This PR introduces a new virtual method and makes the "virtual" in the old const method unnecessary. This PR adds a new data member and makes that data member private (necessitating adding a non-const method and creating a conflict for the Allocator code -- which meter access avenue do I use?). Addressing these significant problems _created_ by this PR is clearly in this PR scope, and it is not "polish" that should be avoided in this PR, especially since there exist a trivial way to address these problems: make the new data member public (with the name matching the old one), remove the word "virtual" from the old const method, and not add the non-const method.

src/mem/Allocator.h
rousskov: > this PR focus is for merging of the two classes. Polishing that is not directly necessary for that merge is out of scope.

This PR _renames_ the moved public data member and makes it _private_. Those two decisions/actions are clearly in this PR scope. Thus, the requested adjustments (of those actions) are also in scope. They also improve the overall PR outcome and are trivial.

src/mem/Allocator.h
rousskov: > elide 4x999 additions of objectSize()

Where would we add objectSize() on every alloc()/freeOne() if, for example, these "fast" counters become a part of "slow" Meter or PoolMeter, removing FastCounters? As detailed elsewhere, byte calculations are indeed unnecessary until meters are merged or printed. We do not need to update a `gb_*` field on every alloc()/freeOne() call, regardless of FastCounters existence.

> 1x999 virtual table lookups of getMeter()

The `meter` is a member of this class so no virtual table lookups are necessary to get to "slow" Meter or PoolMeter fields.

> 2x999 pointer dereference calculations, and lag time loading the separate 'meter' object into CPU cache 999 times

The `meter` is a data member of this Allocator class, just like `counter` is. It is not more or less "separate" than `counter`.

In summary, all these claims of CPU optimizations seem to be gross exaggerations -- about the same amount of calculations would be performed if FastCounter did not exist (with its fields merged into Meter and/or PoolMeter, as needed).

Please note that this change request is not asking to remove FastCounter. At this point, I am only trying to establish the facts. We will decide what the best solution is after we understand what is actually going on in this code.

src/mem/Allocator.h
yadij: see https://github.com/squid-cache/squid/blob/master/src/mem/Pool.h#L139
and https://github.com/squid-cache/squid/blob/master/src/mem/Pool.h#L146

All problems you are stating as new exist in the MemImplementingAllocator current official API. This PR merged the two APIs

The 'meter' member was previously protected (not public) and existence of both accessors made that dangerous. A non-const pool child could access meter directly and alter it during a const method.

src/mem/Allocator.h
yadij: So are you standing by your claim "I do not recommend doing that in this PR"  ?? or actually insisting on the change?

src/mem/Allocator.h
yadij: > At this point, I am only trying to establish the facts. We will decide what the best solution is after we understand what is actually going on in this code.

Ah, "what" is an entirely different topic than "justify this".

The facts known are the documentation comment I wrote. These counters are a pre-existing performance optimization - simply that. Whether it is actually useful or not requires testing I am not about to do here - my understanding of its reason are as stated above.

If you want a better comment text, sure. But I am not intending to polish the logic of optimizations **right now**, halfway through a series of refactors, especially ones that I am not certain of the specific performance benefit/loss.

src/mem/Allocator.h
rousskov: > All problems you are stating as new exist in the MemImplementingAllocator current official API. This PR merged the two APIs

In a hypothetical scenario where a PR mechanically merges two classes without any other changes, the above argument would have been valid[^1]. This PR does not mechanically merge two classes without any other changes. This PR adds new types, renames fields, changes fields visibility, documents controversial APIs, etc. Those changes, once polished/adjusted, are the _right_ thing to do in principle -- the merger should improve code quality rather than make it worse, of course. However, the presence of those (good) changes invalidates "mechanical mergers should avoid API fixes" arguments as inapplicable. We are dealing with a careful manual elimination of an unwanted hierarchy member, not a mechanical merger of two classes.

> The 'meter' member was previously protected (not public) and existence of both accessors made that dangerous.

The old code was problematic. There is no need to preserve/move some of the old problems in/into Allocator. Implementing the trivial changes suggested in this PR will avoid/fix problems instead of preserving them (in some areas) and making them worse (in others).

> A non-const pool child could access meter directly and alter it during a const method.

A const method cannot alter a regular data member, regardless of that member visibility or originating class.

In general:

* A public data member is less dangerous than a combination of a private data member and a virtual accessor.
* A public data member is as dangerous as a public non-virtual modifier method.
* Non-virtual methods are better than virtual methods that are never overridden.

The trivial changes requested here will make the code better/safer. We should not waste time on finding reasons to avoid them.

[^1]: That argument would still not be enough to justify creating bad merged API in most cases (including this one), but at least it would have been _valid_ on the same "mechanical" level.

src/mem/Allocator.h
rousskov: This change request discussed two changes:

* "expose a constant obj_size data member"; please implement that change in this PR
* "remove this method (after this PR adds a constant data member)"; I do not recommend doing that change in this PR

src/mem/Allocator.h
rousskov: > The facts known are the documentation comment I wrote.

The proposed documentation relies on an claim that these counters/actions are a performance optimization. There is currently no (shared) knowledge of that claim validity. In other words, that alleged fact has not been established yet. The already written documentation does not support that fact yet (due to problems detailed elsewhere).

> These counters are a pre-existing performance optimization

I see no evidence of those counters being (as in "existing to provide") a performance optimization.

If you prefer, we can ignore the lack of shared understanding regarding performance optimization status of these counters. AFAICT, that shared understanding is _not_ necessary to merge this PR. We will just need to remove some comments and adjust some names. Please let me know if I should switch to that strategy (instead of asking questions about these counters/actions in hope to correct their documentation/names).

rousskov (commented):
FWIW, I can address the pending change requests myself. Just let me know if you prefer not to spend time on that yourself.

src/mem/Allocator.h
yadij: > The proposed documentation relies on an claim that these counters/actions are a performance optimization. There is currently no (shared) knowledge of that claim validity. In other words, that alleged fact has not been established yet.

AFAICS we have both stated that there is some (non-zero) performance gain from them.

Are you aware of any raison d'être of the variables this PR puts into FastCounters which is not about performance?
If not please accept the claim so we can move forward.

rousskov: > Are you aware of any raison d'être of the variables this PR puts into FastCounters which is not about performance?

I am. Based on that knowledge, I have now suggested specific FastCounters changes in the change request about that class.

As far as _this_ change request is concerned, I suggest simply removing the second sentence of the method description. To move forward, we do not need to agree on the performance impact of this method (as long as neither of the conflicting opinions is documented as the established fact).

src/mem/Allocator.h
rousskov: To move forward despite the unresolved disagreements regarding performance, I suggest removing the new FastCounters class from the PR and simply documenting what each counter means, reusing and polishing PR-provided descriptions:

```C++
    /* the three counters below count events since the last flushing */

    /// the number of alloc() calls
    size_t alloc_calls = 0;

    /// the number of alloc() calls that returned a previously allocated object
    size_t saved_allocs = 0;

    /// the number of freeOne() calls
    size_t free_calls = 0; 
```

Replacing FastCounters (including its description) with the above snippet would address this change request.

We could also add a TODO regarding moving these counters to better locations and other improvements, but I doubt we can easily reach consensus regarding those improvements, so I am _not_ suggesting adding any TODOs here.

The data member names above preserve these counter names in the official code. These names are inconsistent and violate current naming guidelines, but we should not bloat this PR with noisy name changes of problematic counters IMO. These names can (and probably should) be changed in a separate PR focused on improving these counters. As far as these counters are concerned, let's limit this PR to moving these counters to the parent class (and adjusting the code accordingly).

src/mem/Allocator.h
yadij: Done in dc46e76be849276568edce5fd0f04e5279393990
I have done both changes because it makes no sense to expand the scope of this PR to polish the API, only to force an immediate followup PR to re-polish the polishing changes.

src/mem/Allocator.h
yadij: FTR, IMO exposing meter is useful to reduce virtual table size. The other arguments being presented do not make much sense.
To avoid arguing, done in 64d513cf92105fd06905964ae0177f57a6baf675

src/mem/Allocator.h
yadij: > > Are you aware of any raison d'être of the variables this PR puts into FastCounters which is not about performance?
> 
> I am. 

Please share technical detail/reasons instead of obtusely requiring design changes.

yadij:
Given the review insistence on applying polish to the merged API I have merged what was going to be a followup step-8 PR.

src/mem/Allocator.h
rousskov: In the future, if you want to avoid arguing, please avoid posting your own arguments (_any_ arguments and especially arguments that can be misinterpreted as an attack on others!) just before implying that no followup is wanted. Adding that "to avoid arguing" phrase is enough to both imply that no followup is wanted and to record one's disagreement with previous arguments.

src/mem/Allocator.h
rousskov: > Please share technical detail/reasons 

I have done that and even provided (a link to) those technical details/reasons expressed as C++ code. In summary, these data members count the number of various events since the last MemImplementingAllocator flushing event. These data members exist to provide that statistics. They are problematic, on several levels, but we do not need to agree on the exact set of those problems (and their solutions) to make progress with _this_ PR. I am not asking to add any TODOs or any controversial descriptions regarding these data members. We can choose to make progress _despite_ their problems and _despite_ our disagreements about their problems/benefits. For a more specific suggestion on how to move forward regarding those counters, please follow that link: https://github.com/squid-cache/squid/pull/1211#discussion_r1050810870

> instead of obtusely requiring design changes.

Please tone down your comments. We need to cooperate (or at least argue) using technical arguments. There is no need to use words that can be easily misinterpreted as personal attacks on others, especially when others are simply answering your question.

src/mem/Allocator.h
yadij: Done in [52ff68a](https://github.com/squid-cache/squid/pull/1211/commits/52ff68addc2194414a9f6335a8709513dbee799e)

src/mem/Allocator.h
rousskov: I want several changes. As far as **this** change request is concerned, I ([still](https://github.com/squid-cache/squid/pull/1211#discussion_r1050802407)) suggest simply removing the second sentence of the method description. To move forward, we do not need to agree on the performance impact of this method (as long as neither of the conflicting opinions is documented as the established fact).

src/mem/Allocator.h
```diff
@@ -29,38 +36,87 @@ class Allocator : public Interface
...
+    /// the number of malloc()/calloc() calls avoided since last flush
+    size_t countSavedAllocs = 0;
+
+    /// the number of calls to Mem::Allocator::freeOne() since last flush
+    size_t countFreeOne = 0;
+
+    // XXX: no counter for the number of free() calls avoided
+
+    /// brief description of objects returned by alloc()
+    const char *label;
```
rousskov, line 106: Please make this data member const if possible. It clarifies the code, and it is much easier to justify keeping it public this way:

```suggestion
    const char *const label;
```

rousskov (changes requested):
I hope we are almost done here: I see nothing particularly controversial in this set of change requests, and many of them are marked as optional.

src/mem/PoolChunked.cc
```diff
@@ -296,9 +300,9 @@ MemPoolChunked::~MemPoolChunked()
 {
     MemChunk *chunk, *fchunk;
 
-    flushMetersFull();
+    flushCounters();
+    assert(getInUseCount() == 0);
     clean(0);
-    assert(meter.inuse.currentLevel() == 0);
```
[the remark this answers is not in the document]
yadij: This was polish to make the clean+assert more consistent. Doing it the other way has some troubleshooting benefit, so done in [589a79f](https://github.com/squid-cache/squid/pull/1211/commits/589a79ff07302f4b13802982db1e37b740080757)

src/mem/Pool.cc
```diff
@@ -66,79 +65,36 @@ MemPools::setDefaultPoolChunking(bool const &aBool)
...
+        // Do not accumulate peak details as timing for those vary between pools.
+
+        // regenerate gb_* values from original pool stats
+        TheMeter.gb_allocated.count += pool->meter.gb_allocated.count;
+        TheMeter.gb_saved.count += pool->meter.gb_saved.count;
+        TheMeter.gb_freed.count += pool->meter.gb_freed.count;
+        TheMeter.gb_allocated.bytes += pool->meter.gb_allocated.bytes;
+        TheMeter.gb_saved.bytes += pool->meter.gb_saved.bytes;
+        TheMeter.gb_freed.bytes += pool->meter.gb_freed.bytes;
     }
```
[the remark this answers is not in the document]
yadij: Done in[cbc7d9d](https://github.com/squid-cache/squid/pull/1211/commits/cbc7d9d1a0d0a1c3fa123271659b981620bbdd34)

src/mem/PoolMalloc.cc
```diff
@@ -61,39 +62,29 @@ size_t
...
+MemPoolMalloc::MemPoolMalloc(char const *aLabel, size_t aSize) :
+    Mem::Allocator(aLabel, aSize)
 {
 }
 
 MemPoolMalloc::~MemPoolMalloc()
 {
-    assert(meter.inuse.currentLevel() == 0);
     clean(0);
+    assert(getInUseCount() == 0);
```
rousskov, line 87: I recommend keeping the old assert()/clean() order. I know you are trying to make the order in two allocator destructors _consistent_, and that is a noble goal, but I would not try to fix this particular inconsistency in _this_ PR because

* Moving misplaced chunking allocator assert to the right position would contradict the existing comments about that destructor. More (potentially controversial) work is needed to clean up that code. We have discussed that in another change request (that has been correctly addressed already by preserving the problematic assert in problematic code).

* Moving this _correctly placed_ assertion makes the code worse -- it is a step backward. This assert is needed not to double check clean() result -- clean() has nothing to do with in-use objects. This assert exists to make sure we do not destruct prematurely (i.e. while we are still needed by some allocated objects). Ideally, it should be separated from the clean() call by an empty line to emphasize that this is an allocator destruction precondition, not a clean() call precondition (but that polishing can be considered lying outside this PR scope).

In summary, it is better to preserve this inconsistency than to make this destructor code worse.

```suggestion
    assert(getInUseCount() == 0);
    clean(0);
```

I do not insist on this change because this whole code area is problematic, the unwanted move in this PR is not going to break anything badly, and I do not have the time to fight for every little improvement in this otherwise useful PR.

rousskov (approved):
Thank you for addressing my primary concerns. I did not check every low-level change, and I regret the amount of unnecessary/unwanted (IMO) changes in this PR, but I do support this "merger" overall.

PR #1212, opened 2022-12-13
Require C++17

rousskov:
Modern environments support C++17 well. We are wasting significant
amounts of time on emulating such basic C++17 features as std::optional.
We are writing worse code than we can because we lack access to such
C++14 and C++17 features like std::make_unique and std::byte.

The following language-standard-dependent TODOs were completed:

* Switched to [[fallthrough]] attributes.
* Replaced Optional with std::optional.
* Replaced EnableIfType with std::enable_if_t.
* Removed old C++11 type replacements (e.g., xuniform_int_distribution).
* Removed std::allocator declarations deprecated in C++17.
* Made ToSBuf() readable using a fold expression.
* Simplified compile-time "natural type" assertions.
* Used std::container::emplace*() return value where warranted.

Moved BUILDCXX-setting code after AX_CXX_COMPILE_STDCXX adds -std=c++17
to CXX (rather than CXXFLAGS!). Our cf_gen was built with compiler's
default C++ standard version! Ideally, BUILDCXX block should be lowered
further -- other macros might alter CXX -- but its CXXFLAGS adjustment
may affect subsequent code, warranting a careful/dedicated change.


configure.ac
```diff
@@ -75,8 +74,16 @@ AS_IF([test "x${enable_arch_native}" != "xno"],[
...
+user_cxx=`echo "$PRESET_CXXFLAGS" | grep -o -E "(-)std="`
+AS_IF([test "x$user_cxx" = "x"],[
+  # Check for C++17 compiler support
+  # May change CXX.
+  AX_CXX_COMPILE_STDCXX([17],[noext],[mandatory])
+])
+
+# Prerequisite: CXX has been finalized.
+# BUILDCXXFLAGS are defined later, after CXXFLAGS have been finalized.
 AC_ARG_VAR([BUILDCXX],[path to compiler for building compile-time tools. e.g. cf_gen])
```
rousskov, line 87: The diff may mislead here because I had to _move_ AX_CXX_COMPILE_STDCXX call from below BUILDCXX-setting code to above BUILDCXX-setting code for cf_gen.cc sources to compile using C++17. This PR does not change BUILDCXX-setting code (which may need more work) and does not change the AS_IF statement guarding AX_CXX_COMPILE_STDCXX. The PR description documents this bug-fixing move.

configure.ac
```diff
@@ -75,8 +74,16 @@ AS_IF([test "x${enable_arch_native}" != "xno"],[
...
 ])
 
-# might be cross-compiling.
-# NP: BUILDCXXFLAGS defined at the end of configure after CXXFLAGS fully known.
+# If the user did not specify a C++ version.
+user_cxx=`echo "$PRESET_CXXFLAGS" | grep -o -E "(-)std="`
+AS_IF([test "x$user_cxx" = "x"],[
+  # Check for C++17 compiler support
+  # May change CXX.
+  AX_CXX_COMPILE_STDCXX([17],[noext],[mandatory])
```
rousskov, line 82: We used to call AX_CXX_COMPILE_STDCXX_11() rather than the more general AX_CXX_COMPILE_STDCXX() macro that this PR is using. Autoconf-archive merged all individual version-testing macros into AX_CXX_COMPILE_STDCXX(). Modern individual macros simply call the general one. I think it is better to use the general macro because it saves us from embedding an extra m4 source file (each of those macros, even one-liners, have a dedicated m4 file).

src/store/SwapMeta.h
```diff
@@ -169,7 +168,6 @@ DeprecatedSwapMetaType(const RawSwapMetaType type)
         STORE_META_VALID = 7
     };
     return
-        // TODO: simplify with std::underlying_type_t when switching to C++14
```
rousskov, line 172: I removed this TODO after failing to simplify this code using std::underlying_type_t. This controversial code can be simplified, but that simplification is not related to C++14 AFAICT, so it is best to just drop that specific TODO.

lib/hash.cc
```diff
@@ -66,22 +66,22 @@ hash4(const void *data, unsigned int size)
         break;
     case 7:
         HASH4;
-    /* [[fallthrough]] */
+        [[fallthrough]];
     case 6:
```
rousskov, line 70: Please note that these formatting improvements were not manual -- astyle appears to handle `[[fallthrough]]` attributes better than trailing `case` comments.

rousskov (commented):
This switch has been discussed on [squid-dev](http://lists.squid-cache.org/pipermail/squid-dev/2022-December/009832.html).

Please note that Sempaphore CI Classic cannot test C++17 code without heroic efforts. Our CI scripts there will now skip testing this PR code. GitHub Actions do test it, of course.

@kinkie, I expect that some Jenkins adjustments will be necessary as well.

yadij (approved):

src/adaptation/icap/Xaction.cc
```diff
@@ -14,8 +14,8 @@
 #include "adaptation/icap/Launcher.h"
 #include "adaptation/icap/Xaction.h"
 #include "base/AsyncCallbacks.h"
+#include "base/IoManip.h"
```
kinkie, line 17: How did this creep in? It doesn't seem related to the changes in the file

rousskov: It looks unrelated, but it is required to provide our custom std:: optional "printing" operator that used to be in Optional.h.

kinkie: I had figured that out, but hadn't seen that used in the changed code

rousskov: IIRC, it is used for printing the following async callback (parameter). The code in question creates a call(back) object in icapLookupDnsResults() and, hence, has to debugs() it.

```C++
Adaptation::Icap::Xaction::dnsLookupDone(std::optional<Ip::Address> addr)
```

configure.ac
thesamesam: I think this came up downstream at https://bugs.gentoo.org/921688 where the right flags weren't used to build `cf_gen.cc`. We aren't shipping GCC 9 anymore but the user had it installed still.

rousskov: > I think this came up downstream at https://bugs.gentoo.org/921688 where the right flags weren't used to build `cf_gen.cc`.

```
checking whether x86_64-pc-linux-gnu-g++ supports C++17 features with -std=c++17... yes
checking for x86_64-pc-linux-gnu-g++ -std=c++17 ...
...
BUILDCXXFLAGS=-march=core2 -O2
...
libtool: compile:  x86_64-pc-linux-gnu-g++ -std=c++17 ... -march=core2 -O2 -c assert.cc ...
...
x86_64-pc-linux-gnu-g++ -march=core2 -O2 -o cf_gen ./cf_gen.cc ...
```

Yes, the above output looks like a sign of Squid bugs to me.

I hope this cf_gen building bug predates this PR and that this PR changes make fixing that bug a bit easier. The old comment you are responding to here claims that "this PR does not change BUILDCXX-setting code (which may need more work)". The corresponding commit 09835fe message has more on that subject.

It is possible that this bug has been fixed already. I see the expected cf_gen build command in some random build log of v6.6 code:

```
g++ -std=c++17 ... -o cf_gen ./cf_gen.cc
```

However, it is possible that the bug just does not manifest itself in that specific build environment (Ubuntu 22.04).

At any rate, if you can, please file a bug report with Squid [Bugzilla](https://bugs.squid-cache.org/). This old merged PR is not the best place to discuss these issues.

PR #1213, opened 2022-12-17
GCC 12 build error no match for 'operator<<'

yadij:
C++17 fold expression and std::optional printing hits these
errors expanding ostream other than debugs().

### DIFF
--- a/src/base/AsyncJobCalls.h
+++ b/src/base/AsyncJobCalls.h
@@ -11,6 +11,7 @@
 
 #include "base/AsyncJob.h"
 #include "base/CbcPointer.h"
+#include "base/IoManip.h"
 #include "debug/Stream.h"
 
 /**
--- a/src/base/PackableStream.h
+++ b/src/base/PackableStream.h
@@ -9,6 +9,7 @@
 #ifndef SQUID_SRC_BASE_PACKABLESTREAM_H
 #define SQUID_SRC_BASE_PACKABLESTREAM_H
 
+#include "base/IoManip.h"
 #include "base/Packable.h"
 
 #include <ostream>
--- a/src/store/SwapMetaView.cc
+++ b/src/store/SwapMetaView.cc
@@ -84,7 +84,7 @@ Store::SwapMetaView::checkExpectedLength(const size_t expectedLength) const
 {
     if (rawLength != expectedLength)
         throw TextException(ToSBuf("Bad value length in a Store entry meta field expecting a ",
-                                   expectedLength, "-byte value: ", *this), Here());
+                                   expectedLength, "-byte value: ", this), Here());
 }
 
 std::ostream &


src/store/SwapMetaView.cc
```diff
@@ -84,7 +84,7 @@ Store::SwapMetaView::checkExpectedLength(const size_t expectedLength) const
 {
     if (rawLength != expectedLength)
         throw TextException(ToSBuf("Bad value length in a Store entry meta field expecting a ",
-                                   expectedLength, "-byte value: ", *this), Here());
+                                   expectedLength, "-byte value: ", this), Here());
```
yadij, line 87:
```
../../../src/store/SwapMetaView.cc:86:35:   required from here
../../../src/sbuf/Stream.h:66:20: error: no match for 'operator<<' (operand types are 'std::basic_ostream<char>' and 'const Store::SwapMetaView')
   66 |     (out << ... << args);
      |     ~~~~~~~~~~~~~~~^~~~~
```

rousskov: The error message quoted in this change request is about "printing" SwapMetaView. We do have a "printing" operator for SwapMetaView:

```C++
/// writes a short human-readable summary of the given SwapMetaView object
std::ostream &operator <<(std::ostream &, const Store::SwapMetaView &);
```

It is the operator that should be (and, I presume, was) used in this code before C++17 migration AFAICT. I do not know what the correct bug fix is because I do not understand exactly why GCC v12 stopped "seeing" the existing "printing" operator declared for SwapMetaView in this context.

@yadij, does GCC report any "candidates" for this compilation failure? If it does, is there a SwapMetaView-related candidate? Those candidate messages often hint at the true cause of the problem. If you can share them (here or, if there are many candidates, in a GitHub gist), please do.

----

If we cannot find the right fix quickly enough, we can undo the corresponding code-readability improvement (see commit https://github.com/squid-cache/squid/pull/1212/commits/6730aa18c38091be26755017c2e7e25984bdc70c).

rousskov, line 87: The proposed code change "prints" a _pointer_ to the view (i.e. view object's memory address) instead of printing the view itself. This change introduces a runtime bug AFAICT.

src/base/AsyncJobCalls.h
```diff
@@ -11,6 +11,7 @@
 
 #include "base/AsyncJob.h"
 #include "base/CbcPointer.h"
+#include "base/IoManip.h"
```
rousskov, line 14: We should not force manipulators on all AsyncJobCalls users because most calls do not use these manipulators. In general, IoManip.h should be included in .cc files that use/need it, not header files (that do not).

src/base/PackableStream.h
```diff
@@ -9,6 +9,7 @@
 #ifndef SQUID_SRC_BASE_PACKABLESTREAM_H
 #define SQUID_SRC_BASE_PACKABLESTREAM_H
 
+#include "base/IoManip.h"
```
rousskov, line 12: To avoid duplicating discussions, we should revisit this _after_ src/base/AsyncJobCalls.h concern is addressed.

rousskov (changes requested):

src/store/SwapMetaView.cc
yadij: I have placed a copy of "make -k" build log at http://treenet.co.nz/projects/squid/gcc12.log

src/base/AsyncJobCalls.h
yadij: The problem for this is the code in the header file using the std::optional printer. Whenever those templates **attempt** to expand the error appears.

rousskov: > The problem for this is the code in the header file using the std::optional printer.

Some code that uses these templates instantiates them with std::optional. That code should worry about printing the arguments it instantiates. Such code is (currently) quite rare.

These templates and their many other users (i.e. the ones that do not instantiate the call templates with std::optional) should not be exposed to std::optional. std::optional is just one of the numerous possible call parameter types. We do not want to `#include` all of those parameter-declaring headers here. It is harmful and should not be necessary.

Thank you for posting the logs. I will try to figure out what is going on.

rousskov: _Status update_: I believe I figured out why GCC refuses to print `std::optional<Ip::Address>` in icap/Xaction.cc context and found a solution that does not require exposing every async call user to std::optional printing (and such -- the exposure list can only grow if we start going down this path). I will post an update with details after investigating the SwapMetaView problem.

rousskov:
_Status update_: I think I figured it out. Our code was not [ADL](https://en.cppreference.com/w/cpp/language/adl)-friendly, and GCC v12 apparently started enforcing some of the related lookup rules more strictly. I could reproduce similar problems with clang in sandbox [experiments](https://cpp.godbolt.org/z/bb3EEcfbe) as well -- clang error messages helped quite a bit. I posted the corresponding fixes as an unofficial PR against Amos' PR branch at https://github.com/yadij/squid/pull/2.

Those posted fixes also detail for what went wrong (in three commit messages). The most important one is at https://github.com/measurement-factory/squid/commit/eb632a62a1469c60d9cf3c9c38b7cb377b890357

I see [several ways](https://github.com/yadij/squid/pull/2#issue-1505706930) from here to the final solution, but either way, we are getting close! @yadij, your call.

src/base/AsyncJobCalls.h
rousskov: >  I will post an update with details after investigating the SwapMetaView problem.

Done at https://github.com/squid-cache/squid/pull/1213#issuecomment-1360863597.

yadij:
Superseded by PR #1215

PR #1214, opened 2022-12-19
Support "negative" squid-conf-tests

rousskov:
The test suite can now check that Squid rejects a given malformed
configuration file and that the rejection reason matches the expected
one. The latter reduces the probability that a successful "negative"
test outcome would hide a bug (because Squid has rejected a malformed
file but for a reason unrelated to what the test was trying to verify).

For now, enable just one acl regex test case, addressing an old TODO.

To improve "negative" test coverage, we would need to generate test
configurations using a configuration template and a list of problematic
expressions (with corresponding failure messages), but this hard-coded
approach is already better than nothing.

### DIFF
--- a/test-suite/squidconf/bad-regex.conf.instructions
+++ b/test-suite/squidconf/bad-regex.conf.instructions
@@ -1,0 +1,2 @@
+expect-failure (ERROR|FATAL):.*POSIX.regcomp.*failure
+
--- a/test-suite/squidconf/bad-regex.conf.instructions
+++ b/test-suite/squidconf/bad-regex.conf.instructions
@@ -1,2 +1,2 @@
-expect-failure (ERROR|FATAL):.*POSIX.regcomp.*failure
+expect-failure ERROR:.configuration.failure:.POSIX.regcomp.*failure
 
--- a/test-suite/test-squid-conf.sh
+++ b/test-suite/test-squid-conf.sh
@@ -15,6 +15,9 @@ top_builddir=$1
 sbindir=$2
 configFile=$3
 
+# If set, expect non-zero Squid exit code, with a matching stderr message
+failureRegex=""
+
 instructionsFile="$configFile.instructions"
 if test -e $instructionsFile
 then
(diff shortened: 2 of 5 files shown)


test-suite/squidconf/bad-regex.conf.instructions
```diff
@@ -0,0 +1,2 @@
+expect-failure (ERROR|FATAL):.*POSIX.regcomp.*failure
```
yadij, line 1: These patterns should only match the log output produced by the code tested.

For the test enabled by this PR the expected output is:
```
ERROR: configuration failure: POSIX regcomp(3) failure: (13) Invalid preceding regular expression
    regular expression: *
```

```suggestion
expect-failure ERROR:.configuration.failure:.POSIX.regcomp.*failure
```

test-suite/test-squid-conf.sh
```diff
@@ -68,8 +90,34 @@ then
...
 fi
 
-exec $sbindir/squid -k parse -f $configFile
+if test "$result" -eq 0
+then
+    echo "ERROR: Squid successfully parsed malformed $configFile instead of rejecting it"
+    exit 1;
+fi
+
+if ! grep -q -E "$failureRegex" $errorLog
```
yadij, line 112: awk would be a better tool for matching here as it can be given rules for multi-line matches.

yadij (changes requested):

test-suite/test-squid-conf.sh
rousskov: Please feel free to switch to awk, here or in a follow up PR (or when it actually becomes needed). I do not know enough about awk, and do not see the need to switch to awk now to implement that switch myself.

test-suite/squidconf/bad-regex.conf.instructions
rousskov: I committed your suggestion (fdec41e1404566f771181d2d778da6df6c77c002).

> These patterns should only match the log output produced by the code tested.

No constant regular expression can satisfy the above criteria: No mater what regex one writes, I can always invent a future Squid code variation where that regex will match something it should not (or not match something it should). FWIW, the originally proposed expression only matched the intended Squid output, but also allowed for highly anticipated Squid improvement where all fatal configuration ERRORs will become labeled as FATAL. I think that was a slightly better approach, but I (obviously) do not insist on it.

yadij (approved):
thank you

PR #1215, opened 2022-12-21
Fix GCC v12 build errors related to ADL in "printing" code

rousskov:
This change contains three fixes/adjustments described below.

### Fix Store::SwapMetaView printing

    src/sbuf/Stream.h:66:20: error: no match for 'operator<<' (operand
    types are 'std::basic_ostream<char>' and 'const
    Store::SwapMetaView'): (out << ... << args);

Disclaimer: This explanation omits many details that become important in
some other contexts. Most Argument-Dependent Lookup details are
documented at https://en.cppreference.com/w/cpp/language/adl

    // In some printer.h, we define a printing function for any type A.
    // This is similar to an ToSBuf() declaration in sbuf/Stream.h.
    template <typename A>
    void print(ostream &os, const A &argument) { os << argument; }

    // In some n/t.h, we define how to print objects of type N::T.
    // This is similar to store/SwapMetaView.h.
    operator <<(ostream &os, const N::T &argument);

    // In some caller1.cc, we include both headers and call print():
    #include "printer.h"
    #include "n/t.h"
    void caller1(const N::T &argument) { print(std::cout, argument); }

    // In some caller2.cc, we do the same but change #include order:
    #include "n/t.h"
    #include "printer.h"
    void caller2(const N::T &argument) { print(std::cout, argument); }

When looking at "os << argument", the compiler considers two sets of
argument printing operators, formed by the following two sources:

* The usual unqualified name lookup. This set includes N::T printing
  operator if that operator is declared in global namespace somewhere
  above the print() _template declaration_. In the example above, only
  caller2() will have that printing operator in this set, provided that
  operator is declared in global namespace (as it used to be). None of
  the callers will have that printing operator in this set otherwise.

* ADL. This set is computed from the caller point of view. It includes
  N::T printing operator if that operator is declared inside namespace N
  somewhere above the print() _caller_. In the example above, both
  caller1() and caller2() will have that printing operator in this set,
  provided that operator is declared in namespace N (as it is now). None
  of the callers will have that printing operator in this set otherwise.

For code to compile, one of the sets must contain the printing operator.
Given the above outcomes, there is only one sane solution that allows
any caller to instantiate print() with an N::T argument: The argument
printing operator must be declared inside namaspace N! Declaring in
global namespace would require maintaining certain #include order (that
will cause headaches and, eventually, circular dependencies).

In other words, we must rely on ADL, and that means declaring operators
in the namespace of one of their argument types.

### Fix std::optional<Ip::Address> printing (in src/icap/Xaction.cc)

    src/base/AsyncJobCalls.h:115:61: error: no match for 'operator<<'
    (operand types are 'std::basic_ostream<char>' and 'const
    std::optional<Ip::Address>'):
    void print(std::ostream &os) const { os << '(' << arg1 << ')'; }

In this context, both printing operator argument types are in std, but
ADL also looks at template parameters of argument types (if an argument
type is a template). That recursion adds the Ip namespace to the search.

This is a minimal fix. We should move both Ip::Address printers into
ip/print.h or similar, away from Ip::Address users that print nothing.

### Do not declare an overly general std::optional printer

The removed declaration:

* evidently became unused (after the other changes described above);
* places std::optional<N::T> printers in the wrong namespace (global
  instead of N), where ADL cannot find them;
* exposes all I/O manipulators to a, technically, unrelated
  std::optional interface.

### DIFF
--- a/src/adaptation/icap/Xaction.cc
+++ b/src/adaptation/icap/Xaction.cc
@@ -134,6 +134,22 @@ void Adaptation::Icap::Xaction::start()
     Adaptation::Initiate::start();
 }
 
+// TODO: Make reusable by moving this (and the printing operator from
+// ip/Address.h that this code is calling) into ip/print.h or similar.
+namespace Ip {
+
+inline std::ostream &
+operator <<(std::ostream &os, const std::optional<Address> &optional)
+{
+    if (optional.has_value())
+        os << optional.value();
+    else
+        os << "[no IP]";
+    return os;
+}
+
+}
+
 static void
 icapLookupDnsResults(const ipcache_addrs *ia, const Dns::LookupDetails &, void *data)
 {
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -13,7 +13,6 @@
 
 #include <iostream>
 #include <iomanip>
-#include <optional>
 
 /// Safely prints an object pointed to by the given pointer: [label]<object>
 /// Prints nothing at all if the pointer is nil.
@@ -101,18 +100,6 @@ inline AsHex<Integer> asHex(const Integer n) { return AsHex<Integer>(n); }
 
 /// Prints the first n data bytes using hex notation. Does nothing if n is 0.
 void PrintHex(std::ostream &, const char *data, size_t n);
-
-/// prints the value stored inside std::optional (if any)
-template <typename Value>
-inline std::ostream &
-operator <<(std::ostream &os, const std::optional<Value> &optional)
-{
-    if (optional.has_value())
-        os << optional.value();
-    else
-        os << "[no value]";
-    return os;
-}
 
 #endif /* SQUID_SRC_BASE_IO_MANIP_H */
 
--- a/src/base/AsyncJobCalls.h
+++ b/src/base/AsyncJobCalls.h
@@ -11,6 +11,7 @@
 
 #include "base/AsyncJob.h"
 #include "base/CbcPointer.h"
+#include "base/IoManip.h"
 #include "debug/Stream.h"
 
 /**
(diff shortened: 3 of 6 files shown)


src/adaptation/icap/Xaction.cc
```diff
@@ -134,6 +134,22 @@ void Adaptation::Icap::Xaction::start()
     Adaptation::Initiate::start();
 }
 
+// TODO: Make reusable by moving this (and the printing operator from
+// ip/Address.h that this code is calling) into ip/print.h or similar.
```
rousskov, line 138: I marked this PR as a Draft because I was not sure whether this TODO should be addressed in this unofficial PR, in the corresponding official PR, or later, in a followup official PR. @yadij, what is your preference?

src/base/IoManip.h
```diff
@@ -13,7 +13,6 @@
```
rousskov, line 13: The PR description is very long. I am not sure the future commit message is the right place to document how ADL works (in the context of this problem). Should I remove most of that description and simply refer to ADL as a well-understood concept (despite our obviously insufficient ADL understanding, at least prior to this work)?

rousskov (commented):
This PR was open based on Amos [request](https://github.com/yadij/squid/pull/2#issuecomment-1361147169). It explains the problem addressed in #1213 and offers a better (IMO) solution for it.

src/adaptation/icap/Xaction.cc
yadij: This looks like a case for adding ip/gadgets.cc

If this builds without adding problems I am good with it here though.

yadij (approved):

src/adaptation/icap/Xaction.cc
rousskov: It does build without adding problems (for some definition of "adding problems") AFAICT. It just does not make this code reusable. I can make it reusable (i.e. add `ip/gadgets`) in this PR or in a dedicated PR. @yadij, do you have a preference between those two options?

PR #1217, opened 2022-12-24
Fix ./configure --with-filedescriptors=NUMBER

rousskov:
Commit a1c2236 forgot to quote AS_CASE argument of AC_ARG_WITH(),
resulting in the following ./configure-time errors:

    error: --with-filedescriptors expects a numeric argument

We now quote AS_CASE argument of AC_ARG_WITH(), as recommended by
Autoconf and for consistency sake. More (hopefully automated) work is
needed to quote all function call arguments (that should be quoted).

Also use m4 quadrigraphs for quoting safety sake and to attract readers
attention to a risky (from writing in m4 point of view) expression. This
particular improvement is not required to fix the bug.

The other two AC_ARG_WITH() are modified for consistency sake but they
were not broken and their compiled versions have not changed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -370,7 +370,7 @@ AC_ARG_WITH(aufs-threads,
   AS_HELP_STRING([--with-aufs-threads=N_THREADS],
     [Tune the number of worker threads for the aufs object store.]), [
   AS_CASE([$withval],
-    [[[0-9]]*],[squid_opt_aufs_threads=$withval],
+    [@<:@0-9@:>@*],[squid_opt_aufs_threads=$withval],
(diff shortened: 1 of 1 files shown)


rousskov:
Changes in the "compiled" output:

```diff
--- configure   2022-12-24 11:57:43.426496861 -0500
+++ configure.fixed     2022-12-24 11:39:56.685906899 -0500
@@ -52210,2 +52210,3 @@
-  withval=$with_filedescriptors; case $withval in #(
-  0-9*) :
+  withval=$with_filedescriptors;
+  case $withval in #(
+  [0-9]*) :
```

yadij (approved):

PR #1218, opened 2022-12-24
Move Namespace::Foo printing operators to Namespace

rousskov:
... where C++ Argument-Dependent Lookup (ADL) can find them.

These specific printing operators did not encounter build problems yet,
but they were misplaced. See commit XXX for rationale and details.

Also removed unimplemented Mgr::ActionParams printing operator.

Also removed excessive "Namespace::" prefixes inside Namespace in
several printing operators declarations that were already properly
placed in their argument's namespaces (e.g., Ipc::ReadWriteLock).

No functionality changes expected.

### DIFF
--- a/src/DiskIO/IpcIo/IpcIoFile.cc
+++ b/src/DiskIO/IpcIo/IpcIoFile.cc
@@ -74,16 +74,16 @@ operator <<(std::ostream &os, const SipcIo &sio)
 /* IpcIo::Command */
 
 std::ostream &
-operator <<(std::ostream &os, const IpcIo::Command command)
+IpcIo::operator <<(std::ostream &os, const Command command)
 {
     switch (command) {
-    case IpcIo::cmdNone:
+    case cmdNone:
         return os << '-';
-    case IpcIo::cmdOpen:
+    case cmdOpen:
         return os << 'o';
-    case IpcIo::cmdRead:
+    case cmdRead:
         return os << 'r';
-    case IpcIo::cmdWrite:
+    case cmdWrite:
         return os << 'w';
     }
     // unreachable code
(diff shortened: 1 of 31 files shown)


rousskov:
@yadij, the remaining printing operators (found by `git grep -n 'operator.*<<.*std::ostream.*::' \*.h`) look correct to me:

Declared in global namespace and printing types declared inside a class that is declared in global namespace:

```
src/client_side.h:538:std::ostream &operator <<(std::ostream &os, const ConnStateData::PinnedIdleContext &pic);
src/client_side.h:539:std::ostream &operator <<(std::ostream &, const ConnStateData::ServerConnectionContext &);
src/http.h:164:std::ostream &operator <<(std::ostream &os, const HttpStateData::ReuseDecision &d);
```

Declared in Rock namespace and printing a type declared inside a class that is also declared in Rock namespace:

```
src/fs/rock/RockHeaderUpdater.h:77:std::ostream &operator <<(std::ostream &os, const HeaderUpdater::IoCbParams &params)
```

Declared in Security namespace and printing a type (aliased inside a class) that is also declared in Security namespace:

```
src/security/ErrorDetail.h:134:operator <<(std::ostream &os, const ErrorDetail::Pointer &p)
```

Declared in global namespace and printing a type (aliased inside a class) that is also declared in global namespace:

```
src/error/Detail.h:47:std::ostream &operator <<(std::ostream &os, const ErrorDetail::Pointer &);
```

Declared in global namespace and printing a type (aliased inside a namespace) that is also declared in global namespace:

```
src/security/Certificate.h:40:std::ostream &operator <<(std::ostream &, Security::Certificate &);
```

The last one above is a bit tricky because a namespace name is explicitly involved, but it has been studied earlier and determined to be correct:

```C++
// Declared outside Security because all underlying Security::Certificate types
// are declared inside global namespace.
std::ostream &operator <<(std::ostream &, Security::Certificate &);
```

AFAICT, ADL [rules](https://en.cppreference.com/w/cpp/language/adl) ignore type aliases -- they go to the actual type behind those `using` and `typedef` declarations.

Please let me know if I missed any cases that should be changed.

PR #1219, opened 2022-12-28
Fix errorpage.cc build on MacOS

kinkie:
errorpage.cc:153:44: error: implicit instantiation of undefined
    template std::array<HardCodedError, 7> HardCodedErrors


rousskov (approved):
Thank you.

I re-added the error text gist to the PR description (hopefully without violating formatting requirements).

PR #1221, opened 2022-12-29
Warn about some bad from-helper annotations

rousskov:
From-helper annotations may belong to one or more of the following
(partially overlapping) categories:

* A1: Annotations with a name ending with an underscore. Those
  annotations are reserved for admin use. Example: "category_".

* A2: Annotations that this Squid version knows about and treats
  specially. Example: "clt_conn_tag".

* B1: Annotations that this Squid version does not know about at all.
  Future Squids may start using them specially. Example: "detail".

* B2: Annotations unused by this Squid build. Example: Squid
  ./configured with --disable-auth does not use "user" and "password"
  annotations received from an external ACL helper.

* B3: Annotations incompatible with the current helper response
  type/kind. Example: Digest authentication code does not use "ha1"
  annotation in ERR responses received from digest authentication
  helpers.

* B4: Annotations with repeated names (the first of them will be used).
  Example: "user" in OK replies from negotiate authentication helpers.

Warn about received helper annotations in the B1 category.

It is possible to also warn about B2 and B3 annotations, but several
draft implementations imply that such an improvement requires complex
code changes that may also increase Squid code maintenance overheads.

Warning about B4 annotations is not that difficult, but we must decide:

* Whether custom repeated annotation are actually bad (and should be
  reported or even ignored).

* Whether Squid-recognized repeated annotations are bad enough to
  warrant the decreased performance and increased code complexity that
  reporting them would require. In other words, do we want to penalize
  good helpers that do _not_ send any repeated Squid-recognized
  annotations (except, perhaps, "message") to warn about bad helpers
  that do?

### DIFF
--- a/src/helper/Reply.cc
+++ b/src/helper/Reply.cc
@@ -158,9 +158,11 @@ Helper::Reply::CheckReceivedKey(const SBuf &key, const SBuf &value)
     // Squid recognizes these keys (by name) in some helper responses
     static const std::vector<SBuf> recognized = {
         SBuf("clt_conn_tag"),
+        SBuf("group"),
         SBuf("ha1"),
         SBuf("log"),
         SBuf("message"),
+        SBuf("nonce"),
         SBuf("password"),
         SBuf("rewrite-url"),
         SBuf("status"),
@@ -171,9 +173,9 @@ Helper::Reply::CheckReceivedKey(const SBuf &key, const SBuf &value)
         SBuf("user")
     };
 
-    // TODO: Merge with Notes::ReservedKeys(). That list has entries that Squid
-    // sources do _not_ recognize today ("group" and "ttl"), and it is missing
-    // entries from our list ("clt_conn_tag", "store-id", and "token").
+    // TODO: Merge with Notes::ReservedKeys(). That list has an entry that Squid
+    // sources do _not_ recognize today ("ttl"), and it is missing some
+    // recognized entries ("clt_conn_tag", "nonce", store-id", and "token").
 
     // TODO: Skip empty keys (with a dedicated error message).
 
--- a/src/helper/Reply.cc
+++ b/src/helper/Reply.cc
@@ -78,11 +78,11 @@ Helper::Reply::finalize()
             char *w1 = strwordtok(nullptr, &p);
             if (w1 != nullptr) {
                 const char *authToken = w1;
-                notes.addChecked("token",authToken);
+                notes.add("token",authToken);
             } else {
                 // token field is mandatory on this response code
                 result = Helper::BrokenHelper;
-                notes.addChecked("message","Missing 'token' data");
+                notes.add("message","Missing 'token' data");
             }
 
         } else if (!strncmp(p,"AF ",3)) {
@@ -97,15 +97,15 @@ Helper::Reply::finalize()
             if (w2 != nullptr) {
                 // Negotiate "token user"
                 const char *authToken = w1;
-                notes.addChecked("token",authToken);
+                notes.add("token",authToken);
 
                 const char *user = w2;
-                notes.addChecked("user",user);
+                notes.add("user",user);
 
             } else if (w1 != nullptr) {
                 // NTLM "user"
                 const char *user = w1;
-                notes.addChecked("user",user);
+                notes.add("user",user);
             }
         } else if (!strncmp(p,"NA ",3)) {
             // NTLM fail-closed ERR response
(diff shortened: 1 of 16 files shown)


src/helper/Reply.cc
```diff
@@ -151,6 +151,58 @@ isKeyNameChar(char c)
...
+
+    // We do not check custom keys for repetitions because Squid supports them:
+    // The "note" ACL checks all of them and %note prints all of them.
+    if (*key.rbegin() == '_')
+        return; // a custom key
+
+    // To simplify, we allow all recognized keys, even though some of them are
+    // only expected from certain helpers or even only in certain reply types.
+    // To simplify and optimize, we do not check recognized keys for repetitions
+    // because _some_ of them (e.g., "message") do support repetitions.
```
rousskov, line 196: Look for "B4" paragraphs in PR description for an explanation why those warnings were left out of this PR. Same for repeated custom keys.

src/helper/Reply.cc
```diff
@@ -151,6 +151,58 @@ isKeyNameChar(char c)
...
+        return;
+    }
+
+    // We do not check custom keys for repetitions because Squid supports them:
+    // The "note" ACL checks all of them and %note prints all of them.
+    if (*key.rbegin() == '_')
+        return; // a custom key
+
+    // To simplify, we allow all recognized keys, even though some of them are
+    // only expected from certain helpers or even only in certain reply types.
```
rousskov, line 194: > PR description: It is possible to also warn about B2 and B3 annotations, but several draft implementations imply that such an improvement requires complex code changes that may also increase Squid code maintenance overheads.

You can view one such unpolished/unfinished attempt to warn about B2 and B3 annotations at https://github.com/measurement-factory/squid/pull/183. That unofficial PR instruments NotePairs to track explicit Entry "uses" by former findFirst() and find() calls, automatically detecting most B2 and B3 cases. The code is not terrible, and the performance overhead of those changes is relatively small, but the design has significant problems (detailed in the last three paragraphs of that PR description). We also tried two other approaches, but they all had serious shortcomings.

At the end of the day, I think we should start with this simple change and leave further improvements as a long-term goal. If nothing else, this will allow us to collect feedback from admins that use offending helpers. If there is too much negative feedback, we may need to reconsider the current official "just add underscore" approach itself!

src/helper/Reply.cc
```diff
@@ -151,6 +151,58 @@ isKeyNameChar(char c)
...
+        SBuf("store-id"),
+        SBuf("tag"),
+        SBuf("token"),
+        SBuf("url"),
+        SBuf("user")
+    };
+
+    // TODO: Merge with Notes::ReservedKeys(). That list has an entry that Squid
+    // sources do _not_ recognize today ("ttl"), and it is missing some
+    // recognized entries ("clt_conn_tag", "nonce", store-id", and "token").
```
rousskov, line 178: Because if those differences, I think it would be better to merge these two sets in a dedicated PR. Otherwise, the merging noise will make it much more difficult to understand the true meaning/scope of this PR changes.

rousskov (commented):
Sample warning:

```
2022/12/28 14:38:34| WARNING: Unsupported or unexpected from-helper annotation with a name reserved for Squid use: foo=bar
    advice: If this is a custom annotation, rename it to add a trailing underscore: foo_
```

src/helper/Reply.cc
```diff
@@ -176,7 +228,11 @@ Helper::Reply::parseResponseKeys()
         if (v != nullptr && urlDecode && (p-v) > 2) // 1-octet %-escaped requires 3 bytes
             rfc1738_unescape(v);
 
-        notes.add(key, v ? v : ""); // value can be empty, but must not be NULL
+        // TODO: Convert the above code to use Tokenizer and SBuf
+        const SBuf parsedKey(key);
+        const SBuf parsedValue(v ? v : ""); // allow empty values
```
kinkie, line 233: Isn't SBuf(nullptr) the same as SBuf("")?

If so, why the need for the test?

rousskov: > Isn't SBuf(nullptr) the same as SBuf("")?

Yes, the primary effect is the same AFAICT.

> If so, why the need for the test?

We do not. I removed that no longer necessary test in https://github.com/squid-cache/squid/commit/d1f196d4af711f7b5a9413a805fb8deddac16a01

kinkie (approved):
That was my only question.
Code LGTM

PR #1222, opened 2022-12-30
Bug 5254, part 1: Do not leak master process' cache.log to kids

rousskov:
The fork()ed kids unknowingly inherited cache.log file descriptor and,
hence, prevented the underlying file from being deleted on log rotation.

This change does not fully fix the bug because the file is still being
held open by the master process itself. This change was isolated because
it lacks bad/controversial side effects -- it is a simple step forward.

This fix may also help stop leaking kid's cache.log descriptor to
helpers, but that requires more work -- replacing descriptor duping
trick in ipcCreate() with a proper servicing of helper stderr descriptor
via Comm. For now, each helper process still keeps cache.log alive.


src/debug/debug.cc
```diff
@@ -761,8 +762,10 @@ DebugFile::reset(FILE *newFile, const char *newName)
     }
     file_ = newFile; // may be nil
 
-    if (file_)
+    if (file_) {
+        commSetCloseOnExec(fileno(file_));
```
rousskov, line 766: The primary reason I marked this PR as a draft is to ask the following question: Do we care about platforms where commSetCloseOnExec() does not work (e.g., because `FD_CLOEXEC` is not defined)?

AFAICT, if we do want to fully support those platforms, then simply calling commSetCloseOnExec() is not enough (or, arguably, should _never_ be needed!) because we have to loop through all open descriptors and close all of them explicitly before calling one of the exec*() functions anyway.

If we do not care, then it would be nice to add a comment to commSetCloseOnExec(), saying as much.

rousskov: @yadij, @kinkie, any opinion regarding commSetCloseOnExec()? If there are none, then I will proceed with the assumption that "we do not care" but will not "add a comment to commSetCloseOnExec()" since I got no explicit confirmation that we do not actually care, and I am not sure whether we should.

kinkie: according to https://www.gnu.org/software/gnulib/manual/html_node/fcntl_002eh.html , FD_CLOEXEC is undefined on MSVC 14 (an incredibly old version) and mingw. The former is incredibly old, the latter might be relevant for us at some point in the future but I for now I would just state that we could not care for platforms where it's not defined.

rousskov: @yadij, do you agree with @kinkie suggestion to state that we do not care for platforms where FD_CLOEXEC is not defined? This draft PR is blocked on making that decision as detailed in this change request.

kinkie: In addition to my previous statement, according to https://devblogs.microsoft.com/oldnewthing/20111216-00/?p=8873 , FD_CLOEXEC is the default behaviour on Windows. 

In other words, let's move forward and focus on the platforms we have, the discussion about how to better deal with windows belongs there

rousskov: > let's move forward and focus on the platforms we have, the discussion about how to better deal with windows belongs there

I have pushed the corresponding documentation change in commit c82d196 and am marking this PR as ready for review.

kinkie: Looks good to me, the comment is clear and sets clear responsibilities. #shipit

kinkie (approved):
Looks good!

PR #1223, opened 2022-12-30
Bug 5254, p2: Rotate master process' cache.log

rousskov:
Before these changes:

* Master process kept the original cache.log open forever, resulting in
  that file lurking in the filesystem for as long as a Squid instance
  runs, despite log rotations that delete it.

* Multiprocess Squid instances rotated cache.log K times per single
  "squid -k rotate" event, where K is the number of all Squid kids:
  workers, diskers (if any), and Coordinator (if any). Admins saw the
  expected number of rotations only for K=1 instances (workers=1, no
  diskers, and, hence, no Coordinator) -- a lucky accident.

After these changes:

* Master process rotates its cache.log. The original file can be fully
  deleted after the first rotation.

* Squid rotates cache.log 1+K times per "squid -k rotate". The
  additional rotation is performed by the master process itself. Admins
  never see the expected number of rotations.

Thus, this change makes things better for K>1 configurations, but makes
things partially better/worse for K=1 configurations. All configurations
remain broken.

A supported workaround is to use external filename rotation. Each Squid
process closes and reopens the cache_log file: logfile_rotate 0

An unsupported workaround is to configure each process to have its own
cache.log. For example: cache_log cache-${process_number}.log

Both workarounds rely on these code changes and the Bug 5254 p1 fix.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -997,6 +997,7 @@ mainReconfigureFinish(void *)
     reconfiguring = 0;
 }
 
+/// rotates the logs of the worker and other kid processes; \sa masterRotate()
 static void
 mainRotate(void)
 {
@@ -1729,6 +1730,11 @@ mainStartScript(const char *prog)
     }
 }
 
+/*
+ * The master*() functions below are used by the master process in SMP
+ * configurations (i.e. where the master process is not also a worker process).
+ */
+
 /// Initiates shutdown sequence. Shutdown ends when the last running kids stops.
 static void
 masterShutdownStart()
@@ -1759,6 +1765,21 @@ masterReconfigureFinish()
     reconfiguring = 0;
 }
 
+/// Rotates master process logs; \sa mainRotate()
+static void
+masterRotate()
+{
+    if (AvoidSignalAction("log rotation", do_rotate))
+        return;
+
+    // XXX: When all SMP Squid processes use the same cache.log (default), each
+    // process, including this master process, rotates the same log file,
+    // resulting in K excessive rotations for K kid processes.
+    _db_rotate_log(); // cache.log
+
+    // master process should not have any other logs that mainRotate() rotates
+}
+
 /// Reacts to the kid revival alarm.
 static void
 masterReviveKids()
@@ -1777,6 +1798,8 @@ masterCheckAndBroadcastSignals()
         masterShutdownStart();
     if (do_reconfigure)
         masterReconfigureStart();
+    if (do_rotate)
+        masterRotate();
     if (do_revive_kids)
         masterReviveKids();
 


rousskov:
`S-waiting-for-PR`: The full effects described in this PR assume that #1222 fix has been merged. This PR changes work fine without that PR, but the cache.log file will remain lurking in the filesystem until _both_ PRs are merged.

src/main.cc
```diff
@@ -1759,6 +1765,21 @@ masterReconfigureFinish()
...
+static void
+masterRotate()
+{
+    if (AvoidSignalAction("log rotation", do_rotate))
+        return;
+
+    // XXX: When all SMP Squid processes use the same cache.log (default), each
+    // process, including this master process, rotates the same log file,
+    // resulting in K excessive rotations for K kid processes.
+    _db_rotate_log(); // cache.log
```
rousskov, line 1778: The bug behind this XXX is at least as old as SMP support. This PR documents it and adds one more name rotation (for cache.log). Please see PR description for details.

Arguably, this added rotation is the only one that should be happening, but it would take a lot of work to actually achieve that single-rotation goal in SMP Squids -- we may have to refactor all kids to stop logging to files and start sending log records to this master process (via pipes or some such)!

I have an alternative suggestion for addressing this "too many rotations" problem: Let's drop all internal support for filename renaming during log rotation. Squid will still close/reopen log files, of course, but admins will need to use logrotate or a similar external tool to rename the logs.

Whether the harm added by this draft PR (again, see PR description for details) is acceptable depends, in part, on whether we are going to remove renaming support: If we are going to drop it, then this PR should go in (after any other problems are addressed, of course). This uncertainty is the primary reason I marked this PR as a draft.

----

Long-term, we probably want to move towards logging to stdout. Container-using folks would love that, and stdout is easy to understand and use in any production environment. Since stdout cannot support rotations, disabling that support now would make any future transition easier/smaller...

[^1]: Using stdout may still require refactoring all kids to send log records to this master process to ensure that records from different kids do not overlap. Thus, it is not necessarily a _simplification_ from code/development point of view. That move may be a good idea for other reasons. I am mentioning it here just because it may be used as one more reason to drop renaming support.

src/main.cc
```diff
@@ -997,6 +997,7 @@ mainReconfigureFinish(void *)
     reconfiguring = 0;
 }
 
+/// rotates the logs of the worker and other kid processes; \sa masterRotate()
 static void
 mainRotate(void)
```
rousskov, line 1002: Single-process instances (e.g., `squid -N`) use this old mainRotate() function, not the new masterRotate(). In single-process instances, that only process is a _worker_, so the proposed description works OK for them as well. If you can suggest a better wording (that would cover all SMP and non-SMP cases), please do!

yadij:
I have updated the description with bug references. If possible please separate the fixes into two PR.

src/main.cc
yadij: Better to have all squid write to their stderr and a dedicated logging process managing delivery of those sockets to correct cache.log. There is an open wishlist to allow a log daemon helper to get cache.log.

rousskov: Glad we seem to agree regarding the future architecture -- stdout or stderr logging from kid processes, merged at a central collection point. Whether we will use stdout or stderr, and whether that central point is the master process or some dedicated daemon are all secondary details that should be finalized by those working on that feature in the future.

PR #1224, opened 2023-01-04
Maintenance: Consistent use of C++11 "override" specifier

rousskov:
Start following CppCoreGuidelines recommendation C.128: Virtual
functions should specify exactly one of virtual, override, or final.

Used clang-tidy modernize-use-override check to mark overridden methods.
Clang-tidy uses clang AST, so we can only modernize code that clang can
see, but we can enable enough ./configure options for clang to reach
most (possibly all) relevant classes in a single build.

Manual gradual conversion of modified (for other reasons) method
declarations is impractical because some compilers complain when a class
uses an inconsistent _mix_ of virtual and override declarations:

    warning: 'toCbdata' overrides a member function but is not marked
    'override' [-Winconsistent-missing-override]

### cbdata changes

The following manual changes avoid -Winconsistent-missing-override
warnings. Doing these separately, either before or after applying
modernize-use-override, would trigger those warnings.

* Replaced CbdataParent-CBDATA_CLASS with CbdataParent-CBDATA_CHILD
  class inheritance sequences. This fix does not really change anything
  except for making toCbdata() specifiers consistent.

* Replaced CbdataParent-CBDATA_CLASS-CBDATA_CHILD with
  CbdataParent-CBDATA_INTERMEDIATE-CBDATA_CHILD class inheritance
  sequences. This fix removes unused new/delete operators in
  intermediate classes (and associated unused static memory pools) and
  makes toCbdata() specifiers consistent. This fix was difficult to get
  right because of the old design problem discussed below.

While working on the second bullet changes, we first tried to simply
drop the intermediate CBDATA_CLASS sequence element because it should
never be needed/used in class hierarchies ending with CBDATA_CHILD.
Fortunately, CI tests discovered that dropping CBDATA_CLASS converts an
old design problem into runtime bugs: Those intermediate class
constructors (e.g., Mgr::StoreToCommWriter) actually call toCbdata()
when creating connection closure callbacks and such!

During intermediate constructor execution, the class virtual table does
not yet point to toCbdata() defined by CBDATA_CHILD. If we simply remove
CBDATA_CLASS, then, during that intermediate constructor execution, the
virtual table would still point to pure CbdataParent::toCbdata(). Those
intermediate constructors would then crash Squid:

    FATAL: Dying from an exception handling failure;
    exception: [no active exception]

    in OnTerminate () at main.cc:1310
    in std::terminate() () from libstdc++.so.6
    in __cxa_pure_virtual () from libstdc++.so.6
    in CbcPointer<Mgr::StoreToCommWriter>::CbcPointer at CbcPointer.h:94
    in Mgr::StoreToCommWriter::StoreToCommWriter at ...oCommWriter.cc:29
    
We now use lighter CBDATA_INTERMEDIATE() instead of intermediate
CBDATA_CLASS. The resulting code is as risky as it used to be, but at
least we are marking this (old) design problem (in some cases).

We rejected the idea of defining CbdataParent::toCbdata() instead. It
would work fine for linear class inheritance where "this" does not
change -- in those simple class hierarchies, we do not need toCbdata()!
However, such a change would make any constructor-time bugs in multiple
inheritance hierarchies much more difficult to find because the faulty
constructor will work instead of triggering the above FATAL error. The
crashes would happen later, when wrong cbdata pointer is used. While
CBDATA_INTERMEDIATE() cannot fix this design problem, it does not make
it _worse_, so we prefer that solution.

### DIFF
--- a/src/AccessLogEntry.h
+++ b/src/AccessLogEntry.h
@@ -44,11 +44,11 @@ public:
     typedef RefCount<AccessLogEntry> Pointer;
 
     AccessLogEntry();
-    virtual ~AccessLogEntry();
+    ~AccessLogEntry() override;
 
     /* CodeContext API */
-    virtual std::ostream &detailCodeContext(std::ostream &os) const override;
-    virtual ScopedId codeContextGist() const override;
+    std::ostream &detailCodeContext(std::ostream &os) const override;
+    ScopedId codeContextGist() const override;
 
     /// Fetch the client IP log string into the given buffer.
     /// Knows about several alternate locations of the IP
--- a/src/cbdata.h
+++ b/src/cbdata.h
@@ -269,12 +269,18 @@ cbdata_type cbdataInternalAddType(cbdata_type type, const char *label, int size)
        static cbdata_type CBDATA_##type;
 
 /// Starts cbdata-protection in a class hierarchy.
-/// Child classes in the same hierarchy should use CBDATA_CHILD().
+/// Intermediate classes in the same hierarchy must use CBDATA_INTERMEDIATE() if
+/// they risk creating cbdata pointers in their constructors.
+/// Final classes in the same hierarchy must use CBDATA_CHILD().
 class CbdataParent
 {
 public:
     virtual ~CbdataParent() {}
     virtual void *toCbdata() = 0;
+
+private:
+    /// hack: ensure CBDATA_CHILD() after a toCbdata()-defining CBDATA_INTERMEDIATE()
+    virtual void finalizedInCbdataChild() = 0;
 };
 
 /// cbdata-enables a stand-alone class that is not a CbdataParent child
@@ -282,10 +288,25 @@ public:
 /// use this at the start of your class declaration for consistency sake
 #define CBDATA_CLASS(type) CBDATA_DECL_(type, noexcept)
 
-/// cbdata-enables a CbdataParent child class (including grandchildren)
+/// cbdata-enables a final CbdataParent-derived class in a hierarchy
 /// sets the class declaration section to "private"
 /// use this at the start of your class declaration for consistency sake
-#define CBDATA_CHILD(type) CBDATA_DECL_(type, override final)
+#define CBDATA_CHILD(type) CBDATA_DECL_(type, final) \
+      void finalizedInCbdataChild() final {}
+
+/// cbdata-enables a non-final CbdataParent-derived class T in a hierarchy.
+/// Using this macro is required to be able to create cbdata pointers in T
+/// constructors, when the current vtable is still pointing to T::toCbdata()
+/// that would have been pure without this macro, leading to FATAL runtime
+/// OnTerminate() calls. However, assuming that the final cbdata pointer will
+/// still point to T::this is risky -- multiple inheritance changes "this"!
+///
+/// sets the class declaration section to "private"
+/// use this at the start of your class declaration for consistency sake
+#define CBDATA_INTERMEDIATE() \
+    public: \
+        void *toCbdata() override { return this; } \
+    private:
 
 /**
  * Creates a global instance pointer for the CBDATA memory allocator
--- a/src/tests/stub_ipc_Forwarder.cc
+++ b/src/tests/stub_ipc_Forwarder.cc
@@ -7,12 +7,18 @@
  */
 
 #include "squid.h"
-#include "ipc/Forwarder.h"
 
-//Avoid linker errors about Ipc::Forwarder
-void foo_stub_ipc_forwarder();
-void foo_stub_ipc_forwarder()
-{
-    Ipc::Forwarder foo(nullptr,1.0);
-}
+#define STUB_API "ipc/Forwarder.cc"
+#include "tests/STUB.h"
+
+#include "ipc/Forwarder.h"
+Ipc::Forwarder::Forwarder(Request::Pointer, double): AsyncJob("Ipc::Forwarder"), timeout(0) STUB
+Ipc::Forwarder::~Forwarder() STUB
+void Ipc::Forwarder::start() STUB
+bool Ipc::Forwarder::doneAll() const STUB_RETVAL(false)
+void Ipc::Forwarder::swanSong() STUB
+void Ipc::Forwarder::callException(const std::exception &) STUB
+void Ipc::Forwarder::handleError() STUB
+void Ipc::Forwarder::handleTimeout() STUB
+void Ipc::Forwarder::handleException(const std::exception &) STUB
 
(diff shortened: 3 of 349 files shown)


src/AccessLogEntry.h
```diff
@@ -44,11 +44,11 @@ class AccessLogEntry: public CodeContext
     typedef RefCount<AccessLogEntry> Pointer;
 
     AccessLogEntry();
-    virtual ~AccessLogEntry();
+    ~AccessLogEntry() override;
```
rousskov, line 47: In early C++17 days, there were [debates](https://github.com/isocpp/CppCoreGuidelines/issues/721) regarding marking destructors with `override`/`final`. That debate has been [settled](https://github.com/isocpp/CppCoreGuidelines/pull/1448) in favor of marking. Clang-tidy and this change follow that settled recommendation ([C.128](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final)).

src/cbdata.h
```diff
@@ -269,23 +269,44 @@ cbdata_type cbdataInternalAddType(cbdata_type type, const char *label, int size)
...
 /// sets the class declaration section to "private"
 /// use this at the start of your class declaration for consistency sake
 #define CBDATA_CLASS(type) CBDATA_DECL_(type, noexcept)
 
-/// cbdata-enables a CbdataParent child class (including grandchildren)
+/// cbdata-enables a final CbdataParent-derived class in a hierarchy
 /// sets the class declaration section to "private"
 /// use this at the start of your class declaration for consistency sake
-#define CBDATA_CHILD(type) CBDATA_DECL_(type, override final)
+#define CBDATA_CHILD(type) CBDATA_DECL_(type, final) \
```
rousskov, line 294: I removed the `override` specifier here because [C.128](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final) (which I support and based this PR on) says "exactly one". I did not find any other cases where the official code combines these two specifiers.

rousskov (commented):
As explained in the PR description, this PR has to combine automated and manual changes. If you would like to review automated or manual changes in isolation, please use this PR branch commits; all automated changes are in commit 9a8457c.

FWIW, I posted a [script](https://gist.github.com/rousskov/9d982bfaa45bc16713f2484ce5c76c30) that applies clang-tidy modernize-use-override changes (similar the automated changed contained in this PR) to any git branch. They would need to be followed up by manual cbdata adjustments, but those (more focused on rarely changing lines) fixes can probably be cherry-picked from this PR branch without creating lots of merge conflicts.

kinkie:
Thanks for doing this!
I've checked a few pending PRs for merge conflicts. These PR have conflicts:
#1211 , #1179 , #1224 

I didn't check too much what they are about and whether they are abandoned

kinkie:
I agree with the the intent of this going in, don't get me wrong.
In fact, I wouldn't mind to enforce this a part of source-maintenance.sh if clang-tidy can be found on the system running it.

Oops, bad copy paste on my side; it was a quick hack to verify the impact on other ongoing work (at 5% all outstanding PRs, that's remarkably low, too)

rousskov:
> I wouldn't mind to enforce this a part of source-maintenance.sh if clang-tidy can be found on the system running it.

Yes, we may eventually be able to enforce this (and more!) via CI tests, one way or another. For example, we should be able to install all the necessary tools on GitHub Actions, just like we already install gperf and the Daft framework.

Fortunately, until those CI improvements:

* Most violations will be automatically enforced by existing clang build tests: clang enables `-Winconsistent-missing-override` [option](https://clang.llvm.org/docs/DiagnosticsReference.html#winconsistent-missing-override) by default! This is why this PR had to fix those cbdata-related warnings (triggered by inconsistent specifiers added by our cbdata macros that clang-tidy could not, of course, "fix").

* When somebody adds a new API, it would not be very difficult to spot violations during that PR review as well -- there should be no new method declarations with `virtual` in them except in the base class that defines the new interface.

In summary, I doubt this will get out of hand even if CI improvements take a while to materialize.

yadij:
Thank you for for reverting your QA requirements that we a) use both virtual and override, and b) strictly forbid mixing manual changes into a script PR. As per our prior discussions (circa 2010-2017) I am very much in favour of this update.

Not signing off on it yet simply due to inability to do a proper review on large changes.

rousskov:
> Thank you for for reverting your QA requirements that we a) use both virtual and override, and b) strictly forbid mixing manual changes into a script PR.

FWIW, there are no general "QA requirements" (whatever that is) that "strictly forbid mixing manual changes into a script PR". Mixing large automated and manual changes is bad and should be avoided, but there are exceptional situations where such avoidance is not possible or not worth the cost. This is one of those exceptional cases, as documented in the PR description. There is nothing good about this auto/manual mix, but the alternatives are deemed to be overall worse in this specific case.

As for using both `virtual` and `override`, that probably seemed like the best option in August 2015, when the first such combination was added, and when we were probably unaware of [C.128](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final) -- the "very early draft" of CppCoreGuidelines was only [committed](https://github.com/isocpp/CppCoreGuidelines/commit/947cf3a) in September 2015, after Squid commits 640fe8f and 0d253df. As more such combinations entered the code, it would make sense to keep using that reasonable combination for consistency sake (until a global conversion like this one). There is no reversal of any requirements here. Just evolution of code (and our knowledge) based on the best information/tools available to us.

src/tests/stub_ipc_Forwarder.cc
```diff
@@ -7,12 +7,18 @@
...
-void foo_stub_ipc_forwarder();
-void foo_stub_ipc_forwarder()
-{
-    Ipc::Forwarder foo(nullptr,1.0);
-}
+#define STUB_API "ipc/Forwarder.cc"
+#include "tests/STUB.h"
+
+#include "ipc/Forwarder.h"
+Ipc::Forwarder::Forwarder(Request::Pointer, double): AsyncJob("Ipc::Forwarder"), timeout(0) {STUB}
```
rousskov, line 15: We have to initialize the base class in this constructor to avoid the following compiler error. That initialization, if followed by a simple `STUB` expression, confuses Astyle -- it starts needlessly indenting the lines below it, so we have to follow with a less common (but still used) `{STUB}` expression. More details at ebea4b3.

```
tests/stub_ipc_Forwarder.cc:15:17: error: constructor for 'Ipc::Forwarder' must explicitly initialize the base class 'AsyncJob' which does not have a default constructor
```

rousskov:
FWIW, I plan to clear this PR for merging on 2023-01-13 or so unless somebody stops me.

kinkie (approved):
let's move this forward

PR #1226, opened 2023-01-05
Maintenance: Fix "make installcheck" with --enable-gnuregex

rousskov:
... and, hypothetically, old system regex libraries (e.g., 1999 glibc).

    ERROR: Squid successfully parsed malformed
        ...test-suite/squidconf/bad-regex.conf instead of rejecting it

We compile regexes using regcomp(3) with a REG_EXTENDED flag. Under the
hood, that flag gets converted into the RE_SYNTAX_POSIX_EXTENDED syntax
bit set. Glibc changed the definition of that set on 2000-01-18 by
adding a RE_CONTEXT_INVALID_OPS bit. Prior to that, a `*` regex was a
valid expression -- a leading `*` itself matched[^1] an empty string!
Squid copied glibc's bit set macro in 1996 (i.e. without the new bit).

Thus, Squid built with --enable-gnuregex would accept `*` regex, while
Squid linked against a modern regex library would reject it. When 2015
commit 6d2a427 added the `*` test, --enable-gnuregex support was broken,
so the "unstable" test result could not have been noticed. When a recent
commit 12cac45 re-enabled that test, the failures were again missed (due
to another old test-suite bug that ignored failures; to be fixed next).

We now use an invalid regex rejected by system and built-in regex code.

[^1]: https://web.mit.edu/gnu/doc/html/regex_3.html#SEC12

### DIFF
--- a/test-suite/squidconf/bad-regex.conf
+++ b/test-suite/squidconf/bad-regex.conf
@@ -5,4 +5,5 @@
 ## Please see the COPYING and CONTRIBUTORS files for details.
 ##
 
-acl foo browser *
+# REG_EPAREN: Unmatched parenthesis group operator.
+acl foo browser x(y


test-suite/squidconf/bad-regex.conf
```diff
@@ -5,4 +5,5 @@
 ## Please see the COPYING and CONTRIBUTORS files for details.
 ##
 
-acl foo browser *
+# REG_EPAREN: Unmatched parenthesis group operator.
+acl foo browser x(y
```
rousskov, line 9: > This is not correct fix. The `*` pattern was specified since that was commonly used by admin. 

The `*` pattern is a valid regex pattern from regcomp(3) point of view, so `*` cannot be used in a negative test, even if that pattern is "commonly used by admin". This PR fixes that bug.

If you would like Squid to use a different bad regex than the one proposed by this PR, please suggest one, but keep in mind that the bad regex in this PR was picked while studying regcomp() implementation, in hope to avoid recreating the same kind of bug with a different pattern.

> The correct fix is to make the squid library match the upstream original code / behaviour.

Making "squid library match the upstream original code" will _not_ fix this negative test. After such an update, the `*` regex will be successfully parsed by Squid (instead of being rejected) in _more_ build environments (compared to environments where it is successfully parsed today). In other words, updating the code will make the situation worse, not better, as far as this test is concerned.

rousskov: > > Amos: This is not correct fix. The `*` pattern was specified since that was commonly used by admin.
 
> Alex: The `*` pattern is a valid regex pattern from regcomp(3) point of view, so `*` cannot be used in a negative test, even if that pattern is "commonly used by admin". This PR fixes that bug.

Clarification: The `*` pattern is a valid regex pattern from _some_ regcomp(3) implementations point of view. It all depends on whether the RE_CONTEXT_INVALID_OPS (or equivalent) bit is set. We should not use a regex pattern that is accepted by some implementations as valid. The purpose of this test is to check that Squid correctly handles _invalid_ patterns. We should use a pattern that is always invalid (at least in known implementations), regardless of the flags.
 
> > Amos: The correct fix is to make the squid library match the upstream original code / behaviour.

> Alex: Making "squid library match the upstream original code" will _not_ fix this negative test. 

That was wrong, sorry. It _will_ fix this negative test from the exiting/known CI environments point of view. That fact does not make this fix incorrect though. There are many ways to fix this bug. The way proposed in this PR is _better_ than updating built-in code (as far as the bug fix is concerned) because it does not affect Squid behavior and is much simpler. Updating the built-in code would change Squid behavior (in environments that are using that code).

yadij: The Squid relevant code is supposed to exactly replicate the behaviour of the upstream library. Ensuring that admin using the ```*``` pattern are informed of it being invalid. As your details show it has been failing to do so.

For this type of bug a behaviour change is the ideal fix. So that is not a valid argument against do it correctly.

yadij (changes requested):
Please fix the bug(different ```*``` handling from upstream library), not change the pattern to hide it.

rousskov:
Closing in anticipation of #1229 making these tests not fail on known platforms. With that PR change, our time can be spent more productively than arguing about the definition of "upstream library" and how uniform its handling of `*` is. This simple fix (and the complex explanation behind it) would be available should we ever hit this problem with an external regex library.

PR #1227, opened 2023-01-05
Maintenance: Do not hide squid-conf-tests errors

rousskov:
These "make installcheck" failures were ignored since commit 5992004.


rousskov:
I expect CI tests for this PR to _fail_ until #1226 is merged, fixing the (previously ignored) FAILure below:

```
PASS: squid.conf test: squid.conf.default
ERROR: Squid successfully parsed malformed ../../../test-suite/squidconf/bad-regex.conf instead of rejecting it
FAIL: squid.conf test: bad-regex.conf
make[2]: *** [Makefile:1506: squid-conf-tests] Error 1
```

kinkie:
I'm going to start a debate here, we can move it to squid-dev if needed.

I wonder.. do we really really need to bundle a regex library?
It's a headache, it's an old fork of a deprecated library, it's an extraneous body in our codebase. It's rarely tested because all current systems I can think of come with a functioning regex library.

What if we instead detected at configure time that a working regex is missing, and instructed instead the administrators to install gnulib as a prerequisite to build if that was the case?

kinkie:
If there is consensus on removing gnuregex, I'm happy to put in the legwork to make it happen

yadij:
With the move to C++17 we dropped support for the older compilers incomplete-C++11 lack of regex. IMO yes, we can drop the gnuregex now.

rousskov:
> If there is consensus on removing gnuregex, I'm happy to put in the legwork to make it happen

I hope there is consensus. I would support a high-quality PR removing built-in gnuregex code. We certainly do not have to provide that kind of feature, and (all other factors being equal) having less bad code and fewer configuration options is better.

FWIW, the current [plan](https://github.com/squid-cache/squid/pull/1096#issuecomment-1191609567) is to migrate to std::regex. After (or when) that is done, the built-in regex code can be removed without any loss of functionality for those who still rely on it (or without speculating whether there are no such admins left). However, again, I do support an earlier removal, for the reasons given above.

Please note that _this_ PR has pretty much nothing to do with built-in gnuregex or regex in general! The bug fixed here would exist even if builtin gnuregex is removed. The bug and the fix apply to _all_ squid.conf tests, not just the tests that use regexes.

kinkie:
> > If there is consensus on removing gnuregex, I'm happy to put in the legwork to make it happen
> 
> I hope there is consensus. I would support a high-quality PR removing built-in gnuregex code.

It seems so; I'll start work on preparing the PR.

[..]
> However, again, I do support an earlier removal, for the reasons given above.

Good, thanks

> Please note that _this_ PR has pretty much nothing to do with built-in gnuregex or regex in general! The bug fixed here would exist even if builtin gnuregex is removed. The bug and the fix apply to _all_ squid.conf tests, not just the tests that use regexes.

Sure, but it was a nice trigger for followups.

Thanks!

rousskov:
With #1229 merged, this PR is no longer waiting for another PR.

PR #1228, opened 2023-01-05
Fix ./configure --enable-poll and --disable-poll

rousskov:
./configure: line 42685: xyes: command not found
    ./configure: line 42685: xno: command not found

Since commit a1c2236, these options triggered errors messages above, and
--enable-poll did not affect syscall selection for the I/O loop:

    ./configure --enable-poll
    configure: Using epoll for the IO loop.


rousskov:
Current master/v6 code output:

```
./configure --enable-poll
./configure: line 42685: xyes: command not found
configure: Using epoll for the IO loop.
```

PR code output:

```
./configure --enable-poll
configure: choosing user-specified net I/O API poll
configure: Using poll for the IO loop.
```

yadij (approved):
Thank you for catching this

PR #1229, opened 2023-01-06
Remove bundled GnuRegex library

kinkie:
Modern operating systems provide a functioning regex
library, so we don't need to carry one anymore.


CREDITS
```diff
@@ -79,7 +79,7 @@ research project called The Harvest Information Discovery and Access System:
  *   the technology.  The Harvest software is really a reference
  *   implementation of a set of protocols and formats, some of which
  *   we intend to standardize.  We encourage commercial
- *   re-implementations of code complying to this set of standards.  
+ *   re-implementations of code complying to this set of standards.
```
rousskov, line 82: Please undo the out-of-scope whitespace changes above (and further below). There should be no CREDITS changes except for GnuRegex blobs removal.

configure.ac
```diff
@@ -2968,36 +2968,11 @@ AS_IF(test "x$enable_zph_qos" = "xyes"],[
...
-])
-AC_MSG_CHECKING(if GNUregex needs to be compiled)
-AC_MSG_RESULT($enable_gnuregex)
-AS_IF([test "x$enable_gnuregex" = "xyes"],[
-  # for some reason (force-enable, test..) gnuregex was found as needed. Override any system lib
-  REGEXLIB=""
+SQUID_CHECK_REGEX_WORKS
+AS_IF([test "x$squid_cv_regex_works" = "xno"],[
+  AC_MSG_ERROR([Could not detect a functioning regex library.
+                If your OS does not provide one, please install gnulib and retry the build])
```
rousskov, line 2974: Please drop this and SQUID_CHECK_REGEX_WORKS itself. We do not need to (and, hence, should not half-)check, at ./configure time, that non-optional features compile (which is what  SQUID_CHECK_REGEX_WORKS does in this PR). We should rely on the compiler to detect and report any specific problems (when building Squid) instead of poorly duplicating that activity at ./configure time.

rousskov (changes requested):
Thank you for removing that old and problematic code! My change requests are trivial to address, but I consider (one of) them important enough to block this change.

configure.ac
kinkie: Done. It would give us a chance to give a more meaningful error message, but in practice I don't think anyone would ever see it anyway

CREDITS
kinkie: Done. Editor is set to remove white space at end of line and I didn't notice.
Maybe we could clear it in a separate PR and make it part of source-maintenance.sh?

configure.ac
rousskov: > It would give us a chance to give a more meaningful error message

Actually, in many problematic cases, the error message would be more _meaningful_ when it happens during the primary build phase (or later), but this (debatable and varying) message quality is not the primary reason for removing this poor code duplication. Thank you for removing it!

CREDITS
rousskov: > Editor is set to remove white space at end of line and I didn't notice.

I know. We all have been there. FWIW, I recommend running a git diff against the branch forking point _before_ posting a PR _and_ looking at the GitHub-generated diff afterwards. I have caught lots of problems following this simple routine.

>  Maybe we could clear it in a separate PR

That would be nice, especially if you apply that fix to `git grep '\s$'` (with exceptions for RFCs and such) rather than a single file.

> and make it part of source-maintenance.sh?

It already is a part of test-suite/test-sources.sh (that automatically runs on GitHib Actions). Look for `diff --check` there. We should start to enforcing those rules soon (the check failures are currently reported but otherwise ignored).

rousskov (approved):
Thank you.

tools/purge/conffile.cc
```diff
@@ -47,9 +47,14 @@
...
 #include <cstring>
 #include <fstream>
-#include <sys/types.h>
 #include <memory.h>
 
+#if HAVE_REGEX_H
+#include <regex.h>
+#endif
+
+#include <sys/types.h>
```
yadij, line 56: Note the detail from now removed compat/GnuRegex.h :
```
/* POSIX says that <sys/types.h> must be included (by the caller) before
 * <regex.h>.  */
```

yadij (changes requested):

tools/purge/conffile.cc
kinkie: Thanks for catching. According to https://pubs.opengroup.org/onlinepubs/009695399/basedefs/regex.h.html , that' in order to get size_t.

I'll add sys/types.h before regex.h whenever included.

yadij: It's probably not worth keeping these sys/types.h includes at all. squid.h should be pulling it in early already.
With the switch to std::regex coming shortly I don't think it matters if done the way you have applied already, so not complaining about that.

yadij (approved):

yadij:
FYI, I have begun updating my std::regex patch already and will submit followup PR shortly.

kinkie:
> FYI, I have begun updating my std::regex patch already and will submit followup PR shortly.

Awesome!

tools/purge/conffile.cc
rousskov: > ```C++
> /* POSIX says that <sys/types.h> must be included (by the caller) before
>  * <regex.h>.  */
> ```

That assertion used to be correct, but it became false since POSIX 1003.1-2001. See https://sourceware.org/bugzilla/show_bug.cgi?id=2534

Modern regex.h includes the necessary dependencies itself as needed, of course. For example: https://codebrowser.dev/glibc/glibc/posix/regex.h.html#23

src/base/RegexPattern.h
```diff
@@ -9,10 +9,15 @@
 #ifndef SQUID_SRC_BASE_REGEXPATTERN_H
 #define SQUID_SRC_BASE_REGEXPATTERN_H
 
-#include "compat/GnuRegex.h"
 #include "mem/forward.h"
 #include "sbuf/SBuf.h"
 
+#if HAVE_REGEX_H
+// regex.h requires size_t from sys/types.h
+#include <sys/types.h>
```
rousskov, line 17: Please remove these: regex.h does not require manual sys/types.h inclusion since POSIX 1003.1-2001. See https://sourceware.org/bugzilla/show_bug.cgi?id=2534

Modern regex.h includes the necessary dependencies itself as needed, of course. For example: https://codebrowser.dev/glibc/glibc/posix/regex.h.html#23 . If we are modernizing based on recent code, we should take advantage of improvements made since 1996 instead of increasing code pollution with poorly duplicated sys/types.h `#include` statements.

kinkie: Done

tools/purge/conffile.cc
```diff
@@ -47,9 +47,14 @@
 #include <cstdlib>
 #include <cstring>
 #include <fstream>
-#include <sys/types.h>
```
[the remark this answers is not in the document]
kinkie: This was done to follow our sorting conventions on headers.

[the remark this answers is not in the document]
rousskov: I thought that (insufficiently documented?) convention does not apply to cases like this, but perhaps it does. Thank you for clarifying the reasons behind this change. Resolving.

PR #1232, opened 2023-01-08
Use std::regex for RegexPattern

yadij:
Pave the way for modern regex use. C++11 std::regex provides a
portable API for pattern matching and allows future addition of
flags to select different syntax languages at admin choice.

Older code still needing to be upgraded from legacy regcomp(3)
API is not touched. It should be converted to use RegexPattern
objects instead of directly using a std::regex.


src/base/RegexPattern.cc
```diff
@@ -8,34 +8,28 @@
...
-        char errBuf[256];
-        // for simplicity, ignore any error message truncation
-        (void)regerror(errCode, &regex, errBuf, sizeof(errBuf));
-        // POSIX examples show no regfree(&regex) after a regcomp() error;
-        // presumably, regcom() frees any allocated memory on failures
-        throw TextException(ToSBuf("POSIX regcomp(3) failure: (", errCode, ") ", errBuf,
-                                   Debug::Extra, "regular expression: ", pattern), Here());
-    }
-
+    assert(aFlags != 0);
```
yadij, line 20: Due to the current messy situation with pattern syntax' we cannot enforce std::regex::basic flag existence. Instead for now require that the caller sets the flags explicitly.

kinkie:
Looks pretty noncontroversial to me

yadij:
The situation with regex syntax support is currently quite messy. The outdated GNURegex library only really supported BRE (```std::regex::basic```) syntax. AFAIK that library historically being the default on many OS is behind the default for regcomp(3) being the BRE syntax, with ```REG_EXTENDED```/```std::regex::extended``` flag required to use the ERE syntax.

Since ae315d9cf654c284ff53f4b6ed35bf0fc7461536 the squid.conf settings supporting +i/-i flags (and only those) have also supported ERE patterns - but for simplicity when helping users all our documentation and squid-users advise is that admin should only use BRE  in squid.conf. Even though ERE can usually work.
 This PR seeks to retain the status quo during the transition. Leaving mess reduction to later.

One danger we face with this PR is that the default for ```std::regex``` is to use ECMAScript as pattern language. Which has some syntax reserved-characters that can cause significantly weird matches if one is expecting the POSIX BRE/ERE.
 I have done some checking that the callers are all explicitly passing in ```std::regex::extended```, but a second (or more) eyes to verify that would be good.

yadij:
Rebased after 2e0942f94f80eb33c51e3dfa5d1fde35783fd8c8 merge

src/acl/RegexData.cc
```diff
@@ -108,13 +108,13 @@ removeUnnecessaryWildcards(char * t)
 }
 
 static void
-compileRE(std::list<RegexPattern> &curlist, const SBuf &RE, int flags)
+compileRE(std::list<RegexPattern> &curlist, const SBuf &RE, const std::regex::flag_type flags)
 {
     curlist.emplace_back(RE, flags);
```
rousskov, line 113: This whole configuration code (including, but not just compileOptimisedREs()) is clearly written to prioritize runtime performance over configuration-time overheads (and I agree with that tradeoff). AFAICT, we should add `std::regex::optimize` to the flags. If we should not, it would be a good idea to explain (via a source code comment) why not. 

```suggestion
    curlist.emplace_back(RE, flags | std::regex::optimize);
```

Alternatively, this extra flag would go into the RegexPattern constructor itself. We can probably make the same tradeoff claim for all RegexPattern users!

src/base/RegexPattern.cc
```diff
@@ -8,34 +8,29 @@
...
-#include <utility>
 
-RegexPattern::RegexPattern(const SBuf &aPattern, const int aFlags):
+RegexPattern::RegexPattern(const SBuf &aPattern, const std::regex::flag_type aFlags):
     pattern(aPattern),
-    flags(aFlags)
+    regex(aPattern.rawContent(), aPattern.length(), aFlags)
 {
-    memset(&regex, 0, sizeof(regex)); // paranoid; POSIX does not require this
-    if (const auto errCode = regcomp(&regex, pattern.c_str(), flags)) {
```
rousskov, line 23: Have you tried to measure the performance impact of us switching to std::regex? I understand that the performance depends on many variables, but it would be good to confirm that we are not slowing down Squids a lot by this upgrade in a typical/popular environment with some basic regexes (at least).

src/base/RegexPattern.cc
```diff
@@ -8,34 +8,29 @@
...
-        // for simplicity, ignore any error message truncation
-        (void)regerror(errCode, &regex, errBuf, sizeof(errBuf));
-        // POSIX examples show no regfree(&regex) after a regcomp() error;
-        // presumably, regcom() frees any allocated memory on failures
-        throw TextException(ToSBuf("POSIX regcomp(3) failure: (", errCode, ") ", errBuf,
-                                   Debug::Extra, "regular expression: ", pattern), Here());
-    }
-
+    assert(aFlags != 0);
+    assert((aFlags & (std::regex::basic|std::regex::extended)) != 0);
```
rousskov, line 21: We should explain this puzzling assertion. I do not insist on any specific wording, but here is a sketch:

```suggestion
    // this class supports other syntax variations, but its current users must
    // support one of these two for backward compatibility reasons, and we check
    // that they have not forgotten to do so
    assert((aFlags & (std::regex::basic|std::regex::extended)) != 0);
```

rousskov (changes requested):
This is a partial review in hope to help you make progress. I will need to come back at least for double checking the syntax mess you were [concerned](https://github.com/squid-cache/squid/pull/1232#issuecomment-1374793573) about.

src/base/RegexPattern.cc
```diff
@@ -8,34 +8,29 @@
...
 {
-    memset(&regex, 0, sizeof(regex)); // paranoid; POSIX does not require this
-    if (const auto errCode = regcomp(&regex, pattern.c_str(), flags)) {
-        char errBuf[256];
-        // for simplicity, ignore any error message truncation
-        (void)regerror(errCode, &regex, errBuf, sizeof(errBuf));
-        // POSIX examples show no regfree(&regex) after a regcomp() error;
-        // presumably, regcom() frees any allocated memory on failures
-        throw TextException(ToSBuf("POSIX regcomp(3) failure: (", errCode, ") ", errBuf,
-                                   Debug::Extra, "regular expression: ", pattern), Here());
```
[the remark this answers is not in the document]
yadij: Done in [1e75a9f](https://github.com/squid-cache/squid/pull/1232/commits/1e75a9f420e2ac1283c8a9e44d3f36830c4f5b49)

Old output was:
`
2023/02/21 00:47:52| Processing: refresh_pattern -i ) 0 0% 0
2023/02/21 00:47:52| ERROR: configuration failure: Mismatched '(' and ')' in regular expression
2023/02/21 00:47:52| FATAL: Bungled ./foo line 2: refresh_pattern -i ) 0 0% 0
`

New output is:
`
2023/02/21 00:47:52| Processing: refresh_pattern -i ) 0 0% 0
2023/02/21 00:47:52| ERROR: configuration failure: Mismatched '(' and ')' in regular expression: )
    exception location: src/base/RegexPattern.cc(22) RegexPattern
2023/02/21 00:47:52| FATAL: Bungled ./foo line 2: refresh_pattern -i ) 0 0% 0
`

src/base/RegexPattern.cc
```diff
@@ -8,34 +8,29 @@
...
+    assert((aFlags & (std::regex::basic|std::regex::extended)) != 0);
     debugs(28, 2, *this);
 }
 
-RegexPattern::~RegexPattern()
+bool
+RegexPattern::match(const char *str) const
 {
-    regfree(&regex);
+    std::cmatch found;
```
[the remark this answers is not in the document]
yadij: Dropped for now in [e1a6184](https://github.com/squid-cache/squid/pull/1232/commits/e1a618429d55dae569c9bfecf19f4ec760d8070d)

src/base/RegexPattern.h
```diff
@@ -26,33 +24,29 @@ class RegexPattern
...
+    /// a regular expression in the text form
     SBuf pattern;
 
-    /// bitmask of REG_* flags for regcomp(3)
-    const int flags;
-
-    /// a "compiled pattern buffer" filled by regcomp(3) for regexec(3)
-    regex_t regex;
+    /// a "compiled pattern buffer"
+    std::regex regex;
```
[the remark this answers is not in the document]
yadij: Latest constructor is changing it after construction.

src/acl/RegexData.cc
yadij: Done in [b66b827](https://github.com/squid-cache/squid/pull/1232/commits/b66b827af33102ee75dbd7d4c8ce114fb396500f)

src/base/RegexPattern.cc
yadij: Have now done a performance check with a test program (not squid itself) - performing regex search/match 100K times to match a URL. Inefficient, but the pattern is one admin seems to like using.

regcomp() ...  35ms +/- 1ms.
std::regex  ...  480-500ms +/- 10ms

yadij: For the heck of it I tried the other regex syntax and they all turned out worse performance by 30-200ms than POSIX ERE.

yadij: Doh. test did not have optimizations involved. Simply adding `-O2` we get far better numbers

regcomp() ... 35ms +/- 1ms.
std::regex ... 98ms +/- 8ms

src/base/RegexPattern.cc
```diff
@@ -8,34 +8,33 @@
...
-        // for simplicity, ignore any error message truncation
-        (void)regerror(errCode, &regex, errBuf, sizeof(errBuf));
-        // POSIX examples show no regfree(&regex) after a regcomp() error;
-        // presumably, regcom() frees any allocated memory on failures
-        throw TextException(ToSBuf("POSIX regcomp(3) failure: (", errCode, ") ", errBuf,
-                                   Debug::Extra, "regular expression: ", pattern), Here());
+    try {
+        regex = std::regex(pattern.rawContent(), pattern.length(), aFlags|std::regex::nosubs|std::regex::optimize);
+    } catch (const std::regex_error &e) {
+        throw TextException(ToSBuf(e.what(), ": ", pattern), Here());
```
rousskov, line 22: Please catch and annotate _all_ exceptions (and not just std::regex_error) and supply local context instead of relying on the caught exception (that we do not control) to be descriptive enough to explain what configuration step/action has failed. The official code got this right (where applicable), and we should continue to use that error reporting approach. For example (untested, meant as a sketch; as always where I do not indicate otherwise):

```suggestion
    } catch (...) {
        throw TextException(ToSBuf("regex compilation failure:", CurrentException,
                                   Debug::Extra, ": regular expression: ", pattern), Here());
```

src/base/RegexPattern.cc
```diff
@@ -8,34 +8,33 @@
...
 }
 
-RegexPattern::~RegexPattern()
+bool
+RegexPattern::match(const char *str) const
 {
-    regfree(&regex);
+    if (std::regex_search(str, regex))
+        return true;
+    return false;
```
rousskov, line 37: Jut to simplify:

```suggestion
    return std::regex_search(str, regex);
```

src/base/RegexPattern.h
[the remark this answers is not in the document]
rousskov: > Latest constructor is changing it after construction.

Yes, but only because the current implementation did not follow my explicit [suggestion](https://github.com/squid-cache/squid/pull/1232#r1064218193) on how to avoid "changing it after construction" AFAICT. Please try my suggestion and see if that allows us to make this data member constant. I will quote it here of convenience:

> I believe this reporting can be easily done by creating the regex in a simple helper function that takes the same arguments as the RegexPattern constructor, catches STL exceptions and converts them into a TextException with the pattern detail

The catching and throwing code exists already. Please let me know if I should try making the corresponding adjustments instead.

src/base/RegexPattern.h
```diff
@@ -26,33 +24,28 @@ class RegexPattern
...
 
     /// whether the regex differentiates letter case
-    bool caseSensitive() const { return !(flags & REG_ICASE); }
+    bool caseSensitive() const { return !(regex.flags() & std::regex::icase); }
 
     /// whether this is an "any single character" regex (".")
     bool isDot() const { return pattern.length() == 1 && pattern[0] == '.'; }
 
-    bool match(const char *str) const {return regexec(&regex,str,0,nullptr,0)==0;}
+    bool match(const char *) const;
```
rousskov, line 36: Please also add an SBuf matching method and remove the corresponding XXX/c_str() from refreshCheck(). Please check for other places where a RegexPattern::match() caller is using an SBuf or a buffer/length pair. If the latter cases exist, please add a method for that case as well.

Those changes were waiting for this migration to std::regex (or another modern library), and they may offset some of the performance overheads added by this migration.

src/base/RegexPattern.h
```diff
@@ -26,33 +24,28 @@ class RegexPattern
 
 public:
     RegexPattern() = delete;
-    RegexPattern(const SBuf &aPattern, int aFlags);
-    ~RegexPattern();
-
     RegexPattern(RegexPattern &&) = delete; // no copying of any kind
+    RegexPattern(const SBuf &aPattern, std::regex::flag_type);
```
rousskov, line 28: I would promise to add certain flags, so that the callers do not have to bother and can rely on us doing that:

```suggestion
    /// std::regex::nosubs and std::regex::optimize flags are added automatically
    RegexPattern(const SBuf &aPattern, std::regex::flag_type);
```

rousskov (changes requested):
Very good progress, thank you.

src/base/RegexPattern.cc
rousskov: This change request will also make some errors more readable (because the code will assume much less about the STL error text that we do not control). For example, this particular STL implementation ends errors with periods (`.`), resulting in slightly odd/misleading `escape character.: ...` output:

```
2023/02/20 14:10:04| ERROR: configuration failure: Unexpected escape character.: "foo\"
    exception location: src/base/RegexPattern.cc(22) RegexPattern
```

BTW, please disclose regex pattern (enforcement) changes in the PR description. See our CI tests (from where the above error was pasted).

src/base/RegexPattern.h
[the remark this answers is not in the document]
yadij: I saw the suggestion, I do not see a private member being const as more beneficial than avoiding the +1 data copy for function return. Especially since the object itself is apparently complex enough to require a PoolingAllocator.

src/base/RegexPattern.cc
yadij: Done in [0afa2f5](https://github.com/squid-cache/squid/pull/1232/commits/0afa2f57534d6cc9dca9a96b395a133b1904df78)

src/base/RegexPattern.h
yadij: Please note that the only API change to RegexPattern is the type used by flags. Whether the rest of the API supports SBuf (or not) is unrelated to std::regex.

rousskov: > Whether the rest of the API supports SBuf (or not) is unrelated to std::regex

It is related because std::regex, unlike regexec(3), supports matching unterminated buffers. The PR switching from regexec() to std::regex should add the corresponding optimized match() method, especially since Squid code is transitioning to unterminated buffers, and there is at least one existing XXX waiting for such a match() method. The related optimization will also help partially mitigate performance regression added by this PR. It is simple change. I am happy to do it myself if you prefer.

src/base/RegexPattern.h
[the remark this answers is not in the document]
rousskov: > I saw the suggestion, I do not see a private member being const as more beneficial than avoiding the +1 data copy for function return. 

C++ does not work that way. In relevant cases, the suggested code results in _fewer_ operations than the current PR code. See copy elision [documentation](https://en.cppreference.com/w/cpp/language/copy_elision) for technical details.

* Without compiler optimizations (`-O0`; an unimportant in this performance discussion but possibly easier to understand case), the current PR code creates two std::regex objects and then assigns one to another. The suggested code creates just one std::regex object and never uses the std::regex assignment, copy or move operators -- this copy elision is perfectly legal since C++11 and required since C++17.
* `-O2` compilations of the current PR code remove most of the above overhead, but still contain such extra operations as creation and copying of an additional std::locale object (used as an std::regex data member)! The assembly code of the simplest PR-like version of RegexPattern constructor is 128 lines. Its suggested replacement equivalent is only 71 lines long. Blindly counting assembly lines is silly, and I am not an assembly expert, but I did look through those lines, and believe that the primary difference in these two numbers is a good illustration of what better C++ code might yield in terms of the overhead reduction.



We should not be debating the benefits of member initialization versus assignment in such a basic context.

----

> Especially since the object itself is apparently complex enough to require a PoolingAllocator.

This change request is about std::regex objects. Those objects do not use a PoolingAllocator (unfortunately). However, you are correct that std::regex should be treated as "complex enough" to stimulate writing copying-avoidance code, such as the code suggested in this change request.

src/base/RegexPattern.cc
rousskov: This change request was not fully addressed by commit [0afa2f5](https://github.com/squid-cache/squid/pull/1232/commits/0afa2f57534d6cc9dca9a96b395a133b1904df78), but since it is now outdated, I will open another/fresh change request about the same problem.

src/base/RegexPattern.cc
```diff
@@ -8,34 +8,34 @@
...
-        (void)regerror(errCode, &regex, errBuf, sizeof(errBuf));
-        // POSIX examples show no regfree(&regex) after a regcomp() error;
-        // presumably, regcom() frees any allocated memory on failures
-        throw TextException(ToSBuf("POSIX regcomp(3) failure: (", errCode, ") ", errBuf,
-                                   Debug::Extra, "regular expression: ", pattern), Here());
+    try {
+        regex = std::regex(pattern.rawContent(), pattern.length(), aFlags|std::regex::nosubs|std::regex::optimize);
+    } catch (const std::regex_error &e) {
+        throw TextException(ToSBuf("bad regular expression",
+                                   Debug::Extra, "message: ", e.what(),
```
rousskov, line 23: Again, please catch and annotate _all_ exceptions (and not just std::regex_error), be more specific about the context (i.e. what operation has failed when processing a bad regular expression), and more consistent (with the existing code) when reporting the caught exception. Here is a sketch:

```suggestion
    } catch (...) {
        throw TextException(ToSBuf("failed to compile a regular expression",
                                   Debug::Extra, "problem: ", CurrentException,
```

P.S. We could also merge CurrentException printing into the first debugs() message line. Squid currently uses both approaches. I think the overall approach in the current PR code and the above suggestion is overall better despite being more verbose: The less we assume about the structure of the caught exception message, the better. This more verbose approach will also allow us to add a "more lines to follow" marker to multiline messages if we decide that we should do that.

rousskov (changes requested):

src/base/RegexPattern.cc
yadij: This try-catch is specifically to add the `pattern` value to the string in `std::regex_error` (very specifically a syntax error in the `pattern` string). All the rest of the details you are requesting are handled by other try-catch at higher levels of config parser logic.

src/base/RegexPattern.h
yadij: It is unrelated because this API change can be implemented without `std::regex` etc. The scope of this PR is the internal implementation of `RegexPattern`. There are many possible followup PRs touching callers and API which are intentionally left out of scope here.

rousskov: > It is unrelated because this API change can be implemented without `std::regex`

This API could not be implemented earlier. Or, to be more precise, it should not have been implemented earlier (because it would hide potentially significant performance overheads from the caller). The migration to `std::regex` is what makes this API possible.

> The scope of this PR is the internal implementation of RegexPattern.

The scope of this PR is what we decide it is. There are many reasonable ways to define that scope, and adding this API would not make the scope of this PR worse. I think this API improvement should be added here (for all the reasons already documented and more), but I give up.

rousskov (changes requested):

src/base/RegexPattern.h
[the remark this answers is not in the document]
yadij: FTR: with GCC 12.4:
```
  error: moving a local object in a return statement prevents copy elision [-Werror=pessimizing-move]
```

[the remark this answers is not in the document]
rousskov: FWIW, I agree with the compiler -- we should not be moving any local objects here. Please let me know if you want me to address this change request (based on the earlier suggestion).

src/base/RegexPattern.h
yadij: PR 1299 does this.

src/base/RegexPattern.cc
yadij: The downside you are failing to notice is the ongoing bloat of cache.log

Simply put: - the only errors which need the pattern value logged are these `std::regex_error` thrown by the parser when there is a problem **in the pattern**. No other errors are related in any way to the pattern value.

rousskov: > The downside you are failing to notice is the ongoing bloat of cache.log

Adding requested details does not increase "cache.log bloat" IMO:

* It is debatable that std::regex_errors "are the only errors which need the pattern value logged". It is possible that the pattern detail will be useful in some non-regex_error contexts.
* It is clear that the other details suggested in this change requests can be useful for all exception types. This change request is not (just) about the pattern!
* Most likely, these errors are fatal configuration-time errors, not frequently repeated transaction-processing errors.

It is also clear that _excluding_ the pattern detail from exceptions other than std::regex_error requires _extra_ code.

Given all these factors, adding that pattern detail to all exceptions is the best overall solution, even if one is concerned that some errors will get a pattern detail that will be useless in those specific cases.

> No other errors are related in any way to the pattern value.

We cannot reach that conclusion because we do not know what those "other errors" might be: There is no good reason for this catch statement to assume that no exception thrown by the current or future try code block can benefit from reporting the pattern (other than the std::regex_error). It is best to assume the opposite, keeping the code simple while providing _potentially_ useful context information in likely fatal configuration errors.

PR #1233, opened 2023-01-08
Convert MimeEntry to use RegexPattern API

yadij:
Restore 318f2c8d3418a73946ae947d84f7696a381628b1
with modifications for current RegexPattern API.

### DIFF
--- a/src/mime.cc
+++ b/src/mime.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 1996-2022 The Squid Software Foundation and contributors
+ * Copyright (C) 1996-2023 The Squid Software Foundation and contributors
  *
  * Squid software is distributed under GPLv2+ license and includes
  * contributions from numerous individuals and organizations.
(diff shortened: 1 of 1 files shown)


yadij:
formatting muddies the diff bady. Review can see the actual change clearer in commit 318affb

src/mime.cc
```diff
@@ -234,17 +234,14 @@ mimeInit(char *filename)
     char buf[BUFSIZ];
```
kinkie, line 234: While we're at it, would it make sense to make this static?

src/mime.cc
```diff
@@ -262,70 +259,73 @@ mimeInit(char *filename)
...
+                    download_option = 1;
+                else if (!strcmp(option, "+view"))
+                    view_option = 1;
+                else
+                    debugs(25, DBG_IMPORTANT, "ERROR: mimeInit: unknown option: '" << buf << "' (" << option << ")");
+            }
 
-        m = new MimeEntry(pattern,re,type,encoding,mode,view_option,
-                          download_option,icon);
+            m = new MimeEntry(SBuf(pattern),type,encoding,mode,view_option,download_option,icon);
```
kinkie, line 317: Isn't pattern already a SBuf?

src/mime.cc
```diff
@@ -262,70 +259,73 @@ mimeInit(char *filename)
...
+            while ((option = strtok(nullptr, w_space)) != nullptr) {
+                if (!strcmp(option, "+download"))
+                    download_option = 1;
+                else if (!strcmp(option, "+view"))
+                    view_option = 1;
+                else
+                    debugs(25, DBG_IMPORTANT, "ERROR: mimeInit: unknown option: '" << buf << "' (" << option << ")");
+            }
 
-        m = new MimeEntry(pattern,re,type,encoding,mode,view_option,
```
rousskov, line 321: Please continue to _compile_ the pattern here, where we parse/isolate it. In other words, pass/move a RegexPattern object into MimeEntry constructor instead of passing `pattern`. Doing so will allow us to extend/improve the pattern configuration syntax without bothering the MimeEntry class that does not have access to this parsing context. For example, imagine adding a new regex configuration syntax that specifies the regex syntax to be used for the being-configured regex. We want to place that future code somewhere here rather than in MimeEntry constructor...

Doing so would require enabling move support in RegexPattern, which we can easily do after #1232.

src/mime.cc
```diff
@@ -262,70 +259,73 @@ mimeInit(char *filename)
...
-            continue;
-        }
+            xstrncpy(chopbuf, buf, BUFSIZ);
 
-        if ((icon = strtok(nullptr, w_space)) == nullptr) {
-            debugs(25, DBG_IMPORTANT, "ERROR: mimeInit: parse failure: '" << buf << "'");
-            continue;
-        }
+            const auto pattern = SBuf(strtok(chopbuf, w_space));
+            if (pattern.isEmpty()) {
```
rousskov, line 280: Do we have to change `pattern` type and initialization? It feels like an unrelated change.

rousskov (commented):
Thank you for this improvement. I recommend waiting until #1232 landing, to keep regex parsing code/logic inside the configuration parsing code (as detailed in one of the change requests).

src/mime.cc
yadij: If we were doing a larger refactor of the mime code that would be one of the actions. But changing the I.O buffer is out of scope for just using RegexPattern.

src/mime.cc
rousskov: @yadij, please note that this PR was cleared for merging without addressing Francesco's and [my](https://github.com/squid-cache/squid/pull/1233/files?w=0#r1064229147) change requests about `pattern` type (those two change requests are related/linked). Those change requests may be completely wrong, but clearing a PR without changing code or even commenting on them increases the pressure on the reviewers to _block_ future PRs (to make sure their minor polishing requests are not similarly seemingly ignored), increasing overheads for everybody.

PR #1234, opened 2023-01-08
Source Format Enforcement

squidadm:


yadij (approved):

yadij:
OK to test

rousskov (commented):
As you know, I am against merging this PR because we should not update copyright years every year IMO, but I see no reason to repeat that 2021 [discussion](https://github.com/squid-cache/squid/pull/763#pullrequestreview-579383507) this year.

FWIW, these copyright year changes missed src/auth/negotiate/kerberos/negotiate_kerberos_auth.8 AFAICT. Translation-related msgid fields also have an increasingly stale year ranges despite these annual updates (`git grep msgid..Copyright`).

PR #1237, opened 2023-01-11
Dns::LookupDetails: convert String to SBuf

kinkie:


rousskov (changes requested):
Thank you for working on this upgrade. Please let me know if you need my help with any of the change requests.

src/peer_select.cc
```diff
@@ -548,7 +549,7 @@ PeerSelector::noteIps(const Dns::CachedIps *ia, const Dns::LookupDetails &detail
         lastError = nullptr;
         if (fs->code == HIER_DIRECT) {
             lastError = new ErrorState(ERR_DNS_FAIL, Http::scServiceUnavailable, request, al);
-            lastError->dnsError = details.error;
+            lastError->dnsError = SBufToString(details.error);
```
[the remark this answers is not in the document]
kinkie: I am a bit concerned about the scope of this PR balooning out of control. I'll give it a try and push back if it does, leaving it instead to a followup PR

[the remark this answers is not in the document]
rousskov: > The idea to start working on this comes from ... an attempt to turn ipcache ... to a single std::unordered_map and std::nth_element for O(n) ageout instead

Hm... Perhaps I am missing something important, but that plan to sort an unordered map does not sound right. Should not we use ClpMap or a similar map+LRU_index combination instead?!

src/dns/LookupDetails.cc
```diff
@@ -16,8 +16,8 @@ Dns::LookupDetails::print(std::ostream &os) const
 {
     if (wait > 0)
         os << "lookup_wait=" << wait;
-    if (error.size())
-        os << " lookup_err=" << error;
+    if (error && error->length())
```
rousskov, line 19: This and similar .length() changes look like red flags to me: Once we have std::optional, we should stop checking error length. Either there is an error or there is not. Two states; one condition (provided by std::optional).

I have not checked everything, but, most likely, you need to refactor LookupDetails constructor to ensure the following invariant: A present error is never empty.

rousskov (changes requested):
Thank you for making great progress with this PR!

src/dns/LookupDetails.cc
kinkie: I agree with you, in theory. But we don't necessarily know - some initializations are forwarded (e.g. IpCacheLookupForwarder::forwardLookup) and might be passing nullptrs or empty strings. And enforcing such an invariant would only be possible at runtime, and it would constitute an API change.
So my strategy here is to keep on supporting the current API promises, and once we are further along in getting rid of char* things when things will become clearer.

src/peer_select.cc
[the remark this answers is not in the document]
kinkie: > Hm... Perhaps I am missing something important, but a plan to sort[1](#user-content-fn-1-5dfd48221387c3baae028553105c21db) an unordered map does not sound right. Should not we use ClpMap or a similar map+LRU_index combination instead?!

We can, and could (I didn't remember about ClpMap when I started).
my experiment here is trading a lower constant O(1) at insertion time, which is needed to reorder the lru_list, in exchange of a different constant O(n) at ageout time - might be larger or smaller: my idea walks and copies once all elements in the cache, instead of only the oldest ones, but then it only has to free/fix one data structure, not two. Hard to say which path is more expensive although I'm happy to accept considering it a slight higher cost for the propsed solution.
Re: using ClpMap, it might end up being more expensive because it strictly enforces never going over quota, while current ipcache implementation will wait until lru eviction time and go over quota in the meanwhile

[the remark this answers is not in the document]
rousskov: Jut to avoid misunderstanding: This discussion is unrelated to this PR. I am happy to continue it here, but it does not affect this PR state AFAICT.

----

> my experiment here is trading a lower constant O(1) at insertion time, which is needed to reorder the lru_list, in exchange of a different constant O(n) at ageout time

O(1) is a indeed a constant with respect to n -- a fixed cost `c` that does not change with `n`.

O(n) is not a constant -- it is a cost or overhead that is grows linearly with `n` (i.e. a cost function in the shape of `a*n + b`).

It is true that in any specific case (i.e. some fixed `n`, `a`, `b`, and `c`), the actual values of those functions may have any relationship. However, all other factors being equal, we do _not_ want to trade a constant O(1) operations in ClpMap for O(n) operations in some new code (not to mention O(n * log(n)) costs from std::nth_element), just because that relationship might be in our favor in _some_ cases. Please do not go further down that route without securing support for such a violation of basic performance principles.

> ClpMap ... might end up being more expensive because it strictly enforces never going over quota, while current ipcache implementation will wait until lru eviction time and go over quota in the meanwhile

In Squid, spreading costs over all transactions is better than slowing down a few unlucky transactions to pay for a tiny speedup enjoyed by many lucky transactions. We cannot always spread the costs evenly, but we should favor algorithms that do.

Also, "going over quota" is a dangerous bug that can be abused by external traffic (and does increase support overheads). In most cases, the performance benefits of violating configuration are not worth the associated headaches.

Finally, another big ClpMap advantage is that it already exists in the official code. We spent a lot of time getting it there. By default, we should attempt to reuse it rather than work hard on avoiding it.

The existing ipcache code has problems. I do not know whether the current dlink_list-based code (i.e. lru_list and friends) can be easily replaced by ClpMap _before_ some of those problems are solved. However, I think we should plan for (and work towards) that eventual replacement unless there are some serious reasons to prefer a new/different cache organization. I do not see any such reasons in this exchange.

src/dns/LookupDetails.cc
yadij: Alex, This addition of complexity is the direct consequence of requiring std::optional (trilean object) to be used to represent a boolean state. Callers and users now need to explicitly differentiate states {never-set OR nil} from {currently-set AND non-nil}.
The original API used the pointer (SBuf is a decorated smart pointer) being set or non-set (aka 0 length/size). Which is categorically simpler.

src/errorpage.cc
```diff
@@ -828,8 +828,8 @@ ErrorState::Dump(MemBuf * mb)
     if (auth_user_request.getRaw() && auth_user_request->denyMessage())
         str.appendf("Auth ErrMsg: %s\r\n", auth_user_request->denyMessage());
 #endif
-    if (dnsError.size() > 0)
-        str.appendf("DNS ErrMsg: %s\r\n", dnsError.termedBuf());
+    if (dnsError && dnsError->length() > 0)
+        str.appendf("DNS ErrMsg: %s\r\n", dnsError->c_str());
```
yadij, line 832: Please avoid ```c_str()``` whenever possible.
```suggestion
        str.appendf("DNS ErrMsg: " SQUIDSBUFPH "\r\n", SQUIDSBUFPRINT(*dnsError));
```

src/errorpage.cc
```diff
@@ -1212,8 +1212,8 @@ ErrorState::compileLegacyCode(Build &build)
 
     case 'z':
         if (building_deny_info_url) break;
-        if (dnsError.size() > 0)
-            p = dnsError.termedBuf();
+        if (dnsError && dnsError->length() > 0)
```
yadij, line 1215: please use
```suggestion
        if (dnsError && !dnsError->isEmpty())
```

src/fqdncache.cc
```diff
@@ -304,7 +304,7 @@ fqdncacheCallback(fqdncache_entry * f, int wait)
     f->handler = nullptr;
 
     if (cbdataReferenceValidDone(f->handlerData, &cbdata)) {
-        const Dns::LookupDetails details(f->error_message, wait);
+        const Dns::LookupDetails details(SBuf(f->error_message), wait);
```
yadij, line 307: In order to use std::optional this caller code needs to determine whether an error message exists and construct LookupDetails with different constructors - one which sets error, and one which only sets wait time (not yet existing).

src/ipcache.cc
```diff
@@ -280,7 +280,7 @@ IpCacheLookupForwarder::forwardLookup(const char *error)
     // are sequential. Give it just the new, yet-unaccounted-for delay.
     if (receiverObj.set()) {
         if (auto receiver = receiverObj.valid()) {
-            receiver->noteLookup(Dns::LookupDetails(error, additionalLookupDelay()));
+            receiver->noteLookup(Dns::LookupDetails(SBuf(error), additionalLookupDelay()));
```
yadij, line 283: In order to use std::optional this caller code needs to determine whether an error message exists and construct LookupDetails with different constructors - one which sets error, and one which only sets wait time (not yet existing).

src/ipcache.cc
```diff
@@ -445,7 +445,7 @@ ipcacheCallback(ipcache_entry *i, const bool hit, const int wait)
 
     if (hit)
         i->handler.forwardHits(i->addrs);
-    const Dns::LookupDetails details(i->error_message, wait);
+    const Dns::LookupDetails details(SBuf(i->error_message), wait);
```
yadij, line 448: In order to use std::optional this caller code needs to determine whether an error message exists and construct LookupDetails with different constructors - one which sets error, and one which only sets wait time (not yet existing).

Note that this is now the third caller which needs to perform those same duplicate actions just to create a Dns::LookupDetails object "correctly" (avoiding setting the std::optional on non-errors). All of this extra complexity is imposed by using std::optional instead of the old API style of empty string check on use.

yadij (changes requested):
Most of these issues are side effect of std::optional. I would prefer reverting that and going just with the String to SBuf conversion for now.

src/dns/LookupDetails.cc
rousskov: > This addition of complexity is the direct consequence of requiring std::optional

Where do you see additional complexity (in this change request scope)? Here are the sketches for all three options:

Official code:

```C++
if (error.size())
    os << " lookup_err=" << error;
```

With std::optional:

```C++
if (error)
    os << " lookup_err=" << *error;
```

With SBuf:

```C++
if (error.length())
    os << " lookup_err=" << error;
```

I do not see any additional complexity. In fact, one could argue that std::optional makes this code a bit simpler or more readable. Additional complexity does exist in the current PR, but that code should be fixed. This change request is about that fix.

src/fqdncache.cc
rousskov: > In order to use std::optional this caller code needs to determine whether an error message exists and construct LookupDetails with different constructors

There is no such requirement. LookupDetails constructor can work correctly with the existing SBuf API. It is not the most efficient way, but it is simple and is a good _next_ step.

Said that, it is possible, perhaps even likely, that `f->error_message` should be converted to std::optional as well and/or that would we should add another constructor or make some other API changes. I am moving one step at a time here.

I will not repeat this comment in other similar contexts, but it does apply.

kinkie:
We have a deadlock, because two reviewers blocked on different requirements.
TBH, I would choose Amos' suggestion as it changes the fewer assumptions and is the least invasive, while still providing benefits.
@rousskov , are you okay with that?

rousskov:
> We have a deadlock, because two reviewers blocked on different requirements.

IMO, I have resolved that deadlock by correcting false assumptions behind conflicting requirements at https://github.com/squid-cache/squid/pull/1237#discussion_r1068599860

> TBH, I would choose Amos' suggestion as it changes the fewer assumptions and is the least invasive, while still providing benefits.

If "Amos's suggestion" is "use empty `error` member to signal no error", then I see no good reason to keep that design mistake when changing `error` type. This is the best opportunity we will have to fix that mistake, and it is an easy fix. In C++, optional information is usually expressed using pointers (for heap-allocated objects) or std::optional (for copied objects). When changing types, we should use the correct new type unless there are some very compelling reasons to make an exception. BTW, "using a correct type" is not an opinion -- it is a basic engineering principle or what you might call "data" in your conflict resolution formula. Hopefully, we do not need to argue what std::optional is for. In summary, the "data" tells you to do the right thing -- use std::optional for an optional data member.

AFAICT, past objections against using a correct type were based on false assumptions (that I have [detailed](https://github.com/squid-cache/squid/pull/1237#discussion_r1068599860)) -- the final PR code will not be more complex or otherwise "worse" from using that correct type, in any meaningful sense of that phrase. Using std::optional does not change any important assumptions and is not more invasive, in any meaningful sense of that phrase. It does result in better (e.g., safer and, eventually, faster) code.

> @rousskov , are you okay with that?

I am not. I recommend following my suggestions for fixing the current PR code and then, after PR code is ready again, returning to Amos for re-review. Hopefully, by that time, it would be obvious that std::optional was the right solution. I am willing to do this legwork myself (and/or undo its results if I am proven wrong) if you prefer -- just ask.

src/dns/LookupDetails.cc
yadij: The complexity is that you are wrong about this case (for now, with incomplete transition):

> With std::optional:
> 
> ```c++
> if (error)
>     os << " lookup_err=" << *error;
> ```
> 

With creators of LookupDetails initializing the ```error``` member to empty string on non-error responses the recipients if-statement has to be in the form @kinkie used.

src/fqdncache.cc
yadij: AFAICS, kinkie is trying to take a single refactor step in this PR (String->SBuf only). Whereas you are requesting a skip straight to step 3, ignoring the additional changes of step 2 required to make the step 3 be as efficient as you claim it is.

src/dns/LookupDetails.cc
rousskov: @yadij, I believe you are reaching incorrect conclusions because you are looking at (unfinished) PR code. I am looking at how the code will look like when this PR code is fixed.

@kinkie, this is exactly why I have recommended finishing this work _before_ asking Amos to re-review (or concluding that there is some real deadlock here).

src/fqdncache.cc
rousskov: > AFAICS, kinkie is trying to take a single refactor step in this PR (String->SBuf only).

Yes, and stopping at that first step does more harm than good in this particular case. We have to spend time reviewing every `error` user anyway. We should use that to switch to the _correct_ type for this optional field instead of raw SBuf.

> Whereas you are requesting a skip straight to step 3, ignoring the additional changes of step 2 required to make the step 3 be as efficient as you claim it is.

Or you can say that I am requesting to go straight to step 25. The number of steps does not really matter here because these steps are tiny, natural, there are people willing to make them, and this PR is the right/best place to make them.

kinkie: > > AFAICS, kinkie is trying to take a single refactor step in this PR (String->SBuf only).
> 
> Yes, and stopping at that first step does more harm than good in this particular case. We have to spend time reviewing every `error` user anyway. We should use that to switch to the _correct_ type for this optional field instead of raw SBuf.

@yadij is correct, I am taking an incremental approach.
I also agree that moving to std::optional is conceptually more correct, because the current API being "empty string == no error" can be improved upon.
I do not agree that stopping at String -> SBuf is worse than the current status. I entered this specific rabbit hole because of a single caller (ipcache) needing to keep around a c-style string (the hash index) to pass it to DNS lookup which is inefficient and makes memory ownership unclear, and that is the part I would like to focus on next, as we discussed elsewhere in this PR.
So by stopping here, we would be missing an opportunity to improve an API but not make it worse; it would instead give me an opportunity to focus on an interesting use case and get benefits right away.
A way to get a bit of both is to add a boolean getter which for now would check the string length, and eventually move to checking std::optional. We could use it to clarify intent on part of the callers and facilitate the transition.

> > Whereas you are requesting a skip straight to step 3, ignoring the additional changes of step 2 required to make the step 3 be as efficient as you claim it is.
> 
> Or you can say that I am requesting to go straight to step 25. The number of steps does not really matter here because these steps are tiny, natural, there are people willing to make them, and this PR is the right/best place to make them.

I would rather use followup PRs; what do you think of my proposed strategy above?

rousskov: > stopping here, we would be missing an opportunity to improve an API but ... give me an opportunity to focus on an interesting use case and get benefits right away.

Nothing in my change requests prevents your from focusing on whatever you want to focus on or delaying benefits by any amount of time worth talking about, especially given the amount of time already spent on this PR. Even if you did not want to follow my suggestions, you could simply say "if you can do it quickly, please address your change requests and we will see whether the results are worth keeping". That takes approximately one minute for you to write and 30 minutes for me to implement. Compare that with more than an hour we have already _wasted_ on useless discussions.

> A way to get a bit of both is to add a boolean getter

Please see #1241 for what I consider the correct way to address my change requests. That PR does not add any methods.

> I would rather use followup PRs; what do you think of my proposed strategy above?

I would rather merge #1241 which starts with where this PR is right now and then addresses my change requests and Amos' change requests (that did not reach wrong conclusions based on unfinished PR code).

Followup PRs are needed indeed, but #1241 represents a minimal change IMO (not from "a minimal change that compiles" point of view, but from a "minimal change that is worth merging" point of view).

rousskov: @kinkie, please consider closing this PR because its improvements are covered by the merged PR #1241. Thank you.

kinkie:
Reimplemented as PR #1241

PR #1238, opened 2023-01-12
Bug 5256: Intercepting port fails to accept

rousskov:
ERROR: NF getsockopt(ORIGINAL_DST) failed on ... Bad file descriptor
    ERROR: NAT lookup failed to locate original IPs on ...

Ip::Interceptor.LookupNat() needs an open Connection, but commit 7185c9e
supplied a half-baked details object, resulting in ERRORs (showing a
closed connection -- no FD... field) for every otherwise successful
accept(2) attempt on an intercepting port.

Refactored oldAccept() to use exceptions for error handling and false
return result for no-error-but-no-connection cases (Comm::NOMESSAGE).
This refactoring allows us to centralize connection closing code instead
of chasing individual COMM_ERROR and NOMESSAGE cases while still missing
connection closure (if a function called by oldAccept() throws). With
connection closure handled, we can now open the details Connection
earlier, as we have done before commit 7185c9e, meeting current
LookupNat() and other/future code expectations.

Positive side effects of this fix include elimination of the following
old error reporting problems that left admins puzzled why Squid is not
handling new connections with no error messages in non-debugging
cache.log (and with the socket listening queue growing).

* Silent lack of accept() after an ENFILE or EMFILE failure.
* Silent lack of accept() if some function used by oldAccept() throws.
* Silent at level-0 lack of accept() after logging a level-1 ERROR. This
  problem was specific to ALL,0 and similar (rare) configurations.

### DIFF
--- a/src/comm/TcpAcceptor.cc
+++ b/src/comm/TcpAcceptor.cc
@@ -394,10 +394,13 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
         }
     } else if (conn->flags & COMM_INTERCEPTION) { // request the real client/dest IP address from NAT
         details->flags |= COMM_INTERCEPTION;
+        details->fd = sock; // XXX
         if (!Ip::Interceptor.LookupNat(*details)) {
+            details->fd = -1; // XXX
             debugs(50, DBG_IMPORTANT, "ERROR: NAT lookup failed to locate original IPs on " << details);
             return Comm::NOMESSAGE;
         }
+        details->fd = -1; // XXX
     }
 
 #if USE_SQUID_EUI
--- a/src/comm/TcpAcceptor.cc
+++ b/src/comm/TcpAcceptor.cc
@@ -28,6 +28,7 @@
 #include "ip/QosConfig.h"
 #include "log/access_log.h"
 #include "MasterXaction.h"
+#include "sbuf/Stream.h"
 #include "SquidConfig.h"
 #include "StatCounters.h"
 
@@ -271,33 +272,40 @@ Comm::TcpAcceptor::acceptOne()
 
     /* Accept a new connection */
     ConnectionPointer newConnDetails = new Connection();
-    const Comm::Flag flag = oldAccept(newConnDetails);
+    try {
+        if (acceptInto(newConnDetails)) {
+            Assure(newConnDetails->isOpen());
+            CallBack(newConnDetails, [&] {
+                debugs(5, 5, "Listener: " << conn <<
+                       " accepted new connection " << newConnDetails <<
+                       " handler Subscription: " << theCallSub);
+                notify(Comm::OK, newConnDetails);
+            });
+        } else {
+            debugs(5, 5, "try later: " << conn << " handler Subscription: " << theCallSub);
+            newConnDetails->close(); // paranoid manual closure (and may already be closed)
+        }
 
-    if (flag == Comm::COMM_ERROR) {
-        // A non-recoverable error; notify the caller */
-        debugs(5, 5, "non-recoverable error:" << status() << " handler Subscription: " << theCallSub);
-        if (intendedForUserConnections())
-            logAcceptError(newConnDetails);
-        notify(flag, newConnDetails);
-        // XXX: not under async job call protections
-        mustStop("Listener socket closed");
+        SetSelect(conn->fd, COMM_SELECT_READ, doAccept, this, 0);
         return;
+    } catch (...) {
+        const auto debugLevel = intendedForUserConnections() ? DBG_CRITICAL : 3;
+        debugs(5, debugLevel, "ERROR: Stopped accepting connections:" <<
+               Debug::Extra << "error: " << CurrentException);
     }
 
-    if (flag == Comm::NOMESSAGE) {
-        /* register interest again */
-        debugs(5, 5, "try later: " << conn << " handler Subscription: " << theCallSub);
-    } else {
-        // TODO: When ALE, MasterXaction merge, use them or ClientConn instead.
-        CallBack(newConnDetails, [&] {
-            debugs(5, 5, "Listener: " << conn <<
-                   " accepted new connection " << newConnDetails <<
-                   " handler Subscription: " << theCallSub);
-            notify(flag, newConnDetails);
-        });
-    }
+    if (intendedForUserConnections())
+        logAcceptError(newConnDetails);
 
-    SetSelect(conn->fd, COMM_SELECT_READ, doAccept, this, 0);
+    // do not expose subscribers to a "usable" descriptor of a failed connection
+    newConnDetails->close(); // may already be closed
+
+    CallBack(newConnDetails, [&] {
+        notify(Comm::COMM_ERROR, newConnDetails);
+    });
+
+    // TODO: Place under AsyncJob call protections and call mustStop() instead.
+    deleteThis("unrecoverable accept failure");
 }
 
 void
@@ -329,17 +337,11 @@ Comm::TcpAcceptor::notify(const Comm::Flag flag, const Comm::ConnectionPointer &
     }
 }
 
-/**
- * accept() and process
- * Wait for an incoming connection on our listener socket.
- *
- * \retval Comm::OK          success. details parameter filled.
- * \retval Comm::NOMESSAGE   attempted accept() but nothing useful came in.
- *                           Or this client has too many connections already.
- * \retval Comm::COMM_ERROR  an outright failure occurred.
- */
-Comm::Flag
-Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
+/// acceptOne() helper: accept(2) a new TCP connection and fill in its details
+/// \retval false indicates that details do not contain a usable TCP connection,
+/// but future attempts to accept may still be successful
+bool
+Comm::TcpAcceptor::acceptInto(Comm::ConnectionPointer &details)
 {
     ++statCounter.syscalls.sock.accepts;
     struct addrinfo *gai = nullptr;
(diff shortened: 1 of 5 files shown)


src/fd.h
```diff
@@ -11,34 +11,6 @@
 #ifndef SQUID_FD_H_
 #define SQUID_FD_H_
 
-namespace Comm {
-
-/// An open Comm-registered file descriptor guard that, upon creation, registers
-/// the descriptor with Comm and, upon destruction, unregisters and closes the
-/// descriptor (unless the descriptor has been release()d by then).
-class Descriptor
```
rousskov, line 19: I removed Comm::Descriptor as no longer used. AFAIK, there are no known problems with that class, and the long-term plan of using it for controlling descriptors lifetime is still valid, but we should not keep unused code in Squid. We can always re-add it later.

src/comm/TcpAcceptor.cc
```diff
@@ -369,7 +368,10 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
     // Sync with Comm ASAP so that abandoned details can properly close().
     // XXX : these are not all HTTP requests. use a note about type and ip:port details->
     // so we end up with a uniform "(HTTP|FTP-data|HTTPS|...) remote-ip:remote-port"
-    Descriptor sock(rawSock, FD_SOCKET, "HTTP Request");
+    const auto sock = rawSock;
+    fd_open(sock, FD_SOCKET, "HTTP Request");
+    details->fd = sock;
+    details->enterOrphanage();
```
rousskov, line 374: The above change brings us back to where this code was before my buggy commit https://github.com/squid-cache/squid/commit/7185c9e7dabe76fce29ea527c52b6053e53493a5. Instead of using Comm::Descriptor to delay "opening" of the half-baked `details` Connection (to avoid "leaking" an open but unusable Connection object to subscribers), we now close such connections in acceptOne() while correctly handling both explicit exceptions (i.e. former COMM_ERROR cases) and implicit ones (broken before that buggy commit).

src/comm/TcpAcceptor.cc
```diff
@@ -271,33 +272,41 @@ Comm::TcpAcceptor::acceptOne()
...
+    newConnDetails->close(); // may already be closed
 
-    SetSelect(conn->fd, COMM_SELECT_READ, doAccept, this, 0);
+    CallBack(newConnDetails, [&] {
+        notify(Comm::COMM_ERROR, newConnDetails);
+    });
+
+    // XXX: Not under AsyncJob call protections but, if placed there, may cause
+    // problems like making the corresponding HttpSockets entry (if any) stale.
+    mustStop("unrecoverable accept failure");
```
rousskov, line 309: I actually addressed this old XXX in this branch by switching from mustStop() to deleteThis(). deleteThis() does work as intended here, closing the listening socket and destroying the TcpAcceptor job, but I reverted those branch changes because:

1. If this TcpAcceptor object is tied to an HttpSockets FD entry, that entry will become stale. That might be OK, but it would be better to avoid creating this problem anyway (probably by carefully removing HttpSockets!).
2. A later clientConnectionsClose() call will mislead about closing (an already closed!) listening socket. It is not clear to me whether there is other code that also assumes a non-nil `HttpPortList[i]->listenConn` implies an open listening connection.
3. We should avoid adding more/new deleteThis() calls.
4. Properly fixing this old XXX (i.e. addressing the above and similar concerns) can be viewed as being outside this PR scope. It should probably be done in several PRs.

For now, Squid will continue to log this (at debug level 5) after unrecoverable accept errors:

```
AsyncJob.cc(77) mustStop: Comm::TcpAcceptor will STALL, reason: unrecoverable accept failure
```

... and the listening queue may grow (except in those SMP environments where other workers can still accept connections queued at the same port/descriptor, I guess).

src/comm/TcpAcceptor.cc
```diff
@@ -271,33 +272,41 @@ Comm::TcpAcceptor::acceptOne()
...
+            Assure(newConnDetails->isOpen());
+            CallBack(newConnDetails, [&] {
+                debugs(5, 5, "Listener: " << conn <<
+                       " accepted new connection " << newConnDetails <<
+                       " handler Subscription: " << theCallSub);
+                notify(Comm::OK, newConnDetails);
+            });
+        } else {
+            debugs(5, 5, "try later: " << conn << " handler Subscription: " << theCallSub);
+            newConnDetails->close(); // paranoid manual closure (and may already be closed)
```
rousskov, line 286: This explicit closure is not really needed because we are not sharing newConnDetails with anybody in this case. I will remove it if others think it should be removed instead:

```suggestion
```

src/comm/TcpAcceptor.cc
```diff
@@ -271,33 +272,41 @@ Comm::TcpAcceptor::acceptOne()
...
-        notify(flag, newConnDetails);
-        // XXX: not under async job call protections
-        mustStop("Listener socket closed");
+    try {
+        if (acceptInto(newConnDetails)) {
+            Assure(newConnDetails->isOpen());
+            CallBack(newConnDetails, [&] {
+                debugs(5, 5, "Listener: " << conn <<
+                       " accepted new connection " << newConnDetails <<
+                       " handler Subscription: " << theCallSub);
```
rousskov, line 281: I think this and a few other old debugs() can be improved, but I preserved them while moving in hope to avoid arguing about such changes (unrelated to this bug fix).

rousskov (commented):
I asked Bug [5256](https://bugs.squid-cache.org/show_bug.cgi?id=5256#c3) reporter to test the interception side of these changes.

src/comm/TcpAcceptor.cc
yadij: Please do close(). "sharing" is irrelevant. The need to close is about leaving the TCP stack with a hung socket. A Comm::Connection always needs to be closed at the point where it becomes known to be unusable.

src/comm/TcpAcceptor.cc
yadij: This XXX will be resolved by PR #870 when it completes.

yadij (approved):
Just the merge description details to fix. Code LGTM.

yadij:
> Silent lack of accept() after an ENFILE or EMFILE failure.

So-called "silence" for these errors is not a problem. Lack of accept() is the actual problem. AFAIK that was introduced with the switch from C-code to TcpAcceptor AsyncJob behaviour. It should be resolved by calling fdAdjustReserved() and pushing an entry to the AcceptLimiter queue for later retry.

> Silent lack of accept() if some function used by oldAccept() throws.

> Silent at level-0 lack of accept() after logging a level-1 ERROR. This
> problem was specific to ALL,0 and similar (rare) configurations.

This is correctly obeying admin preferences. There is no problem here.

Also, please define what you mean by "Silent". The dictionary meaning of that word contradicts the code - all the above cases have cache.log and/or access.log entries associated with the relevant event.

src/comm/TcpAcceptor.cc
rousskov: > Please do keep the close().

Sure, that is the current plan.

> The need to close is about leaving the TCP stack with a hung socket.

Whether we remove this explicit close() or not, this connection will be closed before we leave this function (because this connection object is not shared). However, I think keeping this explicit close() is slightly better than removing it (because we cannot be sure whether this connection object would never be shared with some other job in some future code, even though such sharing would be a bug -- an explicit close() with a comment clarifies our thinking/assumptions a bit).

> A Comm::Connection always needs to be closed at the point where it becomes known to be unusable.

I disagree with the "always" part. In many cases, the connection should indeed be explicitly closed at that point. In some cases, it should be explicitly closed, but a bit later (e.g., when we reach some common cleanup code in the same function). In some other cases, it is better to rely on RAII and other automated mechanisms than explicit closures. Fortunately, we do not need to agree on this for this PR to be merged.

src/comm/TcpAcceptor.cc
```diff
@@ -354,13 +357,9 @@ Comm::TcpAcceptor::oldAccept(Comm::ConnectionPointer &details)
...
             debugs(50, 5, status() << ": " << xstrerr(errcode));
-            return Comm::NOMESSAGE;
-        } else if (errcode == ENFILE || errcode == EMFILE) {
-            debugs(50, 3, status() << ": " << xstrerr(errcode));
-            return Comm::COMM_ERROR;
+            return false;
         } else {
-            debugs(50, DBG_IMPORTANT, "ERROR: failed to accept an incoming connection: " << xstrerr(errcode));
-            return Comm::COMM_ERROR;
+            throw TextException(ToSBuf("Failed to accept an incoming connection: ", xstrerr(errcode)), Here());
```
rousskov, line 362: > > PR description: Silent lack of accept() after an ENFILE or EMFILE failure.
 
> Amos: So-called "silence" for these errors is not a problem.

It is a fairly serious problem IMO: We just spent several man/hours and almost a week troubleshooting a problematic Squid exactly because those failures did not result in any messages in cache.log. It should have taken us about 10 minutes, but without those cache.log messages, it takes a lot more time to figure out what is actually going on.

> Lack of accept() is the actual problem.

Yes, for some of these errors it is indeed, but the two problems (i.e. silence and lack of accept()) are not mutually exclusive. Both problems are real and co-exist.


> > Silent at level-0 lack of accept() after logging a level-1 ERROR. This
> > problem was specific to ALL,0 and similar (rare) configurations.
 
> This is correctly obeying admin preferences. There is no problem here.

I disagree: First of all, when Squid stops accepting connections on a port, it is a level-0 (i.e. _critical_) error. Some would even argue that such  an error should be fatal (but let's not have that debate here and now). The lack of the corresponding level-0 error in the official code is a bug. Secondly, the level-1 ERROR which _is_ logged in some cases, is _not_ about stopping to accept() future connections (i.e. it is _not_ the missing message that the third bullet point talks about). That level-1 ERROR looks like a regular transaction error to most admins.

> Also, please define what you mean by "Silent". The dictionary meaning of that word contradicts the code - all the above cases have cache.log and/or access.log entries associated with the relevant event.

I think debugging modes (i.e. level-2+) should not count by default -- they are not for admins -- but I made that explicit by adding "non-debugging" to the corresponding paragraph above the bullet points. With that out of the way:

The first two of the three no-accept cases do not have the corresponding cache.log messages at all. The third no-accept case does not have a level-0 message (as detailed in the PR description). The paragraph before the three bullet points already summarizes them as "no error messages in cache.log", effectively confirming what "silent" means in this context. If that is not enough, a [dictionary](https://www.merriam-webster.com/dictionary/silent) meaning of that word -- "making no mention" -- matches the official code AFAICT.

I do not think we should account for access.log entries in this context because a lack of an accept is a Squid port-wide problem rather than a transaction-specific problem. Admins should not be expected to fish for reasons why Squid may have stopped performing one of its primary functions in an access.log of a busy proxy, especially when we do not detail those access.log records enough (yet) to know whether a particular accept(2) error resulted in the lack of further accept(2) calls.

@yadij, I plan to merge this PR this weekend with the current (already adjusted with "non-debugging") PR description unless you speed that up or stop me. If you stop me, please suggest the wording you want to see (or simply edit the current one).

PR #1240, opened 2023-01-14
Drop ProxyProtocol::Header::Pointer typedef

yadij:


src/proxyp/Header.h
```diff
@@ -22,7 +22,6 @@ namespace ProxyProtocol {
 class Header: public RefCountable
 {
 public:
-    typedef RefCount<Header> Pointer;
```
rousskov, line 25: Why do we want to remove this `typedef`? The overall approach looks fairly natural and common in Squid code -- my quick-and-dirty search matched dozens of similar in-class Foo::Pointer declarations that also have an out-of-class FooPointer typedefs (e.g., AsyncJobPointer, AccessLogEntryPointer, BodyPipePointer, etc.). Is there something wrong with Header::Pointer?

yadij: Simplicity. There is no need for two typedefs for the same type meaning.

As for the other classes. They may be similar, I am not including them in this scope.

rousskov: > There is no need for two typedefs for the same type meaning.

Both typedefs can be useful:

* FooPointer declarations are useful in forward declarations (that do not need to know/see Foo declaration).
* Foo::Pointer declarations are useful inside the Foo class, inside classes derived from Foo, and in templates that need access to the pointer type corresponding to the template parameter.

Since both type aliases may be useful, I am against removing either of them.

rousskov: > Alex: Since both type aliases may be useful, I am against removing either of them.

I should add that both aliases are also _used_ in the official code.

@kinkie, what is your opinion about this PR? Do you think my objections should be overruled?

kinkie: I don't think it makes a lot of difference either way, so I won't take any strong position.

This said, if I were to pick a style, I'd favour only adopting Foo::Pointer and collapsing all uses onto that one if at all possible.
Let's toss a coin, really. I don't think this is worth the discussion time

rousskov: > I'd favour only adopting Foo::Pointer and collapsing all uses onto that one if at all possible.

It is not possible unless we want to prohibit declarations that use pointers to Foo but do not need to know Foo detail. IMO, we should support those use cases, not prohibit them.

```C++
using FooPointer = ...
// the following declaration will become impossible
// (without exposing caller to Foo details) if we prohibit FooPointers
void Log(const FooPointer &);
```

Please note that this is _not_ a question of stylistic preference, where "flipping a coin" is sometimes the only (bad) approach we have to use.

There are two sets of actual questions here:

1. Do we want to support all the different use cases that the two kinds of pointers support? If yes, then both kinds should be left in the code. Otherwise, one should be removed (and we need to pick which one -- that is, which use cases we want to sacrifice). There are dozens of such Pointers. If this PR sets a precedent, then we should anticipate lots of similar code changes!
2. Is completely ignoring a negative vote considered an acceptable behavior? Should we start blocking PRs just to make sure our concern was not completely ignored?

> I don't think this is worth the discussion time

Modifying dozens of declarations spread around the code (without any functionality gain) is worth discussing IMHO. Most likely, some of us will spend a lot more time (in the long run) resolving the resulting merge conflicts.

kinkie: Thanks for the example; I agree with you then.
As a matter of stylstic preference, I do like being able to rely on Foo::Pointer being there - just like ::iterator, so I'd leave Foo::Pointer in even if not used, and then also define FooPointer when needed

rousskov: @kinkie, I am glad we are on the same page.

@yadij, if this exchange convinced you that having two different ways to name the same underlying type may be useful, then please keep both in Squid.

kinkie:
@yadij if you agree with the arguments put forward here, how about closing the PR?

As a side note, how about starting to use the "using" syntax for new typedefs?

rousskov:
> As a side note, how about starting to use the "using" syntax for new typedefs?

We are already using the "using" syntax for new and changed typedefs (where appropriate). FWIW, I found only two or three violations in the last 300 master commits (that added more than 40 `using` statements and removed about 60 `typedef` statements).

PR #1242, opened 2023-01-15
Reduce code duplication in ACLCertificateData::parse()

eduard-bagdasaryan:
ACLs ca_cert and user_cert already prohibit key changes. And
server_cert_fingerprint ACL only supports one option name spelling, so,
while "-sha1" is not, technically, a "key", it still cannot be
"changed". Adjust SetAclKey() (added in recent commit 4a3b853) and reuse
it to implement those existing "key change" checks.

### DIFF
--- a/src/ConfigParser.h
+++ b/src/ConfigParser.h
@@ -129,13 +129,6 @@ public:
      */
     static bool NextKvPair(char * &key, char * &value);
 
-    // TODO: Convert into a non-static method after exposing the current parser.
-    /// Extract, validate, and store the ACL key parameter for ACL types
-    /// declared using "acl aclname type key argument..." declaration that
-    /// require unique key values for each aclname+type combination.
-    /// Key comparison is case-insensitive.
-    static void SetAclKey(SBuf &keyStorage, const char *keyParameterName);
-
     /**
      * Preview the next token. The next NextToken() and strtokFile() call
      * will return the same token.
--- a/src/acl/CertificateData.cc
+++ b/src/acl/CertificateData.cc
@@ -16,7 +16,10 @@
 #include "debug/Stream.h"
 #include "wordlist.h"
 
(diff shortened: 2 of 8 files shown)


src/ConfigParser.h
```diff
@@ -129,13 +129,6 @@ class ConfigParser
      */
     static bool NextKvPair(char * &key, char * &value);
 
-    // TODO: Convert into a non-static method after exposing the current parser.
-    /// Extract, validate, and store the ACL key parameter for ACL types
-    /// declared using "acl aclname type key argument..." declaration that
-    /// require unique key values for each aclname+type combination.
-    /// Key comparison is case-insensitive.
-    static void SetAclKey(SBuf &keyStorage, const char *keyParameterName);
```
eduard-bagdasaryan, line 137: This function was moved to Acl::SetKey() without implementation changes, apart from replacing its internal ConfigParser::strtokFile() call with an external parameter. We need this to deal with optional ACL attributes in ACLCertificateData, where an attribute must be extracted and analysed whether it is a genuine ACL key before subjecting it to Acl::SetKey() checks.

src/acl/CertificateData.cc
```diff
@@ -107,14 +105,10 @@ ACLCertificateData::parse()
...
             }
 
-            /* an acl must use consistent attributes in all config lines */
-            if (attribute) {
-                if (strcasecmp(newAttribute, attribute) != 0) {
-                    debugs(28, DBG_CRITICAL, "FATAL: An acl must use consistent attributes in all config lines (" << newAttribute << "!=" << attribute << ").");
-                    self_destruct();
-                    return;
-                }
-            } else {
```
rousskov, line 117: AFAICT, the above deleted lines represent the reduction in code duplication -- the primary focus of this PR. It is not much, but still worth doing IMO. The rest is an "upgrade" of `attribute` to SBuf (to become compatible with SetAclKey()), moving strtokFile() outside SetAclKey() (because the values.insert() code above needs to handle tokens that are ACL values rather than optional ACL attributes), and associated code synchronization changes.

I am aware that `-sha1` processing is currently broken, but this PR manages to avoid touching the buggy code, and will not make fixing that bug more difficult AFAICT. That fix will need a dedicated PR.

rousskov (approved):

PR #1245, opened 2023-01-21
V6 release prep

yadij:


yadij:
Draft until the rest of the release changes are committed.

kinkie (approved):
Looks good!

yadij:
@kinkie, if you want to and have time for any of the todo list in description please feel free to hack away at it before our meeting.

doc/release-notes/release-6.sgml
```diff
@@ -25,7 +25,14 @@ This new release is available for download from <url url="http://www.squid-cache
...
 <p>Although this release is deemed good enough for use in many setups, please note the existence of
 <url url="http://bugs.squid-cache.org/buglist.cgi?query_format=advanced&amp;product=Squid&amp;bug_status=UNCONFIRMED&amp;bug_status=NEW&amp;bug_status=ASSIGNED&amp;bug_status=REOPENED&amp;version=6" name="open bugs against Squid-6">.
-<p>Support for compiling on HPUX with the native HP <em>xcc</em> compiler has been removed. To build on that OS/compiler combination, it is possible to pass these environemnt variables to ./configure: <em>CC="cxx -Ae" RANLIB=":"</em>
+
+<p>Support for compiling on HPUX with the native HP <em>xcc</em> compiler has been removed.
+  To build on that OS/compiler combination, it is possible to pass these environment variables
+  to ./configure: <em>CC="cxx -Ae" RANLIB=":"</em>
+
+<p>This release adds a dependency on C++17 support in any compiler used to build Squid.
+  GCC 8+ and Clang 8+ are known to have working C++17 support and are usable.
```
rousskov, line 34: Do we know for sure that GCC v8 can build this current Squid code? I have not tested that. If we are not sure, I recommend relaxing the above claim:

```suggestion
  GCC 8+ and Clang 8+ support C++17.
```

doc/release-notes/release-6.sgml
```diff
@@ -25,7 +25,14 @@ This new release is available for download from <url url="http://www.squid-cache
...
+  Support for older compilers and OS not supporting these have been removed.
```
rousskov, line 35: We already said that C++17 support is a dependency. I recommend removing this awkwardly worded and an unnecessary narrow statement:

```suggestion
```

doc/release-notes/release-6.sgml
```diff
@@ -130,10 +141,16 @@ This section gives an account of those changes in three categories:
 	<p>This feature has been of limited use since AsyncCalls feature
 	   took over much of the CBDATA functionality.
 
+	<tag>--enable-gnuregex</tag>
+	<p>The GnuRegex library has been removed. C++17 standard library is used instead.
```
rousskov, line 145: If we do not use STL for some regular expressions that were using GnuRegex in old `--enable-gnuregex` builds, then we should remove the second claim:

```suggestion
	<p>The GnuRegex library has been removed.
```

I would also polish the wording further:

```suggestion
	<p>Squid no longer ships with a built-in GnuRegex implementation.
```

... or something like that. Most folks probably do not view Squid distribution as something that contains a GnuRegex _library_ and we probably still support Squid builds that use an external GnuRegex library (indirectly).

doc/release-notes/release-6.sgml
```diff
@@ -130,10 +141,16 @@ This section gives an account of those changes in three categories:
...
+
 	<tag>--enable-kill-parent-hack</tag>
 	<p>This feature has been deprecated for years. Other features such as
 	   <em>--foreground</em> command line argument should be used instead.
 
+	<tag>--enable-leakfinder</tag>
+	<p>Replaced by Valgrind leak detection.
```
rousskov, line 152: Valgrind was already supported, so we did not really _replace_ one with the other.

```suggestion
	<p>Removed. Using Valgrind for leak detection is still supported.
```

ChangeLog
```diff
@@ -1,3 +1,143 @@
...
+	- Fix reconfiguration leaking tls-cert=... memory
+	- Fix X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY handling
+	- Fix "mem_obj->inmem_lo == 0" assertion in StoreEntry::swapOut()
+	- Fix TCP keepalive
+	- Fix SslBump reconfiguration leaking public key memory
+	- Fix ACL-related reconfiguration memory leak
+	- Fix socket accounting for TCP accept()
+	- ... and many documentation changes
+	- ... and much code cleanup and polishing
+	- ... and all fixes from 5.8
```
rousskov, line 139: IMO, v6 changes should be listed relative to some (disclosed!) v5 release. They should not require reading v5 change log. I suggest removing this line and (if we want to preserve the current everything-is-a-list-item format) starting the list with something like

```
	- The changes listed below are relative to Squid v5.8:
```

rousskov (commented):
IMHO, a major release is a good opportunity to improve ChangeLog format:

* Replace leading tabs with 4 spaces, providing more space for the change summary and reducing the number of files that use tabs
* Mention short git commit SHA(s) after each change item that can be associated with a single commit (or, in some special cases, a few commits). Doing so would allow the reader to quickly find more information about the actual change.
* Switch to a more human-friendly (and standard) release date format.

If you agree but prefer to make these or similar changes in a separate PR, I would be happy to fast-track the corresponding "minimal" PR.

kinkie:
> IMHO, a major release is a good opportunity to improve ChangeLog format:
> 
> * Replace leading tabs with 4 spaces, providing more space for the change summary and reducing the number of files that use tabs

I agree with this proposal

> * Mention short git commit SHA(s) after each change item that can be associated with a single commit (or, in some special cases, a few commits). Doing so would allow the reader to quickly find more information about the actual change.

I am skeptical about this proposal.
Whom is the changelog for, and what is the benefit of this change compared to browsing the git log?
If the target of the changelog are humans, I think we need to balance completeness with readability; I see little point of the changelog being a laundry list of everything that happened, costing time to write and drowning the changes in clutter

> * Switch to a more human-friendly (and standard) release date format.

Standard is probably not human friendly: the only uncontroversial format I can think of is yyyy-mm-dd which is not very human friendly. Unfortunately the US habit of writing the month first does not help with clarity elsewhere

rousskov:
> > Mention short git commit SHA(s) after each change item that can be associated with a single commit (or, in some special cases, a few commits). Doing so would allow the reader to quickly find more information about the actual change.

> I am skeptical about this proposal. Whom is the changelog for, 

Primarily for Squid admins. Squid distributors is a secondary audience IMO.

> and what is the benefit of this change compared to browsing the git log?

It is shorter/focused, curated with the needs of the targeted audience in mind. It also does not require cloning the repository or navigating (rather confusing/messy for uninitiated) GitHub rendering of repository logs.

> If the target of the changelog are humans,

Yes, a subset of humans.

> I think we need to balance completeness with readability; I see little point of the changelog being a laundry list of everything that happened, costing time to write and drowning the changes in clutter

Agreed, but I do not know why you are saying that. FWIW, I proposed adding SHA to the existing curated entries (for the reasons stated in my proposal), rather than listing "everything that happened". My proposal will not result in an increase of the number of entries. It will give some admins ability to understand what a cryptic short entry really means and to learn more details about the corresponding change.

> > * Switch to a more human-friendly (and standard) release date format.

> Standard is probably not human friendly: the only uncontroversial format I can think of is yyyy-mm-dd which is not very human friendly. Unfortunately the US habit of writing the month first does not help with clarity elsewhere

I would go with either "January 5, 2023" (disregarding the needs of humans that do not know English month names but are reading the change log written in English) or `yyyy-mm-dd` (disregarding possible confusion about the month/day order which is easily resolved by looking at all entries in aggregate) . Either would be friendlier than the current `dd english_month_abbreviation yyyy` IMHO. I assume there are some date spelling standards that document all those formats ;-).

kinkie:
> > > Mention short git commit SHA(s) after each change item that can be associated with a single commit (or, in some special cases, a few commits). Doing so would allow the reader to quickly find more information about the actual change.
> 
> > I am skeptical about this proposal. Whom is the changelog for,
[...]

> Agreed, but I do not know why you are saying that. FWIW, I proposed adding short SHAs to the existing curated entries (for the reasons stated in my proposal), rather than listing "everything that happened". My proposal will not result in an increase of the number of entries. It will give some admins ability to quickly discover what a cryptic short entry really means and to learn more details about the corresponding change.

Let me show an example of what I mean: [Jenkins' Changelog](https://www.jenkins.io/changelog/) is IMO a good balance between conciseness and details. It allows me as an admin a bird's eye view to enable me to decide whether I should or shouldn't upgrade.

> > > * Switch to a more human-friendly (and standard) release date format.
> 
> > Standard is probably not human friendly: the only uncontroversial format I can think of is yyyy-mm-dd which is not very human friendly. Unfortunately the US habit of writing the month first does not help with clarity elsewhere
> 
> I would go with either "January 5, 2023" (disregarding the needs of humans that do not know English month names but are reading the change log written in English) or `yyyy-mm-dd` (disregarding possible confusion about the month/day order which is easily resolved by looking at all entries in aggregate) . Either would be friendlier than the current `dd english_month_abbreviation yyyy` IMHO. I assume there are some date spelling standards that document all those formats ;-).

There are: YYYY-MM-DD is one of the main ISO date format standards.

rousskov:
> > > > Alex: Mention short git commit SHA(s) after each change item that can be associated with a single commit (or, in some special cases, a few commits). Doing so would allow the reader to quickly find more information about the actual change.

> > > Francesco: I am skeptical about this proposal. ...
 
> Francesco: Let me show an example of what I mean: [Jenkins' Changelog](https://www.jenkins.io/changelog/) is IMO a good balance between conciseness and details. It allows me as an admin a bird's eye view to enable me to decide whether I should or shouldn't upgrade.

Your example matches my proposal very well -- each short entry has a reference to a pull request or a similar change document with details. IMO, referring to a commit message is better that referring to GitHub pull requests, on several levels (I can detail why if needed), but that is a secondary level detail.

So, in summary, I still agree with your reasoning and examples and still do not understand why they were introduced with a negative "I am skeptical about this proposal" foreword. Given your reasoning and examples, I would expect an "I agree with this proposal" foreword.

----

BTW, switching to markdown format would be another possible improvement for our change log. Markdown would improve the presentation of more complex entries when rendered by GitHub and such, without incapacitating console readers.

ChangeLog
yadij: No,
 1) the changelog is long enough already without having to list every change twice.

 2) persons reading a changelog containing a weave of parallel release series do need guidance of what changes are relevant to their version. This line is the mechanism by which that is provided.

 3) ChangeLog is read from backwards from the version of interest. This line documents that any intervening version(s) change lists are to be skipped; until the disclosed version is read. See the v5.0.x release sections of ChangeLog for how that works.

 4) the wording is kept consistent for shared interpretation between human and machine tooling.

ChangeLog
```diff
@@ -1,3 +1,143 @@
+Changes in squid-6.0.1 (06 Feb 2023):
+
+	- Bug 5256: Intercepting port fails to accept
+	- Bug 5241: Block all non-localhost requests by default
+	- Bug 5241: Block to-localhost, to-link-local requests by default
+	- Bug 5232: Fix GCC v12 build [-Wuse-after-free]
+	- Bug 5211: support.cc:355: "!filledCheck->sslErrors" assertion
+	- Bug 5208: Part 1: Restart kids killed by SIGKILL
```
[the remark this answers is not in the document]
yadij: This list is generated with ```git show-branch origin:master origin:v5 a8b7dd82148c88acb1db3227528cd6f4c88c70e4```

yadij:
> IMHO, a major release is a good opportunity to improve ChangeLog format:
> 
>     * Replace leading tabs with 4 spaces, providing more space for the change summary and reducing the number of files that use tabs

This change would have negative side effects in at least two of our major downstream Vendor automation. It is not something we can do lightly, even across release series like this.

> 
>     * Mention short git commit SHA(s) after each change item that can be associated with a single commit (or, in some special cases, a few commits). Doing so would allow the reader to quickly find more information about the actual change.

Many changes have multiple SHA associated. We cannot do that yet. Though it is something I have been looking into for the future automation.


>     * Switch to a more human-friendly (and standard) release date format.
> 

Please explain?  We currently use the ISO 8601 defined (`dd mmm yyyy+`), which is the popular human-readable format everywhere outside USA, China, and Japan.

ChangeLog
```diff
@@ -1,3 +1,143 @@
...
+	- Remove --enable-kill-parent-hack
+	- Remove --disable-loadable-modules
+	- Remove unused/disabled/broken LEAK_CHECK_MODE code
+	- Remove SCO 3.2 support
+	- Remove m88k-specific support
+	- Remove NeXTSTEP support
+	- Remove HPUX compiler support
+	- Remove CBDATA debugging
+	- Require C++17
+	- Add Universally Unique IDentifier (UUID) support
```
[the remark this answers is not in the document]
yadij: Since you point it out, this is unused code we should drop from v6 after branching.

ChangeLog
```diff
@@ -1,3 +1,143 @@
...
+	- Allow sending "squid -k ..." signals to PID 1
+	- Remove bogus "found KEY_PRIVATE" WARNINGs
+	- Add cache_log_message directive
+	- Avoid "BUG #3329: Lost orphan ..." during accept problems
+	- Add tls_key_log to report TLS communication secrets
+	- Do not duplicate free disk slots on diskers restart
+	- Report SMP store queues state (mgr:store_queues)
+	- Add paranoid_hit_validation directive
+	- Remove 8K limit for single access.log line
+	- Add %transport::>connection_id logformat code
```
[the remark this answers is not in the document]
yadij: Nack. The original was correct, this is documenting the release version of the `Add`, not usage. That extra goes in the release notes and squid.conf documentation.

ChangeLog
rousskov: > the changelog is long enough already without having to list every change twice.

Please note that nobody is suggesting to list every change twice.

> persons reading a changelog containing a weave of parallel release series do need guidance of what changes are relevant to their version. 

Yes, and the suggested change will address that need better than the current PR code does.

> This line is the mechanism by which that is provided.

That fact does not address my change request: The change request is saying, in part, "we should not address a need for X by doing Y". A "we are doing Y" response can confirm the alleged problem, but cannot address the change request. Thus, I am not sure why you are stating this.

> This line documents that any intervening v(N-1) change lists are to be skipped; until the disclosed version is read. See the v5.0.x release sections of ChangeLog for how that works.

IMO, this approach does not work well, especially for a new major release. I am not disputing the fact that this approach is used or can be used. I am saying that we can (and should) do better.

> the wording is kept consistent for shared interpretation between human and machine tooling.

IMO, introducing improvements that are inconsistent with problematic past practices can be a step forward and now is a good moment to introduce those improvements and inconsistencies (among with format changes and other adjustments discussed [elsewhere](https://github.com/squid-cache/squid/pull/1245#pullrequestreview-1269568901)).

rousskov:
> > Replace leading tabs with 4 spaces, providing more space for the change summary and reducing the number of files that use tabs

> This change would have negative side effects in at least two of our major downstream Vendor automation. It is not something we can do lightly, even across release series like this.

Let's do it "heavily" then! I mean, there has to be a way for us to improve this file format, and if a major release is not a good time, then there never will be a good time, so we should just do it now.
 
> > Mention short git commit SHA(s) after each change item that can be associated with a single commit (or, in some special cases, a few commits). Doing so would allow the reader to quickly find more information about the actual change.

> Many changes have multiple SHA associated. 

True, but why is that a showstopper? Why not default to provide a SHA of the primary change and deal with special cases specially? For example, we can add `et al.` to cases that have too many commits to list all of them. The overall outcome would still be much better than the current approach!

> We cannot do that yet

Why not?!

> > Switch to a more human-friendly (and standard) release date format.

> We currently use the ISO 8601 defined (`dd mmm yyyy+`),

FWIW, according to ISO web [site](https://www.iso.org/iso-8601-date-and-time-format.html), "ISO 8601 tackles this uncertainty by setting out an internationally agreed way to represent dates: YYYY-MM-DD". Thus, we are currently **not** following ISO 8601 primary recommendation.

> which is the popular human-readable format everywhere outside USA, China, and Japan.

I doubt stand-alone zero-prefixed date numbers like `07` and three-letter English month abbreviations like `Jul` are really popular in human-to-human texts _anywhere_, but does the current format actually prevent us from switching to something better? If not, then let's improve (or find valid arguments not to do so)!

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 5129 pt1: remove Lock use from HttpRequestMethod
+	- Bug 5128: Translation: Fix % i typo in es/ERR_FORWARDING_DENIED
+	- Bug 5090: Must(!request->pinnedConnection()) violation
+	- Bug 5060: Parallel builds are not reliable
+	- Bug 5055: FATAL FwdState::noteDestinationsEnd exception: opening
+	- Bug 5021: Add a script to fix spelling error with codespell
+	- Bug 4946: client_side_request.cc: "request != newRequest"
+	- Bug 4922: Improve ftp://... filename extraction
+	- Bug 4832: '!schemeAccess' assertion on exit
+	- Bug 4696: Fix leaky String move assignment operator
```
rousskov, line 26: Duplicate from v5.1.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 4528: ICAP transactions quit on async DNS lookups
+	- Bug 3193 pt2: NTLM decoder truncating strings
```
rousskov, line 28: Duplicate from v5.7.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 5133: OpenSSL 3.0 support
+	- Bug 5132: Close the tunnel if to-server conn closes after client
+	- Bug 5129 pt1: remove Lock use from HttpRequestMethod
+	- Bug 5128: Translation: Fix % i typo in es/ERR_FORWARDING_DENIED
+	- Bug 5090: Must(!request->pinnedConnection()) violation
+	- Bug 5060: Parallel builds are not reliable
+	- Bug 5055: FATAL FwdState::noteDestinationsEnd exception: opening
+	- Bug 5021: Add a script to fix spelling error with codespell
+	- Bug 4946: client_side_request.cc: "request != newRequest"
+	- Bug 4922: Improve ftp://... filename extraction
```
rousskov, line 24: Duplicate from v5.2.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 5162: mgr:index URL do not produce MGR_INDEX template
+	- Bug 5160: Test suite fails with -flto=auto
+	- Bug 5158: AnyP::Uri::host() mishandles [escaped] IPv6 addresses
+	- Bug 5133: OpenSSL 3.0 support
+	- Bug 5132: Close the tunnel if to-server conn closes after client
+	- Bug 5129 pt1: remove Lock use from HttpRequestMethod
+	- Bug 5128: Translation: Fix % i typo in es/ERR_FORWARDING_DENIED
+	- Bug 5090: Must(!request->pinnedConnection()) violation
+	- Bug 5060: Parallel builds are not reliable
+	- Bug 5055: FATAL FwdState::noteDestinationsEnd exception: opening
```
rousskov, line 21: Duplicate from v5.4.1.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 5169: StoreMap.cc:517 "!s.reading()" assertion
+	- Bug 5162: mgr:index URL do not produce MGR_INDEX template
+	- Bug 5160: Test suite fails with -flto=auto
+	- Bug 5158: AnyP::Uri::host() mishandles [escaped] IPv6 addresses
+	- Bug 5133: OpenSSL 3.0 support
+	- Bug 5132: Close the tunnel if to-server conn closes after client
+	- Bug 5129 pt1: remove Lock use from HttpRequestMethod
+	- Bug 5128: Translation: Fix % i typo in es/ERR_FORWARDING_DENIED
+	- Bug 5090: Must(!request->pinnedConnection()) violation
+	- Bug 5060: Parallel builds are not reliable
```
rousskov, line 20: Duplicate from v5.3.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 5177: clientca certificates sent to https_port clients
+	- Bug 5169: StoreMap.cc:517 "!s.reading()" assertion
+	- Bug 5162: mgr:index URL do not produce MGR_INDEX template
+	- Bug 5160: Test suite fails with -flto=auto
+	- Bug 5158: AnyP::Uri::host() mishandles [escaped] IPv6 addresses
+	- Bug 5133: OpenSSL 3.0 support
+	- Bug 5132: Close the tunnel if to-server conn closes after client
+	- Bug 5129 pt1: remove Lock use from HttpRequestMethod
+	- Bug 5128: Translation: Fix % i typo in es/ERR_FORWARDING_DENIED
+	- Bug 5090: Must(!request->pinnedConnection()) violation
```
rousskov, line 19: Duplicate from v5.5.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 5211: support.cc:355: "!filledCheck->sslErrors" assertion
+	- Bug 5194: Remove all unused debug sections
+	- Bug 5186: noteDestinationsEnd check failed: transportWait
+	- Bug 5177: clientca certificates sent to https_port clients
+	- Bug 5169: StoreMap.cc:517 "!s.reading()" assertion
+	- Bug 5162: mgr:index URL do not produce MGR_INDEX template
+	- Bug 5160: Test suite fails with -flto=auto
+	- Bug 5158: AnyP::Uri::host() mishandles [escaped] IPv6 addresses
+	- Bug 5133: OpenSSL 3.0 support
+	- Bug 5132: Close the tunnel if to-server conn closes after client
```
rousskov, line 16: Duplicate from v5.4.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 5232: Fix GCC v12 build [-Wuse-after-free]
+	- Bug 5211: support.cc:355: "!filledCheck->sslErrors" assertion
+	- Bug 5194: Remove all unused debug sections
+	- Bug 5186: noteDestinationsEnd check failed: transportWait
+	- Bug 5177: clientca certificates sent to https_port clients
+	- Bug 5169: StoreMap.cc:517 "!s.reading()" assertion
+	- Bug 5162: mgr:index URL do not produce MGR_INDEX template
+	- Bug 5160: Test suite fails with -flto=auto
+	- Bug 5158: AnyP::Uri::host() mishandles [escaped] IPv6 addresses
+	- Bug 5133: OpenSSL 3.0 support
```
rousskov, line 15: Duplicate from v5.7.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 5241: Block to-localhost, to-link-local requests by default
+	- Bug 5232: Fix GCC v12 build [-Wuse-after-free]
+	- Bug 5211: support.cc:355: "!filledCheck->sslErrors" assertion
+	- Bug 5194: Remove all unused debug sections
+	- Bug 5186: noteDestinationsEnd check failed: transportWait
+	- Bug 5177: clientca certificates sent to https_port clients
+	- Bug 5169: StoreMap.cc:517 "!s.reading()" assertion
+	- Bug 5162: mgr:index URL do not produce MGR_INDEX template
+	- Bug 5160: Test suite fails with -flto=auto
+	- Bug 5158: AnyP::Uri::host() mishandles [escaped] IPv6 addresses
```
rousskov, line 14: Duplicate from v5.3.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Bug 5241: Block all non-localhost requests by default
+	- Bug 5241: Block to-localhost, to-link-local requests by default
+	- Bug 5232: Fix GCC v12 build [-Wuse-after-free]
+	- Bug 5211: support.cc:355: "!filledCheck->sslErrors" assertion
+	- Bug 5194: Remove all unused debug sections
+	- Bug 5186: noteDestinationsEnd check failed: transportWait
+	- Bug 5177: clientca certificates sent to https_port clients
+	- Bug 5169: StoreMap.cc:517 "!s.reading()" assertion
+	- Bug 5162: mgr:index URL do not produce MGR_INDEX template
+	- Bug 5160: Test suite fails with -flto=auto
```
rousskov, line 13: Duplicate from v5.7.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+
+	- Bug 5256: Intercepting port fails to accept
+	- Bug 5241: Block all non-localhost requests by default
+	- Bug 5241: Block to-localhost, to-link-local requests by default
+	- Bug 5232: Fix GCC v12 build [-Wuse-after-free]
+	- Bug 5211: support.cc:355: "!filledCheck->sslErrors" assertion
+	- Bug 5194: Remove all unused debug sections
+	- Bug 5186: noteDestinationsEnd check failed: transportWait
+	- Bug 5177: clientca certificates sent to https_port clients
+	- Bug 5169: StoreMap.cc:517 "!s.reading()" assertion
```
rousskov, line 11: Duplicate from v5.3.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
+Changes in squid-6.0.1 (06 Feb 2023):
+
+	- Bug 5256: Intercepting port fails to accept
+	- Bug 5241: Block all non-localhost requests by default
+	- Bug 5241: Block to-localhost, to-link-local requests by default
+	- Bug 5232: Fix GCC v12 build [-Wuse-after-free]
+	- Bug 5211: support.cc:355: "!filledCheck->sslErrors" assertion
+	- Bug 5194: Remove all unused debug sections
+	- Bug 5186: noteDestinationsEnd check failed: transportWait
+	- Bug 5177: clientca certificates sent to https_port clients
```
rousskov, line 10: Duplicate from v5.5.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
+Changes in squid-6.0.1 (06 Feb 2023):
+
+	- Bug 5256: Intercepting port fails to accept
+	- Bug 5241: Block all non-localhost requests by default
+	- Bug 5241: Block to-localhost, to-link-local requests by default
+	- Bug 5232: Fix GCC v12 build [-Wuse-after-free]
+	- Bug 5211: support.cc:355: "!filledCheck->sslErrors" assertion
+	- Bug 5194: Remove all unused debug sections
+	- Bug 5186: noteDestinationsEnd check failed: transportWait
```
rousskov, line 9: Duplicate from v5.7.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Remove unused/disabled/broken LEAK_CHECK_MODE code
+	- Remove SCO 3.2 support
+	- Remove m88k-specific support
+	- Remove NeXTSTEP support
+	- Remove HPUX compiler support
+	- Remove CBDATA debugging
+	- Require C++17
+	- ext_kerberos_ldap_group_acl: Support -b with -D
+	- ext_lm_group_acl: Improved username handling
+	- ext_session_acl: fix TDB key lookup
```
rousskov, line 53: Duplicate from v5.7.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Fix StoreMap.cc "anchorAt(anchorId).reading()" assertions
+	- Fix double-free segmentation fault on shutdown
+	- Fix client_side_request.cc:2028 "request->method.id()" assertion
+	- Fix FATAL ServiceRep::putConnection exception: theBusyConns > 0
+	- Fix reconfiguration leaking tls-cert=... memory
+	- Fix X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY handling
+	- Fix "mem_obj->inmem_lo == 0" assertion in StoreEntry::swapOut()
+	- Fix TCP keepalive
+	- Fix SslBump reconfiguration leaking public key memory
+	- Fix ACL-related reconfiguration memory leak
```
rousskov, line 123: Duplicate from v5.1.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Match ./configure --help parameter names with their defaults
+	- Fix typo in manager ACL
+	- Fix SQUID-MIB smilint errors
+	- Fix milliseconds in certain cache.log messages
+	- Fix ignore-cc/act-as-origin in wildcard split-stack ports
+	- Fix comm.cc:644: "address.port() != 0" assertion
+	- Fix StoreMap.cc "anchorAt(anchorId).reading()" assertions
+	- Fix double-free segmentation fault on shutdown
+	- Fix client_side_request.cc:2028 "request->method.id()" assertion
+	- Fix FATAL ServiceRep::putConnection exception: theBusyConns > 0
```
rousskov, line 117: Duplicate from v5.4.1.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Allow sending "squid -k ..." signals to PID 1
+	- Remove bogus "found KEY_PRIVATE" WARNINGs
+	- Avoid "BUG #3329: Lost orphan ..." during accept problems
+	- Report SMP store queues state (mgr:store_queues)
+	- Remove 8K limit for single access.log line
+	- Rename ./configure option --with-libxml2 to --with-xml2
+	- Rename ./configure option --with-libcap to --with-cap
+	- Match ./configure --help parameter names with their defaults
+	- Fix typo in manager ACL
+	- Fix SQUID-MIB smilint errors
```
rousskov, line 110: Duplicate from v5.6.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Honor httpd_suppress_version_string in more contexts
+	- Honor ftp_port worker-queues option
+	- Log early level-0/1 debugs() messages to cache_log
+	- Support reliable zeroing of sensitive buffers
+	- Do not overwrite caching bans
+	- Do not blame cache_peer for 4xx CONNECT responses
+	- Mimic GET reforwarding decisions when our CONNECT fails
+	- Discarded connections do not contribute to forward_max_tries
+	- Honor assertions during shutdown
+	- Kid restart leads to persistent queue overflows, delays/timeouts
```
rousskov, line 93: Duplicate from v5.5.

```suggestion
```

ChangeLog
```diff
@@ -1,3 +1,131 @@
...
+	- Remove m88k-specific support
+	- Remove NeXTSTEP support
+	- Remove HPUX compiler support
+	- Remove CBDATA debugging
+	- Require C++17
+	- ext_kerberos_ldap_group_acl: Support -b with -D
+	- ext_lm_group_acl: Improved username handling
+	- ext_session_acl: fix TDB key lookup
+	- negotiate_wrapper: ensure null-termination of strings
+	- negotiate_kerberos_auth: Initialise default_keytab
```
rousskov, line 55: Duplicate from 5.6.

```suggestion
```

ChangeLog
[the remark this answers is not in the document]
rousskov: > This list is generated with `git show-branch --date-order --no-name v5 master a8b7dd82148c88acb1db3227528cd6f4c88c70e4` and curated

The above fact does not address my concern. In fact, it amplifies it. I have now tried to identify all specific duplicates (i.e. I did what I asked for in the original change request), but I could miss some (and some entries may be missing from v6 changelog -- I did not check for that).

yadij:
> > > Replace leading tabs with 4 spaces, providing more space for the change summary and reducing the number of files that use tabs
> 
> > This change would have negative side effects in at least two of our major downstream Vendor automation. It is not something we can do lightly, even across release series like this.
> 
> Let's do it "heavily" then! I mean, there has to be a way for us to improve this file format, and if a major release is not a good time, then there never will be a good time, so we should just do it now.
> 

Lack of evidence that spaces vs tabs are improvement.

I counter that a better time to change is when the files content editing becomes automated. Which is not yet.

[ FWIW, @kinkie and I are already working on things but the primary focus needs be getting v6.0.1 out since it has a hard deadline to meet regardless of quality. ]


> > > Mention short git commit SHA(s) after each change item that can be associated with a single commit (or, in some special cases, a few commits). Doing so would allow the reader to quickly find more information about the actual change.
> 
> > Many changes have multiple SHA associated.
> 
> True, but why is that a showstopper? Why not default to provide a SHA of the primary change and deal with special cases specially? For example, we can add `et al.` to cases that have too many commits to list all of them. The overall outcome would still be much better than the current approach!
> 
> > We cannot do that yet
> 
> Why not?! 

Because all of these details need to be worked out in a way which can be automated better than the manual curation we have now. We simply do not have the time to do it properly at present and a manual attempt is unlikely to produce a list cleanly matching the end-product of the future automated version.


> > > Switch to a more human-friendly (and standard) release date format.
> 
> > We currently use the ISO 8601 defined (`dd mmm yyyy+`),
> 
> FWIW, according to ISO web [site](https://www.iso.org/iso-8601-date-and-time-format.html), "ISO 8601 tackles this uncertainty by setting out an internationally agreed way to represent dates: YYYY-MM-DD". Thus, we are currently **not** following ISO 8601 primary recommendation.

ISO 8601 **primary** recommendation is the interchange format you mention. It can be said to be "best" for **machine** exchange - but machine friendly was not the criteria you advocated.

> 
> > which is the popular human-readable format everywhere outside USA, China, and Japan.
> 
> I doubt stand-alone zero-prefixed date numbers like `07` and three-letter English month abbreviations like `Jul` are really popular in human-to-human texts _anywhere_,

You are flat wrong about that. It is quite common for hand-written dates. For **human** written dates ISO 8601 lists multiple syntax (yes **not* the YMD primary interchange format being promoted). What we have today is both the "best"[1] of the DMY human syntax **and** the most popular when compared by number of countries it is used[2]. The '0' prefix on day is sometimes elided, sometimes not depending on formality of the context.

[1]  `dd mmm yyyy+` is the "best" in that it avoids month digits and all the USA vs World ordering issues those bring. And does not require full month name, making it fixed-length.
[2] by population usage the `mmmm dd, yyyy+` can be called more popular. But lacks several of the benefits from [1].


> but does the current format actually prevent us from switching to something better? If not, then let's improve (or find valid arguments not to do so)!

Valid argument against: "It works well, and has done so for decades"

IMO you have not substantiated any measurable criteria for "better", nor presented any reasonable argument why change the date syntax at all.

rousskov:
> Lack of evidence that spaces vs tabs are improvement.

Sigh.

> I counter that a better time to change is when the files content editing becomes automated. Which is not yet.

> [ FWIW, @kinkie and I are already working on things but the primary focus needs be getting v6.0.1 out since it has a hard deadline to meet regardless of quality. ]

Glad to hear you are working on improving this file.

BTW, there is no hard deadline to release v6.0.1 -- forcing a manual release to happen at some exact time would be unnecessary and rather cruel to humans doing that manual release. The only hard deadline is for v6 _branching_, which ought to be done _before_ a v6.0.1 release and takes about a minute. In fact, the "deemed" to occur [caveat](https://wiki.squid-cache.org/ReleaseSchedule#notes) removes the need to observe any hard deadlines for the branching event as well: Even if nothing happens at that deadline, we would still know the exact master commit that will be the last/latest that exists in _both_ master and v6 branches, and that git commit is where v6 must be branched of (at _any_ time after that hard deadline).

yadij:
> BTW, there is no hard deadline to release v6.0.1  .....

That description does not match what was formally published in https://wiki.squid-cache.org/ReleaseSchedule. Which documents v6.0.1 happening at `2023-02-05T00:00:00+00:00`. The release packaging commit shall be the first commit after that timestamp on the `v6` branch.

rousskov:
> > Alex: BTW, there is no hard deadline to release v6.0.1  .....
 
> Amos: That description does not match what was formally published in https://wiki.squid-cache.org/ReleaseSchedule. 

I believe the "formally published" [ReleaseSchedule](https://wiki.squid-cache.org/ReleaseSchedule) text matches what I have said (and cited). Hopefully, the clarifications below will help prove that assertion.

> Which documents v6.0.1 happening at `2023-02-05T00:00:00+00:00`. 

Not exactly. The documented `2023-02-05T00:00:00` time (in the v6 maintainer time zone) is for the documented _event_. That event is _not_ a v6.0.1 release. That event is v6 _branching_. See the table "event" heading and the following text: "The tabulated events are nearly instantaneous and should occur (or be deemed occurred) at ... `T00:00:00` in maintainer time zone".

The release itself is related to that event, of course, but, even if we equate release with committing something (which we really should not!), a release probably cannot match that event exactly (branching v6 and committing to v6 probably has to be done as two actions, not one). Furthermore, we explicitly say that the branching event should occur or is _deemed_ to have occurred at that "T0" time. This "deemed" caveat gives humans plenty of flexibility and avoids completely unnecessary and awkward dependency on the exact moment in time.

> The release packaging commit shall be the first commit after that timestamp on the `v6` branch.

That decision is up to the folks doing the release, but there is no such official rule. If needed, there can be 100 git commits exclusive to the v6 branch prior to the v6 commit tagged as `SQUID_6_0_1`. Needless to say, there is an expectation that v6.0.1 release will be available shortly after the deadline, but ReleaseSchedule deadlines and rules only determine the exact master commit that will be the last/latest that exists in both master and v6 branches. Everything else (in the scope of this discussion) is still nearly as flexible as it was before we switched to the fixed release schedule.

Imagine, for example, that a release manager works really hard to make vN.0.1 commit at `T00:00:00`, but, after that release manager creates that commit, the commit fails our CI tests because the corresponding CI files were not updated correctly (despite the fact that all CI tests passed for the PR). That test failure does _not_ imply that we have failed our release obligations, or that we have to make a really defective release to keep our promise! It simply means that one or more vN branch commits are necessary to finalize that particular release. Those commits should be done with a high priority -- we do not want to stall the release -- but they are really not a big deal.

The above is just an example. A power outage or connectivity loss at `T0` is another simple example that does not result in any broken promises or other serious problems. Many other examples are possible. We wrote the rules to only fix/restrict what had to be fixed/restricted. The exact release time is not one of those things, regardless of how we define what "release time" actually means (there is currently no consensus regarding that definition AFAICT).

doc/release-notes/release-6.sgml
```diff
@@ -57,17 +209,68 @@ This section gives an account of those changes in three categories:
...
-	<p>There have been no directives changed.
+	<tag>time units</tag>
+	<p>All directives accepting time values now accept a time unit suffix
+	   from nanosecond to decade
+
+	<tag>sslcrtvalidator_program</tag>
+	<p>Add cache and ttl specifiers to limit resources allocated
+
+	<tag>logformat</tag>
+	<p>added transport::>connection_id tag to record what transport-level
```
yadij, line 233: Please use "New" instead of "added", with upper case for sentence start. Likewise "display" instead of "record".

This directive documents these as "codes", so that terminology should be used.

The angle brackets are reserved syntax. So we have to use the &gt; or &lt; entities. Also the EM tags for highlighting again.

```suggestion
	<p>New <em>transport::&gt;connection_id</em> code to display which transport-level
```

doc/release-notes/release-6.sgml
```diff
@@ -57,17 +209,68 @@ This section gives an account of those changes in three categories:
...
+	   connection the request was received.
+	<p>added busy_time tag to record the cumulative CPU time spent processing
```
yadij, line 235:
```suggestion
	<p>New <em>busy_time</em> code to display the cumulative CPU time spent processing
```

doc/release-notes/release-6.sgml
```diff
@@ -57,17 +209,68 @@ This section gives an account of those changes in three categories:
...
 <sect1>Changes to existing directives<label id="modifieddirectives">
 <p>
 <descrip>
-	<p>There have been no directives changed.
+	<tag>time units</tag>
+	<p>All directives accepting time values now accept a time unit suffix
+	   from nanosecond to decade
+
+	<tag>sslcrtvalidator_program</tag>
+	<p>Add cache and ttl specifiers to limit resources allocated
```
yadij, line 230: These are optional, thus "options".

The "cache" option has not actually changed. Just more verbose text about its use.

The "ttl=infinity" setting is new to the existing "ttl=" option. This text should just mention what it means.

Also, literal syntax should be wrapped in EM tags for highlighting.
```suggestion
	<p>New <em>ttl=infinity</em> option to disable TTL expiry on stored helper responses.
```

doc/release-notes/release-6.sgml
```diff
@@ -57,17 +209,68 @@ This section gives an account of those changes in three categories:
...
+
+	<tag>logformat</tag>
+	<p>added transport::>connection_id tag to record what transport-level
+	   connection the request was received.
+	<p>added busy_time tag to record the cumulative CPU time spent processing
+	   the request, excluding the time spent waiting for external resources.
+	   WARNING: this time is approximate and is known to have bugs and gaps,
+	   so consider it a lower bound
+	<p>added request_attempts tag to record how mnay forwarding attempts were
+	   done for this request
```
yadij, line 240:
```suggestion
	<p>New <em>request_attempts</em> code to display how many forwarding attempts were
	   made for this request.
```

yadij (commented):
Couple of syntax and wording details.

configure.ac
```diff
@@ -131,6 +131,8 @@ AC_SUBST(AR_R)
 
 AC_PATH_PROG(LINUXDOC, linuxdoc, $FALSE)
 AM_CONDITIONAL(ENABLE_RELEASE_DOCS, test "x${ac_cv_path_LINUXDOC}" != "x$FALSE")
+SQUID_RELEASE=`echo $VERSION | cut -d. -f1`
+AC_SUBST(SQUID_RELEASE)
```
rousskov, line 135: Since `SQUID_RELEASE` is generally-useful information that is provided even when `ENABLE_RELEASE_DOCS` is falsy, please separate these two new lines from `ENABLE_RELEASE_DOCS` code above them. I also recommend moving them higher, to make `SQUID_RELEASE` available to a larger section of future `configure.ac` code.

yadij: What future autoconf code do you envision needing the Squid series number earlier than this?
The lines above here are all detecting the build environment/toolchain. This is the point in configure.ac where it begins determining the values going into output files.

I could document those areas better if you like.

rousskov: > What future autoconf code do you envision needing the Squid series number earlier than this?

To save time, I will not try to defend my suggestion to make `SQUID_RELEASE` available to a larger section of future `configure.ac` code. Ignore it if you think it is wrong.

Still, please separate these two new lines from `ENABLE_RELEASE_DOCS` code above them (for the reasons already given).

> I could document those areas better if you like.

I am not requesting any area documentation improvements in this change request, and I do not recommend making them in this PR.

ChangeLog
yadij: Had a meeting with @kinkie today. We will leave this file as-is for now and do a full redesign for a later release soon.

configure.ac
kinkie: I support this going in. It doesn't hurt, and it can help

yadij:
@kinkie, just awaiting your review tick for Anubis to take over.

kinkie (approved):
as discussed, gogogo

rousskov:
FYI: I resolved a trivial merge conflict to let this PR progress further (commit [8d7e388](https://github.com/squid-cache/squid/pull/1245/commits/8d7e388c30e18d26c2294f044beec9ce926733a1)).

PR #1246, opened 2023-01-25
Fix build with clang v15.0.6 [-Wdeprecated-non-prototype]

kinkie:
clang v15.0.6 distributed with Fedora Rawhide complains about function
definitions using K&R style:

    lib/snmplib/mib.c:229:1: error: a function definition without a
    prototype is deprecated in all versions of C and is not supported in
    C2x [-Werror,-Wdeprecated-non-prototype]

PR #1247, opened 2023-01-25
 Fix build with clang v15.0.6 [-Wunused-but-set-variable]

kinkie:
clang v15.0.6 distributed with Fedora Rawhide complains about a variable
that is updated but never used:

    net_db.cc:230:9: error: variable 'removed' set but not used
    [-Werror,-Wunused-but-set-variable]`

### DIFF
--- a/src/icmp/net_db.cc
+++ b/src/icmp/net_db.cc
@@ -227,7 +227,6 @@ netdbPurgeLRU(void)
     netdbEntry **list;
     int k = 0;
     int list_count = 0;
-    int removed = 0;
     list = (netdbEntry **)xcalloc(netdbEntry::UseCount(), sizeof(netdbEntry *));
     hash_first(addr_table);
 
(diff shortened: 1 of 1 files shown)


rousskov (approved):
If this change fixes Squid build somewhere, please update PR title/description accordingly (e.g., see #1246).

I polished PR title, primarily to highlight that this is a code maintenance change (i.e. it does not affect Squid functionality). I removed PR description to fix its formatting -- the PR title now contains essentially the same info. Please review and adjust further as needed.

FWIW, I recommend avoiding `key: text` PR title format except for a few commonly used keys like `Maintenance`, `Bug NNN`, `Docs`, `Language`, and `CI`. Having a small set of commit title keys helps navigating regular `git log` output. On the other hand, using function names as keys is just a distraction/noise -- folks searching for function X changes will use `git log -L...` or similar tools tailored for that purpose.

kinkie:
@rousskov, @yadij, given how trivial this is, can we fast track it?

rousskov:
> @rousskov, @yadij, given how trivial this is, can we fast track it?

It is already on the fast track.

I re-polished PR title/description again, for the usual reasons.

PR #1248, opened 2023-01-28
Drop cache_object protocol support 

eduard-bagdasaryan:
Removing this non-standard protocol (already mentioned as deprecated
in Squid sources) would eliminate duplication and simplify the
existing error-prone forwarding logic (causing CVEs).

Also fixed a squidclient bug where Proxy-Authorization password was
mistakenly sent in the '@' URL part, ending up with Squid parsing two
different passwords: one from the '@' URL part * another from the
Authorization header. We must use only Authorization header for
credentials in cache manager requests.

### DIFF
--- a/src/cache_manager.cc
+++ b/src/cache_manager.cc
@@ -151,15 +151,14 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
 }
 
 /**
- * define whether the URL is a cache-manager URL and parse the action
- * requested by the user. Checks via CacheManager::ActionProtection() that the
- * item is accessible by the user.
+ * Parses the action requested by the user and checks via
+ * CacheManager::ActionProtection() that the item is accessible by the user.
  *
  * Syntax:
  *
- *  scheme "://" authority [ '/squid-internal-mgr' ] path-absolute '?' query-string
+ * [ scheme "://" authority ] '/squid-internal-mgr' path-absolute [ "?" query ] [ "#" fragment ]
  *
- * see RFC 3986 for definitions of scheme, authority, path-absolute, query-string
+ * see RFC 3986 for definitions of scheme, authority, path-absolute, query
  *
  * \returns Mgr::Command object with action to perform and parameters it might use
  */
@@ -169,8 +168,7 @@ CacheManager::ParseUrl(const AnyP::Uri &uri)
     Parser::Tokenizer tok(uri.path());
 
     static const SBuf internalMagicPrefix("/squid-internal-mgr/");
-    if (!tok.skip(internalMagicPrefix) && !tok.skip('/'))
-        throw TextException("invalid URL path", Here());
+    Assure(tok.skip(internalMagicPrefix));
 
     Mgr::Command::Pointer cmd = new Mgr::Command();
     cmd->params.httpUri = SBufToString(uri.absolute());
@@ -227,13 +225,10 @@ CacheManager::ParseHeaders(const HttpRequest * request, Mgr::ActionParams &param
     // TODO: use the authentication system decode to retrieve these details properly.
 
     /* base 64 _decoded_ user:passwd pair */
-    auto basic_cookie(request->header.getAuthToken(Http::HdrType::AUTHORIZATION, "Basic"));
+    const auto basic_cookie(request->header.getAuthToken(Http::HdrType::AUTHORIZATION, "Basic"));
 
-    if (basic_cookie.isEmpty()) {
-        basic_cookie = request->header.getAuthToken(Http::HdrType::PROXY_AUTHORIZATION, "Basic");
-        if (basic_cookie.isEmpty())
-            return;
-    }
+    if (basic_cookie.isEmpty())
+        return;
 
     const auto colonPos = basic_cookie.find(':');
     if (colonPos == SBuf::npos) {
--- a/src/cache_manager.cc
+++ b/src/cache_manager.cc
@@ -227,10 +227,13 @@ CacheManager::ParseHeaders(const HttpRequest * request, Mgr::ActionParams &param
     // TODO: use the authentication system decode to retrieve these details properly.
 
     /* base 64 _decoded_ user:passwd pair */
-    const auto basic_cookie(request->header.getAuthToken(Http::HdrType::PROXY_AUTHORIZATION, "Basic"));
+    auto basic_cookie(request->header.getAuthToken(Http::HdrType::AUTHORIZATION, "Basic"));
 
-    if (basic_cookie.isEmpty())
-        return;
+    if (basic_cookie.isEmpty()) {
+        basic_cookie = request->header.getAuthToken(Http::HdrType::PROXY_AUTHORIZATION, "Basic");
+        if (basic_cookie.isEmpty())
+            return;
+    }
 
     const auto colonPos = basic_cookie.find(':');
     if (colonPos == SBuf::npos) {
--- a/src/cache_manager.cc
+++ b/src/cache_manager.cc
@@ -150,13 +150,6 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
     return cmd->profile->creator->create(cmd);
 }
 
-static const CharacterSet &
-MgrFieldChars()
-{
-    static const CharacterSet actionChars = CharacterSet("mgr-field", "?#").complement();
-    return actionChars;
-}
-
 /**
  * define whether the URL is a cache-manager URL and parse the action
  * requested by the user. Checks via CacheManager::ActionProtection() that the
@@ -182,7 +175,7 @@ CacheManager::ParseUrl(const AnyP::Uri &uri)
     Mgr::Command::Pointer cmd = new Mgr::Command();
     cmd->params.httpUri = SBufToString(uri.absolute());
 
-    const auto &fieldChars = MgrFieldChars();
+    static const auto fieldChars = CharacterSet("mgr-field", "?#").complement();
 
     SBuf action;
     if (!tok.prefix(action, fieldChars)) {
@@ -234,7 +227,7 @@ CacheManager::ParseHeaders(const HttpRequest * request, Mgr::ActionParams &param
     // TODO: use the authentication system decode to retrieve these details properly.
 
     /* base 64 _decoded_ user:passwd pair */
-    const auto basic_cookie(request->header.getAuthToken(Http::HdrType::AUTHORIZATION, "Basic"));
+    const auto basic_cookie(request->header.getAuthToken(Http::HdrType::PROXY_AUTHORIZATION, "Basic"));
 
     if (basic_cookie.isEmpty())
         return;
--- a/src/cache_manager.cc
+++ b/src/cache_manager.cc
@@ -151,14 +151,8 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
 }
 
 static const CharacterSet &
-MgrFieldChars(const AnyP::ProtocolType &protocol)
+MgrFieldChars()
 {
-    // Deprecated cache_object:// scheme used '@' to delimit passwords
-    if (protocol == AnyP::PROTO_CACHE_OBJECT) {
-        static const CharacterSet fieldChars = CharacterSet("cache-object-field", "@?#").complement();
-        return fieldChars;
-    }
-
     static const CharacterSet actionChars = CharacterSet("mgr-field", "?#").complement();
     return actionChars;
 }
@@ -170,7 +164,7 @@ MgrFieldChars(const AnyP::ProtocolType &protocol)
  *
  * Syntax:
  *
- *  scheme "://" authority [ '/squid-internal-mgr' ] path-absolute [ '@' unreserved ] '?' query-string
+ *  scheme "://" authority [ '/squid-internal-mgr' ] path-absolute '?' query-string
  *
  * see RFC 3986 for definitions of scheme, authority, path-absolute, query-string
  *
@@ -188,17 +182,12 @@ CacheManager::ParseUrl(const AnyP::Uri &uri)
     Mgr::Command::Pointer cmd = new Mgr::Command();
     cmd->params.httpUri = SBufToString(uri.absolute());
 
-    const auto &fieldChars = MgrFieldChars(uri.getScheme());
+    const auto &fieldChars = MgrFieldChars();
 
     SBuf action;
     if (!tok.prefix(action, fieldChars)) {
-        if (uri.getScheme() == AnyP::PROTO_CACHE_OBJECT) {
-            static const SBuf menuReport("menu");
-            action = menuReport;
-        } else {
-            static const SBuf indexReport("index");
-            action = indexReport;
-        }
+        static const SBuf indexReport("index");
+        action = indexReport;
     }
     cmd->params.actionName = SBufToString(action);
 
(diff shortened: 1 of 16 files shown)


src/cache_manager.cc
```diff
@@ -188,17 +175,12 @@ CacheManager::ParseUrl(const AnyP::Uri &uri)
     Mgr::Command::Pointer cmd = new Mgr::Command();
```
rousskov, line 175: I cannot attach this change request to the right line in the code, but the `if` statement above still allows non-magical URLs. That was necessary because `cache_object://...` URLs did not have a magic prefix. This PR should not allow non-magical URLs anymore AFAICT.

You may be tempted to fix this by writing a replacement like this:

```C++
if (!tok.skip(internalMagicPrefix))
        throw TextException("invalid URL path", Here());
```

... but I believe the correct replacement is actually this:

```C++
Assure(tok.skip(internalMagicPrefix));
```

Please double check my logic: Bugs notwithstanding, this method must not be called for non-cache-manager URLs. AFAICT, it is not called for non-cache-manager URLs in the official code. We can and should assert that invariant using Assure().

src/cache_manager.cc
```diff
@@ -252,10 +227,13 @@ CacheManager::ParseHeaders(const HttpRequest * request, Mgr::ActionParams &param
     // TODO: use the authentication system decode to retrieve these details properly.
 
     /* base 64 _decoded_ user:passwd pair */
-    const auto basic_cookie(request->header.getAuthToken(Http::HdrType::AUTHORIZATION, "Basic"));
+    auto basic_cookie(request->header.getAuthToken(Http::HdrType::AUTHORIZATION, "Basic"));
 
-    if (basic_cookie.isEmpty())
-        return;
+    if (basic_cookie.isEmpty()) {
+        basic_cookie = request->header.getAuthToken(Http::HdrType::PROXY_AUTHORIZATION, "Basic");
```
rousskov, line 233: Let's clarify why we are adding the second getAuthToken() call:

```suggestion
        // squidclient does and other clients should use Proxy-Authorization to supply mgr credentials
        basic_cookie = request->header.getAuthToken(Http::HdrType::PROXY_AUTHORIZATION, "Basic");
```

We should also add a comment to explain the old/first call IMO:

```C++
// recognize Authorization header for backward compatibility
auto basic_cookie(request->header.getAuthToken(Http::HdrType::AUTHORIZATION, "Basic"));
```

if the above explanation is wrong, please use the right one, of course!

Reordering these calls/checks (i.e. checking what should be the default/primary option) makes sense, but has a potential to break that backward compatibility, so I would not do it in this PR.

tools/cachemgr.cc
```diff
@@ -861,6 +861,11 @@ process_request(cachemgr_request * req)
 
     Ip::Address::FreeAddr(AI);
 
+    // XXX: This Squid does not support receiving cache_object requests, but
+    // very old Squid versions do not support cache manager requests with an
+    // http scheme. We are using cache_object here per "be very conservative"
+    // recommendation archived at
+    // http://lists.squid-cache.org/pipermail/squid-dev/2023-January/009848.html
     l = snprintf(buf, sizeof(buf),
                  "GET cache_object://%s/%s%s%s HTTP/1.0\r\n"
```
rousskov, line 870: @yadij, please check whether this code and its justification matches your _actual_ [recommendation](http://lists.squid-cache.org/pipermail/squid-dev/2023-January/009848.html). If they do not match, please suggest what should be done with this code instead. Please be as specific as possible. Thank you.

FWIW, I recommend upgrading this code to send `http`-based URLs, but the current plan is to follow your recommendation instead.

rousskov (changes requested):
Thank you for such a quick turnaround! We still have a chance to polish and merge this before v6 is branched (if there is consensus to do so).

yadij (changes requested):
Please separate this into a PR for each binary being changed (squid, squidclient, cachemgr.cgi). They should be applied to different Squid release series for improved transition experience.

src/cache_manager.cc
eduard-bagdasaryan: > squidclient does and other clients should use Proxy-Authorization to supply mgr credentials

This was true with the old cache_object scheme when requests were always directed to this Squid instance (i.e., were 'internal' requests), i.e., both Authorization and Proxy-Authorization could probably be used for the same purpose. Now, as we request 'some' cache manager with an http request, we cannot rely on Proxy-Authorization (dropped by the first proxy in chain). @yadij 's objection looks reasonable - we should also adjust squidclient to send Authorization header instead of Proxy-Authorization. @rousskov , do you have any objections?

src/cache_manager.cc
```diff
@@ -150,27 +150,14 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
...
-
 /**
  * define whether the URL is a cache-manager URL and parse the action
  * requested by the user. Checks via CacheManager::ActionProtection() that the
  * item is accessible by the user.
  *
  * Syntax:
  *
- *  scheme "://" authority [ '/squid-internal-mgr' ] path-absolute [ '@' unreserved ] '?' query-string
+ *  scheme "://" authority [ '/squid-internal-mgr' ] path-absolute '?' query-string
```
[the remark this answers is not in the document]
rousskov: > It is better to document with standard RFC3986 definitions instead of coming up with custom syntax definitions.

Unfortunately, using the standard `path-abempty` does not document the actual syntax. The actual syntax does not use two slashes after the `/squid-internal-mgr` prefix.

We have a choice of either using a rather misleading `/squid-internal-mgr` (without a trailing slash that is actually _required_) OR not using `path-absolute` or `path-abempty`. We can use standard `segment`-based syntax instead, but it will only distract from what we are (or should be) trying to document IMO -- hence my suggestions.

I am not going to veto misleading `/squid-internal-mgr` in the official code or the incorrect double slash in one of the suggestions. I do not insist on any changes to this comment. Whatever you can agree on, use it.

tools/cachemgr.cc
rousskov: > Second best plan is to simply convert cachemgr.cgi to http://-only and mark the complex change with XXX for volunteer improvement later (if wanted). This is necessary to remove a lot of the pain caused by lack of (a), and relies on (b) for further reductions

@eduard-bagdasaryan, I recommend using the above "second best" plan for now. If somebody decides to add `http` scheme support detection, they would be in a much better position to test that code against the old Squids they care about.

eduard-bagdasaryan: > Second best plan is to simply convert cachemgr.cgi to http://-only and mark the complex change with XXX

Ok, let's make this change - it looks like there is a consensus about that.

tools/squidclient/squidclient.cc
```diff
@@ -425,17 +425,15 @@ main(int argc, char *argv[])
...
     if (strncmp(url, "mgr:", 4) == 0) {
         char *t = xstrdup(url + 4);
-        const char *at = nullptr;
-        if (!strrchr(t, '@')) { // ignore any -w password if @ is explicit already.
-            at = ProxyAuthorization.password;
+        // XXX: Bail on snprintf() failures
+        snprintf(url, sizeof(url), "http://%s:%hu/squid-internal-mgr/%s", Transport::Config.hostname, Transport::Config.port, t);
+        if (const auto at = strrchr(url, '@')) {
+            *at = 0; // send password in Proxy-Authorization header, not URL
+            pathPassword = at + 1; // embedded @password overwrites -w password further below
```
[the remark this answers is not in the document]
rousskov: @eduard-bagdasaryan, when you fix that bug in the official code (per Amos request), please mention the fix/change in in the PR description.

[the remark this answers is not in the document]
rousskov: Or, better, fix it in a dedicated PR! That dedicated PR will become a prerequisite for this PR (and all the other PRs you end up forking from this PR to address another change request).

src/cache_manager.cc
rousskov: @eduard-bagdasaryan, I do not have any objections. I am glad your work exposed a bug in the official code! If Amos thinks that we should switch `mgr:` to origin `Authorization`, let's switch!

Disclaimer: I have not studied which authentication approach is the right one here. I am just helping polish/document your code changes and following Amos change requests.

src/cache_manager.cc
eduard-bagdasaryan: Agreed. Also we should remove the 'define whether the URL is a cache-manager URL' documentation part then.

tools/squidclient/squidclient.cc
```diff
@@ -424,20 +424,14 @@ main(int argc, char *argv[])
...
-        const char *at = nullptr;
-        if (!strrchr(t, '@')) { // ignore any -w password if @ is explicit already.
-            at = ProxyAuthorization.password;
-        }
-        // embed the -w proxy password into old-style cachemgr URLs
-        if (at)
-            snprintf(url, sizeof(url), "cache_object://%s/%s@%s", Transport::Config.hostname, t, at);
-        else
-            snprintf(url, sizeof(url), "cache_object://%s/%s", Transport::Config.hostname, t);
+        snprintf(url, sizeof(url), "/squid-internal-mgr/%s", t);
```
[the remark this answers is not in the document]
rousskov: @yadij, AFAICT, your change request was made against the first commit in the PR branch (6b6490119fcca368c5b1277a0a2783cc9f908aeb). The PR posted for the review (commit 54c8fd93d940b1bd27c110cbef417238244db616) does not have this code.

tools/squidclient/squidclient.cc
```diff
@@ -457,21 +451,8 @@ main(int argc, char *argv[])
...
-            if (*t == '@') {
-                newhost = t + 1;
-                t = newhost + strcspn(newhost, "@/?");
-            }
-            *t = '\0';
-            host = newhost;
-        }
-    }
+    if (!host)
+        host = xstrdup(Transport::Config.hostname);
```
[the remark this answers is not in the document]
rousskov: Same problem here: This change request is about initial/in-progress code that was not posted for review.

src/cache_manager.cc
eduard-bagdasaryan: >  revert all these basic_cookie change

Done.

tools/cachemgr.cc
```diff
@@ -861,6 +861,11 @@ process_request(cachemgr_request * req)
...
                  "User-Agent: cachemgr.cgi/%s\r\n"
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

tools/squidclient/squidclient.cc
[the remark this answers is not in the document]
eduard-bagdasaryan: >  better, fix it in a dedicated PR!

I decided to fix this bug in this PR because these code blocks (including the new pathPassword variable) are changed  here anyway.

tools/cachemgr.cc
yadij: The split of src/ changes and tools/ changes into separate PRs still remains to complete this item.

>  b) leaving the squid binary removal a Squid version cycle. Giving admin some time to upgrade their tooling.

tools/squidclient/squidclient.cc
```diff
@@ -425,17 +425,15 @@ main(int argc, char *argv[])
...
     if (strncmp(url, "mgr:", 4) == 0) {
         char *t = xstrdup(url + 4);
-        const char *at = nullptr;
-        if (!strrchr(t, '@')) { // ignore any -w password if @ is explicit already.
-            at = ProxyAuthorization.password;
+        // XXX: Bail on snprintf() failures
+        snprintf(url, sizeof(url), "http://%s:%hu/squid-internal-mgr/%s", Transport::Config.hostname, Transport::Config.port, t);
+        if (const auto at = strrchr(url, '@')) {
+            *at = 0; // send password in Authorization header, not URL
+            pathPassword = at + 1; // embedded @password overwrites -w password further below
```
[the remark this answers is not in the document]
eduard-bagdasaryan: Done.

tools/squidclient/squidclient.cc
```diff
@@ -512,8 +510,13 @@ main(int argc, char *argv[])
         }
         if (ProxyAuthorization.user)
             ProxyAuthorization.commit(msg);
-        if (OriginAuthorization.user)
+        if (OriginAuthorization.user) {
+            const auto savedPassword = OriginAuthorization.password;
+            if (pathPassword)
+                OriginAuthorization.password = pathPassword;
```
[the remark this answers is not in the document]
eduard-bagdasaryan: I adjusted the code to abort if either pathPassword or -w/-W have been specified without specifying a username. 

>  We can add pathPassword as an additional (ORed) condition 

We could do this but the old code actually never worked with a missing username: i.e., *Authorization headers were not created if just a password had been provided. We could set username to an empty string in this case (and generate the value from the ':password' string - but this would change the old behavior.

tools/squidclient/squidclient.cc
```diff
@@ -510,10 +508,22 @@ main(int argc, char *argv[])
...
             msg << "Max-Forwards: " << max_forwards << "\r\n";
         }
-        if (ProxyAuthorization.user)
-            ProxyAuthorization.commit(msg);
-        if (OriginAuthorization.user)
-            OriginAuthorization.commit(msg);
+        if (ProxyAuthorization.user) {
+            ProxyAuthorization.commit(msg); // handles missing passwords
+        } else if (ProxyAuthorization.password) {
+            std::cerr << "ERROR: Proxy authentication username is missing\n";
```
rousskov, line 514: Please move this enforcement up/earlier, ASAP after all the named command-line options are parsed by the `while` loop. We should not make these basic configuration  checks when processing each individual URL. A misconfigured squidclient should quit on startup instead. Fixing this will also simplify this convoluted code logic a bit.

Same for OriginAuthorization, of course.

tools/squidclient/squidclient.cc
```diff
@@ -510,10 +508,22 @@ main(int argc, char *argv[])
...
+            exit(EXIT_FAILURE);
+        }
+        if (OriginAuthorization.user) {
+            const auto savedPassword = OriginAuthorization.password;
+            if (pathPassword)
+                OriginAuthorization.password = pathPassword;
+            OriginAuthorization.commit(msg); // handles missing passwords
+            OriginAuthorization.password = savedPassword; // restore the global password setting
+        } else if (pathPassword || OriginAuthorization.password) {
+            std::cerr << "ERROR: WWW authentication username is missing\n";
```
rousskov, line 524: Please move the `pathPassword` part of this check into the `pathPassword`-setting code. That code is in a much better position to report (a URL-specific) failure. Something like `ERROR: Embedding a password in a cache manager command requires providing a username with -U: <actual mgr:command here>`. The moved code will still quit with `EXIT_FAILURE`, of course.

rousskov (changes requested):
Thank you for addressing my earlier concern.

eduard-bagdasaryan:
> separate this into a PR for each binary being changed (squid, squidclient, cachemgr.cgi)

I posted PR1250, PR1251 and PR1252 respectively.

tools/squidclient/squidclient.cc
eduard-bagdasaryan: Done in [PR1251](https://github.com/squid-cache/squid/pull/1251/).

tools/squidclient/squidclient.cc
eduard-bagdasaryan: Done in [PR1251](https://github.com/squid-cache/squid/pull/1251/).

PR #1249, opened 2023-01-30
Remove broken -sha1 option from server_cert_fingerprint

eduard-bagdasaryan:
server_cert_fingerprint support for the sha1 parameter has been broken
for years, probably since its inception (2012 commit 42d3334). The bug
was known since at least 2018 when it was mentioned in Bug 4847
discussion.  The single-dash syntax violates the double-dash pattern
used for other --long ACL options. If fixed, using the option would not
change Squid behavior because SHA1 is the default (and the only
supported) fingerprinting algorithm.

The option was meant to allow admins to be explicit about that default
in case it changes in the future, but implementation bugs derailed that
plan. The fix is not straightforward, and we should be focusing on more
important things.

### DIFF
--- a/doc/release-notes/release-6.sgml
+++ b/doc/release-notes/release-6.sgml
@@ -63,7 +63,9 @@ This section gives an account of those changes in three categories:
 <sect1>Changes to existing directives<label id="modifieddirectives">
 <p>
 <descrip>
-	<p>There have been no directives changed.
+	<tag>server_cert_fingerprint</tag>
+	<p>The <em>-sha</em> argument has been removed.
+	<p>The support for this parameter has been broken for years. It has a single-dash syntax that violates the double-dash pattern used for other long <em>ACL</em> options. After removal, Squid behavior will remain the same because <em>SHA1</em> is the default (and the only supported) fingerprinting algorithm.
 </descrip>
 
 <sect1>Removed directives<label id="removeddirectives">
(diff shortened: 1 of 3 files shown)


doc/release-notes/release-6.sgml
```diff
@@ -63,7 +63,9 @@ This section gives an account of those changes in three categories:
 <sect1>Changes to existing directives<label id="modifieddirectives">
 <p>
 <descrip>
-	<p>There have been no directives changed.
+	<tag>server_cert_fingerprint</tag>
+	<p>The <em>-sha</em> argument has been removed.
+	<p>The support for this parameter has been broken for years. It has a single-dash syntax that violates the double-dash pattern used for other long <em>ACL</em> options. After removal, Squid behavior will remain the same because <em>SHA1</em> is the default (and the only supported) fingerprinting algorithm.
```
yadij, line 68: This removed value was supposedly an optional field in the syntax, thus should be called an "option" instead of a parameter (a mandatory fields in the syntax).

Also, one P tag per setting in the notes please.
```suggestion
	<p>Removed the broken <em>-sha</em> option. <em>SHA1</em> remains the default and only supported fingerprinting algorithm. Configuring it is unnecessary.
```

yadij (changes requested):
LGTM just a release notes text update please.

rousskov (approved):

yadij (approved):

PR #1251, opened 2023-01-31
squidclient: Drop cache_object protocol support

eduard-bagdasaryan:
Use `http` URL scheme for `mgr:command` cache manager requests instead.

Also fixed a bug: When `mgr:command` was used together with a `-w X`
command-line option, squidclient sent that proxy authentication password
`X` in the `cache_object` URL instead of sending the `Y` password from
the `-W Y` option (or no password at all if no `-W` option was given).
If no Authorization header had been sent, Squid's cachemgr_passwd
honored the `-w X` password sent in the cache_object URL.  If
Authorization header had been sent, Squid's cachemgr_passwd honored the
corresponding `-W Y` password, ignoring any password sent in the
cache_object URL.

Now, squidclient uses Authorization HTTP header for sending cache
manager authentication credentials with `mgr:command` requests. Those
credentials are taken either from the `-U` and `-W` command-line options
(if `mgr:command` parameter lacks an embedded password) or from the `-U`
command line option and from the `mgr:command@password` parameter
(otherwise).

squidclient now sends Proxy authentication credentials if and only if
`-u` and `-w` command line options are given. These credentials may be
required to authenticate with the proxy, but they do not affect cache
manager authentication driven by the cachemgr_passwd directive.

Also prohibit specifying a command-line option with a password (`-w` or
`-W`) without specifying the corresponding user name (`-u` or `-U`).
Prior to this change, squidclient did not send the Proxy-Authorization
or Authorization header if the corresponding user name was missing but
did not complain about the problem.

### DIFF
--- a/tools/squidclient/squidclient.cc
+++ b/tools/squidclient/squidclient.cc
@@ -425,17 +425,15 @@ main(int argc, char *argv[])
     }
 #endif
     /* Build the HTTP request */
+    const char *pathPassword = nullptr;
     if (strncmp(url, "mgr:", 4) == 0) {
         char *t = xstrdup(url + 4);
-        const char *at = nullptr;
-        if (!strrchr(t, '@')) { // ignore any -w password if @ is explicit already.
-            at = ProxyAuthorization.password;
+        // XXX: Bail on snprintf() failures
+        snprintf(url, sizeof(url), "http://%s:%hu/squid-internal-mgr/%s", Transport::Config.hostname, Transport::Config.port, t);
+        if (const auto at = strrchr(url, '@')) {
+            *at = 0; // send password in Authorization header, not URL
+            pathPassword = at + 1; // the now-removed embedded @password overwrites OriginAuthorization.password further below
         }
-        // embed the -w proxy password into old-style cachemgr URLs
-        if (at)
-            snprintf(url, sizeof(url), "cache_object://%s/%s@%s", Transport::Config.hostname, t, at);
-        else
-            snprintf(url, sizeof(url), "cache_object://%s/%s", Transport::Config.hostname, t);
         xfree(t);
     }
     if (put_file) {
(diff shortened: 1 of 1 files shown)


rousskov (commented):
I broke PR description by detailing the fixed bug and other improvements while also adding a malformed XXX (that needs to be addressed IMO). The original text said "ending up with Squid parsing two different passwords, one from the '@' URL part [and] another from the Authorization header." That may be true, but _parsing_ an extra password is just a minor unwanted overhead -- I suspect the actual serious _bug_ was in the way Squid honored or _used_ those passwords. It feels like we should explain what that bug was so that it becomes clear how our fix affects admins. I do not know enough to address that PR description XXX myself. Please address it (modifying any text around that XXX as necessary).

Please also double check my PR description wording and adjust as needed.

I do not insist on the two minor code adjustments -- they just try to explain what exactly went wrong (rather than imply that squidclient requires usernames).

yadij (approved):

eduard-bagdasaryan:
> the two minor code adjustments

Done.

> a malformed XXX (that needs to be addressed

Adjusted and checked.

rousskov:
> If Authorization header had been sent, Squid's cachemgr_passwd honored the `-W Y` password, ignoring the `-w X` password sent in the `cache_object` URL.

@eduard-bagdasaryan, what if there was a embedded-in-command password, `-W Y` password, but no `-w X` password? AFAICT, Squid would honor the `Y` password in this case as well. If that is correct, let's say the following instead (i.e. drop the `-w X` restriction):

If Authorization header had been sent, Squid's cachemgr_passwd honored the corresponding `-W Y` password, ignoring any password sent in the `cache_object` URL.

rousskov (dismissed):
Please also adjust release notes.

eduard-bagdasaryan:
> what if there was a embedded-in-command password,

You are right, I adjusted accordingly.

rousskov (approved):

PR #1252, opened 2023-01-31
cachemgr.cgi: Drop cache_object protocol support

eduard-bagdasaryan:


yadij (approved):

kinkie (approved):
LGTM

rousskov (dismissed):
Please adjust release notes.

eduard-bagdasaryan:
> Please adjust release notes.

Can we post these release notes in one place, e.g., PR1250?

rousskov:
> Can we post these release notes in one place, e.g., PR1250?

It is my understanding that we cannot do that _now_ (due to a blocking [review](https://github.com/squid-cache/squid/pull/1250#pullrequestreview-1279121364) for that PR that asks _not_ to add those notes now). IMO, we also should not do that at all (because if we are spending resources on doing separate PRs, then we should make each one as independent as possible).

However, I wrote my "add release notes" request before seeing that blocking PR 1250 review. Since I believe that `cache_object` removal should be done in v6, and not adding release notes to this PR increases our chances of doing that (and clearly does not contradict the wishes of other reviewers that already approved this PR without release notes), I am going to dismiss my change request and clear this PR for merging despite its shortcomings. We can always add release notes later, of course.

PR #1253, opened 2023-02-01
Reject more CONNECT requests with malformed targets

rousskov:
Squid silently ignored many syntax violations, interpreting a malformed
target as a valid host:port address of some origin server and
establishing a CONNECT tunnel with that origin server. While being
"tolerant" probably did not compromise the Squid instance itself, some
known attacks abuse the _difference_ in treatment of malformed requests.
Rejecting malformed requests and closing the connection prevents many
such attacks.

Among other syntax violations, this change rejects bracketed pure IPv4
addresses and bracketed domain names in CONNECT targets. Bracketed IPv6
addresses that include an IPv4address suffix are still accepted (e.g.,
look for ABNF containing ls32 in RFC 3986, Section 3.2.2).

CONNECT target hosts are no longer covered by uri_whitespace: CONNECT
targets containing whitespace are now rejected with in ERR_INVALID_URL
regardless of uri_whitespace and check_hostnames settings. With the
exception of whitespaces in host names when uri_whitespace is set to
"strip" (default), the uri_whitespace directive was already ignored for
all request targets. For example, whitespaces in the port component were
always handled without checking uri_whitespace. In CONNECT context,
stripping whitespace is not only risky but probably is not needed in
practice because user typos should not lead to spaces in CONNECT
targets, and even if they do, virtually all legitimate use cases ought
to fail during certificate validation and similar post-CONNECT checks.


rousskov:
This change was triggered by a Red Team fuzzer discovering that Squid accepts and strips garbage after CONNECT targets (e.g., `example.com:443garbage`), but code analysis and manual testing revealed that we are accepting other forms of syntactically invalid CONNECT targets, including targets with embedded spaces before port numbers and leading port zeros. I think we should be a lot more conservative in this sensitive area.

We should use Tokenizer-based parsing when handling non-CONNECT URIs as well, but that is a much bigger change, with more controversial decisions. IMO, it should be done separately and may require multiple PRs. I left a few TODOs and XXXs about that, but I do not insist on any of those comments -- if you want them removed, just ask!

src/anyp/Uri.cc
```diff
@@ -554,6 +550,88 @@ AnyP::Uri::parseUrn(Parser::Tokenizer &tok)
...
+
+    throw TextException("malformed IPv4 address or host name in uri-host", Here());
+}
+
+/// Extracts and returns an RFC 3986 URI authority port value (with additional
+/// restrictions). The RFC defines port as a possibly empty sequence of decimal
+/// digits. We reject certain ports (that are syntactically valid from the RFC
+/// point of view) because we are worried that Squid and other traffic handlers
+/// may dangerously mishandle unusual (and virtually always bogus) port numbers.
+/// Rejected ports cannot be successfully used by Squid itself.
```
yadij, line 613: No need to be worried.   RFC 9110 section 9.3.6 adds explicit criteria covering all those cases:
> A server MUST reject a CONNECT request that targets an empty or invalid port number

src/anyp/Uri.cc
```diff
@@ -554,6 +550,88 @@ AnyP::Uri::parseUrn(Parser::Tokenizer &tok)
...
+    if (tok.skip('0'))
+        throw TextException("zero or zero-prefixed port", Here());
+
+    int64_t rawPort = 0;
+    if (!tok.int64(rawPort, 10, false)) // port = *DIGIT
+        throw TextException("malformed or missing port", Here());
+
+    Assure(rawPort > 0);
+    // TODO: Move to Uri declaration, use std::optional, adjust its many users.
+    using Port = decltype(port_);
```
yadij, line 626: Please do the first of those TODO items in this PR.

Then you will be able to use the correct return type for this method (`AnyP::Uri::Port` aka `unsigned short` or the better `uint16_t`).

src/anyp/Uri.cc
```diff
@@ -554,6 +550,88 @@ AnyP::Uri::parseUrn(Parser::Tokenizer &tok)
...
+    constexpr auto portMax = 65535; // TODO: Make this a class-scope constant and REuse it.
+    static_assert(std::numeric_limits<Port>::max() >= portMax, "Port type can represent the maximum valid port number");
+    if (Less(portMax, rawPort))
+        throw TextException("huge port", Here());
+
+    return NaturalCast<int>(rawPort);
```
yadij, line 632: This should be down casting to the `Port` type, not the larger `int` type.

src/anyp/Uri.cc
```diff
@@ -554,6 +550,88 @@ AnyP::Uri::parseUrn(Parser::Tokenizer &tok)
...
+    int64_t rawPort = 0;
+    if (!tok.int64(rawPort, 10, false)) // port = *DIGIT
+        throw TextException("malformed or missing port", Here());
+
+    Assure(rawPort > 0);
+    // TODO: Move to Uri declaration, use std::optional, adjust its many users.
+    using Port = decltype(port_);
+    constexpr auto portMax = 65535; // TODO: Make this a class-scope constant and REuse it.
+    static_assert(std::numeric_limits<Port>::max() >= portMax, "Port type can represent the maximum valid port number");
+    if (Less(portMax, rawPort))
```
yadij, line 629: All this complication can be dropped with `Port` defined to the correct type.
```suggestion
    if (Less(std::numeric_limits<Port>::max(), rawPort))
```

yadij (changes requested):

src/anyp/Uri.cc
rousskov: > Please do the first of those TODO items in this PR.

I will have to post another PR that fixes the port type in the official code (and wait for that PR to merge before updating this PR). Doing that polishing change in _this_ PR would obscure important changes (not to mention that it would go against our preferences for reasonably minimal PR).

src/anyp/Uri.cc
rousskov: RFC 9110 does not prohibit empty ports in regular URIs (e.g., look for `If the port subcomponent is empty` prose). This method was written with those future callers/targets in mind, and I will have to adjust this method to handle empty ports differently after the type-changing PR discussed in another change request.

> A server MUST reject a CONNECT request that targets an empty or invalid port number

Empty ports aside, the quoted RFC 9110 text about CONNECT ports relies on RFC 3986 definition of what is "invalid". We go further than basic RFC 3986 validity restrictions. For example, `0`, `0443`, and `443000000000000` are all valid CONNECT port numbers (from the RFCs point of view) that we reject as malformed.

src/anyp/Uri.cc
rousskov: Done in #1255.

src/anyp/Uri.cc
rousskov: This change request got outdated. Please see another (fresh) GitHub thread for why this two-maximums complication should indeed be dropped, but only after we propagate portMax changes to Anyp and other users. I hope that GitHub thread addresses this change request, but, if you insist on further out-of-scope code changes, then let's continue that discussion there.

src/anyp/Uri.cc
```diff
@@ -558,6 +554,89 @@ AnyP::Uri::parseUrn(Parser::Tokenizer &tok)
...
+        throw TextException("zero or zero-prefixed port", Here());
+
+    int64_t rawPort = 0;
+    if (!tok.int64(rawPort, 10, false)) // port = *DIGIT
+        throw TextException("malformed or missing port", Here());
+
+    Assure(rawPort > 0);
+    constexpr KnownPort portMax = 65535; // TODO: Make this a class-scope constant and REuse it.
+    constexpr auto portStorageMax = std::numeric_limits<Port::value_type>::max();
+    static_assert(!Less(portStorageMax, portMax), "Port type can represent the maximum valid port number");
```
rousskov, line 631: > Amos: All this complication can be dropped with Port defined to the correct type.

There are several places (unrelated to this PR) where this 65535 port maximum is used (in various forms), so it is best to address this TODO in another dedicated PR, after migrating the non-CONNECT parsing port (that uses this maximum as well).

In that future dedicated PR, we should be able to get rid of one of the two port maximums, merging them together. However,  _this_ code should not assume that Port::value_type maximum happens to be the maximum valid port number. Future higher-level code defining Port, KnownPort, and KnownPort maximum (for the rest of the code to reuse) can make that assumption, simplifying _this_ code.

rousskov (commented):
I believe all change requests have been addressed as of commit bcb3617. Please re-review.

yadij (approved):

rousskov:
rebuild

PR #1254, opened 2023-02-02
Fix stub for ip/libip.la

yadij:
Use to build disk I/O tests which do not need IP address
operations. This also ensures bitrot does not reappear.

### DIFF
--- a/src/tests/stub_libip.cc
+++ b/src/tests/stub_libip.cc
@@ -13,65 +13,62 @@
 #include "tests/STUB.h"
 
 #include "ip/Address.h"
-namespace Ip
-{
-Address::Address(const struct in_addr &) STUB
-Address::Address(const struct sockaddr_in &) STUB
-Address::Address(const struct in6_addr &) STUB
-Address::Address(const struct sockaddr_in6 &) STUB
-Address::Address(const struct hostent &) STUB
-Address::Address(const struct addrinfo &) STUB
-Address::Address(const char*) STUB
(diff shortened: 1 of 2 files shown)


kinkie (approved):
Tests are green, this can go in.
I'm also happy with fast tracking. @rousskov ?

src/tests/stub_libip.cc
```diff
@@ -12,34 +12,122 @@
 #define STUB_API "ip/libip.la"
 #include "tests/STUB.h"
 
+#include "ip/Address.h"
+namespace Ip
```
rousskov, line 16: Here and elsewhere in this PR, please define stubbed globals and class members the same way we define regular globals and members:

```C++
void Namespace::Class::method() STUB
bool Namespace::Class::StaticMember = false;
int Namespace::Global = 0;
```

rather than 

```C++
namespace Namespace
{
void Class::method() STUB
bool Class::StaticMember = false;
int Global = 0;
}
```

Our regular style is slightly safer when it comes to globals, and style consistency helps searching and adjusting code.

rousskov (commented):
Thank you for this improvement. I have one small change request.

src/tests/stub_libip.cc
yadij: Done in[2323d3c](https://github.com/squid-cache/squid/pull/1254/commits/2323d3c17f7de8e45b28fd13a20e315fc11e9ea8)

rousskov (approved):

PR #1255, opened 2023-02-02
Improve AnyP::Uri::port_ and related port storage types

rousskov:
The old "unsigned short" type is used for various things all over the
code and cannot distinguish a parsed port zero from an absent port. In
theory, it could also hold values exceeding 65535, getting out of sync
with various (poorly duplicated) maximum port value checks.

No significant runtime changes are expected, but some transactions with
unknown request URI ports may now be reported correctly, showing "-"
instead of "0" for the missing port number. In some very special/rare
on_unsupported_protocol cases, Squid might no longer attempt to tunnel
non-HTTP requests to a zero port. Such attempts were failing anyway.

This change leaves many "unsigned short" port use cases untouched. Most
of them should be converted, but most (if not all) of those conversions
should be done in dedicated projects. Here are a few TODO highlights:

* ACLIntRange: Stores integers, is usable for integers, but assumes it
  is being configured with port numbers. We probably should enhance that
  class to handle "all" integers instead of expanding/deepening that
  rather limiting "input is just port numbers" assumption.

* CachePeer and internalRemoteUri(): Likely runtime changes related to
  squid.conf validity.

* Ftp::Server::listenForDataConnection() and friends: Triggers a few
  changes with runtime effects related to comm_local_port() failures.

* Ip::Address, Snmp::Session, fde: Too many low-level changes, some of
  which would be difficult to test.

* Config.Port.icp: Possible runtime changes related to squid.conf and
  command-line arguments validity.

Also used "%hu" instead of "%u" or "%d" to printf() ports. These changes
arguably improve code and might make some pedantic compiler happier, but
they do not fix any "real" bugs because variadic printf() arguments are
automatically promoted from short to int and, hence, printf()
implementations never get and never expect shorts.


src/anyp/Uri.cc
```diff
@@ -362,7 +362,11 @@ AnyP::Uri::parse(const HttpRequestMethod& method, const SBuf &rawUrl)
                 return false;
             *dst = '\0';
 
-            foundPort = scheme.defaultPort(); // may be reset later
+            // If the parsed scheme has no (known) default port, and there is no
+            // explicit port, then we will reject the zero port during foundPort
+            // validation, often resulting in a misleading 400/ERR_INVALID_URL.
+            // TODO: Remove this hack when switching to Tokenizer-based parsing.
+            foundPort = scheme.defaultPort().value_or(0); // may be reset later
```
rousskov, line 369: This diff may look as if this PR adds a "hack", but this PR does not change official code behavior. It only documents it. There is no bug here unless you consider a misleading `400/ERR_INVALID_URL` outcome a bug, and that misleading outcome problem existed before these changes.

src/acl/UrlPort.cc
```diff
@@ -14,6 +14,6 @@
 int
 ACLUrlPortStrategy::match(ACLData<MatchType> * &data, ACLFilledChecklist *checklist)
 {
-    return data->match(checklist->request->url.port());
+    return data->match(checklist->request->url.port().value_or(0));
```
rousskov, line 17: I have adjusted cf.data.pre to explicitly document that port zero in "port" ACL matches unknown ports. This is only a documentation improvement. The official ACL code (implicitly) works the same way.

src/anyp/Uri.cc
```diff
@@ -571,10 +575,12 @@ AnyP::Uri::authority(bool requirePort) const
         authorityWithPort_.append(host());
         authorityHttp_ = authorityWithPort_;
 
-        // authorityForm_ only has :port if it is non-default
-        authorityWithPort_.appendf(":%u",port());
-        if (port() != getScheme().defaultPort())
-            authorityHttp_ = authorityWithPort_;
+        if (port().has_value()) {
+            authorityWithPort_.appendf(":%hu", port().value());
```
rousskov, line 579: Normally, we get here with a port number. If there are some special cases when we get here without the port number, the old code made `authorityWithPort_` look like `host:0`. This PR leaves authorityWithPort_ unchanged -- `host`. I am not 100% sure whether that is an improvement, but it may be -- we are no longer leaking a bogus port number for somebody to pick up. If you disagree, I will adjust this code to append zero port again.

A _proper_ fix would be to leave `authorityWithPort_` completely empty and throw if somebody actually attempts to _use_ it later: Callers should not ask for and should not be given "authority with port" when there is no port. However, that proper  fix is outside this PR scope and it requires serious auditing work (and possibly significant code adjustments).

src/client_side.cc
```diff
@@ -3182,7 +3182,9 @@ ConnStateData::initiateTunneledRequest(HttpRequest::Pointer const &cause, const
         static char ip[MAX_IPSTRLEN];
         connectHost = clientConnection->local.toStr(ip, sizeof(ip));
         connectPort = clientConnection->local.port();
-    } else {
+    }
+
+    if (!connectPort) {
```
rousskov, line 3187: This minor change prevents possible bugs (in the official code) where we may try to tunnel to a zero port.

src/client_side_request.cc
```diff
@@ -614,14 +615,17 @@ ClientRequestContext::hostHeaderVerify()
         // Verify forward-proxy requested URL domain matches the Host: header
         debugs(85, 3, "FAIL on validate URL domain " << http->request->url.host() << " matches Host: " << host);
         hostHeaderVerifyFailed(host, http->request->url.host());
-    } else if (portStr && port != http->request->url.port()) {
+    } else if (portStr && !http->request->url.port()) {
+        debugs(85, 3, "FAIL on validate portless URI matches Host: " << portStr);
+        hostHeaderVerifyFailed("portless URI", portStr);
+    } else if (portStr && port != *http->request->url.port()) {
         // Verify forward-proxy requested URL domain matches the Host: header
```
rousskov, line 622: The new `if` statement above (line 618) is needed so that we can safely dereference url.port() in the other `if` statement on line 621 and 623. AFAICT, this change preserves the original logic.

The official code here is rather messy/problematic and should be refactored, but this PR is not the right place to do that. If you want debugs() or hostHeaderVerifyFailed() parameters to say something else, please provide exact strings (while keeping the peculiarities of the surrounding official code in mind). Thank you.

src/tests/testHttpRequest.cc
```diff
@@ -44,13 +44,12 @@ void
 testHttpRequest::testCreateFromUrl()
 {
     /* vanilla url, implicit method */
-    unsigned short expected_port;
     SBuf url("http://foo:90/bar");
     const auto mx = MasterXaction::MakePortless<XactionInitiator::initHtcp>();
     HttpRequest *aRequest = HttpRequest::FromUrl(url, mx);
-    expected_port = 90;
+    AnyP::KnownPort expected_port = 90;
```
rousskov, line 50: We cannot (easily) use AnyP::Port here because CPPUNIT_ASSERT_EQUAL()s below would not be able to "print" that. However, since the hard-coded port is indeed known in this case, this `KnownPort` usage is not wrong.

rousskov (commented):
This PR is needed because #1253 is [blocked](https://github.com/squid-cache/squid/pull/1253#discussion_r1093732337) on these AnyP::Uri port improvements.

src/anyp/Uri.cc
yadij: Side topic: do you have plans to do that conversion? or shall I resurrect my old project?

src/tests/testHttpRequest.cc
yadij: Please check that situation after adding an `AnyP::Port` stream print operator.

FWIW, I am okay with this as-is.

src/peer_select.cc
```diff
@@ -530,7 +530,7 @@ PeerSelector::noteIp(const Ip::Address &ip)
 
     Comm::ConnectionPointer p = new Comm::Connection();
     p->remote = ip;
-    p->remote.port(peer ? peer->http_port : request->url.port());
+    p->remote.port(peer ? peer->http_port : request->url.port().value_or(0));
```
yadij, line 533: This `.value_or(0)` is broken. The remote port should never be _unknown_ at this point (though it may be `0` for localhost traffic).
If it is unset/unknown something has gone badly wrong with forwarding and the peer/server cannot be selected.

Please either fix (throw an error ?) , or add XXX to document the bug.

yadij (changes requested):

src/anyp/Uri.cc
rousskov: I think the answer depends, in part, on what code you already have. Can you share a pointer to the corresponding git branch in your repository? To avoid misunderstanding, I am _not_ going to review that code -- I understand that it may be very stale and unpolished/unfinished -- I am only going to silently assess how far ahead it is and whether I agree with its overall direction, in order to answer your question.

src/acl/UrlPort.cc
rousskov: Not done. I disagree with some aspects of the suggested out-of-scope TODO and believe that planning this enhancement requires more discussion/research than we should do in this PR. Let's hash it out outside this PR if/when one of us starts working on `port` ACL enhancement.

src/format/Format.cc
```diff
@@ -508,7 +508,7 @@ Format::Format::assemble(MemBuf &mb, const AccessLogEntry::Pointer &al, int logS
 
         case LFT_LOCAL_LISTENING_PORT:
             if (const auto port = FindListeningPortNumber(nullptr, al.getRaw())) {
-                outint = port;
+                outint = *port;
```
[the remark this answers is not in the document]
rousskov: I am skipping this secondary change request until its [parent](https://github.com/squid-cache/squid/pull/1255/files#r1095565843) change request is resolved. FWIW, you can resolve this change request now because the _compiler_ will force me to make this adjustment after/if the code is adjusted the way that parent change request suggests.

src/tests/testHttpRequest.cc
rousskov: I am skipping this secondary change request until its [parent](https://github.com/squid-cache/squid/pull/1255/files#r1095576128) change request is resolved.

src/anyp/Uri.cc
rousskov: Done at 5d9fe29.

src/anyp/Uri.h
```diff
@@ -91,8 +91,8 @@ class Uri
     /// [brackets]. See RFC 3986 Section 3.2.2.
     SBuf hostOrIp() const;
 
-    void port(unsigned short p) {port_=p; touch();}
-    unsigned short port() const {return port_;}
+    void port(const Port p) { port_ = p; touch(); }
+    Port port() const { return port_; }
```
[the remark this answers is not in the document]
rousskov: Done in commit f53d10b.

src/cf.data.pre
```diff
@@ -1219,8 +1219,12 @@ endif
 	acl aclname urlpath_regex [-i] \.gif$ ...
 	  # regex matching on URL path [fast]
 
-	acl aclname port 80 70 21 0-1024...   # destination TCP port [fast]
-	                                      # ranges are allowed
+	acl aclname port 80 70 21 0-1024...
```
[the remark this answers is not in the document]
rousskov: Done in commit 3962b01.

src/client_side.h
```diff
@@ -482,7 +482,7 @@ class ConnStateData:
 
     /// The TLS server host name appears in CONNECT request or the server ip address for the intercepted requests
     SBuf tlsConnectHostOrIp; ///< The TLS server host name as passed in the CONNECT request
-    unsigned short tlsConnectPort = 0; ///< The TLS server port number as passed in the CONNECT request
+    AnyP::Port tlsConnectPort; ///< The TLS server port number as passed in the CONNECT request
```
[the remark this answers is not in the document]
rousskov: Done in commit c18ccde.

rousskov (commented):
Thank you for a prompt review. I believe I have addressed all top-level change requests, satisfying most of them. Please re-review.

src/anyp/Uri.cc
eduard-bagdasaryan: We can simplify a little:
```suggestion
            authorityWithPort_.appendf(":%hu", *port());
```
since we already checked with `has_value()`.

src/client_side.cc
```diff
@@ -3194,7 +3197,7 @@ ConnStateData::initiateTunneledRequest(HttpRequest::Pointer const &cause, const
     }
 
     debugs(33, 2, "Request tunneling for " << reason);
-    ClientHttpRequest *http = buildFakeRequest(connectHost, connectPort, payload);
+    ClientHttpRequest *http = buildFakeRequest(connectHost, *connectPort, payload);
```
eduard-bagdasaryan, line 3200:
```suggestion
    auto http = buildFakeRequest(connectHost, *connectPort, payload);
```

src/anyp/Uri.h
```diff
@@ -54,7 +54,7 @@ class Uri
         hostIsNumeric_ = false;
         *host_ = 0;
         hostAddr_.setEmpty();
-        port_ = 0;
+        port_ = std::nullopt;
```
eduard-bagdasaryan, line 57:
```suggestion
        port_.reset();
```

rousskov: Why do you think reset() is better here?

PR code feels slightly clearer to me -- it explicitly sets `port_` to a well-known "no value" state. The proposed change will result in the same "no value" state, of course, but it is slightly less clear IMO because the reader may not remember the `port_` type and might suspect that its reset() method changes the port to some known default value like 80 or 0.

eduard-bagdasaryan: There is no big difference, as you noted, but it seems to me that reset() is more in line with the calling clear() method - i.e., resetting all values to the 'default' state (equivalent to the default-constructed object). 

> a well-known "no value" state

Methods such as 'reset()' are common for STL types (e.g., smart pointers also have them). On the other hand,  'default' types (such as nullopt) are specific for each type.

rousskov: > Methods such as 'reset()' are common for STL types... On the other hand, 'default' types (such as nullopt) are specific for each type.

I agree that using the same (common) method for many data members is an advantage. IMO, being clear that there is no default port here outweighs that advantage in this std::optional context. We seem agree that the difference is small and unimportant, so if others prefer reset() here, I will accept your suggestion instead.

src/client_side.cc
rousskov: Done in commit b6e56d4.

src/anyp/Uri.cc
yadij: As far a I got was PR #335 to prepare the validation functions for using SBuf produced by a Tokenizer.

src/peer_select.cc
yadij: These Connection are intended to be (if not actually) opened. That requires a valid port number before it is sent to `handlePath()` below. The code leading up to peer selection is responsible for ensuring that the request-target is at least valid as a destination. Otherwise it is not worth even doing the IP lookup on the non-peer destinations.

(IP cache is a "red herring", this is the caller and obviously has access to the `request->url` destination details).

yadij (changes requested):
AFAICS, just the `PeerSelector::noteIp(const Ip::Address &ip)` situation to resolve now.

src/peer_select.cc
rousskov: > Otherwise it is not worth even doing the IP lookup on the non-peer destinations.

In short, I believe there are two reasonable solutions here:

1. Change PeerSelector so that it never selects DIRECT destinations for request-targets with unknown ports, under the assumption that Squid does not have any use cases and does not support any protocols that would be able to use a portless DIRECT destination. This forwarding route selection logic change is clearly outside this PR scope, even if it does not result in any functionality changes for existing Squid installations.

2. Continue to let the caller decide what DIRECT selections are valid (for caller purposes). This is what the current PR code is already doing.

> > Please either fix (throw an error ?) , or add XXX to document the bug.

Added an XXX in commit 41334f0. Please unblock this PR.

src/anyp/Uri.cc
yadij: Sure. All good.

src/client_side_request.cc
yadij: I've spent a while looking this over now and does seem to be both correct and the cleanest current change.

FTR, status quo is what it is because of the incremental discovery of the "safe" or "only partially unsafe" cases in the sparse matrix of input permutations. A refactor risks obscuring which (and how) cases are handled, do not attempt expecting an easy or quick improvement.

src/peer_select.cc
yadij: Dislike, but will stop objecting.

yadij: Oh, as for out of scope: this is the non-tunnel equivalent piece of the forwarding bug half-fixed in this PR with reasoning https://github.com/squid-cache/squid/pull/1255/files/faeb96979ae6cfb0829a2676cb9751efeed48996#r1095191336

yadij (approved):
Thank you for addressing my concerns. Just the discussion with Eduard to go before merge.

rousskov:
> Just the discussion with Eduard to go before merge.

@eduard-bagdasaryan, I am not aware of any ongoing or pending discussions on this PR, but if you are waiting for me, please let me know -- I could have missed it.

src/peer_select.cc
rousskov: > this is the non-tunnel equivalent

The critical difference between the two areas is that the high-level ConnStateData::initiateTunneledRequest() code already had logic for handling missing information -- that whole method is essentially about those missing info checks, and there was no question there that a port is required -- we know that our tunneling code needs ports. It was both easier and appropriate to rely on that existing logic to handle missing ports in that PR code.

In this low-level code, there is no existing framework to properly handle missing ports, no consensus that missing ports are actually a _problem_, and, after establishing that consensus, the missing code would need to be added in very different place(s) -- the problem cannot be properly addressed in _this_ method. That non-trivial out-of-scope change deserves a dedicated PR.

eduard-bagdasaryan:
> I could have missed it.

It looks like all my concerns have been resolved.

PR #1256, opened 2023-02-02
Run cache_peer probes without transaction context

eduard-bagdasaryan:
Even when triggered by a master transaction, these background probes
should not be associated with it. That transaction just wants to know
whether the neighbor is "up" right now. The transaction does not wait
(and is not aware of) the background probe its question may or may not
have triggered.

### DIFF
--- a/src/neighbors.cc
+++ b/src/neighbors.cc
@@ -1145,7 +1145,10 @@ int
 neighborUp(const CachePeer * p)
 {
     if (!p->tcp_up) {
-        peerProbeConnect(const_cast<CachePeer*>(p));
+        // TODO: add and pass the CachePeer context here instead of nullptr
+        CallService(nullptr, [&] {
+            peerProbeConnect(const_cast<CachePeer*>(p));
+        });
         return 0;
     }
 
--- a/src/neighbors.cc
+++ b/src/neighbors.cc
@@ -1145,7 +1145,7 @@ int
 neighborUp(const CachePeer * p)
 {
     if (!p->tcp_up) {
-        // TODO: add and pass the CachePeer context here instead of nullptr
+        // TODO: When CachePeer gets its own CodeContext, pass that context instead of nullptr
         CallService(nullptr, [&] {
             peerProbeConnect(const_cast<CachePeer*>(p));
         });


rousskov (approved):
Thank you. We will work on creating a CachePeer context, but that requires making a couple of non-trivial decisions and will trigger more/unrelated changes. We should not wait with this small "context leak" fix.

yadij:
> These probes are running in a background mode and should not relate to the triggering master transaction.

That is not always true. SNMP for example uses `neighborUp()` which should be directly attributed to the SNMP transaction.

Vice Versa there are calls directly to the `peerProbeConnect()` for background updates external to `neighborUp()`.

rousskov:
> > Original PR description: These probes are running in a background mode and should not relate to the triggering master transaction.
 
> That is not always true. SNMP for example uses `neighborUp()` which should be directly attributed to the SNMP transaction.

I disagree that a _background_ check should be attributed to the SNMP transaction just because that transaction has called neighborUp(). If that transaction were _waiting_ for that check outcome, then the PR description would be wrong -- these would be "foreground" checks, not "background" checks in that hypothetical case. However, neighborUp() callers never wait. They are only asking whether the neighbor is "up" _now_. They do not care (and should not know) what housekeeping asynchronous jobs that question may or may not trigger, and do not wait for any triggered jobs to complete. I added more text to the PR description to say that explicitly.

> Vice Versa there are calls directly to the `peerProbeConnect()` for background updates external to `neighborUp()`.

I have adjusted the PR description to avoid a possible implication that all peerProbeConnect() calls are triggered by master transactions.

Conceptually speaking, when the call is triggered by various updates (rather than a master transaction), the context switch added in this PR is either a correct/necessary switch (because we are switching into the CachePeer context from some other context) or a harmless no-op (because we are already in the CachePeer context), depending on the exact source/context of that update.

yadij:
@rousskov, I think you misunderstand.

SNMP OID `*.1.5.1.1.6` is a state probe for the peer. It initiates a new probe and can be used to accelerate recovery detection of peers that are in DEAD state but actually alive again. That probe is directly part of that SNMP transaction, not a background activity, even though it is asynchronous without a result handler.

rousskov:
> @rousskov, I think you misunderstand.
> 
> SNMP OID `*.1.5.1.1.6` is a state probe for the peer.

I am not an SNMP expert, but you probably mean cachePeerState (`.8`?). To quote `mib.txt`, cachePeerState is "The operational state of this peer". The client asks: What is the current state of this peer? Squid answers with "up" or "down".

> It initiates a new probe

One could try to add to that "... and, BTW, if the peer is down, please poke it as well (unless you are already poking at it)". While that is not documented in `mib.txt`, it may be possible to convince folks that it has to be done. However, all that is pretty much irrelevant to CodeContext and this PR!

> and can be used to accelerate recovery detection of peers that are in DEAD state but actually alive again.

Sure, but that possible acceleration effect is unrelated to SNMP transaction CodeContext tracking. SNMP cachePeerState transaction CodeContext ends/disappears when Squid is done with sending/logging the corresponding SNMP response.

> That probe is directly part of that SNMP transaction, not a background activity, even though it is asynchronous without a result handler.

From CodeContext point of view, one could say that _initiation_ of a probe is a part of that SNMP transaction, but the probe itself _is_ an independent activity with no association (other than initiation) with that transaction. The "background" label might be debatable (and is effectively related to one's point of view -- the definition of a "foreground" activity), but that debate is not relevant to CodeContext. We should not debate that here.

From CodeContext point of view, initiation is important but not _sufficient_ to extend the current transaction CodeContext into the initiated service or transaction.

In most cases, when the transaction does not depend in any way on the result (or even existence) of an asynchronous activity, CodeContext should not associate that transaction with that activity. In most cases, when one activity helps _multiple_ X transactions (especially future X transactions) and may be initiated "spontaneously" (i.e. without any X transaction involvement), that activity should not be associated with CodeContext of a single X transaction.

These rules of thumb help avoid situations where a given transaction CodeContext blows up so much that it becomes associated with lots of completely unrelated activities, sending admins and developers on triage goose chaises. For example, SNMP initiates a peer probe, the probe initiates TCP acceptor (e.g., by closing the probe TCP socket and lowering the total number of busy descriptors), TCP acceptor accepts a new connection, FTP code finally starts reading incoming DATA, that DATA connection has an I/O error, and now we are blaming the SNMP query for some completely unrelated FTP transaction failure!

I see no reason to make an exception here. From SNMP point of view, CodeContext should track SNMP query and response handling. That tracking is sufficient to say that the peer probe was initiated (if it should have been). The probe runs on its own, without CodeContext-related effects on any SNMP transactions.

Some day, we might add "initiator" to CodeContext, to track these weak relationship between otherwise independent transactions.

PR #1257, opened 2023-02-03
Update ChangeLog for latest merges

yadij:


kinkie (approved):
aye sir

rousskov (commented):
FWIW, one other recent entry (master/v6 commit ab04fcb) is currently missing from ChangeLog AFAICT: Removed broken -sha1 option from server_cert_fingerprint.

PR #1258, opened 2023-02-03
Add memFreeRigid to mem/libminimal.la

kinkie:
Complement the already-included memAllocRigid,
to better support unit tests

### DIFF
--- a/src/mem/minimal.cc
+++ b/src/mem/minimal.cc
@@ -81,6 +81,12 @@ memAllocRigid(const size_t netSize)
 }
 
 void
+memFreeRigid(void *buf, size_t)
+{
+    xfree(buf);
+}
+
+void
 memFreeBuf(size_t, void * const buf)
 {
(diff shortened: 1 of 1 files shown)


yadij (approved):
FTR; ideally these `old_api.h` symbols should be dropped. But there is still code using them. So for now this is "good enough".

src/mem/minimal.cc
```diff
@@ -80,6 +80,12 @@ memAllocRigid(const size_t netSize)
     return xmalloc(netSize);
 }
 
+void
+memFreeRigid(void *buf, size_t)
```
rousskov, line 84: To pedantically improve const-correctness and for consistency with other similar use cases in this file (and a significant portion of new code):

```suggestion
memFreeRigid(void * const buf, size_t)
```

rousskov (approved):
Thank you.

PR #1259, opened 2023-02-03
Add transparent_outgoing_ip option

DanWritesCode:
Let's say you have a squid proxy listening on IPs 1.2.3.4 and 1.2.3.5, and it's not used as a load-balancer or cache. With this option enabled, a request coming in on the .5 IP will use the system's .5 IP as outbound, whereas .4 will go through .4.

This allows the proxy user to decide what their outbound IP will be - allowing an entirely new use case for Squid. In order to do this without this option, your squid config would have to look like:

`acl ip1 myip 1.2.3.4
tcp_outgoing_address 1.2.3.4 ip1
acl ip2 myip 1.2.3.5
tcp_outgoing_address 1.2.3.5 ip2`

And so on for each outbound IP you want...

This is not practical for large subnets, and especially not for IPv6.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -1489,6 +1489,12 @@ getOutgoingAddress(HttpRequest * request, const Comm::ConnectionPointer &conn)
         // else no tproxy today ...
     }
 
+    if(Config.onoff.transparent_outgoing_ip) {
+        conn->local = request->clientConnectionManager->clientConnection->local;
+        conn->local.port(0);
+        return;
+    }
+
     if (!Config.accessList.outgoing_address) {
         return; // anything will do.
     }
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -2953,6 +2953,15 @@ DOC_START
 
 DOC_END
 
+NAME: transparent_outgoing_ip
+TYPE: onoff
+LOC: Config.onoff.transparent_outgoing_ip
+DEFAULT: off
+DOC_START
+	If this is enabled, the outgoing IP address will match the
+	local IP address that the request came in through.
+DOC_END
+
 NAME: client_dst_passthru
 TYPE: onoff
 DEFAULT: on
--- a/src/SquidConfig.h
+++ b/src/SquidConfig.h
@@ -333,6 +333,7 @@ public:
         int memory_cache_first;
         int memory_cache_disk;
         int hostStrictVerify;
+        int transparent_outgoing_ip;
         int client_dst_passthru;
         int dns_mdns;
 #if USE_OPENSSL


src/FwdState.cc
```diff
@@ -1508,6 +1508,12 @@ getOutgoingAddress(HttpRequest * request, const Comm::ConnectionPointer &conn)
         // else no tproxy today ...
     }
 
+    if(Config.onoff.transparent_outgoing_ip) {
+        conn->local = request->clientConnectionManager->clientConnection->local;
```
yadij, line 1512: This will crash on Squid generated requests when `request->clientConnectionManager` or `request->clientConnectionManager->clientConnection` are unset.

yadij (changes requested):
Thank you for your interest improving Squid.

Unfortunately this breaks in the presence of any following traffic cases:
 * IPv4 client to IPv6 server or cache_peer
 * IPv6 client to IPv4 server or cache_peer
 * NAT intercepted client
 * PROXY protocol client
 * Squid localhost or link-local listening port

yadij:
IMO we should extend `tcp_outgoing_address` to accept special-case labels instead of just raw-IP. There are several other use-cases that would benefit from that.

src/cf.data.pre
```diff
@@ -2952,6 +2952,15 @@ DOC_START
 
 DOC_END
 
+NAME: transparent_outgoing_ip
+TYPE: onoff
+LOC: Config.onoff.transparent_outgoing_ip
+DEFAULT: off
+DOC_START
+	If this is enabled, the outgoing IP address will match the
+	local IP address that the request came in through.
```
rousskov, line 2961: IMO, this feature should be refactored and reimplemented as a special `tcp_outgoing_address` keyword. For example, the following configuration sketch uses `client-connection-destination-ip` keyword (but I do _not_ insist on that specific keyword spelling and foresee a few other keywords that will be eventually added):

```
tcp_outgoing_address client-connection-destination-ip !localhost
```

I cannot do the required refactoring legwork right now, but I can help with suggestions if you are willing to work on it. Others may be able to offer more help.

Otherwise, the feature is just too limited (i.e. addressing too few/narrow use cases) and too invasive (goes into the middle of a fairly complex and often misunderstood decision making logic) to be officially accepted when there is a proper way to extend an existing directive to cover the new use case.

There are also problems with the feature name and documentation, but I am ignoring them in hope that they will be automatically resolved (or seriously change) during the requested refactoring.

rousskov (changes requested):
Thank you for working on addressing this use case!

rousskov:
Looks like Amos and I independently came to the same conclusions. Let's hope that it is a sign that our change requests point in the right direction :-).

DanWritesCode:
Alex @rousskov, Amos @yadij: Thanks for the review!
I have finally gotten around to adjusting the PR, and it should now more closely reflect the direction we want to take this in. 

There was some trickiness with `parse_acl_address()` calling the method for IP Address parsing (`parse_address()`), since I wanted the 'labels' functionality to be strictly confined to the ACL Address class. However, the changes overall are quite simple.

Let me know what you think!

src/acl/Address.h
```diff
@@ -28,6 +29,7 @@ class Address
     ACLList *aclList;
 
     Ip::Address addr;
+    String label;
```
rousskov, line 32: Please do not add such a large/heavy/deprecated data member as String to Acl::Address, especially when the code is using that member for an (essentially) boolean test.

We should also protect the code using Acl::Address from accidentally accessing "unset" `addr`. The current PR  already introduced a bug where an unset `addr` is used, and there will be more such bugs if we do not act now.

For this PR, using std::optional<Ip::Address> instead of Ip::Address may be sufficient -- the code would interpret the no-address state as a "transparent" state. However, using that approach essentially offloads the heavy lifting to the next developer who needs to add a second "special value" to Acl::Address.

```suggestion
    /// specific IP address (when set) or "use client address" prescription (otherwise)
    std::optional<Ip::Address>;
```


If you want to go for a more future-compatible solution than std::optional, then the best solution for problems like this is probably [std::variant](https://en.cppreference.com/w/cpp/utility/variant) (i.e. a type-safe union). The two existing variants (in this PR) are Ip::Address and UseClientAddress. Here is a declaration sketch:

```suggestion
    struct UseClientAddress {};
    using AddressSource = std::variant<Ip::Address, UseClientAddress>;
    AddressSource addressSource; ///< Ip::Address provider
```

Disclaimer: All suggestions above are untested sketches. All may need adjustments and all do require code changes not shown in the suggestions themselves.

src/cache_cf.cc
```diff
@@ -1551,6 +1551,8 @@ parse_address(Ip::Address *addr)
         addr->setAnyAddr();
     else if ( (!strcmp(token,"no_addr")) || (!strcmp(token,"full_mask")) )
         addr->setNoAddr();
+    else if (!strcmp(token,"transparent"))
```
rousskov, line 1554: Please do not ignore "transparent" label in unsupported contexts:

```
src/cf_parser.cci:        parse_address(&Config.Addrs.client_netmask);
src/cf_parser.cci:        parse_address(&Config.Wccp.router);
src/cf_parser.cci:        parse_address(&Config.Wccp.address);
src/cf_parser.cci:        parse_address(&Config.Wccp2.address);
src/cf_parser.cci:        parse_address(&Config.Addrs.snmp_incoming);
src/cf_parser.cci:        parse_address(&Config.Addrs.snmp_outgoing);
src/cf_parser.cci:        parse_address(&Config.Addrs.udp_incoming);
src/cf_parser.cci:        parse_address(&Config.Addrs.udp_outgoing);
src/cf_parser.cci:        parse_address(&Config.mcast_miss.addr);
```

I did not check all the details, but I would start by splitting parse_address() into two functions:

* parse_address(): Extracts the next token from the parser and calls parseAddressToken() to parse it
* parseAddressToken(): Parses the given token.

PR's parse_acl_address() will then not use parse_address(). Instead, it will extract the token, handle "transparent", and then call parseAddressToken().

Other/old parse_address() callers will remain unchanged.

The above plan also avoids adding a new PeekAtToken() call. We are trying to get rid of that bad hack...

src/cache_cf.cc
```diff
@@ -1588,6 +1590,12 @@ static void
 parse_acl_address(Acl::Address ** head)
```
rousskov, line 1590: Please do not forget to update the dump_acl_address() function. It needs to print tcp_outgoing_addr configuration.

src/cf.data.pre
```diff
@@ -2871,6 +2871,11 @@ DOC_START
 	Processing proceeds in the order specified, and stops at first fully
 	matching line.
 
+	The ipaddr may be one of the following labels (instead of a literal IP address):
+	    transparent: Sets the outgoing IP address to the same IP as the client's connection to the proxy.
```
rousskov, line 2875: "Client's connection to the proxy" has two IP addresses (source and destination), not one. Please rephrase to clearly indicate which of those two IP addresses Squid will try to use. I know that the _example_ below can be interpreted to answer that question, but we should not rely on the example for such a key determination IMO.

rousskov, line 2875: @yadij, to me, "transparent" feels too overloaded/abused of a term to be a good choice in this particular context, especially since the new feature does _not_ select the source IP address of the client TCP connection (something some admins will probably expect if we use "transparent"). I can suggest "use-client-tcp-destination-address", but that is fairly complex/long. We could (re)use "client-dst" or "use-client-dst", but that might be too cryptic? I hope you can suggest an even better name (or pick one of these three alternatives).

rousskov (changes requested):
Excellent progress, thank you! I left a few change requests to move this PR forward. Some polishing will probably required afterwards, but we will get there. Please let me know if you need help with any of the requested changes.

DanWritesCode:
Sorry about the delay in getting back to this, @rousskov. My company is doing a hack-a-thon that allows me some time to work on FOSS. I've implemented most of the suggestions, and will comment on the replies one-by-one. Looking forward to your feedback! 🙂

src/cf.data.pre
DanWritesCode: Coming up with a new name for this must have been the hardest part of the entire PR so far 😂 
I've asked both friends and GPT models for suggestions, and considered everything from "passthrough" to "match-server-ip" and just "dynamic". However, the most precise way to refer to this functionality is by the TCP destination of the client's connection to Squid. With that I settled for `match_client_tcp_dst`, but I'm still open to changing it.

DanWritesCode: I've added a slight clarification here in my latest commit, but it's worth noting the concise example that comes just below that text should clear up any misconceptions that users might have.

src/cache_cf.cc
DanWritesCode: I've done that in my latest changes, but it's worth noting that when I tried searching from my IDE, I couldn't find any instances where `dump_acl_address()` is used. I'm curious to know how that function would get invoked.

src/cache_cf.cc
DanWritesCode: I have implemented this suggestion.

src/acl/Address.h
DanWritesCode: So this is the first time that `std::variant` is being used in this project, and my first time implementing a union pattern like this. At present, I don't think this approach bought us anything compared to the `std::optional` approach (or even a simple bool/int) - and as such, I'm not really confident I implemented this in the proper way. But as you said, the next dev to work on `tcp_outgoing_address` will (probably) find having this useful.

rousskov:
> I've implemented most of the suggestions

Thank you for another successful update!

@DanWritesCode, I will need to review the current result more carefully, but I do see a few minor things that should be adjusted to comply with Squid coding style and C++ best practices. Would you mind if I make a few small commits to your PR branch? It is a much faster process than having to ask for each minor change in a GitHub change request, converge on the boring details, review code changes, readjust, etc. I will briefly explain each change in the corresponding commit message, would be happy to answer questions about those changes, and you will be able to revert any and all of them, of course!

DanWritesCode:
Yes Alex, that is no problem! Feel free to do so.

I initially added you as a collaborator to my fork, but I don't think that is needed so I removed it. Please disregard that email.

rousskov:
_Status update_: This is still on my to-do list. I hope to get back to this by the end of the week.

src/cf.data.pre
```diff
@@ -2851,7 +2851,39 @@ DOC_START
...
+
+	The required address parameter must be one of the following:
+
+		IP: A literal IP address.
+
+		domain: A domain name that Squid attempts to resolve once, at each
+			(re)configuration time, using getaddrinfo(3) or a similar blocking
+			service. DNS resolution errors are fatal.
+
+		match_client_tcp_dst: This keyword instructs Squid to postpone
```
rousskov, line 2864: I recommend `use_` instead of `match_` prefix because there is already a lot of "matching" going on in this context. As far as this client destination IP address is concerned, we are "using" it. I do not insist on this change.

src/cf.data.pre
```diff
@@ -2851,7 +2851,39 @@ DOC_START
...
+			address. Squid silently skips these rules if it cannot determine
+			the corresponding client request or its destination address at
+			tcp_outgoing_address application time.
+
+			For example, if Squid listens on both 10.0.0.1 and 10.0.0.2, but
+			accepts an HTTP request at 10.0.0.2, then matching a
+			tcp_outgoing_address rule with this keyword will set the source
+			address of the outgoing request connection to 10.0.0.2.
+
+		other: Squid also accepts a few currently undocumented keywords.
```
rousskov, line 2877: I did not want to document the other accepted keywords (i.e. "any_addr", "no_addr", and "full_mask"), but I would not object to others doing that. I am also OK with removing this line if others do not like it.

@DanWritesCode, I suspect this may be a controversial topic, and it is out of this PR scope, so I do not recommend documenting those keywords unless @yadij asks you to do that. Removing this (and any other) line if you do not like it is fine, of course.

src/acl/Address.cc
```diff
@@ -18,3 +22,42 @@ Acl::Address::~Address()
...
+    if (std::holds_alternative<Address::MatchClientTcpDst>(source)) {
+        os << "match_client_tcp_dst";
+        return os;
+    }
+
+    const auto &addr = std::get<Ip::Address>(source);
+
+    if (addr.isAnyAddr()) {
+        // XXX: Use squid.conf syntax (e.g., "any_addr") for all special values
+        os << "autoselect";
```
rousskov, line 55: We should fix this XXX, but doing so is outside this PR scope. See commit 6fb1483 message for more details.

rousskov:
OK to test

src/cf.data.pre
```diff
@@ -2868,10 +2900,6 @@ DOC_START
 	  tcp_outgoing_address 2001:db8::1
 	  tcp_outgoing_address 10.1.0.3
 
-	Processing proceeds in the order specified, and stops at first fully
-	matching line.
-
-	Squid will add an implicit IP version test to each line.
 	Requests going to IPv4 websites will use the outgoing 10.1.0.* addresses.
 	Requests going to IPv6 websites will use the outgoing 2001:db8:* addresses.
```
rousskov, line 2904: These two (old) lines belong to the (old) example above them. I moved the two non-example statements ("Processing proceeds..." and "Squid will add...") up, so that they do not split this long example. That split caused problems in earlier branch commits (that accidentally inserted even more text in the middle of that old example), and I bet similar problems may happen in the future.

rousskov (approved):
I am done with my changes. Two bug fixes (commits fa890fc and 519af74). The rest is polishing and minor improvements. I tried to document everything in commit messages and would be happy to answer any questions.

I also adjusted PR title/description to satisfy formatting requirements and to remove details already present in the updated documentation.

@DanWritesCode, please review, retest, and adjust as needed. When you are happy, please request @yadij re-review; I did not do that myself in case you need to make further changes. Thank you!

DanWritesCode:
Hey Alex! Thanks for going through the PR and taking care of all the details. I appreciate it! I'll reply to the individual comments shortly. 
Also, I was also going through the commits and have a question: 

In 03b8df41, how does moving the function up (by a bit) remove the need to polish it? In addition, should the function prototype be in cache_cf.h instead of cache_cf.cc?

src/cf.data.pre
DanWritesCode: I wouldn't mind making the change, and updating docs to reflect it. 
@yadij - do you have an opinion on the naming here by any chance?

src/cf.data.pre
DanWritesCode: One option to consider mentioning here is that "other" can include "any_addr", "no_addr", and "full_mask". This may prompt someone to look into how those options work with this setting, if there is any interest in it, without us having to spend time documenting it.

src/acl/Address.cc
DanWritesCode: Moving this to a STL stream was a good call. The code is so much nicer 👍

src/cache_cf.cc
```diff
@@ -1547,6 +1548,13 @@ parse_address(Ip::Address *addr)
         return;
     }
 
+    parseAddressToken(addr, token);
```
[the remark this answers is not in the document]
DanWritesCode: Ah, I understand now! Thanks for explaining.

DanWritesCode:
Alright, I've gone ahead and re-requested Amos' review.

src/cf.data.pre
yadij: `client_tcp_dst` seems clear enough.

PR #1261, opened 2023-02-05
support TRANSFER_ENCODING with "chunked, chunked"

ShmayaFrankel:


yadij (changes requested):
RFC 9112 section 6.1 forbids this header value
> A sender MUST NOT apply the chunked transfer coding more than once to a message body (i.e., chunking an already chunked message is not allowed).

Treating the broken input as an unsupported encoding is the safest thing Squid can do without completely forbidding access to the broken server. The alternative is closing the server connection and producing a 5xx response to the client.

yadij:
Closing after 10 days with no further feedback.

PR #1262, opened 2023-02-05
Add unit tests for ClpMap

kinkie:


yadij (changes requested):
Partial review to get started.

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,129 @@
...
+{
+    squid_curtime = time(nullptr);
+}
+
+void
+testClpMap::PutGetDelete()
+{
+    testMap m(1024);
+    addData(m, 10);
+    CPPUNIT_ASSERT_EQUAL(static_cast<const int *>(nullptr), m.get("notthere"));
```
rousskov, line 42: CPPUNIT_ASSERT_EQUAL() is best when the printed values tell us something. In this and similar "not found" cases, the printed values are meaningless pointers. I think this would be a lot more readable:

```suggestion
    CPPUNIT_ASSERT(!m.get("notthere"));
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,129 @@
...
+    CPPUNIT_ASSERT_EQUAL(1, *(m.get("1")));
```
rousskov, line 43: This will _crash_ the executable if "1" is not actually there. We want the test to fail instead. I recommend refactoring the test code a little, so that you can say something like this here and in similar cases:

```suggestion
    CPPUNIT_ASSERT_EQUAL(1, Search(m, "1").value_or(0));
```

However, this is actually a weak test. We should test that the map returned not just any pointer to `1`, but the pointer we put there. That means refactoring to end up with something like this:

```suggestion
    CPPUNIT_ASSERT_EQUAL(item1, m.get("1"));
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,129 @@
...
+testClpMap::testMemoryCounter()
+{
+    CPPUNIT_ASSERT_EQUAL(sizeof(int), static_cast<size_t>(DefaultMemoryUsage(int())));
+    CPPUNIT_ASSERT_EQUAL(sizeof(int32_t), static_cast<size_t>(DefaultMemoryUsage(int32_t())));
+    CPPUNIT_ASSERT_EQUAL(sizeof(int64_t), static_cast<size_t>(DefaultMemoryUsage(int64_t())));
+    CPPUNIT_ASSERT_EQUAL(sizeof(char), static_cast<size_t>(DefaultMemoryUsage(char())));
+    char str[10];
+    CPPUNIT_ASSERT_EQUAL(sizeof(str), static_cast<size_t>(DefaultMemoryUsage(str)));
+    CPPUNIT_ASSERT_EQUAL(sizeof(std::string), static_cast<size_t>(DefaultMemoryUsage(std::string())));
+}
```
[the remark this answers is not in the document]
rousskov: FWIW, I do not recommend adding that `TODO` because we would just end up repeating private MemoryCountedFor() implementation code here -- just like testMemoryCounter() already does for DefaultMemoryUsage(). There is no value in that duplication (and there is harm). If the suggested `TODO` stands for something else, please adjust it to clarify.

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,129 @@
...
+
+void
+testClpMap::SetMemLimit()
+{
+    testMap m(2048);
+    addData(m, 1000);
+    auto entriesBefore = m.entries();
+    m.setMemLimit(1024);
+    CPPUNIT_ASSERT(entriesBefore > m.entries());
```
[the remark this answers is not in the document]
rousskov: I agree that we should test what Amos is suggesting to test. However, please note that memoryUsed() for a one-item map is not memPerEntry (i.e. not memory used per each entry) because the map is not guaranteed to use the same amount of memory for each entry. There should be a way to test all of the above _without_ making this assumption.

rousskov (changes requested):
Thank you for adding these tests!

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,129 @@
...
+ */
+
+#include "squid.h"
+#include "testClpMap.h"
+#include "unitTestMain.h"
+
+#include "SquidConfig.h"
+
+#include <ctime>
+
```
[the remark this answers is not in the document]
kinkie: ok

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,129 @@
...
+testClpMap::SetMemLimit()
+{
+    testMap m(2048);
+    addData(m, 1000);
+    auto entriesBefore = m.entries();
+    m.setMemLimit(1024);
+    CPPUNIT_ASSERT(entriesBefore > m.entries());
+}
+
+#include <iostream>
```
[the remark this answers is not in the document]
kinkie: It's actually junk. removing

src/tests/testClpMap.cc
kinkie: ok

src/tests/testClpMap.cc
kinkie: But we are not putting in there a pointer, we are putting in there a value. ClpMap returns a pointer to the internal storage holding the value it copied in

rousskov (commented):
Thank you for making progress with this useful addition. Please note that the current PR branch contains lots of PR-unrelated changes.

src/tests/testClpMap.h
```diff
@@ -0,0 +1,44 @@
...
+#include "base/ClpMap.h"
+
+class testClpMap: public CPPUNIT_NS::TestFixture
+{
+    CPPUNIT_TEST_SUITE(testClpMap);
+    CPPUNIT_TEST(testMemoryCounter);
+    CPPUNIT_TEST(testConstructor);
+    CPPUNIT_TEST(Entries);
+    CPPUNIT_TEST(PutGetDelete);
+    CPPUNIT_TEST(SetMemLimit);
```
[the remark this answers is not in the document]
kinkie: done

src/tests/testClpMap.h
```diff
@@ -0,0 +1,44 @@
...
+    CPPUNIT_TEST(testTtlExpiration);
+    CPPUNIT_TEST_SUITE_END();
+
+protected:
+    using testMap = ClpMap<std::string, int>;
```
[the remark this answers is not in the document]
kinkie: done

src/tests/testClpMap.h
```diff
@@ -0,0 +1,44 @@
...
+
+    // add a standard set of key-values to the map, up to numElems
+    // the keys and values will start at base and count up numElems
+    void addData(testMap &m, int numElems=10, int base=0, testMap::Ttl ttl=10);
```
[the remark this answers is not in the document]
kinkie: After attempting this, I'll push back against the type change, because either we change both paramters to size_t (which makes little sense, because base is not a size), or we get into type cast annoyance, so I'll leave them as ints, which is okay given the use case.
I'm okay with the other suggestions and I'll implement them

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,128 @@
...
+
+void
+testClpMap::testSetMemLimit()
+{
+    TestMap m(2048);
+    addData(m, 1000, 0, 10);
+    auto testEntriesBefore = m.entries();
+    m.setMemLimit(1024);
+    CPPUNIT_ASSERT(testEntriesBefore > m.entries());
+}
```
rousskov, line 114: It would be nice to tell that the map can grow after purging entries. Squid had bugs in similar conditions:

```suggestion

    const auto entriesAfterPurge = m.entries(); // TODO: Move and use in the above assertion
    m.setMemLimit(m.memLimit() * 2);
    // overflow the map with entries again to make sure it can grow after purging
    addData(m, m.memLimit() / sizeof(int), 0, 10); // TODO: Extract into overflowMap()
    CPPUNIT_ASSERT(entriesAfterPurge < m.entries());
}
```

It would be nice to make these checks more strict/accurate, but that requires more code.

P.S. The above TODOs are in scope. I just could not fit those changes into one change request.

rousskov (changes requested):
I left a few more suggestions. Some changes overlap so it may be impossible to just accept them as is, but I hope it is clear what each suggestion is about. Please let me know if you want me to apply any of these changes -- I did not want to create conflicts with your ongoing/active work.

src/tests/testClpMap.h
```diff
@@ -0,0 +1,43 @@
...
+    CPPUNIT_TEST( testEntries );
+    CPPUNIT_TEST( testPutGetDelete );
+    CPPUNIT_TEST( testSetMemLimit );
+    CPPUNIT_TEST( testTtlExpiration );
+    CPPUNIT_TEST_SUITE_END();
+
+protected:
+    using TestMap = ClpMap<std::string, int>;
+
+    // add a standard set of elements to a map
```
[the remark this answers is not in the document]
kinkie: This is a convenience function, meant to add some elements to an array. It is not an API, it's local to the translation unit this file refers to.
Let's keep the complexity of the solution (and the number of iterations here) in line with the complexity of the problem; we are treating this as if it were a core API.
Same thing for Doxygen: do we even WANT to doxygen-document a two-line unit test helper function?

src/tests/testClpMap.h
```diff
@@ -0,0 +1,43 @@
...
+    void addData(TestMap &, int count, int startWith, TestMap::Ttl);
```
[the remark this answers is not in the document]
kinkie: The suggested changes are not better than the current one, as then the question is "what elements"? I'll rename to addSequenceOfElementsToMap , this should be clear then

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,128 @@
...
+#include <ctime>
+
+CPPUNIT_TEST_SUITE_REGISTRATION( testClpMap );
+
+class SquidConfig Config;
+
+void
+testClpMap::addData(TestMap &m, int count, int startWith, TestMap::Ttl ttl)
+{
+    for (auto j = startWith; j < startWith + count; ++j)
```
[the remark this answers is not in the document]
kinkie: See my other comments about keeping things as simple as possible: if one is a size_t and the other is an int, to add them a cast will be needed. And then we might need to signal the compiler that no, we are sure, there will be no overflows. I had actually implemented this, which would turn a two-lines helper function into a five-lines helper function and actually decrease intent clarity.
Could this be made more theoretically correct? Absolutely. Efficient? Of course. Is any of that worth? I contend not. Also, done is better than perfect and we are wasting way too much time on this.
Let me clarify the full context of my objective here:
- I want to use ClpMap for ipcache and fqdncache, as discussed in a different PR
- to do that I need to have const iterators on ClpMap to display cache contents on cachemanager
- to develop the  iterators I want to have higher confidence in the correctness of what I implement
- to have more confidence I want to have unit tests so I can do test-driven development
- to do that I need unit tests, which are currently not in

We are at step zero of what is an actually useful result; it's been two weeks since I started and we are picking nits. At this pace, when will we get to any meaningful result?

src/tests/testClpMap.h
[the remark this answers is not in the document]
yadij: FYI: I prefer to put these in the `private:` section at end of the class. It also helps to clarify that it is not a strangely named test method that got omitted from the `CPPUNIT_TEST()` list.

As for documenting; yes it is worthwhile - if only to assist future dev updating the unit internals.

src/tests/testClpMap.h
```diff
@@ -0,0 +1,43 @@
...
+    void testMemoryCounter();
+    void testConstructor();
+    void testEntries();
+    void testPutGetDelete();
+    void testSetMemLimit();
+    void testTtlExpiration();
+
+public:
+    void setUp() override;
```
yadij, line 39: per Squid coding style the `public:` section(s) should be up above the `protected:` section.

src/tests/testClpMap.h
```diff
@@ -0,0 +1,43 @@
...
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID_BASE_TESTCLPMAP_H
+#define SQUID_BASE_TESTCLPMAP_H
+
+#include "compat/cppunit.h"
+#include "base/ClpMap.h"
```
yadij, line 13: sort please. `b` before `c`
```suggestion
#include "base/ClpMap.h"
#include "compat/cppunit.h"
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,128 @@
...
+    {
+        CPPUNIT_ASSERT_EQUAL(true, m.add(std::to_string(j), j, ttl));
+    }
+}
+
+
```
yadij, line 29:
```suggestion

```

yadij (changes requested):
few more bits of polish

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,128 @@
...
+    CPPUNIT_ASSERT_EQUAL(uint64_t(1024), emptyD.freeMem());
+    CPPUNIT_ASSERT_EQUAL(uint64_t(0), emptyD.memoryUsed());
+    CPPUNIT_ASSERT_EQUAL(size_t(0), emptyD.entries());
+}
+
+void
+testClpMap::testSetMemLimit()
+{
+    TestMap m(2048);
+    addData(m, 1000, 0, 10);
```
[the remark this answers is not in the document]
kinkie: I agree that intent is not clear with the current call, but I argue that the problem is the ClpMap API, which lacks a way to limit the size of the map by number of entries. The current number is guaranteed to overflow because 1k ints will never fit in 2KiB. I'm happy to add the comment but I don't think it's necessarily a good idea to change the overflow-triggering code

src/tests/testClpMap.h
kinkie: Sure, thanks

src/tests/testClpMap.cc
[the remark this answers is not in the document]
kinkie: Given there is no consensus on the suggestions, I'll leave as is

src/tests/testClpMap.cc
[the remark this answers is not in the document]
kinkie: I would argue that ClpMap would be more useful if it didn't try to be too clever about the storage it uses but implemented limits to the number of entries instead, or at least in addition to the current limits.
In fact, I'd rather suggest to build a strategy around that: we test the API we have, extend the API to limit also by number of entries, and test that one instead of kludging around the heuristics about an entry's size

src/tests/testClpMap.cc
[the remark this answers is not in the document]
rousskov: I did take your comments into account when adding my support for Amos' change request. My current theory is that it is possible to write this code using correct types without casts and or other serious complications (despite your claims to the contrary). Let me know when it is a good time for me to try to test that theory. My attempt will not cost you (or me) much and may improve this code, addressing two valid (albeit minor) change requests.

> we are picking nits

I disagree that this is what the focus of the current change requests is. The important change requests are not nitpicking IMO. The vast majority of other change requests are trivially addressable and not worth discussing. I am happy to address nitpicking (or all) change requests if that helps.

> to have more confidence I want to have unit tests so I can do test-driven development

If the current tests give you that confidence, use them! You do not need this PR to be officially merged to use these tests.

> At this pace, when will we get to any meaningful result?

AFAICT, we already have meaningful result for those who want to run these tests. IIRC, there were no complex change requests, and PR delays were primarily on your side -- _you_ determine the speed of progress in this case. That situation may change, of course, but that how it looks to me right now, and, frankly, I envy your position.

FWIW, I am happy to assist with speeding this useful PR up. Just let me know when it is a good time for me to do so. Again, I would have done that already, but I did not want to create conflicts with your active work.

src/tests/testClpMap.cc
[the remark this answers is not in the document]
rousskov: > I argue that the problem is the ClpMap API, which lacks a way to limit the size of the map by number of entries. 

This PR is not the best place to argue about ClpMap API. FWIW, IIRC, this API concern has been discussed already when ClpMap was added, and I believe the current ClpMap API is a better option than a count-based limit despite the difficulties it adds to writing unit tests. We can revisit that API topic, but, again, this testing PR is probably not the best avenue for doing that, especially if you want to speed up its merging.

src/tests/testClpMap.cc
rousskov: This change request was resolved without addressing it. The change request is valid -- only the second suggestion was flawed (as detailed above). I am unresolving this change request.

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,133 @@
...
+    addSequenceOfElementsToMap(m, 1000, 0, 10);
+    const auto testEntriesBefore = m.entries();
+    CPPUNIT_ASSERT(testEntriesBefore > 0);
+    m.setMemLimit(m.memoryUsed() / 2);
+    CPPUNIT_ASSERT(testEntriesBefore > m.entries());
+
+    const auto entriesAfterPurge = m.entries(); // TODO: Move and use in the above assertion
+    m.setMemLimit(m.memLimit() * 2);
+    // overflow the map with entries again to make sure it can grow after purging
+    addSequenceOfElementsToMap(m, m.memLimit() / sizeof(int), 0, 10);  // TODO: Extract into overflowMap()
```
rousskov, line 118: Please implement these TODOs as requested in the original [change request](https://github.com/squid-cache/squid/pull/1262/files#r1111279117). Please note that the overflowMap() code should be used in _both_ "overflow map" cases -- the first case did not get the right changes (yet), but it should be the same code.

src/tests/testClpMap.h
```diff
@@ -0,0 +1,48 @@
...
+    void testMemoryCounter();
+    void testConstructor();
+    void testEntries();
+    void testPutGetDelete();
+    void testSetMemLimit();
+    void testTtlExpiration();
+
+    /// Generate and insert the given number of elements into the given map.
+    /// Each entry is guaranteed to be inserted, but that insertion may purge other entries,
+    /// including entries previously added during the same method call    void addSequenceOfElementsToMap(TestMap &, int count, int startWith, TestMap::Ttl);
```
rousskov, line 43:
```suggestion
    /// including entries previously added during the same method call 
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,133 @@
...
+    CPPUNIT_ASSERT(entriesAfterPurge < m.entries());
+}
+
+void
+testClpMap::testTtlExpiration()
+{
+    TestMap m(2048);
+    m.add(std::to_string(1), 1, 10);
+    CPPUNIT_ASSERT_EQUAL(static_cast<size_t>(1), m.entries());
+    CPPUNIT_ASSERT(static_cast<const int *>(nullptr)!=m.get("1"));
```
rousskov, line 128: I hope the simple expression below works. Please let me know if I missed an earlier comment about it.

```suggestion
    CPPUNIT_ASSERT(m.get("1"));
```

Same for other similar occurrences, of course.

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,133 @@
...
+CPPUNIT_TEST_SUITE_REGISTRATION( testClpMap );
+
+class SquidConfig Config;
+
+void
+testClpMap::addSequenceOfElementsToMap(TestMap &m, int count, int startWith, TestMap::Ttl ttl)
+{
+    for (auto j = startWith; j < startWith + count; ++j)
+    {
+        CPPUNIT_ASSERT_EQUAL(true, m.add(std::to_string(j), j, ttl));
```
rousskov, line 25: To simplify (no value in printing booleans in this case):

```suggestion
        CPPUNIT_ASSERT(m.add(std::to_string(j), j, ttl));
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,133 @@
...
+    addSequenceOfElementsToMap(m, m.memLimit() / sizeof(int), 0, 10);  // TODO: Extract into overflowMap()
+    CPPUNIT_ASSERT(entriesAfterPurge < m.entries());
+}
+
+void
+testClpMap::testTtlExpiration()
+{
+    TestMap m(2048);
+    m.add(std::to_string(1), 1, 10);
+    CPPUNIT_ASSERT_EQUAL(static_cast<size_t>(1), m.entries());
```
rousskov, line 127: Technically, this test is not necessary to test TTL expiration. What matters is that the expiring entry is there and we test that on the next line. I recommend removing it as a distraction:

```suggestion
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,133 @@
...
+    CPPUNIT_ASSERT(static_cast<const int *>(nullptr)!=m.get("1"));
+    squid_curtime += 100;
+    // "1" should have expired
+    CPPUNIT_ASSERT(!m.get("1"));
+}
```
rousskov, line 132: Please test what happens when adding zero-TTL entries. IIRC, they should be accessible during the same second they were added, but not a second later.

rousskov, line 132: Please test that adding an entry with a negative TTLs does not result in that entry being accessible _and_ that the old entry with the same key is purged (or, to be more precise, becomes inaccessible). We spent a ton of time on getting that ClpMap functionality right!

rousskov, line 132: Please test what happens when a same-key entry with shorter TTL is added. It should replace the old entry _and_ the TTL should decrease (i.e. the entry should become inaccessible sooner than the original one would).

rousskov (changes requested):
Thank you for addressing most of my earlier concerns.

src/tests/testClpMap.h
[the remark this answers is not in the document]
rousskov: > As for documenting; yes it is worthwhile - if only to assist future dev updating the unit internals.

FWIW, I wasted at least 10, possibly 15 minutes because I did not understand the now-documented (thank you!) caveats of this complex (in its effects, not implementation) helper function.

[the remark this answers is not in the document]
rousskov: > Same thing for Doxygen: do we even WANT to doxygen-document a two-line unit test helper function?

IMO, yes, because if we do not use Doxygen, somebody will see a non-doxygen comment (e.g., when grepping sources for some specific name) and decide that this is how all (including non-test) functions are documented in Squid, and we will need to ask them to adjust their comments when they post a PR. Or somebody will start arguing whether a particular piece of code is a unit test helper function! Or that a Doxygen comment should be _removed_ from testing code because the latest testing code does not use Doxygen! Etc. Its better to be consistent when adding new code. Fortunately, it does not cost much extra time or effort to add a slash.

src/tests/testClpMap.cc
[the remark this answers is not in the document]
rousskov: See https://github.com/squid-cache/squid/pull/1262#discussion_r1112021206 for the API discussion. It applies to the "support two kind of limits at the same time" suggestion as well.

src/tests/testClpMap.cc
kinkie: > This change request was resolved without addressing it. The change request is valid -- only the second suggestion was flawed (as detailed above). I am unresolving this change request.

Oh, thanks. I had interpreted your concern as only being the second one

src/tests/testClpMap.cc
kinkie: added as testClpMap::testEntriesWithZeroTtl

kinkie: done in testEntriesWithZeroTtl

src/tests/testClpMap.cc
kinkie: done

src/tests/testClpMap.cc
kinkie: All pointer dereferences are now guarded with a nonnull check, we should be good.

src/tests/testClpMap.h
kinkie: done

src/tests/testClpMap.cc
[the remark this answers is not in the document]
yadij: I agree with @rousskov that the memory limit being better for Squid than count limit. It greatly improves the predictability of Squid memory usage - how to control or calculate that is one of the most common queries from users.

src/tests/testClpMap.cc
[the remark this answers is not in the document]
yadij: > However, please note that memoryUsed() for a one-item map is not memPerEntry

Ah right. The key length is variable for this test code.  That could be avoided for test calculations by `addData()` ensuring a fixed-length key string (i.e. 5 bytes "0"-padded).

@rousskov, is that the only variability you see with my suggested test?

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,180 @@
...
+
+void
+testClpMap::testEntriesWithZeroTtl()
+{
+    TestMap m(2048);
+    addSequenceOfElementsToMap(m, 1, 0, 0);
+    CPPUNIT_ASSERT(m.get("0"));  // we get something
+    squid_curtime += 1;
+    CPPUNIT_ASSERT(!m.get("0"));  // we get something
```
yadij, line 167:
```suggestion
    CPPUNIT_ASSERT(m.get("0")); // we get something
    squid_curtime += 1;
    CPPUNIT_ASSERT(!m.get("0")); // expired, we get nothing
```

src/tests/testClpMap.cc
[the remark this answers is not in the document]
rousskov: > @rousskov, is that the only variability you see with my suggested test?

Once I found memPerEntry, I only looked through the high-level C++ comments describing the new tests, not the suggested code itself (because the code depended heavily on the problematic memPerEntry). I plan to revisit this code when this change request is implemented (if it has not been implemented already in the code I reviewed).

Overall, I strongly recommend testing the overall map functionality _without_ assuming exactly how big each entry is, what is the size of various housekeeping information that the map might account for, etc. Otherwise, we risk re-implementing current ClpMap internal code assumptions/decisions in tests, which only makes things worse overall, without significantly increasing the probability of catching implementation bugs.

For example, when we shrink capacity after overfilling the map, we may not be able to predict exactly how many entries will be purged and that is perfectly fine. Instead, we can predict (and should validate) that:

* some entries were purged
* purged entries were purged in LRU order
* some entries are preserved (until we shrink the capacity to zero -- then zero entries should be kept!)
* resulting memory usage (as reported by the map) is at or below the new limit
* adding new entries is still possible

We could also test that, after shrinking the memory limit by half, adding a single entry will purge at least one more entry, but I would _not_ do that because it starts to assume a lot about internal ClpMap decisions. It is possible to imagine a perfectly valid ClpMap implementation that would not pass such a test in various corner cases. We could test that adding, say, 10 new entries after shrinking does purge at least one old entry.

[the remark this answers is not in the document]
rousskov: > > However, please note that memoryUsed() for a one-item map is not memPerEntry ... because the map is not guaranteed to use the same amount of memory for each entry
 
> Ah right. The key length is variable for this test code.

Actually, the primary issue here is not so much the changes in the key length. The issue is that the amount of internal housekeeping information[^1] that the map uses to store items may change as more items are added. For example, the map may allocate a second "bucket" when the first "bucket" gets full (where "bucket" is some undefined internal storage implementation thing). Or create an index but only when the number of entries exceed 100. My point was that we should not (and do not need to) use memPerEntry for this kind of tests.

We could add tests where average memPerEntry or similar stats might be useful. For example, we can test that if the map has more than 10'000 entries, then the housekeeping information does not consume more than 10% of raw/pure entry sizes. However, I doubt we should spend Francesco's time on such tests, in this PR! ;-)

[^1]:  Housekeeping structures do (or should and, hence, eventually will) affect the total reported map memory usage.

rousskov (commented):
Besides missing LRU tests, there are a few minor problems and polishing touches left, but I doubt there is an appetite for working on those items in review-fix-review-fix-review mode. If you give me a chance to polish this code before it is merged, I would appreciate it (just let me know when). If not, then it is not the end of the world (and I might post a followup PR if I get a chance). Thanks again for adding these tests!

src/tests/testClpMap.cc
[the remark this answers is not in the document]
yadij: Oh. That level of housekeeping is too deep for the current ClpMap accounting to even know about.

@kinkie, if you want to defer this change to a future PR I can take a shot at adding a fixed version of my sketch later.

yadij (approved):
More tests can be added later. LGTM now.

src/tests/testClpMap.cc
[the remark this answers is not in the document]
rousskov: > Oh. That level of housekeeping is too deep for the current ClpMap accounting to even know about.

ClpMap already accounts for some housekeeping information at that level. See MemoryCountedFor() where, out of four counted memory consumers, two are about housekeeping overheads.

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,203 @@
...
+void
+testClpMap::testPurgeIsLRU()
+{
+    TestMap m(2048);
+    // TODO: once we have ClpMap iterators we can inspect the contents instead
+    for (int j = 0; j < 10; ++j)
+        CPPUNIT_ASSERT(m.add(std::to_string(j), j, 10));
+    // this makes it the LRU
+    CPPUNIT_ASSERT(m.get("0"));
+    // now overflow the map, checking "0" each time we add an element
```
rousskov, line 193: If we were just "checking", then test would have failed. We are actually making that 0 element the most recently used. I recommend removing correct but misleading (in this context) "this makes it the LRU" assertion above and adjusting this comment to clarify what is actually going on:

```suggestion
    // now overflow the map while keeping "0" the Least Recently Used
```

The above suggestion may also help those code readers that do not know what LRU stands for.

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,203 @@
...
+    for (int j = 100; j < 1000; ++j) {
+        CPPUNIT_ASSERT(m.add(std::to_string(j), j, 10));
+        CPPUNIT_ASSERT(m.get("0"));
+    }
+    CPPUNIT_ASSERT(m.get("0"));
+    CPPUNIT_ASSERT(!m.get("1")); // these should have been aged out
+    CPPUNIT_ASSERT(!m.get("2"));
+    CPPUNIT_ASSERT(!m.get("3"));
+    CPPUNIT_ASSERT(!m.get("4"));
+}
```
rousskov, line 203: Here, I would overflow the map again (using our fillMapWithElements()) and check that "0" is now gone. This would indirectly confirm that there is nothing special about that "0" element except our get() accesses to it in the loop above

```suggestion

    fillMapWithElements(m, 10);
    CPPUNIT_ASSERT(!m.get("0")); // removable when not recently used
}
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,203 @@
...
+        CPPUNIT_ASSERT(m.add(std::to_string(j), j, 10));
+    // this makes it the LRU
+    CPPUNIT_ASSERT(m.get("0"));
+    // now overflow the map, checking "0" each time we add an element
+    for (int j = 100; j < 1000; ++j) {
+        CPPUNIT_ASSERT(m.add(std::to_string(j), j, 10));
+        CPPUNIT_ASSERT(m.get("0"));
+    }
+    CPPUNIT_ASSERT(m.get("0"));
+    CPPUNIT_ASSERT(!m.get("1")); // these should have been aged out
```
rousskov, line 199: Just to clarify and reduce noise:

```suggestion
    // these should have been aged out
    CPPUNIT_ASSERT(!m.get("1"));
```

rousskov (commented):
Thank you for adding an LRU test. The rest of my earlier summary [comment](https://github.com/squid-cache/squid/pull/1262#pullrequestreview-1306419552) still applies.

src/tests/testClpMap.cc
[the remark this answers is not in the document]
kinkie: > For example, when we shrink capacity after overfilling the map, we may not be able to predict exactly how many entries will be purged and that is perfectly fine. Instead, we can predict (and should validate) that:
> 
> * some entries were purged

Isn't this tested in testSetMemLimit ?

> * purged entries were purged in LRU order (in aggregate; i.e. any entry that was kept was more recently "used" than any non-expired entry that was purged)

This is a great idea. Added testClpMap::testPurgeIsLRU

> * some entries are preserved (until we shrink the capacity to zero -- then zero entries should be kept!)

extended testClpMap::testSetMemLimit to check this.
It's a bit kludgy, and it requires making TestMap::MemoryCountedFor public.
By the way, why does it return an optional? I can't see in the code any path that could make it not return a value.

> * resulting memory usage (as reported by the map) is at or below the new limit
> * adding new entries is still possible

> We could also test that, after shrinking the memory limit by half, adding a single entry will purge at least one more entry, but I would _not_ do that because it starts to assume a lot about internal ClpMap decisions. It is possible to imagine a perfectly valid ClpMap implementation that would not pass such a test in various corner cases. We could test that adding, say, 10 new entries after shrinking does purge at least one old entry.

I don't think these cases will actually matter in practice. We could argue that even setting a ClpMap to be capped at 0 size might make no sense in practical terms because that's a degenerate behaviour; but in the end it's okay to add tests for such contingencies, if we stop somewhere sensible

kinkie:
> Besides missing LRU tests, there are a few minor problems and polishing touches left, but I doubt there is an appetite for working on those items in review-fix-review-fix-review mode. If you give me a chance to polish this code before it is merged, I would appreciate it (just let me know when). If not, then it is not the end of the world (and I might post a followup PR if I get a chance). Thanks again for adding these tests!

I'm quite happy with the added tests now. 
Please feel free to improve now, I don't have any large pending changes.

src/tests/testClpMap.cc
```diff
@@ -118,17 +118,51 @@ testClpMap::testSetMemLimit()
...
+        testEntriesBefore = entriesAfterPurge;
+    }
+
+    // set size to 0, all entries should have been removed
+    m.setMemLimit(0);
+    CPPUNIT_ASSERT_EQUAL(size_t(0), m.entries());
+
+    // test that memory used decreases when shrinking
+    size_t curLimit = 10240;
+    size_t lastMemoryUsed;
```
rousskov, line 151: Please avoid undefined variables, especially when that does not buy you anything. To fix this, please do not just move this declaration lower, where we call m.memoryUsed() before the loop. Instead, move this variable _inside_ the loop, adjusting as needed. This change will simplify and otherwise improve this code quite a bit!

src/tests/testClpMap.cc
```diff
@@ -118,17 +118,51 @@ testClpMap::testSetMemLimit()
...
+    m.setMemLimit(curLimit);
+    fillMapWithElements(m, 1000);
+    lastMemoryUsed = m.memoryUsed();
+    while (curLimit > 1024) // stop at any practical size
+    {
+        CPPUNIT_ASSERT(m.memoryUsed() < curLimit);
+        CPPUNIT_ASSERT(m.memoryUsed() <= lastMemoryUsed);
+        // also check that we can still add entries - evicting old ones if needed
+        CPPUNIT_ASSERT(m.add(std::to_string(curLimit+1000), 1, 10));
+        CPPUNIT_ASSERT_EQUAL(1, *m.get(std::to_string(curLimit + 1000)));
```
rousskov, line 161: I do not think we need more magic number for this useful check to work:

```suggestion
        const auto key = std::to_string(curLimit);
        const auto value = TestMap::mapped_type(curLimit);
        CPPUNIT_ASSERT(m.add(key, value, 10));
        CPPUNIT_ASSERT_EQUAL(value, *m.get(key));
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,238 @@
...
+    CPPUNIT_ASSERT_EQUAL(size_t(0), m.entries());
+
+    // test that memory used decreases when shrinking
+    size_t curLimit = 10240;
+    size_t lastMemoryUsed;
+    m.setMemLimit(curLimit);
+    fillMapWithElements(m, 1000);
+    lastMemoryUsed = m.memoryUsed();
+    while (curLimit > 1024) // stop at any practical size
+    {
```
rousskov, line 156: Warning: This change request may depend on other fixes in this review.

If possible, let's reach the natural limit: AFAICT, as long as there is at least one entry, the request to reduce current memory usage should result in less memory used, even if our request was reducing current memory usage by just one byte.

```suggestion
    while (m.entries()) { 
```

You may want to assert that curLimit is positive inside the loop to make sure we are making test progress even if the map is buggy.

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,238 @@
...
+        CPPUNIT_ASSERT(m.memoryUsed() < curLimit);
+        CPPUNIT_ASSERT(m.memoryUsed() <= lastMemoryUsed);
```
rousskov, line 158: Warning: This change request depends on other fixes in this review!

If we asked for less memory, we should get less memory (as long as there are still entries to remove).

```suggestion
        CPPUNIT_ASSERT(m.memoryUsed() < lastMemoryUsed);
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,238 @@
...
+
+    // set size to 0, all entries should have been removed
+    m.setMemLimit(0);
+    CPPUNIT_ASSERT_EQUAL(size_t(0), m.entries());
+
+    // test that memory used decreases when shrinking
+    size_t curLimit = 10240;
+    size_t lastMemoryUsed;
+    m.setMemLimit(curLimit);
+    fillMapWithElements(m, 1000);
```
rousskov, line 153: I would overflow the map with elements on each loop iteration to make things a bit more "dynamic", but I do not insist on this change because it can be argued that the current "pure removal" loop also tests an important invariant.

rousskov (changes requested):

rousskov:
> @rousskov Please feel free to improve now, I don't have any large pending changes.

Thank you. I wrote my review before seeing this comment. Please disregard my review. I will address those change requests.

src/tests/testClpMap.cc
rousskov: I tried and shelved that idea because our existing map overflowing function is not compatible with an empty map. Arguably, it should be, but I ran out of time to polish that.

src/tests/testClpMap.cc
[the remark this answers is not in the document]
rousskov: Done in cd911ae. FWIW, I did not run into any of the [predicted problems](https://github.com/squid-cache/squid/pull/1262/files#r1111673865).

src/mem/PoolingAllocator.h
```diff
@@ -11,6 +11,8 @@
 
 #include "mem/forward.h"
 
+#include <utility>
```
rousskov, line 14: This (correct) change looks out of scope, but it is needed in this PR to compile the new test code:

```
../src/mem/PoolingAllocator.h:31:102: error: ‘forward’ is not a member of ‘std’
../src/mem/PoolingAllocator.h:13:1: did you forget to ‘#include <utility>’?
```

I am not sure why it was not needed during earlier stages of this PR development, but I suspect it could be because a different order of `#include` directives hid this problem.

rousskov (approved):
I have addressed and resolved my own change requests. I tried to briefly document each stand-alone change in the commit messages. More polishing could be done, but I ran out of time.

Please check that you are OK with the current code and adjust as needed (or let me know if I should fix something) or clear it for merging. Thank you.

src/tests/testClpMap.cc
```diff
@@ -153,55 +165,41 @@ testClpMap::testConstructor()
...
-        // also check that we can still add entries - evicting old ones if needed
-        CPPUNIT_ASSERT(m.add(std::to_string(curLimit+1000), 1, 10));
-        CPPUNIT_ASSERT_EQUAL(1, *m.get(std::to_string(curLimit + 1000)));
-        curLimit /= 2;
-        lastMemoryUsed = m.memoryUsed();
-        m.setMemLimit(curLimit);
+    CPPUNIT_ASSERT(m.entries() > entriesAtInitialCapacity);
+
+    // test that memory usage and entry count decrease when the map is shrinking
+    while (m.entries()) {
```
kinkie, line 211: I'm a bit skeptical about this test condition: if there were a bug in ClpMap and entries didn't shrink, this would cause the test to endlessly loop; not the clearest outcome, and unfriendly to automations

src/tests/testClpMap.cc
```diff
@@ -153,55 +165,41 @@ testClpMap::testConstructor()
...
+        // also check that we can still add entries, evicting old ones if needed
+        // (at least as long as there is at least one entry in the map)
+        addOneEntry(m, 0);
```
kinkie, line 214: Note that adding an entry fails with error if the map storage is too small to even contain one entry

src/tests/testClpMap.cc
```diff
@@ -254,7 +254,6 @@ void
 testClpMap::testPurgeIsLRU()
 {
     TestMap m(2048);
-    // TODO: once we have ClpMap iterators we can inspect the contents instead
```
kinkie: In my mind they would, because instead of keeping an external index and very carefully inspecting some elements, we could just look them up, and then do set comparison. Not a big deal tho

kinkie:
@rousskov  Thanks for adding to the testsuite! There's one request from me, once we agree on it, this can be shipped as far as I'm concerned

src/tests/testClpMap.cc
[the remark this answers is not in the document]
yadij: FYI, for some of the cache types this is intended to replace (eg helper responses) there are valid situations where admin set capacity to 0 bytes. Having the ClpMap handle that saves callers individually duplicating all the logic testing if cache is disabled.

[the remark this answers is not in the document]
kinkie: > FYI, for some of the cache types this is intended to replace (eg helper responses) there are valid situations where admin set capacity to 0 bytes. Having the ClpMap handle that saves callers individually duplicating all the logic testing if cache is disabled.

Got you.
The only thing that makes life complicated IMO for callers is that add() returns false if the entry could not be added because memlimit is such that the map can't even store one single entry. This means that either add() return value should be ignored, or that callers have to special case this anyway. Have you considered this?

src/tests/testClpMap.cc
rousskov: This loop cannot run endlessly because it asserts that an unsigned value (m.entries()) always decreases.

src/tests/testClpMap.cc
rousskov: You are right (and I came to the same conclusion after leaving the office last night). Replaced with a TODO in 42708cf.

src/tests/testClpMap.cc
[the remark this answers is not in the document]
rousskov: > Francesco:  By the way, why does TestMap::MemoryCountedFor() return an optional? I can't see in the code any path that could make it not return a value.

Adding an entry with a ULLONG_MAX-byte key[^1] is one of the use cases where MemoryCountedFor() cannot return an accurate answer and, hence, should not return any answer. This is one of the missing test cases, BTW.

The only alternative is an assertion, but since ClpMap is essentially a cache, it feels wrong to kill Squid just because we cannot cache.

[^1]: Which might happen due to `-1` to `unsigned` conversion bugs in caller's implementation of the MemoryUsedBy() API. Similar Squid bugs have been discovered in the past, of course, and I am pretty sure there will be more.


> either add() return value should be ignored, or that callers have to special case this anyway

The API accommodates callers that want to debugs() the outcome or update some indexes/stats. And even among those callers that pay attention to the return value, many will not have to special-case it. Many examples of such paying-attention not-special-casing (in the true/core meaning of that phrase) callers are added in this PR.

rousskov (commented):
@kinkie, I hope I addressed your concerns. Please re-review. Code-wise, there was only one new commit: 42708cf.

src/tests/testClpMap.cc
rousskov: I am not convinced because the code is already "looking them up", without iterators. Moreover, iterators are not really meant for looking up _specific_ values, which is what this code needs AFAICT. However, I would be happy to be proven wrong: We agree that this code needs improvement, and the details are not important at the moment.

src/tests/testClpMap.cc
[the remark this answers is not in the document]
rousskov: > FYI, for some of the cache types this is intended to replace (eg helper responses) there are valid situations where admin set capacity to 0 bytes.

Agreed. I would even assert that a zero-capacity (or equivalent) cache should be "valid" in the _majority_ of use cases, especially where the cache capacity is admin-configurable or dynamically computed from environment characteristics outside code control.

src/tests/testClpMap.cc
kinkie: I think I wasn't clear. What I am considering is a much simpler:

```suggestion
    int maxIterations = 10000;
    while (m.entries() && --maxIterations > 0) {
```

src/tests/testClpMap.cc
```diff
@@ -0,0 +1,346 @@
...
+        const auto memoryUsedBefore = m.memoryUsed();
+        const auto entriesBefore = m.entries();
+
+        const auto newMemoryLimit = memoryUsedBefore/2; // may become zero
+        m.setMemLimit(newMemoryLimit);
+
+        CPPUNIT_ASSERT(m.memoryUsed() <= newMemoryLimit);
+        CPPUNIT_ASSERT(m.entries() < entriesBefore);
+    }
+
```
kinkie, line 223: paired with
```suggestion
    CPPUNIT_ASSERT(maxIterations > 0);
```

src/tests/testClpMap.cc
rousskov: TLDR: I added a source code comment in hope to better address your concern (commit d494db2c3463195a4c4565ba5760140be996e805).

----

Original change request: This loop may be endless if ClpMap is buggy.

Correction: This loop is not endless even if ClpMap is buggy.

Modified change request: Let's consider adding more code, adding more implicit dependencies among constants, and making the loop condition more complex.

I assume that all those changes are meant to make it more obvious that the loop is not endless. IMO, that goal is just not worth the cost, especially when the worst thing that could happen[^1] (with or without) those changes is an endless loop during "make check".

[^1]: And only in a scenario where somebody breaks the testing code itself -- something no solution in this testing code, not even the proposed explicit counter, can protect us from! Again, breaking ClpMap will _not_ break this testing code.

Personally, I would have _more_ questions about the loop with maxIterations code than without it. And I can easily imagine other developers starting to copy (or demand) similar paranoid additions to similar loops in new or modified code!

However, this is _your_ PR. If you insist on adding that code, then do it. My only request is that you also add a comment to explain why that extra code is there. I suggested specific wording in your other change request (because we need to explain the new proposed assertion anyway).

src/tests/testClpMap.cc
rousskov: If (and only if) you decide to add those maxIteration checks despite my concerns:

```suggestion
    assert(!m.entries()); // if this fails, the above loop is broken (and not necessarily ClpMap)
 
```

or, if you insist,

```suggestion
    assert(maxIterations > 0); // if this fails, the above loop is broken (and not necessarily ClpMap)

```

or something like that.

kinkie: By "concerns" do you mean the fact that the test - and not the map - might be broken?
My concern is similar; I'd like to ensure that the test reliably terminates, that's less confusing than the alternative.

We can go one step further and use CPPUNIT_ASSERT_MESSAGE(message, condition) to print the diagnostic out in the test output itself

e.g.
```
CPPUNIT_ASSERT_MESSAGE("endless loop detected, aborting", !m.entries() && maxIterations > 0)
```
or something similar

rousskov: > By "concerns" do you mean the fact that the test - and not the map - might be broken? 

No, I did not mean that. I was concerned that "adding more code, adding more implicit dependencies among constants, and making the loop condition more complex" does more harm than good in this specific case. See https://github.com/squid-cache/squid/pull/1262#discussion_r1116440482 for the associated discussion.

Most of those concerns have not changed. However, I was wrong that this test loop is implemented correctly! I fixed my  bug in branch commit a39baed. Hopefully, that fix addresses your concerns as well. The ball is on your side.

src/tests/testClpMap.cc
rousskov: This thread has been switched to https://github.com/squid-cache/squid/pull/1262/files#r1116614481

src/tests/testClpMap.cc
```diff
@@ -211,7 +211,7 @@ TestClpMap::testMemoryLimit()
     // test that memory usage and entry count decrease when the map is shrinking
     // but prevent endless loops no matter how broken ClpMap implementation is
     auto iterationsLeft = m.entries();
-    CPPUNIT_ASSERT_MESSAGE("ClpMap::entries() returns bogus numbers",
+    CPPUNIT_ASSERT_MESSAGE("ClpMap::entries() returns reasonable numbers",
```
kinkie, line 214: The message should document the failure, see https://cppunit.sourceforge.net/doc/cvs/group___assertions.html#ga1

kinkie:
Confirmed that the message documents failure; changing the message and forcing iterationsLeft to 1 yields:
```
tests/testClpMap.cc:230: Assertion
Test name: TestClpMap::testMemoryLimit
assertion failed
- Expression: iterationsLeft > 0
- endless loop detected while shrinking map

Failures !!!
```

Reverting the message changes; I'll clear for merge in 12 hours unless stopped

src/tests/testClpMap.cc
rousskov: I removed the controversial messages (commit cfea6fa) in hope to make progress. This fiasco also reminded me that it is actually a bad idea to use those messages in the first place. C++17 got it right when they changed the static_assert() API to stop requiring these messages: These messages are more trouble than they are worth, for several reasons! We have discussed this at length [before](https://github.com/squid-cache/squid/pull/785#discussion_r594365508).

> The message should document the failure

I disagree:

* #785 (i.e. commit 2bde9c7) has already established that assertion messages should document the condition. We spent hours on that decision. I support that decision, but if it was wrong, reversing it should be done in a dedicated PR, not here.
* CppUnit [documentation](https://cppunit.sourceforge.net/doc/cvs/group___assertions.html#ga1) and output do not contradict the stated principle IMO. AFAICT, CppUnit treats the message as supplementary output without _requiring_ it to describe either the condition or the failure. I did test CppUnit output before adding those messages, came to the conclusion that it looks OK (and even included a sample in PR commit 79714fc).
* All existing CPPUNIT_ASSERT_MESSAGE() calls in Squid document the condition. If nothing else, it is best to keep the code consistent (and change all messages at once if the decision to change messages is made in a dedicated PR).

Besides a link to the documentation and output (which I interpret in my favor), no _arguments_ have been presented to support the change AFAICT. IMHO, this ordeal is a good illustration why we should avoid those messages completely.

rousskov:
> Confirmed that the message documents failure

I disagree, as detailed in the corresponding [change request](https://github.com/squid-cache/squid/pull/1262#discussion_r1117628929). I hope we can (and think that we should) proceed without those messages instead (commit cfea6fa).

src/tests/testClpMap.cc
kinkie: I agree on avoiding the messages and am satisfied that the condition I was concerned is there.
Let's clear for merge and ship it

rousskov: > Let's clear for merge and ship it

Glad we found an acceptable solution! Please clear the PR for merging when you think it should be cleared.

PR #1266, opened 2023-02-05
Setting receivedWholeAdaptedReply https://bugs.squid-cache.org/show_bug.cgi?id=5187#c6

ShmayaFrankel:


src/client_side_request.cc
```diff
@@ -2115,11 +2115,13 @@ ClientHttpRequest::noteMoreBodyDataAvailable(BodyPipe::Pointer)
...
     }
-
-    if (adaptedBodySource->exhausted())
-        endRequestSatisfaction();
-    // else wait for more body data
-}
+    if (adaptedBodySource->exhausted()) {
+        // XXX Setting receivedWholeAdaptedReply here is a workaround for a
+        // regression, as described in https://bugs.squid-cache.org/show_bug.cgi?id=5187#c6
+        receivedWholeAdaptedReply = true;
```
rousskov, line 2121: Please note that merging this workaround "as is" may, in some cases, resurrect the original bug 5187.

If there is consensus that this change should go in, please restore the useful "else wait for more body data" comment (after the `if` statement) and fix formatting. I would also add a `WARNING: Squid bug 5187 workaround triggered` debugs() message, so that we have a better chance of remembering about and eventually fixing the underlying problem. I can help with these additional steps if needed -- just ask.

My earlier [comments](https://github.com/squid-cache/squid/pull/1267#issuecomment-1418267894) (in another PR) regarding merging your changes directly into v5 apply here as well.

ShmayaFrankel: @rousskov i know it has been a while, was trying to find other solutions around this, but no luck.
i would appreciate any help you can give here..

rousskov: @ShmayaFrankel, just to clarify, are you asking for help with finding a proper solution for this bug or for help with the minor workaround adjustments that I have mentioned in my [comment](https://github.com/squid-cache/squid/pull/1266#discussion_r1096827453)?

PR #1267, opened 2023-02-05
Bug 4981: Work around in-call job invalidation bugs

ShmayaFrankel:
fixes https://bugs.squid-cache.org/show_bug.cgi?id=4981
works well with this patch.


rousskov:
Thank you for posting Bug 4981 workaround patch for v5 inclusion. Most likely, you will need to adjust the PR metainformation to target the master branch (v5 and v6 will then get a backport if their maintainers allow it). See [MergeProcedure](https://wiki.squid-cache.org/MergeProcedure).

FWIW, I recommend using the workaround patch preamble as the PR description (it will become the future commit message body as detailed [elsewhere](https://github.com/measurement-factory/anubis/blob/master/README.md#pr-description)).

PR #1270, opened 2023-02-09
Enable compiler checks for printf-like SBuf methods

eduard-bagdasaryan:
After commit 0f17e25, these two are the only known printf()-like
functions not already covered by these compiler checks. Some helper
debugging hacks do not have PRINTF_FORMAT_ARG2 or similar attributes,
but they are covered because their `__GNUC__` variants use standard
fprintf().

### DIFF
--- a/src/clients/FtpGateway.cc
+++ b/src/clients/FtpGateway.cc
@@ -143,7 +143,8 @@ public:
     void checkUrlpath();
     void buildTitleUrl();
     void writeReplyBody(const char *, size_t len);
-    void printfReplyBody(const char *fmt, ...);
+    // unused
+    void printfReplyBody(const char *fmt, ...) PRINTF_FORMAT_ARG2;
     void completeForwarding() override;
     void processHeadResponse();
     void processReplyBody() override;
--- a/src/clients/FtpGateway.cc
+++ b/src/clients/FtpGateway.cc
@@ -143,8 +143,7 @@ public:
     void checkUrlpath();
     void buildTitleUrl();
     void writeReplyBody(const char *, size_t len);
-    // unused
-    void printfReplyBody(const char *fmt, ...) PRINTF_FORMAT_ARG2;
+    void printfReplyBody(const char *fmt, ...);
     void completeForwarding() override;
     void processHeadResponse();
     void processReplyBody() override;
--- a/src/sbuf/SBuf.h
+++ b/src/sbuf/SBuf.h
@@ -211,13 +211,13 @@ public:
      * \note arguments may be evaluated more than once, be careful
      *       of side-effects
      */
-    SBuf& Printf(const char *fmt, ...);
+    SBuf& Printf(const char *fmt, ...) PRINTF_FORMAT_ARG2;
 
     /** Append operation with printf-style arguments
      * \note arguments may be evaluated more than once, be careful
      *       of side-effects
      */
-    SBuf& appendf(const char *fmt, ...);
+    SBuf& appendf(const char *fmt, ...)  PRINTF_FORMAT_ARG2;
 
     /** Append operation, with vsprintf(3)-style arguments.
      * \note arguments may be evaluated more than once, be careful


eduard-bagdasaryan:
> These two are the only known printf()-like functions not already covered

One more case is now dealt by [PR1271](https://github.com/squid-cache/squid/pull/1271).

rousskov (approved):
Thank you. I wish I had this compiler protection when I was creating bugs you fixed in #1255 commit 7e8c0f2!

Eventually, we should update/modernize `PRINTF_FORMAT_ARG2` definition itself, to enable clang testing as well, but that is out of this PR scope, and what we have now ought to be sufficient because our CI uses GCC.

rousskov:
> PR description: These two are the only known printf()-like functions not already covered

> Eduard: One more case is now dealt by PR1271

Right. To make sure the future commit message from this PR is not lying, I am marking this PR as waiting for that PR 1271.

kinkie (approved):
Good catch! Thanks for identifying this

rousskov:
@yadij, I have restored the original PR description because, IMO, it documents the assumptions/scope of this PR better than its shorter variant. We can discuss this further if really needed.

In the future, please do not make (non-trivial) changes to PR descriptions without logging a comment because (AFAIK) GitHub does not inform humans about such changes, making them very easy to overlook.

yadij (changes requested):
It is bad idea to knowingly lie to downstream vendors and backporters.

rousskov:
> It is bad idea to knowingly lie to downstream vendors and backporters.

Agreed. The above statement implies that you know something that I do not, so please detail your change request: Which printf()-like functions are not covered by these compiler checks (after this PR and its prerequisite PR 1271)?

yadij:
> > It is bad idea to knowingly lie to downstream vendors and backporters.
> 
> Agreed. The above statement implies that you know something about PR description claims that I do not, so please detail your change request: Which printf()-like functions are not covered by these compiler checks (after this PR and its prerequisite PR 1271)?

The issue I have with the current description is precisely that it can only be true when read in the post-1271 context. Yet has no indication that PR 1271 is a prerequisite. Vendors can reasonably be expected to only see the commit patch with this description as an isolated submission to a bug report.

The first sentence also does not add anything useful to the commit. It seems to be self-justification for a change which needs no justification, and is the implicit source of the PR 1271 dependency. So I see no valid reason for it to even be stated.

rousskov:
> > > It is bad idea to knowingly lie to downstream vendors and backporters.

> > Agreed. The above statement implies that you know something about PR description claims that I do not, so please detail your change request: Which printf()-like functions are not covered by these compiler checks (after this PR and its prerequisite PR 1271)?
 
> The issue I have with the current description is precisely that it can only be true when read in the post-1271 context. Yet has no indication that PR 1271 is a prerequisite.

I have now addressed this concern by starting the PR description with "After commit 0f17e25". Please unblock this PR.

> The first sentence also does not add anything useful to the commit. 

The first sentence adds useful (to me) information: The author has checked for other cases where PRINTF_FORMAT_ARG2 is missing and found none.

* Without that information, a high-quality PR review would have to ask the author: "Have you checked for _other_ cases where PRINTF_FORMAT_ARG2 attribute is missing?" Compared to a properly documented PR, additional Q&A wastes time and often becomes impractical after the PR is merged.

* Without that knowledge, one cannot rely on those checks in code refactoring that goes beyond the two SBuf methods. Nothing is guaranteed, of course, but there is an important difference between "I just tagged two SBuf methods" and "I searched for all missing attribute cases and found two SBuf methods to tag"!

> It seems to be self-justification for a change which needs no justification

This PR description is not trying to justify this PR. It provides _additional_ useful information that cannot be derived from the diff. You are correct that the specific code changes in this PR do not need any explicit justification, but a good commit message body is rarely about the code changes that can be easily understood by looking at the commit title and the diff.

----

I continue this part of the discussion in hope to prevent similar PR blocks in the future...

> Vendors can reasonably be expected to only see the commit patch with this description as an isolated submission to a bug report.

Vendors and developers should know that commits must be interpreted in their natural repository context. If somebody submits a patch that looks like a project repository commit, then vendors can reasonably be expected to understand that the description was taken out of its natural context (and act accordingly). If somebody submits a patch that does not look like a project repository commit but reuses a context-sensitive original commit description, then any resulting misinterpretations are on that submitter, not the original commit authors.

In any case, "based on all the changes up to this commit" is the correct commit context _default_. We should not block PRs that use that (very natural) context. Prefixing commit descriptions with "As of the previous commit" phrases should never be required.

This PR was marked as waiting for its prerequisite PR, explicitly establishing that dependency and setting up correct commit context expectations. Eduard even added a GitHub [comment](https://github.com/squid-cache/squid/pull/1270#issuecomment-1424909817) describing that dependency/expectations. In exceptional situations, a reviewer may correctly block a PR until a key prerequisite PR is merged, but the reason for that kind of blocking -- waiting for another PR to land -- should be clearly stated. In case of this PR, a very misleading reason was given instead.

yadij (approved):
I still believe this is unnecessary detail in the description. But at least the core issue is gone now.

PR #1273, opened 2023-02-10
Place more msgs under cache_log_message control, downgrade some

rousskov:
Also support cache_log_message-controlled messages in ::Config-unaware
libraries (e.g., libip) used by tools (e.g., cachemgr.cgi).

Restricting the very first (i.e. "depth 0") "Processing Configuration
File" message does not work during startup (before that file is parsed),
and does not work during reconfiguration (because the old configuration
is reset shortly before logging that line). Future reconfiguration
support improvements may fix the reset problem. Restrictions do work as
expected for included files (i.e. positive "depth" levels).

Restricting "BCP 177 violation" WARNINGs does not work because the
warnings are printed _before_ Squid configuration is parsed. Future
initialization improvements may fix this.

Also downgraded the importance of a few (re)configuration
progress-reporting messages from level 1 to level 2.


src/ssl/support.cc
```diff
@@ -727,7 +728,7 @@ Ssl::InitClientContext(Security::ContextPointer &ctx, Security::PeerOptions &pee
         // TODO: support loading multiple cert/key pairs
         auto &keys = peer.certs.front();
         if (!keys.certFile.isEmpty()) {
-            debugs(83, DBG_IMPORTANT, "Using certificate in " << keys.certFile);
+            debugs(83, Important(78), "Using client certificate in " << keys.certFile);
```
rousskov, line 731: This message text polishing is needed to make IDed messages unique:

```
78 Using client certificate in ...
79 Using certificate in ...
```

src/AuthReg.cc
```diff
@@ -30,18 +32,18 @@
...
-    static const char *digest_type = Auth::Digest::Scheme::GetInstance()->type();
-    debugs(29,DBG_IMPORTANT,"Startup: Initialized Authentication Scheme '" << digest_type << "'");
+    Assure(Digest::Scheme::GetInstance());
+    debugs(29, Important(71), "Startup: Initialized Authentication Scheme 'digest'");
 #endif
 #if HAVE_AUTH_MODULE_NEGOTIATE
-    static const char *negotiate_type = Auth::Negotiate::Scheme::GetInstance()->type();
-    debugs(29,DBG_IMPORTANT,"Startup: Initialized Authentication Scheme '" << negotiate_type << "'");
+    Assure(Negotiate::Scheme::GetInstance());
+    debugs(29, Important(72), "Startup: Initialized Authentication Scheme 'negotiate'");
```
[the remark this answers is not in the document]
yadij: You may recall I have been working on making these modules be initialized by Runners. I was planning to downgrade the level to 2+ at that time.

It seems a bit out of scope to do that in this PR, but I wont object if you want to include. Especially since it will allow merging message ID 68 to 71.

src/acl/Ip.cc
```diff
@@ -13,6 +13,7 @@
 #include "acl/Ip.h"
 #include "cache_cf.h"
 #include "ConfigParser.h"
+#include "debug/Messages.h"
 #include "debug/Stream.h"
```
yadij, line 17: `debug/Message.h` imports `debug/Stream.h` no need to have both.
```suggestion
#include "debug/Messages.h"
```

src/adaptation/icap/ServiceRep.cc
```diff
@@ -18,6 +18,7 @@
 #include "base/TextException.h"
 #include "comm/Connection.h"
 #include "ConfigParser.h"
+#include "debug/Messages.h"
 #include "debug/Stream.h"
```
yadij, line 22:
```suggestion
#include "debug/Messages.h"
```

src/ip/tools.cc
```diff
@@ -9,6 +9,7 @@
 /* DEBUG: section 21    Misc Functions */
 
 #include "squid.h"
+#include "debug/Messages.h"
 #include "debug/Stream.h"
```
yadij, line 13:
```suggestion
#include "debug/Messages.h"
```

src/cache_cf.cc
```diff
@@ -961,7 +961,7 @@ configDoConfigure(void)
 #endif
 
     if (Security::ProxyOutgoingConfig.encryptTransport) {
-        debugs(3, DBG_IMPORTANT, "Initializing https:// proxy context");
+        debugs(3, Important(74), "Initializing https:// proxy context");
```
yadij, line 964: FWIW, I have been thinking for a while that all these "Initializing ..." messages should be level 2+. Just have not gotten around to doing it. If you agree please do that instead.

Same goes for the section 83 "Using ..." messages shown on startup/reconfigure.

yadij (changes requested):
Some things to think on. The only bits I insist on are the `#include` cleanup.

src/AuthReg.cc
```diff
@@ -30,18 +31,18 @@
 void
 Auth::Init()
 {
-    debugs(29,DBG_IMPORTANT,"Startup: Initializing Authentication Schemes ...");
+    debugs(29, Important(68), "Startup: Initializing Authentication Schemes ...");
```
rousskov, line 34: If you are OK with downgrading this message (currently ID 68), please let me know, and I will gladly downgrade it to level 2 as well:

```suggestion
    debugs(29, 2, "Startup: Initializing Authentication Schemes ...");
```

src/AuthReg.cc
```diff
@@ -30,18 +31,18 @@
...
 #endif
 #if HAVE_AUTH_MODULE_DIGEST
     static const char *digest_type = Auth::Digest::Scheme::GetInstance()->type();
-    debugs(29,DBG_IMPORTANT,"Startup: Initialized Authentication Scheme '" << digest_type << "'");
+    debugs(29, 2, "Startup: Initialized Authentication Scheme '" << digest_type << "'");
 #endif
 #if HAVE_AUTH_MODULE_NEGOTIATE
     static const char *negotiate_type = Auth::Negotiate::Scheme::GetInstance()->type();
-    debugs(29,DBG_IMPORTANT,"Startup: Initialized Authentication Scheme '" << negotiate_type << "'");
+    debugs(29, 2, "Startup: Initialized Authentication Scheme '" << negotiate_type << "'");
```
rousskov, line 45: BTW, if you want me to remove somewhat misleading `Startup: ` prefixes from the four messages above, please let me know, and I will do that in this PR. These messages are not specific to Squid startup. Auth::Init() is also called on reconfigurations...

src/AuthReg.cc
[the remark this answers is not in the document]
rousskov: Glad we seem to agree on all counts! Done in 321c28d.

src/acl/Ip.cc
rousskov: Why do you insist on these changes? I do not think they are a good idea: The users of header A are not supposed to know what other headers A includes -- the `#include` list of header A is an internal business of header A (despite the fact that C++ cannot make that information private or hidden from A users). That include list may change at any time, depending on internal A needs, and those changes should not affect any users (that use unaffected header A API). 

The minute time savings during compilation are not worth the inconveniences this "know what the included headers include and optimize to include once" approach causes IMO.

Please reconsider.

src/cache_cf.cc
rousskov: Great, I will do that.

src/AuthReg.cc
yadij: Sure :)

src/acl/Ip.cc
yadij: Beyond the precompiler work, it simplifies the dependency graphs in code documentation. Those graphs can be complicated enough without unnecessary edges causing the graph generator to shuffle object representations around to fit them in.
This is a weak insistence - I do not expect people to remember or even check every include list, but please do when it is known (like now). Eventually we should get this check automated and not have to care.

src/debug/Messages.h
```diff
@@ -73,6 +72,9 @@ class DebugMessages
     Storage messages;
 };
 
+/// Global configuration for DebugMessageLevel() (where/when supported).
+inline DebugMessages *DebugMessagesConfig = nullptr;
```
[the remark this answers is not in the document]
yadij: IMO it perhapse should have. But does fit into this scope, so no problem.

src/AuthReg.cc
```diff
@@ -21,6 +21,7 @@
 #include "auth/negotiate/Scheme.h"
 #endif
 
+#include "debug/Messages.h"
```
yadij, line 24: With the shift to level 2 initialization messages, do we need to add this still?

src/ip/tools.cc
```diff
@@ -78,7 +78,7 @@ Ip::ProbeTransport()
 #else
     debugs(3, 2, "IPv6 transport " << (EnableIpv6?"Available":"Disabled"));
     if (EnableIpv6 != IPV6_OFF) {
-        debugs(3, DBG_CRITICAL, "WARNING: BCP 177 violation. IPv6 transport forced OFF by build parameters.");
+        debugs(3, Critical(67), "WARNING: BCP 177 violation. IPv6 transport forced OFF by build parameters.");
```
rousskov, line 81: > PR description: Restricting "BCP 177 violation" WARNINGs does not work because the warnings are printed before Squid configuration is parsed. Future initialization improvements may fix this.

We do not have to give these messages IDs because controlling them does not work "as is" right now. If you want me to undo these two "BCP 177" changes, please let me know.

rousskov (commented):
I believe I have addressed all change requests. PR title and description have been updated accordingly. Please re-review.

src/acl/Ip.cc
yadij: Thank you for the change despite your disagreement.

src/ip/tools.cc
yadij: This is fine with me. The probes timing may change as the configuration actions are refactored, so it is good to be ready for that here.

yadij (approved):

src/acl/Ip.cc
rousskov: I had no choice in the matter.

PR #413, opened 2019-06-03
Fix GCC-9 build issues

yadij:
GCC-9 continues the development track started with GCC-8
producing more warnings and errors about possible code issues
which Squid use of "-Wall -Werror" turns into hard build
failures:

 error: ‘strncpy’ output may be truncated copying 6 bytes from a
   string of length 6 [-Werror=stringop-truncation]

 error: ‘%s’ directive argument is null
  [-Werror=format-overflow=]

 error: ‘void* memset(void*, int, size_t)’ clearing an object of
  type ... with no trivial copy-assignment; use assignment or
  value-initialization instead [-Werror=class-memaccess]

 error: ‘void* memset(void*, int, size_t)’ clearing an object of
  non-trivial type ...; use assignment or value-initialization
  instead [-Werror=class-memaccess]

Also, segmentation faults with minimal builds have been
identified as std::string template differences between
optimized and non-optimized object binaries. This results in
cppunit (built with optimizations) crashing unit tests when
freeing memory. Workaround that temporarily by removing the use
of --disable-optimizations from minimal builds.

PR #1278, opened 2023-02-14
CI: Disable optimizations in "minimal" build tests

rousskov:
Our "minimal" layer tests had to stop disabling optimizations in 2019
commit 4f3c41b due to problems specific to GCC v9 (v9.1 or earlier). GCC
v9.4 (and later) appear to work OK (at least on Ubuntu 22.04), so let's
try to resume testing --disable-optimizations ./configure option.

### DIFF
--- a/test-suite/buildtests/layer-01-minimal.opts
+++ b/test-suite/buildtests/layer-01-minimal.opts
@@ -30,9 +30,6 @@ MAKETEST="distcheck"
 #   --without-filedescriptors \
 #   --without-build-environment \
 #
-#  GCC-9 cannot (yet) test this due to segmentation faults in std::string template optimizations (via cppunit)
-#	--disable-optimizations \
-#
 # NP: DISTCHECK_CONFIGURE_FLAGS is a magic automake macro for the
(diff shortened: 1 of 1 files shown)


rousskov:
Disclaimer: I only tested with GCC v9.5 and later releases. I have not tested with earlier GCC v9 versions.

If you know how to reproduce the problems mentioned in #413, please see if they still exist with later GCC versions. I do not.

I think we should merge these changes even if GCC v9.0 or v9.1 (not sure which one had problems mentioned in #413) will fail the adjusted "minimal" test-builds.sh test: If an early compiler release is buggy, folks using it can and should upgrade, especially if they want to use the latest Squid master branch. Such compiler bugs should not force developers working with the latest Squid master branch (and using reasonably up-to-date compilers) to manually find and fix build bugs that could have been _prevented_ by testing this useful ./configure option in our CI (e.g., the bug fixed in recent #1276).

yadij:
Squid Project master server is currently on GCC 9.4. I am running a test build now to see if that version is okay.

yadij (approved):

yadij:
Test passed, updating description to mention.

PR #1280, opened 2023-02-16
Source Format Enforcement

rousskov:
NOTICE: File tools/squidclient/squidclient.cc changed: by astyle

### DIFF
--- a/tools/squidclient/squidclient.cc
+++ b/tools/squidclient/squidclient.cc
@@ -441,7 +441,7 @@ main(int argc, char *argv[])
         if (const auto at = strrchr(url, '@')) {
             if (!OriginAuthorization.user) {
                 std::cerr << "ERROR: Embedding a password in a cache manager command requires " <<
-                    "providing a username with -U: mgr:" << t << std::endl;
+                          "providing a username with -U: mgr:" << t << std::endl;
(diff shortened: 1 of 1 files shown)


rousskov:
If this PR is in the way of the ongoing "branching" work, please close it.

yadij:
Thank you for your interest in improving Squid. However please be aware that this change only hides the automation vs ideology problem we are currently working out in the other PR discussions. I would prefer to properly resolve that underlying disagreement/problem. Do you really want to do this manually?

rousskov:
> please be aware that this change only hides the automation vs ideology problem we are currently working out in the other PR discussions. I would prefer to properly resolve that underlying disagreement/problem.

I hope we can "properly resolve" any problems with or without this PR merged. Again, if this PR is in the way of the ongoing "branching" work (or anything else of importance), please close it.

> Do you really want to do this manually?

I assume "this" in the above question is "format sources". Long-term, source code formatting should be automated and enforced, of course. Right now, I would like the current master sources to be formatted (hence, this PR), so that I can stop manually undoing these formatting changes in my master-based branches (after running scripts/source-maintenance.sh to adjust modified sources in those branches).

yadij (approved):

PR #1281, opened 2023-02-17
Ensure StoreEntry presence during store_client lifetime

rousskov:
Most likely, all store_client users already lock their StoreEntry
objects, so this change is unlikely to affect any current use cases.
However, objects like store_client that store StoreEntry objects for
long-term/asynchronous reuse should lock their entries rather than rely
on others to do that for them. Being able to rely on StoreEntry locking
also clarifies existing logic in several store_client methods.

### DIFF
--- a/src/store_client.cc
+++ b/src/store_client.cc
@@ -216,6 +216,9 @@ store_client::store_client(StoreEntry *e) :
     object_ok(true),
     copiedSize(0)
 {
+    Assure(entry);
+    entry->lock("store_client");
+
     flags.disk_io_pending = false;
     flags.store_copying = false;
     ++ entry->refcount;
@@ -228,7 +231,10 @@ store_client::store_client(StoreEntry *e) :
 }
 
 store_client::~store_client()
-{}
+{
+    assert(entry);
+    entry->unlock("store_client");
+}
 
 /* copy bytes requested by the client */
 void
@@ -527,7 +533,7 @@ store_client::readBody(const char *, ssize_t len)
 
     if (len > 0 && rep && entry->mem_obj->inmem_lo == 0 && entry->objectLen() <= (int64_t)Config.Store.maxInMemObjSize && Config.onoff.memory_cache_disk) {
         storeGetMemSpace(len);
-        // The above may start to free our object so we need to check again
+        // recheck for the above call may purge entry's data from the memory cache
         if (entry->mem_obj->inmem_lo == 0) {
             /* Copy read data back into memory.
              * copyInto.offset includes headers, which is what mem cache needs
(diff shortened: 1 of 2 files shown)


src/store_client.cc
```diff
@@ -527,7 +533,7 @@ store_client::readBody(const char *, ssize_t len)
 
     if (len > 0 && rep && entry->mem_obj->inmem_lo == 0 && entry->objectLen() <= (int64_t)Config.Store.maxInMemObjSize && Config.onoff.memory_cache_disk) {
         storeGetMemSpace(len);
-        // The above may start to free our object so we need to check again
+        // recheck for the above call may purge entry's data from the memory cache
```
rousskov, line 536: The old comment was rather misleading: It is not possible to _start_ to free a C/C++ object, so the comment was probably referring to cached data rather than the StoreEntry object itself. The current official storeGetMemSpace() code does not really destroy any locked StoreEntries, and the official code below assumes that `entry` object is still alive (or we would not be able to "check again" anything!). Thus, again, the original comment was probably talking about cached data.

The current storeGetMemSpace() implementation does not purge cached data of locked entries either, but there is no guarantee that it would not start doing that in the future, so the second check is (overall) a legitimate idea even now, when we can finally rely on the `entry` being properly locked.

src/store_client.cc
```diff
@@ -717,15 +723,14 @@ storeUnregister(store_client * sc, StoreEntry * e, void *data)
...
 
+    // We must lock to safely dereference e below, after deleting sc and after
+    // calling CheckQuickAbortIsReasonable().
+    e->lock("storeUnregister");
+
     // XXX: We might be inside sc store_client method somewhere up the call
     // stack. TODO: Convert store_client to AsyncJob to make destruction async.
     delete sc;
 
-    assert(e->locked());
```
rousskov, line 724: This problematic assertion was removed instead of moved earlier/higher because we now know that the entry is locked at least up until the `delete sc` call above. Moving this assertion before that call would be pointless/misleading. After that call, we cannot really check anything because `e` may be gone -- that is why we must lock it ourselves before making that delete call. The official code simply assumed that deleting `sc` cannot unlock an entry, and now that assumption is wrong.

We could assert that the lock count is at least 2, but counting locks like that is conceptually wrong, especially since this particular code does not really care whether `e` remains alive after we are done with it (and that is the only thing this problematic assert was enforcing).

yadij (approved):
Thank you.

Accepting, though IMO the proper fix for this is to make `store_client::e` a smart pointer with reference counting instead of yet more manual lock/unlock calls.

rousskov:
> the proper fix for this is to make `store_client::e` a smart pointer with reference counting instead of yet more manual lock/unlock calls.

I agree that we should create and use such a pointer, of course. However, creation of such a type and replacing numerous raw StoreEntry pointers with it is a fairly complex task that requires non-trivial amounts of time (and deserves a dedicated PR). For now, this minimal change just helps clean up workspace for far more important store_client changes that (as you know) are currently in the works. This cleanup will not be in the way of future smart pointer changes (in fact, it may help!).

PR #1282, opened 2023-02-17
Removed unused storeClientCopyPending()

rousskov:
Unused since 2002 commit edce4d9.


kinkie (approved):
There is a reference to storeClientCopyPending in ChangeLog. Might be worthwhile to add a reference to this as well?

rousskov:
> There is a reference to storeClientCopyPending in ChangeLog. Might be worthwhile to add a reference to this as well?

IMO, audiences targeted by contemporary ChangeLog should not be bothered with notices about removal of code unused since 2002, regardless of what that code was doing in 1997. We should not pollute modern ChangeLog just because 1997 ChangeLog said something. ChangeLog entries are meant to be independent, especially when they are 25+ years apart.

PR #1283, opened 2023-02-17
Removed unused StoreIOState::STFNCB callbacks

rousskov:
Unused since before 2016 commit abf396e that discovered the problem.


kinkie (approved):
Unit tests are green, let's ship it

PR #1288, opened 2023-02-23
Maintenance: Change unit test class names to CamelCase

kinkie:
Besides applying Project CamelCase class naming rules to all CppUnit
test class names, these script-driven changes also adjust affected class
name-dependent code, improve consistency across unit test declarations a
little, and rename a few test classes (e.g. testURL is not TestUri). The
corresponding source code filenames are left unchanged (for now).


rousskov:
Thank you for this fix. I support it and will come back to this PR a bit later to fix/improve its scope under the Easier Done Than Said experiment.

For now, I adjusted the PR title to avoid introducing `Refactor:`, a new overly narrow commit category that, IMO, does not actually apply well to this specific PR. We have discussed commit categories several times already. @kinkie, regardless of how you feel about the topic, is there _anything_ I can do to stop you from creating new categories (or whatever you want to call similar title prefixes!) in ordinary PRs like this one? It would save valuable time...

kinkie:
> For now, I adjusted the PR title to avoid introducing `Refactor:`, a new overly narrow commit category that, IMO, does not actually apply well to this specific PR. We have discussed commit categories several times already. @kinkie, regardless of how you feel about the topic, is there _anything_ I can do to stop you from creating new categories (or whatever you want to call similar title prefixes!) in ordinary PRs like this one? It would save valuable time...

Sorry, I was not aware that we had a concept like "commit categories". To me a PR title is just that, a title, and I try to make it descriptive.
I can be more restrictive and stick to Maintenance, I don't really care about how they are named as long as we are generally aware of what they are about

rousskov:
> I don't really care about how they are named as long as we are generally aware of what they are about

* If the assertion that categories exist and should continue to exist is adopted by the Project, then, someday, somebody will document them somewhere, you will stop adding new ones (without discussion), and I will stop begging you to stop doing that. Needless to say, there will still be conflicts because while an author may think that the official `Maintenance` category applies to their specific PR, the reviewer may think otherwise. Those conflicts should be fairly easily resolved using documented category descriptions/rules (if those descriptions/rules are correctly written).

* If the assertion that categories exist and should continue to exist is rejected by the Project, then we all should stop editing PR titles to change or remove a "new" or "misapplied" category -- there will be no categories! Needless to say, there will still be conflicts because while an author may think that `Maintenance:` prefix is a good idea for their specific PR, the reviewer may think otherwise. Those conflicts would have to be resolved on general "good title" principles.

Today, neither precondition has been satisfied, AFAICT: The Project has not decided whether commit categories exist. Conflicts are resolved using primarily hand-waving with references to existing PR titles and undocumented practices. IMO, this is the worst state out of the three described.

For now, I am just trying to minimize the number of title prefixes _added_, so that it is easier to achieve the outcome described in the first bullet.

rousskov:
> I  will come back to this PR a bit later to fix/improve its scope under the Easier Done Than Said experiment.

@kinkie, I pushed three commits with the bulk of the expected changes, but I need a bit more time to finalize this. Commit fabc7c3 may be of special interest to you already because it adjusts the code related to the classes that your earlier commits have touched. To reduce diff noise, you may want to tell git or GitHub to ignore whitespace differences.

rousskov (approved):
After pushing a few more adjustments, I cannot find more in-scope class names to fix and am content with the current PR code changes. I have updated the PR description to match the current PR state.

The only large known change that I left out of scope is renaming the corresponding source files: That work simply requires more time than I have now. IMO, we do not have to rename in this PR because this PR represents reasonably scoped progress, even without the file name changes.

@kinkie, the ball is on your side! All the current PR changes are script-driven. Please let me know if you see more related things to adjust automatically.

src/tests/testACLMaxUserIP.cc
```diff
@@ -19,13 +19,13 @@
 
 #include <stdexcept>
 
-CPPUNIT_TEST_SUITE_REGISTRATION( testACLMaxUserIP );
+CPPUNIT_TEST_SUITE_REGISTRATION( TestACLMaxUserIP );
```
rousskov, line 22: BTW, this and three more test classes still do not obey CamelCase rules, but I did not fix them (beyond first letter capitalization) because these violations are based on the (also wrong) names of the classes this code is testing (e.g., ACLMaxUserIP). If those tested classes are renamed, we will be able to rename these testing classes as well.

```
class TestACLMaxUserIP: public CPPUNIT_NS::TestFixture
class TestDiskIO: public CPPUNIT_NS::TestFixture
class TestSBuf: public CPPUNIT_NS::TestFixture
class TestSBufList: public CPPUNIT_NS::TestFixture
```

yadij:
This "categories" plan is news to me too.  So my first question is why would be want formal categories or reserved prefixes at all?

As a release manager it is useful to have a way to quickly identify which commit titles can be dropped from the ChangeLog. But that does not need formal categories, just consistent practices.

As such practice I have been trying to do consistently for a long time has these simple rules of thumb:
 * if the patch fixes a known bug => use "Bug NNNN:" as title prefix.
   - this because the old web scripts auto-linked the number to bugzilla.
 * if the patch changes a binary/helper/tool/script that is not the 'squid' one  => use the tool/helper name as title prefix. (eg "squidclient:", "cachemgr.cgi:")
   - highlighting these in ChangeLog has slightly increased importance because they do not get a release notes mention.
 * if the patch is one of a series for some large wiki `Features/Foo` project, => use the project name as title prefix. (eg "SSL-Bump:", "HTTP:", "SourceLayout:")
   - this helps with considering general refactor or cleanup patches for backport (or the reverse, eg my "SourceLayout" rarely qualify, but TLS/SSL usually are worthwhile).
 
Contrast that to labeling every commit as "Fix(|ed|es) ..." regardless of whether it was a bug, or a typo, or refactoring, or missing comments/docs, or everything at once.

rousskov:
> This "categories" plan is news to me too. So my first question is why would be want formal categories or reserved prefixes at all?

There is currently no plan (as far as I know). Formal categories are useful for obvious (IMHO) reasons that I will not repeat here [again](https://github.com/squid-cache/squid/pull/984#issuecomment-1048807082) and that prompt us to add or remove PR title prefixes, including in very [recent](https://github.com/squid-cache/squid/pull/830#pullrequestreview-1026773155) PRs.
 
> As a release manager it is useful to have a way to quickly identify which commit titles can be dropped from the ChangeLog. But that does not need formal categories, just consistent practices.

In a collaborative project using an unstable set of contributors with widely varying communication skills and expertise, "consistent practices" should be documented to reduce overheads, especially when the history of the project does not demonstrate desirable consistency levels. Documenting a consistent prefixing practice effectively introduces "formal categories". It does not really matter what we call them though. It is the agreement regarding their existence and associated categorization rules that matters, not the name of the concept.

> As such practice I have been trying to do consistently for a long time has these simple rules of thumb:
> 
> * if the patch fixes a known bug => use "Bug NNNN:" as title prefix.
> * if the patch changes a binary/helper/tool/script that is not the 'squid' one  => use the tool/helper name as title prefix. (eg "squidclient:", "cachemgr.cgi:")
> * if the patch is one of a series for some large wiki `Features/Foo` project, => use the project name as title prefix. (eg "SSL-Bump:", "HTTP:", "SourceLayout:")
 
I agree with (and use myself) the first two categories. They need a bit more explanations to clarify common gray areas, but I hope the Project can formally adopt them (with those clarifications).

I am strongly against adopting the last rule, for many reasons, at least some of which we have discussed before. To save time, I will not go into them here unless you force me to.

The two additional rules I would like to add are:

* Maintenance: The commit is very unlikely to change Squid runtime behavior. This is a very useful category when backporting (or upporting) changes. It covers source code formatting changes, removal of completely unused code, documentation adjustments, build fixes, etc.

* Avoid adding new categories without discussion. Without this rule, we will continue to argue whether a new title prefix is worth using for a specific PR. If we cannot agree on this rule, then, to reduce overheads, we should add the opposite one: "Avoid editing or removing new categories added by other core developers."

Should we open a [site](https://github.com/squid-cache/squid-cache.github.io) PR to finalize and commit these rules?

src/tests/testACLMaxUserIP.cc
kinkie: ACL is an acronym, as is SBuf. I wouldn't obsess over these

kinkie:
> Should we open a [site](https://github.com/squid-cache/squid-cache.github.io) PR to finalize and commit these rules?

I fully support the first two.
I'm not a fan of Maintenance as described: I feel it's very broad and a vast majority of our PRs would fall under that label, making it just marginally more useful as California's prop 65 warnings. It also may not cover things that are maintenance. Of course, just about any label can have the same weakness.

kinkie:
Regarding THIS PR, it's good to go. I hope that the large number of files it touches won't be an issue if it needs backporting

rousskov:
> I hope that the large number of files it touches won't be an issue if it needs backporting

FWIW, if I were to backport this PR, I would start by running the same script against the target branch instead of cherry-picking commit [41e9c9f](https://github.com/squid-cache/squid/commit/41e9c9f078b45aac6ad49fa93020a862bc278b9b). I am happy to assist with that script application.

rousskov:
> > Should we open a [site](https://github.com/squid-cache/squid-cache.github.io) PR to finalize and commit these rules?
 
> I fully support the first two.

Do you think we should open a [site](https://github.com/squid-cache/squid-cache.github.io) PR to finalize and commit PR title prefix rules, including the first two you fully supported and any other rules that make it into the final set? Or are you against making these rules official and documented (in hope to reduce overheads related to their (non)existence and (mis)application)?

src/tests/testACLMaxUserIP.cc
rousskov: > ACL is an acronym, as is SBuf. I wouldn't obsess over these

We do CamelCase most acronyms in modern Squid sources (including "ACL" and excluding "SBuf"). I would not call a desire to explain why a few non-CamelCase test class names were not converted to CamelCase (in a PR that claims to fix _all_ test class names) an obsession, but I suspect you did not mean for your comment to imply that somebody is "obsessed over these" issues.

kinkie: > > ACL is an acronym, as is SBuf. I wouldn't obsess over these
> 
> We do CamelCase most acronyms in modern Squid sources (including "ACL" and excluding "SBuf"). I would not call a desire to explain why a few non-CamelCase test class names were not converted to CamelCase (in a PR that claims to fix _all_ test class names) an obsession, but I suspect you did not mean for your comment to imply that somebody is "obsessed over these" issues.

Of course I am only referring to my own plans (and lack thereof) about these classes.

PR #1293, opened 2023-03-01
Maintenance: remove some unit test headers

kinkie:
When a unit test class is not reused across translation units, merge the
class declaration with its implementation and remove the header file. In
many cases, these classes are not meant/designed to be reusable that
way, and we should not imply otherwise by publishing a header file.

Excuse src/tests/testAuth.h until we make that test work again.


src/Makefile.am
```diff
@@ -808,8 +808,7 @@ tests_testMem_LDFLAGS = $(LIBADD_DL)
```
rousskov, line 808: > PR description: Most unit test classes are trivial, it's not necessary to have a header file and an implementation file for them, becasuse the definition is never reused

The opening assertion -- "most test classes are trivial" -- is correct but irrelevant. It is the last assertion -- "never reused" -- that actually supports your (correct) decision to remove/merge those headers. Also, those test classes are not meant or not designed to be reused across translation units: Some effort is required to make some of them reusable and we should not false-advertise otherwise.

I have adjusted the PR description to remove the irrelevant assertion and add another argument in support of this PR. I also fixed spelling. Please check and adjust further as needed, of course.

src/Makefile.am
```diff
@@ -808,8 +808,7 @@ tests_testMem_LDFLAGS = $(LIBADD_DL)
 
 check_PROGRAMS += tests/testCharacterSet
 tests_testCharacterSet_SOURCES = \
```
rousskov, line 810: I would very much like to see this PR removing _all_ headers that match the (finalized) removal logic. Are you still working on removing more files? If you are, thank you, and I will be out of your hair until you are done. If you are done, then I would like to understand why most of the remaining headers were not removed. If you need assistance with removing move files, please let me know.

FWIW, here are the currently remaining headers that I know about. Using the same removal logic, many of these should be removed. I have not checked whether any of these are reused.

```
compat/testPreCompiler.h
lib/tests/testRFC1738.h
src/tests/testACLMaxUserIP.h
src/tests/testAuth.h
src/tests/testBoilerplate.h
src/tests/testIpAddress.h
src/tests/testLookupTable.h
src/tests/testMem.h
src/tests/testRock.h
src/tests/testSBuf.h
src/tests/testSBufList.h
src/tests/testStatHist.h
src/tests/testStore.h
src/tests/testStoreController.h
src/tests/testStoreHashIndex.h
src/tests/testString.h
src/tests/testTokenizer.h
src/tests/testURL.h
src/tests/testUfs.h
src/tests/testUriScheme.h
src/tests/testYesNoNone.h
```

src/Makefile.am
```diff
@@ -926,8 +923,7 @@ tests_testURL_LDFLAGS = $(LIBADD_DL)
 
 check_PROGRAMS += tests/testDns
 tests_testDns_SOURCES = \
-	tests/testRFC1035.cc \
-	tests/testRFC1035.h
+	tests/testRFC1035.cc
```
rousskov, line 926: Should we fix capitalization of the affected files in _this_ PR? For example, testRFC1035.cc should (at some point), become TestRfc1035.cc (to match the primary TestRfc1035 class in that file). Should that happen in this PR? I am just asking a question, not implying a "yes" answer :-).

src/Makefile.am
```diff
@@ -808,8 +808,7 @@ tests_testMem_LDFLAGS = $(LIBADD_DL)
 
 check_PROGRAMS += tests/testCharacterSet
```
rousskov, line 809: It feels like the no-header should be the default for unit test classes. Thus, I suggest adjusting `src/tests/testBoilerplate.*` accordingly. I do not insist on that adjustment (until there are at least some test classes that legitimately use header files), but adding a comment to the boilerplate class about the need for a header file would be nice.

rousskov (commented):
I very much support removal of unnecessary header files. Thank you.

src/Makefile.am
kinkie: > I would very much like to see this PR removing _all_ headers that match the (finalized) removal logic

That's my plan that's why the PR is in draft.
I'm currently prioritising fixing the portability and build issues I uncovered first, as these block other development. Once that is done, I'll carry this forward

src/Makefile.am
kinkie: > Should we fix capitalization of the affected files in _this_ PR?

That feels like scope creep, I prefer to focus on one thing at a time, so no. A followup PR can take care of that.

src/Makefile.am
kinkie: I agree.

src/Makefile.am
rousskov: Great plan, thank you! No rush.

BTW, when posting a Draft PR, consider adding a GitHub comment to explain why the PR was marked as a draft -- disclosing your reasons would reduce the number of questions others may have to ask and you may have to answer...

kinkie:
Okay, I've now swept over all trivial headers and removed them.
While doing this, I discovered that testAuth is not being built because it is missing the machinery in src/Makefile.am and as a result has rotten. I'm keeping fixing that out of scope for this PR, there is an initial branch to address that at https://github.com/kinkie/squid/tree/unrot-testauth and it will become its own PR.

src/Makefile.am
kinkie: Adding a comment to testBoilerplate.h

rousskov (changes requested):
Thank you very much for these updates! I added a sentence to the PR description to document one exceptional case.

There are some inconsistent and unnecessary whitespace changes in/around some CPPUNIT_TEST_SUITE_REGISTRATION() calls and moved class declaration lines, but I will not ask you to undo them (beyond what is done by our existing formatting scripts) because I assume that you cannot easily script those improvements.

src/Makefile.am
kinkie: done

rousskov:
Thank you for addressing all of my earlier concerns except one:

```
  NOTICE: File src/tests/testEnumIterator.cc changed: by astyle
  NOTICE: File src/tests/testEvent.cc changed: by scripts/maintenance/sort-includes.pl
  NOTICE: File src/tests/testHttpReply.cc changed: by astyle
  NOTICE: File src/tests/testNetDb.cc changed: by astyle
```

If you have a chance, please apply in-scope automated formatting changes. If you do not have time for this, I will find the time to do it.

kinkie:
> Thank you for addressing all of my earlier concerns except one:

> If you have a chance, please apply in-scope automated formatting changes. If you do not have time for this, I will find the time to do it.

No worries, it takes just a minute. done

rousskov (approved):

PR #1299, opened 2023-03-04
Extend RegexPattern API to handle SBuf 

yadij:


kinkie (approved):

src/acl/RegexData.cc
```diff
@@ -42,11 +42,13 @@ ACLRegexData::lineOptions()
...
 
 bool
-ACLRegexData::match(char const *word)
+ACLRegexData::match(char const *aWord)
 {
-    if (!word)
+    if (!aWord)
         return 0;
 
+    const SBuf word(aWord);
```
rousskov, line 50: It is not possible to simply extend regcomp()-based RegexPattern (and related classes) to support SBuf without secret and expensive to-SBuf conversions like this one. Please refactor to avoid these conversions after #1232 is merged.

rousskov (changes requested):

src/acl/RegexData.cc
yadij: The problem you noticed has nothing to do with the internals of `RegexPattern` and will not be solved by anything in PR #1232.

It is the natural side-effect of config parser and ACL implementations still using `char *` instead of `SBuf`.

rousskov: > The problem you noticed has nothing to do with the internals of `RegexPattern` and will not be solved by anything in PR #1232.

I disagree on both counts.

PR #1300, opened 2023-03-04
Fix several broken or missing copyright statements

yadij:


kinkie (approved):

test-suite/squidconf/bug4832.conf
```diff
@@ -1,2 +1,9 @@
+## Copyright (C) 1996-2023 The Squid Software Foundation and contributors
```
rousskov, line 1: I believe we should remove copyright blobs from all test configuration files that have them (instead of adding them to those that do not): These copyright statements are incompatible with some test configurations, negatively interfere with others, and are unnecessary in these test files (most, if not all, do not contain copyrightable material).

rousskov, line 1: Since this PR is fixing broken test-sutie/squidconf/*.conf copyright blobs, please fix the blob in test-suite/squidconf/mgr_passwd.conf (which has an inconsistent comment style).

yadij: done

PR #1301, opened 2023-03-04
bootstrap.sh: Polish output

yadij:
We use bootstrap.sh a lot in maintenance scripts but do not
need to see its output on success.

 * Fix some incorrect uses of stderr and stdout such that
   all (and only) errors appear on stderr

 * Polish the message display when an error does occur
    detecting one of the tools.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -50,7 +50,8 @@ find_variant()
   done
   if [ "x$found" = "xNOT_FOUND" ]; then
     echo "WARNING: Cannot find $tool version $versions" >&2
-    echo "Trying `$tool --version | head -1`" >&2
+    echo -n "Trying " >&2
+    ($tool --version | head -1) >&2
     found=""
   fi
   echo $found
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -50,7 +50,7 @@ find_variant()
   done
   if [ "x$found" = "xNOT_FOUND" ]; then
     echo "WARNING: Cannot find $tool version $versions" >&2
-    ($tool --version | head -1) >&2
+    echo "Trying `$tool --version 2>&1 | head -1`" >&2
     found=""
   fi
   echo $found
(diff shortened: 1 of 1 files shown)


kinkie (approved):
I don't see a lot of benefit but I don't see any harm either. Ship it

yadij:
> I don't see a lot of benefit but I don't see any harm either. Ship it

The benefit is less text in those NOC email reports without loosing error messages.

rousskov (dismissed):
> We use bootstrap.sh a lot in maintenance scripts but do not need to see its output on success. This option allows execution without the verbose descriptions.

I do not think adding a command line option is the best way to address the problem of unwanted output in successful executions from maintenance scripts. The script already redirects error output to stderr in some cases, and I would recommend improving the remaining warning/error messages to use that (correct) approach[^1]. After that, the caller can use `./bootstrap.sh > /dev/null` to achieve the desired effect.

If you insist on adding a command line option, then I would have to insist that the option is added without sprinkling the code with `test` statements (add a "verbose logging" function like echo_debug() instead) and without silently ignoring unsupported/mistyped command line options (they should become a fatal error instead).

[^1]: Adding a reusable function for that is a good idea, of course, but I would not insist on that additional improvement.

bootstrap.sh
```diff
@@ -50,7 +50,8 @@ find_variant()
   done
   if [ "x$found" = "xNOT_FOUND" ]; then
     echo "WARNING: Cannot find $tool version $versions" >&2
-    echo "Trying `$tool --version | head -1`" >&2
+    echo -n "Trying " >&2
+    ($tool --version | head -1) >&2
```
kinkie, line 54: If we go to the model that stdout can be ignored, while stderr should report error conditions, then these should go to stdout, shouldn't they?

Also, please be aware that not all OSes support "echo -n" (MacOS doesn't, for instance)

kinkie: > This is the report of what went wrong when calling the tools `--version` parameter.
> 
> Thanks for the reminder about `-n` on MacOS. I guess we can do without that line entirely.

A pattern that could work on macos is:
toolversion=`$tool --version | head -1`
```suggestion
echo "Trying $toolversion" >&2 
```

rousskov (approved):

PR #1302, opened 2023-03-05
Reduce stale errno usage

rousskov:
This covers a few easy cases, one of which resulted in wasted triage
time. More changes are needed to cover all potentially stale errno uses.

### DIFF
--- a/src/comm.cc
+++ b/src/comm.cc
@@ -385,6 +385,7 @@ comm_openex(int sock_type,
         AI->ai_protocol = proto;
         debugs(50, 3, "Attempt fallback open socket for: " << addr );
         new_socket = socket(AI->ai_family, AI->ai_socktype, AI->ai_protocol);
+        xerrno = errno;
         debugs(50, 2, "attempt open " << note << " socket on: " << addr);
     }
 
@@ -393,7 +394,7 @@ comm_openex(int sock_type,
          * are failing because the open file table is full.  This
          * limits the number of simultaneous clients */
 
-        if (limitError(errno)) {
+        if (limitError(xerrno)) {
             debugs(50, DBG_IMPORTANT, MYNAME << "socket failure: " << xstrerr(xerrno));
             fdAdjustReserved();
         } else {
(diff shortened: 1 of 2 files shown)


src/comm.cc
```diff
@@ -884,11 +885,11 @@ _comm_close(int fd, char const *file, int line)
     // notify read/write handlers after canceling select reservations, if any
     if (COMMIO_FD_WRITECB(fd)->active()) {
         Comm::SetSelect(fd, COMM_SELECT_WRITE, nullptr, nullptr, 0);
-        COMMIO_FD_WRITECB(fd)->finish(Comm::ERR_CLOSING, errno);
+        COMMIO_FD_WRITECB(fd)->finish(Comm::ERR_CLOSING, 0);
     }
     if (COMMIO_FD_READCB(fd)->active()) {
         Comm::SetSelect(fd, COMM_SELECT_READ, nullptr, nullptr, 0);
-        COMMIO_FD_READCB(fd)->finish(Comm::ERR_CLOSING, errno);
+        COMMIO_FD_READCB(fd)->finish(Comm::ERR_CLOSING, 0);
```
rousskov, line 892: This stale `errno` usage sent me on wild goose chase because it showed up in a (partial) cache.log as EINVAL, which sometimes indicates a Squid bug.

src/comm.cc
```diff
@@ -385,6 +385,7 @@ comm_openex(int sock_type,
         AI->ai_protocol = proto;
         debugs(50, 3, "Attempt fallback open socket for: " << addr );
         new_socket = socket(AI->ai_family, AI->ai_socktype, AI->ai_protocol);
+        xerrno = errno;
```
rousskov, line 388: I am not 100% sure that this code is meant to preserve the _last_ error when trying a second socket() system call after the first failure. If you are sure that this is wrong, please let me know what the replacement C++ comment should say:

```suggestion
        // keep xerrno unchanged because ...XXX...
```

rousskov (commented):
I saw a few other potentially stale `errno` uses, but was not comfortable adjusting that code without testing/investigating, and I do not have time for that right now.

Disclaimer: I have not tested any of these changes.

src/comm.cc
yadij: IIRC this is preserving the *first* errno, so the primary socket failure can be fixed, not just the fallback type.

That said we have had several cycles of fixing errno usage, so we should now check whether the callers of this function actually need it preserved. Perhapse this kludge can all go away.

yadij (changes requested):

src/comm.cc
rousskov: > IIRC this is preserving the _first_ errno, so the primary socket failure can be fixed, not just the fallback type.

I changed the code to preserve the first errno in commit 5dfa03d.

While working on that change, I discovered another bug in this code. There is no trivial or in-scope fix, so I added an XXX (while preserving the current broken official code behavior).
 
> That said we have had several cycles of fixing errno usage, so we should now check whether the callers of this function actually need it preserved. Perhapse this kludge can all go away.

Unfortunately, there are still many (indirect) comm_openex() callers that use `errno`.

Comm should stop setting `errno` for callers (and stop returning invalid descriptor numbers on errors), but that requires a lot of changes outside this PR scope, of course. It can probably be done incrementally, but the first step needs to set a proper framework up. This small PR cannot do that.

yadij (approved):

rousskov:
@kinkie, @yadij, please add these changes to v6. I continue to waste time on chasing wrong v6 bugs because of stale errno usage fixed by this PR. Thank you.

squidadm:
queued for backport to v6

PR #1304, opened 2023-03-10
Replaced clientReplyContext::tempBuffer with old_reqofs

rousskov:
The tempBuffer data member was not actually used as a buffer. We only
used its offset field, and only for saving reqofs (which has a different
type than tempBuffer.offset!). Replaced the buffer with old_reqofs,
consistent with the rest of the "saved stale entry state" code.

Also fixed old_reqsize type to match reqsize and grouped that member
with the other private "saved stale entry state" fields.

Bad old types probably did not trigger runtime failures because the
associated saved numbers are saved at the very beginning of fetching the
entry, when all these accumulation-related counters are still small.

The remaining reqofs and reqsize types are wrong for platforms where
size_t is not uint64_t, but fixing that deserves a dedicated change. For
now, we just made the types of "old_" and "current" members consistent.

### DIFF
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -66,7 +66,6 @@ clientReplyContext::~clientReplyContext()
     /* old_entry might still be set if we didn't yet get the reply
      * code in HandleIMSReply() */
     removeStoreReference(&old_sc, &old_entry);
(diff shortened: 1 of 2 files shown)


kinkie (approved):
Better than before

PR #1305, opened 2023-03-11
Remove some duplicate typedef and forward declarations

yadij:


kinkie (approved):

src/clients/forward.h
```diff
@@ -9,15 +9,12 @@
 #ifndef SQUID_CLIENTS_FORWARD_H
 #define SQUID_CLIENTS_FORWARD_H
 
+#include "base/forward.h"
```
rousskov, line 12: To minimize dependencies among headers, `X.h` should not include `Y.h` when `X.h` does not need `Y.h` to be compiled. In Squid `src/` context, that "to be compiled" phrase means "to be compiled in a hypothetical translation unit that includes `squid.h`, `X.h`, and no other header or code".

Applying the above rule to this PR means that `clients/forward.h` should not add a dependency on `base/forward.h`:

```suggestion
```

Needless to say, other source code adjustments may be necessary to properly remove AsyncJobPointer typedef from `clients/forward.h`. The suggestion in this change request is just the first step and may, in isolation, break the build.

P.S. There is a weaker rule of thumb (based on the above minimization rule) that excepts situations where X cannot be compiled without Y in any _realistic_ translation unit, but even that weaker version cannot justify adding `base/forward.h` to `clients/forward.h`.

src/clients/forward.h
```diff
@@ -9,15 +9,12 @@
...
 #include "sbuf/forward.h"
 
 class FwdState;
 class HttpRequest;
 
-class AsyncJob;
-template <class Cbc> class CbcPointer;
-typedef CbcPointer<AsyncJob> AsyncJobPointer;
```
rousskov, line 19: > PR description: Remove some duplicate pre-defines

I adjusted the PR title to use more accurate C++ terms for "pre-defines": "typedef and forward declarations". I do not insist on the PR title change. Please feel free to revert or adjust as needed.

rousskov (changes requested):
I support the removal, but it should not be done at the expense of adding a `base/forward.h` dependency to `clients/forward.h`, especially since you have started to police indirect duplicates in header inclusion. Please adjust those source files that actually need `base/forward.h` instead. Please see one of the change requests for details and discussion.

rousskov (approved):

PR #1306, opened 2023-03-12
Add libhtml.la for HTML syntax tools

yadij:


src/html/forward.h
```diff
@@ -6,16 +6,16 @@
...
-#else
-extern
-#endif
+/// HTML Document tools
+namespace Html
+{
+// nothing yet
+} // namespace Html
 
 char *html_quote(const char *);
```
rousskov, line 18: As we have discussed [before](https://github.com/squid-cache/squid/pull/1001#discussion_r833316885), forward.h files have a specific purpose/scope: Please restrict new forward.h files to [forward declarations](https://en.cppreference.com/w/cpp/language/class#Forward_declaration) and equivalent incomplete type constructs. If there are no such constructs in a PR, then no forward.h file should be added in that PR.

The html_quote() function declaration can be declared in src/html/quoting.h (if we agreed not to capitalize declaration collections like this to distinguish them from class-focused sources) or src/html/Quoting.h (otherwise).

rousskov (changes requested):
Thank you for reducing lib/ (and C) footprint.

src/html/forward.h
yadij: > Please restrict new forward.h files to [forward declarations](https://en.cppreference.com/w/cpp/language/class#Forward_declaration) and equivalent incomplete type constructs.

Agreed. And it does ... contain the incomplete forward-declaration of a `namespace Html`. With documentation of what that namespace should contain.

> If there are no such constructs in a PR, then no forward.h file should be added in that PR.

Above if-condition being false. Ergo there **should** be a `forward.h` in this PR.

> The html_quote() function declaration can be declared in ...

... or being a single function (not a collection of symbols) it could be declared as a forward-declaration in `forward.h` to reduce `#include`s.

I see no reason to block this PR any further. Please unblock.

rousskov: > it does ... contain the incomplete forward-declaration of a `namespace Html`.

In C++, namespaces cannot be forward-declared.


> > The html_quote() function declaration can be declared in ...
 
> ... or being a single function (not a collection of symbols) it could be declared as a forward-declaration in `forward.h`

In C++, functions cannot be [forward-declared](https://en.cppreference.com/w/cpp/language/class#Forward_declaration). Functions can be fully [_declared_](https://en.cppreference.com/w/cpp/language/function), of course, but `forward.h` is not meant for (function) declarations.

These `forward.h` files have a very specific role/niche -- helping other header files to declare things that use (incompletely declared) names without pulling in the (often heavy) stuff that comes with complete/full declarations of those names.

yadij: Conceding this argument in hopes to make progress.

src/html/Quoting.h
```diff
@@ -6,16 +6,16 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-#ifndef _SQUID_HTML_QUOTE_H
-#define _SQUID_HTML_QUOTE_H
+#ifndef SQUID__SRC_HTML_QUOTING_H
```
kinkie, line 9: why the double underscore in the header guard?

kinkie (approved):

rousskov (changes requested):
Thank you for addressing my earlier concern. I only have one request left, and it is trivial to address. I left specific replacement suggestions.

----

> PR [commit](https://github.com/squid-cache/squid/pull/1306/commits/00a60c6b5d8cad3b37aa7461eb09380231bfd2e9) title: Reviewer demanded changes

FWIW, phrasing like this does not encourage cooperation. "Reviewer requested changes" is a better, "neutral" phrase that cannot be reasonably interpreted as an insult and does not require extra work. Accurately summarizing the actual commit change(s) sometimes help, but it is not required for intermediate PR commits, of course.

src/html/Quoting.h
rousskov: > FYI, the use of two underscores helps reduce the complexity of differentiating these macros from the many other `SQUID_` macros around Squid.

If that is the goal, then we should find a better way to do that IMO.

src/html/Quoting.h
```diff
@@ -6,16 +6,11 @@
...
+#define SQUID__SRC_HTML_QUOTING_H
 
-#ifdef __cplusplus
-extern "C"
-#else
-extern
-#endif
```
rousskov, line 12:
```suggestion
```

rousskov (commented):
LGTM, but if you can fix formatting, please do so.

kinkie (approved):
thanks!

PR #1310, opened 2023-03-17
Fix rock/RockSwapDir.cc "slot->sameKey()" assertion

rousskov:
Introduced in 2014 commit fa2301d, the assertion was too strict: Due to
store_client idiosyncrasies, openStoreIO() may be called long after the
initial cache hit is detected and locked for reading. It is possible
that the StoreEntry object has gone private since that detection and
locking. Private StoreEntry objects lose their public keys. The existing
anchor lock assures that we still point to the right disk entry, but we
cannot check that invariant using the now-private StoreEntry key.

There are no known sightings of this assertion in the wild, but we can
now reproduce it using high-load Web Polygraph tests. It is not clear
whether some other Squid changes made the triggering sequence of events
a bit more likely, or we were just lucky to discover the right workload.

### DIFF
--- a/src/fs/rock/RockSwapDir.cc
+++ b/src/fs/rock/RockSwapDir.cc
@@ -795,7 +795,12 @@ Rock::SwapDir::openStoreIO(StoreEntry &e, StoreIOState::STIOCB * const cbIo, voi
     // public key, but it could have gone private since then (while keeping the
     // anchor lock). The stale anchor key is not (and cannot be) erased (until
     // the marked-for-deletion/release anchor/entry is unlocked is recycled).
-    assert(EBIT_TEST(e.flags, KEY_PRIVATE) || slot->sameKey(static_cast<const cache_key*>(e.key)));
+    const auto ourAnchor = [&]() {
+        if (const auto publicKey = e.publicKey())
+            return slot->sameKey(publicKey);
+        return true; // cannot check
+    };
+    assert(ourAnchor());
 
     // For collapsed disk hits: e.swap_file_sz and slot->basics.swap_file_sz
     // may still be zero and basics.swap_file_sz may grow.
--- a/src/fs/rock/RockSwapDir.cc
+++ b/src/fs/rock/RockSwapDir.cc
@@ -791,7 +791,12 @@ Rock::SwapDir::openStoreIO(StoreEntry &e, StoreIOState::STIOCB * const cbIo, voi
            std::setfill('0') << std::hex << std::uppercase << std::setw(8) <<
            sio->swap_filen);
 
-    assert(slot->sameKey(static_cast<const cache_key*>(e.key)));
+    // When StoreEntry::swap_filen for e was set by our anchorEntry(), e had a
+    // public key, but it could have gone private since then (while keeping the
+    // anchor lock). The stale anchor key is not (and cannot be) erased (until
+    // the marked-for-deletion/release anchor/entry is unlocked is recycled).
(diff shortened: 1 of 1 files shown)


src/fs/rock/RockSwapDir.cc
```diff
@@ -791,7 +791,17 @@ Rock::SwapDir::openStoreIO(StoreEntry &e, StoreIOState::STIOCB * const cbIo, voi
...
+    // When StoreEntry::swap_filen for e was set by our anchorEntry(), e had a
+    // public key, but it could have gone private since then (while keeping the
+    // anchor lock). The stale anchor key is not (and cannot be) erased (until
+    // the marked-for-deletion/release anchor/entry is unlocked is recycled).
+    const auto ourAnchor = [&]() {
+        if (const auto publicKey = e.publicKey())
+            return slot->sameKey(publicKey);
+        return true; // cannot check
+    };
+    assert(ourAnchor());
```
rousskov, line 803:
```suggestion
    assert(EBIT_TEST(e.flags, KEY_PRIVATE) || slot->sameKey(static_cast<const cache_key*>(e.key)));
```

We can come back to raw StoreEntry::key access and casting employed by the official code (as shown in the suggestion above), but it is best to use publicKey() API instead of that dangerous legacy code. It is possible that the bug fixed by this PR would not exist if the original assertion were written using that publicKey() API -- we could have realized that the (public) key may be gone in this context -- but publicKey() did not exist at commit fa2301d time (it was added four years later in commit 4310f8b).

Using publicKey() while still honoring assert()'s "do nothing in optimized NDEBUG builds" principle (and not computing the public key twice), requires a lambda AFAICT, but it is a fairly simple one.

yadij: > Using publicKey() while still honoring assert()'s "do nothing in optimized NDEBUG builds" principle (and not computing the public key twice), requires a lambda AFAICT, but it is a fairly simple one.

Er, FYI `NDEBUG` is just one of multiple problems with optimizing vs assert().
The principle is a simpler: "assert must not have side effects".

yadij: I do not see any side effects of calling publicKey(), nor sameKey() - and both are const methods.

If you are worried about the lambda or hidden side-effects of computing the publicKey, then a protective if-statement that optimizes away may be more suitable.
```suggestion
    if (const auto publicKey = e.publicKey())
        assert(slot->sameKey(publicKey));
```

yadij (approved):
I'm okay with this whether you take my suggestion or not.

src/fs/rock/RockSwapDir.cc
rousskov: >  The principle is a simpler: "assert must not have side effects".

There are many principles associated with assert(). At the _minimum_, like you said, the code inside assert() should not have side effects. Where feasible, we should also aim for the "do-nothing in optimized NDEBUG builds" principle.

Technically, a call to publicKey() in the proposed `if` guard outside assert() might violate both principles: In the current code, it does _not_, but there is no guarantee that publicKey() does not gain side effects or stops being inlined in the future, unbeknownst to this caller. We should prefer code that does not rely on the current implementation of an external (to that code) API.

FWIW, I suspect there is a better (and possibly even a more general) solution here than the merged lambda. I hope we discover it some day (but it is not a big deal if we do not).

P.S. Sorry I could not respond sooner due to travel.

DrDaveD:
The description here says that there are no known sightings of the assertion in the wild, but there is one reported now in [bug 5531](https://bugs.squid-cache.org/show_bug.cgi?id=5531#c7).

PR #1314, opened 2023-03-26
Maintenance: Use C++ stdlib headers

kinkie:
In C++ files, use C++ standard library headers instead of their
(guarded) C equivalents (e.g., use `<cstdlib>` instead of `<stdlib.h>`).


src/acl/external/session/ext_session_acl.cc
```diff
@@ -46,9 +46,7 @@
 #if HAVE_TDB_H
 #include <tdb.h>
 #endif
-#if HAVE_UNISTD_H
-#include <unistd.h>
-#endif
+#include <cunistd>
```
yadij, line 49: These converted headers should be moved up top with the other c++11 headers.
Or maybe better a followup to make `sort-includes.pl` to enforce the squid guidelines header listing better.

scripts/c-to-cpp-includes.pl
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/perl
```
yadij, line 1: Please re-work this as a script in `scripts/maintenance/` to be run by the maintenance automation.
Taking the name+path of the file being processed as `$1` (to detect which ones to skip), its content on `stdin` and send corrected (or not) output to `stdout`.

yadij (changes requested):

scripts/c-to-cpp-includes.pl
```diff
@@ -0,0 +1,48 @@
...
+math
+signal
+stdarg
+stddef
+stdint
+stdio
+stdlib
+string
+time
+unistd
```
yadij, line 24: There is no such header as `cunistd`
```suggestion
```

scripts/c-to-cpp-includes.pl
```diff
@@ -0,0 +1,48 @@
...
+    @files=`git grep -lF $headerguard -- *.cc *.h *.cci| grep -v -e ^compat`;
+    chomp(@files);
+    foreach $file (@files) { 
+        print "file: $file\n";
+        $pre="";
+        open(FILE,"<", $file);
+        read(FILE, $pre ,1000000) || die("read error on file $file: $!");
+        close(FILE) || die("close error on file $file: $!");
+        $post = $pre;
+        $post =~ s/#if(def)? $headerguard\n#include <$header.h>(.*)\n#endif/#include <c$header>$2/m;
```
rousskov, line 39: It will probably take us a while to agree what this script should do and then adjust it to do it well. I suggest dropping this script from the PR and just merging the (polished) changes.

If we want to flag policy violations (after applying these changes), we can add a (much simpler!) check that will _detect_ them. It would be easy for the developer to fix an occasional violation manually. Those occasional simple fixes are not worth the automation overheads in this case IMO (beyond using a script for creation of the initial set of changes, of course).

rousskov (changes requested):
I support the proposed transition to C++ headers that are guaranteed to exist by language standard. Thank you for doing the necessary legwork to make that happen.

scripts/c-to-cpp-includes.pl
rousskov, line 24: I found a few more entries to add to the list: fenv, float, locale, setjmp, uchar, wchar, and wctype. Please note that it does not matter whether some particular Squid code snapshot actually uses them (all but one of these newly found ones are unused in recent master/v7 code): This script should "consider" _all_ possible C library header wrappers (that were not deprecated in C++20, as detailed below), so that we do not worry about script application timing.

AFAICT, those deprecated C library header wrappers correspond to the following tokens: stdbool, complex, tgmath, iso646, and stdalign. The script should _not_ replace the corresponding C headers with deprecated C++ headers IMO.

Disclaimer: I did not check whether some of these headers need special treatment that would preclude their automatic handling. I hope there are none, but if there are, then I would list them in a nearby script comment for completeness/clarity sake.

src/acl/external/session/ext_session_acl.cc
rousskov: > These converted headers should be moved up top with the other c++11 headers.
>  Or maybe better a followup to make sort-includes.pl to enforce the squid guidelines header listing better.

FWIW, I strongly recommend the second alternative -- it would make both changes a lot easier to review correctly, both during scripts development and during official review.

scripts/c-to-cpp-includes.pl
kinkie: Thanks for the additional digging. I agree with you, adding the extra
files to consider and mentioning the ones we won't in a comment

scripts/c-to-cpp-includes.pl
kinkie: That is actually the reason why I put it in scripts/ rather than in maintenance/; I expect this to be a one-off and that regular code reviews will prevent creep, while preserving the work done to write the automation

scripts/c-to-cpp-includes.pl
kinkie: Removed. I'll have to undo and redo the changes in the codebase

scripts/c-to-cpp-includes.pl
kinkie: See other comment about expecting this to be a one-shot. Especially if any discussion is going to happen about the script itself.

scripts/c-to-cpp-includes.pl
rousskov: > I expect this to be a one-off 

Please do not add one-off scripts to official repositories, especially one-off scripts that we would, in that case, need to spend time on (to agree on  the specifics and to polish). Official repositories are for scripts we intend to be reused quite a bit. This is not the only admission criteria, of course, but it is a critical one.

kinkie: > > I expect this to be a one-off
> 
> Please do not add one-off scripts to official repositories, especially one-off scripts that we would, in that case, need to spend time on (to agree on the specifics and to polish). Official repositories are for scripts we intend to be reused quite a bit. This is not the only admission criteria, of course, but it is a critical one.

I disagree on the philosophy, but the point is moot because the output of the script is the relevant part here so I will not discuss it

kinkie:
The script itself will be removed before merging, as discussed

scripts/c-to-cpp-includes.pl
yadij: IMO this should not be a one-off. Maintaining these changes is extra work for reviewers. Which makes no sense when we already have this script that can lift the workload "good enough".

rousskov:
I adjusted PR title/description to use a more standard terminology/capitalization, to avoid unusual escape sequences, and to remove [obsolete](https://github.com/squid-cache/squid/pull/1314#issuecomment-1510723343) text.

kinkie:
> AFAICT, the following legacy headers were left behind in C++ files

Thanks for noticing. These should be covered and I've double checked. Hopefully nothing was left behind

rousskov:
I found three more cases:

* compat/shm.h: I think we should assume this will remain a C++ only file (and include cerrno unconditionally) because it is backed by a C++ source file and, when compiled with a C++ compiler, it defines a function with C++ linkage (shm_portable_segment_name_is_path()). This is the only case I (barely) insist on. It shows what should happen with all the other wrappers as we get rid of legacy C code.
* compat/cpu.h: We should either assume this will remain a C++-only file and include `cerrno` unconditionally OR use the `#if defined(__cplusplus)` trick from `compat/compat_shared.h` to handle `errno.h`. My weak recommendation is the former (because we should eventually get rid of all C files and because it is less work/complexity/ink).
* compat/xis.h: Use the `#if defined(__cplusplus)` trick from `compat/compat_shared.h` to handle `ctype.h`. This file already has the necessary `#if` nearby, so let's use it.

FWIW, I am done looking through sources and do not anticipate new change requests.

kinkie:
> * compat/shm.h: I think we should assume this will remain a C++-only file

Sure, I can do that.
Unlike compat/compat_shared.h however I'm considering adding an explicit  #error clause making the requirement explicit, to clarify intent versus an effectively empty header file

rousskov:
> > * compat/shm.h: I think we should assume this will remain a C++-only file
 
> I'm considering adding an explicit #error clause making the requirement explicit, to clarify intent versus an effectively empty header file

I am not excited about marking our headers as "C++ only":

1. These special markings would have to go into an increasing number of header files. That is the opposite of what we want for special markings (in general). It will also increase tension during review when folks forget to add them (or disagree whether they should be added).
2. When the last C source file is gone, all those markings would have to be removed because they are no longer needed. That is extra work/noise/etc.

In this case, I would rather mark nothing and rely on the compiler to tell me when things went wrong:

> versus an effectively empty header file

* If that "empty" result _compiles_, it is not a big deal -- somebody probably included a C++ header into a C program, but that C program did not need that header file. :yawning_face:

* If that result triggers a compiler error, the developer will be forced to fix the problem they have created before it affects others. Yes, it is possible to come up with exceptional cases where that failure will be delayed, forcing _another_ developer to deal with it, but I would rather risk that than add/remove special markings to/from all C++-only header files in `compat/` (or equivalent).

However, if you insist, I would not object to adding these special markings, as long as they are added _consistently_, following an agreed-in-advance rule for adding them.

compat/xis.h
```diff
@@ -9,11 +9,9 @@
 #ifndef _SQUID_COMPAT_XIS_H
 #define _SQUID_COMPAT_XIS_H
 
-#if HAVE_CTYPE_H
-#include <ctype.h>
-#endif
-
 #if __cplusplus
```
rousskov, line 12: The lines removed above should be _moved_ to the `#else` clause below, right?

kinkie:
On Tue, Apr 18, 2023 at 10:02 PM Alex Rousskov ***@***.***>
wrote:

>
>    - compat/shm.h: I think we should assume this will remain a C++-only
>    file
>
> I'm considering adding an explicit #error clause making the requirement
> explicit, to clarify intent versus an effectively empty header file
>
> I am not excited about marking our headers as "C++ only":
>
>    1. These special markings would have to go into an increasing number
>    of header files. That is the opposite of what we want for special markings
>    (in general). It will also increase tension during review when folks forget
>    to add them (or disagree whether they should be added).
>    2. When the last C source file is gone, all those markings would have
>    to be removed because they are no longer needed. That is extra
>    work/noise/etc.
>
> In this case, I would rather mark nothing and rely on the compiler to tell
> me when things went wrong:
>
> versus an effectively empty header file
>
>
>    -
>
>    If that "empty" result *compiles*, it is not a big deal -- somebody
>    probably included a C++ header into a C program, but that C program did not
>    need that header file. 🥱
>
> There is a problem with this: compat/ headers are often included from each
other (I didn't specify these ones explicitly, but if they aren't, there is
no harm done).


>
>    -
>
>    If that result triggers a compiler error, the developer will be forced
>    to fix the problem they have created before it affects others. Yes, it is
>    possible to come up with exceptional cases where that failure will be
>    delayed, forcing *another* developer to deal with it, but I would
>    rather risk that than add/remove special markings to/from all C++-only
>    header files in compat/ (or equivalent).
>
>
I'm following the philosophy of "fail early, fail spectacularly"

> However, if you insist, I would not object to adding these special
> markings, as long as they are added *consistently*, following an
> agreed-in-advance rule for adding them.
>

It's a corner case; I wouldn't add yet another rule on this. I'd rather the
rule be "if it's in compat/, let's use C headers until we have C files".

-- 
    Francesco

compat/cpu.h
```diff
@@ -9,9 +9,12 @@
 #ifndef SQUID_COMPAT_CPU_H
 #define SQUID_COMPAT_CPU_H
 
-#if HAVE_ERRNO_H
-#include <errno.h> /* for ENOTSUP */
+#ifndef __cplusplus
+#error "This header is only meant to be used from c++ source files"
 #endif
+
+#include <cerrno>
```
rousskov, line 16: Can you give me an example where the `#error` would affect the compilation outcome? I would expect the C compiler to fail on `#include <cerrno>`. If it fails, then I do not think we should bother with  such `#error` clauses. There is just not enough gain from them in that case...

> At https://github.com/squid-cache/squid/pull/1314#issuecomment-1513733817: "if it's in compat/, let's use C headers until we have C files".

IMO, to be consistent, we would then need to prohibit C++ code from `compat/`, which is doable but would make things worse overall (and move us in the opposite direction of the "no C files" goal).

Still, I am not vetoing this idea if you really like it and can come up with a plausible example to address the concern at the beginning of this change request.

rousskov (changes requested):

compat/cpu.h
kinkie: That is a fair point. Removing this

compat/xis.h
rousskov: @kinkie, it looks like this small change request was overlooked when requesting my review. FWIW, I intend to approve this PR once this change request is resolved and the Perl script is [dropped](https://github.com/squid-cache/squid/pull/1314/#r1148711820).

rousskov (changes requested):

compat/xis.h
kinkie: both done. Thanks!

compat/xis.h
```diff
@@ -9,11 +9,8 @@
...
```
rousskov, line 12: The removal of the above lines does not look right to me. This file provides wrappers to standard isfoo() functions. It should include the header that declares those functions because all legitimate xis.h users need those declarations. This change goes against the header inclusion principles that we should formalize and adopt IMO.

It looks like the compiler will force you to fix this, but I wanted to add this comment to emphasize that we should include that standard header (once in each `#if` clause below) not (just) because the compiler forces us to do that.

rousskov (changes requested):

yadij (approved):
Some of the converted lists could still do with sorting by the maintenance script before merge, and @rousskov review issue needs to be fixed.

Other than those I am fine with this now.

compat/xis.h
```diff
@@ -14,6 +14,7 @@
 #endif
```
rousskov, line 14: AFAICT, we should include `<ctype>` in the C++ section below _and_ move these three lines inside the C-only section below. I know that we have been talking about these two basic changes for a while now, but I cannot tell why we cannot arrive at the final result that has `<ctype>` for C++ and these three lines for C.

kinkie: I had test-built it and it worked, so it might be that it's indirectly included. This said, it doesn't hurt and it clarifies intent so why not.
Test-building and adding it.

rousskov: > I had test-built it and it worked

FWIW, the latest published/reviewed version (i.e. commit [63d499c](https://github.com/squid-cache/squid/pull/1314/commits/63d499c8271e231b440af014bd8054fb2ef516cb)) worked because the file _is_ included (once, before C/C++ checks, just like it is included in the official code).

rousskov (approved):

PR #1318, opened 2023-04-01
Fix GCC v13 build [-Warray-bounds]

rousskov:
src/ipc/mem/FlexibleArray.h:34:52: error: array subscript -1 is
    below array bounds of 'int [1]' [-Werror=array-bounds]

We suspect this warning is a GCC v13 regression bug because the callers
marked as problematic by GCC (e.g., Rock::LoadingEntry::LoadingEntry) do
not use "array subscript -1", and the Ipc::StoreMapItems::at() operator
they use even asserts that the subscript is not negative. It might be
GCC bug 107699: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107699

This change replaces the fake one-item array hack with a properly
aligned byte (still used as the "start of the real array" marker).

Also removed some unused and problematic code (instead of polishing it).

### DIFF
--- a/src/ipc/mem/FlexibleArray.h
+++ b/src/ipc/mem/FlexibleArray.h
@@ -31,7 +31,6 @@ public:
     }
 
     Item &operator [](const int idx) { return *(raw() + idx); }
-    const Item &operator [](const int idx) const { return *(raw() + idx); }
 
     Item *raw() { return reinterpret_cast<Item*>(&raw_); }
 
--- a/src/ipc/mem/FlexibleArray.h
+++ b/src/ipc/mem/FlexibleArray.h
@@ -32,10 +32,10 @@ public:
 
     Item &operator [](const int idx) { return *(raw() + idx); }
 
-    Item *raw() { return reinterpret_cast<Item*>(&raw_); }
+    Item *raw() { return reinterpret_cast<Item*>(&start_); }
 
 private:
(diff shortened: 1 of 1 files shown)


src/ipc/mem/FlexibleArray.h
```diff
@@ -27,20 +27,15 @@ class FlexibleArray
 {
 public:
     explicit FlexibleArray(const int capacity) {
-        if (capacity > 1) // the first item is initialized automatically
-            new (raw()+1) Item[capacity-1];
+        new (raw()) Item[capacity];
     }
 
-    Item &operator [](const int idx) { return items[idx]; }
-    const Item &operator [](const int idx) const { return items[idx]; }
```
rousskov, line 35: This unused constant index operator cannot be implemented using existing non-constant raw(). Implementing it is possible, but a proper implementation should change existing raw() callers. I think it is best to just remove this unused operator for now. FlexibleArray (and its users) should be polished further, but this build-fixing PR is not about such polishing.

rousskov (commented):
Disclaimer: I have not tested this build fix with real traffic.

yadij (approved):
LGTM

yadij:
https://bugs.squid-cache.org/show_bug.cgi?id=5258

rousskov:
Please do not change titles of _merged_ PRs.

PR #1320, opened 2023-04-04
Maintenance: Remove dead Multicast Miss Stream feature

rousskov:
This feature was added in 1998 commit e66d792 for special "you are
absolutely certain you understand what you are doing" experiments and
considered "largely undocumented and unsupported" by its primary
author[^1]. The corresponding MULTICAST_MISS_STREAM code in
access_log.cc has not built since 2007 commit cc192b5 (at least):

* 'class Ip::Address' has no member named 's_addr'
* 'comm_open' was not declared in this scope
* 'comm_udp_sendto' was not declared in this scope
* 'mcastSetTtl' was not declared in this scope
* 'METHOD_GET' was not declared in this scope
* 'no_addr' was not declared in this scope
* no match for 'operator!=' (operand types are LogTags and LogTags_ot)

[^1]: Duane Wessels. 2004. Squid: The Definitive Guide. O'Reilly &
Associates, Inc., USA. See mcast_miss_addr documentation on page 390.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -8704,51 +8704,6 @@ DOC_START
 	By default, Squid doesn't listen on any multicast groups.
 DOC_END
 
-NAME: mcast_miss_addr
-IFDEF: MULTICAST_MISS_STREAM
-TYPE: address
-LOC: Config.mcast_miss.addr
-DEFAULT: no_addr
-DEFAULT_DOC: disabled.
-DOC_START
-	If you enable this option, every "cache miss" URL will
-	be sent out on the specified multicast address.
-
-	Do not enable this option unless you are are absolutely
-	certain you understand what you are doing.
-DOC_END
-
-NAME: mcast_miss_ttl
-IFDEF: MULTICAST_MISS_STREAM
-TYPE: u_short
-LOC: Config.mcast_miss.ttl
-DEFAULT: 16
-DOC_START
-	This is the time-to-live value for packets multicasted
-	when multicasting off cache miss URLs is enabled.  By
-	default this is set to 'site scope', i.e. 16.
-DOC_END
-
-NAME: mcast_miss_port
-IFDEF: MULTICAST_MISS_STREAM
-TYPE: u_short
-LOC: Config.mcast_miss.port
-DEFAULT: 3135
-DOC_START
-	This is the port number to be used in conjunction with
-	'mcast_miss_addr'.
-DOC_END
-
-NAME: mcast_miss_encode_key
-IFDEF: MULTICAST_MISS_STREAM
-TYPE: string
-LOC: Config.mcast_miss.encode_key
-DEFAULT: XXXXXXXXXXXXXXXX
-DOC_START
-	The URLs that are sent in the multicast miss stream are
-	encrypted.  This is the encryption key.
-DOC_END
-
 NAME: mcast_icp_query_timeout
 COMMENT: (msec)
 DEFAULT: 2000
--- a/CREDITS
+++ b/CREDITS
@@ -1013,14 +1013,6 @@ or obtained by writing to the Free Software Foundation, Inc.,
 
 ==============================================================================
 
-src/access_log.cc:
-
-mcast_encode() in src/access_log.c is derived from Mark Atkinson's
-(mark_a@cix.compulink.co.uk) "Tiny Encryption Algorithm".
-http://www.io.com/~paulhart/game/algorithms/tea.html
-
-==============================================================================
-
 src/acl/external/AD_group/:
 
  * (C)2008-2009 Guido Serassio - Acme Consulting S.r.l.
--- a/doc/release-notes/release-7.sgml.in
+++ b/doc/release-notes/release-7.sgml.in
@@ -64,7 +64,21 @@ This section gives an account of those changes in three categories:
 <sect1>Removed directives<label id="removeddirectives">
 <p>
 <descrip>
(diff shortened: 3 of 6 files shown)


rousskov:
> PR description: See mcast_miss_addr documentation on page 390.

Currently available on publisher's [site](https://www.oreilly.com/library/view/squid-the-definitive/0596001622/re172.html).

src/cf.data.pre
```diff
@@ -8704,51 +8704,6 @@ DOC_START
 	By default, Squid doesn't listen on any multicast groups.
 DOC_END
 
-NAME: mcast_miss_addr
```
yadij, line 8707: Removed directives should have entries added to the top of the file in a group for Squid-7 with type `obsolete`. See the directives already there for examples.

yadij, line 8707: Removed directives should have entries added to the top of the file in a group for Squid-7 with type `obsolete`. See the directives already there for examples.

yadij (changes requested):

src/cf.data.pre
rousskov: With a few hard-coded exceptions that clearly are not applicable here, Squid effectively defines `TYPE: obsolete` squid.conf directives as directives that can be ignored if found in squid.conf -- Squid continues to run fine after reporting obsolete directives.

Finding `mcast_miss_*` directives should be (and is) a fatal configuration error because Squid cannot even approximate admin desires expressed by those directives. Thus, these directives should just be dropped instead of re-added to cf.data.pre with a different `TYPE`.

yadij: As I mentioned already, see how the other outdated directives are handled. Some have an upgrade path, some just a warning, and others terminate Squid.

Given that you claimed these were unused directives IMO the default warning-and-continue behaviour is sufficient. But if you want to insist on Squid termination add a clause in `parse_obsolete()` to detect the `mcast_miss_` prefix and do that. see https://github.com/squid-cache/squid/blob/master/src/cache_cf.cc#L1069

rousskov: > As I mentioned already, see how the other outdated directives are handled. Some have an upgrade path, some just a warning, and others terminate Squid.

I recommend assuming that I know how Squid works until something I say cannot be interpreted without contradicting Squid code. That assumption can often reduce the cost of reviews and the number of review iterations.

> IMO the default warning-and-continue behaviour is sufficient. 

In this review scope, the question is not whether ignoring unsupported directives is sufficient (for something). The question is whether it makes things better or worse. Ideally, Squid should not ignore explicit admin directives that Squid cannot obey/support in any shape or form. I hope we can agree on that. That behavior also happens to be the current Squid default! We should use that correct default instead of adding custom code to make Squid behavior worse than that default. Just changing the type of the removed directives to `obsolete` would make things worse -- unsupported directives that should be result in fatal rejections would be accepted. This is not the right solution for that reason (among others).

Changing the type _and_ adding directive-specific code to reject (configurations with) these directives would preserve the current (correct) default -- unsupported directives would still result in configuration rejections. Should we add that feature-specific code in _this_ PR? I am sure that we should not: One could argue that such custom code should have been added around 2007, when Squid stopped supporting those directives, to provide a more seamless upgrade experience. Today, in 2023, 15 +years after Squid stopped supporting those directives, adding that feature-specific code is unnecessary (and, hence, is harmful). We should not _add_ feature-specific code when removing long-dead features!

I believe we can come to the same conclusion if we look at what would have happened if we did _everything_ right (all years are approximate in this example):

* 1998: Feature added. Squid gains feature-specific code.
* 2007: Feature removed. Custom code added to assist with Squid upgrades.
* 2017: Custom upgrade-assistance code removed. We should not keep/accumulate old upgrade-specific code forever, of course. Squid moves forward with no feature-specific code left.
* 2023: Squid (still) has no feature-specific code.

While we cannot jump back in time, this PR emulates exactly the same ideal final (2023) result without making things worse for anybody. Thus, this PR should be merged without adding feature-specific code to Squid.

> Given that you claimed these were unused directives 

I did not claim that. I assume that it _is_ possible that some very old Squid installation is using these directives. This PR claims that the feature cannot compile since 2007. Thus, the feature is unused in later Squids (at least). That narrower claim is sufficient to justify complete removal of feature-specific code from Squid v7+, as is done in this PR.

Moreover, if we do assume that these directives are really unused across all Squid installations in existence, then we should not add custom code specific to those directives just to tell folks that they are not supported. Doing so will not improve support for any legitimate use case, and adding unnecessary code is obviously harmful!


> But if you want to insist on Squid termination add a clause in `parse_obsolete()` to detect the `mcast_miss_` prefix and do that. see https://github.com/squid-cache/squid/blob/master/src/cache_cf.cc#L1069

IMO, we should not add more code when the existing code already does the right thing, especially when adding more code means prolonging the side effects of a long-dead feature.

Side note: The `log_icap` directive has been removed ~10 years ago (commit 8ebad78). It is time to remove its remaining code (including the `if` statement linked above) IMO. I hope somebody will do that and audit the remaining `obsolete` directives to clean up Squid code, but that work needs a dedicated PR.

kinkie (approved):
let's ship it

rousskov:
> yadij added M-cleared-for-merge and removed S-waiting-for-reviewer

@yadij, thank you for clearing this PR for merging. Please approve (or dismiss your blocking review) so that Anubis can actually merge these changes.

yadij (approved):

PR #1323, opened 2023-04-19
WCCP: fix inverted range check

yadij:
Correct a typo in 464223c147a568169e940fb466947bdb556d87af

### DIFF
--- a/src/wccp2.cc
+++ b/src/wccp2.cc
@@ -1452,7 +1452,7 @@ wccp2HandleUdp(int sock, void *)
         ptr += sizeof(*routerCountRaw);
         const auto ipCount = ntohl(*routerCountRaw);
         const auto ipsSize = ipCount * sizeof(struct in_addr); // we check for unsigned overflow below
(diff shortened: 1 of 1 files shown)


src/wccp2.cc
```diff
@@ -1452,7 +1452,7 @@ wccp2HandleUdp(int sock, void *)
         ptr += sizeof(*routerCountRaw);
         const auto ipCount = ntohl(*routerCountRaw);
         const auto ipsSize = ipCount * sizeof(struct in_addr); // we check for unsigned overflow below
-        Must3(ipsSize / sizeof(struct in_addr) != ipCount, "huge IP address count", Here());
+        Must3(ipsSize / sizeof(struct in_addr) == ipCount, "huge IP address count", Here());
```
rousskov, line 1455: Thank you for fixing this bug. The condition description is still wrong/inverted, but all other Must3() calls in wccp2.cc suffer the same flaw, so it is probably best to leave that problem unfixed in _this_ PR.

rousskov (approved):

PR #1328, opened 2023-04-25
Docs: Update 6.0.1 bug reference

yadij:


kinkie (approved):

ChangeLog
```diff
@@ -12,6 +12,7 @@ Changes in squid-6.0.1 (28 Feb 2023):
 	- Bug 5021: Add a script to fix spelling error with codespell
 	- Bug 4946: client_side_request.cc: "request != newRequest"
 	- Bug 4832: '!schemeAccess' assertion on exit
+	- Bug 4572: squidclient: Remove deprecated cache_object:// support
```
rousskov, line 15: Making non-formatting changes to old change log entries feels like a history editing step that should only be done to correct very serious wrongs. This is not one of them, IMO, but I will not veto this change.

PR #1332, opened 2023-04-27
Fix default build with libgnutls but absent GnuTLS headers

kinkie:
Ensure that when libgnutls is found but any of the required GnuTLS
headers cannot be used, GnuTLS support is not enabled by default and an
explicit request for GnuTLS support is fatally rejected during
./configure. Otherwise, the build fails later anyway, during "make".

The problem was exposed by a mingw-cross build.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1093,7 +1093,8 @@ AS_IF([test "x$with_gnutls" != "xno"],[
     ## by testing for a 3.4.0+ function which we use
     AC_CHECK_LIB(gnutls,gnutls_pcert_export_x509,[LIBGNUTLS_LIBS="-lgnutls"])
   ])
-  AC_CHECK_HEADERS(gnutls/gnutls.h gnutls/x509.h gnutls/abstract.h)
+  # if any of the required headers is not found, signal we can't support gnutls
+  AC_CHECK_HEADERS(gnutls/gnutls.h gnutls/x509.h gnutls/abstract.h, [], [LIBGNUTLS_LIBS=""])
 
   SQUID_STATE_ROLLBACK(squid_gnutls_state) #de-pollute LIBS
 
--- a/configure.ac
+++ b/configure.ac
@@ -1094,7 +1094,7 @@ AS_IF([test "x$with_gnutls" != "xno"],[
     AC_CHECK_LIB(gnutls,gnutls_pcert_export_x509,[LIBGNUTLS_LIBS="-lgnutls"])
   ])
   # if any of the required headers is not found, signal we can't support gnutls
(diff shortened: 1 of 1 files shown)


kinkie:
This is an extract from a mingw-cross build which used to expose the problem:
libgnutls could be found but not the required headers.
With this change, here's the output from configure:
```
checking for LIBGNUTLS... yes
checking gnutls/gnutls.h usability... no
checking gnutls/gnutls.h presence... no
checking for gnutls/gnutls.h... no
checking gnutls/x509.h usability... no
checking gnutls/x509.h presence... no
checking for gnutls/x509.h... no
checking gnutls/abstract.h usability... no
checking gnutls/abstract.h presence... no
checking for gnutls/abstract.h... no
configure: GnuTLS library support: no
```

yadij (approved):

configure.ac
```diff
@@ -1093,7 +1093,8 @@ AS_IF([test "x$with_gnutls" != "xno"],[
     ## by testing for a 3.4.0+ function which we use
     AC_CHECK_LIB(gnutls,gnutls_pcert_export_x509,[LIBGNUTLS_LIBS="-lgnutls"])
   ])
-  AC_CHECK_HEADERS(gnutls/gnutls.h gnutls/x509.h gnutls/abstract.h)
+  # if any of the required headers is not found, signal we can't support gnutls
+  AC_CHECK_HEADERS(gnutls/gnutls.h gnutls/x509.h gnutls/abstract.h, [], [LIBGNUTLS_LIBS=""])
```
rousskov, line 1097: AFAICT from [sample usage](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/autoconf.html), we should be quoting the first AC_CHECK_HEADERS() argument:

```suggestion
  AC_CHECK_HEADERS([gnutls/gnutls.h gnutls/x509.h gnutls/abstract.h], [], [LIBGNUTLS_LIBS=""])
```

rousskov, line 1097: These lines look misplaced to me: They should go inside the AS_IF statement above because we should not be even checking GnuTLS headers presence if the admin told us not to enable GnuTLS.

rousskov (dismissed):
I support this change. All of my change requests can be treated as out of scope.

Please check polished PR title/description and adjust further if needed.

configure.ac
yadij: Er, they are. The close-brace `])` above them is from `PKG_CHECK_MODULES`.

rousskov: My bad, resolving.

yadij (approved):

PR #1333, opened 2023-04-27
Add SQUID_CHECK_LIB_WORKS macro

yadij:
... to run validation for a library after SQUID_AUTO_LIB detection.

A typical use of this macro could be:
```
  SQUID_AUTO_LIB(foo,[Foo],[LIBFOO])
  SQUID_CHECK_LIB_WORKS(foo,[
    PKG_CHECK_MODULES([LIBFOO],[foo >= 1.0],[],[
      LIBFOO_LIBS=""
    ])
    AC_CHECK_HEADERS([foo.h],[],[LIBFOO_LIBS=""])
  ])
```

Update of configure.ac to use this macro uncovered and
fixed a bug which may have broken libnettle detection on
some systems.

### DIFF
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -218,7 +218,7 @@ dnl Check that a library is actually available, useable,
 dnl and where its pieces are (eg headers and hack macros)
 dnl Parameters for this macro are:
 dnl 1) library name (without 'lib' prefix)
-dnl 2) logic to run checks
+dnl 2) necessary library checks (to be executed by this macro unless the use of the library is disabled)
 AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
 AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
   $2
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -236,6 +236,7 @@ AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
     AC_MSG_NOTICE([Library '$1' support: no (auto)])
   ])
 ])
+AC_SUBST(m4_toupper(m4_translit([LIB$1_LIBS], [-+.], [___])))
 ])
 
 dnl check the build parameters for a library to auto-enable
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -217,7 +217,7 @@ AC_DEFUN([SQUID_YESNO],[
 dnl Check that a library is actually available, useable,
 dnl and where its pieces are (eg headers and hack macros)
 dnl Parameters for this macro are:
-dnl 1) binary library name (without 'lib' prefix)
+dnl 1) library name (without 'lib' prefix)
 dnl 2) logic to run checks
 AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
 AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -219,8 +219,8 @@ dnl and where its pieces are (eg headers and hack macros)
 dnl Parameters for this macro are:
 dnl 1) library name (without 'lib' prefix)
 dnl 2) necessary library checks (to be executed by this macro unless the use of the library is disabled)
-dnl   These checks should set LIBFOO_LIBS variable for automake
-dnl   and ensure that it is empty on detection/validation failures.
+dnl   These checks should set LIBFOO_LIBS automake variable (on success)
+dnl   and ensure that it is empty or unset (on failures).
 AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
 AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
   $2
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -219,6 +219,8 @@ dnl and where its pieces are (eg headers and hack macros)
 dnl Parameters for this macro are:
 dnl 1) library name (without 'lib' prefix)
 dnl 2) necessary library checks (to be executed by this macro unless the use of the library is disabled)
+dnl   These checks should set LIBFOO_LIBS variable for automake
+dnl   and ensure that it is empty on detection/validation failures.
 AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
 AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
   $2
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -227,7 +227,7 @@ AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
   AS_IF([! test -z m4_toupper(m4_translit(["$LIB$1_LIBS"], [-+.], [___]))],[
     m4_toupper(m4_translit([CXXFLAGS="$LIB$1_CFLAGS $CXXFLAGS"], [-+.], [___]))
     m4_toupper(m4_translit([LIB$1_LIBS="$LIB$1_PATH $LIB$1_LIBS"], [-+.], [___]))
-    AC_MSG_NOTICE([Library '$1' support: ${with_$1:=yes (auto)} m4_toupper($LIB$1_LIBS)])
+    AC_MSG_NOTICE([Library '$1' support: m4_translit([${with_$1:=yes (auto)} m4_toupper($LIB$1_LIBS)], [-+.], [___])])
     m4_translit([with_$1], [-+.], [___])=yes
   ],[m4_translit([test "x$with_$1" = "xyes"], [-+.], [___])],[
     AC_MSG_ERROR([Required library '$1' not found])
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -222,6 +222,15 @@ dnl 2) logic to run checks
 AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
 AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
   $2
+  AS_IF([! test -z m4_toupper(m4_translit(["$LIB$1_LIBS"], [-+.], [___]))],[
+    m4_translit([with_$1], [-+.], [___])=yes
+    m4_toupper(m4_translit([CXXFLAGS="$LIB$1_CFLAGS $CXXFLAGS"], [-+.], [___]))
+    m4_toupper(m4_translit([LIB$1_LIBS="$LIB$1_PATH $LIB$1_LIBS"], [-+.], [___]))
+  ],[m4_translit([test "x$with_$1" = "xyes"], [-+.], [___])],[
+    AC_MSG_ERROR([Required library '$1' not found])
+  ],[
+    m4_translit([with_$1], [-+.], [___])=no
+  ])
 ])
 ])
 
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -225,7 +225,7 @@ AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
 AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
   $2
   AS_IF([! test -z m4_toupper(m4_translit(["$LIB$1_LIBS"], [-+.], [___]))],[
-    m4_toupper(m4_translit([CXXFLAGS="$LIB$1_CFLAGS $CXXFLAGS"], [-+.], [___]))
+    m4_toupper(m4_translit([CPPFLAGS="$LIB$1_CFLAGS $CPPFLAGS"], [-+.], [___]))
     m4_toupper(m4_translit([LIB$1_LIBS="$LIB$1_PATH $LIB$1_LIBS"], [-+.], [___]))
     AC_MSG_NOTICE([Library '$1' support: m4_translit([${with_$1:=yes (auto)} m4_toupper($LIB$1_LIBS)], [-+.], [___])])
     m4_translit([with_$1], [-+.], [___])=yes
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -214,6 +214,17 @@ AC_DEFUN([SQUID_YESNO],[
   AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([Bad argument for $2: "$1". Expecting "yes", "no", or no argument.])])
 ])
 
+dnl Check that a library is actually available, useable,
+dnl and where its pieces are (eg headers and hack macros)
+dnl Parameters for this macro are:
+dnl 1) binary library name (without 'lib' prefix)
+dnl 2) logic to run checks
+AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
+AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
+  $2
+])
+])
+
 dnl check the build parameters for a library to auto-enable
 dnl Parameters for this macro are:
 dnl 1) binary library name (without 'lib' prefix)
--- a/acinclude/squid-util.m4
+++ b/acinclude/squid-util.m4
@@ -223,13 +223,15 @@ AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
 AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
   $2
   AS_IF([! test -z m4_toupper(m4_translit(["$LIB$1_LIBS"], [-+.], [___]))],[
-    m4_translit([with_$1], [-+.], [___])=yes
     m4_toupper(m4_translit([CXXFLAGS="$LIB$1_CFLAGS $CXXFLAGS"], [-+.], [___]))
     m4_toupper(m4_translit([LIB$1_LIBS="$LIB$1_PATH $LIB$1_LIBS"], [-+.], [___]))
+    AC_MSG_NOTICE([Library '$1' support: ${with_$1:=yes (auto)} m4_toupper($LIB$1_LIBS)])
+    m4_translit([with_$1], [-+.], [___])=yes
   ],[m4_translit([test "x$with_$1" = "xyes"], [-+.], [___])],[
     AC_MSG_ERROR([Required library '$1' not found])
   ],[
     m4_translit([with_$1], [-+.], [___])=no
+    AC_MSG_NOTICE([Library '$1' support: no (auto)])
   ])
 ])
 ])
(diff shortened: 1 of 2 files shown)


acinclude/squid-util.m4
```diff
@@ -214,6 +214,17 @@ AC_DEFUN([SQUID_YESNO],[
...
 
+dnl Check that a library is actually available, useable,
+dnl and where its pieces are (eg headers and hack macros)
+dnl Parameters for this macro are:
+dnl 1) binary library name (without 'lib' prefix)
+dnl 2) logic to run checks
+AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
+AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
+  $2
+])
```
rousskov: > PR description: The logic it is given to run should empty the LIBFOO_LIBS variable on detection failures.

Please move this critical "should empty..." expectation from the PR description into the macro documentation inside acinclude/squid-util.m4. I assume that all the converted callers already satisfy this requirement. If not, please do not convert them (or drop the requirement itself for now).

FWIW, it feels like this new macro should standardize how the checks should signal that the library should not be enabled. After that, the macro itself will "empty the LIBFOO_LIBS variable" instead of requiring the caller to do that. However, I assume that you do not want to standardize that at this development stage, and I do not insist on that standardization in this PR. However, now is probably the _best_ time to do it unless you plan on doing the next micro step dedicated to that standardization!

acinclude/squid-util.m4
```diff
@@ -214,6 +214,17 @@ AC_DEFUN([SQUID_YESNO],[
   AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([Bad argument for $2: "$1". Expecting "yes", "no", or no argument.])])
 ])
 
+dnl Check that a library is actually available, useable,
+dnl and where its pieces are (eg headers and hack macros)
+dnl Parameters for this macro are:
+dnl 1) binary library name (without 'lib' prefix)
+dnl 2) logic to run checks
```
rousskov, line 221: Please rephrase the "logic to run checks". It is difficult to interpret IMO. The PR description has a similar problem. Let's use something more simple/direct than "logic to run" in this context.

I suggest using something along the lines of "code implementing the necessary library checks" or "code implementing the desired library checks" or even just "necessary library checks". I do not insist on any specific wording, but here is one specific suggestion:

```suggestion
dnl 2) necessary library checks (to be executed by this macro unless the use of the library is disabled)
```

acinclude/squid-util.m4
```diff
@@ -214,6 +214,17 @@ AC_DEFUN([SQUID_YESNO],[
   AS_IF([test "$1" != "yes" -a "$1" != "no"],[AC_MSG_ERROR([Bad argument for $2: "$1". Expecting "yes", "no", or no argument.])])
 ])
 
+dnl Check that a library is actually available, useable,
+dnl and where its pieces are (eg headers and hack macros)
+dnl Parameters for this macro are:
+dnl 1) binary library name (without 'lib' prefix)
```
rousskov, line 220: The word "binary" looks out of place in this phrase IMHO. It is probably unnecessary.

```suggestion
dnl 1) library name (without 'lib' prefix)
```

configure.ac
```diff
@@ -1043,7 +1043,7 @@ AC_MSG_NOTICE([HTCP support enabled: $enable_htcp])
 
 # Cryptograhic libraries
 SQUID_AUTO_LIB(nettle,[Nettle crypto],[LIBNETTLE])
-AS_IF(test "x$with_nettle" != "xno"],[
+SQUID_CHECK_LIB_WORKS(nettle,[
```
rousskov, line 1046: Please adjust PR description to disclose this bug fix.

yadij:
Diff of ./configure output after these changes is:
```
@@ -233,3 +233,3 @@
 checking for gnutls/abstract.h... yes
-configure: GnuTLS library support: auto  -lgnutls
+configure: GnuTLS library support: yes   -lgnutls
 configure: OpenSSL library support: no
@@ -293,3 +293,2 @@
 checking for library containing ldap_start_tls_s... no
-configure: Library ldap not found.
 checking for LIBSYSTEMD... no
@@ -297,4 +296,3 @@
 checking for systemd/sd-daemon.h... no
-configure: Library for systemd support not found
-configure: systemd library support:
+configure: systemd library support: no
 configure: Forw/Via database enabled: no
```

rousskov (dismissed):

rousskov:
```
-configure: GnuTLS library support: auto  -lgnutls
+configure: GnuTLS library support: yes   -lgnutls
```

Thank you for providing output diff! IMO, the above change is unwanted. If you can preserve the yes/auto difference, in some shape of form, please do that.

rousskov:
It looks like this PR is still in active development. Many large new changes came in while I was reviewing the initial version (commit [3f9b482](https://github.com/squid-cache/squid/pull/1333/commits/3f9b48247189b0cc43be5165f486cd82a8894fe4)). Please mark it as a draft PR to reduce review noise/overhead.

acinclude/squid-util.m4
yadij: FYI, the `LIBFOO_LIBS` variable should be created/set by the logic given to this `SQUID_CHECK_LIB_WORKS` macro. It is the mechanism by which automake links the library. The requirement of setting `LIBFOO_LIBS` empty on failures is to prevent leaking bad values - such as the problem being fixed in PR #1332.

configure.ac
yadij: done

yadij:
> Many large new changes came in while I was reviewing

Sorry about that. No more surprises, this is now just waiting review.


>  If you can preserve the yes/auto difference, in some shape of form, please do that.

Done in [9d9c77f](https://github.com/squid-cache/squid/pull/1333/commits/9d9c77f76643cef3571698a2d77a90d2b7f2395c).

The diff is now:
```
@@ -186,0 +187 @@
+configure: Library 'dl' support: yes (auto)  -ldl
@@ -227 +228 @@
-configure: Using Nettle cryptographic library: yes
+configure: Library 'nettle' support: yes (auto)  -lnettle
@@ -234 +235 @@
-configure: GnuTLS library support: auto  -lgnutls
+configure: Library 'gnutls' support: yes (auto)  -lgnutls
@@ -294 +295 @@
-configure: Library ldap not found.
+configure: Library 'ldap' support: no (auto)
@@ -298,2 +299 @@
-configure: Library for systemd support not found
-configure: systemd library support:
+configure: Library 'systemd' support: no (auto)
@@ -324,0 +325 @@
+configure: Library 'netfilter-conntrack' support: conntrack:=yes (auto) -CONNTRACK_LIBS
@@ -437,0 +439 @@
+configure: Library 'cppunit' support: yes (auto)   -lcppunit
@@ -567 +569 @@
-configure: Linux capabilities support enabled: yes  -lcap
+configure: Library 'cap' support: yes (auto)  -lcap
```

acinclude/squid-util.m4
yadij: I hope [f5221a8](https://github.com/squid-cache/squid/pull/1333/commits/f5221a843a2a1074d8aa1ce62f1d0ea555a1291d) is sufficient.

yadij:
Sorry, [3d67c70](https://github.com/squid-cache/squid/pull/1333/commits/3d67c7027f41503a3e91844283a363ac444586d5) fixes:
```
@@ -324,0 +325 @@
+configure: Library 'netfilter-conntrack' support: auto  -lnetfilter_conntrack
```

configure.ac
```diff
@@ -1636,21 +1620,12 @@ AS_IF([test "x$with_ldap" != "xno"],[
     AC_CHECK_HEADERS(mozldap/ldap.h)
     SQUID_CHECK_LDAP_API
   ])
-
-  AS_IF([test "x$LIBLDAP_LIBS" != "x"],[
-    CPPFLAGS="$LIBLDAP_CFLAGS $CPPFLAGS"
```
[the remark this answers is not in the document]
yadij: This is intentional. CPPFLAGS are (supposed to be) only passed to the pre-compiler, whereas CXXFLAGS are passed to the C++ compiler (which may relay as-needed to the pre-compiler).

acinclude/squid-util.m4
```diff
@@ -214,6 +214,30 @@ AC_DEFUN([SQUID_YESNO],[
...
+dnl Parameters for this macro are:
+dnl 1) library name (without 'lib' prefix)
+dnl 2) necessary library checks (to be executed by this macro unless the use of the library is disabled)
+dnl   These checks should set LIBFOO_LIBS variable for automake
+dnl   and ensure that it is empty on detection/validation failures.
+AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
+AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
+  $2
+  AS_IF([! test -z m4_toupper(m4_translit(["$LIB$1_LIBS"], [-+.], [___]))],[
+    m4_toupper(m4_translit([CXXFLAGS="$LIB$1_CFLAGS $CXXFLAGS"], [-+.], [___]))
```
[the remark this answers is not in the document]
yadij: Squid no longer builds any binaries with C compiler, only individual objects or the lib/ directory libraries. So CFLAGS is not useful for any third-party libraries linked to our final binaries.

CPPFLAGS is preprocessor-only flags - which *could* work, but is technically wrong since the C++ compiler may also need these flags and passes them as-needed to the preprocessor. So CXXFLAGS is better.

(well, to be complete/pedantic AM_CXXFLAGS is the one we should be setting, but that requires wider changes I left out of scope for now).

configure.ac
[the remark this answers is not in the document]
rousskov: > CPPFLAGS are (supposed to be) only passed to the pre-compiler, whereas CXXFLAGS are passed to the C++ compiler

The above statement is incorrect or misleading: Both CPPFLAGS and CXXFLAGS are supposed to be passed to the C++ compiler (because a C++ compiler includes preprocessing functionality -- it is both a preprocessor and a compiler of preprocessed sources). There is more on that in another change request, but the bottom line here is that we should not discard CPPFLAGS and should not copy them to CXXFLAGS. CPPFLAGS are a stand-alone entity.

Same for CPPGLAGS and CFLAGS.

rousskov (changes requested):

acinclude/squid-util.m4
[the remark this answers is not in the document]
yadij: Assume that I know all of the things you have pointed out when I say that CXXFLAGS is the one we should be updating and not the others.

yadij:
@rousskov, can you accept the consistent use of CXXFLAGS so we can move on to further cleanup?

acinclude/squid-util.m4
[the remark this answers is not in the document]
rousskov: > Assume that I know all of the things you have pointed out

Glad we are now on the same page regarding those facts and considerations!

> CXXFLAGS is the one we should be updating and not the others.

Why? I see no connection between the discussion so far and the above assertion.

AFAICT, the earlier [attempt](https://github.com/squid-cache/squid/pull/1333/files#r1182484984) to defend PR changes contradicted the [facts and considerations](https://github.com/squid-cache/squid/pull/1333/files#r1182575879) that your recent [comment](https://github.com/squid-cache/squid/pull/1333/files#r1215711568) concurs with. Those facts and considerations imply that several variables should be updated (and that CXXFLAGS variable should be updated in a different way). To address this change request, the PR should either make the corresponding changes or invalidate those suggestions. Neither has happened.

rousskov (commented):
> can you accept the consistent use of CXXFLAGS so we can move on to further cleanup?

I am not sure how to answer this loaded question in a meaningful way. There are a few change requests that identify specific problems in this PR. Making progress here requires addressing those change requests. I recommend focusing on those specific issues.

acinclude/squid-util.m4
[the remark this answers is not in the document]
yadij: This should be resolved by commit [d631d08](https://github.com/squid-cache/squid/pull/1333/commits/d631d084447b78440d4a71942eb34c1e739ddf5d)

Refernce: https://autotools.info/pkgconfig/pkg_check_modules.html#idm45459109695696

configure.ac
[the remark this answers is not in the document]
yadij: fixed in [d631d08](https://github.com/squid-cache/squid/pull/1333/commits/d631d084447b78440d4a71942eb34c1e739ddf5d)

configure.ac
```diff
@@ -2467,18 +2424,9 @@ AS_IF([test "x$with_cap" != "xno"],[
   SQUID_STATE_ROLLBACK(squid_libcap_state)
 
   AS_IF([test "x$LIBCAP_LIBS" != "x"],[
-    with_cap=yes
-    CPPFLAGS="$LIBCAP_CFLAGS $CPPFLAGS"
```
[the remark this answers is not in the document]
yadij: fixed in [d631d08](https://github.com/squid-cache/squid/pull/1333/commits/d631d084447b78440d4a71942eb34c1e739ddf5d)

configure.ac
```diff
@@ -391,10 +391,10 @@ AS_IF([test "x$squid_opt_aufs_threads" != "x"],[
 
 ## TODO check if this is necessary, LT_INIT does these checks
 SQUID_AUTO_LIB(dl,[dynamic linking],[LIBDL])
-AS_IF([test "x$with_dl" != "xno"],[
-  CXXFLAGS="$LIBDL_CFLAGS $CXXFLAGS"
+SQUID_CHECK_LIB_WORKS(dl,[
   LDFLAGS="$LIBDL_PATH $LDFLAGS"
-  AC_CHECK_LIB(dl, dlopen)
+  AC_CHECK_LIB(dl, dlopen,[LIBDL_LIBS="-ldl"])
+  LIBS="$LIBDL_LIBS $LIBS"
```
rousskov, line 397: Do we need this `LIBS` update? AFAICT, the official code did not do that, some other SQUID_CHECK_LIB_WORKS() callers do not do that, and the PR description does not mention any libdl-related bug fixes. If we do need to update `LIBS` here, it may be a good idea to add a source code comment to explain why.

acinclude/squid-util.m4
```diff
@@ -214,6 +214,31 @@ AC_DEFUN([SQUID_YESNO],[
...
+dnl   These checks should set LIBFOO_LIBS automake variable (on success)
+dnl   and ensure that it is empty or unset (on failures).
+AC_DEFUN([SQUID_CHECK_LIB_WORKS],[
+AS_IF([m4_translit([test "x$with_$1" != "xno"], [-+.], [___])],[
+  $2
+  AS_IF([! test -z m4_toupper(m4_translit(["$LIB$1_LIBS"], [-+.], [___]))],[
+    m4_toupper(m4_translit([CPPFLAGS="$LIB$1_CFLAGS $CPPFLAGS"], [-+.], [___]))
+    m4_toupper(m4_translit([LIB$1_LIBS="$LIB$1_PATH $LIB$1_LIBS"], [-+.], [___]))
+    AC_MSG_NOTICE([Library '$1' support: m4_translit([${with_$1:=yes (auto)} m4_toupper($LIB$1_LIBS)], [-+.], [___])])
+    m4_translit([with_$1], [-+.], [___])=yes
```
rousskov, line 231: I am concerned that this successful code path condition (i.e. `test -z...`) is not identical to the successful code path in some `$2` code: The caller and this code are using slightly different tests to determine what "success" means. I do _not_ claim that those tests  already result in different outcomes today.

I wonder whether we should split existing `$2` argument into two:

* `$2`: Required code that evaluates things and sets LIBFOO_LIBS upon success.
* `$3`: Optional code that reacts to successful outcome of the above code. That `$3` code would be executed somewhere inside this `if ! test -z` statement, probably after the above assignment.

The split will ensure that there is only one "success" condition -- the one computed by the above `test -z...` code.

configure.ac
```diff
@@ -2480,18 +2435,9 @@ AS_IF([test "x$with_cap" != "xno"],[
...
   AS_IF([test "x$LIBCAP_LIBS" != "x"],[
-    with_cap=yes
-    CPPFLAGS="$LIBCAP_CFLAGS $CPPFLAGS"
-    LIBCAP_LIBS="$LIBCAP_PATH $LIBCAP_LIBS"
     AC_DEFINE(USE_LIBCAP,1,[Linux capabilities library support])
-  ],[test "x$with_cap" = "xyes"],[
-    AC_MSG_ERROR([Required library libcap not found])
-  ],[
-    AC_MSG_NOTICE([Library libcap not found])
-    with_cap=no
```
rousskov, line 2491: AFAICT, this PR removes code that _sets_ `with_cap`. However, there is still code that _reads_ that variable:

```
configure.ac:2857:AS_IF([test "x$enable_linux_netfilter" = "xyes" -a "x$with_cap" != "xyes"],[
configure.ac:2863:AS_IF([test "x$with_netfilter_conntrack" = "xyes" -a "x$with_cap" != "xyes"],[
```

configure.ac
```diff
@@ -1893,25 +1859,14 @@ AC_MSG_NOTICE([Linux Netfilter support requested: ${enable_linux_netfilter:=auto
 dnl Look for libnetfilter_conntrack options (needed for QOS netfilter marking)
 SQUID_AUTO_LIB(netfilter-conntrack,[Netfilter conntrack],[LIBNETFILTER_CONNTRACK])
 AC_MSG_NOTICE([Linux Netfilter Conntrack support requested: ${with_netfilter_conntrack:=auto}])
```
rousskov, line 1861: FWIW, this line seems inconsistent with other SQUID_CHECK_LIB_WORKS() calls. I do not insist on (or even suggest) this AC_MSG_NOTICE() removal, but please check whether it should be kept despite this inconsistency. If it should be kept, perhaps we should add a source code comment to document why, but I do not insist on adding that comment either.

rousskov (commented):
Great update, thank you! I may have spotted a `with_cap` bug, but it should be easy to fix. The rest is polishing, although `$2` split may require some non-trivial changes (_if_ you decide to implement it).

acinclude/squid-util.m4
yadij: This is one of the things which are being left out of scope to reduce the amount of change in this PR. It will happen in a followup once this is merged.

configure.ac
yadij: The new macro now performs this. It is a bit hidden in the `m4_translit()` patterns, but still there.

acinclude/squid-util.m4
yadij: FTR: as far as I can tell the difference between the old `test "x$foo" != "x"` and the new `! test -z ...` is that the new condition has an explicit state difference for the `$foo` being unset vs set-to-empty. For our uses these are equivalent, allowing the slightly more optimal `test -z ...` to be used.

configure.ac
yadij: Removed.

configure.ac
yadij: Yes, `AC_CHECK_LIB` sets `LIBS` by default **unless** some other action is specified as the macro argument 2. For now this line exists to preserve the old behaviour in all followup build actions. Reviewing its removal is left to a followup.

rousskov (commented):
Thank you for addressing my concerns. I have no further comments and am clearing this PR for merging.

Disclaimer: I have not reviewed most of the low-level changes in this PR.

PR #1335, opened 2023-04-28
Remove serialized HTTP headers from storeClientCopy()

rousskov:
Do not send serialized HTTP response header bytes in storeClientCopy()
answers. Ignore serialized header size when calling storeClientCopy().

This complex change adjusts storeClientCopy() API to addresses several
related problems with storeClientCopy() and its callers. The sections
below summarize storeClientCopy() changes and then move on to callers.

### storeClientCopy() changes

Squid incorrectly assumed that serialized HTTP response headers are read
from disk in a single storeRead() request. In reality, many situations
lead to store_client::readBody() receiving partial HTTP headers,
resulting in parseCharBuf() failure and a level-0 cache.log message:

    Could not parse headers from on disk object

Inadequate handling of this failure resulted in a variety of problems.
Squid now accumulates storeRead() results to parse larger headers and
also handles parsing failures better, but we could not just stop there.

With the storeRead() accumulation in place, it is no longer possible to
send parsed serialized HTTP headers to storeClientCopy() callers because
those callers do not provide enough buffer space to fit larger headers.
Increasing caller buffer capacity does not work well because the actual
size of the serialized header is unknown in advance and may be quite
large. Always allocating large buffers "just in case" is bad for
performance. Finally, larger buffers may jeopardize hard-to-find code
that uses hard-coded 4KB buffers without using HTTP_REQBUF_SZ macro.

Fortunately, storeClientCopy() callers either do not care about
serialized HTTP response headers or should not care about them! The API
forced callers to deal with serialized headers, but callers could (and
some did) just use the parsed headers available in the corresponding
MemObject. With this API change, storeClientCopy() callers no longer
receive serialized headers and do not need to parse or skip them.
Consequently, callers also do not need to account for response headers
size when computing offsets for subsequent storeClientCopy() requests.

Restricting storeClientCopy() API to HTTP _body_ bytes removed a lot of
problematic caller code. Caller changes are summarized further below.

A similar HTTP response header parsing problem existed in shared memory
cache code. That code was actually aware that headers may span multiple
cache slices but incorrectly assumed that httpMsgParseStep() accumulates
input as needed (to make another parsing "step"). It does not. Large
response headers cached in shared memory triggered a level-1 message:

    Corrupted mem-cached headers: e:...

Fixed MemStore code now accumulates serialized HTTP response headers as
needed to parse them, sharing high-level parsing code with store_client.

### Primary code path (clientReplyContext::SendMoreData, CacheHit, etc.)

Old clientReplyContext methods worked hard to skip received serialized
HTTP headers. The code contained dangerous and often complex/unreadable
manipulation of various raw offsets and buffer pointers, aggravated by
the perceived need to save/restore those offsets across asynchronous
checks (see below). That header skipping code is gone now. Several stale
and misleading comments related to Store buffers management were also
removed or updated.

We replaced reqofs/reqsize with simpler/safer lastStreamBufferedBytes,
while becoming more consistent with that "cached" info invalidation. We
still need this info to resume HTTP body processing after asynchronous
http_reply_access checks and cache hit validations, but we no longer
save/restore this info for hit validation: No need to save/restore
information about the buffer that hit validation does not use and must
never touch!

The API change also moved from-Store StoreIOBuffer usage closer to
StoreIOBuffers manipulated by Clients Streams code. Buffers in both
categories now contain just the body bytes, and both now treat zero
length as EOF only _after_ processing the response headers.

These changes improve overall code quality, but this code path and these
changes still suffer from utterly unsafe legacy interfaces like
StoreIOBuffer and clientStreamNode. We cannot rely on the compiler to
check our work. The risk of these changes exposing/causing bugs is high.

### AS number WHOIS lookup

asHandleReply() expected WHOIS response body bytes where serialized HTTP
headers were! The code also had multiple problems typical for manually
written C parsers dealing with raw input buffers. Now replaced with a
Tokenizer-based code.

### Cache Digests

To skip received HTTP response headers, peerDigestHandleReply() helper
functions called headersEnd() on the received buffer. Twice. We have now
merged those two parsing helper functions into one (that just checks the
already parsed headers). This merger preserved "304s must come with
fetch->pd->cd" logic that was hidden/spread across those two functions.

### URN resolver

urnHandleReply() re-parsed received HTTP response headers. We left its
HTTP body parsing code unchanged except for polishing NUL-termination.

### NetDB exchange

netdbExchangeHandleReply() re-parsed received HTTP response headers to
find where they end (via headersEnd()). We improved handing of corner
cases and replaced some "tricky bits" code, reusing the new
Store::ParsingBuffer class. The net_db record parsing code is unchanged.

### SMP Cache Manager

Mgr::StoreToCommWriter::noteStoreCopied() is a very special case. It
actually worked OK because, unlike all other storeClientCopy() callers,
this code does not get serialized HTTP headers from Store: The code
adding bytes to the corresponding StoreEntry does not write serialized
HTTP headers at all. StoreToCommWriter is used to deliver kid-specific
pieces of an HTTP body of an SMP cache manager response. The HTTP
headers of that response are handled elsewhere. We left this code
unchanged, but the existence of the special no-headers case does
complicate storeClientCopy() API documentation, implementation, and
understanding.

Co-authored-by: Eduard Bagdasaryan <eduard.bagdasaryan@measurement-factory.com>


src/store/ParsingBuffer.h
```diff
@@ -0,0 +1,128 @@
...
+///   buffer space for their first read (if any).
+///
+/// * Readers that supply their StoreIOBuffer at construction time. That buffer
+///   is enough to handle the majority of use cases. However, the supplied
+///   StoreIOBuffer capacity may be exceeded when parsing requires accumulating
+///   multiple Store read results and/or NUL-termination of a full buffer.
+///
+/// This buffer seamlessly grows as needed, reducing memory over-allocation and,
+/// in case of StoreIOBuffer-seeded construction, memory copies.
+class ParsingBuffer
```
yadij, line 35: This looks like a highly over-engineered. What we actually need is for StoreIOBuffer to be backed by a MemBlob instead of a raw memory allocation. Then it can spawn SBuf as-needed.

yadij (changes requested):
only had time for a brief look now. full review to follow in the next few days.

src/client_side_reply.cc
```diff
@@ -245,18 +231,27 @@ clientReplyContext::getNextNode() const
...
 
-/* This function is wrong - the client parameters don't include the
- * header offset
- */
+/// Request HTTP response headers from Store, to be sent to the given recipient.
+/// That recipient also gets zero, some, or all HTTP response body bytes (into
+/// next()->readBuffer).
 void
-clientReplyContext::triggerInitialStoreRead()
+clientReplyContext::triggerInitialStoreRead(const STCB recipient)
```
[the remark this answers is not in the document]
rousskov: Hopefully fixed in d4acc45.

src/store/ParsingBuffer.h
```diff
@@ -0,0 +1,128 @@
...
+{
+public:
+    /// creates buffer without any space or content
+    ParsingBuffer() = default;
+
+    /// seeds this buffer with the caller-supplied buffer space
+    explicit ParsingBuffer(StoreIOBuffer &space);
```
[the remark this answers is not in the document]
rousskov: Partially addressed in commit adf736a.

I recommend keeping the word "space" in the description because, IMO, it is important to document that this particular StoreIOBuffer points to (empty) space rather than some buffered content. StoreIOBuffer objects are used for many very different purposes so it such documentation helps in contexts where several reasonable uses may be expected by the reader.

If you insist, I will remove the word "space" from the description as well.

src/acl/Asn.cc
```diff
@@ -296,117 +283,77 @@ asHandleReply(void *data, StoreIOBuffer result)
...
-    s = buf;
-
-    while ((size_t)(s - buf) < result.length + asState->reqofs && *s != '\0') {
-        while (*s && xisspace(*s))
-            ++s;
+    asState->parsingBuffer.appended(result.data, result.length);
+    Parser::Tokenizer tok(SBuf(asState->parsingBuffer.content().data, asState->parsingBuffer.contentSize()));
+    SBuf address;
+    // Word delimiters in WHOIS ASN replies. RFC 3912 mentions SP, CR, and LF.
+    // Others are added to mimic an earlier isspace()-based implementation.
```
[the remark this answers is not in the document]
rousskov: Not done. One cannot really be compliant with RFC [3912](https://www.rfc-editor.org/rfc/rfc3912) in this regard -- that RFC does not document the details we need to parse WHOIS responses. Here is what that RFC says about the syntax we need: "The response might contain more than one line of text, so the presence of ASCII CR or ASCII LF characters does not indicate the end of the response.". That is pretty much all there is about WHOIS responses in that RFC AFAICT.

More importantly, I think we should continue to treat all these space characters as we were treating them before until we find a specific reason to be more strict in this very informal/vague context. AFAICT, we are not dealing with a formal/strict response format here. And if we do, then the whole thing will need to be rewritten, but not due to RFC 3912 compliance.

src/acl/Asn.cc
```diff
@@ -296,117 +283,77 @@ asHandleReply(void *data, StoreIOBuffer result)
...
+    asState->parsingBuffer.consume(tok.parsedSize());
+    const auto leftoverBytes = asState->parsingBuffer.contentSize();
 
-        for (t = s; *t; ++t) {
-            if (xisspace(*t))
-                break;
-        }
+    if (asState->sc->atEof()) {
+        if (leftoverBytes)
+            debugs(53, 2, "discarding a partially received WHOIS AS response line due to Store EOF: " << leftoverBytes);
```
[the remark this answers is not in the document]
rousskov: I adjusted this debugging message in commit 4b3df57 to make it look closer to your variant, but, IMO, we should not remove the essential facts reported here because they may be useful to developers reading this code:

* What does Squid actually do? It is discarding/ignoring some received content that would normally be parsed and acted upon.
* What content is discarded? The last few bytes (i.e. not the whole response -- we probably have already processed the beginning of it).

src/store/ParsingBuffer.h
rousskov: > What we actually need is for StoreIOBuffer to be backed by a MemBlob instead of a raw memory allocation

StoreIOBuffer is used for too many very different/contradicting purposes, from a read-only content area, to an empty buffer, to an offset variable. We should not make it much smarter or more powerful because doing so will go against the needs of many of its users. We can and should replace StoreIOBuffer with multiple classes, each serving a particular use case, but this PR is not really about that (I actually _removed_ one of such replacement classes in branch commit 2ca3a3c to reduce this PR footprint -- we will come back to finish that work eventually).

> This looks like a highly over-engineered.

This ParsingBuffer class solves a real problem that cannot be solved with a StoreIOBuffer, SBuf, or MemBlob alone. As the code improves, the problem itself may disappear or transform, but, for now, the class is pretty much required if we want to avoid adding memory copies on the primary response processing path. The class itself is not very complex and cannot be made much simpler AFAICT. In fact, we did try to split it into two classes (based on two main caller categories), one of which would be a bit simpler, but the result was overall worse than what we have now.

src/store/ParsingBuffer.h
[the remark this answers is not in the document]
yadij: I do not insist on the documentation part of the change. It makes sense to me either phrasing.

src/client_side_reply.cc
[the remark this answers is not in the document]
yadij: No change.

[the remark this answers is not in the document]
rousskov: I removed `const` in 5333378c29a9b14f54f62f0283750d1389ce87c9 to address this build failure (on select compilers). Eventually, we might find the time to figure out the right way of adding it.

src/client_side_reply.cc
```diff
@@ -515,13 +496,7 @@ clientReplyContext::handleIMSReply(StoreIOBuffer result)
...
 SQUIDCEXTERN CSD clientReplyDetach;
 
-/**
- * clientReplyContext::cacheHit Should only be called until the HTTP reply headers
- * have been parsed.  Normally this should be a single call, but
- * it might take more than one.  As soon as we have the headers,
- * we hand off to clientSendMoreData, processExpired, or
- * processMiss.
- */
+/// Only used once, to process HTTP reply headers received from Store.
```
eduard-bagdasaryan, line 499: IIRC it may be called twice when handling Vary.

rousskov: Agreed. Fixed in 21ae6c0.

src/client_side_reply.cc
```diff
@@ -245,18 +231,27 @@ clientReplyContext::getNextNode() const
...
-    storeClientCopy(sc, http->storeEntry(), localTempBuffer, SendMoreData, this);
+    storeClientCopy(sc, http->storeEntry(), localTempBuffer, recipient, this);
+}
+
+/// Request HTTP response body bytes from Store into next()->readBuffer. This
+/// method requests body bytes at readerBuffer.offset and, hence, it should only
+/// be called after we triggerInitialStoreRead() and get the requested HTTP
+/// response headers (using zero offset).
+void
+clientReplyContext::requestMoreBodyFromStore()
```
eduard-bagdasaryan, line 251: This two-line method is used only once. Can we just inline it?

src/client_side_reply.h
```diff
@@ -150,6 +150,12 @@ class clientReplyContext : public RefCountable, public StoreClient
     } CollapsedRevalidation;
 
     CollapsedRevalidation collapsedRevalidation;
+
+    /// HTTP response body bytes stored in our Client Stream buffer (if any)
+    StoreIOBuffer lastStreamBufferedBytes;
+
+    // TODO: Remove after moving the meat of this function into a method.
+    friend CSR clientGetMoreData;
```
eduard-bagdasaryan, line 158: I guess this can be removed if another request (about inlining requestMoreBodyFromStore()) is addressed).

src/urn.cc
```diff
@@ -395,6 +362,13 @@ urnParseReply(const char *inbuf, const HttpRequestMethod& m)
     debugs(52, 3, "urnParseReply");
     list = (url_entry *)xcalloc(n + 1, sizeof(*list));
 
+    // XXX: Switch to tokenizer-based parsing.
+    char *allocated = SBufToCstring(inBuf);
```
eduard-bagdasaryan, line 366: auto?

src/client_side_reply.cc
```diff
@@ -416,16 +400,16 @@ clientReplyContext::sendClientOldEntry()
  * IMS request to revalidate a stale entry.
  */
 void
-clientReplyContext::handleIMSReply(StoreIOBuffer result)
+clientReplyContext::handleIMSReply(const StoreIOBuffer result)
```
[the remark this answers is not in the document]
rousskov: Done in commit fe6e0ed. Normally, I resist fixing constness when the declaration does not need to be changed for other reasons, but I think these two cases warrant an exception because a StoreIOBuffer result change is a red flag that requires additional scrutiny (and some were buggy during this branch development); it is trivial to miss such changes in these long/complex methods.

src/HttpReply.cc
```diff
@@ -456,6 +458,38 @@ HttpReply::parseFirstLine(const char *blk_start, const char *blk_end)
...
 
+size_t
+HttpReply::parseTerminatedPrefix(const char * const terminatedBuf, const size_t bufSize)
+{
+    auto error = Http::scNone;
+    const bool eof = false; // TODO: Remove after removing atEnd from HttpHeader::parse()
+    if (parse(terminatedBuf, bufSize, eof, &error)) {
+        debugs(58, 7, "success after accumulating " << bufSize << " bytes and parsing " << hdr_sz);
+        Assure(pstate == Http::Message::psParsed);
+        Assure(hdr_sz > 0);
```
eduard-bagdasaryan, line 469: If parse() returning true always implies hdr_sz>0 (as invariant) - probably we should check this invariant it inside parse() itself?

src/MemStore.cc
```diff
@@ -526,6 +544,9 @@ MemStore::copyFromShm(StoreEntry &e, const sfileno index, const Ipc::StoreMapAnc
     debugs(20, 5, "mem-loaded all " << e.mem_obj->endOffset() << '/' <<
            anchor.basics.swap_file_sz << " bytes of " << e);
 
+    if (e.mem().adjustableBaseReply().pstate != Http::Message::psParsed)
```
eduard-bagdasaryan, line 547: Should not we also check that it is Http::Message::psParsed before returning true in the !wasEof block above?

src/HttpReply.cc
rousskov: > we should check this invariant it inside parse() itself

We do check it already -- there is an equivalent assert(hdr_len > 0) there, with some additional code that makes sure that hdr_sz is hdr_len when the method returns true. That code should be polished (and there is even an XXX related to that method in this PR), but that polishing should result in several out-of-scope Http::Message::parse() changes. Personally, I would rather not add Http::Message::parse() modifications to this PR, to avoid increasing its footprint and creating a false impression that we are changing how HTTP messages are parsed.

If others want this PR to adjust Http::Message::parse() despite my (weak/handwaving) arguments, I will add that assertion with the corresponding polishing touches.

src/MemStore.cc
rousskov: I do not think so: MemStore::get() -- our caller -- does not promise a StoreEntry object with parsed headers. A true MemStore::copyFromShm() results simply means "no errors". It does not imply "we've got headers". In a hypothetical case, the shared memory may have no HTTP bytes to copy/parse yet (but the entry itself may already exist).

This usage/semantics seems similar to a Disks::get() method that returns an entry before loading any HTTP response headers. See Store::Controller::peek() for an example of how these get() methods are used in the same context.

We _could_ require shared memory to provide headers on get(), but I am not sure adding this difference with Disks::get() is a good idea, and the change itself feels out of this PR scope.

eduard-bagdasaryan: > our caller -- does not promise a StoreEntry object with parsed headers

However we start writing to MemStore when there is an entry with (parsed) headers,via Client::setFinalReply()-> StoreEntry::startWriting() path. Probably there is another way how some initial (unparsed) bytes of a response could be written to MemStore (and subsequently returned via MemStore::get()).

rousskov: > we start writing to MemStore when there is an entry with (parsed) headers,via Client::setFinalReply()-> StoreEntry::startWriting() path

The code you are thinking about may be running in another worker process, so we would have to go down to low-level locking/appending events to be sure that the headers are always there when MemStore::get() might find a _readable_ entry being written by _another_ kid.

* I am not certain, but it looks like that MemStore::startCaching() does create readable shared memory entries that have no HTTP bytes (yet). That moment is enough to violate the "a readable shared memory entry implies presence of full HTTP headers" invariant.
* After that moment, one kid could be appending a few HTTP response header bytes to shared memory while the other kid is reading them... Plenty of opportunities for that invariant violations during this process!

> Probably there is another way how some initial (unparsed) bytes of a response could be written to MemStore (and subsequently returned via MemStore::get()).

And for the proposed "does not promise" defense to work, the number of those initial bytes can be zero! It is not the number of bytes that matters but the availability of a readable shared memory entry. That entry may be empty!

One might hope that startCaching() itself is not called until all response headers are written to local memory (from where MemStore would copy them to shared memory), but at least one StoreEntry::swapOut() caller does not seem to care about our `ENTRY_FWD_HDR_WAIT` hack, and MemStore itself does not care about that flag at all. AFAICT, StoreEntry::swapOut() can be called for an ENTRY_FWD_HDR_WAIT entry without any headers in local memory.

More importantly, there is no enforced invariant (AFAICT) that would preclude some (current or future) code from making a memory cache entry available for reading before it has full HTTP response headers. If there is such an invariant, then we should probably fix MemStore code, but I would argue that such a fix is best developed _after_ this PR because the problem (if any) is far from trivial and is mostly unrelated to this PR focus/changes.

FWIW, I see no such invariant in MemStore::startCaching() and copyToShm() code. As mentioned earlier, current MemStore writing code appears to not care about HTTP header boundary (in a sense related to this discussion) and higher-level swapOut() code might run quite early, before the headers are written to local memory.

eduard-bagdasaryan: > MemStore::startCaching() does create readable shared memory entries that have no HTTP bytes (yet).

I checked the code and think that this is correct: though copyToShm() call follows startCaching(), these both operations are separate (not atomic) - another process may start reading between these two calls.

rousskov:
@yadij, please note that this important PR has been waiting for your review for more than a month.

yadij (approved):

PR #1343, opened 2023-05-07
Do not populate StoreEntry basics from Transients

rousskov:
client_side_reply.cc:1078: "http->storeEntry()->objectLen() >= 0"
    inside clientReplyContext::storeOKTransferDone()

Since commit 24c9378, Transients are not supposed to maintain "basic"
cache entry information (i.e. StoreMapAnchor::Basics a.k.a. StoreEntry
STORE_META_STD fields). Using often-at-their-defaults Transient basics
may unexpectedly erase valuable StoreEntry information obtained from one
of the caches (e.g., swap_file_sz) and lead to assertions.

Consequently, do not populate unused Transients entry basics either.

Eventually, we may find a good way to parameterize StoreMapAnchor to
avoid keeping those unused fields in Transients StoreMap.

### DIFF
--- a/src/MemObject.h
+++ b/src/MemObject.h
@@ -169,6 +169,20 @@ public:
     class XitTable
     {
     public:
+        /// associate our StoreEntry with a Transients entry at the given index
+        void open(const int32_t anIndex, const Io anIo)
+        {
+            index = anIndex;
+            io = anIo;
+        }
+
+        /// stop associating our StoreEntry with a Transients entry
+        void close()
+        {
+            index = -1;
+            io = Store::ioDone;
+        }
+
         int32_t index = -1; ///< entry position inside the in-transit table
         Io io = ioUndecided; ///< current I/O state
     };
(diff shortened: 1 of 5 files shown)


rousskov:
High-load Web Polygraph tests occasionally trigger this assertion, and we have collected enough breadcrumbs to identify the likely underlying bug. However, we do not know the _exact_ scenario that triggers it. One scenario that is not real/possible but may be _close_ to what is actually going on is described [elsewhere](https://github.com/measurement-factory/squid/commit/7ba699b). A similar real scenario would explain why these assertions are rare even though the buggy code itself is used all the time -- the SMP race condition required for a known StoreEntry::swap_file_sz to disappear due to this bug ought to be very rare.

PR branch commit cc718a4 focuses on the bug fix itself -- you may want to start there if you want to study the bug itself. All other current changes (i.e. the fixup commit c41dc48) cleanup no-longer-used function parameters and inline a now-trivial function.

PR #1345, opened 2023-05-09
Do not apply custom debugs() format to Debug::Extra lines

rousskov:
Some debugs() callers set and leave custom std::ostream formats like
std::hex. Such behavior should be considered natural/correct because,
from those callers point of view, they see and control the entire
debugging message. Stream formats obviously must not cross debugs()
boundaries, and, from those callers point of view, those boundaries are
the message they explicitly call with. Here is a hypothetical example:

    debugs(33, DBG_CRITICAL, "Invalid QoS flag: " << std::hex << flag);

When we automatically add extra information to debugging messages (e.g.,
to detail which transaction has failed, what listening port is
misconfigured, or where the exception was thrown from), the "extra"
lines are added in a completely different code context. That detailing
code also sees a seemingly isolated printing context and should not be
expected to remember to reset debugging stream flags "just in case":

    os << Debug::Extra << "listening port: " << s.port();

The combination of natural code expectations, may result in very
misleading messages because, for example, hexadecimal numbers may look
just like decimal ones. We have seen that happening to transaction IDs
and here is an example based on the above debugs() that complains about
an invalid QoS flag on http_port 80 (but logs port 50):

    2023/05/09 16:44:13| Invalid QoS flag: deadbeef
        listening port: 50

This change would break any _deliberate_ format "bleeds" from primary
debugs() messages to Debug::Extra lines, but no such bleeding cases are
known, and they are too confusing/fragile to be encouraged/supported.

### DIFF
--- a/src/debug/Stream.h
+++ b/src/debug/Stream.h
@@ -62,7 +62,6 @@ public:
         friend class DebugMessageHeader;
 
         void rewind(const int aSection, const int aLevel);
-        void formatStream();
         Context *upper; ///< previous or parent record in nested debugging calls
         std::ostringstream buf; ///< debugs() output sink
         bool forceAlert; ///< the current debugs() will be a syslog ALERT
@@ -110,8 +109,9 @@ public:
     /// configures the active debugging context to write syslog ALERT
     static void ForceAlert();
 
-    /// prefixes each grouped debugs() line after the first one in the group
-    static std::ostream& Extra(std::ostream &os) { return os << "\n    "; }
+    /// Prefixes each grouped debugs() line after the first one in the group.
+    /// Resets debugging stream format to the initial/default one.
+    static std::ostream& Extra(std::ostream &os);
 
     /// silently erases saved early debugs() messages (if any)
     static void ForgetSaved();
@@ -170,6 +170,7 @@ public:
     static void SettleSyslog();
 
 private:
+    static void FormatStream(std::ostream &);
     static void LogMessage(const Context &);
 
     static Context *Current; ///< deepest active context; nil outside debugs()
--- a/src/debug/Stream.h
+++ b/src/debug/Stream.h
@@ -111,7 +111,7 @@ public:
 
     /// Prefixes each grouped debugs() line after the first one in the group.
     /// Resets debugging stream format to the initial/default one.
-    static std::ostream& Extra(std::ostream &os);
+    static std::ostream& Extra(std::ostream &);
 
     /// silently erases saved early debugs() messages (if any)
     static void ForgetSaved();
--- a/src/debug/debug.cc
+++ b/src/debug/debug.cc
@@ -1277,7 +1277,7 @@ Debug::Context::Context(const int aSection, const int aLevel):
     forceAlert(false),
     waitingForIdle(false)
 {
-    formatStream();
+    FormatStream(buf);
 }
 
 /// Optimization: avoids new Context creation for every debugs().
@@ -1292,14 +1292,12 @@ Debug::Context::rewind(const int aSection, const int aLevel)
 
     buf.str(CompiledDebugMessageBody());
     buf.clear();
-    // debugs() users are supposed to preserve format, but
-    // some do not, so we have to waste cycles resetting it for all.
-    formatStream();
+    FormatStream(buf);
 }
 
 /// configures default formatting for the debugging stream
 void
-Debug::Context::formatStream()
+Debug::FormatStream(std::ostream &buf)
 {
     const static std::ostringstream cleanStream;
     buf.flags(cleanStream.flags() | std::ios::fixed);
@@ -1307,6 +1305,17 @@ Debug::Context::formatStream()
     buf.precision(2);
     buf.fill(' ');
     // If this is not enough, use copyfmt(cleanStream) which is ~10% slower.
+}
+
+std::ostream &
+Debug::Extra(std::ostream &os)
+{
+    // Prevent previous line formats bleeding onto this line: Previous line code
+    // may not even be aware of some detailing code automatically adding extras.
+    FormatStream(os);
+
+    os << "\n    ";
+    return os;
 }
 
 void
--- a/src/tests/stub_debug.cc
+++ b/src/tests/stub_debug.cc
@@ -40,6 +40,16 @@ _db_rotate_log(void)
 {}
 
 void
+Debug::FormatStream(std::ostream &buf)
+{
+    const static std::ostringstream cleanStream;
+    buf.flags(cleanStream.flags() | std::ios::fixed);
+    buf.width(cleanStream.width());
+    buf.precision(2);
+    buf.fill(' ');
+}
+
+void
 Debug::LogMessage(const Context &context)
 {
     if (context.level > DBG_IMPORTANT)
@@ -51,6 +61,14 @@ Debug::LogMessage(const Context &context)
     fprintf(stderr, "%s| %s\n",
             "stub time", // debugLogTime(current_time),
             context.buf.str().c_str());
+}
+
+std::ostream &
+Debug::Extra(std::ostream &os)
+{
+    FormatStream(os);
+    os << "\n    ";
+    return os;
 }
 
 bool Debug::StderrEnabled() STUB_RETVAL(false)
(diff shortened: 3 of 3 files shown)


rousskov:
@eduard-bagdasaryan, I hope this fix addresses the problem with misleading logged transaction IDs that you have been bitten by recently.

kinkie:
> and here is an example based on the above debugs() that complains about
> an invalid QoS flag on http_port 80

Would it be possible to include an example of the output of this same message after the proposed code change?

src/debug/Stream.h
```diff
@@ -110,8 +109,9 @@ class Debug
     /// configures the active debugging context to write syslog ALERT
     static void ForceAlert();
 
-    /// prefixes each grouped debugs() line after the first one in the group
-    static std::ostream& Extra(std::ostream &os) { return os << "\n    "; }
+    /// Prefixes each grouped debugs() line after the first one in the group.
+    /// Resets debugging stream format to the initial/default one.
+    static std::ostream& Extra(std::ostream &os);
```
eduard-bagdasaryan, line 114:
```suggestion
    static std::ostream& Extra(std::ostream &);
```

eduard-bagdasaryan:
> I hope this fix addresses the problem with misleading logged transaction IDs

I am OK with your changes.

> no such bleeding cases are known

I quickly looked through Debug::Extra users and have not found such cases.

rousskov:
I cannot figure out why Jenkins [fails](https://build.squid-cache.org/job/5-pr-test/5629/) to test this PR. My "Retry" and "Matrix Reloaded" attempts have [failed](https://build.squid-cache.org/job/5-pr-test/5630/) with similar (and seemingly bogus) errors referencing the latest PR commit (done via GitHub UI):

```
Aborting build: 6ce51c8f6ac163fb44d1ad0628053438fad3ae5e has not been merged into master
```

FWIW, the previous Jenkins tests (before that PR branch commit) were [successful](https://build.squid-cache.org/job/5-pr-test/5627/).

@kinkie, if you know how to move this Jenkins testing forward, please help.

kinkie:
> I cannot figure out why Jenkins [fails](https://build.squid-cache.org/job/5-pr-test/5629/) to test this PR. My "Retry" and "Matrix Reloaded" attempts have [failed](https://build.squid-cache.org/job/5-pr-test/5630/) with similar (and seemingly bogus) errors referencing the latest PR commit (done via GitHub UI):
> 
> ```
> Aborting build: 6ce51c8f6ac163fb44d1ad0628053438fad3ae5e has not been merged into master
> ```
> 
> FWIW, the previous Jenkins tests (before that PR branch commit) were [successful](https://build.squid-cache.org/job/5-pr-test/5627/).
> 
> @kinkie, if you know how to move this Jenkins testing forward, please help.

This blocker is due to custom logic in our pre-environment setup script:

echo "Verify commit hash ..."
if [ "$sha1" = "$ghprbActualCommit" ]; then
   echo "Aborting build: $ghprbActualCommit has not been merged into $ghprbTargetBranch"
   exit 1
fi

I do not know the purpose of that code; I would expect that squashing the commits and force-pushing will reset things end work around the issue.

eduard-bagdasaryan:
> why Jenkins [fails](https://build.squid-cache.org/job/5-pr-test/5629/) to test this PR

This failure looks strange to me. The build [output](https://build.squid-cache.org/job/5-pr-test/5629/console) had the following line:

> GitHub pull request #1345 of commit 6ce51c8f6ac163fb44d1ad0628053438fad3ae5e, has merge conflicts.

But three lines below it reports that PR "now has NO merge conflicts".
Conflicts may explain the [ "$sha1" = "$ghprbActualCommit" ] check failure, because in this case $sha1 is the [head](https://github.com/jenkinsci/ghprb-plugin/issues/563#issuecomment-325833595) of the PR branch, and not the PR merge commit, as it should be. On the other hand, GitHub does not show any conflicts in this PR.

rousskov:
> Eduard: $sha1 is the head of the PR branch, and not the PR merge commit, as it should be.

Based on sha1 and ghprbActualCommit definitions in Jenkins ghprb-plugin [README](https://github.com/jenkinsci/ghprb-plugin/blob/ghprb-1.42.2/README.md#creating-a-job) (that you effectively linked to, thank you!), the error message in the custom Jenkins script is wrong or very misleading. I suggest the following fix:

```diff
- echo "Aborting build: $ghprbActualCommit has not been merged into $ghprbTargetBranch"
+ echo "ERROR: Cannot find a GitHub merge commit for this $ghprbTargetBranch PR.
+ echo "Aborting build: Will not test PR branch (commit $ghprbActualCommit) in isolation."
```

@eduard-bagdasaryan, if that GitHub merge commit actually exists, please try to figure out why the plugin(?) cannot find it. It may help us to automate a workaround for this problem. If that commit does not exist, then there is nothing we can do for this plugin configuration (besides fixing the script error message as discussed above).

We do test the staged commit later, but, IIRC, that set of tests is different from the set of tests that ghprb-plugin uses. To avoid being dependent on GitHub merge commit, we could switch ghprb-plugin from testing the GitHub merge commit to testing the PR branch itself, but, if my recollection is correct, such a switch would make ghprb-plugin tests slightly less applicable to the targeted branch than they are today (in cases where the PR branch is not based on the latest targeted branch). It might also introduce an additional different with GitHub Actions tests -- I have not checked, but I presume those tests test GitHub merge commit instead of the latest PR branch commit.

kinkie:
In recent days Gihub had reliability issues on their API. Could it be that we are chasing a glitch?

rousskov:
> it looks like problems were caused by the PR 'mergeable' flag instability

> Currently, ... 'mergeable' flag is true for this PR

@eduard-bagdasaryan, let's assume that all of the above is true. Given your explanation, I would expect the plugin to successfully start testing this PR now. However, my attempt has [failed](https://build.squid-cache.org/job/5-pr-test/5639/). Can you tell why it fails? Or am I misinterpreting your explanation as implying that it should succeed?

rousskov:
> In recent days Gihub had reliability issues on their API. Could it be that we are chasing a glitch?

I assume GitHub behavior has triggered this problem. This has happened many times before and will happen again. My focus is on things we can actually control (rather than GitHub hiccups themselves): I am trying to see whether we can automatically handle such expected problems better, reducing long-term waste related to investigations, test reruns, delayed merges, frustrated new PR contributors, etc., etc.

eduard-bagdasaryan:
> my attempt has failed

The [log](https://build.squid-cache.org/job/5-pr-test/5639/console) does not contain PR builder debugging - It looks like that manual triggering build does not activate the plugin, which is normally triggered via a web hook.

rousskov:
Resending the event did not any visible (to me) effect. I am now trying to rebuild using Jenkins plugin command:

rebuild

rousskov:
rebuild

rousskov:
The plugin `rebuild` command [worked](https://build.squid-cache.org/job/5-pr-test/5643/).

BTW, that sha1 plugin variable/parameter that we discussed earlier, is not always a commit SHA, so that custom script test is not always comparing one SHA with another:

```
18:08:59 + [ origin/pr/1345/merge = 6ce51c8f6ac163fb44d1ad0628053438fad3ae5e ]
```

The script message change I [suggested](https://github.com/squid-cache/squid/pull/1345#issuecomment-1544417988) earlier should still work.

In summary, there are two positive outcomes here: We can fix the error message in the custom script, and we have discovered a Jenkins plugin command that can rebuild PRs without fake commits or unnecessary merges (that would, among other things, erase any prior PR approvals!). On the negative side, we still do not know how to _automate_ waiting for the merge commit creation.

eduard-bagdasaryan:
@kinkie , if you don't mind, I am going to apply Alex' suggestions on Jenkins.
Also, do you have any ideas why the latest Jenkins build failed (for one node): `fatal: unable to access 'https://github.com/kinkie/support-tools.git/': gnutls_handshake() failed: The TLS connection was non-properly terminated.`

kinkie:
> @kinkie , if you don't mind, I am going to apply Alex' suggestions on Jenkins. Also, do you have any ideas why the latest Jenkins build failed (for one node): `fatal: unable to access 'https://github.com/kinkie/support-tools.git/': gnutls_handshake() failed: The TLS connection was non-properly terminated.`

Go right ahead, no need to wait for my permission.
re: the failure. No idea. I have seen some failures to connect to github at times during builds. It might be github, or it might be some deep packet inspection snafu at my ISP, who knows.

eduard-bagdasaryan:
rebuild

PR #1346, opened 2023-05-10
Do not leak Security::CertErrors created in X509_verify_cert()

rousskov:
ACLFilledChecklist was using a raw C pointer for handling cbdata-managed
Security::CertErrors. Some sslErrors users knew about hidden cbdata
requirements, some did not, resulting in inconsistent locking/unlocking
and associated memory leaks. Upgrading ACLFilledChecklist::sslErrors to
smart CbcPointer fixes those leaks (and simplifies code).

### DIFF
--- a/src/acl/FilledChecklist.h
+++ b/src/acl/FilledChecklist.h
@@ -87,11 +87,12 @@ public:
     char *snmp_community;
 #endif
 
+    // TODO: RefCount errors; do not ignore them because their "owner" is gone!
     /// TLS server [certificate validation] errors, in undefined order.
     /// The errors are accumulated as Squid goes through validation steps
     /// and server certificates. They are cleared on connection retries.
     /// For sslproxy_cert_error checks, contains just the current/last error.
-    const Security::CertErrors *sslErrors;
+    CbcPointer<Security::CertErrors> sslErrors;
 
     /// Peer certificate being checked by ssl_verify_cb() and by
     /// Security::PeerConnector class. In other contexts, the peer
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1524,11 +1524,10 @@ bool ConnStateData::serveDelayedError(Http::Stream *context)
             bool allowDomainMismatch = false;
             if (Config.ssl_client.cert_error) {
                 ACLFilledChecklist check(Config.ssl_client.cert_error, nullptr);
-                check.sslErrors = new Security::CertErrors(Security::CertError(SQUID_X509_V_ERR_DOMAIN_MISMATCH, srvCert));
+                const auto sslErrors = std::make_unique<Security::CertErrors>(Security::CertError(SQUID_X509_V_ERR_DOMAIN_MISMATCH, srvCert));
+                check.sslErrors = sslErrors.get();
                 clientAclChecklistFill(check, http);
                 allowDomainMismatch = check.fastCheck().allowed();
-                delete check.sslErrors;
-                check.sslErrors = nullptr;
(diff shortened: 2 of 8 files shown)


src/client_side.cc
```diff
@@ -1524,11 +1524,10 @@ bool ConnStateData::serveDelayedError(Http::Stream *context)
             bool allowDomainMismatch = false;
             if (Config.ssl_client.cert_error) {
                 ACLFilledChecklist check(Config.ssl_client.cert_error, nullptr);
-                check.sslErrors = new Security::CertErrors(Security::CertError(SQUID_X509_V_ERR_DOMAIN_MISMATCH, srvCert));
+                const auto sslErrors = std::make_unique<Security::CertErrors>(Security::CertError(SQUID_X509_V_ERR_DOMAIN_MISMATCH, srvCert));
```
rousskov, line 1527: This `std::make_unique<CertErrors>(CertError(...))` code is fairly basic but duplicated. I did not add a function to encapsulate it, primarily because we should be reference-counting these errors _instead_ (and implementing that requires serious out-of-scope changes). I did add a TODO about reference counting.

src/acl/FilledChecklist.h
```diff
@@ -87,11 +87,12 @@ class ACLFilledChecklist: public ACLChecklist
     char *snmp_community;
 #endif
 
+    // TODO: RefCount errors; do not ignore them because their "owner" is gone!
     /// TLS server [certificate validation] errors, in undefined order.
     /// The errors are accumulated as Squid goes through validation steps
     /// and server certificates. They are cleared on connection retries.
     /// For sslproxy_cert_error checks, contains just the current/last error.
-    const Security::CertErrors *sslErrors;
+    CbcPointer<Security::CertErrors> sslErrors;
```
rousskov, line 95: The loss of `const` here and elsewhere is unfortunate but necessary because CbcPointer calls non-const toCbdata().

rousskov (commented):
This PR is related to a squid-dev [thread](http://lists.squid-cache.org/pipermail/squid-dev/2023-March/009862.html).

src/client_side.cc
yadij: Given that the new type of `check.sslErrors` performs cdata/reference counting, why the additional `std::make_unique`?

src/ssl/ServerBump.cc
```diff
@@ -62,13 +62,12 @@ Ssl::ServerBump::attachServerSession(const Security::SessionPointer &s)
     serverSession = s;
 }
 
-const Security::CertErrors *
+Security::CertErrors *
```
yadij, line 65: This looks like it may be another source of leak risks. Can it become a `Pointer` reference yet?

rousskov: This is a stored errors _accessor_ method rather than errors sharing/extracting/making method. If errors were guaranteed to exist, then the method would just return a reference to CertErrors instead of a pointer. The pointer is (naturally) used to be able to indicate lack of errors.

I do not see a leak risk in this particular API, but if somebody were to change this method return type to avoid raw pointers, then the adjusted method would return something like `std::optional< std::reference_wrapper<CertErrors> >`. A raw pointer is a much simpler equivalent of that construct!

src/client_side.cc
yadij: I see. Thank you.

yadij (approved):

yadij:
No more v5 releases expected. Stop auto-backport attempts.

PR #1347, opened 2023-05-11
Fix store_client caller memory leak on certain errors

rousskov:
When a storeUnregister() code path destroys store_client before the
latter has a chance to deliver the answer, the cbdataReferenceDone()
call in store_client::finishCallback() is not reached, keeping the
callback data (e.g., clientReplyContext) alive forever. These
storeClientCopy() "cancellations" may happen, for example, when the
client-to-Squid connection is closed while store_client waits for Store.

Use CallbackData to guarantee cbdataReferenceDone() when store_client is
destructed before it can finishCallback(). These synchronous callbacks
will be replaced with AsyncCalls. For now, we use the "discouraged"
CallbackData API to accommodate the existing legacy callbacks.

Probably broken since 2002 commit fa80a8e.


src/store_client.cc
```diff
@@ -899,7 +897,7 @@ store_client::dumpStats(MemBuf * output, int clientNumber) const
     if (_callback.pending())
         return;
 
-    output->appendf("\tClient #%d, %p\n", clientNumber, _callback.callback_data);
+    output->appendf("\tClient #%d, %p\n", clientNumber, this);
```
rousskov, line 900: This adjustment changes which raw pointer is printed for each store_client in `mgr:objects`, `mgr:vm_objects`, and `mgr:openfd_objects` reports. The printed pointer can still be used to identify the corresponding storeClientCopy() caller in a debugging log. In some sense, printing `this` is better because a single storeClientCopy() caller often owns multiple store_client objects at the same time, but that is not why I made this change. I made this otherwise unnecessary/out-of-scope change because CallbackData API does not expose a raw pointer value.

We could enhance CallbackData API, of course, but exposing that raw pointer is a bit dangerous, and we should refactor to give store_client a unique instance ID instead of using frequently recycled memory addresses anyway. That refactoring needs quite a bit of noisy out-of-scope code changes, including replacing all these ugly appendf() calls with std::ostream-based code.

src/store_client.cc
```diff
@@ -176,14 +176,12 @@ store_client::finishCallback()
 
     cmp_offset = result.offset + result.length;
     STCB *temphandler = _callback.callback_handler;
-    void *cbdata = _callback.callback_data;
+    const auto cbdata = _callback.cbData.validDone();
     _callback = Callback(nullptr, nullptr);
```
yadij, line 180: Please also fix this line.
```suggestion
    _callback = Callback();
```

That may expose the issues with C++11 initialization of function pointers we have hit elsewhere. Namely that the constructor needs to set them explicitly, not implicitly.

src/store_client.cc
yadij: FTR, displaying the raw-pointer value as an ID here is a security issue. It exposes operating system internal layouts to external viewers, violating OS memory protections. If there is any other abstraction we can use as a unique client ID that would be much better.
I do not insist on an ID mechanism being created yet, but if an alternative is already available please switch to that value instead of `this`.

yadij (changes requested):

yadij (approved):

PR #1348, opened 2023-05-12
Update libsspwin32

yadij:
Move library files to lib/sspi/ for better modularity.

Add SQUID_CHECK_WIN32_SSPI autoconf test for dependency
checking by helpers.

Add missing HAVE_FOO_H wrappers around includes.
Fixes Squid coding style compliance.

Remove unnecessary __cplusplus protections.

Remove many redundant includes from helpers code.

Remove CygWin hack which is unreachable with precompiler
conditions.

Fix several structure initialization compiler errors.

### DIFF
--- a/acinclude/win32-sspi.m4
+++ b/acinclude/win32-sspi.m4
@@ -5,11 +5,9 @@
 ## Please see the COPYING and CONTRIBUTORS files for details.
 ##
 
-# check if the Windows SSPI requirements are available and work.
-# first argument is the variable containing the result
-#   (will be set to "yes" or "no")
-# second argument is the commands to run on success
-#
+# Checks whether the Windows SSPI requirements are available and work.
+# Sets squid_cv_win32_sspi to "yes" or "no".
+# The argument is the commands to run on success.
 AC_DEFUN([SQUID_CHECK_WIN32_SSPI],[
   AC_CHECK_HEADERS([w32api/windows.h windows.h],[
     squid_cv_win32_sspi=yes
--- a/acinclude/win32-sspi.m4
+++ b/acinclude/win32-sspi.m4
@@ -11,13 +11,12 @@
 # second argument is the commands to run on success
 #
 AC_DEFUN([SQUID_CHECK_WIN32_SSPI],[
-  AC_CACHE_CHECK([whether Windows SSPI is available],[$1],[
-    AC_CHECK_HEADERS([w32api/windows.h windows.h],[
-      $1=yes
-      # optional headers
-      AC_CHECK_HEADERS([wchar.h tchar.h])
-      # required headers
-      AC_CHECK_HEADERS([ntsecapi.h security.h sspi.h],,[$1=no],[
+  AC_CHECK_HEADERS([w32api/windows.h windows.h],[
+    squid_cv_win32_sspi=yes
+    # optional headers
+    AC_CHECK_HEADERS([wchar.h tchar.h])
+    # required headers
+    AC_CHECK_HEADERS([ntsecapi.h security.h sspi.h],,[squid_cv_win32_sspi=no],[
 #define SECURITY_WIN32
 #if HAVE_WINDOWS_H
 #include <windows.h>
@@ -33,8 +32,7 @@ AC_DEFUN([SQUID_CHECK_WIN32_SSPI],[
 #if HAVE_SSPI_H
 #include <sspi.h>
 #endif
-      ])
     ])
   ])
-  AS_IF([test "x$$1" = "xyes"],[$2])
+  AS_IF([test "x$squid_cv_win32_sspi" = "xyes"],[$1])
 ])
--- a/acinclude/win32-sspi.m4
+++ b/acinclude/win32-sspi.m4
@@ -1,0 +1,40 @@
+## Copyright (C) 1996-2023 The Squid Software Foundation and contributors
+##
+## Squid software is distributed under GPLv2+ license and includes
+## contributions from numerous individuals and organizations.
+## Please see the COPYING and CONTRIBUTORS files for details.
+##
+
+# check if the Windows SSPI requirements are available and work.
+# first argument is the variable containing the result
+#   (will be set to "yes" or "no")
+# second argument is the commands to run on success
+#
+AC_DEFUN([SQUID_CHECK_WIN32_SSPI],[
+  AC_CACHE_CHECK([whether Windows SSPI is available],[$1],[
+    AC_CHECK_HEADERS([w32api/windows.h windows.h],[
+      $1=yes
+      # optional headers
+      AC_CHECK_HEADERS([wchar.h tchar.h])
+      # required headers
+      AC_CHECK_HEADERS([ntsecapi.h security.h sspi.h],,[$1=no],[
+#define SECURITY_WIN32
+#if HAVE_WINDOWS_H
+#include <windows.h>
+#elif HAVE_W32API_WINDOWS_H
+#include <w32api/windows.h>
+#endif
+#if HAVE_NTSECAPI_H
+#include <ntsecapi.h>
+#endif
+#if HAVE_SECURITY_H
+#include <security.h>
+#endif
+#if HAVE_SSPI_H
+#include <sspi.h>
+#endif
+      ])
+    ])
+  ])
+  AS_IF([test "x$$1" = "xyes"],[$2])
+])
--- a/lib/sspi/sspwin32.cc
+++ b/lib/sspi/sspwin32.cc
@@ -14,18 +14,20 @@
 #include "sspi/sspwin32.h"
 #include "util.h"
 
-#if HAVE_WINDOWS_H && HAVE_SSPI_H
-
 // FARPROC is an exception on Windows to the -Wcast-function-type sanity check.
 // suppress the warning only when casting FARPROC
 template <typename T>
 T
 farproc_cast(FARPROC in)
 {
+#if defined(__GCC__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wcast-function-type"
     return reinterpret_cast<T>(in);
 #pragma GCC diagnostic pop
+#else
+    return reinterpret_cast<T>(in);
+#endif
 }
 
 typedef struct _AUTH_SEQ {
@@ -611,5 +613,3 @@ const char * WINAPI SSP_ValidateNegotiateCredentials(PVOID PAutenticateBuf, int 
     return NULL;
 }
 #endif /* HAVE_AUTH_MODULE_NEGOTIATE */
-
-#endif /* HAVE_WINDOWS_H && HAVE_SSPI_H */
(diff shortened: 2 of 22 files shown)


yadij:
This is draft because for me the library does not yet build. PR #1085 changed the C-style casting `(X) GetProcAddress(...)` to `reinterpret_cast<X>(GetProcAddress(...))` but neither form builds for me in GCC-12 cross-compile. That should be resolved before this merges, but not required - we have a lot of good progress here either way.

yadij:
Also, this requires PR #1341 to be merged first (changes are included here).

rousskov (changes requested):
> Also, this requires PR https://github.com/squid-cache/squid/pull/1341 to be merged first 

Agreed. Please ping me when that is settled, and I will review.

yadij:
Since PR #1341 is seemingly stalled on idealistic pedantry I have rewritten this PR to not depend on it.

kinkie:
Would you mind including some output from a configure run on mingw? The code looks good enough to me but it's the easiest way to double check it :)

acinclude/win32-sspi.m4
```diff
@@ -0,0 +1,40 @@
...
+## Please see the COPYING and CONTRIBUTORS files for details.
+##
+
+# check if the Windows SSPI requirements are available and work.
+# first argument is the variable containing the result
+#   (will be set to "yes" or "no")
+# second argument is the commands to run on success
+#
+AC_DEFUN([SQUID_CHECK_WIN32_SSPI],[
+  AC_CACHE_CHECK([whether Windows SSPI is available],[$1],[
```
rousskov, line 14: PR build [log](https://github.com/squid-cache/squid/suites/13813788092/artifacts/766560267):

```
checking whether Windows SSPI is available... checking for w32api/windows.h... no
checking for windows.h... (cached) no

configure: helper auth/basic/SSPI ... found but cannot be built
```

I see a few problems in the above output (but please ignore any specifics that are inaccurate -- no need to argue about these low-level details):

1. The first line appears to contain what should be two separate lines.
2. The outcome for "checking whether Windows SSPI is available..." (or "checking for w32api/windows.h...") check is not visible.
3. There is an empty line after "checking for windows.h... (cached) no"

Official build [log](https://github.com/squid-cache/squid/suites/13796614278/artifacts/765425688) looks much better:

```
checking for w32api/windows.h... no
checking for windows.h... (cached) no
configure: helper auth/basic/SSPI ... found but cannot be built
```

Please fix and check for other problems by comparing "before" and "after" ./configure output.

lib/sspi/sspwin32.cc
```diff
@@ -9,9 +9,23 @@
...
 #include "ntlmauth/ntlmauth.h"
-#include "sspwin32.h"
+#include "sspi/sspwin32.h"
 #include "util.h"
 
+#if HAVE_WINDOWS_H && HAVE_SSPI_H
+
+// FARPROC is an exception on Windows to the -Wcast-function-type sanity check.
+// suppress the warning only when casting FARPROC
+template<typename T>
```
rousskov, line 19: To improve (IMO) readability, I recommend using the following (more prevalent in Squid code) formatting for this line:

```suggestion
template <typename T>
```

lib/sspi/sspwin32.cc
```diff
@@ -9,9 +9,23 @@
...
+T
+farproc_cast(FARPROC in)
```
rousskov, line 21: In C++, `foo_cast<>` means "cast to foo". This function does not cast to FARPROC. It casts _from_ FARPROC.

I would also use CamelCase for this helper name so that the readers of the code using it know that it is not a standard cast:

```suggestion
FarProcToFunction(FARPROC in)
```

or, to follow Windows naming scheme,

```suggestion
FarProcToProc(FARPROC in)
```

or, simply,

```suggestion
ToFunction(FARPROC in)
```

or

```suggestion
ToProc(FARPROC in)
```

As usual, I do not insist on any specific spelling.

lib/sspi/sspwin32.cc
```diff
@@ -9,9 +9,23 @@
...
+{
+#pragma GCC diagnostic push
```
rousskov, line 23:
```
warning C4068: unknown pragma 'GCC'
```

To protect us from compiler warnings similar to the above, please wrap these pragmas in #ifdefs. There are examples in the official code.

AFAICT, we can assume that modern clang and GCC versions support these pragmas but compilers like MSVC [do not](https://godbolt.org/z/W51oT6rKn). Adding a `__GNUC__` wrapper helps in my tests, but there may be a better way.

rousskov (changes requested):

lib/sspi/sspwin32.cc
yadij: > In C++, `foo_cast<>` means "cast to foo". 

WTF? that is a strange interpretation. It does not make any sense when applied to the standard X_cast operators.

It is supposed to be  ```[mechanism]_cast<foo>(expression)``` in this case the mechanism is Windows FARPROC type alias (inheritance, interfacing, puny-type, functor, function pointer, ... whatever it is).

> This function does not cast to FARPROC. It casts _from_ FARPROC.

Indeed, it casts toward the **type** template parameter.

lib/sspi/sspwin32.cc
yadij: Where did you get that warning from?

The ```GCC``` prefix is supposed to be standard way of defining the rest of the line as GCC-only and to be ignored by all other C++ compilers.

acinclude/win32-sspi.m4
yadij: Fixed in [de3c084](https://github.com/squid-cache/squid/pull/1348/commits/de3c0845687cb525af643f1b55fec2870180ed3a)

Output now looks like:
```
checking for w32api/windows.h... no
checking for windows.h... (cached) yes
checking for wchar.h... (cached) yes
checking for tchar.h... yes
checking for ntsecapi.hhh... no
checking for security.h... yes
checking for sspi.h... yes
configure: helper auth/basic/SSPI ... found but cannot be built
```

lib/sspi/sspwin32.cc
rousskov: > Alex: In C++, foo_cast<> means "cast to foo". This function does not cast to FARPROC. It casts from FARPROC.

> Amos: It is supposed to be `[mechanism]_cast<foo>(expression)`

Agreed. I was wrong -- I went too far in that generalization. Sorry.

I still think we should use CamelCase for this helper name so that the readers of the code using it know that it is not a standard cast, but I do not insist on this change.

lib/sspi/sspwin32.cc
rousskov:
```
warning C4068: unknown pragma 'GCC'
```
 
> > Alex: AFAICT, we can assume that modern clang and GCC versions support these pragmas but compilers like MSVC [do not](https://godbolt.org/z/W51oT6rKn).

> Amos: Where did you get that warning from?

The problem is general, but that particular warning came from an MSVC test that was linked from the original change request: https://godbolt.org/z/W51oT6rKn

> The GCC prefix is supposed to be standard way of defining the rest of the line as GCC-only and to be ignored by all other C++ compilers.

It _is_ ignored, with a warning.

As for GCC prefix being "standard way", I see no such language/mechanism in pragma [documentation](https://en.cppreference.com/w/cpp/preprocessor/impl). There is a requirement to _ignore_ pragmas that are not recognized. AFAICT, that requirement does not prohibit issuing warnings (or supporting GCC pragmas in a non-GCC compiler for that matter).

BTW, these problems with pragmas is one of the many examples illustrating why I consider defaulting to `-Werror` a mistake, but this change request is not about that.

acinclude/win32-sspi.m4
```diff
@@ -0,0 +1,38 @@
...
+## Squid software is distributed under GPLv2+ license and includes
+## contributions from numerous individuals and organizations.
+## Please see the COPYING and CONTRIBUTORS files for details.
+##
+
+# check if the Windows SSPI requirements are available and work.
+# first argument is the variable containing the result
+#   (will be set to "yes" or "no")
+# second argument is the commands to run on success
+#
```
rousskov, line 12: The description got stale:

```suggestion
# Checks whether the Windows SSPI requirements are available and work.
# Sets squid_cv_win32_sspi to "yes" or "no".
# The argument is the commands to run on success.
```

The above suggestion also removes an empty trailing line in that comment. I do not insist on that removal because we are very inconsistent with regard to these comments styling (including `#`, `##`, and `dnl` variants and these empty lines), but I think it is better to converge towards no empty lines because our C++ descriptions do not use them and because they introduce an arguably unwanted distance from the thing being described, increasing the chances of description/code separation.

I also recommend rephrasing "requirements are available and work", but I do not know enough about SSPI to suggest specific improvements -- the macro code does not make the "work" part obvious (to me).

acinclude/win32-sspi.m4
```diff
@@ -0,0 +1,38 @@
...
+AC_DEFUN([SQUID_CHECK_WIN32_SSPI],[
+  AC_CHECK_HEADERS([w32api/windows.h windows.h],[
+    squid_cv_win32_sspi=yes
```
rousskov, line 15: If possible, please move this assignment inside the second AC_CHECK_HEADERS(). I assume the code "works" as is, but this assignment looks like the wrong "default" here IMHO.

rousskov (changes requested):

lib/sspi/sspwin32.cc
yadij: Fixed in [9a918aa](https://github.com/squid-cache/squid/pull/1348/commits/9a918aaaacb1a9d5c0a7c834a591c7cfaf5a6005)

yadij:
I think we are up-to-date now. Please re-review.

acinclude/win32-sspi.m4
rousskov: Right, I only looked at the lines below the `yes` assignment, so my "second" AC_CHECK_HEADERS is your "third" -- we are talking about the same AC_CHECK_HEADERS() line that is described in PR code as "required headers".

> the if-found and if-not-found cases get run for each of them.

Ah, I did not realize that! I consider this change request addressed, thank you.

rousskov (commented):
> PR description: This adds AC_CACHE to optimize build performance.

Please update the PR description to match the current code. I did not check the whole description, but the sentence quoted above is stale.

I see no problems in the current PR code, but I have not checked most of the low-level details.

yadij:
Description refreshed.

@rousskov, are you wishing to do that low-level review? if not please mark as cleared.

rousskov:
> @rousskov, are you wishing to do that low-level review? if not please mark as cleared.

Cleared for merge. FWIW, my "did not review X" disclaimers do not imply that I intend to review X (unless I explicitly say otherwise; in which case I would block the PR or self-request a review). That disclaimer is there primarily to let the PR author know that they are on their own with regard to X.

PR #1350, opened 2023-05-15
Rename class ACL to Acl::Node

yadij:
Also a lot of style cleanup on touched code
and lines documenting class ACL

### DIFF
--- a/src/acl/Gadgets.cc
+++ b/src/acl/Gadgets.cc
@@ -30,7 +30,7 @@
 #include <set>
 #include <algorithm>
 
-typedef std::set<ACL*> AclSet;
+using AclSet = std::set<Acl::Node *>;
 /// Accumulates all ACLs to facilitate their clean deletion despite reuse.
 static AclSet *RegisteredAcls; // TODO: Remove when ACLs are refcounted
 
@@ -76,9 +76,7 @@ aclIsProxyAuth(const char *name)
 
     debugs(28, 5, "aclIsProxyAuth: called for " << name);
 
-    ACL *a;
-
-    if ((a = ACL::FindByName(name))) {
+    if (const auto a = Acl::Node::FindByName(name)) {
         debugs(28, 5, "aclIsProxyAuth: returning " << a->isProxyAuth());
         return a->isProxyAuth();
     }
@@ -218,7 +216,7 @@ aclParseAclList(ConfigParser &, Acl::Tree **treep, const char *label)
 }
 
 void
-aclRegister(ACL *acl)
+aclRegister(Acl::Node *acl)
 {
     if (!acl->registered) {
         if (!RegisteredAcls)
@@ -231,7 +229,7 @@ aclRegister(ACL *acl)
 /// remove registered acl from the centralized deletion set
 static
 void
-aclDeregister(ACL *acl)
+aclDeregister(Acl::Node *acl)
 {
     if (acl->registered) {
         if (RegisteredAcls)
@@ -246,13 +244,13 @@ aclDeregister(ACL *acl)
 
 /// called to delete ALL Acls.
 void
-aclDestroyAcls(ACL ** head)
+aclDestroyAcls(Acl::Node ** head)
 {
     *head = nullptr; // Config.aclList
     if (AclSet *acls = RegisteredAcls) {
         debugs(28, 8, "deleting all " << acls->size() << " ACLs");
         while (!acls->empty()) {
-            ACL *acl = *acls->begin();
+            const auto acl = *acls->begin();
             // We use centralized deletion (this function) so ~ACL should not
             // delete other ACLs, but we still deregister first to prevent any
             // accesses to the being-deleted ACL via RegisteredAcls.
--- a/src/acl/Node.h
+++ b/src/acl/Node.h
@@ -6,10 +6,9 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-#ifndef SQUID_ACL_H
-#define SQUID_ACL_H
+#ifndef SQUID_ACL_NODE_H
+#define SQUID_ACL_NODE_H
 
-#include "acl/forward.h"
 #include "acl/Options.h"
 #include "cbdata.h"
 #include "defines.h"
@@ -26,7 +25,7 @@ namespace Acl {
 /// the ACL type name known to admins
 typedef const char *TypeName;
 /// a "factory" function for making ACL objects (of some ACL child type)
-typedef ACL *(*Maker)(TypeName typeName);
+typedef Node *(*Maker)(TypeName typeName);
 /// use the given ACL Maker for all ACLs of the named type
 void RegisterMaker(TypeName typeName, Maker maker);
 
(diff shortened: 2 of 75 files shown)


yadij:
This small piece of long overdue cleanup resolves a large number of collisions between our `class ACL` and Windows `struct ACL`.

src/acl/Acl.cc
```diff
@@ -306,13 +307,13 @@ ACL::ParseAclLine(ConfigParser &parser, ACL ** head)
...
 {
     return false;
 }
 
 void
-ACL::parseFlags()
+Acl::ACL::parseFlags()
 {
     Acl::Options allOptions = options();
     for (const auto lineOption: lineOptions()) {
```
rousskov, line 319: Please remove `Acl::` namespace prefixes from the old code that this PR moves inside that Acl namespace, like the allOptions declaration line above. This change request applies to all such occurrences.

src/acl/Acl.h
```diff
@@ -36,12 +36,9 @@ void RegisterMaker(TypeName typeName, Maker maker);
 /// Key comparison is case-insensitive.
 void SetKey(SBuf &keyStorage, const char *keyParameterName, const char *newKey);
 
-} // namespace Acl
-
 /// A configurable condition. A node in the ACL expression tree.
 /// Can evaluate itself in FilledChecklist context.
 /// Does not change during evaluation.
-/// \ingroup ACLAPI
 class ACL
```
rousskov, line 42: Please consider renaming the `ACL` class  to `Node` or `Condition` while we modify all these lines.
 
* The ACL name itself is not precise enough to describe what this class is about. Too many things are Access Control List-related in ACL context. We have a whole namespace with that name for them! And this class is not even a _List_ in most contexts (its invasive `next` field is _not_ used to build ACL lists for configuration directives -- we have Acl::InnerNode for that).

* The all-caps ACL spelling goes against our own CamelCase style that we do apply to abbreviations like HTTP and ICP. Many other related classes and the Acl namespace itself use that style.

* Avoiding a name spelling clash with Windows ACL typedef helps with code search and understanding.
 
If we have to change all these lines to fix Windows build, then let's address the class naming problem, so that we do not have to change them again. Renaming will add a few more changes, but it is probably worth it long-term.

I do not insist on renaming.

src/acl/Acl.h
```diff
@@ -26,7 +26,7 @@ namespace Acl {
 /// the ACL type name known to admins
 typedef const char *TypeName;
 /// a "factory" function for making ACL objects (of some ACL child type)
-typedef ACL *(*Maker)(TypeName typeName);
+typedef Acl::ACL *(*Maker)(TypeName typeName);
 /// use the given ACL Maker for all ACLs of the named type
```
rousskov, line 30: Please do not add unnecessary `Acl::` namespace prefixes inside the Acl namespace, like the Maker declaration line above. This change request applies to all such additions.

The current PR code relies on the fact that moving ACL inside the namespace already avoids the clash with Windows ACL typedef. We do not have to always prefix inside the namespace, but if we had to, the current PR code would not work (because it does not and _cannot_ always prefix).

src/acl/Gadgets.cc
```diff
@@ -79,9 +79,7 @@ aclIsProxyAuth(const char *name)
 
     debugs(28, 5, "aclIsProxyAuth: called for " << name);
 
-    ACL *a;
-
-    if ((a = ACL::FindByName(name))) {
+    if (auto a = Acl::ACL::FindByName(name)) {
```
rousskov, line 82:
```suggestion
    if (const auto a = Acl::ACL::FindByName(name)) {
```

I normally do not ask you to fix `const` in such contexts, but since moving `a` declaration is itself a change outside this PR scope, I think it is best to finish this out-of-scope polishing started by this PR in this PR. Same for similar `const` requests in this review. As usual, if some of them do not compile, then just paste a compiler error message to address them.

src/acl/Gadgets.cc
```diff
@@ -249,13 +247,13 @@ aclDeregister(ACL *acl)
...
 void
-aclDestroyAcls(ACL ** head)
+aclDestroyAcls(Acl::ACL ** head)
 {
     *head = nullptr; // Config.aclList
     if (AclSet *acls = RegisteredAcls) {
         debugs(28, 8, "deleting all " << acls->size() << " ACLs");
         while (!acls->empty()) {
-            ACL *acl = *acls->begin();
+            auto acl = *acls->begin();
```
rousskov, line 256:
```suggestion
            const auto acl = *acls->begin();
```

src/cache_cf.cc
```diff
@@ -4708,7 +4708,7 @@ static void parse_ftp_epsv(acl_access **ftp_epsv)
         *ftp_epsv = nullptr;
 
         if (ftpEpsvDeprecatedAction == Acl::Answer(ACCESS_DENIED)) {
-            if (ACL *a = ACL::FindByName("all"))
+            if (auto a = Acl::ACL::FindByName("all"))
```
rousskov, line 4711:
```suggestion
            if (const auto a = Acl::ACL::FindByName("all"))
```

src/external_acl.cc
```diff
@@ -1152,7 +1152,7 @@ ExternalACLLookup::checkForAsync(ACLChecklist *checklist)const
 {
     /* TODO: optimise this - we probably have a pointer to this
      * around somewhere */
-    ACL *acl = ACL::FindByName(AclMatchedName);
+    auto acl = Acl::ACL::FindByName(AclMatchedName);
```
rousskov, line 1155:
```suggestion
    const auto acl = Acl::ACL::FindByName(AclMatchedName);
```

rousskov (changes requested):
Thank you for working on this code improvement.

I wish this PR did not contain many out-of-scope polishing touches. I do not insist on undoing those changes, but I cannot be sure I have noticed/reviewed all of them either. Please avoid such polishing in future PRs.

src/acl/Acl.h
yadij: done

src/acl/Acl.cc
yadij: done as far as I can tell in latest code.

rousskov (changes requested):

I only insist on one followup change -- the one about the new XXX.

If possible, please also auto-format sources. [CI](https://github.com/squid-cache/squid/actions/runs/6372235008/job/17294843264?pr=1350) detects many "malformed" changes.

src/acl/Node.h
```diff
@@ -213,7 +211,6 @@ class acl_proxy_auth_match_cache
 
 /// \ingroup ACLAPI
 /// XXX: find a way to remove or at least use a refcounted ACL pointer
-extern const char *AclMatchedName;  /* NULL */
-
-#endif /* SQUID_ACL_H */
+extern const char *AclMatchedName;  /* nullptr */
 
+#endif /* SQUID_ACL_NODE_H */
```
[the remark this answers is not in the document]
kinkie: What if we just turned this on its head and moved to `#pragma once` ? 
It has been reportedly [supported by everyone](https://en.wikipedia.org/wiki/Pragma_once#Portability) for about 10 years, and would just make this whole point moot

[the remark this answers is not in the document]
rousskov: According to the [page](https://en.wikipedia.org/wiki/Pragma_once) you referenced, non-standard `#pragma once` comes with several significant caveats (besides the obvious lack of standardized universal support). Some of those caveats might be applicable to some Squid builds.

Include guards (i.e. the standard approach we have been using) seem to be an overall better alternative. I want to believe that we can agree on how to name new/changed ones -- it seems such a simple and non-controversial task! Right now, I do not see any significant signs that we cannot.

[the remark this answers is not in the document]
kinkie: yes, but.

In theory, there is no snandardised support. In practice all compilers support it.
In theory, there can be issues with symbolic links, in practice, we don't use symbolic links in our build chain.
Also, in theory it's easy to come up with consensus, in practice we're spending time on this

Let me see if I can come up with something automated as part of source-formatting, this could prove a viable alternate.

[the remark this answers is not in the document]
kinkie: Okay; PR #1630 can address this topic, hopefully for good, by automatically enforcing a standard

kinkie:
@yadij this has been addressed elsewhere I think. Can it be closed?

PR #1351, opened 2023-05-16
Maintenance: Remove a stale sponsor URL

kinkie:
SpinUp has ceased operations. Their URL is used by an unrelated entity.

### DIFF
--- a/SPONSORS.list
+++ b/SPONSORS.list
@@ -13,7 +13,7 @@ DigitalOcean - https://www.digitalocean.com/
 	DigitalOcean has donated droplets from their cloud infrastructure
 	to host most of Squid Project's continuous integration farm.
 
-SpinUp - https://SpinUp.com
+SpinUp
 
 	SpinUp has donated cloud resources to host our main website, wiki
 	and mailing lists.
--- a/SPONSORS.list
+++ b/SPONSORS.list
@@ -13,6 +13,11 @@ DigitalOcean - https://www.digitalocean.com/
 	DigitalOcean has donated droplets from their cloud infrastructure
 	to host most of Squid Project's continuous integration farm.
 
+SpinUp - https://SpinUp.com
+
+	SpinUp has donated cloud resources to host our main website, wiki
+	and mailing lists.
+
 The Measurement Factory - http://www.measurement-factory.com/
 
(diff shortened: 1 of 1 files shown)


SPONSORS.list
```diff
@@ -13,11 +13,6 @@ DigitalOcean - https://www.digitalocean.com/
 	DigitalOcean has donated droplets from their cloud infrastructure
 	to host most of Squid Project's continuous integration farm.
 
-SpinUp - https://SpinUp.com
-
-	SpinUp has donated cloud resources to host our main website, wiki
```
rousskov, line 18: > PR description: SpinUp has ceased operations

Their "_has_ donated" entry is currently located in the v6 section. I am OK with removing them from that `@Squid-6` section if they have not actually provided any resources since v5 was branched.

If they have provided resources for v5 or earlier releases, then they should be moved to the right blob rather than completed deleted AFAICT.

In either case, this PR would be effectively fixing a bug. I support that correction. I just do not know which action would restore a match with reality (but it is very unlikely that the current version does that).

kinkie: Let me be clear about the problem I'm trying to fix: that company doesn't exist anymore, and the URL we are referencing belongs to another, unrelated, entity.

In other words, the crediting is now moot, and while I visited the site and it's not dodgy, we are still promoting someone who is not the donor

yadij: Removing just the URL would be appropriate then.

rousskov: > Amos: Removing just the URL would be appropriate then.

Done (under the Easier Done Than Said experiment, although its primary rationale does not fully apply here). I have also update PR title/description accordingly.

@yadij, if you are OK with this updated PR, please clear it for merging. This PR still addresses the problem [described](https://github.com/squid-cache/squid/pull/1351#discussion_r1199568290) by Francesco. We can always come back to the bigger issue of stale entries in a dedicated PR, of course.

kinkie:
Oh wait, I can't approve 🤣

rousskov:
> Oh wait, I can't approve rofl

You have already approved it (implicitly; by authoring the PR). You just needed to clear the PR for merging. I have done that for you now.

PR #1352, opened 2023-05-17
Refactor low level I/O traffic accounting

yadij:
Move the FD read/write byte counting to the class fde
read/write operators instead of having the callers perform
manual accouting with fd_bytes().

This fixes traffic double-counting in FTP and
under-counting of UDP protocols, Disker I/O, possibly TLS
tunnels as well.

Also, resolves MinGW-w64 symbol collisions between winnt.h
definitions and FD_READ/RD_WRITE enums.


kinkie:
I like this!

src/dns_internal.cc
```diff
@@ -1334,6 +1327,7 @@ idnsRead(int fd, void *)
     while (max) {
         --max;
         len = comm_udp_recvfrom(fd, rbuf, SQUID_UDP_SO_RCVBUF, 0, bugbypass);
+        fd_table[fd].bytesRead(len);
```
rousskov, line 1330: Passing a negative `len` value to bytesRead(size_t) will create havoc. Please move this new line to where the old fd_bytes() call was. Please double check for similar bugs in the PR -- I have not.

src/fde.h
```diff
@@ -97,6 +112,12 @@ class fde
     /// record a transaction on this FD
     void noteUse() { ++pconn.uses; }
 
+    /// record bytes being read and return the accumulated total for this FD.
+    uint64_t bytesRead(size_t n=0) { if (n > 0) bytes_read += n; return bytes_read; }
```
rousskov, line 116: Please split this method into two, one to increment the counter and one to get the total. The split will simplify (and might even speed up) both methods!

* I suggest naming the first method noteBytesRead() or recordBytesRead(), but there may be better names.
* I would keep the name of the constant getter method the same -- bytesRead(), but totalBytesRead() also works OK IMO.
* To simplify searches and references, I would _not_ use the same name for both methods.

In the first method -- noteBytesRead() -- please do not waste cycles/ink on the positivity check: Zeros are rare and harmless in this context, while the `> 0` check may create a false impression that negative arguments are OK (when, in fact, they are caller bugs).

Same for bytesWritten(), of course.

src/comm.cc
```diff
@@ -131,6 +131,7 @@ comm_udp_recvfrom(int fd, void *buf, size_t len, int flags, Ip::Address &from)
     int x = recvfrom(fd, buf, len, flags, AI->ai_addr, &AI->ai_addrlen);
     from = *AI;
     Ip::Address::FreeAddr(AI);
+    fd_table[fd].bytesRead(x);
```
rousskov, line 134: I think we should not add this line into this low-level function because this low-level function cannot be certain that:

* comm_udp_recvfrom() is (and will always be) called for an fd_table-registered FD
* callers do not (and will never) call bytesRead() themselves
 
Instead, let the callers of low-level functions continue to update the byte counters until all such calls disappear behind fde or we find a better/safer way to address this uncertainty. Yes, we will miss a few calls, but that is a lesser evil (for now).

Please check for other similar cases in the PR -- I have not.

rousskov (changes requested):
Thank you for working on this improvement. I only had time for three critical issues, but this review should allow you to make progress.

P.S. Improving "const-correctness" for various new PR variables would be nice.

src/dns_internal.cc
yadij: No. The bytesRead() implementation explicitly handles negative values for exactly this type of case.

src/fde.h
yadij: Zeros may be harmless but negatives are not, and they are relatively common. See the original fd_bytes() implementation for comparision. Placing the if-condition inside this function de-duplicates all callers having to perform the identical check themselves prior to calling it.

kinkie: > Zeros may be harmless but negatives are not, and they are relatively common. See the original fd_bytes() implementation for comparision. Placing the if-condition inside this function de-duplicates all callers having to perform the identical check themselves prior to calling it.

Isn't size_t guaranteed to be unsigned?

rousskov: > Isn't size_t guaranteed to be unsigned?

Yes, it [is](https://en.cppreference.com/w/c/types/size_t), and the proposed methods should continue to use `size_t` (not `int` that was used by fd_bytes()).

And, yes, the callers should check the I/O result before interpreting it as the number of bytes read or written. Those checks are a part of the I/O APIs they are using.

yadij: Ah, that is what this is all about. Thank you @kinkie for pointing out the real problem.

src/comm.cc
yadij: One of the guarantees with FDs is that they are unique for the duration of their lifetime. If the FD is not registered the corresponding fde will be reserved and showing `closed` state. Recording traffic for them regardless of state is a good idea to avoid some of the bugs being fixed here recurring and harmless in the case the fde object was not actively `open`ed.

src/fde.h
yadij: fixed in [bb0ee01](https://github.com/squid-cache/squid/pull/1352/commits/bb0ee01abadf8c307baaff54043962e528ef2c2e)

src/dns_internal.cc
yadij: fixed in [bb0ee01](https://github.com/squid-cache/squid/pull/1352/commits/bb0ee01abadf8c307baaff54043962e528ef2c2e)

src/fde.h
```diff
@@ -84,8 +84,25 @@ class fde
     /// use I/O methods that maintain an internal-to-them buffer
     void useBufferedIo(READ_HANDLER *, WRITE_HANDLER *);
 
-    int read(int fd, char *buf, int len) { return readMethod_(fd, buf, len); }
-    int write(int fd, const char *buf, int len) { return writeMethod_(fd, buf, len); }
+    /// call the read method set by fde::setIo() requesting len more bytes.
```
rousskov, line 87: I recommend avoiding (re)documenting what that "read method" API is:

```suggestion
    /// call the read method set by setIo()
```

If you insist on (re)documenting, then please reduce the inaccuracy of the current (re)documentation attempt:

```suggestion
    /// call the read method set by setIo(), requesting to read up to len more bytes.
```

The above suggestions also avoid using Foo inside Foo.

src/fde.h
```diff
@@ -84,8 +84,25 @@ class fde
...
+    /// Performs I/O counting for all bytes received.
+    /// \returns the number of bytes actually received
```
rousskov, line 89: This function does not always return "the number of bytes actually received"

```suggestion
    /// \returns the read method result
```

src/fde.h
```diff
@@ -84,8 +84,25 @@ class fde
...
+        auto n = readMethod_(fd, buf, len);
+        if (n > 0)
+            bytesRead(n);
+        return n;
+    }
+
+    /// call the WRITE_HANDLER set by fde::setIo() with len more bytes.
+    /// Performs I/O counting for all bytes written.
+    /// \returns the number of bytes actually written
+    int write(int fd, const char *buf, int len) {
```
rousskov, line 100: Please apply read() changes to the write() method documentation as well.

Making `n` (and other variables) constant would be nice.

src/fde.h
```diff
@@ -97,6 +114,18 @@ class fde
     /// record a transaction on this FD
     void noteUse() { ++pconn.uses; }
 
+    /// record bytes being read and return the accumulated total for this FD.
+    void bytesRead(const size_t n) { bytes_read += n; }
+
+    /// \sa bytes_read
```
rousskov, line 120: Just to minimize the number of links documentation readers have to click to learn what this method does:

```suggestion
    /// \copydoc totalBytesRead_
```

The specific suggestion above accounts for the renaming suggestion in the other change request. Adjust as needed.

src/fde.h
```diff
@@ -97,6 +114,18 @@ class fde
...
+    uint64_t totalBytesRead() const { return bytes_read; }
+
+    /// record bytes being written and return the accumulated total for this FD.
+    void bytesWritten(const size_t n) { bytes_written += n; }
+
+    /// \sa bytes_written
```
rousskov, line 126:
```suggestion
    /// \copydoc totalBytesWritten_
```

src/fde.h
```diff
@@ -184,6 +210,9 @@ class fde
     // I/O methods connect Squid to the device/stack/library fde represents
     READ_HANDLER *readMethod_ = nullptr; ///< imports bytes into Squid
     WRITE_HANDLER *writeMethod_ = nullptr; ///< exports Squid bytes
+
+    uint64_t bytes_read = 0; ///< aggregate bytes read
+    uint64_t bytes_written = 0; ///< aggregate bytes written
```
rousskov, line 215: Since we are (correctly!) changing the type of these data members, and have to do it in a rather subtle way that may hide certain bugs from the compiler, please _rename_ them. This renaming will also allow us to follow the current Squid naming style better and be more consistent with the two other closely related data member names in this class.

```suggestion
    uint64_t totalBytesRead_ = 0; ///< aggregate bytes received
    uint64_t totalBytesWritten_ = 0; ///< aggregate bytes sent
```

I like using synonyms instead of saying "foo is foo", but I do not insist on the above comment adjustment. Here is a version without it:

```suggestion
    uint64_t totalBytesRead_ = 0; ///< aggregate bytes read
    uint64_t totalBytesWritten_ = 0; ///< aggregate bytes written
```

Needless to say, the other PR code will need to be adjusted accordingly, but it is a straightforward (and useful) change.

src/DiskIO/IpcIo/IpcIoFile.cc
```diff
@@ -784,6 +784,9 @@ diskerWriteAttempts(IpcIoMsg &ipcIo)
                    " on " << attempts << " try: " << xstrerr(ipcIo.xerrno));
             ipcIo.len = wroteSoFar;
             return; // bail on error
+        } else {
+            // XXX: setup fd_table[TheFile] properly in DiskerOpen so we can use FD_WRITE_METHOD().
+            fd_table[TheFile].bytesWritten(result);
```
rousskov, line 789: Please remove this new `else` and move the bytesWritten() call/comment down, using the wroteNow variable that is computed a few lines below, where the "else" case is already being handled in the official code.

rousskov (changes requested):
Thank you for addressing two of the three critical issues identified in my first review. I have updated the thread about the remaining one. I also had a bit more time to look over PR code and added a few new change requests; all are easy to address.

src/fde.h
yadij: done in [ce5f427](https://github.com/squid-cache/squid/pull/1352/commits/ce5f4270e3dbfeba2771540abf1ff2e46f2bd02d)

src/DiskIO/IpcIo/IpcIoFile.cc
yadij: done in [3f5541c](https://github.com/squid-cache/squid/pull/1352/commits/3f5541c90c50f6afef45114c34a8915f433d10b2)

src/fde.h
```diff
@@ -84,8 +84,25 @@ class fde
     /// use I/O methods that maintain an internal-to-them buffer
     void useBufferedIo(READ_HANDLER *, WRITE_HANDLER *);
 
-    int read(int fd, char *buf, int len) { return readMethod_(fd, buf, len); }
-    int write(int fd, const char *buf, int len) { return writeMethod_(fd, buf, len); }
+    /// call the read method set by setIo()
```
rousskov, line 87: I accidentally dropped the terminating period in my earlier suggestion, probably while being confused by the missing `call` capitalization. Sorry. That period may become important for Doxygen "briefs" and improves overall multiline comment consistency/shape, so please fix this:

```suggestion
    /// Call the read method set by setIo(). 
```

While we are at it, besides setIo(), there are actually at least two more public functions that set I/O methods: useDefaultIo() and useBufferedIo(). Let's avoid the implication that only setIo() matters here. For example:

```suggestion
    /// Call the read method set by setIo() or similar. 
```

or just drop that non-essential detail completely:

```suggestion
    /// Call the previously set read I/O method.
```

src/fde.h
```diff
@@ -84,8 +84,25 @@ class fde
...
+    /// Performs I/O counting for all bytes received.
+    /// \returns the read method result
+    int read(int fd, char *buf, int len) {
+        auto n = readMethod_(fd, buf, len);
+        if (n > 0)
+            bytesRead(n);
+        return n;
+    }
+
+    /// call the write method set by setIo()
```
rousskov, line 97:
```suggestion
    /// Call the write method set by setIo().
```
or, better,

```suggestion
    /// Call the write method set by setIo() or similar.
```

or simply

```suggestion
    /// Call the previously set write I/O method.
```

src/fde.h
```diff
@@ -97,6 +114,18 @@ class fde
     /// record a transaction on this FD
     void noteUse() { ++pconn.uses; }
 
+    /// record bytes being read and return the accumulated total for this FD.
```
rousskov, line 117: This description got stale -- the method no longer returns anything. We should also polish it to emphasize that Squid read the bytes in question (rather than being in the process of reading them) -- the I/O being recorded here is over. For example:

```suggestion
    /// Record bytes received from a single I/O call.
    /// Whenever possible, configure I/O methods and use read() instead.
```

src/fde.h
```diff
@@ -97,6 +114,18 @@ class fde
...
+    void bytesRead(const size_t n) { bytes_read += n; }
+
+    /// \copydoc totalBytesRead_
+    uint64_t totalBytesRead() const { return bytes_read; }
+
+    /// record bytes being written and return the accumulated total for this FD.
```
rousskov, line 123:
```suggestion
    /// Record bytes sent by a single I/O call.
    /// Whenever possible, configure I/O methods and use write() instead.
```

src/snmp_core.cc
```diff
@@ -445,6 +447,8 @@ snmpConstructReponse(SnmpRequest * rq)
     if (RespPDU != nullptr) {
         snmp_build(&rq->session, RespPDU, rq->outbuf, &rq->outlen);
         comm_udp_sendto(rq->sock, rq->from, rq->outbuf, rq->outlen);
+        // XXX: setup fd_table[] properly so we can use FD_WRITE_METHOD()
+        fd_table[rq->sock].bytesWritten(rq->outlen);
```
rousskov, line 451: The new code should not be used when comm_udp_sendto() fails.

```suggestion
        const auto result = comm_udp_sendto(rq->sock, rq->from, rq->outbuf, rq->outlen);
        if (result > 0) {
            // XXX: setup fd_table[] properly so we can use FD_WRITE_METHOD()
            fd_table[rq->sock].bytesWritten(result);
        }
```

src/wccp.cc
```diff
@@ -357,6 +364,9 @@ wccpAssignBuckets(void)
                   buf,
                   wab_len + WCCP_BUCKETS + cache_len,
                   0);
+    // XXX: setup fd_table[] properly so we can use FD_WRITE_METHOD()
+    fd_table[sock].bytesWritten(wab_len + WCCP_BUCKETS + cache_len);
```
rousskov, line 368: Same here: Missing protection from comm_udp_send() failures. GitHub will not let me post a specific correction for this one.

src/wccp2.cc
```diff
@@ -1640,12 +1642,19 @@ wccp2HereIam(void *)
                                 router,
                                 &service_list_ptr->wccp_packet,
                                 service_list_ptr->wccp_packet_size);
+                // XXX: setup fd_table[] properly so we can use FD_WRITE_METHOD()
+                fd_table[theWccp2Connection->fd].bytesWritten(service_list_ptr->wccp_packet_size);
```
rousskov, line 1646: Same here: Missing protection from comm_udp_sendto() failures. GitHub will not let me post a specific correction for this one.

src/wccp2.cc
```diff
@@ -2024,10 +2033,16 @@ wccp2AssignBuckets(void *)
                                     tmp_rtr,
                                     &wccp_packet,
                                     offset);
+                    // XXX: setup fd_table[] properly so we can use FD_WRITE_METHOD()
+                    fd_table[theWccp2Connection->fd].bytesWritten(offset);
                 } else {
-                    if (send(theWccp2Connection, &wccp_packet, offset, 0) < static_cast<int>(offset)) {
+                    const auto x = send(theWccp2Connection, &wccp_packet, offset, 0);
+                    if (x < static_cast<int>(offset)) {
```
rousskov, line 2040: We can simplify and avoid (paranoid) 64-to-32 bit value truncation concerns:

```suggestion
                    if (x < 0) {
```

src/wccp2.cc
```diff
@@ -2024,10 +2033,16 @@ wccp2AssignBuckets(void *)
                                     tmp_rtr,
                                     &wccp_packet,
                                     offset);
+                    // XXX: setup fd_table[] properly so we can use FD_WRITE_METHOD()
+                    fd_table[theWccp2Connection->fd].bytesWritten(offset);
```
rousskov, line 2037: Same here: Missing protection from comm_udp_sendto() failures. GitHub will not let me post a specific correction for this one.

src/wccp2.cc
```diff
@@ -2024,10 +2033,16 @@ wccp2AssignBuckets(void *)
...
                 } else {
-                    if (send(theWccp2Connection, &wccp_packet, offset, 0) < static_cast<int>(offset)) {
+                    const auto x = send(theWccp2Connection, &wccp_packet, offset, 0);
+                    if (x < static_cast<int>(offset)) {
                         int xerrno = errno;
                         debugs(80, 2, "ERROR: failed to send WCCPv2 HERE_I_AM packet to " << tmp_rtr << " : " << xstrerr(xerrno));
+                    } else {
+                        // XXX: setup fd_table[] properly so we can use FD_WRITE_METHOD()
+                        fd_table[theWccp2Connection->fd].bytesWritten(x);
```
rousskov, line 2045: > PR description: Move the FD read/write byte counting to the class fde read/write operators ... This fixes ... under-counting of UDP protocols

This PR does not move UDP byte accounting to fde class methods. Instead, it fixes UDP undercounting by adding missing accounting calls to UDP handling code. The current plan is to switch UDP handling code to fde accounting as well, but that has not happened yet. Please adjust the PR description accordingly.

While you are fixing this, please replace "operators" with a more accurate "methods" or "functions".

rousskov (changes requested):
Another partial review to help make steady progress.

src/comm.cc
rousskov: > quite a lot of new calls to bytesRead/bytesWritten with XXX markers are added as a result of this request

Addressing the problem identified in this change request did _not_ require the addition of those calls.

I am not against improving UDP accounting, including improving UDP accounting by adding those (temporary) calls with XXXs, but that improvement is a _separate_ issue/decision. I have not asked for that accounting improvement and have not even suggested it. In fact, I have suggested the _opposite_ approach -- "keeping a few (translated) legacy calls" that had to be kept to avoid accounting regressions, resulting in "virtually no new bytesRead() calls".

PR #1355, opened 2023-05-20
squidclient: MinGW-w64 build fixes

yadij:


kinkie (approved):
Ship it!

rousskov (changes requested):

yadij:
This change is independent of PR #1341

tools/squidclient/Transport.cc
```diff
@@ -249,8 +249,8 @@ Transport::Write(const void *buf, size_t len)
 #endif
     } else {
 
-#if _SQUID_WINDOWS_
-        return send(conn, buf, len, 0);
+#if _SQUID_WINDOWS_ || _SQUID_MINGW_
```
rousskov, line 252: This condition should not be added because its addition changes nothing in the official code: In the official code, `_SQUID_MINGW_` cannot be true if `_SQUID_WINDOWS_` is false. The situation will change after PR 1341 lands. This PR can be merged before that if these `_SQUID_MINGW_` are removed, but since we will probably want to add these conditions after PR 1341 lands, we probably should wait for that PR to land first (to avoid changing this code twice). Your call though.

yadij: PR #1341 has landed. Making this request obsolete.

tools/squidclient/Transport.cc
```diff
@@ -249,8 +249,8 @@ Transport::Write(const void *buf, size_t len)
...
+        return send(conn, static_cast<const char *>(buf), len, 0);
 #else
         alarm(Config.ioTimeout);
         return write(conn, buf, len);
```
rousskov, line 256: I am concerned about us adding MinGW-specific hacks to high-level code. For example, there are other send(2) calls in high-level Squid code. Some do not use `char *` buffers AFAICT. There ought to be write(2) calls as well. If we end up adding similar MinGW wrappers to many of them, we would be making our code worse overall, even though each specific change may look "small".

Similarly, AFAICT, this change effectively disables squidclient timeout feature on MinGW. What criteria do we use to decide which features are essential for MinGW and which can be easily dropped on MinGW? For example, perhaps the squidclient feature _itself_ can be dropped on MinGW? There ought to be other HTTP clients supported in that environment. I do not know the criteria you use. Can you share them?

If we decide that we must support squidclient on MinGW, then I see two possible next steps:

A. Investigate what it would take for MinGW builds to use the ordinary write() code path instead of the exceptional send() path. If MinGW can use that ordinary code path without heroic efforts, great! We get fewer exceptional handling in high-level Squid code now and can solve the send() argument type problem later.

B. If using write(2) is not feasible, provide a send(2) portability wrapper that can be used without casting and `#ifs` throughout Squid high-level code. AFAICT, old Windows code attempted to do that in compat/os/mswindows.h (but we may need a different solution for MinGW).

rousskov (changes requested):

yadij:
Closing as squidclient tool has now been removed from Squid.

PR #1356, opened 2023-05-20
Fix fatalf() undefined for environments without syslog

yadij:
src/debug/debug.cc:1087
error: 'fatalf' was not declared in this scope

### DIFF
--- a/src/debug/debug.cc
+++ b/src/debug/debug.cc
@@ -11,6 +11,7 @@
 #include "squid.h"
 #include "base/TextException.h"
 #include "debug/Stream.h"
+#include "fatal.h"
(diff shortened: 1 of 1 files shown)


kinkie (approved):
I’m surprised this only affects environments without syslog but.. sure!

rousskov (approved):

rousskov:
> I’m surprised this only affects environments without syslog

```C++
#if HAVE_SYSLOG
    _db_set_syslog(facility);
#else
    (void)facility;
    // TODO: Throw.
    fatalf("Logging to syslog not available on this platform");
#endif
```

PR #1357, opened 2023-05-20
Fix time_t type conflicts in libdebug

yadij:
error: cannot convert 'const long int*' to 'const time_t*'

On Windows, time_t is 64-bit, and long is not.


rousskov (approved):
Thank you for fixing this, especially given such a low quality if the official debugLogTime() code!

PR #1358, opened 2023-05-20
MinGW-w64: enable native file locking

yadij:
flock() and related API symbols are not defined by MinGW.


rousskov (changes requested):

yadij:
This change is independent of PR #1341

src/base/File.cc
```diff
@@ -34,15 +34,15 @@ FileOpeningConfig::ReadOnly()
     FileOpeningConfig cfg;
 
     /* I/O */
-#if _SQUID_WINDOWS_
+#if _SQUID_WINDOWS_ || _SQUID_MINGW_
```
rousskov, line 37: This condition should not be added because its addition changes nothing in the official code: In the official code, `_SQUID_MINGW_` cannot be true if `_SQUID_WINDOWS_` is false. The situation will change after PR #1341 lands. AFAICT, this PR only changes these conditions, so, if they are removed, we will be left with the official code. Thus, there is nothing to merge here until PR 1341 lands.

rousskov (dismissed):

PR #1359, opened 2023-05-20
MinGW-w64: fix TOS socket API

yadij:
error: cannot convert 'int*' to 'const char*'


rousskov (dismissed):

yadij:
This change is independant of the changes in PR #1341

src/ip/QosConfig.cc
```diff
@@ -533,7 +533,11 @@ Ip::Qos::setSockTos(const int fd, tos_t tos, int type)
     // unless we pass it a 32-bit variable storing 8-bits of data.
     // NP: it is documented as 'int' for all systems, even those like Linux which accept 8-bit char
     //     so we convert to a int before setting.
+#if _SQUID_MINGW_
```
rousskov, line 536: > This change is independant of the changes in PR #1341

This PR can certainly be viewed as independent, but if I do that, then I would have to block this PR because `_SQUID_MINGW_` semantics on the official code is not yet compatible with this PR (even though the code may _compile_ -- the compiler does not understand macro semantics). It will become compatible after PR 1341 is merged. I prefer to just wait for that PR to land, so that we can move steadily forward in one direction.

Same for the other similar recent PRs, of course.

yadij: I do not understand what you mean. PR #1343 does not change _SQUID_MINGW_ in any way. The macro exists today in `master` branch and is set for exactly the build environment which this new code is needed.

rousskov: You probably meant "PR 1341". Please see https://github.com/squid-cache/squid/pull/1341#discussion_r1199839754, where the discussion about those PR 1341 changes is (or should be) happening.

rousskov: > Alex: This PR can certainly be viewed as independent, but if I do that, then I would have to block this PR because `_SQUID_MINGW_` semantics on the official code is not yet compatible with this PR (even though the code may _compile_ -- the compiler does not understand macro semantics).

I found a way to let this PR merge independent from PR 1341: Squid already has a bit of code that uses `_SQUID_MINGW_` outside `_SQUID_WINDOWS_` sections. We can merge this new code before PR 1341 lands because this PR uses the same (old) approach (and it does not conflict with PR 1341 changes). I am removing my negative vote from this and similar recent PRs and will re-review them accordingly.

src/ip/QosConfig.cc
```diff
@@ -533,7 +533,11 @@ Ip::Qos::setSockTos(const int fd, tos_t tos, int type)
     // unless we pass it a 32-bit variable storing 8-bits of data.
     // NP: it is documented as 'int' for all systems, even those like Linux which accept 8-bit char
     //     so we convert to a int before setting.
```
rousskov, line 535: This comment will become stale if this PR changes are committed as is. If the above claim is still valid for all non-Windows environments (that we know about), then we can simply move this comment inside the new `#else` section and replace "all systems" with "all non-Windows systems" or even "all other systems".

When adjusting the comment, please drop the NP prefix and the related indentation.

```
       IP_TOS (since Linux 1.0)
              Set or receive the Type-Of-Service (TOS) field that is
              sent with every IP packet originating from this socket.
              It is used to prioritize packets on the network.  TOS is a
              byte.  There are some standard TOS flags defined:
```

FWIW, at least some modern Linux [manpages](https://www.man7.org/linux/man-pages/man7/ip.7.html) document the TOS value as a "byte" not "8-bit char". Internally, Linux kernel [probably](https://gist.github.com/wenjianhn/0f7a9a1e36018a42515c?permalink_comment_id=4302154#gistcomment-4302154) stores these values using (the equivalent of) uint8_t. `tos_t` is (the equivalent of) uint8_t. None of that matches "8-bit char" exactly IMHO.

I would also replace "it" with "the underlying socket option value" or something like that to emphasize that we are talking about the option-specific thing the optvalue pointer points at rather than the (generic) optvalue pointer. The latter is `const void *` on modern Linuxes (at least).

src/ip/QosConfig.cc
```diff
@@ -533,7 +533,11 @@ Ip::Qos::setSockTos(const int fd, tos_t tos, int type)
     // unless we pass it a 32-bit variable storing 8-bits of data.
```
rousskov, line 533: Please move this comment inside the `#else` statement were FreeBSD code will now live.

Changing our `int` to `int32_t` to match the claim in the comment is probably out of this PR scope.

src/ip/QosConfig.cc
```diff
@@ -533,7 +533,11 @@ Ip::Qos::setSockTos(const int fd, tos_t tos, int type)
...
     // NP: it is documented as 'int' for all systems, even those like Linux which accept 8-bit char
     //     so we convert to a int before setting.
+#if _SQUID_MINGW_
+    const char bTos = tos;
```
rousskov, line 537: Can you point to a document that defines the expected `IP_TOS` _value_ for MinGW to be a `char`? I am worried that we are fixing a compiler warning but adding a (silent) runtime bug here.

P.S. I understanding that using `const char` type for the underlying value changes the setsockopt() option value argument type from `int *` to `const char *` and fixes the compiler warning. However, the type of the underlying value is not determined by what _pointer_ type setsockopt() declaration expects. The same setsockopt() function works with all socket options, so its pointer type cannot determine the underlying option value type. We already had to fix the underlying type to match OS expectations in commit b4878a4. I want to make sure we are not re-introducing a problem that commit fixed.

Some Windows [documentation](https://learn.microsoft.com/en-us/windows/win32/winsock/ipproto-ip-socket-options) says that IP_TOS socket option is a DWORD boolean value (that should not be used), but I am _not_ suggesting that documentation is relevant to MinGW builds. I am only mentioning it here to illustrate the variety of underlying types that OSes might expect.

rousskov (changes requested):

src/ip/QosConfig.cc
yadij: Hmm, you are right. The DWORD is documented as a "boolean" type, which in turn is specified as "integer" sized. So we should just cast the pointer while leaving the type actually be `int`.

Updated in [9470fea](https://github.com/squid-cache/squid/pull/1359/commits/9470fea0f8df48a345b178ebe143e09805e701b3).

yadij: Regarding the "Do not use" comment in their documentation, it then states that QoS API should be used to set DiffServ instead. However the QoS API referenced has been obsoleted and removed in Windows Vista. It looks like the `Socket Options` documentation is a bit outdated.

src/ip/QosConfig.cc
```diff
@@ -531,15 +531,22 @@ Ip::Qos::setSockTos(const int fd, tos_t tos, int type)
 {
     // Bug 3731: FreeBSD produces 'invalid option'
     // unless we pass it a 32-bit variable storing 8-bits of data.
-    // NP: it is documented as 'int' for all systems, even those like Linux which accept 8-bit char
-    //     so we convert to a int before setting.
+    // NOTE: setsockopt(.., IP_TOS, ...) 'optval' parameter
+    //       is documented as 'int' for all systems
+    //       despite what the function parameter type is.
+    //       So we convert to int before setting.
```
rousskov, line 537: Please drop the `NOTE:` prefix. Every comment is a NOTE by default, and there is no good reason to maintain the associated indentation. I would also simplify to avoid using the first "parameter" word, to clarify what type we are talking about. Together with the other change request suggestion, this comment will become very simple:

```suggestion
    // setsockopt(.., IP_TOS, ...) option value is documented to be 'int' for all systems
```

or even

```suggestion
    // IP_TOS option value is documented to be 'int' for all systems
```

We could add a comment about casting to the place where we actually cast, but I would not do that because Squid has dozens of such casts and some of them are already documented. There is no need to document this particular cast in this particular place IMO[^1]. There is nothing special about this particular cast, and it is virtually unrelated to the type of the option value (technically, the explicit cast could be avoided if the type was `char`, but the reader of this code does not really need to know that fact).

[^1]: especially since we should eventually migrate to a cast-free wrapper as discussed in the other change request

rousskov (changes requested):

Sakatok (approved):

yadij:
Waiting on PR #1373

src/WinSvc.cc
```diff
@@ -965,7 +965,7 @@ static int Win32SockInit(void)
         } else {
             opt = opt | SO_SYNCHRONOUS_NONALERT;
 
-            if (::setsockopt(INVALID_SOCKET, SOL_SOCKET, SO_OPENTYPE, (char *) &opt, optlen)) {
+            if (::setsockopt(INVALID_SOCKET, SOL_SOCKET, SO_OPENTYPE, reinterpret_cast<char *>(&opt), optlen)) {
```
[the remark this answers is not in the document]
yadij: As demonstrated in PR #1373 this is a deceptive request. Acceding will result in a fall down the slippery slope of ever increasing perfection requests. Throwing this PR further out of the intended scope of fixing a **few** MinGW compile errors.

yadij:
No longer going to wait for PR #1373 which has already started devolving into requests for perfection.

src/WinSvc.cc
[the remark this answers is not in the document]
rousskov: > As demonstrated in PR #1373 this is a deceptive request. Acceding will result in a fall down the slippery slope of ever increasing perfection requests. Throwing this PR further out of the intended scope of fixing a **few** MinGW compile errors.

I disagree with the above assessment and speculation. However, may not need to agree on this to make progress: If you refactor this PR so that it is just "fixing a few MinGW compile errors", without modifying so many setsockopt() calls (most of which are not specific and some are probably not even applicable to MinGW builds), then this change request will become stale/inapplicable and can be resolved on those grounds. I do not know whether (and am not implying that) such refactoring is possible.

rousskov (changes requested):

kinkie (approved):

src/client_side.cc
```diff
@@ -2134,7 +2134,7 @@ ConnStateData::start()
             (transparent() || port->disable_pmtu_discovery == DISABLE_PMTU_ALWAYS)) {
 #if defined(IP_MTU_DISCOVER) && defined(IP_PMTUDISC_DONT)
         int i = IP_PMTUDISC_DONT;
-        if (setsockopt(clientConnection->fd, SOL_IP, IP_MTU_DISCOVER, &i, sizeof(i)) < 0) {
+        if (setsockopt(clientConnection->fd, SOL_IP, IP_MTU_DISCOVER, reinterpret_cast<char *>(&i), sizeof(i)) < 0) {
```
rousskov, line 2137: TBD

src/WinSvc.cc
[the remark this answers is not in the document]
rousskov: AFAICT, [this change request](https://github.com/squid-cache/squid/pull/1359#discussion_r1204726369) has not been addressed when PR branch was force-pushed from 02b82a1 to 213df95 and my re-review requested a few hours ago.

rousskov (changes requested):
Please address the earlier change request at https://github.com/squid-cache/squid/pull/1359#discussion_r1204726369

src/WinSvc.cc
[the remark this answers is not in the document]
yadij: It was addressed by authors decision not to implement your requested change.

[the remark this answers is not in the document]
rousskov: > > AFAICT, [this change request](https://github.com/squid-cache/squid/pull/1359#discussion_r1204726369) has not been addressed when PR branch was force-pushed from [02b82a1](https://github.com/squid-cache/squid/commit/02b82a17493c4097e5ba316f6a44c3ed6d4706a1) to [213df95](https://github.com/squid-cache/squid/commit/213df95d46de47975799e6a4879eca5ac93245c5) and my re-review requested a few hours ago.

> It was addressed by authors decision not to implement your requested change.

Thank you for disclosing that decision; I was not aware of that decision when my re-review was requested.

This change request stands. Doing nothing does not address my blocking concern.

rousskov:
> [Alex](https://github.com/squid-cache/squid/pull/1915#discussion_r1797092133): Please do not block PRs just because you do not see value in them (when other core developers do see value).

I am not blocking this PR because I do not see its value. I am currently blocking this PR because its value is overwhelmed by its harm, while addressing my concern (and reversing that relationship) does not require heroic efforts.

PR #1362, opened 2023-05-21
Remove obsolete caddr_t

yadij:


lib/radix.c
```diff
@@ -243,10 +243,10 @@ rn_satsifies_leaf(char *trial, register struct squid_radix_node *leaf, int skip)
 
 struct squid_radix_node *
 squid_rn_match(void *v_arg, struct squid_radix_node_head *head) {
-    caddr_t v = v_arg;
+    char *v = v_arg;
     register struct squid_radix_node *t = head->rnh_treetop, *x;
-    register caddr_t cp = v, cp2;
-    caddr_t cplim;
+    register char *cp = v, *cp2;
```
kinkie, line 248: AFAIK the register modifier is ignored by now; will it make sense to remove it while we're here?

yadij: IMO best to do that in a separate PR.

lib/radix.c
```diff
@@ -98,7 +98,7 @@ static char *rn_zeros, *rn_ones;
 #define rm_leaf rm_rmu.rmu_leaf /* extra field would make 32 bytes */
 
 /* Helper macros */
-#define squid_Bcmp(a, b, l) (l == 0 ? 0 : memcmp((caddr_t)(a), (caddr_t)(b), (u_long)l))
+#define squid_Bcmp(a, b, l) (l == 0 ? 0 : memcmp((char *)(a), (char *)(b), (u_long)l))
```
rousskov, line 101: This badly written macro is unused[^1]. It is best to delete it instead of adjusting its _problematic_ argument casts.

```suggestion
```

[^1]: I did not check carefully, but it looks like it is unused since inception in 2001 commit 06cdab8 or, in its original `Bcmp()` spelling, in 1997 commit 76f729d.

rousskov (commented):
Thank you for removing this obsolete type.

Disclaimer: I did not check all the replacements.

rousskov (approved):

PR #1367, opened 2023-05-25
Fix userinfo percent-encoding

rousskov:
%X expects an unsigned int, and that is what we were giving it. However,
to get to the correct unsigned int value from a (signed) char, one has
to cast to an unsigned char (or equivalent) first.

Broken since inception in commit 7b75100.

Also adjusted similar (commented out) ext_edirectory_userip_acl code.

### DIFF
--- a/src/acl/external/eDirectory_userip/ext_edirectory_userip_acl.cc
+++ b/src/acl/external/eDirectory_userip/ext_edirectory_userip_acl.cc
@@ -1555,7 +1555,7 @@ MainSafe(int argc, char **argv)
         /* BINARY DEBUGGING *
                     local_printfx("while() -> bufa[%" PRIuSIZE "]: %s", k, bufa);
                     for (i = 0; i < k; ++i)
-                      local_printfx("%02X", bufa[i]);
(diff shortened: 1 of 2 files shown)


src/acl/external/eDirectory_userip/ext_edirectory_userip_acl.cc
```diff
@@ -1555,7 +1555,7 @@ MainSafe(int argc, char **argv)
         /* BINARY DEBUGGING *
                     local_printfx("while() -> bufa[%" PRIuSIZE "]: %s", k, bufa);
                     for (i = 0; i < k; ++i)
-                      local_printfx("%02X", bufa[i]);
+                      local_printfx("%02X", static_cast<unsigned int>(static_cast<unsigned char>(bufa[i])));
```
rousskov, line 1558: There are several similar (but working) commented out places in this source file that use a different variant of the same conversion (quoted below). If you prefer to keep this kind of code consistent in this file, please let me know, and I will use that variant instead (without testing). Long term, we should be migrating towards a cast though.

```C++
        int c = bufa[i];
        if (c < 0)
            c = c + 256;
```

yadij (approved):

squidadm:
queued for backport to v6

PR #1368, opened 2023-05-25
Restore errno in %err_detail for ERR_CONNECT_FAIL

rousskov:
Squid was sometimes logging %err_code/%err_detail as
ERR_CONNECT_FAIL/WITH_SERVER. It now logs
ERR_CONNECT_FAIL/WITH_SERVER+errno=111 (or similar).

When dealing with two error details, Squid was ignoring the latter one.
The new ErrorDetails code combines multiple details, reusing the
existing Security::ErrorDetail::brief() "a+b+c" syntax.

The new detail accumulation functionality may help detail other errors.

At least some of the logged errno details were lost in commit ba3fe8d.


src/cf.data.pre
```diff
@@ -4686,7 +4686,13 @@ DOC_START
 		sn	Unique sequence number per log line entry
 		err_code    The ID of an error response served by Squid or
 				a similar internal error identifier.
-		err_detail  Additional err_code-dependent error information.
+
+		err_detail  Additional err_code-dependent error information. Multiple
+			details are separated by the plus sign ('+'). Admins should not
```
rousskov, line 4691: I am not sure the `+` sign is the best choice here. On one hand, that is what some OpenSSL errors are using, and some admin scripts might be able to handle that delimiter better already. On the other hand, we normally use commas for lists, and using `+` might make it more difficult to distinguish a group of OpenSSL error details from other error details. Which delimiter do _you_ consider the best and why?

rousskov (commented):
test|master/v7 (b7fe9e6)|PR (412d550)|diff|
|---:|---------------|----------------|----|
| **GET directly to a non-listening origin** |  |  |  |
| %err_code | `ERR_CONNECT_FAIL` | `ERR_CONNECT_FAIL` |  |
| %err_detail | `WITH_SERVER` | `WITH_SERVER+errno=111` | ! |
| **CONNECT directly to a non-listening origin** |  |  |  |
| %err_code | `ERR_CONNECT_FAIL` | `ERR_CONNECT_FAIL` |  |
| %err_detail | `errno=111` | `errno=111` |  |
| **GET through a cache_peer to a non-listening origin** |  |  |  |
| %err_code | `-` | `-` |  |
| %err_detail | `-` | `-` |  |
| **CONNECT through a cache_peer to a non-listening origin** |  |  |  |
| %err_code | `-` | `-` |  |
| %err_detail | `-` | `-` |  |
| **GET through a non-listening cache_peer** |  |  |  |
| %err_code | `ERR_CONNECT_FAIL` | `ERR_CONNECT_FAIL` |  |
| %err_detail | `WITH_SERVER` | `WITH_SERVER+errno=111` | ! |
| **CONNECT through a non-listening cache_peer** |  |  |  |
| %err_code | `ERR_CONNECT_FAIL` | `ERR_CONNECT_FAIL` |  |
| %err_detail | `errno=111` | `errno=111` |  |

src/cf.data.pre
yadij: IMO the syntax (whatever it is or becomes) should be the same as Squid will use in the `Proxy-Status:` header `detail=` parameter defined by RFC 9209 section 2.1.5.

yadij (changes requested):
Overall this PR has all the hallmarks of over-engineering a complex solution to re-implement STL container functionality.

Please just define the `ErrorDetails` container as a `list` or `forward_list` and use the available `std::merge` calls available since C++17.

src/error/Details.h
```diff
@@ -0,0 +1,51 @@
...
+#define SQUID_SRC_ERROR_DETAILS_H
+
+#include "error/Detail.h"
+#include "mem/PoolingAllocator.h"
+
+#include <vector>
+
+/// Multiple details of a single error, in "canonical" order without duplicates.
+/// Our "canonical" order is the approximate detail discovery order (e.g., the
+/// order of Error::update() calls). This class isolates multi-detail storage
```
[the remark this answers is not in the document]
yadij: I do not think those words mean what you think they do.

src/error/Details.h
```diff
@@ -0,0 +1,51 @@
...
+    /// adds unique details (if any) from the given collection
+    void mergeMany(const ErrorDetails &);
+
+    /* ErrorDetail API */
+    virtual SBuf brief() const override;
+    virtual SBuf verbose(const HttpRequestPointer &) const override;
+
+private:
+    /// known unique details in canonical order
+    std::vector<ErrorDetailPointer, PoolingAllocator<ErrorDetailPointer> > details;
```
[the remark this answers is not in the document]
rousskov: TLDR: `std::vector` optimizes random access while `std::forward_list` makes such access very inefficient and awkward. This difference is real, but it does not automatically mean `std::forward_list` is a better choice for this private data member. We have to consider all factors. Availability of random access interfaces is a secondary concern here because the (private) code in question does not need to protect others from erroneous random access operations and erroneous random access inside this class is very unlikely. Performance concerns make `std::vector` a better choice overall.

> The definition you created for "canonical" forbids insertion or prepending entries,

As you know, both `vector` and `forward_list` support insertion of entries in arbitrary positions.  Canonical order makes such operations unnecessary in the current code. However, if you prefer "forbids", let's assume that interpretation for the sake of the argument.

> and reading in any order other than forward

That is not technically accurate: One could, in theory, meaningfully render events (stored in canonical order) in reverse, for example. However, again, if you prefer to think of this as a "read forward only" storage, let's assume that interpretation for the sake of the argument.

> `vector` violates all of those properties by default, `forward_list` only violates prepend by default. therefore it is clearly the better type.

Yes, `forward_list` is a slightly better type with respect to restricting access. However, it is worse type _overall_ in this context, and that is what matters. If we were talking about some publicly accessible storage where wrong access order would have been bad, the conclusion could have been different. Here, we are talking about a private data member in a small class that can and does exercise tight control over storage operations to prevent bad ones. In this case, performance matters more than (unwanted) availability of fast random access.

src/error/Details.h
yadij, line 19:
```suggestion
/// Multiple details of a single error, in temporal order of discovery without duplicates.
/// The order/timing of Error::update() calls is used to approximate detail discovery time.
/// This class isolates multi-detail storage
```

src/error/Details.h
```diff
@@ -0,0 +1,51 @@
...
+
+protected:
+    // use Merge() instead
+    ErrorDetails() = default;
+
+    /// adds the given detail unless we already have it
+    void mergeOne(const ErrorDetail &);
+
+    /// adds unique details (if any) from the given collection
+    void mergeMany(const ErrorDetails &);
```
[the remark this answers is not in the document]
yadij: This is **not** a self-documenting improvement. Both of these are performing the same semantic action (**merge**). The logic difference between type A and type B merging is already documented by the simple act of the merge being done on an A vs B object.

yadij (changes requested):

src/error/Details.h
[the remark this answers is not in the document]
yadij: Done in https://github.com/squid-cache/squid/pull/1368#discussion_r1247756096

rousskov: I committed your suggestion (b9d27b0), fixed description formatting (6f52d32), and then removed stale "canonical" references (c0b1ed3).

rousskov:

src/error/Details.cc
```diff
@@ -0,0 +1,117 @@
...
+    // error storage (e.g., ALE) twice from the same detail source (e.g., the
+    // same HttpRequest object reachable via two different code/pointer paths)
+    // or retag the error with the same context information (e.g., WITH_CLIENT).
+    // In all those cases, comparing detail object addresses is enough to detect
+    // duplicates. In other cases (if they exist), a duplicate detail would
+    // imply inefficient or buggy error detailing code. Instead of spending ink
+    // and CPU cycles on hiding them, this comparison may expose those (minor)
+    // problems (in hope that they get fixed). Detail consumers should not care.
+    return &d1 == &d2;
+}
```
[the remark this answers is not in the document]
yadij: The API of the `Same(X,Y)` implicitly promises to compare the objects themselves. Not just their addresses.

Also how is `this comparison may expose those (minor) problems (in hope that they get fixed).` supposed to materialize from using a function which does nothing but trivial raw-pointer comparison?
AFAICT the case of duplicates with non-identical raw-pointer values are left to the user or admin spotting them in outputs and caring enough to report. Which would occur anyway without this function wrapper.

Therefore, IMO it is simpler (and thus better) to use the regular equality checks (`==` or `!=`) in the callers without this function.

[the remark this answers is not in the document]
yadij: FTR; after core meeting we agreed to move this to `ErrorDetail::compare()` method. Awaiting that update for next review.

src/error/Details.h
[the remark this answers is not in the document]
yadij: FTR; at core meeting we decided to always present `ErrorDetails` or equivalent collection objects to callers to avoid the issues with juggling a mix of single instances and set containers through a single API.

src/error/Details.h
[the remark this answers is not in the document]
yadij: Closing for now. We can open the topic of most optimal container type after the core meeting agreed redesign occurs.

src/error/Error.h
```diff
@@ -11,29 +11,53 @@
 
 #include "error/Detail.h"
 #include "error/forward.h"
+#include "mem/PoolingAllocator.h"
 
 #include <iosfwd>
+#include <vector>
+
+/// zero or more details of a single error
+using ErrorDetails = std::vector<ErrorDetailPointer, PoolingAllocator<ErrorDetailPointer> >;
```
rousskov, line 20: FWIW, I have tried to implement this using std::list, but it only resulted in more memory allocations/deallocations (because a list has to allocate and deallocate every list node while a vector allocates and deallocates all nodes at once).

Lists can be spliced with fewer memory operations, but the current branch code no longer needs to splice, and the earlier std::forward_list code that was still splicing was not really outperforming an std::vector implementation either (both had about the same number of memory operations for a few errors and std::vector was often winning when handling more details). The relevant use cases just do not benefit from using lists, even though std::vector has to occasionally reallocate.

STL lists may be used where container elements are expensive or impossible to copy; ErrorDetail::Pointer is cheap to copy. STL lists may also be more efficient when we need to delete or add nodes in the middle of a container, but we do not need to do that when storing error details.

src/error/Error.h
```diff
@@ -11,29 +11,53 @@
...
-    Error(const err_type c, const ErrorDetailPointer &d): category(c), detail(d) {}
+    Error(const err_type c, const ErrorDetailPointer &d): Error(c) { update(d); }
 
     explicit operator bool() const { return category != ERR_NONE; }
 
     /// if necessary, stores the given error information (if any)
     void update(const Error &);
 
-    /// convenience wrapper for update(Error)
-    void update(const err_type c, const ErrorDetailPointer &d) { update(Error(c, d)); }
```
rousskov, line 30: We could keep the old simple implementation, but it becomes quite expensive when we migrate from a single error detail to many details (regardless of the STL container type).

src/error/Error.h
```diff
@@ -11,29 +11,53 @@
...
+
+    /// if necessary, stores the given error category (if any), leaving details unchanged;
+    /// the first non-ERR_NONE category wins;
+    /// more convenient alternative to calling update(Error(c))
+    void update(err_type);
+
+    /// records an additional error detail (if any), leaving category unchanged
+    /// more convenient and faster alternative to calling update(Error(ERR_NONE, d))
+    /// \param detail either nil or a pointer to a new or an already known detail
+    void update(const ErrorDetailPointer &detail);
```
rousskov, line 50: We do not have to add this and the other new update() method. Earlier implementations did not have them. However, having these methods actually simplifies Error code (and improves caller code readability).

src/errorpage.cc
```diff
@@ -1348,15 +1348,12 @@ ErrorState::BuildHttpReply()
     // Make sure error codes get back to the client side for logging and
     // error tracking.
     if (request) {
-        if (detail)
-            request->detailError(type, detail);
-        else
-            request->detailError(type, SysErrorDetail::NewIfAny(xerrno));
+        request->error.update(type, detail);
+        request->error.update(SysErrorDetail::NewIfAny(xerrno));
```
rousskov, line 1352: It is possible to add more or more complex Error methods to handle this better, but I did not do that because this code should not exist at all. There is already a TODO that would refactor how ErrorState handles these error details. When we are done, I suspect that ErrorState will start using Error objects in ALE and/or HttpRequest to handle them, eliminating this error-prone code.

src/errorpage.cc
```diff
@@ -1348,15 +1348,12 @@ ErrorState::BuildHttpReply()
...
     } else if (ale) {
-        if (detail)
-            ale->updateError(Error(type, detail));
-        else
-            ale->updateError(Error(type, SysErrorDetail::NewIfAny(xerrno)));
+        Error err(type, detail);
+        err.update(SysErrorDetail::NewIfAny(xerrno));
+        ale->updateError(err);
```
rousskov, line 1356: Unlike HttpRequest, ALE does not expose its error storage, so we cannot, without heroic efforts, keep this code symmetric to the more efficient HttpRequest code above. I suspect we will eventually adjust ALE to expose its error, but that requires a few noisy changes and should probably be done _after_ ErrorState refactoring mentioned in another change request comment. For now, I think this lack of symmetry is acceptable in this (already problematic) code.

rousskov (commented):
I refactored multi-detail support to avoid merge/splice complexities without adding significant overheads (in most cases). It took a while to find the right complexity/performance balance. I hope it was worth it. @yadij, please re-review.

More performance-related improvements are possible, but they require exposing stored Error objects in ALE, ClientHttpRequest, and other classes, which would create a lot of noise unrelated to fixing this bug (and at least one more that is waiting for this fix to land). We should do that refactoring in followup PRs IMO.

yadij (approved):

PR #1371, opened 2023-06-01
Stop leaking PeerDigests on reconfiguration

eduard-bagdasaryan:
peerDigestCreate() cbdata-locked the digest twice. The second lock was a
"self lock" -- a lock without storing the locked pointer somewhere. That
self lock was introduced (with an XXX) in 2002 commit fa80a8e. It did
not have the corresponding unlock, and Squid leaked the digest object.
That leak became conditional in 2018 commit b56b37c: Since then, Squid
was missing one digest unlock only when the CachePeer object was gone
before a peerDigestDestroy() call, as happens during reconfiguration.

Also removed a pair of excessive cbdata locks/unlocks (that triggered
this leak): Long-term users should lock when they start storing a
pointer to the cbdata-protected object and unlock when they stop.

Also converted PeerDigest::peer to CbcPointer, addressing a TODO.

Also fixed subtle problems with detecting gone CachePeer and gone
PeerDigest objects in peerDigestFetchedEnough().

Also removed stale peerNoteDigestGone() declaration.

### DIFF
--- a/src/peer_digest.cc
+++ b/src/peer_digest.cc
@@ -656,7 +656,7 @@ peerDigestFetchedEnough(DigestFetchState * fetch, char *buf, ssize_t size, const
      * cases marked with '?!' should not happen */
 
     if (!reason) {
-        if (!(pd = fetch->pd))
+        if (!pdcb_valid || !(pd = fetch->pd))
             reason = "peer digest disappeared?!";
         else
             host = pd->host;
--- a/src/peer_digest.cc
+++ b/src/peer_digest.cc
@@ -10,6 +10,7 @@
 
 #include "squid.h"
 #if USE_CACHE_DIGESTS
+#include "base/IoManip.h"
 #include "CacheDigest.h"
 #include "CachePeer.h"
 #include "event.h"
@@ -65,19 +66,10 @@ static const time_t GlobDigestReqMinGap = 1 * 60;   /* seconds */
 
 static time_t pd_last_req_time = 0; /* last call to Check */
 
-PeerDigest::PeerDigest(CachePeer * p)
+PeerDigest::PeerDigest(CachePeer * const p):
+    peer(p),
+    host(peer->host) // if peer disappears, we will know its name
 {
-    assert(p);
-
-    /*
-     * DPW 2007-04-12
-     * Lock on to the peer here.  The corresponding cbdataReferenceDone()
-     * is in peerDigestDestroy().
-     */
-    peer = cbdataReference(p);
-    /* if peer disappears, we will know it's name */
-    host = p->host;
-
     times.initialized = squid_curtime;
 }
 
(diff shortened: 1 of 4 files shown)


src/peer_digest.cc
```diff
@@ -65,19 +66,10 @@ static const time_t GlobDigestReqMinGap = 1 * 60;   /* seconds */
 
 static time_t pd_last_req_time = 0; /* last call to Check */
 
-PeerDigest::PeerDigest(CachePeer * p)
+PeerDigest::PeerDigest(CachePeer * const p):
+    peer(p),
+    host(peer->host) // if peer disappears, we will know its name
 {
-    assert(p);
```
rousskov, line 70: This may seem like an assertion removal (and a possibly nil `peer` pointer dereference above), but the new `peer->...` call above asserts when `p` is nil, so the safety check in the official code are preserved by this change.

src/peer_digest.cc
```diff
@@ -699,7 +650,7 @@ peerDigestFetchedEnough(DigestFetchState * fetch, char *buf, ssize_t size, const
     const char *reason = nullptr;  /* reason for completion */
     const char *no_bug = nullptr;  /* successful completion if set */
     const int pdcb_valid = cbdataReferenceValid(fetch->pd);
-    const int pcb_valid = cbdataReferenceValid(fetch->pd->peer);
+    const int pcb_valid = pdcb_valid && fetch->pd->peer.valid();
```
rousskov, line 653: This is essentially a bug fix: The official code could dereference an invalid `fetch->pd` pointer here. I added the corresponding sentence to the PR description.

src/peer_digest.cc
```diff
@@ -699,7 +650,7 @@ peerDigestFetchedEnough(DigestFetchState * fetch, char *buf, ssize_t size, const
...
 
     /* test possible exiting conditions (the same for most steps!)
      * cases marked with '?!' should not happen */
```
rousskov, line 656: Please add `!pdcb_valid || ` as the first condition of the `if (!(pd = fetch->pd))` statement below. Since we are fixing `fetch->pd` dereference bugs in this method, let's fix the last(?) one of them.

src/peer_digest.cc
```diff
@@ -717,7 +668,7 @@ peerDigestFetchedEnough(DigestFetchState * fetch, char *buf, ssize_t size, const
     /* continue checking (with pd and host known and valid) */
 
     if (!reason) {
-        if (!cbdataReferenceValid(pd->peer))
+        if (!pd->peer)
```
rousskov, line 671: This is another (potential) bug fix in this PR AFAICT: The official code does not interpret nil `pd->peer` as a `reason`-setting problem[^1], but should. Again, I adjusted PR description to disclose these fixes.

[^1]: cbdataReferenceValid(nullptr) returns true

src/peer_digest.cc
```diff
@@ -289,7 +240,7 @@ peerDigestCheck(void *data)
 static void
 peerDigestRequest(PeerDigest * pd)
 {
-    CachePeer *p = pd->peer;
+    const auto p = pd->peer.get(); // TODO: Replace with a reference.
```
rousskov, line 243: FWIW, I agree that this TODO should be left for a follow up polishing PR to decrease noise in this bug-fixing PR.

rousskov (commented):
LGTM with one small change (an additional bug fix). My other comments simply state my understanding of this tricky bug-infested code. They do not require a (re)action unless I got something wrong.

src/peer_digest.cc
eduard-bagdasaryan: Done (70cb627).

rousskov (approved):

PR #1372, opened 2023-06-01
Destroy an idle PeerDigest after its CachePeer disappears

eduard-bagdasaryan:
Delaying this destruction until the next peerDigestCheck() event results
in accumulation of PeerDigests (and related objects) during frequent
reconfigurations.


rousskov (approved):
Thank you for this step towards decent reconfiguraiton support.

yadij (approved):

PR #1373, opened 2023-06-03
Make TCP SetSocketOption template public

yadij:


src/comm/Tcp.h
```diff
@@ -25,6 +27,30 @@ class TcpKeepAlive
...
+/// setsockopt(2) wrapper
+template <typename Option>
+inline bool
+SetSocketOption(const int fd, const int level, const int optName, const Option &optValue)
+{
+    static_assert(std::is_trivially_copyable<Option>::value, "setsockopt() expects POD-like options");
+    static_assert(!std::is_same<Option, bool>::value, "setsockopt() uses int to represent boolean options");
+    if (setsockopt(fd, level, optName, reinterpret_cast<char *>(const_cast<Option *>(&optValue)), sizeof(optValue)) < 0) {
+        const auto xerrno = errno;
+        debugs(5, DBG_IMPORTANT, "ERROR: setsockopt(2) failure: " << xstrerr(xerrno));
```
rousskov, line 39: We _can_ leave DBG_IMPORTANT hard-coded like this for now because all the existing callers use it, but some #1359 callers are not compatible with this debugging level. I recommend moving the debugging level into a wrapper parameter in _this_ PR, so that #1359 can use these wrappers without waiting for another PR that solves the debug level problem and without mixing debugging level API changes with lots of "mechanical" caller changes (that #1359 is currently focusing on).

We can make debugging level the last parameter. If you do that, please make it a _required_ one (despite a few existing caller changes):

* defaulting to DBG_IMPORANT would hide this somewhat important and surprising side effect from caller code readers
* defaulting to level 2 or higher may eventually result in callers that should report an important error but forget to do so (neither the copy-pasting developer nor a reviewer will have enough visual clues to spot the problem)
* defaulting to any level may make future cache_log_message control (if any) more harsh/controversial, defeating one of the the primary purposes of that directive -- avoiding costly arguments regarding the "best" message level.
* having no default parameter may allow future wrapper versions to throw on errors by default -- the explicit parameter will then be used to disable that new default throwing behavior

We can even move _both_ the debugging section and the debugging level into wrapper parameters. In that case, in may be a good idea to make them the first two parameters. I am concerned that he API will be a bit verbose and callers may look a bit odd (because these wrappers are not a debugging function). Your call.

The last alternative is to actually throw, as suggested by the TODO comment below. That would solve all these problems nicely, but requires more work in other places.

In summary, I am OK with all of the options itemized below:

1. Do nothing in this PR; another PR will have to solve this problem.
2. Move the debugging level to the last required wrapper parameter.
3. Move the debugging section and level to (first?) wrapper parameters.
4. Throw (instead of using debugs()) to report an error.

rousskov (changes requested):
Thank you for generalizing our setsockopt() wrappers.

yadij:
Superceded by PR #2061

PR #1379, opened 2023-06-14
Fix key equality comparison in LookupTable map

rousskov:
An std::unordered_map with case-insensitive keys must use a
case-insensitive key equality comparison. lookupTable_t used (the
default) std::equal_to<SBuf> comparison which is case-sensitive.

The full extent of this bug effects is unknown, but Squid was
mishandling Cache-Control and Surrogate-Control directives with
non-canonical (i.e. other than all-lower-case) spelling. Similar
problems affected WWW-Authenticate Digest parameter names, but the
related code remains inconsistent, with both case-sensitive and
case-insensitive checks applied to some of the key names in
Auth::Digest::Config::decode().

Also removed a similarly buggy (and, technically, unused) "typical use"
example from the CaseInsensitiveSBufHash class description. C++
documentation and Squid code are better sources of usage examples when
it comes to STL-used concepts like hash function objects.

This minimal fix excludes LookupTable class polishing.

The bug was introduced in 2015 commit 81ab22b.

### DIFF
--- a/src/base/LookupTable.h
+++ b/src/base/LookupTable.h
@@ -70,7 +70,7 @@ public:
     }
 
 private:
-    typedef std::unordered_map<const SBuf, EnumType, Hasher> lookupTable_t;
+    using lookupTable_t = std::unordered_map<const SBuf, EnumType, Hasher, CaseInsensitiveSBufEqual>;
     lookupTable_t lookupTable;
     EnumType invalidValue;
 };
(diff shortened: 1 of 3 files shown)


rousskov:
Manual micro-tests using a similarly broken map and map keys that differ in case only show duplicate entries and unexpected lookup misses. Simple testing code is available at https://godbolt.org/z/zWW3oY1E9

kinkie (approved):
Looks good, the doc clearly specifies it should be case-insensitive
Thanks, and good catch!

yadij (approved):

PR #1380, opened 2023-06-15
Honor DNS RR TTLs larger than negative_dns_ttl

rousskov:
Since 2017 commit fd9c47d, Squid was effectively ignoring DNS RR TTLs
that exceeded negative_dns_ttl (i.e. 60 seconds by default) because the
"find the smallest TTL across the DNS records seen so far" code in
ipcache_entry::updateTtl() mistook the "default" ipcache_entry::expires
value as the one based on an earlier seen DNS record.

In most cases, this bug decreased IP cache hit ratio.

Existing fqdncache code does not suffer from the same bug because
fqdncacheParse() always resets fqdncache_entry::expires instead of
updating it incrementally. ipcacheParse() has to update incrementally
because it is called twice per entry, once with an A answer and once
with an AAAA answer.

Ideally, ipcache_entry::expires should be made optional to eliminate
awkward "first updateTtl() call" detection, but doing so well requires
significant code changes, so that entries without a known expiration
value are not cached forever _unless_ they were loaded from /etc/hosts.
And those changes should probably be propagated to fqdncache.cc.

### DIFF
--- a/src/ipcache.cc
+++ b/src/ipcache.cc
@@ -136,7 +136,12 @@ public:
 
     hash_link hash;     /* must be first */
     time_t lastref;
-    time_t expires;
+
+    /// The earliest time this entry will be considered stale. Entries without
+    /// an expiration time (e.g., current "static" entries) are considered
+    /// fresh. \sa ipcacheExpiredEntry()
+    std::optional<time_t> expires;
+
     ipcache_addrs addrs;
     IpCacheLookupForwarder handler;
     char *error_message;
@@ -330,7 +335,8 @@ ipcache_get(const char *name)
 static int
 ipcacheExpiredEntry(ipcache_entry * i)
 {
-    /* all static entries are locked, so this takes care of them too */
+    if (!i->expires)
+        return 0; // presumably a current "static" entry
 
     if (i->locks != 0)
         return 0;
@@ -339,7 +345,7 @@ ipcacheExpiredEntry(ipcache_entry * i)
         if (0 == i->flags.negcached)
             return 1;
 
-    if (i->expires > squid_curtime)
+    if (*i->expires > squid_curtime)
         return 0;
 
     return 1;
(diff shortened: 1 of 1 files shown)


rousskov:
> PR description: Ideally, ipcache_entry::expires should be made optional to eliminate awkward "first updateTtl() call" detection, but doing so well requires significant code changes, so that entries without a known expiration value are not cached forever unless they were loaded from /etc/hosts. And those changes should probably be propagated to fqdncache.cc.

FWIW, I actually implemented the initial changes implied above in branch commit 5fa0cf4 (reverted since then), but they were not enough, and those partial changes arguably created more problems than they solved. The above text represents the right direction for future improvements IMO, but this bug can use a more dedicated/smaller fix, while those future improvements may wait for basic ipcache and fqdncache infrastructure upgrade (dlink removal, entry locking fixes, etc.) -- there are a lot of low-level inconsistencies and gotchas in that code, complicating the "let's make ipcache_entry::expires optional" plan.

If there is no consensus regarding the paragraph quoted above, I will just remove it from the PR description.

rousskov:
Unfortunately, this fix is _not_ related to the recently discussed bug [5185](https://bugs.squid-cache.org/show_bug.cgi?id=5185), even though both problems are related to IP cache entries and may affect IP cache hit ratio.

src/ipcache.cc
```diff
@@ -543,8 +543,16 @@ ipcache_entry::updateTtl(const unsigned int rrTtl)
                                 Config.positiveDnsTtl); // largest value allowed
 
     const time_t rrExpires = squid_curtime + ttl;
-    if (rrExpires < expires)
+    const auto firstUpdate = addrs.size() <= 1;
+    if (firstUpdate) {
```
yadij, line 547: Please avoid the code bloat. Just do the check without adding extra variables.
```suggestion
    if (addrs.size() <= 1) {
```

yadij (changes requested):

src/ipcache.cc
rousskov: Done at [b6f23c6](https://github.com/squid-cache/squid/pull/1380/commits/b6f23c6890b3b97f5baee0033fd57d70b94eccc9).

yadij (approved):

PR #1381, opened 2023-06-15
Do not use raw pointers to index CARP CachePeers

eduard-bagdasaryan:
Simplified and improved code safety by using CbcPointers instead.

Also fixed mgr:carp Cache Manager reports to detail relevant
cache_peers instead of all cache_peers. When mgr:carp report was added
in 2000 commit 8ee9b49, Squid did not index (or even distinguish!)
CARP cache_peers, and the reporting loop naturally iterated through all
cache_peers. 2002 commit b399543 added identification and indexing of
CARP peers but forgot to adjust the reporting loop.

Very similar changes will be applied to userhash and sourcehash
cache_peers. 2008 commit 63104e2 simply copied problematic CARP code to
add userhash and sourcehash cache_peer support. This change adds a few
reusable types with those upcoming improvements in mind.

### DIFF
--- a/src/CachePeers.h
+++ b/src/CachePeers.h
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 1996-2023 The Squid Software Foundation and contributors
+ *
+ * Squid software is distributed under GPLv2+ license and includes
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID_CACHEPEERS_H
+#define SQUID_CACHEPEERS_H
+
+#include "base/forward.h"
+#include "CachePeer.h"
+#include "mem/PoolingAllocator.h"
+
+#include <vector>
+
+/// Weak pointers to zero or more Config.peers.
+/// Users must specify the selection algorithm and the order of entries.
+using SelectedCachePeers = std::vector< CbcPointer<CachePeer>, PoolingAllocator< CbcPointer<CachePeer> > >;
+
+/// Temporary, local storage of raw pointers to zero or more Config.peers.
+using RawCachePeers = std::vector<CachePeer *, PoolingAllocator<CachePeer*> >;
+
+#endif /* SQUID_CACHEPEERS_H */
+
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -199,6 +199,7 @@ squid_SOURCES = \
 	CacheManager.h \
 	CachePeer.cc \
 	CachePeer.h \
+	CachePeers.h \
 	ClientInfo.h \
 	ClientRequestContext.h \
 	CollapsedForwarding.cc \
--- a/src/carp.cc
+++ b/src/carp.cc
@@ -10,6 +10,7 @@
 
 #include "squid.h"
 #include "CachePeer.h"
+#include "CachePeers.h"
 #include "carp.h"
 #include "HttpRequest.h"
 #include "mgr/Registration.h"
@@ -22,8 +23,9 @@
 
 #define ROTATE_LEFT(x, n) (((x) << (n)) | ((x) >> (32-(n))))
 
-static int n_carp_peers = 0;
-static CachePeer **carp_peers = nullptr;
+/// CARP cache_peers ordered by their CARP weight
+static SelectedCachePeers TheCarpPeers;
+
 static OBJH carpCachemgr;
 
 static int
@@ -44,27 +46,22 @@ void
 carpInit(void)
 {
     int W = 0;
-    int K;
-    int k;
     double P_last, X_last, Xn;
-    CachePeer *p;
-    CachePeer **P;
     char *t;
     /* Clean up */
 
-    for (k = 0; k < n_carp_peers; ++k) {
-        cbdataReferenceDone(carp_peers[k]);
-    }
-
-    safe_free(carp_peers);
-    n_carp_peers = 0;
+    TheCarpPeers.clear();
 
     /* initialize cache manager before we have a chance to leave the execution path */
     carpRegisterWithCacheManager();
 
     /* find out which peers we have */
 
-    for (p = Config.peers; p; p = p->next) {
+    RawCachePeers rawCarpPeers;
+    for (auto p = Config.peers; p; p = p->next) {
+        if (!cbdataReferenceValid(p))
+            continue;
+
         if (!p->options.carp)
             continue;
 
@@ -73,24 +70,16 @@ carpInit(void)
         if (p->weight == 0)
             continue;
 
-        ++n_carp_peers;
+        rawCarpPeers.push_back(p);
 
         W += p->weight;
     }
 
-    if (n_carp_peers == 0)
+    if (rawCarpPeers.empty())
         return;
 
-    carp_peers = (CachePeer **)xcalloc(n_carp_peers, sizeof(*carp_peers));
-
-    /* Build a list of the found peers and calculate hashes and load factors */
-    for (P = carp_peers, p = Config.peers; p; p = p->next) {
-        if (!p->options.carp)
-            continue;
-
-        if (p->weight == 0)
-            continue;
-
+    /* calculate hashes and load factors */
+    for (const auto p: rawCarpPeers) {
         /* calculate this peers hash */
         p->carp.hash = 0;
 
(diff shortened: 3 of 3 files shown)


src/CachePeers.h
```diff
@@ -0,0 +1,26 @@
...
+#include "mem/PoolingAllocator.h"
+
+#include <vector>
+
+/// Weak pointers to zero or more Config.peers.
+/// Users must specify the selection algorithm and the order of entries.
+using SelectedCachePeers = std::vector< CbcPointer<CachePeer>, PoolingAllocator< CbcPointer<CachePeer> > >;
+
+/// Temporary, local storage of raw pointers to zero or more Config.peers.
+using RawCachePeers = std::vector<CachePeer *, PoolingAllocator<CachePeer*> >;
```
yadij, line 23: Please remove this type. We should never be storing raw-pointers

src/CachePeers.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include "base/forward.h"
+#include "CachePeer.h"
+#include "mem/PoolingAllocator.h"
+
+#include <vector>
+
+/// Weak pointers to zero or more Config.peers.
+/// Users must specify the selection algorithm and the order of entries.
+using SelectedCachePeers = std::vector< CbcPointer<CachePeer>, PoolingAllocator< CbcPointer<CachePeer> > >;
```
yadij, line 20: The name for this type is bad "selected" is a use-specific detail that should be part of a local variable name, not a global type.

src/CachePeers.h
yadij, line 23: Please remove this type. We should not be storing raw-pointers for these in any new code. The temporary vector should be storing the relevant `Pointer` type instead.

src/CachePeers.h
```diff
@@ -0,0 +1,26 @@
+/*
```
yadij, line 1: Please remove this file. Absent a `src/peer/forward.h` for defining things the type should be declared in `CachePeer.h` or local within the `carp.cc` where it is exclusively used.

yadij (changes requested):

src/CachePeers.h
rousskov: > the type should be declared in CachePeer.h or local within the carp.cc where it is exclusively used.

This type should not be declared in CachePeer.h because many CachePeer users do not need to know about various _collections_ of cache peers and, hence, do not need to be exposed to the associated overheads.

When we propagate this PR changes to the remaining two cache peer kinds (userhash and sourcehash) that have the same code/needs, this header file will be used by three source files -- no more exclusive use. Those userhash and sourcehash changes have been implemented already, so we know their needs well and will post the corresponding PR(s) shortly after this PR lands; that code is just waiting to be synced with the CARP changes happening during this PR review... This is a minimal PR focusing on one cache peer kind out of three similar ones.

src/CachePeers.h
rousskov: > "selected" is a use-specific detail that should be part of a local variable name, not a global type.

When several users share the same detail, expressing that shared concern via a type name is often a good idea. In this particular case, it is a good idea to emphasize that this collection of CachePeer object is not a collection of _all_ configured cache_peers. This PR fixes at least one bug that is directly related to the difference between "all peers" and "selected peers". Moreover, for cbdata-protected types, the difference between "selected" and "all" cache_peers affects stored pointer type/validity and essentially _requires_ different container types (hence, different names).

This type name reflects those ideas, even though it cannot and does not not define how the selection is performed by each type user. Needless to say, a specific type user (that wants to reuse this type in several places) can create a user-specific alias that will identify the selection process by name. Doing so without duplicating code requires this type alias to exist.

src/CachePeers.h
rousskov: > We should not be storing raw-pointers for these in any new code.

I disagree with that assertion. Given the expense associated with smart pointers, especially cbdata pointers, _temporary_ storage of raw pointers is the best solution for use cases like this:

```C++
if (const auto raw = smart.valid())
    raw->doSomething();
```

This type is used for exactly the same purpose; the only difference is that the corresponding method call has to be split across two loops, necessitating (temporary) storage between the loops.

> The temporary vector should be storing the relevant `Pointer` type instead.

Why? The current PR code that uses this type for temporary (function execution-scoped) storage of selected/validated cache peers is better than the proposed code that would abuse a cbdata-protected pointer for the same purpose.

The only viable alternative that I can see is storing [references](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper) instead of raw pointers. However, since we then need to create CbcPointers, I think the current approach is slightly better -- creating a long-term stored pointer from a reference raises arguably bigger red flags that temporary storing a raw pointer.

src/CachePeers.h
yadij: Disagree. The reason you have presented against my request can be applied to every single API ever created as an argument for placing every single symbol they define in a unique filename. That is not a reasonable way to define APIs.
All I am asking for is to use the Best Practice of aggregating related API symbols into the same files.

src/CachePeers.h
yadij: Under the principles of C++ RAII the use of ```vector``` may result in either leaked memory allocation, or crashes on double-free. Assuming that all code changes in future will always correctly account for both those risks is a **Bad Practice**.

Relating to performance, the heavy memory operations from ```Pointer``` are largely offset against the removal of delayed conversion to ```Pointer```. The remaining weight is largely due to inefficiency in our cbdata ```Pointer``` design which is a separate performance issue which will gain Squid much larger benefits if fixed at the definition itself.

yadij (dismissed):
Unresolved discussions from earlier review.

src/CachePeers.h
rousskov: > The reason you have presented against my request can be applied to every single API ever created as an argument for placing every single symbol they define in a unique filename. That is not a reasonable way to define APIs.

Agreed. An obvious misapplication of a reason does not prove that the reason itself is invalid. We do not want to put every name into its own unique file or put all names into the same file. Both extremes are obviously wrong.

>  All I am asking for is to use the Best Practice of aggregating related API symbols into the same files.

The PR code already does that. It is just difficult to see that there are enough CachePeers (plural) stuff to warrant the creation of a dedicated file because we use minimal PRs. We already have code changes that will add more CachePeers (plural) stuff into CachePeers.h. Those APIs do not need to go into CachePeer (singular) files, for the reasons already stated. There is a CachePeer API and a CachePeers API. Each deserve their own files.

src/CachePeers.h
rousskov: > This is a type for storing a non-singular amount of CachePeer - thus "CachePeers" is more correct.

If we had just one kind of CachePeer object collection, that conclusion would be correct. In reality, there are several different collections of CachePeer objects, including these three:

* `CachePeers` is the (future name) of the type representing `Config.peers`, `Config.npeers`, and associated functionality -- the object that owns all configured peers. Besides CachePeer ownership, it encapsulates configuration-related and peer polling functionality that other containers do not care about. This type does not exist as a named C++ type in the official code or this PR, but the underlying code does exist. It just lacks proper encapsulation. We are working on that, but we need to get this and several other "minimal" helper PRs out of the way first.

* `SelectedCachePeers` may not contain all configured peers and, today, it does not own any of them (because it is a collection of _weak_ cbdata pointers). Like CachePeers, it is suitable for long-term storage of peers, but without the configuration-related baggage that the CachePeers type in the first bullet has to carry.

* `RawCachePeers`, like SelectedCachePeers, may not contain all configured peers and does not own any of them (because it is a collection of raw pointers). Unlike the other two types, it is not suitable for long-term storage of peers.

The three types address three different needs in the current official code. This PR introduces two of them. Another PR (already in the works) will introduce CachePeers. Eventually, these types will change and some may disappear, but different kinds of CachePeer collections will remain for the foreseeable future.

> "selected" has no relevance at the level of type definition - it should be in the local variable name for callers where that property is true/relevant.

Selection will always be true/relevant for this type. I am open to spelling improvements, of course, but CachePeers is not a good spelling option for SelectedCachePeers because we want to emphasize that these are not necessarily all configured peers and because CachePeers spelling is best used for the primary collection of all configured CachePeer objects (as detailed in the first bullet above). And the difference between the two types is not just about a subset of peers or CachePeer ownership. There is a lot more baggage that the type in the first bullet has to deal with and other types do not care about.

rousskov:

kinkie:
let's ship it :)

PR #1382, opened 2023-06-15
Bug 5278: Log %err_code for "early" request handling errors

rousskov:
Certain "early" request handling errors (e.g., ERR_TOO_BIG,
ERR_UNSUP_REQ, ERR_UNSUP_HTTPVERSION, ERR_PROTOCOL_UNKNOWN, and
ERR_INVALID_URL) are handled with nil ClientHttpRequest::request object.
ErrorState relied on that object to record error type and details, so
%err_code expanded to "-" in the corresponding access.log records.

Now, ErrorState also uses ALE (where available). ALE already handles the
combination of an early error and late HttpRequest (e.g., the fake one
clientReplyContext::createStoreEntry() creates when handling the error).

This minimal fix does not address known problems with ErrorState::detail
and ErrorState::BuildHttpReply(). It also duplicates related xerrno
detailing code. We will address some of these problems shortly,
including the increased code duplication problem.

### DIFF
--- a/src/errorpage.cc
+++ b/src/errorpage.cc
@@ -1352,6 +1352,11 @@ ErrorState::BuildHttpReply()
             request->detailError(type, detail);
         else
             request->detailError(type, SysErrorDetail::NewIfAny(xerrno));
+    } else if (ale) {
+        if (detail)
+            ale->updateError(Error(type, detail));
+        else
+            ale->updateError(Error(type, SysErrorDetail::NewIfAny(xerrno)));
     }
 
     return rep;


src/errorpage.cc
```diff
@@ -1352,6 +1352,11 @@ ErrorState::BuildHttpReply()
             request->detailError(type, detail);
         else
             request->detailError(type, SysErrorDetail::NewIfAny(xerrno));
+    } else if (ale) {
+        if (detail)
+            ale->updateError(Error(type, detail));
+        else
+            ale->updateError(Error(type, SysErrorDetail::NewIfAny(xerrno)));
     }
```
rousskov, line 1360: I am essentially duplicating the xerrno-handling code snippet here. There is already an ErrorState TODO about encapsulating that detail extraction/creation logic:

```C++
    // TODO: Replace type, xerrno and detail with Error while adding a virtual
    // Error::Detail::sysError() method to extract errno in detailError().
    /// type-specific detail about the transaction error;
    /// overwrites xerrno;
    ErrorDetail::Pointer detail;
```

I will address that TODO in a dedicated PR, removing this (and other, similar) code duplication in the process. Unlike this simple bug-fixing PR, that followup refactoring PR may not be suitable for backporting, so I want to get this bug fix (and another ErrorState fix that is waiting for #1368) merged first.

src/errorpage.cc
```diff
@@ -1352,6 +1352,11 @@ ErrorState::BuildHttpReply()
...
```
rousskov, line 1361: This small fix does not address some of the Bug 5278 concerns. For example, it does not affect parsing and logging of HTTP headers behind an unexpected relative URL. However, it is a significant improvement because we now inform the admins of what went wrong by logging `%err_code`. The (somewhat controversial) processing/logging of related _details_ can be left to future work/PRs.

yadij (approved):

PR #1383, opened 2023-06-16
CI: Remove pass-through test-functionality test wrappers

rousskov:
Instead of requiring a custom test wrapper for each test and, hence,
creating an ever-increasing number of pass-through wrappers that do
nothing useful, use a custom test wrapper if and only if it exists. By
default (i.e. when there is no custom wrapper), just run the named test.

As a positive side effect, this change also simplifies running tests
that are not on the $default_tests list hard-coded in main():

    test-suite/test-functionality.sh some-new-test-i-am-working-on

### DIFF
--- a/test-suite/test-functionality.sh
+++ b/test-suite/test-functionality.sh
@@ -165,14 +165,6 @@ run_confirmed_test() {
     return $result
 }
 
-check_pconn() {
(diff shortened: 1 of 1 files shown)


kinkie (approved):
looks legit

PR #1384, opened 2023-06-19
Windows: Drop obsolete WinSock v1 library

yadij:
wsock32 library and its winsock.h are for Windows 95
and older version socket support. We no longer need
Squid to run on such old Windows machines.

### DIFF
--- a/acinclude/os-deps.m4
+++ b/acinclude/os-deps.m4
@@ -705,7 +705,7 @@ AC_CACHE_CHECK(for _res.nsaddr_list, ac_cv_have_res_nsaddr_list,
 ])
 
 
-dnl checks the ws2_32 library to use
+dnl checks whether to use the ws2_32 library
 dnl may set ac_cv_func_select as a side effect
 AC_DEFUN([SQUID_CHECK_WINSOCK_LIB],[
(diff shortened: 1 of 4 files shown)


acinclude/os-deps.m4
```diff
@@ -715,16 +705,14 @@ AC_CACHE_CHECK(for _res.nsaddr_list, ac_cv_have_res_nsaddr_list,
 ])
 
 
-dnl checks the winsock library to use (ws2_32 or wsock32)
+dnl checks the ws2_32 library to use
```
rousskov, line 708: Just a wording polish. AFAICT, this code may end up selecting no library, selecting ws2_32 library, or decide to use some other (unknown to this code by name) winsock library from LIBS.

Simple/direct:

```suggestion
dnl checks whether to use the ws2_32 library
```

or, to match our own `AC_MSG_CHECKING([for winsock library])` inside this code:

```suggestion
dnl checks the winsock library to use
```

I do not insist on this change, and do not want to spend more time on this minor issue. If neither suggestion is an improvement in your opinion, please just resolve this change request without a discussion.

rousskov (approved):

PR #1385, opened 2023-06-19
Windows: Fix symbol collisions

yadij:
'near' is a macro defined by the Windows API.
We cannot use it for anything.


src/errorpage.cc
```diff
@@ -1430,13 +1435,13 @@ ErrorState::compile(const char *input, bool building_deny_info_url, bool allowRe
 
 /// react to a compile() error
 /// \param msg  description of what went wrong
-/// \param near  approximate start of the problematic input
+/// \param data  approximate start of the problematic input
```
rousskov, line 1438: Please avoid `data`, `state`, `object` and similar names that can mean nearly anything. For the `near` data member and the related parameters, I can suggest approximateInput, inputArea, or inputStart as a replacement (and an improvement compared to "near"), but I do not insist on any specific spelling.

rousskov (changes requested):
Thank you for fixing this name clash.

src/errorpage.cc
yadij: The choice of name was given because this data can **be** literally anything.

rousskov: > The choice of name was given because this data can **be** literally anything.

While any individual character inside this c-string parameter can be nearly anything[^1], the parameter should be naming that string _as a whole_. That string (as a whole) has a fairly specific meaning. That meaning allows this function to actually _use_ that parameter. That meaning is already documented (this change request is attached to that documentation). The new parameter name should (continue to) reflect that meaning.

If you would like to emphasize the fact that this input has not been validated or sanitized, then I can suggest adding the word "raw" to the already proposed parameter names -- approximateRawInput, rawInputArea, or rawInputStart -- but I do not insist on that addition or any specific spelling.

[^1]: any character but NUL

yadij: Exactly. "data" is the most specific name for what is contained.

rousskov: Repeating assertions, especially those that have been discussed and proven false, does not help advance this PR forward.

yadij: My problem is this is **not** a "c-string", nor the input, nor the start of anything in the input.

Would you accept "ptr"  ?

rousskov: > Would you accept "ptr" ?

I would rather not because "ptr" has the same problem as "data" plus a few additional problems.

> My problem is this is not a "c-string", nor the input, nor the start of anything in the input.

I believe that assertion is false, but we do not have to debate its correctness because most of the names I proposed do not claim that this parameter is "a c-string", "the input", or "the start of anything in the input": approximateInput, inputArea, approximateRawInput, and rawInputArea.

FWIW, all existing callers (that I could find) pass `build.input` as the actual parameter:

```C++
noteBuildError("Bad @Squid{logformat} sequence", build.input);
bypassBuildErrorXXX("Unsupported deny_info %code", build.input);
bypassBuildErrorXXX("Unsupported error page %code", build.input);
```

AFAICT, that `build.input` is a c-string that comes from sources like these:

```
rawDetail.c_str()
ErrorPage::IsDenyInfoUri(page_id)
error_text[page_id]
localeTmpl.text()
```

It is also treated as a c-string, twice, inside this ErrorPage::BuildErrorPrinter::print() method:

```C++
if (strlen(near) > maxContextLength)
```

and

```
os << near;
```

kinkie: In order to move forward this PR, I've made a proposal to call the parameter errorLocation. Hopefully this will break the tie and allow moving this forward

kinkie (approved):

src/errorpage.h
```diff
@@ -207,7 +207,7 @@ class ErrorState
     HttpReplyPointer response_;
 
 private:
-    void noteBuildError_(const char *msg, const char *near, const bool forceBypass);
+    void noteBuildError_(const char *, const char *, const bool forceBypass);
```
rousskov, line 210: We tend to name parameters in declarations when their meaning cannot be easily guessed (e.g., from the parameter type). This is one of such cases:

```suggestion
    void noteBuildError_(const char *msg, const char *errorLocation, bool forceBypass);
```

src/errorpage.cc
rousskov: IMO, errorLocation is not a good choice, especially because it clashes badly with inputLocation name that is already used in the same context. Unlike the inputLocation parameter, this `near` parameter is not really a _location_ of X. It is X itself. Here, X is, roughly speaking, "problematic input" (i.e. something that triggered an error). Not an "error" itself.

However, in the interest of progress, I am _not_ going to block this PR if it uses errorLocation intstead of `data` -- errorLocation is better than `data` and, given the unfortunate resistance to better names, can be considered "good enough". Thus, I am resolving this change request.

This change request was not specific to the code line it was attached to. I have added the corresponding change requests to other changes introducing unwanted `data` names.

rousskov (changes requested):
Thank you for trying to push this PR forward. I do not insist on fixing any problems other than the removal of `data` (i.e. the old change request).

kinkie (approved):
Simple refactoring, clearing for merge

PR #1386, opened 2023-06-19
Fix several missing includes

yadij:
error: fd_table was not declared in this scope

### DIFF
--- a/src/icp_v2.cc
+++ b/src/icp_v2.cc
@@ -23,6 +23,7 @@
 #include "comm/Connection.h"
 #include "comm/Loops.h"
 #include "fd.h"
+#include "fde.h"
 #include "HttpRequest.h"
 #include "icmp/net_db.h"
 #include "ICP.h"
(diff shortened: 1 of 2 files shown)


src/icp_v2.cc
```diff
@@ -23,6 +23,7 @@
 #include "comm/Connection.h"
 #include "comm/Loops.h"
 #include "fd.h"
+#include "fde.h"
```
rousskov, line 26: > `error: fd_table was not declared in this scope`

How does fd_table get into src/icp_v2.cc code? I do not see any direct use or any FD_READ_METHOD()/FD_WRITE_METHOD() calls in src/icp_v2.cc, but my build environment is different. I am wondering whether we should `#include` this header somewhere else instead (to prevent not just this, but similar build errors in the future).

src/snmp_core.cc
```diff
@@ -18,6 +18,7 @@
 #include "comm/Connection.h"
 #include "comm/Loops.h"
 #include "fatal.h"
+#include "fde.h"
```
rousskov, line 21: Same question here: What src/snmp_core.cc scope does `fd_table was not declared in this scope` error message refer to?

PR #1387, opened 2023-06-19
Fix PRIuSIZE definition

yadij:
This printf format-string helper was defined twice
with inconsistent criteria and values.

Define an M4 macro to perform the code checks in
./configure and allow other types to also reduce
their tests in future.

Also, fix several format strings which were using
bad %codes. Identified by MinGW cross-compile.

### DIFF
--- a/acinclude/printf.m4
+++ b/acinclude/printf.m4
@@ -1,0 +1,33 @@
+## Copyright (C) 1996-2023 The Squid Software Foundation and contributors
+##
+## Squid software is distributed under GPLv2+ license and includes
+## contributions from numerous individuals and organizations.
+## Please see the COPYING and CONTRIBUTORS files for details.
+##
+
+dnl check which printf code to use for a type
+dnl $1 - the type being checked
+dnl $2 - the PRIxxx macro to define
+dnl $3 - a list of possible codes
+AC_DEFUN([SQUID_CHECK_PRINTF_CODES],[
+  AH_TEMPLATE([$2],[printf display of $1])
+  AC_CACHE_CHECK([which printf code displays $1],
+                 [squid_cv_printf_$2],[
+    SQUID_STATE_SAVE($2)
+    CXXFLAGS="$CXXFLAGS -Werror=format"
+    for code in $3; do
+      AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+#include <cstdio>
+      ]],[[$1 val={};printf("%$code", val);]])],[
+        squid_cv_printf_$2=$code
+        break
+      ])
+    done
+    SQUID_STATE_ROLLBACK($2)
+  ])
+  AS_IF([test -z "$squid_cv_printf_$2"],
+    AC_MSG_FAILURE([no printf support for $1 (tried: $3)])
+  )
+  AC_DEFINE_UNQUOTED($2,["$squid_cv_printf_$2"])
+  AC_MSG_RESULT($squid_cv_printf_$2)
+])
--- a/configure.ac
+++ b/configure.ac
@@ -25,6 +25,7 @@ m4_include([acinclude/ldap.m4])
 m4_include([acinclude/nettle.m4])
 m4_include([acinclude/pam.m4])
 m4_include([acinclude/pkg.m4])
+m4_include([acinclude/printf.m4])
 m4_include([acinclude/tdb.m4])
 m4_include([acinclude/lib-checks.m4])
 m4_include([acinclude/ax_cxx_compile_stdcxx.m4])
@@ -2393,6 +2394,8 @@ AC_TYPE_INT64_T
 AC_TYPE_UINT64_T
 AC_TYPE_PID_T
 AC_TYPE_SIZE_T
+AC_CHECK_SIZEOF(size_t)
+SQUID_CHECK_PRINTF_CODES([size_t],[PRIuSIZE],[zu llu lu u I64u I32u])
 AC_TYPE_SSIZE_T
 AC_TYPE_OFF_T
 AC_TYPE_UID_T
@@ -2402,7 +2405,6 @@ AC_CHECK_SIZEOF(int64_t)
 AC_CHECK_SIZEOF(long)
 #need the define for overflow checks
 AC_CHECK_SIZEOF(off_t)
-AC_CHECK_SIZEOF(size_t)
 
 dnl On Solaris 9 x86, gcc may includes a "fixed" set of old system include files
 dnl that is incompatible with the updated Solaris header files.
--- a/src/acl/external/LM_group/ext_lm_group_acl.cc
+++ b/src/acl/external/LM_group/ext_lm_group_acl.cc
@@ -569,7 +569,7 @@ main(int argc, char *argv[])
         if ((p = strchr(buf, '\r')) != NULL)
             *p = '\0';      /* strip \r */
 
-        debug("Got '%s' from Squid (length: %d).\n", buf, strlen(buf));
+        debug("Got '%s' from Squid (length: %" PRIuSIZE ").\n", buf, strlen(buf));
 
         if (buf[0] == '\0') {
             SEND_BH(HLP_MSG("Invalid Request."));
--- a/src/ssl/crtd_message.cc
+++ b/src/ssl/crtd_message.cc
@@ -129,7 +129,7 @@ std::string Ssl::CrtdMessage::compose() const
 {
     if (code.empty()) return std::string();
     char buffer[10];
-    snprintf(buffer, sizeof(buffer), "%zd", body.length());
+    snprintf(buffer, sizeof(buffer), "%" PRIuSIZE, body.length());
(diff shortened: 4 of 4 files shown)


acinclude/printf.m4
```diff
@@ -0,0 +1,30 @@
...
+      AC_LINK_IFELSE([AC_LANG_PROGRAM([[
+#include <cstdio>
+      ]],[[$1 val={};printf("%$code", val);]])],[
+        squid_cv_printf_$2=$code
+        break
+      ])
+    done
+    SQUID_STATE_ROLLBACK($2)
+  ])
+  AC_DEFINE_UNQUOTED($2,["$squid_cv_printf_$2"])
```
rousskov, line 28: AFAICT, this AC_DEFINE_UNQUOTED() defines an empty PRIuSIZE when the loop above does not find any supported %code. Instead, please either warn and use the first %code from `$3` OR fail with an error.

acinclude/printf.m4
```diff
@@ -0,0 +1,30 @@
...
+dnl check which printf code to use for a type
+dnl $1 - the type being checked
+dnl $2 - the PRIxxx macro to define
+dnl $3 - a list of possible codes
+AC_DEFUN([SQUID_CHECK_PRINTF_CODES],[
+  AH_TEMPLATE([$2],[printf display of $1])
+  AC_CACHE_CHECK([which printf code displays $1],
+                 [squid_cv_printf_$2],[
+    SQUID_STATE_SAVE($2)
+    CXXFLAGS="-Werror=format"
```
rousskov, line 17: We cannot expect all compilers to support that option. I am worried that when the compiler ignores this option, we may build Squid with the wrong printf `%code`, and when the compiler refuses to run with that option, we may not find the right printf() code at all.

Also, we should not just completely overwrite CXXFLAGS because that may break building the test program below when building _anything_ requires custom CXXFLAGS (supplied by the admin).

Since the first given %code is supposed to be the most portable/universal, and since `-Wformat` is enabled by `-Wall` in GCC, I suggest removing this CXXFLAGS customization. We may still use the wrong %code in some cases, but we will cover more cases with simpler code this way. Moreover, the official code is already using this approach AFAICT.

```suggestion
```

If removing this line means that we can also remove SQUID_STATE_SAVE/SQUID_STATE_ROLLBACK, please remove those as well, of course.

configure.ac
```diff
@@ -2393,18 +2394,17 @@ AC_TYPE_INT64_T
 AC_TYPE_UINT64_T
 AC_TYPE_PID_T
 AC_TYPE_SIZE_T
+AC_CHECK_SIZEOF(size_t)
+SQUID_CHECK_PRINTF_CODES([size_t],[PRIuSIZE],[zu llu lu u I64u I32u])
```
rousskov, line 2398: AFAICT, all master/v7 code using PRIuSIZE today is C++ code. Can we just switch that code to [std::printf/etc.](https://en.cppreference.com/w/cpp/io/c/fprintf) and `%zu` format code instead of trying to guess the right format code at ./configure time? AFAICT, doing so will improve Squid code and automatically address all other change requests.

rousskov (changes requested):
Thank you for fixing these bugs!

acinclude/printf.m4
yadij: Regarding compiler support: All the compilers we officially support accept this flag. If we need to add support for one that has different options we can adjust this line at that time.

Why we need it: this format check is disabled in the current set of flags used Squid-wide. Without it any of the options that size_t can implicitly convert to (regardless of value wrap/loss) will be detected as valid format codes.

I will fix this to be an append to user flags though. Thanks for that catch.

configure.ac
yadij: Hmm, good point. The odd codes were there for MSVC which we still hope to support again one day, but looking into that it was updated in 2013 to support %zu.

acinclude/printf.m4
rousskov: > All the compilers we officially support accept this flag.

I do not recall us agreeing on the list of supported compilers but, more importantly, the primary problem here is _not_ with the compilers we test with:

> If we need to add support for one that has different options we can adjust this line at that time.

The problem here is not so much with adjusting when we _know_ that we need to adjust. The problem is an admin building Squid with a compiler that does not support this option, with the build _silently_ failing to detect the right %code and using the wrong %code because of that lack of support (without our or admin knowledge about any of that).

> this format check is disabled in the current set of flags used Squid-wide

What makes you think that? GCC documentation says that `-Wall` enables `-Wformat`. Clang [documentation](https://clang.llvm.org/docs/DiagnosticsReference.html#wformat) says that `-Wformat` is enabled by default, and my [test](https://godbolt.org/z/r5YPd3sn6) supports that claim. Squid enables `-Wall` and `-Werror` by default (where supported).

> Without it any of the options that size_t can implicitly convert to (regardless of value wrap/loss) will be detected as valid format codes.

I do not think that is an accurate assumption because of the -Wall and defaults discussed above.

In summary, explicitly adding `-Wformat` is not needed for common Squid builds and may create/cause problems in some environments. Omitting that option may create problems in some other environments, but those problems are likely to extend to the whole Squid code, not just this test case. Thus, it looks like we should not add this compiler flag here. When/if we discover specific environments where it (or its equivalent) is needed, we will enable it globally for those cases, not just here.

However, perhaps I am missing something that has not been (dis)covered yet. If you remove this configure.ac line, does the overall outcome change in your MinGW test environment? If it does, there may be another bug we need to fix somewhere!

configure.ac
rousskov: Glad to hear that. Let's focus on this promising direction! We may discover that it is not working, but I think it is worth investigating.

rousskov: #1388 removes PRIuSIZE.

yadij:
Status update FTR; leaving this in Draft state for now in case we want to use the M4 macro for checking any other format codes. There is some work doing cleanup of other similar macros and getting a ```-Wformat```-clean build before we make a final call on use/drop-ing this PR.

acinclude/printf.m4
```diff
@@ -0,0 +1,30 @@
...
+    for code in $3; do
+      AC_LINK_IFELSE([AC_LANG_PROGRAM([[
```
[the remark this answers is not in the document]
yadij: Done in [3d746d2](https://github.com/squid-cache/squid/pull/1387/commits/3d746d2cdce91fbae23109ef3faaf1b5c2803c61)

acinclude/printf.m4
yadij: Done in [3d746d2](https://github.com/squid-cache/squid/pull/1387/commits/3d746d2cdce91fbae23109ef3faaf1b5c2803c61).

FTR; I am going by https://github.com/squid-cache/squid/blob/master/doc/release-notes/release-6.sgml.in#L27 for my definition of "officially supported compilers".

This line was tested with GCC 12 in both MinGW-w64 cross-build and Linux native build. Using default ```./configure``` options. Several order permutations of the format selection and also ```%az``` (invalid) were used. Only the explicit addition of this flag caused ``%u``` to produce the expected error about bad down-casting on my 64-bit machine and continuing through to find ```%zu``` or ```%llu``` whichever was next (those being equivalent on 64-bit builds).
 There are some weird things going on with flags set (or not) by some tests, and misuse of some flags in our configure scripts. For now I think this is "good enough" and removal can wait on future work cleaning up that mess.

acinclude/printf.m4
yadij: Fixed in [3d746d2](https://github.com/squid-cache/squid/pull/1387/commits/3d746d2cdce91fbae23109ef3faaf1b5c2803c61)

acinclude/printf.m4
```diff
@@ -0,0 +1,33 @@
...
+dnl check which printf code to use for a type
+dnl $1 - the type being checked
+dnl $2 - the PRIxxx macro to define
+dnl $3 - a list of possible codes
+AC_DEFUN([SQUID_CHECK_PRINTF_CODES],[
+  AH_TEMPLATE([$2],[printf display of $1])
+  AC_CACHE_CHECK([which printf code displays $1],
+                 [squid_cv_printf_$2],[
+    SQUID_STATE_SAVE($2)
+    CXXFLAGS="$CXXFLAGS -Werror=format"
```
rousskov, line 17: > There are some weird things going on with flags set (or not) by some tests

Yes, sounds like it.

> For now I think this is "good enough" and removal can wait on future work cleaning up that mess.

Since this addition is a temporary and _problematic_ workaround for other (yet-to-be-identified) ./configure bugs, it should be marked as such with a source code comment. For example,

```suggestion
    CXXFLAGS="$CXXFLAGS -Werror=format" # XXX: This should not be needed for most builds (and breaks the test for some others)
```

acinclude/printf.m4
rousskov: Moved to https://github.com/squid-cache/squid/pull/1387#discussion_r1237259277

PR #1388, opened 2023-06-20
Drop PRIuSIZE macro

yadij:
C++11 requires compiler support for %zu so
we no longer need to check for it.

### DIFF
--- a/compat/types.h
+++ b/compat/types.h
@@ -125,21 +125,6 @@
 #endif
 #endif
 
-#ifndef PRIuSIZE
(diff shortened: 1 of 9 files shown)


rousskov (approved):
Thank you for removing this ugly hack!

> C++11 requires compiler support for %zu

FWIW, I am not sure that (correct) claim is applicable because C++ requirements are for std::printf() and friends, not the C functions we are using. I suggest adding something like `, and printf() implementations we know about support it, `, but I do not insist on this adjustment.

PR #1389, opened 2023-06-21
Keep ::helper objects alive while in use by helper_servers

eduard-bagdasaryan:
Squid creates ::helper objects to manage a given configured helper. For
each ::helper object, Squid may create one or more helper_server objects
to manage communication with individual helper processes. A
helper_server object uses its so called "parent" ::helper object to
access configuration and for helper process death notification purposes.
There are no checks that the "parent" object is still alive when used.

The same problem applies to statefulhelper and helper_stateful_server.

helper_server code evidently attempted to extend their "parent" lifetime
using cbdata, but that does not work, especially in C++ world where the
object is destructed (and, hence, enters an undefined state) even though
its top-level memory is preserved by cbdata. Non-trivial members like
helper::queue (std::queue) and statefulhelper::reservations
(std::unordered_map) release their internal memory when destructed.

We now refcount ::helper objects to keep them alive until the last
object using them is gone. This does not result in reference loops
because the ::helper object uses raw dlink_node pointers to store its
helper_servers.

The following helpers (listed here by their container names) were
destructed while possibly still in use by their helper_server objects:
external_acl::theHelper, Ssl::CertValidationHelper::ssl_crt_validator,
Ssl::Helper::ssl_crtd. The following helpers are not destructed during
reconfiguration: redirectors, storeIds, basicauthenticators,
ntlmauthenticators, negotiateauthenticators, and digestauthenticators
(even though casual reading of relevant code may suggest otherwise).

This bug fix does not address or mark many remaining helper bugs.

### DIFF
--- a/src/auth/basic/Config.cc
+++ b/src/auth/basic/Config.cc
@@ -36,7 +36,7 @@
 /* Basic Scheme */
 static AUTHSSTATS authenticateBasicStats;
 
-HelperPointer basicauthenticators = nullptr;
+HelperPointer basicauthenticators;
 
 static int authbasic_initialised = 0;
 
--- a/src/auth/basic/Config.cc
+++ b/src/auth/basic/Config.cc
@@ -36,7 +36,7 @@
 /* Basic Scheme */
 static AUTHSSTATS authenticateBasicStats;
 
-helper *basicauthenticators = nullptr;
+HelperPointer basicauthenticators = nullptr;
 
 static int authbasic_initialised = 0;
 
@@ -109,7 +109,6 @@ Auth::Basic::Config::done()
         helperShutdown(basicauthenticators);
     }
 
-    delete basicauthenticators;
     basicauthenticators = nullptr;
 
     if (authenticateProgram)
@@ -305,7 +304,7 @@ Auth::Basic::Config::init(Auth::SchemeConfig *)
         authbasic_initialised = 1;
 
         if (basicauthenticators == nullptr)
-            basicauthenticators = new helper("basicauthenticator");
+            basicauthenticators = helper::Make("basicauthenticator");
 
         basicauthenticators->cmdline = authenticateProgram;
 
--- a/src/auth/basic/Config.cc
+++ b/src/auth/basic/Config.cc
@@ -36,7 +36,7 @@
 /* Basic Scheme */
 static AUTHSSTATS authenticateBasicStats;
 
-HelperPointer basicauthenticators;
+Helper::ClientPointer basicauthenticators;
 
 static int authbasic_initialised = 0;
 
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -46,7 +46,7 @@
 
 static AUTHSSTATS authenticateDigestStats;
 
-HelperPointer digestauthenticators = nullptr;
+HelperPointer digestauthenticators;
 
 static hash_table *digest_nonce_cache;
 
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -46,7 +46,7 @@
 
 static AUTHSSTATS authenticateDigestStats;
 
-helper *digestauthenticators = nullptr;
+HelperPointer digestauthenticators = nullptr;
 
 static hash_table *digest_nonce_cache;
 
@@ -525,7 +525,7 @@ Auth::Digest::Config::init(Auth::SchemeConfig *)
         authdigest_initialised = 1;
 
         if (digestauthenticators == nullptr)
-            digestauthenticators = new helper("digestauthenticator");
+            digestauthenticators = helper::Make("digestauthenticator");
 
         digestauthenticators->cmdline = authenticateProgram;
 
@@ -559,7 +559,6 @@ Auth::Digest::Config::done()
     if (!shutting_down)
         return;
 
-    delete digestauthenticators;
     digestauthenticators = nullptr;
 
     if (authenticateProgram)
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -46,7 +46,7 @@
 
 static AUTHSSTATS authenticateDigestStats;
 
-HelperPointer digestauthenticators;
+Helper::ClientPointer digestauthenticators;
 
 static hash_table *digest_nonce_cache;
 
--- a/src/auth/negotiate/Config.cc
+++ b/src/auth/negotiate/Config.cc
@@ -32,7 +32,7 @@
 
 static AUTHSSTATS authenticateNegotiateStats;
 
-StatefulHelperPointer negotiateauthenticators = nullptr;
+StatefulHelperPointer negotiateauthenticators;
 
 static int authnegotiate_initialised = 0;
 
--- a/src/auth/negotiate/Config.cc
+++ b/src/auth/negotiate/Config.cc
@@ -32,7 +32,7 @@
 
 static AUTHSSTATS authenticateNegotiateStats;
 
-statefulhelper *negotiateauthenticators = nullptr;
+StatefulHelperPointer negotiateauthenticators = nullptr;
 
 static int authnegotiate_initialised = 0;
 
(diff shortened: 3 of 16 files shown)


src/auth/basic/Config.cc
```diff
@@ -36,7 +36,7 @@
 /* Basic Scheme */
 static AUTHSSTATS authenticateBasicStats;
 
-helper *basicauthenticators = nullptr;
+HelperPointer basicauthenticators = nullptr;
```
yadij, line 39: Pointer types should never need explicit ```nullptr``` instantiation.
```suggestion
HelperPointer basicauthenticators;
```

src/auth/negotiate/Config.cc
```diff
@@ -32,7 +32,7 @@
 
 static AUTHSSTATS authenticateNegotiateStats;
 
-statefulhelper *negotiateauthenticators = nullptr;
+StatefulHelperPointer negotiateauthenticators = nullptr;
```
yadij, line 35:
```suggestion
StatefulHelperPointer negotiateauthenticators;
```

src/auth/digest/Config.cc
```diff
@@ -46,7 +46,7 @@
 
 static AUTHSSTATS authenticateDigestStats;
 
-helper *digestauthenticators = nullptr;
+HelperPointer digestauthenticators = nullptr;
```
yadij, line 49:
```suggestion
HelperPointer digestauthenticators;
```

src/auth/ntlm/Config.cc
```diff
@@ -33,7 +33,7 @@
 /* NTLM Scheme */
 static AUTHSSTATS authenticateNTLMStats;
 
-statefulhelper *ntlmauthenticators = nullptr;
+StatefulHelperPointer ntlmauthenticators = nullptr;
```
yadij, line 36:
```suggestion
StatefulHelperPointer ntlmauthenticators;
```

src/redirect.cc
```diff
@@ -54,8 +54,8 @@ class RedirectStateData
 
 static HLPCB redirectHandleReply;
 static HLPCB storeIdHandleReply;
-static helper *redirectors = nullptr;
-static helper *storeIds = nullptr;
+static helper::Pointer redirectors = nullptr;
+static helper::Pointer storeIds = nullptr;
```
yadij, line 58:
```suggestion
static helper::Pointer redirectors;
static helper::Pointer storeIds;
```

src/helper/forward.h
```diff
@@ -24,6 +26,9 @@ class Request;
 
 } // namespace Helper
 
+using HelperPointer = RefCount<helper>;
+using StatefulHelperPointer = RefCount<statefulhelper>;
```
yadij, line 30: Please avoid defining multiple different types referring to the same thing. Specifically the ```helper::Pointer``` and ```statefulhelper::Pointer``` types duplicate these in a confuse-able way.

FWIW, my rule of thumb is to prefer the class-scoped definitions whenever possible. However if the code needs global-scope types (as this helper code might), move everything to use those instead. So each class with a ```Pointer``` in public ```.h``` files only has one of the two styles. Type symbols defined locally in a ```.cc``` function do not count for this rule.

yadij, line 30: If you keep these new types, please move into the ```namespace Helper``` scope.

src/helper.h
```diff
@@ -60,26 +61,15 @@ class HelperServerBase;
...
-        last_queue_warn(0),
-        last_restart(0),
-        timeout(0),
-        retryTimedOut(false),
-        retryBrokenHelper(false),
-        eom('\n') {
-        memset(&stats, 0, sizeof(stats));
-    }
+    static Pointer Make(const char *name);
+
```
yadij, line 72: Please revert the code style to using C++ construction instead of a maker-factory style.

```suggestion
explicit helper(const char *name) : id_name(name) {}
```

src/helper.h
```diff
@@ -140,23 +134,25 @@ class helper
...
 public:
+    using Pointer = RefCount<statefulhelper>;
     typedef std::unordered_map<Helper::ReservationId, helper_stateful_server *> Reservations;
 
-    inline statefulhelper(const char *name) : helper(name) {}
     inline ~statefulhelper() {}
 
 public:
+    static Pointer Make(const char *name);
+
```
yadij, line 145: Please revert the code style to using C++ construction instead of a maker-factory style.

```suggestion
explicit statefulhelper(const char *name) : helper(name) {}
```

yadij (changes requested):

eduard-bagdasaryan (commented):
Partially done at 14593dc.

src/helper/forward.h
rousskov: IMO, the primary reason we should not move these aliases is because such a move would create a "Foo inside Foo" problem: Helper::HelperPointer does not look good and Helper::Pointer is just wrong. The problem here is in the name of the `Helper` namespace or the name of the `::helper` class (that this PR does not and should not change).

I hope we will rename the `::helper` class eventually. If we can agree on that future class name _now,_ then we can move these new pointers inside the Helper namespace using that new name in _this_ PR (and then rename the classes in another dedicated PR).

I can suggest Helper::Client (for the current `::helper` class) and Helper::Session (for the current `helper_server`). That naming scheme matches what the corresponding classes represent well. Squid already uses similar Client terminology for HTTP and ICAP clients.

That (future) naming scheme would result in _this_ PR using Helper::ClientPointer and Helper::StatefulClientPointer. Another dedicated PR will then rename the other/existing classes and move them into the Helper namespace (and into the src/helper/ directory?). We just need to merge a few existing bug fixes (like this PR) so it should not take long.

@yadij, would you support this new Client/Session naming scheme for helper-related classes?

rousskov:

rousskov:
rebuild

rousskov:
> @yadij, please note that this PR has been waiting for your review for more than 20 days. Under the "Timeouts for stalled PRs" experiment, it may be force-merged in 10 days.

Those 10 days also passed. Under the "Timeouts for stalled PRs" [experiment](http://lists.squid-cache.org/private/squid-board/2023-June/001806.html), I am manually dismissing the blocking review and will clear this PR for merging. If there are any objections, please voice them _now_. CC: @yadij, @kinkie.

yadij:
Hmm, I did not get any notifications about this.

src/helper.h
yadij: > > Amos: Please revert the code style to using C++ construction instead of a maker-factory style.
> 
> Why?! PR code (returning a smart pointer) is safer than raw `new` (returning a pointer that can easily leak).

All callers are immediately and unconditionally de-referencing the pointer - regardless of whether it is raw or smart. That invalidates the claim of their being any difference in safety.

> The new API also directs new helper users to proper pointer storage. As an added bonus, this `X::Make()` upgrade significantly reduces a chance that we have missed a raw X allocation somewhere in this PR or will miss it somewhere in backporting.


> Overall, it is an improvement that should not be reverted IMO.

IMO simplicity is better API design. The result of `new` should be stored in a `Pointer` - that is a constant for all types in Squid with `Pointer` defined. This is gratuitously forcing developers to work with an unusual (for C++) allocation mechanism.

> 
> > Eduard: the factory style would involve some dynamic type selection
> 
> Let's ignore unimportant terminology differences. The essence of this change request is clear -- replace `X::Make()` with `new X()`. I see no justification for going back to raw pointers, but perhaps I am missing something. In either case, we can save time by _not_ arguing about the definition of "maker-factory style".

Nod. This is not a request to "go back to raw pointers" it is a request to remove a layer of abstraction between the construction and Pointer storage.

src/helper/forward.h
yadij: I am okay with that naming. Yes.

yadij (changes requested):
Maker and Pointer definitions still need to be resolved.

src/helper/forward.h
eduard-bagdasaryan: Done (72c531e).

yadij (approved):

PR #1392, opened 2023-06-21
Replaced ACLStrategised, enabling other ACL improvements

rousskov:
The ACLStrategised class stands in the way of several ACL bug fixes and
improvements because its design forces us to place ACL-like match()
methods inside non-ACL classes, creating two parallel matching
hierarchies: one rooted in the ACL class and one rooted in the Strategy
template. The two APIs have similar methods, but Strategy-based objects
lie outside the ACL hierarchy and cannot be treated as ACL objects.

ACLStrategised pairs an ACL-like matching algorithm (a Strategy-derived
class) with an ACLData-derived class. The need to combine the two is
genuine, but the same combination is best supported without creating a
parallel hierarchy of Strategy classes. The new ACL-derived
ParameterizedNode base class accomplishes that, addressing the old
ACLStrategised design XXX.

Strategy-derived classes were not pooled at all! With these changes, all
formerly ACLStrategised classes get individual memory pools, typically
one per acltype, with proper acltype-based naming in mgr:mem reports. No
other functionality changes intended.


src/acl/ParameterizedNode.h
```diff
@@ -0,0 +1,56 @@
...
+protected:
+    /* ACL API */
+    void prepareForUse() override { parameters->prepareForUse(); }
+    void parse() override { parameters->parse(); }
+    SBufList dump() const override { return parameters->dump(); }
+    bool empty() const override { return parameters->empty(); }
+    const Acl::Options &lineOptions() override { return parameters->lineOptions(); }
+
+    /// items this ACL is configured to match; never nil after construction
+    std::unique_ptr<Parameters> parameters;
```
rousskov, line 42: What name do you prefer for this above std::unique_ptr<Parameters> class member (called `parameters` in this PR code), keeping in mind that this class member lives inside a ParameterizedNode class that lives inside an Acl namespace:

* `parameters` -- matches the meaning of ACLData objects it stores; we do not have consistent terminology for this, but those objects are known as ACL parameters (among other, more problematic names);

* `data` -- matches the type name of ACLData objects it stores and minimizes PR code changes, but tells nothing to the code reader (nearly everything can be called "data");

* `options` -- please do not propose "options" because that word is already used for ACL-global flags/options and ACL line parameter flags/options (see ACL::options() and ACL::lineOptions());

* `arguments` -- please do not propose "arguments" because that word is already used to describe "ACL parameter values", "ACL-global options", and "ACL line options" collectively/together (i.e. anything that follows the `acl aclname acltype` directive prefix);

* any better options?

src/acl/SourceAsn.h
```diff
@@ -9,17 +9,22 @@
...
+#include "ip/forward.h"
 
-class ACLChecklist;
-
-class ACLSourceASNStrategy : public ACLStrategy<Ip::Address>
+namespace Acl
 {
 
+/// a "src_as" ACL
+class SourceAsnCheck: public ParameterizedNode< ACLData<Ip::Address> >
```
rousskov, line 20: This PR has to modify this and similar ACLFooStrategy class names because they are no longer Strategy-derived classes. We are using this renaming as an excuse to also move these classes into the Acl namespace. We should still use a common class naming pattern, of course. This PR uses a Check suffix, so ACLFooStrategy becomes Acl::FooCheck. If you have better naming pattern ideas, please discuss!

rousskov (commented):
This _draft_ PR currently converts only a few representative ACLStrategised cases. I am posting this PR early so that you can request conversion changes _before_ we propagate the same conversion approach to the remaining ~34 classes. I also have two specific naming questions (detailed in dedicated change requests).

src/acl/ParameterizedNode.h
```diff
@@ -0,0 +1,56 @@
...
+
+protected:
+    /* ACL API */
+    void prepareForUse() override { parameters->prepareForUse(); }
+    void parse() override { parameters->parse(); }
+    SBufList dump() const override { return parameters->dump(); }
+    bool empty() const override { return parameters->empty(); }
+    const Acl::Options &lineOptions() override { return parameters->lineOptions(); }
+
+    /// items this ACL is configured to match; never nil after construction
```
eduard-bagdasaryan, line 41: However, as you noted above, it is 'never nil' after a derived class construction (not this class construction). Also, can we  somehow guarantee (during compile time) that this derived class initialization occurred?

rousskov:
> I am posting this [draft] PR early so that you can request conversion changes _before_ we propagate the same conversion approach to the remaining ~34 classes. I also have two specific naming questions (detailed in dedicated change requests).

FYI: We will start applying similar changes to those remaining ~34 ACLStrategised cases in a couple of days, after this draft PR age crosses the 10 days boundary. If you intended to provide feedback on the approach or answer those two naming questions, please do so soon, before each correction would have to be propagated to dozens of classes!

src/acl/ParameterizedNode.h
yadij: The wording does "smell" a bit to me, but cannot point at specifically why right now.

src/acl/ParameterizedNode.h
```diff
@@ -0,0 +1,58 @@
...
+{
+
+/// An ACL that manages configured parameters using a given Parameters class P.
+/// That Parameters class must support an ACLData<> or equivalent API.
+template <class P>
+class ParameterizedNode: public ACL
+{
+public:
+    using Parameters = P;
+
```
yadij, line 28: Please just use the template definition to name the type instead of a separate sub-type.

```suggestion
/// An ACL that manages configured parameters using a given Parameters class.
/// That Parameters class must support an ACLData<> or equivalent API.
template <class Parameters>
class ParameterizedNode: public ACL
{
public:

```

yadij (commented):
Despite the descriptions provided so far I am finding it very hard so far to tell what this template is actually supposed to be doing. It looks like you are just shuffling things around, but the renaming is obfuscating which parts are the strategy vs the config data.

Also, what is the performance difference between these two techniques?

src/acl/ParameterizedNode.h
```diff
@@ -0,0 +1,58 @@
...
+#include <memory>
+
+namespace Acl
+{
+
+/// An ACL that manages configured parameters using a given Parameters class P.
+/// That Parameters class must support an ACLData<> or equivalent API.
+template <class P>
+class ParameterizedNode: public ACL
```
rousskov, line 24: > I am finding it very hard so far to tell what this template is actually supposed to be doing. 

Sorry to hear that. FWIW, I believe the current "ACL that manages configured parameters using a given Parameters class" definition is a good one. I cannot improve it without knowing what causes the difficulty you are describing.

> It looks like you are just shuffling things around, but the renaming is obfuscating which parts are the strategy vs the config data.

* The "config data" is ParameterizedNode::data. It used to be ACLStrategized::data. No significant changes here.
* The "transaction data extraction and matching strategy" is ParameterizedNode::match(), which is our regular ACL::match(). It used to be ACLStrategized::matcher->match() which is _not_ a part of the ACL class/hierarchy, causing the problems described in the PR description and addressed in this PR.


> Also, what is the performance difference between these two techniques?

The new types remove one forwarding hop (from ACLStrategised to ACLStrategised::matcher), so they ought to be a tiny bit faster, but this PR is not a performance optimization. The focus here is on solving serious architectural problems detailed in the PR description. We are lucky that these solutions do not hurt performance (AFAICT), but they would be necessary even if they did.

src/acl/ParameterizedNode.h
rousskov: > Please just use the template definition to name the type instead of a separate sub-type.

Why? The current approach is better because it gives derived classes and external users a named type they can use. If we remove that alias as suggested, then derived classes would have to add duplicated aliases of their own (if they can guess the underlying type correctly), while some external users would have no good way of guessing the type at all.

For example, imagine writing a template function `foo(T &acl)` that accepts any ParameterizedNode acl object. How would that function get to the parameters type if we remove the alias? Today, that function can just use T::Parameters.

FWIW, this aliasing of template parameter types is standard practice, for similar [reasons](https://stackoverflow.com/questions/63104551/why-alias-a-template-parameter-in-c-standard-containers).

src/acl/ParameterizedNode.h
rousskov: > The wording does "smell" a bit to me, but cannot point at specifically why right now.

We can try to improve the "wording" later -- it is not duplicated, so it would be easy to adjust as needed. Finalizing the field _name_ **now** is important because it is duplicated -- all derived ACLs have to use that data member by name.

src/acl/ParameterizedNode.h
rousskov: > The new types remove one forwarding hop (from ACLStrategised to ACLStrategised::matcher), so they ought to be a tiny bit faster

When I wrote the above, I did not realize that Strategy-derived classes are not pooled at all in the official code! With this PR changes, that problem is gone (with the Strategy template class itself) and all formerly ACLStrategised classes get individual memory pools, typically one per acltype, with proper acltype-based naming in mgr:mem reports. That improves reporting and may help with memory leak triage while pooling might speed up reconfiguration a little.

Still, this PR is not really a performance optimization. The focus here is on solving serious architectural problems detailed in the PR description. We are lucky that these solutions do not hurt performance (AFAICT), but they would be necessary even if they did.

src/AclRegs.cc
```diff
@@ -108,6 +106,85 @@
...
+
+    /// Replaces generic memory allocator label X set by our MEMPROXY_CLASS(X)
+    /// with an admin-friendly label based on the given acltype-like name.
+    /// Normally, our class constructor sets the right allocator label using the
+    /// actlype name, but that algorithm results in unstable and misleading
+    /// labels when the same instantiation of this template class is used for
+    /// _multiple_ acltype names. Calling this method corrects that behavior.
+    /// \prec this method must be called at most once
+    /// \prec if called, this method must be called before the class constructor
+    static void PreferAllocatorLabelPrefix(const char * const suffix)
```
rousskov, line 135: If we remove this and related pool relabeling code in FinalizePoolLabel(), we will essentially go back to the situation in the official code, where all ACLStrategized-based objects are allocated under one admin-unfriendly pool name, regardless of their acltype. The number of the corresponding pool lines in mgr:report will change because the ACLStrategized template was parameterized by ACLData-derived MatchType and this FinalizedParameterizedNode template is parameterized by the ACL-derived Parent class.

This and related pool relabeling code produces acltype-specific pool names. I think this minor code addition is worth it (and it will become even simpler when Mem::AllocatorProxy is finally removed). Here are a few related mgr:mem output line prefixes produced be this PR code:

```
acltype=myportname        112 ...
acltype=port              112 ...
acltype=method            112 ...
acltype=ssl::server_name+ 128 ...
acltype=at_step           112 ...
acltype=srcdomain+        112 ...
acltype=req_header        112 ...
acltype=url_regex         112 ...
```

src/acl/ParameterizedNode.h
rousskov: I decided to preserve the old `data` name to avoid creating otherwise unnecessary and out-of-scope code changes in this PR. We can rename this class member to `parameters` (or whatever we can agree on) later, of course.

src/acl/AdaptationService.h
```diff
@@ -9,15 +9,21 @@
...
+#include "acl/Data.h"
+#include "acl/ParameterizedNode.h"
 
-/// \ingroup ACLAPI
-class ACLAdaptationServiceStrategy : public ACLStrategy<const char *>
+namespace Acl
 {
 
+/// an "adaptation_service" ACL
+class AdaptationServiceCheck: public ParameterizedNode< ACLData<const char *> >
```
rousskov, line 19: I did not rename this and other old source files even though we are changing the primary class names in most of them. We should eventually rename (and perhaps even merge trivial .cc and .h files in most cases), but I believe such renaming deserves a _dedicated_ PR to avoid creating a lot of out-of-scope noise here. This PR did not create this old problem: The old file names were often inconsistent with primary class names (e.g., this file was no called AdaptationServiceStrategy.h or similar). Moreover, some class names were (and remain) rather problematic (e.g. official code has both ACLRequestHeaderStrategy and ACLHTTPReqHeaderStrategy that became RequestHeaderCheck and HttpReqHeaderCheck).

rousskov (commented):
I finished converting ACLStrategised-related classes and marked the PR as ready for review.

kinkie (approved):
Thanks for writing this and for the explanation. Sounds legit, let's ship it

rousskov:
Thank you, Francesco!

@yadij, you are _not_ blocking this PR, so you do not have to review it, but I will keep it open for two more days in case you need more time to come back to it. If I can merge sooner, please let me know :-).

yadij (changes requested):

src/acl/ParameterizedNode.h
rousskov: > I do not think we have any code actually in Squid that needs that ability.

In this PR, there is at least one declaration in another class that already uses that ability:

```C++
FinalizedParameterizedNode(TypeName typeName, Parameters * const params)
```

The only way to avoid the above usage is to convert that simple constructor into a _template_ that accepts any `params` type even though the class itself needs a rather specific type. Such a conversion will make code more complex and compiler errors about mismatching ACLData types even more difficult to decipher (because the mismatch will happen at a lower code level). IMO, this "standard practice" `using` declaration is an acceptable price to pay for avoiding those complications (and it may get more reuse in the future).

src/acl/ServerName.h
```diff
@@ -20,12 +19,15 @@ class ACLServerNameData : public ACLDomainData {
     bool match(const char *) override;
 };
 
-class ACLServerNameStrategy : public ACLStrategy<char const *>
+namespace Acl
 {
 
+/// an "ssl::server_name" or "ssl::server_name_regex" ACL
+class ServerNameCheck: public ParameterizedNode< ACLData<const char *> >
```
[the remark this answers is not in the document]
rousskov: >  What do you think about having a code level namespace matching the ssl:: config namespace? I think it would be a good idea so that we can have different FooCheck classes for the distinct protocol scopes.

Yeah, I was thinking about namespaces for ACLs as well. I suspect that we are using the Acl namespace wrong -- it should not be simply/automatically strapped on top of anything that has to do with ACL-derived code. Instead, it should only be used for code that provides ACL fundamentals to the rest of Squid code base -- an Acl namespace is for the ACL "library" or "service" (in general sense, not the literal "libacls.a" library and such), not the code _using_ that library or service.

If we switch to that approach, many of the ACL-derived classes (including many of those touched in in this PR and others!) should be moved from their Acl namespace into the global namespace (or, in some rare cases, into other namespaces). They should also be moved outside the `src/acl` directory! IMO, those moves should be done in dedicated PR (or PRs), moving all classes (or, for example, moving all protocol Foo-related classes).

For example, this Acl::ServerNameCheck class may become ServerNameCheck (because it checks names using some non-TLS sources of information) or TlsSeverNameCheck (because it starts with or focuses on TLS sources of information). Note that it should not become Security::ServerNameCheck because the same reasoning that applies to the Acl namespace applies to the Security namespace as well.

N.B. I do not think we should focus on matching squid.conf namespaces with C++ namespaces. The two (obviously related) concepts operate in rather different environments, facing very different users and constraints. When there is a good match, great. When there is not, it is not a big deal. In case of `ssl::`, there is not a good match because Squid code uses `ssl` namespace for narrow, legacy reasons.

I did not want to discuss any of that in this PR because the topic is complex, controversial, and orthogonal to this PR changes. If we try to solve the two problems at once in this PR, we will reduce the number of times the same lines of code will be (trivially) modified, but we will muddy the important changes that this PR is focusing on and will significantly delay progress. I am certain that those changes, if any, should be done in a separate PR or PRs, after we develop consensus regarding namespace use. For now, this PR simply converts ACLFooStrategized into Acl::FooCheck and that is a good step forward as far as names are concerned, regardless of future changes, even if some of those future changes move all of the affected classes elsewhere.

----

> Sine this PR is including these renames.

Just to avoid a misunderstanding: This PR renames ACLFooStrategy classes because the corresponding code lines must be modified for in-scope reasons and because the old names are clearly bad, on several levels. I can undo all the ACLFooStrategy renaming, but it will not have a significant effect on the diff noise (and may complicate future class moves away from the Acl namespace).

rousskov (commented):
@yadij, I hope I have addressed all your change requests. Please re-review.

kinkie (approved):
I'm happy for this to go in

src/acl/ServerName.h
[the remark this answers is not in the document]
yadij: I think we are in rough agreement on the direction of ACL cleanup and use of `namespace Acl`. Future cleanup PRs/discussions can clarify further.

PR #1403, opened 2023-07-02
Maintenance: Include inttypes instead of stdint

yadij:
With current C/C++ standard, intypes.h includes stdint.h.

### DIFF
--- a/acinclude/nettle.m4
+++ b/acinclude/nettle.m4
@@ -15,8 +15,8 @@ AC_DEFUN([SQUID_CHECK_NETTLE_BASE64],[
   AH_TEMPLATE(HAVE_NETTLE34_BASE64,[set to 1 if Nettle 3.4 API will link])
   SQUID_STATE_SAVE(squid_nettle_base64_state)
   AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+#   include <cinttypes>
 #   include <cstddef>
-#   include <cstdint>
 #   include <nettle/base64.h>
   ]],[[
     char inData[10]; inData[0] = '\0';
--- a/compat/types.h
+++ b/compat/types.h
@@ -31,7 +31,9 @@
 #if HAVE_STDDEF_H
 #include <stddef.h>
 #endif
-#if HAVE_INTTYPES_H
+#if __cplusplus
+#include <cinttypes>
+#elif HAVE_INTTYPES_H
 #include <inttypes.h>
 #endif
 #if HAVE_SYS_BITYPES_H
@@ -48,15 +50,6 @@
 /******************************************************/
 /* Typedefs for missing entries on a system           */
 /******************************************************/
-
-/*
- * Ensure that standard type limits are defined for use
- */
-#if __cplusplus
-#include <cstdint>
-#elif HAVE_STDINT_H
-#include <stdint.h>
-#endif
 
 /* explicit bit sizes */
 #if !defined(UINT32_MIN)
--- a/src/base/Stopwatch.h
+++ b/src/base/Stopwatch.h
@@ -10,7 +10,6 @@
 #define SQUID_SRC_BASE_STOPWATCH_H
 
 #include <chrono>
-#include <cinttypes>
 
 /// Quickly accumulates related real-time (a.k.a. "physical time" or "wall
 /// clock") periods. Continues to run if the caller is blocked on a system call.
--- a/src/base/Stopwatch.h
+++ b/src/base/Stopwatch.h
@@ -10,7 +10,7 @@
 #define SQUID_SRC_BASE_STOPWATCH_H
 
 #include <chrono>
-#include <cstdint>
+#include <cinttypes>
 
 /// Quickly accumulates related real-time (a.k.a. "physical time" or "wall
 /// clock") periods. Continues to run if the caller is blocked on a system call.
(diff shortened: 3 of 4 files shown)


kinkie (approved):
Land all the things

acinclude/nettle.m4
```diff
@@ -15,8 +15,8 @@ AC_DEFUN([SQUID_CHECK_NETTLE_BASE64],[
   AH_TEMPLATE(HAVE_NETTLE34_BASE64,[set to 1 if Nettle 3.4 API will link])
   SQUID_STATE_SAVE(squid_nettle_base64_state)
   AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+#   include <cinttypes>
 #   include <cstddef>
-#   include <cstdint>
```
rousskov, line 19: > PR description: With current C/C++ standard the intypes.h includes stdint.h

Agreed. Due to the above guarantee, we do not need to include stdint.h if we have to include intypes.h.

> PR title (and this code change): Include inttypes instead of stdint

Why? The two headers are different. If some code only needs (smaller) stdint.h, why should that code include (larger) inttypes.h that it does not need?

Please note that I am not saying that this particular AC_LANG_PROGRAM() or any other specific place changed in this PR does not need inttypes.h. I suspect some of them needed that header, triggering this PR; I have not checked any specific locations for that code needs. It is the "let's always include the larger header" approach (implied by the PR title/description) that raises a red flag, even if some code does need to include the larger header. If all code changes in this PR are required, then let's adjust this PR title/description to remove that implication. For example:

```
Fix ... build by including missing inttypes.h

Also, standard C/C++ intypes.h includes stdint.h, so we do not need to
explicitly include the latter when we need to include the former.
```

compat/types.h
```diff
@@ -49,15 +51,6 @@
...
 /******************************************************/
 
-/*
- * Ensure that standard type limits are defined for use
- */
-#if __cplusplus
-#include <cstdint>
-#elif HAVE_STDINT_H
-#include <stdint.h>
-#endif
```
rousskov, line 59: This removal makes sense to me: We have (most likely) included inttypes.h above so we do not need to include stdint.h here.

rousskov (changes requested):

acinclude/nettle.m4
yadij: This line is being touched ensure that the configure test is checking things work with the same includes that Squid itself uses later for the actual build.

> > PR description: With current C/C++ standard the intypes.h includes stdint.h
> 
> Agreed. Due to the above guarantee, we do not need to include stdint.h if we have to include intypes.h.
> 
> > PR title (and this code change): Include inttypes instead of stdint
> 
> Why? The two headers are different. If some code only needs (smaller) stdint.h, why should that code include (larger) inttypes.h that it does not need?
> 

Why? because these files are globally included the ```cstdint``` and ```stdint.h``` is redundant with C99 and C++11.

> Please note that I am not saying that this particular AC_LANG_PROGRAM() or any other specific place changed in this PR does not need inttypes.h. I suspect some of them needed that header, triggering this PR; I have not checked any specific locations for that code needs.

FTR, the trigger for this PR was your request in other reviews to use the STD definitions for printf ```%PRI...``` codes. That led to reviewing the ```compat/types.h``` content against C++11 guaranteed available symbols.

This is basic code maintenance.

rousskov, line 19: > This line is being touched ensure that the configure test is checking things work with the same includes that Squid itself uses later for the actual build.

We should either rely on "intypes.h includes stdint.h" guarantee or never use that guarantee. If, starting with this PR, we rely on that guarantee, then this change should be undone[^1] because the existing configure test is already "checking things work with the same includes that Squid itself uses later for the actual build". For that "actual build", Squid will explicitly include cinttypes, but we know that Squid will "use" cstdint as well!

[^1]: Assuming this code is using the correct header in the first place, of course. If this code actually needs cinttypes, then this change is not wrong, just out of this PR scope. My change request is based on the quoted rationale. I have not checked whether this code uses any cinttypes APIs that are not provided by cstdint.

IMO, we should not tell folks to include B.h instead of A.h when the standard tells them that the API they use is defined in A.h. Two changes in this PR, including this one, imply the opposite. I am against such changes and consider their stated rationale contradicting the guarantee that this PR is based on.

src/base/Stopwatch.h
```diff
@@ -10,7 +10,7 @@
 #define SQUID_SRC_BASE_STOPWATCH_H
 
 #include <chrono>
-#include <cstdint>
+#include <cinttypes>
```
rousskov, line 13: > Why? because these files are globally included

If these files are globally included (and we are relying on them to be globally included when making decisions about our code), should not this include be _removed_ instead?

```suggestion
```

acinclude/nettle.m4
rousskov: I am resolving this change request and using the information you have provided in more specific change requests.

rousskov (changes requested):
I have updated my review based on your feedback, thank you.

P.S. I also polished PR description wording a bit. Please feel free to undo those changes if you do not like them or adjust further as needed.

kinkie:
I think we should close this PR without merging, and take a more minimalistic approach, replacing <inttypes.h> with <cinttypes> which offers a stronger guararntee.
We have it in configure.ac, acinclude/init.m4, compat/types.h

kinkie (changes requested):
I don't think it makes sense to do this; I propose we close the PR as indicated in my ast comment

PR #1404, opened 2023-07-02
Remove unnecessary net compatibility workarounds

kinkie:
All modern network stacks provide inet_ntop(), inet_pton(),
getaddrinfo(), and getnameinfo() APIs.

Also removed related Solaris 9 hacks, effectively discontinuing support
for that platform. Solaris 9 is no longer supported by Oracle.


configure.ac
```diff
@@ -2718,7 +2718,7 @@ AC_REPLACE_FUNCS(\
 	tempnam \
 )
 
-AC_CHECK_DECLS([getaddrinfo,getnameinfo,inet_ntop,inet_pton,InetNtopA,InetPtonA],,,[
+AC_CHECK_DECLS([InetNtopA,InetPtonA],,,[
```
yadij, line 2721: I have been reviewing the Sockets API fro MinGW recently and it seems modern Windows API (Vista+) is included in the set which provides the POSIX syscalls. I think we can drop these ```InetPtonA``` and ```InetNtopA``` as well.

compat/Makefile.am
```diff
@@ -29,14 +29,6 @@ libcompatsquid_la_SOURCES = \
...
 	fdsetsize.h \
-	getaddrinfo.cc \
-	getaddrinfo.h \
-	getnameinfo.cc \
-	getnameinfo.h \
-	inet_ntop.cc \
-	inet_ntop.h \
-	inet_pton.cc \
-	inet_pton.h \
 	initgroups.h \
```
rousskov, line 32: I have adjusted PR title/description. I started that edit to fix M-failed-description, but then discovered that the PR removes more than "functions" and edited some more. Please review and adjust further as needed.

> Adjusted PR description: All modern network stacks provide inet_ntop(), inet_pton(),
getaddrinfo(), and getnameinfo() APIs.

The code remaining in compat/os/opensolaris_10_netdb.h still defines `struct addrinfo`, `EAI_OVERFLOW`, and various other getaddrinfo() API components (see commit 3729717). PR description can be interpreted as implying that Squid only supports "modern network stacks". Based on that assumption, we should remove all those Solaris replacements, possibly undoing the entire hack introduced in commit 3729717 and, hence, effectively dropping Solaris 9 support (if it is still considered supported by anyone).

I do not insist on any changes.

rousskov (commented):
Thank you for this cleanup. I hope your assumption regarding modern TCP/IP stacks is correct and relevant to all Squid platforms we want to support.

kinkie:
> Thank you for this cleanup. I hope your assumption regarding modern TCP/IP stacks is correct and relevant to all Squid platforms we want to support.

I checked the manual pages of the environments we support, I can try a build on freebsd and openbsd before merge; mingw (and therefore windows) will be double checked as part of the ongoing efforts there

compat/Makefile.am
kinkie: Solaris 9 is not supported by Oracle anymore; I'll remove

configure.ac
kinkie: removed

compat/Makefile.am
kinkie: Removed

rousskov:
> I can try a build on freebsd and openbsd before merge

Your call.

compat/Makefile.am
rousskov: Glad those hacks can be removed now. I added the corresponding note to the PR description.

kinkie:
> > I can try a build on freebsd and openbsd before merge
> 
> Your call.

https://build.squid-cache.org/job/anybranch-openbsd-matrix/17/COMPILER=clang,label=openbsd/console

kinkie:
> > > I can try a build on freebsd and openbsd before merge
> > 
> > 
> > Your call.
> 
> https://build.squid-cache.org/job/anybranch-openbsd-matrix/17/COMPILER=clang,label=openbsd/console

Test failed but due to an issue with the testing framework, the code built fine.

rousskov (approved):
Thank you for this cleanup!

Disclaimer: I cannot verify the claims in the PR description.

squidadm:
queued for backport to v6

PR #1406, opened 2023-07-03
Bug 5283: squidclient mgr:info returns 403

yadij:
Since commit a4e35bd8ffe51cda83e63faab79bda5a838c7e1a
squidclient has been sending an http:// scheme URL instead of
cache_object:// Presence of cache_object:// scheme alone used
to be enough to determine the use of an internal request, so
the checks of hostname did not need to care about 'localhost'
being sent instead of the actual assigned hostname.

Now that the default squidclient URL is http://localhost:3128/
this additional possible hostname matters.


src/internal.cc
```diff
@@ -175,6 +175,9 @@ internalHostnameIs(const char *arg)
         if (0 == strcmp(arg, w->key))
             return 1;
 
+    if (0 == strcmp(arg, "localhost"))
+        return 1;
+
     return 0;
```
rousskov, line 181: > PR description: Presence of cache_object:// scheme alone used to be enough to determine the use of an internal request, so the checks of hostname did not need to care about 'localhost' being sent instead of the actual assigned hostname. Now that the default squidclient URL is http://localhost:3128/ this additional possible hostname matters.

Since Squid supported non-cache_object internal requests before that scheme was removed, I do not think we can claim that hostname checks were not needed before the removal and are needed after the removal. AFAICT, those checks were needed before and are still needed now to identify internal URLs matching the current Squid instance.

I need more time to suggest a specific fix for the PR description.

src/internal.cc
```diff
@@ -175,6 +175,9 @@ internalHostnameIs(const char *arg)
         if (0 == strcmp(arg, w->key))
             return 1;
 
+    if (0 == strcmp(arg, "localhost"))
+        return 1;
```
rousskov, line 179: The implied assumption that any Squid instance that receives a localhost URL listens on localhost is wrong AFAICT: An admin can configure Squid that does not listen on any IP addresses associated with the `localhost` name and send that Squid a request with a localhost URL.

How about checking for localhost here only when the relevant http_port does not specify an IP address or specifies an IP address that `localhost` resolves to? AFAICT, such a conditional check would take care of most common use cases without mishandling 

1. correct cache manager requests sent to a localhost Squid instance _through_ a non-localhost Squid instance
2. bad cache manager requests sent to a non-localhost Squid instance but targeting a localhost Squid instance

rousskov (dismissed):

src/internal.cc
rousskov: > Alex: How about checking for localhost here only when the relevant http_port does not specify an IP address or specifies an IP address that localhost resolves to?

Moreover, the above logic can and should be extended to apply to _any_ requested domain name or IP that reach this function -- there is nothing specific about localhost here: If the requested target (including "localhost") matches[^1] the listening port address, return a truthy value.

[^1]: matches after the name resolution to get the targeted IP addresses (if such a resolution is necessary), of course

> Amos at https://bugs.squid-cache.org/show_bug.cgi?id=5283#c6:  Better checks will need to be implemented long-term.

I am not against merging this code change now if you think that is the best way forward, but we should add the corresponding comment. For example, "XXX: This check mishandles rare cases where this localhost-targeting request was not received on a localhost-associated port."

src/internal.cc
rousskov: > I need more time to suggest a specific fix for the PR description.

Done: I updated PR description text accordingly. Please adjust further as needed.

I consider this change request resolved.

src/internal.cc
yadij: Added the XXX in [07b4652](https://github.com/squid-cache/squid/pull/1406/commits/07b4652c0c715ed2ed4bfd24900cffd2c82d1013) since this should be fixed in v6.1 and the full fix will take longer and more testing.

src/internal.cc
```diff
@@ -163,6 +163,8 @@ internalHostname(void)
     return host;
 }
 
+// XXX: This check mishandles cases where the listening port hostname
+//      does not match the request-target hostname.
```
rousskov, line 167: This XXX documents the old/existing problem unrelated to this PR workaround and, arguably, to this instance-wide internalHostnameIs() function. I am not against this XXX, but I recommend moving it to the internalHostnameIs() caller.

There are at least two (very different) problems related to the internal request detection code:

1. We do not have enough port-specific checks. We can document this old problem in this PR as discussed above.
2. The new check added by this PR is flawed. We should document _that_ problem. The XXX above is not enough. I have proposed the corresponding XXX text earlier, but I do not insist on any specific wording. That XXX should go where the new `if` statement is.

yadij: The flaw itself is not new at all. Just more obvious with the new check being added. This whole function does not verify that any of the names actually match the listening port's configured  host:port or defaultsite= settings.
IMO, doing a complex fix just for "localhost" is nowhere near enough to resolve the issue, so I placed it on the function.

yadij:
Converting this to draft because it missed the 6.1 release, making it no longer urgent. We can wait for the full refactor requested.

rousskov:
> We can wait for the full refactor requested.

Your call. I have not requested refactoring, but I agree that Squid is missing a lot of logic in this "Who am I (to this request)?" area.

FWIW, I expect the final solution to still respond with errors to "default" (i.e. localhost-based) cache manager requests in _some_ environments.  I am also concerned that squidclient is using an absolute URL when sending a direct request to what feels to me like an origin server service (rather than a proxy service), inviting unwanted forwarding. I doubt there is consensus about these and related aspects of the final solution, so if you are going to make changes, it may be a good idea to discuss the design principles they will be based on.

src/client_side.cc
```diff
@@ -1638,7 +1638,10 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
 #endif
 
     if (internalCheck(request->url.path())) {
-        if (internalHostnameIs(request->url.host()) && request->url.port() == getMyPort()) {
+        const bool allowLocalhost = (conn->port->s.isLocalhost() || conn->port->s.isAnyAddr());
+        const bool isMyHostname = (allowLocalhost && strcmp(request->url.host(), "localhost") == 0) ||
+                                  internalHostnameIs(request->url.host());
+        if (isMyHostname && request->url.port() == conn->port->s.port()) {
```
rousskov, line 1644: Please move all these checks inside one dedicated function.

FWIW, I recommend starting that method with the universal (and fast) port number check. It will simplify the function code structure because you can simply `return false` after that check fails and `return true` from any confirmed alternative (validated after that port number check). Adding debugging to those `return` statements may be a good idea (and will help self-document that code).

src/client_side.cc
```diff
@@ -1638,7 +1638,10 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
...
             debugs(33, 2, "internal URL found: " << request->url.getScheme() << "://" << request->url.authority(true));
```
rousskov, line 1645: > We can wait for the full refactor requested.

Does this updated PR represent that "full refactor"? Or should we still treat this PR as a temporary hack to relieve bug 5283 pressures?

src/client_side.cc
```diff
@@ -1638,7 +1638,10 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
 #endif
 
     if (internalCheck(request->url.path())) {
-        if (internalHostnameIs(request->url.host()) && request->url.port() == getMyPort()) {
+        const bool allowLocalhost = (conn->port->s.isLocalhost() || conn->port->s.isAnyAddr());
+        const bool isMyHostname = (allowLocalhost && strcmp(request->url.host(), "localhost") == 0) ||
```
rousskov, line 1642: If we can rely on url.host() being lower-case, please disclose that invariant in the Uri::host() description. Otherwise, please use strcasecmp().

src/client_side.cc
```diff
@@ -1638,7 +1638,10 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
 #endif
 
     if (internalCheck(request->url.path())) {
-        if (internalHostnameIs(request->url.host()) && request->url.port() == getMyPort()) {
+        const bool allowLocalhost = (conn->port->s.isLocalhost() || conn->port->s.isAnyAddr());
```
rousskov, line 1641: Please avoid "allow" or "deny" in variable names that do not contain the final allow/deny decision because such naming makes the code more difficult to understand.

```suggestion
        const auto receivedAtLocalhost = (conn->port->s.isLocalhost() || conn->port->s.isAnyAddr());
```

rousskov (changes requested):

src/client_side.cc
yadij: The parts of the refactor which were colliding with this have been merged as 101694e4332f35dc3b5e4729839fcd2daad858a5 allowing this PR to progress.

src/client_side.cc
yadij: Done in [f1015fa](https://github.com/squid-cache/squid/pull/1406/commits/f1015fa9f7ccf7c9d8122f534ba5bd40b03b5a3f)

src/client_side.cc
yadij: Looking into this, parsed URLs enforce lower case hostnames. But the `host(...)` setter method does not  - so `append_domain` at least allowing any string input means its case cannot be trusted.
Changing to `srcasecmp()` as requested.

src/internal.cc
```diff
@@ -163,18 +163,31 @@ internalHostname(void)
...
+
+    // accept hostname_aliases directive
+    for (const auto *w = Config.hostnameAliases; w; w = w->next) {
+        if (strcasecmp(url.host(), w->key) == 0)
+            return true;
+    }
 
-    return 0;
+    return false; // no match
 }
```
rousskov, line 193: To undo an unwanted or out-of-scope "EOF formatting" change immediately below this change request:

```suggestion
}

```

src/client_side.cc
```diff
@@ -1636,7 +1636,7 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
 #endif
 
     if (internalCheck(request->url.path())) {
-        if (internalHostnameIs(request->url.host()) && request->url.port() == getMyPort()) {
+        if (internalWithMyHostname(request->url, *(conn->port))) {
```
rousskov, line 1639: FWIW, I see no need to add those parenthesis to dereference a pointer: If there is a way for the alternative `(*conn)->port()` expression to compile, then we have a much bigger problem that cannot be solved by sprinkling parenthesis around the code! We can and should keep it simple:

```suggestion
        if (internalWithMyHostname(request->url, *conn->port)) {
```

rousskov (changes requested):
Excellent progress, thank you! The PR description got a bit stale, but we can update it once the primary code changes settle.

The two primary change requests are marked with "(this is a primary change request)". Most of my other change requests are simple polishing improvements. I am happy to implement all of them myself if you prefer.

src/client_side.cc
yadij: These brackets are for developer readability, to clarity that the `port` pointer is being dereferenced not the `conn` pointer.

src/internal.cc
```diff
@@ -163,18 +163,31 @@ internalHostname(void)
...
+    // different ports is a fast NO
+    if (url.port() != listener.s.port())
+        return false;
+
+    // "localhost" in URL can match wildcard IP, or explicit localhost-IP listeners
+    const bool receivedAtLocalhost = (listener.s.isLocalhost() || listener.s.isAnyAddr());
+    if (receivedAtLocalhost && strcasecmp(url.host(), "localhost") == 0)
+        return true;
+
+    // accel ports with defaultsite= can also match
```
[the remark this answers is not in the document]
yadij: NAK, that text can be inferred from the code (eg a "foo is foo" comment). This is attempting to document that on accel ports defaultsite **can/maybe** match the the URL domain. Avoiding any implication of requirement.

src/internal.cc
```diff
@@ -163,18 +163,31 @@ internalHostname(void)
...
+    if (listener.defaultsite && strcasecmp(url.host(), listener.defaultsite) == 0)
+        return true;
+
+    // accept visible_hostname, unique_hostname, or machines hostname()
```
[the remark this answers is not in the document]
yadij: This is supposed to be documenting what detail/setting(s) the test case is looking for, not how the check is implemented.

src/internal.h
```diff
@@ -27,7 +28,8 @@ bool internalStaticCheck(const SBuf &urlPath);
 char *internalLocalUri(const char *dir, const SBuf &name);
 char *internalRemoteUri(bool, const char *, unsigned short, const char *, const SBuf &);
 const char *internalHostname(void);
-int internalHostnameIs(const char *);
+/// whether the given URI matches this Squid hostname and listening IP:port
+bool internalWithMyHostname(const AnyP::Uri &, const AnyP::PortCfg &);
```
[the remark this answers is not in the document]
yadij: Edited your change suggestion to remove a typo `((`.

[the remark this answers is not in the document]
yadij: Retaining the `internalFoo()` naming scheme since this is part of the `internal.h` API. If/When it is moved elsewhere it can be renamed appropriately.

src/internal.cc
```diff
@@ -163,18 +163,31 @@ internalHostname(void)
...
-    if (0 == strcmp(arg, internalHostname()))
-        return 1;
-
-    for (w = Config.hostnameAliases; w; w = w->next)
-        if (0 == strcmp(arg, w->key))
-            return 1;
+    // different ports is a fast NO
+    if (url.port() != listener.s.port())
+        return false;
+
```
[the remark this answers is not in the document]
yadij: Done in [bfc32f7](https://github.com/squid-cache/squid/pull/1406/commits/bfc32f7d6daa5eddb93c56c76517d12f5d7af949)

src/internal.cc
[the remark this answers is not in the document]
yadij: re-worded in [bfc32f7](https://github.com/squid-cache/squid/pull/1406/commits/bfc32f7d6daa5eddb93c56c76517d12f5d7af949) to make a more consistent case documentation

src/internal.cc
```diff
@@ -163,18 +163,31 @@ internalHostname(void)
     return host;
 }
 
-int
-internalHostnameIs(const char *arg)
+bool
+internalWithMyHostname(const AnyP::Uri &url, const AnyP::PortCfg &listener)
```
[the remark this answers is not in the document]
yadij: Done in [bfc32f7](https://github.com/squid-cache/squid/pull/1406/commits/bfc32f7d6daa5eddb93c56c76517d12f5d7af949)

src/tests/stub_internal.cc
```diff
@@ -13,4 +13,4 @@
 #include "tests/STUB.h"
 
 char *internalLocalUri(const char *, const SBuf &) STUB_RETVAL(nullptr)
-
+bool internalWithMyHostname(const AnyP::Uri &, const AnyP::PortCfg &) STUB_RETVAL(false)
```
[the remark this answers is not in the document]
yadij: Done in [bfc32f7](https://github.com/squid-cache/squid/pull/1406/commits/bfc32f7d6daa5eddb93c56c76517d12f5d7af949)

src/internal.cc
```diff
@@ -162,18 +162,36 @@ internalHostname(void)
...
-    for (w = Config.hostnameAliases; w; w = w->next)
-        if (0 == strcmp(arg, w->key))
-            return 1;
+    // different port numbers is a fast NO
+    if (url.port() != listeningPort.s.port())
+        return false;
+
+    // accept raw-IP URL matching a port configured with explicit IP:port settings
+    if (url.hostIsNumeric() && !listeningPort.s.isAnyAddr() && url.hostIP() == listeningPort.s)
+        return true;
```
rousskov, line 174: Should not we reject specific IP mismatches?

```suggestion
    if (url.hostIsNumeric() && !listeningPort.s.isAnyAddr())
        return url.hostIP() == listeningPort.s;
```

If we want to preserve the official code logic with regard to an internalHostname() and hostname_aliases match (for hostIsNumeric() cases), then we would _also_ need to move this check further down, after those checks. The answer to that question may depend on whether we want to support IP addresses as internalHostnames and hostname_aliases. 

Another potential related concern here is whether it is possible to specify a listening port IP mask like 127.0.0.255. I cannot check that quickly right now, but if it is possible, then the equality comparison for IPs is wrong.

rousskov (changes requested):
I also updated PR description but it needs more work. For example, it would be nice to describe a specific use case where we still reject same-host mgr:info requests that were allowed before cache_object removal.

src/internal.cc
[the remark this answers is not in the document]
yadij: You have this backwards. This comment as-is documents what this caller is trying to accept.

Calling `internalHostname()` is a de-duplication of the code we would otherwise have to write here to test these three configuration cases against the URL.

src/internal.cc
yadij: Thanks. Yes we should.

PR #1407, opened 2023-07-03
Maintenance: Remove unused xstrtoul() from the global scope

kinkie:
xstrtoul() is only used by xstrtoui() in the same module, no need to
export it to the global namespace

### DIFF
--- a/compat/xstrto.cc
+++ b/compat/xstrto.cc
@@ -57,7 +57,7 @@
 
 #include <cerrno>
 
-bool
+static bool
 xstrtoul(const char *s, char **end, unsigned long *value,
          unsigned long min, unsigned long max)
 {
--- a/compat/xstrto.h
+++ b/compat/xstrto.h
@@ -9,11 +9,11 @@
 #ifndef _SQUID_XSTRTO_H
 #define _SQUID_XSTRTO_H
 
-// these functions are not used by the remaining Squid C code.
+// this function is not used by the remaining Squid C code
 #if defined(__cplusplus)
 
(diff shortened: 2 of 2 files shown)


rousskov (approved):
> PR description: xstrtoul is only used by xstrstoui in the same module, no need to export it to the global namespace

I am not against these (technically correct) changes. I doubt they are worth the effort, especially if somebody discovers they need xstrtoul() and re-adds that function to the header later. IMHO, the energy spent on this polishing would be better spent on replacing both xstrtoui() and xstrtoul() with modern code, but I understand that requires a lot more work, and I am not asking for it.

Please note that I added a "Maintenance:" prefix to the PR title. Under the Easier Done experiment: I fixed PR title/description typos and fixed PR description formatting. I also fixed stale `__cplusplus` explanation and updated xstrtoui() Doxygen comment to better match Squid documentation style (see commit 34b7d7d). Please check.

kinkie:
This header file might be tickling in the wrong way clang++ on testPreCompiler on ubuntu-focal on arm64 . It was a 2 minutes effort, it won't hurt anyway. Thanks!

rousskov:
> This header file might be tickling in the wrong way clang++ on testPreCompiler on ubuntu-focal on arm64.

Disclosing those details could have been useful IMHO.

> It was a 2 minutes effort

The _total_ cost exceeds 20 minutes. Not a big deal, of course, but it is not a negligible amount of time either.

squidadm:
queued for backport to v6

PR #1408, opened 2023-07-04
Do not leak memory when handling cache manager requests

rousskov:
Also adjusted Cache-Control APIs to prevent similar bugs. These changes
also speed up processing a bit and simplify most of the affected code.
The now-gone "just remove the old CC" putCc() misfeature was unused.

The leak was introduced by commit 92a5adb: PutCommonResponseHeaders()
incorrectly assumed that putCc(pointerToX) takes ownership of X.

Detected by Coverity. CID 1534779: Resource leak (RESOURCE_LEAK).

### DIFF
--- a/src/HttpHeader.cc
+++ b/src/HttpHeader.cc
@@ -1040,15 +1040,14 @@ HttpHeader::putAuth(const char *auth_scheme, const char *realm)
 }
 
 void
-HttpHeader::putCc(const HttpHdrCc * cc)
+HttpHeader::putCc(const HttpHdrCc &cc)
 {
-    assert(cc);
     /* remove old directives if any */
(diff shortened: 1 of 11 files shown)


kinkie (approved):
Seems it could be simplified further, but I want to avoid scope creep. Shipit

yadij (approved):

rousskov:
> The leak was introduced by commit 92a5adb73b7a50d725fd122c6d27cf8ce1c109f3

@kinkie, @yadij, please note that v6 is suffering from this memory leak (since v6 commit 0298bbb). FWIW, the fix in master/v7 (i.e. commit 182faab merged in this PR) applies cleanly to v6.8 and appears to work in my quick-and-dirty tests.

squidadm:
queued for backport to v6

PR #1409, opened 2023-07-05
Support read-only ClpMap iteration by ClpMap users

kinkie:
This change makes it possible for anticipated future ClpMap users (e.g.,
stat_ipcache_get()) to report their cache contents. No changes to
ClpMap::Entry and related old types except making them public.

The alternative solution -- a visitor design pattern -- was rejected
because `for` loops are a bit easier to read than for_each() loops.

No Squid functionality changes expected.


src/tests/testClpMap.h
```diff
@@ -0,0 +1,43 @@
...
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID_BASE_TESTCLPMAP_H
+#define SQUID_BASE_TESTCLPMAP_H
+
+#include "compat/cppunit.h"
+#include "base/ClpMap.h"
+
+class testClpMap: public CPPUNIT_NS::TestFixture
```
rousskov, line 15: I will make a few PR corrections and adjustments later, but you do not need to wait for me to remove this header file and related unnecessary/stale changes.

rousskov (changes requested):
Thank you for adding iterators to ClpMap.

What iteration contexts do you plan to use these iterators for in the foreseeable future? If possible, please provide (in a PR comment) a couple of references to specific existing code (that should be converted in future PRs), to make it easier to check whether the ClpMap parts exposed by this PR are adequate/sufficient for those known use cases.

kinkie:
The use cases that got me into developing this are ipcache and fqdncache. ClpMap would work well for them, but we need to display the contents of the cache in cachemgr.

src/Makefile.am
```diff
@@ -815,10 +815,11 @@ check_PROGRAMS += tests/testClpMap
 tests_testClpMap_SOURCES = \
 	tests/testClpMap.cc
 nodist_tests_testClpMap_SOURCES = \
+	base/ClpMap.h \
```
rousskov, line 818: Why do we want to list this particular header file in `nodist_tests_testClpMap_SOURCES`? It was not listed before this PR, and other header files (included by testClpMap.cc) remain unlisted in this PR, so this looks like an out of scope (and possibly unnecessary) change, but perhaps I am missing something.

kinkie: I think this failed some build until I did, but now after retesting it seems to work. Removing.

src/base/ClpMap.h
```diff
@@ -114,19 +108,26 @@ class ClpMap
...
-    /// It may return expired entries. Order is not guaranteed
-    ConstEntriesIterator cbegin() { return entries_.cbegin(); }
-    ConstEntriesIterator cend() { return entries_.cend(); }
+    /// Read-only traversal of all cached entries in LRU order, least recently
+    /// used entry first. Stored expired entries (if any) are included. Any map
+    /// modification may invalidate these iterators and their derivatives.
+    ConstEntriesIterator cbegin() const { return entries_.cbegin(); }
+    ConstEntriesIterator cend() const { return entries_.cend(); }
+    ConstEntriesIterator begin() const { return cbegin(); }
+    ConstEntriesIterator end() const { return cend(); }
```
kinkie, line 117: While I see the handiness of doing this, won't it break the minimum surprise principle?

kinkie (commented):
Thanks for the improvements, @rousskov !

src/base/ClpMap.h
rousskov: The only surprise I can imagine here is that somebody (incorrectly) tries to modify the map using begin(), and the compiler (correctly) stops them. I think that price is worth paying for supporting range-based for loops because those loops are safer and easier to write/read. I wish C++ did not create this trade-off for us, but, unfortunately, we have to pick the lesser evil (AFAICT).

I expect future IP cache dumping code to use range-based for loops. I am not sure anybody would need to use explicit cbegin()/cend() in the foreseeable future, but I am not against adding those trivial methods, for completeness sake.

rousskov: BTW, if you want us to be explicit about it, we can do this:

```suggestion
    /// range-based `for` loop support; \sa cbegin()
    ConstEntriesIterator begin() const { return cbegin(); }
    ConstEntriesIterator end() const { return cend(); }
```

rousskov (dismissed):
@kinkie, I am done making changes and am OK with the current code. Please review, including the adjusted PR title/description.

kinkie:
Sounds good!
I might look into an (additional) iterator that skips over expired entries as a followup PR, there might be a use case for it

rousskov (approved):

kinkie:
> P.S. IIRC, IP cache should report expired entries.

I agree. That's what it's doing now, among other things

PR #1411, opened 2023-07-07
GCC complains because memrchr is not defined

stranger777:
And code must be *memrchr?

GCC complains because memrchr is not defined. Maybe it's a typo and code must be *memrchr? I'm not sure but it seems logically for the code

### DIFF
--- a/src/sbuf/SBuf.cc
+++ b/src/sbuf/SBuf.cc
@@ -711,7 +711,7 @@ SBuf::rfind(char c, SBuf::size_type endPos) const
     if (length() == 0)
         return endPos;
 
-    const void *i = memrchr(buf(), (int)c, (size_type)endPos);
+    const void *i = *memrchr(buf(), (int)c, (size_type)endPos);
 
     if (i == nullptr)
         return npos;


rousskov:
> GCC complains because memrchr is not defined.

Thank you for trying to fix this build problem! I have not seen it reported recently. We have fixed similar problems in 2014 (see commit ee4d4e8), and it is possible that something else have changed to resurrect them, or you have a rather unusual build environment. What official commit SHA are you building? What is your build environment?

Do you see HAVE_STRING_H defined in your include/autoconf.h after you run ./configure?

Do you see HAVE_MEMRCHR defined in your include/autoconf.h after you run ./configure?

You may also want to share the relevant ./configure output lines:

```
checking for string.h... yes
checking for memrchr... yes
```

FWIW, in my environment, memrchr() declaration is provided in the system string.h header (/usr/include/string.h), and "man memrchr" suggests the same location. Where is that name declared in your environment? Does your system have string.h that declares memrchr?

> Maybe it's a typo and code must be *memrchr?

FWIW, this code compiles in many environments since 2016, looks OK to me, should not compile if you dereference that memrchr() result AFAICT, and such a typo would not explain a "memrchr not defined" error.

yadij:
@stranger777, additional to @rousskov's questions please indicate what OS you are using that gets these results.

kinkie:
@stranger777 , any updates? It's been a few months now, and I am planning to close the PR unless you have any news as noone else can currently reproduce it

yadij:
Closing for lack of progress as we are trying to reduce the PR queue.

PR #1400, opened 2023-06-29
Bug 5187: Work around REQMOD satisfaction regression

briskycat:
Commit ba3fe8d broke ICAP REQMOD satisfaction transactions. In some
cases, this workaround may resurrect Squid Bug 5187. Triage available at
https://bugs.squid-cache.org/show_bug.cgi?id=5187#c6

PR #1412, opened 2023-07-09
Bug 5187: Work around REQMOD satisfaction regression in v6

briskycat:
Commit ba3fe8d broke ICAP REQMOD satisfaction transactions. In some cases, this workaround may resurrect Squid Bug 5187. Triage available at https://bugs.squid-cache.org/show_bug.cgi?id=5187#c6


briskycat:
This is not relevant any longer (as per https://github.com/squid-cache/squid/pull/1400#issuecomment-1649731683).

PR #1413, opened 2023-07-10
Bug 5185: IPcache lookup of domain names is case-sensitive

sachin1gupta1980:


sachin1gupta1980:
I am not sure how can we test it in local setup. If someone can help with steps  for running test cases in local setup, please let me know.

rousskov:
@sachin1gupta1980, thank you for opening this PR that addresses many #1398 concerns. Would you be able to help with the TODO items in the updated PR description?

FWIW, for most of them, implementing a simple loop that tests a million or so hashes of randomly generated 0-16 byte c-strings may be enough. See RandomUuid(void) constructor for a hint on generating 0-16 byte c-strings (and if you use the same seed, you will get the same sequence of strings for all tests, allowing for a cleaner hashX-vs-hashY comparison). You can _temporary_ place that test code inside SquidMain(), before the configuration is parsed, and exit after the test, so that you do not need to worry about linking with hash.c (and can use relevant Squid functions). There are many other ways of doing all this, of course!

P.S. You may be able to address CI [failure](https://github.com/squid-cache/squid/actions/runs/5504516363/jobs/10039544679?pr=1413) by formatting changed sources with something like the following command:

```sh
./scripts/source-maintenance.sh --with-astyle /name/of/your/astyle-v3.1-executable
```

kinkie:
@sachin1gupta1980 , would you like to continue developing this or should someone else carry on the torch?
Thanks!

kinkie:
I'd like to do away with hash_table and instead use std::containers, but until such an endeavour is underatken, this will help

kinkie (approved):

lib/tests/testHash.cc
```diff
@@ -0,0 +1,45 @@
...
+};
+void
+TestLookupTable::testLookupTableLookup()
+{
+   // Test thet hash4/casehash4 returns equal result for lower case string
+    for(const char * domain : lowerCaseString)
+        CPPUNIT_ASSERT_EQUAL(casehash4(domain), hash4(domain));
+       
+   // Test thet hash4/casehash4 returns  result mix case string
+    for(const char * domain : lowerCaseString)
```
rousskov, line 42: I did not want to block this _draft_ PR with broken CI tests and many unfinished TODO items, but since it has been approved by @kinkie, I have to block it.

Please ping me when top TODO items are completed (or if you want to challenge their validity), lib/tests/testHash.cc is updated/integretated with the current unit test harness, this test case is fixed to match its description, and GitHub Actions CI tests are all green. At that time, I will polish the testing code (if still needed) and re-review.

rousskov (changes requested):

lib/tests/testHash.cc
sachin1gupta1980: I am sorry, I could not get to review comments. I will update the test cases and update PR

PR #1414, opened 2023-07-11
Drop ClientHttpRequest::flags::internal

yadij:
This flag is unnecessary when HttpRequest::flags::internal
is available, which is almost all code needing to detect
internal traffic.

The exception to this is parsing prior to HttpRequest
creation, which already uses internalCheck() API instead.

### DIFF
--- a/src/client_side.cc
+++ b/src/client_side.cc
@@ -1405,9 +1405,6 @@ ConnStateData::parseHttpRequest(const Http1::RequestParserPointer &hp)
         /* internal URL mode */
         /* prepend our name & port */
         http->uri = xstrdup(internalLocalUri(nullptr, hp->requestUri()));
-        // We just re-wrote the URL. Must replace the Host: header.
-        //  But have not parsed there yet!! flag for local-only handling.
-        http->flags.internal = true;
 
     } else if (port->flags.accelSurrogate) {
         /* accelerator mode */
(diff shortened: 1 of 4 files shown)


src/client_side.cc
```diff
@@ -1405,9 +1405,6 @@ ConnStateData::parseHttpRequest(const Http1::RequestParserPointer &hp)
         /* internal URL mode */
         /* prepend our name & port */
         http->uri = xstrdup(internalLocalUri(nullptr, hp->requestUri()));
-        // We just re-wrote the URL. Must replace the Host: header.
-        //  But have not parsed there yet!! flag for local-only handling.
-        http->flags.internal = true;
```
rousskov, line 1410: I assume you have tested that the PR code does the same thing as the official code after we get here with a relative URL and the Host header that does not match the URL authority component created by internalLocalUri(). If you have not tested that, please do!

rousskov (changes requested):
Thank you for removing this poorly duplicated flag.

Please consider using backticks to quote HttpRequest:flags:internal in the PR title and/or description, to stop GitHub from interpreting that `HttpRequest:flags:internal` sequence as markdown.

src/client_side.cc
yadij: I have tested that, yes it does AFAICT.

src/client_side.cc
```diff
@@ -1655,8 +1653,6 @@ clientProcessRequest(ConnStateData *conn, const Http1::RequestParserPointer &hp,
             request->flags.disableCacheUse("cache manager URL");
     }
 
-    request->flags.internal = http->flags.internal;
```
rousskov, line 1658: Disclaimer: I have not checked all the low-level details, especially whether the removal of this assignment may change Squid behavior in some (hypothetical) cases where `http->flags.internal` was false but `request->flags.internal` was true (and, hence, became false here in the official code, due to this assignment).

rousskov (approved):

PR #1416, opened 2023-07-11
Encapsulate HttpHeaderEntry length calculation

eduard-bagdasaryan:


rousskov (dismissed):
FYI: We are finishing a bug fix that needs this method as well.

yadij (dismissed):

rousskov (approved):
@yadij, I had to resolve a trivial conflict with a recent change, and that removed your approval. If you can re-approve, this PR will be merged sooner.

P.S. After we removed some unused code, addEntry() code is no longer duplicated, but we have a pending change that will duplicate that code again, so it is a good idea to merge this. I adjusted PR title accordingly.

yadij (approved):

PR #1418, opened 2023-07-12
MinGW: Fix 'error: cannot convert size_t*'

yadij:
The WinSock2 API definitions use size_t instead of
POSIX socklen_t for struct addrinfo members.

### DIFF
--- a/src/comm.cc
+++ b/src/comm.cc
@@ -128,7 +128,7 @@ comm_udp_recvfrom(int fd, void *buf, size_t len, int flags, Ip::Address &from)
     debugs(5,8, "comm_udp_recvfrom: FD " << fd << " from " << from);
     struct addrinfo *AI = nullptr;
     Ip::Address::InitAddr(AI);
-    int x = recvfrom(fd, buf, len, flags, AI->ai_addr, &AI->ai_addrlen);
+    int x = recvfrom(fd, buf, len, flags, AI->ai_addr, reinterpret_cast<socklen_t*>(&(AI->ai_addrlen)));
     from = *AI;
     Ip::Address::FreeAddr(AI);
     return x;
@@ -182,7 +182,7 @@ comm_local_port(int fd)
 
     Ip::Address::InitAddr(addr);
 
-    if (getsockname(fd, addr->ai_addr, &(addr->ai_addrlen)) ) {
+    if (getsockname(fd, addr->ai_addr, reinterpret_cast<socklen_t*>(&(addr->ai_addrlen))) ) {
         int xerrno = errno;
         debugs(50, DBG_IMPORTANT, "ERROR: " << MYNAME << "Failed to retrieve TCP/UDP port number for socket: FD " << fd << ": " << xstrerr(xerrno));
         Ip::Address::FreeAddr(addr);
--- a/src/comm/ConnOpener.cc
+++ b/src/comm/ConnOpener.cc
@@ -432,7 +432,7 @@ Comm::ConnOpener::lookupLocalAddress()
     struct addrinfo *addr = nullptr;
     Ip::Address::InitAddr(addr);
 
-    if (getsockname(conn_->fd, addr->ai_addr, &(addr->ai_addrlen)) != 0) {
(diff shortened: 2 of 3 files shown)


src/comm.cc
```diff
@@ -128,7 +128,7 @@ comm_udp_recvfrom(int fd, void *buf, size_t len, int flags, Ip::Address &from)
     debugs(5,8, "comm_udp_recvfrom: FD " << fd << " from " << from);
     struct addrinfo *AI = nullptr;
     Ip::Address::InitAddr(AI);
-    int x = recvfrom(fd, buf, len, flags, AI->ai_addr, &AI->ai_addrlen);
+    int x = recvfrom(fd, buf, len, flags, AI->ai_addr, reinterpret_cast<socklen_t*>(&(AI->ai_addrlen)));
```
rousskov, line 131: > The WinSock2 API definitions use size_t instead of POSIX socklen_t for struct addrinfo members.

AFAICT, on a platform where ai_addrlen member is size_t, but recvfrom() expects a pointer to socklen_t (i.e. `unsigned int`), our code is buggy beyond minor type compatibility issues -- we supply a pointer to an integer that has a different _size_ in environments where sizeof(int) is not sizeof(size_t). Most popular 64-bit environments are like that! A pointer reinterpreting type cast evidently shuts the compiler up, but it does not fix the actual bug.

BTW, WinSock2 [docs](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-recvfrom) appear to suggest that their recvfrom() does not expect POSIX socklen_t `fromlen` argument, making things even worse: They show recvfrom() parameter as being `int` rather than `unsigned int` (which is what POSIX socklen_t is according to other [docs](https://pubs.opengroup.org/onlinepubs/7908799/xns/syssocket.h.html)).

How does "native" Windows code use addrinfo with recvfrom()?

----

P.S. Linus discussed a similar problem back in [1998](https://yarchive.net/comp/linux/socklen_t.html)!

rousskov (changes requested):

src/comm.cc
kinkie: https://github.com/squid-cache/squid/blob/master/compat/os/mswindows.h#L626 seems to be doing the right thing. 
I'd say the big question is how to have mingw use that wrapper cleanly, or abstract it better if possible at all

kinkie (changes requested):
rely on compat/os/mswindows.h wrappers instead

src/comm.cc
rousskov: > https://github.com/squid-cache/squid/blob/master/compat/os/mswindows.h#L626 seems to be doing the right thing. 

Does it? To me, that Squid::recvfrom() helper is not doing anything _relevant_ at all -- it is using a socklen_t argument without casting. However, the macro wrapping that helper function appears to use the same problematic cast as the one discussed in this change request (for recvfrom() callers that supply an address of a  size_t ai_addrlen as `n`):

```C++
#define recvfrom(s,b,l,f,r,n) Squid::recvfrom(s,b,l,f,r,reinterpret_cast<socklen_t*>(n))
```

That cast may be OK in that macro usage contexts (I did not check), but it is not OK in general (AFAICT, as detailed in this change request).

> I'd say the big question is how to have mingw use that wrapper cleanly, or abstract it better if possible at all

I doubt that the best solution here can be based on that recvfrom() wrapper quoted above. I suspect the correct solution here is to use a local length variable of an appropriate type (and to copy from/to the caller's variable of an appropriate type).

kinkie: Huh? Is the wrapper not doing

    int ifl=*fl;
    if ((result = ::recvfrom(_get_osfhandle(s), (char *)b, l, f, fr, &ifl)) == SOCKET_ERROR) {

If anything, the preprocessor macro should do away with the reinterpret_cast because it's a noop at best, a bug at worst

rousskov: Ah, I see what you mean now. Yes, Squid::recvfrom() is copying (one way but not the other, which is a bug!). One can consider that a step forward, I guess, but we are still very far away from the correct overall implementation, including the most critical bits -- the wrapper function interface and the integration with the system recvfrom() interface!

yadij:
This PR has now been duplicated in PR #2046 and superseded.

PR #1419, opened 2023-07-12
Remove outdated Win32 select() module

yadij:
This code has quite a lot of bitrot compared to
the POSIX select(2) code in ModSelect.cc.

Modern Windows and MinGW apparently support
the POSIX API. So we should be able to just
use the normal ModSelect now. Any adjustments
should be added to ModSelect.cc when later
testing proves a need.


kinkie (approved):
simplify all the things

yadij:
@rousskov, it really is just bitrot and code de-duplication. Years of small optimizations and re-ordering of code in the Linux file without the same changes made here. Combined with the already mentioned fact that Windows has improved its POSIX support. Nothing that I can point at specifically has any urgency.

PR #1420, opened 2023-07-13
Miss if a 304 update would exceed reply_header_max_size

eduard-bagdasaryan:
Fetch the resource unconditionally when a 304 (Not Modified) response to
an internal cache revalidation request grows cached HTTP response
headers beyond the reply_header_max_size limit.


src/store.cc
```diff
@@ -1442,8 +1443,14 @@ StoreEntry::updateOnNotModified(const StoreEntry &e304)
     // update reply before calling timestampsSet() below
     const auto &oldReply = mem_obj->freshestReply();
     const auto updatedReply = oldReply.recreateOnNotModified(e304.mem_obj->baseReply());
-    if (updatedReply) // HTTP 304 brought in new information
+    if (updatedReply) { // HTTP 304 brought in new information
+        if (updatedReply->prefixLen() > Config.maxReplyHeaderSize) {
+            throw TextException(ToSBuf("cannot update the cached response because its updated ",
+                                       updatedReply->prefixLen(), "-byte header would exceed ",
+                                       Config.maxReplyHeaderSize, "-byte reply_header_max_size"), Here());
+        }
```
rousskov, line 1451: This change is the essence of this bug-fixing PR. The other changes are needed to provide the approximate message prefix length to this code (without duplicating related magical code) and to avoid code duplication when handling this problem/condition in two updateOnNotModified() callers.

src/http/StatusLine.h
```diff
@@ -44,7 +44,7 @@ class StatusLine
     /// retrieve the status code for this status line
     Http::StatusCode status() const { return status_; }
 
-    /// retrieve the reason string for this status line
+    /// the reason-phrase string; never nil
     const char *reason() const;
```
rousskov, line 48: I adjusted the existing description to document the never-nil guarantee used by the new caller of this method. This PR does not change how this method works. A lot can be polished in this and related classes, but this PR is not about that.

src/HttpReply.h
```diff
@@ -132,6 +132,10 @@ class HttpReply: public Http::Message
     /// \retval 0 implies that more data is needed to parse the response prefix
     size_t parseTerminatedPrefix(const char *, size_t);
 
+    /// approximate size of a "status-line CRLF headers CRLF" sequence
+    /// \sa HttpRequest::prefixLen()
+    size_t prefixLen() const;
```
rousskov, line 137: The name of this new method -- prefixLen() -- was chosen to match the name of the existing HttpRequest::prefixLen() method. It is not the best possible name (e.g., approximatePrefixLength() would be better), but I think it is best to keep the two method names in sync. We could even make them virtual, but there is no need to add that overhead right now.

If you would like, I can followup with a PR that renames the two methods, but please do not make that renaming a precondition for _this_ PR, to avoid creating a yet another long chain of minimal PRs getting stuck in the review queue.

rousskov, line 137: Regarding "approximate": It would be nice to return the exact packInto() length here, but the existing code is not ready for that: This method uses a combination of the existing StatusLine::reason() and Http::One::ResponseParser::firstLineSize() code. That combination does not quite match the existing packInto() code. In very rare cases, packed prefix length will be different. This approximation is _not_ a problem for the existing method callers, and fixing this mismatch would require non-trivial and controversial changes.

src/HttpRequest.cc
```diff
@@ -368,9 +368,7 @@ httpRequestPack(void *obj, Packable *p)
 int
 HttpRequest::prefixLen() const
 {
-    return method.image().length() + 1 +
-           url.path().length() + 1 +
-           4 + 1 + 3 + 2 +
+    return Http::One::RequestParser::RequestLineSize(method.image().length(), url.path().length()) +
```
rousskov, line 371: This change preserves the symmetry between request and response handling code. It also reduces code duplication. The magical-looking math is preserved in this transition to Http::One::RequestParser::RequestLineSize().

src/client_side_reply.cc
```diff
@@ -440,12 +440,18 @@ clientReplyContext::handleIMSReply(const StoreIOBuffer result)
...
+        if (!Store::Root().updateOnNotModified(old_entry, *http->storeEntry())) {
+            old_entry->release(true);
+            restoreState();
+            http->updateLoggingTags(LOG_TCP_MISS);
+            processMiss();
+            return;
+        }
+
+        http->updateLoggingTags(LOG_TCP_REFRESH_UNMODIFIED);
+        http->request->flags.staleIfHit = false; // old_entry is no longer stale
```
rousskov, line 454: The diff does not show this well, but these two "new" code lines is the _old_ 304 handling code. No changes on this "successful 304 update" code path.

src/http/one/RequestParser.cc
```diff
@@ -14,11 +14,17 @@
 #include "SquidConfig.h"
 
 Http1::Parser::size_type
-Http::One::RequestParser::firstLineSize() const
+Http::One::RequestParser::RequestLineSize(const size_t methodLength, const size_t requestTargetLength)
```
rousskov, line 17: See Http::One::ResponseParser::StatusLineSize() comment regarding the name change here.

src/http/one/ResponseParser.cc
```diff
@@ -18,11 +18,11 @@
 const SBuf Http::One::ResponseParser::IcyMagic("ICY ");
 
 Http1::Parser::size_type
-Http::One::ResponseParser::firstLineSize() const
+Http::One::ResponseParser::StatusLineSize(const AnyP::ProtocolVersion &msgProtocol, const size_t reasonLength)
```
rousskov, line 21: Regarding the name change: To avoid duplication of this non-trivial code, we needed to convert this method into a static function member. Long-term, this non-parsing code should not be inside a (parser) class at all -- it should be a free function -- but we could not move it outside in this PR because the method implementation uses some (needlessly) protected constants from the parent class. When the parent class code is adjusted to avoid hiding those constants, this should become a free function. And that free function will use this response-specific name to avoid clashing with a similar request-specific function.

src/store/Controller.cc
```diff
@@ -708,20 +708,29 @@ Store::Controller::updateOnNotModified(StoreEntry *old, StoreEntry &e304)
...
-        return;
+    try {
+        if (!old->updateOnNotModified(e304)) {
+            debugs(20, 5, "updated nothing in " << *old << " with " << e304);
+            return true;
+        }
+    } catch (...) {
+        debugs(20, DBG_IMPORTANT, "ERROR: Failed to update a cached response: " << CurrentException <<
+               Debug::Extra << "cache entry that needed revalidation: " << *old <<
+               Debug::Extra << "304/revalidation response: " << e304);
```
rousskov, line 723: I am not sure we should report these low-level details to cache.log (at level 1). They might be handy in developer-driven triage, but even that assertion is questionable, and admins would not know what to do with this information. If others think that these details should not be dumped, let's remove them!

N.B. A part of StoreEntry gists printed by these two lines is the StoreEntry object memory address. Exposing that info has been flagged as a potential threat in the past.

src/http/one/ResponseParser.cc
```diff
@@ -18,11 +18,11 @@
 const SBuf Http::One::ResponseParser::IcyMagic("ICY ");
 
 Http1::Parser::size_type
-Http::One::ResponseParser::firstLineSize() const
+Http::One::ResponseParser::StatusLineSize(const AnyP::ProtocolVersion &msgProtocol, const size_t reasonPhraseLength)
```
rousskov, line 21: Regarding the name change: To avoid duplication of this non-trivial code, we needed to convert this method into a static function member. Long-term, this non-parsing code should not be inside a (parser) class at all -- it should be a free function -- but we could not move it outside in this PR because the method implementation uses some (needlessly) protected constants from the parent class. When the parent class code is adjusted to avoid hiding those constants, this should become a free function. And that free function will use this response-specific name to avoid clashing with a similar request-specific function.

src/HttpReply.h
yadij: Case in point what I am requesting you do **not** try to (de-)duplicate the parser input normalization calculation for this on-wire output calculation.

yadij (changes requested):

yadij:
Do you have a bug reference for this? or is it just a theoretical issue being resolved?

rousskov:
> Do you have a bug reference for this? or is it just a theoretical issue being resolved?

@yadij, you should have received a notification from GitHub about this PR, dated July 13, 2023, shortly after this PR was published. In my copy of that notification, the subject line starts with `Re: [squid-cache/squid] Factory/O`....

src/http/one/ResponseParser.cc
rousskov: > To know what size will be generated on output please add a method to `class StatusLine` calculating what length its output will be.

Done in commit 78aaaf9.

src/store/Controller.cc
rousskov: Removed in branch commit f7d15b9.

src/http/one/RequestParser.cc
rousskov: > Please drop these changes.

Done in branch commit dd6fc48.

rousskov (commented):
@yadij, I removed the changes you did not like and added Http::StatusLine::packedLength() that duplicates some of the Http::StatusLine::packInto() logic instead. This decreases the overall footprint of this PR at the expense of more code duplication. It is possible to avoid this code duplication, but not without quite a few out-of-scope changes, so perhaps this cost is acceptable. Please re-review.

@eduard-bagdasaryan, please check and retest the changes I have pushed and adjust as needed.

yadij (approved):

eduard-bagdasaryan:
> check and retest the changes I have pushed

Done, no problems spotted.

PR #1422, opened 2023-07-13
Do not use static initialization to register modules

rousskov:
ERROR: ... Unknown authentication scheme 'ntlm'.

When a translation unit does not contain main() and its code is not used
by the rest of the executable, the linker may exclude that translation
unit from the executable. This exclusion by LTO may happen even if that
code _is_ used to initialize static variables in that translation unit:
"If no variable or function is odr-used from a given translation unit,
the non-local variables defined in that translation unit may never be
initialized"[^1].

[^1]: https://en.cppreference.com/w/cpp/language/initialization

For example, src/auth/ntlm/Scheme.o translation unit contains nothing
but NtlmAuthRr class definition and static initialization code. The
linker knows that the rest of Squid does not use NtlmAuthRr and excludes
that translation unit from the squid executable, effectively disabling
NTLM module registration required to parse "auth_param ntlm" directives.

The problem does affect existing NTLM module, and may affect any future
module code as we reduce module's external footprint. This change
converts all RegisteredRunner registrations from using side effects of
static initialization to explicit registration calls from SquidMain().
Relying on "life before main()" is a design bug. This PR fixes that bug
with respect to RegisteredRunner registrations.

Due to indeterminate C++ static initialization order, no module can
require registration before main() starts. Thus, moving registration
timing to the beginning of SquidMain() should have no negative effects.

The new registration API still does not expose main.cc to any module
details (other than the name of the registration function itself). We
still do not need to #include any module headers into main.cc. Compiler
or linker does catch most typos in RegisteredRunner names.

Unfortunately, explicit registration still cannot prevent bugs where we
forget to register a module or fail to register a module due to wrong
registration code guards. Eventually, CI will expose such bugs.

### DIFF
--- a/src/base/RunnersRegistry.h
+++ b/src/base/RunnersRegistry.h
@@ -128,27 +128,46 @@ bool UseThisStatic(const void *);
         RegisterRunner(new Who) > 0 && \
         UseThisStatic(& Who ## _Registered_);
 
-/// helps DefineRunnerRegistrator() and CallRunnerRegistrator() to use the same
-/// registration function name
-#define DeclareRunnerRegistrator(Who) \
-    void Register ## Who ## Now()
+/// helps DefineRunnerRegistrator() and CallRunnerRegistrator() (and their
+/// namespace-sensitive variants) to use the same registration function name
+#define NameRunnerRegistrator_(Namespace, ClassName) \
+    Register ## ClassName ## In ## Namespace()
 
-/// Define registration code for the given RegisteredRunner class. A matching
-/// CallRunnerRegistrator(Who) call should run this code before any possible use
-/// of the being-registered module.
-#define DefineRunnerRegistrator(Who) \
-    DeclareRunnerRegistrator(Who); \
-    void Register ## Who ## Now() { \
-        const auto registered = RegisterRunner(new Who); \
+/// helper macro that implements DefineRunnerRegistrator() and
+/// DefineRunnerRegistratorIn() using supplied constructor expression
+#define DefineRunnerRegistrator_(Namespace, ClassName, Constructor) \
+    void NameRunnerRegistrator_(Namespace, ClassName); \
+    void NameRunnerRegistrator_(Namespace, ClassName) { \
+        const auto registered = RegisterRunner(new Constructor); \
         assert(registered); \
     }
 
-/// convenience macro: register one RegisteredRunner kid
-#define CallRunnerRegistrator(Who) \
+/// Define registration code for the given RegisteredRunner-derived class known
+/// as ClassName in Namespace. A matching CallRunnerRegistratorIn() call should
+/// run this code before any possible use of the being-registered module.
+/// \sa DefineRunnerRegistrator() for classes declared in the global namespace.
+#define DefineRunnerRegistratorIn(Namespace, ClassName) \
+    DefineRunnerRegistrator_(Namespace, ClassName, Namespace::ClassName)
+
+/// Define registration code for the given RegisteredRunner-derived class known
+/// as ClassName (in global namespace). A matching CallRunnerRegistrator() call
+/// should run this code before any possible use of the being-registered module.
+/// \sa DefineRunnerRegistratorIn() for classes declared in named namespaces.
+#define DefineRunnerRegistrator(ClassName) \
+    DefineRunnerRegistrator_(Global, ClassName, ClassName)
+
+/// Register one RegisteredRunner kid, known as ClassName in Namespace.
+/// \sa CallRunnerRegistrator() for classes declared in the global namespace.
+#define CallRunnerRegistratorIn(Namespace, ClassName) \
     do { \
-        DeclareRunnerRegistrator(Who); \
-        Register ## Who ## Now(); \
+        void NameRunnerRegistrator_(Namespace, ClassName); \
+        NameRunnerRegistrator_(Namespace, ClassName); \
     } while (false)
+
+/// Register one RegisteredRunner kid, known as ClassName (in the global namespace).
+/// \sa CallRunnerRegistratorIn() for classes declared in named namespaces.
+#define CallRunnerRegistrator(ClassName) \
+    CallRunnerRegistratorIn(Global, ClassName)
 
 #endif /* SQUID_BASE_RUNNERSREGISTRY_H */
 
--- a/src/base/RunnersRegistry.h
+++ b/src/base/RunnersRegistry.h
@@ -118,16 +118,6 @@ protected:
     debugs(1, 2, "running " # m); \
     RunRegistered(&m)
 
-/// convenience function to "use" an otherwise unreferenced static variable
-bool UseThisStatic(const void *);
-
-// TODO: Remove, together with UseThisStatic(), after migrating to DefineRunnerRegistrator()
-/// convenience macro: register one RegisteredRunner kid as early as possible
-#define RunnerRegistrationEntry(Who) \
-    static const bool Who ## _Registered_ = \
-        RegisterRunner(new Who) > 0 && \
-        UseThisStatic(& Who ## _Registered_);
-
 /// helps DefineRunnerRegistrator() and CallRunnerRegistrator() (and their
 /// namespace-sensitive variants) to use the same registration function name
 #define NameRunnerRegistrator_(Namespace, ClassName) \
--- a/src/base/RunnersRegistry.h
+++ b/src/base/RunnersRegistry.h
@@ -121,11 +121,34 @@ protected:
 /// convenience function to "use" an otherwise unreferenced static variable
 bool UseThisStatic(const void *);
 
+// TODO: Remove, together with UseThisStatic(), after migrating to DefineRunnerRegistrator()
 /// convenience macro: register one RegisteredRunner kid as early as possible
 #define RunnerRegistrationEntry(Who) \
     static const bool Who ## _Registered_ = \
         RegisterRunner(new Who) > 0 && \
         UseThisStatic(& Who ## _Registered_);
+
+/// helps DefineRunnerRegistrator() and CallRunnerRegistrator() to use the same
+/// registration function name
+#define DeclareRunnerRegistrator(Who) \
+    void Register ## Who ## Now()
+
+/// Define registration code for the given RegisteredRunner class. A matching
+/// CallRunnerRegistrator(Who) call should run this code before any possible use
+/// of the being-registered module.
+#define DefineRunnerRegistrator(Who) \
+    DeclareRunnerRegistrator(Who); \
+    void Register ## Who ## Now() { \
+        const auto registered = RegisterRunner(new Who); \
+        assert(registered); \
+    }
+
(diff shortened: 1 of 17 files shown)


src/base/RunnersRegistry.h
```diff
@@ -121,11 +121,34 @@ class IndependentRunner: public RegisteredRunner
 /// convenience function to "use" an otherwise unreferenced static variable
 bool UseThisStatic(const void *);
 
+// TODO: Remove, together with UseThisStatic(), after migrating to DefineRunnerRegistrator()
```
rousskov, line 124: This is a draft PR because it lacks conversion of the remaining dozen of RunnerRegistrationEntry() calls and the removal of the corresponding macros. I am posting this draft PR to solicit feedback about the proposed registration interface (so that we do not need to modify those dozens of calls multiple times).

To minimize the number of uncaught registration typos, I also would like to reject duplicate registrations (at startup time), but I need to double check that the corresponding (trivial) RegisterRunner_() enhancement will not break IndependentRunners first. We can now also add "registering such-and-such" debugging so it is easy to tell whether Squid recognized that a given module must be registered. These enhancements do not have to be done in this PR though.

rousskov (commented):
In my quick-and-dirty tests, this draft PR fixes the NTLM problem described on [squid-users](http://lists.squid-cache.org/pipermail/squid-users/2023-July/025959.html), but the underlying issue is bigger than NTLM code -- it has the potential to affect any dynamically registered module (and we should see more and more such registrations as we improve Squid code!).

src/main.cc
```diff
@@ -1440,6 +1440,14 @@ StartUsingConfig()
 int
 SquidMain(int argc, char **argv)
 {
+    // We must register all modules before the first RunRegisteredHere() call.
+    // We do it ASAP/here so that we do not need to move this code when we add
+    // earlier hooks to the RegisteredRunner API. This collection of
+    // registration calls is not a RegisteredRunner "event" in itself.
+#if HAVE_AUTH_MODULE_NTLM
+    CallRunnerRegistrator(NtlmAuthRr);
+#endif
```
kinkie, line 1449: Could we push it out to its own InitialiseSubmodules function?
Not a big deal but would make intent clearer

rousskov: Good idea! Done in commit 947b372.

> InitialiseSubmodules

I also emphasized that these module registrations do not _initialize_ those modules. We are (or should be), essentially, remembering a pointer to initialization (and, in general, event handling) code here, not calling the code behind that pointer.

kinkie (dismissed):

src/base/RunnersRegistry.h
rousskov: > This is a draft PR because it lacks conversion of the remaining dozen of RunnerRegistrationEntry() calls and the removal of the corresponding macros.

Done (as of branch commit c9edc47). This is no longer a draft PR.

> TODO: Reject duplicate registrations (at Squid startup time).

> To minimize the number of uncaught registration typos, I also would like to reject duplicate registrations (at startup time), but I need to double check that the corresponding (trivial) RegisterRunner_() enhancement will not break IndependentRunners first.

After examining the related code, I dropped the above TODO. I believe the enhancement mentioned above should be done in a dedicated PR because it requires a few code changes that have nothing to do with this PR. Technically, the official code also allows duplicate registrations, so this PR does not add that problem, and does not have to solve it.

src/dns_internal.cc
```diff
@@ -210,10 +210,10 @@ class ConfigRr : public RegisteredRunner
     void endingShutdown() override;
 };
 
-RunnerRegistrationEntry(ConfigRr);
-
 } // namespace Dns
 
+DefineRunnerRegistratorIn(Dns, ConfigRr);
```
rousskov, line 215: I had to move this and other in-namespace registrator function definitions into the global namespace. We cannot simply define the registrator function for Foo::Class inside the corresponding namespace Foo because the caller of that function -- RegisterModules() in main.cc -- would not be able to declare that registrator function: Functions cannot open namespaces (to add declarations to them). And without such declaration inside RegisterModules(), a call to that registrator function there would generate compiler "not declared in this scope" errors.

We could add a third group of macros that will open the namespace and declare registrator functions there as needed, but the corresponding code noise and code duplication are not worth the namespace purity in this case. Similarly, we could add a whole dedicated header file that main.cc will `#include` instead of self-declaring these registrator functions, but that would create even more noise! The current solution is the lesser of all these evils IMO.

src/main.cc
```diff
@@ -1437,9 +1437,52 @@ StartUsingConfig()
...
+    CallRunnerRegistrator(PeerPoolMgrsRr);
+    CallRunnerRegistrator(SharedMemPagesRr);
+    CallRunnerRegistrator(SharedSessionCacheRr);
+    CallRunnerRegistrator(TransientsRr);
+    CallRunnerRegistratorIn(Dns, ConfigRr);
+    CallRunnerRegistratorIn(Rock, SwapDirRr);
+
+#if USE_SQUID_ESI && HAVE_LIBEXPAT
+    CallRunnerRegistratorIn(Esi, ExpatRr);
+#endif
```
rousskov, line 1463: Guards like these are the biggest drawback of this solution IMO. RegisterModules() has to know what are the preconditions for the corresponding module registrator existence. The only solution for this problem (besides the buggy static initialization-based one that this PR removes), is automated generation of module registration code at Squid build time, when the list of enabled (to build) modules becomes known. That solution requires a lot of work and will complicate this code comprehension further. I do not think it is worth it at this point, but we can consider it if these module registration guards go out of hand.

src/main.cc
```diff
@@ -1437,9 +1437,52 @@ StartUsingConfig()
...
+
+#if USE_SQUID_ESI && HAVE_LIBXML2
+    CallRunnerRegistratorIn(Esi, Libxml2Rr);
+#endif
+
+#if HAVE_AUTH_MODULE_NTLM
```
rousskov, line 1469: I was not sure whether this condition is supposed to be written as in the suggestion below and picked the simplest variant that appears to work. Please apply the suggestion below if it is the recommended way of testing NTLM authentication module availability.

```suggestion
#if USE_AUTH && HAVE_AUTH_MODULE_NTLM
```

yadij: the one you have in the PR code is correct.

yadij (commented):
The problem description for this PR is incorrect. It predicates the solution and extends the change across other non-problematic runner registrations on the basis that a call from main() is required.

However the problem case situation appears when LTO optimization is performed by compilers linking convenience libraries to each other (`auth/ntlm/libntlmauth.;a` module is linked to `auth/libauth.la` before linking to `main.o`). This breaks the `UseThisStatic()` workaround in `RunnerRegistrationEntry()` used to avoid the same problem linking convenience libraries to `.o` objects.

The best long term solution is to make the optional modules proper dynamic libraries. as was attempted in PR #1053 and PR #1052. That way their `UseThisStatic()` are preserved in the `dlopen()` load sequence and runners initialized on library loading.

yadij:
@rousskov, if you are able to replicate the issue reliably can you test adding the ```__attribute__((used))``` or ```[[gnu::used]]``` attributes to the definition of ```UseThisStatic()``` in `base/RunnersRegistry.h` - that may be what we need to inform the optimizer that the code is actually needed.

rousskov:
> The problem description for this PR is incorrect.

What exactly is incorrect? I am happy to adjust, but I do not know what description problems you see. I did adjust the description to be more precise about why this code is being removed. If that does not help, please be more specific.

> It predicates the solution and extends the change across other non-problematic runner registrations on the basis that a call from main() is required.

AFAICT, there is a misunderstanding of what problems this PR is fixing. This PR is fixing two problems:

1. Broken NTLM module registration. This is a surface-level problem that has many possible workarounds, including several mentioned in your PR comments so far. This PR addresses this problem by fixing its underlying cause (described in the next item). NTLM module registration problem triggered this PR work, but this PR is not specific to that problem.

2. Squid reliance on static initialization side effects also known as "life before main()". Relying on "life before main()" is a design bug. While "life before main()" is supported by C++ language, developers should avoid that language misfeature. This PR removes one such bug from Squid code. FWIW, we are not alone in our desire to avoid this problematic design. For example, according to "Static Initializers are a problem" [summary](https://github.com/rust-lang/rust/issues/111921) section, Chromium bans them, and Google almost bans them.

> However the problem case situation appears when LTO optimization is performed by compilers linking convenience libraries to each other (`auth/ntlm/libntlmauth.;a` module is linked to `auth/libauth.la` before linking to `main.o`). This breaks the `UseThisStatic()` workaround in `RunnerRegistrationEntry()` used to avoid the same problem linking convenience libraries to `.o` objects.

```
$ nm src/auth/.libs/libauth.a | grep -E 'NtlmAuthRr|UseThisStatic'
    U _Z13UseThisStaticPKv
    0000000000000011 b _ZL22NtlmAuthRr_Registered_
    ... and other NtlmAuthRr symbols ...
$ nm src/auth/.libs/libauth.a  | grep -E 'NtlmAuthRr|UseThisStatic'
    U _Z13UseThisStaticPKv
    0000000000000011 b _ZL22NtlmAuthRr_Registered_
    ... and other NtlmAuthRr symbols ...
$ nm src/squid  | grep -E 'NtlmAuthRr|UseThisStatic'
    00000000007083ae T _Z13UseThisStaticPKv
    ... no NtlmAuthRr symbols at all ...
```

As you can see, in my tests, unreachable (from the linker point of view) code is only removed when linking `src/squid` executable -- both `src/auth/ntlm/.libs/libntlm.a` and `src/auth/.libs/libauth.a` static libraries still have NtlmAuthRr code. That makes a lot of sense for me because, at library building time, the linker cannot tell whether NtlmAuthRr code will be unused. However, I am happy to assume that the same code could have been removed earlier in another environment! None of this PR reasoning relies on that removal being at executable linking stage. It is just what happens in my tests (and what makes sense).



> The best long term solution is to make the optional modules proper dynamic libraries.

Any future dynamic libraries we author should not re-introduce the "life before main()" design bug fixed in this PR, even if it can get away with it because some tools are not smart enough to delete unreachable code at dynamic linking time. Instead, such a dynamic library should be initialized by calling a well-known or configurable library initialization function, without relying on any C++ static initialization side effects.

> if you are able to replicate the issue reliably can you test adding the `__attribute__((used))` or `[[gnu::used]]` attributes to the definition of `UseThisStatic()` in `base/RunnersRegistry.h` - that may be what we need to inform the optimizer that the code is actually needed.

Adding `__attribute__((used))` did not help in my tests, but note that the optimizer already thinks that UseThisStatic() is used/needed (see `nm` output above). That function _is_ included in the executable.

The problem is not UseThisStatic(). The problem is that the entire NtlmAuthRr code is deemed unreachable from, roughly speaking, main(). It does not matter what functions that unreachable code uses. What matters is whether _its_ code is used by _others_. That translation unit does not provide anything the executable calls/uses, directly or indirectly, and it gets removed.

yadij (approved):

squidadm:
queued for backport to v6

PR #1421, opened 2023-07-13
Bug 5154: Reject IPv6 address when IPv6 is disabled

FireBurn:
This is a continuaton of the work by Jeffrey Kintscher

PR #1423, opened 2023-07-18
Allow unit tests to customize their initialization code

rousskov:
CppUnit provides TestFixture::setUp() and tearDown() methods, but those
methods are executed before and after _each_ test case. Many single
TestFixture-derived classes register many test cases that need to run
some test program setup/initialization code just once. Counting past
setUp() calls would increase noise and likely introduce bugs.

Some of our test cases may actually need before/after cleanup -- the
functionality already provided by TestFixture::setUp()/tearDown()
methods.  Moreover, when multiple TestFixture-derived classes share the
same test executable, their collections of test cases are executed in
essentially random order (see commit 27685ef). In those use cases,
insuring one-time initialization via setUp() hacks would be especially
awkward and error-prone.

With this solution, a test program with custom needs just needs to
define a TestProgram-derived class that implements its one-time startup
logic. The same TestProgram class hierarchy might also prove useful in
future custom test execution adjustments.

Also migrated away from the "include main()" unit test design to avoid
adding risky hacks that register custom TestProgram-derived classes and
to reduce "magic" in test code (while following our style guidelines).
Every test programs now declares its own (trivial) main() rather than
getting it from include/unitTestMain.h.

### DIFF
--- a/include/unitTestMain.h
+++ b/include/unitTestMain.h
@@ -21,9 +21,6 @@
 #include <cppunit/TestResultCollector.h>
 #include <cppunit/TestRunner.h>
 
-#include <memory>
-#include <stdexcept>
-
 /// implements test program's main() function while enabling customization
 class TestProgram
 {
@@ -37,14 +34,14 @@ public:
     /// Implements main(), combining all the steps.
     /// Must be called from main().
     /// \returns desired main() result.
-    int run();
+    int run(int argc, char *argv[]);
 
 private:
     bool runTests();
 };
 
 int
-TestProgram::run()
+TestProgram::run(int, char *[])
 {
 #if ENABLE_DEBUG_SECTION
     Debug::Levels[ENABLE_DEBUG_SECTION] = 99;
@@ -84,37 +81,6 @@ TestProgram::runTests()
     outputter.write();
 
     return result.wasSuccessful();
-}
-
-/// TestProgram object registered by RegisterTestProgram() (or nil).
-static std::unique_ptr<TestProgram> TestProgram_;
-
-/// Creates and registers a CustomTestProgram object. Optionally "uses" the
-/// static defined by RegisterTestProgram (by taking that variable address).
-/// Use RegisterTestProgram_() instead.
-template <class CustomTestProgram>
-static bool
-RegisterTestProgram_(void * = nullptr)
-{
-    // Avoid assert(): Some tests do not link with libcompatsquid.la (XXX?).
-    // assert(!TestProgram_);
-    if (TestProgram_)
-        throw std::runtime_error("attempt to register more than one TestProgram");
-
-    TestProgram_.reset(new CustomTestProgram());
-    return true;
-}
-
-/// A helper macro to create and register a CustomTestProgram object.
-#define RegisterTestProgram(CustomTestProgram) \
-    static bool CustomTestProgramRegistration_ = RegisterTestProgram_<CustomTestProgram>(&CustomTestProgramRegistration_)
-
-int
-main(int, char *[])
-{
-    if (!TestProgram_)
-        RegisterTestProgram(TestProgram);
-    return TestProgram_->run();
 }
 
 #endif /* SQUID_INCLUDE_UNITTESTMAIN_H */
--- a/include/unitTestMain.h
+++ b/include/unitTestMain.h
@@ -34,10 +34,6 @@ public:
     /// Does nothing by default.
     virtual void startup() {}
 
-    /// Runs after all tests, regardless of their outcome.
-    /// Does nothing by default.
-    virtual void shutdown() {}
-
     /// Implements main(), combining all the steps.
     /// Must be called from main().
     /// \returns desired main() result.
@@ -56,8 +52,6 @@ TestProgram::run()
 
     startup();
     const auto result = runTests();
-    shutdown();
-
     return result ? 0 : 1;
 }
 
--- a/include/unitTestMain.h
+++ b/include/unitTestMain.h
@@ -21,13 +21,51 @@
 #include <cppunit/TestResultCollector.h>
 #include <cppunit/TestRunner.h>
 
+#include <memory>
+#include <stdexcept>
+
+/// implements test program's main() function while enabling customization
+class TestProgram
+{
+public:
+    virtual ~TestProgram() = default;
+
+    /// Runs before all tests.
+    /// Does nothing by default.
+    virtual void startup() {std::cerr << "XXX1: Default Before\n";}
+
+    /// Runs after all tests, regardless of their outcome.
+    /// Does nothing by default.
+    virtual void shutdown() {std::cerr << "XXX9: Default After\n";}
+
+    /// Implements main(), combining all the steps.
+    /// Must be called from main().
+    /// \returns desired main() result.
+    int run();
+
+private:
+    bool runTests();
+};
+
 int
-main(int, char *[])
+TestProgram::run()
 {
 #if ENABLE_DEBUG_SECTION
     Debug::Levels[ENABLE_DEBUG_SECTION] = 99;
 #endif
 
+    startup();
+    const auto result = runTests();
+    shutdown();
+
+    return result ? 0 : 1;
+}
+
+/// runs all tests registered with CPPUNIT_TEST_SUITE_REGISTRATION() calls
+/// \returns whether all tests were successful
+bool
+TestProgram::runTests()
+{
     // Create the event manager and test controller
     CPPUNIT_NS::TestResult controller;
 
@@ -51,7 +89,38 @@ main(int, char *[])
     CPPUNIT_NS::CompilerOutputter outputter( &result, std::cerr );
     outputter.write();
 
-    return result.wasSuccessful() ? 0 : 1;
+    return result.wasSuccessful();
+}
+
+/// TestProgram object registered by RegisterTestProgram() (or nil).
+static std::unique_ptr<TestProgram> TestProgram_;
+
+/// Creates and registers a CustomTestProgram object. Optionally "uses" the
+/// static defined by RegisterTestProgram (by taking that variable address).
+/// Use RegisterTestProgram_() instead.
+template <class CustomTestProgram>
+static bool
+RegisterTestProgram_(void * = nullptr)
+{
+    // Avoid assert(): Some tests do not link with libcompatsquid.la (XXX?).
+    // assert(!TestProgram_);
+    if (TestProgram_)
+        throw std::runtime_error("attempt to register more than one TestProgram");
+
+    TestProgram_.reset(new CustomTestProgram());
+    return true;
+}
+
+/// A helper macro to create and register a CustomTestProgram object.
+#define RegisterTestProgram(CustomTestProgram) \
+    static bool CustomTestProgramRegistration_ = RegisterTestProgram_<CustomTestProgram>(&CustomTestProgramRegistration_)
+
+int
+main(int, char *[])
+{
+    if (!TestProgram_)
+        RegisterTestProgram(TestProgram);
+    return TestProgram_->run();
 }
 
 #endif /* SQUID_INCLUDE_UNITTESTMAIN_H */
--- a/include/unitTestMain.h
+++ b/include/unitTestMain.h
@@ -32,11 +32,11 @@ public:
 
     /// Runs before all tests.
     /// Does nothing by default.
-    virtual void startup() {std::cerr << "XXX1: Default Before\n";}
+    virtual void startup() {}
 
     /// Runs after all tests, regardless of their outcome.
     /// Does nothing by default.
-    virtual void shutdown() {std::cerr << "XXX9: Default After\n";}
+    virtual void shutdown() {}
 
     /// Implements main(), combining all the steps.
     /// Must be called from main().
--- a/src/tests/testACLMaxUserIP.cc
+++ b/src/tests/testACLMaxUserIP.cc
@@ -63,8 +63,6 @@ public:
     void startup() override;
 };
 
-RegisterTestProgram(MyProgram);
-
 void
 MyProgram::startup()
 {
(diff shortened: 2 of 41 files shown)


src/tests/testHttpRequest.cc
```diff
@@ -24,9 +24,6 @@ class TestHttpRequest : public CPPUNIT_NS::TestFixture
     CPPUNIT_TEST(testSanityCheckStartLine);
     CPPUNIT_TEST_SUITE_END();
 
-public:
-    void setUp() override;
```
rousskov, line 28: > PR description: CppUnit provides TestFixture::setUp() and tearDown() methods, but those methods are executed before and after _each_ test case.

In fact, the whole TestFixture object is created and destroyed for each test case, and the destruction waits for all test cases (within the same TestFixture class?) to complete. Here is the output from temporary branch debugging (removed in branch commit a3ab387) that illustrates that sequence of events:

```
$ src/tests/testHttpRequest
SKIP: mem/libmem.la MemPools (not implemented).
SKIP: security/libsecurity.la PeerOptions (not implemented).
SKIP: store/libstore.la Controller (not implemented).
XXX1b: Custom Before
SKIP: mem/libmem.la Init (not implemented).
XXX2: Default TestHttpRequest ctor
XXX2: Default TestHttpRequest ctor
XXX2: Default TestHttpRequest ctor
............
OK (12)
XXX8: Default TestHttpRequest dtor
XXX8: Default TestHttpRequest dtor
XXX8: Default TestHttpRequest dtor
XXX9: Default After
SKIP: store/libstore.la ~Controller (not implemented).
```

src/tests/testACLMaxUserIP.cc
```diff
@@ -56,10 +53,21 @@ TestACLMaxUserIP::testDefaults()
...
+};
+
+RegisterTestProgram(MyProgram);
+
 void
-TestACLMaxUserIP::setUp()
+MyProgram::startup()
 {
-    CPPUNIT_NS::TestFixture::setUp();
     Acl::RegisterMaker("max_user_ip", [](Acl::TypeName name)->ACL* { return new ACLMaxUserIP(name); });
```
rousskov, line 71: This was probably a bug waiting to be discovered -- we were registering the same ACL maker function twice, once for each TestACLMaxUserIP test case.

include/unitTestMain.h
```diff
@@ -51,7 +89,38 @@ main(int, char *[])
...
+    if (TestProgram_)
+        throw std::runtime_error("attempt to register more than one TestProgram");
+
+    TestProgram_.reset(new CustomTestProgram());
+    return true;
+}
+
+/// A helper macro to create and register a CustomTestProgram object.
+#define RegisterTestProgram(CustomTestProgram) \
+    static bool CustomTestProgramRegistration_ = RegisterTestProgram_<CustomTestProgram>(&CustomTestProgramRegistration_)
```
rousskov, line 116: Note that these static initialization tricks are not going to be removed by LTO because they are located in the same translation unit as main() -- our test programs `#include` this file, and this file defines main(). Said that, if you know of a better way to handle this registration, please let me know. I could not find anything that would not require changing all test files, even those that do not need to customize startup/shutdown. If you think we should modify all of them to avoid these tricks, please let me know.

To see two (for now) usage examples of this macro outside of main(), grep this PR for RegisterTestProgram calls.

rousskov (commented):
This draft PR only converts two tests. I am posting it for feedback before all other tests that have custom setUp() or tearDown() methods that need conversion are converted. There are probably quite a few of them! I also left a comment regarding a specific LTO concern that you may want to help address.

FWIW, this PR was triggered by observing repeated BCP 177 warnings from the previously missing Ip::ProbeTransport() call that we are adding to some of the unit tests in #1421. This work may help address TODOs described in commits 27685ef and 23b7963 as well.

yadij (approved):

include/unitTestMain.h
```diff
@@ -21,13 +21,51 @@
...
 int
-main(int, char *[])
+TestProgram::run()
 {
 #if ENABLE_DEBUG_SECTION
     Debug::Levels[ENABLE_DEBUG_SECTION] = 99;
 #endif
 
+    startup();
+    const auto result = runTests();
```
kinkie, line 58: What if runTests() throws an exception? Is there any risk that we will leak system resources (e.g. shm segments)?
Would it make sense to catch, shutdown, rethrow? Or to extend the API to have an explicit on-exception handler? Or could that role be handled by the virtual descriptor, and in that case we should document it?

rousskov: > Your question made me think that adding shutdown() method was premature. I will be removing that method.

Done in commit c69f505.

include/unitTestMain.h
rousskov: > If you think we should modify all of them to avoid these tricks, please let me know.

I decided to remove these tricks, migrating away from "include main()" unit test design in commit 8908d8b. This comment details my reasoning:

As you know, official code defines main() in include/unitTestMain.h which is included by nearly all tests. This approach avoids duplicating main() definition (containing a simple single-line body) in each test file. However, placing main() in a header file:

* confuses new developers that are looking for main() as a well-known test program starting point; the code looks surprising or "magical";

* forces us to add clunky/risky/obscure RegisterTestProgram() interface (the subject of this comment thread) that uses static initialization side effects to "inject" additional functionality into TestProgram;

* violates basic developer expectations and Squid code style guidelines;

* complicates tests that need to customize their startup (about 25% of tests already contain some customization and, bugs notwithstanding, the portion of such tests is expected to _grow_). Arguably, these tests do not actually duplicate main() implementation because their main()s create a test-specific object rather than the "default" TestProgram object!

Tiny main() code duplication is the lesser evil here.

src/tests/testRock.cc
```diff
@@ -97,14 +103,24 @@ TestRock::setUp()
     if (Ipc::Mem::Segment::BasePath == nullptr)
         Ipc::Mem::Segment::BasePath = ".";
 
+    // diff reduction; TODO: inline
+    finishStartup();
```
rousskov, line 107: I plan to inline this call (before merging this PR). Inlining now will make it more difficult to see what the new startup code has changed and what it has preserved. Fortunately, this was the only set of tests with complicated setup changes.

src/tests/testRock.cc
```diff
@@ -97,14 +103,24 @@ TestRock::setUp()
...
+}
+
+void
+TestRock::setUp()
+{
+    CPPUNIT_NS::TestFixture::setUp();
+
+    if (0 > system ("rm -rf " TESTDIR))
+        throw std::runtime_error("Failed to clean test work directory");
```
rousskov, line 116: This is old code, in the old per-test-case initialization sequences. No changes here despite what the diff misrepresents.

include/unitTestMain.h
```diff
@@ -21,13 +21,42 @@
...
+    /// Must be called from main().
+    /// \returns desired main() result.
+    int run(int argc, char *argv[]);
+
+private:
+    bool runTests();
+};
+
 int
-main(int, char *[])
```
rousskov, line 25: With main() removal, we should rename this header file. I think that is best done in the renaming-only PR that will also rename many of the test case files. We have used this noise-reduction strategy in recent commit 41e9c9f as well.

rousskov (commented):
@kinkie, @yadij, thank you for prompt and useful initial feedback about this draft PR. I am done modifying this PR (with one diff reduction hack removal postponed until merge time) and am marking it as ready for regular review.

yadij (changes requested):

include/unitTestMain.h
```diff
@@ -21,13 +21,42 @@
 #include <cppunit/TestResultCollector.h>
 #include <cppunit/TestRunner.h>
 
+/// implements test program's main() function while enabling customization
+class TestProgram
```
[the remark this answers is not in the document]
rousskov: Classes representing basic concepts like "a program" or "an application" are very common indeed. New Squid developers reading this code should recognize familiar patterns.

> FME; What has changed since my almost identical proposal was vetoed (circa 2008) as `Bad Practice`?

I cannot answer that question without seeing that (proposal and its) veto. FWIW, I glanced through Subject lines of squid-dev postings from 2007 to 2009 and could not find anything relevant mentioning "application" or "program"; I also looked through same Subject lines mentioning "RFC". Needless to say, this search is not accurate or exhaustive, and I probably just missed your proposal. Can you share a pointer?

src/tests/testRock.cc
rousskov: Done in commit 6810ed1.

src/tests/testStore.cc
```diff
@@ -203,3 +203,23 @@ TestStore::testSwapMetaTypeClassification()
...
+public:
+    /* TestProgram API */
+    void startup() override;
+};
+
+void
+MyTestProgram::startup()
+{
+    Mem::Init();
+}
```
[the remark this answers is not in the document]
rousskov: Done in commit eba8c16 to avoid arguing.

src/tests/testString.cc
```diff
@@ -37,10 +34,16 @@ class TestString : public CPPUNIT_NS::TestFixture
...
+    /* TestProgram API */
+    void startup() override;
+};
 
 void
-TestString::setUp()
+MyTestProgram::startup()
 {
     Mem::Init();
 }
```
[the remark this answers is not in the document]
rousskov: Done in commit eba8c16 to avoid arguing.

include/unitTestMain.h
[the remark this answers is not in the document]
yadij: Sorry I got the year wrong. Latest proposal was 19 June 2017 "[PATCH] helper process library" - mailing list mirrors appear to have dropped the patch from my latest revision.
If you are okay with this MFC design now I can revamp the old patch and open a PR proposing its use for helpers alongside this one for test binaries.

include/unitTestMain.h
yadij: I only agreed with that delay because you indicated that there was major followup work that a delay would help simplify. I assume this PR is that followup change.

FTR; I am accepting this delayed rename for these as well on the assumption that the two large sets of `src/tests/` files needing rename are not exactly identical. Please do submit the rename PR ASAP though, having separate renames does make code archaeology harder with how git does renaming.

yadij (approved):

include/unitTestMain.h
yadij: It is out-of-scope for this PR, but perhapse we should have a team discussion on redesigning these to be proper unit tests instead of the smoke-tests they currently are.

include/unitTestMain.h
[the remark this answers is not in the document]
rousskov: > Latest proposal was 19 June 2017 "[PATCH] helper process library"

AFAICT, my [review](http://lists.squid-cache.org/pipermail/squid-dev/2017-June/009003.html) of that [proposal](http://lists.squid-cache.org/pipermail/squid-dev/2017-June/009002.html) matches this PR very well -- this PR fixes the _same_ main() problems (in the official unit testing code) that 2017 proposal would have introduced into Squid helper code and that my 2017 review has identified.

For example, the TestProgram class implemented in this PR matches the Helper::Instance or Helper::Application class fix suggested in my 2017 review, and the following main() function sketched in my 2017 review perfectly matches the main() functions added in this PR:

```C++
// 2017 review sketch:
int
main(int argc, char *argv[])
{
    return MyInstance().run(argc, argv);
}
```

BTW, instead of proposing a Program or Application class, that 2017 [patch](//lists.squid-cache.org/pipermail/squid-dev/attachments/20170619/13fb64c1/attachment-0001.bin) proposed a C++ _namespace_:

```C++
// 2017 patch (with comments stripped):
namespace HelperProcess {
extern std::string Name;
bool Start();
bool Options(int argc, char *argv[]);
void DoOneRequest(std::string &buffer);
} // namespace HelperProcess
```


> What has changed since my almost identical proposal was vetoed as `Bad Practice`?

I can now answer that question. Nothing has changed at the high-level AFAICT: 2017 proposal had serious architectural flaws that were identified in my 2017 review, with specific suggestions on how to address them. This PR does not have those flaws.

> If you are okay with this MFC design now

The Microsoft Foundation Class library ([MFC](https://learn.microsoft.com/en-us/cpp/mfc/mfc-application-architecture-classes?view=msvc-170)) design is a red herring: It was not proposed or reviewed in 2017. It is not under review here and now. The TestProgram class introduced in this PR has similarities with MFC [CWinApp](https://learn.microsoft.com/en-us/cpp/mfc/reference/cwinapp-class?view=msvc-170) class and with numerous "application" or "program" classes in numerous other libraries and programs. There was and still is nothing wrong with basic Program or Application class design per se, but it was not a part of that 2017 proposal.


> I can revamp the old patch and open a PR proposing its use for helpers

If you resume working on that helper refactoring, please address my 2017 concerns before posting a PR. Most of them are critical to getting the overall design right. Also, to avoid redoing a lot of changes, please do not convert more than one helper until there is an agreement that the PR is on the right track (but be prepared to convert them all at the end of the process).

include/unitTestMain.h
rousskov: > I only agreed with that delay because you indicated that there was major followup work that a delay would help simplify. I assume this PR is that followup change.

I cannot find the indications you are thinking about. I [said](https://github.com/squid-cache/squid/pull/1288#pullrequestreview-1312473019) the following regarding delaying renames in recent commit https://github.com/squid-cache/squid/commit/41e9c9f078b45aac6ad49fa93020a862bc278b9b, but that does not match your recollection above: "The only large known change that I left out of scope is renaming the corresponding source files: That work simply requires more time than I have now. IMO, we do not have to rename in this PR because this PR represents reasonably scoped progress, even without the file name changes."

> Please do submit the rename PR ASAP though

I was hoping that Francesco will do that renaming so that I can focus on other things. This PR changes require renaming a single file -- unitTestMain.h.  Commit https://github.com/squid-cache/squid/commit/41e9c9f078b45aac6ad49fa93020a862bc278b9b changes (and inconsistencies in existing official code) require renaming dozens of files. Said that, if I am made responsible for solving this problem as well, then I _will_ rename them all. Just let me know when you decide that I should be the one to do it.

@kinkie, do you plan to post a renaming map and then apply reviewed renaming changes? If you plan to work on that, do you need my help?

PR #1424, opened 2023-07-20
Do not use invasive lists to store CachePeers

eduard-bagdasaryan:
Using invasive lists for CachePeer objects gives no advantages, but
requires maintaining non-standard error-prone code that leads to
excessive locking and associated memory overheads.

Also fixed a neighbors_init() bug that resulted in accessing an already
deleted "looks like this host" CachePeer object.


src/CachePeers.cc
```diff
@@ -0,0 +1,56 @@
...
+    // Remember the number of polls to keep shifting each poll starting point,
+    // to avoid always polling the same group of peers before other peers and
+    // risk overloading that first group with requests.
+    if (!pollIndex)
+        ++peerPolls_;
+
+    // subtract 1 to set the very first pos to zero
+    const auto pos = (peerPolls_ - 1 + pollIndex) % size();
+
+    return storage[pos].get();
```
rousskov, line 27: This code re-implements the existing `first_ping` feature. This implementation avoids the dangers of having a raw pointer/iterator getting out of sync with its globally-modifiable container.

The code looks a bit strange/unusual because we need to support cache peer scans where each new scan starts with the "next" peer (for load balancing purposes) but iterates through all peers. The start of each scan is detected by zero `pollIndex`. There is similar code in Squid (e.g., storeDirSelectSwapDirRoundRobin()). This implementation avoids caller code duplication and makes the caller code much safer compared with the old ad-hoc approach. It also preserves idiosyncrasies of the related official code -- no `first_ping` functionality changes are expected here.

It is possible to completely refactor these cache peer scans, moving this code outside CachePeers, but those improvements will change the functionality of fairly esoteric Squid features. They are also quite difficult to test. I think they should not be attempted without a pressing need, and definitely not in _this_ PR (that focuses on basic CachePeer storage improvements).

rousskov (dismissed):
Untangling individual CachePeers from (storing) each other is a yet another step towards smooth reconfiguration support.

src/CachePeers.cc
yadij: IMO we should redesign how all these polling features work and interact. I agree on doing that as a separate project.

For now, please pass around the relevant `Pointer` objects instead of raw-pointers. The whole reason they exist is to avoid/solve the dangers/issues of raw-pointers you appear to be worried about.

src/SquidConfig.h
```diff
@@ -242,7 +243,7 @@ class SquidConfig
     size_t tcpRcvBufsz;
     size_t udpMaxHitObjsz;
     wordlist *mcast_group_list;
-    CachePeer *peers;
+    CachePeers *peers;
     int npeers;
```
yadij, line 247: You appear to have a `CurrentCachePeers()` static global that provides the appropriate API for accessing the configuration container. Please replace these `SquidConfig::peers` and `SquidConfig::npeers` members with a static container inside the `CurrentCachePeers()` function.

src/CachePeers.h
```diff
@@ -0,0 +1,61 @@
...
+    /// a remote cache hit. Never nil. \sa neighborsUdpPing()
+    /// \param iteration a 0-based index of a loop scanning all peers
+    CachePeer *nextPeerToPing(size_t iteration);
+
+private:
+    /// cache_peers in configuration/parsing order
+    Storage storage;
+
+    /// total number of completed peer scans by nextPeerToPing()-calling code
+    uint64_t peerPolls_ = 0;
```
yadij, line 49: I am not sure I understand why we need this feature right now.
 * If it can be dropped please do,
OR
 * If we need to keep it, then this should be a `static` member of `CachePeer`.

With that changed this class can disappear entirely and the global config object provided by `CurrentCachePeers()` just be a `std::vector` of a `Pointer` type.

src/CachePeers.h
```diff
@@ -0,0 +1,61 @@
...
+};
+
+/// All configured cache_peers that are still available/relevant.
+/// \returns an empty container if no cache_peers were configured or all
+/// configured cache_peers were removed (e.g., by DeleteConfigured()).
+const CachePeers &CurrentCachePeers();
+
+/// destroys the given peer after removing it from the set of configured peers
+void DeleteConfigured(CachePeer *);
```
yadij, line 58: Calls to this should be replaced by `CurrentCachePeers()->erase(Foo);`

yadij (changes requested):
Thank you for taking on this much needed cleanup.

Here is my initial review with the major design issue in the current approach. They key issue is removing the `SquidConfig` member(s). That will allow a great deal of simplification.

src/SquidConfig.h
rousskov: > You appear to have a CurrentCachePeers() static global that provides the appropriate API for accessing the configuration container

CurrentCachePeers() is a global convenience function that provides safe access to code that wants to work with a _constant_ container of still-available configured cache peers without worrying about their existence. It is essential for using C++ range-based for loops (which usually improve code quality but are not appropriate for all use cases).

> Please replace these SquidConfig::peers and SquidConfig::npeers members with a static container inside the CurrentCachePeers() function.

It is not possible to use CurrentCachePeers() to store SquidConfig::peers because CurrentCachePeers() provides read-only access. We can, of course, move SquidConfig::peers to a static variable inside CachePeers.cc and provide new functions to write that variable, but I believe all of that work should be declared outside _this_ minimal PR scope.

Removing SquidConfig::peers would create significant out-of-scope PR noise and complications. Even cf.data.pre would need to be modified, and if we start modifying that, I can imagine many serious improvements that suddenly become in-scope! This _minimal_ PR improves SquidConfig::peers _type_ (upgrading it from a raw invasive list to a proper container).  This PR should not have to _also_ change where SquidConfig::peers are stored. To keep this PR focused/minimal, any old code that can successfully use Config.peers should continue to do so in _this_ PR.

src/CachePeers.cc
rousskov: > please pass around the relevant Pointer objects instead of raw-pointers

This method should not return a pointer at all! Fixed in c01fe3a.

rousskov (commented):
@yadij, I hope I addressed all of your concerns. Please re-review.

src/carp.cc
```diff
@@ -58,7 +58,9 @@ carpInit(void)
     /* find out which peers we have */
 
     RawCachePeers rawCarpPeers;
-    for (auto p = Config.peers; p; p = p->next) {
+    for (const auto &peer: CurrentCachePeers()) {
+        const auto p = peer.get();
+
         if (!cbdataReferenceValid(p))
             continue;
```
rousskov, line 65: These two lines from the official code will be gone after #1454 is merged. We can remove them now, proactively, but that would be an out-of-scope change and it may create a merge conflict that would otherwise may be avoided.

rousskov (approved):
@eduard-bagdasaryan, thank you for resolving those merge conflicts.

rousskov:

src/CachePeers.h
yadij: > the `first_ping` variable it is effectively replacing

Aha. That was the detail I missed. Thank you for clarifying.

src/carp.cc
yadij: ... said merge has happened.

src/SquidConfig.h
yadij: Hmm, that is wrong. The type change brings an API change which has already required a change to all code lines using the config object.
The only additional change I see from this request is logic using+maintaining the now unnecessary `SquidConfig::npeers`  value.

src/neighbors.cc
```diff
@@ -551,16 +510,13 @@ neighbors_init(void)
...
 
     neighborsRegisterWithCacheManager();
 
     if (Comm::IsConnOpen(icpIncomingConn)) {
+        RawCachePeers peersToRemove;
 
-        for (thisPeer = Config.peers; thisPeer; thisPeer = next) {
-            next = thisPeer->next;
-
+        for (const auto &thisPeer: CurrentCachePeers()) {
```
yadij, line 519: The temporary `peersToRemove` list and extra work can be elided by making this for-loop iterate backwards over the STL container and removing peers as it finds them.

```suggestion
        for (const auto thisPeer = Config.peers->rbegin(); thisPeer++; thisPeer != Config.peers->rend()) {
```

src/neighbors.cc
```diff
@@ -571,17 +527,22 @@ neighbors_init(void)
...
+                peersToRemove.push_back(thisPeer.get());
+                break; // avoid warning about (and removing) the same CachePeer twice
             }
         }
+
+        while (peersToRemove.size()) {
+            const auto p = peersToRemove.back();
+            peersToRemove.pop_back();
+            DeleteConfigured(p);
+        }
```
yadij, line 539: ... with accompanying:
```suggestion
                Config.peers->erase(thisPeer);
                break; // avoid warning about (and removing) the same CachePeer twice
            }
        }
```

yadij (changes requested):

src/neighbors.cc
eduard-bagdasaryan: Removing elements from a container while iterating over it is not a good idea, since erase() [invalidates](https://en.cppreference.com/w/cpp/container/vector/erase) iterators (e.g., in vectors). And in fact, CachePeers::Storage is a vector.

src/CachePeers.h
```diff
@@ -13,8 +13,51 @@
...
+    /// deletes a previously add()ed CachePeer object
+    void remove(CachePeer *);
+
+    /// the number of currently stored (i.e. added and not removed) cache_peers
+    size_t size() const { return storage.size(); }
+
+    /* peer iterators forming a sequence for C++ range-based for loop API */
+    using const_iterator = Storage::const_iterator;
+    const_iterator begin() const { return storage.cbegin(); }
+    const_iterator end() const { return storage.cend(); }
```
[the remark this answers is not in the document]
eduard-bagdasaryan: I am not against this improvement, but I could not find existing examples with return type deduction in Squid code. Are we going to use this new style here (and in similar future cases)?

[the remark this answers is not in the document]
rousskov: > Are we going to use this new style here (and in similar future cases)?

Ideally, we should use `auto` where using it is better than not using it. Hopefully, we can avoid an endless stream of change requests asking folks to add or remove `auto` or arguing about special/tricky cases like this one.

My recommendation is to use `auto` here just to avoid arguing about it.

P.S. IIIRC, a recent PR used return type deduction; it is probably not merged yet.

src/CachePeers.h
rousskov: In summary, AFAICT, the current arguments in support of this change request are:

* The additional code duplication that the requested changes will create will be small (for now).
* The existing callers will not assert (because they are working correctly).
* The existing callers do not care about the invariant protected by the assertion.
 
There is no consensus regarding the validity of the last argument listed above.

The arguments in support of PR code include:

* The change request will lead to compiler errors. Fixing those compiler errors while providing safe constant peer iteration is not trivial, likely to lead to more change requests.
* PR code reduces code duplication.
* PR code helps alert developers and reviewers to the associated code risks.
* The assertions may detect new bugs in future callers. Official code has bugs in the same area.

Please resolve this change request without further arguments because the existing arguments, after two review iterations, are still not even close to justifying the requested changes.

A detailed response follows.

----

> > > Amos: Calls to this should be replaced by `CurrentCachePeers()->erase(Foo);`
 
> > Alex: That replacement is not possible because CurrentCachePeers() provides safe read-only access.

@yadij, please note that this change request is asking for a minor change (i.e. removing a simple wrapper function) that is not technically _possible_ (without additional work to make the changed code compile while preserving safety guarantees). Even if this change is implemented despite the lack of consensus, this change request creates non-trivial work for very little (and disputed) gain. Please consider resolving this change request without further arguments on those grounds alone.

> > However, even if we were to remove or work around that safety feature, why should those calls be replaced? I believe the current solution is significantly better:

> > * The caller does not just want to erase an item from a collection. The caller wants to destroy the CachePeer object as well. The difference can be critical and providing visual clues about it (in the form of the word "delete") is a good idea IMO.
 
> Amos: The STL API distinction between `remove` and `erase` is already well-known. We can safely abide by that without needing our home-grown wrappers and terminology.

The alleged well-known difference between STL remove() and STL erase() or our alleged ability to do "abide by that difference" is not the point here! The point is that we want the caller to know that they are deleting (including in C++ sense) the cache peer object. The proposed refactoring will not let the caller know that, even if the caller is perfectly aware of well-known facts (which _is_ a reasonable design assumption).

> > * The caller wants to delete (what the caller believes to be) a configured peer P. To avoid code that deletes wrong peers, we should assert that P is one of the configured peers. In STL containers, erase() variations that accept objects tolerate objects that are not present in the container -- those erase() methods return the number of removed elements (0 or 1). We do not want to imply that this removal has similar forgiving semantics.

> AFAICT none of our code cares (nor needs to) about whether the `CachePeer` object is part of the list or not before the action. All that matters is that it is not there afterward.

While a caller might indeed not care[^1] or, much more likely, may forget to care, this wrapper defends a useful invariant -- those who remove cache peers from the current configuration (a major, disruptive action!) specify a peer that actually exists in the current configuration. A buggy caller may leave the peer object they do want to remove while specifying some stale peer pointer. This wrapper will detect such problems. If we start hitting these assertions, we may eventually decide to relax the guarantees, but it is best to _start_ with reasonable strict guarantees that are more likely to catch bugs. Future relaxing is easier than future restricting!

[^1]: Whether existing callers do not care is debatable, but I am trying to avoid a yet another debate. 

> > * Unlike the proposed replacement, DeleteConfigured() does not result in caller code duplication.
 
> There are only two callers in current code.

Yes, the suggested code duplication will be (currently) small, but it is still duplication. Even small duplication often leads to bugs. I should not be defending the lack of duplication! It is a clear positive.


> Amos: * `neighborsUdpAck()` has a raw-pointer being used before and after the act of deletion. If that peer is not listed we have bigger issues, and do not care since it is being dropped from availability effective immediately. Adding the assert/Assure looks okay, but IMO is paranoid.

The assertion will detect some of those bigger issues if they appear. I care about that. The assertion may be a bit paranoid for a given caller fixed by this PR, but it is there for all callers, including future ones. It is not paranoid when all callers are considered. It protects a useful invariant. As discussed above, we may relax that invariant later if needed, but that possibility does not make this assertion paranoid (quite the opposite, actually!).

> * `neightbors_init()` is used by startup/reconfigure. All objects are guaranteed to be part of the list,

Reconfiguration is a long process. By the time neightbors_init() is called, some peers might have been already removed or invalidated. However, those potential complications are not actually relevant here! neightbors_init() iterates currently configured cache_peers. Naturally, it will never trigger the assertion designed to detect pointers to non-currently-configured cache_peers unless its iteration itself becomes very buggy (e.g., if it tries to delete the same peer twice). In that case, the assertion would be very useful!

src/SquidConfig.h
rousskov: > The only additional change I see from this request is logic using+maintaining the now unnecessary `SquidConfig::npeers` value.

Satisfying this change request would also require creating new code/logic to store/maintain a place where configured cache peers are stored and manipulated. That change, in turn, brings significant cache_peer parser changes into this PR scope (at least!). All that work is currently outside this PR scope. The benefits of moving that storage away from Config are also debatable, but I am trying to avoid that out-of-scope debate.

If we really want minimal PRs, we should stop insisting on changes that _can_ be considered outside a reasonable PR scope, even if there is strong consensus that those changes represent a step forward. We will improve the affected cache peer storage and parsing code in the foreseeable future -- this PR is one of dozens PRs in the project that, collectively, solves many long-term configuration problems. Please let us move forward using these small but well-scoped steps.

---

> Hmm, that is wrong. 

FWIW, I do not know what "that" is in this context, so I am addressing the other/specific comments only.

> The type change brings an API change which has already required a change to all code lines using the config object.

Even the above "all code lines" assertion is incorrect: Not all code lines using Config.peers required a change in this PR. For example, some of the lines testing whether cache peers were configured at all remained as is. There are other use cases where the type change did not matter.

src/CachePeers.h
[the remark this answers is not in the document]
eduard-bagdasaryan: Done, I also adjusted size(), just in case.

rousskov (approved):

src/neighbors.cc
yadij: One more reason for it not to be a vector. Also the vector type is why I suggested `rbegin()` instead of a normal forward or ranged-for scan.
Anyway, this is optimization, I do not insist.

src/SquidConfig.h
yadij: IMO this request is part of the policy change we agreed on in todays developer meeting. About preferring not to use globals. That was the intention behind this request.
Since we now have a formal policy about it and `SquidConfig` is a pre-existing global I will accept shifting this request to a followup PR.

yadij (approved):

yadij (approved):
My remaining issues are deferred to followup PR changes. So allowing this current implementation to go in.

src/SquidConfig.h
rousskov: Thank you for clearing this PR!

I would like to clarify one aspect in hope to reduce future conflicts:

> IMO this request is part of the policy change we agreed on in todays developer meeting. About preferring not to use globals

The "no new globals" policy does not apply to Config.peers because a non-static data member is not a "global". The policy does not even apply to Config itself because that global is grandfathered (i.e. not "new"). Thus, this and similar change requests should not be viewed as a "part of [that] policy change".

Config.peers field may stay the same or may change. If it changes, it will not be because of the "no new globals" policy. Same for the other Config data members.

Config global may change because of the intent behind the "no new globals" policy. Whether it will be eventually replaced with CurrentConfig() or just adjusted to rely on static zero-initialization of its POD data members (individual data member declaration bugs notwithstanding) remains to be seen.

PR #1425, opened 2023-07-21
Fix memory leak when reconfiguring multiline all-of ACLs

rousskov:
Normally, Acl::InnerNode::add() automatically registers stored ACL nodes
for future cleanup, but when we find the second all-of rule/line (with
the same ACL name), we do not add() the newly created OrNode and have to
explicitly register it to avoid memory leaks on reconfiguration.

Leaking since all-of ACL support was added in 2013 commit 6f58d7d.

### DIFF
--- a/src/acl/AllOf.cc
+++ b/src/acl/AllOf.cc
@@ -10,6 +10,7 @@
 #include "acl/AllOf.h"
 #include "acl/BoolOps.h"
 #include "acl/Checklist.h"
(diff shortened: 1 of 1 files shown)


yadij (approved):
Approving for the bug fix.

FTR; I still have major design objections to an `OrNode` being created by configuring an `all-of` ACL type. But that is a discussion for another time.

squidadm:
queued for backport to v6

PR #1426, opened 2023-07-24
CacheManager: require /squid-internal-mgr/ URL path prefix

rousskov:
ERROR: Squid BUG: assurance failed: tok.skip(internalMagicPrefix)
    exception location: cache_manager.cc(173) ParseUrl

Due to poor code duplication, commit 92a5adb accidentally classified
URLs without a trailing slash in the magical prefix as valid cache
manager URLs, triggering the above ERRORs. We were denying such
"slashless" cache manager URLs (as invalid internal URLs) prior to that
commit. Since that commit, the ERRORs triggered by that commit
effectively denied them as well. Denying them properly results in
simpler/smaller code (than allowing them would), so we should avoid a UI
change and continue to deny them, at least for now.

This change also reduces duplication of magic prefix definitions. Other
pending work will completely eliminate that duplication in src/ code.


src/internal.cc
```diff
@@ -86,8 +86,7 @@ internalStaticCheck(const SBuf &urlPath)
 bool
 ForSomeCacheManager(const SBuf &urlPath)
 {
-    static const SBuf mgrPfx("/squid-internal-mgr");
```
rousskov, line 89: This was the poorly duplicated prefix definition that resulted in `Squid BUG: assurance failed: tok.skip(internalMagicPrefix)` ERRORs.

src/cache_manager.cc
```diff
@@ -152,6 +152,13 @@ CacheManager::createRequestedAction(const Mgr::ActionParams &params)
     return cmd->profile->creator->create(cmd);
 }
 
+const SBuf &
+CacheManager::MagicUrlPathPrefix()
+{
+    static const SBuf magic("/squid-internal-mgr/");
+    return magic;
+}
```
rousskov, line 160: > Current PR title: Resume treating slashless cache manager URLs as invalid

What I want to say is something like "Resume treating squid-internal-mgr$ URL paths as invalid squid-internal URLs" but that is too long.

We could say "squid-internal-mgr$ URL paths are invalid squid-internal URLs" but that variant drops the fairly critical "resume treating" part -- this PR is fixing a recently introduced bug, not changing manager URL syntax or core Squid behavior.

If you have better ideas, please change this PR title.

src/tests/testCacheManager.cc
```diff
@@ -180,35 +205,45 @@ TestCacheManager::testParseUrl()
...
+                    bits.append(action);
+                    bits.append(param);
+                    bits.append(frag);
+                    mgrUrl.path(bits);
+                    mgr->testInvalidUrl(mgrUrl, "insufficient magic");
+                }
+            }
+        }
 
+        {
```
rousskov, line 234: This is just a diff reducing/clarifying addition. Please let me know if you want me to delete it (and shift the enclosed code left).

src/internal.cc
```diff
@@ -86,8 +86,7 @@ internalStaticCheck(const SBuf &urlPath)
...
-    return urlPath.startsWith(mgrPfx);
+    return urlPath.startsWith(CacheManager::MagicUrlPathPrefix());
 }
```
rousskov, line 90: > PR description: ... Denying them properly results in simpler/smaller code (than allowing them would), so we should avoid a UI change and continue to deny them, at least for now.

We have actually implemented code that would start allowing "slashless" cache manager URLs, so the above claim is not a speculation, but avoiding the UI change is the primary reason to continue to deny them in this small bug-fixing PR.

src/CacheManager.h
```diff
@@ -34,6 +34,9 @@ class CacheManager
 public:
     typedef std::vector<Mgr::ActionProfilePointer> Menu;
 
+    /// initial URL path characters that identify cache manager requests
+    static const SBuf &MagicUrlPathPrefix();
```
yadij, line 38: IMO we should use the formal terminology here. Under IETF nomenclature these are called "well-known" prefixes. Even though Squid does not use the RFC 8615 "/.well-known/" string.

yadij (changes requested):

src/cache_manager.cc
yadij: Title updated. Please ensure that the requests missing these '/' are resulting in 404 errors, not a 5xx status.

src/internal.cc
yadij: Not sure I follow what you are saying here. IF you already have a proper fix implemented why bother with a hack.

FWIW ...

1) `/squid-internal-mgr` without a trailing slash semantically means that the request reaches the mgr, but the mgr logic does not have any resource defined for that URI. So it should result in 404.

2) this PR is going into `master` branch. Avoiding UI change is not a consideration for code going there - just add appropriate documentation in the PR.

src/tests/testCacheManager.cc
yadij: yes, drop this indentation please.

src/CacheManager.h
rousskov: Done in commit 4a3ec07.

src/tests/testCacheManager.cc
rousskov: Done in commit f8efc2c.

src/internal.cc
rousskov: >> We have actually implemented code that would start allowing "slashless" cache manager URLs

> IF you already have a proper fix implemented why bother with a hack

This PR does not contain a hack. It contains a proper bug fix for a recently introduced bug.

Properly handling slashless URLs (after they are allowed) requires a lot more than just "allowing" slashless URLs. However, just allowing that URL alone already requires quite a bit of code changes. We know the latter because we have implemented those changes. The changes we implemented are not enough for a "proper fix", but any "proper fix" will contain similar changes (and a lot more).

>> avoiding the UI change is the primary reason

> Avoiding UI change is not a consideration for code going [into `master` branch].

It is a consideration for PR scope. This is a minimal PR fixing a very recent bug by restoring very old pre-bug behavior. Nothing else. We could do a lot more, but we do not have to. The current minimal scope is reasonable.

rousskov:
> yadij changed the title "Resume treating slashless cache manager URLs as invalid" to "CacheManager: require path-abolute URLs"

@yadij, please note that the bug and this PR are not about path-absolute URLs:

```
* /squid-internal-mgr/ -- old prefix (used since the prefix was added in 2011 commit e37bd29)
* /squid-internal-mgr  -- buggy prefix accidentally added (to _some_ contexts) by recent commit 92a5adb
* /squid-internal-mgr/ -- old prefix restored by this PR
```

yadij (approved):

rousskov:
@yadij, please note that the "require path-abolute URLs" part of your PR title does not match this PR code (as detailed in https://github.com/squid-cache/squid/pull/1426#issuecomment-1648903892) and contains an "abolute" typo. I am not going to touch that title to avoid derailing or delaying this approved PR, but please consider changing that title if you think it should be changed.

rousskov:
@yadij, just a reminder about this ready-to-merge PR (approved by you and with passed staging checks!) waiting for your [action](https://github.com/squid-cache/squid/pull/1426#issuecomment-1654217590).

rousskov:
> @yadij, just a reminder about this ready-to-merge PR (approved by you and with passed staging checks!) waiting for your [action](https://github.com/squid-cache/squid/pull/1426#issuecomment-1654217590).

I timed out and fixed the title myself. I hope I found a correct and non-controversial variation.

yadij:
manual backport

PR #1428, opened 2023-07-27
Bug 4981: Work around in-call job invalidation bugs

ShmayaFrankel:
Bug 4981 is one known case of such invalidation, but this workaround is
much broader than that bug context. We can speculate that architectural
problems described in commit e3b6f15 are behind (some of) these bugs.

### DIFF
--- a/src/base/AsyncJobCalls.h
+++ b/src/base/AsyncJobCalls.h
@@ -179,7 +179,7 @@ JobDialer<Job>::dial(AsyncCall &call)
         if (!job) {
             debugs(call.debugSection, DBG_CRITICAL, "ERROR: Squid BUG: Job invalidated during " <<
                    call.name << " that threw exception: " << e.what());
-            return; // See also: Bug 4981, commit e3b6f15, and XXX in Http::Stream class description.
+            return; // see also: bug 4981, commit e3b6f15, and XXX in Http::Stream class description
         }
         job->callException(e);
     }
--- a/src/base/AsyncJobCalls.h
+++ b/src/base/AsyncJobCalls.h
@@ -176,16 +176,14 @@ JobDialer<Job>::dial(AsyncCall &call)
     } catch (const std::exception &e) {
         debugs(call.debugSection, 3,
                call.name << " threw exception: " << e.what());
-        if (!job.valid())
-        {
+        if (!job.valid()) {
             debugs(call.debugSection, DBG_CRITICAL,
                    "BUG: job invalidated during " << call.name << " that threw exception: " << e.what());
             return; // See also: Bug 4981, commit e3b6f15, and XXX in Http::Stream class description.
         }
         job->callException(e);
     }
-    if (!job.valid())
-    {
+    if (!job.valid()) {
         debugs(call.debugSection, DBG_CRITICAL,
                "BUG 4981: job invalidated during " << call.name);
         return;
--- a/src/base/AsyncJobCalls.h
+++ b/src/base/AsyncJobCalls.h
@@ -177,7 +177,7 @@ JobDialer<Job>::dial(AsyncCall &call)
         debugs(call.debugSection, 3,
                call.name << " threw exception: " << e.what());
         if (!job) {
-            debugs(call.debugSection, 3, "ERROR: Squid BUG: Job invalidated during " <<
+            debugs(call.debugSection, DBG_CRITICAL, "ERROR: Squid BUG: Job invalidated during " <<
                    call.name << " that threw exception: " << e.what());
             return; // see also: bug 4981, commit e3b6f15, and XXX in Http::Stream class description
         }
@@ -185,7 +185,7 @@ JobDialer<Job>::dial(AsyncCall &call)
     }
 
     if (!job) {
-        debugs(call.debugSection, 3, "ERROR: Squid BUG: Job invalidated during " << call.name);
+        debugs(call.debugSection, DBG_CRITICAL, "ERROR: Squid BUG: Job invalidated during " << call.name);
         return;
     }
     job->callEnd(); // may delete job
--- a/src/base/AsyncJobCalls.h
+++ b/src/base/AsyncJobCalls.h
@@ -22,7 +22,7 @@
  * below implement for us, calling the job's method with the right params.
  */
 template <class Job>
-class JobDialer : public CallDialer
+class JobDialer: public CallDialer
 {
 public:
     typedef Job DestClass;
@@ -41,7 +41,7 @@ protected:
 
 private:
     // not implemented and should not be needed
-    JobDialer &operator=(const JobDialer &);
+    JobDialer &operator =(const JobDialer &);
 };
 
 /// schedule an async job call using a dialer; use CallJobHere macros instead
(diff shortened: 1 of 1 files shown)


rousskov (commented):
@ShmayaFrankel, I pushed a few polishing commits into your PR branch[^1] and updated PR title/description to match Squid commit message format. Please feel free to undo or adjust any changes -- this is your PR.

@yadij, @kinkie, I am not against this crude workaround going in because I am not aware of anybody working on fixing the old underlying problem. I am not sure disabling the corresponding assertions is the lesser evil here, especially since we do not see a lot of complaints about the bug -- keeping fairly critical assertions alive and asking a few affected admins to either patch their Squids or [facilitate](https://wiki.squid-cache.org/SquidFaq/AboutSquid#how-to-add-a-new-squid-feature-enhance-of-fix-something) a proper fix might be the lesser evil. I certainly do not insist on this change.

[^1]: At the time of writing, GitHub does not reflect those 5 new commits in this PR, but does show them in your branch [summary](https://github.com/squid-cache/squid/compare/master...ShmayaFrankel:squid:4981-workaround). This is probably a temporary GitHub glitch -- I cannot explain it. I expect the PR state to be automatically synced/fixed in the foreseeable future. This PR will show (at least) 7 commits when that happens. After that happens, I will come back to enable CI tests.

src/base/AsyncJobCalls.h
```diff
@@ -176,11 +176,21 @@ JobDialer<Job>::dial(AsyncCall &call)
     } catch (const std::exception &e) {
         debugs(call.debugSection, 3,
                call.name << " threw exception: " << e.what());
+        if (!job.valid())
+        {
```
yadij, line 180:
```suggestion
        if (!job.valid()) {
```

src/base/AsyncJobCalls.h
```diff
@@ -176,11 +176,21 @@ JobDialer<Job>::dial(AsyncCall &call)
...
+            debugs(call.debugSection, DBG_CRITICAL,
+                   "BUG: job invalidated during " << call.name << " that threw exception: " << e.what());
+            return; // See also: Bug 4981, commit e3b6f15, and XXX in Http::Stream class description.
+        }
         job->callException(e);
     }
-
+    if (!job.valid())
+    {
```
yadij, line 188:
```suggestion
    if (!job.valid()) {
```

yadij (changes requested):
just some polish needed

rousskov:
> just some polish needed

@yadij, that polish has been done already. See the footnote in https://github.com/squid-cache/squid/pull/1428#pullrequestreview-1550320037

rousskov:
> I expect the PR state to be automatically synced/fixed in the foreseeable future.

I timed out and pushed a trivial commit [7673c49](https://github.com/squid-cache/squid/pull/1428/commits/7673c496d1faee8cfa93e7369530f36b0e0bda3c). It did wake GitHub up as far as this PR state is concerned, so we are back to regular/expected PR state, with all earlier PR branch commits visible on PR pages.

rousskov:
OK to test

ShmayaFrankel:
Once approved, please also tag for v6 backport

ShmayaFrankel:
@yadij could please review after @rousskov commits?

src/base/AsyncJobCalls.h
```diff
@@ -176,9 +176,18 @@ JobDialer<Job>::dial(AsyncCall &call)
...
+        if (!job) {
+            debugs(call.debugSection, DBG_CRITICAL, "ERROR: Squid BUG: Job invalidated during " <<
+                   call.name << " that threw exception: " << e.what());
+            return; // see also: bug 4981, commit e3b6f15, and XXX in Http::Stream class description
+        }
         job->callException(e);
     }
 
+    if (!job) {
+        debugs(call.debugSection, DBG_CRITICAL, "ERROR: Squid BUG 4981: Job invalidated during " << call.name);
```
yadij, line 188: Please do not put bug numbers into the logs. This tends towards an ongoing series of user reports when distros LTS continue to ship buggy Squid versions long after we fix it. Use a code comment instead.

src/base/AsyncJobCalls.h
```diff
@@ -176,9 +176,18 @@ JobDialer<Job>::dial(AsyncCall &call)
...
+        if (!job) {
+            debugs(call.debugSection, 3, "ERROR: Squid BUG: Job invalidated during " <<
+                   call.name << " that threw exception: " << e.what());
+            return; // see also: bug 4981, commit e3b6f15, and XXX in Http::Stream class description
+        }
         job->callException(e);
     }
 
+    if (!job) {
+        debugs(call.debugSection, 3, "ERROR: Squid BUG: Job invalidated during " << call.name);
```
rousskov, line 188: > Amos: Are you sure that DBG_CRITICAL is the right level for these?

These checks are essentially failed assertions. Assertion failures should be reported at the highest level because they indicate that Squid may be violating all sorts of guarantees and invariants going forward.

I will restore DBG_CRITICAL (instead of the current level 3) after posting this review, but I am not going to block this PR if it ends up using DBG_IMPORTANT instead.

> If there is nothing the admin can do to avoid it perhapse level-1 would be better.

IIRC, we do not have consensus about the (difference in) meaning of debugging levels 0 and 1. Until that consensus is established, "better" is essentially undefined or, to be more precise, has many conflicting definitions. I am happy to cooperate towards establishing that consensus, but I assume that you do not want to use _this_ PR to do so.

FWIW, I disagree with all of the following statements, listed here  _without_ an implication that you are making or implying these statements:

* Level-0 messages are for things admins can prevent
* Level-1 messages are for things admins cannot prevent
* Assertions (and equivalent) are things admins cannot prevent
* Concerns about debugging noise is what determines whether level 0 or level 1 should be used.

... where "prevent" is short for "can do something to avoid in the future".

The correct way to control debugging noise[^1] among level-0/1 cache.log messages is [cache_log_message](http://www.squid-cache.org/Doc/config/cache_log_message/).

[^1]: as defined by the admin

src/base/AsyncJobCalls.h
rousskov: > Please do not put bug numbers into the logs. This tends towards an ongoing series of user reports when distros LTS continue to ship buggy Squid versions long after we fix it. Use a code comment instead.

I strongly disagree with that recommendation. IMO, bug reports about fixed bugs is a much smaller problem than a "BUG" message without attribution, both for admins and the Project. And that smaller problem can be reduced further by other means, such as improving Bugzilla rendering of closed bug reports.

Where applicable, we have been putting bug numbers into these BUG lines for a very long time and should continue to do so IMO. At the very least, until we develop consensus about this policy change, we should not ask authors to change (correct) bug attributions.

ShmayaFrankel:
@yadij @rousskov should i restore the bug number?

src/base/AsyncJobCalls.h
```diff
@@ -176,9 +176,18 @@ JobDialer<Job>::dial(AsyncCall &call)
     } catch (const std::exception &e) {
         debugs(call.debugSection, 3,
                call.name << " threw exception: " << e.what());
+        if (!job) {
+            debugs(call.debugSection, DBG_CRITICAL, "ERROR: Squid BUG: Job invalidated during " <<
+                   call.name << " that threw exception: " << e.what());
```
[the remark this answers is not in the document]
yadij: You seem to have forgotten bug 3279. Each and every side effect got reported individually as "The Bug" Just because the number got placed in the logs. I was wading through long debug traces that turned out to be third-party software issues for 8 years.

These exception handlers are even more likely to be hit due to various bugs (unrelated to each other) elsewhere in the code. Resulting in the same hell all over again, but worse.

yadij:
> @yadij @rousskov should i restore the bug number?

NO please.

FTR, @rousskov has answered the one question I had. IMO this is good to go in now as-is.

yadij (approved):

src/base/AsyncJobCalls.h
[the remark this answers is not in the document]
rousskov: > You seem to have forgotten ...

FWIW, I recommend finding ways to express ideas without negative remarks about collaborators.

> You seem to have forgotten bug 3279. Each and every side effect got reported individually as "The Bug" Just because the number got placed in the logs.

I disagree with the "because the number got placed in the logs" assertion. There are other, IMO much more accurate, explanations for those reports.
 
> These exception handlers are even more likely to be hit due to various bugs (unrelated to each other) elsewhere in the code. Resulting in the same hell all over again, but worse.

I see no connection between "using a bug number in debugs()" and the alleged (but unspecified) hellish effects.

If anything, the number of reports if we use a specific bug number is likely to be _smaller_ than the number of reports if we do not use a specific bug number because, in the latter case, fewer admins would know that this bug is already known to us and, hence, more admins will report them.

AFAICT, the effort to address those reports should not depend in any significant way on whether there is a bug number on that debugs() line.

I consider reports (collected under that bug number) a small but positive step towards fixing the underlying Squid bug(s).

squidadm:
queued for backport to v6

rousskov:
@ShmayaFrankel, if you are _not_ happy with how our script will record your commit-based name and email into Squid CONTRIBUTORS file, please let me know, and we will do a manual override. You can see that script results at the bottom of the "Check source-maintenance" diff inside the "Run ./test-suite/test-sources.sh" section in [CI console log](https://github.com/squid-cache/squid/actions/runs/5842160575/job/15842983683).

ShmayaFrankel:
> @ShmayaFrankel, if you are _not_ happy with how our script will record your commit-based name and email into Squid CONTRIBUTORS file, please let me know, and we will do a manual override. You can see that script results at the bottom of the "Check source-maintenance" diff inside the "Run ./test-suite/test-sources.sh" section in [CI console log](https://github.com/squid-cache/squid/actions/runs/5842160575/job/15842983683).

@rousskov please use:
Shmaya Frankel(shmaya.frankel@gmail.com)
Thanks!

rousskov:
@ShmayaFrankel, done at https://github.com/squid-cache/squid/pull/1476/commits/949afca82500bed404067f756bb1af4bbbb89c95.

PR #1429, opened 2023-07-27
Bug 5290: pure virtual call in Ftp::Client constructor

rousskov:
FATAL: Dying from an exception handling failure;
    exception: [no active exception]

Converting `this` to CbcPointer in a constructor of an abstract class
like Ftp::Client does not work because our virtual toCbdata() method
remains pure until the final/child class constructor runs.

Conceptually, the bug was probably introduced in 2013 commit 434a79b,
when FTP class hierarchy grew, making Ftp::Client an abstract class, but
the trigger was recent commit 337b9aa that removed CBDATA_CLASS() from
Ftp::Client class declaration. We discovered, described, and addressed
several such bugs in that commit, but we missed this case.

### DIFF
--- a/src/clients/FtpClient.cc
+++ b/src/clients/FtpClient.cc
@@ -196,16 +196,10 @@ Ftp::Client::Client(FwdState *fwdState):
 
     ctrl.last_command = xstrdup("Connect to server");
 
-    // and hope that the caller calls watchForCtrlClosure()
-}
-
-void
-Ftp::Client::watchForCtrlClosure()
-{
     typedef CommCbMemFunT<Client, CommCloseCbParams> Dialer;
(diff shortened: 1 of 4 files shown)


rousskov:
Disclaimer: I have not tested this fix. I have [asked](https://bugs.squid-cache.org/show_bug.cgi?id=5290#c2) the bug reporter to test it.

rousskov (commented):
Status update: We have one [confirmation](http://lists.squid-cache.org/pipermail/squid-users/2023-July/026001.html) that this change fixes bug 5290.

yadij (approved):

squidadm:
queued for backport to v6

PR #1430, opened 2023-07-28
Cover OnTerminate() calls unrelated to exception handling

rousskov:
The C++ standard lists many reasons[^1] for calling std::terminate().
All of them deal with an "exception handling failure". However, when
Squid bugs lead to an undefined behavior (e.g., by calling a pure
virtual function), some compilers also call std::terminate(). In those
cases, there may be no active exception, and our std::terminate()
handler's report about an "exception handling failure" (with "no active
exception") was confusing and misleading.

Also do not describe an exception when there is none. Callers that know
that the exception exists may continue to use CurrentException(),
especially if they do not need to report that exception on a dedicated
debugs() line. Other callers, including OnTerminate(), should use the
new CurrentExceptionExtra() manipulator.

[^1]: https://en.cppreference.com/w/cpp/error/terminate

### DIFF
--- a/src/base/TextException.cc
+++ b/src/base/TextException.cc
@@ -65,25 +65,41 @@ operator <<(std::ostream &os, const TextException &ex)
     return os;
 }
 
+/// prints the current exception (which presence has been verified by the caller)
+static std::ostream &
+CurrentException_(std::ostream &os)
+{
+    try {
+        throw; // re-throw to recognize the exception type
+    }
+    catch (const TextException &ex) {
+        os << ex; // optimization: this is a lot cheaper than what() below
+    }
+    catch (const std::exception &ex) {
+        os << ex.what();
+    }
+    catch (...) {
+        os << "[unknown exception type]";
+    }
+    return os;
+}
+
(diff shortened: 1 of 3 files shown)


rousskov:
```
2023/07/27 07:20:48| FATAL: Dying from an exception handling failure; exception: [no active exception]
```

This improvement was triggered by bug [5290](https://bugs.squid-cache.org/show_bug.cgi?id=5290). Bug 5290 is not related to C++ exceptions and their handling, but that bug report title was even changed to contain the word "exception" at some point, matching misleading Squid cache.log output:

```
           What    |Removed                     |Added
----------------------------------------------------------------------------
            Summary|ftp related crash in        |pure virtual call exception
                   |6.1-20230714-rdde7b8e63     |in Ftp::StartGateway
```

This PR does not fix bug 5290 (see #1429 for that fix).

This PR adjusts OnTerminate() output to avoid misleading statements and noise. For example, when a pure virtual function is called, the new output is:

```
2023/07/28 10:08:17| FATAL: Dying after an undetermined failure
```

PR #1432, opened 2023-07-29
ESI: Fix build [-Wsingle-bit-bitfield-constant-conversion]

kinkie:
clang 16, the default on current fedora rawhide and centos stream 9,
complains about several ESI places:

    error: implicit truncation from 'int' to a one-bit wide bit-field
    changes value from 1 to -1
    [-Werror,-Wsingle-bit-bitfield-constant-conversion]

Turn 1-bit flags in ESI data structures from ints to unsigned ints to
avoid the problem.

### DIFF
--- a/src/esi/Context.h
+++ b/src/esi/Context.h
@@ -75,20 +75,20 @@ public:
     ClientHttpRequest *http;
 
     struct {
-        int passthrough:1;
-        int oktosend:1;
-        int finished:1;
+        unsigned int passthrough:1;
+        unsigned int oktosend:1;
+        unsigned int finished:1;
 
         /* an error has occurred, send full body replies
          * regardless. Note that we don't fail midstream
          * because we buffer until we can not fail
          */
-        int error:1;
+        unsigned int error:1;
 
-        int finishedtemplate:1; /* we've read the entire template */
-        int clientwantsdata:1; /* we need to satisfy a read request */
-        int kicked:1; /* note on reentering the kick routine */
-        int detached:1; /* our downstream has detached */
+        unsigned int finishedtemplate:1; /* we've read the entire template */
+        unsigned int clientwantsdata:1; /* we need to satisfy a read request */
+        unsigned int kicked:1; /* note on reentering the kick routine */
+        unsigned int detached:1; /* our downstream has detached */
     } flags;
 
     err_type errorpage; /* if we error what page to use */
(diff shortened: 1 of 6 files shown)


yadij (approved):
While this is technically okay, I question whether we can avoid using 1-bit flags entirely. Perhapse converting these flags to `bool` would be better?

kinkie:
> While this is technically okay, I question whether we can avoid using 1-bit flags entirely. Perhapse converting these flags to `bool` would be better?

I thought about it, this is the least invasive change that would fix the issue.
Converting to bool would also mean converting all the literals

rousskov (approved):
AFAICT, this PR removes the last set of `signed int` single-bit fields from Squid. Thank you, Francesco!

FWIW, here are the remaining definitely signed (the first two) and possibly signed (four enums) bit fields that I could find. Improving that particular code will be difficult.

```
src/Store.h:    sfileno swap_filen:25;
src/Store.h:    sdirno swap_dirn:7;

src/Store.h:    mem_status_t mem_status:3;
src/Store.h:    ping_status_t ping_status:3;
src/Store.h:    store_status_t store_status:3;
src/Store.h:    swap_status_t swap_status:3;
```

> While this is technically okay, I question whether we can avoid using 1-bit flags entirely. Perhapse converting these flags to `bool` would be better?

Yes, it would be. If somebody works on that, I recommend converting all such fields (that can be converted to a simple bool) in one PR. I agree with Francesco that such a conversion deserves a dedicated (and more complex/noisy) PR.

rousskov:
@kinkie, I adjusted PR title to be more precise and to-the-point[^1]. I adjusted PR description to look better and address M-failed-description. I do not insist on any of those changes. Please check, adjust further as needed, and clear for merging.

[^1]: The conversion was not really "broken" AFAICT -- the _boolean_ value of those flags was correct despite that warning. If I am wrong, then this PR does not just fix the clang build. It actually fixes many ESI bugs, changing ESI behavior! With all the problems in ESI code, such widespread breakage seems unlikely to me.

kinkie:
> 1. If I am wrong, then this PR does not just fix the clang build. It actually fixes many ESI bugs, changing ESI behavior! Even with all the problems in ESI code, such widespread breakage seems unlikely to me.

I think the underlying code works practice, based on the assumption that C(++) treats 0 as "false-y" and non-zero as "true-y". But this change is more explicit about these bits being bools and signals intent better. bool would be even better, and non-bit fields would probably trade core use for speed. It'll be for some other time.

Thanks for the quick review; this unblocks upgrading the docker containers in the CI farm

squidadm:
queued for backport to v6

PR #1437, opened 2023-07-30
Remove Splay::walk() method

rousskov:
Splay template class provided two traversal APIs: walk() and visit().
Primary code is only using visit(). Our tests were only testing walk().


rousskov (commented):
I hope this PR will help simplify and land #1431, but it is a step forward on its own.

A lot in the ancient Splay testing code can be improved, but doing so well requires changing a lot of code unrelated to this minimal PR, and I resisted the temptation.

yadij (approved):
Thank you for removing this code bloat.

PR #1438, opened 2023-07-31
Fix: Compilation error due to possible string over-read in src/acl/BoolOps.cc

oguzhanmeteozturk:
While attempting to build the Squid package version 6.1-2 on a fresh Ubuntu Jammy environment, I encountered a compilation error in the src/acl/BoolOps.cc file. The issue is related to a potential string over-read as indicated by the following error message:

```bash
__builtin_strlen' reading 1 or more bytes from a region of size 0 [-Werror=stringop-overread]
Must(strlen(acl->name) <= sizeof(name)-2)
```

My build environment is Ubuntu 22.04 Jammy, with gcc-12. The sources were obtained directly from http://http.debian.net/debian/pool/main/s/squid and the package was built using 

```bash
dpkg-buildpackage -rfakeroot -b -us -uc.
```

I am not familiar with the codebase and my understanding of C++ is limited. However, to address the error, I tried to adjust the assert line in the Acl::NotNode::NotNode(ACL *acl) function as follows:

```c++
// From
assert(acl);
// To
assert(acl && acl->name && *acl->name);
```

I attempted to ensure that the strlen(acl->name) is not called when acl->name is a null pointer or an empty string, hopefully preventing the over-read error. The change seemed to resolve the compilation error in my local environment.

However, as I am not deeply familiar with the project, I would greatly appreciate any feedback or further guidance on whether this is an appropriate fix and the potential ramifications of this change.

Thank you for your time and consideration.

### DIFF
--- a/src/acl/BoolOps.cc
+++ b/src/acl/BoolOps.cc
@@ -16,7 +16,7 @@
 
 Acl::NotNode::NotNode(ACL *acl)
 {
-    assert(acl);
+    assert(acl && acl->name && *acl->name);
     Must(strlen(acl->name) <= sizeof(name)-2);
     name[0] = '!';
     name[1] = '\0';


kinkie:
Makes sense, thanks for this contribution!

kinkie:
Ok to test

src/acl/BoolOps.cc
```diff
@@ -16,7 +16,7 @@
 
 Acl::NotNode::NotNode(ACL *acl)
 {
-    assert(acl);
+    assert(acl && acl->name && *acl->name);
```
rousskov, line 19: Thank you for finding a workaround for this compilation error!

> I attempted to ensure that the strlen(acl->name) is not called when acl->name is a null pointer or an empty string, hopefully preventing the over-read error. 

`acl->name` cannot be a null pointer because Acl::name is a fixed-size array (rather than a c-string pointer). Here is its declaration in Acl.h:

```C++
    char name[ACL_NAME_SZ];
```

Thus, I assume that it is the non-empty condition that silences the warning in your environment: `*acl->name`.

However, the warning still does not make sense to me because there is nothing wrong with calling strlen() on an empty string! In other words, there is no "overread" even if `acl->name` is empty.

This constructor code will work "correctly" with an empty name, so I would prefer to see a very different solution, but since I cannot offer (a path to) one until ACL::name is replaced with SBuf, I am OK with merging this variant (if it works):

```suggestion
    assert(acl);
    assert(*acl->name); // suppresses -Wstringop-overread in strlen() below
```

I [failed](https://godbolt.org/z/fxeE75oGo) to reproduce this warning. Does the above variant work in your environment?

If the above suggestion does not suppress that warning, please try this variant:

```suggestion
    assert(acl);
    assert(acl->name); // suppresses -Wstringop-overread in strlen() below
```


> The change seemed to resolve the compilation error in my local environment.

FWIW, the warning smells like a false positive to me -- a GCC bug.  There are dozens of similar bugs in GCC, including GCC v12. They have even created a meta bug for them at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97048.

We will also need to fix PR [title](https://github.com/measurement-factory/anubis/blob/master/README.md#pr-title) and [description](https://github.com/measurement-factory/anubis/blob/master/README.md#pr-description) (i.e. future official [commit message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message)), but let's finalize the code first.

rousskov (changes requested):

PR #1439, opened 2023-08-01
Source Format Enforcement

rousskov:
This change is a reference point for automated CONTRIBUTORS updates.

### DIFF
--- a/src/tests/testCacheManager.cc
+++ b/src/tests/testCacheManager.cc
@@ -54,7 +54,7 @@ CacheManagerInternals::testValidUrl(const AnyP::Uri &url)
         (void)ParseUrl(url);
     } catch (...) {
         std::cerr << "\nFAIL: " << url <<
-            Debug::Extra << "error: " << CurrentException << "\n";
+                  Debug::Extra << "error: " << CurrentException << "\n";
         CPPUNIT_FAIL("rejected a valid URL");
     }
 }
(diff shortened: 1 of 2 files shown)


src/tests/testCacheManager.cc
```diff
@@ -54,7 +54,7 @@ CacheManagerInternals::testValidUrl(const AnyP::Uri &url)
         (void)ParseUrl(url);
     } catch (...) {
         std::cerr << "\nFAIL: " << url <<
-            Debug::Extra << "error: " << CurrentException << "\n";
```
kinkie, line 57: Doesn't seem related to the PR

src/tests/testCacheManager.cc
```diff
@@ -65,7 +65,7 @@ CacheManagerInternals::testInvalidUrl(const AnyP::Uri &url, const char *const pr
     try {
         (void)ParseUrl(url);
         std::cerr << "\nFAIL: " << url <<
-            Debug::Extra << "error: should be rejected due to '" << problem << "'\n";
```
kinkie, line 68: Doesn't seem related to the PR

kinkie (changes requested):
Seems there's spurious changes

src/tests/testCacheManager.cc
rousskov: This "Source Format Enforcement" PR (as all "Source Format Enforcement" PRs?) is the result of executing our `./scripts/source-maintenance.sh` script. That script may format sources and update CONTRIBUTORS. Both adjustment happened in this PR. Both are in-scope. Does this summary address your concerns?

kinkie (approved):
Aha yes. It's the first comment that threw me off

rousskov:
> It's the first comment that threw me off

If you recall, that or a similar comment is required to advance the "manually audited version" of the CONTRIBUTORS file. See collectAuthors() description in scripts/source-maintenance.sh or `git grep -10 vettedCommitPhraseRegex`.

PR #1440, opened 2023-08-01
CI: More HTTP caching and revalidation tests

rousskov:
Use Daft cache-response test to monitor for bugs in basic caching code,
including bugs like the one fixed by commit c203754. Unlike the old
proxy-collapsed-forwarding test that uses concurrent requests, this test
varies basic response properties.

Use Daft accumulate-headers-after-304 test to monitor for bugs like the
one fixed by commit 55e1c6e. Unlike old proxy-update-headers-after-304,
this test focuses on certain _problematic_ HTTP 304 header updates.

### DIFF
--- a/test-suite/test-functionality.sh
+++ b/test-suite/test-functionality.sh
@@ -215,7 +215,9 @@ main() {
         local default_tests="
             pconn
             proxy-update-headers-after-304
+            accumulate-headers-after-304
             upgrade-protocols
+            cache-response
             proxy-collapsed-forwarding
             busy-restart
             truncated-responses


kinkie (approved):
It's not entirely clear to me how this gets integrated, but it won't hurt

rousskov:
> It's not entirely clear to me how this gets integrated

As you know, GitHub Actions CI is driven by our `.github/workflows/default.yaml` configuration. The `functionality-tests` section in that configuration prepares the virtual environment and starts our `test-suite/test-functionality.sh` script. That script downloads and runs named tests.

If one installs node.js and follows environment conventions similar to those on GitHub Actions CI virtual nodes, one can run the same `test-suite/test-functionality.sh` script locally, against their own Squid, without using GitHub Actions CI.

kinkie:
Got you. I was missing the part where tests get downloaded. Thanks!

PR #1444, opened 2023-08-06
v6 Next Backports

squidadm:


yadij:
OK to test

yadij (approved):

configure.ac
```diff
@@ -5,7 +5,7 @@
 ## Please see the COPYING and CONTRIBUTORS files for details.
 ##
 
-AC_INIT([Squid Web Proxy],[6.1-VCS],[http://bugs.squid-cache.org/],[squid])
+AC_INIT([Squid Web Proxy],[6.2-VCS],[https://bugs.squid-cache.org/],[squid])
```
rousskov, line 8: Should not the URI scheme change land in master first?

yadij: Ideally yes, though its a bit late to get it in this release. I have a followup coming shortly for this and another out-of-sync bit.

PR #1449, opened 2023-08-09
Maintenance: use SBuf instead of SquidString in CommonPool

kinkie:


src/CommonPool.h
```diff
@@ -28,11 +28,11 @@ class CommonPool
 
 public:
     static CommonPool *Factory (unsigned char _class, CompositePoolNode::Pointer&);
-    char const* theClassTypeLabel() const {return typeLabel.termedBuf();}
+    const SBuf &theClassTypeLabel() const { return typeLabel; }
```
rousskov, line 31: If you have the time, please rename this method to classTypeLabel() -- the old name goes against our current naming style, and all callers are already affected by the type change. Otherwise, this can be merged as is because that renaming can be viewed as out of scope.

rousskov (approved):
Thank you for this improvement. I pushed a commit to prevent compiler warnings about meaningless "const" in the return result (and to avoid unnecessary SBuf copies).

src/CommonPool.h
kinkie: Sure.
Update coming up

rousskov (approved):

PR #1450, opened 2023-08-09
Use SBuf in CompositeSelectionDetails and DelayTaggedBucket

kinkie:


src/DelayTagged.h
```diff
@@ -30,10 +31,10 @@ class DelayTaggedBucket : public RefCountable
     typedef RefCount<DelayTaggedBucket> Pointer;
 
     void stats(StoreEntry *)const;
-    DelayTaggedBucket(String &aTag);
+    DelayTaggedBucket(const SBuf &aTag);
     ~DelayTaggedBucket() override;
     DelayBucket theBucket;
-    String tag;
+    SBuf tag;
```
rousskov, line 37: From compiler point of view, we can make this (and possibly the other new SBuf field) `const` AFAICT, but I did _not_ do it because I am not sure whether this tag is meant to be preserved throughout DelayTaggedBucket lifetime and because one could consider such a change out of this PR scope. @kinkie, your call.

rousskov (approved):
@kinkie, thank you for removing another pair of Strings!

Per our "Easier Done" experiment, I fixed one `#include` and polished formatting of branch-modified sources, including applying automated include header order fixes. I also removed one RefCount copy in the modified code and made the PR title more specific. Please review and undo/adjust as needed, as usual.

src/DelayTagged.h
kinkie: It can work as const. I'll request a re-review soon

rousskov: ... and please check that other new SBuf field as well if you decide to go down this path.

kinkie: CompositeSelectionDetails seems to be never modified after construction.
Moving the data setter in the constructor, I could make both always-present fields const

rousskov (approved):

PR #1451, opened 2023-08-09
Maintenance: use SBufs for ErrorDetailEntry data fields

kinkie:
Also fixed (re)configuration crashes when an error-details.txt entry is
missing a required "detail" or "descr" field.


src/HttpHeaderTools.cc
```diff
@@ -233,6 +234,19 @@ httpHeaderParseQuotedString(const char *start, const int len, String *val)
     return 1;
 }
 
+/**
+ * Temporary bridge for the String-based variant to support conversion
+ * from SquidString to SBuf
+ * TODO: make this the primary implementation
+ */
+int httpHeaderParseQuotedString(const char *start, const int len, SBuf &val)
```
rousskov, line 242: Most likely, the long-term "primary" function should have a different API. For example, I would expect it return an SBuf (and an exception thrown on errors, as usual). The closer you can get to that correct long-term interface in this PR, when you are adding the very first callers for that function, the better! Otherwise, we would be creating new technical debt (i.e. wrong HTTP quoted string parsing API) while trying to get rid of the old technical debt (i.e. String).

src/ssl/ErrorDetailManager.cc
```diff
@@ -61,7 +62,8 @@ Ssl::ErrorDetailsList::getErrorDescr(Security::ErrorCode value)
 {
     const ErrorDetails::const_iterator it = theList.find(value);
     if (it != theList.end()) {
-        return it->second.descr.termedBuf();
+        SBuf rv(it->second.descr);
+        return rv.c_str();
```
rousskov, line 66: Returning a pointer to a temporary local variable. Please address all such bugs.

rousskov (changes requested):
Thank you for starting this conversion. It is more difficult than it may seem. I left a few comments in hope to help you make progress, but please do not overlook PR build failures because, I suspect, they may require non-trivial adjustments as well.

src/HttpHeaderTools.cc
```diff
@@ -233,6 +234,19 @@ httpHeaderParseQuotedString(const char *start, const int len, String *val)
...
+{
+    String s;
+    auto rv = httpHeaderParseQuotedString(start, len, &s);
+    val = StringToSBuf(s);
```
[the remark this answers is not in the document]
kinkie: Doing what you suggest would balloon the scope of the PR.

I would prefer an approach where
1. data members and simple interfaces get ported. There are 58 files including SquidString right now
2. once they are ported, get rid of the bridge methods like this one; one PR per method, building unit tests first to ensure non regressions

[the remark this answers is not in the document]
rousskov: > Doing what you suggest would balloon the scope of the PR.

The above assertion is false because suggested option 1 keeps the scope of this PR exactly the same.

> I would prefer an approach where [we first introduce bridge functions and then get rid of them].

I am not against bridge functions per se, and suggested option 1 arguably continues to use a bridge function. A bridge function is the best way forward in many cases, but each suggested conversion should be evaluated on its own. Attempts to confine all future conversions to one conversion technique (e.g., bridge functions) seem counter productive to me because Squid code is too varied for one technique to be the right choice in all conversion cases.

As you know, I support minimal conversion PRs and even prefer those whenever they are feasible. Needless to say, there are many ways to convert and even many ways to create bridge functions. Naturally, there are certain principles or invariants we should observe _regardless_ of the PR size or the overall conversion approach. This change request is about violations of some of those principles: 

* avoid performance regressions (option 2 avoids such a regression but requires more work)
* if you have to create a performance regression, let the potential caller know (option 1 does that)

src/security/ErrorDetail.cc
```diff
@@ -676,7 +676,7 @@ Security::ErrorDetail::err_descr() const
     if (!error_no)
         return "[No Error]";
 #if USE_OPENSSL
-    if (const char *err = detailEntry.descr.termedBuf())
+    if (const char *err = detailEntry.descr.c_str())
```
rousskov, line 679: [Same](https://github.com/squid-cache/squid/pull/1451/files#r1289104529) problem in this newly added change: Unlike the old condition, the new condition is always true.

src/HttpHeaderTools.cc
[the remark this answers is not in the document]
kinkie: > > Doing what you suggest would balloon the scope of the PR.
> 
> The above assertion is false because suggested option 1 keeps the scope of this PR exactly the same.

You are right; I had improperly focused only on part of your answer.

> > I would prefer an approach where [we first introduce bridge functions and then get rid of them].
> 
> I am not against bridge functions per se, and suggested option 1 arguably continues to use a bridge function. A bridge function is the best way forward in many cases, but each suggested conversion should be evaluated on its own. Attempts to confine all future conversions to one conversion technique (e.g., bridge functions) seem counter productive to me because Squid code is too varied for one technique to be the right choice in all conversion cases.
> 
> As you know, I support minimal conversion PRs and even prefer those whenever they are feasible. Needless to say, there are many ways to convert and even many ways to create bridge functions. Naturally, there are certain principles or invariants we should observe _regardless_ of the PR size or the overall conversion approach. This change request is about violations of some of those principles:
> 
> * avoid performance regressions (option 2 avoids such a regression but requires more work)
> * if you have to create a performance regression, let the potential caller know (option 1 does that)

I'll for for option 1 for now. I would like to push through getting rid of SquidStrig

src/HttpHeaderTools.cc
```diff
@@ -233,6 +234,19 @@ httpHeaderParseQuotedString(const char *start, const int len, String *val)
     return 1;
 }
 
+/**
+ * Temporary bridge for the String-based variant to support conversion
+ * from SquidString to SBuf
+ * TODO: make this the primary implementation
```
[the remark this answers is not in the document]
kinkie: Done. I had put it next to the current one but I have no strong preference

src/security/ErrorDetail.cc
```diff
@@ -662,7 +662,7 @@ Security::ErrorDetail::err_code() const
 {
 #if USE_OPENSSL
     // try detailEntry first because it is faster
-    if (const char *err = detailEntry.name.termedBuf())
+    if (const char *err = detailEntry.name.c_str())
```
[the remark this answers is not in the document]
kinkie: I think we can get away with an isEmpty() test here. An uninitialised detailEntry will have a default-constructed SBuf which has zero length

src/HttpHeaderTools.cc
kinkie: Done in the SBuf-based variant

src/HttpHeaderTools.cc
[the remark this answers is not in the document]
kinkie: Some of these are borne out of wanting to avoid issues such as [this other comment](https://github.com/squid-cache/squid/pull/1451#discussion_r1289147284), coupled with the opportunity that these methods have few callers. I'll see where it leads and that it doesn't blow the scope up to much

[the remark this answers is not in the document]
rousskov: > I'll see where it leads and that it doesn't blow the scope up to much

Sounds like a good plan to me. When assessing where these changes have led, please keep in mind the essence of my previous [comment](https://github.com/squid-cache/squid/pull/1451#discussion_r1291613871): The added long-term API changes will need to be adjusted before they can land -- there is currently a significant amount of expected additional work in this PR. triggered by those long-term API changes.

src/ssl/ErrorDetailManager.cc
kinkie: I think using a static variable should be enough for that

src/ssl/ErrorDetailManager.cc
```diff
@@ -56,23 +57,24 @@ Ssl::ErrorDetailsList::getRecord(Security::ErrorCode value, ErrorDetailEntry &en
     return false;
 }
 
-const char *
+const SBuf
 Ssl::ErrorDetailsList::getErrorDescr(Security::ErrorCode value)
 {
     const ErrorDetails::const_iterator it = theList.find(value);
```
kinkie, line 63: @rousskov , I must be missing something really important. Such as.. where is theList initialised?

src/ssl/ErrorDetailManager.cc
rousskov: Using a static variable in such a context is unsafe because it silently screws up two callers in close proximity with each other. When possible, we should avoid such time bombs.

kinkie:
commit 61fdffb makes it clear that an ErrorDetailEntry is either all set or not set,
and adds a way to test it, making intent clear in callers

src/security/ErrorDetail.cc
kinkie: Addressed in 61fdffb

src/ssl/ErrorDetailManager.cc
kinkie: What is the best way to do it? SBuf::copy() to a static buffer is very kludgy (e.g. it doesn't guarantee zero-termination)

src/security/ErrorDetail.cc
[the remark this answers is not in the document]
kinkie: Addressed in commit 61fdffb

kinkie:
@rousskov , can you check the status I'm at? Thanks!

rousskov:
> @rousskov , can you check the status I'm at?

Sure, I flipped this PR to my side. Please give me a few days.

src/security/ErrorDetail.cc
```diff
@@ -716,18 +724,20 @@ Security::ErrorDetail::err_lib_error() const
...
         {"ssl_cn", &ErrorDetail::cn},
         {"ssl_notbefore", &ErrorDetail::notbefore},
         {"ssl_notafter", &ErrorDetail::notafter},
-        {"err_name", &ErrorDetail::err_code},
+        {"err_name", &ErrorDetail::err_code_legacy},
         {"ssl_error_descr", &ErrorDetail::err_descr},
         {"ssl_lib_error", &ErrorDetail::err_lib_error}
     };
 
+    // TODO why are we doing this? can't we just lookup the map above?
```
rousskov, line 740: I bet we are looping because the original code (before 83b053a) was looping and changing from a loop to a map lookup is not trivial. The original code had to loop because it used a raw/POD array.

src/security/ErrorDetail.cc
```diff
@@ -716,18 +724,20 @@ Security::ErrorDetail::err_lib_error() const
 size_t
 Security::ErrorDetail::convert(const char *code, const char **value) const
 {
+    // TODO change this to return SBuf
```
rousskov, line 727: There are a lot of problems in this official code, and these problems are in the way of doing the Right Thing in this PR. I am too busy to fix this quickly myself. If you can help, I can suggest the next step: This method should take an std::ostream reference (instead of a pointer to a c-string). The error extractors in the map below should print their information into the supplied stream (instead of returning a string).

rousskov (dismissed):

src/ssl/ErrorDetailManager.cc
rousskov: > I hope to find the answer to that question when I resume working on this PR.

I do not have a complete/polished answer, but its beginnings are sketched at https://github.com/squid-cache/squid/pull/1451#discussion_r1297800747.

src/security/ErrorDetail.cc
rousskov: > This method should take an std::ostream reference (instead of a pointer to a c-string). The error extractors in the map below should print their information into the supplied stream (instead of returning a string).

Done in 326b13c, but I have not reviewed, tested, and polished the result yet.

rousskov: > Done in [326b13c](https://github.com/squid-cache/squid/commit/326b13c98494cca1d95241d052f6db7bf425f495), but I have not reviewed, tested, and polished the result yet.

I pushed more fixes (as of commit 51ba982), but I am not done yet.

src/ssl/ErrorDetailManager.cc
```diff
@@ -24,6 +26,25 @@ void Ssl::errorDetailClean()
     Ssl::ErrorDetailsManager::Shutdown();
 }
 
+/// ErrorDetailEntry constructor helper that extracts a quoted HTTP field value
+static SBuf
+SlowlyParseQuotedField(const char * const description, const HttpHeader &parser, const char * const fieldName)
+{
+    String fieldValue;
+    if (!parser.hasNamed(fieldName, strlen(fieldName), &fieldValue))
+        throw TextException(ToSBuf("Missing ", description), Here());
```
rousskov, line 35: This check is missing in the official code. Without this check httpHeaderParseQuotedString() crashes in the unlikely event an error-details.txt entry is missing a required "detail" or "descr" field. I have disclosed this fix in the updated PR description.

httpHeaderParseQuotedString() also has parsing bugs. I do not think we should touch that old function in this PR. I also do not think we should adjust our SlowlyParseQuotedString() wrapper to compensate for some of those httpHeaderParseQuotedString() problems. The wrapper should remain a wrapper.

src/ssl/ErrorDetailManager.cc
```diff
@@ -42,40 +63,11 @@ class ErrorDetailFile : public TemplateFile
...
-    const ErrorDetails::const_iterator it = theList.find(value);
-    if (it != theList.end()) {
-        return it->second.descr.termedBuf();
-    }
-
-    return nullptr;
-}
-
-const char *
-Ssl::ErrorDetailsList::getErrorDetail(Security::ErrorCode value)
```
rousskov, line 71: I have removed getErrorDescr() and getErrorDetail() (and the corresponding ErrorDetailsManager methods) because they had to be modified to work with SBuf _and_ these "find record and extract the requested field" methods were only making things more complex. An updated findRecord() interface addresses old caller needs without requiring a wrapper function for every field and "forwarding" those wrappers through the ErrorDetailsManager class to the ultimate consumer.

The diff does not show some of these modifications well, unfortunately.

rousskov (commented):
@kinkie, I am done with adjusting this PR. I hope the current version is OK, but there is a lot of noise, and I could have easily missed something. The ball is on your side.

src/security/ErrorDetail.h
```diff
@@ -80,16 +84,17 @@ class ErrorDetail: public ::ErrorDetail
 private:
     ErrorDetail(ErrorCode err, int aSysErrorNo);
 
+    // TODO: Rename to printCamelCaseDetailName()?
```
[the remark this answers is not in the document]
kinkie: Removing the TODO. Hopefully over time we will rename things,
there is no point in singling anythign specific out

src/ssl/ErrorDetailManager.cc
kinkie: okay, reverting, adding to my personal TODO list to look into httpHeaderParseQuotedString

src/ssl/ErrorDetailManager.cc
kinkie: ok, thanks!

src/security/ErrorDetail.h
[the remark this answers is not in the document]
rousskov: > there is no point in singling anythign specific out

The point of this specific TODO is documented in this change request. Do you think we cannot agree on the new names? That is the precondition for that TODO removal. It is not clear (to me) whether that precondition has been met.

src/ssl/ErrorDetailManager.cc
kinkie: Okay; I notice that Http::SlowlyParseQuotedString has some checking and uses exceptions for the fail path, so it seems compatible with the intent here. At this point, this wrapper is useless at all, I can just call Http::SlowlyParseQuotedString, as this function only has two callsites.

src/ssl/ErrorDetailManager.cc
```diff
@@ -24,6 +26,24 @@ void Ssl::errorDetailClean()
     Ssl::ErrorDetailsManager::Shutdown();
 }
 
+/// ErrorDetailEntry constructor helper that extracts a quoted HTTP field value
+static SBuf
+SlowlyParseQuotedField(const char * const description, const HttpHeader &parser, const char * const fieldName)
+{
+    String fieldValue;
+    parser.hasNamed(fieldName, strlen(fieldName), &fieldValue);
```
rousskov, line 34: Calling hasNamed() without checking its result does not make much sense. Before we decide how to write this code, could you please explain what was the motivation behind commit 088a048c92fd49daa1606b686edfd613a39c0d09? Commit message does not answer that question AFAICT.

I am aware of the following two statements, but they do not help me answer that "why commit https://github.com/squid-cache/squid/commit/088a048c92fd49daa1606b686edfd613a39c0d09" question:

>> Alex at https://github.com/squid-cache/squid/pull/1451#discussion_r1337658245: This check is missing in the official code. Without this check httpHeaderParseQuotedString() crashes in the unlikely event an error-details.txt entry is missing a required "detail" or "descr" field. I have disclosed this fix in the updated PR description.

>> httpHeaderParseQuotedString() also has parsing bugs. I do not think we should touch that old function in this PR. I also do not think we should adjust our SlowlyParseQuotedString() wrapper to compensate for some of those httpHeaderParseQuotedString() problems. The wrapper should remain a wrapper.

> Francesco at https://github.com/squid-cache/squid/pull/1451#discussion_r1347716925: Okay; I notice that Http::SlowlyParseQuotedString has some checking and uses exceptions for the fail path, so it seems compatible with the intent here. At this point, this wrapper is useless at all, I can just call Http::SlowlyParseQuotedString, as this function only has two callsites.

src/security/ErrorDetail.cc
```diff
@@ -714,31 +749,32 @@ Security::ErrorDetail::err_lib_error() const
  \retval 0 for unsupported codes
 */
 size_t
-Security::ErrorDetail::convert(const char *code, const char **value) const
+Security::ErrorDetail::convertErrorCodeToDescription(const char * const code, std::ostream &os) const
```
rousskov, line 752: This renaming is rather misleading because this method does not convert ErrorCode (yes, there is such a thing in this context) to its description. This code expands `%code`. I recommend leaving this method name unchanged in this PR.

If (and only if) you insist on renaming, I can suggest:

```suggestion
Security::ErrorDetail::expandPercentCode(const char * const code, std::ostream &os) const
```

or

```suggestion
Security::ErrorDetail::substitutePercentCode(const char * const code, std::ostream &os) const
```

or something like that...

rousskov (changes requested):

src/security/ErrorDetail.cc
kinkie: I like expandPercentCode; "convert" is very  nondescriptive

src/ssl/ErrorDetailManager.cc
kinkie: hasNamed() is misnamed; the third argument is an OUT parameter, which returns into the provided string, maybe this is the source of the confusion. Can we do without checking the return value? Yes, because if the header is not found, an empty string will be returned, and Http::SlowlyParseQuotedString will call httpHeaderParseQuotedString which will fail because the first character is not a ```"``` .

I had interpreted your request in this way; but yes, if you ask me, one more free check would help

kinkie: I misinterpreted your comment as a suggestion to remove the check. Accepted your suggestion reinstating it, I'm good with the code as is. I don't think there's any more pending requests for improvements, are there?

rousskov: > I don't think there's any more pending requests for improvements, are there?

I do not see any. Please feel free to clear for merge when you are comfortable with PR code.

rousskov:
> @rousskov , ready at any time; it's only missing an explicit approval

This PR was missing an `M-cleared-for-merge` label. It does not need an explicit approval (because your implicit approval is enough to merge this PR on a "slow burner" path).

FYI: It takes Anubis a few minutes to scan all PRs after a manual PR change, so you will see a stale "PR approval" check right after manual PR changes.

As of now, if all PR tests succeed and no new merging problems (or last minute objections) surface, then this PR should be automatically staged and merged.

PR #1454, opened 2023-08-11
Fix cbdata assertion in carpInit()

rousskov:
... after commit e7959b5

    FATAL: assertion failed: cbdata.cc:276: "c->locks > 0"

We cannot test cbdata validity of CachePeer pointers stored in
Config.peers because Config.peers own CachePeer objects. By definition,
owners should not lock their objects and do not need to test object
validity: Owners determine the lifetime of those objects! Naturally,
unlocked objects must not be tested for validity by others because to
test an object validity one has to have a lock that preserves object
metadata even after the object is invalidated (by its owner).

### DIFF
--- a/src/carp.cc
+++ b/src/carp.cc
@@ -59,9 +59,6 @@ carpInit(void)
 
     RawCachePeers rawCarpPeers;
     for (auto p = Config.peers; p; p = p->next) {
-        if (!cbdataReferenceValid(p))
-            continue;
-
         if (!p->options.carp)
             continue;
 


rousskov:
AFAICT, we have a new CI test that would have caught this bug before it was merged, but that test is waiting for the bug fix in #1368 to land... Hopefully, I would be able to rewrite that fix and finally add that cache_peer test in the foreseeable future.

yadij:
FTR, CachePeer should be RefCounted instead of CBDATA. The configuration list holds one lock, and each transaction using is holds another. That way reconfigure just drops the main lock without causing nil-pointer problems to transaction logging of ongoing activity. This pattern works well for most of the complex configuration objects.

yadij (approved):

rousskov:
> CachePeer should be RefCounted

Yes, and we are actively working towards that goal. It takes about 8 minimal PRs to get there. The fifth PR in that series (#1424) has been waiting for review for three weeks:

* 97269a7 Destroy an idle PeerDigest after its CachePeer disappears
* 0f78379 Stop leaking PeerDigests on reconfiguration
* e373298 Forget non-peer access details
* e7959b5 Do not use raw pointers to index CARP CachePeers
* #1424
* Do not use raw pointers to index srchash and userhash CachePeers
* RefCount CachePeers

rousskov:
> yadij added backport-to-v6 labels
>  changed the title [to move `e7959b5` into PR description]

@yadij, I recommend undoing both actions quoted above. This PR fixes a very recent bug in master/v7 (e7959b56a56634766df84324ef8383540ae36118). That bug has not been backported to v6 and is currently _not_ earmarked for backporting. If Francesco and you decide to backport that change to v6, it should be, IMO, backported with this fix squashed into a single backporting commit.

This is just a recommendation regarding this PR metadata. I do not insist on any changes.

PR #1456, opened 2023-08-12
Add tls_outgoing_options_for_retries

rousskov:
acl downgrade ...
    tls_outgoing_options cipher=ALL@SECLEVEL=4
    tls_outgoing_options_for_retries cipher=ALL@SECLEVEL=3 if downgrade

This directive is handy in environments where Squid should communicate
using strict TLS options by default (see tls_outgoing_options) but may
relax those restrictions to accommodate certain older origin servers
that require inferior (but still deemed "secure enough") settings.

Popular browsers also retry with special TLS options after certain TLS
failures, so we are not inventing something new here. Unfortunately,
Squid cannot rely on those client features, for several reasons. For
example, Squid may not have (or may be unable to relay) enough TLS
server information to trigger browser-initiated retries. Squid may also
be dealing with a user agent that does not retry (to Squid admin's
satisfaction).

### DIFF
--- a/src/ConfigOption.h
+++ b/src/ConfigOption.h
@@ -24,21 +24,32 @@ namespace Configuration {
 /// (i.e. the Config objects and similar/related global state). To facilitate
 /// reuse, implementations/specializations should also be independent from any
 /// specific configuration directive name and its squid.conf location.
-///
-/// TODO: Support multi-directive components of various kinds.
 template <class T>
 class Component
 {
 public:
-    /* the code adding "TYPE: T" to cf.data.pre must specialize these */
+    /* code adding "REPETITIONS: banned" to cf.data.pre must specialize these */
 
     /// creates a new T instance using the given parser; never returns nil
     static T Parse(ConfigParser &);
 
-    /// reports the current T instance configuration in squid.conf format
+    /// reports configuration parameters of the given T instance in squid.conf format
     static void Print(std::ostream &, const T &);
 
-    /// destroys Parse() result
+    /* code adding "REPETITIONS: update" to cf.data.pre must specialize these */
+
+    /// creates a new T instance for one or more ParseAndUpdate() calls
+    static T Create();
+
+    /// parses the first and any subsequent directive occurrence, updating T
+    static void ParseAndUpdate(T &, ConfigParser &);
+
+    /// reports the given T instance configuration in squid.conf format
+    static void PrintDirectives(std::ostream &, const T &, const char *directiveName);
+
+    /* code adding "TYPE: T" to cf.data.pre must specialize this */
+
+    /// destroys Parse() or Create() result
     static void Free(T);
 };
 
--- a/src/ConfigOption.h
+++ b/src/ConfigOption.h
@@ -44,8 +44,9 @@ public:
     /// parses the first and any subsequent directive occurrence, updating T
     static void ParseAndUpdate(T &, ConfigParser &);
 
+    // TODO: Upgrade to std::ostream after upgrading Security::PeerOptions::dumpCfg()
     /// reports the given T instance configuration in squid.conf format
-    static void PrintDirectives(std::ostream &, const T &, const char *directiveName);
+    static void PrintDirectives(StoreEntry *, const T &, const char *directiveName);
 
     /* code adding "TYPE: T" to cf.data.pre must specialize this */
 
--- a/src/ResolvedPeers.h
+++ b/src/ResolvedPeers.h
@@ -22,7 +22,7 @@ public:
     explicit ResolvedPeerPath(const Comm::ConnectionPointer &conn): connection(conn) {}
 
     Comm::ConnectionPointer connection; ///< (the address of) a path
-    Security::PeerOptionsPointer tlsOptions; ///< custom TLS session options (or nil)
+    Security::PeerContextPointer tlsContext; ///< custom TLS session options (or nil)
     bool available = true; ///< whether this path may be extracted
 };
 
@@ -54,7 +54,7 @@ public:
 
     /// makes the previously extracted path available for extraction at its
     /// original position and resets TLS parameters to use for that path
-    void reinstatePath(const PeerConnectionPointer &, const Security::PeerOptionsPointer &tlsOptions);
+    void reinstatePath(const PeerConnectionPointer &, const Security::PeerContextPointer &tlsContext);
 
     /// extracts and returns the first queued address
     PeerConnectionPointer extractFront();
@@ -129,7 +129,7 @@ public:
 
     PeerConnectionPointer() = default;
     PeerConnectionPointer(std::nullptr_t): PeerConnectionPointer() {} ///< implicit nullptr conversion
-    PeerConnectionPointer(const Comm::ConnectionPointer &conn, const size_type pos, const Security::PeerOptionsPointer &tlsParameters): connection_(conn), position_(pos), tlsOptions_(tlsParameters) {}
+    PeerConnectionPointer(const Comm::ConnectionPointer &conn, const size_type pos, const Security::PeerContextPointer &tlsParameters): connection_(conn), position_(pos), tlsContext_(tlsParameters) {}
 
     /* read-only pointer API; for Connection assignment, see finalize() */
     explicit operator bool() const { return static_cast<bool>(connection_); }
@@ -140,7 +140,7 @@ public:
     operator const Comm::ConnectionPointer&() const { return connection_; }
 
     /// custom TLS session options (or nil)
-    auto &tlsOptions() const { return tlsOptions_; }
+    auto &tlsContext() { return tlsContext_; }
 
     /// upgrade stored peer selection details with a matching actual connection
     void finalize(const Comm::ConnectionPointer &conn) { connection_ = conn; }
@@ -159,8 +159,8 @@ private:
     size_type position_ = npos;
     friend class ResolvedPeers;
 
-    /// \copydoc tlsOptions()
-    Security::PeerOptionsPointer tlsOptions_;
+    /// \copydoc tlsContext()
+    Security::PeerContextPointer tlsContext_;
 };
 
 /// summarized ResolvedPeers (for debugging)
--- a/src/ResolvedPeers.h
+++ b/src/ResolvedPeers.h
@@ -140,7 +140,7 @@ public:
     operator const Comm::ConnectionPointer&() const { return connection_; }
 
     /// custom TLS communication settings (or nil)
-    auto &tlsContext() { return tlsContext_; }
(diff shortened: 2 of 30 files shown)


src/security/PeerOptions.h
```diff
@@ -145,15 +150,88 @@ class PeerOptions
...
+/// TLS client code.
+class FuturePeerContext: public RefCountable
+{
+public:
+    FuturePeerContext(PeerOptions &o, const ContextPointer &c): options(o), raw(c)
+    {
+    }
+
+    PeerOptions &options; ///< TLS context configuration
+    const ContextPointer &raw; ///< TLS context configured using options
```
rousskov, line 202: Q1: Should I post a PR that fixes bug 5293 while using FuturePeerContext shim as necessary to prevent lots of code changes that have pretty much nothing to do with that bug? The alternative is to post that bug-fixing PR without the shim (and with a lot of noise).

This shim adds a small performance overhead (we need to allocate those trivial shim objects whenever a new TLS connection is created), but if we promptly remove it in a followup PR, then it is not a big problem.

src/ConfigOption.h
```diff
@@ -24,21 +24,33 @@ namespace Configuration {
...
+
+    /// creates a new T instance for one or more ParseAndUpdate() calls
+    static T Create();
+
+    /// parses the first and any subsequent directive occurrence, updating T
+    static void ParseAndUpdate(T &, ConfigParser &);
+
+    // TODO: Upgrade to std::ostream after upgrading Security::PeerOptions::dumpCfg()
+    /// reports the given T instance configuration in squid.conf format
+    static void PrintDirectives(StoreEntry *, const T &, const char *directiveName);
```
rousskov, line 49: Q2: Should we upgrade PeerOptions::dumpCfg() and related code to use std::ostream (in a yet another separate PR)? 

Configuration::Components enhancement adds PrintDirectives() API that uses StoreEntry to report directives configuration. Its existing sister Print() API (for squid.conf directives that cannot be repeated) uses std::ostream. Ideally, the new API should use std::ostream as well, but we cannot do that easily because the existing PeerOptions::dumpCfg() code on that new API path still uses StoreEntry and has to be changed throughout to use std::ostream, introducing a lot of unrelated noise.

The alternative here is to add an API using StoreEntry and wait for a better opportunity to upgrade that (and all the new code that will start using that new API by then).

src/security/Session.cc
```diff
@@ -180,13 +180,14 @@ CreateSession(const Security::ContextPointer &ctx, const Comm::ConnectionPointer
...
-        return CreateSession(ctx, c, Security::ProxyOutgoingConfig, Security::Io::BIO_TO_SERVER, squidCtx);
-
-    auto *peer = c->getPeer();
-    return CreateSession(ctx, c, peer->secure, Security::Io::BIO_TO_SERVER, squidCtx);
+    // TODO: We cannot make ctx constant because CreateSession() takes
+    // non-constant ctx.options (PeerOptions). It does that because GnuTLS
+    // needs to call PeerOptions::updateSessionOptions(), which is not constant
+    // because it compiles options (just in case) every time. To achieve
+    // const-correctness, we should compile PeerOptions once, not every time.
+    return CreateSession(ctx.raw, c, ctx.options, Security::Io::BIO_TO_SERVER, squidCtx);
```
rousskov, line 190: Q3: Should we postpone fixing old (but now marked) problems with PeerOptoons being modified at transaction handling time?

The alternative is to start revamping that extremely messy class so that we can make sure that the options are parsed before first use by transactions (and, hence, remain constant once they are parsed). This refactoring will require at least one dedicated PR. I am especially concerned with the affected experimental GnuTLS code that would be difficult to test and collaborate on. On the bright side, fixing those old problems would allow us to improve const-correctness of new PeerContext code now, when that code is added.

src/adaptation/icap/Xaction.cc
```diff
@@ -49,13 +49,15 @@ class IcapPeerConnector: public Security::PeerConnector {
...
 
     /* Security::PeerConnector API */
     bool initialize(Security::SessionPointer &) override;
     void noteNegotiationDone(ErrorState *error) override;
-    Security::ContextPointer getTlsContext() override {
-        return icapService->sslContext;
+    Security::FuturePeerContextPointer peerContext() const override {
+        // XXX: We should be able to keep the options part constant.
+        auto &cfg = const_cast<Adaptation::ServiceConfig&>(icapService->cfg());
+        return MakeLikeFuture(cfg.secure, icapService->sslContext);
```
rousskov, line 60: This is one more place relevant to const-correctness question Q3 that got marked in this PR.

rousskov (commented):
This draft PR has three key changes:

1. Security::PeerContext: Pairs TLS context with its TLS options, addressing bug [5293](https://bugs.squid-cache.org/show_bug.cgi?id=5293). To implement this without widespread out-of-scope changes in PeerOptions and  ContextPointer-using code, I had to add a FuturePeerContext shim. Branch commit 2668068 has more info.

2. `REPETITIONS: update`: Support multi-directive Configuration::Components and corresponding cf.data.pre entries. Adding this support addresses Components TODO. Now, new directives (like the one added in this PR) can use that configuration API and, when we have the time, more of the old directives (e.g. http_access) can now be converted. This change may also help with generating better squid.conf documentation. Branch commit ab65c42 has more info.

3. The tls_outgoing_options_for_retries feature itself, of course. The code "works".

My plan is to extract the first two changes into individual precursor PRs (while keeping this PR as as draft). If you have any objections, please let me know ASAP.

I am posting this draft PR to ask three specific questions in the change requests attached to this comment. My own answer to all those questions is "yes". If your answer is "no" (or if you see some better alternatives than the ones I have mentioned), please let me know ASAP. Thank you!

src/security/Session.cc
yadij: Agreed on the postpone. At least it is out-of-scope for this PR.

FTR; the GnuTLS code has proven to be quite stable and is in use by a large number (~2k) of Debian installs. IMO we can stop calling the forward-proxy parts experimental as of v7/master.

src/cf_gen.cc
```diff
@@ -97,6 +97,9 @@ class Entry
     LineList cfgLines; ///< between CONFIG_START and CONFIG_END
     int array_flag = 0; ///< TYPE is a raw array[] declaration
 
+    /// whether this is a "REPETITIONS: update" entry
```
yadij, line 100: IMO calling the option **REPEATS** would help avoid problems from developer typos.

src/ConfigOption.h
yadij: From core meeting discussion:
 Changes included here are useful to simplify the feature addition, technical debt induced by incomplete transition is acceptable in light of how this overlaps the ongoing ConfigParser and CacheManager refactoring.

src/ResolvedPeers.h
```diff
@@ -124,7 +129,7 @@ class PeerConnectionPointer
 
     PeerConnectionPointer() = default;
     PeerConnectionPointer(std::nullptr_t): PeerConnectionPointer() {} ///< implicit nullptr conversion
-    PeerConnectionPointer(const Comm::ConnectionPointer &conn, const size_type pos): connection_(conn), position_(pos) {}
+    PeerConnectionPointer(const Comm::ConnectionPointer &conn, const size_type pos, const Security::PeerContextPointer &tlsParameters): connection_(conn), position_(pos), tlsContext_(tlsParameters) {}
```
yadij, line 132:
```suggestion
    PeerConnectionPointer(const Comm::ConnectionPointer &conn, const size_type pos, const Security::PeerContextPointer &ctx):
        connection_(conn),
        position_(pos),
        tlsContext_(ctx)
    {}
```

yadij (changes requested):

src/ConfigOption.h
rousskov: > Should we upgrade PeerOptions::dumpCfg() and related code to use std::ostream (in a yet another separate PR)?

That upgrade is done at #1460.

src/cf.data.pre
```diff
@@ -3076,6 +3076,82 @@ DOC_START
...
+	See tls_outgoing_options for documentation of recognized options,
+	including "ciphers=..." and "cafile=...".
+
+	This directive is consulted whenever an attempt to establish a TLS
+	connection with an origin server fails with ERR_SECURE_CONNECT_FAIL:
+
+	* If no rules are configured or if no configured rules matched, Squid does
+	  not treat ERR_SECURE_CONNECT_FAIL outcome specially (i.e. Squid may
+	  still attempt to reforward the request to a different IP address, using
+	  regular reforwarding logic).
```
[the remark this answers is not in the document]
kinkie: While I agree this could be improved, I don't think it's unclear or misleading to
my - and here I proxy for people with a moderate amount of understanding.
I correctly understood the target use case for this option by just reading this help
text, that's a pretty high bar to clear already.
The kind of subtleties I see here may not even belong to a configuration template,
but to a book.
I'd defer any improvements to a followup PR, but TBH even if there were none,
I would be happy

src/cf.data.pre
```diff
@@ -3076,6 +3076,82 @@ DOC_START
...
+
+	Directives are evaluated in configuration order. The first matching
+	directive wins and stops further evaluation. A directive without ACLs
+	always matches when evaluated.
+
+	This feature has no effect on TLS cache_peers. Admins should use
+	cache_peer directives to customize TLS settings for those connections.
+	However, TLS connections through plain-text cache_peers may be retried.
+
+	This directive is handy in environments where Squid should communicate
```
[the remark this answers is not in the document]
kinkie: I think this text can belong here: release notes are not likely to be read by a new administrator
setting up a new squid, like they would by an existing administrator updating an older install.
These belong to both places, so I'm happy with the current status

src/cf_gen.cc
kinkie: The cardinality of the set of people affected by this is depressingly low.
Not worth discussing

src/security/PeerOptions.h
kinkie: I trust this to be done. Thanks!

kinkie (commented):
LGTM; please proceed

kinkie:
Overall shape is good enough for me.
Prequisite PR will preceed.
I don't think it's a good idea to backport the feature to v6.
Thanks!

src/security/PeerOptions.h
rousskov: > OK, we will try to do a small fix using a shim to facilitate v6 backporting. That matches the first change described in the https://github.com/squid-cache/squid/pull/1456#pullrequestreview-1575490152 for this Draft PR.

Posted as #1895. Once that PR lands in master/v7, I will update this PR accordingly, of course.

PR #1464, opened 2023-08-28
implement debugs_once

kinkie:
Implement a debugs() variant (named debugs_once()) that only generates output once per lifetime of a
squid process.

This can be used to avoid spamming the cache.log with messages such
as the one in src/helper/Reply.cc:200

### DIFF
--- a/src/debug/Stream.h
+++ b/src/debug/Stream.h
@@ -169,6 +169,9 @@ public:
     /// are expected.
     static void SettleSyslog();
 
+    /// returns true once for each different filename and line number, then false
+    static bool Once(const char * aFileName, const uint32_t aLineNo);
+
 private:
     static void FormatStream(std::ostream &);
     static void LogMessage(const Context &);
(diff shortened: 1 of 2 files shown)


kinkie:
I'm looking for feedback such as:
- should we reset the once() counters at reconfigure?
- should we use a DBG_ONCE bit-flag to the debug section instead of a copypasta-ed function?

The implementation works - tested by hand-patching main()

rousskov (changes requested):
> * should we reset the once() counters at reconfigure?

I hope the "use the existing feature" answer below will address this question as well.



> should we use a DBG_ONCE bit-flag to the debug section instead of a copypasta-ed function?

The currently proposed duplication of complex debugs() logic is a non-starter because it violates basic engineering principles (I can expand on that if needed).

The "used once" signal should come in a form of a debugging level flag/object/adjustment, just like cache_log_message implementation does.


I suspect that the best way to control the default number of messages is via the existing `cache_log_message limit=1` code/functionality and possibly even via that existing configuration interface (i.e. adding the corresponding DEFAULT lines). Please investigate those options.

kinkie:
Closing; this is superseded by PR #1481

PR #1466, opened 2023-08-30
Prep for 6.3

yadij:


kinkie (approved):

rousskov (approved):
We should stop propagating minor release change log updates through master.

PR #1468, opened 2023-09-04
Report all AsyncJob objects (mgr:jobs)

eduard-bagdasaryan:
This new cache manager report is useful in triage and test automation.
It complements mgr:objects, mgr:events, and mgr:filedescriptors reports.

### DIFF
--- a/src/base/AsyncJob.cc
+++ b/src/base/AsyncJob.cc
@@ -190,8 +190,16 @@ void
 AsyncJob::stat(StoreEntry *e)
 {
     PackableStream os(*e);
-    for (const auto job: Jobs_)
-        os << job->status() << "\n";
+    const char *ident = "    ";
+    for (const auto job: Jobs_) {
+        os << ident << job->id.prefix() << job->id.value << ":\n";
+        os << ident << ident << "type: " << job->typeName << '\n';
+        os << ident << ident << "status:" << job->status() << '\n';
+        if (!job->started_)
+            os << ident << ident << "started: false\n";
+        if (job->stopReason)
+            os << ident << ident << "stopped: " << job->stopReason << '\n';
+    }
 }
 
 void
--- a/src/base/AsyncJob.cc
+++ b/src/base/AsyncJob.cc
@@ -195,7 +195,7 @@ AsyncJob::ReportAllJobs(StoreEntry *e)
     // this loop uses YAML syntax, but AsyncJob::status() still needs to be adjusted to use YAML
     const char *indent = "    ";
     for (const auto job: Jobs_) {
-        os << indent << job->id.prefix() << job->id.value << ":\n";
+        os << indent << job->id << ":\n";
         os << indent << indent << "type: '" << job->typeName << "'\n";
         os << indent << indent << "status:" << job->status() << '\n';
         if (!job->started_)
--- a/src/base/AsyncJob.cc
+++ b/src/base/AsyncJob.cc
@@ -196,7 +196,7 @@ AsyncJob::ReportAllJobs(StoreEntry *e)
     const char *indent = "    ";
     for (const auto job: Jobs_) {
         os << indent << job->id.prefix() << job->id.value << ":\n";
-        os << indent << indent << "type: " << job->typeName << '\n';
+        os << indent << indent << "type: '" << job->typeName << "'\n";
         os << indent << indent << "status:" << job->status() << '\n';
         if (!job->started_)
             os << indent << indent << "started: false\n";
--- a/src/base/AsyncJob.cc
+++ b/src/base/AsyncJob.cc
@@ -16,6 +16,7 @@
 #include "base/TextException.h"
 #include "cbdata.h"
 #include "mgr/Registration.h"
+#include "mem/PoolingAllocator.h"
 #include "MemBuf.h"
 #include "Store.h"
 
--- a/src/base/AsyncJob.cc
+++ b/src/base/AsyncJob.cc
@@ -15,9 +15,9 @@
 #include "base/PackableStream.h"
 #include "base/TextException.h"
 #include "cbdata.h"
-#include "mgr/Registration.h"
 #include "mem/PoolingAllocator.h"
 #include "MemBuf.h"
+#include "mgr/Registration.h"
 #include "Store.h"
 
 #include <ostream>
--- a/src/base/AsyncJob.cc
+++ b/src/base/AsyncJob.cc
@@ -20,6 +20,8 @@
 
 InstanceIdDefinitions(AsyncJob, "job");
 
+std::set<AsyncJob *> AsyncJob::Jobs;
+
 void
 AsyncJob::Start(const Pointer &job)
 {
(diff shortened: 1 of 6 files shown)


src/main.cc
```diff
@@ -1203,6 +1203,8 @@ mainInitialize(void)
     FwdState::initModule();
     SBufStatsAction::RegisterWithCacheManager();
 
+    AsyncJob::RegisterWithCacheManager();
```
eduard-bagdasaryan, line 1206: We are trying to reduce the number of explicit dependencies in main.cc (see commit 230d441). However, today, main.cc already knows about the AsyncJob class, and replacing this call with CallRunnerRegistrator() would require adding a few extra code lines because AsyncJob.cc does not use a RegisteredRunner yet. Should we do this refactoring in this PR?

eduard-bagdasaryan:
Here is a sample of mgr:jobs output for one kid:

```
by kid1 {
    job169:
        type: Mgr::StoreToCommWriter
        status: [ job169]
    job167:
        type: HttpStateData
        status: [ job167]
    job163:
        type: ClientHttpRequest
        status: [ job163]
        started: false
    job162:
        type: Http1::Server
        status: [ job162]
    job2:
        type: Ipc::UdsOp
        status: [ job2]
    job5:
        type: Comm::TcpAcceptor
        status: FD 27, [::] [ job5]
} by kid1

```

rousskov (approved):
@eduard-bagdasaryan, thank you for adding this useful report. I committed two minor polishing changes. Please let me know if they are problematic.

src/base/AsyncJob.cc
```diff
@@ -12,14 +12,21 @@
...
 #include "MemBuf.h"
+#include "mgr/Registration.h"
+#include "Store.h"
 
 #include <ostream>
+#include <unordered_set>
 
 InstanceIdDefinitions(AsyncJob, "job");
 
+static std::unordered_set<AsyncJob *, std::hash<AsyncJob *>, std::equal_to<AsyncJob *>, PoolingAllocator<AsyncJob *> > Jobs_;
```
yadij, line 28: Please do not use global statics. Correct initialization keeps coming up as problematic for these.
A variable which is only used internally by the `class AsyncJob` can be a private member, otherwise prefer to wrap with a function for cross-unit access.

src/base/AsyncJob.cc
```diff
@@ -179,3 +188,26 @@ const char *AsyncJob::status() const
...
+void
+AsyncJob::ReportAllJobs(StoreEntry *e)
+{
+    PackableStream os(*e);
+    // this loop uses YAML syntax, but AsyncJob::status() still needs to be adjusted to use YAML
+    const char *indent = "    ";
+    for (const auto job: Jobs_) {
+        os << indent << job->id << ":\n";
+        os << indent << indent << "type: '" << job->typeName << "'\n";
+        os << indent << indent << "status:" << job->status() << '\n';
```
yadij, line 200: YAML "flow sequence" should be wrapped in square brackets:
```suggestion
        os << indent << indent << "status: [" << job->status() << "]\n";
```

src/base/AsyncJob.cc
```diff
@@ -179,3 +188,26 @@ const char *AsyncJob::status() const
     return buf.content();
 }
 
+void
+AsyncJob::ReportAllJobs(StoreEntry *e)
+{
+    PackableStream os(*e);
+    // this loop uses YAML syntax, but AsyncJob::status() still needs to be adjusted to use YAML
```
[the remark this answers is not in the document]
yadij: IMO `status()` is currently best interpreted as a YAML flow sequence https://yaml.org/spec/1.2.2/#flow-sequences

src/base/AsyncJob.cc
```diff
@@ -179,3 +188,26 @@ const char *AsyncJob::status() const
...
+    const char *indent = "    ";
+    for (const auto job: Jobs_) {
+        os << indent << job->id << ":\n";
+        os << indent << indent << "type: '" << job->typeName << "'\n";
+        os << indent << indent << "status:" << job->status() << '\n';
+        if (!job->started_)
+            os << indent << indent << "started: false\n";
+        if (job->stopReason)
+            os << indent << indent << "stopped: '" << job->stopReason << "'\n";
```
yadij, line 204: Reason text may contain single-quote characters, yes? which break this quoting style.
There are two alternatives;

Either use double-quotes and ensure that reason text is appropriately slash-escaped per YAML:
```suggestion
            os << indent << indent << "stopped: \"" << Format::QuoteYamlString(job->stopReason) << "\"\n";
```
(this will require a YAML specific encoder added to `src/format/Quoting.cc` until we add a dedicated YAML library).

or, use the YAML "plain style" lines:
```suggestion
            os << indent << indent << "stopped: " << job->stopReason << "\n";
```

yadij (changes requested):

rousskov (commented):
@yadij, we will add YAML string encoder to make progress, but I hope I can convince you to consider all other change requests addressed without PR code changes. Please re-review those change requests.

src/base/AsyncJob.cc
yadij: Punted to future PR.

src/base/AsyncJob.cc
yadij: Update: developer meeting today decided to start a policy of "no new globals".

yadij: Update: developer meeting today decided to start a policy of "no new globals".

yadij (changes requested):

src/base/AsyncJob.cc
rousskov: Branch commit b05b183 adheres to the new "no new globals" policy.

src/base/AsyncJob.cc
```diff
@@ -12,14 +12,27 @@
...
+#include <unordered_set>
 
 InstanceIdDefinitions(AsyncJob, "job");
 
+/// a set of all AsyncJob objects in existence
+static auto &
+AllJobs()
+{
+    static const auto jobs = new std::unordered_set<AsyncJob *, std::hash<AsyncJob *>, std::equal_to<AsyncJob *>, PoolingAllocator<AsyncJob *> >();
+    return *jobs;
```
rousskov, line 33: Some of the old Squid code does not allocate the object on heap, exposing us to `static` _destruction_ order [fiasco](https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2) risks. This implementation follows the recommended [practice](https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use) for use cases without important side effects in (never-called) destructors.

Most new code should follow this pattern for wrapping would-be-globals AFAICT.

rousskov (approved):

src/base/AsyncJob.cc
rousskov: The documentation for the new "no new globals" policy has been proposed at https://github.com/squid-cache/squid-cache.github.io/pull/6. The code in that documentation proposal matches this PR code.

yadij (approved):

PR #1471, opened 2023-09-08
Drop helpless helper requests

eduard-bagdasaryan:
When a helper has no server programs left (e.g., because all started
programs, if any, have quit and no new programs could be started), drop
all queued helper requests, so that their corresponding master
transactions do not get stuck.

Existing handleFewerServers() already handles this situation by killing
Squid, but we must drop queued requests in other places (that can be
found by locating code that adjusts childs.n_active) because:

* that existing Squid-killing code effectively excludes several use
  cases (e.g., it ignores common startup=0 helper configurations);

* handleFewerServers() may not be called in some relevant cases
  (e.g., when no new servers were needed after a reconfiguration);

* handleFewerServers() may be followed by a helperOpenServers() call
  (that may open new servers for handling queued requests); we do
  not want to kill queued requests in those cases.

Squid may recover and successfully start another helper later, but
transaction X progress must not rely on an unrelated transaction Y
actions: If transaction Y does not show up or does not result in a
successful helper program start, then transaction X will continue to be
queued at the helper level, possibly forever (even request timeouts are
handled by individual helper server objects that we lack here).

Also significantly simplified HelperServerClosed() implementations.


rousskov (approved):
Thank you for making a yet another small step towards smooth reconfiguration support (and simplifying convoluted HelperServerClosed() methods on the way).

src/helper.cc
```diff
@@ -869,8 +874,35 @@ helper::handleKilledServer(HelperServerBase *srv, bool &needsNewServers)
 
         if (childs.needNew() > 0) {
             srv->flags.shutdown = true;
-            needsNewServers = true;
+            helperOpenServers(this);
```
yadij, line 877: What about the `class statefulhelper`s which need this logic to call `helperOpenStatefulServer(this)` instead ?

yadij (changes requested):

rousskov (changes requested):

src/helper.cc
eduard-bagdasaryan: > Facilitate future reuse of ::helper class PR

This is now [PR1480](https://github.com/squid-cache/squid/pull/1480/files) .

rousskov: > Wait for a prerequisite minimal PR that makes (the equivalent of) helperOpenServers() virtual.

Done in recent official commit bd71920. That virtual method was named openSessions(). This PR has been updated to use that method, fixing this bug.

rousskov (approved):
@eduard-bagdasaryan, thank you for fixing this PR.

@yadij, please re-review.

yadij (approved):

yadij:
Sorry this took so long. Clearing for merge ASAP.

PR #1472, opened 2023-09-09
Log %err_code for ERR_RELAY_REMOTE transactions

rousskov:
For ERR_RELAY_REMOTE transactions, Squid was logging %err_code as "-"
because BuildHttpReply() was not updating HttpRequest::error or ALE.
That update was missing because the pre-computed response in those
transactions triggered a premature exit from BuildHttpReply().

BuildHttpReply() should not be updating errors at all, but significant
code refactoring required to fix that problem needs a dedicated change.

### DIFF
--- a/src/errorpage.cc
+++ b/src/errorpage.cc
@@ -1277,8 +1277,7 @@ HttpReply *
 ErrorState::BuildHttpReply()
 {
     // Make sure error codes get back to the client side for logging and
-    // error tracking. XXX: ErrorState should only reflect already recorded
-    // information instead of being responsible for updating records.
+    // error tracking.
     if (request) {
         request->error.update(type, detail);
         request->error.update(SysErrorDetail::NewIfAny(xerrno));
--- a/src/errorpage.cc
+++ b/src/errorpage.cc
@@ -678,15 +678,16 @@ ErrorState::NewForwarding(err_type type, HttpRequestPointer &request, const Acce
     return new ErrorState(type, status, request.getRaw(), ale);
 }
 
-ErrorState::ErrorState(err_type t) :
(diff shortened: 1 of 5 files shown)


src/errorpage.cc
```diff
@@ -1345,17 +1355,6 @@ ErrorState::BuildHttpReply()
...
-    // Make sure error codes get back to the client side for logging and
-    // error tracking.
-    if (request) {
-        request->error.update(type, detail);
-        request->error.update(SysErrorDetail::NewIfAny(xerrno));
-    } else if (ale) {
-        Error err(type, detail);
-        err.update(SysErrorDetail::NewIfAny(xerrno));
-        ale->updateError(err);
-    }
```
rousskov, line 1357: Moving this code up (without changes), before the `if (response_)` check, fixes the primary cause of this bug.

The PR also adjusts the ErrorState constructor dedicated to ERR_RELAY_REMOTE responses to get ALE because the moved code updates ALE, even though those updates are probably not related to the current ERR_RELAY_REMOTE code paths. That improvement also proves that this PR did not miss any cases where ERR_RELAY_REMOTE ErrorState was created without ALE.

This messy code needs quite a bit of additional work. We will follow up with at least one (but probably two) PRs that should address out-of-scope problems already marked in ErrorState declaration and new missing error updates problems discovered while working on this PR.

test-suite/test-functionality.sh
```diff
@@ -214,6 +214,7 @@ main() {
     then
         local default_tests="
             pconn
+            dead-peer
```
rousskov, line 217: This regression test is dedicated to the fixed bug and another logging bug fixed in recent master/v7 commit ea3f56e.

yadij (approved):

PR #1474, opened 2023-09-13
Do not use raw pointers to index sourcehash CachePeers 

eduard-bagdasaryan:
Simplified and improved code safety by using CbcPointers for sourcehash
cache_peers, as we have done for CARP peers in recent commit e7959b5.

Also fixed mgr:sourcehash Cache Manager reports to detail relevant
cache_peers instead of all cache_peers. This problem existed since
inception (2008 commit f7e1d9c) as detailed in recent commit e7959b5.

Also applied the new "no new globals" policy to CARP peering code, to
keep improved CARP and sourcehash peering code in sync.

### DIFF
--- a/src/carp.cc
+++ b/src/carp.cc
@@ -24,7 +24,12 @@
 #define ROTATE_LEFT(x, n) (((x) << (n)) | ((x) >> (32-(n))))
 
 /// CARP cache_peers ordered by their CARP weight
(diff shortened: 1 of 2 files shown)


rousskov (approved):
Thank you for propagating commit e7959b56a56634766df84324ef8383540ae36118 changes to sourcehash cache_peers.

PR #1477, opened 2023-09-15
WIP: helpers API cleanup

yadij:


src/helper.cc
```diff
@@ -557,19 +557,20 @@ helper::prepSubmit()
...
 
-    submit(buf, callback, data); // will send or queue
+    submit(buf, callback, data, reservation); // will send or queue
     return true; // request submitted or queued
 }
 
 /// dispatches or enqueues a helper requests; does not enforce queue limits
 void
-helper::submit(const char *buf, HLPCB * callback, void *data)
+helper::submit(const char *buf, HLPCB * callback, void *data, const Helper::ReservationId &)
```
rousskov, line 573: None of the concepts represented by the current `::helper` class should know about helper reservations. Reservation is a concept exclusive to the `statefulhelper` class. We should not add reservation-specific parameters to `::helper`, even if these reservations are going to be always nil for stateless helpers.

rousskov (changes requested):
I strongly recommend postponing this work until we land the promised "minimal" PR(s) that [rename](https://github.com/squid-cache/squid/pull/1389#discussion_r1237075585) helper classes and [add virtual method support](https://github.com/squid-cache/squid/pull/1471#discussion_r1324659044) to Helper::Client. @eduard-bagdasaryan is working on those changes now, and I expect them to be posted for the official review very soon. 

After those minimal PRs land, we can start chipping away at various specific helper problems that require virtual methods to be solved. IMO, this strategy provides better working conditions for individual followup PRs and reduces conflicts among them.

src/helper.cc
yadij: Well then. Since the objected-to detail is a hard prerequisite of this PR change I am closing it entirely. Will see what can be done after the @eduard-bagdasaryan  changes.

PR #1478, opened 2023-09-16
Bug 5300: cachemgr.cgi assertion

yadij:


kinkie:
Has this been tested?

yadij:
Yes

rousskov (commented):
> yadij requested a review from rousskov

I have to decline this opportunity to review cachemgr.cgi code because, as you probably know, I believe this whole feature should be removed from the Squid Project instead of being improved and supported by us (for the reasons I have already articulated many times but can repeat here if needed). This and other recently discovered cachemgr.cgi bugs (that went unnoticed since Squid v4) are a case in point.

kinkie (approved):
I agree we should remove cachemgr.cgi, but it is out there and people are using it, however disappointing that may be. Hopefully cachemgr.js or something similar will be our next stop

rousskov:
> I agree we should remove cachemgr.cgi,

I am very glad to hear that! I hope you will post a PR removing that feature.

> but it is out there and people are using it

Those facts do not preclude cachemgr.cgi removal from master/v7 IMO, especially given the tool low use (which the recent discovery of old problems helped demonstrate). The mountain of technical debt is one of the primary existential threats this Project is facing. This feature is a large boulder on that mountain; its impact goes well beyond the 1320 lines of tools/cachemgr.cc code.

> Hopefully cachemgr.js or something similar will be our next stop

I hope that next stop will not be inside this repository or, given the current amount of resources, this Project.

squidadm:
queued for backport to v6

yadij:
queued for backport to v5

PR #1479, opened 2023-09-16
Bug 5301: cachemgr.cgi not showing new manager interface URLs

yadij:
Also fix several related UI issues uncovered during testing:

* Prune the list of servers accessible via the CGI tool login.
 Their responses would be badly mangled if accessed via
 the old tools parse logic.
Also, hide the old login form if all servers use the new
manager interface.

* Ensure the 'menu' report is always used by default after
the CGI tool login. This prevents errors about MGR_INDEX
not being available on recent Squid releases. Restoring the
expected CGI tool behavior.


kinkie (approved):
looks good but oh boy..

squidadm:
queued for backport to v6

PR #1484, opened 2023-09-20
Maintenance: delete empty comm_poll.h

kinkie:
Unused since 2011 commit d841c88.
Empty since 2008 commit 5acc9f3.


rousskov (approved):
Thank you. I updated PR description to document when we stopped using this file.

PR #1486, opened 2023-09-21
Maintenance: remove openssl-ism in Certificate acl

kinkie:
When GnuTLS support was introduced, the class definitions
for Certificate and CertificateData were not updated to
use the generic Security names.

### DIFF
--- a/src/acl/Certificate.h
+++ b/src/acl/Certificate.h
@@ -17,7 +17,7 @@ namespace Acl
 {
 
 /// a "user_cert" or "ca_cert" ACL
-class ClientCertificateCheck : public ParameterizedNode<ACLData<Security::Certificate *>>
+class ClientCertificateCheck: public ParameterizedNode<ACLData<Security::Certificate *> >
 {
 public:
     /* ACL API */
--- a/src/acl/Certificate.h
+++ b/src/acl/Certificate.h
@@ -17,7 +17,7 @@ namespace Acl
 {
 
 /// a "user_cert" or "ca_cert" ACL
-class ClientCertificateCheck: public ParameterizedNode< ACLData<X509 *> >
+class ClientCertificateCheck : public ParameterizedNode<ACLData<Security::Certificate *>>
 {
 public:
     /* ACL API */
--- a/src/acl/Certificate.h
+++ b/src/acl/Certificate.h
@@ -17,7 +17,7 @@ namespace Acl
 {
 
 /// a "user_cert" or "ca_cert" ACL
-class ClientCertificateCheck: public ParameterizedNode<ACLData<Security::Certificate *> >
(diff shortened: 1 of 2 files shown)


src/acl/Certificate.h
```diff
@@ -17,7 +17,7 @@ namespace Acl
 {
 
 /// a "user_cert" or "ca_cert" ACL
-class ClientCertificateCheck: public ParameterizedNode< ACLData<X509 *> >
+class ClientCertificateCheck: public ParameterizedNode< ACLData<Security::Certificate *> >
```
rousskov, line 20: I believe this change is slightly wrong because it implies that ClientCertificateCheck can be used in GnuTLS builds. That implication is false -- the class implementation is still OpenSSL-specific. IMO, it is best to keep ClientCertificateCheck declaration OpenSSL-specific to match its implementation.

rousskov, line 20: I have removed the "so that it can pass stricter validation checksSecurity" reason in the original PR description because, IMO, we should not be testing OpenSSL-specific code outside of OpenSSL builds (even if this file can be successfully tested today after this change). For example, I should be able to add an OpenSSL-specific field to ClientCertificateCheck _without_ breaking any tests!

rousskov (commented):
`-0`: I think this change is slightly wrong, but I will not block it.

src/acl/Certificate.h
kinkie: I'm fine with that, but then I propose to guard the whole file with USE_OPENSSL as an alternative

rousskov: > I propose to guard the whole file with USE_OPENSSL as an alternative

This double-guarding of headers is one of the options discussed at https://github.com/squid-cache/squid/pull/1463#discussion_r1333133479. I hope this alternative is not going to be accepted and recommend waiting for the outcome of that discussion.

yadij:
FTR; This change depends on the PR #139 changes for sufficient testing to prove this change works.

src/acl/Certificate.h
rousskov: > [Amos](https://github.com/squid-cache/squid/pull/1486#issuecomment-1734597361): This change indirectly depends on the PR #139 changes for sufficient testing to prove this change works.

FWIW, I see no such dependency. This change obviously "works" -- changing a type alias does not affect compiled code. This class use and implementation are still OpenSSL-specific, so this PR does not depend on (or prevent) any GnuTLS-specific improvements.

I hope that this PR is closed without merging after #1463 merges, for the reasons outlined in this [change request](https://github.com/squid-cache/squid/pull/1486/files#r1333206353). AFAICT, in #1463, after this PR was posted, Francesco found a nice way to observe the "do not check what we do not make" principle, making this controversial change unnecessary.

kinkie:
as per discussion, dropping this

PR #1487, opened 2023-09-21
testheaders.sh: force-remove temporary files

kinkie:
At high levels of build parallelism on systems using GNU coreutils,
sometimes "make check" hangs on a request to confirm
removal of a read-only file temporary file.

Force-remove temporary test files to ensure removal is noniteractive.

### DIFF
--- a/test-suite/testheaders.sh
+++ b/test-suite/testheaders.sh
@@ -41,7 +41,7 @@ EOF
             echo "Fail."
             exitCode=1
         fi
-        rm $t.cc $t.o
+        rm -f $t.cc $t.o
     fi
     test $exitCode -eq 0 || break
 done


test-suite/testheaders.sh
```diff
@@ -41,7 +41,7 @@ EOF
             echo "Fail."
             exitCode=1
         fi
-        rm $t.cc $t.o
+        rm -f $t.cc $t.o
```
rousskov, line 44: > "make check" hangs on a request to confirm removal of a read-only file temporary file.

What makes a temporary file read-only?

kinkie: Automake internals. I haven't investigated the causes, only the symptoms, and this whole script is going to be removed as part of PR #1463

rousskov (approved):

PR #1489, opened 2023-09-22
Converted helperOpenServers() to a virtual Helper::Client member

eduard-bagdasaryan:
addressing a TODO.

No Squid functionality changes are expected.

### DIFF
--- a/src/helper.cc
+++ b/src/helper.cc
@@ -193,7 +193,7 @@ helper_stateful_server::~helper_stateful_server()
 }
 
 void
-helperOpenServers(const Helper::Client::Pointer &hlp)
+Helper::Client::openSessions()
 {
     char *s;
     char *progname;
@@ -208,6 +208,8 @@ helperOpenServers(const Helper::Client::Pointer &hlp)
     int wfd;
     void * hIpc;
     wordlist *w;
+    // Helps reducing diff. TODO: remove
+    const auto hlp = this;
 
     if (hlp->cmdline == nullptr)
         return;
@@ -327,18 +329,15 @@ helperOpenServers(const Helper::Client::Pointer &hlp)
     helperKickQueue(hlp);
 }
 
-/**
- * DPW 2007-05-08
- *
- * helperStatefulOpenServers: create the stateful child helper processes
- */
 void
-helperStatefulOpenServers(const statefulhelper::Pointer &hlp)
+statefulhelper::openSessions()
 {
     char *shortname;
     const char *args[HELPER_MAX_ARGS+1]; // save space for a NULL terminator
     char fd_note_buf[FD_DESC_SZ];
     int nargs = 0;
+    // Helps reducing diff. TODO: remove
+    const auto hlp = this;
 
     if (hlp->cmdline == nullptr)
         return;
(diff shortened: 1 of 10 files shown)


src/helper.cc
```diff
@@ -891,38 +890,30 @@ Helper::Client::handleFewerServers(const bool madeProgress)
...
+    srv->parent->sessionClosed(*srv);
     srv->dropQueued();
-
     delete srv;
 }
 
-// XXX: Almost duplicates Helper::Session::HelperServerClosed() because helperOpenServers() is not a virtual method of the `Helper::Client` class
-// TODO: Fix the `Helper::Client` class hierarchy to use virtual functions.
+// XXX: Essentially duplicates Helper::Session::HelperServerClosed() until we add SessionBase::helper().
 void
```
rousskov, line 912: This PR removes the previously disclosed blocker for removing this code duplication (i.e. the old XXX removed above). Unfortunately, we have discovered (the hard way!) another blocker (i.e. this new XXX). The new blocker is difficult to spot because both HelperServerClosed() method bodies have identical "spelling" now, but there is no `srv->helper` variable in the common base of the two `srv` classes (i.e. there is no SessionBase::helper data member).

Please do not ask to address this XXX in this minimal PR. I have checked, and adding helper() or a similar API will trigger many out-of-scope changes, some of which may be controversial. Let's continue to improve this messy helper code with small (but correct) steps.

rousskov (dismissed):
Thank you. I left a few comments in case others want to review this PR. I am approving without requesting any changes.

src/helper.cc
```diff
@@ -208,6 +208,8 @@ helperOpenServers(const Helper::Client::Pointer &hlp)
     int wfd;
     void * hIpc;
     wordlist *w;
+    // Helps reducing diff. TODO: remove
+    const auto hlp = this;
```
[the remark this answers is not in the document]
yadij: Porting should not be an issue as none of these refactors qualify for backport.

More importantly IMO, is that both of these `openSessions()` methods need a refactor to be better C++ logic and de-duplicate a lot of their actions. Authors choice of scope on whether to include that in-scope - it is clearly related to this PR change, but is NOT required to be done urgently.

src/helper.cc
```diff
@@ -1244,7 +1235,7 @@ Enqueue(Helper::Client * const hlp, Helper::Xaction * const r)
     /* do this first so idle=N has a chance to grow the child pool before it hits critical. */
     if (hlp->childs.needNew() > 0) {
         debugs(84, DBG_CRITICAL, "Starting new " << hlp->id_name << " helpers...");
```
yadij, line 1237: Please take advantage of the chance to drop this overly verbose `debugs()` - better info is displayed by the `openSessions()` methods.

```suggestion
```

src/helper.cc
```diff
@@ -1273,7 +1264,7 @@ StatefulEnqueue(statefulhelper * hlp, Helper::Xaction * r)
...
```
yadij, line 1266: Please take advantage of the chance to drop this overly verbose `debugs()` - better info is displayed by the `openSessions()` methods.

```suggestion
```

yadij (commented):
There are some possible additions within scope IMO. I do not insist on any of these though.

src/helper.cc
eduard-bagdasaryan: I am not sure whether the openSession() message is better, but there are few differences between them:
* this message level is 0 while the openSession()'s is 1
* openSession() prints nothing is hlp->cmdline is nil
* hlp->id_name is a helper category, while openSession() prints program name instead.

src/helper.cc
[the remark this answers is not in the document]
rousskov: > More importantly IMO, is that both of these `openSessions()` methods need a refactor to be better C++ logic and de-duplicate a lot of their actions. Authors choice of scope on whether to include that in-scope

Agreed.

> Porting should not be an issue as none of these refactors qualify for backport.

Distinguishing genuine changes from noise becomes a porting issue when the targeted branch has also modified the same code. This may not matter for backporting to other official Squid branch, but porting activities are not limited to those use cases.

src/helper.cc
rousskov: @eduard-bagdasaryan, you are right, but this line is 95% noise, and we will be better off without it. We can improve openSession() debugging later as needed. I will merge the suggested change from GitHub.

rousskov (approved):

PR #1490, opened 2023-09-22
Reduce helper callback code duplication

eduard-bagdasaryan:
No Squid functionality changes are expected.


src/helper.cc
```diff
@@ -1201,7 +1205,7 @@ helperStatefulHandleRead(const Comm::ConnectionPointer &conn, char *, size_t len
         if (r && cbdataReferenceValid(r->request.data)) {
             r->reply.finalize();
             r->reply.reservationId = srv->reservationId;
-            r->request.callback(r->request.data, r->reply);
+            hlp->callBack(*r);
```
rousskov, line 1208: By encapsulating code related to calling back the requestor, we were able to remove a few cbdataReferenceValid() and cbdataReferenceValidDone() calls, but we left in place a couple of them, including this one.

It is possible to get rid of this cbdataReferenceValid() call but doing so requires either calling finalize() in vain (which some might view as a performance regression for the affected cases) or refactoring code to ignore repeated attempts to finalize a response (which is a rather significant change, especially given the existence of at least three different response finalization approaches in the current code). Removing this check will also affect the existing non-trivial and controversial `called` logic (we actually tried to refactor that code and it quickly snowballed into several out-of-scope changes). Selecting the best path forward and implementing these improvements should be done in dedicated future PRs IMO.

src/helper.cc
```diff
@@ -1560,23 +1564,22 @@ Helper::Session::checkForTimedOutRequests(bool const retry)
...
         debugs(84, 2, "Request " << r->request.Id << " timed-out, remove it from queue");
-        void *cbdata;
         bool retried = false;
         if (retry && r->request.retries < MAX_RETRIES && cbdataReferenceValid(r->request.data)) {
             debugs(84, 2, "Retry request " << r->request.Id);
             ++r->request.retries;
             parent->submitRequest(r);
             retried = true;
-        } else if (cbdataReferenceValidDone(r->request.data, &cbdata)) {
+        } else if (cbdataReferenceValid(r->request.data)) {
```
rousskov, line 1573: This is another cbdataReferenceValid() left in the code, for very similar reasons.

The two cbdataReferenceValid() calls in this code can be merged into one by introducing a higher level `if` statement, but that would create a lot of PR noise, so this optimization is best avoided in this PR IMO.

src/helper.cc
```diff
@@ -123,18 +123,14 @@ Helper::SessionBase::closeWritePipeSafely(const char * const id_name)
 }
 
 void
-Helper::SessionBase::dropQueued()
+Helper::SessionBase::dropQueued(Client &client)
 {
     while (!requests.empty()) {
         // XXX: re-schedule these on another helper?
```
rousskov, line 129: Supplying Client info to this method is inline with addressing that potential (and controversial) XXX. Most likely, this code will be eventually refactored to extract queued requests away from this dying session, but that refactoring is outside this PR scope, of course.

rousskov (approved):
Thank you. I left a few comments in case others want to review this PR. I am approving without requesting any changes.

src/helper.cc
rousskov: Glad we agree that those cbdataReferenceValid() calls should be left as is in this PR. I hope you can approve this PR based on that agreement.

FWIW, replacing raw cbdata-protected pointers with CbcPointer or equivalent is likely (e.g., during transition to async callbacks) but would not really change the essence of this problem: Either we continue to avoid unnecesasry helper work when the requester is gone (which requires checking whether the requester is gone) OR we simplify code (by removing requestor checks that are essentially performance optimizations). We will determine the best way forward when we have to. This PR does not need to solve this particular puzzle.

rousskov (approved):

PR #1492, opened 2023-09-24
Y2038: Use time_t for commSetConnTimeout() timeout parameter

kinkie:
Change commSetConnTimeout() "timeout" parameter from int to time_t, to
match the common caller type and improve Year 2038-safety on systems
with 32-bit int.

Detected by Coverity. CID 1545129: Use of 32-bit time_t (Y2K38_SAFETY).

### DIFF
--- a/src/comm.cc
+++ b/src/comm.cc
@@ -592,7 +592,7 @@ commUnsetFdTimeout(int fd)
 }
 
 int
-commSetConnTimeout(const Comm::ConnectionPointer &conn, int timeout, AsyncCall::Pointer &callback)
+commSetConnTimeout(const Comm::ConnectionPointer &conn, time_t timeout, AsyncCall::Pointer &callback)
 {
     debugs(5, 3, conn << " timeout " << timeout);
     assert(Comm::IsConnOpen(conn));
--- a/src/dns_internal.cc
+++ b/src/dns_internal.cc
@@ -833,7 +833,7 @@ idnsDoSendQueryVC(nsvc *vc)
     vc->busy = 1;
 
     // Comm needs seconds but idnsCheckQueue() will check the exact timeout
-    const int timeout = (Config.Timeout.idns_query % 1000 ?
+    const time_t timeout = (Config.Timeout.idns_query % 1000 ?
                          Config.Timeout.idns_query + 1000 : Config.Timeout.idns_query) / 1000;
     AsyncCall::Pointer nil;
 
--- a/src/dns_internal.cc
+++ b/src/dns_internal.cc
@@ -834,7 +834,7 @@ idnsDoSendQueryVC(nsvc *vc)
 
     // Comm needs seconds but idnsCheckQueue() will check the exact timeout
     const auto timeout = (Config.Timeout.idns_query % 1000 ?
-                         Config.Timeout.idns_query + 1000 : Config.Timeout.idns_query) / 1000;
+                          Config.Timeout.idns_query + 1000 : Config.Timeout.idns_query) / 1000;
     AsyncCall::Pointer nil;
 
     commSetConnTimeout(vc->conn, timeout, nil);
--- a/src/helper.cc
+++ b/src/helper.cc
@@ -1600,7 +1600,8 @@ Helper::Session::requestTimeout(const CommTimeoutCbParams &io)
                                      CommTimeoutCbPtrFun(Session::requestTimeout, srv));
 
     const time_t timeSpent = srv->requests.empty() ? 0 : (squid_curtime - srv->requests.front()->request.dispatch_time.tv_sec);
-    const auto timeLeft = max(static_cast<time_t>(1), srv->parent->timeout - timeSpent);
+    const time_t minimumNewTimeout = 1; // second
+    const auto timeLeft = max(minimumNewTimeout, srv->parent->timeout - timeSpent);
 
     commSetConnTimeout(io.conn, timeLeft, timeoutCall);
 }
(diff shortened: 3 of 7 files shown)


src/dns_internal.cc
```diff
@@ -833,7 +833,7 @@ idnsDoSendQueryVC(nsvc *vc)
     vc->busy = 1;
 
     // Comm needs seconds but idnsCheckQueue() will check the exact timeout
-    const int timeout = (Config.Timeout.idns_query % 1000 ?
+    const time_t timeout = (Config.Timeout.idns_query % 1000 ?
```
rousskov, line 836: The actual type here is currently time_msec_t rather than time_t. Those two types are rather different (e.g., only one of them is _unsigned_ in most environments). If the compiler is happy, I suggest removing this type conversion from _this_ line because this code is not written to handle that conversion correctly:

```suggestion
    const auto timeout = (Config.Timeout.idns_query % 1000 ?
```

The above will also improve code consistency: There are quite a few similar `auto` lines in the existing code. I found only one other line that uses an explicit time_t variable like this PR code does.

src/dns_internal.cc
```diff
@@ -833,7 +833,7 @@ idnsDoSendQueryVC(nsvc *vc)
...
                          Config.Timeout.idns_query + 1000 : Config.Timeout.idns_query) / 1000;
```
rousskov, line 837: Please also consider fixing Ipc::UdsOp::setTimeout() which forwards an `int` timeout argument to commSetConnTimeout().

rousskov, line 837: If you are fixing this caller, please also fix a very similar `int` caller in src/helper.cc near line 1603:

```C++
const int timeLeft = max(1, (static_cast<int>(srv->parent->timeout) - timeSpent));
```

The explicit cast here is probably meant to make the max() template compile. There are several ways to adjust that, including something like that:

```C++
const time_t minimumNewTimeout = 1; // second
const auto timeLeft = max(minimumNewTimeout, srv->parent->timeout - timeSpent);
```

rousskov (changes requested):
Thank you for improving this code.

We are using negative commSetConnTimeout() timeout values while time_t is not, technically, guaranteed to be signed. However, this PR can still be considered a step forward even though it does not solve that (unlikely to affect us in real world) problem because nearly all commSetConnTimeout() callers do use time_t already.

rousskov:
I adjusted PR title/description, primarily to fix a few minor technical issues and typos. Please adjust further as needed.

src/dns_internal.cc
kinkie: Thanks for the advice. Accepting

kinkie: Thanks for catching this, it had slipped past me. Fixing.

rousskov: > I'd rather static_cast<time_t>(1) ; is there any reason not to?

I prefer the two-variable suggestion for two reasons:

* It is arguably a lot more readable.
* It documents the effect of having that `1` in std::max().

I do not insist on my variation.

src/helper.cc
```diff
@@ -1599,8 +1599,8 @@ Helper::Session::requestTimeout(const CommTimeoutCbParams &io)
     AsyncCall::Pointer timeoutCall = commCbCall(84, 4, "Helper::Session::requestTimeout",
                                      CommTimeoutCbPtrFun(Session::requestTimeout, srv));
 
-    const int timeSpent = srv->requests.empty() ? 0 : (squid_curtime - srv->requests.front()->request.dispatch_time.tv_sec);
-    const int timeLeft = max(1, (static_cast<int>(srv->parent->timeout) - timeSpent));
+    const time_t timeSpent = srv->requests.empty() ? 0 : (squid_curtime - srv->requests.front()->request.dispatch_time.tv_sec);
+    const time_t timeLeft = max(static_cast<time_t>(1), srv->parent->timeout - timeSpent);
```
rousskov, line 1603:
```suggestion
    const auto timeLeft = max(static_cast<time_t>(1), srv->parent->timeout - timeSpent);
```

rousskov (approved):

rousskov:
I adjusted PR description again to use the original/shorter version of the Y2K38_SAFETY defect wording. I did not have access to that version earlier because Coverity service was down. Sorry.

yadij:
RFC: long-term we are better to be moving to the `std::chrono` types instead of `time_t`. Is it possible to start that conversion here without major trouble?

rousskov:
> long-term we are better to be moving to the `std::chrono` types instead of `time_t`.

Yes, of course.

> Is it possible to start that conversion here without major trouble?

I think the answer depends, in part, on whether we want to start supporting sub-second timeouts. I cannot be sure without studying this issue, but I suspect that we do. If we do, then, for commSetConnTimeout(), the migration to std::chrono should probably start at the timeout storing/handling code, not its users (that this PR is focusing on) because while it is possible to safely convert time_t to the right std::chrono type, it is not possible to safely convert any sub-second std::chrono type to time_t.

FWIW, my "use PackableStream to print time_t" suggestions in https://github.com/squid-cache/squid/pull/1493#pullrequestreview-1642676855 and https://github.com/squid-cache/squid/pull/1494#pullrequestreview-1642656295 _are_ tiny steps towards that std::chrono conversion.

src/comm.cc
```diff
@@ -592,7 +592,7 @@ commUnsetFdTimeout(int fd)
 }
 
 int
-commSetConnTimeout(const Comm::ConnectionPointer &conn, int timeout, AsyncCall::Pointer &callback)
+commSetConnTimeout(const Comm::ConnectionPointer &conn, time_t timeout, AsyncCall::Pointer &callback)
```
rousskov, line 595: I did not realize this by looking at the PR diff, but I now see two more problems related to this (merged) PR scope:

* This function and commUnsetConnTimeout() return type should be `time_t` rather than `int`.
* The C-style `timeout` cast inside this function should be removed.

@kinkie, if you can fix these problems, please do.

The latest Coverity report (defect 1547031) triggered this comment.

PR #1493, opened 2023-09-25
Y2038: improve printing of time settings

kinkie:
Avoid truncation errors when printing time_t-based squid.conf directives
on platforms with 32-bit int and 64-bit time_t. Also avoid similar
errors when printing time_msec-based directives on platforms with 32-bit
int.

Detected by Coverity. CID 1529622: Use of 32-bit time_t (Y2K38_SAFETY).

### DIFF
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -2950,7 +2950,8 @@ parse_TokenOrQuotedString(char **var)
 static void
 dump_time_t(StoreEntry * entry, const char *name, time_t var)
 {
-    storeAppendPrintf(entry, "%s %ld seconds\n", name, static_cast<long>(var));
+    PackableStream os(*entry);
+    os << name << ' ' << var << " seconds\n";
 }
 
 void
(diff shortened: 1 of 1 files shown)


src/cache_cf.cc
```diff
@@ -2950,7 +2950,7 @@ parse_TokenOrQuotedString(char **var)
 static void
 dump_time_t(StoreEntry * entry, const char *name, time_t var)
 {
-    storeAppendPrintf(entry, "%s %d seconds\n", name, (int) var);
+    storeAppendPrintf(entry, "%s %ld seconds\n", name, static_cast<long>(var));
```
rousskov, line 2953: I am not excited about this change because it makes the problem more difficult to see rather than fixing it: `long int` is [guaranteed](https://en.cppreference.com/w/cpp/language/types) to have 32 bits, not 64 bits as this fix (description) may want us to believe. I made the same mistake in commit 5ef3cbb, but I did not have Coverity results to help me see the light.

Want a proper minimal fix? Cast to and printf() int64_t.

Want to stop kicking the can down the road? Convert dump_time_t() to use PackableStream instead. See SBufStatsAction::dump(), dump_refreshpattern(), Mem::Stats(), and similar functions for examples.

rousskov:
I adjusted PR description to better match the proposed fix. Please adjust further as needed.

rousskov (approved):
Thank you for improving this fix. I have adjusted PR description again to better match the proposed fix. Please adjust further as needed.

kinkie:
Works for me, let's ship this

squidadm:
queued for backport to v6

PR #1494, opened 2023-09-25
Y2038: Fix cache_peer connect-timeout reporting

kinkie:
... on systems with 64-bit time_t and 32-bit int.

Also fixed `htcp=oldsquid` reporting broken since 2015 commit 4ac1880.

This fix affects mgr:config and mgr:server_list cache manager reports.

### DIFF
--- a/src/neighbors.cc
+++ b/src/neighbors.cc
@@ -1,3 +1,4 @@
+
 /*
  * Copyright (C) 1996-2023 The Squid Software Foundation and contributors
  *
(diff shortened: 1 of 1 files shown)


src/neighbors.cc
```diff
@@ -1453,7 +1453,7 @@ dump_peer_options(StoreEntry * sentry, CachePeer * p)
         storeAppendPrintf(sentry, " ttl=%d", p->mcast.ttl);
 
     if (p->connect_timeout_raw > 0)
-        storeAppendPrintf(sentry, " connect-timeout=%d", (int)p->connect_timeout_raw);
+        storeAppendPrintf(sentry, " connect-timeout=%ld", p->connect_timeout_raw);
```
rousskov, line 1456: I am not excited about this change because it makes the problem more difficult to see rather than fixing it: `long int` is [guaranteed](https://en.cppreference.com/w/cpp/language/types) to have 32 bits, not 64 bits as this fix (description) may want us to believe. I made the same mistake in commit 5ef3cbb, but I did not have Coverity results to help me see the light.

Want a better minimal fix? Cast to and printf() int64_t.

Want to stop kicking the can down the road? Convert the entire dump_peer_options() to use PackableStream instead. See SBufStatsAction::dump(), dump_refreshpattern(), Mem::Stats(), and similar functions for examples.

kinkie: done

src/neighbors.cc
```diff
@@ -1367,131 +1367,133 @@ neighborDumpPeers(StoreEntry * sentry)
...
         if (p->options.htcp_oldsquid || p->options.htcp_no_clr || p->options.htcp_no_purge_clr || p->options.htcp_only_clr) {
             bool doneopts = false;
             if (p->options.htcp_oldsquid) {
-                storeAppendPrintf(sentry, "oldsquid");
+                os << "oldsquid";
                 doneopts = true;
             }
             if (p->options.htcp_no_clr) {
-                storeAppendPrintf(sentry, "%sno-clr",(doneopts?",":"="));
+                os << (doneopts ? ',' : '=') << "no-clr";
```
rousskov, line 1428: Instead of error-prone hacks like doneopts, we should add a Delimiter stream manipulator and use it in contexts like this. I did not do it in this PR, but I am not against including it in this PR.

src/neighbors.cc
```diff
@@ -1367,131 +1367,133 @@ neighborDumpPeers(StoreEntry * sentry)
...
 
 #if USE_HTCP
     if (p->options.htcp) {
-        storeAppendPrintf(sentry, " htcp");
+        os << " htcp";
         if (p->options.htcp_oldsquid || p->options.htcp_no_clr || p->options.htcp_no_purge_clr || p->options.htcp_only_clr) {
             bool doneopts = false;
             if (p->options.htcp_oldsquid) {
-                storeAppendPrintf(sentry, "oldsquid");
+                os << "oldsquid";
```
rousskov, line 1424: This reporting is broken in the official code as well (missing `=`). I did not fix it, but I am not against adding this fix to this PR scope, especially if that fix is added in this form (that would improve code safety during refactoring and facilitate future code improvements discussed in another review comment):

```suggestion
                os << (doneopts ? ',' : '=') << "oldsquid";
```

rousskov (approved):
I fixed one PR bug (commit eda290d) and polished, including PR title/description. Please review and adjust further as needed.

rousskov:
```C++
            p->connect_timeout_raw = xatoi(token + 16);
```

AFAICT, cache_peer connect-timeout _parsing_ is similarly broken. Fixing that can be considered outside this PR scope, but I am not against adding that fix to this PR. Your call.

src/neighbors.cc
kinkie: I ported like-for-like, but I agree with fixing this

src/neighbors.cc
kinkie: I'd prefer to not expand scope, it is a thing we do way too often. We might add an XXX if you want, and I'll put that idea in my queue

kinkie:
> ```c++
>             p->connect_timeout_raw = xatoi(token + 16);
> ```
> 
> AFAICT, cache_peer connect-timeout _parsing_ is similarly broken. Fixing that can be considered outside this PR scope, but I am not against adding that fix to this PR. Your call.

Let's keep this PR as a single focus. Added fixing the parsing to my TODO list

src/neighbors.cc
rousskov: This is a TODO rather than an XXX, even though the lack of this feature probably triggered a bug in the official code (the bug you fixed in recent PR commit 4fcf7ce).

To avoid a misunderstanding, I am (still) not suggesting or asking for any changes in this PR, not even the addition of that TODO.

rousskov (approved):

src/neighbors.cc
```diff
@@ -1367,131 +1367,133 @@ neighborDumpPeers(StoreEntry * sentry)
...
 
 #if USE_HTCP
     if (p->options.htcp) {
-        storeAppendPrintf(sentry, " htcp");
+        os << " htcp";
         if (p->options.htcp_oldsquid || p->options.htcp_no_clr || p->options.htcp_no_purge_clr || p->options.htcp_only_clr) {
             bool doneopts = false;
             if (p->options.htcp_oldsquid) {
-                storeAppendPrintf(sentry, "oldsquid");
+                os << (doneopts ? ',' : '=') << "oldsquid";
```
rousskov, line 1424: @kinkie, FYI: I added the following paragraphs to the PR description to disclose this fix:

Also fixed `htcp=oldsquid` reporting broken since 2015 commit 4ac1880.

rousskov: BTW, this fix should resurrect Coverity defect 740350 (logically dead code). If that happens, we should classify that new/resurrected defect as "intentional". It will go away with the introduction of Delimiter or a similar manipulator.

kinkie: yes

kinkie:
v6 backporting doesn't apply cleanly. Aborting backport

PR #1496, opened 2023-09-26
Do not use raw pointers to index userhash CachePeers

eduard-bagdasaryan:
Simplified and improved code safety by using CbcPointers for userhash
cache_peers, as we have done for CARP peers in recent commit e7959b5.

Also fixed mgr:userehash Cache Manager reports to detail relevant
cache_peers instead of all cache_peers. This problem existed since
inception (2008 commit f7e1d9c) as detailed in recent commit e7959b5.


rousskov (approved):
> Similar sourcehash changes are already in master [2d5ac43](https://github.com/squid-cache/squid/commit/2d5ac435c3bc120b6ddb3458d6360139b09eb891).

Approving after superficial review, based on the above statement.

yadij (approved):

PR #1498, opened 2023-10-01
RFC 9112: Improve HTTP chunked encoding compliance

yadij:


yadij:
@rousskov, please check the functionality-tests is correct given the change this PR causes to chunked encoding error handling.

rousskov:
> please check the functionality-tests is correct

Will do.

src/http/one/TeChunkedParser.cc
```diff
@@ -121,7 +126,7 @@ Http::One::TeChunkedParser::parseChunkMetadataSuffix(Tokenizer &tok)
     // bad or insufficient input, like in the code below. TODO: Expand up.
     try {
         parseChunkExtensions(tok); // a possibly empty chunk-ext list
-        skipLineTerminator(tok);
+        tok.skipRequired("CRLF after [chunk-ext]", Http1::CrLf());
```
rousskov, line 129: > please check the functionality-tests is correct

Done. I believe our CI tests revealed a bug in PR changes. You should be able to trigger the same bug using pretty much any basic chunked-encoded message. FWIW, the failing test was using a 15-byte (before encoding) chunk-encoded body like this:

```
f\r\n
body-bytes-here\r\n
```

The _combination_ of old parseChunkExtensions() and new tok.skipRequired() does not work because parseChunkExtensions() consumes any BWS. In case of a simple canonical chunk, parseChunkExtensions() consumes the leading CR in the chunk-terminating CRLF (that PR code now _requires_ "after [chunk-ext]"). That innocent CR looks like BWS to Http::One::ParseBws().

To fix this, we should fix parseChunkExtensions() (at least) to consume nothing but chunk extensions. Those extensions may include BWS, of course, but if we want to be strict in its caller, then parseChunkExtensions()[^1] cannot consume BWS that is _not_ a part of a chunk extension. 

The same restrictions apply to any code that goes before the new tok.skipRequired() calls -- that code can no longer consume CR from the terminating CRLF.

[^1]: parseChunkExtensions() itself and the code parseChunkExtensions() calls, of course.

rousskov (changes requested):

src/http/one/TeChunkedParser.cc
yadij: Ah. Thank you. I have adjusted the parser in [4a39435](https://github.com/squid-cache/squid/pull/1498/commits/4a39435cf8568fabae2bbb8c84d753569bf04918) to also be more strict about the BWS it consumes from those two places, no longer skipping the relaxed whitespace characters should meet the need without re-opening the bugs handling IBM server and ICAP responses.

src/http/one/TeChunkedParser.cc
```diff
@@ -132,13 +137,22 @@ Http::One::TeChunkedParser::parseChunkMetadataSuffix(Tokenizer &tok)
...
+        // Bug 4492: IBM_HTTP_Server sends SP after chunk-size
+        // causing RFC 7230 Errata #4667
+
+        // Use strict BWS definition here to avoid consuming
+        // from our callers "[chunk-ext] CRLF" sequence when
+        // using the relaxed_header_parser.
+        if (const auto count = tok.skipAll(CharacterSet::WSP)) {
+            // Generating BWS is a MUST-level violation so warn about it as needed.
+            debugs(33, ErrorLevel(), "found " << count << " BWS octets");
+        }
```
rousskov, line 156: Please do change portions of the existing grammar/code "to avoid consuming from our callers". The latest PR attempt (i.e. PR commit 4a39435cf8568fabae2bbb8c84d753569bf04918) is buggy AFAICT -- it does not cover some cases -- but it is the whole _approach_ that is flawed. We should change the approach rather than chase and fix individual PR bugs (and then wait for similar bugs to appear in future code).

Official Http::One::TeChunkedParser::parseChunkExtensions() is buggy: It promises to parse `chunk-ext` (as defined by the corresponding ABNF), but it also parses other sequences (e.g., `BWS`). This PR should fix _that_ bug. It currently does not (it replaces some `BWS` with other sequences and increases code complexity/duplication, but the underlying bug remains unaddressed).

I suggest renaming parseChunkExtensions() parameter from tok to callerTok and updating callerTok at the existing commit point (at the end of each loop iteration). All current PR changes inside this method should be removed. Here is how the fixed version of the method loop body would look like (GitHub does not allow me to post this as a suggestion):

```C++
        auto tok = callerTok;

        ParseBws(tok); // Bug 4492: IBM_HTTP_Server sends SP after chunk-size

        if (!tok.skip(';'))
            return; // reached the end of extensions (if any)

        parseOneChunkExtension(tok);
        buf_ = tok.remaining(); // got one extension
        callerTok = tok;
```

In the PR diff, you should see no loop changes except three new lines: two at the top (including an empty line) and one at the bottom of the loop body code.

Http::One::TeChunkedParser::parseOneChunkExtension() has a very similar bug and needs a very similar fix (instead of the current PR changes).

I did not check other TeChunkedParser methods. Please do so.

rousskov (changes requested):

src/http/one/TeChunkedParser.cc
yadij: FTR; the issue currently is that "relaxed" `ParseBws()` consumes characters outside the [`BWS` definition](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.3).
The software violating the RFC sender requirements is known to use `SP` so we can be strict with these syntax tolerances. Remember the goal behind this PR is to increase standard compliance, not reduce it.

rousskov: > FTR; the issue currently is that "relaxed" `ParseBws()` consumes characters outside the [`BWS` definition](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.3).

While there are many "current issues" that can be discussed (in their proper contexts), _this_ change request is about TeChunkedParser problems that are _not_ tied to any specific ParseBws() implementation. Addressing this change request does not imply, does not require, and cannot be done by changing ParseBws() (or replacing ParseBws() calls with ad hoc code that follows RFC definition of BWS).

> Remember the goal behind this PR is to increase standard compliance, not reduce it.

Addressing this change request will fix bugs without reducing standard compliance.

I recommend carefully rereading the change request with the intent to understand the actual problem it describes and the solution it outlines. Neither are related to the first response to this change request (partially quoted above).

yadij: Done by [8caa384](https://github.com/squid-cache/squid/pull/1498/commits/8caa3843216b26b6028ae0b40e83acc34db2583b)

src/http/one/TeChunkedParser.cc
```diff
@@ -158,8 +166,10 @@ Http::One::ChunkExtensionValueParser::Ignore(Tokenizer &tok, const SBuf &extName
...
 /// chunk-ext = *( BWS ";" BWS chunk-ext-name [ BWS "=" BWS chunk-ext-val ] )
 void
-Http::One::TeChunkedParser::parseOneChunkExtension(Tokenizer &tok)
+Http::One::TeChunkedParser::parseOneChunkExtension(Tokenizer &callerTok)
 {
+    auto tok = callerTok;
+
     ParseBws(tok); // Bug 4492: ICAP servers send SP before chunk-ext-name
 
     const auto extName = tok.prefix("chunk-ext-name", CharacterSet::TCHAR);
```
rousskov, line 175: We must also update callerTok for valueless chunk-ext (see an early `return` a few lines below). According to the chunk-ext ABNF, BWS after chunk-ext-name may not belong to chunk-ext[^1], so we have to update callerTok here, after successfully parsing chunk-ext-name (and before unconditionally consuming BWS). Here is one way to do that:

[^1]: BWS after chunk-ext-name does not belong to chunk-ext if this chunk-ext has no optional `BWS "=" BWS chunk-ext-val` component.

```suggestion
    const auto extName = tok.prefix("chunk-ext-name", CharacterSet::TCHAR);
    callerTok = tok; // in case we determine that this is a valueless chunk-ext
```

rousskov (changes requested):
Only one "excessive BWS consumption" bug left AFAICT. Fortunately, it is easy to fix.

rousskov (approved):

yadij:
@rousskov, @kinkie - sorry guys the last minute conflict fix and changelog update reset the review approvals. Would either of you mind re-approving for fast track please (and can then drop the other re-review requested). Thank you.

kinkie (dismissed):

rousskov (changes requested):

ChangeLog
```diff
@@ -5,6 +5,7 @@ Changes in squid-6.4 (15 Oct 2023):
 	- Bug 5301: cachemgr.cgi not showing new manager interface URLs
 	- Bug 5300: cachemgr.cgi assertion
 	- HTTP: Improve handling of empty lines received prior to request-line
+	- RFC 9112: Improve HTTP chunked encoding compliance
```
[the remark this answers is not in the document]
yadij: With the constant delays this PR has been facing PR #1510 has already gone in first.

rousskov (approved):

kinkie (approved):

squidadm:
queued for backport to v6

yadij:
queued for backport to v5

PR #1504, opened 2023-10-09
Recognize internal requests created by adaptation/redirection

eduard-bagdasaryan:
Before this fix, Squid set flags.internal for virgin requests but not
for adapted/redirected requests, leaving post-adaptation request
processing code in an inconsistent state, forwarding the
internalCheck()-compliant adapted/redirected requests as regular
requests, triggering forwarding loops and complicating code refactoring.

### DIFF
--- a/src/client_side_request.cc
+++ b/src/client_side_request.cc
@@ -1623,6 +1623,13 @@ ClientHttpRequest::initRequest(HttpRequest *aRequest)
 }
 
 void
+ClientHttpRequest::resetRequest(HttpRequest *newRequest)
+{
+    const auto uriChanged = request->effectiveRequestUri() != newRequest->effectiveRequestUri();
+    resetRequestXXX(newRequest, uriChanged);
+}
+
+void
 ClientHttpRequest::resetRequestXXX(HttpRequest *newRequest, const bool uriChanged)
 {
     assert(request != newRequest);
@@ -1990,8 +1997,7 @@ ClientHttpRequest::handleAdaptedHeader(Http::Message *msg)
     assert(msg);
 
     if (HttpRequest *new_req = dynamic_cast<HttpRequest*>(msg)) {
-        const auto uriChanged = request->effectiveRequestUri() != new_req->effectiveRequestUri();
-        resetRequestXXX(new_req, uriChanged);
+        resetRequest(new_req);
         assert(request->method.id());
     } else if (HttpReply *new_rep = dynamic_cast<HttpReply*>(msg)) {
         debugs(85,3, "REQMOD reply is HTTP reply");
--- a/src/client_side_request.cc
+++ b/src/client_side_request.cc
@@ -1222,9 +1222,6 @@ ClientRequestContext::clientRedirectDone(const Helper::Reply &reply)
                     new_request->url = tmpUrl;
                     debugs(61, 2, "URL-rewriter diverts URL from " << old_request->effectiveRequestUri() << " to " << new_request->effectiveRequestUri());
 
-                    // update the new request to flag the re-writing was done on it
-                    new_request->flags.redirected = true;
-
                     // unlink bodypipe from the old request. Not needed there any longer.
                     if (old_request->body_pipe != nullptr) {
                         old_request->body_pipe = nullptr;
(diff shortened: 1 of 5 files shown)


yadij (commented):
IMO this whole `flags.internal` mechanism should be dropped. The logic this PR moves into `ClientHttpRequest::checkForInternalAccess()` should be moved to `FwdState::Start()` where the `internalStart()` is actually performed - after all such things like adaptation.

rousskov:
> IMO this whole `flags.internal` mechanism should be dropped.

The flag should be dropped if the result of the corresponding (increasingly expensive) calculations is used just once (e.g., in FwdState::Start()). That consolidation of use cases has not happened yet. I doubt it will ever happen (in part, for the reasons detailed below). If it does happen, the flag will be dropped.

> `ClientHttpRequest::checkForInternalAccess()` [logic] should be moved to `FwdState::Start()` where the `internalStart()` is actually performed - after all such things like adaptation.

It is not possible to confine checkForInternalAccess() logic to FwdState::Start() because internal request decisions are made (and must continue to be made) both pre-cache and post-cache. FwdState is post-cache.

FWIW, we are finishing work on another minimal PR that will address an old XXX and clarify these internal request decisions further.

src/client_side_request.cc
```diff
@@ -1222,17 +1223,14 @@ ClientRequestContext::clientRedirectDone(const Helper::Reply &reply)
                     new_request->url = tmpUrl;
```
rousskov, line 1223:
```C++
// prevent broken helpers causing too much damage. If old URL == new URL skip the re-write.
if (urlNote != nullptr && strcmp(urlNote, http->uri))
```

@yadij, this comment/discussion is outside this minimal PR scope, but it may be useful to know what can be done with the above URI comparison long-term. As you know, this code is comparing from-helper URI (`urlNote`) and `http->uri` to detect URI changes. ICAP/eCAP adaptation code (e.g., resetRequest() in this PR) is comparing two effectiveRequestUri()s. AFAICT, the two comparison methods ought to yield different results in some corner cases. Do you think this redirector code can eventually be switched to comparing effectiveRequestUri()s instead?

FWIW, this code is already _reporting_ effectiveRequestUri() difference in the debugs() below, not the difference between URIs actually being compared above...

rousskov (approved):

yadij:
Removing merge clearance because this PR has not actually passed the process criteria https://wiki.squid-cache.org/MergeProcedure#votes

rousskov:
> Removing merge clearance because this PR has not actually passed the process criteria https://wiki.squid-cache.org/MergeProcedure#votes

Anubis does not merge PRs that "have not actually passed process criteria", regardless of `M-cleared-for-merge` label presence. That label does not violate any merging rules we have.

Removing `M-cleared-for-merge` label creates extra work for humans (finding PRs that should be labeled) and extra delays (ready PRs are not getting merged, waiting for humans to find them). As a partial workaround, I used to apply `S-waiting-for-committer` label to reduce that extra "which PRs should I label?" work and associated delays. You have asked me not to do that, and I stopped.

@yadij, please do not remove `M-cleared-for-merge` labels set by others if you are not blocking a PR (with a negative vote or a request for review).

rousskov:
Per meeting discussion, I have added a `S-could-use-an-approval` label (for us mortals) and restored the `M-cleared-for-merge` label (for Anubis).

kinkie (approved):

src/client_side_request.cc
yadij: IIRC that use of `http->uri` was an optimization to get the URL from way back when `HttpRequest::uri` has issues with being partial vs absolute vs canonical. Correct code today would always do the `tmpUrl.parse(urlNote);` and only do the performance optimizing skip `if (tmpUrl != old_request->uri)`

As the comment implies this whole condition is (almost) purely performance fix for badly coded helpers. We are just trying to avoid the CPU cycles it takes to clone everything (with possible associated state loss from bugs).

PS. Long term I hope that `http->uri` can disappear completely - we should always be referring to whatever `HttpRequest` is "active" (virgin vs adapted vs redirected).

PR #1505, opened 2023-10-10
Maintenance: fix outdated stub_errorpage.cc contents

yadij:


kinkie (approved):

src/tests/stub_errorpage.cc
```diff
@@ -7,17 +7,31 @@
...
+ErrorState *ErrorState::NewForwarding(err_type, HttpRequestPointer &, const AccessLogEntryPointer &) STUB_RETVAL(nullptr)
+HttpReply *ErrorState::BuildHttpReply(void) STUB_RETVAL(nullptr)
+void ErrorState::validate() STUB
+void errorInitialize(void) STUB
+void errorClean(void) STUB
+void errorSend(const Comm::ConnectionPointer &, ErrorState *) STUB
 void errorAppendEntry(StoreEntry *, ErrorState * ) STUB
-bool strHdrAcptLangGetItem(const String &, char *, int, size_t &) STUB_RETVAL(false)
-void TemplateFile::loadDefault() STUB
+err_type errorReservePageId(const char *, const SBuf &) STUB_RETVAL(err_type(0))
```
rousskov, line 28: Why move these three unchanged definitions? If possible, please keep their positions unchanged (from diff point of view).

rousskov (dismissed):

src/tests/stub_errorpage.cc
yadij: Side effect of updating the file to the latest stub layout style. Each line in the stub is being kept in the same order as the matching .h it came from relative to the other lines - which in turn enables semi-automated stub creation.

src/tests/stub_errorpage.cc
```diff
@@ -7,17 +7,31 @@
  */
 
 #include "squid.h"
-#include "errorpage.h"
+#include "AccessLogEntry.h"
```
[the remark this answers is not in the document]
yadij: "invalid use of incomplete type" errors. `HttpReply.h` seems to pull it in somewhere anyway so we do not need this specially. Thanks for the catch.

src/tests/stub_errorpage.cc
rousskov: > Each line in the stub is being kept in the same order as the matching .h it came from relative to the other lines

I object to introducing that kind of requirement (for humans) until we automate stub maintenance.

src/tests/stub_errorpage.cc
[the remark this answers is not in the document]
rousskov: The header is still needed AFAICT:

```
In file included from ../../../src/http/one/forward.h:12,
                 from ../../../src/http/forward.h:12,
                 from ../../../src/store_key_md5.h:15,
                 from ../../../src/CacheDigest.h:15,
                 from ../../../src/globals.h:12,
                 from ../../../src/sbuf/SBuf.h:17,
                 from ../../../src/HttpBody.h:12,
                 from ../../../src/HttpReply.h:13,
                 from ../../../src/tests/stub_errorpage.cc:10:
../../../src/base/RefCount.h: In instantiation of ‘void RefCount<C>::dereference(const C*) [with C = AccessLogEntry]’:
../../../src/base/RefCount.h:42:9:   required from ‘RefCount<C>::~RefCount() [with C = AccessLogEntry]’
../../../src/tests/stub_errorpage.cc:17:96:   required from here
../../../src/base/RefCount.h:120:31: error: invalid use of incomplete type ‘const class AccessLogEntry’
  120 |         if (tempP_ && tempP_->unlock() == 0)
```

I would love to understand _why_ recent commit a7b75c6 does not protect us from these [errors](https://github.com/squid-cache/squid/actions/runs/6480816264/job/17597092743?pr=1505), but I cannot insist on developing that knowledge as a part of _this_ PR. CC: @kinkie.

You will need to re-add that header to pass build tests.

src/tests/stub_errorpage.cc
yadij: It is not a requirement. It is how I generate these stubs. This file was completely re-generated so its contents will not line up exactly with the ad-hoc human content that was previously there.

src/tests/stub_errorpage.cc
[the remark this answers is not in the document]
yadij: done in 8a3f94a9f542ed0a7cee8192a7511b832df6c4b4

src/tests/stub_errorpage.cc
rousskov: > Alex: Why move these three unchanged definitions? If possible, please keep their positions unchanged (from diff point of view).

> Amos: Side effect of updating the file to the latest stub layout style. Each line in the stub is being kept in the same order as the matching .h it came from relative to the other lines

> Alex: I object to introducing that kind of requirement (for humans) until we automate stub maintenance.

> Amos: It is not a requirement. It is how I generate these stubs.

Aha, so there is no relevant "latest stub layout style" as far as Project is concerned. There is some generator program that introduces unwanted changes into PRs like this one. A developer can use whatever tools they want, of course, but I think they should clean up after their tools. This change request is asking for such a cleanup: Please do not reorder unchanged stubs in PRs (even if your generator reorders unchanged stubs).

rousskov:
> Unless there's any reason to defer, I'm planning to clear this for merge tomorrow - it seems the key concerns were reportedly addressed, let's move forward :)

No objections. FWIW, merging without addressing my (secondary) concerns may encourage creation of similar problems in future PRs, which may, in turn, encourage blocking of those future PRs. I discourage interference in the form of merging problematic PRs because such 3rd party actions only complicate the problem space. Needless to say, if you want similar code changes in Squid, then you can post the corresponding non-problematic PR and get that merged in two days instead.

src/tests/stub_errorpage.cc
kinkie: I've restored the ordering. TBH I find these concerns do not meet the bar for relevance, but
that goes both ways so I've restored the previous state

kinkie (approved):

kinkie:
> > Unless there's any reason to defer, I'm planning to clear this for merge tomorrow - it seems the key concerns were reportedly addressed, let's move forward :)
> 
> No objections. FWIW, merging without addressing my (secondary) concerns may encourage creation of similar problems in future PRs, which may, in turn, encourage blocking of those future PRs. I discourage interference in the form of merging problematic PRs because such 3rd party actions only complicate the problem space. Needless to say, if you want similar code changes in Squid, then you can post the corresponding non-problematic PR and get that merged in two days instead.

Restored ordering of the lines, and merging.
I'd like to clean up stalled PRs, we have too many. Let's merge or close them if they have fundamental flaws

rousskov:
>  I'd like to clean up stalled PRs, we have too many.

Good luck! FWIW, I was not blocking this PR from going in for the last two months.

> Let's merge or close them if they have fundamental flaws.

"Merging as is" and "closing" are not the only two possible actions, of course. "Has fundamental flaws" and "has no fundamental flaws" are not the only meaningful characteristics of a PR (and defining the boundaries of a "fundamental flaw" is going to be very difficult). In fact, the vast majority of old PRs can be refactored to avoid fundamental flaws, but it often requires significant work, including conflict resolution.

PR #1502, opened 2023-10-08
Fix handling of zero cache_peers with --enable-cache-digests

eduard-bagdasaryan:
The bug was introduced in recent commit 2e24d0b.

PR #1506, opened 2023-10-10
Fix crash in neighborsDigestSelect

kinkie:
Fixed null pointer dereference in neighborsDigestSelect
which would manifest in case of USE_CACHE_DIGESTS 
and no configured cache_peers

### DIFF
--- a/src/neighbors.cc
+++ b/src/neighbors.cc
@@ -765,32 +765,34 @@ neighborsDigestSelect(PeerSelector *ps)
 
     storeKeyPublicByRequest(request);
 
-    for (size_t i = 0; i < Config.peers->size(); ++i) {
-        const auto p = &Config.peers->nextPeerToPing(i);
+    if (Config.peers) {
+        for (size_t i = 0; i < Config.peers->size(); ++i) {
+            const auto p = &Config.peers->nextPeerToPing(i);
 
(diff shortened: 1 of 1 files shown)


src/neighbors.cc
```diff
@@ -765,32 +765,34 @@ neighborsDigestSelect(PeerSelector *ps)
 
     storeKeyPublicByRequest(request);
 
-    for (size_t i = 0; i < Config.peers->size(); ++i) {
-        const auto p = &Config.peers->nextPeerToPing(i);
+    if (Config.peers) {
```
kinkie, line 768: Lots of formatting changes due to the need to indent a block. This is the only line actually added

kinkie:
run-tested

src/neighbors.cc
rousskov: Looks like a dupe of #1502 to me. I would go with #1502 instead of this noisier and slightly more dangerous change.

rousskov (changes requested):

kinkie:
Doh. #1502 had escaped me. Approving that, hopefully can be fast-tracked for merge then

rousskov:
> #1502 had escaped me. Approving that, hopefully can be fast-tracked for merge then

Yes, it was already on the fast track. Currently waiting for Jenkins tests of the staged commit...

PR #1507, opened 2023-10-11
Squid version 3.4 (and older) End-Of-Life

yadij:
We are about to hit the 10 year anniversary of v3.4
release. With all known downstream OS vendors now
distributing at least Squid version 3.5.


yadij:
I was hoping to also drop the v3.5 notes. But there are still 3 vendors shipping Squid-3.5:
 * Endian (3.5.28)
 * Funtoo (3.5.27)
 * SMEServer (3.5.20)

kinkie (approved):

rousskov:
> I was hoping to also drop the v3.5 notes. But there are still 3 vendors shipping Squid-3.5

IMO, the reason to keep v3.5 notes in this branch is unrelated to this PR because vendors shipping Squid v3.5 do not ship the branch _this_ PR is modifying. If we are dropping any notes, then we should drop all v3, v4, v5 and v6 notes from the master branch: Vendors shipping those Squid versions are _not_ shipping our master branch. Any modification of vN notes are vN specific and, hence, do not (and should not) go through master.

yadij:
We still disagree on that then, so that is another reason for leaving 3.5 out of scope.

rousskov:
> PR title: Squid version 3.4 (and older) End-Of-Life

Is not Squid version 3.5 "End-Of-Life" as well? What is your definition of "End-Of-Life"?

PR #1509, opened 2023-10-11
v6 Typo fix

squidadm:


yadij:
Test of the CI automation whitelisting. Successful, we do not actually need this change in v6 yet.

PR #1511, opened 2023-10-11
Refactor Security::BlindPeerConnector constructor

kinkie:
Avoid having to include HttpRequest.h in
BlindPeerConnector.h by moving the
definition of the constructor to the .cc file

### DIFF
--- a/src/security/BlindPeerConnector.cc
+++ b/src/security/BlindPeerConnector.cc
@@ -90,7 +90,7 @@ Security::BlindPeerConnector::BlindPeerConnector(HttpRequestPointer &aRequest,
         const Comm::ConnectionPointer &aServerConn,
         const AsyncCallback<EncryptorAnswer> &aCallback,
         const AccessLogEntryPointer &alp,
-        const time_t timeout) :
+        time_t timeout) :
     AsyncJob("Security::BlindPeerConnector"),
     Security::PeerConnector(aServerConn, aCallback, alp, timeout)
 {
--- a/src/security/BlindPeerConnector.cc
+++ b/src/security/BlindPeerConnector.cc
@@ -86,3 +86,13 @@ Security::BlindPeerConnector::noteNegotiationDone(ErrorState *error)
     }
 }
 
+Security::BlindPeerConnector::BlindPeerConnector(HttpRequestPointer &aRequest,
+        const Comm::ConnectionPointer &aServerConn,
+        const AsyncCallback<EncryptorAnswer> &aCallback,
+        const AccessLogEntryPointer &alp,
+        const time_t timeout) :
+    AsyncJob("Security::BlindPeerConnector"),
+    Security::PeerConnector(aServerConn, aCallback, alp, timeout)
+{
+    request = aRequest;
+}
--- a/src/tests/stub_libsecurity.cc
+++ b/src/tests/stub_libsecurity.cc
@@ -24,7 +24,7 @@ BlindPeerConnector::BlindPeerConnector(HttpRequestPointer &, const Comm::Connect
                                        const time_t) :
     AsyncJob("Security::BlindPeerConnector"),
     Security::PeerConnector(aServerConn, aCallback, alp, 0)
-STUB_NOP
+{STUB_NOP}
 
 bool BlindPeerConnector::initialize(Security::SessionPointer &) STUB_RETVAL(false)
 Security::ContextPointer BlindPeerConnector::getTlsContext() STUB_RETVAL(Security::ContextPointer())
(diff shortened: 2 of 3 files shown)


yadij (approved):
One small improvement for the stub, but I do not insist on that.

kinkie:
@yadij the change works, please re-approve

rousskov (approved):
Thank you.

yadij (approved):

src/security/BlindPeerConnector.cc
```diff
@@ -86,3 +86,13 @@ Security::BlindPeerConnector::noteNegotiationDone(ErrorState *error)
     }
 }
 
+Security::BlindPeerConnector::BlindPeerConnector(HttpRequestPointer &aRequest,
+        const Comm::ConnectionPointer &aServerConn,
+        const AsyncCallback<EncryptorAnswer> &aCallback,
+        const AccessLogEntryPointer &alp,
+        const time_t timeout) :
```
yadij, line 93: With @rouskov removal of the `const` we must keep the definition and declarations matching.
```suggestion
        time_t timeout) :
```

src/tests/stub_libsecurity.cc
```diff
@@ -18,6 +18,14 @@
 CBDATA_NAMESPACED_CLASS_INIT(Security, BlindPeerConnector);
 namespace Security
 {
+BlindPeerConnector::BlindPeerConnector(HttpRequestPointer &, const Comm::ConnectionPointer & aServerConn,
+                                       const AsyncCallback<EncryptorAnswer> & aCallback,
+                                       const AccessLogEntryPointer &alp,
+                                       const time_t) :
```
yadij, line 24: With @rouskov removal of the `const` we must keep the definition and declarations matching.
```suggestion
                                       time_t) :
```

yadij (changes requested):

src/security/BlindPeerConnector.cc
rousskov: > we must keep the definition and declarations matching

They do match: In C++, `const`  in `const time_t` in function declaration is meaningless garbage -- no declaration user can benefit from that `const` being there (that is why some compilers warn about it) while `const`  in `const time_t foo` in function definition is a meaningful const-correctness feature -- both humans and compilers can benefit from it (matching `time_t` in the declaration).

Please withdraw your change requests.

kinkie:
@yadij , are you satisfied with the feedback from Alex? I'd like to push this forward

src/security/BlindPeerConnector.cc
yadij: FYI, One of the compilers that complain about "pointless" use of `const` is doxygen. Which complains as well when defintion/declaration are different `const`-ness.

rousskov: > FYI, One of the compilers that complain about "pointless" use of `const` is doxygen. Which complains as well when defintion/declaration are different `const`-ness.

If Doxygen complains about both pointless const in declarations AND a C++-allowed differences between definitions and declarations, then we have no better choice than to ignore those complaints because there is no way for us to satisfy Doxygen without losing const-correctness completely. 

BTW, there are dozens of these "different" declaration/definition cases in Squid already, and there will be more. Evidently, Doxygen can grok this code even if it (incorrectly) warns about it.

> Also, as you should know the reason being that when `const` is pointless, it became pointless because the variable value was _copied_ into the stack by `lvalue`. So the function/method does not have access to change the original caller variable even if it changes the variable it got. Even telling the compiler the variable is `const` is pointless - both caller and receiver already know it is `lvalue` passing.

I am not sure what the purpose of the above paragraph is, but the reason to add `const` to parameters in function definitions is to tell both the human reader and the compiler that the parameter cannot be modified by the function (including functions it calls). Many humans find it easier to read and refactor code when they know that a variable cannot be modified in the middle of it. Compilers make sure that invariant actually holds (and, in some rare complex cases, might even benefit from that `const` knowledge when optimizing code). In addition, using `const` in this context makes writing better code a bit easier -- fewer exceptions to worry about when following the "default to const" or "start with const" principle (that we should all follow to compensate for this C/C++ weakness IMHO).

src/security/BlindPeerConnector.cc
```diff
@@ -86,3 +86,13 @@ Security::BlindPeerConnector::noteNegotiationDone(ErrorState *error)
     }
 }
 
+Security::BlindPeerConnector::BlindPeerConnector(HttpRequestPointer &aRequest,
+        const Comm::ConnectionPointer &aServerConn,
+        const AsyncCallback<EncryptorAnswer> &aCallback,
+        const AccessLogEntryPointer &alp,
+        time_t timeout) :
```
rousskov, line 93: This should be constant, but I am not going to insist on restoring this in hope that this PR fiasco is not going to become a precedent for prohibiting marking parameters as constant in function implementations.

```suggestion
        const time_t timeout) :
```

Const-correctness is a good thing. We should not be arguing about those markers in function definitions or, at the very least, we should not be asking PR authors to _remove_ them.

yadij (approved):

PR #1513, opened 2023-10-12
Fix and improve html_quote()

kinkie:
Fixed an html_quote() bug: Unwanted space characters in decimal numeric
character references (e.g., `&#  7;` for ASCII BEL character).

Encode single quote character (`'`) as `&apos;` rather than `&#39;`.

Reduced encoding complexity from O(5*n) to O(n), where n is input string
length (and 5 is the number of supported character references)!

Added unit tests.


kinkie:
I'd love to transform this to sbuf-based, but it would require a second implementation, because
squidclient uses it so it needs to be in lib/, but SBuf is not useable there.

Thaoughts?

yadij:
> I'd love to transform this to sbuf-based, but it would require a second implementation, because squidclient uses it so it needs to be in lib/, but SBuf is not useable there.
> 

That is not a strict requirement, just convenience.  The `tools/` binaries can link against built objects in the `src/` directory - I have PR #1306 already open moving it.

src/tests/testHtmlQuote.cc
```diff
@@ -0,0 +1,74 @@
...
+#include "unitTestMain.h"
+
+#include <cstring>
+#include <iostream>
+
+class testHtmlQuote: public CPPUNIT_NS::TestFixture
+{
+    CPPUNIT_TEST_SUITE(testHtmlQuote);
+    CPPUNIT_TEST(test_html_quote_cstr);
+    // CPPUNIT_TEST(testPerformance);
```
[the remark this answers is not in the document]
kinkie: I actually plan to remove it before merging. I mostly used it to quantify the speed improvements coming out of using CharacterSet, but that's not really relevant for correctness, and to show reviewers what I used so that they can pick nits at it.

kinkie:
> > I'd love to transform this to sbuf-based, but it would require a second implementation, because squidclient uses it so it needs to be in lib/, but SBuf is not useable there.
> 
> That is not a strict requirement, just convenience. The `tools/` binaries can link against built objects in the `src/` directory - I have PR #1306 already open moving it.

SBuf brings in the whole MemPools subsystem, which is a big dependency.
I would rather issue a different challenge: why are we even distributing cachemanager anymore? We could morph it into a convenience wrapper for curl now.

rousskov (changes requested):
I support fixing and improving html_quote(). I would very much prefer that it is fixed and improved _once_, in this PR, as opposed to first in this PR and then again when we migrate it to an SBuf-friendly Tokenizer-based implementation.

> I'd love to transform this to sbuf-based, but it would require a second implementation, because squidclient uses it so it needs to be in lib/, but SBuf is not useable there.

I am blocking this PR to ask you to submit a prerequisite minimal PR that moves squidclient to src/. I am aware that Amos [says](https://github.com/squid-cache/squid/pull/1513#issuecomment-1759578542) that such a move is not necessary. If you refuse to move squidclient, I will find the time to move it myself.

N.B. I am against using SBuf outside of src/. I can detail why if necessary.

src/tests/testHtmlQuote.cc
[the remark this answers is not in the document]
rousskov: Thank you for showing what you use for performance testing. Please keep it until we are ready to merge this PR (and completely remove it then). I would also enable it until then, but I did not study the details and do not insist on it.

rousskov:
> I would rather issue a different challenge: why are we even distributing cachemanager anymore? We could morph it into a convenience wrapper for curl now.

I assume you meant squidclient.

I support removing squidclient completely (instead of moving it to src/) in a dedicated prerequisite PR. In fact, I think it is a great idea!

I see no real need for leaving a curl wrapper in place, but please prove me wrong if I missed some important reasons (other than "people still use squidclient"). If you decide to argue for leaving a wrapper, please keep Squid bug [5283](https://bugs.squid-cache.org/show_bug.cgi?id=5283) in mind -- it already precludes using just `squidclient mgr:foo` with many (most?) modern production Squids today unless the user is willing to jump through many hoops (which they are better equipped to do with curl or wget IMHO).

kinkie:
> > I would rather issue a different challenge: why are we even distributing cachemanager anymore? We could morph it into a convenience wrapper for curl now.
> 
> I assume you meant squidclient.

yes, sorry

> I support removing squidclient completely (instead of moving it to src/) in a dedicated prerequisite PR. In fact, I think it is a great idea!
> 
> I see no real need for leaving a curl wrapper in place, but please prove me wrong if I missed some important reasons (other than "people still use squidclient"). If you decide to argue for leaving a wrapper, please keep Squid bug [5283](https://bugs.squid-cache.org/show_bug.cgi?id=5283) in mind -- it already precludes using just `squidclient mgr:foo` with many (most?) modern production Squids today unless the user is willing to jump through many hoops (which they are better equipped to do with curl or wget IMHO).

What I'm thinking is to replace our current squidclient with a shell script that turns 'squidclient mgr:foo' into curl http://localhost/internal-squid-mgr/foo' and so on, for convenience

kinkie:
> I support fixing and improving html_quote(). I would very much prefer that it is fixed and improved _once_, in this PR, as opposed to first in this PR and then again when we migrate it to an SBuf-friendly Tokenizer-based implementation.


> > I'd love to transform this to sbuf-based, but it would require a second implementation, because squidclient uses it so it needs to be in lib/, but SBuf is not useable there.
> 
> I am blocking this PR to ask you to submit a prerequisite minimal PR that moves squidclient to src/. I am aware that Amos [says](https://github.com/squid-cache/squid/pull/1513#issuecomment-1759578542) that such a move is not necessary. If you refuse to move squidclient, I will find the time to move it myself.

TBH I think that this PR has enough value, between the bug fix and the unit test that allowed me to identify the bug in the first place, that I would prefer to do it in two - or even three - steps:
1. this
2. remove or rework squidclient
3. with squid being the only client of this, rework this (with the benefit of being able to do TDD)  


> N.B. I am against using SBuf outside of src/. I can detail why if necessary.

I agree with this general sentiment. The effort and risk that would be required to port all of SBuf dependencies out of src/ is not worthwhile in my opinion.

rousskov:
> > I see no real need for leaving a curl wrapper in place, but please prove me wrong if I missed some important reasons (other than "people still use squidclient"). If you decide to argue for leaving a wrapper, please keep Squid bug [5283](https://bugs.squid-cache.org/show_bug.cgi?id=5283) in mind -- it already precludes using just `squidclient mgr:foo` with many (most?) modern production Squids today unless the user is willing to jump through many hoops (which they are better equipped to do with curl or wget IMHO).

> What I'm thinking is to replace our current squidclient with a shell script that turns 'squidclient mgr:foo' into curl http://localhost/internal-squid-mgr/foo' and so on, for convenience

My "no real need[^1]" comment was written under the assumption that you were thinking exactly that. Glad my assumption was correct. The comment still stands (including its assertion that there is not much "convenience" left due to bug 5283 complications).

[^1]: And there is clearly harm: A lot of work is required to produce _and_ maintain a quality wrapper.

rousskov:
> I think that this PR has enough value

It is not the insufficient PR "value" that concerns me. It is the total cost. I expect that most of the cost of _landing_ this PR (in its current shape and form) and then landing an SBuf replacement PR has not been paid yet, and that I will be the one paying it. I want to increase the efficiency of us moving forward together by reordering the steps you prefer. So far, I have not heard arguments why we should not.

kinkie:
> So far, I have not heard arguments why we should not.

My main argument is that the current change is very straightforward and simple, the cost should be low.
I am happy to continue moving this to src/, but it will involve:

1. deciding what library to place it in (libformat? a new one?)
2. reimplementing the function
3. adapting callers (there's not many, thankfully)

I'm very happy to do it, I'd just like to checkpoint the current state first.
Please check if in practice there is a lot of work to get the current iteration merged. If you see that, I can agree and put in the extra effort. If on the other hand it's straightforward, let's check this in and I will start phase 2 tomorrow.

src/tests/testHtmlQuote.cc
```diff
@@ -0,0 +1,74 @@
...
+testHtmlQuote::test_html_quote_cstr()
+{
+    CPPUNIT_ASSERT_EQUAL(std::string(""), std::string(html_quote("")));
+    CPPUNIT_ASSERT_EQUAL(std::string("bar"), std::string(html_quote("bar")));
+    CPPUNIT_ASSERT_EQUAL(std::string("foo&lt;bar&gt;gazonk"), std::string(html_quote("foo<bar>gazonk")));
+    CPPUNIT_ASSERT_EQUAL(std::string("foo&amp;bar"), std::string(html_quote("foo&bar")));
+    CPPUNIT_ASSERT_EQUAL(std::string("some&#39;thing"), std::string(html_quote("some'thing")));
+    CPPUNIT_ASSERT_EQUAL(std::string("some&quot;thing"), std::string(html_quote("some\"thing")));
+    CPPUNIT_ASSERT_EQUAL(std::string("&#31;"), std::string(html_quote("\x1f")));
+}
```
rousskov, line 41: Please add the following test cases:

* one test for every HTML entity (see htmlEntities)
* a single test with a string that has nothing but a single HTML entity (of your choice)
* a single test with a string that has nothing but a code point (of your choice) that needs `&#nnn` encoding

lib/html_quote.cc
```diff
@@ -7,47 +7,25 @@
...
-    },
-    {
-        0, NULL
-    }
+const std::unordered_map<char,const char *> htmlEntities = {
+    {'<', "&lt;"},
+    {'>', "&gt;"},
+    {'"', "&quot;"},
+    {'&', "&amp;"},
+    {'\'', "&#39;"}
```
rousskov, line 22: It would be good to add a C++ comment that documents what this non-entity is doing in the entities map (or in the refactored code). In other words, why this is not `&apos;`... I do not insist on this change.

src/tests/testHtmlQuote.cc
rousskov, line 41: Pease add a loop that encodes each and every possible byte value. You would not be able to fully validate the encoding result (without heroic efforts), but please do check that the output length is positive and,

* if the output length is one, then the output character is the input character;
* if the first output character is `&`, then the last character is `;`;
* if the first output character is `&` and the second character is `#`, then output length is at least 4 (`&#1;`) and at most 6 (`&#255;`).

I do not insist on adding the check in the last/third bullet point above.

lib/html_quote.cc
```diff
@@ -7,47 +7,25 @@
...
 
 #include "squid.h"
+#include "base/CharacterSet.h"
 #include "html_quote.h"
 
-#if HAVE_STRING_H
-#include <string.h>
-#endif
+#include <cstring>
+#include <unordered_map>
```
rousskov, line 15: If possible, please auto-format changed sources. CI log [shows](https://github.com/squid-cache/squid/actions/runs/6495896158/job/17641739507?pr=1513) formatting changes.

rousskov (changes requested):
> > It is not the insufficient PR "value" that concerns me. It is the total cost. I expect that most of the cost of landing this PR (starting from its current shape and form) and then landing an SBuf replacement PR has not been paid yet, and that I will be the one paying it. I want to increase the efficiency of us moving forward together by reordering the steps you prefer. So far, I have not heard arguments why we should not.

> My main argument is that the current change is very straightforward and simple

... but not ready to be merged. It needs more work. Some of that work will be wasted because we will rewrite this code in the followup PR (during step 2). I see no good reason to create that waste.

> the cost should be low.

Agreed, it should be, but it is not, and that is what matters. If this PR were easy to review and ready to merge, sure! It is neither. FWIW, I spent ~95 minutes on this review iteration alone.

> I am happy to continue moving this to src/, but it will involve:
> 
> 1. deciding what library to place it in (libformat? a new one?)
> 2. reimplementing the function
> 3. adapting callers (there's not many, thankfully)

Agreed re "will involve", although steps 2 and 3 should be either essentially combined or reordered (and there may be more steps).

> I'd just like to checkpoint the current state first.

Yes, I am aware and do understand.

> Please check if in practice there is a lot of work to get the current iteration merged. If you see that, I can agree and put in the extra effort. If on the other hand it's straightforward, let's check this in and I will start phase 2 tomorrow.

Done. There is, as I expected, quite a bit of work left here -- see change requests for details. It is your call whether to proceed with adjusting this PR or move to the other steps.

lib/html_quote.cc
```diff
@@ -7,47 +7,25 @@
  */
 
 #include "squid.h"
+#include "base/CharacterSet.h"
```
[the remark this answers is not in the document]
yadij: This issues would be resolved by merging PR #1306 first. I have updated and re-requested review. If we can get that is now @kinkie is free to change the API and rename the function within the new namespace in one step as this PR.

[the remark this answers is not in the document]
rousskov: > PR #1306 is now merged. Resolving

Merging of PR #1306, by itself, does not address this change request. Unresolving.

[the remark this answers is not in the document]
rousskov: And merging master into this PR branch without properly resolving conflicting file locations (i.e. merge commit 183966d) does not address this change request either. The followup merge fix (i.e. commit cc648ed) does resolve this change request.

src/tests/testHtmlQuote.cc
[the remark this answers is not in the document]
kinkie: There is little point in leaving it enabled, as it's tricky to obtain a baseline measure.
It might come in handy to assess the efficiency of an alternate
quoting implementation that uses SBufs directly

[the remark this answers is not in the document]
rousskov: > There is little point in leaving it enabled

Just to clarify: The point of leaving it enabled is to allow others working on this PR to run this code without modifying sources. This added convenience is minor, of course.

src/tests/testHtmlQuote.cc
kinkie: Github copilot did most of the legwork :)
More than that, it exposes that this function takes signed chars, but the actual type that would make sense is unsigned

lib/html_quote.cc
kinkie: There is actually no reason why it can't be &apos; that I know of.
In fact, why don't we enrich this table with the ASCII part of the list from https://www.w3schools.com/html/html_entities.asp ?

lib/html_quote.cc
kinkie: source-formatted

src/html/Quoting.cc
```diff
@@ -14,8 +14,7 @@
 #include <unordered_map>
 
 /*
- *  html_do_quote - Returns a static buffer containing the quoted
- *  string.
+ *  html_do_quote - Returns a static buffer containing the quoted string.
```
rousskov, line 17: This description belongs to the header file, where it should use a single-line Doxygen format. For example:

```
/// a static buffer containing an HTML-encoded version of the given c-string
```

Adjust as needed, of course. It would be nice to reference HTML encoding standard that we are implementing here, but I do not know whether it exists.

lib/html_quote.cc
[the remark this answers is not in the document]
rousskov: > Alex: The followup merge fix (i.e. commit cc648edebf78a8976ba3964816404473a5d4448c) does resolve this change request.

> Francesco: I think it should be fine now, is it not?

Yes, it is fine. I have marked this change request as resolved yesterday.

src/html/Quoting.cc
```diff
@@ -7,81 +7,51 @@
...
+#include <cstring>
+#include <unordered_map>
 
-/*
- *  html_do_quote - Returns a static buffer containing the quoted
- *  string.
- */
 char *
 html_quote(const char *string)
 {
```
rousskov, line 18: > https://www.w3.org/MarkUp/HTMLPlus/htmlplus_59.html#0 seems to be the closest I can find to an authority.

I think we should refer to https://html.spec.whatwg.org/#character-references or https://html.spec.whatwg.org/multipage/syntax.html#syntax-charref (which covers the general `&#n;` encoding mechanism and refers to https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references for specific named character references).
 
> Quoting from https://www.w3.org/MarkUp/HTMLPlus/htmlplus_13.html: "There are a large number of entities defined by the ISO, covering most languages and symbols for publishing and mathematics. Requiring all browsers to support these would be impractical"

I do not think we care about that problem (in this PR scope) because we do not support UTF-8 or other standards that actually allow for that "larger number of entities". We only care about a handful of named character references we use and the general (decimal) numeric character reference mechanism for encoding bytes.

```suggestion
{
    // See https://html.spec.whatwg.org/#character-references
```

P.S. The references you found are for "HTML+ discussion documents". The references I recommend are for the HTML standard, albeit a "living" one.

src/tests/testHtmlQuote.cc
kinkie: added

src/tests/testHtmlQuote.cc
```diff
@@ -0,0 +1,87 @@
...
+    CPPUNIT_ASSERT_EQUAL(std::string("some&apos;thing"), std::string(html_quote("some'thing")));
+    CPPUNIT_ASSERT_EQUAL(std::string("some&quot;thing"), std::string(html_quote("some\"thing")));
+
+    for (unsigned char ch = 1; ch < 0xff; ++ch) {
+        unsigned char buf[2] = {ch, '\0'};
+        char *quoted = html_quote(reinterpret_cast<char *>(buf));
+        if (strlen(quoted) == 1) {
+            CPPUNIT_ASSERT_EQUAL(static_cast<int>(ch), static_cast<int>(quoted[0]));
+        } else {
+            CPPUNIT_ASSERT_EQUAL(';', quoted[strlen(quoted)-1]);
```
rousskov, line 47: This code assumes that strlen(quoted) is positive, but we only know that it is not equal to 1.

```suggestion
            CPPUNIT_ASSERT(strlen(quoted) > 1);
            CPPUNIT_ASSERT_EQUAL(';', quoted[strlen(quoted)-1]);
```

kinkie: adding

lib/html_quote.cc
```diff
@@ -73,19 +50,16 @@ html_quote(const char *string)
         /* Walk thru the list of HTML Entities that must be quoted to
          * display safely
          */
-        for (i = 0; htmlstandardentities[i].code; i++) {
-            if (ch == htmlstandardentities[i].code) {
-                escape = htmlstandardentities[i].quote;
-                break;
-            }
+        if (htmlSpecialCharacters[ch]) {
+            escape = htmlEntities.at(ch); // guaranteed to exist
```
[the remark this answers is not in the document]
kinkie: @rousskov , is HtmlSpecialCharacters in the latest commit similar to what you had in mind?

Using it the performance of html_encode gets more than doubled over what's in trunk (the simple test runs in 1.41s while in trunk - IIRC - it's about 3.2 )

lib/html_quote.cc
```diff
@@ -73,19 +50,16 @@ html_quote(const char *string)
...
         }
         /* Encode control chars just to be on the safe side, and make
          * sure all 8-bit characters are encoded to protect from buggy
          * clients
          */
         if (!escape && (ch <= 0x1F || ch >= 0x7f) && ch != '\n' && ch != '\r' && ch != '\t') {
             static char dec_encoded[7];
-            snprintf(dec_encoded, sizeof dec_encoded, "&#%3d;", (int) ch);
+            snprintf(dec_encoded, sizeof dec_encoded, "&#%d;", static_cast<int>(ch));
```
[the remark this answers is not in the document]
kinkie: That's a very good point; I'll extract that bit into its own PR, no sense in keeping it pending

src/html/Quoting.cc
```diff
@@ -9,81 +9,74 @@
...
-    },
-    {
-        '&', "&amp;"
-    },
-    {
-        '\'', "&#39;"
-    },
-    {
-        0, NULL
+    for (uint32_t ch = 1; ch < 256; ++ch) {
```
yadij, line 38: This misses out the `&#0;` character reference for ASCII `NUL`.

src/html/Quoting.h
```diff
@@ -9,6 +9,7 @@
 #ifndef SQUID__SRC_HTML_QUOTING_H
 #define SQUID__SRC_HTML_QUOTING_H
 
+/// obtain a static buffer containing an HTML-encoded version of the given c-string
```
yadij, line 12:
```suggestion
/** Obtain a static buffer containing an HTML-encoded version of the given c-string.
 *
 * HTML reserved characters are replaced with character references
 * per https://html.spec.whatwg.org/#character-references
 */
```

yadij (changes requested):
Two minor changes, and one big question .....

 What is the scope of this PR now?
 Is it going to stay as C-string manipulation, or get another round of refactor to SBuf before merge?

src/html/Quoting.cc
kinkie: Yes. However, as the input to the function is null-terminated strings, we have no way to receive that input

kinkie:
> Is it going to stay as C-string manipulation, or get another round of refactor to SBuf before merge?

My strategy would be to merge as c-string, then refactor to SBuf and merge that separately

src/html/Quoting.cc
rousskov: You are both right. Please include NUL because

* HtmlSpecialCharacters class is not documented to be specific/internal to html_quote(const char *);
* Including it will reduce the probability of bugs in future refactoring changes where input becomes something other than a c-string.

```suggestion
    for (uint32_t ch = 0; ch < 256; ++ch) {
```

src/html/Quoting.cc
```diff
@@ -9,81 +9,74 @@
...
-static struct {
-    unsigned char code;
-    const char *quote;
-} htmlstandardentities[] =
+#include <array>
+#include <cstring>
+
+class HtmlSpecialCharacters {
+private:
+    std::array<const char *, 256> escapeMap = {};
```
rousskov, line 17: Can we use SBuf to store escape sequences, removing the need for adding this (_currently buggy_) new class (and some associated hacks like maxEscapeLength with it)?

```suggestion
    std::array<SBuf, 256> escapeMap = {};
```

src/html/Quoting.cc
```diff
@@ -9,81 +9,74 @@
...
 
-/*
- *  html_do_quote - Returns a static buffer containing the quoted
- *  string.
- */
 char *
 html_quote(const char *string)
 {
-    static char *buf = nullptr;
+    // for syntax, see https://html.spec.whatwg.org/#character-references
```
rousskov, line 59: This reference was moved to the header file.

```suggestion
```

src/html/Quoting.cc
```diff
@@ -9,81 +9,74 @@
...
-         * clients
-         */
-        if (!escape && (ch <= 0x1F || ch >= 0x7f) && ch != '\n' && ch != '\r' && ch != '\t') {
-            static char dec_encoded[7];
-            snprintf(dec_encoded, sizeof dec_encoded, "&#%3d;", (int) ch);
-            escape = dec_encoded;
-        }
+        escape = htmlSpecialCharacters[ch];
+
         if (escape) {
```
rousskov, line 80:
```suggestion
        if (const auto escape = htmlSpecialCharacters[ch]) {
```

... and drop the wrong-scope `escape` declaration above.

src/html/Quoting.cc
```diff
@@ -9,81 +9,74 @@
...
-    if (!buf || strlen(string) * 6 > bufsize) {
+    if (buf == nullptr || strlen(string) * 6 > bufsize) {
         xfree(buf);
         bufsize = strlen(string) * 6 + 1;
-        buf = static_cast<char *>(xcalloc(bufsize, 1));
+        buf = static_cast<char*>(xcalloc(bufsize, 1));
     }
     for (src = string, dst = buf; *src; src++) {
-        const char *escape = NULL;
+        const char *escape = nullptr;
```
rousskov, line 75:
```suggestion
```

src/html/Quoting.cc
```diff
@@ -9,81 +9,74 @@
...
-    /* XXX This really should be implemented using a MemPool, but
-     * MemPools are not yet available in lib...
+    /* XXX This really should be implemented using a MemPool
      */
-    if (!buf || strlen(string) * 6 > bufsize) {
+    if (buf == nullptr || strlen(string) * 6 > bufsize) {
         xfree(buf);
         bufsize = strlen(string) * 6 + 1;
-        buf = static_cast<char *>(xcalloc(bufsize, 1));
+        buf = static_cast<char*>(xcalloc(bufsize, 1));
```
rousskov, line 72: Please undo these unnecessary, partially controversial (and arguably unwanted) changes, starting from the XXX comment and ending with `         buf = static_cast<char...` below it.

src/tests/testHtmlQuote.cc
```diff
@@ -0,0 +1,90 @@
...
+        if (strlen(quoted) == 1) {
+            CPPUNIT_ASSERT_EQUAL(static_cast<int>(ch), static_cast<int>(quoted[0]));
+        } else {
+            CPPUNIT_ASSERT(strlen(quoted) > 1);
+            CPPUNIT_ASSERT_EQUAL(';', quoted[strlen(quoted)-1]);
+        }
+        if (strlen(quoted) > 1 && quoted[0] == '&' && quoted[1] == '#') {
+            CPPUNIT_ASSERT(strlen(quoted) > 3);
+            CPPUNIT_ASSERT(strlen(quoted) <= 6);
+        }
```
rousskov, line 57: AFAICT, we can simplify because the first character of a multicharacter sequence is always `&`, right?

```suggestion
            CPPUNIT_ASSERT_EQUAL('&', quoted[0]);
            CPPUNIT_ASSERT_EQUAL(';', quoted[strlen(quoted)-1]);
            if (quoted[1] == '#') {
                CPPUNIT_ASSERT(strlen(quoted) > 3);
                CPPUNIT_ASSERT(strlen(quoted) <= 6);
            }
        }
```

src/tests/testHtmlQuote.cc
```diff
@@ -0,0 +1,90 @@
...
+    CPPUNIT_ASSERT_EQUAL(std::string("&#163;"), std::string(html_quote("\xa3")));
+
+    for (unsigned char ch = 1; ch < 0xff; ++ch) {
+        unsigned char buf[2] = {ch, '\0'};
+        auto quoted = html_quote(reinterpret_cast<char *>(buf));
+
+        if (strlen(quoted) == 1) {
+            CPPUNIT_ASSERT_EQUAL(static_cast<int>(ch), static_cast<int>(quoted[0]));
+        } else {
+            CPPUNIT_ASSERT(strlen(quoted) > 1);
```
rousskov, line 51: If the logic in the simplification change request below is correct, then we can be more assertive here:

```suggestion
            CPPUNIT_ASSERT(strlen(quoted) >= 3);
```

rousskov (changes requested):
I think we are getting close to the final version, thank you!

src/html/Quoting.cc
kinkie: This would have the effect of pulling libsbuf, libbase, mempools into everythin in $(top_builddir)/tools , via libhtml.la , so.. no. I could use (for now) std::string

rousskov: > This would have the effect of pulling libsbuf, libbase, mempools into everythin in $(top_builddir)/tools

You are right. It looks like an unfortunate typo in [previous](https://github.com/squid-cache/squid/pull/1513#issuecomment-1759609500) [analysis](https://github.com/squid-cache/squid/pull/1513#issuecomment-1759663076) resulted in this tools/cachemgr problem unaddressed by the recent official commit 80cba6d. Sigh. FWIW, I would (still) enthusiastically support removing cachemgr from this repository.

> I could use (for now) std::string

We should not use std::string in new src/ code (I can detail the reasons if needed, but we have discussed this before). There are some narrow exceptions to that rule, but src/html/Quoting.cc does not really qualify. The correct solution, IMO, is to remove cachemgr. We could move cachemgr into src/ as a workaround, but that would only prolong that tool's agony.

We could grant Quoting.cc a temporary exception for std::string use to save time required to fix HtmlSpecialCharacters class, but such an exception may encourage even more wasteful attempts to sprinkle src/ with std::string. Fortunately, that exception is not necessary to solve the problem this change request is focusing on.

I recommend a different solution:

1. Remove HtmlSpecialCharacters class. This way you do not have to fix it (or even discuss how it is broken).
2. "Leak" std::array items memory. These items are allocated only once in program lifetime. It is acceptable to never destroy them (Squid has lots of similar "leaks" and even _recommends_ them as the best solution for the "No new globals" rule which is very applicable to this array).

After the above changes, the beginning of html_quote() may look like this:

```C++
static const auto escapeSequences = MakeEscapeSequences();
```

... where MakeEscapeSequences() is a `static` function that allocates (and returns a pointer to) std::array<const char *, 256>. Try using `auto` for MakeEscapeSequences() return type so that you do not have to duplicate or name that array type -- it should only be spelled out once.

kinkie: Done and working, with a slight difference: instead of a pointer, I'm returning a reference to a static variable in MakeEscapeSequences(). Is it an acceptable change? What are the pros/cons, if any?

src/html/Quoting.cc
kinkie: done, resolving

src/html/Quoting.cc
kinkie: ok

src/html/Quoting.cc
rousskov: > instead of a pointer, I'm returning a reference to a static variable in MakeEscapeSequences(). Is it an acceptable change?

No, returning a reference to a disappearing (static) variable is not acceptable. See the documentation for the "[No new globals](https://wiki.squid-cache.org/DeveloperResources/SquidCodingGuidelines#rule-no-new-globals)" rule for explanations, especially the last paragraph.

src/html/Quoting.cc
kinkie: done

lib/html_quote.cc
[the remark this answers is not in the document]
kinkie: implemented. resolving

src/html/Quoting.cc
kinkie: Hm.. "adjusted to return a constant reference" in the "no new globals" spec is confusing :\

I'm not particularly keen to mix pointers with operator[]. Since it's done only once, what about copying? I'll use pointers if you insist

rousskov: I insist (and can explain why if needed). I can adjust this myself if you prefer.

rousskov: > "adjusted to return a constant reference" in the "no new globals" spec is confusing :\

In what way? The sketched function returns a constant reference. It can be adjusted to return a constant reference (instead of a non-constant one).

BTW, you _can_ implement this using the sketch from "no new globals". Doing so will eliminate pointers to the array from html_quote().

kinkie:
I think all outstanding remarks have been tackled

yadij (approved):

src/html/Quoting.cc
```diff
@@ -9,50 +9,48 @@
...
 
-/*
- *  html_do_quote - Returns a static buffer containing the quoted
- *  string.
- */
 char *
 html_quote(const char *string)
 {
-    static char *buf = nullptr;
+    static const auto htmlSpecialCharacters = MakeEscapeSequences();
```
rousskov, line 48: We do not want to copy this array, even after migration to SBuf:

```suggestion
    static const auto &htmlSpecialCharacters = MakeEscapeSequences();
```

I also recommend renaming htmlSpecialCharacters variable to `escapeSequences` because it does not actually contain or supply special characters in any applicable sense, but I do not insist on that naming fix.

```suggestion
    static const auto &escapeSequences = MakeEscapeSequences();
```

src/html/Quoting.cc
```diff
@@ -9,50 +9,48 @@
...
- */
-static struct {
-    unsigned char code;
-    const char *quote;
-} htmlstandardentities[] =
+#include <array>
+#include <cstring>
+
 
+static const auto& MakeEscapeSequences()
```
rousskov, line 16: Please follow Squid Style (and run formatting to fix a couple of other [problems](https://github.com/squid-cache/squid/actions/runs/6630677868/job/18012571733?pr=1513)):

```suggestion
static const auto &
MakeEscapeSequences()
```

rousskov, line 16: With the current implementation, this is not a "maker" function (i.e. a function that gives you a new thing every time you call it). This is an accessor to escape sequences (that have complex initialization logic). Please rename it to EscapeSequences(), AllEscapeSequences(), or something like that _and_ either assert() that it is never called twice (simply because we do not want to spend time on supporting multiple calls) or adjust it a bit to return the previously initialized array on repeated calls. You can use `escapeMap` for that assert.

src/html/Quoting.cc
```diff
@@ -9,50 +9,48 @@
...
-    /* NOTE: The quoted form MUST not be larger than 6 character.
-     * see close to the MemPool commend below
+    static std::array<std::pair<unsigned char, const char *>, 5> const escapePairs = {
+        std::make_pair('<', "&lt;"),
+        std::make_pair('>', "&gt;"),
+        std::make_pair('"', "&quot;"),
+        std::make_pair('&', "&amp;"),
+        std::make_pair('\'', "&apos;")
+    };
+    static auto escapeMap = new std::array<const char *, 256>{};
```
rousskov, line 25: As we discussed on the conference call, I do recommend migrating this array from c-string to SBuf in this PR after removing tools/cachemgr.cgi, but it is your call.

src/html/Quoting.cc
```diff
@@ -9,50 +9,48 @@
...
+    for (uint32_t ch = 0; ch < 256; ++ch) {
+        if ((ch <= 0x1F || ch >= 0x7f) && ch != '\n' && ch != '\r' && ch != '\t') {
+            (*escapeMap)[ch] = static_cast<char *>(xcalloc(maxEscapeLength, 1));
+            snprintf(const_cast<char*>((*escapeMap)[ch]), sizeof escapeMap[ch], "&#%d;", static_cast<int>(ch));
+        }
+    }
+    for (auto &pair: escapePairs) {
+        (*escapeMap)[pair.first] = static_cast<char *>(xcalloc(maxEscapeLength, 1));
+        xstrncpy(const_cast<char*>((*escapeMap)[pair.first]), pair.second, maxEscapeLength);
     }
```
rousskov, line 40: If we migrate escapeMap to SBuf, then this loop can be unrolled. Unrolling it will remove escapePairs, will save a few code lines, and might arguably simplify this code a bit. Your call.

src/tests/testHtmlQuote.cc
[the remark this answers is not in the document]
rousskov: After all performance related improvements are completed (and PR description updated accordingly), please do not forget to remove testPerformance().

rousskov (changes requested):
I found one problem (missing reference in htmlSpecialCharacters type). The rest is polishing.

src/html/Quoting.cc
kinkie: done

src/html/Quoting.cc
kinkie: Waiting for PR #1542 to land

src/html/Quoting.cc
kinkie: done in 446feb58d9

kinkie:
Interestingly, using SBuf as a backing storage comes with a 3.5x performance regression.

src/tests/testHtmlQuote.cc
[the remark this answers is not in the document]
kinkie: Done

src/html/Quoting.cc
kinkie: done

yadij (dismissed):

rousskov (approved):
Thank you for addressing all my previous concerns and significantly improving this code. I pushed a few minor polishing touches and adjusted PR title/description, including the addition of the following change disclosure:

> Encode single quote character (`'`) as `&apos;` rather than `&#39;`.

FWIW, I am not sure we should have made the above encoding change, but I hope it does not hurt, and I am not objecting to it.

I am OK with this PR going in "as is", but I will not clear this PR for merging myself to give @kinkie and @yadij a chance to review the last set of changes and adjust things further if needed.

kinkie:
LGTM. Thanks for improving this! Clearing for merge

PR #1516, opened 2023-10-13
Fix and improve annotation reporting

kinkie:
We just wanted to remove legacy printf()-like calls from Notes.cc, but
realized that finding correct replacement for that code is complicated
because some of the calls were broken, and the true meaning or purpose
of the affected annotation reporting methods was elusive. This change
combines several related fixes and improvements detailed below.

### Fix reporting method names and their descriptions

Humans could not easily figure out the difference between Note::dump(),
Note::toString(), Notes::dump(), Notes::toString(), and
NotePairs::toString() methods, especially since Note and Notes classes
had both, implying some important difference. The toString() name is
very generic. The dump() name is used (differently) in Configuration
code and ACL class hierarchy; these classes are used by that code, but
they do not belong to that hierarchy. Bugs and the variety of
annotation-related use cases increased doubts and confusion.

The new task/format-specific names for Note and Notes methods fix this.

### Fix annotate_client and annotate_transaction mgr:config reporting

```diff
-acl markAsX annotate_client name1: value1
-name2: value2a,value2b
-
+acl markAsX annotate_client name1=value1 name2=value2a,value2b
```

Multi-name annotations were split across several lines and used an
incorrect name/value separator. The "acl" directive line was followed by
an extra new line.

### Fix note directive mgr:config reporting

```diff

-note name1 "value1a,value1b"(note name1... line) 
-
-note name2 "value2"(note name2... line) 
-
+note name1 "value1a,value1b"
+note name2 "value2" fromTrustedClient toTrustedServer
```

Each "note" directive was followed by bogus "(note...)" suffix instead
of ACL names (if any). The "note" directive line was followed by an
extra new line.

### Improve debugging of annotations in helper responses

```diff
- ... externalAclHandleReply: reply={result=OK, notes={tag: 1; x_: Y }}
+ ... externalAclHandleReply: reply={result=OK, notes={tag=1 x_=Y }}
```

The contents of old notes{...} did not match what helpers were sending.
The new format matches helper output in all simple cases (at least).


src/Notes.cc
```diff
@@ -116,12 +116,11 @@ Note::dump(StoreEntry *entry, const char *k)
...
 
 SBuf
-Note::toString(const char *sep) const
+Note::toString(const SBuf &sep) const
 {
     SBuf result;
     for (const auto &val: values)
-        result.appendf("%.*s: %.*s%s", key().length(), key().rawContent(),
-                       val->value().length(), val->value().rawContent(), sep);
+        result.append(key()).append(": ",2).append(val->value()).append(sep);
```
yadij, line 123: `SBufStream` is probably better here than chaining `append()`

kinkie: Given all the contents are SBuf, chaining appends should be more efficient

src/Notes.h
```diff
@@ -11,6 +11,7 @@
 
 #include "acl/forward.h"
 #include "base/RefCount.h"
+#include "sbuf/SBuf.h"
```
rousskov, line 14: New header code does not store SBuf. Can we just do something like this instead?

```suggestion
#include "sbuf/forward.h"
```

src/Notes.cc
```diff
@@ -258,14 +257,15 @@ Notes::dump(StoreEntry *entry, const char *key)
         n->dump(entry, key);
 }
 
-const char *
+SBuf
 Notes::toString(const char *sep) const
 {
     static SBuf result;
+    SBuf separator(sep);
```
rousskov, line 264: This adds a memory allocation that is not really necessary. I am pretty sure the whole thing needs to be converted to use std::ostream (instead of SBuf) and asList(), but I do not have the time to provide a blueprint right now. I can suggest starting with removing this performance regression.

rousskov (changes requested):

src/Notes.cc
kinkie: SBufStream instantiates one additional SBuf for the sink plus a bit of code between SBufStream, PackableStream, std::streambuf and the per-type operator<<'s .

Looking at the code it is not a lot, but in this case feels very avoidable

src/Notes.cc
kinkie: I argue it is not a significant performance regression; the alternative is to call strlen(sep).
In practical terms, this code has only one caller, where the separator is not even specified so the default ("\r\n") is used

src/Notes.cc
yadij: Where is the second one coming from? That could be well worth removing if true.

FWIW, IIRC the stream should be writing to the buffer provided to its constructor.

kinkie: > Where is the second one coming from? That could be well worth removing if true.
> 
> FWIW, IIRC the stream should be writing to the buffer provided to its constructor.

Doh, you're right. SBufStream::sink_ is a reference.

kinkie: > To avoid misunderstanding, I am participating in this change request discussion, but I do not insist on any work (or suggest any changes not backed by that extra research work) associated with this change request.

Thanks for making this explicit!

src/Notes.h
kinkie: Yes, changing

src/Notes.cc
kinkie: I wasn't aware of AsList(). It's nice; I'll use it

kinkie:
Waiting for PR #1306  to land before resuming work here

kinkie:
Okay, things should be in a state where merging could be feasible.
The code is slightly better, but there is a lot of work to do which is outside the scope for this PR.

Changing toString requires changing externally-facing promises that are made in squid.conf, and implemented via src/format/Format.cc in very opaque ways - suffice to say that around Format.cc:1395 , the separator is written added to the output in a find() function; in fact the whole Format::Format::assemble should probably be rewritten entirely.

src/Notes.cc
```diff
@@ -108,8 +110,8 @@ void
 Note::dump(StoreEntry *entry, const char *k)
 {
     for (const auto &v: values) {
-        storeAppendPrintf(entry, "%s %.*s %s",
-                          k, key().length(), key().rawContent(), ConfigParser::QuoteString(SBufToString(v->value())));
+        storeAppendPrintf(entry, "%s " SQUIDSBUFPH " %s",
+                          k, SQUIDSBUFPRINT(key()), ConfigParser::QuoteString(SBufToString(v->value())));
```
rousskov, line 114: Please use `PackableStream os(*entry)` and avoid preserving storeAppendPrintf() (among other long-term advantages). This is especially relevant in a PR that focuses on removing appendf() calls!

P.S. You do _not_ have to convert dump_acl_list() below to use a stream here.

src/acl/AnnotationData.cc
```diff
@@ -23,8 +24,9 @@ SBufList
 ACLAnnotationData::dump() const
 {
     SBufList sl;
-    if (const char *strNotes = notes->toString())
-        sl.push_back(SBuf(strNotes));
+    auto strNotes = notes->toString();
+    if (!strNotes.isEmpty())
+        sl.push_back(strNotes);
```
rousskov, line 29: Can we tell that notes are empty before printing them? Something along these lines, perhaps?

```suggestion
    if (!notes->empty())
        sl.push_back(notes->toString());
```

src/Notes.h
```diff
@@ -100,6 +102,7 @@ class Note: public RefCountable
     Values values; ///< The possible values list for the note
 };
 
+
```
rousskov, line 105:
```suggestion
```

src/Notes.h
rousskov: The promised change did not happen or was reverted. Unresolving. The question in the change request remains open.

rousskov (changes requested):
> Changing toString requires changing externally-facing promises 

FWIW, it would help (me) to know what those blocking promises are, exactly.

> that are made in squid.conf, and implemented via src/format/Format.cc in very opaque ways - suffice to say that around `Format.cc:1395`, the separator is written added to the output in a find() function

AFAICT, that find() call locates an individual note value in a multi-valued note using a user-configured value separator. Think about finding value `private` in a `Cache-Control: max-age=0,private` note. That find() call should be unrelated to _printing_ notes and note values.

I left a few simple change requests to make progress. I will have to come back to this once the blocking "promises" in the above excuse are known.

src/Notes.cc
```diff
@@ -40,7 +42,7 @@ Note::Value::Value(const char *aVal, const bool quoted, const char *descr, const
         if (!valueFormat->parse(theValue.c_str())) {
             delete valueFormat;
             SBuf exceptionMsg;
-            exceptionMsg.Printf("failed to parse annotation value %s", theValue.c_str());
+            exceptionMsg.Printf("failed to parse annotation value ").append(theValue);
             throw TexcHere(exceptionMsg.c_str());
```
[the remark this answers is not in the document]
kinkie: done in b6a1d69aed

src/Notes.h
kinkie: done in ebc2484b09

src/acl/AnnotationData.cc
kinkie: Yep, you're right

kinkie:
> > Changing toString requires changing externally-facing promises
> 
> FWIW, it would help (me) to know what those blocking promises are, exactly.

From cf.data.pre:
```
                        By default, multiple note values are separated with ","
                        and multiple notes are separated with "\r\n".
                        When logging named notes with %{name}note, the
                        explicitly configured separator is used between note
                        values. When logging all notes with %note, the
                        explicitly configured separator is used between
                        individual notes. There is currently no way to
                        specify both value and notes separators when logging
                        all notes with %note.
```


> > that are made in squid.conf, and implemented via src/format/Format.cc in very opaque ways - suffice to say that around `Format.cc:1395`, the separator is written added to the output in a find() function
> 
> AFAICT, that find() call locates an individual note value in a multi-valued note using a user-configured value separator. Think about finding value `private` in a `Cache-Control: max-age=0,private` note. That find() call should be unrelated to _printing_ notes and note values.

I suppose it's a matter of understanding what we mean by "printed".
If a note has multiple values, they are squashed together, separated by the separator supplied to find(), in the result returned by find(). We can argue that since find doesn't output anything, it's not printing.
But the content that is eventually printed to log is formed within find(), which is counterintuitive to me, as opposed for instance to find() returning a SBufList that then the caller will be in charge to format however it wants to.


> I left a few simple change requests to make progress. I will have to come back to this once the blocking "promises" in the above excuse are known.

rousskov:
Thank you for detailing what you meant by "externally-facing promises".

> > AFAICT, that find() call locates an individual note value in a multi-valued note using a user-configured value separator. Think about finding value `private` in a `Cache-Control: max-age=0,private` note. That find() call should be unrelated to _printing_ notes and note values.
 
> I suppose it's a matter of understanding what we mean by "printed". If a note has multiple values, they are squashed together, separated by the separator supplied to find(), in the result returned by find().

The above "note has multiple values" description assumes that Squid can always tell, without any additional information, that a note has multiple values. That assumption is false as detailed below.

> ...  We can argue that since find doesn't output anything, it's not printing. But the content that is eventually printed to log is formed within find(), which is counterintuitive to me, as opposed for instance to find() returning a SBufList that then the caller will be in charge to format however it wants to.

A hypothetical generic find() would not be able to correctly return the list of values in some cases because Squid does not and cannot know what an individual annotation value is without admin supplying that value syntax knowledge in squid.conf by providing a individual value separator.

For example, what is are the values in a `Foo: bar,baz;boo+error=2` annotation? There is no single correct answer to that question! To one admin, the answer is `bar` and `baz;boo+error=2`. To another, it is `bar,baz;boo` and `error=2`. The admin has to tell us what value syntax they use. Without that information, we only know that the annotation is named "Foo" and has one or more values.

Once we know the separator, our find() function then can find individual values. This find() step is not about printing/reporting/etc. It is about _parsing_!

And, yes, we also need to deal with multiple same-name annotations, and we have a default value separator for the parser (and, IIRC, at least two default delimiters for the printer -- annotation delimiter and individual annotation value delimiter).

kinkie:
> Thank you for detailing what you meant by "externally-facing promises".
> 
> > > AFAICT, that find() call locates an individual note value in a multi-valued note using a user-configured value separator. Think about finding value `private` in a `Cache-Control: max-age=0,private` note. That find() call should be unrelated to _printing_ notes and note values.
> 
> > I suppose it's a matter of understanding what we mean by "printed". If a note has multiple values, they are squashed together, separated by the separator supplied to find(), in the result returned by find().
> 
> The above "note has multiple values" description assumes that Squid can always tell, without any additional information, that a note has multiple values. That assumption is false as detailed below.
> 
> > ...  We can argue that since find doesn't output anything, it's not printing. But the content that is eventually printed to log is formed within find(), which is counterintuitive to me, as opposed for instance to find() returning a SBufList that then the caller will be in charge to format however it wants to.
> 
> A hypothetical generic find() would not be able to correctly return the list of values in some cases because Squid does not and cannot know what an individual annotation value is without admin supplying that value syntax knowledge in squid.conf by providing a individual value separator.
> 
> For example, what is are the values in a `Foo: bar,baz;boo+error=2` annotation? There is no single correct answer to that question! To one admin, the answer is `bar` and `baz;boo+error=2`. To another, it is `bar,baz;boo` and `error=2`. The admin has to tell us what value syntax they use. Without that information, we only know that the annotation is named "Foo" and has one or more values.
> 
> Once we know the separator, our find() function then can find individual values. This find() step is not about printing/reporting/etc. It is about _parsing_!

Sorry, I don't follow you.
In this configuration example:
```
acl an1 annotate_transaction n1+=n1v1 n1+=n1v2
acl an2 annotate_transaction n1+=n1v3 n2+=n2v1
note n3 n3v1 all
logformat notes %/note /// %/{n1}note /// %{n1:;}note
http_access allow all an1 an2
access_log daemon:/tmp/notes.log logformat=notes all
```
the output is:
```
"n1: n1v1\r\nn1: n1v2\r\nn1: n1v3\r\nn2: n2v1\r\nn3: n3v1\r\n" /// n1v1,n1v2,n1v3 /// n1v1;n1v2;n1v3
```

I think I understand better now.
We have 3 toString functions, that are used in different contexts, and it's not clear which is which.
Note::toString is used for Format's %{name}note . NotePairs::toString is used for %note , and Notes::toString which is used for ACLAnnotationData::dump 

And the latter is right now broken:
```
acl an2 annotate_transaction , n1: n1v3
n2: n2v1

acl an1 annotate_transaction , n1: n1v1
n1: n1v2
```


> 
> And, yes, we also need to deal with multiple same-name annotations, and we have a default value separator for the parser (and, IIRC, at least two default delimiters for the printer -- annotation delimiter and individual annotation value delimiter).

rousskov:
> > For example, what is are the values in a `Foo: bar,baz;boo+error=2` annotation? There is no single correct answer to that question! To one admin, the answer is `bar` and `baz;boo+error=2`. To another, it is `bar,baz;boo` and `error=2`. The admin has to tell us what value syntax they use. Without that information, we only know that the annotation is named "Foo" and has one or more values.

> > Once we know the separator, our find() function then can find individual values. This find() step is not about printing/reporting/etc. It is about _parsing_!
 
> Sorry, I don't follow you. 

I may be able to help if you detail what prevents you from following the above example.

> In this configuration example: ...

Please note that you configuration example does not match my example and cannot illustrate why find() is not about printing but about parsing. Try something like this (untested!) example instead:

```
# annotate using a single annotation
# the annotation value interpretation/parsing is up to the recipient
acl mark4 annotate_transaction n4=4a;4b,4c+4d
http_access allow mark4

# matches only when annotation named n4 has a `4d` value (in a `+` separated list of values).
acl n4with4d note -m=+ n4 4d

# logs these two values for that n4 annotation: `4a` and `4b,4c+4d`
# none of these two values are a `4d` value from above,
# even though the annotation has not changed!
logformat xsquid n4=%note{n4:;}
access_log access_log stdio:/tmp/access.log logformat=xsquid n4with4d
```

All of the above should work during the same HTTP transaction! For the above to work, we need find() to know how to parse an annotation to find individual values in a specific context; each context comes with its own value delimiters. 

What happens with those values after find() extracts them is a different story.

> We have 3 toString functions, that are used in different contexts, and it's not clear which is which. Note::toString is used for Format's %{name}note . NotePairs::toString is used for %note , and Notes::toString which is used for ACLAnnotationData::dump
 
> And the latter is right now broken:
> 
> ```
> acl an2 annotate_transaction , n1: n1v3
> n2: n2v1
> 
> acl an1 annotate_transaction , n1: n1v1
> n1: n1v2
> ```

This bug may help us improve this code without breaking promises, but that remains to be seen.

kinkie:
@rousskov kind reminder, could you review this? Together with PR 1574 this will solve an outstanding bug

src/acl/AnnotationData.cc
```diff
@@ -23,8 +24,8 @@ SBufList
 ACLAnnotationData::dump() const
 {
     SBufList sl;
-    if (const char *strNotes = notes->toString())
-        sl.push_back(SBuf(strNotes));
+    if (!notes->empty())
+        sl.push_back(notes->dump());
```
rousskov, line 28: I assume this is the only place where we call the new Notes::dump() method. Please convert that method to accept an std::ostream reference and use SBufStream here to convert dump() output into an SBufList. The current PR code essentially does the same, but uses SBufStream inside Notes::dump() which complicates that method reuse and slows down the ongoing transition from SBufList to streams.

src/Notes.h
```diff
@@ -247,7 +249,7 @@ class NotePairs: public RefCountable
 
     /// Convert NotePairs list to a string consist of "Key: Value"
     /// entries separated by sep string.
-    const char *toString(const char *sep = "\r\n") const;
+    SBuf toString(const char *sep = "\r\n") const;
```
rousskov, line 252: Just for my information, does PR code ever call this method without arguments (i.e. relying on the CRLF default)? I know that we provide a CRLF argument in some cases, but are there any calls of this method without any argument in the PR code?

rousskov (changes requested):
> @rousskov kind reminder, could you review this?

I have not forgotten and will work on this PR as soon as I get a chance. I left a few minor change requests to help you make progress. This PR may need more changes, but I cannot know what serious changes, if any, will be needed by quickly looking through the diff. I hope to find time to work on this within 10 days.

FWIW, I plan to start with preparing a _comprehensive_ manual test in the form of a squid.conf snippet (that tickles all relevant dumping/toString() variants in all relevant ways) and collect past and current outputs (access.log and mgr:config?) to see what exactly has changed and what output bugs, if any, remain. If somebody does that legwork for me, I will not need to find less time for the followup review.

> Together with PR 1574 this will solve an outstanding bug

If this PR fixes a bug, then its PR title/description are stale and should be updated accordingly. That is another action that, ideally, should not be on _my_ to-do list.

src/Notes.h
kinkie: Yes. In fact, that's the reason why the `acl annotate_transaction` output is currently broken.
In fact, as far as I can tell this code is only really used in cachemgr, and is mildly overengineered.

rousskov: > Alex: does PR code ever call this method without arguments (i.e. relying on the CRLF default)? ... are there any calls of this method without any argument in the PR code?

> Francesco: Yes.

Where?

kinkie: Doh, sorry, I got confused.
There used to be one such caller, ACLAnnotationData::dump. The PR removes it, so now AFAICT there's none.

rousskov:
_Status update_: I fixed build and an ACL configuration reporting bug, but more changes are needed to finish this PR. I will come back to this.

rousskov (approved):
I am done polishing this code and am approving this PR. I have tried to document what this PR fixed in the updated PR title/description. Please review and retest.

kinkie:
@rousskov thanks for working on fixing and polishing this PR so much!
I'll give it a check, but it looks great

kinkie:
runtime-tested, clearing for merge

PR #1517, opened 2023-10-13
Fix stack buffer overflow when parsing Digest Authorization

nonsleepr:
The bug was discovered and detailed by Joshua Rogers at
https://megamansec.github.io/Squid-Security-Audit/digest-overflow.html
where it was filed as "Stack Buffer Overflow in Digest Authentication".

### DIFF
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -833,6 +833,7 @@ Auth::Digest::Config::decode(char const *proxy_auth, const HttpRequest *request,
                 debugs(29, 9, "Found noncecount '" << digest_request->nc << "'");
             } else {
                 debugs(29, 9, "Invalid nc '" << value << "' in '" << temp << "'");
+                digest_request->nc[0] = 0;
             }
             break;
 
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -828,6 +828,7 @@ Auth::Digest::Config::decode(char const *proxy_auth, const HttpRequest *request,
         case DIGEST_NC:
             if (value.size() == 8) {
                 // for historical reasons, the nc value MUST be exactly 8 bytes
+                static_assert(sizeof(digest_request->nc) == 8 + 1);
                 xstrncpy(digest_request->nc, value.rawBuf(), value.size() + 1);
                 debugs(29, 9, "Found noncecount '" << digest_request->nc << "'");
             } else {
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -827,8 +827,7 @@ Auth::Digest::Config::decode(char const *proxy_auth, const HttpRequest *request,
 
         case DIGEST_NC:
             if (value.size() == 8) {
-                // For historical reasons, the nc value MUST be exactly 8 hexadecimal
-                // digits.
+                // for historical reasons, the nc value MUST be exactly 8 bytes
                 xstrncpy(digest_request->nc, value.rawBuf(), value.size() + 1);
                 debugs(29, 9, "Found noncecount '" << digest_request->nc << "'");
             } else {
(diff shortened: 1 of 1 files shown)


yadij:
OK to test

MegaManSec:
Note: an alternative patch is provided in https://github.com/squid-cache/squid/security/advisories/GHSA-phqj-m8gv-cq4g but the one in this PR is probably more appropriate.

nonsleepr:
> Note: an alternative patch is provided in https://github.com/squid-cache/squid/security/advisories/GHSA-phqj-m8gv-cq4g but the one in this PR is probably more appropriate.

The advisory isn't public though.

@MegaManSec, since all (I assume all) vulns found were disclosed, it might make sense to open GHSAs, that way the info would propagate to various threat intel platforms and bring more attention to the issues.

src/auth/digest/Config.cc
```diff
@@ -826,11 +826,14 @@ Auth::Digest::Config::decode(char const *proxy_auth, const HttpRequest *request,
             break;
 
         case DIGEST_NC:
-            if (value.size() != 8) {
+            if (value.size() == 8) {
+                // For historical reasons, the nc value MUST be exactly 8 hexadecimal
+                // digits.
```
rousskov, line 831: Since this code does not validate the contents of `nc`, I would relax the comment to avoid implying otherwise:

```suggestion
                // for historical reasons, the nc value MUST be exactly 8 bytes
```

src/auth/digest/Config.cc
```diff
@@ -826,11 +826,14 @@ Auth::Digest::Config::decode(char const *proxy_auth, const HttpRequest *request,
...
+                xstrncpy(digest_request->nc, value.rawBuf(), value.size() + 1);
```
rousskov, line 832: To explain why this terrible code "works" and protect it from refactoring bugs, I would add an assertion like this (untested) one:

```suggestion
                static_assert(sizeof(digest_request->nc) == 8 + 1);
                xstrncpy(digest_request->nc, value.rawBuf(), value.size() + 1);
```

src/auth/digest/Config.cc
```diff
@@ -826,11 +826,14 @@ Auth::Digest::Config::decode(char const *proxy_auth, const HttpRequest *request,
...
+                debugs(29, 9, "Found noncecount '" << digest_request->nc << "'");
+            } else {
                 debugs(29, 9, "Invalid nc '" << value << "' in '" << temp << "'");
```
rousskov, line 835: AFAICT, this code is inside a loop, and, hence, might see a dozen different `nc` values, some valid, some not. The last value wins. Unless it is invalid (then some earlier value wins; if any). Such inconsistencies often trigger bugs. Unfortunately, this code only validates values _after_ they are stored, but we can at least ensure that belated validation code always looks at the last seen value (or nothing):

```suggestion
                debugs(29, 9, "Invalid nc '" << value << "' in '" << temp << "'");
                digest_request->nc[0] = 0;
```

rousskov (approved):
Thank you for posting this fix! The changes looks OK to me. The overall surrounding code quality is so low, that it is difficult to decide what should be adjusted while fixing this bug. I have left a few suggestions, but I do not insist on any changes.

I fixed PR description [formatting](https://github.com/measurement-factory/anubis/blob/master/README.md#pr-description) and polished PR title for precision. Please adjust further as needed.

@MegaManSec, you are still mentioned in the PR description, so you may want to validate its wording.

rousskov (dismissed):

yadij (approved):

rousskov (approved):

squidadm:
queued for backport to v6

zmanion:
This (and the other bugs identified by @MegaManSec) would be much easier to track with CVE IDs (which do not require fixes, just sufficient and public evidence of vulnerability). If help assigning and publishing CVE IDs is wanted, let me know.

PR #1519, opened 2023-10-14
Upgrade AsList separators to support c-strings

kinkie:


src/base/IoManip.h
```diff
@@ -109,17 +108,17 @@ class AsList
 public:
     explicit AsList(const Container &c): container(c) {}
 
-    /// a character to print before the first item (if any)
-    auto &prefixedBy(const char ch) { prefix = ch; return *this; }
+    /// a c-string to print before the first item (if any). Caller must ensure lifetime.
```
rousskov, line 111: FWIW, I have considered using SBuf to eliminate these lifetime concerns, but decided that we should not do it (yet) since existing callers do not need this, and there is a chance that no future caller will need this.

rousskov (approved):
Thank you for this improvement.

I have edited PR title/description and committed a small correction. Please adjust further as needed.

Please avoid adding things like "This is a prerequisite to use it to implement PR #1516" to PR descriptions. PR description is a future commit message body. This information _is_ useful, but it should be added as a second/regular PR comment, especially when we do not know for sure what will happen in #1516 yet. This way, we can merge this PR without developing consensus about another PR destiny and without adding "forward" references to PRs that might never merge or merge in a very different state.

src/base/IoManip.h
```diff
@@ -130,11 +129,11 @@ operator <<(std::ostream &os, const AsList<Container> &manipulator)
     for (const auto &item: manipulator.container) {
         if (!opened) {
             if (manipulator.prefix)
-                os << *manipulator.prefix;
+                os << manipulator.prefix;
             opened = true;
         } else {
             if (manipulator.delimiter)
-                os << *manipulator.delimiter;
+                os << manipulator.delimiter;
```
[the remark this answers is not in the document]
kinkie: I had the same thought

src/base/IoManip.h
kinkie: Again, thought. In practice, it's highly likely that the separators will be literals, so no lifetime concerns.
OTOH, bringing SBuf in would mean that all the required machinery would also need to be pulled in.
FWIW, I considered using std::string, but it also felt overkill for this specific purpose

yadij: std::string may be more efficient long-term with the C++17 `""s` syntax constants. But that can be an optimization issue for later.

yadij (approved):

PR #1522, opened 2023-10-16
v5 Next Backports

squidadm:


kinkie:
Three options here: either backporting commit 512b96be30 (PR #1429), hand-merging, or stopping tests for clang on too-modern OSes. @yadij , what do you think is the best way forward?

yadij:
Making sure the backport does not break older OS is the goal. Which the `static_assert` touchup seems to have done. The OS complaining about bitmasks are new enough not to be needing this v5 patch.

PR #1530, opened 2023-10-20
negotiate_wrapper_auth: protect from responses over 64KB

rousskov:
... received from NTLM and Kerberos helpers.

This code uses MAX_AUTHTOKEN_LEN (~64KB) buffers to read response lines.
fgets(3) guarantees to terminate the supplied buffer, but it does not
return nil when the input line is larger than the buffer. We have
already detected such "Oversized message" cases for fgets(stdin) calls,
but not for fgets(FDNOUT) and fgets(FDKOUT) calls.

### DIFF
--- a/src/auth/negotiate/wrapper/negotiate_wrapper.cc
+++ b/src/auth/negotiate/wrapper/negotiate_wrapper.cc
@@ -224,6 +224,13 @@ processingLoop(FILE *FDKIN, FILE *FDKOUT, FILE *FDNIN, FILE *FDNOUT)
                         LogTime(), PROGRAM);
                 return 0;
             }
+
+            if (!memchr(tbuff, '\n', sizeof(tbuff) - 1)) {
+                fprintf(stderr, "%s| %s: Oversized NTLM helper response\n",
+                        LogTime(), PROGRAM);
+                return 0;
+            }
+
             /*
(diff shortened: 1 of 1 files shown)


src/auth/negotiate/wrapper/negotiate_wrapper.cc
```diff
@@ -224,6 +224,13 @@ processingLoop(FILE *FDKIN, FILE *FDKOUT, FILE *FDNIN, FILE *FDNOUT)
                         LogTime(), PROGRAM);
                 return 0;
             }
+
+            if (!memchr(tbuff, '\n', sizeof(tbuff) - 1)) {
+                fprintf(stderr, "%s| %s: Oversized NTLM helper response\n",
+                        LogTime(), PROGRAM);
```
rousskov, line 230: This PR code mimics a similar memchr() check and "Oversized" error message located earlier in this code, where fgets(stdin) is called.

rousskov (commented):
I have not tested this code. I do not insist on merging this PR. I found this minor bug when working on an essentially unrelated PR... If you would like to adjust something, please feel free to edit the PR branch or PR metadata.

yadij (approved):

kinkie (approved):

PR #1531, opened 2023-10-21
fuzzing add fuzz testing support

0x34d:
Purpose:
This PR aims to integrate `squid-cache` with the [oss-fuzz](https://google.github.io/oss-fuzz) platform for ongoing fuzz testing.

Run:

```bash
export CFLAGS="-O1 -fno-omit-frame-pointer -gline-tables-only -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link"
export CXXFLAGS="-O1 -fno-omit-frame-pointer -gline-tables-only -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link"
export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"

./bootstrap.sh
./configure --disable-shared --enable-fuzzer=$LIB_FUZZING_ENGINE
make -j$(nproc)

make -C lib/ fuzzRFC1738
make -C src/ fuzzHttp1Parser fuzzRFC1035
```


src/Makefile.am
```diff
@@ -2746,3 +2746,56 @@ tests_testEventLoop_LDADD = \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 tests_testEventLoop_LDFLAGS = $(LIBADD_DL)
+
+if ENABLE_FUZZ_DECODE
+EXTRA_PROGRAMS =+ fuzzHttp1Parser fuzzRFC1035
+
+fuzzHttp1Parser_SOURCES = \
```
yadij, line 2753: We already have extensive edge case testing for the `Http1::Parser` class. I do not believe fuzzer adds any benefit here.

lib/Makefile.am
```diff
@@ -77,3 +77,13 @@ tests_testRFC1738_LDADD= \
 tests_testRFC1738_LDFLAGS = $(LIBADD_DL)
 
 testHeaders: $(top_srcdir)/include/*.h
+
+if ENABLE_FUZZ_DECODE
+EXTRA_PROGRAMS = fuzzRFC1738
+fuzzRFC1738_SOURCES = tests/fuzzRFC1738.cc
```
yadij, line 83: I do not see any benefit here. This function has a deterministic mapping of inputA to outputB. While we are currently lacking a test of the mapping, a fuzzer is the least efficient way to test.

yadij (changes requested):

lib/Makefile.am
0x34d: I agree with that.

src/Makefile.am
0x34d: Well, this PR primarily aims to integrate `squid-cache` into the [oss-fuzz](https://google.github.io/oss-fuzz) infrastructure for continuous fuzzing.

rousskov: > We already have extensive edge case testing for the Http1::Parser class. I do not believe fuzzer adds any benefit here.

FWIW, I disagree that testing Http1::Parser using fuzzers adds no benefit.

src/tests/fuzzHttp1Parser.cc
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    SBuf ioBuf;
+    Http1::RequestParser hp;
+
+    ioBuf.clear();
+    ioBuf.append(data, size);
+
+    hp.clear();
+    hp.parse(ioBuf);
```
rousskov, line 47: If fuzzer is able to find a Squid bug, how will that bug manifest itself in this test? A crash?

What happens if Http::One::RequestParser::parse() throws an exception?

BTW, did the proposed tests flagged any problems in covered Squid parsers?

src/Makefile.am
```diff
@@ -2746,3 +2746,56 @@ tests_testEventLoop_LDADD = \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 tests_testEventLoop_LDFLAGS = $(LIBADD_DL)
+
+if ENABLE_FUZZ_DECODE
+EXTRA_PROGRAMS =+ fuzzHttp1Parser fuzzRFC1035
+
```
rousskov, line 2752: I am not yet convinced that the design proposed by this PR (i.e. a dedicated test program for each parsing function) is the right one because it requires building stand-alone programs to test (often rather deeply embedded/integrated) parsers. A similar design for some of the existing Squid unit tests has wasted numerous hours on maintaining the surrounding scaffolding just to make the tests compile, usually with very little payoff.

Are there any oss-fuzz integration alternatives where fuzzing does _not_ require creating dedicated unit tests and, roughly speaking, uses primary Squid code, with Squid own main() function, Squid own configuration code, etc.? If yes, what are their pros and cons compared to the proposed integration approach?

src/tests/fuzzHttp1Parser_seed_corpus.zip
rousskov, line 1: If possible, please point me to some documentation that explains the role of these seed files and how they are supposed to be created. At the first glance, the current contents of this file suggests testing with a handful of basic input sequences (which could be fed to Squid without all this integration cost!), but perhaps that first impression is totally wrong.

rousskov (changes requested):
Thank you for offering oss-fuzz integration.

IMO, Squid Project should accept high-quality PRs that offer fuzzing integration (after fixing bugs exposed by the proposed integration, if any). However, arriving at such integration requires making a few difficult Squid design decisions. I have serious concerns about decisions made by this PR and left two primary change requests in hope to facilitate the corresponding discussion.

I am ignoring secondary implementation problems in the current PR. They can be addressed after the architectural issues are resolved.

P.S. I am aware of two other attempts to integrate Squid with oss-fuzz, but all of them failed. The last one was back in April 2021. Back then, the idea did not receive enough Project support to proceed beyond initial private steps.

MegaManSec:
FWIW, I have no problem with sharing my (extremely edited) Squid code which I used in the fuzzing segment of my research.

A few branches exist: URN response parsing, ESI markup tag fuzzer, FTP response parsing, client request parsing, server response parsing, and server response (with caching of data) parsing. Unfortunately I didn't keep good track of other fuzzing harnesses, and random commits in a 'master' branch include: "Gopher server response", "URI parsing", "client http headers"(HttpHeader::parse), HttpHdrCc, HttpHdrSc, httpHeaderParseQuotedString, httpHeaderParseQuotedString, chunked encoding (handleRequestBodyData), SquidMD5Init/Update/Final/MD5, HttpReply::parse, htcpHandleMsg.

In my branches like client request parsing, server response parsing, and server response (with caching of data) parsing, I was able to rewrite Squid to be a fully fledged binary which could use afl++'s persistent mode and effectively accept unlimited amount of requests/responses and parse them completely and only drop them once all "the work had been done" in relation to parsing, and in the case of server response, caching.

Let me give an example of a very early example of the "client request" fuzzing method. Note: this will not work alone, and there's a lot of small changes that were needed for this to be succesful, and changes have to be made to ensure high code coverage, such as:  "bool intercepted = true;" in src/RequestFlags.h. Removing some calls to "Must(IsConnOpen(conn));", sendComplete, commSetConnTimeout, and rewriting "AsyncJob::Start" calls to not be Asyncjobs (AsyncJob::Start(new HttpStateData(fwd)); becomes auto p = new HttpStateData(fwd); p->start();" for example but the correct "order" must be set. Setting want.ai_flags = AI_NUMERICHOST; in src/ip/Address.cc to avoid DNS lookups, and specifically defining the file descriptors as stdin was already required. ClientHttpRequest::doCallouts was NOP'd, Comm::Write was NOP'd, some of PeerSelector::resolveSelected was NOP'd, TunnelStateData::Connection::initConnection was NOP'd, part of ACLChecklist::goAsync was NOP'd, commSetNonBlocking was NOP'd, and so on. There are a lot of things to change.

`void Server::doitServer(const char *buf, size_t len) {
	reader = NULL;
//	maybeMakeSpaceAvailable();
	CommIoCbParams rd(this);
	rd.flag = Comm::OK;
	rd.size = len;
	inBuf.append(buf,len);

	statCounter.client_http.kbytes_in += len;
	if (!receivedFirstByte_)
		receivedFirstByte(); //does nothing..
}

__AFL_FUZZ_INIT();
int main() {
   Mem::Init();
   AnyP::UriScheme::Init();
   storeFsInit();
   Acl::Init();
   Store::Init();
   parseConfigFile(DEFAULT_CONFIG_FILE);
//    setMaxFD();
    fde::Init();
   comm_init();
   httpHeaderInitModule();
   errorInitialize();
storeInit();
//Dns::Init();
Debug::parseOptions("ALL,3");
	__AFL_INIT();
	unsigned char *b = __AFL_FUZZ_TESTCASE_BUF;
	while(__AFL_LOOP(500)) {
		size_t len = __AFL_FUZZ_TESTCASE_LEN;
		if(len >= 4096) continue;
		char buf[4096];
		memset(buf,'\0',4096);
		memcpy(buf,b,len);
		MasterXaction::Pointer mx = new MasterXaction(XactionInitiator::initClient);
		mx->squidPort = new AnyP::PortCfg();
		mx->squidPort->transport = Http::ProtocolVersion(1,1);
		mx->tcpClient = new Comm::Connection;
		ConnStateData *serv = Http::NewServer(mx);
		serv->doitServer(buf,len);
		serv->clientParseRequests();
	}
	exit(0);

}`

I could work on cleaning up the patches to make it more.. observable, but I don't know how useful it would be? It would take until the end of the year as I have other things to focus on.

Note: if this does ever end up in oss-fuzz, I think it would be kind of the contributor to donate the $ reward to The Squid Project. just my 2cents abotu that.

Cheers,
Josh

rousskov:
> Let me give an example of a very early example of the "client request" fuzzing method. Note: this will not work alone, and there's a lot of small changes that were needed for this to be succesful. ... I could work on cleaning up the patches ..., but I don't know how useful it would be?

@MegaManSec, thank you for providing that example and sharing your design! FWIW, compared to parser call wrappers in the current PR, your example looks closer to what Squid, IMHO, needs as far as the overall traffic fuzzing design is concerned, but I hope it is possible to get a lot further in that direction. I realize that it is much easier to implement fuzzing that way for a specific test project, but we do not want to sprinkle official Squid code with "turn this off when fuzzing" conditionals and then work hard on maintaining the integrity of the mutilated code to avoid false negatives during tests. Hopefully, we do not really have to do much of that, but I do not know enough about fuzzing needs to be sure.

The primary bootstrap information I am currently missing is the expected number of tests for a single test point. For example, how many distinct requests are we expecting to feed the HTTP request parser with? This PR fuzzHttp1Parser_seed_corpus files appear to contain a handful of tests. The __AFL_LOOP(500) example above may imply 500.  Is ~1000 tests the expected number, ballpark? Or am I completely misinterpreting these signals? I have a related change request: https://github.com/squid-cache/squid/pull/1531#discussion_r1367778912

src/tests/fuzzHttp1Parser.cc
0x34d: > If fuzzer is able to find a Squid bug, how will that bug manifest itself in this test? A crash?

A crash report is generated by `oss-fuzz infra` and an email is sent to the [contact](https://google.github.io/oss-fuzz/getting-started/new-project-guide/#primary). After downloading the crash file, the bug can be [reproduced](https://google.github.io/oss-fuzz/advanced-topics/reproducing).




> What happens if Http::One::RequestParser::parse() throws an exception?

Done thanks for pointing it out  https://github.com/squid-cache/squid/pull/1531/commits/2aec1c227fbe198d9e040bb20ac4ba05aa640c9c

> BTW, did the proposed tests flagged any problems in covered Squid parsers?

Yes, a heap-buffer-overflow in `rfc1035MessageUnpack`;  I send a bug report to `squid-bugs@lists.squid-cache.org`

src/Makefile.am
0x34d: > I am not yet convinced that the design proposed by this PR (i.e. a dedicated test program for each parsing function) is the right one because it requires building stand-alone programs to test (often rather deeply embedded/integrated) parsers. A similar design for some of the existing Squid unit tests has wasted numerous hours on maintaining the surrounding scaffolding just to make the tests compile, usually with very little payoff.

Yes, each parser requires its own dedicated fuzzer for effective testing. Combining all tests into a single program is bad. For example, if the fuzzer generates a DNS query, this query would also be parsed by the HTTP1 parser, resulting in false execution paths. This will confuse the feedback fuzzing used by OSS-Fuzz{LibFuzzer, AFL, and Honggfuzz} this will generate bad low-quality input for fuzzing.


> Are there any oss-fuzz integration alternatives where fuzzing does not require creating dedicated unit tests and, roughly speaking, uses primary Squid code, with Squid own main() function, Squid own configuration code, etc.? If yes, what are their pros and cons compared to the proposed integration approach?

Unfortunately, that can't be done.

src/tests/fuzzHttp1Parser_seed_corpus.zip
0x34d: Documentation on [seed](https://llvm.org/docs/LibFuzzer.html#corpus).

For `rfc1035MessageUnpack`, the base is [testRFC1035.cc](https://github.com/squid-cache/squid/blob/master/src/tests/testRFC1035.cc). All the `const char *buf` data is used as seed.

For `Http::One::RequestParser::parse`, the base is [testHttp1Parser.cc](https://github.com/squid-cache/squid/blob/master/src/tests/testHttp1Parser.cc). However, for these seeds, I use my own old HTTP data files.

0x34d:
What @MegaManSec offers is great. The Squid project can use the -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION macro for those small changes

0x34d:
> The primary bootstrap information I am currently missing is the expected number of tests for a single test point. For example, how many distinct requests are we expecting to feed the HTTP request parser with? This PR fuzzHttp1Parser_seed_corpus files appear to contain a handful of tests. The __AFL_LOOP(500) example above may imply 500. Is ~1000 tests the expected number, ballpark? Or am I completely misinterpreting these signals? I have a related change request: https://github.com/squid-cache/squid/pull/1531#discussion_r1367778912

usually 50 million to 100 million execution.

MegaManSec:
>FWIW, compared to parser call wrappers in the current PR, your example looks closer to what Squid, IMHO, needs as far as the overall traffic fuzzing design is concerned, but I hope it is possible to get a lot further in that direction.

Both are useful. For example, if a 'overall Squid fuzzer' which handles HTTP requests and HTTP responses, then DNS will be completely unfuzzed since it (should) never even touch the DNS code (at least for receiving DNS responses) The same then goes for, let's say, ICAP.

Something else I did which is probably also not really possible with oss-fuzz is once I had fuzzed the HTTP requests and HTTP responses with high enough code coverage, I ran a real Squid instance which would treat every request as a a request for localhost:1234. Then, a script would randomly send one request to Squid, and localhost:1234 would randomly send a response. This way, bugs which required a specific request and then a specific response, were picked up. I don't think this would be possible in oss-fuzz, unless some ["magic byte"](https://blogs.opera.com/security/2021/05/fuzzing-http-proxies-privoxy-part-1/) is used, but this can be slow.

lib/tests/fuzzRFC1738.cc
```diff
@@ -0,0 +1,36 @@
...
+    if (size < kMinInputLength || size > kMaxInputLength) {
+        return 0;
+    }
+
+    char *data_in = (char*)calloc(size + 1, sizeof(char));
+    memcpy(data_in, data, size);
+
+    rfc1738_unescape(data_in);
+
+    rfc1738_do_escape(data_in, RFC1738_ESCAPE_UNSAFE);
```
MegaManSec: Would it be sensible to assert if an decode(encode(data))) != data, for example?

0x34d: This test has been removed. Anyway, it's not cryptography, so differential fuzzing isn't worth it.

MegaManSec: >it's not cryptography, so differential fuzzing isn't worth it.

I don't see the relevance of cryptography and whether it's "worth it". Differential fuzzing of these functions can pick up security-related issues such as request smuggling, for example.

rousskov:
> Typically, 50 million to 100 million executions, or until it stops finding new paths in the last 3 hours.

Thank you for that estimate! 100 million test cases at, say, 100 microseconds each is 2.8 hours. This is doable for testing staged PR commits, especially if we can parallelize some tests. It also does not preclude expanding testing to longer code paths than a basic parser function, but we need to be careful to keep response times under 100 microseconds (the smaller the better, of course).

0x34d:
> > Typically, 50 million to 100 million executions, or until it stops finding new paths in the last 3 hours.
> 
> Thank you for that estimate! 100 million test cases at, say, 100 microseconds each is 2.8 hours. This is doable for testing staged PR commits, especially if we can parallelize some tests. It also does not preclude expanding testing to longer code paths than a basic parser function, but we need to be careful to keep response times under 100 microseconds (the smaller the better, of course).


If the project integrates with `oss-fuzz`, you don't need to worry about it. At most, for PR fuzzing, [ci-fuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) with `300sec` is enough, and heavy loading is done by Google's cluster.

If you want to fuzz it on a local machine, use around 8 threads of libFuzzer for 12-16 hours, and of course, watch out for roadblocks.

lib/tests/fuzzRFC1738.cc
0x34d: file is deprecated

rousskov:
> > FWIW, compared to parser call wrappers in the current PR, your example looks closer to what Squid, IMHO, needs as far as the overall traffic fuzzing design is concerned, but I hope it is possible to get a lot further in that direction.

> Both are useful.

Usefulness is not the question. The question is whether the overall cost of fuzzing design/approach X is worth its expected benefit (in Squid context). If the answer is "yes" for designs X and Y, then the question is whether the cost of doing _both_ X and Y is worth the expected benefit. As of now, I have a fast-growing suspicion that the "one program per test" approach is inferior to "embedded in Squid" design based on cost/benefit analysis, but there are still a few unknowns I need to resolve to be sure.

> For example, if a 'overall Squid fuzzer' which handles HTTP requests and HTTP responses, then DNS will be completely unfuzzed since it (should) never even touch the DNS code (at least for receiving DNS responses) The same then goes for, let's say, ICAP.

Right now, I am not comparing "overall Squid fuzzer" and "parser P fuzzer". I am comparing "function F fuzzer is a stand-alone program" (i.e. this PR) and "function F fuzzer as a optional part of a Squid executable (or equivalent)" (i.e. the extrapolation of your [example](https://github.com/squid-cache/squid/pull/1531#issuecomment-1773908420)).

> Something else I did which is probably also not really possible with oss-fuzz is once I had fuzzed the HTTP requests and HTTP responses with high enough code coverage, I ran a real Squid instance which would treat every request as a a request for localhost:1234. Then, a script would randomly send one request to Squid, and localhost:1234 would randomly send a response.

Yes, we already have (hand-written) tests that do that sort of thing. Those particular hand-written tests are way too slow to complete 100 million auto-fuzzed test cases in a few hours. It remains to be seen whether we can design to afford "whole Squid" fuzzers (for the lack of a better term). I suspect that the answer is "yes, we can". Enabling that future possibility is one of the items in my cost-benefit analysis discussed above.

rousskov:
> > > Typically, 50 million to 100 million executions, or until it stops finding new paths in the last 3 hours.
> > 
> > 
> > Thank you for that estimate! 100 million test cases at, say, 100 microseconds each is 2.8 hours. This is doable for testing staged PR commits, especially if we can parallelize some tests. It also does not preclude expanding testing to longer code paths than a basic parser function, but we need to be careful to keep response times under 100 microseconds (the smaller the better, of course).
 
> If the project integrates with `oss-fuzz`, you don't need to worry about it. 

I have to worry about it because oss-fuzz is not magic: If, for example, one test case takes ten seconds, then oss-fuzz will not complete 100 million test cases in reasonable time.

> At most, for PR fuzzing, [ci-fuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) with `300sec` is enough

It is "enough" for a certain mean test case response time. Based on your earlier estimate of 100 million test cases, we now know a ballpark upper bound for an average test case: 100 microseconds.

And, yes, I am pretty sure we will require CIFuzz or similar CI integration. Running all fuzzing tests on a "local machine" is not a particularly relevant use case IMO (rerunning a single _failed_ fuzzing test case is).

0x34d:
> I have to worry about it because oss-fuzz is not magic: If, for example, one test case takes ten seconds, then oss-fuzz will not complete 100 million test cases in reasonable time.

100 million is not a magic number, this is for my private testing. on `oss-fuzz cluster` fuzzes fire up when a merge happens in a git repo.

image

rousskov:
> > I have to worry about it because oss-fuzz is not magic: If, for example, one test case takes ten seconds, then oss-fuzz will not complete 100 million test cases in reasonable time.
> 
> 100 million is not a magic number, this is for my private testing. on `oss-fuzz cluster` fuzzes fire up when a merge happens in a git repo. it's less about execution and more about coverage.

We need an test case count (i.e. an acceptable test case response time) estimate to evaluate fuzzing design approaches. "Enough coverage" is not helpful for that purpose AFAICT. If 100 million is not a reasonable ballpark estimate for this purpose, please say so.

0x34d:
> > > I have to worry about it because oss-fuzz is not magic: If, for example, one test case takes ten seconds, then oss-fuzz will not complete 100 million test cases in reasonable time.
> > 
> > 
> > 100 million is not a magic number, this is for my private testing. it's less about execution and more about coverage.
> 
> We need a test case count (i.e. an acceptable test case response time) estimate to evaluate fuzzing design approaches. "Enough coverage" is not helpful for that purpose AFAICT. If 100 million is not a reasonable ballpark estimate for this purpose, please say so.


Set any number of test count for `parser by parser`, then try to get LCOV coverage for those tests and examine their behavior. you need to know what you're doing. 
Sometimes you think you're fuzzing a decoder, but in reality, you're hitting the same brick wall because there is a crc32.

src/tests/fuzzHttp1Parser_seed_corpus.zip
rousskov, line 1: Do we have to commit zip files? Would it be possible to commit raw input files instead, so that each of them can be maintained/manipulated separately using git commands? I do not see zip archives mentioned in LibFuzzer corpus [documentation](https://llvm.org/docs/LibFuzzer.html#corpus), but I did not dig deeper.

rousskov:
> > We need a test case count (i.e. an acceptable test case response time) estimate to evaluate fuzzing design approaches. "Enough coverage" is not helpful for that purpose AFAICT. If 100 million is not a reasonable ballpark estimate for this purpose, please say so.
 
> Set any number of test count for `parser by parser`, then try to get LCOV coverage for those tests and examine their behavior. you need to know what you're doing. Sometimes you think you're fuzzing a decoder, but in reality, you're hitting the same brick wall because there is a crc32.

To avoid misunderstanding, I am counting on you, the PR author, to provide the corresponding reasonable estimates. I do understand that the actual test count (to achieve reasonable coverage) depends on many factors. I would like to avoid doing the legwork required for me, a PR reviewer, to obtain those estimates.

rousskov (changes requested):

MegaManSec:
@rousskov: Is it possible to start a "Github discussion" around this, since this discussion goes beyond this PR?
I have some diffs that I can provide to give an indication of how I approached fuzzing. We could always move this to a mailing list, but it's a bit easier for me here.

I'm a bit surprised to hear about the rfc1035MessageUnpack bug, as I also fuzzed that:
```
diff --git a/src/dns/rfc1035.cc b/src/dns/rfc1035.cc
index 737b3c5b8..419441d89 100644
--- a/src/dns/rfc1035.cc
+++ b/src/dns/rfc1035.cc
@@ -748,6 +748,19 @@ rfc1035SetQueryID(char *buf, unsigned short qid)
     unsigned short s = htons(qid);
     memcpy(buf, &s, sizeof(s));
 }
+/*__AFL_FUZZ_INIT();
+
+int main() {
+   unsigned char *b = __AFL_FUZZ_TESTCASE_BUF;
+   while(__AFL_LOOP(1000)) {
+      size_t len = __AFL_FUZZ_TESTCASE_LEN;
+      if(len >= 8192) continue;
+               rfc1035_message *message = NULL;
+               rfc1035MessageUnpack((char*)b, len, &message);
+   }
+   return 0;
+}*/
+
```

Also an 'lcov' coverage report of when I ran a standard Squid installation with the large list of previously generated unique (in the sense of code coverage) requests & server combinations: https://megamansec.github.io/squid-report/ may be interesting for someone.

rousskov:
> @rousskov: Is it possible to start a "Github discussion" around this, since this discussion goes beyond this PR?

Enabling GitHub Discussions is a Project decision. Personally, I would not object to it if others really want them, although I expect that adding GitHub Discussions to the two existing mailing lists (squid-users and squid-dev) will increase net Project costs/overhead for the foreseeable future. That overhead problem can be addressed by shutting down mailing lists (i.e. moving completely to Discussions), but that, in itself, is a controversial action that is likely to generate a lot of negativity and is unlikely to gain a lot of traction, I suspect.

> I have some diffs that I can provide to give an indication of how I approached fuzzing. We could always move this to a mailing list, but it's a bit easier for me here.

If you have Squid code to share but it is not ready or not appropriate for merging, I recommend posting a Draft PR instead. In the description, state that you are _not_ suggesting to merge these changes; you are suggesting to discuss related topics X and Y instead. Please note that most of us are busy addressing the vulnerabilities you have published; I would not expect a lot of feedback in the near term.

src/tests/fuzzHttp1Parser.cc
```diff
@@ -0,0 +1,54 @@
...
+    ioBuf.clear();
+    ioBuf.append(data, size);
+
+    try {
+        hp.parse(ioBuf);
+    } catch (...) { 
+        /* Only memory bugs not exception */
+    }
+
+    return 0;
```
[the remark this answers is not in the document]
0x34d: Adding asserts can cause non-reproducible bugs and many false positives.

0x34d:
> In the design I am thinking of, the right parser fuzzer(s) will be selected and activated, one at a time, by a command line parameter (to squid executable) and/or a special HTTP/FTP request (accepted by a running Squid instance). Would that approach be compatible with the "feedback fuzzing used by OSS-Fuzz{LibFuzzer, AFL, and Honggfuzz}"?


Ah, won't work. `command line parameter` are not supported in `oss-fuzz`.
but we can compile each fuzzer one by one with macro.

0x34d:
@rousskov, I can't create a fuzzing harness for `FTP, URN, WCCPv2, and Gopher, etc.` as there are no good unit tests available for it.

`fuzzHttp1Parser` is based on `tests/testHttp1Parser`, and `fuzzRFC1035` is based on `tests/testDns`. Both of these fuzzers work exactly the same as their corresponding unit tests.

`tests/fuzzRFC1035.cc for tests/testRFC1035.cc` and `tests/fuzzHttp1Parser.cc for tests/testHttp1Parser.cc` in linker `$(FUZZ_DECODE_ENGINE) for $(LIBCPPUNIT_LIBS)` are super dumb, super easy.

You don't have to maintain anything crazy.

MegaManSec:
Harnesses for FTP, URN, and so on aren't that different. There are also a lot of HTTP parsers. Some of the changes for that based on my git log (which went under improvements later). All of these examples are missing other minor changes that are required (like configuration loading, in some cases, or hard-coding configurations to avoid problems):

```
+
+/**
+ * Fuzz how Squid reads Gopher server replies.
+ * Due to async calls, we largely just initalize everything before calling gopherToHTML() then clean up.
+ * Most code is taken from gopherReadReply().
+ */
+
+int main() {
+       FwdState *f;
+       GopherStateData *gopherState;
+       char *buf;
+       size_t len;
+
+       __AFL_INIT();
+       buf = (char *)__AFL_FUZZ_TESTCASE_BUF;
+       while(__AFL_LOOP(UINT_MAX)) {
+               len = __AFL_FUZZ_TESTCASE_LEN;
+               if(len > 9000) continue;
+               gopherState = new GopherStateData(f);
+
+               gopherState->HTML_header_added = 1;
+               gopherState->conversion = GopherStateData::HTML_CSO_RESULT; // GopherStateData::HTML_INDEX_RESULT; TODO: how to compare these two?
+               gopherState->cso_recno = 0;
+               gopherToHTML(gopherState, buf, len);
+
+               delete gopherState;
+       }
+       return 0;
+
+}
+
```

```
+/**
+ * Fuzz how Squid handles Urn REPLIES.
+ * Normally the entrypoint here is urnStart and unfortunately requires quite a bit of information.
+ * We alrgely emulate UrnState::start for this and just handle the 'reply' from the server.
+ * before calling urnHandleReply.
+ */
+#ifdef __AFL_HAVE_MANUAL_CONTROL
+int main() {
+       Http1::ResponseParser *hp;
+       StoreIOBuffer *tempBuffer;
+       AccessLogEntry *ale;
+       int len;
+       char *buf;
+       __AFL_INIT();
+       buf = (char *)__AFL_FUZZ_TESTCASE_BUF;
+
+       while(__AFL_LOOP(100000))
+       {
+               len = __AFL_FUZZ_TESTCASE_LEN;
+               if(len > (URN_REQBUF_SZ-1))
+                       continue;
+
+               SBuf inBuf((const char*) buf);
+               hp = new Http1::ResponseParser; //must be http compliant.
+               bool parsedOk = hp->parse(inBuf);
+               if(!parsedOk) {
+                       inBuf.clear();
+                       delete hp;
+                       continue;
+               }
+               inBuf.clear();
+
+               tempBuffer = new StoreIOBuffer(len, 0, (char *)buf);
+               ale = new AccessLogEntry();
+               CodeContext::Reset(ale);
+               const auto anUrn = new UrnState(ale);
+               visible_appname_string = "fuzzer";
+               strcpy(anUrn->reqbuf, (const char*)buf);
+               anUrn->entry = new StoreEntry();
+               anUrn->entry->lock("UrnState::start");
+               anUrn->reqofs = 0;
+               urnHandleReply(anUrn, *tempBuffer);
+               ale->unlock();
+               ale->unlock();
+               delete hp;
+               delete ale;
+               delete tempBuffer;
+               delete anUrn->entry;
+//             delete anUrn;
+       }
+exit(0);
+}
+#endif
```

```
+/**
+  * Fuzz a bunch of URL parsers. TODO: find other url parsers and see if they always are the same :)
+  *
+  */
+
+int uri_main() {
+//     Mem::Init();
+//     AnyP::UriScheme::Init();
+       static const CharacterSet uiChars = CharacterSet(UserInfoChars()).remove('%').rename("userinfo-reserved"); //Stolen from ::absolute() :)
+       Config.appendDomain = "google.com";
+       Config.appendDomainLen = strlen("google.com")+1;
+       Config.uri_whitespace = URI_WHITESPACE_CHOP;
+       char *buf;
+       size_t len;
+       __AFL_INIT();
+       buf = (char *)__AFL_FUZZ_TESTCASE_BUF;
+       while(__AFL_LOOP(10000)) {
+               len = __AFL_FUZZ_TESTCASE_LEN;
+               if(len >= 2048) continue;
+
+               auto *urlPointer = new AnyP::Uri;
+               auto *urlPointer2 = new AnyP::Uri;
+               SBuf sbuf(buf);
+               urlPointer->parse(Http::METHOD_CONNECT, sbuf);
+               urlPointer2->parse(Http::METHOD_GET, sbuf);
+
+               SBuf s = urlPointer->Encode(sbuf, uiChars);
+//             char *uccwr = urlCanonicalCleanWithoutRequest(sbuf,Http::METHOD_CONNECT , AnyP::PROTO_HTTP); //consider Http::METHOD_CONNECT/Http:METHOD_GET
+//             char *uccwr2 = urlCanonicalCleanWithoutRequest(sbuf,Http::METHOD_GET , AnyP::PROTO_HTTP);
+               urlIsRelative(buf);
+               urlPointer->addRelativePath(buf);
+               char *cleanurl = urlPointer->cleanup(buf);
+               Config.uri_whitespace = URI_WHITESPACE_STRIP;
+
+               urlPointer->clear();
+               urlPointer2->clear();
+      urlPointer->parse(Http::METHOD_CONNECT, sbuf);
+      urlPointer2->parse(Http::METHOD_GET, sbuf);
+
+
+               char *cleanurl2 = urlPointer->cleanup(buf);
+               matchDomainName(cleanurl2, cleanurl, mdnHonorWildcards);
+               matchDomainName(cleanurl, cleanurl2, mdnHonorWildcards);
+               matchDomainName(cleanurl, cleanurl2, mdnRejectSubsubDomains);
+               matchDomainName(cleanurl2, cleanurl, mdnRejectSubsubDomains);
+
+               if(cleanurl2)
+                       xfree(cleanurl2);
+               if(cleanurl)
+                       xfree(cleanurl);
+               sbuf.clear();
+               s.clear();
+               urlPointer->clear();
+               urlPointer2->clear();
+               delete urlPointer2;
+               delete urlPointer;
+       }
+return 0;
+}
+
```

```
+/**
+  * Fuzz CLIENT HTTP Headera (Client -> Server)
+  */
+__AFL_FUZZ_INIT();
+int main() {
+               Mem::Init();
+               httpHeaderInitModule();
+               parseConfigFile(DEFAULT_CONFIG_FILE);
+               Debug::parseOptions("ALL,3");
+                __AFL_INIT();
+               unsigned char *b = __AFL_FUZZ_TESTCASE_BUF;
+               while(__AFL_LOOP(500)) {
+                       size_t len = __AFL_FUZZ_TESTCASE_LEN;
+                       if(len>=65534) continue;
+                       char buf[65535];
+                       memset(buf,'\0', 65535);
+                       memcpy(buf, b, len);
+                       size_t hdr_sz;
+
+                       HttpHeader *hd = new HttpHeader(hoRequest);
+                       Http::ContentLengthInterpreter clen;
+
+                       hd->parse(buf, len, true, hdr_sz, clen);
+                       SBuf basic = hd->getAuthToken(Http::HdrType::AUTHORIZATION, "Basic");
+                       basic.clear();
+                       hd->removeHopByHopEntries();
+
+                       hd->clean();
+                       delete hd;
+               }
+       return 0;
+}
+
```

```
+/*int main() {
+       Mem::Init();
+       httpHeaderInitModule();
+       Debug::log_stderr = 1;
+//     Debug::parseOptions("ALL,3");
+       unsigned char *b = __AFL_FUZZ_TESTCASE_BUF;
+//     __AFL_INIT();
+       while(__AFL_LOOP(1))
+       {
+               char *buf = (char *)b;
+               size_t len = __AFL_FUZZ_TESTCASE_LEN;
+               if(len > 65534) continue;
+
+               String str(buf);
+               HttpHdrCc *cc=new HttpHdrCc();
+               cc->parse(str);
+               str.clean();
+               delete cc;
+       }
+       return 0;
+}*/

```

```
commit 660f221619c7688d16f6b55c633bb05b974a724c
Author: Your Name <you@example.com>
Date:   Fri Feb 26 00:37:33 2021 +0000

    Fuzz Squid's MD5 implementation.. TODO: Make this a bit better with incremental adding.. See https://guidovranken.com/2019/05/14/differential-fuzzing-of-cryptographic-libraries/

diff --git a/src/store_key_md5.cc b/src/store_key_md5.cc
index df86b7ebb..ddd790c36 100644
--- a/src/store_key_md5.cc
+++ b/src/store_key_md5.cc
@@ -12,6 +12,7 @@
 #include "HttpRequest.h"
 #include "md5.h"
 #include "store_key_md5.h"
+#include <openssl/md5.h>
 
 static cache_key null_key[SQUID_MD5_DIGEST_LENGTH];
 
@@ -93,6 +94,28 @@ storeKeyPrivate()
     return reinterpret_cast<cache_key*>(&key);
 }
 
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
+
+       static cache_key digest[SQUID_MD5_DIGEST_LENGTH];
+       static cache_key digest2[SQUID_MD5_DIGEST_LENGTH];
+       SquidMD5_CTX M;
+       SquidMD5Init(&M);
+//     for(int i=0; i < Size; i++) {
+               SquidMD5Update(&M, Data, 1);
+//     }
+       SquidMD5Final(digest, &M);
+
+       MD5(Data,Size,digest2);
+
+       if(memcmp(digest,digest2,SQUID_MD5_DIGEST_LENGTH) != 0) {
+               fprintf(stderr, "oops...\n");
+               fflush(stderr);
+               abort();
+       }
+       exit(0);
+}
+
+
 const cache_key *
 storeKeyPublic(const char *url, const HttpRequestMethod& method, const KeyScope keyScope)
 {
```

```
commit 87bb20d90716d2da64e1e4a666123350c02c640c
Author: Your Name <you@example.com>
Date:   Thu Feb 25 15:04:36 2021 +0000

    Header Tools

diff --git a/src/HttpHeaderTools.cc b/src/HttpHeaderTools.cc
index 1ea6bc4ff..3d89348cd 100644
--- a/src/HttpHeaderTools.cc
+++ b/src/HttpHeaderTools.cc
@@ -162,7 +162,28 @@ httpHeaderParseOffset(const char *start, int64_t *value, char **endPtr)
  * something went wrong, returns non-zero on success.
  * Un-escapes quoted-pair characters found within the string.
  * start should point at the first double-quote.
- */
+ /*/
+/*__AFL_FUZZ_INIT();
+int main() {
+   Mem::Init();
+   httpHeaderInitModule();
+       parseConfigFile(DEFAULT_CONFIG_FILE);
+   unsigned char *b = __AFL_FUZZ_TESTCASE_BUF;
+   while(__AFL_LOOP(100))
+   {
+      char *buf = (char *)b;
+      size_t len = __AFL_FUZZ_TESTCASE_LEN;
+
+      String str;
+               httpHeaderParseQuotedString(buf, len, &str);
+               str.clean();
+
+
+               SBuf key = httpHeaderQuoteString(buf);
+               key.clear();
+   }
+   return 0;
+}*/
 int
 httpHeaderParseQuotedString(const char *start, const int len, String *val)
 {
```

```
+/**
+  * Some really bad technique to fuzz HTCP packets.
+  * In the future, I should probably somehow pack the packets 'manually' such as is done in htcpTstReply().
+  * Until then, we're a dumb fuzzer!
+  */
+/*__AFL_FUZZ_INIT();
+
+int main() {
+       Mem::Init();
+       Acl::Init();
+       parseConfigFile(DEFAULT_CONFIG_FILE);
+       unsigned char *b = __AFL_FUZZ_TESTCASE_BUF;
+       while(__AFL_LOOP(1000)) {
+               size_t len = __AFL_FUZZ_TESTCASE_LEN;
+               if(len >= 8192) continue;
+               static Ip::Address from;
+               htcpHandleMsg(( char *)b, len, from);
+       }
+       return 0;
+}*/
```

And so on.

Future discussion I'll move to a different PR to avoid spamming here.

rousskov:
> > In the design I am thinking of, the right parser fuzzer(s) will be selected and activated, one at a time, by a command line parameter (to squid executable) and/or a special HTTP/FTP request (accepted by a running Squid instance). Would that approach be compatible with the "feedback fuzzing used by OSS-Fuzz{LibFuzzer, AFL, and Honggfuzz}"?

> Ah, won't work. `command line parameter` are not supported in `oss-fuzz`.

Noted. Does oss-fuzz support some way to select a fuzzer (i.e. the function being tested) when using libfuzzer as a library?

> but we can compile each fuzzer one by one with macro.

Yes, we can, but I hope there is a better interface available for oss-fuzz integration. Perhaps one binary (the one that oss-fuzz sees) can exec another (with the right parameter)?


>>> Unfortunately, that can't be done.

> > What makes you think that?

> it can be one

I assume you meant "it can be done". If so, this is a complete reversal of the previous assessment. Please be more careful when rejecting ideas because I am relying on your oss-fuzz expertise to do so correctly: I do not know enough about oss-fuzz and cannot double check everything.

> > LibFuzzer [documentation](https://llvm.org/docs/LibFuzzer.html#using-libfuzzer-as-a-library) ... explicitly covers the design where main() is provided by the program being tested itself.

> but this is bloat and slows down the fuzzer speed.

As for that unspecified "bloat", I see a lot more of (often very expensive as far as development/maintenance time is concerned!) Makefile/dependencies bloat in this PR than in using-libfuzzer-as-a-library approach, even if we have to use macros for every fuzzer, but perhaps you were talking about some other kind of "bloat" that I do not know about?

How/why does using libfuzzer as a library "slow down the fuzzer speed"? How significant is that slowdown in Squid context?

rousskov:
> I can't create a fuzzing harness for `FTP, URN, WCCPv2, and Gopher, etc.` as there are no good unit tests available for it.

Understood. Please do not attempt to create more fuzzing harnesses (for any protocols/parsers) until we settle on the overall approach to fuzzing in Squid. This PR already has enough fuzzer variety to decide on that overall approach IMHO.

> You don't have to maintain anything crazy.

This is where you are crossing into Squid-specific area where I can rely on my expertise to reject inaccurate assessments: The proposed fuzzHttp1Parser_SOURCES approach can be considered "crazy" expensive in terms of maintenance and future code refactoring costs. I bet you quickly copied existing tests_testHttp1Parser_SOURCES to bootstrap fuzzHttp1Parser_SOURCES, but that does not mean you are not duplicating a big problem. And this particular parser is not the worst example of what is yet to come if we continue in this direction -- Squid developers have wasted a lot of time _and_ rejected great Squid improvement opportunities because of those costs.

0x34d:
> This is where you are crossing into Squid-specific area where I can rely on my expertise to reject inaccurate assessments: The proposed fuzzHttp1Parser_SOURCES approach can be considered "crazy" expensive in terms of maintenance and future code refactoring costs. I bet you quickly copied existing tests_testHttp1Parser_SOURCES to bootstrap fuzzHttp1Parser_SOURCES, but that does not mean you are not duplicating a big problem. And this particular parser is not the worst example of what is yet to come if we continue in this direction -- Squid developers have wasted a lot of time and rejected great Squid improvement opportunities because of those costs.

I know, right know I'm fixing a build failure in the project so I can relate.

rousskov:
> I know, right know I'm fixing a build failure in the project so I can relate.

Hopefully, that negative experience will increase your cooperation in finding a better oss-fuzz integration approach (than duplication of existing problems with Squid unit testing approach). ;-)

src/Makefile.am
yadij: What benefits exactly/specifically do you see us getting from fuzzing the `Http1::Parser` class?
 This is a class where the edge cases are small for each syntax token and well-known.

IME fuzzing is most useful for smoke testing larger integrated systems - such as caching, HTTP relay. The level where unit tests become overly large and burdensome to write a unique test for each and every edge case. A fuzzer test alongside Daft tests, CoAdvisor and performance measurements would be good to add IMO.

src/tests/fuzzHttp1Parser.cc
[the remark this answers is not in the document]
0x34d: The goal is to find memory corruption, not exception.

PR #1534, opened 2023-10-23
Maintenance: consolidate testBoilerpate header file

kinkie:


src/tests/testBoilerplate.cc
```diff
@@ -7,11 +7,32 @@
...
 #include "unitTestMain.h"
 
 #include <stdexcept>
 
+/*
+ * demonstration test file, as new idioms are made they will
+ * be shown in the TestBoilerplate source.
+ *
+ * If a class declaration is only used in more than one source file,
+ * then move the declaration to a separate header file
```
rousskov, line 20: "only used in more than one" phrasing sounds odd to me.

If you insist on having this text:

```suggestion
 *
 * If the class declaration is used in more than one source file,
 * then place that declaration into a dedicated header file instead.
```

But my recommendation is to remove it. There is no (reasonable) way to reuse a class declaration across source files without placing it into a header file. We do not need to document this:

```suggestion
```

src/tests/testBoilerplate.cc
```diff
@@ -7,11 +7,32 @@
...
+ */
+
+class TestBoilerplate: public CPPUNIT_NS::TestFixture
+{
+    CPPUNIT_TEST_SUITE( TestBoilerplate );
+    /* note the statement here and then the actual prototype below */
+    CPPUNIT_TEST( testDemonstration );
+    CPPUNIT_TEST_SUITE_END();
+
+public:
```
rousskov, line 30:
```suggestion
```

rousskov (commented):
Thank you for this improvement. I do not insist on any changes.

rousskov (approved):

yadij (approved):

PR #1535, opened 2023-10-23
Fix build on MacOS

kinkie:
To use std::nullptr_t it is necessary to include <cstddef>
on MacOS Ventura 13.4.1 
with clang version 14.0.3 (Xcode 14.3.1)


rousskov (approved):
Thank you. I changed title to be more specific. If you can add more details (e.g., compiler name/version) please do so. There are many recent examples in `git log`.

yadij (approved):

PR #1536, opened 2023-10-23
Improve handling of expanding HTTP header values

rousskov:
Squid manipulations often increase HTTP header value length compared to
the corresponding raw value received by Squid. Raw header length is
checked against request_header_max_size and reply_header_max_size that
default to 64KB, making the raw value safe to store in a String object
(by default). However, when the increased length of a manipulated value
exceeds String class limits, Squid leaks memory, asserts, or possibly
stalls affected transactions. The long-term fix for this problem is a
complete String elimination from Squid sources, but that takes time.

Known manipulations may effectively concatenate headers and/or increase
header value length by 50%. This workaround makes such known increases
safe by essentially tripling String class limits:

    (64KB + 64KB) * 150% = 3 * 64KB

https://megamansec.github.io/Squid-Security-Audit/response-memleaks.html
where it was filed as "Memory Leak in HTTP Response Parsing".

### DIFF
--- a/src/SquidString.h
+++ b/src/SquidString.h
@@ -140,7 +140,16 @@ private:
 
     size_type len_ = 0;  /* current length  */
 
-    static const size_type SizeMax_ = 65535; ///< 64K limit protects some fixed-size buffers
+    /// An earlier 64KB limit was meant to protect some fixed-size buffers, but
+    /// (a) we do not know where those buffers are (or whether they still exist)
+    /// (b) too many String users unknowingly exceeded that limit and asserted.
+    /// We are now using a larger limit to reduce the number of (b) cases,
+    /// especially cases where "compact" lists of items grow 50% in size when we
+    /// convert them to canonical form. The new limit is selected to withstand
+    /// concatenation and ~50% expansion of two HTTP headers limited by default
+    /// request_header_max_size and reply_header_max_size settings.
+    static const size_type SizeMax_ = 3*64*1024 - 1;
+
     /// returns true after increasing the first argument by extra if the sum does not exceed SizeMax_
     static bool SafeAdd(size_type &base, size_type extra) { if (extra <= SizeMax_ && base <= SizeMax_ - extra) { base += extra; return true; } return false; }
 
--- a/src/http.cc
+++ b/src/http.cc
@@ -1895,8 +1895,9 @@ HttpStateData::httpBuildRequestHeader(HttpRequest * request,
 
         String strFwd = hdr_in->getList(Http::HdrType::X_FORWARDED_FOR);
 
-        // if we cannot double strFwd size, then it grew past 50% of the limit
-        if (!strFwd.canGrowBy(strFwd.size())) {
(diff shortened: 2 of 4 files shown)


src/SquidString.h
```diff
@@ -140,7 +140,16 @@ class String
...
-    static const size_type SizeMax_ = 65535; ///< 64K limit protects some fixed-size buffers
+    /// An earlier 64KB limit was meant to protect some fixed-size buffers, but
+    /// (a) we do not know where those buffers are (or whether they still exist)
+    /// (b) too many String users unknowingly exceeded that limit and asserted.
+    /// We are now using a larger limit to reduce the number of (b) cases,
+    /// especially cases where "compact" lists of items grow 50% in size when we
+    /// convert them to canonical form. The new limit is selected to withstand
+    /// concatenation and ~50% expansion of two HTTP headers limited by default
+    /// request_header_max_size and reply_header_max_size settings.
+    static const size_type SizeMax_ = 3*64*1024 - 1;
```
rousskov, line 151: This new comment states that we do not know where those "some fixed-size buffers" are or whether they exist. I could not find such buffers. To qualify, a buffer would have to meet these two criteria:

* Be problem-free when SizeMax_ remains at the official 64KB setting
* Become problematic when SizeMax_ increases

If you know where such buffers are in the current official Squid code, please share that information. In that case, this PR may need more work to accommodate those buffers!

src/http.cc
```diff
@@ -1895,8 +1895,9 @@ HttpStateData::httpBuildRequestHeader(HttpRequest * request,
 
         String strFwd = hdr_in->getList(Http::HdrType::X_FORWARDED_FOR);
 
-        // if we cannot double strFwd size, then it grew past 50% of the limit
-        if (!strFwd.canGrowBy(strFwd.size())) {
+        // Detect unreasonably long header values. And paranoidly check String
+        // limits: a String ought to accommodate two reasonable-length values.
+        if (strFwd.size() > 32*1024 || !strFwd.canGrowBy(strFwd.size())) {
```
rousskov, line 1900: This change preserves existing limits for forwarding loop detection. IMO, there is no reason to increase _those_ limits (i.e. there is no reason to allow even longer forwarding loops just because we are not going to crash if we allow them).

kinkie:
The approach looks like a sensible patch; has it been tested in any way?

rousskov:
> The approach looks like a sensible patch; has it been tested in any way?

Yes, of course. All PRs are supposed to be tested unless noted otherwise. Factory PRs, including this one, follow that rule.

Needless to say, it is not feasible to test all code paths affected by this change in the maximum String buffer size. I did try to find all limit users, and analyzed the ones I found.

We plan to add the corresponding CI regression tests, but fixing other Zero Day vulnerabilities with published reproducers takes priority.

yadij (approved):

kinkie (approved):

squidadm:
queued for backport to v6

PR #1538, opened 2023-10-24
RFC 1123: Fix date parsing

rousskov:
The bug was discovered and detailed by Joshua Rogers at
https://megamansec.github.io/Squid-Security-Audit/datetime-overflow.html
where it was filed as "1-Byte Buffer OverRead in RFC 1123 date/time
Handling".

### DIFF
--- a/src/time/rfc1123.cc
+++ b/src/time/rfc1123.cc
@@ -44,7 +44,13 @@ make_month(const char *s)
     char month[3];
 
     month[0] = xtoupper(*s);
+    if (!month[0])
+        return -1; // protects *(s + 1) below
+
     month[1] = xtolower(*(s + 1));
+    if (!month[1])
+        return -1; // protects *(s + 2) below
+
     month[2] = xtolower(*(s + 2));
 
     for (i = 0; i < 12; i++)


kinkie (approved):

src/time/rfc1123.cc
```diff
@@ -44,7 +44,13 @@ make_month(const char *s)
...
 
     month[0] = xtoupper(*s);
+    if (!month[0])
+        return -1; // protects *(s + 1) below
+
     month[1] = xtolower(*(s + 1));
+    if (!month[1])
+        return -1; // protects *(s + 2) below
+
     month[2] = xtolower(*(s + 2));
```
rousskov, line 54: This is not how we should write code these days, but improving this code would require rewriting the whole function (and more), which is not a good idea for this minimal fix. AFAICT from looking at optimized assembly code, this fix has reasonable performance impact; I cannot make it faster without expanding the scope of these changes. The entire make_month() function (and several of its siblings!) should be replaced with safer, optimized versions anyway.

src/time/rfc1123.cc
```diff
@@ -44,7 +44,13 @@ make_month(const char *s)
     char month[3];
 
     month[0] = xtoupper(*s);
+    if (!month[0])
+        return -1; // protects *(s + 1) below
```
rousskov, line 48: The new safety checks do not alter Squid behavior: Any (unlikely) overread side effects aside, we continue to accept valid 3-letter months and reject everything else AFAICT.

src/time/rfc1123.cc
kinkie: I agree with all you write. Maybe I can look into these
after html_quote is done

yadij: Agreed. Fix first polish later.

yadij (approved):

squidadm:
queued for backport to v6

PR #1543, opened 2023-10-25
Exit without asserting when helper process startup fails

rousskov:
... to dup() after fork() and before execvp().

Assertions are for handling program logic errors. Helper initialization
code already handled system call errors correctly (i.e. by exiting the
newly created helper process with an error), except for a couple of
assert()s that could be triggered by dup(2) failures.

https://megamansec.github.io/Squid-Security-Audit/ipc-assert.html
where it was filed as 'Assertion in Squid "Helper" Process Creator'.

### DIFF
--- a/src/ipc.cc
+++ b/src/ipc.cc
@@ -22,6 +22,11 @@
 
 #include <chrono>
 #include <thread>
+#include <cstdlib>
+
+#if HAVE_UNISTD_H
+#include <unistd.h>
+#endif
 
 static const char *hello_string = "hi there\n";
 #ifndef HELLO_BUF_SZ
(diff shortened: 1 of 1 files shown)


kinkie (approved):
The forecast vulnerability seems very theoretical and low impact to me, but the code still deserves a fix
Thanks for doing it!

yadij (approved):

squidadm:
queued for backport to v6

yadij:
Withdrawing v5 backport due to lambda support issues.

MegaManSec:
FYI, I am more than happy to reject GHSA-xggx-9329-3c27 as a security issue, and considering the resource exhaustion that would be required to perform that would in of itself be a vulnerability, it's difficult to see how classing GHSA-xggx-9329-3c27 as a security issue is correct.

rousskov:
> the resource exhaustion ... would in of itself be a vulnerability

Since descriptors are a finite resource, I doubt that exhausting them while correctly serving traffic can be considered a Squid security vulnerability. If there is nothing Squid can do to prevent a given problem, then that problem is not a (meaningful) Squid security vulnerability, right? Certain scalability limits could be a security vulnerability for a given deployment, of course, but that problem is not about Squid itself being vulnerable AFAICT.

MegaManSec:
Agreed.

PR #1545, opened 2023-10-26
Add Tokenizer::skipOne() to fetch the char skipped

yadij:


src/parser/Tokenizer.cc
```diff
@@ -168,6 +168,17 @@ Parser::Tokenizer::skipOne(const CharacterSet &chars)
     return false;
 }
 
+char
+Parser::Tokenizer::skipOne()
+{
+    if (buf_.isEmpty())
+        throw InsufficientInput();
```
rousskov, line 175: This is not the method we discussed in https://github.com/squid-cache/squid/pull/1540#discussion_r1372563952. This method throws on EOF. The method we discussed returns std::nullopt in that case (otherwise, we would not be able to use it in an `if` statement as sketched).

If you want to keep this skipOne() method, please give it a `description` parameter -- all Tokenizer methods that throw have a description parameter. That is how we distinguish them from the non-throwing methods...

To implement the method we discussed, add another skipOne(void) method that returns std::nullopt on EOF. The throwing skipOne(description) should call skipOne(void).

I recommend doing both, but I cannot insist on it.

src/parser/Tokenizer.cc
```diff
@@ -168,6 +168,17 @@ Parser::Tokenizer::skipOne(const CharacterSet &chars)
...
+
+    debugs(24, 8, "skipping one of any character");
+    const auto result = consume(1);
+    return result[0];
```
rousskov, line 179: I doubt the compiler will be able to optimize the creation of a temporary SBuf away in this case. I would help the compiler by doing something like this instead:

```suggestion
    const auto result = buf_[0];
    consume(1);
    debugs(24, 8, "character code: " << int(result));
    return result;
```

N.B. debugs() will print "skipOne" so we do not need to say "skipping one" or some such.

src/parser/Tokenizer.h
```diff
@@ -109,6 +109,12 @@ class Tokenizer
      */
     bool skipOne(const CharacterSet &discardables);
 
+    /** Skips a single character from the buffer.
+     *
+     * \return the character which was skipped
```
rousskov, line 114:
```suggestion
     * \return the skipped character
```

rousskov (changes requested):
I welcome this Tokenizer enhancement, but it needs a bit of work.

PR #1546, opened 2023-10-26
Docs: update cfgman tarball automation

yadij:
Fix script issues:

    NOTICE: unknown line 'IF ...'
    NOTICE: unknown line 'ENDIF'
    NOTICE: unknown line 'POSTSCRIPTUM: ...'

Add display of which cf.data.pre line each such NOTICE
(and some debugs) was found on.

Add missing HAVE_AUTH_* definitions.

Fix case typo of if/endif so regex can detect it.

Also, fix indentation typo in 'acl' directive texts.

### DIFF
--- a/scripts/www/build-cfg-help.pl
+++ b/scripts/www/build-cfg-help.pl
@@ -352,8 +352,7 @@ while (<>) {
         print "NOTICE: line $.: unknown line '$_'\n";
     }
 }
-my $condition;
-foreach $condition (@ifelse) {
+foreach my $condition (@ifelse) {
     print "ERROR: missing ENDIF for '$condition'\n";
 }
 end_options;
--- a/scripts/www/build-cfg-help.pl
+++ b/scripts/www/build-cfg-help.pl
@@ -247,9 +247,12 @@ sub update_defaults()
     }
 }
 
+my $linecount = 0;
+my @ifelse = {};
 while (<>) {
     chomp;
     last if (/^EOF$/);
+    $linecount++;
     if ($_ =~ /^NAME: (.*)$/) {
         my (@aliases) = split(/ /, $1);
         $data = {};
@@ -267,8 +270,28 @@ while (<>) {
         $data->{'default_doc'} = "";
         $data->{'default_if_none'} = "";
 
-        print "DEBUG: new option: $name\n" if $verbose;
+        print "DEBUG: line $linecount: new option: $name\n" if $verbose;
         next;
+    } elsif ($_ =~ /^IF (.*)$/) {
+        my $cond = $1;
+        push(@ifelse, $1);
+        if (! exists $defines{$1}) {
+            print "NOTICE: line $linecount: unknown ./configure option '$1'\n";
+        } else {
+            $cond = $defines{$1};
+        }
+        if ($state eq "doc") {
+            $data->{"doc"} .= "if " . $cond . "\n";
+        } elsif ($state eq "comment") {
+            $comment .= "if " . $cond . "\n";
+        }
+    } elsif ($_ =~ /^ENDIF$/) {
+        pop(@ifelse);
+        if ($state eq "doc") {
+            $data->{"doc"} .= "endif\n";
+        } elsif ($state eq "comment") {
+            $comment .= "endif\n";
+        }
     } elsif ($_ =~ /^COMMENT: (.*)$/) {
         $data->{"comment"} = $1;
     } elsif ($_ =~ /^TYPE: (.*)$/) {
@@ -280,6 +303,11 @@ while (<>) {
         } else {
             $data->{"default"} .= "$name $1\n";
         }
+    } elsif ($_ =~ /^POSTSCRIPTUM: (.*)$/) {
+        if ($data->{"default"} eq "none") {
+            $data->{"default"} = "";
+        }
+        $data->{"default"} .= "$name $1\n";
     } elsif ($_ =~ /^DEFAULT_DOC: (.*)$/) {
         $data->{"default_doc"} .= "$1\n";
     } elsif ($_ =~ /^DEFAULT_IF_NONE: (.*)$/) {
@@ -323,7 +351,7 @@ while (<>) {
     } elsif (/^#/) {
         next;
     } elsif ($_ ne "") {
-        print "NOTICE: unknown line '$_'\n";
+        print "NOTICE: line $linecount: unknown line '$_'\n";
     }
 }
 end_options;
--- a/scripts/www/build-cfg-help.pl
+++ b/scripts/www/build-cfg-help.pl
@@ -248,7 +248,7 @@ sub update_defaults()
 }
 
 my $linecount = 0;
-my @ifelse = {};
+my @ifelse = ();
 while (<>) {
     chomp;
     last if (/^EOF$/);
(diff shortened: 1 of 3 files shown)


scripts/www/build-cfg-help.pl
```diff
@@ -247,9 +247,12 @@ ()
     }
 }
 
+my $linecount = 0;
```
rousskov, line 250: If you have not already, please consider using `$.` (or equivalent) instead of maintaining your own line count. See "man perlvar" for more details, including `@ARGV` handing that is happening in the loop below:

```
$. is reset when the filehandle is closed, but not when an open filehandle is reopened without an intervening "close()".  For
more details, see "I/O Operators" in perlop.  Because "<>" never does an explicit close, line numbers increase across "ARGV"
files (but see examples in "eof" in perlfunc).
```

scripts/www/build-cfg-help.pl
```diff
@@ -247,9 +247,12 @@ ()
...
+my @ifelse = {};
 while (<>) {
     chomp;
     last if (/^EOF$/);
+    $linecount++;
```
rousskov, line 255: If you insist on maintaining your own line count, please increment this line counter after reading a line, before any code that might exit the loop.

This is one of the two problems insist on addressing.

scripts/www/build-cfg-help.pl
```diff
@@ -267,8 +270,28 @@ ()
         $data->{'default_doc'} = "";
         $data->{'default_if_none'} = "";
 
-        print "DEBUG: new option: $name\n" if $verbose;
+        print "DEBUG: line $linecount: new option: $name\n" if $verbose;
         next;
+    } elsif ($_ =~ /^IF (.*)$/) {
+        my $cond = $1;
+        push(@ifelse, $1);
+        if (! exists $defines{$1}) {
```
rousskov, line 278: It is very easy to accidentally create a hash entry in Perl. Where all hash entries must be defined, I recommend using `defined` instead of `exists` for testing:

```suggestion
        if (! defined $defines{$1}) {
```

scripts/www/build-cfg-help.pl
```diff
@@ -247,9 +247,12 @@ ()
     }
 }
 
+my $linecount = 0;
+my @ifelse = {};
```
rousskov, line 251: This is an array, not a hash:

```suggestion
my @ifelse = ();
```

This is one of the two problems insist on addressing.

scripts/www/build-cfg-help.pl
```diff
@@ -323,7 +351,7 @@ ()
     } elsif (/^#/) {
         next;
     } elsif ($_ ne "") {
-        print "NOTICE: unknown line '$_'\n";
+        print "NOTICE: line $linecount: unknown line '$_'\n";
     }
 }
 end_options;
```
rousskov, line 357: Please report unclosed `IF`s here. We now have `@ifelse` we can trivially check...

scripts/www/build-cfg-help.pl
```diff
@@ -267,8 +270,28 @@ ()
         $data->{'default_doc'} = "";
         $data->{'default_if_none'} = "";
 
-        print "DEBUG: new option: $name\n" if $verbose;
+        print "DEBUG: line $linecount: new option: $name\n" if $verbose;
         next;
+    } elsif ($_ =~ /^IF (.*)$/) {
+        my $cond = $1;
+        push(@ifelse, $1);
```
rousskov, line 277: I would push the current line number so that warnings about unclosed `IF`s (see another change request) can tell the user where the broken `IF` started.

```suggestion
        push(@ifelse, "$.: $1");
```

src/cf_gen_defines
```diff
@@ -24,6 +24,8 @@ BEGIN {
 	define["FOLLOW_X_FORWARDED_FOR&&LINUX_NETFILTER"]="--enable-follow-x-forwarded-for and --enable-linux-netfilter"
 	define["FOLLOW_X_FORWARDED_FOR&&USE_ADAPTATION"]="--enable-follow-x-forwarded-for and (--enable-icap-client and/or --enable-ecap)"
 	define["FOLLOW_X_FORWARDED_FOR&&USE_DELAY_POOLS"]="--enable-follow-x-forwarded-for and --enable-delay-pools"
+	define["HAVE_AUTH_MODULE_BASIC"]="--enable-auth-basic"
+	define["HAVE_AUTH_MODULE_DIGEST"]="--enable-auth-digest"
```
rousskov, line 28: What effect, if any, does this correction have on various squid.conf files generated during Squid build?

rousskov (changes requested):
Thank you for this improvement. I do not insist on any changes except the two explicitly marked requests.

src/cf_gen_defines
yadij: Both this and the `IF`/`ENDIF` tracking change the squid.conf.documented and HTML output to use the `./configure` option(s) listed in this array instead of the code internal `USE_*`/`HAVE_*` macro names.

For example, text at http://www.squid-cache.org/Doc/config/auth_param/
```
IF HAVE_AUTH_MODULE_BASIC
	=== Basic authentication parameters ===

...
		processing and similar.

ENDIF
IF HAVE_AUTH_MODULE_DIGEST
	=== Digest authentication parameters ===
...

ENDIF
```
becomes
```
if --enable-auth-basic
	=== Basic authentication parameters ===

...
		processing and similar.

endif
if --enable-auth-digest
	=== Digest authentication parameters ===
...

endif
```

scripts/www/build-cfg-help.pl
yadij: Done in [ae14820](https://github.com/squid-cache/squid/pull/1546/commits/ae148203f6b1b2afb2497bc48191d21fbdad8be5)

scripts/www/build-cfg-help.pl
yadij: Done in[b01c6f5](https://github.com/squid-cache/squid/pull/1546/commits/b01c6f574b2fdc5ff5c722d88a506e2047bc703c)

src/cf_gen_defines
rousskov: Great, thank you.

scripts/www/build-cfg-help.pl
```diff
@@ -351,9 +349,13 @@ ()
     } elsif (/^#/) {
         next;
     } elsif ($_ ne "") {
-        print "NOTICE: line $linecount: unknown line '$_'\n";
+        print "NOTICE: line $.: unknown line '$_'\n";
     }
 }
+my $condition;
+foreach $condition (@ifelse) {
```
rousskov, line 356:
```suggestion
foreach my $condition (@ifelse) {
```

scripts/www/build-cfg-help.pl
```diff
@@ -351,9 +349,13 @@ ()
...
+    print "ERROR: missing ENDIF for '$condition'\n";
```
rousskov, line 357: Since we now include the line number in `@ifelse`:

```suggestion
    print "ERROR: missing ENDIF to match $condition\n";
```

It would be even better to split the popped item (which is not a "condition" anymore) and say something like "to match ... on line ...".

rousskov (approved):
Thank you for addressing my concerns. I do not insist on any further changes.

rousskov (approved):

PR #1547, opened 2023-10-26
Remove disabled classful networks code

kinkie:
Removed ifdef-d code in networkFromInaddr() and mark that function as
static. The code was disabled since 1997 commit 429fdbe.


rousskov (approved):
Thank you. I adjusted PR title/description, to emphasize that the removed code was disabled and inform that it was disabled since 1997.

yadij (approved):
Wow. That should have been gone a decade ago. Thanks.

PR #1551, opened 2023-10-27
Bug 5312: Startup aborts if OPEN_MAX exceeds RLIMIT_NOFILE

alanc:
FATAL: Event loop exited with failure

The DP_POLL ioctl on Solaris fails with EINVAL when the number of
supplied descriptors (OPEN_MAX) is higher than the current RLIMIT_NOFILE
setting. When it comes to the maximum number of descriptors, we should
use Squid_MaxFD (which already reflects RLIMIT_NOFILE setting).

### DIFF
--- a/src/comm/ModDevPoll.cc
+++ b/src/comm/ModDevPoll.cc
@@ -17,10 +17,7 @@
  *
  * Ported by Peter Payne from Squid 2.7.STABLE9 comm_devpoll.c
  * on August 11, 2010 at 3pm (GMT+0100 Europe/London).
- *
- * Last modified 2010-10-08
  */
-
 /*
  * There are several poll types in Squid, ALL of which are compiled and linked
  * in. Thus conditional compile-time flags are used to prevent the different
@@ -47,10 +44,6 @@
 #endif
 
 #define DEBUG_DEVPOLL 0
-
-// OPEN_MAX is defined in <climits>
-#define DEVPOLL_UPDATESIZE  OPEN_MAX
-#define DEVPOLL_QUERYSIZE   OPEN_MAX
 
 /* TYPEDEFS */
 typedef short pollfd_events_t; /* type of pollfd.events from sys/poll.h */
@@ -128,7 +121,7 @@ comm_flush_updates(void)
  * @param events events to register (usually POLLIN, POLLOUT, or POLLREMOVE)
  */
 static void
-comm_update_fd(int fd, int events)
+comm_update_fd(int fd, pollfd_events_t events)
 {
     debugs(
         5,
@@ -178,20 +171,23 @@ Comm::SelectLoopInit(void)
     /* allocate memory first before attempting to open poll device */
     /* This tracks the FD devpoll offset+state */
     devpoll_state = (struct _devpoll_state *)xcalloc(
-                        SQUID_MAXFD, sizeof(struct _devpoll_state)
+                        Squid_MaxFD, sizeof(struct _devpoll_state)
                     );
 
-    /* And this is the stuff we use to read events */
+    /* This is the stuff we use to read events.  If it's larger than
+       the current RLIMIT_NOFILE, the Solaris kernel returns EINVAL. */
+    dpoll_nfds = Squid_MaxFD;
     do_poll.dp_fds = (struct pollfd *)xcalloc(
-                         DEVPOLL_QUERYSIZE, sizeof(struct pollfd)
+                         dpoll_nfds, sizeof(struct pollfd)
                      );
-    dpoll_nfds = DEVPOLL_QUERYSIZE;
 
-    devpoll_update.pfds = (struct pollfd *)xcalloc(
-                              DEVPOLL_UPDATESIZE, sizeof(struct pollfd)
-                          );
+    /* This is the stuff we use to write requests to change tracking state.
+       It's also limited to the current RLIMIT_NOFILE by the Solaris kernel. */
     devpoll_update.cur = -1;
-    devpoll_update.size = DEVPOLL_UPDATESIZE;
+    devpoll_update.size = Squid_MaxFD;
+    devpoll_update.pfds = (struct pollfd *)xcalloc(
+                              devpoll_update.size, sizeof(struct pollfd)
+                          );
 
     /* attempt to open /dev/poll device */
     devpoll_fd = open("/dev/poll", O_RDWR);


alanc:
More details are explained in https://bugs.squid-cache.org/show_bug.cgi?id=5312

yadij:
OK to test

yadij (dismissed):

alanc:
The force push just fixed the MAX_OPEN -> OPEN_MAX typo in the commit title.

rousskov:
> The force push just fixed the MAX_OPEN -> OPEN_MAX typo in the commit title.

@alanc, just FYI: We do not use PR branch commit titles and messages when merging changes into master. Anubis, the bot automating such merging, uses GitHub PR title and PR description for the squashed master [commit](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message). We need to add a greeting message to explain that and other caveats to new contributors...

src/comm/ModDevPoll.cc
```diff
@@ -48,9 +48,8 @@
 
 #define DEBUG_DEVPOLL 0
 
-// OPEN_MAX is defined in <climits>
-#define DEVPOLL_UPDATESIZE  OPEN_MAX
-#define DEVPOLL_QUERYSIZE   OPEN_MAX
+#define DEVPOLL_UPDATESIZE  Squid_MaxFD
+#define DEVPOLL_QUERYSIZE   Squid_MaxFD
```
rousskov, line 52: We should get rid of these non-standard (and potentially clashing) macros completely and use Squid_MaxFD in ModDevPoll code instead.

```suggestion
```

To avoid code duplication, the lines using these values should be reordered so that each code snippet is using Squid_MaxFD exactly once. For example:

```C++
    dpoll_nfds = Squid_MaxFD;
    do_poll.dp_fds = (struct pollfd *)xcalloc(
                         dpoll_nfds, sizeof(struct pollfd)
                     );
```

and

```C++
    devpoll_update.size = Squid_MaxFD;
    devpoll_update.pfds = (struct pollfd *)xcalloc(
                              devpoll_update.size, sizeof(struct pollfd)
                          );
```

If both snippets should use the same number of descriptors, then the second snippet should become

```C++
    devpoll_update.size = dpoll_nfds;
    devpoll_update.pfds = (struct pollfd *)xcalloc(
                              devpoll_update.size, sizeof(struct pollfd)
                          );
```

src/comm/ModDevPoll.cc
```diff
@@ -48,9 +48,8 @@
 
 #define DEBUG_DEVPOLL 0
 
-// OPEN_MAX is defined in <climits>
-#define DEVPOLL_UPDATESIZE  OPEN_MAX
-#define DEVPOLL_QUERYSIZE   OPEN_MAX
```
rousskov, line 53: What about SQUID_MAXFD at the very beginning of Comm::SelectLoopInit()? Should that become Squid_MaxFD as well? If not, it would be best to add a C++ comment there explaining why the three contexts (i.e. devpoll_state, do_poll, and devpoll_update) should use two different counters (i.e. Squid_MaxFD and SQUID_MAXFD).

rousskov (dismissed):
Thank you for this bug fix!

I cannot object to this fix because it makes some things better without making other things worse, but I think we should finish this update by addressing several closely related problems, as detailed in my change requests. I can implement those change requests if you prefer -- just let me know! Both look trivial to me, but I may be missing some hidden caveats.

src/comm/ModDevPoll.cc
alanc: I'm new to squid code, so I don't quite understand the relationship between SQUID_MAXFD and Squid_MaxFD - I see that SQUID_MAXFD is set at compile time, and Squid_MaxFD is set at runtime, but I don't see what SQUID_MAXFD is used for outside of making sure that if USE_SELECT is defined, that FD_SETSIZE is at least SQUID_MAXFD.

src/comm/ModDevPoll.cc
alanc: While the original code used the same value for both, it is not required to do so.  Both must be no larger than the current RLIMIT_NOFILES, since you can't request to poll on more file descriptors than you can possibly open, nor receive more poll responses than the number of file descriptors that are currently open.

DEVPOLL_UPDATESIZE is the number of fd's that can fit in the buffer used to batch up requests to add/remove fd's from the list /dev/poll is monitoring.  Having it be Squid_MAXFD allows requesting to add or remove every fd at once, and is likely overkill - perhaps this should be a fixed value such as 128 or 256 (the old OPEN_MAX value).  The cost of having it lower is simply making more write() system calls if a list of fd's larger than the buffer size is being presented all at once.

DEVPOLL_QUERYSIZE is the maximum number of fd's that can be returned in a single call to the DP_POLL ioctl() [the equivalent of a poll() or select() system call in this backend]. Having it be Squid_MAXFD allows being notified of a result for every fd at once, and may also be overkill, but I don't know what value to suggest here.

src/comm/ModDevPoll.cc
```diff
@@ -17,10 +17,8 @@
  *
  * Ported by Peter Payne from Squid 2.7.STABLE9 comm_devpoll.c
  * on August 11, 2010 at 3pm (GMT+0100 Europe/London).
- *
- * Last modified 2010-10-08
  */
-
+re
```
yadij, line 21: this typo is completely breaking the build.
```suggestion

```

yadij (changes requested):

src/comm/ModDevPoll.cc
rousskov: I have not checked all the details, but, IIRC, the following is true about Squid_MaxFD:

* Squid_MaxFD is the maximum number of open descriptors a running Squid process supports;
* Squid_MaxFD is the size of fde::Table (a.k.a. fd_table; see fde::Init());
* all open descriptors are smaller than Squid_MaxFD.

AFAICT, that is all we need to know to answer the question in this change request.

rousskov (changes requested):
FWIW, I am quite surprised by the recent PR updates. I expected this PR to do what its current description says -- use Squid_MaxFD for everything.

I hope this PR does not need to make all these complex decisions to fully address Bug 5312 (i.e. it is sufficient to just make all the three arrays Squid_MaxFD in size), but I may be missing something.

I left a few specific questions to facilitate progress. At the end of the day, if there are good reasons to make those complex decisions in this PR, we will need to document, via C++ comments, the rationale behind those new constants (i.e., why "256" and why "1024").

src/comm/ModDevPoll.cc
rousskov: This change request is outdated. I split/moved the discussion to https://github.com/squid-cache/squid/pull/1551#discussion_r1375553125 and https://github.com/squid-cache/squid/pull/1551#discussion_r1375554045. I am resolving this change request.

src/comm/ModDevPoll.cc
```diff
@@ -48,19 +45,15 @@
 
 #define DEBUG_DEVPOLL 0
 
-// OPEN_MAX is defined in <climits>
-#define DEVPOLL_UPDATESIZE  OPEN_MAX
-#define DEVPOLL_QUERYSIZE   OPEN_MAX
+// Maximum number of fd's to buffer requests to change polling state
+#define DEVPOLL_UPDATESIZE  256
```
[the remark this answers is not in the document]
alanc: My thought was simply that why should we allocate more memory than we might need, but as these two allocations each take only 64-bytes per fd, the default 4095 fd's will only add up to 256kb for each, and the apparently maximum of 32768 would only be 2mb for each, both of which were a ton of memory when I started programming, but are likely not so noticeable on a modern server running squid.  I can simplify down to Squid_MAXFD for both.

src/comm/ModDevPoll.cc
```diff
@@ -48,19 +45,15 @@
...
+// Maximum number of fd's to receive polling updates for in a single syscall
+#define DEVPOLL_QUERYSIZE   1024
```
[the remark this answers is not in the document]
alanc: I'm not an expert either, but I don't know of any reasons beyond attempting to optimize for memory usage over number of system calls/loop iterations made.  I'll defer to your judgement here.

src/comm/ModDevPoll.cc
```diff
@@ -232,11 +226,11 @@ Comm::SetSelect(int fd, unsigned int type, PF * handler, void *client_data, time
     if (!F->flags.open) {
         /* remove from poll set */
         comm_update_fd( fd, POLLREMOVE );
-        devpoll_state[fd].state = 0;
+        F->epoll_state = 0;
         return;
     }
 
-    pollfd_events_t state_old = devpoll_state[fd].state;
+    pollfd_events_t state_old = (pollfd_events_t) F->epoll_state;
```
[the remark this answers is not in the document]
alanc: In this PR?  It just avoided having to decide what the right size would be.

In another PR it helps solve https://bugs.squid-cache.org/show_bug.cgi?id=5314 in which some code paths are trying to access `devpol_state` when `Comm::SelectLoopInit()` had not yet been called, but the `fde` struct was properly initialized.  I think the epoll code would similarly crash if you just changed it to use a similar allocation strategy.

I can switch back to leaving the old allocation for now and just adjusting it to Squid_MaxFD for this PR.

src/comm/ModDevPoll.cc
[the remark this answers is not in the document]
rousskov: > why should we allocate more memory than we might need

... because Squid might need all of it (and Squid will crash if we do not allocate enough memory for these arrays).

It would be possible to refactor Squid code so that it wastes less memory without crashing. However, that refactoring requires a lot of work, saving a few megabytes of RAM is not a sufficient incentive to perform that work IMO, and this PR is not about that kind of optimization.

src/comm/ModDevPoll.cc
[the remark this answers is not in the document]
rousskov: Thank you. Since this is not a performance optimization PR, we do not need to make complex performance optimization decisions. Let's keep it simple for now.

src/comm/ModDevPoll.cc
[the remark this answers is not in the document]
rousskov: > In this PR? It just avoided having to decide what the right size would be.

Understood. Since removing that array (in this PR) creates more problems than it solves, I think you made the right call to restore it.

> In another PR...

Let's deal with that other problem in another PR. On the surface of it, I doubt that removing this array will solve that initialization problem there, but we do not need to discuss it _here_.

rousskov (approved):
Thank you for addressing my concerns. I have adjusted PR title/description to supply (assumed) logged symptoms and fix [formatting](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message).

I am approving this PR and requesting @yadij re-review (I believe his earlier concern has been addressed as well).

yadij (approved):
FTR; this will need a re-work when the `max_filedescriptors` directive becomes reconfigurable. However since it is currently limited to a full restart on changes this PR change is sufficient.

PR #1553, opened 2023-10-29
Fix infinite recursion when parsing HTTP chunks

rousskov:
This change stops infinite HttpStateData recursion with at-max-capacity
inBuf. Such inBuf prevents progress in the following call chain:

* processReply()
* processReplyBody() and decodeAndWriteReplyBody()
* maybeReadVirginBody()
* maybeMakeSpaceAvailable() -- tries but fails to quit processing
* processReply()

HttpStateData::maybeMakeSpaceAvailable() no longer calls processReply(),
preventing recursion.

maybeReadVirginBody() now aborts transactions that would otherwise get
stalled due to full read buffer at its maximum capacity. This change
requires that all maybeReadVirginBody() callers do actually need more
response data to make progress. AFAICT, that (natural) invariant holds.

We moved transaction stalling check from maybeMakeSpaceAvailable() into
its previous callers. Without that move, maybeMakeSpaceAvailable() would
have to handle both abortTransaction() and delayRead() cases. Besides
increased code complexity, that would trigger some premature delayRead()
calls (at maybeReadVirginBody() time). Deciding whether to delay socket
reads is complicated, the delay mechanism is expensive, and delaying may
become unnecessary by the time the socket becomes readable, so it is
best to continue to only delayRead() at readReply() time, when there is
no other choice left.

maybeReadVirginBody() mishandled cases where progress was possible, but
not _immediately_ -- it did nothing in those cases, probably stalling
transactions when maybeMakeSpaceAvailable() returned false but did not
call processReply(). This is now fixed: maybeReadVirginBody() now starts
waiting for the socket to be ready for reading in those cases,
effectively passing control to readReply() that handles them.

maybeReadVirginBody() prematurely grew buffer for future socket reads.
As a (positive) side effect of the above refactoring, we now delay
buffer growth until the actual read(2) time, which is best for
performance. Most likely, this premature buffer growth was an accident:
maybeReadVirginBody() correctly called maybeMakeSpaceAvailable() with
doGrow set to false. However, maybeMakeSpaceAvailable() misinterpreted
doGrow as a "do not actually do it" parameter. That bug is now gone.

This recursion bug was discovered and detailed by Joshua Rogers at
https://megamansec.github.io/Squid-Security-Audit/
where it was filed as "Chunked Encoding Stack Overflow".

### DIFF
--- a/src/http.cc
+++ b/src/http.cc
@@ -54,6 +54,7 @@
 #include "RefreshPattern.h"
 #include "rfc1738.h"
 #include "SquidConfig.h"
+#include "SquidMath.h"
 #include "StatCounters.h"
 #include "Store.h"
 #include "StrList.h"
@@ -1195,16 +1196,24 @@ HttpStateData::readReply(const CommIoCbParams &io)
      * Plus, it breaks our lame *HalfClosed() detection
      */
 
-    Must(maybeMakeSpaceAvailable(true));
-    CommIoCbParams rd(this); // will be expanded with ReadNow results
-    rd.conn = io.conn;
-    rd.size = entry->bytesWanted(Range<size_t>(0, inBuf.spaceSize()));
+    const auto moreDataPermission = canBufferMoreReplyBytes();
+    if (!moreDataPermission) {
+        abortTransaction("ready to read required data, but the read buffer is full and cannot be drained");
+        return;
+    }
 
-    if (rd.size <= 0) {
+    const auto readSizeMax = maybeMakeSpaceAvailable(moreDataPermission.value());
+    // TODO: Move this logic inside maybeMakeSpaceAvailable():
+    const auto readSizeWanted = readSizeMax ? entry->bytesWanted(Range<size_t>(0, readSizeMax)) : 0;
+
+    if (readSizeWanted <= 0) {
         delayRead();
         return;
     }
 
+    CommIoCbParams rd(this); // will be expanded with ReadNow results
+    rd.conn = io.conn;
+    rd.size = readSizeWanted;
     switch (Comm::ReadNow(rd, inBuf)) {
     case Comm::INPROGRESS:
         if (inBuf.isEmpty())
@@ -1513,8 +1522,9 @@ HttpStateData::processReplyBody()
             }
 
             flags.do_next_read = true;
+            maybeReadVirginBody();
+            return;
         }
-        break;
 
         case COMPLETE_PERSISTENT_MSG: {
             debugs(11, 5, "processReplyBody: COMPLETE_PERSISTENT_MSG from " << serverConnection);
@@ -1567,7 +1577,8 @@ HttpStateData::processReplyBody()
             return;
         }
 
-    maybeReadVirginBody();
+    // TODO: Remove "else" after ENTRY_ABORTED handling above to simplify this logic.
+    assert(!"unreachable code because all ENTRY_ABORTED and persistentConnStatus() outcomes are handled above");
 }
 
 bool
@@ -1586,8 +1597,10 @@ HttpStateData::maybeReadVirginBody()
     if (!Comm::IsConnOpen(serverConnection) || fd_table[serverConnection->fd].closing())
         return;
 
-    if (!maybeMakeSpaceAvailable(false))
+    if (!canBufferMoreReplyBytes()) {
+        abortTransaction("more response bytes required, but the read buffer is full and cannot be drained");
         return;
+    }
 
     // XXX: get rid of the do_next_read flag
     // check for the proper reasons preventing read(2)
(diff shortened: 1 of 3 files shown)


src/http.cc
```diff
@@ -1605,40 +1616,78 @@ HttpStateData::maybeReadVirginBody()
...
+    // Config.readAheadGap like this creates huge read buffers for large
+    // read_ahead_gap values. TODO: Switch to using tcp_recv_bufsize as the
+    // primary read buffer capacity factor.
+    //
+    // TODO: Cannot reuse throwing NaturalCast() here. Consider removing
+    // .value() dereference in NaturalCast() or add/use NaturalCastOrMax().
+    const auto configurationPreferences = NaturalSum<size_t>(Config.readAheadGap).value_or(SBuf::maxSize);
+
+    // TODO: Honor TeChunkedParser look-ahead and trailer parsing requirements
+    // (when explicit configurationPreferences are set too low).
```
rousskov, line 1640: These TODOs and XXXs as well as many other related problems were being addressed by #736. Some day, I will update and resume pushing that old PR forward, but I am trying to minimize the changes in this PR. If others object to any of the above TODOs and XXXs, I will remove them.

src/http.cc
```diff
@@ -1605,40 +1616,78 @@ HttpStateData::maybeReadVirginBody()
...
 {
-    // how much we are allowed to buffer
-    const int limitBuffer = (flags.headers_parsed ? Config.readAheadGap : Config.maxReplyHeaderSize);
-
-    if (limitBuffer < 0 || inBuf.length() >= (SBuf::size_type)limitBuffer) {
-        // when buffer is at or over limit already
-        debugs(11, 7, "will not read up to " << limitBuffer << ". buffer has (" << inBuf.length() << "/" << inBuf.spaceSize() << ") from " << serverConnection);
-        debugs(11, DBG_DATA, "buffer has {" << inBuf << "}");
-        // Process next response from buffer
-        processReply();
```
rousskov, line 1619: This misplaced clause caused the infinite recursion.

src/http.cc
```diff
@@ -1605,40 +1616,78 @@ HttpStateData::maybeReadVirginBody()
...
+        return 0;
     }
 
-    // just report whether we could grow or not, do not actually do it
-    if (doGrow)
-        return (read_size >= 2);
-
     // we may need to grow the buffer
     inBuf.reserveSpace(read_size);
-    debugs(11, 8, (!flags.do_next_read ? "will not" : "may") <<
```
rousskov, line 1637: I have removed "will not" clause from this debugs() line because we only "make space available" when we are on a "true do_next_read" path. We may not actually read for various reasons, but not because flags.do_next_read is false.

rousskov (commented):
These changes fix the problem in my tests. I am posting this PR as a Draft because I would like to run more tests and double check that there are no places where we should add the removed processReply() call. However, I think it is fairly safe to start reviewing this rushed fix when you get a chance, at least to make sure you are OK with various TODOs and XXXs being left as they are now.

rousskov:
> Alex: I am posting this PR as a Draft because I would like to ... double check that there are no places where we should add the removed processReply() call.

Done, with the following conclusion: I do not think more processReply() calls are needed in this PR. We already call processReply() after we received more data from a peer and after we are done handling a 1xx control message. The only other condition/event I can think of that would require us to parse already buffered from-peer bytes is when the adaptation drains a previously overflowed BodyPipe buffer. In that Client::noteMoreBodySpaceAvailable() code, we call maybeReadVirginBody() instead of processReply() but that should work OK for now AFAICT, and, more importantly improving that fragile code (with some rather heavy implicit assumptions!) is outside _this_ PR scope.

I am marking this PR as ready for review. We still plan on doing more testing in the background, and I do _not_ intend to merge this PR (even if it is approved) until that testing is completed.

src/http.cc
```diff
@@ -1195,16 +1196,24 @@ HttpStateData::readReply(const CommIoCbParams &io)
...
-    CommIoCbParams rd(this); // will be expanded with ReadNow results
-    rd.conn = io.conn;
-    rd.size = entry->bytesWanted(Range<size_t>(0, inBuf.spaceSize()));
+    const auto moreDataPermission = canBufferMoreReplyBytes();
+    if (!moreDataPermission) {
+        abortTransaction("ready to read required data, but the read buffer is full and cannot be drained");
+        return;
+    }
+
+    const auto readSizeMax = maybeMakeSpaceAvailable(moreDataPermission.value());
```
eduard-bagdasaryan, line 1205: BTW, is there a similar problem in Server code, i.e., do we need some 'canBufferMoreRequestBytes()' check before calling Server::maybeMakeSpaceAvailable()?

eduard-bagdasaryan:
>maybeReadVirginBody() correctly called maybeMakeSpaceAvailable() with doGrow set to false. [...] maybeMakeSpaceAvailable() misinterpreted doGrow as a "do not actually do it" parameter. 

This should also mean that the other maybeMakeSpaceAvailable(true) call in HttpStateData::readReply() worked incorrectly, i.e., it did not grow the buffer.

yadij (approved):

rousskov (commented):
> Alex: We still plan on doing more testing in the background, and I do not intend to merge this PR (even if it is approved) until that testing is completed.

Others have merged this PR. We are still running those tests. Eduard or I will follow up if those tests expose any significant problems associated with this PR changes.

> Eduard: This should also mean that the other maybeMakeSpaceAvailable(true) call in HttpStateData::readReply() worked incorrectly, i.e., it did not grow the buffer.

Yes, here is how #736 described that bug: 

> Also fixed commit https://github.com/squid-cache/squid/commit/1ad68518bc9cf11d68e486759be9d88adf7b16f8 to delay buffer allocations until we are ready
to read from the server. That old commit made buffer allocations in
maybeMakeSpaceAvailable() optional so that they can be delayed until we
actually read from the peer socket. Delaying buffer allocation until
read(2) is the right thing to do. The callers (correctly) passed a
boolean flag to trigger the allocation, but the method interpreted the
flag as if it meant "do not allocate". That is, commit tried to delay
allocations but actually allocated early.

There is more in that PR description, including the explanation why the code "worked" despite doing the opposite of what was intended/expected.

yadij:
@rousskov preparing the advisory for announcement. Was this patch sufficient to fix the issue or are/were there additional changes necessary?

rousskov:
> Was this patch sufficient to fix the issue or are/were there additional changes necessary?

AFAIK, commit 50c5af88bd74208103722f87593a974bb0f0b8e9 fixes infinite processReply-maybeMakeSpaceAvailable-processReply recursion when parsing HTTP chunks. Additional changes are necessary to fix other known problems in response reading/buffering code, along the lines of #736, but none of those "other known problems" are infinite chunks parsing recursion IIRC.

PR #1554, opened 2023-10-30
Move WHOIS code to libclients.la

yadij:
No logic changes


kinkie (approved):
LGTM. Thanks!

rousskov (approved):
IMO, this move is a bit premature because WhoisState is not a Client child but should become one (assuming the feature is not completely removed, of course).

yadij:
> IMO, this move is a bit premature because WhoisState is not a Client child but should become one (assuming the feature is not completely removed, of course).

That is in part why I am not including any code change in this PR. This one is about removing one more dependency from squid_SOURCES. The refactor is for later.

kinkie:
> That is in part why I am not including any code change in this PR. This one is about removing one more dependency from squid_SOURCES. The refactor is for later.

FWIW, I fully support this move as it serves to the strategic goal to make Squid more modular

PR #1555, opened 2023-10-30
Remove SquidString include from src/ssl/helper.cc

kinkie:


rousskov (approved):

rousskov (commented):
I adjusted PR title/description to make the former more precise (there is no SSL helper but there are two helpers defined in that file) and remove the latter.

yadij (approved):

PR #1556, opened 2023-10-30
Refactor Ssl::ErrorDetailsManager to use SBuf

kinkie:
Have Ssl::ErrorDetailsList use SBuf instead of String, and
ErrorDetailsManager::cache index on SBuf and not std::string

### DIFF
--- a/src/ssl/ErrorDetailManager.cc
+++ b/src/ssl/ErrorDetailManager.cc
@@ -98,7 +98,7 @@ Ssl::ErrorDetailsList::Pointer
 Ssl::ErrorDetailsManager::getCachedDetails(const char * const lang) const
 {
     Cache::iterator it;
-    it = cache.find(lang);
+    it = cache.find(SBuf(lang));
     if (it != cache.end()) {
         debugs(83, 8, "Found template details in cache for language: " << lang);
         return it->second;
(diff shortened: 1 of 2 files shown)


yadij (dismissed):

src/ssl/ErrorDetailManager.cc
```diff
@@ -98,7 +98,7 @@ Ssl::ErrorDetailsList::Pointer
 Ssl::ErrorDetailsManager::getCachedDetails(const char * const lang) const
 {
     Cache::iterator it;
-    it = cache.find(lang);
+    it = cache.find(SBuf(lang));
```
rousskov, line 101: I hope you will follow up with a PR that will change `lang` type to SBuf by upgrading strHdrAcptLangGetItem() to return optional SBuf (instead of filling a fixed-size buffer).

src/ssl/ErrorDetailManager.h
```diff
@@ -84,7 +82,7 @@ class ErrorDetailsManager
     /// cache the given error details list.
     void cacheDetails(const ErrorDetailsList::Pointer &errorDetails) const;
 
-    typedef std::map<std::string, ErrorDetailsList::Pointer> Cache;
+    typedef std::map<SBuf, ErrorDetailsList::Pointer> Cache;
```
rousskov, line 85: We try to upgrade typedefs when changing them:

```suggestion
    using Cache = std::map<SBuf, ErrorDetailsList::Pointer>;
```

rousskov (approved):
Thank you. I do not insist on any changes.

src/ssl/ErrorDetailManager.cc
kinkie: Eventually, yes. As a general approach, my strategy is:
1. remove SquidString, starting with low-hanging fruits
2. go chase char* and convert them as much as possible

kinkie:
Sadly changing the typedef invalidated approvals. @rousskov , @yadij , please re-approve at your convenience

rousskov (approved):

yadij (approved):

PR #1557, opened 2023-10-30
FTP: Ignore credentials with a NUL-prefixed username

sp-andwei:
FATAL: FTP login parsing destroyed username info

https://megamansec.github.io/Squid-Security-Audit/ftp-fatal.html
where it was filed as "FTP Authentication Crash".

### DIFF
--- a/src/clients/FtpGateway.cc
+++ b/src/clients/FtpGateway.cc
@@ -402,6 +402,12 @@ Ftp::Gateway::loginParser(const SBuf &login, bool escaped)
     if (login.isEmpty())
         return;
 
+    if (!login[0]) {
+        debugs(9, 2, "WARNING: Ignoring FTP credentials that start with a NUL character");
+        // TODO: Either support credentials with NUL characters (in any position) or ban all of them.
+        return;
+    }
+
     const SBuf::size_type colonPos = login.find(':');
 
     /* If there was a username part with at least one character use it.
@@ -1042,9 +1048,8 @@ Ftp::Gateway::checkAuth(const HttpHeader * req_hdr)
     /* Test URL login syntax. Overrides any headers received. */
     loginParser(request->url.userInfo(), true);
 
-    /* name is missing. that's fatal. */
-    if (!user[0])
-        fatal("FTP login parsing destroyed username info");
+    // XXX: We we keep default "anonymous" instead of properly supporting empty usernames.
+    Assure(user[0]);
 
     /* name + password == success */
     if (password[0])


yadij (approved):
LGTM

yadij:
OK to test

src/clients/FtpGateway.cc
```diff
@@ -1042,9 +1042,9 @@ Ftp::Gateway::checkAuth(const HttpHeader * req_hdr)
     /* Test URL login syntax. Overrides any headers received. */
     loginParser(request->url.userInfo(), true);
 
-    /* name is missing. that's fatal. */
-    if (!user[0])
-        fatal("FTP login parsing destroyed username info");
+    /* empty user is not an error; if someone is intent on passing that to
+     * us, go ahead and pass it to the server; this will almost certainly
+     * fail but at least fail gracefully */
```
rousskov, line 1047: This comment contradicts Ftp::Gateway::loginParser() logic (at least [^1]): There, the logic is documented as (I am paraphrasing for clarity): if the username is empty, retain what we have already. Since this code starts with a non-empty "anonymous" value, I believe this fatal() call should be unreachable in correctly written code. The fact that it is reachable, implies (to me), that there is a bug somewhere -- somewhere we are erasing the previously set username.

This is especially bad because this method logic is a sequence of username (and password) extraction steps, with the last successful step winning. The fact that this fatal() code is reachable means there is also a danger that we are erasing a non-empty non-default username somewhere.

Instead of forwarding what looks like garbage and contradicting ourselves, would it be possible to fix Ftp::Gateway::loginParser() so that it decides whether username is present (and updates `user` field) _after_ rfc1738_unescape()? AFAICT, that simple change would address the actual bug here. After that, this if+fatal() statement should be replaced with Assure().

[^1]: The new comment also contradicts the intent behind the buggy fatal() call if that intent was to assert that our code did not overwrite the default "anonymous" value.

rousskov (changes requested):

src/clients/FtpGateway.cc
andwei: Hmhm. I think we could keep it consistent with the rest of the code by simply adding a `|| strlen(login.c_str()) == 0` check to the `Ftp::Gateway::loginParser` to prevent empty usernames from overwriting anything that came before.

That being said, I think that this consistency is bought by continueing to not treat the special case of an empty username correctly (see @MegaManSec comment about empty usernames). Looking at the Uri class, I see that ftp://@<host> is eventually treated the same as ftp://<host> with regard to username -- "anonymous" is used, although the former example provides a username "".  `Ftp::Gateway::loginParser` checks the size of the `SBuf &login` and return directly if it's empty. That fails for the offending `AG00` base64 sequence because for that check SBuf is not treated as a c-string and leaves us with the `'\0` username that is not expected.

Thus, by following the intention of the current code and fixing the Parser in that way, we accept that an empty username equals "no username", which could be ok given that "empty username" probably does not have an overwhelming relevance in practice.

Please let me know what you think

PS: A fix to actually handle the empty username correctly seems a bit more of an effort since it would involve the Uri Parser, because the distinction between empty and none is already killed there, AFAICS.

rousskov, line 1047: >  I think we could keep it consistent with the rest of the code by simply adding a `|| strlen(login.c_str()) == 0` check to the `Ftp::Gateway::loginParser` to prevent empty usernames from overwriting anything that came before.

Please use a `!login[0]` test instead:

```C++
if (login.isEmpty())
    return;

if (!login[0]) {
    debugs(9, 2, "WARNING: Ignoring FTP credentials that start with a NUL character");
    // TODO: Either support credentials with NUL characters (in any position) or ban all of them.
    return;
}
```

I would also add `Assure(user[0])` after rfc1738_unescape() but that is optional. AFAICT, that terrible function cannot make `user` empty in this context.

Finally, I would replace the fatal() `if` statement with:

```suggestion
    // XXX: We we keep default "anonymous" instead of properly supporting empty usernames.
    Assure(user[0]);
```

rousskov: > I think that this consistency is bought by continueing to not treat the special case of an empty username correctly

I bet you are right.

> A fix to actually handle the empty username correctly seems a bit more of an effort

Agreed. That correct handling cannot be implemented well without introducing new types and changing quite a bit of code, including low-quality FTP code.

My recommendation for _this_ PR is to preserve consistent (but incorrect) treatment of empty user names (or what we decide, due to lack of appropriate support in code, is an equivalent of an empty user name). When/if somebody decides to improve Squid FTP compliance with regard to empty user names, they can post a PR dedicated to that improvement.

I am resolving this change request and opening another one to discuss your code change proposal.

P.S. Just to avoid a misunderstanding: My primary goal here is to make sure that this PR does not make things worse (e.g., by adding a comment that is not consistent with the rest of the code). I am not suggesting or insisting on any specific solution to address (or work around) the vulnerability.

andwei: I changed the commit according to your requests. Looks good to me (log is from a backport of the relevant change to v4.16) 
`printf "GET ftp://ftp.dlptest.com/ HTTP/1.1\r\nAuthorization: Basic AG00\r\n\r\n" | nc <squid_host> <squid_proxy_port>`
causes (excerpt)
```
2023/11/03 12:43:40.571 kid1| 9,2| src/clients/FtpClient.cc(810) writeCommand: ftp<< USER anonymous^M  
2023/11/03 12:43:40.642 kid1| 9,2| src/clients/FtpClient.cc(1127) parseControlReply: ftp>> 331 Please specify the password.   
2023/11/03 12:43:40.642 kid1| 9,2| src/clients/FtpClient.cc(810) writeCommand: ftp<< PASS anonymous@foo^M 
2023/11/03 12:43:43.571 kid1| 9,2| src/clients/FtpClient.cc(1127) parseControlReply: ftp>> 530 Login incorrect. 
...
2023/11/03 12:43:43.577 kid1| 11,2| src/http/Stream.cc(272) sendStartOfMessage: HTTP Client REPLY:                                                                                                                                       
---------                                                                                                                                                                                                                                                     
HTTP/1.1 401 Unauthorized^M 
...
X-Squid-Error: ERR_FTP_FORBIDDEN 530^M     
```

rousskov: > Hope I did not mess up the whole review process by force-pushing

You did not, but there is usually no reason to force-push. In the vast majority of cases, it only makes things more difficult for humans working on the PR. You also do not need to merge master into your PR branch (unless there are conflicts).

src/clients/FtpGateway.cc
```diff
@@ -402,6 +402,12 @@ Ftp::Gateway::loginParser(const SBuf &login, bool escaped)
     if (login.isEmpty())
         return;
 
+    if (!login[0]) {
+        debugs(9, 2, "WARNING: Ignoring FTP credentials that start with a NUL character");
```
rousskov, line 406: I am concerned that this debugging line was not present in the test output you have shared (quoted below). Are you sure the test is tickling the new code? Was the same test crashing Squid before?

> ```
> 2023/11/03 12:43:40.571 kid1| 9,2| src/clients/FtpClient.cc(810) writeCommand: ftp<< USER anonymous^M  
> 2023/11/03 12:43:40.642 kid1| 9,2| src/clients/FtpClient.cc(1127) parseControlReply: ftp>> 331 Please specify the password.   
> 2023/11/03 12:43:40.642 kid1| 9,2| src/clients/FtpClient.cc(810) writeCommand: ftp<< PASS anonymous@foo^M 
> 2023/11/03 12:43:43.571 kid1| 9,2| src/clients/FtpClient.cc(1127) parseControlReply: ftp>> 530 Login incorrect. 
> ...
> 2023/11/03 12:43:43.577 kid1| 11,2| src/http/Stream.cc(272) sendStartOfMessage: HTTP Client REPLY:                                                                                                                                       
> ---------                                                                                                                                                                                                                                                     
> HTTP/1.1 401 Unauthorized^M 
> ...
> X-Squid-Error: ERR_FTP_FORBIDDEN 530^M     
> ```

rousskov (approved):
Thanks a lot for adjusting _and_ testing!

rousskov:
I have adjusted PR title/description (i.e. the future commit [message](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message)) to match the current solution.

src/clients/FtpGateway.cc
andwei: Sorry for the late reply. It did tickle the code in question. It was admittedly rather stupid to not include that m(
```

yadij (approved):

squidadm:
queued for backport to v6

PR #1561, opened 2023-10-31
Bug 5309: frequent "lowestOffset () <= target_offset" assertion

rousskov:
Recent commit 122a6e3 left store_client::readOffset() unchanged but
should have adjusted it to match changed copyInto.offset semantics:
Starting with that commit, storeClientCopy() callers supply HTTP
response _body_ offset rather than HTTP response offset.

This bug decreased readOffset() values (by the size of stored HTTP
response headers), effectively telling Store that we are not yet done
with some of the MemObject/mem_hdr bytes. This bug could cause slightly
higher transaction memory usage because the same response bytes are
trimmed later. This bug should not have caused any assertions.

However, the old mem_hdr::freeDataUpto() code that uses readOffset() is
also broken -- the assertion in that method only "works" when
readOffset() returns values matching a memory node boundary. The smaller
values returned by buggy readOffset() triggered buggy assertions.

This minimal fix removes the recent store_client::readOffset() bug
described above. We will address old mem_hdr problems separately.

### DIFF
--- a/src/MemObject.cc
+++ b/src/MemObject.cc
@@ -167,7 +167,7 @@ struct LowestMemReader : public unary_function<store_client, void> {
 
     void operator() (store_client const &x) {
         if (x.getType() == STORE_MEM_CLIENT)
-            current = std::min(current, x.readOffset());
+            current = std::min(current, x.discardableHttpEnd());
     }
 
     int64_t current;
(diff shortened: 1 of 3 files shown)


yadij (approved):

rousskov:
> PR description: the old mem_hdr::freeDataUpto() code that uses readOffset() is also broken -- the assertion in that method only "works" when readOffset() returns values matching a memory node boundary. The smaller values returned by buggy readOffset() triggered buggy assertions. ... We will address old mem_hdr problems separately.

Done in #1562.

squidadm:
queued for backport to v6

PR #1563, opened 2023-11-01
v6 Next Backports

squidadm:


rousskov:
@kinkie, @yadij: Based on bug 5309 [feedback](https://bugs.squid-cache.org/show_bug.cgi?id=5309#c20) so far, we should either backport both #1561 and #1562 or none of them. As of now, this PR only includes the former.

yadij:
Scheduled #1562 as well. thanks

PR #1568, opened 2023-11-03
Fix ipv4 and expand ipv6 ACL parameter matching

rousskov:
The undocumented "ipv4", "ipv6", and "all" parameters can be used in
src, dst, and localip ACLs. This refactoring fixes "ipv4" matching and
expands "ipv6" scope as detailed further below. These parameters remain
undocumented.

Together, the fixed ipv4 parameter and the adjusted ipv6 parameter now
match any IP address known to Squid. So does the undocumented "all"
parameter (which is currently used to implement the "all" dst ACL).

Also speed up matching of ACLs containing "ipv4", "ipv6", and "all"
parameters (at the cost of adding two boolean tests to other src, dst,
and localip ACLs). This performance improvement is a side effect of the
corresponding code improvements and is not the focus of these changes.

Also improved mgr:config reporting of ACLs containing these special
parameters. They are now reported using their names (e.g., "all") rather
than IP addresses (e.g., `::/0` instead of "all").

### Fix ipv4 ACL parameter to match (any IPv4 address)

The "ipv4" parameter is supposed to match any IPv4 address, but it
matched none because an unmasked[^1] IPv4 input address was compared to
an all-0s address[^2], while any IPv4 address in Squid has some 1s due
to IPv4-to-IPv6 mapping used by Ip::Address. Those 1s could not match.

The "ipv4" parameter did match one address -- an IPv6 address `::`[^2].

This matching was possibly broken since inception (2009 commit 7764e92).

[^1]: Calling mask.setNoAddr() sets all sin6_addr bits to 1 while
calling mask.applyMask(0, AF_INET) does nothing at all.

[^2]: We did not set q->addr1. Ip::Address default constructor sets all
Ip::Address bits to 0. For ACL IP comparison purposes, that is
equivalent to an IPv6 `::` address (all sin6_addr bits are zero).

### Expand ipv6 ACL parameter scope to match all IPv6 addresses

The "ipv6" parameter was not matching many IPv6 addresses, including
some frequently seen ones like `::1` (i.e. IPv6 localhost). Now it
matches any IP address that Squid code itself considers an IPv6 address,
reducing surprises among admins that need to treat all IPv6 addresses
specially.

### DIFF
--- a/src/acl/Ip.cc
+++ b/src/acl/Ip.cc
@@ -221,104 +221,6 @@ acl_ip_data::FactoryParse(const char *t)
 
     debugs(28, 5, "aclIpParseIpData: " << t);
 
-    /* Special ACL RHS "all" matches entire Internet */
-    if (strcmp(t, "all") == 0) {
-        debugs(28, 9, "aclIpParseIpData: magic 'all' found.");
-        q->addr1.setAnyAddr();
-        q->addr2.setEmpty();
-        q->mask.setAnyAddr();
-        return q;
-    }
-
-    /* Detect some old broken strings equivalent to 'all'.
-     * treat them nicely. But be loud until its fixed.  */
-    if (strcmp(t, "0/0") == 0 || strcmp(t, "0.0.0.0/0") == 0 || strcmp(t, "0.0.0.0/0.0.0.0") == 0 ||
-            strcmp(t, "0.0.0.0-255.255.255.255") == 0 || strcmp(t, "0.0.0.0-0.0.0.0/0") == 0) {
-
-        debugs(28,DBG_CRITICAL, "ERROR: '" << t << "' needs to be replaced by the term 'all'.");
-        debugs(28,DBG_CRITICAL, "SECURITY NOTICE: Overriding config setting. Using 'all' instead.");
-        q->addr1.setAnyAddr();
-        q->addr2.setEmpty();
-        q->mask.setAnyAddr();
-        return q;
-    }
-
-    /* Special ACL RHS "ipv4" matches IPv4 Internet
-     * A nod to IANA; we include the entire class space in case
-     * they manage to find a way to recover and use it */
-    if (strcmp(t, "ipv4") == 0) {
-        q->addr1.setAnyAddr();
-        q->addr1.setIPv4();
-        q->mask = q->addr1;
-        return q;
-    }
-
-    /* Special ACL RHS "ipv6" matches IPv6-Unicast Internet */
-    if (strcmp(t, "ipv6") == 0) {
-        debugs(28, 9, "aclIpParseIpData: magic 'ipv6' found.");
-        r = q; // save head of the list for result.
-
-        /* 0000::/4 is a mix of localhost and obsolete IPv4-mapping space. Not valid outside this host. */
-
-        /* Future global unicast space: 1000::/4 */
-        q->addr1 = "1000::";
-        q->mask.setNoAddr();
-        q->mask.applyMask(4, AF_INET6);
-
-        /* Current global unicast space: 2000::/4 = (2000::/4 - 3000::/4) */
-        q->next = new acl_ip_data;
-        q = q->next;
-        q->addr1 = "2000::";
-        q->mask.setNoAddr();
-        q->mask.applyMask(3, AF_INET6);
-
-        /* Future global unicast space: 4000::/2 = (4000::/4 - 7000::/4) */
-        q->next = new acl_ip_data;
-        q = q->next;
-        q->addr1 = "4000::";
-        q->mask.setNoAddr();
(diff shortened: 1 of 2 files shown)


src/acl/Ip.cc
```diff
@@ -471,13 +374,64 @@ acl_ip_data::FactoryParse(const char *t)
...
+    /* Detect some old broken strings equivalent to 'all'.
+     * treat them nicely. But be loud until its fixed.  */
+    if (strcmp(token, "0/0") == 0 ||
+            strcmp(token, "0.0.0.0/0") == 0 ||
+            strcmp(token, "0.0.0.0/0.0.0.0") == 0 ||
+            strcmp(token, "0.0.0.0-255.255.255.255") == 0 ||
+            strcmp(token, "0.0.0.0-0.0.0.0/0") == 0) {
+
+        debugs(28,DBG_CRITICAL, "ERROR: '" << token << "' needs to be replaced by the term 'all'.");
+        debugs(28,DBG_CRITICAL, "SECURITY NOTICE: Overriding config setting. Using 'all' instead.");
```
rousskov, line 416: I left this [2009](https://github.com/squid-cache/squid/commit/7764e92dbe388e6152b559910d4769d8c737b19f) override unchanged. It should probably be removed (with /0 addresses converted into a fatal configuration error?), but that felt like an out of scope change.

src/acl/Ip.cc
```diff
@@ -221,103 +221,6 @@ acl_ip_data::FactoryParse(const char *t)
...
-        q->mask.setAnyAddr();
-        return q;
-    }
-
-    /* Special ACL RHS "ipv4" matches IPv4 Internet
-     * A nod to IANA; we include the entire class space in case
-     * they manage to find a way to recover and use it */
-    if (strcmp(t, "ipv4") == 0) {
-        q->mask.setNoAddr();
-        q->mask.applyMask(0, AF_INET);
```
rousskov, line 251: This is the problematic `ipv4` code that resulted in no IPv4 matches (and a single IPv6 match!). Please see PR description for details.

src/acl/Ip.cc
```diff
@@ -221,103 +221,6 @@ acl_ip_data::FactoryParse(const char *t)
...
-        return q;
-    }
-
-    /* Special ACL RHS "ipv6" matches IPv6-Unicast Internet */
-    if (strcmp(t, "ipv6") == 0) {
-        debugs(28, 9, "aclIpParseIpData: magic 'ipv6' found.");
-        r = q; // save head of the list for result.
-
-        /* 0000::/4 is a mix of localhost and obsolete IPv4-mapping space. Not valid outside this host. */
```
rousskov, line 260: These are now included. I see absolutely no reason to exclude them from `ipv6`, even if they are "not valid outside this host". IMO, `ipv6` is not a "Does the attacker use this IP address correctly?" ACL value. It is just a "will Squid treat this IP address as an IPv6 address?" ACL value. Squid may see these addresses (e.g., in request URLs) and admins have legitimate use cases for matching them when they want to match "all IPv6 addresses" (e.g., to work around certain recently discovered IPv6 handling bugs and their still-under-investigation side effects)...

The same reasoning applies to a few other IPv6 ranges that were excluded in the rules below.

src/acl/Ip.cc
kinkie: Are we sure it's a good idea to replace the config instead of aborting and letting the admin update their settings?
The replaced config also matches all ipv6, which may or may not be what the admin intends

kinkie (commented):
I like the simplification and spirit, with one concern

src/acl/Ip.cc
rousskov: > Are we sure it's a good idea to replace the config instead of aborting and letting the admin update their settings?

FWIW, IMO, replacing the config in this particular case (including the fact that this is master/v7 code) is bad. That is why I said that this official 2009 workaround code should be removed or replaced (in a dedicated PR).

> The replaced config also matches all ipv6, which may or may not be what the admin intends

Yes, that is one of this official code problems.

Just to avoid misunderstanding: You are asking a question about official code that this PR does not change in any material way (only moves and synchronizes it with the `all` mechanism changes done in this PR). As I said, I believe this code should be removed or replaced with a different (and probably fatal) error, but all of those changes are outside _this_ PR scope.

If you prefer, I can return this code to its original place and do not update it to use the PR-adjusted `all` mechanism, but that feels wrong to me because this old code explicitly says `using 'all'` _and_ has the effect of using `all`.

kinkie: If the behaviour is what is currently shipped, then there is no reason to change here.

I would support a different PR changing the behaviour from warn-and-change to warn-and-exit

kinkie (approved):

yadij (approved):

PR #1573, opened 2023-11-05
Use PackableStream for dump_acl

kinkie:


src/cache_cf.cc
```diff
@@ -1490,16 +1491,12 @@ free_SBufList(SBufList *list)
...
-                          name,
-                          ae->name,
-                          ae->typeString());
-        SBufList tail;
-        tail.splice(tail.end(), ae->dumpOptions());
-        tail.splice(tail.end(), ae->dump()); // ACL parameters
-        dump_SBufList(entry, tail);
+        os << name << " " << ae->name << " " << ae->typeString() << " "
+           << AsList(ae->dumpOptions()).delimitedBy(" ") << " "
+           << AsList(ae->dump()).delimitedBy(" ") << "\n";
```
rousskov: The above printed an extra space for ACLs without options and and extra space for ACLs without data parameters. I fixed that in PR branch commit 3df4b49. That commit also got rid of ACL::dumpOptions() that was using std::stream to build an SBufList that AsList() in this PR code was then dumping back to std::stream.

src/acl/Acl.cc
```diff
@@ -322,25 +323,14 @@ ACL::parseFlags()
...
-    // this check also works around dump_SBufList() adding ' ' after empty items
-    if (!myOptions.empty()) {
-        SBufStream stream;
-        stream << myOptions;
-        const SBuf optionsImage = stream.buf();
-        if (!optionsImage.isEmpty())
-            result.push_back(optionsImage);
-    }
-    return result;
+    os << directiveName << ' ' << name << ' ' << typeString() << options() <<
```
rousskov, line 331: No space before options() because individual Acl::Option printers add a leading space. Technically, that space addition is wrong/misplaced, but fixing that is out of scope _and_ will require quite a few code changes.

src/acl/Acl.h
```diff
@@ -82,7 +82,11 @@ class ACL
 
     virtual void prepareForUse() {}
 
-    SBufList dumpOptions(); ///< \returns approximate options configuration
+    // TODO: Find a way to make options() and this method constant
+    /// Prints aggregated "acl" (or similar) directive configuration, including
+    /// the given directive name, ACL name, ACL type, and ACL parameters. The
+    /// printed parameters are collected from all same-name "acl" directives.
+    void dumpWhole(const char *directiveName, std::ostream &);
```
rousskov, line 89: We should not name this method dump() because there is a virtual dump() method already. In a sense, maybe that inconsistency is a good thing because this method has to print (an equivalent of) more than one "acl" directive (due to acl parameter aggregation by ACL objects).

rousskov (approved):
@kinkie, I fixed space handling in "acl" dumps and pushed "streaming" changes one level down, to simplify. Please adjust further as needed.

yadij (approved):
Passing under the "Good enough for now" criteria. Other projects are taking on the additional polish and refactoring this needs for completion.

src/acl/Acl.h
yadij: IMO that is fine for now. The whole "dump" thing is going away with the manager report project.

PR #1574, opened 2023-11-06
Fix acl annotate_transaction reporting in mgr:config

kinkie:
Possibly broken since commit 4eac340 that stopped printing `-m=` because
it customized/specialized too much of the TypedOption::print() code.

### DIFF
--- a/src/acl/CharacterSetOption.h
+++ b/src/acl/CharacterSetOption.h
@@ -29,9 +29,8 @@ TypedOption<CharacterSetOptionValue>::import(const SBuf &rawValue) const
 }
 
 template <>
-inline
-void
-TypedOption<CharacterSetOptionValue>::print(std::ostream &os) const
+inline void
+TypedOption<CharacterSetOptionValue>::printValue(std::ostream &os) const
 {
     recipient_->value.printChars(os); // TODO: Quote if needed.
 }
--- a/src/acl/Options.h
+++ b/src/acl/Options.h
@@ -182,10 +182,18 @@ public:
     {
         if (configured()) {
             os << ' ' << (disabled() ? offName : onName);
-            if (valued())
-                os << '=' << recipient_->value;
+            if (valued()) {
+                os << '=';
+                printValue(os);
+            }
         }
         // else do not report the implicit default
+    }
+
+    /// used in print() to print the option if configured
+    void printValue(std::ostream &os) const
+    {
+        os << '=' << recipient_->value;
     }
 
 private:
--- a/src/acl/Options.h
+++ b/src/acl/Options.h
@@ -190,14 +190,9 @@ public:
         // else do not report the implicit default
     }
 
-    /// used in print() to print the option if configured
-    void printValue(std::ostream &os) const
-    {
-        os << recipient_->value;
-    }
-
 private:
     void import(const SBuf &rawValue) const { recipient_->value = rawValue; }
+    void printValue(std::ostream &os) const { os << recipient_->value; }
 
     // The "mutable" specifier demarcates set-once Option kind/behavior from the
     // ever-changing recipient of the actual admin-configured option value.
--- a/src/acl/Options.h
+++ b/src/acl/Options.h
@@ -193,7 +193,7 @@ public:
     /// used in print() to print the option if configured
     void printValue(std::ostream &os) const
     {
-        os << '=' << recipient_->value;
+        os << recipient_->value;
     }
 
 private:


kinkie:
The configuration:
acl an1 annotate_transaction -m=! n1+=n1v1 n1+=n1v2
acl an2 annotate_transaction n1+=n1v3 n2+=n2v1

gets currently rendered as:
acl an2 annotate_transaction , n1: n1v3
n2: n2v1
acl an1 annotate_transaction ! n1: n1v1
n1: n1v2

With the PR code:
acl an2 annotate_transaction n1: n1v3
n2: n2v1
acl an1 annotate_transaction -m=! n1: n1v1
n1: n1v2

kinkie:
The rendering happens down a pretty deep rabbit hole, it is a generic framework
which however appears to be only ever used for this ACL type

src/acl/CharacterSetOption.h
```diff
@@ -33,7 +33,10 @@ inline
 void
 TypedOption<CharacterSetOptionValue>::print(std::ostream &os) const
 {
-    recipient_->value.printChars(os); // TODO: Quote if needed.
+    if (recipient_->valued) {
+        os << " -m=";
+        recipient_->value.printChars(os); // TODO: Quote if needed.
```
rousskov, line 38: Even if this change fixes output in one particular use cases, the code does not look right, on several levels (that I can detail if needed). FWIW, the official code is also broken -- I am not defending it.

Here is the default code that we are overwriting with this print() customization:

```C++
    void print(std::ostream &os) const override
    {
        if (configured()) {
            os << ' ' << (disabled() ? offName : onName);
            if (valued())
                os << '=' << recipient_->value;
        }
        // else do not report the implicit default
    }
```

AFAICT, we want to customize the `<< recipient_->value` part (i.e. how the CharacterSet value is printed if it should be printed), not the whole option printing mechanism (that has other useful bits that we do _not_ want to throw away).

If we do not customize at all, then the default CharacterSet printing operator will print the character set name, with decorations:

```C++
std::ostream&
operator <<(std::ostream &s, const CharacterSet &c)
{
    s << "CharacterSet(" << c.name << ')';
    return s;
}
```

I believe the correct solution here is to extract value printing from the above print() into a printValue(std::ostream &) method, so that TypedOption<CharacterSetOptionValue>::printValue() can then customize _just that part_ of print() while leaving the rest of print() alone.

@kinkie, if you can take a stub at this extraction, please do. Otherwise, I will find the time to do it.

rousskov (changes requested):
Thank you for detailing output changes!

src/acl/CharacterSetOption.h
kinkie: Sure, I can do this; it feels way overengineered still, but it will work

kinkie: Done.

Here's the output with the current code:
```

acl an2 annotate_transaction n1: n1v3
n2: n2v1

acl an1 annotate_transaction -m=! n1: n1v1
n1: n1v2


```
The newlines are fixed elsewhere

kinkie:
There's a bit of a mess with the commits from master, but I don't have the time to fix it now. Tomorrow, hopefully

src/acl/CharacterSetOption.h
```diff
@@ -29,10 +29,10 @@ TypedOption<CharacterSetOptionValue>::import(const SBuf &rawValue) const
 }
 
 template <>
-inline
-void
-TypedOption<CharacterSetOptionValue>::print(std::ostream &os) const
+inline void
+TypedOption<CharacterSetOptionValue>::printValue(std::ostream &os) const
 {
+    os << '=';
```
rousskov, line 35: This part belongs to the common TypedOption::print() code AFAICT -- all configured() and valued() options should print `=` before the value; that symbol is a part of ACL option value syntax.

rousskov (approved):
I fixed a bug that probably resulted in two `=` being printed and polished. I also adjusted PR title/description to specify which cache manager report got fixed and the likely origin of the fixed bug. Please adjust further as needed.

I have not tested this PR code.

kinkie:
> > I fixed a bug that probably resulted in two `=` being printed and polished. I also adjusted PR title/description to specify which cache manager report got fixed and the likely origin of the fixed bug. Please adjust further as needed.
> > I have not tested this PR code.
> 
> Thanks for that, I can do that now.

I had run-tested the code, must have slipped by me as I was squashing the commits. Thanks for catching the error.

Run-testing;
```
acl an2 annotate_transaction n1: n1v3
n2: n2v1

acl an1 annotate_transaction -m=! n1: n1v1
n1: n1v2
```

Approving and clearing for merge. Thanks!

PR #1575, opened 2023-11-07
Report/abort on any catastrophic rock cache_dir indexing failure

eduard-bagdasaryan:
kid11| 0,3| TextException.cc(110) Throw: mem/PageStack.cc:111: false
    kid11| 93,2| AsyncJob.cc(129) callException: false
    kid11| 93,5| AsyncJob.cc(84) mustStop: Rock::Rebuild will stop
    kid11| Finished rebuilding storage from disk.

Some of the serious cache_dir indexing errors (such as job-stopping
exceptions) were not logged at level 1. After such an error, Squid would
end up using a small (possibly empty) subset of cache_dir slots, usually
without admin knowledge. Other serious errors were reported and treated
as fatal. We now report all serious errors and treat them as fatal.

Also added more rebuild information to reported errors.

Also removed the "Do you need to run 'squid -z' ..." hint. "squid -z" is
only useful for ENOENT, but SwapDir::init() essentially prevents ENOENT
during Rebuild by not starting a Rebuild job when cache_dir is missing.


eduard-bagdasaryan:
Here is an example of level-1 error output, where Squid attempts to deal with an invalid rock db file:

before:

    Loading cache_dir #0 from /bad/rock  
    Do you need to run 'squid -z' to initialize storage?
    Closing HTTP(S) port [::]:3128
    listening port: 3128
    Not currently OK to rewrite swap log.
    storeDirWriteCleanLogs: Operation aborted.
    FATAL: Rock cache_dir[0] rebuild of /bad/rock failed: cannot read db header.

after:

    Loading cache_dir #0 from /bad/rock
    FATAL: Cannot rebuild rock cache_dir index for /bad/rock
        problem: cannot read db header
        scan progress: 0%
        exception location: rock/RockRebuild.cc(723) failure

rousskov (dismissed):

rousskov (approved):

kinkie (approved):

PR #1576, opened 2023-11-07
Reduce ACLChecklist::AsyncState to a function pointer

eduard-bagdasaryan:
ExternalACLLookup::checkForAsync() -- the code that starts communication
with the external ACL helper -- relies on two implicit assumptions:

* AclMatchedName global matches this->name during ACLExternal::match().
* FindByName("x") result never changes during same-Checklist evaluation.

The first assumption might hold[^1], but the second assumption will fail
once we finally start refcounting ACLs instead of allowing
reconfiguration to break in-flight transactions waiting for a slow ACL
match with ACCESS_DUNNO answers. Today, reconfiguration invalidates all
ACL objects, making repeated same-Checklist checkForAsync() calls
impossible across reconfiguration. Tomorrow, such calls will happen and
violation of the second assumption will lead to assertions or worse[^2].
Thus, removing that assumption is a precondition on ACL refcounting and
smooth reconfiguration support.

ExternalACLLookup used FindByName() to find the ACL being matched. The
source code comment suggested that we "have a pointer to this around
somewhere". That pointer is "this" pointer of an ACL calling goAsync().
It is available in the call stack two frames higher, but is not stored
by the Checklist in any usable form.

To remove that FindByName() call, we could just add a second goAsync()
parameter to pass the ACL pointer to Checklist. Doing that would require
changing all the goAsync() callers and the corresponding
ACLChecklist::AsyncState API. Since that very complex and confusing API
was actually used as a basic function pointer, we decided it is best to
replace the has-to-be-changed-anyway API with a function pointer. That
plan worked well, removing a lot of unnecessary and confusing code!

[^1]: It is difficult to be sure: AclMatchedName global was named and
added for a very different purpose, and no API enforces that invariant.

[^2]: When ACLs are refcountered, ACLExternal::match() will continue to
set AclMatchedName global to the name (e.g., "x") of an ACL object that
resumes matching after Squid reconfigures and receives a helper reply.
However, the same FindByName("x") call may now return nil or a
completely different ACL object (still named "x" in the new
configuration, but possibly no longer an ACLExternal object)!

### DIFF
--- a/src/acl/DestinationDomain.cc
+++ b/src/acl/DestinationDomain.cc
@@ -19,10 +19,9 @@
 static void LookupDone(const char *, const Dns::LookupDetails &, void *data);
 
 static void
-StartLookup(ACLChecklist &cl, const ACL &)
+StartLookup(ACLFilledChecklist &cl, const ACL &)
 {
-    const auto checklist = Filled(&cl);
-    fqdncache_nbgethostbyaddr(checklist->dst_addr, LookupDone, checklist);
+    fqdncache_nbgethostbyaddr(cl.dst_addr, LookupDone, &cl);
 }
 
 static void
--- a/src/acl/DestinationDomain.cc
+++ b/src/acl/DestinationDomain.cc
@@ -16,28 +16,22 @@
 #include "fqdncache.h"
 #include "HttpRequest.h"
 
-DestinationDomainLookup DestinationDomainLookup::instance_;
+static void LookupDone(const char *, const Dns::LookupDetails &, void *data);
 
-DestinationDomainLookup *
-DestinationDomainLookup::Instance()
+static void
+StartLookup(ACLChecklist &cl, const ACL &)
 {
-    return &instance_;
-}
-
-void
-DestinationDomainLookup::checkForAsync(ACLChecklist *cl) const
-{
-    ACLFilledChecklist *checklist = Filled(cl);
+    const auto checklist = Filled(&cl);
     fqdncache_nbgethostbyaddr(checklist->dst_addr, LookupDone, checklist);
 }
 
-void
-DestinationDomainLookup::LookupDone(const char *, const Dns::LookupDetails &details, void *data)
+static void
+LookupDone(const char *, const Dns::LookupDetails &details, void *data)
 {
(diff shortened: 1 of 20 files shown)


src/external_acl.cc
```diff
@@ -1139,32 +1143,13 @@ externalAclShutdown(void)
...
-void
-ExternalACLLookup::checkForAsync(ACLChecklist *checklist)const
-{
-    /* TODO: optimise this - we probably have a pointer to this
-     * around somewhere */
-    ACL *acl = ACL::FindByName(AclMatchedName);
-    assert(acl);
-    ACLExternal *me = dynamic_cast<ACLExternal *> (acl);
-    assert (me);
-    ACLExternal::ExternalAclLookup(checklist, me);
```
rousskov, line 1158: This is that problematic FindByName() call that was blocking ACL refcounting. This PR addresses the above optimization TODO as well, but the related speedup ought to be negligible in most environments, and it is not the focus of this PR.

src/acl/DestinationDomain.cc
```diff
@@ -16,28 +16,21 @@
...
-DestinationDomainLookup DestinationDomainLookup::instance_;
+static void LookupDone(const char *, const Dns::LookupDetails &, void *data);
 
-DestinationDomainLookup *
-DestinationDomainLookup::Instance()
+static void
+StartLookup(ACLFilledChecklist &cl, const ACL &)
 {
-    return &instance_;
+    fqdncache_nbgethostbyaddr(cl.dst_addr, LookupDone, &cl);
```
rousskov, line 24: The diff does not make it obvious, but we were declaring a hierarchy of AsyncState classes just to make async function calls like this fqdncache_nbgethostbyaddr() call. Those classes are now replaced by simple static functions like this StartLookup(), simplifying Checklist logic quite a bit!

In the foreseeable future, we would be able to get rid of those static functions as well, but that requires improving all async functions like fqdncache_nbgethostbyaddr() to become fully async, which requires non-trivial work that deserves dedicated PRs. Today, those functions sometimes call their callbacks immediately, before returning to the caller, creating all sorts of problems.

rousskov (dismissed):

rousskov (approved):

rousskov (approved):

PR #1577, opened 2023-11-08
Fix qos_flows confguration reporting

kinkie:
mgr:config output contained garbage when qos_flows was not configured.


kinkie:
With configuration:
```
qos_flows mark local-hit=0x01 sibling-hit=0x02 parent-hit=0x03 miss=0xf0/0x10
```
cachemgr output is:
```
qos_flows mark local-hit=0x1 sibling-hit=0x2 parent-hit=0x3 miss=0xf0/0x10 disable-preserve-miss
```
with master, cachemgr output is:
```
qos_flowsmark local-hit=0x01 sibling-hit=0x02 parent-hit=0x03 miss=0xF0/0x10 disable-preserve-miss
host_verify_strict off
```

When qos_flows is not configured, cachemgr output is:
```
http_port [::]:3128 name=3128 connection-auth=on tls-disable
https_port [::]:3128 name=3128 connection-auth=on tls-disable
<E8>D<A5> tUhost_verify_strict off
```

src/ip/QosConfig.cc
```diff
@@ -465,64 +467,61 @@ Ip::Qos::Config::parseConfigLine()
...
-    char *p = entry;
     if (isHitTosActive()) {
-
-        p += snprintf(p, 11, "%s", name); // strlen("qos_flows ");
-        p += snprintf(p, 4, "%s", "tos");
+        os << name << " tos";
 
         if (tosLocalHit > 0) {
-            p += snprintf(p, 16, " local-hit=0x%02X", tosLocalHit);
+            os << " local-hit=" << AsHex(tosLocalHit);
```
rousskov, line 476: For consistency sake, please use the lowercase asHex() wrapper rather than instantiating an object of the templated class AsHex. Same for other stream manipulators. IIRC, in some special cases, the compiler will not be able to guess the template type when instantiating an object, but will be able to guess it when calling a function. That is why we usually provide those wrappers...

src/ip/QosConfig.cc
```diff
@@ -465,64 +467,61 @@ Ip::Qos::Config::parseConfigLine()
...
         }
         if (tosSiblingHit > 0) {
-            p += snprintf(p, 18, " sibling-hit=0x%02X", tosSiblingHit);
+            os << " sibling-hit=" << AsHex(tosSiblingHit);
```
rousskov, line 479: The old code specified the width and, hence, printed leading zeros for single-digit hex values. If this PR should preserve that functionality, let's preserve it (e.g., by adding/using the corresponding AsHex method(s)). Otherwise, please note this output format change in the PR description.

src/ip/QosConfig.h
```diff
@@ -240,18 +242,14 @@ class Config
...
-        (X).dumpConfigLine(temp,n); \
-            storeAppendPrintf(e, "%s", temp); \
-    } while(0);
-
 } // namespace Qos
 
 } // namespace Ip
 
+/* legacy parser access wrappers */
+inline void parse_QosConfig(Ip::Qos::Config * theConfig) { theConfig->parseConfigLine(); }
```
rousskov, line 250: We tend to use `theX` format for class data members (and only in legacy code; modern code uses `x_` instead).

```suggestion
inline void parse_QosConfig(Ip::Qos::Config * const c) { c->parseConfigLine(); }
```

rousskov (changes requested):
Thank you very much for fixing this annoying bug!

rousskov:
I adjusted PR title/description to use better-known terms and fix formatting. I stripped implementation details that the diff makes obvious. Please adjust further as needed.

src/ip/QosConfig.cc
kinkie: In practice it seems to be working, as shown in the test output.
But I agree it's a good idea to make sure it's not by accident, so I'll add that as well

src/ip/QosConfig.cc
```diff
@@ -465,64 +467,61 @@ Ip::Qos::Config::parseConfigLine()
...
-            p += snprintf(p, 23, " disable-preserve-miss");
+            os << " disable-preserve-miss";
         }
         if (preserveMissTos && preserveMissTosMask != 0) {
-            p += snprintf(p, 16, " miss-mask=0x%02X", preserveMissTosMask);
+            os << " miss-mask=" << AsHex(preserveMissTosMask);
         }
-        p += snprintf(p, 2, "\n");
+        os << "\n";
+
```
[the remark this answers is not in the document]
kinkie: No harm in adding it

src/ip/QosConfig.cc
rousskov: > I'll post a separate PR to add this snippet of knowledge to the IoManip.h file

Please note that it would be difficult to document this detail further without duplicating that documentation. We already have many such wrappers (at least two in IoManip.h and five more for various AsyncCalls: `git grep -E 'helper.*to (ease|simplify).*creation'`). We could hope that folks would not see any similar AsHex usage in the existing code and use asHex() instead, but that obviously does not really work well.

Alternatively, we can stop thriving for consistency in this area -- use AsHex() and asHex() interchangeably, without any change requests to rename.

A third option I can think of is to prove that asHex() is not needed -- that the cases such wrappers are meant to address do not apply to AsHex class. In that case, we can remove asHex() completely and solve the consistency problem that way!

src/ip/QosConfig.cc
rousskov: > In practice it seems to be working, as shown in the test output.

The test [output](https://github.com/squid-cache/squid/pull/1577#issuecomment-1802098544) shows that the current PR code is _not_ "working" (as far as this change request is concerned):

```
qos_flows mark local-hit=0x1 sibling-hit=0x2 parent-hit=0x3 ...
```

kinkie:
waiting for PR #1578 for the toHex() width specifier

kinkie:
Bugs should be fixed now. As soon as PR #1578 is in, this should be ready to go. Requesting review in the meanwhile

src/ip/QosConfig.cc
```diff
@@ -465,64 +467,61 @@ Ip::Qos::Config::parseConfigLine()
...
-    char *p = entry;
     if (isHitTosActive()) {
-
-        p += snprintf(p, 11, "%s", name); // strlen("qos_flows ");
-        p += snprintf(p, 4, "%s", "tos");
+        os << name << " tos";
 
         if (tosLocalHit > 0) {
-            p += snprintf(p, 16, " local-hit=0x%02X", tosLocalHit);
+            os << " local-hit=0x" << asHex(tosLocalHit).minDigits(2);
```
rousskov, line 476: `%X` uses [uppercase](https://en.cppreference.com/w/c/io/fprintf). The stream default is lowercase. To preserve formatting, we should add .upperCase() to all these asHex() calls, right?

When fixing this, please put upperCase() in front of minDigits() rather than behind it (see commit 8d85b10 in your asHex() PR).

BTW, this is as use case that can benefit greatly from using a local function to avoid all this code duplication. For example:

```C++
template <class Integer>
static auto asQosHex(const Integer n) { return asHex(n).upperCase().minDigits(2); }
```

rousskov (changes requested):

src/ip/QosConfig.cc
kinkie: I'll call it asQosConfigHex, but sure

src/ip/QosConfig.h
```diff
@@ -187,7 +189,7 @@ class Config
      * objects are part of the dump function must be self-contained.
      * which means no StoreEntry references. Just a basic char* buffer.
      */
```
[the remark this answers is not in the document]
kinkie: done

src/ip/QosConfig.h
```diff
@@ -187,7 +189,7 @@ class Config
...
-    void dumpConfigLine(char *entry, const char *name) const;
+    void dumpConfigLine(std::ostream &, const char *) const;
```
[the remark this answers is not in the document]
kinkie: done

src/ip/QosConfig.h
```diff
@@ -240,18 +242,14 @@ class Config
...
-    } while(0);
-
 } // namespace Qos
 
 } // namespace Ip
 
+/* legacy parser access wrappers */
+inline void parse_QosConfig(Ip::Qos::Config * c) { c->parseConfigLine(); }
+inline void free_QosConfig(Ip::Qos::Config *) {}
+void dump_QosConfig(StoreEntry *, const char *, const Ip::Qos::Config &);
```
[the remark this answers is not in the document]
kinkie: ok

rousskov (approved):
Thank you for addressing my concerns. I pushed a few polishing commits and simplified PR title.

yadij (approved):

PR #1578, opened 2023-11-08
Improve and broadly use asHex()

kinkie:
Replaced all raw std::hex uses for integers outside of IoManip code.

Also fixed std::ostream flags restoration in AsHex printing code.

Also fixed or polished a few level-2+ debugs() statements.

Also added AsHex unit tests.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -2757,3 +2757,18 @@ tests_testEventLoop_LDADD = \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 tests_testEventLoop_LDFLAGS = $(LIBADD_DL)
+
+check_PROGRAMS += tests/testIoManip
+tests_testIoManip_SOURCES = \
+	tests/testIoManip.cc
+nodist_tests_testIoManip_SOURCES = \
+	tests/stub_SBuf.cc \
+	tests/stub_debug.cc \
+	tests/stub_libmem.cc \
+	tests/stub_libtime.cc
+tests_testIoManip_LDADD = \
+	base/libbase.la \
+	$(LIBCPPUNIT_LIBS) \
+	$(COMPAT_LIB) \
+	$(XTRA_LIBS)
+tests_testIoManip_LDFLAGS = $(LIBADD_DL)
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1726,6 +1726,24 @@ EXTRA_DIST += \
 	tests/testACLMaxUserIP.cc
 endif
 
+## Tests of html/*
+
+check_PROGRAMS += tests/testHtmlQuote
+tests_testHtmlQuote_SOURCES = \
+	tests/testHtmlQuote.cc
+nodist_tests_testHtmlQuote_SOURCES = \
+	$(TESTSOURCES) \
+	tests/stub_debug.cc \
+	tests/stub_libmem.cc
+tests_testHtmlQuote_LDADD= \
+	html/libhtml.la \
+	sbuf/libsbuf.la \
+	base/libbase.la \
+	$(LIBCPPUNIT_LIBS) \
+	$(COMPAT_LIB) \
+	$(XTRA_LIBS)
+tests_testHtmlQuote_LDFLAGS = $(LIBADD_DL)
+
 ## Tests of http/* and HTTP Protocol objects
 
 check_PROGRAMS += tests/test_http_range
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -95,9 +95,7 @@ operator <<(std::ostream &os, const AsHex<Integer> number)
 {
     const auto oldFlags = os.flags();
     const auto savedFill = os.fill('0');
-    if (number.upperCase_)
-        os << std::uppercase;
-    os << std::hex <<
+    os << std::hex << (number.upperCase_ ? std::uppercase : std::nouppercase) <<
        std::setw(number.width) <<
        number.io_manip <<
        std::setw(0);
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -75,7 +75,7 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
     return os;
 }
 
-/// std::ostream manipulator to print integers as hex numbers prefixed by 0x
+/// std::ostream manipulator to print integers as hex numbers
 /// Normally used through the asHex() convenience function
 template <class Integer>
 class AsHex
@@ -85,7 +85,7 @@ public:
     auto &minDigits(const size_t w) { width = w; return *this; }
     auto &upperCase(bool u = true) { upperCase_ = u; return *this; }
     Integer io_manip; ///< the integer to print
-    size_t width = 0; ///< the minimum number of digits to print after the 0x prefix
+    size_t width = 0; ///< the minimum number of digits to print
     bool upperCase_ = false; ///< output in uppercase?
 };
 
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -111,7 +111,7 @@ template <class Integer>
 class AsHex
 {
 public:
-    explicit AsHex(const Integer n) : io_manip(n) {}
+    explicit AsHex(const Integer n): io_manip(n) {}
 
     /// Sets the minimum number of digits to print. If the integer has fewer
     /// digits than the given width, then we also print leading zero(s).
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -13,6 +13,7 @@
 
 #include <iostream>
 #include <iomanip>
+#include <optional>
 
 /// Safely prints an object pointed to by the given pointer: [label]<object>
 /// Prints nothing at all if the pointer is nil.
@@ -82,11 +83,22 @@ class AsHex
 {
 public:
     explicit AsHex(const Integer n) : io_manip(n) {}
+
+    /// Sets the minimum number of digits to print. If the integer has fewer
+    /// digits than the given width, then we print leading zero(s). Otherwise,
+    /// this method has no effect.
     auto &minDigits(const size_t w) { width = w; return *this; }
-    auto &upperCase(bool u = true) { upperCase_ = u; return *this; }
+
+    /// Print hex digits in upper (or, with a false parameter value, lower) case.
+    auto &upperCase(const bool u = true) { forceCase = u; return *this; }
+
     Integer io_manip; ///< the integer to print
-    size_t width = 0; ///< the minimum number of digits to print
-    bool upperCase_ = false; ///< output in uppercase?
+
+    ///< \copydoc minDigits(). The default is to use stream's field width.
+    std::optional<size_t> width;
+
+    ///< \copydoc upperCase(). The default is to use stream's std::uppercase flag.
+    std::optional<bool> forceCase;
 };
 
 template <class Integer>
@@ -94,11 +106,19 @@ inline std::ostream &
 operator <<(std::ostream &os, const AsHex<Integer> number)
 {
     const auto oldFlags = os.flags();
-    const auto savedFill = os.fill('0');
-    os << std::hex << (number.upperCase_ ? std::uppercase : std::nouppercase) <<
-       std::setw(number.width) <<
-       number.io_manip;
-    os.fill(savedFill);
+    const auto oldFill = os.fill();
+
+    if (number.forceCase)
+        os << (*number.forceCase ? std::uppercase : std::nouppercase);
+
+    if (number.width) {
+        os.width(*number.width);
+        os.fill('0');
+    }
+
+    os << std::hex << number.io_manip;
+
+    os.fill(oldFill);
     os.flags(oldFlags);
     return os;
 }
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -15,6 +15,35 @@
 #include <iomanip>
 #include <optional>
 
+/// \section Custom manipulator tuning methods
+///
+/// Our convenience manipulator/wrapper classes often have methods that tune
+/// their "printing" effects (e.g., AsHex::minDigits()). STL streams also have
+/// manipulators that tune how subsequent operator "<<" parameters are printed
+/// (e.g., std::setw()). The calling code can also print various decorations
+/// (i.e. prefixes and suffixes). The following principles are useful when
+/// deciding what manipulator methods to add and how to implement them:
+///
+/// \li Add a manipulator method if callers would otherwise have to restore
+/// stream format after calling the manipulator. For example, AsHex::toUpper()
+/// frees callers from doing `std::uppercase << asHex(n) << std::nouppercase`.
+///
+/// \li Add a manipulator method if callers would otherwise have to use
+/// conditionals to get the same effect. For example, AsList::prefixedBy() frees
+/// callers from doing `(c.empty() ? "" : "/") << asList(c)`.
+///
+/// \li Add a manipulator method if callers would otherwise have to repeat a
+/// combination of actions to get the right effect. For example,
+/// AsList::minDigits() prevents duplication of the following caller code:
+/// `std::setfill('0') << std::setw(8) << asHex(n)`.
+///
+/// \li Avoid adding a manipulator method that can be _fully_ replaced with a
+/// _single_ caller item. For example, do not add AsX::foo() if callers can do
+/// `bar << asX(y)` or `asX(y) << bar` and get exactly the same effect.
+///
+/// \li Manipulators should honor existing stream formatting to the extent
+/// possible (e.g., AsHex honors std::uppercase by default).
+
 /// Safely prints an object pointed to by the given pointer: [label]<object>
 /// Prints nothing at all if the pointer is nil.
 template <class Pointer>
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -114,17 +114,18 @@ public:
     explicit AsHex(const Integer n) : io_manip(n) {}
 
     /// Sets the minimum number of digits to print. If the integer has fewer
-    /// digits than the given width, then we print leading zero(s). Otherwise,
-    /// this method has no effect.
-    auto &minDigits(const size_t w) { width = w; return *this; }
+    /// digits than the given width, then we also print leading zero(s).
+    /// Otherwise, this method has no effect.
+    auto &minDigits(const size_t w) { forcePadding = w; return *this; }
 
     /// Print hex digits in upper (or, with a false parameter value, lower) case.
     auto &upperCase(const bool u = true) { forceCase = u; return *this; }
 
     Integer io_manip; ///< the integer to print
 
-    ///< \copydoc minDigits(). The default is to use stream's field width.
-    std::optional<size_t> width;
+    ///< \copydoc minDigits(). The default is to use stream's field width and
+    ///stream's fill character.
+    std::optional<size_t> forcePadding;
 
     ///< \copydoc upperCase(). The default is to use stream's std::uppercase flag.
     std::optional<bool> forceCase;
@@ -140,8 +141,8 @@ operator <<(std::ostream &os, const AsHex<Integer> number)
     if (number.forceCase)
         os << (*number.forceCase ? std::uppercase : std::nouppercase);
 
-    if (number.width) {
-        os.width(*number.width);
+    if (number.forcePadding) {
+        os.width(*number.forcePadding);
         os.fill('0');
     }
 
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -83,11 +83,9 @@ class AsHex
 public:
     explicit AsHex(const Integer n) : io_manip(n) {}
     auto &minDigits(const size_t w) { width = w; return *this; }
-    auto &printPrefix(bool p) { printPrefix_ = p; return *this; }
     auto &upperCase(bool u = true) { upperCase_ = u; return *this; }
     Integer io_manip; ///< the integer to print
     size_t width = 0; ///< the minimum number of digits to print after the 0x prefix
-    bool printPrefix_ = true; ///< output 0x prefix
     bool upperCase_ = false; ///< output in uppercase?
 };
 
@@ -97,8 +95,6 @@ operator <<(std::ostream &os, const AsHex<Integer> number)
 {
     const auto oldFlags = os.flags();
     const auto savedFill = os.fill('0');
-    if (number.printPrefix_)
-        os << "0x" ;
     if (number.upperCase_)
         os << std::uppercase;
     os << std::hex <<
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -99,7 +99,7 @@ operator <<(std::ostream &os, const AsHex<Integer> number)
        std::setw(number.width) <<
        number.io_manip;
     os.fill(savedFill);
-    os.setf(oldFlags);
+    os.flags(oldFlags);
     return os;
 }
 
--- a/src/base/IoManip.h
+++ b/src/base/IoManip.h
@@ -111,6 +111,12 @@ template <class Integer>
 class AsHex
 {
 public:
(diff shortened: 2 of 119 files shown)


src/base/IoManip.h
```diff
@@ -76,27 +76,31 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
 }
 
 /// std::ostream manipulator to print integers as hex numbers prefixed by 0x
+/// Normally used through the asHex() convenience function
 template <class Integer>
 class AsHex
 {
 public:
-    explicit AsHex(const Integer n): io_manip(n) {}
+    explicit AsHex(const Integer n, const size_t w=0) : io_manip(n), width(w) {}
```
rousskov, line 84: Please use a named method to configure output, not an (anonymous from the caller point of view) constructor parameter. See class Raw for examples. Among other benefits, this approach does not result in duplication of new construction parameters in the asHex() wrapper.

The functionality this PR is currently adding goes way beyond specifying the width (which could probably be done using an existing std::setw() manipulator _in the asHex() caller_). Thus, the new method name should not be width() or similar. I can suggest `withLeadingZeros(size_t minDigits)`.

src/base/IoManip.h
```diff
@@ -76,27 +76,31 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
 }
 
 /// std::ostream manipulator to print integers as hex numbers prefixed by 0x
+/// Normally used through the asHex() convenience function
 template <class Integer>
 class AsHex
```
rousskov, line 81: Please consider applying the new functionality to std::hex code inside Eui::Eui48::lookup(), Rock::SwapDir::createStoreIO(), Rock::SwapDir::createUpdateIO(), Rock::SwapDir::openStoreIO() and other contexts. Search for "std::hex" and "std::setfill". Keep std::uppercase.

I do not insist on this application, only consideration.

src/base/IoManip.h
```diff
@@ -76,27 +76,31 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
...
 {
 public:
-    explicit AsHex(const Integer n): io_manip(n) {}
+    explicit AsHex(const Integer n, const size_t w=0) : io_manip(n), width(w) {}
     Integer io_manip; ///< the integer to print
+    size_t width; ///< the width to use number of digits to print
```
rousskov, line 86:
```suggestion
    size_t width; ///< the minimum number of digits to print, not counting the 0x prefix
```

src/base/IoManip.h
```diff
@@ -76,27 +76,31 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
...
 template <class Integer>
 inline std::ostream &
 operator <<(std::ostream &os, const AsHex<Integer> number)
 {
     const auto oldFlags = os.flags();
-    os << std::hex << std::showbase << number.io_manip;
+    os << "0x" << std::hex <<
+        std::setfill('0') << std::setw(number.width) <<
+        number.io_manip;
     os.setf(oldFlags);
```
rousskov, line 97: Please remember and restore the fill character that the new code is changing. It is not a flag. See how PrintHex() does it.

Please check whether width is automatically reset after the number is printed. If it is not, we need to remember and restore it as well.

src/tests/testSBuf.cc
```diff
@@ -1190,6 +1203,17 @@ TestSBuf::testSBufHash()
     }
 }
 
+void
+TestIoManip::testAsHex()
+{
+    SBufStream os;
+    os << asHex(0xa0);
+    CPPUNIT_ASSERT_EQUAL(SBuf("0xa0"), os.buf());
```
rousskov, line 1211: You can simplify this testing code a lot using this pattern:

```C++
    CPPUNIT_ASSERT_EQUAL(SBuf("0xa0"), ToSBuf(asHex(0xa0)));
```

src/tests/testSBuf.cc
```diff
@@ -106,9 +107,21 @@ class TestSBuf : public CPPUNIT_NS::TestFixture
     void testStdStringOps();
     void testIterators();
     void testSBufHash();
+    void testAsHex();
 };
 CPPUNIT_TEST_SUITE_REGISTRATION( TestSBuf );
 
+class TestIoManip : public CPPUNIT_NS::TestFixture
```
rousskov, line 114: IoManip is unrelated to SBuf. We should not add this new test into testSBuf.cc, even if the test is using SBuf.

src/tests/testSBuf.cc
```diff
@@ -1190,6 +1203,17 @@ TestSBuf::testSBufHash()
...
 
+void
+TestIoManip::testAsHex()
+{
+    SBufStream os;
+    os << asHex(0xa0);
+    CPPUNIT_ASSERT_EQUAL(SBuf("0xa0"), os.buf());
+    os.clearBuf();
+    os << asHex(0xa0,4);
+    CPPUNIT_ASSERT_EQUAL(SBuf("0x00a0"), os.buf());
```
rousskov, line 1214: If you are going to test this, please do add a test where the number of digits higher than the limit and where the limit is zero.

```C++
CPPUNIT_ASSERT_EQUAL(SBuf("0x1234"), ToSBuf(asHex(0x1234).withLeadingZeros(2));
CPPUNIT_ASSERT_EQUAL(SBuf("0x4"), ToSBuf(asHex(0x00004).withLeadingZeros(0));
```

src/base/IoManip.h
```diff
@@ -76,27 +76,31 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
...
     return os;
 }
 
-/// a helper to ease AsHex object creation
+/// helpers to ease AsHex object creation
```
rousskov, line 101: Undo because there is only one helper:

```suggestion
/// a helper to ease AsHex object creation
```

rousskov (changes requested):

src/base/IoManip.h
kinkie: Unfortunately setw() won't work (I tried), with std::hex, because the output would then look like '0000x0abcd'. This is also the reason why I have changed how the prefix is added

src/base/IoManip.h
kinkie: Added fill character reset.

Regarding resetting width, it's guaranteed to happen every time are sent to the stream (among other cases) " arithmetic types or void pointers (at Stage 3 of num_put::put())".

The fun thing is, I can see nothing in this code that constrains it to be used with integer types only. The width is reset in other cases, such as when char, c-string and basic_string are output, so there could be a scenario where some one misuses this code to output something which is not a number and doesn't get reduced to a c-string in its operator<<.

I'll still add a setw(0) because it gives us a hard guarantee at a probably low cost, but it could be debaeable

src/tests/testSBuf.cc
kinkie: sure, thanks

src/tests/testSBuf.cc
rousskov: >  this can be some throwaway tests for development and this is a handy place to place them.

You would need to define the implications of "throwaway tests for development" for this argument to be usable in a discussion. For example, if the implication is that these tests should be thrown away at the end of this PR, then please say so. Removing them will obviously address this change request.

>  It could also be something we could be interested into as a model moving forward, to get out of the stub hell we currently have to maintain

If you want to propose a "model" change, please do so first/separately/properly. I am happy to discuss such a proposal. Until then, arguments based on a new but insufficiently specified model are not admissible IMO.

yadij (approved):

src/tests/testSBuf.cc
yadij: I agree with Alex on this one. Please move this test to a separate unit -see `src/Makefile.am` line 807 and later for where to place the tests of `src/base/` code.

src/base/IoManip.h
kinkie: For Eui things, I think the best solution is to implement an operator<< for the relevant types, which is
out of scope here.

I've gone over the list of std::hex users, I think I caught all those for which it makes sense right away.

src/tests/testSBuf.cc
kinkie: @rousskov went ahead and implemented it while I was away. Thanks for doing that!

rousskov (changes requested):
Thank you for making excellent progress with this PR! I will need to request or make a few adjustments later, but I left two change requests that may help us keep moving forward at a decent speed.

src/base/IoManip.h
```diff
@@ -76,20 +76,36 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
...
 /// std::ostream manipulator to print integers as hex numbers prefixed by 0x
+/// Normally used through the asHex() convenience function
 template <class Integer>
 class AsHex
 {
 public:
-    explicit AsHex(const Integer n): io_manip(n) {}
+    explicit AsHex(const Integer n) : io_manip(n) {}
+    auto &minDigits(const size_t w) { width = w; return *this; }
+    auto &printPrefix(bool p) { printPrefix_ = p; return *this; }
```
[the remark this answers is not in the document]
kinkie: done

src/tests/testSBuf.cc
kinkie: Sigh.. I crossed wires with html_quote. Too much enthusiasm

src/Makefile.am
```diff
@@ -2757,3 +2757,18 @@ tests_testEventLoop_LDADD = \
 	$(COMPAT_LIB) \
 	$(XTRA_LIBS)
 tests_testEventLoop_LDFLAGS = $(LIBADD_DL)
+
+check_PROGRAMS += tests/testIoManip
```
rousskov, line 2761:
```
FAIL: tests/testIoManip
```

@kinkie, if you do not mind, please check why https://github.com/squid-cache/squid/actions/runs/7021325784?pr=1578 CI logs do not show any details about the current testIoManip failure. If I just failed to find those details in the logs, please let me know what I missed. I wonder if there is something wrong with our CI scripts that hides those details...

I do not insist on this investigation or any related changes (because I do not know whether the problem, if any, is related to this PR).

rousskov (changes requested):
Thank you for addressing several large change requests! I added two more (including, AFAICT, a bug fix) to keep this PR moving forward.

src/base/IoManip.h
```diff
@@ -76,20 +76,30 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
 }
 
 /// std::ostream manipulator to print integers as hex numbers prefixed by 0x
```
yadij, line 78: The "prefixed by 0x" part of this documentation appears not to be true.

AFAICS most of the uses do actually need to display the value with that prefix. So perhapse it would be worth printing the prefix by default with a `noPrefix()` method to prevent its display. That would clean up a fair bit of mess in the debugs.

yadij (changes requested):

src/base/IoManip.h
kinkie: That was my original implementation, using printPrefix(false) - positives are easier to understand.
Based on feedback from @rousskov , I’ve removed that code; for one, the cases where printing a prefix is NOT wanted are more than the opposite. The prefix can be printed by the surrounding code easily

rousskov (dismissed):
I pushed many changes, including a few bug fixes, and adjusted PR description accordingly. @kinkie, please review and retest.

Here is a list of arguably noteworthy branch commits (with the most important ones in bold face):

* 1530d9e Fix directory components reporting in Fs::Ufs::RebuildState::getNextFile()
* 3e48bb6 **Prohibit printing of pointers and floating point numbers asHex()**
* 9e67753 fixup: Undo out-of-scope branch ExceptionErrorDetail format changes
* e912dc7 fixup: Restored printing devpoll_state[fd].state as hex
* d7b52b1 fixup: Fix support for smaller integer types like char
* efd5e4b75e835c41677a1934029e56211af57495 fixup: Added **a negative value test** and more simple value tests
* 41014b6 fixup: Removed AsHex tests that were moved to testIoManip
* 809692c Converted ldap_escape_value()
* 1fef97c Updated (unused and broken) icpPktDump()
* 7398ef6 **Deduplicated Eui48 address printing code**
* afb4f3b fixup: Restored printing epoll_state as hex
* 79d67f5 **Documented the principles behind custom manipulator tuning methods**
* f0b584d **Re-implemented AsHex printing to honor stream formatting**
* 4326dc9 fixup: Fixed stream resetting between test cases
* 7322cc8 **Fixed stream flags restoration code**
* 468646a fixup: Do not setw() twice

There is one more raw std::hex use in IoManip.cc. We can easily replace it at the expense of PrintHex() performance (we would be moving AsHex stream manipulations inside each loop iteration; clang was not capable of optimizing that overhead away in my [quick test](https://godbolt.org/z/Kc551jYs5)).

The three remaining raw std::hex uses outside of IoManip code do not apply to integers AFAICT. All three are probably wrong -- this official code should be adjusted to remove 0x, std::hex, and std::dec, but those adjustments are outside this PR scope (and are difficult for me to test):

```
src/icmp/IcmpSquid.cc:    debugs(37, 4, "Pinger handle: 0x" << std::hex << hIpc << std::dec << ", PID: " << pid
);

src/unlinkd.cc:    debugs(2, 4, "Unlinkd handle: 0x" << std::hex << hIpc << std::dec << ", PID: " << pid);

src/wccp2.cc:                debugs (80, 5,  "checking cache list: (" << std::hex << cache_address.s_addr << ":" <<  router_list_ptr->local_ip.s_addr << ")");
```

src/Makefile.am
kinkie: dismissing stale comment; tests are run and passing

kinkie: And by the way: thanks for improving the tests to the level you did, @rousskov !

src/base/IoManip.h
```diff
@@ -13,6 +13,36 @@
...
+/// combination of actions to get the right effect. For example,
+/// AsList::minDigits() prevents duplication of the following caller code:
+/// `std::setfill('0') << std::setw(8) << asHex(n)`.
+///
+/// \li Avoid adding a manipulator method that can be _fully_ replaced with a
+/// _single_ caller item. For example, do not add AsX::foo() if callers can do
+/// `bar << asX(y)` or `asX(y) << bar` and get exactly the same effect.
+///
+/// \li Manipulators should honor existing stream formatting to the extent
+/// possible (e.g., AsHex honors std::uppercase by default).
```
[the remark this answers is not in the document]
kinkie: Looks good! I corrected a small mistake

src/Makefile.am
rousskov: > dismissing stale comment; tests are run and passing

Please note that my request was _not_ about fixing not running or failing tests. It was about our CI logs missing test failure details. Sooner or later, testIoManip will fail for some PR. It would be nice to avoid dealing with the same problem then. However, it is possible that the "missing failure details" problem, whatever it was, has been unknowingly fixed since then.

I do not insist on any changes related to this change request.

kinkie: I tried injecting a failure into testIoManip, here is the outcome of the test:

```
[...]
PASS: tests/testEventLoop
FAIL: tests/testIoManip
============================================================================
Testsuite summary for Squid Web Proxy 7.0.0-VCS
============================================================================
# TOTAL: 36
# PASS:  35
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See src/test-suite.log
Please report to https://bugs.squid-cache.org/
============================================================================
make[4]: *** [Makefile:6202: test-suite.log] Error 1
make[4]: Leaving directory '/home/kinkie/src/squid/btbuild/src'
make[3]: *** [Makefile:6310: check-TESTS] Error 2
make[3]: Leaving directory '/home/kinkie/src/squid/btbuild/src'
make[2]: *** [Makefile:6653: check-am] Error 2
make[2]: Leaving directory '/home/kinkie/src/squid/btbuild/src'
make[1]: *** [Makefile:6094: check-recursive] Error 1
make[1]: Leaving directory '/home/kinkie/src/squid/btbuild/src'
make: *** [Makefile:6655: check] Error 2
make: Leaving directory '/home/kinkie/src/squid/btbuild/src'
```

The test framework left behind ```src/tests/testIoManip.log ``` and ```src/tests/testIoManip.trs```

src/tests/testIoManip.log contains:
```
SKIP: mem/libmem.la MemPools (not implemented).
.F
testIoManip.cc:155:Assertion
Test name: TestIoManip::testAsHex
equality assertion failed
- Expected: 1
- Actual  : 0

Failures !!!
Run: 1   Failure total: 1   Failures: 1   Errors: 0
FAIL tests/testIoManip (exit status: 1)
```

and src/tests/testIoManip.trs contains
```
:test-result: FAIL
:global-test-result: FAIL
:recheck: yes
:copy-in-global-log: yes
```

It should be easy to add a reporting step for failures

src/base/RandomUuid.cc
```diff
@@ -91,22 +91,15 @@ RandomUuid::serialize() const
...
-    const auto savedFlags = os.flags();
-    const auto savedFill = os.fill('0');
-
-    os << std::hex;
-
     os <<
-       std::setw(8) << timeLow  << '-' <<
-       std::setw(4) << timeMid << '-' <<
-       std::setw(4) << timeHiAndVersion << '-' <<
-       std::setw(2) << +clockSeqHiAndReserved << std::setw(2) << +clockSeqLow << '-';
```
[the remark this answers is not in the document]
kinkie: Noted to do as a followup

kinkie:
This looks like a plausible candidate for merge.
@rousskov , @yadij , could you review and clear at your earliest convenience?

rousskov (dismissed):
I made two trivial corrections, but the PR still looks good to me overall.

kinkie:
> I made two trivial corrections, but the PR still looks good to me overall.

Great! As soon as @yadij has a chance to weigh in and hopefully retract his objection to merge, this can go in

src/base/IoManip.h
```diff
@@ -75,22 +105,61 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
...
+    /// Otherwise, this method has no effect.
+    auto &minDigits(const size_t w) { forcePadding = w; return *this; }
+
+    /// Print hex digits in upper (or, with a false parameter value, lower) case.
+    auto &upperCase(const bool u = true) { forceCase = u; return *this; }
+
     Integer io_manip; ///< the integer to print
+
+    ///< \copydoc minDigits(). The default is to use stream's field width and
+    ///stream's fill character.
```
yadij, line 133: The "///<" indicates that this line is documentation for the previous (`io_manip`) member. Looks like a typo to me.
Also, it seems best to have `\copydoc` on its own line since that will pull in a whole set of self-contained documentation syntax.
```suggestion
    /// \copydoc minDigits()
    /// The default is to use stream's field width and stream's fill character.
```

src/base/IoManip.h
```diff
@@ -75,22 +105,61 @@ operator <<(std::ostream &os, const RawPointerT<Pointer> &pd)
...
+    std::optional<size_t> forcePadding;
+
+    ///< \copydoc upperCase(). The default is to use stream's std::uppercase flag.
```
yadij, line 136: Same as with above.
```suggestion
    /// \copydoc upperCase()
    /// The default is to use stream's std::uppercase flag.
```

src/ipcache.cc
```diff
@@ -729,7 +730,7 @@ ipcache_gethostbyname(const char *name, int flags)
 {
     ipcache_entry *i = nullptr;
     assert(name);
-    debugs(14, 3, "ipcache_gethostbyname: '" << name  << "', flags=" << std::hex << flags);
+    debugs(14, 3, "ipcache_gethostbyname: '" << name  << "', flags=" << asHex(flags));
```
yadij, line 733:
```suggestion
    debugs(14, 3, "'" << name  << "', flags=" << asHex(flags));
```

src/tests/testIoManip.cc
```diff
@@ -0,0 +1,161 @@
...
+    // padding with zeros works even for zero values
+    CPPUNIT_ASSERT_EQUAL(std::string("0000"), toStdString(asHex(0).minDigits(4)));
+
+    // minDigits() does not truncate
+    CPPUNIT_ASSERT_EQUAL(std::string("1"), toStdString(asHex(0x1).minDigits(0)));
+    CPPUNIT_ASSERT_EQUAL(std::string("12"), toStdString(asHex(0x12).minDigits(1)));
+    CPPUNIT_ASSERT_EQUAL(std::string("123"), toStdString(asHex(0x123).minDigits(2)));
+
+    // upperCase() forces uppercase
+    CPPUNIT_ASSERT_EQUAL(std::string("A"), toStdString(asHex(0xA).upperCase()));
```
yadij, line 94: Please make this a test of how the default treats lowercase as well as already-uppercase.
```suggestion
    CPPUNIT_ASSERT_EQUAL(std::string("A"), toStdString(asHex(0xA).upperCase()));
    CPPUNIT_ASSERT_EQUAL(std::string("A"), toStdString(asHex(0xa).upperCase()));
```

src/tests/testIoManip.cc
```diff
@@ -0,0 +1,161 @@
...
+    CPPUNIT_ASSERT_EQUAL(std::string("1A2B"), toStdString(asHex(0x1a2b).upperCase(true)));
+
+    std::ostringstream ss;
+
+    // upperCase(false) forces lowercase
```
yadij, line 99: Why is this test case significantly different from the earlier tests of `true` ?
Surely we do not need this special `ss` variable management.

src/tests/testIoManip.cc
```diff
@@ -0,0 +1,161 @@
...
+
+void
+TestIoManip::testAsHex()
+{
+    // basic values
+    CPPUNIT_ASSERT_EQUAL(std::string("0"), toStdString(asHex(0)));
+    CPPUNIT_ASSERT_EQUAL(std::string("123abc"), toStdString(asHex(0x123abc)));
+
+    // large values
+    CPPUNIT_ASSERT_EQUAL(std::string("7fffffffffffffff"), toStdString(asHex(std::numeric_limits<int64_t>::max())));
```
yadij, line 56: What about negative values such as `-1` or `std::numeric_limits<int64_t>::min()` ?

yadij (changes requested):

src/tests/testIoManip.cc
rousskov: Francesco is right. We should not test that very basic C++ features work as expected.

src/tests/testIoManip.cc
rousskov: > Why is this test case significantly different from the earlier tests of `true` ?

The earlier tests could use and, hence, used default std::ostream flags (including lowercase). This test (and the other ones that use `ss`) cannot use those defaults and, hence, need an explicit stream object to set the flags on.

> Surely we do not need this special `ss` variable management.

If you see a good way to remove `ss` while still testing what is being tested, please post a specific suggestion.

src/tests/testIoManip.cc
rousskov: Negative values are tested in the next blob. Look for `// negative values` title and a comment that explains why we do not support them in a meaningful way.

The same comment implies that testing negative values of unknown (e.g., `-1`) or very large (e.g., `std::numeric_limits<int64_t>::min()`) may not be such a good idea due to (implicit) casting and associated undefined behavior on over/underflows. I would not violently oppose adding those two lines/cases, but I did remove them (after adding them myself) and do not recommend adding them back. 

The primary point here is that std::hex and AsHex do not support negative hex numbers (in a meaningful way). We cannot check for negative numbers and compile time and (currently, in the official code and this PR) do not want to throw at runtime.

rousskov (approved):

src/tests/testIoManip.cc
yadij: Ah right. Sorry I missed that.

yadij (approved):

kinkie:
Manual backport to v6 in #1624

PR #1584, opened 2023-11-10
Bug 5318: peer_digest.cc:399: "fetch->pd && receivedData.data"

rousskov:
Recent commit 122a6e3 removed HTTP response headers from store_client
responses. That removal created the possibility of an empty
StoreIOBuffer at the beginning of the feeding sequence. Pending Bug 5317
fix will make such buffers even more frequent. Existing store_client
recipients have varying requirements with regard to empty response
buffers, as documented in store_client::finishCallback(). We missed this
requirement conflict in Cache Digest code. This fix adjusts Cache
Digests code to be compatible with empty StoreIOBuffer representation in
current store_client code.

### DIFF
--- a/src/peer_digest.cc
+++ b/src/peer_digest.cc
@@ -349,11 +349,11 @@ peerDigestHandleReply(void *data, StoreIOBuffer receivedData)
         return;
     }
 
-    assert(fetch->pd && receivedData.data);
+    assert(fetch->pd);
     /* The existing code assumes that the received pointer is
      * where we asked the data to be put
      */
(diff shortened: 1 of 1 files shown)


rousskov (commented):
I am posting this as a Draft PR because this code is completely untested and should not be committed without tests. I do not have the time to test this immediately and am asking bug [5318](https://bugs.squid-cache.org/show_bug.cgi?id=5318) reporters to test this PR.

yadij (commented):
6 days with only good feedback from testing I am clearing this for merge into the next release.

PR #1585, opened 2023-11-13
Bug 5319: QOS Netfilter MARK preservation is always disabled

niconorsk:
Default ./configure options and explicit --enable-zph-qos enabled ZPH
QOS support (USE_QOS_TOS) as expected but did not enable QOS Netfilter
MARK preservation support (USE_LIBNETFILTERCONNTRACK). For example,
qos_flows directive became available, but clientside_mark and
client_connection_mark ACL types were not recognized.

The missing opening bracket before AS_IF() condition injected a trailing
closing bracket into that condition, resulting in an always-false
condition for setting USE_LIBNETFILTERCONNTRACK.

Broken since 2022 commit a1c2236.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2852,7 +2852,7 @@ SQUID_DEFINE_BOOL(USE_QOS_TOS,${enable_zph_qos:=yes},
           [Enable Zero Penalty Hit QOS. When set, Squid will alter the
            TOS field of HIT responses to help policing network traffic])
 AC_MSG_NOTICE([ZPH QOS enabled: $enable_zph_qos])
-AS_IF(test "x$enable_zph_qos" = "xyes"],[
+AS_IF([test "x$enable_zph_qos" = "xyes"],[
   AC_MSG_NOTICE([QOS netfilter mark preservation enabled: $with_netfilter_conntrack])
   SQUID_DEFINE_BOOL(USE_LIBNETFILTERCONNTRACK,${with_netfilter_conntrack:=no},
     [Enable support for QOS netfilter mark preservation])


yadij (approved):

yadij:
OK to test

rousskov (approved):
@niconorsk, thank you for fixing this bug!

@yadij, I have adjusted PR title/description to fix [formatting](https://github.com/measurement-factory/anubis/blob/master/README.md#pr-title) and to detail the non-obvious effects of this bug. While working on that, I realized that the bug affects some "default" ./configure runs where `--enable-zph-qos` is _not_ given because we try to enable ZPH QOS by default. However, this is not my area of expertise, and I could miss something important. Please double check and adjust further as needed.

yadij (approved):

niconorsk:
@rousskov I'm by no means an expert on this, but I just tested a configure without the explicit `--enable-zph-qos` and it did enable mark preservation which I think is the desired behaviour.

rousskov:
> I'm by no means an expert on this, but I just tested a configure without the explicit `--enable-zph-qos` and it did enable mark preservation which I think is the desired behaviour.

Yes, Squid should enable mark preservation without explicit `--enable-zph-qos` (where possible) AFAICT.

I am not sure which code flavor you have tested:

* If you have tested this PR code, then we are on the same page -- this PR fixes Squid behavior with _and_ without an explicit  `--enable-zph-qos` (in environments where mark preservation is supported). In this case, this PR description looks correct to me.

* If you have tested broken official code (i.e. Squid code without this PR fix), and it does enable mark preservation without an explicit `--enable-zph-qos`, then my PR description is partially wrong and should be fixed!

@niconorsk, please clarify. I am postponing merging this PR to make sure we get the PR description right.

niconorsk:
@rousskov The first bullet point is what I have tested.

squidadm:
queued for backport to v6

PR #1589, opened 2023-11-17
Limit the number of allowed X-Forwarded-For hops 

p4zuu:
Squid will ignore all X-Forwarded-For elements listed after the first 64
addresses allowed by the follow_x_forwarded_for directive. A different
limit can be specified by defining a C++ SQUID_X_FORWARDED_FOR_HOP_MAX
macro, but that macro is not a supported Squid configuration interface
and may change or disappear at any time.

Squid will log a cache.log ERROR if the hop limit has been reached.

This change works around problematic ACLChecklist and/or slow ACLs
implementation that results in immediate nonBlockingCheck() callbacks.
Such callbacks have caused many bugs and development complications. In
clientFollowXForwardedForCheck() context, they lead to indirect
recursion that was bound only by the number of allowed XFF entries,
which could reach thousands and exhaust Squid process call stack.

This recursion bug was discovered and detailed by Joshua Rogers at
https://megamansec.github.io/Squid-Security-Audit/xff-stackoverflow.html
where it was filed as "X-Forwarded-For Stack Overflow".

### DIFF
--- a/src/ClientRequestContext.h
+++ b/src/ClientRequestContext.h
@@ -83,7 +83,7 @@ public:
     ErrorState *error = nullptr; ///< saved error page for centralized/delayed processing
     bool readNextRequest = false; ///< whether Squid should read after error handling
 #if FOLLOW_X_FORWARDED_FOR
-    uint32_t currentXffHopNumber = 0; ///< number of X-Forwarded-For header values processed so far
+    size_t currentXffHopNumber = 0; ///< number of X-Forwarded-For header values processed so far
 #endif
 };
 
--- a/src/ClientRequestContext.h
+++ b/src/ClientRequestContext.h
@@ -77,11 +77,13 @@ public:
     bool no_cache_done = false;
     bool interpreted_req_hdrs = false;
     bool toClientMarkingDone = false;
+    bool readNextRequest = false; ///< whether Squid should read after error handling
 #if USE_OPENSSL
     bool sslBumpCheckDone = false;
 #endif
+
     ErrorState *error = nullptr; ///< saved error page for centralized/delayed processing
-    bool readNextRequest = false; ///< whether Squid should read after error handling
+
 #if FOLLOW_X_FORWARDED_FOR
     size_t currentXffHopNumber = 0; ///< number of X-Forwarded-For header values processed so far
 #endif
--- a/src/ClientRequestContext.h
+++ b/src/ClientRequestContext.h
@@ -82,6 +82,9 @@ public:
 #endif
     ErrorState *error = nullptr; ///< saved error page for centralized/delayed processing
     bool readNextRequest = false; ///< whether Squid should read after error handling
+#if FOLLOW_X_FORWARDED_FOR
+    uint32_t currentXffHopNumber = 0; ///< number of X-Forwarded-For header values processed so far
+#endif
 };
 
 #endif /* SQUID_CLIENTREQUESTCONTEXT_H */
--- a/src/ClientRequestContext.h
+++ b/src/ClientRequestContext.h
@@ -77,11 +77,11 @@ public:
     bool no_cache_done = false;
     bool interpreted_req_hdrs = false;
     bool toClientMarkingDone = false;
-    bool readNextRequest = false; ///< whether Squid should read after error handling
 #if USE_OPENSSL
     bool sslBumpCheckDone = false;
 #endif
 
+    bool readNextRequest = false; ///< whether Squid should read after error handling
     ErrorState *error = nullptr; ///< saved error page for centralized/delayed processing
 
 #if FOLLOW_X_FORWARDED_FOR
--- a/src/client_side_request.cc
+++ b/src/client_side_request.cc
@@ -442,9 +442,11 @@ clientFollowXForwardedForCheck(Acl::Answer answer, void *data)
                 /* override the default src_addr tested if we have to go deeper than one level into XFF */
                 Filled(calloutContext->acl_checklist)->src_addr = request->indirect_client_addr;
             }
-            if (++calloutContext->currentXffHopNumber < SQUID_X_FORWARDED_FOR_HOP_MAX)
+            if (++calloutContext->currentXffHopNumber < SQUID_X_FORWARDED_FOR_HOP_MAX) {
                 calloutContext->acl_checklist->nonBlockingCheck(clientFollowXForwardedForCheck, data);
-            return;
+                return;
+            }
+            // fall through to resume clientAccessCheck() processing
         }
     }
 
--- a/src/client_side_request.cc
+++ b/src/client_side_request.cc
@@ -75,6 +75,9 @@
 #endif
 
 #if FOLLOW_X_FORWARDED_FOR
+#if !defined(MAX_X_FORWARDED_HOP)
+#define MAX_X_FORWARDED_HOP 16
+#endif
 static void clientFollowXForwardedForCheck(Acl::Answer answer, void *data);
 #endif /* FOLLOW_X_FORWARDED_FOR */
 
@@ -437,7 +440,8 @@ clientFollowXForwardedForCheck(Acl::Answer answer, void *data)
                 /* override the default src_addr tested if we have to go deeper than one level into XFF */
                 Filled(calloutContext->acl_checklist)->src_addr = request->indirect_client_addr;
             }
-            calloutContext->acl_checklist->nonBlockingCheck(clientFollowXForwardedForCheck, data);
+            if (++calloutContext->currentXffHopNumber < MAX_X_FORWARDED_HOP)
+                calloutContext->acl_checklist->nonBlockingCheck(clientFollowXForwardedForCheck, data);
             return;
         }
     }
--- a/src/client_side_request.cc
+++ b/src/client_side_request.cc
@@ -75,9 +75,11 @@
 #endif
 
 #if FOLLOW_X_FORWARDED_FOR
+
 #if !defined(SQUID_X_FORWARDED_FOR_HOP_MAX)
 #define SQUID_X_FORWARDED_FOR_HOP_MAX 16
 #endif
+
 static void clientFollowXForwardedForCheck(Acl::Answer answer, void *data);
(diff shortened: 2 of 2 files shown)


rousskov (changes requested):
Thank you for this fix. I do not think we need a limit long-term (we should avoid using stack for this iteration instead), but this PR is much better than nothing!

I will polish this implementation a bit, but I hope you can help by addressing two change requests first. If you do not have the time to do that, please let me know!

src/client_side_request.cc
```diff
@@ -437,6 +438,7 @@ clientFollowXForwardedForCheck(Acl::Answer answer, void *data)
                 /* override the default src_addr tested if we have to go deeper than one level into XFF */
                 Filled(calloutContext->acl_checklist)->src_addr = request->indirect_client_addr;
             }
+            calloutContext->currentHopNumber += 1;
             calloutContext->acl_checklist->nonBlockingCheck(clientFollowXForwardedForCheck, data);
```
yadij, line 442:
```suggestion
            if (++calloutContext->currentHopNumber < MAX_X_FORWARDED_HOP)
                calloutContext->acl_checklist->nonBlockingCheck(clientFollowXForwardedForCheck, data);
```

src/ClientRequestContext.h
```diff
@@ -82,6 +82,9 @@ class ClientRequestContext : public RefCountable
 #endif
     ErrorState *error = nullptr; ///< saved error page for centralized/delayed processing
     bool readNextRequest = false; ///< whether Squid should read after error handling
+#if FOLLOW_X_FORWARDED_FOR
+    uint32_t currentHopNumber = 0;
```
yadij, line 86: This name is a bit too generic for a very XFF-specific value.
```suggestion
    uint32_t currentXffHopNumber = 0;
```

rousskov: I was going to polish this manually later, but if we are doing this polish via change requests, then I can suggest:

```suggestion
    size_t xffAddressesChecked = 0;
```

or, if you really insist,

```suggestion
    size_t currentXffHopNumber = 0;
```

Rationale: We should not be counting bits in this context, and this data member is counting the number of address checks performed rather than some hop position or "number" (e.g., if we skip an XFF address for some reason, the counter should not go up even though we would be looking at the next address after that skip).

yadij: AIUI this is a recursion counter to limit the stack size for this old semi-async code. A full fix would store the unprocessed XFF values and step over them in async Calls.

FTR; this logic must never "skip" XFF addresses. Every value MUST meet the criteria of being a valid representation of a raw-IP accepted by the ACLs - or the whole process stops and uses the previously found IP.

rousskov: I disagree with some of the above assertions, but since they do not really contradict my code change recommendations, I will save time by ignoring those disagreements.

src/client_side_request.cc
```diff
@@ -398,7 +399,7 @@ clientFollowXForwardedForCheck(Acl::Answer answer, void *data)
     ClientHttpRequest *http = calloutContext->http;
     HttpRequest *request = http->request;
 
-    if (answer.allowed() && request->x_forwarded_for_iterator.size() != 0) {
+    if (answer.allowed() && request->x_forwarded_for_iterator.size() != 0 && calloutContext->currentHopNumber < MAX_X_FORWARDED_HOP) {
```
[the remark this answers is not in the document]
p4zuu: It should be fixed with @yadji's suggestion (https://github.com/squid-cache/squid/pull/1589#discussion_r1397877228) now applied

src/ClientRequestContext.h
p4zuu: I went for `currentXffHopNumber` since you both seemed agree

p4zuu:
> Thank you for this fix. I do not think we need a limit long-term (we should avoid using stack for this iteration instead), but this PR is much better than nothing!
> 
> I will polish this implementation a bit, but I hope you can help by addressing two change requests first. If you do not have the time to do that, please let me know!

Thanks for the suggestions! I have time to polish it :)

src/client_side_request.cc
```diff
@@ -75,6 +75,7 @@
 #endif
 
 #if FOLLOW_X_FORWARDED_FOR
+#define MAX_X_FORWARDED_HOP 16
```
yadij, line 78: If this is intended for build-time tuning this `#define` needs to be wrapped to allow builders value to overwrite this default.

```suggestion
#if !defined(MAX_X_FORWARDED_HOP)
#define MAX_X_FORWARDED_HOP 16
#endif
```

Otherwise,
  If not the limit should be set as ` static const size_t` at the point in the code just before it is used. see other comment for that.

src/client_side_request.cc
```diff
@@ -437,7 +438,9 @@ clientFollowXForwardedForCheck(Acl::Answer answer, void *data)
                 /* override the default src_addr tested if we have to go deeper than one level into XFF */
                 Filled(calloutContext->acl_checklist)->src_addr = request->indirect_client_addr;
             }
-            calloutContext->acl_checklist->nonBlockingCheck(clientFollowXForwardedForCheck, data);
+            if (++calloutContext->currentXffHopNumber < MAX_X_FORWARDED_HOP) {
+                calloutContext->acl_checklist->nonBlockingCheck(clientFollowXForwardedForCheck, data);
+            }
```
yadij, line 443: No need for `{...}` around single code lines.

```suggestion
            if (++calloutContext->currentXffHopNumber < MAX_X_FORWARDED_HOP)
                calloutContext->acl_checklist->nonBlockingCheck(clientFollowXForwardedForCheck, data);
```

OR, if the `#define MAX_X_FORWARDED_HOP` is not needed to be build-time tuned this could be:
```suggestion
            static const uint32_t maxXffHops = 16;
            if (++calloutContext->currentXffHopNumber < maxXffHops)
                calloutContext->acl_checklist->nonBlockingCheck(clientFollowXForwardedForCheck, data);
```

src/ClientRequestContext.h
```diff
@@ -82,6 +82,9 @@ class ClientRequestContext : public RefCountable
 #endif
     ErrorState *error = nullptr; ///< saved error page for centralized/delayed processing
     bool readNextRequest = false; ///< whether Squid should read after error handling
+#if FOLLOW_X_FORWARDED_FOR
+    uint32_t currentXffHopNumber = 0;
```
yadij, line 86:
```suggestion
    uint32_t currentXffHopNumber = 0; ///< number of X-Forwarded-For header values processed so far
```

yadij (approved):
Looks good to me. These are purely code polishing suggestions.

src/client_side_request.cc
p4zuu: I prefer allowing builders to change the `MAX_X_FORWARDED_HOP` value. I am not sure if the default value is high enough not to break some usages. What do you think?

src/client_side_request.cc
```diff
@@ -75,6 +75,7 @@
 #endif
 
 #if FOLLOW_X_FORWARDED_FOR
+#define MAX_X_FORWARDED_HOP 16
 static void clientFollowXForwardedForCheck(Acl::Answer answer, void *data);
```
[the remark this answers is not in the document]
p4zuu: Updated :)

rousskov (approved):
I pushed an important bug fix and a few improvements/polishing touches. Please see commit messages for details, but here are a few highlights:

* 95b25e3 fixup: Do not stall after reaching the hops limit!
* 596ae3d: Raise hops limit from 16 to 64
* 95f72d6: Inform the admin if we have to ignore XFF addresses
* d3cd0a1: fixup: Use a Squid-specific [macro] name

I fixed PR description formatting and polished/adjusted PR title/description a bit to correct minor technical glitches and detail important (IMO) aspects of this change. Please adjust further as needed.

This PR can be merged as is AFAICT, but I am not clear it for merging right now to give you a chance to review my changes.

src/client_side_request.cc
rousskov: I changed the default to 64. Please see branch commit 596ae3d for rationale.

src/ClientRequestContext.h
```diff
@@ -80,8 +80,13 @@ class ClientRequestContext : public RefCountable
 #if USE_OPENSSL
     bool sslBumpCheckDone = false;
 #endif
-    ErrorState *error = nullptr; ///< saved error page for centralized/delayed processing
+
     bool readNextRequest = false; ///< whether Squid should read after error handling
+    ErrorState *error = nullptr; ///< saved error page for centralized/delayed processing
```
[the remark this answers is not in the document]
kinkie: I'm happy to approve the PR once this trivial request is satisfied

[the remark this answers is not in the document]
rousskov: > I'm happy to approve the PR once this trivial request is satisfied

Francesco, this comment was not requesting a PR change but explaining an existing (and seemingly out of scope) change. In other words, the data member move discussed in my comment has been implemented already (branch commit eddb0bfe08e896c3831f647ea8c427c228328212). I left that "I do not insist on this change" GitHub comment just in case others want to _undo_ that change.

Sorry for the confusion. In the future, I will try to remember to explicitly mark similar comments as comments that explain existing changes rather than requesting future changes.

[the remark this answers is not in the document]
kinkie: > Sorry for the confusion

Nothing to be sorry about, I should have read the code more carefully

rousskov:
OK to test

kinkie (approved):

rousskov:
> feature label to indicate need for documentation of new "-DSQUID_X_FORWARDED_FOR_HOP_MAX=" build option for Squid-6.6.

Please do _not_ include that unsupported build option in Squid documentation. The PR description (i.e. the future commit message) already says that (problematic) interface is unsupported. That macro is a hack. We hope that nobody needs to adjust it. If somebody needs to adjust it, we made that easy for them, but we do not want to encourage folks to investigate (whether they need) adjustments and are not responsible for their outcome.

PR #1590, opened 2023-11-17
Remove deprecated string memory pools API

yadij:
Also added 32-byte, 64-byte, 128-byte, 256-byte, 512-byte, and 1024-byte
memory pools.

### DIFF
--- a/src/mem/forward.h
+++ b/src/mem/forward.h
@@ -73,7 +73,6 @@ void memFreeBuf(size_t size, void *);
 void memFreeRigid(void *, size_t net_size);
 FREE *memFreeBufFunc(size_t size);
 int memInUse(mem_type);
-void memDataInit(mem_type, const char *, size_t, int, bool doZero = true);
 void memCheckInit(void);
 size_t memStringCount();
 
--- a/src/mem/forward.h
+++ b/src/mem/forward.h
@@ -39,6 +39,12 @@ typedef void FREE(void *);
 /// Types of memory pool which do not yet use MEMPROXY_CLASS() API
 typedef enum {
     MEM_NONE,
+    MEM_32B_BUF,
+    MEM_64B_BUF,
+    MEM_128B_BUF,
+    MEM_256B_BUF,
+    MEM_512B_BUF,
+    MEM_1K_BUF,
     MEM_2K_BUF,
     MEM_4K_BUF,
     MEM_8K_BUF,
--- a/src/mem/forward.h
+++ b/src/mem/forward.h
@@ -73,7 +73,6 @@ void memFreeBuf(size_t size, void *);
 void memFreeRigid(void *, size_t net_size);
 FREE *memFreeBufFunc(size_t size);
 int memInUse(mem_type);
-void memDataInit(mem_type, const char *, size_t, int, bool doZero = true);
 size_t memStringCount();
 
 #endif /* _SQUID_SRC_MEM_FORWARD_H */
--- a/src/mem/forward.h
+++ b/src/mem/forward.h
@@ -64,13 +64,11 @@ void memCleanModule(void);
 void memConfigure(void);
 /// Allocate one element from the typed pool
 void *memAllocate(mem_type);
-void *memAllocString(size_t net_size, size_t * gross_size);
 void *memAllocBuf(size_t net_size, size_t * gross_size);
 void *memAllocRigid(size_t net_size);
 void *memReallocBuf(void *buf, size_t net_size, size_t * gross_size);
 /// Free a element allocated by memAllocate()
 void memFree(void *, int type);
-void memFreeString(size_t size, void *);
 void memFreeBuf(size_t size, void *);
 void memFreeRigid(void *, size_t net_size);
 FREE *memFreeBufFunc(size_t size);
--- a/src/mem/old_api.cc
+++ b/src/mem/old_api.cc
@@ -430,7 +430,7 @@ Mem::Init(void)
     memDataInit(MEM_16K_BUF, "16K Buffer", 16384, 10, false);
     memDataInit(MEM_32K_BUF, "32K Buffer", 32768, 10, false);
     memDataInit(MEM_64K_BUF, "64K Buffer", 65536, 10, false);
-    // TODO: make these objects use MEMPROXY_CLASS() and drop the doZero parameter
+    // TODO: Carefully stop zeroing these objects memory and drop the doZero parameter
     memDataInit(MEM_DREAD_CTRL, "dread_ctrl", sizeof(dread_ctrl), 0, true);
     memDataInit(MEM_DWRITE_Q, "dwrite_q", sizeof(dwrite_q), 0, true);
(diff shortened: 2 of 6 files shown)


src/mem/old_api.cc
```diff
@@ -439,6 +428,12 @@ Mem::Init(void)
      * that are never used or used only once; perhaps we should simply use
```
rousskov, line 428:
```diff
-      * Starting with generic 2kB - 64kB buffr pools, then specific object types.
+      * Starting with generic buffer pools, then specific object types.
```

This PR makes comment a few lines above stale.

rousskov (dismissed):
Thank you.

src/mem/old_api.cc
```diff
@@ -310,7 +281,25 @@ memFindBufSizeType(size_t net_size, size_t * gross_size)
...
+    } else if (net_size <= 256) {
+        type = MEM_256B_BUF;
+        size = 256;
+    } else if (net_size <= 512) {
+        type = MEM_512B_BUF;
+        size = 512;
+    } else if (net_size <= 1024) {
+        type = MEM_1K_BUF;
+        size = 1024;
+    } else if (net_size <= 2 * 1024) {
```
rousskov, line 302: I added "Also added a few new pools for 2KB and smaller buffers." to the PR description because these added pools did not exist before from memAllocBuf() callers point of view and because set of added pools is not equivalent to the string pools in the code below (that were used by now-converted memAllocString() callers, among others):

```C++
static const PoolMeta PoolAttrs[mem_str_pool_count] = {
    {"Short Strings", Mem::Allocator::RoundedSize(36)}, /* to fit rfc1123 and similar */
    {"Medium Strings", Mem::Allocator::RoundedSize(128)}, /* to fit most urls */
    {"Long Strings", Mem::Allocator::RoundedSize(512)},
    {"1KB Strings", Mem::Allocator::RoundedSize(1024)},
```

We are changing a lot more here than the API, and we should document that.

Please adjust PR description further as needed, of course.

src/mem/old_api.cc
yadij: Done in [0738130](https://github.com/squid-cache/squid/pull/1590/commits/07381308af2c283c803542334c4e6bdec64cb32c)

yadij:
@rousskov how important is the `xmalloc()` vs `xcalloc()` difference for `memAllocRigid()` ?
  if we can skip the TODO about zeroing the buffer, a lot more of the old string pools logic becomes dead code.

rousskov:
> how important is the `xmalloc()` vs `xcalloc()` difference for `memAllocRigid()` ? if we can skip the TODO about zeroing the buffer, a lot more of the old string pools logic becomes dead code.

I do not know the answer to your question, but I will try to find out under the assumption that you are willing to remove more old code in this PR if the answer is something like "unimportant". Needless to say, the more crap we can safely remove, the better!

yadij:
@rousskov FWIW, with the other removals done in [01e18dd](https://github.com/squid-cache/squid/pull/1590/commits/01e18dd74c14d1b8ae04a938238e527012562719) it has become clear(er) that the xcalloc() was only being used for the very-large random-size buffers. The "standard" buffer pools are not being zeroed.

src/mem/old_api.cc
```diff
@@ -432,22 +411,29 @@ Mem::Init(void)
...
     memDataInit(MEM_2K_BUF, "2K Buffer", 2048, 10, false);
     memDataInit(MEM_4K_BUF, "4K Buffer", 4096, 10, false);
     memDataInit(MEM_8K_BUF, "8K Buffer", 8192, 10, false);
     memDataInit(MEM_16K_BUF, "16K Buffer", 16384, 10, false);
     memDataInit(MEM_32K_BUF, "32K Buffer", 32768, 10, false);
     memDataInit(MEM_64K_BUF, "64K Buffer", 65536, 10, false);
-    memDataInit(MEM_DREAD_CTRL, "dread_ctrl", sizeof(dread_ctrl), 0);
-    memDataInit(MEM_DWRITE_Q, "dwrite_q", sizeof(dwrite_q), 0);
-    memDataInit(MEM_MD5_DIGEST, "MD5 digest", SQUID_MD5_DIGEST_LENGTH, 0);
+    // TODO: make these objects use MEMPROXY_CLASS() and drop the doZero parameter
```
rousskov, line 433: I disagree with the suggested MEMPROXY_CLASS() solution (because, in part, MD5_DIGEST-using objects should not become MEMPROXY_CLASS), but I agree with the _essential_ parts of this TODO. Here is a variant that may avoid disagreements:

```suggestion
    // TODO: Carefully stop zeroing these objects memory and drop the doZero parameter
```

I do _not_ insist on the comment using the word "Carefully". FWIW, before suggesting that word, I have looked at how these buffers are used right now. Here are my draft notes:

MEM_DREAD_CTRL zeroing is a waste because the corresponding dread_ctrl object fields are fully initialized upon allocation and uninitialized padding is irrelevant in this context. The type should be upgraded to a pooled C++ class. I recommend doing that in a dedicated PR.

MEM_DWRITE_Q zeroing may be essential in the current (arguably buggy) code because one of the dwrite_q users does _not_ initialize the dwrite_q::file_offset field -- there is a corresponding XXX in diskCombineWrites(). The type should be upgraded to a pooled C++ class, but resolving that XXX is not trivial and deserves a dedicated PR.

MEM_MD5_DIGEST zeroing is a waste because the buffer is immediately filled with SQUID_MD5_DIGEST_LENGTH bytes. Most likely, MEM_MD5_DIGEST-based cache_key type should be changed from a buffer to two uint64_t integers, but a clean upgrade requires a lot of work and warrants a dedicated PR.

rousskov (changes requested):

src/mem/old_api.cc
```diff
@@ -224,32 +222,12 @@ memFree(void *p, int type)
...
-    ++StrCountMeter;
-    StrVolumeMeter += *gross_size;
-    return xcalloc(1, net_size);
-}
-
 void *
 memAllocRigid(size_t net_size)
 {
-    // TODO: Use memAllocString() instead (after it stops zeroing memory).
+    // TODO: Use memAllocBuf() instead (after it stops zeroing memory).
```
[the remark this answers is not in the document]
yadij: Thank you. I will followup this PR with full removal of the `GetStrPool()` and related code.

rousskov (dismissed):

src/mem/forward.h
```diff
@@ -73,7 +73,7 @@ void memFreeBuf(size_t size, void *);
 void memFreeRigid(void *, size_t net_size);
 FREE *memFreeBufFunc(size_t size);
 int memInUse(mem_type);
-void memCheckInit(void);
+void memDataInit(mem_type, const char *, size_t, int, bool doZero = true);
```
rousskov, line 76: Please remove this static function declaration from forward.h and src/tests/stub_libmem.cc files (at least). AFAICT, those no-longer-wanted lines were resurrected in [b1bc5f7](https://github.com/squid-cache/squid/pull/1590/commits/b1bc5f74878a177a168d89300533574569cee693).

kinkie (approved):
@yadij , is there any reason to delay merging this?

PR #1591, opened 2023-11-18
Move memory initialization check

yadij:
mainInitialize() is far too late to be validating that memory
pools created with memDataInit() have been correctly initialized.
They will have already been used by debugging, and many other
components early-setup logic.

Instead run the check at the last possible moment before
declaring the memory pools to be fully initialized.


rousskov (approved):
Thank you.

PR #1592, opened 2023-11-20
Stop zeroing huge memAllocBuf() buffers

yadij:
memAllocBuf() buffers smaller than 64KB were not zeroed before this
change. Larger buffers (a.k.a. "huge buffers") were zeroed with
xcalloc(). We believe it is safe to stop zeroing those huge buffers
because all known code that allocates huge buffers also allocates
smaller ones for the same purpose. If some of that code relied on
zeroing for years, we would expect to see problems with smaller buffers.

Removing xcalloc() allows removal of mem*Rigid(), memBufStats() and all
string pools API complications.

The cache manager mgr:mem report section for string
statistics is also removed.

### DIFF
--- a/src/mem/old_api.cc
+++ b/src/mem/old_api.cc
@@ -197,7 +197,7 @@ memAllocBuf(size_t net_size, size_t * gross_size)
         return memAllocate(type);
     else {
         ++HugeBufCountMeter;
-        HugeBufVolumeMeter += *gross_size;
+        HugeBufVolumeMeter += net_size;
         return xmalloc(net_size);
     }
 }
--- a/src/mem/old_api.cc
+++ b/src/mem/old_api.cc
@@ -40,23 +40,14 @@ static void memFree16K(void *);
 static void memFree32K(void *);
 static void memFree64K(void *);
 
-/* local prototypes */
-static void memStringStats(std::ostream &);
-
 /* module locals */
 static double xm_time = 0;
 static double xm_deltat = 0;
-
-/* string pools */
-#define mem_str_pool_count 6
 
 struct PoolMeta {
     const char *name;
     size_t obj_size;
 };
-
-static Mem::Meter StrCountMeter;
-static Mem::Meter StrVolumeMeter;
 
 static Mem::Meter HugeBufCountMeter;
 static Mem::Meter HugeBufVolumeMeter;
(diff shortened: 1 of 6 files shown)


yadij:
Changes built on PR 1590 and just waiting on that PR to merge first.

src/tests/stub_libmem.cc
```diff
@@ -39,18 +39,12 @@ void *memAllocate(mem_type)
...
-}
-
 void *
 memAllocBuf(size_t net_size, size_t * gross_size)
 {
-    *gross_size=net_size;
-    return xcalloc(1, net_size);
+    if (gross_size)
+        *gross_size = net_size;
+    return xmalloc(net_size);
```
rousskov, line 47: Why is it safe to stop zeroing this memory? Please adjust PR description to answer that question.

rousskov (changes requested):

src/tests/stub_libmem.cc
yadij: done

rousskov (changes requested):
If you are OK with my PR title/description adjustments, then we are almost done here! Please ping me after merging master with prerequisite PR changes into this PR branch.

src/mem/old_api.cc
```diff
@@ -350,7 +198,7 @@ memAllocBuf(size_t net_size, size_t * gross_size)
     else {
         ++HugeBufCountMeter;
         HugeBufVolumeMeter += *gross_size;
-        return xcalloc(1, net_size);
```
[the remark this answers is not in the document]
yadij: On inspection this should not be using `memReallocBuf()` at all. It is the first allocation of the buffer. Switching to that and clearing it with `memset()` in the upcoming rebase.

src/tests/stub_libmem.cc
```diff
@@ -39,16 +39,12 @@ void *memAllocate(mem_type)
...
-{
-    return xmalloc(net_size);
-}
-
 void *
 memAllocBuf(size_t net_size, size_t * gross_size)
 {
-    *gross_size=net_size;
-    return xcalloc(1, net_size);
+    if (gross_size)
```
rousskov, line 45: If this PR effectively makes the undocumented memAllocBuf() gross_size address _optional_, it should make the corresponding change in memReallocBuf() as well, to keep the two sister functions in sync: All buffers allocated using memAllocBuf() ought to remain compatible with memReallocBuf().

Please do not forget to update _all_ memReallocBuf() implementations, including the one in src/mem/minimal.cc.

src/mem/old_api.cc
```diff
@@ -329,7 +198,7 @@ memAllocBuf(size_t net_size, size_t * gross_size)
     else {
         ++HugeBufCountMeter;
         HugeBufVolumeMeter += *gross_size;
```
rousskov, line 200: This PR makes gross_size parameter optional in src/mem/minimal.cc and src/tests/stub_libmem.cc, but the parameter is still required in this (primary) memAllocBuf() implementation. This PR also adds a memAllocBuf() caller that supplies nil gross_size and can crash this code.

Please adjust to keep all memAllocBuf() implementations in sync. For example, we can make the parameter optional in all memAllocBuf() implementations, including this one:

```suggestion
        HugeBufVolumeMeter += net_size;
```

rousskov (changes requested):

src/tests/stub_libmem.cc
yadij: It is **effectively** optional now, the callers have not changed their expectations nor requirements. Whenever it can be provided it should be provided.

The `memReallocBuf` suffers from the design flaw that looking up which pool is responsible for deallocating `oldbuf` requires knowing `oldbuf` size. That is passed using `gross_size` - so this parameter is **not** optional for `memReallocBuf()`.

Dropping `memReallocBuf()` entirely is in the planning for a followup PR as soon as these larger removals get out of the way.

rousskov: > `gross_size` is **not** optional for `memReallocBuf()`.

Agreed. Sorry I missed that fact when writing my change request.

> Dropping `memReallocBuf()` entirely is in the planning for a followup PR

I am resolving my change request under the assumption that you will promptly post a followup PR that removes memReallocBuf() and, hence, address the problem this change request was dedicated to -- a new dangerous inconsistency between memAllocBuf() and memReallocBuf().

rousskov (approved):

kinkie (approved):
ship it!

PR #1593, opened 2023-11-20
Make static analysis tools happier about peerDigestLookup()

JetStreamSinner:
Some tools thought that `key` could be nil. Refactor a bit to convince
them otherwise.

### DIFF
--- a/src/neighbors.cc
+++ b/src/neighbors.cc
@@ -732,7 +732,7 @@ peerDigestLookup(CachePeer * p, PeerSelector * ps)
     assert(p->digest->cd);
     /* does digest predict a hit? */
 
-    if (!p->digest->cd->contains(key))
+    if (key && !p->digest->cd->contains(key))
         return LOOKUP_MISS;
 
     debugs(15, 5, "HIT for cache_peer " << *p);
--- a/src/neighbors.cc
+++ b/src/neighbors.cc
@@ -707,9 +707,9 @@ peerDigestLookup(CachePeer * p, PeerSelector * ps)
 #if USE_CACHE_DIGESTS
     assert(ps);
     HttpRequest *request = ps->request;
-    const cache_key *key = request ? storeKeyPublicByRequest(request) : nullptr;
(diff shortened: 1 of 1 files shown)


src/neighbors.cc
```diff
@@ -732,7 +732,7 @@ peerDigestLookup(CachePeer * p, PeerSelector * ps)
     assert(p->digest->cd);
     /* does digest predict a hit? */
 
-    if (!p->digest->cd->contains(key))
+    if (key && !p->digest->cd->contains(key))
```
rousskov, line 735: > Found by static analysis tools

Can those tools prove that key may be nil here? If not, then I hope that humans can and, hence, should, assert that key is never nil in this context. If you agree, please refactor the code in the beginning of peerDigestLookup() accordingly:

```C++
    assert(ps);
    HttpRequest *request = ps->request;
    assert(request);

    assert(p);
```

And then just do this here (instead of dangerously storing ephemeral storeKeyPublicByRequest() result as if for reuse):

```suggestion
    if (!p->digest->cd->contains(storeKeyPublicByRequest(request)))
```

We could `assert(key)` above, but I hope that static analysis will discover that key cannot be nil in this code variation without that misleading noise.

We should document that storeKeyPublicByRequest() never returns nil, but it may be better to spend that energy on changing cache_key type to a non-pointer instead.

rousskov (changes requested):
Thank you for working on this poorly written code and making static analysis tools happier. I hope there is a better solution for this problem.

src/neighbors.cc
JetStreamSinner: Got it. Sorry, for shallow analysis, will be more attentive next time. 
I will take into account your suggestions and add required changes

rousskov: No problem at all! Looking forward to your changes.

rousskov (approved):

rousskov:
OK to test

yadij (approved):

PR #1594, opened 2023-11-20
Remove global AclMatchedName from ACL::match() virtual methods

eduard-bagdasaryan:
... and use ACL::name instead, which became accessible since 8319d47.


yadij (approved):
thank you

rousskov (approved):

PR #1599, opened 2023-11-22
Add "virtual" specifier to class destructor

JetStreamSinner:
Ftp::CtrlChannel allocating dynamic memory in constructor by call:
buf = static_cast<char*>(memAllocBuf(4096, &size));
but since its destructor not virtual only desctructor of base class
called and memory not freed

### DIFF
--- a/src/clients/FtpClient.h
+++ b/src/clients/FtpClient.h
@@ -74,7 +74,7 @@ class CtrlChannel: public Ftp::Channel
 {
 public:
     CtrlChannel();
-    ~CtrlChannel();
+    virtual ~CtrlChannel();
 
     char *buf;
     size_t size;
--- a/src/clients/FtpClient.h
+++ b/src/clients/FtpClient.h
@@ -63,6 +63,8 @@ public:
      * the callback will still happen and needs to be handled (usually dropped).
      */
     Comm::ConnectionPointer listenConn;
+    
+    virtual ~Channel() = default;
 
 private:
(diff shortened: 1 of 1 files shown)


rousskov (changes requested):
> since [Ftp::CtrlChannel] destructor not virtual only desctructor of base class
called and memory not freed

AFAICT, both CtrlChannel and Channel destructors are called when (and only when) a Client object is destroyed:

```C++
class Client ...
{
    ...
    CtrlChannel ctrl; ///< FTP control channel state
    DataChannel data; ///< FTP data channel state
};
```

Channel objects are never `delete`d using a Channel pointer. The compiler always knows the exact Channel-based object type -- and, hence, the compiler always calls the right Channel destructor (e.g., CtrlChannel) when a Channel-based object is being destroyed.

N.B. Channel has no virtual methods because it does not need any.

The only related code improvement I can think of in this context is to [prohibit](https://godbolt.org/z/cdf51njbP) `delete` operation for Channel objects (on the grounds that Channel is a base class without a virtual destructor). I am not requesting this change, but feel free to refactor this PR accordingly.

AFAICT, all other operations on channel objects are safe or cannot be made safe.

JetStreamSinner:
Well, it seems that ```Channel``` and its child classes do not intended to use outside Client class and always stack based, so I think its better to leave it as it without changes promoted by me. Thank you for explanation, I going to close this pr

PR #1602, opened 2023-11-27
v5 Next Backports

squidadm:


kinkie (approved):

yadij:
Accepting as build errors only occur on OS which have already updated to ship v6

PR #1604, opened 2023-11-27
Avoid UB when copying AnyP::Uri

jtstrs:
Without a self-assignment check, a memcpy(3) call in Uri copy assignment
operator could result in undefined behavior. Let the compiler generate
all this code instead.

### DIFF
--- a/src/anyp/Uri.h
+++ b/src/anyp/Uri.h
@@ -34,22 +34,12 @@ class Uri
 public:
     Uri(): hostIsNumeric_(false) { *host_ = 0; }
     Uri(AnyP::UriScheme const &aScheme);
-    Uri(const Uri &other) {
-        scheme_ = other.scheme_;
-        userInfo_ = other.userInfo_;
-        memcpy(host_, other.host_, sizeof(host_));
-        hostIsNumeric_ = other.hostIsNumeric_;
-        hostAddr_ = other.hostAddr_;
-        port_ = other.port_;
-        path_ = other.path_;
-        touch();
-    }
 
-    Uri &operator=(const Uri &other) {
-        Uri tmp(other);
-        swap(tmp);
-        return *this;
-    }
+    Uri(const Uri &other) = default;
+    Uri &operator=(const Uri &other) = default;
+
+    Uri(Uri &&other) = default;
+    Uri &operator=(Uri &&other) = default;
 
     void clear() {
         scheme_=AnyP::PROTO_NONE;
@@ -58,17 +48,6 @@ public:
         hostAddr_.setEmpty();
         port_ = std::nullopt;
         touch();
-    }
-
-    void swap(Uri& other)
-    {
-        std::swap(scheme_, other.scheme_);
-        std::swap(userInfo_, other.userInfo_);
-        memcpy(host_, other.host_, sizeof(host_));
-        std::swap(hostIsNumeric_, other.hostIsNumeric_);
-        std::swap(hostAddr_, other.hostAddr_);
-        std::swap(port_, other.port_);
-        std::swap(path_, other.path_);
     }
 
     void touch(); ///< clear the cached URI display forms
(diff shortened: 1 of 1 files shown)


src/anyp/Uri.h
```diff
@@ -57,6 +59,18 @@ class Uri
         port_ = std::nullopt;
         touch();
     }
+
+    void swap(Uri& other)
+    {
+        std::swap(scheme_, other.scheme_);
+        std::swap(userInfo_, other.userInfo_);
+        memcpy(host_, other.host_, sizeof(host_));
```
kinkie, line 67: is host_ not swapped?

src/anyp/Uri.h
```diff
@@ -35,17 +35,19 @@ class Uri
...
-        scheme_ = o.scheme_;
-        userInfo_ = o.userInfo_;
-        memcpy(host_, o.host_, sizeof(host_));
-        hostIsNumeric_ = o.hostIsNumeric_;
-        hostAddr_ = o.hostAddr_;
-        port_ = o.port_;
-        path_ = o.path_;
+        scheme_ = other.scheme_;
+        userInfo_ = other.userInfo_;
+        memcpy(host_, other.host_, sizeof(host_));
```
rousskov, line 40: Would it be accurate to say that the official assignment operator implementation is correct except that this memcpy() should be memmove()?

N.B. This comment does not request any changes. I am just trying to understand the true scope of the problem.

rousskov (changes requested):
If possible, please do something like this _instead_:

```C++
Uri(const Uri &) = default;
Uri &operator=(const Uri &) = default;
```

If default implementations break something, please let me know what they break.

P.S. If the above works, we should also check whether adding default move methods is a good idea. I have not checked that.

src/anyp/Uri.h
JetStreamSinner: Yes, using ```memcpy``` with overlapping sequences is UB, but I think that
```Obj a = a;```
is pretty rare things although that's valid and compilable code.
I think that usage of ```memmove``` is good idea for handling this use case

src/anyp/Uri.h
rousskov: I am interpreting your answer as "yes, memcpy() is the only bug here".

> AFAIK adding the missing if-statement for self-assign check will optimize away...

As for an optimization opportunity, when self-assignments are very rare, adding _unnecessary_ `if` statements is the _opposite_ of optimization: In such contexts, it is overall better to (safely) copy in all cases, then to pay the price of avoiding unnecessary (but exceptionally rare) copies. In most contexts, self-assignments are very rare.

src/anyp/Uri.h
rousskov: Please focus on the `= default` solution first. Using memmove() is unnecessary if that `= default` solution works.

JetStreamSinner: Sure

jtstrs:
> If possible, please do something like this _instead_:
> 
> ```c++
> Uri(const Uri &) = default;
> Uri &operator=(const Uri &) = default;
> ```
> 
> If default implementations break something, please let me know what they break.
> 
> P.S. If the above works, we should also check whether adding default move methods is a good idea. I have not checked that.

Done. It's seems ok, but I wondering about ```touch``` method, can absent of this method lead to side effects,
What do you think?

rousskov (approved):
Thank you for fixing this bug. I am clearing this PR for merging after polishing code formatting and updating PR title/description.

> can [absence of a touch() call] lead to side effects ... ?

The touch() method is a Uri cache clearing method. That cache is internal to each Uri object. AFAICT, we are now correctly copying that cache instead of clearing it. I see that change as a problem-free improvement, but please correct me if I missed some caveats.

P.S. AFAICT, AnyP::Uri cache can be safely copied. If this cache cannot be safely copied for some reason, then it should be re-implemented to make copies impossible (i.e. so that attempts to copy it do not compile).

rousskov:
OK to test

yadij (approved):

PR #1606, opened 2023-11-29
Remove unnecessary checks for nil Store::Controller::swapDir

jtstrs:
... and renamed that data member. No functionality changes expected.

### DIFF
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -203,7 +203,7 @@ Store::Controller::configure()
 
     // TODO: move this into a memory cache class when we have one
     const int64_t memMax = static_cast<int64_t>(min(Config.Store.maxInMemObjSize, Config.memMaxSize));
-    const int64_t disksMax = swapDir ? swapDir->maxObjectSize() : 0;
+    const int64_t disksMax = swapDir->maxObjectSize();
     store_maxobjsize = std::max(disksMax, memMax);
 }
 
--- a/src/store/Controller.cc
+++ b/src/store/Controller.cc
@@ -33,7 +33,7 @@
 int Store::Controller::store_dirs_rebuilding = 1;
 
 Store::Controller::Controller() :
-    swapDir(new Disks),
+    disks(new Disks),
     sharedMemStore(nullptr),
     localMemStore(false),
     transients(nullptr)
@@ -45,7 +45,7 @@ Store::Controller::~Controller()
 {
     delete sharedMemStore;
     delete transients;
-    delete swapDir;
+    delete disks;
 
     if (store_table) {
         hashFreeItems(store_table, destroyStoreEntry);
@@ -66,7 +66,7 @@ Store::Controller::init()
         }
     }
 
-    swapDir->init();
+    disks->init();
 
     if (Transients::Enabled() && IamWorkerProcess()) {
         transients = new Transients;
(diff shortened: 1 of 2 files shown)


kinkie:
I agree this test doesn't seem to make any sense here

kinkie:
Ok to test

src/store/Controller.cc
```diff
@@ -203,7 +203,7 @@ Store::Controller::configure()
 
     // TODO: move this into a memory cache class when we have one
     const int64_t memMax = static_cast<int64_t>(min(Config.Store.maxInMemObjSize, Config.memMaxSize));
-    const int64_t disksMax = swapDir ? swapDir->maxObjectSize() : 0;
+    const int64_t disksMax = swapDir->maxObjectSize();
```
rousskov, line 206: > Removed unneccessary check for `swapDir` function in `Controller` since its assuming that `swapDir` will be valid at the begining of configure function.

This removal is not wrong, but since there is similar Controller code, it would be much better to remove _all_ such "unnecessary Controller::swapDir checks". Please do that in this PR. Each PR has a cost...

> Btw, does it make sense to check `swapDir` ptr at all?

On a low level, it does not, but the overall situation is a bit more complicated. Disk caching is an optional feature. Currently, Controller code is stuck between two correct designs:

1. Lack of a disk cache is best handled by a Disks object. The swapDir member is never nil.
2. Lack of a disk cache is best handled by Controller. The swapDir member may be nil. Disks class assumes that there is at least one cache_dir configured.

I prefer the first option because it should result in simpler, safer code. The current Controller code leans heavily towards that alternative. Addressing my change request will complete Controller migration to that design.

BTW, please use [PR description](https://github.com/measurement-factory/anubis/blob/master/README.md#pr-description) for the future commit message body. Use regular PR comments to ask questions. This approach both reduces the amount of manual work necessary to merge your PR and preserves the original discussion when/if PR description is edited.

P.S. The `swapDir` name itself is a bug. The correct name should have been something like `disks` or `cacheDirs`. Fixing that naming bug may not be worth the noise that fix will create and can be viewed as outside _this_ PR scope IMO.

rousskov (changes requested):

src/store/Controller.cc
jtstrs: Got it. Thank you, will take into account all your comments

src/store/Controller.h
```diff
@@ -149,7 +149,7 @@ class Controller: public Storage
     void checkTransients(const StoreEntry &) const;
     void checkFoundCandidate(const StoreEntry &) const;
 
-    Disks *swapDir; ///< summary view of all disk caches
+    Disks *disks; ///< summary view of all disk caches (including none); never nil
```
rousskov, line 152: > Alex: The `swapDir` name itself is a bug. The correct name should have been something like `disks` or `cacheDirs`. Fixing that naming bug may not be worth the noise that fix will create and can be viewed as outside this PR scope IMO.

FWIW, I am still not sure the renaming noise is worth its benefits, but I am clearing this PR for merging. Thank you for removing all these unnecessary and misleading checks!

I have polished PR-modified code and synced PR title/description with the latest changes.

rousskov (approved):

kinkie (approved):

PR #1608, opened 2023-11-30
Bug 5274: Successful tunnels logged as TCP_TUNNEL/500

rousskov:
Stop calling retryOrBail() when the tunneled Squid-server connection
(that we have committed to use) closes. Our retryOrBail() is dedicated
to handling errors. Most[^1] serverClosed() calls are _not_ related to
errors because our tunneling code abuses asynchronous connection closure
callbacks for TunnelStateData work termination. Depending on the
transaction details (e.g., TLS interception vs. true CONNECT), calling
retryOrBail() on these no-error code paths may result in retryOrBail()
"catch all other errors" code creating bogus ERR_CANNOT_FORWARD errors.

Most tunneling errors are already detailed, and retryOrBail() does not
have enough information to correctly detail the remaining ones anyway.
Removing this retryOrBail() call selects the arguably lesser evil.

The client-Squid connection closure callback, clientClosed(), already
uses the same logic.

This change does not resurrect Bug 5132 fixed by commit 752fa20 that
added the now-replaced retryOrBail() call to serverClosed(). That commit
fixed the leak by calling deleteThis() (via retryOrBail()). Our
finishWritingAndDelete() call preserves that logic. That commit also
claimed to allow more retries, but that claim was a mistake: To-server
closure callback registration (e.g. commitToServer()) bans retries.

[^1]: The fact that severClosed() is called for both successful and
problematic outcomes prevents TunnelStateData from properly handling
certain (rare) errors. We tried to fix that as well, but the changes
quickly snowballed, so we left a few XXXs instead.

### DIFF
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -260,6 +260,7 @@ private:
     /// resumes operations after the (possibly failed) HTTP CONNECT exchange
     void tunnelEstablishmentDone(Http::TunnelerAnswer &answer);
 
+    void finishWritingAndDelete(Connection &);
     void deleteThis();
 
     void cancelStep(const char *reason);
@@ -309,7 +310,9 @@ TunnelStateData::serverClosed()
 {
     server.noteClosure();
 
-    retryOrBail(__FUNCTION__);
+    request->hier.stopPeerClock(false);
+
+    finishWritingAndDelete(client);
 }
 
 /// TunnelStateData::clientClosed() wrapper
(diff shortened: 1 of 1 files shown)


src/tunnel.cc
```diff
@@ -324,12 +327,48 @@ void
...
+    // closure callback should be dedicated to handling rare connection closures
+    // originated _outside_ of TunnelStateData (e.g., during shutdown). In all
+    // other cases, our own close()-calling code must detail the
+    // closure-triggering error (if any) _and_ clear all callbacks: Our code
+    // does not need to be (asynchronously) notified of the closure that it
+    // itself has initiated! Until that (significant) refactoring,
+    // serverClosed() and clientClosed() callbacks will continue to mishandle
+    // those rare closures as regular ones, and access.log records will continue
+    // to lack some tunneling error indicators/details.
+    //
```
rousskov, line 368: I did try to address both XXXs above. This PR represents a third attempt to fix this bug. The two previous attempts snowballed too much. I abandoned the first attempt after realizing that fixing tunnel reading code would require switching internal buffers to SBuf, triggering lots of complex and out-of-scope changes. The second attempt was a lot more modest -- just fixing close() callers -- but that snowballed as well because there are many of them and some are rather problematic. There are just too many problem in this tunneling code... We will fix them, but doing that using minimal PRs requires attacking a few low-level problems first.

Even this small PR that only changes serverClosed() to call the last portion of the old retryOrBail() code path has to select the lesser of the two evils: After these changes, common successful outcomes should be fixed, but shutdown and similar rare "external" premature Squid-server connection closures may not be detailed at all (as opposed to being detailed incorrectly).

This comment does not request any changes.

src/tunnel.cc
```diff
@@ -324,12 +327,48 @@ void
 TunnelStateData::clientClosed()
 {
     client.noteClosure();
+    finishWritingAndDelete(server);
```
rousskov, line 330: More debugging aside, this PR does not change how clientClosed() works. We just moved the old code into finishWritingAndDelete() and rearranged it a bit to add debugging and XXXs.

This comment does not request any changes.

rousskov (commented):
> PR description: ... commit https://github.com/squid-cache/squid/commit/752fa2083698a88533ef23d88490f43f153f7a4c ... also claimed to allow more retries, but that claim was a mistake: To-server closure callback registration (e.g. commitToServer()) bans retries.

@eduard-bagdasaryan has [spotted](https://github.com/squid-cache/squid/pull/957#discussion_r780836591) that false claim much earlier than I did, but the corresponding PR had already been merged (and I probably did not even realize that Eduard's comment was invalidating a PR description claim).

inventvenkat:
Hi Maintainers, When can I expect an official backport for this fix?

kinkie:
> Hi Maintainers, When can I expect an official backport for this fix?

let's see how clean the backport is

rousskov:
> > Hi Maintainers, When can I expect an official backport for this fix?
> 
> let's see how clean the backport is

Clarification: What Francesco means is "let's see whether cherry-picking this PR commit into v6 succeeds and the result passes built-in tests". It probably does -- you can track #1641 progress that attempts to merge these changes automatically. If it fails, one can certainly backport these changes manually, but that (obviously) requires more work.

squidadm:
queued for backport to v6

rousskov:
> Is there a work around for this issue for squid v6.6?

Not that I know of (unless you accept an untested source code [patch](https://github.com/squid-cache/squid/commit/debf3f17be7761ea4992864a828f42ee773dfbaf.patch) as a workaround).

PR #1611, opened 2023-12-05
Docs: Describe more ACL effects on (re)authentication

rousskov:
Existing documentation was

* silent about %ul, max_user_ip, ident, and ident_regex side effects;
* silent about adapted_http_access context triggering authentication;
* vague about (re)authentication triggers.

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -588,14 +588,31 @@ DOC_START
 	different helper, but not unconfigure the helper completely.
 
 	Please note that while this directive defines how Squid processes
-	authentication it does not automatically activate authentication.
-	To use authentication you must in addition make use of ACLs based
-	on login name in http_access (proxy_auth, proxy_auth_regex or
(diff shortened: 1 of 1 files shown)


rousskov (commented):
This PR was triggered by admin questions. The admin was trying to avoid authentication for certain requests and was surprised that those requests still triggered authentication even though they did not meet the documented criteria.

FWIW, I tried to verify statements in the updated text (using basic authentication for test cases), but authentication caveats is not my expertise area, and I could have missed some important caveats. I would prefer not to merge this PR without @yadij blessing (at least).

We should either document or remove known caveats, but I do not insist on any specific documentation text. If you can think of a better wording, please edit the proposed text or post specific replacement suggestions.

kinkie (approved):
Thanks for clarifying the documentation!

yadij (approved):
There may be other things missing. But this update is good.

PR #1612, opened 2023-12-05
Docs: Describe surprising side effects of auth_param basic

rousskov:
acl badGuys proxy_auth Bob
    http_access deny badGuys

Admins may be surprised that their proxy_auth ACLs do not match users
with logins identical to those listed as proxy_auth ACL values. For
example, a user logged in as "Bob" will no match the above ACL if Basic
authentication is used without an explicit "casesensitive on" setting.
In fact, the above ACL cannot match any user in that environment!

### DIFF
--- a/src/cf.data.pre
+++ b/src/cf.data.pre
@@ -763,11 +763,24 @@ IF HAVE_AUTH_MODULE_BASIC
 		use the max_user_ip ACL in an http_access rule.
 
 	"casesensitive" on|off
-		Specifies if usernames are case sensitive. Most user databases
-		are case insensitive allowing the same username to be spelled
-		using both lower and upper case letters, but some are case
-		sensitive. This makes a big difference for user_max_ip ACL
-		processing and similar.
+		Specifies whether upper case letters in client-sent usernames are
+		preserved. By default and when explicitly set to "off", a username
+		extracted from Proxy-Authorization or Authorization request header is
+		forced to lower case before user credentials are checked or stored.
+
+		Most user databases are case insensitive, allowing the same username to be
+		spelled using both lower and upper case letters. For such databases, any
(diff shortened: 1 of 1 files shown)


rousskov (commented):
This PR was triggered by surprising test results obtained while working on #1611: I was not expecting a case sensitivity setting, especially the default one, to result in altered usernames.

FWIW, I tried to verify statements in the updated text, but authentication caveats is not my expertise area, and I could have missed something important. I would prefer not to merge this PR without @yadij blessing (at least).

We should either document or remove this caveat, but I do not insist on any specific documentation text. If you can think of a better wording, please edit the proposed text or post specific replacement suggestions.

kinkie (approved):

rousskov:
> FYI, there is likely a bug involved here.

I agree that the current behavior can be considered a bug (or a combination of bugs). We probably disagree on the specifics of that bug.

I recommend merging this documentation (if we agree that it represents the current behavior that is unlikely to change in a few days) and then wait for improvements and fixes. When those improvements/fixes come, they will be reflected in the already merged documentation, assisting us with their impact comprehension.

Regardless of whether these documentation improvements are merged, I recommend discussing any authentication configuration/implementation changes elsewhere. I see several significant red flags in the [proposed changes](https://github.com/squid-cache/squid/pull/1612#issuecomment-1841734391), but it feels wrong to start reviewing that proposal _here_, especially if we agree that those changes are not going to be trivial and should be done in dedicated PR(s). If you are going to work on improving authentication along the lines of that proposal, please make an RFC on squid-dev, open a Draft PR (without any code changes at first), or something like that.

kinkie:
@yadij any comments? I'll clear for merge on Dec 17th unless you have any, and then we can add looking into the bug to our backlog

yadij (commented):
.

PR #1613, opened 2023-12-05
Avoid UB when packing a domain name

jtstrs:
rfc1035NamePack() called rfc1035LabelPack() with a nil label buffer.
Feeding memcpy() a nil buffer is undefined behavior, even if size is 0.

### DIFF
--- a/src/dns/rfc1035.cc
+++ b/src/dns/rfc1035.cc
@@ -137,11 +137,11 @@ rfc1035NamePack(char *buf, size_t sz, const char *name)
         off += rfc1035LabelPack(buf + off, sz - off, t);
     xfree(copy);
 
-    // Put terminate NULL label at the end
-    *(buf + off) = 0;
-    off++;
+    // Add a terminating 0-size label
+    assert(off < sz);
+    buf[off] = 0;
+    ++off;
 
-    assert(off <= sz);
     return off;
 }
 
--- a/src/dns/rfc1035.cc
+++ b/src/dns/rfc1035.cc
@@ -100,14 +100,16 @@ rfc1035HeaderPack(char *buf, size_t sz, rfc1035_message * hdr)
 static int
 rfc1035LabelPack(char *buf, size_t sz, const char *label)
 {
+    assert(label);
+    assert(!strchr(label, '.'));
+
     int off = 0;
-    size_t len = label ? strlen(label) : 0;
-    if (label)
-        assert(!strchr(label, '.'));
+    size_t len = strlen(label);
     if (len > RFC1035_MAXLABELSZ)
         len = RFC1035_MAXLABELSZ;
     assert(sz >= len + 1);
-    *(buf + off) = (char) len;
+
+    *buf = (char) len;
     off++;
     memcpy(buf + off, label, len);
     off += len;
@@ -134,7 +136,11 @@ rfc1035NamePack(char *buf, size_t sz, const char *name)
     for (t = strtok(copy, "."); t; t = strtok(nullptr, "."))
         off += rfc1035LabelPack(buf + off, sz - off, t);
     xfree(copy);
-    off += rfc1035LabelPack(buf + off, sz - off, nullptr);
+
+    // Put terminate NULL label at the end
+    *(buf + off) = 0;
+    off++;
+
     assert(off <= sz);
     return off;
 }
--- a/src/dns/rfc1035.cc
+++ b/src/dns/rfc1035.cc
@@ -104,12 +104,11 @@ rfc1035LabelPack(char *buf, size_t sz, const char *label)
     assert(!strchr(label, '.'));
 
     int off = 0;
-    size_t len = strlen(label);
+    auto len = strlen(label);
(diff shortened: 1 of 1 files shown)


jtstrs:
Using ```memcpy``` is undefined behiour according to
https://en.cppreference.com/w/cpp/string/byte/memcpy
> If either dest or src is an [invalid or null pointer], 
> the behavior is undefined, even if count is zero.

src/dns/rfc1035.cc
```diff
@@ -134,7 +136,11 @@ rfc1035NamePack(char *buf, size_t sz, const char *name)
     for (t = strtok(copy, "."); t; t = strtok(nullptr, "."))
         off += rfc1035LabelPack(buf + off, sz - off, t);
     xfree(copy);
-    off += rfc1035LabelPack(buf + off, sz - off, nullptr);
+
+    // Put terminate NULL label at the end
+    *(buf + off) = 0;
```
rousskov, line 141: I am not sure, but it looks like the proposed code might overwrite the supplied buffer. For example, that overwrite may happen during rfc1035NamePack("", 0, "test").

AFAICT, the official code asserts in that case because the second (i.e. post-loop or terminating) rfc1035LabelPack() call runs out of space; the corresponding assertion is inside rfc1035LabelPack(): `assert(sz >= len + 1)`. PR effectively replaces that assertion (during buffer-termination) with custom and overly optimistic code.

We do not want either outcome, but if we have to pick between the two evils in this terrible code, it is better to assert than to overwrite.

P.S. When you modify C++ code, please use const and [AAA](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/) as much as possible.

rousskov (changes requested):
Thank you for working on fixing this bug!

src/dns/rfc1035.cc
yadij: Also, please use explicit `'\0'` character instead of `0` to avoid injecting more than 1 byte.

rousskov: > For example, that overrun may happen during rfc1035NamePack("", 0, "test").

I suspect that my (extreme) example does not really work because strtok("") will return a non-nil string and we will will enter the loop (and then assert inside rfc1035LabelPack() instead of overflowing the buffer).

However, I still think that this change request is valid -- the underlying problem does exist; we just need a more complex test to trigger it. Perhaps something like the test case below would trigger buffer overflow in the new termination code?

```C++
char buf[2];
// the loop will pack char(1) and character 'A', returning offset 2;
// the termination code will write into an off-limits  `buf + 2` memory
rfc1035NamePack(buf, sizeof(buf), "A"); // needs 3 bytes but is given 2
```

jtstrs: Such cases should handle by assertion at the end of ```rfc1035NamePack``` function. (There is also true for test suggested by you) https://godbolt.org/z/r511j47b7
But that assertion happening only after editing of buffer, so probably moving this assertion before adding null-term label should solve possible problem, i.e.

```
// Ensure that no buffer overflow there
assert(off + 1 <= sz);
// Put terminate NULL label at the end
*(buf + off) = 0;
off++;

return off;
```

What do you think about this proposal?

rousskov: > Such cases should handle by assertion at the end of rfc1035NamePack function. (There is also true for test suggested by you) https://godbolt.org/z/r511j47b7

I agree that such cases should be handled by an assertion in rfc1035NamePack(), but rfc1035NamePack() must assert _before_ overflowing the buffer, not _after_ overflowing it! After an overflow, the problem that the assertion needs to prevent cannot be prevented; the game has been lost.

>  But that assertion happening only after editing of buffer, so probably moving this assertion before adding null-term label should solve possible problem

Exactly. Here is a (slightly) better sketch:

```C++
// add a terminating 0-size label
assert(off < sz);
buf[off] = 0;
++off;

return off;
```

jtstrs: Got it

rousskov (approved):
Thank you for addressing my primary concern. I am approving this PR after polishing its code and PR metadata a little. @jtstrs, please double check my changes.

src/dns/rfc1035.cc
```diff
@@ -134,8 +136,12 @@ rfc1035NamePack(char *buf, size_t sz, const char *name)
     for (t = strtok(copy, "."); t; t = strtok(nullptr, "."))
         off += rfc1035LabelPack(buf + off, sz - off, t);
     xfree(copy);
-    off += rfc1035LabelPack(buf + off, sz - off, nullptr);
```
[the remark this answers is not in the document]
yadij: Either implementation is fine with me.

yadij (approved):
We still have the UB implied by potential `memcpy()` mistakes. But replacing the one in label packing with `memove()` can be done when the other ones are updated.

rousskov:
OK to test

kinkie:
OK to test

kinkie (approved):

PR #1614, opened 2023-12-06
Declutter FD_WRITE and FD_READ

kinkie:
Avoid clashes with same-name constants defined by MS WIndows Socket API.

Also removed a few related ununsed macros.

### DIFF
--- a/compat/os/mswindows.h
+++ b/compat/os/mswindows.h
@@ -286,8 +286,6 @@ struct timezone {
 
 #undef FD_CLOSE
 #undef FD_OPEN
-#undef FD_READ
-#undef FD_WRITE
 
 #ifndef EISCONN
 #define EISCONN WSAEISCONN
--- a/compat/os/mswindows.h
+++ b/compat/os/mswindows.h
@@ -284,9 +284,6 @@ struct timezone {
 #define _PATH_DEVNULL "NUL"
 #endif
 
-#undef FD_CLOSE
-#undef FD_OPEN
-
 #ifndef EISCONN
 #define EISCONN WSAEISCONN
 #endif
--- a/src/fd.h
+++ b/src/fd.h
@@ -20,7 +20,7 @@ enum class FdOps {
 void fd_close(int fd);
 void fd_open(int fd, unsigned int type, const char *);
 void fd_note(int fd, const char *);
-void fd_bytes(int fd, int len, const FdOps type);
+void fd_bytes(int fd, int len, const FdOps);
 void fdDumpOpen(void);
 int fdUsageHigh(void);
 void fdAdjustReserved(void);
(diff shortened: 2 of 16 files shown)


yadij (dismissed):

src/fd.h
```diff
@@ -11,10 +11,16 @@
 #ifndef SQUID_FD_H_
 #define SQUID_FD_H_
 
+/// what kind of operation to count traffic for in fd_bytes()
```
rousskov, line 14: We should also make it possible to reuse these values in other, similar contexts, without changing this description:

```suggestion
/// distinguishes reading/importing I/O operations from their writing/exporting counterparts
```

I do not insist on any specific wording.

compat/os/mswindows.h
```diff
@@ -286,8 +286,6 @@ struct timezone {
 
 #undef FD_CLOSE
 #undef FD_OPEN
```
rousskov, line 288: If possible:

```suggestion
```

src/tests/stub_fd.cc
```diff
@@ -18,7 +18,7 @@ fde *fde::Table = nullptr;
 int fdNFree(void) STUB_RETVAL(-1)
 void fd_open(int, unsigned int, const char *) STUB
 void fd_close(int) STUB
-void fd_bytes(int, int, unsigned int) STUB
+void fd_bytes(int, int, const FdOps) STUB
```
rousskov, line 21: There is no variable here for `const` to matter:

```suggestion
void fd_bytes(int, int, FdOps) STUB
```

rousskov (changes requested):
Thank you for fixing this.

P.S. I will polish fd_bytes() implementation to match the new types.

src/fd.h
```diff
@@ -11,10 +11,16 @@
...
+enum class FdOps {
+    READ,
+    WRITE
```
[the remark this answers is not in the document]
kinkie: Done in 6cb1de634e (to be pushed shortly)

src/fd.h
```diff
@@ -11,10 +11,16 @@
 #ifndef SQUID_FD_H_
 #define SQUID_FD_H_
 
+/// what kind of operation to count traffic for in fd_bytes()
+enum class FdOps {
```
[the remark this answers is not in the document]
kinkie: Done in soon-to-be-pushed 4c522f9732

src/fd.h
kinkie: I don't have any preference, rewording as suggested

src/tests/stub_fd.cc
kinkie: De-const-ed in to-be-pushed 73b392fa36

yadij (dismissed):

rousskov (approved):
Thank you for this cleanup. Your changes allowed us to replace the runtime assertion with a compile-time check. I adjusted fd_bytes() accordingly, polished PR title/description, and approved this PR.

rousskov:
Jenkins, test this please.

PR #1616, opened 2023-12-06
Remove broken and disabled icpPktDump()

kinkie:


yadij:
IMO this should be altered to the level-2 debug output for ICP protocol instead of dropped entirely.

kinkie:
I don't think this is useful at all as it is, as its use case is not particularly relevant (ICP has been stable for over a decade now) and can be better served by not-inline tools.

To use this, someone would have to:
1. know this exists
2. recompile squid
3. run it in production, dumping to log all messages from all peers at all time

Compared to:
1. tcpdump, prefiltering with appropriate network parameters
2. use a protocol inspection tool such as wireshark

I can see this as having a valid use case when the ICP parser was being developed, but I wouldn't say that's currently relevant anymore

src/icp_v2.cc
```diff
@@ -604,24 +604,6 @@ icpHandleIcpV2(int fd, Ip::Address &from, char *buf, int len)
...
-icpPktDump(icp_common_t * pkt)
-{
-    Ip::Address a;
-
-    debugs(12, 9, "opcode:     " << std::setw(3) << pkt->opcode  << " " << icp_opcode_str[pkt->opcode]);
-    debugs(12, 9, "version: "<< std::left << std::setw(8) << pkt->version);
-    debugs(12, 9, "length:  "<< std::left << std::setw(8) << ntohs(pkt->length));
-    debugs(12, 9, "reqnum:  "<< std::left << std::setw(8) << ntohl(pkt->reqnum));
-    debugs(12, 9, "flags:   "<< std::left << std::hex << std::setw(8) << ntohl(pkt->flags));
-    a = (struct in_addr)pkt->shostid;
```
rousskov, line 618: I agree with both of you: This code should be removed (but _not_ because some external tool can offer similar debugging in some circumstances) and, ideally, new always-compiled high-quality debugging code should be added. Those two changes do not have to happen in a _single_ PR because the being-removed code does not compile if enabled (and was probably dead/unused for a very long time):

```
icp_v2.cc:618:30: error: no matching function for call to ‘in_addr::in_addr(uint32_t&)’
  618 |     a = (struct in_addr)pkt->shostid;
```

Thus, I am approving this PR (with adjusted PR title/description).

rousskov (approved):

PR #1619, opened 2023-12-13
Polished TunnelStateData::usePinned()

jtstrs:
... to avoid creating an impression of nil connManager dereference and
to reduce unwanted differences with FwdState::usePinned().

### DIFF
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -1453,7 +1453,7 @@ TunnelStateData::startConnecting()
 void
 TunnelStateData::usePinned()
 {
-    assert(request);
+    Must(request);
     const auto connManager = request->pinnedConnection();
     Comm::ConnectionPointer serverConn = nullptr;
 
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -1453,8 +1453,10 @@ TunnelStateData::startConnecting()
 void
 TunnelStateData::usePinned()
 {
-    Must(request);
+    Assure(request);
     const auto connManager = request->pinnedConnection();
+    Assure(connManager);
+
     try {
         const auto serverConn = ConnStateData::BorrowPinnedConnection(request.getRaw(), al);
         debugs(26, 7, "pinned peer connection: " << serverConn);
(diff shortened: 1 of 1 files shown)


src/tunnel.cc
```diff
@@ -1453,8 +1453,10 @@ TunnelStateData::startConnecting()
 void
 TunnelStateData::usePinned()
 {
-    Must(request);
+    Assure(request);
```
rousskov, line 1456: Please undo this out-of-scope and distracting improvement:

```suggestion
    Must(request);
```

In the future, please avoid out-of-scope changes (by default).

rousskov (changes requested):
> PR description: Fix potentinal use of nullptr in usePinned

There is no potential for a nil pointer dereference here AFAICT, but this PR does highlight at least two "code quality" problems that can and should be addressed. I left the corresponding detailed change request.

BTW, in most cases, if there is a known nil pointer dereference, it cannot be _properly_ addressed by adding an assertion: Assertions are meant for conditions we believe to be always true (rather than for properly handling a known false case).

rousskov (approved):
Thank you.

rousskov:
I adjusted PR title/description to match the current PR code. @jtstrs, in the future, please keep that metadata up-to-date because it will be [used](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message) for the commit message when merging this PR.

rousskov:
@jtstrs, Squid automation scripts identify the author of your contributions as `Alex <bigalex934@gmail.com>` because that is what you have in your GitHub account. That string will be added to CONTRIBUTORS _unless_ you tell me to adjust it (e.g., to add a last name or to change the email address). If you would like to adjust, please let me know!

jtstrs:
>  I adjusted PR title/description to match the current PR code. @jtstrs, in the future, please keep that metadata up-to-date because it will be [used](https://github.com/measurement-factory/anubis/blob/master/README.md#commit-message) >  for the commit message when merging this PR.

Good, will take it into account

> @jtstrs, Squid automation scripts identify the author of your contributions as Alex <bigalex934@gmail.com> because that is what you have in your GitHub account. That string will be added to CONTRIBUTORS unless you tell me to adjust it (e.g., to add a last name or to change the email address). Not adding anything to CONTRIBUTORS is also allowed (but discouraged). If you would like to adjust something, please let me know!

Its ok, but anyway thank you for highlighting that

rousskov:
> Its ok, but anyway thank you for highlighting that

The latest [application](https://github.com/squid-cache/squid/actions/runs/7207176081/job/19633500502?pr=1619#step:4:52) of the same script picked up another variation:

```diff
+ Alex <bigalex934@gmail.com>
+ Alexey Simakov <bigalex934@gmail.com>
```

If you do not have another preference, we will use the latter/second variant because it has more information.

kinkie (approved):

PR #1621, opened 2023-12-17
Add SBuf::push_back()

kinkie:
This allows using STL algorithms via std::back_inserter and such.

### DIFF
--- a/src/sbuf/SBuf.h
+++ b/src/sbuf/SBuf.h
@@ -190,10 +190,10 @@ public:
      */
     SBuf& append(const SBuf & S);
 
-    /// Append a single character. The character may be NUL (\0).
+    /// \copydoc push_back(char c)
     SBuf& append(const char c) { push_back(c); return *this; }
 
-    /// Append a single character, which may be '\0'
+    /// Append a single character. The character may be NUL (\0).
     void push_back(const char c);
 
     /** Append operation for C-style strings.
--- a/src/sbuf/SBuf.h
+++ b/src/sbuf/SBuf.h
@@ -191,10 +191,10 @@ public:
     SBuf& append(const SBuf & S);
 
     /// Append a single character. The character may be NUL (\0).
-    SBuf& append(const char c);
+    SBuf& append(const char c) { push_back(c); return *this; }
 
     /// Append a single character, which may be '\0'
-    void push_back(const char c) { append(c); }
+    void push_back(const char c);
 
     /** Append operation for C-style strings.
      *
--- a/src/sbuf/SBuf.h
+++ b/src/sbuf/SBuf.h
@@ -193,7 +193,7 @@ public:
     /// Append a single character. The character may be NUL (\0).
     SBuf& append(const char c);
 
-    /// Append a single character, compatibly with std::string::push_back()
+    /// Append a single character, which may be '\0'
     void push_back(const char c) { append(c); }
 
     /** Append operation for C-style strings.
--- a/src/sbuf/SBuf.h
+++ b/src/sbuf/SBuf.h
@@ -96,6 +96,7 @@ public:
     typedef MemBlob::size_type size_type;
     typedef SBufIterator const_iterator;
     typedef SBufReverseIterator const_reverse_iterator;
+    using value_type = char;
     static const size_type npos = 0xffffffff; // max(uint32_t)
 
     /// Maximum size of a SBuf. By design it MUST be < MAX(size_type)/2. Currently 256Mb.
@@ -191,6 +192,9 @@ public:
 
     /// Append a single character. The character may be NUL (\0).
     SBuf& append(const char c);
+
+    /// Append a single character, compatibly with std::string::push_back()
+    void push_back(const char c) { append(c); }
 
     /** Append operation for C-style strings.
      *
--- a/src/sbuf/SBuf.h
+++ b/src/sbuf/SBuf.h
@@ -190,11 +190,11 @@ public:
      */
     SBuf& append(const SBuf & S);
 
-    /// \copydoc push_back(char c)
+    /// \copydoc push_back(char)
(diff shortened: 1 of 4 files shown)


kinkie:
The starting use case for this for me is rfc1738_unescape, which I want to convert to a less awkward API.
Having SBuf::push_back should enable a templatized, common implementation supporting both SBuf and std::string,
thus both Squid and helpers' use cases

yadij:
> The starting use case for this for me is rfc1738_unescape, which I want to convert to a less awkward API.

Instead of changing the `rfc1738_*()` API to SBuf, please add the missing `static SBuf AnyP::Uri::Decode(const SBuf &)` to match the new `AnyP::Uri::Encode`

src/tests/testSBuf.cc
```diff
@@ -1190,6 +1193,23 @@ TestSBuf::testSBufHash()
...
+        SBuf src("foo"), dst;
+        std::copy(src.begin(), src.end(), std::back_inserter(dst));
+        CPPUNIT_ASSERT_EQUAL(dst, src);
+    }
+
+    {
+        SBuf src("foo"), dst;
+        std::copy_if(src.begin(), src.end(), std::back_inserter(dst),
+            [](char c) { return c != 'o'; });
+        CPPUNIT_ASSERT_EQUAL(dst, SBuf("f"));
```
yadij, line 1209: Please make this test a little more clear that the string is not just being mangled badly.
```suggestion
        SBuf src("foaro"), dst;
        std::copy_if(src.begin(), src.end(), std::back_inserter(dst),
            [](char c) { return c != 'o'; });
        CPPUNIT_ASSERT_EQUAL(dst, SBuf("far"));
```

src/sbuf/SBuf.h
```diff
@@ -192,6 +193,9 @@ class SBuf
     /// Append a single character. The character may be NUL (\0).
     SBuf& append(const char c);
 
+    /// Append a single character, compatibly with std::string::push_back()
+    void push_back(const char c) { append(c); }
```
yadij, line 197: Please do these the other way around. Implementing our custom `append()` API in terms of the STL one.

```suggestion
    /// \copydoc push_back(const char c)
    /// The character may be NUL (\0).
    SBuf& append(const char c) { push_back(c); return *this; }

    /// Append a single character, compatibly with std::string::push_back()
    void push_back(const char c);
```

This will need relevant SBuf.cc changes.

yadij (changes requested):

src/sbuf/SBuf.h
```diff
@@ -192,6 +193,9 @@ class SBuf
     /// Append a single character. The character may be NUL (\0).
     SBuf& append(const char c);
 
+    /// Append a single character, compatibly with std::string::push_back()
```
rousskov, line 196: Please drop the ", compatibly with std::string::push_back()" part. This method is _not_ compatible with std::string::push_back(), and we are not trying to be compatible with std::string::push_back(). We are trying to be compatible with std::back_inserter and similar STL code AFAICT, but that fact does not need to be documented explicitly IMO.

```suggestion
    /// Append a single character.
```

I am against sprinkling SBuf code with comments about NULs, but if you or others insist on preserving that part of the comment:

```suggestion
    /// Append a single character. The character may be NUL (\0).
```

I have removed the corresponding controversial text from the PR description.

rousskov:
> The starting use case for this for me is rfc1738_unescape, which I want to convert to a less awkward API. Having SBuf::push_back should enable a templatized, common implementation supporting both SBuf and std::string, thus both Squid and helpers' use cases

FWIW, the above plan sounds like a big red flag to me. We want rfc1738_unescape() to be fast and easy to maintain. It is unlikely that we will achieve that if we focus on std::string compatibility. Modern helper classes should use SBuf, not std::string.

rousskov (changes requested):

kinkie:
> > The starting use case for this for me is rfc1738_unescape, which I want to convert to a less awkward API. Having SBuf::push_back should enable a templatized, common implementation supporting both SBuf and std::string, thus both Squid and helpers' use cases
> 
> FWIW, the above plan sounds like a big red flag to me. We want rfc1738_unescape() to be fast and easy to maintain. It is unlikely that we will achieve that if we focus on std::string compatibility. Modern helper classes should use SBuf, not std::string.

In Squid, I fully agree with you. I would 
Helpers on the other hand, don't have accesso to SBuf, and are better off with std::string than c-string.

This is an attempt to cover both use cases with the same implementation, which would result in better correctness and testability.

rousskov:
> Helpers on the other hand, don't have accesso to SBuf

What makes you think that?

```
src/security/cert_generators/file/Makefile.am:  $(top_builddir)/src/sbuf/libsbuf.la \
src/security/cert_generators/file/certificate_db.cc:#include "sbuf/Stream.h"
src/security/cert_generators/file/security_file_certgen.cc:#include "sbuf/Stream.h"
```

> This is an attempt to cover both use cases with the same implementation, which would result in better correctness

I do not expect much impact on correctness but it is likely result in worse performance because SBuf and std::string have different performance tradeoffs.

src/sbuf/SBuf.h
kinkie: Corrected

src/sbuf/SBuf.h
kinkie: done

src/tests/testSBuf.cc
kinkie: I changed to use the standard string used throughout the test, this should offer a greater confidence in the result

rousskov (approved):
Thank you for addressing my primary concern. I am approving this PR after committing a few minor polishing touches.

rousskov:
I also adjusted PR title to be a bit more precise, but I do not insist on that adjustment.

yadij (approved):

PR #1624, opened 2023-12-21
Improve and broadly use asHex()

kinkie:
Backport PR #1578


yadij:
Does this fix a specific bug in v6? it is rather a lot of code change for backport to stable release.

kinkie:
Not necessarily, no. But it is a prerequisite for fixing the qos cache
manager reporting

@mobile

yadij (approved):

PR #1625, opened 2023-12-22
Bug 5329: cbdata.cc:276 "c->locks > 0" assertion on reconfigure

eduard-bagdasaryan:
Recent commit 0f78379 correctly removed an excessive cbdata lock of a
CachePeer::digest object in peerDigestCreate() but accidentally lost
another digest lock while inlining peerDigestCreate(). The resulting
excessive unlocking triggered reconfiguration assertions.
    
This change restores the lost lock as a short-term fix.

Long-term, CachePeer code should be fixed to become an exclusive[^1]
PeerDigest owner (i.e. creating and deleting its cbdata-protected digest
object without locking, unlocking, or checking locks). That improvement
is already in the works, but it requires significant code refactoring.

[^1]: Shared PeerDigest ownership (i.e. reference counting instead of
explicit delete and cbdata) does not work well in this context due to
circular references.

### DIFF
--- a/src/cache_cf.cc
+++ b/src/cache_cf.cc
@@ -2404,7 +2404,7 @@ parse_peer(CachePeers **peers)
 #if USE_CACHE_DIGESTS
     if (!p->options.no_digest) {
         const auto pd = new PeerDigest(p);
-        p->digest = cbdataReference(pd);
+        p->digest = cbdataReference(pd); // CachePeer destructor unlocks
     }
 #endif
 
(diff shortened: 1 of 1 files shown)


rousskov (approved):
This code continues to be conceptually wrong -- object owners should not cbdata-lock/unlock/check _their_ objects. Violations of that cbdata design invariant lead to lock mismatching bugs like Bug 5329 in recent commit 0f783794b8cbdd6342cc2d36343db72f04d2519c and the memory leak bug fixed by that commit. Factory is already working on fixing the underlying problem, but the changes required to do that properly are substantial/noisy. This minimal fix is the best next step we could come up with to address the immediate needs of those running master/v7 code.

kinkie (approved):

PR #1626, opened 2023-12-25
Add AnyP::Uri::Decode()

kinkie:
Implement percent-decoding according to RFC 3986. Currently unused.
TODO: Upgrade rfc1738_unescape() callers.


rousskov (changes requested):
I have three serious concerns about this PR: One is about exposing the entire Squid code to bad fromhex() interface and two are about the proposed AnyP::Uri::Decode() functionality/API (`%%` decoding and pct-encoding error handling). Please see individual change requests for details.

Where do you plan to use AnyP::Uri::Decode(). Do you plan to replace all or most rfc1738_unescape() calls with it? Knowing your immediate plans may help in future review iterations...

P.S. There are many small problems in the current PR (typos, out of scope changes, etc.). I marked some and can fix all I can find later, after the primary problems are addressed.

src/anyp/Uri.cc
```diff
@@ -81,6 +81,66 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+        // ch is '%', let's look ahead
+        ++i;
+        if (i == end) {
+            // '%' is the last character, so it's not encoded
+            output.append(ch);
+            break;
+        }
+        const auto ch1 = *i;
+        if (ch1 == '%') {
+            // met "%%"; output "%" and continue
```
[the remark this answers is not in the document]
kinkie: As this is a parser, I chose to err on the side of accepting input which might be invalid.
If there is consensus, I can remove that bit; @yadij what do you think?

kinkie:
> I have three serious concerns about this PR: One is about exposing the entire Squid code to bad fromhex() interface and two are about the proposed AnyP::Uri::Decode() functionality/API (`%%` decoding and pct-encoding error handling). Please see individual change requests for details.

I thought about what to do on fromhex() before choosing this path;
I felt that avoiding code duplication would be a better tradeoff than duplicating functionality.

> Where do you plan to use AnyP::Uri::Decode(). Do you plan to replace all or most rfc1738_unescape() calls with it? Knowing your immediate plans may help in future review iterations...

I embarked on this path as suggested in a PR #1621 [comment](https://github.com/squid-cache/squid/pull/1621#issuecomment-1859217798). Generally speaking, yes. Porting to this method will help move SBuf adoption forward, starting with some use cases, where a c-string is copied and then decoded.

> P.S. There are many small problems in the current PR (typos, out of scope changes, etc.). I marked some and can fix all I can find later, after the primary problems are addressed.
Thanks

src/anyp/Uri.cc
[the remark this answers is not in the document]
rousskov: > As this is a parser, I chose to err on the side of accepting input which might be invalid.

A parser that "accepts" invalid input needs a very good justification for doing so. "Being a parser" is not a justification (arguably, "being a parser" justifies the _opposite_ behavior!). A parser that, besides accepting invalid input, produces one of several possible output variations from that invalid input also needs to justify the chosen variation.

These basic principles are especially important when parser output is meant, in part, for external consumption (e.g., a decoded and re-encoded request URL is then forwarded to a server) because all external consumers lose the ability to detect previously invalid input (triggering the usual set of "hops A and B disagree on what the URL is" vulnerabilities) and because some external consumers may rely on the original/invalid input (i.e. the proxy is perceived as the one breaking things that "work" without it).

> If there is consensus, I can remove that bit; @yadij what do you think?

FWIW, there is certainly no consensus about the current code.

P.S. The above thoughts apply to all cases where the proposed parser is accepting invalid input, of course (`%%` is just one of them).

src/tests/testURL.cc
```diff
@@ -81,9 +83,43 @@ TestUri::testDefaultConstructor()
...
+void
+TestUri::testEncodeDecode()
+{
+    std::vector<std::pair<const char *, const char *>>
+        testCasesEncode = {
+            {"foo", "foo"},
+            {"foo%", "foo%25"},
+            {"fo%o", "fo%25o"},
+            {"fo%%o", "fo%25%25o"}
+            } ,
```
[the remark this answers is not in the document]
kinkie: Ok, done

rousskov:
> > I have three serious concerns about this PR: One is about exposing the entire Squid code to bad fromhex() interface ...
 
> I thought about what to do on fromhex() before choosing this path; I felt that avoiding code duplication would be a better tradeoff than duplicating functionality.

AFAICT, "avoiding code duplication vs. duplicating functionality" is a false dichotomy in this case. Here are a few options we can pick from (and there may be more):

1. make unsafe fromhex() interface public (i.e. current PR)
2. provide a new public hex-character-to-integer function with safe interface and the same (or, optionally, faster) implementation (i.e. "duplicate functionality")
3. provide a new public hex-character-to-integer function with safe interface that uses/calls fromhex() internally
4. provide a new public pct-encoding-to-char function with safe interface

I see no good reason to pick option 1 -- at least two and possibly all three other options are better IMO.
 
> > Where do you plan to use AnyP::Uri::Decode(). Do you plan to replace all or most rfc1738_unescape() calls with it? Knowing your immediate plans may help in future review iterations...
 
> Generally speaking, yes. Porting to this method will help move SBuf adoption forward, starting with some use cases, where a c-string is copied and then decoded.

To provide context for deciding how to resolve current problems, I recommend starting with at least two specific use cases: Acl::UrlPathCheck::match() and parse_peer(). How should these callers handle invalid input?

src/tests/testURL.cc
```diff
@@ -81,9 +83,43 @@ TestUri::testDefaultConstructor()
...
+            {"foo%%", "foo%"},
+            {"foo%25", "foo%"},
+            {"foo%2", "foo%2"},
+            {"fo%o", "fo%o"},
+            {"fo%2o", "fo%2o"},
+            {"fo%25o", "fo%o"},
+            {"fo%%o", "fo%o"},
+            {"fo%25%25o", "fo%%o"},
+            {"fo%20o", "fo o"}
+            }
```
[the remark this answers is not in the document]
kinkie: Thanks for suggesting this, this error case has shown that the implementation cannot handle gracefully these cases. I'll have to rework the implementation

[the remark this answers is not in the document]
rousskov: FWIW, "gracefully" sounds like another red flag to me. I recommend agreeing on how invalid input should be handled _before_ you "rework the implementation".

src/anyp/Uri.cc
```diff
@@ -81,6 +81,66 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+        if (i == end) {
+            // end is one-past-last, so ch2 is actually garbage
+            output.append(ch).append(ch1);
+            break;
+        }
+        const auto ch2 = *i;
+        const auto hex2 = HextDigitToInt(ch2);
+        if (hex2 == -1) {
+            // ch2 is not a hex digit, so it's not encoded
+            output.append(ch).append(ch1).append(ch2);
```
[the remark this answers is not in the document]
kinkie: The RFC reads:
`A percent-encoded octet is encoded as a character triplet`, or in BNF: `pct-encoded = "%" HEXDIG HEXDIG`, so the %4 should be copied over, and we are left with `%4A`.

The case of `%4%%41` is more interesting, because given that %% -> % is not mentioned in the RFC, we could say right now we are not compliant, and by the reading of the RFC we should output %4%A.

I think the key to decide what to do is in the last paragraph of [RFC 3986 section 2.4](https://datatracker.ietf.org/doc/html/rfc3986#section-2.4), which says:
```
Because the percent ("%") character serves as the indicator for
percent-encoded octets, it must be percent-encoded as "%25" for that
octet to be used as data within a URI
```
Which means that %4% is an invalid %-encoded string (as is %%).

src/anyp/Uri.cc
```diff
@@ -81,6 +81,66 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+        }
+        const auto hex1 = HextDigitToInt(ch1);
+        if (hex1 == -1) {
+            // ch1 is not a hex digit, so it's not encoded
+            output.append(ch).append(ch1);
+            continue;
+        }
+        ++i;
+        if (i == end) {
+            // end is one-past-last, so ch2 is actually garbage
```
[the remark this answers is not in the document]
kinkie: changed

src/anyp/Uri.cc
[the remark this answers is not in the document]
rousskov: > The RFC reads: `A percent-encoded octet is encoded as a character triplet`, or in BNF: `pct-encoded = "%" HEXDIG HEXDIG`,

Yes.

> so the %4 should be copied over

Why?! I see no logical connection between the two statements and nothing in the URI grammar that allows us to "copy over" bare `%`. As I said [earlier](https://github.com/squid-cache/squid/pull/1626#discussion_r1436201046) (and you may have implied in your "the key to decide" response quoted below), the grammar prohibits `%` outside of `pct-encoded` triplets. If those bare `%` are banned by the grammar, why should we copy them instead of rejecting malformed input?


> The case of `%4%%41` is more interesting ... by the reading of the RFC we should output %4%A.

Why?! I see no sequence of URI grammar rules that would yield that output for that input. What sequence produces that output?


> I think the key to decide what to do is in the last paragraph of [RFC 3986 section 2.4](https://datatracker.ietf.org/doc/html/rfc3986#section-2.4), which says:
> 
> ```
> Because the percent ("%") character serves as the indicator for
> percent-encoded octets, it must be percent-encoded as "%25" for that
> octet to be used as data within a URI
> ```

... and that prose is implied by the RFC grammar rules AFAICT -- bare `%` cannot appear anywhere except as the indicator for percent-encoded octets: `unreserved`, `sub-delims`, and other URI grammar rules do not allow `%` characters.
 
> Which means that %4% is an invalid %-encoded string (as is %%).

Agreed. And `%X` and `%AX`.

src/anyp/Uri.cc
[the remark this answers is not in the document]
yadij: IMO, in this case "Accept" means not decoding the sequence.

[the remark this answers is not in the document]
yadij: Changed my mind after considering https://github.com/squid-cache/squid/pull/1626#discussion_r1436560296 . The decoder should throw when invalid `pct-encoded` input is received.

src/tests/testURL.cc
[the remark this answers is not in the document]
yadij: The "ignore invalid '%`" behaviour can be wrapped in tolerant-parse exceptions if and **only** if actually proven needed by valid traffic. We should prefer the strict parse and validation.

src/anyp/Uri.cc
```diff
@@ -81,6 +81,66 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
     return output;
 }
 
+SBuf
+AnyP::Uri::Rfc3986Encode(const SBuf &buf)
```
yadij, line 85: Please just have the callers pass the CharacterSet they want encoded. Each caller case has a different set even within URI sections.

yadij (changes requested):

src/anyp/Uri.cc
kinkie: This is still possible using the Encode function, that's not going away under this proposal.
I special-cased RFC3986 for convenience; do you still think that this use case is not common enough to warrant it?

src/anyp/Uri.cc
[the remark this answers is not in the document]
kinkie: > Changed my mind after considering [#1626 (comment)](https://github.com/squid-cache/squid/pull/1626#discussion_r1436560296) . The decoder should throw when invalid `pct-encoded` input is received.

Yeah, that was how my thoughts evolved as well. Good, I see we have consensus that invalid input should be rejected. Alex and I also discussed using std::optional<SBuf> as a possible way to signal errors; I think I like throwing more and for now I'll use that, but the point is not yet settled

src/anyp/Uri.cc
```diff
@@ -81,6 +81,48 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+        ++i;
+        if (i == end)
+            throw TextException("incomplete %-encoded triplet at end of input", Here());
+        const auto ch2 = *i;
+        const auto hex1 = HextDigitToInt(ch1);
+        const auto hex2 = HextDigitToInt(ch2);
+        if (hex1 == -1 || hex2 == -1)
+            throw TextException("invalid %-encoded triplet", Here());
+
+        const auto decoded = hex1 << 4 | hex2;
```
rousskov, line 120: Please make `decoded` a character instead of secretly relying on the current lack of SBuf::append(int) or a similar method or another implicit conversion.

```suggestion
        const auto decoded = static_cast<char>(hex1 << 4 | hex2);
```

src/anyp/Uri.h
```diff
@@ -113,6 +113,13 @@ class Uri
     /// the provided set of expected characters.
     static SBuf Encode(const SBuf &, const CharacterSet &expected);
 
+    /// %-encode all characters except "unreserved" ones, as defined by RFC 3986
+    static SBuf Rfc3986Encode(const SBuf &);
+
+    /// %-decode characters an a buffer. Throws TextException if the input
+    /// is invalid
```
rousskov, line 120: By default, a function may throw. A parsing function that does not return std::optional value should be _expected_ to throw on parsing errors. By default, a throwing Squid code throws TextException. We do not want to start documenting all those defaults/expectations for every function, even if we have not fully embraced those defaults as developers yet.

```suggestion
    /// %-decoded version of the given buffer
```

or

```suggestion
    /// %-decodes the given buffer
```

include/rfc1738.h
```diff
@@ -64,6 +64,9 @@ extern char *rfc1738_do_escape(const char *url, int flags);
  */
 extern void rfc1738_unescape(char *url);
 
+/// convert a single-digit hex ascii character to its corresponding value
+extern int HextDigitToInt(char);
```
yadij, line 68: IMO the reduction of future dependencies on this deprecated code is worth duplicating the logic in this helper function in the `anyp/Uri.cc` where the new decoder is. Leaving the `include/` and `lib/` code untouched.

src/anyp/Uri.cc
yadij: I think it will be not common enough, and touching the encoder should be out of scope for this PR. If I am wrong about the utility it and/or other wrappers can be easily added in a followup PR.

src/anyp/Uri.cc
```diff
@@ -81,6 +81,48 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+        output.append(decoded);
+    }
```
yadij, line 122: This being a parser, we should use a `Tokenizer`. 
The logic of this loop then becomes:
```suggestion
    Parser::Tokenizer tok(buf);
    while(!tok.atEnd()) {
        static const auto unencodedOctets = CharacterSet("unencoded","%").compliment();
        output.append(tok.prefix("unencoded", unencodedOctets));

        if (tok.atEnd())
            continue;

        if (tok.skipOne('%')) {
            int64_t hex1 = 0, hex2 = 0;
            if (tok.int64(hex1, 16, false, 1) && tok.int64(hex2, 16, false, 1)) {
                output.append(char((hex1<<4)|hex2));
                continue;
            }
        }

        throw TextException("invalid %-encoded triplet", Here());
    }
```

yadij (changes requested):

src/anyp/Uri.cc
rousskov: Switching to Tokenizer is a step forward and will reduce the amount of work I will need to do in this PR. Keeping the following rules of thumb in mind may reduce that work further:

* avoid atEnd() checks inside the parsing loop
* avoid `if` statements inside the parsing loop[^1]
* avoid Tokenizer::int64() when parsing strict grammars
* use names of grammar elements from the RFC
* keep in mind that Tokenizer::foo(description, ...) methods throw on bad and insufficient input
* InsufficientInput exceptions must not spread beyond parsing code[^2]

[^1]: Some `if` statements are unavoidable, of course. The proposed loop has three (including one complex statement that should really be counted as two). I hope we can reduce that count to one.

[^2]: Unlike TextExceptions, InsufficientInput exceptions are dedicated to parsing needs. They are not general-purpose exceptions. If "final" parsing code decides to use helper parser classes/functions that may throw InsufficientInput exceptions, that parsing code is responsible for shielding the rest of Squid from them.

src/anyp/Uri.cc
kinkie: done, using a type declaration instead of a static_cast

src/anyp/Uri.cc
kinkie: done, using a slightly different flow than suggested by @yadij

yadij (changes requested):

src/anyp/Uri.cc
```diff
@@ -81,6 +103,50 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+    SBuf token;
+    static const auto unencodedOctets = CharacterSet("unencoded", "%").complement();
+
+    tok.prefix(token, unencodedOctets);
+    output.append(token);
+
+    while (!tok.atEnd())
+    {
+        if (!tok.skip('%')) // should never happen
+            throw TextException("incomplete %-encoded triplet at end of input", Here());
```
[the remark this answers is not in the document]
kinkie: I'm trying to navigate between your and Alex' suggestion; while I welcome the scrutiny over my proposal, let's walk through the code because I think that while the message could very well be improved, the code itself should be okay

1.  Lines 125-126's purpose is ensure that we are at the end of input because there's no '%' in the input, or that the first character of the unparsed input is a '%'.
2. if it's the former case, then the condition at line 128 is false, the loop is never executed, and we just return the input.
3. If it's the latter, then the conditions at line 130 should always be true. We could argue that this could be an assertion instead an exception, or that the text could be clearer. I'm happy to follow any suggestion here
4. lines 133-134 extract the encoded hex characters, and there's exactly two. The decoding is fairly straightforward
5. the invariant [2] should be guaranteed at every iteration of the while loop, by the conditions at lines 144-145. The test at line 144 is meant to guard against the case where the encoded digits are at the end of input; in that case the token output parameter would be let unchanged resulting in duplicate output

The main difference with your proposal is that the invariant at [2] simplifies the code structure of the inner loop, avoiding the need of multiple atEnd() tests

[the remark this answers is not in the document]
rousskov: > Francesco: I'm trying to navigate between your and Alex' suggestion

I assume you are referring to https://github.com/squid-cache/squid/pull/1626#discussion_r1437200159 as my suggestion. FWIW, I appreciate you removing atEnd() checks inside the loop, but there are still too many `if` statements AFAICT. I did not check further. To minimize conflicting suggestions, I will come back to this PR after you make Amos happy.

> We could argue that this could be an assertion instead an exception

Rules of thumb:

* If X should never happen regardless of input, then Assure(!X) or assert(!X). In most input parsing cases, Assure() is the right choice because the problem is likely confined to the problematic transaction and the nearby code already throws for other reasons.
* If X may only happen when parsing invalid input, then throw TextException.

rousskov (changes requested):

src/anyp/Uri.cc
```diff
@@ -81,6 +103,50 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+
+    Parser::Tokenizer tok(buf);
+    SBuf token;
+    static const auto unencodedOctets = CharacterSet("unencoded", "%").complement();
+
+    tok.prefix(token, unencodedOctets);
+    output.append(token);
+
+    while (!tok.atEnd())
+    {
```
yadij, line 129: After "refactor step 1a and 1b" we now have `unencoded` characters handled first and at each loop ending.

Refactor step 2: This code can be de-duplicated by moving the unencoded string handling into the start of the loop (and up from the back).

```suggestion
    {
        static const auto unencodedOctets = CharacterSet("unencoded", "%").complement();
        output.append(tok.prefix("unencoded", unencodedOctets));

```

src/anyp/Uri.cc
```diff
@@ -81,6 +103,50 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+        if (!tok.prefix(token, HexDigits, 2) || token.length() != 2)
+            throw TextException("incomplete or invalid %-encoded triplet", Here());
+
+        const auto hex1 = HextDigitToInt(token[0]);
+        const auto hex2 = HextDigitToInt(token[1]);
+        if (hex1 == -1 || hex2 == -1)
+            throw TextException("invalid %-encoded triplet", Here());
+
+        const char decoded = hex1 << 4 | hex2;
+        output.append(decoded);
```
yadij, line 142: * All of this logic is only relevant if "%" was actually the next character received.
* also, solves the issue of step 1 hitting `Assure` after consuming to the end of `tok`

Refactor step 3:
```suggestion
        if (tok.skip('%')) {
            if (!tok.prefix(token, HexDigits, 2) || token.length() != 2)
                throw TextException("incomplete or invalid %-encoded triplet", Here());

            const auto hex1 = HextDigitToInt(token[0]);
            const auto hex2 = HextDigitToInt(token[1]);
            if (hex1 == -1 || hex2 == -1)
                throw TextException("invalid %-encoded triplet", Here());

            const char decoded = hex1 << 4 | hex2;
            output.append(decoded);
        } else if (!tok.atEnd())
            throw TextException("'%' expected but not found", Here());
```

yadij, line 142: `Tokenizer` has built-in hex logic. We can use that instead of inventing a new `HexDigitToIntMap` algorithm.
In order to require exactly-2 characters we do have to parse them individually, placing a limit of 1 handles that fine.

Refactor step 4:
```suggestion
            int64_t hex1 = 0, hex2 = 0;
            if (tok.int64(hex1, 16, false, 1) && tok.int64(hex2, 16, false, 1)) {
                output.append(char((hex1<<4)|hex2));
            } else
                throw TextException("incomplete or invalid %-encoded triplet", Here());
```

yadij (changes requested):
Here are four refactor steps optimizing the logic you have. Notice that they essentially implement Alex's "rules of thumb", though not word-for-word as he stated. Also, that when followed one end up with (almost) what I suggested initially.

src/anyp/Uri.cc
```diff
@@ -81,6 +103,50 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+
+    SBuf output;
+    output.reserveSpace(buf.length()); // worst case: buf is not encoded
+
+    Parser::Tokenizer tok(buf);
+    SBuf token;
+    static const auto unencodedOctets = CharacterSet("unencoded", "%").complement();
+
+    tok.prefix(token, unencodedOctets);
+    output.append(token);
```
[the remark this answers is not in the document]
kinkie: Aaah, now I understand! Cool; I had missed this Tokenizer::prefix() overlaod as it is not positioned next to the main one in the header file

src/anyp/Uri.cc
```diff
@@ -81,6 +103,50 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+        const auto hex1 = HextDigitToInt(token[0]);
+        const auto hex2 = HextDigitToInt(token[1]);
+        if (hex1 == -1 || hex2 == -1)
+            throw TextException("invalid %-encoded triplet", Here());
+
+        const char decoded = hex1 << 4 | hex2;
+        output.append(decoded);
+
+        if (tok.prefix(token, unencodedOctets))
+            output.append(token);
```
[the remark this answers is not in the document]
kinkie: Yes, the bool-returning Tokenizer::prefix() doesn't do that for us. Good to learn this!

src/anyp/Uri.cc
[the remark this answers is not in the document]
yadij: Actually I made a mistake here on these `1a` and `1b` steps. The `SBuf prefix()` method throws if it fails to find any unencoded characters between two encoded characters (eg. `%01%02`).

We will have to retain the `bool prefix()` usage or now:
```
    if (tok.prefix(token, unencodedOctets))
        output.append(token);
````

moving of it into the loop remains suggested for the de-duplication.

**Comment by rousskov** [comment]
@yadij, we discussed adding a boolean Tokenizer data member to mark input that will not grow in the future. BinaryTokenizer has a private member for that already:

```C++
    bool expectMore_; ///< whether more data bytes may arrive in the future
```

I do _not_ yet know whether this PR needs such a member in Tokenizer, but I am not against adding it (in principle) as long as we keep Tokenizer and BinaryTokenizer classes in sync with regard to this concept. In Tokenizer, the member should be set in the class constructor and reset() methods. Unfortunately, BinaryTokenizer constructor default (expectMore=false) is different from the current implicit Tokenizer default. Thus, we would have to change one or the other and adjust all affected callers.

If we are considering adding expectMore_ to Tokenizer, we should also consider splitting Tokenizer into two classes, the legacy class that does not throw (and does not need expectMore_), and a new class that will get the existing three throwing "description" methods (plus expectMore_). Alternatively, that "new class" code can be merged with BinaryTokenizer -- each tokenizer method can be designed to treat input as "text" or "binary", as appropriate for that method. The primary design difference between BinaryTokenizer and classic Tokenizer are insufficient input handling and commit points, not "text" vs. "binary"!

IMO, all these changes/decisions should be done in a dedicated minimal PR because they affect a lot of code unrelated to this PR _and_ represent an elevated danger for that code.

----

I kind of hope that this PR does not need any of the above changes, but if you think it does, then we need to work on a prerequisite PR that provides the missing expectMore_ pieces. To avoid a lot of additional difficult work that might be unnecessary for this PR, I suggest the following plan:

1. If you can improve Decode() by using a Tokenizer that supports expectMore_, then write Decode() as if Tokenizer already supports expectMore_. The code may not compile, but that is OK at this point.
2. We review/finalize Decode() implementation sketch. It may not compile, but that is still OK at this point.
3. If, by this time, Decode() still needs Tokenizer that supports expectMore_, then we open a prerequisite PR that adds that support.

P.S. As I mentioned during the meeting, we are also adding a few expectMore-related methods to BinaryTokenizer in a internal Factory PR we are getting ready for the official review. I do not think this PR needs them, so I am not going to make matters more complex by suggesting them in this PR at this point.

kinkie:
> adding a boolean Tokenizer data member to mark input that will not grow in the future

I do not think we need this concept for this PR. It might be needed in the future, but I am aware of no use case where it could be needed. Main use cases for rfc1738_unescape are either in helper protocols or in FTP, where there is external framing

src/anyp/Uri.cc
kinkie: Adopted. Thanks!

kinkie: Adopted. Thanks!

src/anyp/Uri.cc
kinkie: Adopted.

src/anyp/Uri.cc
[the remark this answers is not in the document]
kinkie: Done

yadij:
The `Tokenizer::expectMore` mechanism will allow easy removal of some if-statements from the loop.
If @rousskov is happy to allow those if-statements as "good enough" for now, then this PR does not need to wait for that change PR (it is out of scope here).

yadij (dismissed):
LGTM, given current `Tokenizer` ability.

yadij:
ping @rousskov .

AFAICT all issues have been resolved to "good enough" standard. Absent any feedback I expect to clear this for merge under the 10-day criteria in 2-3 days.

rousskov:
> AFAICT all issues have been resolved to "good enough" standard. Absent any feedback I expect to clear this for merge under the 10-day criteria in 2-3 days.

I will review this PR as soon as I can, but it may not happen until next week. While you can clear this PR for merging at any time, it would still need to pass my review to actually merge (due to the current negative vote).

src/anyp/Uri.cc
```diff
@@ -81,6 +84,45 @@ AnyP::Uri::Encode(const SBuf &buf, const CharacterSet &ignore)
...
+}
+
+SBuf
+AnyP::Uri::Decode(const SBuf &buf)
+{
+    if (buf.isEmpty())
+        return buf;
+
+    SBuf output;
+    output.reserveSpace(buf.length()); // worst case: buf is not encoded
```
rousskov, line 100: AFAICT, this optimization code attempts to reduce the number of memory (re)allocations in the parsing loop below, but that attempt fails badly on valid input that has no percent characters. Does not such input represent a _dominant_ use case (at least for some of the expected callers)?

rousskov (commented):
Excellent progress, thank you. I think we are almost done with this PR!

src/anyp/Uri.cc
kinkie: We don't know, or control it.
The current behaviour mimics the one of the function it aims to replace, but it's also possible that SBuf's heuristics work well enough, at the cost of multiple data copies on encoded-heavy strings.
An optimization that might be interesting is to flip a flag if there is any decoding taking place at all, and if there is none just return the input SBuf. This could help with small optimizations elsewhere (e.g. SBuf::operator =), or by a slightly increased probability of sharing the backing store

rousskov: > We don't know, or control it.

That sounds like an argument for _not_ adding potentially expensive optimizations!

> The current behaviour mimics the one of the function it aims to replace.

Which one? rfc1738_unescape() does not have this behavior.

> Why do you say it fails badly in case of no percent characters?

I say that because the proposed reserveSpace() call leads to an extra buffer allocation and an extra memcpy() in those cases. Without that "optimization", SBuf::append() does not allocate and does not copy the buffer when appending to virgin SBufs. It uses cheap reference counting instead.

I removed that "optimization" in commit 6d70432, addressing this change request.

> if anything we shouldn't call this "worst case"

Yes, that phrasing was not ideal. It is now gone (as a side effect of other changes).

> One possible optimization could be to have a bool switch set to false once per decoding loop, and flipped to true if an encoded triplet is found. Then we can use this information to return the input SBuf instead of the output one. We would still have iterated and copied to output, but thus would allow some optimizations downstream.
> 
> Alternatively, we could special-case and optimize the not-encoded case by adding this code block before the decoding loop, this should essentially make the not-encoded case zero-copy (this has been tested and works):
> 
> ```
>     if (tok.prefix(token, unencodedOctets)) { //[1]
>         if (tok.atEnd())
>             return buf;
>         else
>             output.append(token); //[2]
>     }
> ```
> 
> On top of this, another possible optimization (at the cost of complexity) is to have the output.reserveSpace call happen on just before line marked [2] here, and on the else-branch of the condition at [1] which is needed because if the first character of the encoded string is %, prefix() will return false. This would make the not-encoded case zero-copy and minimal-allocation

I believe the current (as of branch commit 28dc639) implementation is simpler and at least as efficient as the ones discussed above. If your tests reveal unexpected allocations or copies in the no-percent cases, let's revisit this topic.

src/anyp/Uri.cc
rousskov: > Amos: Please just have the callers pass the CharacterSet they want encoded. ... I think it will be not common enough, and touching the encoder should be out of scope for this PR. If I am wrong about the utility it and/or other wrappers can be easily added in a followup PR.

I agree.

> Francesco: Done.

I did not find any signs that the changes requested by Amos were actually implemented. Now done (branch commit 65eeca9).

rousskov (approved):
I am approving this PR after fixing one minor branch bug (commit 01966b0), removing unwanted optimization, removing [questionable](https://github.com/squid-cache/squid/pull/1626/#discussion_r1436715111) Rfc3986Encode(), and doing some polishing (including adjusting tests and PR title/description).

I am not clearing this PR for merging to give others a chance to review my changes.

yadij (approved):

PR #1628, opened 2024-01-05
maintenance: remove testPreCompiler

kinkie:
Precompilers are a very well established and tested technology,
remove this test which provides very little signal


rousskov (approved):
I agree that we should not test compiler support for basic C++ features, including support for basic preprocessor features.

yadij:
IFAIK this was (is?) useful for catching issues with cross-build environments. Even though we no longer support builds on the OS this was created for, the cross-builds still sometimes have issues with incorrectly defined `USE_*` and `HAVE_*` macros.

kinkie:
> IFAIK this was (is?) useful for catching issues with cross-build environments. Even though we no longer support builds on the OS this was created for, the cross-builds still sometimes have issues with incorrectly defined `USE_*` and `HAVE_*` macros.

Sure, but what's tested here is that #ifdef and #if and #undef work as expected, not that USE_ and HAVE_ are set correctly.

Ten years have passed since squid3 (which is mentioned in the comments), and since then the variety of compilers and OSes we target has decreased quite substantially, and for those we have significant evidence that they work.
Furthermore, I don't think we had a meaningful failure of this test in years (but on the opposite side it's an opportunity for flakiness)

PR #1629, opened 2024-01-05
Activate OpenSSL KTLS for SSL Bump

ss3git:
Adding a build option "--with-openssl-ktls" which activates OpenSSL KTLS function (SSL_OP_ENABLE_KTLS).

This code is tested to work using squid v6.6 on FreeBSD 13.2 with OpenSSL 1.1.1t-freebsd  or  OpenSSL 3.1.4.

```
Requirements (test environment):
 FreeBSD kernel with KERN_TLS enabled
 OpenSSL built with KTLS enabled

% kldload ktls_ocf.ko
% sysctl kern.ipc.tls.enable=1


Results after some ssl bump traffic:

% sysctl -a | grep ipc.tls.stats.ocf
kern.ipc.tls.stats.ocf.retries: 0
kern.ipc.tls.stats.ocf.separate_output: 0
kern.ipc.tls.stats.ocf.inplace: 238607
kern.ipc.tls.stats.ocf.tls13_chacha20_encrypts: 32
kern.ipc.tls.stats.ocf.tls13_chacha20_decrypts: 162
kern.ipc.tls.stats.ocf.tls13_gcm_encrypts: 222917
kern.ipc.tls.stats.ocf.tls13_gcm_decrypts: 284010
kern.ipc.tls.stats.ocf.tls12_chacha20_encrypts: 265
kern.ipc.tls.stats.ocf.tls12_chacha20_decrypts: 1021
kern.ipc.tls.stats.ocf.tls12_gcm_encrypts: 15064
kern.ipc.tls.stats.ocf.tls12_gcm_decrypts: 14736
kern.ipc.tls.stats.ocf.tls11_cbc_encrypts: 329
kern.ipc.tls.stats.ocf.tls10_cbc_encrypts: 0

### DIFF
--- a/src/security/Session.cc
+++ b/src/security/Session.cc
@@ -136,6 +136,10 @@ CreateSession(const Security::ContextPointer &ctx, const Comm::ConnectionPointer
         const int fd = conn->fd;
 
 #if USE_OPENSSL
+		#if USE_OPENSSL_KTLS
+	    SSL_set_options(session.get(), SSL_OP_ENABLE_KTLS);
+	    #endif
+    	
         // without BIO, we would call SSL_set_fd(ssl.get(), fd) instead
         if (BIO *bio = Ssl::Bio::Create(fd, type)) {
             Ssl::Bio::Link(session.get(), bio); // cannot fail
--- a/src/ssl/bio.cc
+++ b/src/ssl/bio.cc
@@ -80,6 +80,10 @@ Ssl::Bio::Create(const int fd, Security::Io::Type type)
 
     if (BIO *bio = BIO_new(useMethod)) {
         BIO_int_ctrl(bio, BIO_C_SET_FD, type, fd);
+        #if USE_OPENSSL_KTLS
+        BIO *bio_sock = BIO_new_socket(fd, BIO_NOCLOSE);
+        bio = BIO_push(bio, bio_sock);
+        #endif
         return bio;
     }
     return nullptr;
@@ -108,7 +112,11 @@ int Ssl::Bio::write(const char *buf, int size, BIO *table)
 #if _SQUID_WINDOWS_
     const int result = socket_write_method(fd_, buf, size);
 #else
+    #if USE_OPENSSL_KTLS
+    const int result = BIO_write(BIO_next(table), buf, size);
+    #else
     const int result = default_write_method(fd_, buf, size);
+    #endif
 #endif
     const int xerrno = errno;
     debugs(83, 5, "FD " << fd_ << " wrote " << result << " <= " << size);
@@ -131,7 +139,11 @@ Ssl::Bio::read(char *buf, int size, BIO *table)
 #if _SQUID_WINDOWS_
     const int result = socket_read_method(fd_, buf, size);
 #else
+    #if USE_OPENSSL_KTLS
+    const int result = BIO_read(BIO_next(table), buf, size);
+    #else
     const int result = default_read_method(fd_, buf, size);
+    #endif
 #endif
     const int xerrno = errno;
     debugs(83, 5, "FD " << fd_ << " read " << result << " <= " << size);
@@ -286,7 +298,7 @@ Ssl::ServerBio::readAndGive(char *buf, const int size, BIO *table)
         return giveBuffered(buf, size);
 
     if (record_) {
-        const int result = readAndBuffer(table);
+        const int result = readAndBuffer(table, size);
         if (result <= 0)
             return result;
         return giveBuffered(buf, size);
@@ -300,7 +312,7 @@ Ssl::ServerBio::readAndGive(char *buf, const int size, BIO *table)
 int
 Ssl::ServerBio::readAndParse(char *buf, const int size, BIO *table)
 {
-    const int result = readAndBuffer(table);
+    const int result = readAndBuffer(table, size);
     if (result <= 0)
         return result;
 
@@ -308,7 +320,11 @@ Ssl::ServerBio::readAndParse(char *buf, const int size, BIO *table)
         if (!parser_.parseHello(rbuf)) {
             // need more data to finish parsing
             BIO_set_retry_read(table);
+            #if USE_OPENSSL_KTLS
+            return giveBuffered(buf, size);
+            #else
             return -1;
+            #endif
         }
         parsedHandshake = true; // done parsing (successfully)
     }
@@ -324,10 +340,15 @@ Ssl::ServerBio::readAndParse(char *buf, const int size, BIO *table)
 /// Reads more data into the read buffer. Returns either the number of bytes
 /// read or, on errors (including "try again" errors), a negative number.
 int
-Ssl::ServerBio::readAndBuffer(BIO *table)
+Ssl::ServerBio::readAndBuffer(BIO *table, const int size)
 {
+    #if USE_OPENSSL_KTLS
+    char *space = rbuf.rawAppendStart(size);
+    const int result = Ssl::Bio::read(space, size, table);
+    #else
     char *space = rbuf.rawAppendStart(SQUID_TCP_SO_RCVBUF);
     const int result = Ssl::Bio::read(space, SQUID_TCP_SO_RCVBUF, table);
+    #endif
     if (result <= 0)
         return result;
 
@@ -549,7 +570,11 @@ squid_bio_ctrl(BIO *table, int cmd, long arg1, void *arg2)
         case BIO_CTRL_WPENDING:
     */
     default:
+        #if USE_OPENSSL_KTLS
+        return BIO_ctrl(BIO_next(table), cmd, arg1, arg2);
+        #else
         return 0;
+        #endif
 
     }
 
--- a/configure.ac
+++ b/configure.ac
@@ -1110,6 +1110,7 @@ SQUID_CHECK_LIB_WORKS(gnutls,[
 dnl User may specify OpenSSL is needed from a non-standard location
 SQUID_OPTIONAL_LIB(openssl,[OpenSSL],[LIBOPENSSL])
 AH_TEMPLATE(USE_OPENSSL,[OpenSSL support is available])
+AH_TEMPLATE(USE_OPENSSL_KTLS,[OpenSSL Kernel TLS Offload])
 ## OpenSSL is default disable due to licensing issues on some OS
 AS_IF([test "x$with_openssl" = "xyes"],[
   CPPFLAGS="$LIBOPENSSL_CFLAGS $CPPFLAGS"
@@ -1169,6 +1170,10 @@ AS_IF([test "x$with_openssl" = "xyes"],[
     CXXFLAGS="$LIBOPENSSL_CFLAGS $CXXFLAGS"
     SSLLIB="$LIBOPENSSL_PATH $LIBOPENSSL_LIBS $SSLLIB"
     AC_DEFINE(USE_OPENSSL,1,[OpenSSL support is available])
+    
+    AS_IF([ test "x$with_openssl_ktls" = "xyes" ],[
+      AC_DEFINE(USE_OPENSSL_KTLS,1,[Use OpenSSL Kernel TLS Offload])
+    ])
 
     # check for API functions
     SQUID_CHECK_LIBCRYPTO_API
--- a/src/cf_gen_defines
+++ b/src/cf_gen_defines
@@ -43,6 +43,7 @@ BEGIN {
 	define["USE_IDENT"]="--enable-ident-lookups"
 	define["USE_LOADABLE_MODULES"]="--enable-shared"
 	define["USE_OPENSSL"]="--with-openssl"
+	define["USE_OPENSSL_KTLS"]="--with-openssl-ktls"
 	define["USE_QOS_TOS"]="--enable-zph-qos"
 	define["USE_SQUID_ESI"]="--enable-esi"
 	define["USE_SQUID_EUI"]="--enable-eui"
--- a/src/ssl/bio.h
+++ b/src/ssl/bio.h
@@ -172,7 +172,7 @@ public:
 private:
     int readAndGive(char *buf, const int size, BIO *table);
     int readAndParse(char *buf, const int size, BIO *table);
-    int readAndBuffer(BIO *table);
+    int readAndBuffer(BIO *table, const int size);
     int giveBuffered(char *buf, const int size);
 
     /// SSL client features extracted from ClientHello message or SSL object

### CODE REVIEW

File: src/ssl/bio.cc [hunk 1]
```diff
@@ -300,15 +312,19 @@ Ssl::ServerBio::readAndGive(char *buf, const int size, BIO *table)
...
+    const int result = readAndBuffer(table, size);
     if (result <= 0)
         return result;
 
     try {
         if (!parser_.parseHello(rbuf)) {
             // need more data to finish parsing
             BIO_set_retry_read(table);
+            #if USE_OPENSSL_KTLS
+            return giveBuffered(buf, size);
```
rousskov, line 324: The old/official code suggests that we do not want to give TLS libraries (whatever they are) the raw bytes until we parse them.  PR code does not seem to match that logic. Could you please explain, in a GitHub comment (for now), why we change the logic here if USE_OPENSSL_KTLS?

src/security/Session.cc
```diff
@@ -136,6 +136,10 @@ CreateSession(const Security::ContextPointer &ctx, const Comm::ConnectionPointer
         const int fd = conn->fd;
 
 #if USE_OPENSSL
+		#if USE_OPENSSL_KTLS
+	    SSL_set_options(session.get(), SSL_OP_ENABLE_KTLS);
```
rousskov, line 140: Security::PeerOptions::updateSessionOptions() also sets options. If SSL_OP_ENABLE_KTLS option can be set by adding a new ssl_options entry (and, hence, recognizing a new peer-specific setting in squid.conf), please do so instead of hard-coding this setting (for all peers)?

If the above or a similar plan works, we may be even be able to avoid introducing USE_OPENSSL_KTLS!

rousskov (changes requested):
Thank you very much for working on adding KTLS support! This work may require several iterations, but I hope this incomplete review helps you make steady progress.

src/ssl/bio.cc
```diff
@@ -80,6 +80,10 @@ Ssl::Bio::Create(const int fd, Security::Io::Type type)
 
     if (BIO *bio = BIO_new(useMethod)) {
         BIO_int_ctrl(bio, BIO_C_SET_FD, type, fd);
+        #if USE_OPENSSL_KTLS
+        BIO *bio_sock = BIO_new_socket(fd, BIO_NOCLOSE);
```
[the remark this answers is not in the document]
ss3git: SSL_free() seems to call BIO_free_all() for linked BIOs.

https://github.com/squid-cache/squid/blob/ee7240ff8b1548eb76853711ea64ea7343cba13e/src/security/Session.cc#L94

https://github.com/openssl/openssl/blob/856d4c6ebc1aec2a714ce497af9c3e48c50e689d/ssl/ssl_lib.c#L1209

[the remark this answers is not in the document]
rousskov: > SSL_free() seems to call BIO_free_all() for linked BIOs.

Aha, that fact (together with the fact that we currently always call Ssl::Bio::Link() after Ssl::Bio::Create(), linking the BIO object to the SSL object) explains why I cannot find any _relevant_ BIO_free() calls in the current code. Thank you. I am resolving this change request.

src/ssl/bio.cc
ss3git: Reverted: this change was necessary at the early stage of prototyping to successfully establish the connection, but it seems to be no longer required.

https://github.com/squid-cache/squid/pull/1629/commits/d279eaec8649a6d293e6ff5687ac560d86a360d9

ss3git: Just reverting this may not be a good idea.

When using KTLS, perhaps the upper layer OpenSSL function (SSL_*) does not want this Ssl::ServerBio::read function to read more data from the socket (file descriptor) than the OpenSSL function is expecting because at some specific point the OpenSSL function changes the socket option.

https://github.com/openssl/openssl/blob/5963aa8c196d7c5a940a979299a07418527932af/include/internal/ktls.h#L70C3-L70C3


I cannot reproduce the issue that I was encountering while prototyping though, depending on the timing of packet arrival, the current (old/official) code potentially read more than expected data from the socket, which has been OK for non-ktls session but not for ktls.

Perhaps we have to consider some alternative way.

src/security/Session.cc
ss3git: Just reverting this may not be a good idea.

When using KTLS, perhaps the upper layer OpenSSL function (SSL_*) does not want this Ssl::ServerBio::read function to read more data from the socket (file descriptor) than the OpenSSL function is expecting because at some specific point the OpenSSL function changes the socket option.

https://github.com/openssl/openssl/blob/5963aa8c196d7c5a940a979299a07418527932af/include/internal/ktls.h#L70C3-L70C3


I cannot reproduce the issue that I was encountering while prototyping though, depending on the timing of packet arrival, the current (old/official) code potentially read more than expected data from the socket, which has been OK for non-ktls session but not for ktls.

Perhaps we have to consider some alternative way.

src/ssl/bio.cc
ss3git: Just reverting this may not be a good idea.

When using KTLS, perhaps the upper layer OpenSSL function (SSL_*) does not want this Ssl::ServerBio::read function to read more data from the socket (file descriptor) than the OpenSSL function is expecting because at some specific point the OpenSSL function changes the socket option.

https://github.com/openssl/openssl/blob/5963aa8c196d7c5a940a979299a07418527932af/include/internal/ktls.h#L70C3-L70C3


I cannot reproduce the issue that I was encountering while prototyping though, depending on the timing of packet arrival, the current (old/official) code potentially read more than expected data from the socket, which has been OK for non-ktls session but not for ktls.

Perhaps we have to consider some alternative way.

ss3git: https://github.com/squid-cache/squid/pull/1629/commits/ba88d4a6af19628e74a966c25897cb52e71b4c68

This commit is obviously not for release, but (I guess) something to limit the read size like this would be needed to ensure ktls establishment if keeping the logic (we do not want to give TLS libraries the raw bytes until we parse them) is mandatory. (Is it for DoS attack prevention or something?)

rousskov: > Is it for DoS attack prevention or something?

Unfortunately, I do not know the answer. I can try to research it, but it will take time. Right now, I can speculate that the primary goal here is _not_ attack prevention but something like making sure we can still splice the TCP connection (i.e. preventing OpenSSL _sending_ anything on the socket) if necessary. It is also possible that we wanted to finish configuring OpenSSL session state (using parsed Hello information and squid.conf settings) as necessary.

Without knowing what this old code does (or was supposed to do), I cannot fully review your changes, including the sketch in ba88d4a6af19628e74a966c25897cb52e71b4c68.

ss3git: I've changed the code to PEEK the socket while getting Hello information.
https://github.com/squid-cache/squid/pull/1629/commits/d1c9d68d79f67154c1048c95b6f8f114b4b96f38

Hopefully this satisfies both of correct ktls establishment and "not-giving the raw bytes to the upper layer till Hello."

Actually it seems to be working on FreeBSD, however, I was not sure if it also works on every platform that squid supports so reintroduced the build switch macro in the later commits.

src/security/Session.cc
ss3git: supported options=ENABLE_KTLS in squid.conf

https://github.com/squid-cache/squid/pull/1629/commits/14c81ec77ce855444453a6b2d822da7f549d4acd

src/ssl/bio.cc
```diff
@@ -131,7 +139,11 @@ Ssl::Bio::read(char *buf, int size, BIO *table)
 #if _SQUID_WINDOWS_
     const int result = socket_read_method(fd_, buf, size);
 #else
+    #if USE_OPENSSL_KTLS
```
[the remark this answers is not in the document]
ss3git: This topic might be outdated because of refactor.

src/ssl/bio.cc
```diff
@@ -80,6 +80,10 @@ Ssl::Bio::Create(const int fd, Security::Io::Type type)
 
     if (BIO *bio = BIO_new(useMethod)) {
         BIO_int_ctrl(bio, BIO_C_SET_FD, type, fd);
+        #if USE_OPENSSL_KTLS
+        BIO *bio_sock = BIO_new_socket(fd, BIO_NOCLOSE);
+        bio = BIO_push(bio, bio_sock);
```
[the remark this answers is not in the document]
ss3git: Done, but also refactored..

src/ssl/bio.cc
```diff
@@ -108,7 +112,11 @@ int Ssl::Bio::write(const char *buf, int size, BIO *table)
 #if _SQUID_WINDOWS_
     const int result = socket_write_method(fd_, buf, size);
 #else
+    #if USE_OPENSSL_KTLS
+    const int result = BIO_write(BIO_next(table), buf, size);
```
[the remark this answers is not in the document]
ss3git: Done, but also refactored..

include/squid.h
```diff
@@ -61,5 +61,11 @@ using namespace Squid;
 
 #define LOCAL_ARRAY(type, name, size) static type name[size]
 
+#if USE_OPENSSL
+#if _SQUID_FREEBSD_
+#define OPENSSL_KTLS_SUPPORT 1
+#endif
+#endif
+
```
yadij, line 69: PR description indicates there is a `./configure` option added. But no logic change adds that option.

Either the option should be added and this chunk dropped entirely.
OR this chunk moved to `compat/os/freebsd.h` since the change is OS-specific to FreeBSD, with PR description updated to correctly state what is being added.

yadij (changes requested):
There are a lot of minor C-style code that need to be polished yup to the C++ style we use in Squid. Please let me know when this PR is ready for a final polishing review.

For now, there is one portability changed needed.

src/ssl/bio.cc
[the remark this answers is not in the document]
yadij: AFAICS, as of [124aa58](https://github.com/squid-cache/squid/pull/1629/commits/124aa58b201c3d25c0acccdc31178606f1cb820c) the refactor broke these things even worse than before.  Now when the feature is enabled-but-not-configured on Windows the entirely wrong `default_*_method` gets called instead of `socket_*_method`.

src/ssl/bio.cc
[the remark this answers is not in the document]
yadij: Issues remains, see discussion on `Ssl::Bio::read`. Whatever solution is used for that needs to also be applied here symmetrically.

include/squid.h
ss3git: Hi.

My first intention in this PR was to add the configurable option "--with-openssl-ktls" to enable "USE_OPENSSL_KTLS" (for every session), so I had changed configure.ac etc.:
 https://github.com/squid-cache/squid/pull/1629/commits/1f7e2da4cebed8a32f101245facfbe21689d01fa

However, rousskov suggested me, if logically possible, that not adding the build option was desireble, and instead it should be just configurable (e.g. http_port and tls_outgoing_options):
 https://github.com/squid-cache/squid/pull/1629#discussion_r1443285736

So, I cancelled (reverted) configure.ac or related changes, and introduced OPENSSL_KTLS_SUPPORT (which is not user-configurable but OS-dependent) to separate codes for KTLS enabled or disabled platforms.


Regarding "OS-specific to FreeBSD," it's because I can test only on the platform (at least now).
Linux, for example, may also be able to support (can define OPENSSL_KTLS_SUPPORT) with or without modifications of code in the future, if some one could test it.
Even though, if changing the place to define OPENSSL_KTLS_SUPPORT or renaming it is desired, there is no reason not to do so.

ss3git:
Hi all,

Just letting you know that I don't have any planned commit now.

Unfortunately, my holiday will be over soon, so I may not be able to make a quick response hereafter. If you have questions asking my original intention for example, I will try to answer to them as possible, but for other things like refactoring requests, I cannot say I will do soon. (Please go ahead and change the code for that if anyone can do.)

src/ssl/bio.cc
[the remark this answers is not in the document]
yadij: Yes, the code should account for Windows possibly adding support. 
You could simplify the code a bit too with precompiler blocks, like so:
```
    int result;
#if OPENSSL_KTLS_SUPPORT
    if (BIO_next(table))
        result = BIO_write(BIO_next(table), buf, size);
    else
#endif
#if _SQUID_WINDOWS_
    result = socket_write_method(fd_, buf, size);
#else
    result = default_write_method(fd_, buf, size);
#endif
```

yadij:
Updated commit description to clarify latest change and pass Anubis requirements.

Testing details dropped:
```
Requirements (test environment):
 FreeBSD kernel with KERN_TLS enabled
 OpenSSL built with KTLS enabled

% kldload ktls_ocf.ko
% sysctl kern.ipc.tls.enable=1

Results after some ssl bump traffic:

% sysctl -a | grep ipc.tls.stats.ocf
kern.ipc.tls.stats.ocf.retries: 0
kern.ipc.tls.stats.ocf.separate_output: 0
kern.ipc.tls.stats.ocf.inplace: 238607
kern.ipc.tls.stats.ocf.tls13_chacha20_encrypts: 32
kern.ipc.tls.stats.ocf.tls13_chacha20_decrypts: 162
kern.ipc.tls.stats.ocf.tls13_gcm_encrypts: 222917
kern.ipc.tls.stats.ocf.tls13_gcm_decrypts: 284010
kern.ipc.tls.stats.ocf.tls12_chacha20_encrypts: 265
kern.ipc.tls.stats.ocf.tls12_chacha20_decrypts: 1021
kern.ipc.tls.stats.ocf.tls12_gcm_encrypts: 15064
kern.ipc.tls.stats.ocf.tls12_gcm_decrypts: 14736
kern.ipc.tls.stats.ocf.tls11_cbc_encrypts: 329
kern.ipc.tls.stats.ocf.tls10_cbc_encrypts: 0
```

yadij:
@rousskov, please re-review latest code after my changes.

yadij (approved):

src/ssl/bio.cc
```diff
@@ -80,6 +80,12 @@ Ssl::Bio::Create(const int fd, Security::Io::Type type)
 
     if (BIO *bio = BIO_new(useMethod)) {
         BIO_int_ctrl(bio, BIO_C_SET_FD, type, fd);
+        if (enable_ktls) {
+#if OPENSSL_KTLS_SUPPORT
+            if (auto *bio_sock = BIO_new_socket(fd, BIO_NOCLOSE))
```
rousskov, line 85: Why do we need to add this BIO? In other words, why do we need _two_ custom descriptor-reading BIOs instead of one?

src/security/PeerOptions.cc
```diff
@@ -801,6 +806,19 @@ Security::PeerOptions::updateSessionOptions(Security::SessionPointer &s)
...
 }
 
+#if OPENSSL_KTLS_SUPPORT
+void
+Security::PeerOptions::updateSessionOptionsOnlyKtls(Security::SessionPointer &s)
+{
+    parseOptions();
+    
+    debugs(83, 5, "set OpenSSL options (KTLS only) for session=" << s << ", Options=" << (parsedOptions & SSL_OP_ENABLE_KTLS));
+    SSL_clear_options(s.get(), SSL_OP_ENABLE_KTLS);
```
rousskov, line 816: Why do we need to clear options before setting them in this context, but not in other contexts where SSL_set_options() is called? PR commit 124aa58b201c3d25c0acccdc31178606f1cb820c message suggests that this was done "to support squid -k reconfigure", but reconfiguration should not affect _existing_ SSL objects. I have not checked, but I would expect this `s` to be a new SSL object, created after Squid started to reconfigure itself.

rousskov (changes requested):

src/ssl/bio.cc
ss3git: I have only a little time, so short answer is:

OpenSSL's KTLS implementations are hidden as static here:
https://github.com/openssl/openssl/blob/01088602522015906654877ad2730ce805f3f925/crypto/bio/bss_sock.c#L38

- Then BIO_new_socket() (or BIO_s_socket()) is needed if squid wants to use SSL_*, is what I thought.
https://github.com/openssl/openssl/blob/01088602522015906654877ad2730ce805f3f925/crypto/bio/bss_sock.c#L58

src/security/PeerOptions.cc
ss3git: > "but I would expect this s to be a new SSL object, created after Squid started to reconfigure itself"

This was what I thought (expected) at first too, but actually, the test result was different.
I suspect some underlying issue around here (either in squid or openssl).

src/ssl/bio.cc
```diff
@@ -324,17 +346,83 @@ Ssl::ServerBio::readAndParse(char *buf, const int size, BIO *table)
...
     return result;
 }
 
+#if OPENSSL_KTLS_SUPPORT
+/// Read and give everything to our parser. (KTLS support ver.)
+/// When/if parsing is finished (successfully or not), start giving to OpenSSL.
+///
+/// \note If using KTLS, we must not consume more socket data than the caller (OpenSSL) expects.
+/// Otherwise, the session establishment potentially fails depending on the timing of packet arrival.
+/// Therefore, we peek the socket here and ensure we do not consume the socket data more than required.
```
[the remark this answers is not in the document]
ss3git: **This comment is explaining what I have understood so far and it might be wrong.**

> "all KTLS-supporting implementations must peek at the socket"

This part is perhaps not true.
General implementation just needs to use SSL_* with the regular BIO.
Squid might be some exception because it has to intercept the traffic using the wrapper BIO.
I agree that the code is a kind of workaround to satisfy squid's current requirement.

I was searching if there is any technical reference.
So far, what I could find is only this one that is telling reading the socket after negotiation might be undefined behaviour in FreeBSD.

https://gist.github.com/vladak/8fc4bb65f68a55eb98630b5ab5c6a4b9

> receive needs to be done using recvmsg(2)
> Once TLS receive is enabled by a successful set of the TCP_RXTLS_ENABLE socket option, all data read from the socket is returned as decrypted TLS records. Each received TLS record must be read from the socket using recvmsg(2). Each received TLS record will contain a TLS_GET_RECORD control message along with the decrypted payload. The control message contains a struct tls_get_record which includes fields from the TLS record header. If an invalid or corrupted TLS record is received, recvmsg(2) will fail with one of the following errors...
> 
> TODO: what happens if the app calls read(2) on the socket ?

include/squid.h
```diff
@@ -61,5 +61,11 @@ using namespace Squid;
 
 #define LOCAL_ARRAY(type, name, size) static type name[size]
 
+#if USE_OPENSSL
+#if _SQUID_FREEBSD_
+#define OPENSSL_KTLS_SUPPORT 1
```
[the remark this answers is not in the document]
ss3git: I understand what you say. I wanted to minimize degradation risks from my standing point.

As you suggested (and I agreed) in another comment, analysis of lower layer implmenentation will also be necessary.
If leaving unclosed PR is difficult, I guess this one should be rejected for now, until someone (specifically Linux user) can take over it sometime later.

Thank you for reviewing at any rate.

src/ssl/bio.cc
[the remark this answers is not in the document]
rousskov: I found some related OpenSSL code and [discussion](https://github.com/openssl/openssl/pull/7848#discussion_r245650388). AFAICT, the situation is more nuanced (and arguably worse) than I expected:

* OpenSSL has a `read_ahead` mode that enables reading more socket bytes than OpenSSL needs to parse the next TLS element. If that mode is enabled, OpenSSL will usually request more socket bytes than the handshake. I speculate that `read_ahead` applies to handshake reading/parsing. Thus, our attempts to prevent "read past the handshake" by relying on OpenSSL to tell us how much to reach will fail.
* Since 2022, OpenSSL [enables](https://github.com/openssl/openssl/blob/0a22436ea5826d0089db7f1cd97b7c90135ca165/ssl/record/methods/ktls_meth.c#L439C14-L442) read_ahead mode when KTLS is in use: "With KTLS we always try and read as much as possible and fill the buffer". Thus, our attempts to prevent "read past the handshake" appear to work against OpenSSL desires.
* Kernel KTLS API (SOL_TLS, tls12_crypto_info_aes_gcm_128 and friends) probably does not provide the ability to report extra bytes read to the kernel (extra bytes read while parsing the handshake), especially bytes from an incomplete TLS record. The kernel may assume that it starts reading from the TLS record boundary.

Combining all of the above, I _speculate_ that OpenSSL will often attempt to read more bytes than the size of TLS handshake (while processing that handshake and before giving decryption control to the kernel) and, hence, will deal with any extra bytes read. When KTLS is enabled for a TLS connection, OpenSSL may assume that any "extra" bytes will not split a TLS record (because initial TLS records are usually not split across socket writes or reads?).

I failed to find OpenSSL documentation that outlines requirements for KTLS-compatible BIOs.

Based on the above, I hope we do not need to treat KTLS I/Os specially in our custom BIO, but, AFAICT, there is currently not enough information available to confirm or reject that conclusion.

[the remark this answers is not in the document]
rousskov: > reading the socket after negotiation might be undefined behaviour

Yes, that makes a lot of sense: There can be only one socket reader in this context. Once kernel starts reading the socket, the user must not.

What we do not know is whether it is OK for BIO, _during negotiation_, to read more bytes than OpenSSL has asked for. I suspect the answer depends on whether OpenSSL has code to get the buffered data from our BIO (kind of "flushing" our read BIO buffers -- there is OpenSSL code to flush write BIO buffers IIRC). I suspect that OpenSSL code does not exist and, hence, we cannot read more than OpenSSL asked (even if OpenSSL itself asks to read more than the handshake!) because those buffered bytes will be effectively lost, breaking TLS.

src/ssl/bio.cc
rousskov: Thank you for providing those details. Unfortunately, they are not enough for me to connect the dots and answer the question in this change request.

ss3git: OK, I know I jumped into there.

This is related to what I prototyped first, so I would like to explain step-by-step.

Firstly, I share a thread I read as the starting point.
https://github.com/openssl/openssl/issues/14595

Here, mattcaswell mentions:

> KTLS will be used if:
>
> OpenSSL has been compiled with the "enable-ktls" option
> KTLS has not been explicitly disabled at run time via SSL_MODE_NO_KTLS_TX and SSL_MODE_NO_KTLS_RX
> You are using a suitable KTLS aware BIO (BIO_s_connect(), or BIO_s_socket())
> KTLS is compatible with the ciphersuites, TLS protocol versions and extensions that have been negotiated (this may vary from one > > > platform to another based on kernel support)

In short, using the BIO_METHOD(s) that BIO_s_connect() or BIO_s_socket() offers is the way to activate KTLS. Also, I thought BIO_s_socket() should be the appropriate one that Squid can use to intercept traffic.

Note:
 - The type BIO_METHOD (a.k.a. struct bio_method_st) is defined in OpenSSL internal header.
 - BIO_new_socket() that returns "BIO *" which is internally linked to the KTLS-aware BIO_METHOD.
 - SSL_set_fd()/SSL_set_rfd()/SSL_set_wfd() also internally generates a BIO which links to the KTLS-aware BIO_METHOD.



Based on the knowledge, I did some prototyping.

The result was:
 1. apply SSL_set_fd(fd), and not link the squid-wrapper BIO -- failed.
 2. apply SSL_set_wfd(fd), and only link the squid-wrapper BIO for receive side by SSL_set0_rbio(ssl, fd): (partially) succeeded.
 3. opposite of 2 -- failed.

Note:
 - fd is the file descriptor which had been used by squid.
 - No.2 offloads only tx side. Also, it should have other issues that I didn't realize at that time.



I started to dig into the receive side. The resulting point was the currently concerned function Ssl::ServerBio::read().

Using BIO_new_socket() and chaining the returned BIO to the squid-wrapper BIO, with some tuning of Ssl::ServerBio::read(), should be the way to go; was the conclusion of my observation.

and thereafter, probably you know.


Regarding direct answer to your comment, I do not exclude the possibility that the squid-wrapper includes everything the KTLS-aware BIO_METHOD offers, but I concern that results in Squid re-implements every related function of OpenSSL. (I may be wrong of course.)

rousskov: Your comment was very helpful, thank you! I do not yet understand whether any custom BIO can be considered a "suitable KTLS aware BIO" by OpenSSL. Do you know what specific checks/tests OpenSSL is using to decide whether the configured BIO (chain) is "suitable"?

FWIW, I understand that we can create and chain two BIOs, one custom (via BIO_new()) and one via BIO_new_socket(), like this PR is going, and that chain will be (allegedly, in your tests) considered "KTLS aware BIO" by OpenSSL.

If creating a single "suitable" custom BIO is not practical for some reason, then another alternative to consider is complete removal of Squid custom BIO: If Squid can parse the handshake by peeking at the socket (while being called from OpenSSL), then Squid ought to be able to parse the handshake by peeking at the socket without any OpenSSL calls/guidance (and then create OpenSSL connection without any custom BIO as/if needed).

ss3git: > Do you know what specific checks/tests OpenSSL is using to decide whether the configured BIO (chain) is "suitable"?

Not directly answering to your question, but this change in squid_bio_ctrl() is the one to make the entire chain of BIOs "KTLS-aware" in this context, I think.

```
#if OPENSSL_KTLS_SUPPORT
        if (BIO_next(table))
            return BIO_ctrl(BIO_next(table), cmd, arg1, arg2);
#endif
```

It passes though unknown (by squid wrapper) control messages, including BIO_CTRL_SET_KTLS etc., to the next BIO (which is KTLS-ware in this case).
https://github.com/openssl/openssl/blob/bee1e83a18d6820b4264cbefca9ff5576ff679c2/crypto/bio/bss_sock.c#L185

For the remaining part of your comment, I may post comments later if I had time..

ss3git: The comment in OpenSSL internal header below suggests BIO_CTRL_SET_KTLS is open, however, it is suspected that this is true.

https://github.com/openssl/openssl/blob/bee1e83a18d6820b4264cbefca9ff5576ff679c2/include/internal/bio.h#L40

```
/* Changes to these internal BIOs must also update include/openssl/bio.h */
# define BIO_CTRL_SET_KTLS                      72
# define BIO_CTRL_SET_KTLS_TX_SEND_CTRL_MSG     74
# define BIO_CTRL_CLEAR_KTLS_TX_CTRL_MSG        75
```

I cannot find the BIO_CTRL_SET_KTLS message defined in my local include/openssl/bio.h, and adding the case condition actually makes the compilation fail.

```
bio.cc:628:10: error: use of undeclared identifier 'BIO_CTRL_SET_KTLS'
    case BIO_CTRL_SET_KTLS:
         ^
1 error generated.
```

If the above is the truth and I'm understanding it correctly, it is almost impossible to implement a single "KTLS-aware" custom BIO outside OpenSSL unless statically linking a specific version of OpenSSL library, since the upper layer SSL_*() uses the BIO_CTRL message but the custom BIO linked to ssl instance does not understand the message in a regular manner.

  Note: KTLS-aware here means that the BIO interoperates with other KTLS related functions of OpenSSL.

So, 

> If creating a single "suitable" custom BIO is not practical for some reason

This might have been reinforced unfortunatelly.

ss3git: > then another alternative to consider is complete removal of Squid custom BIO: If Squid can parse the handshake by peeking at the socket (while being called from OpenSSL), then Squid ought to be able to parse the handshake by peeking at the socket without any OpenSSL calls/guidance (and then create OpenSSL connection without any custom BIO as/if needed).


I wonder if this difference is not important for non-blocking i/o.


```
int Ssl::Bio::write(const char *buf, int size, BIO *table)
{
    :
    BIO_clear_retry_flags(table);
    if (result **<** 0) {
        const bool ignoreError = ignoreErrno(xerrno) != 0;
        debugs(83, 5, "error: " << xerrno << " ignored: " << ignoreError);
        if (ignoreError)
            BIO_set_retry_write(table);
    }

    return result;
}
```

```
static int sock_write(BIO *b, const char *in, int inl)
{
    :
    BIO_clear_retry_flags(b);
    if (ret **<=** 0) {
        if (BIO_sock_should_retry(ret))
            BIO_set_retry_write(b);
    }
    return ret;
}
```

src/ssl/bio.cc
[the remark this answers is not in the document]
ss3git: If record_ is set before negotiation finishes (it look like so), readAndBuffer(table, SQUID_TCP_SO_RCVBUF)  must be readAndBuffer(table, size) (for KTLS).

I might've misread it before.

```
/// Read and give everything to OpenSSL.
int
Ssl::ServerBio::readAndGive(char *buf, const int size, BIO *table)
{
    // If we have unused buffered bytes, give those bytes to OpenSSL now,
    // before reading more. TODO: Read if we have buffered less than size?
    if (rbufConsumePos < rbuf.length())
        return giveBuffered(buf, size);

    if (record_) {
        const int result = readAndBuffer(table, SQUID_TCP_SO_RCVBUF);
        if (result <= 0)
            return result;
        return giveBuffered(buf, size);
    }

    return Ssl::Bio::read(buf, size, table);
}
```

src/ssl/bio.cc
```diff
@@ -286,7 +308,7 @@ Ssl::ServerBio::readAndGive(char *buf, const int size, BIO *table)
         return giveBuffered(buf, size);
 
     if (record_) {
-        const int result = readAndBuffer(table);
+        const int result = readAndBuffer(table, SQUID_TCP_SO_RCVBUF);
```
ss3git, line 311: Limit the read size to "size" when KTLS is enabled.

src/ssl/bio.cc
ss3git: Supposing the comparator difference wasn't important, the logic still has a difference in the case result:-1 and errno:EAGAIN.
At least direct return from BIO_read() breaks the function (makes it incompatible with squid logic).
It is strongly suspected that "complete removal of Squid custom BIO" is simply possible.

work:
```
int
Ssl::Bio::read(char *buf, int size, BIO *table)
{
    errno = 0;
    int result;
#if OPENSSL_KTLS_SUPPORT
    if (BIO_next(table))
        **result =** BIO_read(BIO_next(table), buf, size);
    else
#endif
```

does not work:
```
int
Ssl::Bio::read(char *buf, int size, BIO *table)
{
    errno = 0;
    int result;
#if OPENSSL_KTLS_SUPPORT
    if (BIO_next(table))
        **return** BIO_read(BIO_next(table), buf, size);
    else
#endif
```

src/ssl/bio.cc
ss3git: https://github.com/squid-cache/squid/pull/1629/commits/fa5d875cfb5444425bcdaecdedafa8a0e2e998e2

include/squid.h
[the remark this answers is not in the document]
ss3git: removed OPENSSL_KTLS_SUPPORT

https://github.com/squid-cache/squid/pull/1629/commits/8f45fdc9213684a73458727eb3ef3c0eb8eeb143

src/security/PeerOptions.cc
ss3git: Here are the logs for that.


<1> Initial condition

Config with  tls_outgoing_options options=ENABLE_KTLS and http_port options=ENABLE_KTLS

```
:22.815 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bbe6000, Options=8
:22.815 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bbe6000
:22.822 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bc5f000, Options=8
:22.822 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bc5f000
:22.831 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a29a000, Options=8
:22.831 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a29a000
:22.839 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a473000, Options=8
:22.839 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a473000
:22.847 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87c1d5000, Options=8
:22.847 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87c1d5000
:22.863 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87c1e9000, Options=8
:22.863 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87c1e9000
:22.904 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a479000, Options=8
:22.904 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a479000
:22.927 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a47d000, Options=8
:22.927 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a47d000
:23.488 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a473000, Options=8
:23.488 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a473000
:23.499 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a32c000, Options=8
:23.499 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a32c000
:23.513 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a32e000, Options=8
:23.513 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a32e000
:23.525 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a332000, Options=8
:23.525 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a332000
:23.533 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a334000, Options=8
:23.534 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a334000
:23.542 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a413000, Options=8
:23.542 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a413000
:23.552 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a47f000, Options=8
:23.552 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a47f000
:23.749 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bb1a000, Options=8
:23.749 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bb1a000
:23.753 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a198000, Options=8
:23.753 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a198000
:23.760 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a422000, Options=8
:23.760 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a422000
:23.772 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a425000, Options=8
:23.772 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a425000
:23.791 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bb74000, Options=8
:23.791 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bb74000


Options is correctly recognized as 8.
Every session is created with KTLS. (expected behavior)
```


<2> without SSL_clear_options()

remove tls_outgoing_options options=ENABLE_KTLS and http_port options=ENABLE_KTLS
squid -k reconfigure

```
:04.447 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x82ca51000, Options=0
:04.447 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x82ca51000
:04.455 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a28d000, Options=0
:04.455 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a28d000
:04.464 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bf3f000, Options=0
:04.464 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bf3f000
:04.486 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bf41000, Options=0
:04.486 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bf41000
:04.494 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a2c4000, Options=0
:04.494 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a2c4000
:04.507 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bef5000, Options=0
:04.507 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bef5000
:04.557 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a329000, Options=0
:04.557 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87a329000
:04.677 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bf41000, Options=0
:04.677 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87bf41000
:05.019 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bb71000, Options=0
:05.019 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bb71000
:05.031 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87be9e000, Options=0
:05.031 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87be9e000
:05.045 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bea0000, Options=0
:05.045 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bea0000
:05.057 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bea2000, Options=0
:05.058 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bea2000
:05.066 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bff4000, Options=0
:05.066 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87bff4000
:05.076 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87be7a000, Options=0
:05.076 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87be7a000
:05.086 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87be7c000, Options=0
:05.086 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87be7c000
:05.283 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87be7e000, Options=0
:05.283 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87be7e000
:05.287 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87bf4c000, Options=0
:05.287 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87bf4c000
:05.295 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a4e9000, Options=0
:05.295 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a4e9000
:05.314 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a4eb000, Options=0
:05.314 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a4eb000
:05.326 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87a4ed000, Options=0
:05.326 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 1 for TLS session=0x87a4ed000


Options is correctly recognized as 0.
However, some of the sessons are created with KTLS. (not expected behavior)
Perhaps the server side state remains even after the reconfiguration.
```

<3> The same test with SSL_clear_options() (Initial condition is same)

remove tls_outgoing_options options=ENABLE_KTLS and http_port options=ENABLE_KTLS
squid -k reconfigure

```
:42.426 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d364000, Options=0
:42.426 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d364000
:42.441 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d366000, Options=0
:42.441 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d366000
:42.444 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d377000, Options=0
:42.444 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d377000
:42.462 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d513000, Options=0
:42.462 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d513000
:42.467 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87e92e000, Options=0
:42.467 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87e92e000
:42.486 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87eb7f000, Options=0
:42.486 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87eb7f000
:42.652 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87ec14000, Options=0
:42.652 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87ec14000
:42.973 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d366000, Options=0
:42.973 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d366000
:42.980 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d2b7000, Options=0
:42.980 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d2b7000
:43.124 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d2e2000, Options=0
:43.124 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d2e2000
:43.128 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d3ad000, Options=0
:43.128 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d3ad000
:43.452 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87ec14000, Options=0
:43.453 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87ec14000
:43.551 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d418000, Options=0
:43.551 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d418000
:43.567 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87ea83000, Options=0
:43.567 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87ea83000
:43.571 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d41a000, Options=0
:43.571 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d41a000
:43.581 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87eb7f000, Options=0
:43.581 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87eb7f000
:43.589 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87e99e000, Options=0
:43.589 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87e99e000
:43.591 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87ea76000, Options=0
:43.591 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87ea76000
:43.613 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d3a8000, Options=0
:43.613 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d3a8000
:43.728 kid1| 83,5| PeerOptions.cc(815) updateSessionOptionsOnlyKtls: set OpenSSL options (KTLS only) for session=0x87d3aa000, Options=0
:43.728 kid1| 83,5| Session.cc(142) CreateSession: KTLS: 0 for TLS session=0x87d3aa000


Options is correctly recognized as 0.
Every session is created without KTLS. (expected behavior)
```

src/ssl/bio.cc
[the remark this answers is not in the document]
ss3git: By this commit https://github.com/squid-cache/squid/pull/1629/commits/fa5d875cfb5444425bcdaecdedafa8a0e2e998e2,

> during negotiation, to read more bytes than OpenSSL has asked for.

"read more bytes than OpenSSL during cipher negotiation" is not expected to occur any more. (read more bytes than OpenSSL occurs but only for Hello.)


> (even if OpenSSL itself asks to read more than the handshake!)

This is difficult, but if it really happens, the native OpenSSL socket BIO should have the same issue.

src/ssl/bio.cc
ss3git: My comment above was not correct.
I had to set (copy) corresponding flags to "table" for this evaluation.

ss3git:
Looks like it is working on Linux too.

```
$ uname -a
Linux ubuntu2310 6.5.0-9-generic #9-Ubuntu SMP PREEMPT_DYNAMIC Sat Oct  7 01:35:40 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

$ lsmod | grep tls
tls                   151552  0

$ cat /proc/net/tls_stat
TlsCurrTxSw                             3
TlsCurrRxSw                             3
TlsCurrTxDevice                         0
TlsCurrRxDevice                         0
TlsTxSw                                 2406
TlsRxSw                                 1496
TlsTxDevice                             0
TlsRxDevice                             0
TlsDecryptError                         0
TlsRxDeviceResync                       0
TlsDecryptRetry                         0
TlsRxNoPadViolation                     0

$ /usr/local/squid/sbin/squid -v
Squid Cache: Version 6.6
Service Name: squid

This binary uses OpenSSL 3.0.10 1 Aug 2023. configure options:  '--with-large-files' '--enable-ssl-crtd' '--with-openssl=/usr/local/ssl' '--enable-security-cert-generators' '--enable-security-cert-validators--enable-linux-netfilter' '--with-default-user=daemon' --enable-ltdl-convenience
```

rousskov:
Thank you for making excellent progress with this PR! If you have time, please fix current GitHub CI test failures, but that can wait until PR code settles. On my side, I need to review the new/current code and (hopefully) clear the old red flags and red flags raised by new comments, but that will take time; I do not expect much progress until February 12, 2024.

ss3git:
FYI,
I have run a very rough benchmark of this PR.

```
<environment>

   ----------------------     10GbEther    ------------------------- 
  | squid @ freebsd 13.2 | -------------- | apache @ windows        |
   ----------------------                 |                         |
    Intel Atom C3708                      | wget @ freebsd @ vmware |
     equipped with                         ------------------------- 
    Intel QuickAssist Tech.
    (QAT)
  
  
<benchmark tool>
  
    https_proxy = squid:bump_port
    wget -O /dev/null https://apache/some_large_file.bin
    (only 1 session)
  
  
<squid builds>
  
  (a)4k
    equivalent to the plain squid (v6.6) or v6.6 with this PR.
    HTTP_REQBUF_SZ (defined in src/http/forward.h) = 4096
  
  (b)16k
    buffer expanded.
    HTTP_REQBUF_SZ = 16384
  
  (c)64k_patched
    buffer expanded, with an extra patch for src/security/Session.cc below.
    HTTP_REQBUF_SZ = 65536
  
      @@ tls_read_method(int fd, char *buf, int len)
               }
           }
           int i = SSL_read(session, buf, len);
      +    while( i > 0 && len - i > 0 ){
      +        int i_t = SSL_read(session, buf+i, len-i);
      +
      +        if ( i_t <= 0 ){
      +            break;
      +        }
      +
      +        i += i_t;
      +    };
       #elif USE_GNUTLS
           int i = gnutls_record_recv(session, buf, len);
       #endif
  
  
<benchmark results>

 TLS_AES_256_GCM_SHA384
                    throuhput (squid cpu usage)
  1. no KTLS
   (a)4k          :  450 Mbps (100%)
   (b)16k         :  800 Mbps (100%)
   (c)64k_patched : 1100 Mbps (100%)
  
  2. KTLS without QAT
   (a)4k          :  700 Mbps (100%)
   (b)16k         : 1500 Mbps (100%)
   (c)64k_patched : 2200 Mbps ( 90%)
  
  3. KTLS with QAT
   (a)4k          :  350 Mbps ( 50%)
   (b)16k         : 1100 Mbps ( 75%)
   (c)64k_patched : 1100 Mbps ( 45%)
  
 TLS_AES_128_GCM_SHA256
                    throuhput (squid cpu usage)
  1. no KTLS
   (a)4k          :  450 Mbps (100%)
   (b)16k         :  850 Mbps (100%)
   (c)64k_patched : 1100 Mbps (100%)
                                    
  2. KTLS without QAT               
   (a)4k          :  730 Mbps (100%)
   (b)16k         : 1500 Mbps (100%)
   (c)64k_patched : 2300 Mbps ( 85%)
                                    
  3. KTLS with QAT                  
   (a)4k          :  360 Mbps ( 50%)
   (b)16k         : 1200 Mbps ( 80%)
   (c)64k_patched : 1150 Mbps ( 45%)

 Note:
  - Squid is the bottleneck if the cpu usage is 100%.
  - For others, the bottleneck may be in the kernel, QAT, or benchmark environment.
  - 4k buffer seems to be too small for QAT in this benchmark. (overhead is larger.)

```

 From this limited test, the hard-coded HTTP_REQBUF_SZ might be insufficient to maximize the benefits of KTLS.
 (I am not going to touch it in this PR of course.)

ss3git:
Hi all,

After a year of usage, I tried to further boost the performance by multi-threading the OpenSSL I/O (with stream buffer size tuning).

https://github.com/squid-cache/squid/compare/v6...ss3git:squid:master_ssl_multithread?expand=1

Here are the results of throughput benchmark:
image
image

Note:
-	Simple multi-threading (w/o KTLS) apparently surpassed that with KTLS enabled in this benchmark, however, Squid is subjectively responsive when KTLS is enabled. I would personally enable both KTLS and MT.
-	The current implementation depends on Kqueue, so it does not work on Linux yet.

Although I have implemented this only for my personal interest, it could be applicable for some general cases where the multi-workers (processes) option does not apply. If potential demand exists, please let me know, then I may send PR sometime later.

rousskov:
Thank you for sharing these new results! While I am not sure four concurrent wget sessions represent an important use case, these results support the idea of investigating the related improvements (i.e., KTLS, threads, and bigger buffers) further.

FWIW, it could help folks like me to properly grok/use these results if these measurements are accompanied with some sort of model/assertions/statements that _explain_ them, so that the reader can extrapolate the measurements beyond the current basic "KTLS helps X", "threads help X", and "bigger buffers help X" implications (where "X" is something like "response throughput in a specific test environment"). This is especially important when many of the reported lines have _negative_ slope -- suggesting that these changes may hurt Squid performance at bigger scale (i.e. in some arguably more important use cases than the ones shown on graphs).

Regarding threading improvements: I cannot easily tell from the referenced raw diff how the thread code is separated from the rest of Squid thread-unsafe code. I suspect proposed threads cannot call debugs() and cannot allocate memory using Squid APIs, but there is no compile-time protection from those mistakes. Threaded code might already use thread-unsafe `fde` structures that might be accessed by the main thread in some cases. These and similar concerns raise a lot of red flags! IMHO, such safety concerns would make it very difficult to correctly offload non-trivial Squid functionality into a thread, and the performance gains of doing so (compared to using SMP Squid) would have to be _huge_ for Squid Project to accept such risks. My personal recommendation (based on these general considerations combined with poor diff understanding) is to table that threading direction/idea instead of trying to prove its merits.

ss3git:
Hello @rousskov.

I never insist that my implementation should be officially adopted. I’m doing this as noted purely for my interest (and needs in some degree) and just hoping it might be a hint for someone else. I would like to put some drawings here that hopefully express my current observation.

image

rousskov:
@ss3git, yes, I think it has been always clear that you do not insist on official adoption. And thank you for illustrations of the three mechanisms (i.e. KTLS, larger buffers, and threads); they may become handy when discussing the corresponding tradeoffs and improvement claims.

ss3git:
Regarding my multi-threading implementation, I do not think we should spend too much time to discuss deeply here.

I leave some brief note about the design (regardless of whether correct or not) for someone who might read my dirty code.

-	An I/O thread is created for each TLS session (roughly) at the timing of the first usage of tls_read_method/tls_write_method after its negotiation (bump started). The config can limit the maximum number of concurrent I/O threads. If configured to 0, or the number of currently running I/O threads reaches the limit, the I/O is not threaded until the condition changes (by reconfigutation or other I/O termination), and the behavior (I/O path) is same as the official squid.


    - Pipes between the parent (Squid main thread) and the child (I/O thread) are created along with the thread creation. 
image


    - A coule of fde* elements, which newly include the pipe information (file descriptors), mutex and flags, are arranged by the parent, and they are shared with the child. (As @rousskov says, including them in fd_table[] may be risky.)

                struct {
                    int ssl_threaded;   // 0: non-thread, 1: threaded, -1: threading failed
                    int piped_read_fd;
                    int piped_write_fd;
                    int piped_read_fd_at_thread;
                    int piped_write_fd_at_thread;
                    uint64_t ssl_traffic_counter_read;
                    uint64_t ssl_traffic_counter_write;
                    int ssl_max_write_size;
                    int real_fd;
                    int destroying;
                    pthread_mutex_t ssl_mutex;
                    pthread_t th;
                    pthread_attr_t attr;
                    int error_flag;
                } ssl_th_info;
            

    - The parent hereafter uses the pipes to read/write data for the session in place of SSL_read/SSL_write.
    
    - Select modules (currently only ModKqueue is supported) called by the parent also monitor the file handlers of the pipes instead of the one of original TLS.

    - The child reads data from the pipe and writes it to the TLS side, and vice versa.

-	During most of the lifetime of the child, the pipes are the main interface with the parent.
    - mutex and flags are used to know if the parent is willing to destroy the child.

-	At the end of each session, the parent destroys the child by destroy_child(), which is currently called in commStartTlsClose().

    - The parent and child communicate with each other using “ssl_mutex” and some flags such as “destroying“ and “error_flag”. (An ideal implementation could replace them by some atomic operations.)
    
    - After confirming that the child has stopped, the parent cleans up the fde* area (for pipes) used and ssl_th_info of itself.

src/security/PeerOptions.cc
yadij: FWIW, yes the problem with (`<2>`) results is likely that this test uses a "reconfigure" instead of a full shutdown and re-start of Squid between tests. Sessions are cached and re-used by `libssl` itself when the squid PID and process initialization are unchanged.

yadij: That said, we should support Squid being reconfigured in such a way. So clearing is probably a good idea - other options may have been suffering the same issue quietly.

rousskov: > However, some of the sessons are created with KTLS. (not expected behavior)

The word "some" underscores my original concerns: We may be resetting too much, adjusting pre-reconfiguration SSL objects that should not change (because they are already in use by ongoing transactions) instead of creating new ones (for use by new transactions). Squid [harsh] reconfiguration is designed as "free/clear all configuration objects and then configure everything from scratch". There are some exceptions, but that is the general design principle. It is not yet clear to me whether this PR violates that principle (and then has to work around the problems it introduces) or was simply suffering from (and now works around) earlier violations. In either case, more changes will be needed, either to remove the violation or to add a comment explaining odd/unusual/exceptional reset.

Can you confirm that this `s` is "a new SSL object, created after Squid started to reconfigure itself"? The answer was not clear (to me) from the debugging already provided.

ss3git: @yadij @rousskov 

> Can you confirm that this s is "a new SSL object, created after Squid started to reconfigure itself"? 

updateSessionOptionsOnlyKtls() is currently called only in CreateSession(), so I believe the _s_ is a new one (session).

However, the _ctx_ which the new session derives the options from is cached and may be reused at ConnStateData::getTlsContextFromCache(). (The issue is not reproducible if temporarily disabled the reuse path.)

Which is the expected spec of Squid?
 - a reconfiguration should not affect the sessions created during the lifetime of cached _ctx_.
 - a reconfiguration should be immediately applied to the new sessions.

PR #1630, opened 2024-01-07
Maintenance: automate header guards

kinkie:
Add a source-maintenance plugin to maintain a
header guard standard defined as SQUID_FILE_PATH_H.

### DIFF
--- a/scripts/maintenance/standard-header-guards.pl
+++ b/scripts/maintenance/standard-header-guards.pl
@@ -1,0 +1,61 @@
+#!/usr/bin/perl
+
+## Copyright (C) 1996-2023 The Squid Software Foundation and contributors
+##
+## Squid software is distributed under GPLv2+ license and includes
+## contributions from numerous individuals and organizations.
+## Please see the COPYING and CONTRIBUTORS files for details.
+##
+
+# source-maintenance plugin to standardize the header guards
+
+use strict;
+use warnings;
+
+my $filename = $ARGV[0];
+
+if ($filename !~ /\.h$/i || $filename =~ /^scripts\/boilerplate.h/i) {
+    while (<>) { print; }
+    exit(0);
+}
+
+my @accumulate=();
+my $current_guard = undef;
+
+my $first_ifndef_pos = undef;
+my $last_endif_pos = undef;
+while (<>) {
+    push(@accumulate, $_);
+    if (!defined($first_ifndef_pos) && /^#ifndef\s+(\w+_H)/ ) {
+        $current_guard = $1;
+        $first_ifndef_pos = $#accumulate;
+    } elsif (/^#endif/) {
+        $last_endif_pos = $#accumulate;
+    }
+}
+
+die("Cannot decect header guard #ifndef in $filename")
+    unless defined($first_ifndef_pos);
+die("cannot detect header guard in $filename")
+    unless defined($current_guard);
+die("no #endif in $filename - incomplete file or header guard?")
+    unless defined($last_endif_pos);
+die("last endif in $filename is not after first ifndef")
+    unless ($last_endif_pos > $first_ifndef_pos);
+die("first #ifndef in $filename seems to be the last line in the file")
+    unless ($first_ifndef_pos < $#accumulate);
+die ("#define $current_guard doesn't immediately follow first #ifndef in $filename")
+    unless ($accumulate[$first_ifndef_pos+1] =~ /^#define\s+$current_guard/);
+for (@accumulate[$last_endif_pos+1..$#accumulate]) {
+    die("unexpected content '$_' after last #ifndef in $filename") unless(/^$/);
+}
+
+my $new_guard = $filename;
+$new_guard =~ s/[\/\-\.]/_/g;
+$new_guard = "SQUID_".uc($new_guard);
+
+$accumulate[$first_ifndef_pos] = "#ifndef $new_guard\n";
+$accumulate[$first_ifndef_pos+1] = "#define $new_guard\n";
+$accumulate[$last_endif_pos] = "#endif /* $new_guard */\n";
+
+print @accumulate;


kinkie:
Please advise whether the PR should include the changes from the execution or to leave that as the subsequent run of source-maintenance

rousskov:
> Please advise whether the PR should include the changes from the execution

IMO, it should not because keeping large automated charges outside decreases this PR noise and helps with porting. However, I would not object to the inclusion if others insist on it. It is not a big deal either way.

Thank you for posting the diff!

kinkie:
Small change needed in src/dns/rfc3596.cc as it relies on header guard internals, breaking layering

src/dns/rfc3596.cc
```diff
@@ -25,7 +25,7 @@
 #include <strings.h>
 #endif
 
-#ifndef SQUID_RFC1035_H
+#ifndef RFC1035_MAXHOSTNAMESZ
```
yadij, line 28: Since you are changing this manually, please update to match our current requirements for non-file wrappers:
```suggestion
#if !defined(RFC1035_MAXHOSTNAMESZ)
```

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,64 @@
...
+    if (/^#endif/ ) {
+        if ($#accumulate >= 0) {
+            print join("", @accumulate);
+            @accumulate=($_);
+            next;
+        }
+    }
+    push(@accumulate, $_);
+}
+print "#endif /* $new_guard */\n\n";
```
yadij, line 64: Missing empty line at the end of the file. This is needed to avoid issues with some of our tools.
```suggestion
print "#endif /* $new_guard */\n\n";

```

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,64 @@
...
+}
+
+# print preamble
+while (<>) {
+    last if (/^#ifndef\s+(\w+_H)/);
+    print;
+}
+
+# we are at end of fine, no preamble found
+exit(0) unless defined($_);
```
yadij, line 41: Please consistently use either `exit(0)` or `exit 0`. Currently there is a mix.

yadij (changes requested):

scripts/maintenance/standard-header-guards.pl
kinkie: Good catch; changing to also use function-call syntax for calls to die()

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,64 @@
...
```
rousskov, line 41: Please quit with an error if a `.h` file in `src/` does not have a guard (that the script can detect). I counted approximately 15 such files (that all should be fixed AFAICT):

```
$ git grep -L -P '^#ifndef\s+(\w+_H)' '*.h' | fgrep src/
src/acl/AnnotateClient.h
src/acl/AnnotateTransaction.h
src/acl/external/kerberos_ldap_group/support.h
src/adaptation/ecap/Registry.h
src/auth/State.h
src/auth/basic/NIS/nis_support.h
src/auth/basic/RADIUS/radius-util.h
src/auth/basic/RADIUS/radius.h
src/auth/digest/LDAP/ldap_backend.h
src/auth/digest/eDirectory/edir_ldapext.h
src/auth/digest/eDirectory/ldap_backend.h
src/auth/digest/file/text_backend.h
src/auth/negotiate/kerberos/negotiate_kerberos.h
src/tests/STUB.h
src/tests/TestSwapDir.h
```

P.S. Please fix the "fine" typo in the comment while you are adjusting this code.

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,64 @@
...
+    if (/^#endif/ ) {
+        if ($#accumulate >= 0) {
+            print join("", @accumulate);
+            @accumulate=($_);
+            next;
+        }
+    }
+    push(@accumulate, $_);
+}
+print "#endif /* $new_guard */\n\n";
```
rousskov, line 64: Please refactor to quit with an error if the expected terminating `#endif` is not found (instead of silently adding a matching `#endif`) and if there are unexpected accumulated leftovers after the expected terminating `#endif` (instead of silently dropping them).

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,64 @@
...
+    print;
+}
+
+# we are at end of fine, no preamble found
+exit(0) unless defined($_);
+
+# need to rematch as $1 is in scope of loop
+/^#ifndef\s+(\w+_H)/;
+die("no header guard found") unless defined $1;
+my $current_guard = $1;
```
rousskov, line 46: Instead of rematching to set current_guard after the guard-searching loop, I recommend setting current_guard inside that loop itself.

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,64 @@
...
+<>;
+/^#define\s+(\w+)_H/;
+die("header new_guard mismatch") unless $1 eq $current_guard;
+
+print "#ifndef $new_guard\n#define $new_guard\n";
```
rousskov, line 51: For readability sake:

```suggestion
print "#ifndef $new_guard\n";
print "#define $new_guard\n";
```

rousskov (changes requested):
Thank you for adding this script.

rousskov:
I removed temporary references from the PR description and removed RELATIVE from the pattern there. IMHO, "relative" is confusing without a reference point (i.e. "relative to what?"). I do not think we have to detail that -- the script and adjusted sources will speak for themselves. I do not insist on removing RELATIVE though.

scripts/maintenance/standard-header-guards.pl
kinkie: ok

scripts/maintenance/standard-header-guards.pl
kinkie: The side effect of this change is that it requires including in the PR all the problematic header files, otherwise the source-maintenance Github check will fail this requiring some hoop-jumping to land the code. Now done

kinkie:
@rousskov I think we should be good. I had to hardcode ignoring headers in scripts/ , I couldn't think of any better way to avoid clobbering scripts/boilerplate.h without affecting its function

scripts/maintenance/standard-header-guards.pl
rousskov: Yes, and this is a _good_ side effect IMO. If others disagree, we can fix those files in a prerequisite minimal PR instead.

rousskov:
> I had to hardcode ignoring headers in scripts/ , I couldn't think of any better way to avoid clobbering scripts/boilerplate.h without affecting its function

My [suggestion](https://github.com/squid-cache/squid/pull/1630#discussion_r1445233575) was to limit the corresponding fatal errors to headers in `src/`. I will evaluate your alternative approach when I review this PR.

kinkie:
> > I had to hardcode ignoring headers in scripts/ , I couldn't think of any better way to avoid clobbering scripts/boilerplate.h without affecting its function
> 
> My [suggestion](https://github.com/squid-cache/squid/pull/1630#discussion_r1445233575) was to limit the corresponding fatal errors (and, hence, formatting) to headers in `src/`. I will evaluate your alternative approach when I review this PR.

we also have heaaders in include/ and lib/. In practice it seems to work well, the only file that shouldn't have header guards seems to be boilerplate.h. Looking forward to hear if you agree with my approach

src/adaptation/ecap/Registry.h
```diff
@@ -8,3 +8,7 @@
 
 // TBD
 
+#ifndef SQUID_SRC_ADAPTATION_ECAP_REGISTRY_H
+#define SQUID_SRC_ADAPTATION_ECAP_REGISTRY_H
+
+#endif /* SQUID_SRC_ADAPTATION_ECAP_REGISTRY_H */
```
yadij, line 14:
```suggestion
#endif /* SQUID_SRC_ADAPTATION_ECAP_REGISTRY_H */

```

src/acl/AnnotateTransaction.h
```diff
@@ -6,8 +6,8 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-#ifndef SQUID_ACLANNOTATETRANSACTION
-#define SQUID_ACLANNOTATETRANSACTION
+#ifndef SQUID_ACLANNOTATETRANSACTION_H
+#define SQUID_ACLANNOTATETRANSACTION_H
```
yadij, line 10: It would be best to add the new macro syntax to avoid unnecessary double-touching and adjust the `#endif` as well for clarity that this was not broken earlier.
```suggestion
#ifndef SQUID_SRC_ACL_ANNOTATETRANSACTION_H
#define SQUID_SRC_ACL_ANNOTATETRANSACTION_H
```

src/auth/basic/NIS/nis_support.h
```diff
@@ -6,5 +6,9 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
+#ifndef SQUID_SRC_AUTH_BASIC_NIS_NIS_SUPPORT_H
+#define SQUID_SRC_AUTH_BASIC_NIS_NIS_SUPPORT_H
+
 extern char * get_nis_password(char *user, char *nisdomain, char *nismap);
 
+#endif /* SQUID_SRC_AUTH_BASIC_NIS_NIS_SUPPORT_H */
```
yadij, line 14:
```suggestion
#endif /* SQUID_SRC_AUTH_BASIC_NIS_NIS_SUPPORT_H */

```

src/auth/basic/RADIUS/radius-util.h
```diff
@@ -8,9 +8,13 @@
...
+#ifndef SQUID_SRC_AUTH_BASIC_RADIUS_RADIUS_UTIL_H
+#define SQUID_SRC_AUTH_BASIC_RADIUS_RADIUS_UTIL_H
+
 // uses the squid utilities
 #include "util.h"
 
 /* util.c */
 uint32_t        get_ipaddr (char *);
 
+#endif /* SQUID_SRC_AUTH_BASIC_RADIUS_RADIUS_UTIL_H */
```
yadij, line 20:
```suggestion
#endif /* SQUID_SRC_AUTH_BASIC_RADIUS_RADIUS_UTIL_H */

```

src/auth/digest/LDAP/ldap_backend.h
```diff
@@ -10,8 +10,12 @@
...
 
+#ifndef SQUID_SRC_AUTH_DIGEST_LDAP_LDAP_BACKEND_H
+#define SQUID_SRC_AUTH_DIGEST_LDAP_LDAP_BACKEND_H
+
 #include "auth/digest/LDAP/digest_common.h"
 
 extern int LDAPArguments(int argc, char **argv);
 extern void LDAPHHA1(RequestData * requestData);
 
+#endif /* SQUID_SRC_AUTH_DIGEST_LDAP_LDAP_BACKEND_H */
```
yadij, line 21:
```suggestion
#endif /* SQUID_SRC_AUTH_DIGEST_LDAP_LDAP_BACKEND_H */

```

src/auth/digest/eDirectory/ldap_backend.h
```diff
@@ -10,7 +10,11 @@
...
  */
 
+#ifndef SQUID_SRC_AUTH_DIGEST_EDIRECTORY_LDAP_BACKEND_H
+#define SQUID_SRC_AUTH_DIGEST_EDIRECTORY_LDAP_BACKEND_H
+
 #include "auth/digest/eDirectory/digest_common.h"
 extern int LDAPArguments(int argc, char **argv);
 extern void LDAPHHA1(RequestData * requestData);
 
+#endif /* SQUID_SRC_AUTH_DIGEST_EDIRECTORY_LDAP_BACKEND_H */
```
yadij, line 20:
```suggestion
#endif /* SQUID_SRC_AUTH_DIGEST_EDIRECTORY_LDAP_BACKEND_H */

```

src/auth/negotiate/kerberos/negotiate_kerberos.h
```diff
@@ -148,3 +151,4 @@ char *get_ad_groups(char *ad_groups, krb5_context context, krb5_pac pac);
 #endif
 int check_k5_err(krb5_context context, const char *msg, krb5_error_code code);
 
+#endif /* SQUID_SRC_ACL_EXTERNAL_KERBEROS_LDAP_GROUP_SUPPORT_H */
```
yadij, line 154:
```suggestion
#endif /* SQUID_SRC_ACL_EXTERNAL_KERBEROS_LDAP_GROUP_SUPPORT_H */

```

src/tests/STUB.h
```diff
@@ -53,5 +54,4 @@
 /** Same as STUB_RETREF(). TODO: Remove */
 #define STUB_RETSTATREF(x) STUB_RETREF(x)
 
-#endif /* STUB */
-
+#endif /* SQUID_SRC_TESTS_STUB_H */
```
yadij, line 57:
```suggestion
#endif /* SQUID_SRC_TESTS_STUB_H */

```

src/tests/TestSwapDir.h
```diff
@@ -6,8 +6,8 @@
  * Please see the COPYING and CONTRIBUTORS files for details.
  */
 
-#ifndef TEST_TESTSWAPDIR
-#define TEST_TESTSWAPDIR
+#ifndef SQUID_SRC_TESTS_TESTWAPDIR_H
+#define SQUID_SRC_TESTS_TESTWAPDIR_H
```
yadij, line 10:
```suggestion
#ifndef SQUID_SRC_TESTS_TESTSWAPDIR_H
#define SQUID_SRC_TESTS_TESTSWAPDIR_H
```

src/tests/TestSwapDir.h
```diff
@@ -41,5 +41,4 @@ class TestSwapDir : public SwapDir
 
 typedef RefCount<TestSwapDir> TestSwapDirPointer;
 
-#endif  /* TEST_TESTSWAPDIR */
-
+#endif /* SQUID_SRC_TESTS_TESTWAPDIR_H */
```
yadij, line 44:
```suggestion
#endif /* SQUID_SRC_TESTS_TESTSWAPDIR_H */
```

src/adaptation/ecap/Registry.h
```diff
@@ -8,3 +8,7 @@
 
 // TBD
 
+#ifndef SQUID_SRC_ADAPTATION_ECAP_REGISTRY_H
+#define SQUID_SRC_ADAPTATION_ECAP_REGISTRY_H
```
yadij, line 12:
```suggestion
#ifndef SQUID_SRC_ADAPTATION_ECAP_REGISTRY_H
#define SQUID_SRC_ADAPTATION_ECAP_REGISTRY_H

// TBD
```

or just drop this unused file entirely.

yadij (changes requested):
Bunch of polish to the new additions.

yadij:
> > > I had to hardcode ignoring headers in scripts/ , I couldn't think of any better way to avoid clobbering scripts/boilerplate.h without affecting its function
> > 
> > 
> > My [suggestion](https://github.com/squid-cache/squid/pull/1630#discussion_r1445233575) was to limit the corresponding fatal errors (and, hence, formatting) to headers in `src/`. I will evaluate your alternative approach when I review this PR.
> 
> we also have heaaders in include/ and lib/. In practice it seems to work well, the only file that shouldn't have header guards seems to be boilerplate.h. Looking forward to hear if you agree with my approach

I would like to see the other directories have at least a once-over by this script when the flag-day change goes in to make sure we are consistent everywhere.  `boilerplate.h` could be manually omitted at that time, with a followup rearrangement on how the boilerplate is done to better fit in with this script and with non-C/C++ files using it.

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,72 @@
...
+
+## Copyright (C) 1996-2023 The Squid Software Foundation and contributors
+##
+## Squid software is distributed under GPLv2+ license and includes
+## contributions from numerous individuals and organizations.
+## Please see the COPYING and CONTRIBUTORS files for details.
+##
+
+# USAGE: standard-header-new_guards.pl filename.h >filename.h.new
+#
```
rousskov, line 11: Already stale and unnecessary -- the interface is known/defined by this script being a source-maintenance plugin. And the script is not really meant to be used as a stand-alone program.

```suggestion
```

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,72 @@
...
+# source-maintenance plungin to standardise the header guards
```
rousskov, line 12:
```suggestion
# source-maintenance plugin to standardize C/C++ header guard presence and spelling
```

or, at the very least:

```suggestion
# source-maintenance plugin to standardize the header guards
```

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,72 @@
...
+# - anything until the last #endif clause in the file is copied verbatim
+# - the last #endif clause is replaced with the closing header guard and an empty line
+
+use strict;
+use warnings;
+
+my $filename = $ARGV[0];
+my $new_guard = $filename;
+$new_guard =~ s/[\/\-\.]/_/g;
+$new_guard = "SQUID_".uc($new_guard);
```
rousskov, line 27: Please declare this and other variables as late as possible. Doing so makes code more readable, clarifies its intent, reduces chances of leaving unused variables behind, etc.

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,72 @@
...
+if ($filename !~ /\.h$/i || $filename =~ /^scripts\//i) {
+    while (<>) { print; }
+    exit(0);
+}
+
+my @accumulate=();
+my $current_guard = undef;
+
+# print preamble
+while ($_ = <>) {
```
rousskov, line 38:
```suggestion
while (<>) {
```

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,72 @@
...
+
+if ($filename !~ /\.h$/i || $filename =~ /^scripts\//i) {
+    while (<>) { print; }
+    exit(0);
+}
+
+my @accumulate=();
+my $current_guard = undef;
+
+# print preamble
```
rousskov, line 37:
```suggestion
# load preamble
```

rousskov (changes requested):

scripts/maintenance/standard-header-guards.pl
kinkie: from its neighbour sort-includes.pl:

```
# USAGE: sort-includes.pl filename.cc >filename.cc.sorted
```

This said, I have zero attachment to a comment; removing it

scripts/maintenance/standard-header-guards.pl
kinkie: "standardise" is the proper spelling in UK English,
but sure

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,72 @@
...
+# USAGE: standard-header-new_guards.pl filename.h >filename.h.new
+#
+# source-maintenance plungin to standardise the header guards
+#
+# Behaviour:
+# - files whose name does not end in .h are passed verbatim
+# - anything before the first #ifdef SOMETHING_ENDING_IN_H is copied verbatim
+# - the existing header guard is replaced with SQUID_RELATIVE_PATH_OF_FILE_H
+# - anything until the last #endif clause in the file is copied verbatim
+# - the last #endif clause is replaced with the closing header guard and an empty line
```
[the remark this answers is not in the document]
kinkie: removed

scripts/maintenance/standard-header-guards.pl
kinkie: This already is as late as possible; both variables are set in the loop
immediately below, and may not be scoped in that loop because they are
referenced outside the loop after it finishes

kinkie: done (old stale comment had some skew in line numbers, sorry for that)

scripts/maintenance/standard-header-guards.pl
kinkie: changed

scripts/maintenance/standard-header-guards.pl
kinkie: done

kinkie: changed

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,72 @@
...
+        if ($#accumulate >= 0) {
+            print join("", @accumulate);
+            @accumulate=($_);
+            next;
+        }
+    }
+    push(@accumulate, $_);
+}
+
+shift(@accumulate); # remove last #endif
```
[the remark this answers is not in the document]
kinkie: reworked. Checks are now all consolidated in a single block

src/auth/digest/eDirectory/ldap_backend.h
kinkie: done

src/auth/negotiate/kerberos/negotiate_kerberos.h
kinkie: done

src/tests/STUB.h
kinkie: done

src/tests/TestSwapDir.h
kinkie: done

src/tests/TestSwapDir.h
kinkie: done

src/auth/digest/LDAP/ldap_backend.h
kinkie: done by running the script

src/auth/basic/RADIUS/radius-util.h
kinkie: done

src/auth/basic/NIS/nis_support.h
kinkie: done

yadij:
Applied the two polish bits myself. LGTM for the bits I have been reviewing. Now just waiting on @rousskov review.

yadij (approved):

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,57 @@
...
+my $first_ifdef_pos = undef;
+my $last_ifdef_pos = undef;
+while (<>) {
+    if (!defined($first_ifdef_pos) && /^#ifndef\s+(\w+_H)/ ) {
+        $current_guard = $1;
+        $first_ifdef_pos = $#accumulate+1;
+    } elsif (/^#endif/) {
+        $last_ifdef_pos = $#accumulate+1;
+    }
+    push(@accumulate, $_);
```
rousskov, line 34: Please push first, before the `if` statements above. Then you will not have to remember to compensate with `+1` inside those `if` statements...

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,57 @@
...
+}
+
+die("cannot detect header guard in $filename") unless defined($current_guard);
+die("no #endif in $filename") unless defined($last_ifdef_pos);
+die ("#define $current_guard doesn't immediately follow first #ifndef in $filename")
+    unless ($accumulate[$first_ifdef_pos+1] =~ /^#define\s+$current_guard/);
```
rousskov, line 40: The current PR code is more complex/tricky than it needs to be, but I do not insist on simplifying it further. In this particular case, the complexity may lead to overread of the accumulate array though:

```suggestion
    unless ($first_ifdef_pos < $#accumulate && $accumulate[$first_ifdef_pos+1] =~ /^#define\s+$current_guard/);
```

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,57 @@
...
+
+if ($filename !~ /\.h$/i || $filename =~ /^scripts\/boilerplate.h/i) {
+    while (<>) { print; }
+    exit(0);
+}
+
+my @accumulate=();
+my $current_guard = undef;
+
+my $first_ifdef_pos = undef;
```
rousskov, line 25:
```suggestion
my $first_ifndef_pos = undef;
```

or, given the current (complex) code structure:

```suggestion
my $guard_ifdef_pos = undef;
```

... or something like that. I do not insist on this change, but I find the current "ifdef/ifndef" mismatch rather confusing.

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,57 @@
...
+my $last_ifdef_pos = undef;
```
rousskov, line 26:
```suggestion
my $last_endif_pos = undef;
```

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,57 @@
...
+die("cannot detect header guard in $filename") unless defined($current_guard);
+die("no #endif in $filename") unless defined($last_ifdef_pos);
+die ("#define $current_guard doesn't immediately follow first #ifndef in $filename")
+    unless ($accumulate[$first_ifdef_pos+1] =~ /^#define\s+$current_guard/);
+for (@accumulate[$last_ifdef_pos+1..$#accumulate]) {
+    die("unexpected content '$_' after last #ifndef in $filename") unless(/^$/);
+}
+
+my $new_guard = $filename;
+$new_guard =~ s/[\/\-\.]/_/g;
```
rousskov, line 46: I would replace everything except alphanumeric characters, but I do not insist on this change.

scripts/maintenance/standard-header-guards.pl
```diff
@@ -0,0 +1,57 @@
...
+$new_guard = "SQUID_".uc($new_guard);
```
rousskov, line 47: If we replace guard lines inside the accumulation array, we would not have to split array printing into several groups with complex start/end points. I do not insist on that approach though.

rousskov (approved):
Excellent progress. I think we are almost done here. Since I may not be able to re-review quickly, I am not formally requesting any changes, but please address as many of my change requests as possible before merging this.

scripts/maintenance/standard-header-guards.pl
rousskov: Yes, I am aware. Guess how many times that proper spelling variation is used in Squid sources?

scripts/maintenance/standard-header-guards.pl
kinkie: Good point. Replacing with $first_ifndef_pos

scripts/maintenance/standard-header-guards.pl
kinkie: fair point. doing that

scripts/maintenance/standard-header-guards.pl
kinkie: Overread can only happen if the first ifndef is the last line in the file (and if it is, source formatting is the last of our problems and this is likely ).

There are therefore only two cases where this can happen:
- no endif/truncated file - this is caught by the test in line 38
- last endif is before the first ifdef - I'm adding an explicit test for it
- I will add an explicit test for this scenario, with a more useful error message; I would expect it will never ever trigger tho

scripts/maintenance/standard-header-guards.pl
kinkie: For the files we currently have, I can't spot any obvious issues; I'll leave as-is for now

kinkie:
> The current PR approach probably "works" but it is "neither here nor there" IMHO. I am not violently against it, but I think we can and should do better (by moving in either direction outlined in the bullets above).

Current approach only skips boilerplate.h

rousskov (commented):
Thank you for addressing my primary concerns! Unfortunately, I do not have the time to check and simplify this further. I am not requesting any changes.

kinkie:
In order to avoid disruption for followup PRs due to codeQL checks failing after merging this script,
I'll adopt this merge strategy:

1. I'll extract the header changes from this PR and land them as their own PR
2. I'll run the script once and land the produced changes as their own PR
3. I'll land the script

yadij (approved):

kinkie:
Will readd backporting after having landed the v6-specific script-generated changes.

squidadm:
queued for backport to v6

PR #1631, opened 2024-01-08
Rename class ACL to Acl::Node

kinkie:
There is a name clash between Squid's ACL and
MS Windows' ACL (http://tinyurl.com/ns-winnt-acl).

### DIFF
--- a/src/acl/Acl.cc
+++ b/src/acl/Acl.cc
@@ -118,13 +118,13 @@ AclNode *
 AclNode::FindByName(const char *name)
 {
     AclNode *a;
-    debugs(28, 9, "AclNode::FindByName '" << name << "'");
+    debugs(28, 9, "name=" << name);
 
     for (a = Config.aclList; a; a = a->next)
         if (!strcasecmp(a->name, name))
             return a;
 
-    debugs(28, 9, "AclNode::FindByName found no match");
+    debugs(28, 9, "found no match");
 
     return nullptr;
 }
@@ -364,7 +364,7 @@ AclNode::cacheMatchAcl(dlink_list * cache, ACLChecklist *checklist)
         auth_match = (acl_proxy_auth_match_cache *)link->data;
 
         if (auth_match->acl_data == this) {
-            debugs(28, 4, "AclNode::cacheMatchAcl: cache hit on acl '" << name << "' (" << this << ")");
+            debugs(28, 4, "cache hit on acl '" << name << "' (" << this << ")");
             return auth_match->matchrv;
         }
 
@@ -373,7 +373,7 @@ AclNode::cacheMatchAcl(dlink_list * cache, ACLChecklist *checklist)
 
     auth_match = new acl_proxy_auth_match_cache(matchForCache(checklist), this);
     dlinkAddTail(auth_match, &auth_match->link, cache);
-    debugs(28, 4, "AclNode::cacheMatchAcl: miss for '" << name << "'. Adding result " << auth_match->matchrv);
+    debugs(28, 4, "miss for acl '" << name << "'. Adding result " << auth_match->matchrv);
     return auth_match->matchrv;
 }
 
@@ -427,7 +427,7 @@ AclNode::~AclNode()
 void
 AclNode::Initialize()
 {
-    AclNode *a = Config.aclList;
+    auto *a = Config.aclList;
     debugs(53, 3, "AclNode::Initialize");
 
     while (a) {
--- a/src/acl/Acl.cc
+++ b/src/acl/Acl.cc
@@ -31,16 +31,16 @@ const char *AclMatchedName = nullptr;
 
 namespace Acl {
 
-/// Acl::AclNode type name comparison functor
+/// Acl::Node type name comparison functor
 class TypeNameCmp {
 public:
     bool operator()(TypeName a, TypeName b) const { return strcmp(a, b) < 0; }
 };
 
-/// Acl::AclNode makers indexed by Acl::AclNode type name
+/// Acl::Node makers indexed by Acl::Node type name
 typedef std::map<TypeName, Maker, TypeNameCmp> Makers;
 
-/// registered Acl::AclNode Makers
+/// registered Acl::Node Makers
 static Makers &
 TheMakers()
 {
@@ -48,9 +48,9 @@ TheMakers()
     return Registry;
 }
 
-/// creates an Acl::AclNode object of the named (and already registered) Acl::AclNode child type
+/// creates an Acl::Node object of the named (and already registered) Acl::Node child type
 static
-Acl::AclNode *
+Acl::Node *
 Make(TypeName typeName)
 {
     const auto pos = TheMakers().find(typeName);
@@ -103,19 +103,19 @@ Acl::SetKey(SBuf &keyStorage, const char *keyParameterName, const char *newKey)
 }
 
 void *
-Acl::AclNode::operator new (size_t)
+Acl::Node::operator new (size_t)
 {
-    fatal ("unusable Acl::AclNode::new");
+    fatal ("unusable Acl::Node::new");
     return (void *)1;
 }
 
-void Acl::AclNode::operator delete(void *)
+void Acl::Node::operator delete(void *)
 {
-    fatal ("unusable Acl::AclNode::delete");
+    fatal ("unusable Acl::Node::delete");
 }
 
-Acl::AclNode *
-Acl::AclNode::FindByName(const char *name)
+Acl::Node *
+Acl::Node::FindByName(const char *name)
 {
     debugs(28, 9, "name=" << name);
 
@@ -128,7 +128,7 @@ Acl::AclNode::FindByName(const char *name)
     return nullptr;
 }
 
-Acl::AclNode::AclNode() :
+Acl::Node::Node() :
     cfgline(nullptr),
     next(nullptr),
     registered(false)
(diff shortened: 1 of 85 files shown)


src/cache_cf.cc
```diff
@@ -4724,7 +4724,7 @@ static void parse_ftp_epsv(acl_access **ftp_epsv)
         *ftp_epsv = nullptr;
 
         if (ftpEpsvDeprecatedAction == Acl::Answer(ACCESS_DENIED)) {
-            if (ACL *a = ACL::FindByName("all"))
+            if (AclNode *a = AclNode::FindByName("all"))
```
yadij, line 4727:
```suggestion
            if (auto *a = AclNode::FindByName("all"))
```

src/acl/InnerNode.cc
```diff
@@ -57,7 +57,7 @@ Acl::InnerNode::lineParse()
             ++t;
 
         debugs(28, 3, "looking for ACL " << t);
-        ACL *a = ACL::FindByName(t);
+        AclNode *a = AclNode::FindByName(t);
```
yadij, line 60:
```suggestion
        auto *a = AclNode::FindByName(t);
```

src/acl/Gadgets.cc
```diff
@@ -246,16 +246,16 @@ aclDeregister(ACL *acl)
...
-aclDestroyAcls(ACL ** head)
+aclDestroyAcls(AclNode ** head)
 {
     *head = nullptr; // Config.aclList
     if (AclSet *acls = RegisteredAcls) {
         debugs(28, 8, "deleting all " << acls->size() << " ACLs");
         while (!acls->empty()) {
-            ACL *acl = *acls->begin();
-            // We use centralized deletion (this function) so ~ACL should not
+            AclNode *acl = *acls->begin();
```
yadij, line 255:
```suggestion
            auto *acl = *acls->begin();
```

src/acl/Gadgets.cc
```diff
@@ -76,9 +76,9 @@ aclIsProxyAuth(const char *name)
 
     debugs(28, 5, "aclIsProxyAuth: called for " << name);
 
-    ACL *a;
+    AclNode *a;
```
yadij, line 80:
```suggestion
```

src/acl/Gadgets.cc
```diff
@@ -76,9 +76,9 @@ aclIsProxyAuth(const char *name)
...
 
-    if ((a = ACL::FindByName(name))) {
+    if ((a = AclNode::FindByName(name))) {
```
yadij, line 81:
```suggestion
    if (const auto *a = AclNode::FindByName(name)) {
```

src/acl/Acl.cc
```diff
@@ -417,18 +417,18 @@ ACL::requiresRequest() const
...
     AclMatchedName = nullptr; // in case it was pointing to our name
 }
 
 void
-ACL::Initialize()
+AclNode::Initialize()
 {
-    ACL *a = Config.aclList;
-    debugs(53, 3, "ACL::Initialize");
+    AclNode *a = Config.aclList;
```
yadij, line 430:
```suggestion
    auto *a = Config.aclList;
```

src/acl/Acl.cc
```diff
@@ -373,7 +373,7 @@ ACL::cacheMatchAcl(dlink_list * cache, ACLChecklist *checklist)
 
     auth_match = new acl_proxy_auth_match_cache(matchForCache(checklist), this);
     dlinkAddTail(auth_match, &auth_match->link, cache);
-    debugs(28, 4, "ACL::cacheMatchAcl: miss for '" << name << "'. Adding result " << auth_match->matchrv);
+    debugs(28, 4, "AclNode::cacheMatchAcl: miss for '" << name << "'. Adding result " << auth_match->matchrv);
```
yadij, line 376:
```suggestion
    debugs(28, 4, "miss for acl '" << name << "'. Adding result " << auth_match->matchrv);
```

src/acl/Acl.cc
```diff
@@ -364,7 +364,7 @@ ACL::cacheMatchAcl(dlink_list * cache, ACLChecklist *checklist)
         auth_match = (acl_proxy_auth_match_cache *)link->data;
 
         if (auth_match->acl_data == this) {
-            debugs(28, 4, "ACL::cacheMatchAcl: cache hit on acl '" << name << "' (" << this << ")");
+            debugs(28, 4, "AclNode::cacheMatchAcl: cache hit on acl '" << name << "' (" << this << ")");
```
yadij, line 367:
```suggestion
            debugs(28, 4, "cache hit on acl '" << name << "' (" << this << ")");
```

src/acl/Acl.cc
```diff
@@ -103,48 +103,48 @@ Acl::SetKey(SBuf &keyStorage, const char *keyParameterName, const char *newKey)
...
-    debugs(28, 9, "ACL::FindByName '" << name << "'");
+    AclNode *a;
+    debugs(28, 9, "AclNode::FindByName '" << name << "'");
 
     for (a = Config.aclList; a; a = a->next)
         if (!strcasecmp(a->name, name))
             return a;
 
-    debugs(28, 9, "ACL::FindByName found no match");
+    debugs(28, 9, "AclNode::FindByName found no match");
```
yadij, line 127:
```suggestion
    debugs(28, 9, "found no match");
```

src/acl/Acl.cc
```diff
@@ -103,48 +103,48 @@ Acl::SetKey(SBuf &keyStorage, const char *keyParameterName, const char *newKey)
...
 
-ACL *
-ACL::FindByName(const char *name)
+AclNode *
+AclNode::FindByName(const char *name)
 {
-    ACL *a;
-    debugs(28, 9, "ACL::FindByName '" << name << "'");
+    AclNode *a;
+    debugs(28, 9, "AclNode::FindByName '" << name << "'");
```
yadij, line 121:
```suggestion
    debugs(28, 9, "name=" << name);
```

yadij (changes requested):
Some minor polish to touched lines.

kinkie:
This stab has a smaller scope than the PR #1350, as it only focuses on renaming the class.
With this change, an AclNode is the inner representation of an administrator-visible ACL

src/acl/Acl.h
```diff
@@ -42,31 +42,31 @@ void SetKey(SBuf &keyStorage, const char *keyParameterName, const char *newKey);
 /// Can evaluate itself in FilledChecklist context.
 /// Does not change during evaluation.
 /// \ingroup ACLAPI
-class ACL
+class AclNode
```
rousskov, line 45: Why is this class renamed to AclNode and not to Acl::Node? Why is it better to modify more than 50 source code files twice (once in this PR and then again in a followup PR that will have to move the same class into its proper namespace) while performing essentially the same renaming change?

rousskov (dismissed):

src/acl/Acl.h
kinkie: The reason is searchability. A unique name, although more verbose than needed, enables using many more tools than a clean, but ambiguous without specialist knowledge, name

rousskov: Acl::Node is a globally unique name. Do we really need to argue whether C++ namespaces are better than enhanced searchability by C++-unaware tools?!

kinkie: We don't, and I'm not attached to the name.

<sarcastic, but not wrong>
Thankfully all that's needed now is to `sed -i~ 's/AclNode/Node/g' $(git ls-files)` ; the other way around is not as easy, even with specialised tools
</sarcastic>

If that is the only objection to merging this PR, I'll do that right away.

rousskov: > I'm not attached to the name.

It is not about the spelling of the name. It is about properly using C++ namespaces. FWIW, I was shocked to see AclName being seriously proposed as a candidate when we already have and actively use the Acl namespace.

> Thankfully all that's needed now is to `sed -i~ 's/AclNode/Node/g' $(git ls-files)`

That substitution would not work AFAICT because AclNode was not declared inside the Acl namespace. You need something like `s/AclNode/Acl::Node/g`.

> the other way around is not as easy, even with specialised tools

Mechanical[^1] renaming of official ACL to Acl::Node is as easy/difficult as mechanical renaming of official ACL to AclNode. The primary difference is that only one of the two targets is acceptable for long-term use.

[^1]: Here, I use that term to mean "without the use of tools that understand C++".

Mechanical renaming of Acl::Node to Acl::Foo is currently also easy because there are very few bare Node names in the current code.

In general, yes, namespaces may complicate or prohibit certain mechanical actions. That is a (small) negative to consider when evaluating whether to introduce a _new_ namespace. However, those difficulties are pretty much irrelevant when it comes to using the _existing_ applicable namespace!

> If that is the only objection to merging this PR, I'll do that right away.

I reserve the right to request (or implement) more changes after I see (or implement) the outcome of this critical correction. The compiler might ask us to fix a few problems as well. However, FWIW, I will not insist on you removing unnecessary `Acl::` prefixes from Acl::Node occurrences inside Acl namespace; that "Foo inside Foo" style violation can be considered acceptable for this special renaming PR IMO.

kinkie: AclNode -> Acl::Node is now done, including hoisting the class into its own header. Unfortunately the implementation of the ACL types registry is too coupled to Node's internals to do the same with the implementation file without out-of-scope surgery

rousskov (commented):
I have adjusted PR title/description to match the current PR. I did not check most of the other details, but I do not object to this PR going in.

IMO, a lot (and possibly most) of these renames should be automatically polished further to remove unnecessary `Acl::` prefixes inside Acl namespace. Some "ACL" replacements in comments made those comments even less clear and should be manually adjusted/undone. I can do both, but not in the immediate future.

As promised, I am not going to block this PR on those polishing grounds despite the long-term pain created by numerous excessive Acl:: prefixes.

kinkie:
> numerous excessive Acl:: prefixes

I should have caught and removed the vast majority of these

kinkie:
@yadij , there's your outstanding negative vote. Hopefully your doubts have been addressed now

yadij (approved):
LGTM

PR #1633, opened 2024-01-11
Maintenance: remove unused asnAclInitialize declaration

kinkie:


rousskov (dismissed):
asnAclInitialize() is unused since 2003 commit b0dd28b that introduced a general prepareForUse() replacement.

rousskov (approved):

PR #1635, opened 2024-01-13
Create libasync

kinkie:
Move Async and Events related code
to own library in src/async


kinkie:
Note: I have intentionally avoided creating Async namespace, this PR touches so much code already.
If there is consensus to do it, I'll do that as a followup PR

kinkie:
The idea for this PR stems from a [comment](https://github.com/squid-cache/squid/pull/1548#issuecomment-1880348532) in PR #1548

rousskov:
> I have intentionally avoided creating Async namespace, this PR touches so much code already.

FWIW, I do not think creating an Async namespace is a good idea (in any PR).

yadij:
> > I have intentionally avoided creating Async namespace, this PR touches so much code already.
> 
> FWIW, I do not think creating Async namespace is a good idea (in any PR).

And yet it was you who decided all these objects should consume the `Async` prefix (a.k.a. namespace).
You have a better alternative to rename them as?


IMO the `AsyncJob` and **related** pieces are sufficiently complex and self-contained to deserve their own library+namespace like this.

However, @kinkie please do not merge the `EventLoop` mechanism into it. That should have a separate library of its own.

yadij:
@kinkie, also be aware the `AsyncEngine` is an interface for the `Engine` mechanism, not the part of `AsyncJob` mechanism.

kinkie:
> > I have intentionally avoided creating Async namespace, this PR touches so much code already.
> 
> FWIW, I do not think creating Async namespace is a good idea (in any PR).

Sure, but could you highlight why? It would be a deviation from standard practice, isn't it?

src/AsyncEngine.cc
```diff
@@ -8,4 +8,3 @@
 
 #include "squid.h"
 #include "AsyncEngine.h"
-
```
yadij, line 11: File should no longer be touched.

src/FwdState.cc
```diff
@@ -27,7 +28,6 @@
 #include "comm/Loops.h"
 #include "CommCalls.h"
 #include "errorpage.h"
-#include "event.h"
```
yadij, line 30: This `event.h` and corresponding `event.cc` code is all part of the `EventLoop` mechanism. Not part of `AsyncJob` code, even though the two are similar and interact.

src/Makefile.am
```diff
@@ -16,15 +16,15 @@ LOADABLE_MODULES_SOURCES = \
...
+SUBDIRS = async mem time debug base anyp helper dns html ftp parser comm error eui acl format clients sbuf servers fs repl store DiskIO proxyp
 
 if ENABLE_AUTH
 SUBDIRS += auth
 AUTH_LIBS= auth/libauth.la
 AUTH_ACL_LIBS= auth/libacls.la
 endif
 
-SUBDIRS	+= http ip icmp ident log ipc mgr
+SUBDIRS	+=  http ip icmp ident log ipc mgr
```
yadij, line 27: typo 
```suggestion
SUBDIRS	+= http ip icmp ident log ipc mgr
```

src/async/forward.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "base/forward.h"
+
+class AsyncCall;
+class AsyncCallQueue;
+class AsyncJob;
+class CallDialer;
+class DelayedAsyncCalls;
+
+template <class Answer> class AsyncCallback;
```
yadij, line 20:
```suggestion
template<class Answer> class AsyncCallback;
```

src/async/forward.h
```diff
@@ -0,0 +1,27 @@
...
+using AsyncCallPointer = RefCount<AsyncCall>;
+template<class Job> class JobWait;
+
+template<class Cbc> class CbcPointer;
+typedef CbcPointer<AsyncJob> AsyncJobPointer;
```
yadij, line 25:
```suggestion
using AsyncJobPointer = CbcPointer<AsyncJob>;
```

yadij (changes requested):

rousskov (changes requested):
> Alex: FWIW, I do not think creating Async namespace is a good idea (in any PR).

> Francesco: Sure, but could you highlight why?

I am not convinced that every Squid class must be in a namespace[^1]. If some Squid classes should not be in a namespace, then it is reasonable to question whether, say, AsyncCall should be. To me, AsyncCall feels similar to SBuf -- both do not really need a dedicated namespace[^1] AFAICT.

I failed to find an _applicable_ industry guideline regarding the use of namespaces in similar contexts.

The value of numerous users using "Async::Call" (instead of existing "AsyncCall") just so that AsyncCall implementation can simply say "Call" is rather questionable.

If _all_ deferred calls (whatever we decide that or a similar term means) should migrate to (the variation of) AsyncCall API, then we may even decide to drop "Async" itself. One large change is better than two (with the same end result) in this case.

I will change my position if somebody proves that we should change 800+ lines of code to migrate AsyncCall and AsyncJob to Async namespace, but I believe that change proposal must come first, before I am forced to spend even more time on discussing a change that has not been really proposed yet.

[^1]: Other than the existing default global namespace and "Squid" namespace that we might use more aggressively for reasons unrelated to this discussion.

> Francesco: It would be a deviation from standard practice, isn't it?

You will need to define "standard practice" for me to be able to answer that question. I can interpret that term many different ways. Needless to say, Squid does not use namespaces for some code grouped inside src directories, and I am not aware of any requirement to use a namespace for all src directories code. Each directory or group of related names deserves a dedicated decision IMO.


> Amos: And yet it was you who decided all these objects should consume the `Async` prefix (a.k.a. namespace).

I have been begging to stop posting reviews of people and yet here we go again.

* Squid classes used "Async prefix" before "all these objects" were introduced
* a class name prefix is not the same as a C++ namespace
* not "all these objects" consume the "Async prefix"


> Amos: You have a better alternative to rename them as?

I am not convinced they should be renamed at this time.


> Amos: IMO the `AsyncJob` and **related** pieces are sufficiently complex and self-contained to deserve their own library+namespace like this.

FWIW, I am (still) OK with grouping AsyncCall-related code in a dedicated directory and convenience library. The devil is in the details (e.g., what code to move, what directory name to use, whether to introduce a new namespace, whether to rename anything, etc.). For example, we can move without introducing a new namespace. I believe those details are being hammered out [elsewhere](  https://github.com/squid-cache/squid-cache.github.io/pull/20), and this PR should continue to wait for the conclusion of those discussions.

> Amos: the `EventLoop` mechanism ... should have a separate library of its own.

FWIW, there is currently no consensus on whether EventLoop deserves a library of its own.

PR #1636, opened 2024-01-14
Remove external LM_group ACL helper

kinkie:
The LM_group helper has not built since at least Squid v3.5.
Remove it as likely not useful.


kinkie:
@yadij , can you help remove the leftover bits in the manuals tranlsations?

kinkie:
The problem is actually different: there is a name clash between our WinSvc.h file and the <winsvc.h> system header

PR #1642, opened 2024-01-17
Remove AclMatchedName from ACL::ParseAclLine()

rousskov:
ACL parsing code needs to know the aclname parameter of the being-parsed
acl directive to report various errors. Most admins recognize their ACLs
by these names so reporting aclnames improves UX. Since before 1999
commit b6a2f15, Squid used a "temporary" and "ugly" trick that supplied
aclname via the unrelated global variable called AclMatchedName (which
has its own set of problems). Some ACL parsing code used AclMatchedName
in cache.log messages, but most ACL-related problems were still reported
without that information.

Passing ACL::name to each parsing-related function via an extra
parameter is not just ugly but impractical because some the low-level
parsing functions should not really know about ACLs. Instead, we reuse
existing CodeContext mechanism to report parsing context information (in
this case -- aclname).

We plan to enhance parsing context to cover directives other than "acl"
(without modifying every directive parser, of course), but this first
small step significantly reduces configuration code exposure to
AclMatchedName, unblocking ACL-related smooth reconfiguration
improvements.


src/acl/Acl.cc
```diff
@@ -268,22 +305,11 @@ Acl::Node::ParseAclLine(ConfigParser &parser, Node ** head)
         new_acl = 0;
     }
 
-    /*
-     * Here we set AclMatchedName in case we need to use it in a
-     * warning message in Acl::SplayInserter::Merge().
-     */
-    AclMatchedName = A->name;   /* ugly */
```
rousskov, line 275: Removal of this "ugly" hack (and the code that uses it) is the focus of this PR, but I expect this PR changes to have long-term positive effects on configuration parsing in general. BTW, the work on this PR also triggered discovery of problems fixed in recent commit 0898d0f; we had even tried to fix them in this PR (before we realized how deep that well goes).

src/acl/Acl.cc
```diff
@@ -211,7 +237,16 @@ Acl::Node::ParseAclLine(ConfigParser &parser, Node ** head)
         return;
     }
 
-    xstrncpy(aclname, t, ACL_NAME_SZ);
+    SBuf aclname(t);
+    CallParser(ParsingContext::Pointer::Make(aclname), [&] {
+        ParseNamed(parser, head, aclname.c_str()); // TODO: Convert Node::name to SBuf
```
rousskov, line 242: IMHO, this c-string-to-SBuf conversion does not qualify for a "performance regression" XXX because the now-removed code was also copying the parsed token (probably for no good reason). All this noise (and remaining copying in Acl::Node constructor) will go away when we convert Acl::Node::name to SBuf, but that change requires a dedicated PR. Said that, if others insist on this PR adding an XXX comment here, I will do that without arguing.

test-suite/squidconf/bad-acl-dstdomain-dupe.conf.instructions
```diff
@@ -1,25 +1,31 @@
...
+    acl name: test22
 
 WARNING: Ignoring .example.com because it is already covered by .example.com
-    advice: Remove value .example.com from the ACL named test31
-
+    advice: Remove value .example.com from the ACL
 WARNING: Ignoring c.example.com because it is already covered by .example.com
-    advice: Remove value c.example.com from the ACL named test31
+    advice: Remove value c.example.com from the ACL
+    acl name: test31
```
rousskov, line 22: We managed to keep ACL names in all these checking patterns except when the ACL name is _repeated_ (because there are multiple warnings for the same ACL X, each containing the expected "acl name: X" line). In those cases, test-squid-conf.sh lacks code to check that there are _two_ "acl name: X" lines. There are some [hairy shell tricks](https://unix.stackexchange.com/questions/335276/grep-v-how-to-exclude-only-the-first-or-last-n-lines-that-match) that would allow us to implement those checks, but I do not think we should complicate test-squid-conf.sh so much. This problem is best solved after/if we migrate test-squid-conf.sh to a better scripting language.

test-suite/squidconf/bad-acl-dstdomain-dupe.conf
```diff
@@ -6,7 +6,7 @@
 ##
 
 acl test11 dstdomain .example.com example.com
-acl test12 dstdomain example.com .example.com
+acl test12 dstdomain example.org .example.org
```
rousskov, line 9: We slightly changed these and similar test cases to make their WARNING lines unique (after those lines lost the ACL name that was moved into another/dedicated `acl name: ...` line).

src/base/CodeContext.h
```diff
@@ -106,20 +106,38 @@ class CodeContextGuard
...
+
+/// To supply error-reporting code with parsing context X (where the error
+/// occurred), parsing code should use this function when initiating parsing
+/// inside that context X.
+/// \sa CallAndRestore_()
+template <typename Fun>
+inline void
+CallParser(const CodeContext::Pointer &parsingContext, Fun &&parse)
+{
+    CallAndRestore_(parsingContext, parse);
```
rousskov, line 140: The implementation of the new CallParser() is currently exactly the same as the implementation of CallBack(), but their purposes are very different. We moved the common code into CallAndRestore_() helper so that there is no code duplication here, but I believe it is best to keep referring to the two use cases using two unique names rather than trying to come up with something generic (which we tried and did not like the results).

rousskov (commented):
These comments are meant to clarify what this PR is doing. They do not request any code changes.

PR #1644, opened 2024-01-21
Throw, not self_destruct(), on qos_flow configuration errors

jtstrs:


jtstrs:
If its preferred to leave this as it. 
Probably there make sense to mark
that this is expected behaviour
more explicitly

yadij:
There are several other items of polish that need doing to this method. Most accessible of those being the `debugs(x,y, ZZ); self_destruct();` sequences should also be replaced with `throw TextException(ToSBuf(ZZ), Here());` if you want to do a more general polish of this function.

And thank you for the interest in improving Squid.

src/ip/QosConfig.cc
```diff
@@ -325,35 +326,30 @@ Ip::Qos::Config::parseConfigLine()
...
+                throw TextException(ToSBuf("ERROR: Invalid parameter 'mark' in qos_flows option. ", 
+                                           "Linux Netfilter marking not available on this platform."), Here());
 #endif // SO_MARK && USE_LIBCAP
             } else if (strncmp(token, "tos",3) == 0) {
                 preserveMissTos = true;
                 tos = true;
             } else {
-                preserveMissTos = true;
-                tos = true;
+                throw TextException(ToSBuf("Unsupported use case"), Here());
```
rousskov, line 339: If this code stays, please report the unsupported token value and, if possible, tell the admin what that token is (e.g. "qos_flows value type name"?) instead of using a generic "use case" description. Imagine an admin looking at a configuration line containing _multiple_ tokens, trying to understand which one is "unsupported"...

rousskov (changes requested):
Replacing debugs()+self_destruct() with throw is a step forward, thank you. The current PR description does not appear to match PR changes. Perhaps it is missing an "Also, " prefix? Please keep that PR description in sync as you address change requests.

src/ip/QosConfig.cc
jtstrs: Your right.
I will update this error message with meaningful description.
Thank you

src/ip/QosConfig.cc
```diff
@@ -325,35 +326,30 @@ Ip::Qos::Config::parseConfigLine()
...
+                throw TextException(ToSBuf("Invalid parameter 'mark' in qos_flows option. ",
+                                           "Linux Netfilter marking not available on this platform."), Here());
 #endif // SO_MARK && USE_LIBCAP
             } else if (strncmp(token, "tos",3) == 0) {
                 preserveMissTos = true;
                 tos = true;
             } else {
-                preserveMissTos = true;
-                tos = true;
+                throw TextException(ToSBuf("Bad qos_flaws token value ", &token[0]), Here());
```
rousskov, line 339: AFAICT, there is [consensus](https://github.com/squid-cache/squid/pull/1644#issuecomment-1902855838) that we want to get rid of this "default to tos" behavior. Throwing an exception here is the right thing to do. However, throwing an exception is _not_ enough! The missing pieces include:

1. Update cf.data.pre to document the new/updated qos_flows configuration syntax. FWIW, without that update, I have no idea what syntax this PR is trying to support!
2. Move tos/mark determination away from the `while` loop (assuming the first tos/mark parameter is now required and must not be repeated).
3. Fix `qos_flaws` spelling in the new exception text.
4. Replace `token value` in the exception text with the name of the now-required qos_flows parameter from item 1 and suggest two of the supported values for that parameter (i.e. "tos" and "mark").
5. Update "Default to TOS for backwards compatibility" comment (currently at the start of the loop).
6. Update PR title/description to focus on this functionality change.
7. Update doc/release-notes/release-7.sgml.in to mention this functionality change.

FWIW, I recommend splitting this PR into two:

* A PR that replaces self_destruct() with exceptions. I recommend using this PR for that improvement. I have note checked carefully, but it is possible that all you need to do is to undo this `else` change.
* A PR that removes backward compatibility by updating this `else` clause and refactoring (see the seven items above). I recommend opening a new PR dedicated to that functionality change, after this PR merges.

rousskov (changes requested):

src/ip/QosConfig.cc
jtstrs: Good, will make as you suggested.
Could you check the rest changes?

src/ip/QosConfig.cc
```diff
@@ -325,13 +326,11 @@ Ip::Qos::Config::parseConfigLine()
...
-                       << "Linux Netfilter marking not available without LIBCAP support.");
-                self_destruct();
+                throw TextException(ToSBuf("Invalid parameter 'mark' in qos_flows option. ",
+                                           "Linux Netfilter marking not available on this platform."), Here());
 #else // SO_MARK && USE_LIBCAP
-                debugs(3, DBG_CRITICAL, "ERROR: Invalid parameter 'mark' in qos_flows option. "
-                       << "Linux Netfilter marking not available on this platform.");
-                self_destruct();
+                throw TextException(ToSBuf("Invalid parameter 'mark' in qos_flows option. ",
+                                           "Linux Netfilter marking not available on this platform."), Here());
```
rousskov, line 333: We do not really need ToSBuf() in this and similar contexts where just string literals are concatenated[^1], but since other nearby exceptions have to use ToSBuf(), I think it is OK to use ToSBuf() here as well. Your call.

[^1]: Removing a comma between the two arguments will allow C++ to concatenate [string literals](https://en.cppreference.com/w/cpp/language/string_literal) instead of ToSBuf() doing the same at runtime AFAICT.

src/ip/QosConfig.cc
```diff
@@ -346,14 +345,12 @@ Ip::Qos::Config::parseConfigLine()
...
-                    debugs(3, DBG_CRITICAL, "ERROR: Bad mark local-hit value " << &token[10]);
-                    self_destruct();
+                    throw TextException(ToSBuf("Bad mark local-hit value ", &token[10]), Here());
                 }
             } else {
                 unsigned int v = 0;
                 if (!xstrtoui(&token[10], nullptr, &v, 0, std::numeric_limits<tos_t>::max())) {
-                    debugs(3, DBG_CRITICAL, "ERROR: Bad TOS local-hit value " << &token[10]);
-                    self_destruct();
+                    throw TextException(ToSBuf("Bad TOS local-hit value ", &token[10]), Here());
```
rousskov, line 353: The `&token[10]` expression is ugly and dangerous, but I do not think this PR should change that because a proper fix will require some out-of-scope refactoring. Most of these lines will need to be changed again as we migrate to better parsing routines anyway (and that migration will remove that expression), so there is no particular pressure to improve them further right now, in this simple PR IMO. I am not requesting any related changes here.

rousskov (approved):
LGTM, with or without the string literals changes discussed in a dedicated change request. I plan to clear this PR for merging after 48 hours if others do not intervene.

P.S. I have fixed `#include` formatting, adjusted PR title (to be more specific), and removed PR description (as no longer necessary after the title adjustment).

rousskov:
OK to test

yadij (approved):

PR #1645, opened 2024-01-21
NTLM/Negotiate: Fix crash on bad helper TT responses

jtstrs:
Helper lookup may be made without a client HTTP Request,
(stored in lm_request->request). But in Helper::TT cases the
lm_request->request was dereferenced without any checks.


kinkie:
Ok to test

yadij (approved):
LGTM. We could perhapse handle this better with an exception. But the caller logic is not prepared for that eventuality - Handling it as a failed login is "good enough" for an immediate fix.

kinkie:
> LGTM. We could perhapse handle this better with an exception. But the caller logic is not prepared for that eventuality - Handling it as a failed login is "good enough" for an immediate fix.

could we maybe leave a XXX comment behind?

kinkie (approved):

squidadm:
queued for backport to v6

PR #1654, opened 2024-01-30
Maintenance: automate header guards 1/3

kinkie:
In preparation for merging the automated header-guards maintenance
script, merge the manual changes required for that script to succeed.

### DIFF
--- a/lib/rfcnb/byteorder.h
+++ b/lib/rfcnb/byteorder.h
@@ -34,6 +34,9 @@
  * int manipulation
  */
 
+#ifndef SQUID_LIB_RFCNB_BYTEORDER_H
+#define SQUID_LIB_RFCNB_BYTEORDER_H
+
 #undef CAREFUL_ALIGNMENT
 
 /* we know that the 386 can handle misalignment and has the "right"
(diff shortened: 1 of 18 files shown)


yadij (approved):

rousskov (approved):
> This is a precursor to land PR #1630

I removed this comment from the PR description. IMO, we should avoid using GitHub PR numbers in commit messages, especially while we are still abusing those numbers in backported commit titles.

I also updated PR description in hope to clarify the effect of these manual changes.

Disclaimer: After reading https://github.com/squid-cache/squid/pull/1630#issuecomment-1916379917, I still do not understand why these changes need to be committed separately from the new script. In other words, I do not understand why committing these changes with the script would result in "codeQL checks failing".

kinkie:
> Disclaimer: After reading [#1630 (comment)](https://github.com/squid-cache/squid/pull/1630#issuecomment-1916379917), I still do not understand why these changes need to be committed separately from the new script. In other words, I do not understand why committing these changes with the script would result in "codeQL checks failing".

I believe we should backport the script to 6.X: it does not affect functionality and we benefit by having the code alignment.
however while the script and manual adjustments should be easily backportable, the automated changes may not work: any file added, removed, or renamed would require fixing the merge one way or the other.

The backport strategy needs to be the same as for this one: backport the manual adjustments, run the script and land a specific 6.X PR, and only then land the script. Any other strategy will require manual fixes at some point or another.

rousskov:
> The backport strategy needs to be the same as for this one: backport the manual adjustments, run the script and land a specific 6.X PR, and only then land the script. Any other strategy will require manual fixes at some point or another.

There is nothing wrong with manual fixes per se. Clicking buttons to backport changes is obviously convenient, but that convenience does not always trump other considerations. FWIW, I am getting increasingly concerned that we are making things worse in master (and creating extra work) in order to avoid a bit of manual backporting work in v6.

PR #1658, opened 2024-02-01
Use AnyP::Uri::Decode() for url_regex checks

kinkie:


yadij (changes requested):
Two things for this specific case:

 1) IMO, the conversion to `Decode()` should be done only where the `rfc1738_unescape()` is forcing use of `char*` in a series of `SBuf` operations. This case we have a necessary conversion to `char*` regardless of the decode. So not (yet) a good candidate for removal of `rfc1738_unescape()`.

2) The `effectiveRequestUri()` in current v7 code is normalized with inappropriate encoding removed. Please investigate the users of this `match()` method and if possible just drop the `rfc1738_unescape()` entirely.

kinkie:
> 1. IMO, the conversion to `Decode()` should be done only where the `rfc1738_unescape()` is forcing use of `char*` in a series of `SBuf` operations. This case we have a necessary conversion to `char*` regardless of the decode. So not (yet) a good candidate for removal of `rfc1738_unescape()`.

The main problem here is that ACLServerNameData::match() should use SBuf, not char* . I have refrained from doing that in order to avoid balooning the scope of this PR

yadij:
Hmm. `ACLServerNameData::match()` should not be using `request->effectiveRequestUri()` value in any way. It should be using the `request->uri.host()` value directly.

src/acl/Url.cc
```diff
@@ -11,16 +11,15 @@
 #include "squid.h"
 #include "acl/FilledChecklist.h"
 #include "acl/Url.h"
+#include "anyp/Uri.h"
```
rousskov, line 14: > Amos: Two things for this specific case:
> 
> 1. IMO, the conversion to `Decode()` should be done only where the `rfc1738_unescape()` is forcing use of `char*` in a series of `SBuf` operations.

I disagree: IMO, **all** rfc1738_unescape() should be eventually converted (or removed, etc.). We may prioritize certain use cases over others, of course, but the overall goal should be to get rid of the very problematic rfc1738_unescape().

> This case we have a necessary conversion to `char*` regardless of the decode.

SBuf returned by Decode() can be converted to a c-string so the necessity of that conversion alone is not a disqualifying condition.

> So not (yet) a good candidate for removal of `rfc1738_unescape()`.

That may still be true (I do not know), but not for the reasons given.

> 2. The `effectiveRequestUri()` in current v7 code is normalized with inappropriate encoding removed.

What do you mean by "inappropriate" here?

I do not see any decoding in effectiveRequestUri(). In fact, the method _encodes_ userinfo (via Uri::absolute()) and probably uses _encoded_ URL whitespace in URI_WHITESPACE_ENCODE cases (see AnyP::Uri::parse()). What am I missing?

src/acl/Url.cc
```diff
@@ -11,16 +11,15 @@
...
 int
 Acl::UrlCheck::match(ACLChecklist * const ch)
 {
     const auto checklist = Filled(ch);
 
-    char *esc_buf = SBufToCstring(checklist->request->effectiveRequestUri());
-    rfc1738_unescape(esc_buf);
+    char *esc_buf = SBufToCstring(AnyP::Uri::Decode(checklist->request->effectiveRequestUri()));
     int result = data->match(esc_buf);
     xfree(esc_buf);
```
rousskov, line 24: Please do not allocate c-strings where that allocation can be avoided:

```suggestion
    // TODO: Consider refactoring so that effectiveRequestUri() returns decoded URI.
    auto decodedUri = AnyP::Uri::Decode(checklist->request->effectiveRequestUri());
    const auto result = data->match(decodedUri.c_str());
```

The above suggestion also follows AAA and improves const-correctness. Please always _start_ with `const auto` or equivalent to minimize the volume of such corrections.

src/acl/Url.cc
rousskov, line 14:
```diff
- Maintenance: adopt AnyP::Uri::Decode [1/n]
+ Use AnyP::Uri::Decode() for url_regex checks
```

I have adjusted PR title to remove the implication that these changes have no important runtime effects: The new Decode() is quite different from the old rfc1738_unescape() in its treatment of invalid encodings! I also removed `[1/n]` counter because it has virtually no useful purpose[^1] while requiring maintenance/synchronization across PRs, potentially creating conflicts and disagreements.

This comment does not request any PR changes but please keep it in mind when working on other PRs.

[^1]: One can find relevant PRs/commits by searching for "Use AnyP::Uri::Decode" and "Drop rfc1738_unescape" in titles and ordering those PRs/commits any way they want.

rousskov (changes requested):
Thank you very much for using a piecemeal approach to rfc1738_unescape() replacement! I hope this approach helps us increase discussions efficiency and make steady progress. I recommend keeping the number of open related PRs capped at two.

rousskov:
> Amos: `ACLServerNameData::match()` should not be using `request->effectiveRequestUri()` value in any way. It should be using the `request->uri.host()` value directly.

This discussion feels outside this PR scope. Are we sure we want to continue it (in this PR)? If at all possible, let's focus on the specific conversion this PR is proposing. Acl::UrlCheck::match (i.e. url_regex acl) does not use ACLServerNameData (used by server_name acl), right?

src/acl/Url.cc
kinkie: > I disagree: IMO, **all** rfc1738_unescape() should be eventually converted (or removed, etc.). We may prioritize certain use cases over others, of course, but the overall goal should be to get rid of the very problematic rfc1738_unescape().

This is my general intent

rousskov: > What I am getting from all this is that our goal contexts still differ despite the weekly meetings discussion of what the goal and approach should be.

I do not recall any discussions that contradicted what I have said in my response. I also do not recall us finalizing "goal and approach" of rfc1738_unescape() replacements in any relevant way. If you believe that we have agreed on something specific that (is relevant to this or related future PRs and) contradicts my comments, then I recommend posting that alleged agreement for a confirmation/review. In my experience, the probability of the three of us having the _same_ interpretation of an informal verbal discussion about complex technical matters is asymptotically approaching zero.

> Also, refer to meaning of "IMO".

AFAICT, the meaning of the phrase "in my opinion" does not contradict what I have said. I do not know why "IMO meaning" is mentioned twice in your response. If you feel it is important to clarify this, please detail your concerns.

rousskov (approved):
I have committed my suggestion after Amos' "thumbs up" reaction to it. Please feel free to merge if you are happy with the new code and it passes your tests.

kinkie:
Thanks @rousskov ; clearing for merge

yadij (approved):

PR #1661, opened 2024-02-03
Maintenance: Detect mismatching .instructions files

eduard-bagdasaryan:


rousskov (approved):
@eduard-bagdasaryan, please double check that my adjustments did not screw something up. Thank you.

rousskov:
I asked Eduard to add mismatching .instruction files detection because we had a case of a buggy internal PR that passed all the tests simply because one of the (correct) instruction files had a typo in its filename and, hence, was ignored by "make distcheck". This PR should prevent similar accidents.

The "opposite" test is not possible -- not all configuration files have a corresponding instruction file.

eduard-bagdasaryan:
> double check that my adjustments

Your adjustments look correct.

yadij (approved):
FTR; you may recall issues like this were amongst the reasons I put forward against use of this separation between instructions and test data.

rousskov:
> you may recall issues like this were amongst the reasons I put forward against use of this separation between instructions and test data.

Yes, this is one of the ([undisclosed](https://github.com/squid-cache/squid/pull/1125#discussion_r951626783) but natural) "cons" of the current design. I am glad that we will now partially mitigate it.

PR #1666, opened 2024-02-09
Fix MemObject.cc:123: "!updatedReply_" assertion

rousskov:
MemStore.cc(506) copyFromShm: entry 1284 copied ...
    FATAL: assertion failed: MemObject.cc:123: "!updatedReply_"

The code loading a response from the shared memory cache incorrectly
assumed that the being-loaded response could not have been updated by an
HTTP 304 (Not Modified) reply and called adjustableBaseReply() that is
banned for updated responses. The goal of that call was to determine
whether the cached response header has been parsed. That determination
can be made without using a method that is banned for updated responses.

StoreClient and clientReplyContext had very similar checks that used the
right/safe approach because their current code did not need an immediate
access to an "adjustable" response. We have updated all external
psParsed checks to reduce chances that this bug resurfaces.

### DIFF
--- a/src/client_side_reply.cc
+++ b/src/client_side_reply.cc
@@ -1062,7 +1062,7 @@ clientReplyContext::storeNotOKTransferDone() const
     assert(mem != nullptr);
     assert(http->request != nullptr);
 
-    if (mem->baseReply().pstate != Http::Message::psParsed)
+    if (!http->storeEntry()->hasParsedReplyHeader())
         /* haven't found end of headers yet */
         return 0;
 
--- a/src/store.cc
+++ b/src/store.cc
@@ -228,6 +228,19 @@ StoreEntry::bytesWanted (Range<size_t> const aRange, bool ignoreDelayPools) cons
 }
 
 bool
+StoreEntry::hasParsedReplyHeader() const
+{
+    if (mem_obj) {
+        const auto &reply = mem_obj->baseReply();
+        if (reply.pstate == Http::Message::psParsed) {
+            debugs(20, 7, reply.hdr_sz);
+            return true;
+        }
+    }
+    return false;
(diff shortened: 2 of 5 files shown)


src/store.cc
```diff
@@ -227,6 +227,19 @@ StoreEntry::bytesWanted (Range<size_t> const aRange, bool ignoreDelayPools) cons
     return mem_obj->mostBytesWanted(aRange.end, ignoreDelayPools);
 }
 
+bool
+StoreEntry::hasParsedReplyHeader() const
+{
+    if (mem_obj) {
+        const auto &reply = mem_obj->baseReply();
+        if (reply.pstate == Http::Message::psParsed) {
+            debugs(20, 7, reply.hdr_sz);
```
rousskov, line 236: (Eventual) access to hdr_sz is one of the two primary reasons for calling this method. Printing hdr_sz value helps in triage, but I do not insist on this additional debugging. If you are against this debugging, please let me know.

src/client_side_reply.cc
```diff
@@ -1062,7 +1062,7 @@ clientReplyContext::storeNotOKTransferDone() const
     assert(mem != nullptr);
     assert(http->request != nullptr);
 
-    if (mem->baseReply().pstate != Http::Message::psParsed)
+    if (!http->storeEntry()->hasParsedReplyHeader())
```
yadij, line 1065: Please consider moving the `mem` object creation after this now that it is no longer needed beforehand. I do not insist on this.

yadij (approved):

src/client_side_reply.cc
rousskov: Sorry, this PR was merged before I saw your change request.

PR #1671, opened 2024-02-13
Fix type mismatch in new/delete of addrinfo::ai_addr

jtstrs:
ERROR: AddressSanitizer: new-delete-type-mismatch...:
    size of the allocated type: 28 bytes;
    size of the deallocated type: 16 bytes.
    in Ip::Address::InitAddr(addrinfo*&) src/ip/Address.cc:678
    in Ip::Address::FreeAddr(addrinfo*&) src/ip/Address.cc:690

* `delete` used addrinfo::ai_addr type (i.e. sockaddr)
* some `new` calls use sockaddr_in type
* some `new` calls use sockaddr_in6 type

Mismatching new/delete types result in undefined behavior. However,
since these are all PODs, the bug may not have resulted in runtime
problems in most environments, even though sockaddr_in6 is usually
larger than sockaddr and sockaddr_in tructure (e.g., 28 vs. 16 bytes).

### DIFF
--- a/src/ip/Address.cc
+++ b/src/ip/Address.cc
@@ -43,7 +43,7 @@
 namespace Ip {
 
 template <class SockAddrType>
-static void AllocateAddrMember(struct addrinfo &);
+static SockAddrType &AllocateAddrMember(struct addrinfo &);
 static void FreeAddrMember(struct addrinfo &);
 
 } // namespace Ip
@@ -642,9 +642,9 @@ Ip::Address::getAddrInfo(struct addrinfo *&dst, int force) const
         dst->ai_protocol = IPPROTO_UDP;
 
     if (force == AF_INET6 || (force == AF_UNSPEC && isIPv6()) ) {
-        AllocateAddrMember<struct sockaddr_in6>(*dst);
-        getSockAddr(*(reinterpret_cast<struct sockaddr_in6*>(dst->ai_addr)));
-        dst->ai_family = (reinterpret_cast<struct sockaddr_in6*>(dst->ai_addr))->sin6_family;
+        auto &ai_addr = AllocateAddrMember<struct sockaddr_in6>(*dst);
+        getSockAddr(ai_addr);
+        dst->ai_family = ai_addr.sin6_family;
 #if 0
         /**
          * Enable only if you must and please report to squid-dev if you find a need for this.
@@ -660,9 +660,9 @@ Ip::Address::getAddrInfo(struct addrinfo *&dst, int force) const
 #endif
 
     } else if ( force == AF_INET || (force == AF_UNSPEC && isIPv4()) ) {
-        AllocateAddrMember<struct sockaddr_in>(*dst);
-        getSockAddr(*(reinterpret_cast<struct sockaddr_in*>(dst->ai_addr)));
-        dst->ai_family = (reinterpret_cast<struct sockaddr_in*>(dst->ai_addr))->sin_family;
+        auto &ai_addr = AllocateAddrMember<struct sockaddr_in>(*dst);
+        getSockAddr(ai_addr);
+        dst->ai_family = ai_addr.sin_family;
     } else {
         IASSERT("false",false);
     }
@@ -695,19 +695,27 @@ Ip::Address::FreeAddr(struct addrinfo *&ai)
     ai = nullptr;
 }
 
+/// Allocates and fills with zeros a SockAddrType object.
+/// Initializes addrinfo::ai_addr and ai_addrlen with the allocated object.
+/// \returns the allocated object
 template <class SockAddrType>
-void
+SockAddrType &
 Ip::AllocateAddrMember(struct addrinfo &ai)
 {
     static_assert(
         std::is_same<SockAddrType, struct sockaddr_in>::value ||
         std::is_same<SockAddrType, struct sockaddr_in6>::value,
         "FreeAddrMember() supports this addrinfo::ai_addr type");
-    ai.ai_addr = reinterpret_cast<struct sockaddr*>(new SockAddrType);
+
+    const auto ai_addr = new SockAddrType;
     // We do not use `new SockAddrType{}` above instead of memset() below
     // because, since C++14, doing so may not initialize SockAddrType padding.
-    memset(ai.ai_addr, 0, sizeof(SockAddrType));
-    ai.ai_addrlen = sizeof(SockAddrType);
+    memset(ai_addr, 0, sizeof(*ai_addr));
+
+    ai.ai_addr = reinterpret_cast<struct sockaddr*>(ai_addr);
+    ai.ai_addrlen = sizeof(*ai_addr);
+
+    return *ai_addr;
 }
 
 void
--- a/src/ip/Address.cc
+++ b/src/ip/Address.cc
@@ -685,8 +685,7 @@ Ip::Address::InitAddr(struct addrinfo *&ai)
 void
 Ip::Address::FreeAddr(struct addrinfo *&ai)
 {
-    if (!ai)
-        return;
+    if (ai == nullptr) return;
 
     FreeAddrMember(*ai);
 
--- a/src/ip/Address.cc
+++ b/src/ip/Address.cc
@@ -710,6 +710,7 @@ Ip::AllocateAddrMember(struct addrinfo &ai)
     const auto ai_addr = new SockAddrType;
     // We do not use `new SockAddrType{}` above instead of memset() below
     // because, since C++14, doing so may not initialize SockAddrType padding.
+    static_assert(std::is_trivial<SockAddrType>::value, "can memset() this addrinfo::ai_addr type");
     memset(ai_addr, 0, sizeof(*ai_addr));
 
     ai.ai_addr = reinterpret_cast<struct sockaddr*>(ai_addr);
--- a/src/ip/Address.cc
+++ b/src/ip/Address.cc
@@ -725,15 +725,14 @@ Ip::AllocateAddrMember(struct addrinfo &ai)
 void
 Ip::FreeAddrMember(struct addrinfo &ai)
 {
-    if (!ai.ai_addr) {
+    if (!ai.ai_addr)
         return;
-    }
 
-    if (ai.ai_addrlen == sizeof(struct sockaddr_in)) {
+    if (ai.ai_addrlen == sizeof(struct sockaddr_in))
         delete reinterpret_cast<struct sockaddr_in*>(ai.ai_addr);
-    } else if (ai.ai_addrlen == sizeof(struct sockaddr_in6)) {
+    else if (ai.ai_addrlen == sizeof(struct sockaddr_in6))
         delete reinterpret_cast<struct sockaddr_in6*>(ai.ai_addr);
-    } else {
+    else {
         debugs(14, DBG_CRITICAL, "ERROR: Squid Bug: Unexpected addrinfo::ai_addr size: " << ai.ai_addrlen <<
                Debug::Extra << "sockaddr_in size: " << sizeof(struct sockaddr_in) <<
                Debug::Extra << "sockaddr_in6 size: " << sizeof(struct sockaddr_in6));
--- a/src/ip/Address.cc
+++ b/src/ip/Address.cc
@@ -707,6 +707,8 @@ Ip::AllocateAddrMember(struct addrinfo * const ai)
         std::is_same<SockAddrType, struct sockaddr_in6>::value,
         "FreeAddrMember() supports this addrinfo::ai_addr type");
     ai->ai_addr = reinterpret_cast<struct sockaddr*>(new SockAddrType);
+    // We do not use `new SockAddrType{}` above instead of memset() below
+    // because, since C++14, doing so may not initialize SockAddrType padding.
     memset(ai->ai_addr, 0, sizeof(SockAddrType));
     ai->ai_addrlen = sizeof(SockAddrType);
 }
--- a/src/ip/Address.cc
+++ b/src/ip/Address.cc
@@ -44,13 +44,9 @@
 
 namespace Ip {
 
-enum class SockAddrType {
-    SockAddrIn,
-    SockAddrIn6
-};
-
-static void initSockAddr(struct addrinfo *, SockAddrType);
-static void releaseSockAddr(struct addrinfo *);
+template <class SockAddrType>
+static void AllocateAddrMember(struct addrinfo *);
+static void FreeAddrMember(struct addrinfo *);
 
 } // namespace Ip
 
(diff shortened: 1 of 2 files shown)


kinkie:
I think there is merit to this patch, however it is probably incorrect:

If you see the code around line 593 (operator =):
```
        // attempt to handle partially initialised addrinfo.
        // such as those where data only comes from getsockopt()
        if (s.ai_addr != nullptr) {
            if (s.ai_addrlen == sizeof(struct sockaddr_in6)) {
                operator=(*((struct sockaddr_in6*)s.ai_addr));
                return true;
            } else if (s.ai_addrlen == sizeof(struct sockaddr_in)) {
                operator=(*((struct sockaddr_in*)s.ai_addr));
                return true;
            }
        }
```

I think this patch should operate in a similar way and not blindly cast, but rely on the ai_addrlen to decide wheteher to cast.
Could you look into this?

kinkie:
> Sure, will update this. Probably I missed something in my analysis, I've based current change on assumption thats `InitAddr` and `FreeAddr` is pair function and there is always expected that `FreeAddr` get `ai->ai_addr` as `sockaddr_in6` type

Maybe I missed out on that as well, but if that is the case, an assertion that the `ai_addrlen` be consistent with this assumption wouldn't hurt

yadij (dismissed):
LGTM, I have added the relevant code polish to pass our current style guidelines

yadij:
OK to test

src/ip/Address.cc
```diff
@@ -695,9 +696,11 @@ Ip::Address::InitAddr(struct addrinfo *&ai)
 void
 Ip::Address::FreeAddr(struct addrinfo *&ai)
 {
-    if (ai == nullptr) return;
+    if (!ai)
+        return;
 
-    if (ai->ai_addr) delete ai->ai_addr;
+    if (ai->ai_addr)
+        delete static_cast<struct sockaddr*>(struct sockaddr_in6*>(ai->ai_addr);
```
rousskov, line 703: Please keep the two `delete ... ai_addr` calls in sync. I have not studied whether any of them need changing, but if one of them needs changing, then the other one must be changed as well.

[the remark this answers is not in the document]
rousskov: I share Francesco's concern. Unfortunately, even more (than suggested) changes are needed to do this right because the current official code _sometimes_ does not set ai_addrlen before or immediately after allocating ai_addr. We need to refactor _all_ constructing code, so that we can rely on ai_addrlen during destruction, and I strongly recommend encapsulating this ai_addr memory allocation/destruction logic into a pair of dedicated functions.

I found three places where ai_addr memory is allocated and two places where it is freed. Please double check and keep all of them in sync.

rousskov (changes requested):

rousskov:
Please also double check the adjusted PR title/description.

@jtstrs, have you seen this bug result in actual memory leaks with some common memory management library? I am just curious. It is a bug regardless of whether those memory leaks are common in some typical environments...

jtstrs:
@rousskov 
Actually there is just an ```new-delete-type-mismatch``` was caught with Address Sanitizer.
Memory leak was my assumption, after few checks. Sorry, if it added some unclearance

There typical log from ASan for this case:
```
=================================================================
==(squid-1)==20157==ERROR: AddressSanitizer: new-delete-type-mismatch on 0x603000115990 in thread T0:
object passed to delete has wrong type:
size of the allocated type: 28 bytes;
size of the deallocated type: 16 bytes.
#0 0x7f900f51b5a5 in operator delete(void*, unsigned long) (/usr/lib64/libasan.so.5+0x10a5a5)
1 0xefd1c7 in Ip::Address::FreeAddr(addrinfo*&) SQUID_PATH//squid-5.9/src/ip/Address.cc:690

0x603000115990 is located 0 bytes inside of 28-byte region [0x603000115990,0x6030001159ac)
allocated by thread T0 here:
#0 0x7f900f51a11f in operator new(unsigned long) (/usr/lib64/libasan.so.5+0x10911f)
1 0xefd05a in Ip::Address::InitAddr(addrinfo*&) SQUID_PATH/squid-5.9/src/ip/Address.cc:678

SUMMARY: AddressSanitizer: new-delete-type-mismatch (/usr/lib64/libasan.so.5+0x10a5a5) in operator delete(void*, unsigned long)
==(squid-1)==20157==HINT: if you don't care about these errors you may set ASAN_OPTIONS=new_delete_type_mismatch=0

Command: (squid-1) --kid squid-1 -f /etc/squid/squid_captive.conf
```

kinkie:
> @rousskov Actually there is just an `new-delete-type-mismatch` was caught with Address Sanitizer. Memory leak was my assumption. Sorry, if it added some unclearance

It looks to me like you found a genuine issue.
Not only that, Ip::Address is used quite a lot, which means it's also impactful.
@rousskov 's comment is spot on, I reckon that the best path forward is to ensure ai_addrlen is consistent with the actual size of what is being stored in ai_addr, and to use that information. Long term, the whole class should probably be redesigned but that's against the actual short-term objective of fixing the problem quickly

rousskov:
> Francesco: It looks to me like you found a genuine issue.

I would like to second Francesco's opinion: @jtstrs, your research/PR is very valuable, even if there are usually no memory leaks associated with the new/delete mismatch bugs you have uncovered.

@jtstrs, would you like to refactor PR changes to isolate ai_addr construction (including correct length setting and zeroing) and destruction into dedicated functions or should we take a stab at that first? Either way is fine; just let us know.

jtstrs:
I would prefer to stick approach with an encapsulation of construction/deleting ```ai_addr```

rousskov:
> I'll prefer to stick an approach with encapsulation of construction/deleting `ai_addr`

OK, we will wait for your changes before polishing this up. No rush!

src/ip/Address.cc
```diff
@@ -683,32 +677,73 @@ Ip::Address::InitAddr(struct addrinfo *&ai)
     }
 
     // remove any existing data.
-    if (ai->ai_addr) delete ai->ai_addr;
-
-    ai->ai_addr = (struct sockaddr*)new sockaddr_in6;
-    memset(ai->ai_addr, 0, sizeof(struct sockaddr_in6));
-
-    ai->ai_addrlen = sizeof(struct sockaddr_in6);
+    FreeAddrMember(*ai);
```
rousskov, line 680: We can also move this call into a new `else` clause for the above addrinfo allocation `if` statement.

src/ip/Address.cc
```diff
@@ -683,32 +677,73 @@ Ip::Address::InitAddr(struct addrinfo *&ai)
...
+
+    if (ai.ai_addrlen == sizeof(struct sockaddr_in)) {
+        delete reinterpret_cast<struct sockaddr_in*>(ai.ai_addr);
+    } else if (ai.ai_addrlen == sizeof(struct sockaddr_in6)) {
+        delete reinterpret_cast<struct sockaddr_in6*>(ai.ai_addr);
+    } else {
+        debugs(14, DBG_CRITICAL, "ERROR: Squid Bug: Unexpected addrinfo::ai_addr size: " << ai.ai_addrlen <<
+               Debug::Extra << "sockaddr_in size: " << sizeof(struct sockaddr_in) <<
+               Debug::Extra << "sockaddr_in6 size: " << sizeof(struct sockaddr_in6));
+        // leak memory
```
rousskov, line 740: Or we can assert() instead. Throwing in cleanup code is not a good idea.

rousskov (approved):
I pushed a few polishing adjustments and can now approve this PR. I am not requesting any further code changes in this PR.

@jtstrs, please review and _retest_ my changes. I could have missed something! Thank you.

jtstrs:
Changes looks good from my side.
Thank you for polishing

jtstrs:
I'll check it patch with ASan again and add comment with results little bit later

src/ip/Address.cc
```diff
@@ -683,32 +677,73 @@ Ip::Address::InitAddr(struct addrinfo *&ai)
...
+
+/// Deallocates addrinfo::ai_addr member of the given structure and adjusts that
+/// structure accordingly. Safe for structures with a nil ai_addr member.
+/// \sa Ip::AllocateAddrMember()
+void
+Ip::FreeAddrMember(struct addrinfo &ai)
+{
+    if (!ai.ai_addr) {
+        return;
+    }
```
yadij, line 730: Current style guidelines is not to add braces around single-line if-statements.
```suggestion
    if (!ai.ai_addr)
        return;
```

src/ip/Address.cc
```diff
@@ -683,32 +677,73 @@ Ip::Address::InitAddr(struct addrinfo *&ai)
...
+
+    if (ai.ai_addrlen == sizeof(struct sockaddr_in)) {
+        delete reinterpret_cast<struct sockaddr_in*>(ai.ai_addr);
+    } else if (ai.ai_addrlen == sizeof(struct sockaddr_in6)) {
+        delete reinterpret_cast<struct sockaddr_in6*>(ai.ai_addr);
```
yadij, line 735: Current style guidelines is not to add braces around single-line if-statements.
```suggestion
    if (ai.ai_addrlen == sizeof(struct sockaddr_in))
        delete reinterpret_cast<struct sockaddr_in*>(ai.ai_addr);
        
    else if (ai.ai_addrlen == sizeof(struct sockaddr_in6))
        delete reinterpret_cast<struct sockaddr_in6*>(ai.ai_addr);

```

src/ip/Address.cc
```diff
@@ -683,32 +677,73 @@ Ip::Address::InitAddr(struct addrinfo *&ai)
...
+    } else {
```
yadij, line 736: Squid also stores `struct sockaddr_storage` when the exact address type is unknown before being filled with values.

(Please double-check my suggestion. I am not sure if the `ai_addrlen` value actually survives the syscall()'s which fill in the address value.)

```suggestion
    else if (ai.ai_addrlen == sizeof(struct sockaddr_storage))
        delete reinterpret_cast<struct sockaddr_storage*>(ai.ai_addr);

    else {
```

src/ip/Address.cc
yadij: `assert()` would also be a bad idea for now. There are `struct sockaddr_un` stored for UDS sockets `Comm::Connection`. They have variable-length allocations.

yadij (changes requested):

rousskov (changes requested):

jtstrs:
Im little bit lost track of the conversation.
@rousskov, now I supposed to push my efforts to getting rid of a dynamic allocations for all usages of ```addrinfo``` related to ```recv``` and ```getsockname``` syscalls in scope of current PR?

rousskov:
> now I supposed to push my efforts to getting rid of a dynamic allocations for all usages of `addrinfo` related to `recv` and `getsockname` syscalls in scope of current PR?

I recommend a slightly different approach: Find one addrinfo use case where addrinfo object is dynamically allocated, its length argument is supplied to some system call like recvfrom() to be updated, but the object as a whole is not used for anything (e.g. it is not then given to some other function to do something else). That is the simplest use case I can think of. Post a pull request dedicated to removing that one unnecessary addrinfo usage. Make sure the new code handles any reasonable (for the given context) address and errors. Once that PR lands, we will know how to address several similar (and possibly even all of the remaining) cases. If that plan works, eventually, we will get back to this PR with no dynamic addrinfo allocation cases left except for those this PR already handles correctly.

Icmp4::Recv() is a candidate but there might be even simpler use cases (Icmp4::Recv() does feed that addrinfo object to one of the fifty[^1] Ip::Address assignment operators, but it is possible that you will find another Ip::Address method that accomplishes what we want without adding one).

[^1]: "Fifty" is a wild exaggeration, of course. There are "only" eight AFAICT.

jtstrs:
Got it

kinkie:
@jtstrs , any news? Can I help you move this forward?

jtstrs:
Sorry, didnt start yet. 
I've some medical issues. 
I'll continue work after ~2 weeks

kinkie:
> Sorry, didnt start yet

No worries, and best luck!

jtstrs:
I've started analysis of things which we discussed above, and I'm afraid, that I could'nt found any "dummy" usages
of addrinfo structure, even Icmp* classes use information provided by recvfrom function for some logging purposes (I mean Icmp::Log(...)). So I'm not sure that there is possible to fully remove any usage of addrinfo, but I think that there is a few places, where we can get rid of from InitAddr usages.

rousskov:
> > Find one addrinfo use case where addrinfo object is dynamically allocated, its length argument is supplied to some system call like recvfrom() to be updated, but the object as a whole is not used for anything (e.g. it is not then given to some other function to do something else).

> I've started analysis of things which we discussed above, and I'm afraid, that I could'nt found any "dummy" usages of addrinfo structure, even Icmp* classes use information provided by recvfrom function for some logging purposes (I mean Icmp::Log(...)). So I'm not sure that there is possible to fully remove any usage of addrinfo

Icmp::Log() does not take addrinfo. It takes Ip::Address. Thus, Icmp::Log() is a red herring on the path to removing excessive addrinfo use.

If you are studying Icmp4::Recv(), then decide what recvfrom() argument to use to store the "from" address and then how to convert that filled "from" address to Ip::Address (i.e. to `preply.from`). That new argument type should be byte-compatible with sockaddr structure layout (i.e. the type required by recvfrom()), of course. Most likely, that argument will not be allocated on the heap.

One possible (but not guaranteed) way to find an acceptable solution for the above problem is the following process:

1. What is the overall best solution for correctly calling recvfrom() without using any 3rd party libraries?
2. How to implement that solution in Icmp4::Recv() context?

src/ip/Address.cc
yadij: FTR; these were an attempt to standardize on one C-type for the IP::Address API. It did not get completed and the individual `sockaddr_*` type methods removed. `struct addrinfo` was chosen as it is the formal type used by modern IPv6-enabled APIs. It was begun before I learned of `sockaddr_storage` which is essentially the C equivalent of what Ip::Address is supposed to be.

In hindsight the syscalls using any `sockaddr_*` type should have been converted to `sockaddr_storage`. So we have `addrinfo` for just the DNS related APIs and `sockaddr_storage` for the socket APIs. Dropping support of the other methods from `Ip::Address`

rousskov: @jtstrs, I believe Amos' recollection supports my [recommendation](https://github.com/squid-cache/squid/pull/1671#issuecomment-1969862840) and gives you an additional hint with regard to the right type to use for at least some of the system calls you are considering converting away from addrinfo.

jtstrs:
Got it, thanks for hints, will proceed analysis then

jtstrs:
Probably we could use ```sockaddr_storage``` type inside of Ip::Address class since its supposed to be wrapper for C api (according to @yadij said) and also wrap all system api calls inside of separate module like it's done in for example comm.cc:comm_udp_from, thus we get all interuction with this api in one place, also later there is possible to implement some wrappers for socket class instead of working with it handlers directly. What do you think?

rousskov:
> Probably we could use `sockaddr_storage` type inside of Ip::Address class ... and also wrap all system api calls inside of separate module ... What do you think?

Ip::Address does need a lot of changes, but a major development like changing Ip::Address storage type is way outside this PR scope (regardless of its long-term merits). Correctly creating a new module to "wrap all system API calls" is also a huge undertaking. With all due respect, this kind of development should not be _driven_ by a new Squid developer -- correct changes on those paths require too many Squid-specific (and difficult) decisions.

Please start small instead! Let's see how we can solve a few isolated/small problems first. Use Ip::Address as needed; add or adjust a method or two if you must, but stop if class changes snowball beyond that. Use sockaddr_storage as/if needed. Eliminate one or two specific addrinfo use cases. We will then decide how to generalize your fixes.

This PR changes may affect Ip::Address, its future rewrites, and even creation of new modules, of course, but I hope that we do not need to cross that bridge right here, right now, before we have settled a single addrinfo removal change.

yadij:
Alternative solution taken up in PR #2136

PR #1672, opened 2024-02-13
Bug 5343: Fix GCC v14 not finding std::find()

kinkie:
Reply.cc:198: error: no matching function for call to find(...)

The required STL header was missed in 2023 commit 27c3677.

PR #1673, opened 2024-02-13
Bug 5343: Fix remaining GCC v14 build issues

yadij:


src/base/InstanceId.h
```diff
@@ -49,6 +49,7 @@ class InstanceId
     typedef ValueType Value; ///< id storage type
 
     InstanceId() {change();}
+    InstanceId(const InstanceId &); ///< no copying; IDs are unique
```
kinkie, line 52: How about marking this explicitly?
No difference in practice, but more clarity on intent

```suggestion
    // work around gcc-14 bug
    InstanceId(const InstanceId &) = delete; ///< no copying; IDs are unique
```

rousskov, line 52: AFAICT, this PR change starts allowing InstanceId copying by assignment. There are two ways to fix that, depending on whether we want to (continue to) ban moving:

To allow[^1] moving InstanceId objects by explicitly declaring default move constructor and move assignment operator:

```suggestion
    // no copying (because IDs are unique) but moving is OK
    InstanceId(const InstanceId &) = delete;
    InstanceId &operator=(const InstanceId &) = delete;
    InstanceId(InstanceId &&) = defailt;
    InstanceId &operator=(InstanceId &&) = default;
```

To (continue to) prohibit moving InstanceId objects, use this idiomatic shorter variant that prohibits everything at once:

```suggestion
    InstanceId(InstanceId &&) = delete; // no copying/moving of any kind
```


[^1]: A user-declared copy constructor inhibits move members. See page 25+ of the well-known "Special Members" [slides](https://accu.org/conf-docs/PDFs_2014/Howard_Hinnant_Accu_2014.pdf).

rousskov (changes requested):

yadij:
All build tests work here for my Debian environment. Waiting now on PR #1672 to go in first and the wider build farm confirmation.

rousskov (approved):
Thank you for fixing this PR. I have adjusted PR title to match #1672 and removed PR description (that contained a PR comment rather than a future commit message). IMO, the bug number is enough to link these two changes together, but I am not against referencing the first commit from the second PR as well (or even merging the two PRs into one!).

uhliarik:
I can confirm that this PR fixes build failure on Fedora rawhide which contains GCC v14. Thanks for the fix!

kinkie (approved):
Shipit

squidadm:
queued for backport to v6

PR #1676, opened 2024-02-15
Portability: use printf(1) instead of echo -n

kinkie:
MacOS' implementation of echo(1) does not support the -n flag.

### DIFF
--- a/scripts/source-maintenance.sh
+++ b/scripts/source-maintenance.sh
@@ -616,12 +616,12 @@ printRawAmFile ()
     echo "## $GeneratedByMe"
     echo
 
-    printf "$1 ="
+    printf "%s =" "$1"
     # Only some files are formed from *.po filenames, but all such files
     # should list *.lang filenames instead.
     git ls-files $2$3 | sed -e s%$2%%g -e 's%\.po%\.lang%g' | while read f; do
         echo " \\"
-        printf "    ${f}"
+        printf "    %s" "${f}"
     done
     echo ""
 }
--- a/scripts/source-maintenance.sh
+++ b/scripts/source-maintenance.sh
@@ -620,7 +620,7 @@ printRawAmFile ()
     # Only some files are formed from *.po filenames, but all such files
     # should list *.lang filenames instead.
     git ls-files $2$3 | sed -e s%$2%%g -e 's%\.po%\.lang%g' | while read f; do
-        printf ' \\\n\t%s' "${f}" 
+        printf ' \\\n\t%s' "${f}"
     done
     printf '\n'
 }
--- a/scripts/source-maintenance.sh
+++ b/scripts/source-maintenance.sh
@@ -620,10 +620,9 @@ printRawAmFile ()
     # Only some files are formed from *.po filenames, but all such files
     # should list *.lang filenames instead.
     git ls-files $2$3 | sed -e s%$2%%g -e 's%\.po%\.lang%g' | while read f; do
-        echo " \\"
-        printf "    %s" "${f}"
+        printf ' \\\n\t%s' "${f}" 
     done
-    echo ""
+    printf '\n'
 }
 
 generateAmFile ()
--- a/scripts/source-maintenance.sh
+++ b/scripts/source-maintenance.sh
@@ -616,12 +616,12 @@ printRawAmFile ()
     echo "## $GeneratedByMe"
     echo
 
-    echo -n "$1 ="
+    printf "$1 ="
     # Only some files are formed from *.po filenames, but all such files
     # should list *.lang filenames instead.
     git ls-files $2$3 | sed -e s%$2%%g -e 's%\.po%\.lang%g' | while read f; do
         echo " \\"
-        echo -n "    ${f}"
+        printf "    ${f}"
     done
     echo ""
 }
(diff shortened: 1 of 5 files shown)


yadij (dismissed):

scripts/update-pot.sh
```diff
@@ -49,7 +49,7 @@ msgcat ./pot/*.pot -s -o errpages.pot.new &&
 
 ## Update all existing dictionaries with the new content ...
 for f in `ls -1 ./*.po` ; do
-	echo -n "Update: ${f} ... "
+	printf "Update: ${f} ... "
```
rousskov, line 52: Here and in several similar places where external input becomes a part of the printf format string, please use a "%s" or a similar format string instead. I am aware that the probability of a file name or similar input containing `%` or other special characters is small in these contexts. Still we should not abuse the format string like that.

rousskov (commented):
Thank you for fixing this. I do not insist on any changes, but please do consider addressing my change request.

I have shortened PR description by removing the information already provided in PR title.

kinkie:
Note: I've changed the output in errors/, because:
- the current shell syntax is very noisy and silences all output
- po2html takes a long while, and so the no-newline approach works very poorly because output will overlap on parallel builds. Will now output two lines per language

@yadij review was dismissed by the changes; would you mind re-reviewing?

scripts/source-maintenance.sh
```diff
@@ -616,12 +616,12 @@ printRawAmFile ()
...
     echo
 
-    echo -n "$1 ="
+    printf "%s =" "$1"
     # Only some files are formed from *.po filenames, but all such files
     # should list *.lang filenames instead.
     git ls-files $2$3 | sed -e s%$2%%g -e 's%\.po%\.lang%g' | while read f; do
         echo " \\"
-        echo -n "    ${f}"
+        printf "    %s" "${f}"
```
yadij, line 624: These lines are just supposed to produce folded-whitespace. Now that we are using `printf` here the initial `echo` can be merged into the format string properly and that weird extra line elided.

```
    printf " \\\n\t%s" "${f}"
```

yadij (changes requested):

scripts/source-maintenance.sh
rousskov: FWIW, I find the current PR code (two code lines producing two output lines) more readable than the proposed single-line code that prints two lines.

kinkie: I agree. @yadij is changing this a dealbreaker for you?

yadij: I believe that @rousskov already mistaking this as adding **two** lines of output is a good evidence of why this should **not** be done as two statements.

This code loop is actually only printing one line in the yoda-style form: `line-terminator line-text`

If this code were producing a normal text list for humans this code would look like:
 . `printf "\n%s" "${f}"`

However, automake build tools require folded-whitespace lines. Which have a fancy terminator:
 . ` line := *CHAR ( SP '\' LF HTAB )`
instead of the usual
 . ` line := *CHAR ( *CR LF )`

And thus we end up with:
 . `printf " \\\n\t%s" "${f}"`

rousskov: > I believe that @rousskov already mistaking this as ...

Again and again, please do not review people.

FWIW, I know what this code does and (still) think that two code lines are better than one in this case. Furthermore, since this PR does not have to modify the first `echo` line at all, one can classify this change request itself (and, hence, the discussion of its merits) to be outside this PR scope.

If you do want to polish this code in this PR, then adding comments (to clarify that the first code line terminates the "current" output line and the second code line starts the "next" output line) may be a good idea, but it may more than one iteration to get those comments right.

yadij: The reason for my request is to remove an old hack which is no longer necessary. If you are going to insist on splitting the line terminator across multiple lines. Why not print three lines? that makes as much sense as insisting a hack remains untouched. IMO the fix both easy and relevant to the current PR.

kinkie: The tab had been expanded in translation sometime, and was showing up as a number of spaces.
printf "\t" is clearer, so switching to that.
For the sake of consistency, I'll also turn the terminating `echo ""` into a printf '\n'

yadij (dismissed):

yadij (approved):

rousskov:
> this PR got stuck because Jenkins build [failed.](https://build.squid-cache.org/job/5-pr-auto/1884/COMPILER=clang,OS=debian-testing,label=pr-merge-builder/console). The commit shows a misleading green check because Jenkins did not report failure status.

Just to clarify, this PR is waiting for Jenkins to report some of the expected status checks for a staged commit to GitHub: All of the _reported_ statuses are green, but Jenkins did not report some of the statuses at all. The unreported check(s) have failed on Jenkins side, but GitHub does not know that. Similar Jenkins problems appear to affect pre-staging Jenkins checks as well (e.g., in #1708).

kinkie:
Jenkins had a bunch of defunct ssh connections. I restarted it, taking the chance to update it,
It should be okay now

yadij (approved):

squidadm:
queued for backport to v6

PR #1677, opened 2024-02-15
Remove unnecessary C linkage and unused code

kinkie:
Also upgrade lib/util.c to C++.


yadij (approved):
Thank you for these removals.

lib/util.cc
```diff
@@ -73,19 +73,6 @@ xint64toa(int64_t num)
     return buf;
```
rousskov, line 73: Some `#include` statements in this used-to-be-C file can probably be upgraded (including the removal of their guards), but I do not insist on that.

rousskov (approved):
Thank you. I have simplified the future commit message, disclosed util.c upgrade, and clarified "dead" with "unused".

PR #1683, opened 2024-02-16
Fix debugging for responses that Expire at check time

rousskov:
Since 2000 commit 65fa5c6, our level-3 debugging mislead about Expires
being less than the check time when the two times were identical. The
actual checked conditions are correct: Roughly speaking, the response
with Expires value T is considered expired at that time T.

Also dropped extra (and inconsistent) trailing space on debugs() lines.
This space was added by the same 2000 commit, probably accidentally.

### DIFF
--- a/src/refresh.cc
+++ b/src/refresh.cc
@@ -144,14 +144,10 @@ refreshStaleness(const StoreEntry * entry, time_t check_time, const time_t age, 
         sf->expires = true;
 
         if (entry->expires > check_time) {
-            debugs(22, 3, "FRESH: expires " << entry->expires <<
-                   " >= check_time " << check_time << " ");
-
+            debugs(22, 3, "FRESH: expires " << entry->expires << " > check_time " << check_time);
             return -1;
         } else {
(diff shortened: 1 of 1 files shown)


src/refresh.cc
```diff
@@ -144,14 +144,10 @@ refreshStaleness(const StoreEntry * entry, time_t check_time, const time_t age,
...
 
         if (entry->expires > check_time) {
-            debugs(22, 3, "FRESH: expires " << entry->expires <<
-                   " >= check_time " << check_time << " ");
-
+            debugs(22, 3, "FRESH: expires " << entry->expires << " > check_time " << check_time);
             return -1;
         } else {
-            debugs(22, 3, "STALE: expires " << entry->expires <<
-                   " < check_time " << check_time << " ");
```
rousskov, line 153: A minimal change would have been this one-character patch:

```diff
-                   " < check_time " << check_time << " ");
+                   " <= check_time " << check_time << " ");
```

yadij (approved):

squidadm:
queued for backport to v6

PR #1686, opened 2024-02-18
Maintenance: Use SQUID_CHECK_LIB_WORKS for PSAPI.dll detection

yadij:
Fixes an old bug with psapi.h detection not properly setting
HAVE_PSAPI_H.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -431,28 +431,20 @@ SQUID_STATE_ROLLBACK(LIBATOMIC)
 AC_SUBST(ATOMICLIB)
 
 SQUID_AUTO_LIB(psapi,[Windows PSAPI.dll],[LIBPSAPI])
-AS_IF([test "x$with_psapi" != "xno"],[
+SQUID_CHECK_LIB_WORKS(psapi,[
   SQUID_STATE_SAVE(squid_psapi_state)
   CPPFLAGS="$LIBPSAPI_CFLAGS $CPPFLAGS"
   LIBS="$LIBPSAPI_PATH $LIBS"
-  AC_CHECK_LIB([psapi],[K32GetProcessMemoryInfo],[LIBPSAPI_LIBS="-lpsapi"])
-  AC_CHECK_HEADER([psapi.h],,,[
+  AC_CHECK_LIB([psapi],[K32GetProcessMemoryInfo],[
+    LIBPSAPI_LIBS="-lpsapi"
+    AC_CHECK_HEADERS([psapi.h],,,[
 #if HAVE_WINDOWS_H
 #include <windows.h>
 #endif
+    ])
   ])
   SQUID_STATE_ROLLBACK(squid_psapi_state)
-  AS_IF([test "x$LIBPSAPI_LIBS" != "x"],[
-    CXXFLAGS="$LIBPSAPI_CFLAGS $CXXFLAGS"
-    LIBPSAPI_LIBS="$LIBPSAPI_PATH $LIBPSAPI_LIBS"
-  ],[test "x$with_psapi" = "xyes"],[
-    AC_MSG_ERROR([Required library PSAPI.dll not found])
-  ],[
-    AC_MSG_NOTICE([Library PSAPI.dll not found.])
-  ])
 ])
-AC_MSG_NOTICE([Using Windows PSAPI.dll library: ${with_psapi:=auto}])
-AC_SUBST(LIBPSAPI_LIBS)
 
 AC_SEARCH_LIBS([shm_open], [rt])
 AS_IF([test "x$ac_cv_search_shm_open" != "xno"],[


src/tools.cc
```diff
@@ -195,40 +195,28 @@ squid_getrusage(struct rusage *r)
...
-            } else {
-                CloseHandle( hProcess );
-                return;
+        // On Windows NT and later call PSAPI.DLL for process Memory information
+        auto hProcess = OpenProcess(PROCESS_QUERY_INFORMATION|PROCESS_VM_READ, FALSE, GetCurrentProcessId());
+        // Microsoft CRT doesn't have getrusage function,
+        // so we get process CPU time information from PSAPI.DLL.
+        FILETIME ftCreate, ftExit, ftKernel, ftUser;
+        if (GetProcessTimes(hProcess, &ftCreate, &ftExit, &ftKernel, &ftUser)) {
+            int64_t *ptUser = (int64_t *)&ftUser;
```
kinkie, line 204: why not using auto here?

configure.ac
```diff
@@ -432,28 +432,20 @@ SQUID_STATE_ROLLBACK(LIBATOMIC)
...
   LIBS="$LIBPSAPI_PATH $LIBS"
-  AC_CHECK_LIB([psapi],[K32GetProcessMemoryInfo],[LIBPSAPI_LIBS="-lpsapi"])
-  AC_CHECK_HEADER([psapi.h],,,[
+  AC_CHECK_LIB([psapi],[K32GetProcessMemoryInfo],[
+    LIBPSAPI_LIBS="-lpsapi"
+    AC_CHECK_HEADER([psapi.h],,,[
 #if HAVE_WINDOWS_H
 #include <windows.h>
 #endif
+    ])
```
rousskov, line 445: >  If the system header file header-file is compilable, execute shell commands action-if-found, otherwise execute action-if-not-found. If you just want to define a symbol if the header file is available, consider using AC_CHECK_HEADERS instead.

If I am interpreting AC_CHECK_HEADER() documentation (quoted above) correctly, this AC_CHECK_HEADER() call does not define HAVE_PSAPI_H. The same is true for the official configure.ac code that compiles into this (after the actual header presence check):

```shell
if test "x$ac_cv_header_psapi_h" = xyes
then :

fi
```

If the above analysis is correct, then Squid never defines HAVE_PSAPI_H and never actually includes psapi.h despite attempting to do so.

This comment is just FYI, in case you thought this code does something useful. I am not requesting any changes here.

src/tools.cc
rousskov: Yes, but start with a `const auto` attempt and downgrade to `auto` if you have to.

src/tools.cc
```diff
@@ -195,40 +195,28 @@ squid_getrusage(struct rusage *r)
...
-                int64_t tKernel64 = *ptKernel / 10;
-                r->ru_utime.tv_sec =(long)(tUser64 / 1000000);
-                r->ru_stime.tv_sec =(long)(tKernel64 / 1000000);
-                r->ru_utime.tv_usec =(long)(tUser64 % 1000000);
-                r->ru_stime.tv_usec =(long)(tKernel64 % 1000000);
-            } else {
-                CloseHandle( hProcess );
-                return;
+        // On Windows NT and later call PSAPI.DLL for process Memory information
+        auto hProcess = OpenProcess(PROCESS_QUERY_INFORMATION|PROCESS_VM_READ, FALSE, GetCurrentProcessId());
```
rousskov, line 199: If this PR polishes this code, please consider checking for OpenProcess() errors.

rousskov (commented):
I have not done a thorough review of this problematic Windows-specific code, and I am not requesting any changes.

src/tools.cc
```diff
@@ -195,40 +195,28 @@ squid_getrusage(struct rusage *r)
...
+        // Microsoft CRT doesn't have getrusage function,
+        // so we get process CPU time information from PSAPI.DLL.
+        FILETIME ftCreate, ftExit, ftKernel, ftUser;
+        if (GetProcessTimes(hProcess, &ftCreate, &ftExit, &ftKernel, &ftUser)) {
+            int64_t *ptUser = (int64_t *)&ftUser;
+            int64_t tUser64 = *ptUser / 10;
+            int64_t *ptKernel = (int64_t *)&ftKernel;
+            int64_t tKernel64 = *ptKernel / 10;
+            r->ru_utime.tv_sec =(long)(tUser64 / 1000000);
```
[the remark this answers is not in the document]
yadij: I do not have a build environment to ensure that a `static_cast` is correct, these may need to be `reinterpret_cast` instead. The changes done I am completely confident are just style that will not affect the built behaviour.

configure.ac
yadij: done in [7909a08](https://github.com/squid-cache/squid/pull/1686/commits/7909a08807a1958112f4d089abb10d9058cf4d00)

configure.ac
```diff
@@ -432,28 +432,20 @@ SQUID_STATE_ROLLBACK(LIBATOMIC)
...
-AS_IF([test "x$with_psapi" != "xno"],[
+SQUID_CHECK_LIB_WORKS(psapi,[
   SQUID_STATE_SAVE(squid_psapi_state)
   CPPFLAGS="$LIBPSAPI_CFLAGS $CPPFLAGS"
   LIBS="$LIBPSAPI_PATH $LIBS"
-  AC_CHECK_LIB([psapi],[K32GetProcessMemoryInfo],[LIBPSAPI_LIBS="-lpsapi"])
-  AC_CHECK_HEADER([psapi.h],,,[
+  AC_CHECK_LIB([psapi],[K32GetProcessMemoryInfo],[
+    LIBPSAPI_LIBS="-lpsapi"
+    AC_CHECK_HEADERS([psapi.h],,,[
```
rousskov, line 441: If Squid was never including psapi.h prior to this fix (despite attempting to do so), then I recommend mentioning it in the PR description.

yadij: I am not sure about "never" being true. It very likely was included indirectly by one of the other API files. For example; `ws2tcpip.h` and `winsock2.h` extend the `windows.h` API in quite a lot of unexpected ways.

src/tools.cc
kinkie: I'd keep polishing this code out of scope.
For instance, we can by now safely assume that (WIN32_OS_version >= _WIN_OS_WINNT) is always true, and we could factor it away.

src/tools.cc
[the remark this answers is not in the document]
kinkie: fair enough

src/tools.cc
rousskov: > I'd keep polishing this code out of scope.

Me too, but this PR has already crossed that line (unfortunately). That is why my change request has a precondition.

rousskov (commented):
I have no new suggestions, and I am not requesting any changes. Thank you for fixing psapi.h inclusion bug!

src/tools.cc
yadij: Very well, in order to make progress dropping all the src/tools.cc changes.

kinkie (approved):
Let's ship it, no point in holding back

rousskov:
I have updated PR description that became stale after recent PR branch changes.

PR #1687, opened 2024-02-18
Maintenance: update --with-heimdal-krb5 detection

yadij:


kinkie:
Looks good, but one question: if both gnugss and heimdal are present, which one should we favour and why?

yadij:
The existing code prefers MIT, Heimdal, GSS in that order. At present I am not changing that due to lack of understanding the specific reasons for preferences.

AFAIK some OS (re-)implementations of libgss lack a fully functional Kerberos or GSSAPI feature set. But it is still the most widely available by default in Linux and BSD distributions. So worth checking for.

kinkie (approved):
Thanks

rousskov (commented):
I am unable to efficiently review most of the changes in this PR.

PR #1688, opened 2024-02-18
Rename --with-gnugss to --with-gss

yadij:
Updating the library detection to use pkg-config and
bringing relevant macros up to current style.


rousskov (commented):
Sorry, I cannot efficiently review these changes. I trust they have been tested.

kinkie (approved):
Looks good!

PR #1690, opened 2024-02-20
Maintenance: MacOS compatiblity fix and github CI action

kinkie:
Improve MacOS compatibility
Add Github action workflow to test against regressions


kinkie:
Landing this will likely require changes to Anubis configuration

kinkie:
Output from successful run: https://github.com/kinkie/squid/actions/runs/7970186528/job/21757301890

.github/workflows/macos.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Checkout sources
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 5
+    - name: Prepare build
+      run: |
+        export PKG_CONFIG_PATH="`brew list openldap | grep ldap.pc | head -1 | xargs dirname`:`brew list cyrus-sasl | grep libsasl2.pc | head -1 | xargs dirname`";
+        export CFLAGS="-Wno-compound-token-split-by-macro ";
+        export CPPFLAGS="`pkg-config --cflags ldap libsasl2`";
+        export LDFLAGS="`pkg-config --libs ldap libsasl2`";
```
yadij, line 30: It would be best to update `configure.ac` to do these.

.github/workflows/macos.yaml
```diff
@@ -0,0 +1,38 @@
...
+        ./bootstrap.sh && \
+        mkdir btbuild && \
+        cd btbuild && \
+        ../configure
+    - name: Build It!
+      run: gmake -C btbuild -j $(nproc) all && gmake -C btbuild -j $(nproc) check
```
yadij, line 36: Make should be capable of building multiple targets. Like so:
```suggestion
      run: gmake -C btbuild -j $(nproc) all check
```

... though, `all` should be redundant as `check` is supposed to do that and more.

yadij (changes requested):

.github/workflows/macos.yaml
kinkie: In practice it isn't; the check target randomly fails at high levels of parallelism, I suspect that's due to us mixing noinst_*_SOURCES and *_SOURCES

.github/workflows/macos.yaml
kinkie: ldap-related configure.ac is a mess :\
Note that this is only relevant on macos, because:

.github/workflows/macos.yaml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  push:
+    branches: [ "master", "auto" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build:
+
+    runs-on: macos-latest
```
rousskov, line 12: Please move this build test to github/workflows/default.yaml matrix (or prove that it should not be integrated with the existing build test in default.yaml despite significant duplication of non-trivial code/logic).

.github/workflows/macos.yaml
rousskov: > ldap-related configure.ac is a mess

IMO, that assertion is not a valid excuse to create _new mess_ in GitHub Actions configuration.

Please note that it is perfectly fine to (modify configure.ac slightly) to make GitHub Actions work on macos _without_ enabling LDAP features (on macos). We do not have to spend a lot of resources to enable LDAP testing on macos in this PR. Moreover, the more custom code we need to add to enable such testing, the higher the probability that we should not add that custom code at all (and should effectively disable that testing instead).

rousskov (changes requested):

kinkie:
Closing this PR for now to not make too much noise; I'll continue the work in the branch aiming to:
1. simplify the macos specific parts
2. document what can't be simplified
3. fold into the default.yaml matrix

Once these are done, I'll reapply for joining. There might be other PRs coming to improve configure.ac

rousskov:
> There might be other PRs coming to improve configure.ac

... and perhaps a PR dedicated to that obsolete/deprecated vfork() call upgrade?

PR #1691, opened 2024-02-20
Fix marking of problematic cached IP addresses

rousskov:
Since inception in 2017 commit fd9c47d, Dns::CachedIps::have() always
returned position zero after finding a matching IP address (at zero or
positive position). The bug affected two callers:

* markAsBad() always marked the first stored address (as bad);
* forgetMarking() always cleared the first stored address marking.

Buggy markings led to Squid sometimes not attempting to use a working
address (e.g., IPv4) while using a known problematic one (e.g., IPv6).


kinkie (approved):
doh! Thanks for finding this

squidadm:
queued for backport to v6

PR #1692, opened 2024-02-20
Fix misleading Dns::CachedIps::restoreGoodness() debugging

rousskov:


src/ipcache.cc
```diff
@@ -974,9 +974,10 @@ Dns::CachedIps::restoreGoodness(const char *name)
         for (auto &cachedIp: ips)
             cachedIp.forgetMarking();
         badCount_ = 0;
+        debugs(14, 3, "cleared all " << size() << " bad IPs for " << name);
+        // fall through to reset goodPosition and report the current state
     }
     Must(seekNewGood(name));
-    debugs(14, 3, "cleared all IPs for " << name << "; now back to " << *this);
```
rousskov, line 979: This line was claiming that all IPs were cleared even when none were cleared (because the IPs are cleared in the `if` statement above). This misleading debugging complicated (and was discovered during) #1691 triage, but it does not affect the bug fixed there.

We do not need this debugging line at all because seekNewGood() provides enough debugging already.

kinkie (approved):
Thanks!

squidadm:
queued for backport to v6

PR #1696, opened 2024-02-21
Maintenance: increase portability of cf_gen automake rules

kinkie:
The MacOS linker uses a subdirectory named [target].dSYM to hold debug
information. Ensure it is removed by 'make clean'.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -708,7 +708,7 @@ uninstall-local: squid.conf.default
 	@$(SHELL) $(top_srcdir)/scripts/remove-cfg.sh "$(RM)" $(DESTDIR)$(DEFAULT_CONFIG_FILE) squid.conf.default
 
 distclean-local:
-	test -d "cf_gen.dSYM" && rm -rf "cf_gen.dSYM" || true
+	test -d "cf_gen.dSYM" && rm -rf "cf_gen.dSYM" || $(TRUE)
 
 CLEANFILES += cf.data squid.conf.default squid.conf.documented \
 	test_tools.cc *.a
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -156,9 +156,11 @@ bin_PROGRAMS =
 libexec_PROGRAMS = \
 	$(UNLINKD)
 
-cf_gen_SOURCES = cf_gen.cc cf_gen_defines.cci
+cf_gen_SOURCES = cf_gen.cc
+nodist_cf_gen_HEADER = cf_gen_defines.cci
 ## cf_gen must be stand-alone executable. It is a purely build-time executable.
 cf_gen_LDADD=
+cf_gen.$(OBJEXT): cf_gen_defines.cci
 
 ## cf_gen.cc needs src/cf_gen_defines.cci
 AM_CPPFLAGS += -I$(top_builddir)/src
@@ -622,6 +624,10 @@ test_cache_digest: test_cache_digest.o CacheDigest.o debug.o globals.o store_key
 ## If autodependency works well this is not needed anymore
 cache_cf.o: cf_parser.cci
 
+# cf_gen builds the configuration files.
+cf_gen$(EXEEXT): $(cf_gen_SOURCES) $(cf_gen_DEPENDENCIES) cf_gen_defines.cci
+	$(BUILDCXX) $(BUILDCXXFLAGS) -o $@ $(srcdir)/cf_gen.cc -I$(srcdir) -I$(top_builddir)/include/ -I$(top_builddir)/src
+
 # squid.conf.default is built by cf_gen when making cf_parser.cci
 squid.conf.default squid.conf.documented: cf_parser.cci
 	true
@@ -700,6 +706,9 @@ install-data-local: install-sysconfDATA install-dataDATA
 uninstall-local: squid.conf.default
 	@$(SHELL) $(top_srcdir)/scripts/remove-cfg.sh "$(RM)" $(DESTDIR)$(DEFAULT_MIME_TABLE) $(srcdir)/mime.conf.default
 	@$(SHELL) $(top_srcdir)/scripts/remove-cfg.sh "$(RM)" $(DESTDIR)$(DEFAULT_CONFIG_FILE) squid.conf.default
+
+distclean-local:
+	test -d "cf_gen.dSYM" && rm -rf "cf_gen.dSYM" || $(TRUE)
 
 CLEANFILES += cf.data squid.conf.default squid.conf.documented \
 	test_tools.cc *.a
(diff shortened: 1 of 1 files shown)


src/Makefile.am
```diff
@@ -707,6 +707,9 @@ uninstall-local: squid.conf.default
 	@$(SHELL) $(top_srcdir)/scripts/remove-cfg.sh "$(RM)" $(DESTDIR)$(DEFAULT_MIME_TABLE) $(srcdir)/mime.conf.default
 	@$(SHELL) $(top_srcdir)/scripts/remove-cfg.sh "$(RM)" $(DESTDIR)$(DEFAULT_CONFIG_FILE) squid.conf.default
 
+distclean-local:
```
rousskov, line 710: FWIW, I could not quickly figure out whether .dSYM should be deleted during "make clean" or "make distclean". Are you sure that "make clean" is not the right target for this?

src/Makefile.am
```diff
@@ -707,6 +707,9 @@ uninstall-local: squid.conf.default
...
+	test -d "cf_gen.dSYM" && rm -rf "cf_gen.dSYM" || $(TRUE)
```
rousskov, line 711: > The cf_gen target may leak this subdirectory for cf_gen

What makes cf_gen special? In other words, why do we need to add this special cleanup for cf_gen but not for other binaries? Perhaps the problem should be addressed by modifying how cf_gen is built?

rousskov (changes requested):
I committed a polishing fix; please check.

I am not requesting any changes at this point, just more info.

src/Makefile.am
kinkie: I think it could be be because it's the only noinst_PROGRAM we are buildingq

rousskov: Does temporary making cf_gen an installed program solves the problem (without this PR changes)?

src/Makefile.am
yadij: AIUI, if an object is generated by make `clean` should erase it. `distclean` is for the extra bits that can be re-generated by another build tool, eg re-running `configure`.

src/Makefile.am
kinkie: > Does temporary making cf_gen an installed program solves the problem (without this PR changes)?

Even if it did pass the build test, it would make no sense to keep it in, because by doing so it would cause cf_gen to be installed to a target system. The only reason to do it would be to possibly file a bug against autoconf

src/Makefile.am
kinkie: yes, clean-local works. Updating

src/Makefile.am
kinkie: Okay, I figured it out.
TL;DR: we override too much of autotools to build cf_gen. Updates coming up

rousskov (approved):
I trust you have tested that this change did not affect what files are archived by "make dist".

rousskov:
I polished PR description a little. I do not insist on those changes.

src/Makefile.am
```diff
@@ -624,10 +622,6 @@ test_cache_digest: test_cache_digest.o CacheDigest.o debug.o globals.o store_key
 ## If autodependency works well this is not needed anymore
 cache_cf.o: cf_parser.cci
 
-# cf_gen builds the configuration files.
-cf_gen$(EXEEXT): $(cf_gen_SOURCES) $(cf_gen_DEPENDENCIES) cf_gen_defines.cci
```
yadij, line 628: Removal of this rule causes cross-compile to break with "cf_gen.exe: cannot execute binary file: Exec format error"

yadij (changes requested):
This breaks build for me

src/Makefile.am
kinkie: reverted

src/Makefile.am
```diff
@@ -156,11 +156,9 @@ bin_PROGRAMS =
 libexec_PROGRAMS = \
 	$(UNLINKD)
 
-cf_gen_SOURCES = cf_gen.cc
-nodist_cf_gen_HEADER = cf_gen_defines.cci
+cf_gen_SOURCES = cf_gen.cc cf_gen_defines.cci
```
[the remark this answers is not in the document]
kinkie: yes. Reverted.

src/Makefile.am
```diff
@@ -707,6 +707,9 @@ uninstall-local: squid.conf.default
 	@$(SHELL) $(top_srcdir)/scripts/remove-cfg.sh "$(RM)" $(DESTDIR)$(DEFAULT_MIME_TABLE) $(srcdir)/mime.conf.default
 	@$(SHELL) $(top_srcdir)/scripts/remove-cfg.sh "$(RM)" $(DESTDIR)$(DEFAULT_CONFIG_FILE) squid.conf.default
 
+clean-local:
+	test -d "cf_gen.dSYM" && rm -rf "cf_gen.dSYM" || $(TRUE)
```
rousskov, line 711: Automake [documentation](https://www.gnu.org/software/automake/manual/html_node/Clean.html) suggests a simpler variant that uses a recipe line prefix to [handle errors](https://www.gnu.org/software/make/manual/make.html#Errors-in-Recipes):

```suggestion
	-rm -rf cf_gen.dSYM
```

There is also no need to "quote" cf_gen.dSYM AFAICT (the above suggestion remove quotes).

I have not tested the above. If this simpler version works, we should use it IMO.

N.B. Squid already uses `-` recipe line prefixes in contrib/Makefile.am (at least).

rousskov (changes requested):

src/Makefile.am
kinkie: The theoretical difference is that the existing line will not try to remove the directory if it's not there, where the proposed change will remove it unconditionally but then not fail if it can't (e.g. because it doesn't exist).

I'm not attached to the existing version and will accept the change to allow the PR to continue, but I doubt it's worth a blocking change request

yadij (approved):
LGTM. I have adjusted the description again to document what the current change is.

src/Makefile.am
rousskov: > The theoretical difference is that the existing line will not try to remove the directory if it's not there, where the proposed change will remove it unconditionally but then not fail if it can't (e.g. because it doesn't exist).

Not exactly: AFAICT, `rm` does check whether the file is there (because it wants to get filesystem information about the file, very similar to what `test -d` would have to do), and `rm -rf` does not fail when the directory does not exist (it exits with zero status code instead).

There are, however, other, more important (IMO) differences. The proposed version is a lot simpler and does not introduce a yet another way of deleting things in Squid Automake files.
 
> I'm not attached to the existing version and will accept the change to allow the PR to continue, but I doubt it's worth a blocking change request

Thank you for accepting this change.

rousskov (approved):

squidadm:
queued for backport to v6

PR #1697, opened 2024-02-21
negotiate_wrapper: Do not use vfork()

kinkie:
POSIX.1-2001 marks vfork(2) OBSOLETE.
POSIX.1-2008 removes the specification of vfork(2).
MacOS system headers declare vfork(2) as deprecated.
We only use vfork(2) in negotiate_wrapper, where it is not necessary.

### DIFF
--- a/src/auth/negotiate/wrapper/negotiate_wrapper.cc
+++ b/src/auth/negotiate/wrapper/negotiate_wrapper.cc
@@ -367,7 +367,7 @@ main(int argc, char *const argv[])
         exit(EXIT_FAILURE);
     }
 
-    if  (( fpid = vfork()) < 0 ) {
+    if  (( fpid = fork()) < 0 ) {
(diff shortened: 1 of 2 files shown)


rousskov (approved):
@kinkie, please check my PR title/description and required.m4 adjustments.

kinkie:
> @kinkie, please check my PR title/description and required.m4 adjustments.

vfork is only deprecated on MacOS. On other OSes (e.g. Linux) it's supported and working as expected

kinkie:
Changed the title, everything else is clear

rousskov:
> vfork is only deprecated on MacOS. 

That is enough to call that API deprecated, especially since POSIX obsoleted and then removed that API (for all OSes). Said that, I am OK with the removal of that "deprecated" detail from the title.

yadij (approved):

squidadm:
queued for backport to v6

PR #1707, opened 2024-02-26
Fix Controller.cc TheRoot assertion during shutdown

rousskov:
FATAL: assertion failed: Controller.cc:933: "TheRoot"

Store is an essential service, used by a lot of Squid code. As was
already established in 2022 commit 23b7963, this service should be
available during shutdown. That commit correctly removed explicit Store
service termination, but missed the fact that the reference-counting
TheRoot pointer (that provides access to the Store Controller singleton)
gets _automatically_ destroyed during C++ cleanup. This change removes
TheRoot reference counting, making that Controller singleton immortal.

Squid asserted when exiting with active entries in the shared memory
cache because TheRoot destruction leads to Controller destruction,
Controller destructor cleans up cache index, and that cleanup code may
result in calls to Store::Root() that dereferences destroyed TheRoot.

These assertions were seen when Squid was shutdown cleanly (e.g., using
SIGTERM) and when a kid process was exiting due to a fatal() error.

Making Store::Controller singleton immortal means that the class
destructor is never called. Fortunately, the destructor did nothing
particularly useful; Store flushing is performed by Controller::sync()
which is explicitly called during early stages of clean shutdown. The
now-unused destructor code was removed: Implementing this destructor in
a meaningful way (while avoiding accessing a being-destructed global
Store!) requires heroic efforts (which would be wasted since the
destructor is never actually called).

Also made Store Controller singleton available on the first use,
complying with Store's "essential, widely used service" status and
removing the need for an explicit Store::Init(void) call.

Also removed unit tests that required support for dynamic replacement of
Store Controller singleton. The minuscule value of the removed tests did
not justify the costs of supporting a replaceable Store Controller.

Also removed Store::FreeMemory(). Its implementation contradicted Store
status as an essential service. The function was only used by unit
tests, and its removal addresses the corresponding commit 23b7963 TODO.


yadij (approved):
"Good Enough". Though IMO ideally we should have a `Runner` to handle cleaning up  Store during shutdown and on reconfigure, as much to ensure that all still-relevant objects have a current copy swapped out to disk as to free memory.

Also, have you verified that this does not cause any `MemPool` shutdown to assert when the `Controller` and/or `Store`'s it manages are holding allocations?

yadij:
> Also removed unit tests that required support for dynamic replacement of
Store Controller singleton. The minuscule value of the removed tests did
not justify the costs of supporting a replaceable Store Controller.

IIRC, this API feature was intended to support not having any cache related objects instantiated when Squid was configured with all disk + memory caches disabled. Just noting the details, I do not insist on retaining such support.

rousskov:
Thank you for approving this PR.

> Though IMO ideally we should have a `Runner` to handle cleaning up Store during shutdown and on reconfigure, as much to ensure that all still-relevant objects have a current copy swapped out to disk as to free memory.

The swapout you are thinking about is not (and should not be) done by Controller destructor removed in this PR. It is (and/or should be) done by Controller::sync() and friends, as detailed in the PR description. We should use a RegisteredRunner API to initiate Store flushing, but that future improvement is not related to this PR.

> Also, have you verified that this does not cause any `MemPool` shutdown to assert when the `Controller` and/or `Store`'s it manages are holding allocations?

If MemPool asserts on shutdown because some allocated objects were not freed, then it is broken for the same fundamental reasons: Like Store, memory pools are an essential basic service that must be immortal (i.e. it must not be shut down, and it cannot know when it is no longer in use). I could not find the assert mentioned in the above question, and I hope it does not exist, but please point me to it if it does (keeping the note below in mind)!

N.B. AFAICT: memClean() that main.cc calls during shutdown does not assert. MemPools object is immortal and never destroys individual pools stored in MemPools::pools -- their destructors are not called during shutdown. memCleanModule() is declared but not defined and should be removed together with memInitModule().

> IIRC, [dynamic replacement of Store Controller] API feature was intended to support not having any cache related objects instantiated when Squid was configured with all disk + memory caches disabled. Just noting the details, I do not insist on retaining such support.

It is not clear to me whether it is a good idea to introduce a Controller-derived class dedicated to a cacheless configuration, but if we do decide that it is a good idea, we can still implement it -- this PR does not make such implementation significantly more difficult. The switch to a cacheless Controller should not be done externally to the Controller code anyway, so removing the public Controller replacement API in this PR is a step forward regardless of that (uncertain) future.

kinkie (approved):

PR #1709, opened 2024-02-27
Maintenance: update --with-mit-krb5 detection

yadij:


acinclude/krb5.m4
```diff
@@ -8,6 +8,51 @@
...
+  ])
+  CPPFLAGS="$LIBMIT_KRB5_CFLAGS $CPPFLAGS"
+  LIBS="$LIBMIT_KRB5_LIBS $LIBS"
+  missing_required=no
+  AC_MSG_NOTICE([Try to find Kerberos libraries in given path])
+  AC_CHECK_LIB(krb5, [main], [LIBMIT_KRB5_LIBS="$LIBMIT_KRB5_LIBS -lkrb5"],[
+    AC_MSG_WARN([library 'krb5' is required for Solaris Kerberos])
+    missing_required=yes
+  ])
+  AC_CHECK_LIB(gss, [main], [LIBMIT_KRB5_LIBS="$LIBMIT_KRB5_LIBS -lgss"],[
```
rousskov, line 33: Can we reuse the GSS check that is done elsewhere? The fact that two checks are different is a red flag. If they have to be different, it would be good to add a source code comment to explain why.

acinclude/krb5.m4
```diff
@@ -8,6 +8,51 @@
...
+    AC_DEFINE(USE_SOLARIS_KRB5,1,[Solaris Kerberos support is available])
+    SQUID_CHECK_KRB5_SOLARIS_BROKEN_KRB5_H
+    AS_IF([test "x$squid_cv_broken_krb5_h" = "xyes"],[
+      AC_DEFINE(HAVE_BROKEN_SOLARIS_KRB5_H, 1, [Define to 1 if Solaris krb5.h is broken for C++])
+      AC_MSG_WARN([You have a broken Solaris <krb5.h> system include.])
+      AC_MSG_WARN([Please see http://bugs.opensolaris.org/bugdatabase/view_bug.do?bug_id=6837512])
+      AC_MSG_WARN([If you need Kerberos support you will have to patch])
+      AC_MSG_WARN([your system. See contrib/solaris/solaris-krb5-include.patch])
+    ])
+    AC_CHECK_HEADERS(gssapi.h gssapi/gssapi.h gssapi/gssapi_krb5.h)
```
rousskov, line 48: Similar concern here: We already check gssapi.h elsewhere, under different conditions.

yadij: Not to worry, autoconf uses result caching to reduce work when the cases are the same. And when they are different we do need to re-check with the different `$PATH` anyway.

acinclude/krb5.m4
yadij: FFWIW I suspect we can probably do without this bit and adjust the `--with-gss` logic to allow detecting it on Solaris. 

However, I/we lack a Solaris machine to test out logic changes like that properly. So for now I am leaving this as-is.

rousskov: > we lack a Solaris machine to test out logic changes like that properly. 

... and yet this PR is changing Solaris-specific code. In this context, I see no substantial difference between the untested changes already done and the untested changes my request implies. IMO, in this case, it is best to stop propagating problematic Solaris-specific code that we cannot test then to modify it without testing.

acinclude/krb5.m4
rousskov: I am not worried about autoconf ability to set the value of a HAVE_FOO macro. I am worried about the code that may be assuming that a true value implies that certain preconditions (specific to that code) were met. With two independent checks, there are two sets of preconditions, so that assumption (if it exists in code) will be wrong.

On the other hand, if there are no such assumptions in our code, then we should be testing the corresponding headers unconditionally, like we already do with many other headers.

Either way, this code is problematic.

rousskov (commented):
The responses to my requests did not address my concerns, but I am still not insisting on any changes here. I do not know enough about this code and am not going to fight for its quality.

kinkie:
I move that we remove any configure magic targeting Solaris.

Let's face it, the likelihood of Squid being deployed on Solaris is next to zero.
We haven't been able to do any tests on Solaris ever since Oracle decided they didn't want the FOSS community to work on it. Oracle can port it themselves, if they want, or they can contact us with donations of licenses or even better cloud compute time to tinker with it. Until that happens, we have our hands full with other platforms that are much more likely to be useful

rousskov:
> I move that we remove any configure magic targeting Solaris.

If admins can workaround the lack of removed magic with ./configure parameters (and such), then you have my full support to remove the corresponding tricks (in a dedicated PR). The same principle goes for any other exotic environment or very old versions of popular environments: Our ./configure should have no unnecessary environment-specific workarounds, especially when those workarounds only apply to rarely seen cases. Many of such workarounds come with a significant cost to Squid code and Squid developers. They are misplaced and should be moved to documentation and support channels.
 
> the likelihood of Squid being deployed on Solaris is next to zero.

Currently, that (largely irrelevant IMO) probability is close to one because we still have relatively fresh bug reports filed by folks running Squid on Solaris:

* 2023-10-27: Bug [5314](https://bugs.squid-cache.org/show_bug.cgi?id=5314): segfaults in ModDevPoll Comm::SetSelect()   
* 2023-10-27: Bug [5312](https://bugs.squid-cache.org/show_bug.cgi?id=5312): ModDevPoll aborts on startup if fd_limit is less than OPEN_MAX  
* 2022-09-30: Bug [5242](https://bugs.squid-cache.org/show_bug.cgi?id=5242): WCCPv2 seems broken in Ssquid 5.4, 5.5, 5.7   
* 2020-08-23: Bug [5073](https://bugs.squid-cache.org/show_bug.cgi?id=5073): Compile error: index was not declared in this scope   
* 2018-05-31: Bug [4857](https://bugs.squid-cache.org/show_bug.cgi?id=4857): Polish ModDevPoll.cc  
* 2018-01-25: Bug [4813](https://bugs.squid-cache.org/show_bug.cgi?id=4813): Wrong CPU statistics in cachemgr  
* 2017-03-15: Bug [4686](https://bugs.squid-cache.org/show_bug.cgi?id=4686): assertion failed: client_side_reply.cc:1641: "reply == NULL"  
* 2017-02-18: Bug [4673](https://bugs.squid-cache.org/show_bug.cgi?id=4673): assertion failed: http.cc:1576: "!Comm::MonitorsRead(serverConnection->fd)"   
* 2017-01-17: Bug [4662](https://bugs.squid-cache.org/show_bug.cgi?id=4662): build errors with LibreSSL 2.4.4  
* 2017-01-16: Bug [4661](https://bugs.squid-cache.org/show_bug.cgi?id=4661): 4.0.17 compile error 'warning: _XPG4_2 redefined' with GCC on Solaris 10  

However, the frequency of such reports is clearly low. For comparison, there were ~25 such reports in 2016 year alone.

yadij:
IMO we have enough interest to keep it around, but not enough to waste our resources spending more time on this discussion. The shuffling (non-)change in this PR at least makes the `configure.ac` logic easier to work with regardless of how the Solaris situation goes.
 We just need to accept that this is "Good Enough" (not actively breaking anything new) and let those with Solaris interest+testing ability work on further improvements.

kinkie (approved):
good enough

rousskov:
> Francesco: I move that we remove any configure magic targeting Solaris.

> Alex: If admins can workaround the lack of removed magic with ./configure parameters (and such), then you have my full support to remove all the corresponding tricks (in a dedicated PR).

> Amos: IMO we have enough interest to keep it around

I do not see enough interest to keep any Solaris configuration magic that can be replaced with custom ./configure parameters and such. I have posted the indicators of some Squid-on-Solaris interest in general (that I know about), but that interest is not sufficient to keep unnecessary magic IMO.

> Amos: but not enough to waste our resources spending more time on this discussion.

IMO, Solaris magic removal (i.e. the topic of this discussion) has the potential to save a lot of resources long-term. The result depends on us, of course, but it can result in a very positive outcome rather than waste.

> Amos: The shuffling (non-)change in this PR at least makes the `configure.ac` logic easier to work with regardless of how the Solaris situation goes. We just need to accept that this is "Good Enough" (not actively breaking anything new) and let those with Solaris interest+testing ability work on further improvements.

FWIW, this discussion has not (and does not have to in the future) block this PR.

PR #1711, opened 2024-02-28
Bug 5343: Fix GCC v14 not finding std::find()

yadij:
Reply.cc:198: error: no matching function for call to find(...)

The required STL header was missed in 2023 commit 27c3677.

### DIFF
--- a/src/ResolvedPeers.cc
+++ b/src/ResolvedPeers.cc
@@ -13,6 +13,8 @@
 #include "ResolvedPeers.h"
 #include "SquidConfig.h"
 
+#include <algorithm>
+
 ResolvedPeers::ResolvedPeers()
 {
     if (Config.forward_max_tries > 0)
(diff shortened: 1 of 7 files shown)


kinkie (approved):

rousskov:
Please do not make commits with two PR numbers at the end of the commit message title. I have already objected (without results) to using original PRs in backported commits. Using two PR numbers would make the situation even worse.

Please do not misinterpret this comment as an assertion that you were planning to do what this comment is asking to avoid.

yadij:
Please notice how attempts to modify the title have no effect on the backport commit:
https://github.com/squid-cache/squid/commit/a6ff3200ef6b343b6af410f68d9cf7fce9f98fa8

No matter how many times you ask for things to be done differently Alex, the software systems we use do not comply to your desires.

rousskov:
> No matter how many times you ask for things to be done differently Alex, the software systems we use do not comply to your desires.

Software is not the problem. My concerns can be addressed; I have even provided specific suggestions for doing so. I have already reviewed and invalidated all specific excuses for not addressing my concerns. I see no reason to repeat that discussion until there is a new excuse.

PR #1713, opened 2024-02-28
Bug 5069: Keep listening after getsockname() error

rousskov:
ERROR: Stopped accepting connections:
    error: getsockname() failed to locate local-IP on ...

In many cases, these failures are intermittent client-triggered errors
(e.g., client shut down the accepted socket); Squid will successfully
accept other connections and, hence, should keep listening for them.

### DIFF
--- a/src/comm/TcpAcceptor.cc
+++ b/src/comm/TcpAcceptor.cc
@@ -381,7 +381,10 @@ Comm::TcpAcceptor::acceptInto(Comm::ConnectionPointer &details)
     if (getsockname(sock, gai->ai_addr, &gai->ai_addrlen) != 0) {
         int xerrno = errno;
         Ip::Address::FreeAddr(gai);
-        throw TextException(ToSBuf("getsockname() failed to locate local-IP on ", details, ": ", xstrerr(xerrno)), Here());
+        debugs(50, DBG_IMPORTANT, "ERROR: Closing accepted TCP connection after failing to obtain its local IP address" <<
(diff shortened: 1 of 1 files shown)


rousskov:
Sample output from patched code (the error was forced/injected by manually corrupting the file descriptor value for this test):

```
2024/02/28 10:20:22| ERROR: Closing accepted TCP connection after failing to obtain its local IP address
    accepted connection: conn33 local=[::] remote=127.0.0.1:59464 FD 20 flags=129
    getsockname(2) error: (9) Bad file descriptor
    listening port: 3128
```

kinkie (approved):
Looks good, thanks!

yadij:
FWIW this bug looks like just another symptom bug 3565 which already has PR #857 (blocked over disagreement on scope).

rousskov:
> FWIW this bug looks like just another symptom bug 3565 which already has PR #857

The bug fixed in this PR is pretty much unrelated to bug 3565 or PR #857 AFAICT: In modern Squid code, the exception removed by this PR was always caught by the caller, without killing Squid or causing any other damage unique to the modified code path. Comm::TcpAcceptor::acceptOne() -- the only acceptInto() caller -- already handles acceptInto() exceptions and false results. This PR does not remove the last thrown acceptInto() exception and does not add the first false result.

rousskov:
Bug reported has [confirmed](https://bugs.squid-cache.org/show_bug.cgi?id=5069#c7) that this fix works for them.

squidadm:
queued for backport to v6

PR #1715, opened 2024-03-01
Maintenance: update --with-systemd detection

yadij:
Drop support for obsolete systemd v209 and older.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1259,20 +1259,13 @@ SQUID_CHECK_LIB_WORKS(ldap,[
 ])
 
 SQUID_AUTO_LIB(systemd,[systemd API for start-up notification],[LIBSYSTEMD])
-AH_TEMPLATE(USE_SYSTEMD,[systemd support is available])
 SQUID_CHECK_LIB_WORKS(systemd,[
   SQUID_STATE_SAVE(squid_systemd_state)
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -1448,20 +1448,13 @@ SQUID_CHECK_LIB_WORKS(ldap,[
...
-  ])
-  AC_CHECK_HEADERS(systemd/sd-daemon.h)
+  PKG_CHECK_MODULES(LIBSYSTEMD,[libsystemd >= 209],[
+    AC_CHECK_HEADERS(systemd/sd-daemon.h)
+  ],[:])
   SQUID_STATE_ROLLBACK(squid_systemd_state)
-
-  AS_IF([test "x$LIBSYSTEMD_LIBS" != "x"],[
-    AC_DEFINE(USE_SYSTEMD,1,[systemd support is available])
-  ])
```
kinkie, line 1464: Seems incomplete. We are relying on USE_SYSTEMD in src/client_side.cc, and this PR would not continue defining it, would it?

kinkie (approved):
LGTM

rousskov:
> Drop support for obsolete systemd v209 and older.

I do not think we should use "Maintenance:" title prefix for PRs that drop support for something (because it devalues the prefix as an indication of a "safe" functionality-preserving change), but I am not going to fight about it.

PR #1716, opened 2024-03-04
mkrelease: ensure creation of translation:

kinkie:
Ensure presence of needed packages for release automation


yadij (approved):

mkrelease.sh
```diff
@@ -31,6 +31,18 @@ if test "$rev" != "$checkrev" ; then
...
+po2html=`which po2html`
+if test -z "$po2html" ; then
+    echo "cannot find po2html"
+    exit 1
+fi
+po2txt=`which po2txt`
+if test -z "$po2txt" ; then
+    echo "cannot find po2txt"
+    exit 1
+fi
```
rousskov, line 43: This hack cannot be the right fix, even though I am sure it helps. A proper solution ought to be different, especially since Squid Makefiles do not call `po2txt` and `po2html` (they call `$(PO2TEXT)` and `$(PO2HTML)` and those variables might be influenced by things outside this script and this hack).

I am not going to block this hack addition, but I recommend figuring out why things do not fail properly without it (and then addressing that/core problem).

P.S. PR title is misleading. This hack does not ensure any creation. It only facilitates failure (in some cases).

kinkie: You are right; this is a hack is not the best fix. The right fix would be to ensure that the --enable-translations configure option aborts the build if the tools can't be found, instead of silently disabling them as it does right now.
This would in turn have further requirements and consequences, making it an effective rabbit hole.
It would be a lot of work for a system that has exactly two users and is used at most once a month; it's just not worth our time to do anything better than this.
Thanks for not blocking this PR; as you say it's not the best choice, but it helps avoid problems like it happened with the 6.7 release, which was missing the translated pages because nothing noticed that the tools needed to generate them were missing

rousskov: > this is a hack is not the best fix

IMO, the fact that this PR does not fix the problem that should be fixed should have been disclosed in the PR when it was posted. A proper disclosure would have saved me quite a bit of time!

> This would in turn have further requirements and consequences, making it an effective rabbit hole.

FWIW, to make a convincing rabbit hole argument one has to demonstrate inevitability of at least a few significant complications getting us deeper and deeper into the rabbit hole. From where I sit, this looks like a lame excuse rather than an argument advancing a valid position.

> it helps avoid problems like it happened with the 6.7 release

So would a solution.

squidadm:
queued for backport to v6

PR #1717, opened 2024-03-04
Maintenance: update --with-netfilter-conntrack detection

yadij:
Fixes pollution of global LIBS, LDFLAGS and CPPFLAGS
when the library and required headers are not
detected as usable.


kinkie (approved):
looks good

configure.ac
```diff
@@ -1455,13 +1455,15 @@ AC_MSG_NOTICE([Linux Netfilter support requested: ${enable_linux_netfilter:=auto
 dnl Look for libnetfilter_conntrack options (needed for QOS netfilter marking)
 SQUID_AUTO_LIB(netfilter-conntrack,[Netfilter conntrack],[LIBNETFILTER_CONNTRACK])
 SQUID_CHECK_LIB_WORKS(netfilter-conntrack,[
-  LDFLAGS="$LIBNETFILTER_CONNTRACK_PATH $LDFLAGS"
+  SQUID_STATE_SAVE(squid_netfilter_conntrack_state)
+  PKG_CHECK_MODULES([LIBNETFILTER_CONNTRACK],[libnetfilter_conntrack],[:],[:])
   CPPFLAGS="$LIBNETFILTER_CONNTRACK_CFLAGS $CPPFLAGS"
-  AC_SEARCH_LIBS([nfct_query],[netfilter_conntrack],[LIBNETFILTER_CONNTRACK_LIBS="-lnetfilter_conntrack"])
+  LIBS="$LIBNETFILTER_CONNTRACK_PATH $LIBNETFILTER_CONNTRACK_LIBS $LIBS"
```
kinkie, line 1461: besides consistency, what is the reason to touch LIBS here? the following tests only check for headers

configure.ac
```diff
@@ -1455,13 +1455,15 @@ AC_MSG_NOTICE([Linux Netfilter support requested: ${enable_linux_netfilter:=auto
...
   AC_CHECK_HEADERS([ \
     libnetfilter_conntrack/libnetfilter_conntrack.h \
-    libnetfilter_conntrack/libnetfilter_conntrack_tcp.h
-  ],,[LIBNETFILTER_CONNTRACK_LIBS=""])
+    libnetfilter_conntrack/libnetfilter_conntrack_tcp.h \
```
kinkie, line 1464: According to a quick experiment I've just run, the backslash continuations are not needed (at least with autoconf 2.71). While removing them might be out of scope for this PR, adding them is unnecessary

yadij: Thanks for he info. Leaving it for consistency with every other headers list.

configure.ac
yadij: consistency is the main reason, to ensure that these checks still work properly when the LIBS are set instead of LDFLAGS.

PR #1718, opened 2024-03-04
Maintenance: update --with-cap detection

yadij:


src/tools.cc
```diff
@@ -1132,8 +1131,8 @@ keepCapabilities(void)
 static void
 restoreCapabilities(bool keep)
 {
-    /* NP: keep these two if-endif separate. Non-Linux work perfectly well without Linux syscap support. */
-#if USE_LIBCAP
+    /* keep these two if-endif separate. Non-Linux work perfectly well without Linux syscap support. */
```
rousskov, line 1134: If we are polishing this comment: I could not easily tell what exactly should be kept separate. I suggest mentioning specific condition(s) instead. For example:

```suggestion
    // Check HAVE_LIBCAP separately from _SQUID_LINUX_: Some non-Linux cases work without LIBCAP.
```

or

```suggestion
    // Check HAVE_LIBCAP separately from _SQUID_LINUX_: Non-Linux cases may benefit from LIBCAP but work without it.
```

... or whatever gets the reader closer to the actual reasons behind the existing code logic.

yadij: Fixed in [1f09e63](https://github.com/squid-cache/squid/pull/1718/commits/1f09e63bfe697ac380eb8eeabbb1329080511941)

src/tools.cc
```diff
@@ -1132,8 +1131,12 @@ keepCapabilities(void)
 static void
 restoreCapabilities(bool keep)
 {
-    /* NP: keep these two if-endif separate. Non-Linux work perfectly well without Linux syscap support. */
-#if USE_LIBCAP
+#if _SQUID_LINUX_ && !HAVE_LIBCAP
+    /* Linux requires syscap support from libcap. */
+    Ip::Interceptor.StopTransparency("Missing needed capability support.");
+#endif
```
rousskov, line 1137: To avoid this Linux code falling through into what is arguably marked as "Non-Linux" code at the end of this function:

```suggestion
    return;
#endif
```


Alternatively, consider refactoring a little to arrive at this overall simpler (and focused on the primary use case) code shape:

```C++
#if HAVE_LIBCAP
    ...
    return;
#endif

#if _SQUID_LINUX_
    Ip::Interceptor.StopTransparency("Linux requires syscap support from libcap.");
    (void)keep;
#else
    /* Non-Linux transparent proxy works with or without libcap support. */
    (void)keep;
#endif
```

No "keep these two if-endif separate" comment is required with the above code shape either IMO. The code is self-documenting. The only disadvantage I can see with this simpler shape is that Ip::Interceptor.StopTransparency() call will be compiled (but never executed) on Linux with LIBCAP support. Your call.

yadij: The suggested change causes unreachable code errors on the `#else` clause at the end. Going with the alternative.

kinkie (approved):
LGTM!

PR #1720, opened 2024-03-04
Use ERR_READ_ERROR for read-from-client I/O errors

eduard-bagdasaryan:
ERR_CLIENT_GONE is still used for unexpected zero-size reads on
client-to-Squid connections. The two cases are now distinct.

### DIFF
--- a/src/servers/Server.cc
+++ b/src/servers/Server.cc
@@ -176,7 +176,7 @@ Server::doClientRead(const CommIoCbParams &io)
         LogTagsErrors lte;
         lte.timedout = rd.xerrno == ETIMEDOUT;
         lte.aborted = !lte.timedout; // intentionally true for zero rd.xerrno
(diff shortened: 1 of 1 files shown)


eduard-bagdasaryan:
The fix was [suggested](https://github.com/squid-cache/squid/pull/1582/files#r1388884278) by @rousskov .

rousskov:
> The fix was [suggested](https://github.com/squid-cache/squid/pull/1582/files#r1388884278) by @rousskov .

Yes, to me, ERR_READ_ERROR feels more appropriate for reporting network read errors than ERR_CLIENT_GONE, especially when the latter is used for a no-error zero-read case.

Once this PR lands (or is rejected), we will follow up with the code deduplication PR suggested in the comment referenced above.

rousskov (approved):

kinkie (approved):
Agreed

yadij (approved):

PR #1721, opened 2024-03-05
Remove unused strnstr() replacement

yadij:
Squid code does not actually use strnstr() so this copy of FreeBSD
code does not need to be shipped.


kinkie (approved):
This is great 👍

rousskov (approved):
Thank you.

PR #1723, opened 2024-03-06
De-duplicate I/O error-detailing code

eduard-bagdasaryan:
This improvement eliminates code duplication introduced
in recent commit b850f8b.

### DIFF
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -240,11 +240,7 @@ FwdState::updateAleWithFinalError()
     if (!err || !al)
         return;
 
-    LogTagsErrors lte;
-    if (err->xerrno == ETIMEDOUT || err->type == ERR_READ_TIMEOUT)
-        lte.timedout = true;
-    else if (err->type != ERR_NONE)
-        lte.aborted = true;
+    const auto lte = LogTagsErrors::FromErrno(err->type == ERR_READ_TIMEOUT ? ETIMEDOUT : err->xerrno);
     al->cache.code.err.update(lte);
     if (!err->detail) {
         static const auto d = MakeNamedErrorDetail("WITH_SERVER");
--- a/src/LogTags.cc
+++ b/src/LogTags.cc
@@ -18,6 +18,15 @@ LogTagsErrors::update(const LogTagsErrors &other)
     aborted = aborted || other.aborted;
 }
 
+LogTagsErrors
+LogTagsErrors::FromErrno(const int errNo)
+{
+    LogTagsErrors lte;
+    lte.timedout = errNo == ETIMEDOUT;
+    lte.aborted = !lte.timedout; // intentionally true for zero errNo
+    return lte;
+}
+
 /* LogTags */
 
 // old deprecated tag strings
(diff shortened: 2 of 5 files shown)


eduard-bagdasaryan:
This PR addresses a PR1582 [suggestion](This improvement eliminates code duplication introduced
by https://github.com/measurement-factory/squid/commit/b850f8b2161635fba683e9f55e8a2f068937174a fix).

src/FwdState.cc
```diff
@@ -240,11 +240,7 @@ FwdState::updateAleWithFinalError()
     if (!err || !al)
         return;
 
-    LogTagsErrors lte;
-    if (err->xerrno == ETIMEDOUT || err->type == ERR_READ_TIMEOUT)
-        lte.timedout = true;
-    else if (err->type != ERR_NONE)
-        lte.aborted = true;
+    const auto lte = LogTagsErrors::FromErrno(err->type == ERR_READ_TIMEOUT ? ETIMEDOUT : err->xerrno);
```
rousskov, line 243: Eduard and I tried to figure out why this code treated ERR_NONE `err->type` specially (commit 21f90159029aeb05fca55a6f48f120afb5f2b705), but failed. I speculate that this code was negatively influenced by Ftp::Client::failed() or similar ugly methods that try hard to trick the reader into thinking that the method creates ErrorState objects with ERR_NONE `type`.

We believe it is best to move on to simpler and less duplicated code here until we find specific reasons to make exceptions (which we will have to explain when re-adding).

This comment does not request any code changes. I am just explaining why PR code ignores the special ERR_NONE condition in this refactoring.

rousskov (dismissed):
Thank you for this cleanup!

src/FwdState.cc
yadij: I think it was there because client disconnect after sending the whole reply is not an abort/error.
Though I am also unable to easily find any path which leads there without setting the `err->type`.

```
FwdState::completed()
{
...
    if (entry->store_status == STORE_PENDING) {
        if (entry->isEmpty()) {
...
        } else {
            updateAleWithFinalError(); // if any
            if (storedWholeReply_)
                entry->completeSuccessfully(storedWholeReply_);
...
```

yadij (changes requested):
Just one nit. LGTM otherwise.

src/LogTags.cc
```diff
@@ -18,6 +18,15 @@ LogTagsErrors::update(const LogTagsErrors &other)
     aborted = aborted || other.aborted;
 }
 
+LogTagsErrors
+LogTagsErrors::FromErrno(const int errNo)
+{
+    LogTagsErrors lte;
+    lte.timedout = errNo == ETIMEDOUT;
```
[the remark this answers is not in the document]
rousskov: Changed in [ffb3251](https://github.com/squid-cache/squid/pull/1723/commits/ffb32515eefd472a26eb554120d937d3d505201c).

@yadij, if you are going to block PRs based on this style preference, please work on updating [official style guidelines](https://wiki.squid-cache.org/DeveloperResources/SquidCodingGuidelines) to reflect that requirement and investigate whether astyle can be configured to apply it.

kinkie (approved):
The design of LogTagErrors feels weird, but that's for another PR to deal with, if at all

yadij (approved):

rousskov (approved):

PR #1725, opened 2024-03-07
Improve bounds checking in rfc1035NameUnpack

kinkie:
Code audit by Peter J Philipp found that under certain conditions
it is possible to craft a DNS message which causes Squid to read
one extra byte.
Add an extra check to prevent this from happening

### DIFF
--- a/src/dns/rfc1035.cc
+++ b/src/dns/rfc1035.cc
@@ -265,6 +265,11 @@ rfc1035NameUnpack(const char *buf, size_t sz, unsigned int *off, unsigned short 
                 RFC1035_UNPACK_DEBUG;
                 return 1;
             }
+            /* before copying sizeof(unnsigned short), ensure we respect bounds */
+            if ((*off) + sizeof(s) > sz) {
+                RFC1035_UNPACK_DEBUG;
+                return 1;
+            }
             memcpy(&s, buf + (*off), sizeof(s));
             s = ntohs(s);
             (*off) += sizeof(s);


src/dns/rfc1035.cc
```diff
@@ -265,6 +265,11 @@ rfc1035NameUnpack(const char *buf, size_t sz, unsigned int *off, unsigned short
...
                 return 1;
             }
+            /* before copying compression offset value, ensure it is inside the buffer */
+            if ((*off) + sizeof(s) > sz) {
+                RFC1035_UNPACK_DEBUG;
+                return 1;
+            }
             memcpy(&s, buf + (*off), sizeof(s));
             s = ntohs(s);
             (*off) += sizeof(s);
```
rousskov, line 275: AFAICT, the `if` statement misleadingly described as a "sanity check" below this line (I cannot attach a change request there) should now be removed or converted into a (paranoid) assertion:

```C++
if ((*off) + sizeof(s) > sz) return; // OK: new check
(*off) += sizeof(s);
if ((*off) > sz) return; // XXX: old check that we are now doing earlier
```

I suggest removing that old check (i.e. not adding a paranoid assertion).

rousskov (changes requested):
Thank you for posting this fix. If Peter wants to be credited beyond the commit message, please do not forget to adjust PR metadata accordingly.

I polished PR description (including to use name punctuation from emails) and pushed a minor comment adjustment. Please adjust further as needed.

AFAICT, we are effectively moving an existing check here. My change request is meant to finalize that move. I can remove that code myself if you prefer, but I thought it might be easier to discuss it before that removal...

src/dns/rfc1035.cc
kinkie: I agree it is effectively a duplicate check, and can be removed

rousskov:
> If Peter wants to be credited beyond the commit message, please do not forget to adjust PR metadata accordingly.

@kinkie, just to avoid a misunderstanding: I have approved this PR, but I do not intend to clear it for merging due to the above concern. The ball is on your side.

PR #1728, opened 2024-03-09
ESI: Disable by default

yadij:


rousskov (changes requested):
Thank you for implementing this long overdue change!

kinkie (dismissed):
❤️

configure.ac
```diff
@@ -820,19 +820,22 @@ AC_ARG_ENABLE(delay-pools,
...
+                 [Enable ESI for accelerators. ESI requires expat or xml2 library.
                   Enabling ESI will cause squid reverse proxies to be capable
-                  of the Edge Acceleration Specification (www.esi.org).]),
-                 [squid_opt_use_esi=$enableval],[])
+                  of the Edge Acceleration Specification (www.esi.org).])
+],[
+  SQUID_YESNO($enableval,[--enable-esi])
+])
+AC_MSG_NOTICE([Enable ESI processor: ${enable_esi:=no (auto)}])
+AC_DEFINE_UNQUOTED(USE_SQUID_ESI,(HAVE_LIBEXPAT||HAVE_LIBXML2))
```
rousskov, line 833:
```C++
/* Define to 1 if you have the expat library */
/* #undef HAVE_LIBEXPAT */

#define USE_SQUID_ESI (HAVE_LIBEXPAT||HAVE_LIBXML2)
```

We should not define USE_SQUID_ESI in terms of other macros, especially often-undefined macros. I am not going to block this PR because it has a huge value of disabling ESI by default, but please change USE_SQUID_ESI value to be based on enable_esi alone.

rousskov (commented):
> [yadij](https://github.com/yadij) dismissed [rousskov](https://github.com/rousskov)’s [stale review](https://github.com/squid-cache/squid/pull/1728#pullrequestreview-1926575682) March 11, 2024 06:13

Please never dismiss my negative reviews (regardless of whether you think you have addressed all of my concerns).

configure.ac
yadij: No. That suggestion would cause compiler errors with `--enables-esi --without-expat --without-xml2` or equivalent missing-library build environments.
This simple definition replaces the complex and delicate nested set of `AS_IF` clauses that were at line 881. In a way which cannot ever produce a wrong (mis)match.

FTR; supporting non-defined conditionals such as you seem worried about are why we prohibit `#ifndef` for testing these wrappers. See all the existing .cc file checks like `#if USE_SQUID_ESI && HAVE_EXPAT` - when either or both of the macros may be undefined.

... and yes I have thoroughly tested this.

rousskov: > No. That suggestion would cause compiler errors with `--enables-esi --without-expat --without-xml2` or equivalent missing-library build environments.

The combination of ./configure options you are probably thinking about will not cause compiler errors AFAICT; it will cause ./configure "ESI processor requires libxml2 or libexpat" errors, and that is the expected/desired outcome:

```sh
AS_IF([test "x$enable_esi" = "xyes" -a "x$LIBXML2_LIBS" = "x" -a "x$LIBEXPAT_LIBS" = "x"],[
  AC_MSG_ERROR([ESI processor requires libxml2 or libexpat])
])
```

> This simple definition replaces the complex and delicate nested set of `AS_IF` clauses that were at line 881.

The `enable_esi`-based definition I am rooting for ought to be simple as well.

> In a way which cannot ever produce a wrong (mis)match.

That assertion is false: The PR version will fail if some other code starts using one of the two libraries. As we discussed elsewhere recently, a particular library use is a separate concern from ESI use, even though current ESI code does require one of the two libraries and even though no other current code uses those libraries.

yadij: Done in [c7cd396](https://github.com/squid-cache/squid/pull/1728/commits/c7cd39690b35be19ee7d86625fcb74bd633cfcc0)

rousskov (commented):
Thank you.

kinkie (approved):

PR #1729, opened 2024-03-09
Implement iterator for wordlist

kinkie:
Also removed ToSBufList(): Unused since inception (2013 commit 6a17a36).


kinkie:
First use case is a refactoring of ErrorState::Dump which will be forthcoming (it's already implemented)

kinkie:
I guess the main debate point is whether we should use the stream fill character as separator, or wether we should hardcode space or something else.

yadij:
Not objecting, just wondering if we are better off converting `wordlist` to another type when we run into them.

kinkie:
> Not objecting, just wondering if we are better off converting `wordlist` to another type when we run into them.

My attempts in the past have resulted in being rabbit holes. My plan is that by implementing quick-win higher level abstractions, it will make the replacement easier

src/wordlist.h
```diff
@@ -65,5 +67,7 @@ char *wordlistChopHead(wordlist **);
 /// convert a wordlist to a SBufList
 SBufList ToSBufList(wordlist *);
 
+std::ostream& operator<<(std::ostream &os, const wordlist wl);
```
rousskov, line 70: If possible, please specialize AsList<wordlist> instead. AFAICT, doing so will fix at least three problems with the current PR code, offer more formatting flexibility to callers, and reduce future code changes related to wordlist elimination from Squid.

When specializing, please try to avoid duplicating the existing manipulator.container [loop](https://github.com/squid-cache/squid/blob/6410d312077b71a1f5f619390c2d467b1fe4dc74/src/base/IoManip.h#L203-L213). Instead, try adding worldlist::begin(), worldlist::end() (at least).

rousskov (changes requested):

rousskov:
> > just wondering if we are better off converting `wordlist` to another type when we run into them.

> implementing quick-win higher level abstractions, it will make the replacement easier

Yes, wordlist should be converted, and, yes, individual conversions are often difficult. This PR (if done correctly) may help during transition to better container types. I hope that this PR can move wordlist closer to standard container APIs (that can be reused during future conversions) instead of adding special (i.e. wordlist-only) "higher level abstractions" (that would then need to be converted again). I left the corresponding [change request](https://github.com/squid-cache/squid/pull/1729#discussion_r1518881823).

src/wordlist.h
kinkie: Done.

Build-tested with this (why, oh why doesn't github support stacked diffs!)

```
diff --git a/src/errorpage.cc b/src/errorpage.cc
index a56998dd72..5f35a5e44a 100644
--- a/src/errorpage.cc
+++ b/src/errorpage.cc
@@ -10,6 +10,8 @@
 
 #include "squid.h"
 #include "AccessLogEntry.h"
+#include "base/IoManip.h"
+#include "base/PackableStream.h"
 #include "cache_cf.h"
 #include "clients/forward.h"
 #include "comm/Connection.h"
@@ -864,7 +866,9 @@ ErrorState::Dump(MemBuf * mb)
         str.appendf("FTP Request: %s\r\n", ftp.request);
         str.appendf("FTP Reply: %s\r\n", (ftp.reply? ftp.reply:"[none]"));
         str.append("FTP Msg: ", 9);
-        wordlistCat(ftp.server_msg, &str);
+        PackableStream s(str);
+        s << AsList(*(ftp.server_msg));
+        // wordlistCat(ftp.server_msg, &str);
         str.append("\r\n", 2);
     }
 ```

File: src/wordlist.h [hunk 2]
```diff
@@ -12,6 +12,28 @@
 #include "globals.h"
 #include "sbuf/List.h"
 
+#include <iterator>
+
+class wordlist;
+
+class WordlistIterator
+{
+    const wordlist *w;
```
yadij, line 21: Class layout style require this to be in an explicit `private:` section at the end of the class.

src/wordlist.h
```diff
@@ -20,6 +42,7 @@ class wordlist
 {
     MEMPROXY_CLASS(wordlist);
     friend char *wordlistChopHead(wordlist **);
+    using iterator = WordlistIterator;
```
yadij, line 45: It this needed? it appears unused. Or perhapse the `begin()` and `end()` methods should return `iterator` instead.

yadij (changes requested):

src/wordlist.h
kinkie: added

yadij (changes requested):

kinkie:
I’ll recheck I pushed. Thanks and sorry

@mobile

src/wordlist.h
```diff
@@ -62,8 +91,18 @@ void wordlistDestroy(wordlist **);
  */
 char *wordlistChopHead(wordlist **);
 
-/// convert a wordlist to a SBufList
-SBufList ToSBufList(wordlist *);
```
rousskov, line 66: This removal is outside this PR scope, and I do not insist on it. I removed ToSBufList() because, if it stays, then its implementation should be modified to use the new iterator. Reverting the last branch commit (25e74e7) will re-introduce that modification.

This comment does not request any changes.

src/wordlist.cc
```diff
@@ -34,21 +34,8 @@ wordlistAdd(wordlist ** list, const char *key)
...
-ToSBufList(wordlist *wl)
-{
-    SBufList rv;
-    while (wl != nullptr) {
-        rv.push_back(SBuf(wl->key));
-        wl = wl->next;
-    }
-    return rv;
+    for (const auto &word: *w)
+        mb->appendf("%s\n", word);
```
rousskov, line 38: A few other places can be converted to use the new iterators, but I do not think they should be converted (in this PR). IMO,. it makes sense to convert this internal worldlist code from dogfooding point of view, but other code is not trivial to convert and/or should be refactored to avoid worldlist in the first place.

This comment does not request any changes.

rousskov (approved):
I have committed a few minor fixes and polishing touches. I have adjusted PR description accordingly. @kinkie, please adjust further as needed and merge when you are happy with the code.

src/wordlist.h
kinkie: I think wordlist has a functional place, and SBufList should take that place.
I agree the implementation of the converter should be moved to the iterator, so I'm okay with the removal,
reimplementation can happen when needed.

kinkie:
@yadij are you satisfied with the current shape of the code?

yadij (approved):

src/wordlist.h
```diff
@@ -29,7 +28,7 @@ class WordlistIterator
     auto operator ==(const WordlistIterator &rhs) const { return this->w == rhs.w; }
     auto operator !=(const WordlistIterator &rhs) const { return this->w != rhs.w; }
 
-    inline reference operator *() const;
+    inline value_type operator *() const;
```
rousskov, line 30: FTR: We were forced to return value_type here (and, hence, remove std::input_iterator_tag iterator_category) because of unexplained build [failure](https://build.squid-cache.org/job/5-pr-test/7163/COMPILER=gcc,OS=centos-stream-8,label=pr-test-builder/console) with GCC on centos-stream-8:

```
20:02:14 ../../../../src/wordlist.h: In member function 'const char* const& WordlistIterator::operator*() const':
20:02:14 ../../../../src/wordlist.h:104:15: error: returning reference to temporary [-Werror=return-local-addr]
20:02:14      return w->key;
```

rousskov (approved):

PR #1733, opened 2024-03-12
Fix ./configure after strnstr() replacement removal

rousskov:
configure: line 51104: SQUID_CHECK_FUNC_STRNSTR: command not found

The macro call should have been removed in recent commit 9d3433c.


kinkie (approved):
thanks for catching this

PR #1730, opened 2024-03-09
Refactor and improve ErrorState::Dump

kinkie:
Rework the internals for generating output in ErrorState::Dump,
used for expanding the '%W' token in error page templates.

Also fix a bug with excessive html-quoting of the output.

PR #1735, opened 2024-03-13
Use PackableStream for dump_peers

kinkie:
Also improve formatting slightly

### DIFF
--- a/src/neighbors.cc
+++ b/src/neighbors.cc
@@ -1503,67 +1503,65 @@ static void
 dump_peers(StoreEntry *sentry, CachePeers *peers)
 {
     char ntoabuf[MAX_IPSTRLEN];
+    PackableStream os(*sentry);
     int i;
 
     if (!peers) {
-        storeAppendPrintf(sentry, "There are no neighbors installed.\n");
+        os << "There are no neighbors installed.\n";
         return;
     }
 
     for (const auto &peer: *peers) {
         const auto e = peer.get();
         assert(e->host != nullptr);
-        storeAppendPrintf(sentry, "\n%-11.11s: %s\n",
-                          neighborTypeStr(e),
-                          e->name);
-        storeAppendPrintf(sentry, "Host       : %s/%d/%d\n",
-                          e->host,
-                          e->http_port,
-                          e->icp.port);
-        storeAppendPrintf(sentry, "Flags      :");
+        os << "\n" << std::setw(11) << std::left <<
+            neighborTypeStr(e) << ": " << e->name << "\n";
+        os << "Host       : " << e->host << '/' << e->http_port << '/' <<
+            e->icp.port << "\n";
+        os << "Flags      :";
+        os.flush();
         dump_peer_options(sentry, e);
 
-        for (i = 0; i < e->n_addresses; ++i) {
-            storeAppendPrintf(sentry, "Address[%d] : %s\n", i,
-                              e->addresses[i].toStr(ntoabuf,MAX_IPSTRLEN) );
-        }
+        for (i = 0; i < e->n_addresses; ++i)
+            os << "Address[" << i << "] : " << e->addresses[i].toStr(ntoabuf, MAX_IPSTRLEN) << "\n";
 
-        storeAppendPrintf(sentry, "Status     : %s\n",
-                          neighborUp(e) ? "Up" : "Down");
-        storeAppendPrintf(sentry, "FETCHES    : %d\n", e->stats.fetches);
-        storeAppendPrintf(sentry, "OPEN CONNS : %d\n", e->stats.conn_open);
-        storeAppendPrintf(sentry, "AVG RTT    : %d msec\n", e->stats.rtt);
+        os << "Status     : " << (neighborUp(e) ? "Up" : "Down") << "\n" <<
+            "FETCHES    : " << std::setw(8) << std::right << e->stats.fetches << "\n" <<
+            "OPEN CONNS : " << std::setw(8) << std::right << e->stats.conn_open << "\n" <<
+            "AVG RTT    : " << std::setw(8) << std::right << e->stats.rtt << " msec\n";
 
         if (!e->options.no_query) {
-            storeAppendPrintf(sentry, "LAST QUERY : %8d seconds ago\n",
-                              (int) (squid_curtime - e->stats.last_query));
+            if (e->stats.last_query > 0)
+                os << "LAST QUERY : " << std::setw(8) << std::right <<
+                    (squid_curtime - e->stats.last_query) << " seconds ago\n";
+            else
+                os << "LAST QUERY : none sent\n";
 
             if (e->stats.last_reply > 0)
-                storeAppendPrintf(sentry, "LAST REPLY : %8d seconds ago\n",
-                                  (int) (squid_curtime - e->stats.last_reply));
+                os << "LAST REPLY : " << std::setw(8) << std::right <<
+                    (squid_curtime - e->stats.last_reply) << " seconds ago\n";
             else
-                storeAppendPrintf(sentry, "LAST REPLY : none received\n");
+                os << "LAST REPLY : none received\n";
 
-            storeAppendPrintf(sentry, "PINGS SENT : %8d\n", e->stats.pings_sent);
-
-            storeAppendPrintf(sentry, "PINGS ACKED: %8d %3d%%\n",
-                              e->stats.pings_acked,
-                              Math::intPercent(e->stats.pings_acked, e->stats.pings_sent));
+            os << "PINGS SENT : " << std::setw(8) << std::right << e->stats.pings_sent << "\n" <<
+                "PINGS ACKED: " << std::setw(8) << std::right << e->stats.pings_acked << " " <<
+                Math::intPercent(e->stats.pings_acked, e->stats.pings_sent) << "\n";
         }
 
-        storeAppendPrintf(sentry, "IGNORED    : %8d %3d%%\n", e->stats.ignored_replies, Math::intPercent(e->stats.ignored_replies, e->stats.pings_acked));
+        os << "IGNORED    : " << std::setw(8) << std::right << e->stats.ignored_replies << " " <<
+            Math::intPercent(e->stats.ignored_replies, e->stats.pings_acked) << "\n";
 
         if (!e->options.no_query) {
-            storeAppendPrintf(sentry, "Histogram of PINGS ACKED:\n");
+            os << "Histogram of PINGS ACKED:\n";
 #if USE_HTCP
 
             if (e->options.htcp) {
-                storeAppendPrintf(sentry, "\tMisses\t%8d %3d%%\n",
-                                  e->htcp.counts[0],
-                                  Math::intPercent(e->htcp.counts[0], e->stats.pings_acked));
-                storeAppendPrintf(sentry, "\tHits\t%8d %3d%%\n",
-                                  e->htcp.counts[1],
-                                  Math::intPercent(e->htcp.counts[1], e->stats.pings_acked));
+                os << "\tMisses\t" << std::setw(8) << std::right << e->htcp.counts[0] << " " <<
+                    std::setw(3) << std::right <<
+                    Math::intPercent(e->htcp.counts[0], e->stats.pings_acked) << "\n" <<
+                    "\tHits\t" << std::setw(8) << std::right << e->htcp.counts[1] << " " <<
+                    std::setw(3) << std::right <<
+                    Math::intPercent(e->htcp.counts[1], e->stats.pings_acked) << "\n";
             } else {
 #endif
 
@@ -1571,10 +1569,10 @@ dump_peers(StoreEntry *sentry, CachePeers *peers)
                     if (e->icp.counts[op] == 0)
                         continue;
 
-                    storeAppendPrintf(sentry, "    %12.12s : %8d %3d%%\n",
-                                      icp_opcode_str[op],
-                                      e->icp.counts[op],
-                                      Math::intPercent(e->icp.counts[op], e->stats.pings_acked));
+                    os << "    " << std::setw(12) << std::setprecision(12) <<
+                        std::right << icp_opcode_str[op] << " : " << std::setw(8) << std::right <<
+                        e->icp.counts[op] << " " << std::setw(3) << std::right <<
+                        Math::intPercent(e->icp.counts[op], e->stats.pings_acked) << "\n";
                 }
 
 #if USE_HTCP
@@ -1586,11 +1584,11 @@ dump_peers(StoreEntry *sentry, CachePeers *peers)
         }
 
         if (e->stats.last_connect_failure) {
-            storeAppendPrintf(sentry, "Last failed connect() at: %s\n",
-                              Time::FormatHttpd(e->stats.last_connect_failure));
+            os << "Last failed connect() at: " <<
+                Time::FormatHttpd(e->stats.last_connect_failure) << "\n";
         }
 
-        storeAppendPrintf(sentry, "keep-alive ratio: %d%%\n", Math::intPercent(e->stats.n_keepalives_recv, e->stats.n_keepalives_sent));
+        os << "keep-alive ratio: " << Math::intPercent(e->stats.n_keepalives_recv, e->stats.n_keepalives_sent) << "\n";
     }
 }
 
--- a/src/neighbors.cc
+++ b/src/neighbors.cc
@@ -1545,11 +1545,13 @@ dump_peers(StoreEntry *sentry, CachePeers *peers)
 
             os << "PINGS SENT : " << std::setw(8) << std::right << e->stats.pings_sent << "\n" <<
                 "PINGS ACKED: " << std::setw(8) << std::right << e->stats.pings_acked << " " <<
-                Math::intPercent(e->stats.pings_acked, e->stats.pings_sent) << "\n";
+                std::setw(3) << std::right <<
+                Math::intPercent(e->stats.pings_acked, e->stats.pings_sent) << "%\n";
         }
 
         os << "IGNORED    : " << std::setw(8) << std::right << e->stats.ignored_replies << " " <<
-            Math::intPercent(e->stats.ignored_replies, e->stats.pings_acked) << "\n";
+            std::setw(3) << std::right <<
+            Math::intPercent(e->stats.ignored_replies, e->stats.pings_acked) << "%\n";
 
         if (!e->options.no_query) {
             os << "Histogram of PINGS ACKED:\n";
@@ -1558,10 +1560,10 @@ dump_peers(StoreEntry *sentry, CachePeers *peers)
             if (e->options.htcp) {
                 os << "\tMisses\t" << std::setw(8) << std::right << e->htcp.counts[0] << " " <<
                     std::setw(3) << std::right <<
-                    Math::intPercent(e->htcp.counts[0], e->stats.pings_acked) << "\n" <<
+                    Math::intPercent(e->htcp.counts[0], e->stats.pings_acked) << "%\n" <<
                     "\tHits\t" << std::setw(8) << std::right << e->htcp.counts[1] << " " <<
                     std::setw(3) << std::right <<
-                    Math::intPercent(e->htcp.counts[1], e->stats.pings_acked) << "\n";
+                    Math::intPercent(e->htcp.counts[1], e->stats.pings_acked) << "%\n";
             } else {
 #endif
 
@@ -1572,7 +1574,7 @@ dump_peers(StoreEntry *sentry, CachePeers *peers)
                     os << "    " << std::setw(12) << std::setprecision(12) <<
                         std::right << icp_opcode_str[op] << " : " << std::setw(8) << std::right <<
                         e->icp.counts[op] << " " << std::setw(3) << std::right <<
-                        Math::intPercent(e->icp.counts[op], e->stats.pings_acked) << "\n";
+                        Math::intPercent(e->icp.counts[op], e->stats.pings_acked) << "%\n";
                 }
 
 #if USE_HTCP
@@ -1588,7 +1590,7 @@ dump_peers(StoreEntry *sentry, CachePeers *peers)
                 Time::FormatHttpd(e->stats.last_connect_failure) << "\n";
         }
 
-        os << "keep-alive ratio: " << Math::intPercent(e->stats.n_keepalives_recv, e->stats.n_keepalives_sent) << "\n";
+        os << "keep-alive ratio: " << Math::intPercent(e->stats.n_keepalives_recv, e->stats.n_keepalives_sent) << "%\n";
     }
 }
 


kinkie:
Runtime test:

master:
```
Sibling    : localhost
Host       : localhost/13128/0
Flags      : default connection-auth=auto tls-disable
Address[0] : 127.0.0.1
Status     : Down
FETCHES    : 0
OPEN CONNS : 0
AVG RTT    : 0 msec
LAST QUERY : 1710344842 seconds ago
LAST REPLY : none received
PINGS SENT :        0
PINGS ACKED:        0   0%
IGNORED    :        0   0%
Histogram of PINGS ACKED:
Last failed connect() at: 13/Mar/2024:15:47:16 +0000
keep-alive ratio: 0%
```

PR:
```
Sibling    : localhost
Host       : localhost/13128/0
Flags      : default connection-auth=auto tls-disable
Address[0] : 127.0.0.1
Status     : Down
FETCHES    :        0
OPEN CONNS :        0
AVG RTT    :        0 msec
LAST QUERY : none sent
LAST REPLY : none received
PINGS SENT :        0
PINGS ACKED:        0   0%
IGNORED    :        0   0%
Histogram of PINGS ACKED:
Last failed connect() at: 13/Mar/2024:15:54:31 +0000
keep-alive ratio: 0%
```

Delta:
```diff
--- /tmp/before	2024-03-13 22:54:54
+++ /tmp/after	2024-03-13 22:54:53
@@ -3,14 +3,14 @@
 Flags      : default connection-auth=auto tls-disable
 Address[0] : 127.0.0.1
 Status     : Down
-FETCHES    : 0
-OPEN CONNS : 0
-AVG RTT    : 0 msec
-LAST QUERY : 1710344842 seconds ago
+FETCHES    :        0
+OPEN CONNS :        0
+AVG RTT    :        0 msec
+LAST QUERY : none sent
 LAST REPLY : none received
 PINGS SENT :        0
 PINGS ACKED:        0   0%
 IGNORED    :        0   0%
 Histogram of PINGS ACKED:
-Last failed connect() at: 13/Mar/2024:15:47:16 +0000
+Last failed connect() at: 13/Mar/2024:15:54:31 +0000
 keep-alive ratio: 0%
```

yadij (dismissed):
there are line indentation issues. Pplease run `scripts/source-maintenance.sh` and merge the relevant adjustments.

src/neighbors.cc
```diff
@@ -1503,78 +1503,78 @@ static void
...
+        for (i = 0; i < e->n_addresses; ++i)
+            os << "Address[" << i << "] : " << e->addresses[i].toStr(ntoabuf, MAX_IPSTRLEN) << "\n";
 
-        storeAppendPrintf(sentry, "Status     : %s\n",
-                          neighborUp(e) ? "Up" : "Down");
-        storeAppendPrintf(sentry, "FETCHES    : %d\n", e->stats.fetches);
-        storeAppendPrintf(sentry, "OPEN CONNS : %d\n", e->stats.conn_open);
-        storeAppendPrintf(sentry, "AVG RTT    : %d msec\n", e->stats.rtt);
+        os << "Status     : " << (neighborUp(e) ? "Up" : "Down") << "\n" <<
+            "FETCHES    : " << std::setw(8) << std::right << e->stats.fetches << "\n" <<
```
rousskov, line 1529: I recommend going in the opposite direction: Instead of adding/duplicating tricks that attempt to make output "pretty" (and arguably fail to accomplish that in many basic cases), I would drop these attempts and just output `key: value` lines. I do not insist on that change in direction.

BTW, if you would like to make output YAML-compliant, please go ahead. Again, I do not insist on that.

rousskov (changes requested):
Thank you for posting sample output and the diff!

src/neighbors.cc
```diff
@@ -1503,78 +1503,78 @@ static void
...
+        os << "Flags      :";
+        os.flush();
         dump_peer_options(sentry, e);
 
-        for (i = 0; i < e->n_addresses; ++i) {
-            storeAppendPrintf(sentry, "Address[%d] : %s\n", i,
-                              e->addresses[i].toStr(ntoabuf,MAX_IPSTRLEN) );
-        }
+        for (i = 0; i < e->n_addresses; ++i)
+            os << "Address[" << i << "] : " << e->addresses[i].toStr(ntoabuf, MAX_IPSTRLEN) << "\n";
```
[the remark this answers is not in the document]
kinkie: Won't this have the same problem of emitting an URL instead of the naked IP?
I have another draft PR aiming to change the awkward API to instead return a SBuf (as an additional call / wrapper, at least during the transition)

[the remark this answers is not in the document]
rousskov: > Won't this have the same problem of emitting an URL instead of the naked IP?

It is difficult for me to answer this loaded question in a meaningful way. FWIW, my change request does not contradict https://github.com/squid-cache/squid/pull/1730#discussion_r1519722148. If you see specific problems that make the requested switch a bad idea, let's discuss!

----

> I have another draft PR aiming to change the awkward API to instead return a SBuf (as an additional call / wrapper, at least during the transition)

That plan raises a few red flags for me. FWIW, I recommend limiting the corresponding initial changes to a couple of use cases to avoid rewriting a lot of code.

src/neighbors.cc
```diff
@@ -1503,78 +1503,78 @@ static void
...
-                          e->http_port,
-                          e->icp.port);
-        storeAppendPrintf(sentry, "Flags      :");
+        os << "\n" << std::setw(11) << std::left <<
+            neighborTypeStr(e) << ": " << e->name << "\n";
+        os << "Host       : " << e->host << '/' << e->http_port << '/' <<
+            e->icp.port << "\n";
+        os << "Flags      :";
+        os.flush();
         dump_peer_options(sentry, e);
```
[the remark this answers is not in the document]
kinkie: Done. I've also taken the opportunity to improve some other bits and bobs

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+    if (options.default_parent)
+        os << " default";
+
+    if (options.roundrobin)
+        os << " round-robin";
+
+    if (options.carp)
+        os << " carp";
+
+#if USE_AUTHu
```
rousskov, line 152:
```suggestion
#if USE_AUTH
```

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+        os << " ttl=" << mcast.ttl;
+
+    if (connect_timeout_raw > 0)
+        os << " connect-timeout=" << connect_timeout_raw;
+
+    if (connect_fail_limit != PEER_TCP_MAGIC_COUNT)
+        os << " connect-fail-limit=" << connect_fail_limit;
+
+#if USE_CACHE_DIGESTS
+
```
rousskov, line 229:
```suggestion
```

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+    if (digest_url)
+        os << " digest-url=" << digest_url;
+
```
rousskov, line 232:
```suggestion
```

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+    if (connection_auth == 0)
+        os << " connection-auth=off";
+    else if (connection_auth == 1)
+        os << " connection-auth=on";
+    else if (connection_auth == 2)
+        os << " connection-auth=auto";
+
+    secure.dumpCfg(os, "tls-");
+    os << '\n';
```
rousskov, line 261: I recommend moving this line to callers. It is not appropriate for this method to decide where (or whether) new lines should be added after the options. I do not insist on this change.

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+}
+
+void
+CachePeer::reportStatistics (std::ostream& os)
+{
+    os << "\n" << std::setw(11) << std::left << typeString() <<
```
rousskov, line 267: Please move these opening new line to the caller (where it will probably become a closing new line).

src/CachePeer.h
```diff
@@ -48,6 +48,12 @@ class CachePeer
     /// \returns the effective connect timeout for the given peer
     time_t connectTimeout() const;
 
+    /// output to stream a representation of the peer's statistics
+    void reportStatistics (std::ostream& s);
+
+    /// output to stream a representation of the peer's configuration options
+    void dumpOptions(std::ostream& );
```
rousskov, line 55:
```suggestion
    void dumpOptions(std::ostream &) const;
```

And please check all other new methods for const-correctness.

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+        ": " << n_addresses << "\n" <<
```
rousskov, line 268: This part of the dump should be revised to restore the lost peer name reporting. I also strongly recommend changing the format from the old `<type>: <name>` to

```
name: <name>
type: <type>
```

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+
+    char ntoabuf[MAX_IPSTRLEN];
+    for (int i = 0; i < n_addresses; ++i)
+        os << "Address[" << i << "] : " <<
+            addresses[i].toStr(ntoabuf, MAX_IPSTRLEN) << "\n";
+
+    os << "Status     : " << (neighborUp(this) ? "Up" : "Down") << "\n" <<
+        "FETCHES    : " << stats.fetches << "\n" <<
+        "OPEN CONNS : " << stats.conn_open << "\n" <<
+        "AVG RTT    : " << stats.rtt << " msec\n";
```
rousskov, line 282: How about migrating away from ALLCAPS and cryptic abbreviations?

BTW, if you would like to make output YAML-compliant, please go ahead.

I do not insist on any of these things.

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+
+    if (!options.no_query) {
```
rousskov, line 284: I would drop this condition. We should be reporting what has happened (instead of trying to avoid reporting what should not be happening based on the current configuration). This change request works best with the removal of "none sent/received" lines discussed in other change requests.

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+        if (stats.last_query > 0)
+            os << "LAST QUERY : " <<
+                (squid_curtime - stats.last_query) << " seconds ago\n";
+        else
+            os << "LAST QUERY : none sent\n";
```
rousskov, line 289: I recommend not reporting what has not happened. If nothing else, the proposed text could mislead the reader into thinking that Squid did send some queries, but Squid was not able to send the last few queries (i.e. the last query sending attempts have failed).

```suggestion
```

src/CachePeer.cc
```diff
@@ -123,6 +125,230 @@ CachePeer::connectTimeout() const
...
+
+        if (stats.last_reply > 0)
+            os << "LAST REPLY : " <<
+                (squid_curtime - stats.last_reply) << " seconds ago\n";
+        else
+            os << "LAST REPLY : none received\n";
```
rousskov, line 295: Similarly, I would remove this line, even though it was reported before.

```suggestion
```

rousskov (changes requested):
This PR needs a lot more small fixes than I have time for right now. I will come back to this later, but I left a few change request to facilitate progress.

kinkie:
> there are line indentation issues. Pplease run `scripts/source-maintenance.sh` and merge the relevant adjustments.

done

src/CachePeer.cc
kinkie: good idea

src/neighbors.cc
```diff
@@ -1503,78 +1503,78 @@ static void
...
-                          neighborUp(e) ? "Up" : "Down");
-        storeAppendPrintf(sentry, "FETCHES    : %d\n", e->stats.fetches);
-        storeAppendPrintf(sentry, "OPEN CONNS : %d\n", e->stats.conn_open);
-        storeAppendPrintf(sentry, "AVG RTT    : %d msec\n", e->stats.rtt);
+        os << "Status     : " << (neighborUp(e) ? "Up" : "Down") << "\n" <<
+            "FETCHES    : " << std::setw(8) << std::right << e->stats.fetches << "\n" <<
+            "OPEN CONNS : " << std::setw(8) << std::right << e->stats.conn_open << "\n" <<
+            "AVG RTT    : " << std::setw(8) << std::right << e->stats.rtt << " msec\n";
 
         if (!e->options.no_query) {
```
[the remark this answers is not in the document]
kinkie: I agree. Most of the time there is no difference, and where there is, your idea is sensible.

src/neighbors.cc
```diff
@@ -1503,78 +1503,78 @@ static void
...
-            storeAppendPrintf(sentry, "LAST QUERY : %8d seconds ago\n",
-                              (int) (squid_curtime - e->stats.last_query));
+            if (e->stats.last_query > 0)
+                os << "LAST QUERY : " << std::setw(8) << std::right <<
+                    (squid_curtime - e->stats.last_query) << " seconds ago\n";
+            else
+                os << "LAST QUERY : none sent\n";
```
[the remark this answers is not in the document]
kinkie: removed

src/neighbors.cc
kinkie: For now I'll leave the current format, minus the formatting

src/CachePeer.h
```diff
@@ -65,6 +71,7 @@ class CachePeer
     char *host = nullptr;
 
     peer_t type = PEER_NONE;
+    const char * typeString() const;
```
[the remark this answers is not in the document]
kinkie: I thought about that, and originally decided to keep it close to the type it refers to. Happy to move it

src/CachePeer.cc
kinkie: done. Also moved n_addresses to its own line

src/CachePeer.cc
kinkie: I agree on the all caps, removing that. I'd defer YAML to another time

src/CachePeer.cc
kinkie: removed

src/neighbors.cc
```diff
@@ -1503,78 +1503,78 @@ static void
...
             if (e->stats.last_reply > 0)
-                storeAppendPrintf(sentry, "LAST REPLY : %8d seconds ago\n",
-                                  (int) (squid_curtime - e->stats.last_reply));
+                os << "LAST REPLY : " << std::setw(8) << std::right <<
+                    (squid_curtime - e->stats.last_reply) << " seconds ago\n";
             else
-                storeAppendPrintf(sentry, "LAST REPLY : none received\n");
-
-            storeAppendPrintf(sentry, "PINGS SENT : %8d\n", e->stats.pings_sent);
+                os << "LAST REPLY : none received\n";
```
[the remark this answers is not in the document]
kinkie: removed

kinkie:
Current sample output:

```
Name       : 127.0.0.1
Type       : Sibling
Addresses  : 1
Host       : 127.0.0.1/13129/0
Flags      : default connection-auth=auto tls-disable
Address[0] : 127.0.0.1
Status     : Down
Fetches    : 0
Open conns : 0
Avg rtt    : 0 msec
Pings sent : 0
Pings acked: 0 0%
Ignored    : 0 0%
Histogram of pings acked:
Last failed connect() at: 15/Mar/2024:05:27:58 +0000
keep-alive ratio: 0%
```

src/CachePeer.cc
```diff
@@ -123,6 +126,207 @@ CachePeer::connectTimeout() const
...
+           (squid_curtime - stats.last_reply) << " seconds ago\n";
+
+    os << "Pings sent : "  << stats.pings_sent << '\n' <<
+       "Pings acked: " << stats.pings_acked << " " <<
+       Math::intPercent(stats.pings_acked, stats.pings_sent) << "%\n";
+
+    os << "Ignored    : " << stats.ignored_replies << " " <<
+       Math::intPercent(stats.ignored_replies, stats.pings_acked) << "%\n";
+
+    os << "Histogram of pings acked:\n";
```
rousskov, line 285: Please do not print this section header when the section itself is empty (which is a common or even default case). The best way to address this change request could be adding a simple AtMostOnce class to IoManip.h:

```C++
const auto histogramHeader = AtMostOnce("Histogram of pings acked:\n");
if (options.htcp) {
    os << histogramHeader << ...;
} else {
    for (...) {
        if (...) continue;
        os << histogramHeader << ...;
}
```

Needless to say, this class will be useful in many other contexts and can be later extended to support types other than string literals.

src/CachePeer.cc
```diff
@@ -123,6 +126,207 @@ CachePeer::connectTimeout() const
...
+
+#if USE_HTCP
+    if (options.htcp) {
+        os << "\tMisses\t" << htcp.counts[0] << " " <<
```
rousskov, line 289: Please do not use tabs to avoid mixing tabs and spaces for output formatting. My earlier suggestion for (also) avoiding setw() still stands.

src/CachePeer.cc
```diff
@@ -123,6 +126,207 @@ CachePeer::connectTimeout() const
...
+           Math::intPercent(htcp.counts[0], stats.pings_acked) << "%\n" <<
+           "\tHits\t" << htcp.counts[1] << " " <<
+           Math::intPercent(htcp.counts[1], stats.pings_acked) << "%\n";
+    } else {
+#endif
+        for (auto op : WholeEnum<icp_opcode>()) {
+            if (icp.counts[op] == 0)
+                continue;
+
+            os << "    " << std::setw(12) << std::setprecision(12) <<
```
rousskov, line 299: The official code does not use 12-digit precision. Moreover, this code does not print floating point numbers so setprecision() cannot be useful here.

The best way to address this and other related (but not yet spelled out) concerns is to add an AsPercent class to IoManip.h. The constructor should take part and whole arguments...

src/CachePeer.cc
```diff
@@ -123,6 +126,207 @@ CachePeer::connectTimeout() const
...
+           addresses[i].toStr(ntoabuf, MAX_IPSTRLEN) << '\n';
+
+    os << "Status     : " << (neighborUp(this) ? "Up" : "Down") << '\n' <<
+       "Fetches    : " << stats.fetches << '\n' <<
+       "Open conns : " << stats.conn_open << '\n' <<
+       "Average RTT: " << stats.rtt << " msec\n";
+
+    if (stats.last_query > 0)
+        os << "Last query : " <<
+           (squid_curtime - stats.last_query) << " seconds ago\n";
```
rousskov, line 272: Please use braces for multiline `if` statements, including this one. Please place the opening brace on the same line as the `if` condition (other PR `if` statements violate this "ENFORCED" [rule](https://wiki.squid-cache.org/DeveloperResources/SquidCodingGuidelines#enforced-rules) and should be fixed).

src/CachePeer.cc
```diff
@@ -123,6 +126,207 @@ CachePeer::connectTimeout() const
...
+       "Host       : " << host << '/' << http_port << '/' <<
+       icp.port << '\n';
+    os << "Flags      :";
+    dumpOptions(os);
+    os << '\n';
+
+    char ntoabuf[MAX_IPSTRLEN];
+    for (int i = 0; i < n_addresses; ++i)
+        os << "Address[" << i << "] : " <<
+           addresses[i].toStr(ntoabuf, MAX_IPSTRLEN) << '\n';
```
rousskov, line 263: I would simplify by using operator "<<" for Ip::Address printing. We are changing output format for other reasons anyway.

src/CachePeer.cc
```diff
@@ -123,6 +126,207 @@ CachePeer::connectTimeout() const
...
+
+    os << "Status     : " << (neighborUp(this) ? "Up" : "Down") << '\n' <<
+       "Fetches    : " << stats.fetches << '\n' <<
+       "Open conns : " << stats.conn_open << '\n' <<
```
rousskov, line 267: Do you know whether `Open conns : 5` represents a valid YAML key:value pair? And, if yes, does the key value include the space after `conns`?

rousskov (changes requested):

src/CachePeer.cc
yadij: FTR; in cache manager "text" the HTAB character has semantic value. cachemanager.cgi interprets HTAB at the beginning of a line delimit a table row (HTML TR tags), HTAB within a line delimit table cells (HTML TD tags).

src/CachePeer.cc
kinkie: Done and waiting for PR #1742

yadij: not done AFAICT

kinkie: Right, that is waiting for AtMostOnce. I can resume this now that that one is unblocked

src/CachePeer.cc
kinkie: The whole thing is not well-formed YAML, because keys need to be unique in an object.
Reworking with indentation so that it is

src/CachePeer.cc
kinkie: ok

src/CachePeer.cc
kinkie: AsPercent seems a bit overkill for now, I'll just avoid setting the precision - I suspect
it comes out of printf() formatting, which we are doing away with anyway

kinkie:
Current sample output:

```yaml
- localhost:
  type: sibling
  host: localhost/3129/0
  flags: default connection-auth=auto tls-disable
  addresses: [ "[::1]" ]
  status: Down
  fetches: 0
  open connections: 0
  average RTT: 0 msec
  pings sent: 0
  pings acked: 0 0%
  replies ignored: 0 0%
  last failed connection at: 09/Apr/2024:15:42:02 +0700
  keep-alive ratio: 0%

- localhost.localdomain:
  type: sibling
  host: localhost.localdomain/3130/0
  flags: default connection-auth=auto tls-disable
  addresses: [  ]
  status: Down
  fetches: 0
  open connections: 0
  average RTT: 0 msec
  pings sent: 0
  pings acked: 0 0%
  replies ignored: 0 0%
  keep-alive ratio: 0%

```

I've validated it at https://www.yamllint.com/; it's valid syntax.

src/CachePeer.cc
kinkie: Now the output is valid yaml

src/CachePeer.cc
kinkie: done

src/CachePeer.cc
kinkie: removed the precision, if we rely on yaml instead of spacing it's completely unnecessary anyway

src/CachePeer.cc
```diff
@@ -114,6 +117,208 @@ CachePeer::connectTimeout() const
...
+        os << " connection-auth=auto";
+
+    secure.dumpCfg(os, "tls-");
+}
+
+void
+CachePeer::reportStatistics (std::ostream& os)
+{
+    os <<
+       "- " << name << ":\n" <<
```
rousskov, line 242:
```
- localhost:
  type: sibling

- localhost.localdomain:
  type: sibling
```

> it's valid syntax.

Thank you for doing the legwork required to reach this important milestone!

The current valid YAML output is also more-or-less human-friendly -- another requirement met.

Unfortunately, the current output is not machine-friendly enough. With some YAML/JSON libraries, this output is nearly unusable because getting the name of a cache_peer would require such heroic (and fragile) efforts as enumerating all cache_peer hash keys to find the key that is actually a cache_peer name.

Fortunately, this is easy to fix. I suggest the following as a starting point:


```suggestion
    os << "-";
    
    os << "  name: " << name << "\n" <<
```

or, to make each cache_peer reporting line format the same (at the expense of more lines in the report), add a new line after "-"

```suggestion
    os << "-\n";
    
    os << "  name: " << name << "\n" <<
```

src/CachePeer.cc
```diff
@@ -114,6 +117,208 @@ CachePeer::connectTimeout() const
...
+           Math::intPercent(htcp.counts[1], stats.pings_acked) << "%\n";
+    } else {
+#endif
+        for (auto op : WholeEnum<icp_opcode>()) {
+            if (icp.counts[op] == 0)
+                continue;
+
+            os << sectionHeader;
+            os << "    " <<
+               std::right << icp_opcode_str[op] << ": " <<
```
rousskov, line 292: Specifying right/left alignment only makes sense when also setting output field width, right?

```suggestion
            os << "    " << icp_opcode_str[op] << ": " <<
```

src/CachePeer.cc
```diff
@@ -114,6 +117,208 @@ CachePeer::connectTimeout() const
...
+       "  open connections: " << stats.conn_open << '\n' <<
+       "  average RTT: " << stats.rtt << " msec\n";
+
+    if (stats.last_query > 0) {
+        os << "  last query:" <<
+           (squid_curtime - stats.last_query) << " seconds ago\n";
+    }
+
+    if (stats.last_reply > 0) {
+        os << " last reply: " <<
```
rousskov, line 264: Missing one space in indentation? Should we add a yamlIndent(n = 2) manipulator to make this indentation more "explicit" and harder to screw up?

```suggestion
        os << yamlIndent << "last reply: " <<
```

Those lines that need to double the indentation can use either

```C++
        os << yamlIndent(4) << "htcp hits: " << ...
```

or (to reduce work for those implementing yamlIndent at the expense of callers):

```C++
        os << yamlIndent << yamlIndent << "htcp hits: " << ...
```

or, with more work for those implementing yamlIndent, even

```C++
        os << yamlIndent*2 << "htcp hits: " << ...
```

src/CachePeer.cc
```diff
@@ -114,6 +117,208 @@ CachePeer::connectTimeout() const
...
+    else if (connection_auth == 1)
+        os << " connection-auth=on";
+    else if (connection_auth == 2)
+        os << " connection-auth=auto";
+
+    secure.dumpCfg(os, "tls-");
+}
+
+void
+CachePeer::reportStatistics (std::ostream& os)
```
rousskov, line 239: To mark this output as YAML-converted, I suggest renaming `os`:

```suggestion
CachePeer::reportStatistics (std::ostream &yaml)
```

Eventually, we may even introduce an std::ostream wrapper to simplify YAML formatting code, but renaming now will still be useful then.

rousskov (changes requested):

src/CachePeer.cc
kinkie: sure

src/CachePeer.cc
kinkie: I thought about that, but didn't want to go there yet
I would call it indent() or spaces(), no need to be too specific here

src/CachePeer.cc
kinkie: It's even easier than that. Since the peer name is unique due to how config works, we can just do away with the array and have a map, keyed on the peer name.

This is what it looks like:
```yaml
localhost:
  type: sibling
  host: localhost/3129/0
  flags: default connection-auth=auto tls-disable
  addresses: [ "[::1]" ]
  status: Down
  fetches: 0
  open connections: 0
  average RTT: 0 msec
  pings sent: 0
  pings acked: 0 0%
  replies ignored: 0 0%
  last failed connection at: 10/Apr/2024:18:00:55 +0700
  keep-alive ratio: 0%

localhost.localdomain:
  type: sibling
  host: localhost.localdomain/3130/0
  flags: default connection-auth=auto tls-disable
  addresses: [  ]
  status: Down
  fetches: 0
  open connections: 0
  average RTT: 0 msec
  pings sent: 0
  pings acked: 0 0%
  replies ignored: 0 0%
  keep-alive ratio: 0%
```
It is valid yaml, and should address your concern

src/CachePeer.cc
rousskov: > I thought about that, but didn't want to go there yet

Indentation mistakes in this PR are enough reasons to add/use yamlIndent(), indent(), or spaces() IMO.

>  I would call it indent() or spaces(), no need to be too specific here

* yamlIndent() looks ugly but makes it easy to select the default number of spaces specifically for YAML output.
* indent() looks much better than yamlIndent(), but I doubt we will have two-space indentation in many contexts other than YAML. That is why I though of yamlIndent()... If we switch to four-space indentation for YAML, then I am OK with indent() or indent(n = 4) because that default works great in most text formatting contexts.
* spaces(n) look great and is more versatile, but having a default parameter for a function called spaces() is a bit odd IMHO so we would have to always say spaces(2) or spaces(4) explicitly for default indentation cases. This is acceptable IMO, but not ideal.

src/CachePeer.cc
rousskov: > It is valid yaml, and should address your concern

Agreed, but there is a catch. If we switch to a map, we are making the task of restoring configured cache_dir order (which is important in some contexts) much more difficult for consumers that load this report into an order-destroying hash (because that is what their yaml2foo library provides for YAML maps). If you switch to a map, please add a new field to each cache_dir entry to specify a 1-based configuration index (e.g., "configuration index: 2" for localhost.localdomain cache_peer in your example).

I am not an YAML conversion expert, so if an expert tells us that my concern is unfounded because all popular converters preserve order in maps, then we do not need to add a configuration index.

N.B. I am aware that we report cache_dirs in configuration order.

src/CachePeer.cc
kinkie: I'd go for spaces(n) as it keeps things simple and is highly visible.
If we wanted to have something stateful, it could be an indent object, with up() and down() methods, but is it really worth it?

rousskov: > I'd go for spaces(n) as it keeps things simple and is highly visible.

I assume you mean spaces() function _without_ a default value for its `n` parameter. I am OK with that option.

Please make it a constexpr function that only takes one parameter and returns a string literal, a `const char *` pointer, or equivalent. If you need help with that, please ask! Otherwise, we would have to use a stream manipulator pattern which is a lot of work producing worse result (please do not add an SBuf-returning or similar heavy/noisy/memory-allocating function!).

src/CachePeer.cc
```diff
@@ -114,6 +117,206 @@ CachePeer::connectTimeout() const
...
+       name << ":\n" <<
+       "  type: " << typeString() << '\n' <<
+       "  host: " << host << '/' << http_port << '/' << icp.port << '\n';
+    yaml << "  flags:";
+    dumpOptions(yaml);
+    yaml << '\n';
+
+    std::vector<Ip::Address> addr;
+    std::copy(addresses, addresses+n_addresses, std::back_inserter(addr));
+    yaml << "  addresses: [ " << AsList(addr).prefixedBy("\"").delimitedBy("\", \"").suffixedBy("\"") << " ]\n";
```
yadij, line 251: The prefix and suffix can just be hard-coded with the brackets to make this simpler.
```suggestion
    yaml << "  addresses: [ \"" << AsList(addr).delimitedBy("\", \"") << "\" ]\n";
```

src/CachePeer.cc
```diff
@@ -114,6 +117,206 @@ CachePeer::connectTimeout() const
...
+    yaml << "  replies ignored: " << stats.ignored_replies << " " <<
+       Math::intPercent(stats.ignored_replies, stats.pings_acked) << "%\n";
+
+    auto sectionHeader = AtMostOnce("  histogram of pings acked:\n");
+
+#if USE_HTCP
+    if (options.htcp) {
+        yaml << sectionHeader;
+        yaml << "    htcp misses: " << htcp.counts[0] << " " << Math::intPercent(htcp.counts[0], stats.pings_acked) << "%\n"
+             << "    htcp hits: " << htcp.counts[1] << " " << Math::intPercent(htcp.counts[1], stats.pings_acked) << "%\n";
```
yadij, line 281: If you are making this formally YAML. These two values should be separated and the line use set-brackets to bind them as a pair.
```suggestion
        yaml << "    htcp misses: { count: " << htcp.counts[0] << ", ratio: " << Math::intPercent(htcp.counts[0], stats.pings_acked) << "% }\n"
             << "    htcp hits:  { count: " << htcp.counts[1] << ", ratio: " << Math::intPercent(htcp.counts[1], stats.pings_acked) << "% }\n";
```

... same for the equivalent ICP lines below.

src/CachePeer.cc
```diff
@@ -114,6 +117,206 @@ CachePeer::connectTimeout() const
...
+    } else {
+#endif
+        for (auto op : WholeEnum<icp_opcode>()) {
+            if (icp.counts[op] == 0)
+                continue;
+
+            yaml << sectionHeader;
+            yaml << "    " <<
+               std::right << icp_opcode_str[op] << ": " <<
```
yadij, line 290: Please take advantage of the changes to add the "icp" prefix which this report was missing
```suggestion
               std::right << "icp " << icp_opcode_str[op] << ": " <<
```

yadij (changes requested):

src/CachePeer.cc
kinkie: Not really, that would turn the case where the list of addresses is empty to an empty string

src/CachePeer.cc
kinkie: .. or the whole thing could be considered as a string from yaml's perspective. I'm fine either way, do you have any strong preference?

src/CachePeer.cc
rousskov: This unreadable code code needs to be improved, but the correct fix is different. I did not say anything in my review because there are several overlapping problems in these two lines that may warrant a prerequisite PR if we start addressing them.

If you want to start moving in that direction, add AsList::quoted(void) or AsList::quotedValues(void).

```suggestion
    yaml << "  addresses: [ " << AsList(addr).quoted().delimitedBy(", ") << " ]\n";
```

src/CachePeer.cc
rousskov: > These two values should be separated and the line use set-brackets to bind them as a pair.

I am not against this change. FWIW, I said nothing about this in my review because it is not clear to me where to draw the line: Many other places in this output use a single YAML value for what is actually a "compound" value. Ideally, from a machine (i.e. non-human) consumer point of view, even "0 msec" should be converted to avoid post-YAML parsing. We _can_ convert all of these values, but that is quite a bit of work and will reduce output readability for human consumers.

FWIW, if we start converting these things, I will request the removal of the corresponding code duplication like the one in the suggested replacement. Again, that is not a showstopper, it can be done, and it will improve this code quality, but keep that additional work in mind when deciding where to draw the line.

src/CachePeer.cc
kinkie: PR #1779 implements AsList::quotedBy()

Waiting for that to be reviewed and merged

kinkie:
Waiting for PR #1779

src/CachePeer.cc
```diff
@@ -114,6 +117,206 @@ CachePeer::connectTimeout() const
     return Config.Timeout.peer_connect;
 }
 
+void
+CachePeer::dumpOptions(std::ostream &yaml) const
```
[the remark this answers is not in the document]
kinkie: Going back to "os"

src/neighbors.cc
```diff
@@ -1367,230 +1351,22 @@ neighborDumpPeers(StoreEntry * sentry)
...
 {
-    char ntoabuf[MAX_IPSTRLEN];
-    int i;
+    PackableStream os(*sentry);
 
     if (!peers) {
-        storeAppendPrintf(sentry, "There are no neighbors installed.\n");
+        os << "There are no neighbors installed.\n";
         return;
     }
```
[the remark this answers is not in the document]
kinkie: If we wanted to be super strict, the yaml tree should probably be something like:
```
number of neighbours: <number>
neighbours:
  <name1>:
    spec1: value
    spec2: value
    ...
  <name2>:
    spec1: value
    spec2: value
    ...
```

What you propose is good for a human, but might confuse automated clients because there are two different data types used as values for a map.
I don't have any strong opinion, what do you think @yadij ?

src/CachePeer.cc
kinkie: ok

src/CachePeer.cc
kinkie: Started using quoted() early in commit 1d10221d03, resolving

src/CachePeer.cc
kinkie: I would keep with my prior assessment and not convert, favouring humans.
If we really wanted to target machine consumers, we should probably be looking into supporting Prometheus' native format

src/neighbors.cc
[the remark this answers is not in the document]
kinkie: after 4798ca005c, the output format is now:

```yaml
cache_peers number: 2
cache_peers:
  - name: localhost
    type: sibling
    host: localhost/3129/0
    flags: default tls-disable
    addresses: [ "[::1]" ]
    status: Down
    fetches: 0
    open connections: 0
    average RTT: 0 msec
    pings sent: 0
    pings acked: 0 0%
    replies ignored: 0 0%
    last failed connection at: 13/Apr/2024:04:15:56 +0700
    keep-alive ratio: 0%
  - name: localhost.localdomain
    type: sibling
    host: localhost.localdomain/3130/0
    flags: default tls-disable
    addresses: [  ]
    status: Down
    fetches: 0
    open connections: 0
    average RTT: 0 msec
    pings sent: 0
    pings acked: 0 0%
    replies ignored: 0 0%
    keep-alive ratio: 0%
```
and
```yaml
cache_peers number: 0
```
in case there are no peers configured

src/CachePeer.cc
```diff
@@ -114,6 +117,206 @@ CachePeer::connectTimeout() const
...
+}
+
+void
+CachePeer::reportStatistics (std::ostream& yaml)
+{
+    yaml <<
+       name << ":\n" <<
+       "  type: " << typeString() << '\n' <<
+       "  host: " << host << '/' << http_port << '/' << icp.port << '\n';
+    yaml << "  flags:";
```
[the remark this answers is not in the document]
kinkie: done in 781cc657c1

src/CachePeer.cc
```diff
@@ -114,6 +117,206 @@ CachePeer::connectTimeout() const
...
+    secure.dumpCfg(yaml, "tls-");
+}
+
+void
+CachePeer::reportStatistics (std::ostream& yaml)
+{
+    yaml <<
+       name << ":\n" <<
+       "  type: " << typeString() << '\n' <<
+       "  host: " << host << '/' << http_port << '/' << icp.port << '\n';
```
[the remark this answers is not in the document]
kinkie: done in cc1542711d (without capitalization)

kinkie:
Current output format:

```yaml
cache_peers number: 2
cache_peers:
  - name: localhost
    type: sibling
    http address: localhost:3129
    options: default tls-disable
    addresses: [ "[::1]" ]
    status: Down
    fetches: 0
    open connections: 0
    average RTT: 0 msec
    pings sent: 0
    pings acked: 0 0%
    replies ignored: 0 0%
    last failed connection at: 13/Apr/2024:18:01:11 +0700
    keep-alive ratio: 0%
  - name: localhost.localdomain
    type: sibling
    http address: localhost.localdomain:3130
    options: default tls-disable
    addresses: [  ]
    status: Down
    fetches: 0
    open connections: 0
    average RTT: 0 msec
    pings sent: 0
    pings acked: 0 0%
    replies ignored: 0 0%
    keep-alive ratio: 0%
```

kinkie:
Possible thing to investigate: whether to have this action return `application/yaml` content type

yadij:
FWIW, @rousskov vetoed the change to allow reports to have their own `Content-Type`.

rousskov:
> FWIW, @rousskov vetoed the change to allow reports to have their own `Content-Type`.

FWIW, I did not.

rousskov:
> Possible thing to investigate: whether to have this action return `application/yaml` content type

That would be nice, but we cannot do that easily right now because this _action_ (as a whole) does not return YAML in SMP Squids. To get to YAML, we need to fix[^1] how SMP Squids "bracket" from-kid output, replacing current `by kidN { ... } by kidN` "brackets" with some YAML-compatible variation. Here is one option that limits bracketing modifications and mostly preserves their current advantages, but there may be better options:

```
by kid1: {
...
} # by kid1

by kid2: {
...
} # by kid2
```

Bracketing changes do not belong to _this_ PR.

[^1]: More fixes may be necessary because each kid-specific blob may need to be adjusted as well to keep aggregated output valid; I have not investigated what exact changes, if any, are needed there. Ideally, we should check/decide _before_ this PR goes in so that we do not have to modify non-SMP output twice. However, I cannot insist on that precondition.

src/CachePeer.cc
```diff
@@ -114,6 +118,209 @@ CachePeer::connectTimeout() const
...
+             Math::intPercent(htcp.counts[1], stats.pings_acked) << "%\n";
+    } else {
+#endif
+        for (auto op : WholeEnum<icp_opcode>()) {
+            if (icp.counts[op] == 0)
+                continue;
+
+            yaml << spaces(4) << sectionHeader;
+            yaml << spaces(6) <<
+                 "icp " << icp_opcode_str[op] << ": " <<
```
rousskov, line 295: Why should this PR add an `icp ` prefix to icp_opcode_str strings that all start with an `ICP_` prefix?

src/neighbors.cc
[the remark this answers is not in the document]
rousskov: > the output format is now: ...

Looks good to me overall, thank you! 

I would prefer "cache_peer directives:" to "cache_peers number:", but that is subjective, and I do not insist on that change.

If that does not break YAML, I would probably add an empty line between peers to improve readability, but that is rather subjective and I do not insist on that change.

src/CachePeer.cc
```diff
@@ -114,6 +118,209 @@ CachePeer::connectTimeout() const
...
+
+void
+CachePeer::reportStatistics (std::ostream& yaml)
+{
+    yaml <<
+         spaces(2) << "- name: " << name << "\n" <<
+         spaces(4) << "type: " << typeString() << '\n' <<
+         spaces(4) << "http address: " << host << ':' << http_port << '\n';
+    if (icp.port)
+        yaml << spaces(4) << "icp address:" << host << ':' << icp.port << '\n';
```
rousskov, line 246: I think we ought to use all-caps for these protocol name abbreviations. This is how the standards and the vast majority of folks spell them. Why be different in this for-human output that does not have readability, clashes, and other concerns that C++ names have?

```suggestion
         spaces(4) << "HTTP address: " << host << ':' << http_port << '\n';
    if (icp.port)
        yaml << spaces(4) << "ICP address:" << host << ':' << icp.port << '\n';
```

src/CachePeer.cc
```diff
@@ -114,6 +118,209 @@ CachePeer::connectTimeout() const
...
+         Math::intPercent(stats.ignored_replies, stats.pings_acked) << "%\n";
+
+    auto sectionHeader = AtMostOnce("histogram of pings acked:\n");
+
+#if USE_HTCP
+    if (options.htcp) {
+        yaml << spaces(4) << sectionHeader;
+        yaml << spaces(6) << "htcp misses: " << htcp.counts[0] << " " <<
+             Math::intPercent(htcp.counts[0], stats.pings_acked) << "%\n" <<
+             spaces(6) << "htcp hits: " << htcp.counts[1] << " " <<
```
rousskov, line 285:
```suggestion
        yaml << spaces(6) << "HTCP misses: " << htcp.counts[0] << " " <<
             Math::intPercent(htcp.counts[0], stats.pings_acked) << "%\n" <<
             spaces(6) << "HTCP hits: " << htcp.counts[1] << " " <<
```

src/CachePeer.cc
```diff
@@ -114,6 +118,209 @@ CachePeer::connectTimeout() const
...
+        os << " connection-auth=off";
+    else if (connection_auth == 1)
+        os << " connection-auth=on";
+    // else default connection-auth=auto
+
+    secure.dumpCfg(os, "tls-");
+}
+
+void
+CachePeer::reportStatistics (std::ostream& yaml)
```
rousskov, line 239: What prevents this method from being `const`? If nothing or if a trivial change would allow it, please make it `const`.

src/sbuf/Spaces.h
```diff
@@ -0,0 +1,16 @@
...
+ * contributions from numerous individuals and organizations.
+ * Please see the COPYING and CONTRIBUTORS files for details.
+ */
+
+#ifndef SQUID_SRC_SBUF_SPACES_H
+#define SQUID_SRC_SBUF_SPACES_H
+
+#include "sbuf/SBuf.h"
+
+const SBuf spaces(size_t count);
```
rousskov, line 14: FWIW, I will try to find the time to check whether this can become a constexpr (which may even improve some code in callers). If I succeed, I will fix the return type (extra/unwanted `const`), the lack of documentation for this new global function, and NoNewGlobals implementation violation at that time.

rousskov (changes requested):
Great progress, thank you! This is an incomplete review that should still allow you to move forward. I will come back to this PR later.

src/CachePeer.cc
kinkie: Nothing. Making it const.

src/CachePeer.cc
kinkie: Mostly for consistency, and with machine parsers in mind. Personal preference comes into play as well.
I think it's a nit, I'll change it if you feel strongly about it

src/CachePeer.cc
kinkie: Thanks, removing.

src/CachePeer.cc
rousskov: > Alex: Why be different in this for-human output that does not have readability, clashes, and other concerns that C++ names have?

> Francesco: Mostly for consistency

The proposed format (change) is not consistent with current Squid output and, more importantly, is very inconsistent with subject area norms. That problem trumps any desire to make for-admin output text consistent with for-developers source code names (if that is what you are implying). Not to mention that there are plenty of Squid source code names where HTTP is (correctly) formatted as `Http`, `http`, or (incorrectly) `HTTP`, so no single variant will be truly consistent with C++ code names.

> and with machine parsers in mind

Why would a machine parser care whether the token it is parsing contains "HTTP" or "http"?!

> I'll change it if you feel strongly about it

I hope this decision can be made based on (objective) merits rather than the (subjective) strength of our convictions.

src/CachePeer.cc
kinkie: done

src/CachePeer.cc
kinkie: > I hope this decision can be made based on (objective) merits rather than the (subjective) strength of our convictions.

I don't think there's a measurable benefit by taking either choice, so I'm not sure we could converge on some objective merit. But that's not a big deal, I have no attachment to either choice so I'm changing it

src/sbuf/Spaces.h
kinkie: Based on our conversation, I'd hold off on this experiment; spaces (or its successor) will likely need to hold some state

kinkie:
Current output:

```yaml
cache_peers number: 3
cache_peers:
  - name: "localhost"
    type: parent
    HTTP address: localhost:3129
    ICP address: localhost:3129
    options:  default tls-disable
    addresses: [ "[::1]" ]
    status: Down
    fetches: 0
    open connections: 0
    average RTT: 0 msec
    pings sent: 0
    pings acked: 0 0%
    replies ignored: 0 0%
    last failed connection at: 17/Apr/2024:09:46:22 +0100
    keep-alive ratio: 0%
  - name: "127.0.0.1"
    type: parent
    HTTP address: 127.0.0.1:3130
    ICP address: 127.0.0.1:3130
    options:  default tls-disable
    addresses: [ "127.0.0.1" ]
    status: Down
    fetches: 0
    open connections: 0
    average RTT: 0 msec
    pings sent: 0
    pings acked: 0 0%
    replies ignored: 0 0%
    last failed connection at: 17/Apr/2024:09:46:22 +0100
    keep-alive ratio: 0%
  - name: "::1"
    type: parent
    HTTP address: ::1:3131
    ICP address: ::1:3131
    options:  default tls-disable
    addresses: [ "[::1]" ]
    status: Down
    fetches: 0
    open connections: 0
    average RTT: 0 msec
    pings sent: 0
    pings acked: 0 0%
    replies ignored: 0 0%
    last failed connection at: 17/Apr/2024:09:46:22 +0100
    keep-alive ratio: 0%
```

rousskov (changes requested):
Another incomplete review to help you make progress.

PR #1738, opened 2024-03-14
Convert loadable_modules to SBufList

yadij:
no behaviour changes.

### DIFF
--- a/src/LoadableModule.cc
+++ b/src/LoadableModule.cc
@@ -15,7 +15,7 @@
 // is seen by the static linker, the linker will complain.
 
 LoadableModule::LoadableModule(const SBuf &aName):
-        theName(aName)
+    theName(aName)
 {
     // Initialise preloaded symbol lookup table.
     LTDL_SET_PRELOADED_SYMBOLS();
--- a/src/LoadableModule.cc
+++ b/src/LoadableModule.cc
@@ -14,7 +14,8 @@
 // Note: We must use preprocessor instead of C ifs because if dlopen()
 // is seen by the static linker, the linker will complain.
 
-LoadableModule::LoadableModule(const String &aName): theName(aName), theHandle(nullptr)
+LoadableModule::LoadableModule(const SBuf &aName):
+        theName(aName)
 {
     // Initialise preloaded symbol lookup table.
     LTDL_SET_PRELOADED_SYMBOLS();
@@ -62,7 +63,7 @@ LoadableModule::unload()
 void *
 LoadableModule::openModule()
 {
-    return lt_dlopen(theName.termedBuf());
+    return lt_dlopen(theName.c_str());
 }
 
 bool
--- a/src/LoadableModule.h
+++ b/src/LoadableModule.h
@@ -19,8 +19,8 @@ public:
     ~LoadableModule();           // unloads if loaded
 
     bool loaded() const;
-    const SBuf &name() const { return theName; }
-    const SBuf &error() const { return theError; }
+    const auto &name() const { return theName; }
+    const auto &error() const { return theError; }
 
     void load(); // throws Texc
     void unload(); // throws Texc
--- a/src/LoadableModule.h
+++ b/src/LoadableModule.h
@@ -15,7 +15,7 @@
 class LoadableModule
 {
 public:
-    LoadableModule(const SBuf &aName);
+    explicit LoadableModule(const SBuf &aName);
     ~LoadableModule();           // unloads if loaded
 
     bool loaded() const;
--- a/src/LoadableModule.h
+++ b/src/LoadableModule.h
@@ -9,26 +9,26 @@
 #ifndef SQUID_SRC_LOADABLEMODULE_H
 #define SQUID_SRC_LOADABLEMODULE_H
 
-#include "SquidString.h"
+#include "sbuf/SBuf.h"
 
 // wrapper for dlopen(3), libltdl, and friends
 class LoadableModule
 {
 public:
-    LoadableModule(const String &aName);
+    LoadableModule(const SBuf &aName);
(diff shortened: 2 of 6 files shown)


kinkie (dismissed):
I assume it has been runtime tested? If so, go go go

src/LoadableModule.h
```diff
@@ -9,26 +9,26 @@
...
 
-#include "SquidString.h"
+#include "sbuf/SBuf.h"
 
 // wrapper for dlopen(3), libltdl, and friends
 class LoadableModule
 {
 public:
-    LoadableModule(const String &aName);
+    LoadableModule(const SBuf &aName);
```
rousskov, line 18: If possible without heroic efforts, let's block unwanted implicit conversions:

```suggestion
    explicit LoadableModule(const SBuf &aName);
```

src/LoadableModules.cc
```diff
@@ -8,29 +8,30 @@
...
+#include "sbuf/List.h"
 
 static void
-LoadModule(const char *fname)
+LoadModule(const SBuf &fname)
 {
     debugs(1, DBG_IMPORTANT, "Loading Squid module from '" << fname << "'");
 
-    LoadableModule *m = new LoadableModule(fname);
+    auto *m = new LoadableModule(fname);
```
rousskov, line 20: This line did not have to be changed AFAICT, but if this PR changes it, let's improve its const-correctness.

```suggestion
    const auto m = new LoadableModule(fname);
```

src/LoadableModules.cc
```diff
@@ -8,29 +8,30 @@
...
 void
-LoadableModulesConfigure(const wordlist *names)
+LoadableModulesConfigure(const SBufList &names)
 {
-    int count = 0;
-    for (const wordlist *i = names; i; i = i->next, ++count)
-        LoadModule(i->key);
+    unsigned int count = 0;
+    for (const auto &which : names) {
+        LoadModule(which);
```
rousskov, line 32: Please use a better fitting variable name:

```suggestion
    for (const auto &name: names) {
        LoadModule(name);
```

src/LoadableModules.cc
```diff
@@ -8,29 +8,30 @@
...
 }
 
 void
-LoadableModulesConfigure(const wordlist *names)
+LoadableModulesConfigure(const SBufList &names)
 {
-    int count = 0;
-    for (const wordlist *i = names; i; i = i->next, ++count)
-        LoadModule(i->key);
+    unsigned int count = 0;
```
rousskov, line 30: This PR does not have to change this type, but since it does, let's use a more appropriate type for counting things inside a container:

```suggestion
    size_t count = 0;
```

Alternatively, remove `count` and just report names.size(), simplifying this code further!

src/LoadableModules.h
```diff
@@ -9,11 +9,13 @@
...
 #define SQUID_SRC_LOADABLEMODULES_H
 
+#include "sbuf/forward.h"
+
 // TODO: add reporting for cachemgr
 // TODO: add reconfiguration support
 
-class wordlist;
-void LoadableModulesConfigure(const wordlist *names);
+/// list of library filenames to load dynamically
```
rousskov, line 17: The added description does not match the function it describes (it kind of matches its parameter instead).

```suggestion
/// dynamically load named libraries, in the listed order
```

or

```suggestion
/// Dynamically loads libraries identified by their filenames.
/// The libraries are loaded in the listed order.
```

or something like that.

src/LoadableModule.h
```diff
@@ -9,26 +9,26 @@
...
     ~LoadableModule();           // unloads if loaded
 
     bool loaded() const;
-    const String &name() const { return theName; }
-    const String &error() const { return theError; }
+    const SBuf &name() const { return theName; }
+    const SBuf &error() const { return theError; }
```
rousskov, line 23: I do not insist on this, but we can apply AAA here as well:

```suggestion
    const auto &name() const { return theName; }
    const auto &error() const { return theError; }
```

src/LoadableModule.cc
```diff
@@ -14,7 +14,8 @@
 // Note: We must use preprocessor instead of C ifs because if dlopen()
 // is seen by the static linker, the linker will complain.
 
-LoadableModule::LoadableModule(const String &aName): theName(aName), theHandle(nullptr)
+LoadableModule::LoadableModule(const SBuf &aName):
+        theName(aName)
```
rousskov, line 18: Just to fix [formatting](https://github.com/squid-cache/squid/actions/runs/8277010739/job/22646550395#step:4:153):

```suggestion
    theName(aName)
```

rousskov (changes requested):

src/LoadableModule.cc
yadij: done in [fd858aa](https://github.com/squid-cache/squid/pull/1738/commits/fd858aaa13035510319477ae023f7064982d4576)

src/LoadableModules.cc
yadij: done in [fd858aa](https://github.com/squid-cache/squid/pull/1738/commits/fd858aaa13035510319477ae023f7064982d4576)

src/LoadableModules.cc
yadij: done in [fd858aa](https://github.com/squid-cache/squid/pull/1738/commits/fd858aaa13035510319477ae023f7064982d4576)

src/LoadableModules.cc
yadij: done in [fd858aa](https://github.com/squid-cache/squid/pull/1738/commits/fd858aaa13035510319477ae023f7064982d4576)

rousskov (approved):
Thank you.

PR #1739, opened 2024-03-15
WIP: implement Ip::Address::asText()

kinkie:
Progressively migrate all users of Ip::Address::toStr, toUrl, toHostStr to
a new manipulator that clarifies how to format the output and simplifies

### DIFF
--- a/src/tests/testIpAddress.cc
+++ b/src/tests/testIpAddress.cc
@@ -75,10 +75,9 @@ protected:
     void testtoStr();
     void testtoUrl_fromInAddr();
     void testtoUrl_fromSockAddr();
+    void testAddressText();
     void testgetReverseString();
     void testMasking();
-
-    void testAddressText();
 
     // bugs.
     void testBugNullingDisplay();


rousskov:
@kinkie, I removed undocumented `S-waiting-for-feedback` label. We already have `S-waiting-for-more-reviewers` that is better because it is compatible (as in "does not overlap") with existing `S-waiting-for-reviewer`.

src/peer_sourcehash.cc
```diff
@@ -137,28 +137,25 @@ peerSourceHashRegisterWithCacheManager(void)
...
-    char ntoabuf[MAX_IPSTRLEN];
 
     if (SourceHashPeers().empty())
         return nullptr;
 
     assert(ps);
     HttpRequest *request = ps->request;
 
-    key = request->client_addr.toStr(ntoabuf, sizeof(ntoabuf));
+    const auto key = request->client_addr.toStrAsSBuf();
```
yadij, line 152: This use-case does not need any new API. It can just:
```suggestion
    const auto key = ToSBuf(request->client_addr);
```

src/ip/Address.h
```diff
@@ -211,6 +212,7 @@ class Address
      \return pointer to buffer received.
      */
     char* toStr(char *buf, const unsigned int blen, int force = AF_UNSPEC) const;
+    SBuf toStrAsSBuf(int force = AF_UNSPEC) const;
```
yadij, line 215: IMO the `AsSBuf` portion of the naming is more harm than good. We should just be deprecating the `char*` API and otherwise keeping the naming the same.
```suggestion
    SBuf toStr(int force = AF_UNSPEC) const;
    /// Deprecated. Use the SBuf method instead.
    char* toStr(char *buf, const unsigned int blen, int force = AF_UNSPEC) const;
```

yadij (changes requested):

src/peer_sourcehash.cc
rousskov: Do we want to guarantee that two different Squid instances produce the same hash for the same request?

* If yes, then we may want to avoid changing format/hash to avoid  breaking that guarantee for instances running different Squid versions. We could still change if there is a compelling reason to do so, of course, but I do not see such a reason here.
* Otherwise, we should not covert the IP address to text to compute a hash.

src/ip/Address.h
rousskov: I do agree that toStrAsSBuf() is a non-starter as far as spelling is concerned. However, FWIW, following the above renaming [suggestion](https://github.com/squid-cache/squid/pull/1739#discussion_r1526454769) will _not_ address my primary [concerns](https://github.com/squid-cache/squid/pull/1739#pullrequestreview-1939010545) and, hence, will not move this PR forward.

yadij:
> There are three primary problems with existing Ip::Address::toStr():
> 
...
>     2. Give-me-a-large-enough-buffer API is usually awkward to use, especially in modern C++ code. Its usage also results in a lot of code duplication.
> 

Er, these are APIs for use in old C-code. They have never been true C++ APIs. It looks to me like this update to use SBuf is part of fixing that.

>     3. The "force IP version" part of API is a bad idea with worse implementation, resulting in ugly code and bugs, possibly including security vulnerabilities.
> 

FTR; That is an issue forced on us by:
 * kernel folks insisting on Dual-Stack TCP implementations, and
 * sysadmin insisting on IPv4-only networks, and
 * ASCII-based protocols requiring specific address representations [1]
 * and a number of other more uncommon use-cases.

[1] long-term we should use the `IPvFuture` representation from [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2) whenever possible. I do not think the real world is ready for that yet, case in point being our own IP parser.

> 
> When improving this API and (eventually) modifying all its callers, we should aim at solving all these problems. The current PR comes short because it only attempts to solve the second toStr() problem. The current PR is arguably worse than simply replacing existing SBuf-needing ip.toStr(...) calls with ToSBuf(ip)!

Falls foul of your issue (1). ToSBuf() uses the stream operator which was designed solely for `debugs()` uses and is equivalent to `toUrl()` not `toStr()`. It is only good for strings which can cope with URL-format ("[]" wrappers).


> ### Ip::AddressText asText(const Ip::Address &)
> 
> To address the above three (and other) problems, I propose to introduce an asText() function that returns an object of an Ip::AddressText class. 

I support this plan.

rousskov:
> > To address the above three (and other) problems, I propose to introduce an asText() function that returns an object of an Ip::AddressText class.
 
> I support this plan.

Glad to hear that! The next step is to decide how to address problem 1 (i.e. Current APIs do a poor job distinguishing various output formats). I see two basic options/directions:

1a. Default to "minimal" (for the lack of a better word) format (i.e. no brackets and no port). Optional method calls add bells and whistles. We use this approach with AsList and, to some degree, AsHex. Cons: This "minimal" format is arguably the most dangerous one in this case; copy-pasted code does not bring reviewer attention to wrong format for the new location.

1b. Default to "bracketed with port" format. Method calls adjust as needed. Cons: This default is wrong for some use cases; copy-pasted code does not bring reviewer attention to wrong format for the new location.

2. Require each caller to make formatting choices. Cons: Should use an uglier enum-based interfaces illustrated below:

```C++
os << AsText(ip); // Compile error because the caller forgot to make two required formatting choices
os << AsText(ip, Ip::Text::withoutPort); // Compile error because the caller forgot to make bracketing choice
os << AsText(ip, Ip::Text::withoutBrackets, Ip::Text::withPort); // implicitly throws exception for IPv6 addresses
os << AsText(ip, Ip::Text::throwOnIpv6, Ip::Text::withPort); // explicitly throws exception for IPv6 addresses
os << AsText(ip, Ip::Text::withBrackets, Ip::Text::withPort); // OK: all required explicit choices were made
```

I think we should use option 2 for safety sake. It is not too ugly[^1].

[^1]: It is _possible_ to implement compile-time guarantees of option 2 while still using backeted(), withPort(), and similar methods from options 1a and 1b (instead of enum constants), but that requires a lot more IoManip code and does not scale well with the number of required formatting choices. I am against that option and am not showing its details here.

You might be able to convince me that option 1b is good enough if it uses "bracketed with port" default.

Which option do you prefer?

P.S. Again, the spelling of all these new names is a separate/secondary issue. Do not worry about it for now!

----

> > 2. Give-me-a-large-enough-buffer API is usually awkward to use, especially in modern C++ code. Its usage also results in a lot of code duplication.

> Er, these are APIs for use in old C-code. They have never been true C++ APIs.

Why "Er"? Seems like we agree that this give-me-a-large-enough-buffer API is a problem for the current Squid code.
 
> > 3. The "force IP version" part of API is a bad idea with worse implementation, resulting in ugly code and bugs, possibly including security vulnerabilities.

> FTR; That is an issue forced on us by: ...

I strongly disagree: None of the specific things you have mentioned force this part of toStr() API and its implementation.

> long-term we should use the `IPvFuture` representation from [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2) whenever possible.

I disagree that IPvFuture format should be used for IPv4 and IPv6 addresses today and in the foreseeable future. I also doubt we should add support for that format today and in the foreseeable future -- we need to fix serious, fundamental problems with the existing IPv4/IPv6 code before throwing a third (and very rarely used) format into the mix.

> > The current PR is arguably worse than simply replacing existing SBuf-needing ip.toStr(...) calls with ToSBuf(ip)!
 
> Falls foul of your issue (1).

Yes, among other problems. I did not say that "simply replacing" is a good solution. We do not need to evaluate it because there is a better one.

kinkie:
> I think we should use option 2 for safety sake. It is not too ugly

I don't like it much, I'd prefer modifier functions like done in AsList. 
If we will go this way, I'd rather use or-ed (|) arguments to avoid having to use varargs (or template varargs).
I'm okay to default to "bracketed with port"; we can also add convenience shortcuts (e.g. "bare()" to signify a combination of flags that is unbracketed, no port)

rousskov:
> I don't like it much, 

_Why_? Do you disagree that it is safer? Or do you think safety (i.e., getting the right text for the given context) is less important in this context than some other (hopefully to be disclosed) considerations?

> I'd prefer modifier functions like done in AsList.

That AsList approach (i.e. optional formatting configuration methods) looks nicer but leads to bugs like the one fixed in 0dc9f3b25fd74362705e5ab6515ae8c34155533d. These bugs would be difficult to detect in some Ip::Address contexts and some of them will have security implications (e.g., wrong port due to port defaults or perhaps IPv6 addresses that will be misinterpreted as IPv4 addresses with ports).

kinkie:
> > I don't like it much,
> 
> _Why_? Do you disagree that it is safer? Or do you think safety (i.e., getting the right text for the given context) is less important in this context than some other (hopefully to be disclosed) considerations?

The most important thing is verbosity; this approach turns a relatively simple operation into an
an unwieldy syntax, and the proposal seems to rely on varargs as well

> > I'd prefer modifier functions like done in AsList.
> 
> That AsList approach (i.e. optional formatting configuration methods) looks nicer but leads to bugs like the one fixed in [0dc9f3b](https://github.com/squid-cache/squid/commit/0dc9f3b25fd74362705e5ab6515ae8c34155533d). These bugs would be difficult to detect in some Ip::Address contexts and some of them will have security implications (e.g., wrong port due to port defaults or perhaps IPv6 addresses that will be misinterpreted as IPv4 addresses with ports).

That is a fair point

rousskov:
> the proposal seems to rely on varargs as well

For the known use cases, the implementation does not need varargs AFAICT: There is a finite (and small!) number of valid combinations of required formatting choices.

src/ip/Address.cc
```diff
@@ -1043,3 +1054,44 @@ Ip::Address::getInAddr(struct in_addr &buf) const
...
+    if ( ip_.isAnyAddr() ) {
+        if (ip_.isIPv6())
+            os << "::";
+        if (ip_.isIPv4())
+            os << "0.0.0.0";
+    } else {
+
+        char buf[INET6_ADDRSTRLEN];
+        if (ip_.isIPv6()) {
+            inet_ntop(AF_INET6, &ip_.mSocketAddr_.sin6_addr, buf, INET6_ADDRSTRLEN);
```
kinkie, line 1078: Should we check for errors here? Legacy code doesn't

src/peer_sourcehash.cc
kinkie: Fair point.
I think we should try to preserve behaviour in a major version,
but I wouldn't feel constrained to do the same across
major versions.

At the same time, I would keep this change as its own PR

src/ip/Address.h
kinkie: It is now gone

src/peer_sourcehash.cc
kinkie: > This use-case does not need any new API. It can just:

Adopted

kinkie:
@rousskov , @yadij , this seems like a nice checkpoint for collecting feedback about the API.

Current state:
- implemented Ip::AddressText . Format can be specified with booleans to the constructor or with functions.
- the text emitting function (Ip::AddressText::print) is a simplified port of Ip::Address::toStr()
- I've implemented unit tests to check for feature parity; please advise if they should cover additional cases

Next steps, if there are no significant objections to the current state,
- reimplement toStr(), toUrl() and toHostStr() on top of AddressText, extending it as needed

Once that is done, I'd start the cleanup, porting callsites to the AddressText API; this might offer the opportunity for deeper changes to move away from the "pass me a large enough buffer" paradigm.

What do you think?

src/peer_sourcehash.cc
rousskov: > I wouldn't feel constrained to do the same across major versions.

Is this hash computation (de facto) standardized across different products (including non-Squid-based caches) or can we assume that this is a Squid-specific concern?

If it is a Squid-specific concern: The "we can change things across versions" logic is sound, but it does not work as well when dealing with distributed cache hierarchies where nodes may run different versions. I would not object to a breaking change like that if others are sure that it is warranted, but in that case, again, we should not covert the IP address to text to compute a hash.

The non-obvious sensitivity of this code to address formatting warrants a C++ comment, especially if we migrate to something implicit like ToSBuf().

> I would keep this change as its own PR

Thank you. I assume "this change" is "possible change of peer hash values".

src/ip/Address.h
```diff
@@ -34,12 +35,41 @@
...
+class AddressText
+{
+public:
+    explicit AddressText(const Address &ip, bool printPort = false, bool printBrackets = true);
+    AddressText &withPort(bool b = false)
+    {
+        printPort_ = b;
+        return *this;
+    };
+    AddressText &bracketed(bool b = true)
```
rousskov, line 49: Please document this and other methods.

In this method description, please disclose that bracketed(true) does not imply the address will be printed with brackets. This behavior is rather counter-intuitive!

In this method description, please disclose that brackets (if any) are going to surround only the IP porting of address.

src/ip/Address.h
```diff
@@ -34,12 +35,41 @@
 namespace Ip
 {
 
+class Address;
+
+class AddressText
+{
+public:
+    explicit AddressText(const Address &ip, bool printPort = false, bool printBrackets = true);
```
rousskov, line 43: Please do not allow configuration with anonymous boolean parameters. Most humans cannot reliably remember what AddressText(ip, false, true) means!

```suggestion
    explicit AddressText(const Address &ip);
```

src/ip/Address.h
```diff
@@ -34,12 +35,41 @@
 namespace Ip
 {
 
+class Address;
+
```
rousskov, line 39: This forward declaration should be provided by ip/forward.h.

src/ip/Address.h
rousskov: > It is now gone

Still here and still used. I cannot provide requested feedback until this is gone.

src/ip/Address.h
```diff
@@ -220,7 +254,7 @@ class Address
      \param len byte length of buffer available for writing.
      \return pointer to buffer received.
      */
-    char* toUrl(char *buf, unsigned int len) const;
+    char *toUrl(char *buf, unsigned int len) const;
```
rousskov, line 257: Out of respect for reviewers, before asking for reviews, please look through PR diff and remove clearly unwanted out-of-scope changes like this one.

rousskov (changes requested):

src/ip/Address.h
kinkie: it is. Removing.

src/ip/Address.h
kinkie: done

src/ip/Address.h
kinkie: Making it explicit that brackets only apply to IPv6 addresses

rousskov (changes requested):

PR #1742, opened 2024-03-18
Add AtMostOnce stream manipulator

kinkie:


src/base/IoManip.h
```diff
@@ -220,5 +220,39 @@ operator <<(std::ostream &os, const AsList<Container> &manipulator)
...
+};
+
+/// a helper to ease AtMostOnce object creation
+template <class T>
+inline auto atMostOnce(const T &t) { return AtMostOnce<T>(t); }
+
+template <class T>
+inline std::ostream& operator<<(std::ostream& os, AtMostOnce<T>& a) {
+    return a.print(os);
+}
```
yadij, line 255: How exactly is the **once** property supposed to be guaranteed when every call to this generates a new temporary instance of `class AtMostOnce` ?

yadij (changes requested):

src/base/IoManip.h
kinkie: Yes it does. See TestIoManip::testAtMostOnce for intended use patterns.
Unless I'm missing something

kinkie:
It works and should be generic, although the way it is implemented places the burden of inner object lifetime on the caller.
There can be two ways out, if that is an undesired behaviour:
1. copy the to-be-printed object
2. does a pattern exist to take ownership of a reference?

src/base/IoManip.h
kinkie: The target pattern is:

```c++
  auto once = atMostOnce("some heading\n");
  ostream << once << "some text\n";
  ostream << once << "some more text" << anObject << '\n';
```

rousskov: > Amos: How exactly is the once property supposed to be guaranteed when every call to this generates a new temporary

> Francesco: Yes it does.

You are both right. When "this" is atMostOnce() function, every call generates a new instance. When "this" is "<<" operator, every call does not generate a new instance. The latter is how this new feature is expected to be used (as illustrated by Francesco).

We need to research whether the helper function is needed in this (and similar) cases. If it is not needed, we should not be adding it here (and in similar cases). I know such helper functions are needed in _some_ cases, but I do not remember what exactly defines that need/those cases. I will research this when I start reviewing this PR, but if you can find the answer, please do!

kinkie: AFAIK these helper functions are used to facilitate template argument deduction, so that
```
auto a1 = atMostOnce("const char *");
and 
std::string s1("std::string");
auto a2 = atMostOnce(s1);
```
work out of the box

kinkie: Actually, after a quick experiment, it turns out to not be needed.
Removing.

rousskov: In general, such helper functions are not needed in some cases and are needed in others. It would be good to know the actual difference instead of adding/removing code based on possibly incomplete tests.

rousskov (changes requested):
This PR needs [research](https://github.com/squid-cache/squid/pull/1742#discussion_r1528564068) and the usual set of many little polishing changes. I will do them when I find the time.

src/base/IoManip.h
kinkie: A [Microsoft devblog article](https://devblogs.microsoft.com/cppblog/how-to-use-class-template-argument-deduction/) can help. 

> Class Template Argument Deduction (CTAD) is a C++17 Core Language feature that reduces code verbosity [...]
> C++98 through C++14 performed template argument deduction for function templates [..]
> However, class templates didn’t benefit from these rules. If you wanted to construct a std::pair from two ints, you had to say std::pair<int, int> p(11, 22);, despite the fact that the compiler already knows that the types of 11 and 22 are int

kinkie:
> This PR needs [research](https://github.com/squid-cache/squid/pull/1742#discussion_r1528564068) 

See https://github.com/squid-cache/squid/pull/1742#discussion_r1529660490

rousskov:
> I still predict the same issue when this manipulator is used like all the other stream manipulators. Specifically using the pattern: `... << AtMostOnce("foo") << ...`

That usage will be a bug, but I do not understand _why_ you are saying the above. We all seem to agree that this PR makes such bugs possible. Are you blocking this PR because AtMostOnce() should not be added at all due to these incorrect usage risks? Are you blocking this PR because there is a way to avoid or reduce incorrect usage that still needs to be implemented? Something else? Please clarify.

kinkie:
@yadij , @rousskov , I don't think there's any open change requests

src/base/IoManip.h
```diff
@@ -220,5 +220,48 @@ operator <<(std::ostream &os, const AsList<Container> &manipulator)
...
+/// Prints AtMostOnce argument if needed. The argument is not constant to
+/// prevent wrong usage:
+///
+/// \code
+/// /* Compiler error: cannot bind non-const lvalue reference to an rvalue */
+/// os << AtMostOnce(x);
+/// \endcode
+template <class T>
+inline auto &
+operator <<(std::ostream &os, AtMostOnce<T> &a) {
```
rousskov, line 261: > > Amos: I still predict the same issue when this manipulator is used like all the other stream manipulators. Specifically using the pattern: `... << AtMostOnce("foo") << ...`
> 
> Alex: That usage will be a bug

Fortunately, it will be a compile-time bug. In other words, code like `<< AtMostOnce("foo") <<` does not compile. FWIW, I was kind of hoping for that outcome, but did not want to claim that until I actually tested PR code. I have now adjusted AtMostOnce documentation to reflect that protection.

src/tests/testIoManip.cc
```diff
@@ -153,6 +155,31 @@ TestIoManip::testAsHex()
...
+        CPPUNIT_ASSERT_EQUAL(std::string("text1"), ss.str());
+        ss << textOnce;
+        ss << textOnce;
+        CPPUNIT_ASSERT_EQUAL(std::string("text1"), ss.str());
+    }
+
+    {
+        std::ostringstream ss;
+        const std::string s("text2");
+        auto textOnce = AtMostOnce(s);
```
rousskov, line 174: If we want to document std::string lifetime considerations, we could add a comment like this:

```suggestion
        // Cannot create std::string when creating textOnce because the string may be
        // destroyed before we are done with textOnce:
        // auto textOnce = AtMostOnce(std::string("do not do this"));
        const std::string s("text2");
        auto textOnce = AtMostOnce(s);
```

I did not add it myself because it feels a bit too obvious and because few developers will notice/read this comment when working with AtMostOnce. I am not against adding it.

rousskov (approved):
@kinkie, I pushed a few small changes. Please adjust further as needed.

src/tests/testIoManip.cc
kinkie: It won't hurt, and I see value in having tests as a source of live documentation if anyone needs it, however trivial

rousskov (approved):

kinkie:
@yadij ping

kinkie:
@yadij any chance you could review PR 1742?

yadij (approved):

PR #1743, opened 2024-03-18
NoNewGlobals for digestFieldsLookupTable

kinkie:
Detected by Coverity. CID 1554677: Initialization or destruction

### DIFF
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -80,13 +80,6 @@ DigestAttrs[] = {
     {nullptr, DIGEST_INVALID_ATTR}
 };
 
-static const auto&
-DigestFieldsLookupTable()
-{
-    static const auto *t = new LookupTable<http_digest_attr_type>(DIGEST_INVALID_ATTR, DigestAttrs);
-    return *t;
-}
-
 /*
  *
  * Nonce Functions
@@ -778,7 +771,8 @@ Auth::Digest::Config::decode(char const *proxy_auth, const HttpRequest *request,
         }
 
         /* find type */
-        const http_digest_attr_type t = DigestFieldsLookupTable().lookup(keyName);
+        static const auto digestFieldsLookupTable = new LookupTable<http_digest_attr_type>(DIGEST_INVALID_ATTR, DigestAttrs);
+        const auto t = digestFieldsLookupTable->lookup(keyName);
 
         switch (t) {
         case DIGEST_USERNAME:
--- a/src/auth/digest/Config.cc
+++ b/src/auth/digest/Config.cc
@@ -80,6 +80,13 @@ DigestAttrs[] = {
     {nullptr, DIGEST_INVALID_ATTR}
 };
 
+static const auto&
+DigestFieldsLookupTable()
+{
+    static const auto table = new LookupTable<http_digest_attr_type>(DIGEST_INVALID_ATTR, DigestAttrs);
+    return *table;
+}
+
 /*
  *
  * Nonce Functions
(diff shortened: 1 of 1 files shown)


src/auth/digest/Config.cc
```diff
@@ -774,6 +771,7 @@ Auth::Digest::Config::decode(char const *proxy_auth, const HttpRequest *request,
         }
 
         /* find type */
+        static LookupTable<http_digest_attr_type> DigestFieldsLookupTable(DIGEST_INVALID_ATTR, DigestAttrs);
```
rousskov, line 774: While unlikely, this local static might be destroyed _before_ the last call to this function in future Squid code. Thus, this change does not fully address the problem identified by Coverity. Consider using the safer pattern from the "No new globals" rule instead.

rousskov (commented):
> Fix Coverity Scan defects cid-14554677

In my test, Coverity Scan cannot find a defect with that CID -- "No issues found". Please double check the ID. If that ID is correct, can you provide (in a GitHub comment) a link to the defect report, please?

src/auth/digest/Config.cc
kinkie: ok

src/auth/digest/Config.cc
```diff
@@ -774,7 +771,8 @@ Auth::Digest::Config::decode(char const *proxy_auth, const HttpRequest *request,
         }
 
         /* find type */
-        const http_digest_attr_type t = DigestFieldsLookupTable.lookup(keyName);
+        static const auto digestFieldsLookupTable = new LookupTable<http_digest_attr_type>(DIGEST_INVALID_ATTR, DigestAttrs);
```
rousskov, line 774: Does not this line still suffer from the same underlying problem, but applied to DigestAttrs global? If it does, then it may be best to create a no-globals style function after all, and move DigestAttrs into that function!

kinkie: I think the DigestAttrs global is safe:
looking at where it lies in the resulting binary (on MacOS)
```
00000001006c05f0 (__DATA_CONST,__const) non-external __ZL11DigestAttrs
```

I found a reference in https://github.com/golang/go/issues/38830 to what it means, and it seems safe enough. The main question is how reliable and portable this is; it might be worthwhile to move everything there for portability reasons

rousskov: > I think the DigestAttrs global is safe

I think it may be safe because LookupTableRecord is currently a constant POD. However, I see no reason not to encapsulate that used-once constant while we are fixing initialization of the global that uses that constant. Your call though.

kinkie: I'll encapsulate it. If anything because I feel it offers a stronger guarantee.

yadij (approved):

rousskov:
> > Fix Coverity Scan defects cid-14554677
> 
> In my test, Coverity Scan cannot find a defect with that CID -- "No issues found". Please double check the ID. If that ID is correct, can you provide (in a GitHub comment) a link to the defect report, please?

@kinkie, any feedback regarding the above concern?

kinkie:
> > > Fix Coverity Scan defects cid-14554677
> > 
> > 
> > In my test, Coverity Scan cannot find a defect with that CID -- "No issues found". Please double check the ID. If that ID is correct, can you provide (in a GitHub comment) a link to the defect report, please?
> 
> @kinkie, any feedback regarding the above concern?

The branch name was corrected, there was a copy-paste error in the description.
Fixed, and clearing for merge

rousskov:
> The branch name was corrected, there was a copy-paste error in the description.

FWIW, I virtually never look at the PR branch name. That name is not important because it does not go into the merged commit. It may also be misleading because development may start with rather different ideas in mind before a better/final solution is found.

Thank you for fixing the PR description!

rousskov:
> Thank you for fixing the PR description!

I adjusted it further to match the format used by similar PRs.

PR #1745, opened 2024-03-18
NoNewGlobals: HttpUpgradeProtocolAccess::ProtoOther

kinkie:
Detected by Coverity. CID 1554674: Initialization or destruction

### DIFF
--- a/src/HttpUpgradeProtocolAccess.cc
+++ b/src/HttpUpgradeProtocolAccess.cc
@@ -49,7 +49,7 @@ HttpUpgradeProtocolAccess::configureGuard(ConfigParser &parser)
 {
     const auto rawProto = parser.NextToken();
     if (!rawProto)
-        throw TextException(ToSBuf("expected a protocol name or ", ProtoOther), Here());
+        throw TextException(ToSBuf("expected a protocol name or ", ProtoOther()), Here());
 
     if (ProtoOther().cmp(rawProto) == 0) {
         aclParseAccessLine(cfg_directive, parser, &other);
--- a/src/HttpUpgradeProtocolAccess.cc
+++ b/src/HttpUpgradeProtocolAccess.cc
@@ -17,7 +17,6 @@
 
 #include <algorithm>
 
-const SBuf HttpUpgradeProtocolAccess::ProtoOther("OTHER");
 
 ProtocolView::ProtocolView(const char * const start, const size_t len):
     ProtocolView(SBuf(start, len))
@@ -53,7 +52,7 @@ HttpUpgradeProtocolAccess::configureGuard(ConfigParser &parser)
     if (!rawProto)
         throw TextException(ToSBuf("expected a protocol name or ", ProtoOther), Here());
 
-    if (ProtoOther.cmp(rawProto) == 0) {
+    if (ProtoOther().cmp(rawProto) == 0) {
         aclParseAccessLine(cfg_directive, parser, &other);
         return;
     }
--- a/src/HttpUpgradeProtocolAccess.cc
+++ b/src/HttpUpgradeProtocolAccess.cc
@@ -17,7 +17,6 @@
 
 #include <algorithm>
 
-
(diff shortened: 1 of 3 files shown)


src/HttpUpgradeProtocolAccess.h
```diff
@@ -82,7 +82,7 @@ class HttpUpgradeProtocolAccess
     typedef std::deque<NamedGuard> NamedGuards;
 
     /// pseudonym to specify rules for "all other protocols"
-    static const SBuf ProtoOther;
+    static const inline SBuf& ProtoOther();
```
rousskov, line 85:
```suggestion
    inline static const SBuf &ProtoOther();
```

src/HttpUpgradeProtocolAccess.h
```diff
@@ -114,7 +114,14 @@ HttpUpgradeProtocolAccess::forApplicable(const ProtocolView &offer, const Visito
         seenApplicable = true; // may already be true
     }
     if (!seenApplicable && other) // OTHER is applicable if named rules were not
-        (void)visitor(ProtoOther, other);
+        (void)visitor(ProtoOther(), other);
+}
+
+const inline SBuf &
```
rousskov, line 120: For consistency with the existing code, please put `inline` first:

```suggestion
inline const SBuf &
```

or

```suggestion
inline const auto &
```

src/HttpUpgradeProtocolAccess.h
```diff
@@ -114,7 +114,14 @@ HttpUpgradeProtocolAccess::forApplicable(const ProtocolView &offer, const Visito
...
+HttpUpgradeProtocolAccess::ProtoOther()
+{
+    const static auto *b = new SBuf("OTHER");
+    return *b;
```
rousskov, line 124: Please follow [guideline style](https://wiki.squid-cache.org/DeveloperResources/SquidCodingGuidelines#rule-no-new-globals), at least for consistency sake:

```suggestion
    static const auto proto = new SBuf("OTHER");
    return *proto;
```

The above is also using an arguably more relevant/natural name for the variable than "b".

src/HttpUpgradeProtocolAccess.cc
```diff
@@ -53,7 +51,7 @@ HttpUpgradeProtocolAccess::configureGuard(ConfigParser &parser)
     if (!rawProto)
         throw TextException(ToSBuf("expected a protocol name or ", ProtoOther), Here());
```
rousskov, line 52: This should fix the build:

```suggestion
        throw TextException(ToSBuf("expected a protocol name or ", ProtoOther()), Here());
```

rousskov (changes requested):
I also polished PR title/description to be more precise in Squid style guideline reference and provide Coverity defect type ID. Please adjust further as needed.

kinkie:
Thanks!

src/HttpUpgradeProtocolAccess.h
```diff
@@ -114,7 +114,14 @@ HttpUpgradeProtocolAccess::forApplicable(const ProtocolView &offer, const Visito
         seenApplicable = true; // may already be true
     }
     if (!seenApplicable && other) // OTHER is applicable if named rules were not
-        (void)visitor(ProtoOther, other);
+        (void)visitor(ProtoOther(), other);
+}
+
+inline const SBuf &
```
rousskov, line 120: We can make this auto as well. We do not have a policy about it (yet?). I do not insist on this change. Your call.

```suggestion
inline const auto &
```

rousskov (approved):
LGTM, with or without the suggested change.

src/HttpUpgradeProtocolAccess.h
kinkie: I thought about it. The declaration needs it, and at that point I'd rather keep a consistent signature in the implementation, even if in theory we could make it auto

yadij (approved):

PR #1746, opened 2024-03-18
NoNewGlobals for MapLabel

kinkie:
Detected by Coverity. CID 1554668: Initialization or destruction

### DIFF
--- a/src/Transients.cc
+++ b/src/Transients.cc
@@ -25,11 +25,11 @@
 #include <limits>
 
 /// shared memory segment path to use for Transients map
-static const auto&
+inline static const auto&
 MapLabel()
 {
-    static const auto* ml = new SBuf("transients_map");
-    return *ml;
+    static const auto proto = new SBuf("transients_map");
+    return *proto;
 }
 
 Transients::Transients(): map(nullptr), locals(nullptr)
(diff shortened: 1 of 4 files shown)


rousskov (changes requested):
When removing #1745 changes from this PR, please apply all relevant #1745 [review suggestions](https://github.com/squid-cache/squid/pull/1745#pullrequestreview-1943615284) to this PR as well.

rousskov (dismissed):
LGTM after a minor polishing touch, thank you.

src/Transients.cc
```diff
@@ -25,7 +25,12 @@
 #include <limits>
 
 /// shared memory segment path to use for Transients map
-static const SBuf MapLabel("transients_map");
+inline static const auto &
```
yadij, line 28: I assume this was done as a side effect of @rousskov request to match/copy PR #1745 review changes.

AIUI, The reasons for using `inline static` pertain specifically to the logic being in a shared header, they become counter-productive when the logic is defined as a `static` function within a unit (`.cc` file).

```suggestion
static const auto &
```

yadij (changes requested):

src/Transients.cc
kinkie: Fine by me; the compiler can inline if it chooses to

PR #1747, opened 2024-03-19
NoNewGlobals for Adaptation::Config::metaHeaders

kinkie:
Detected by Coverity. CID 1554662: Initialization or destruction

### DIFF
--- a/src/adaptation/icap/ModXact.cc
+++ b/src/adaptation/icap/ModXact.cc
@@ -1482,7 +1482,7 @@ void Adaptation::Icap::ModXact::makeRequestHeaders(MemBuf &buf)
         makeUsernameHeader(request, buf);
 
     // Adaptation::Config::metaHeaders
-    for (auto &h: Adaptation::Config::metaHeaders()) {
+    for (const auto &h: Adaptation::Config::metaHeaders()) {
         HttpRequest *r = virgin.cause ?
                          virgin.cause : dynamic_cast<HttpRequest*>(virgin.header);
         Must(r);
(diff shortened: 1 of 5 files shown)


yadij (approved):
Assuming this builds fine.

rousskov (commented):
LGTM overall. I will come back to this to try to improve AAA and const-correctness of the adjusted/move code.

src/adaptation/icap/ModXact.cc
```diff
@@ -1482,7 +1482,7 @@ void Adaptation::Icap::ModXact::makeRequestHeaders(MemBuf &buf)
         makeUsernameHeader(request, buf);
 
     // Adaptation::Config::metaHeaders
-    for (auto h: Adaptation::Config::metaHeaders) {
+    for (const auto h: Adaptation::Config::metaHeaders()) {
```
rousskov, line 1485: I removed a branch-added reference here in commit 33c0fba: It is probably safe to use a reference (to a Note::Pointer) in the current code, but I am a bit worried that this history-updating loop might start modifying Adaptation::Config::metaHeaders() global, invalidating the loop variable `h`. This is a bit paranoid, but this branch is not about optimizing this loop iteration, so let's not take these risks here.

rousskov (approved):
> LGTM overall. I will come back to this to try to improve AAA and const-correctness of the adjusted/move code.

Done.

rousskov (approved):

src/adaptation/icap/ModXact.cc
rousskov: > I removed a branch-added reference here in commit https://github.com/squid-cache/squid/commit/33c0fba498b2b07ce56438793bddbbceb29994e6: ...

I had to revert this change because some compilers [complain](https://github.com/squid-cache/squid/actions/runs/8410743742) about that arguably safer code:

    error: loop variable creates a copy [-Werror=range-loop-construct]

Looks like we cannot easily protect ourselves from Adaptation::Config::metaHeaders() modification. That is not a big deal because such modification is unlikely.

PR #1748, opened 2024-03-19
Make BodyPipe::MaxCapacity constexpr

kinkie:
Detected by Coverity. CID 1554661: Initialization or destruction

### DIFF
--- a/src/adaptation/icap/ModXact.cc
+++ b/src/adaptation/icap/ModXact.cc
@@ -42,8 +42,6 @@
 CBDATA_NAMESPACED_CLASS_INIT(Adaptation::Icap, ModXact);
 CBDATA_NAMESPACED_CLASS_INIT(Adaptation::Icap, ModXactLauncher);
 
-static const size_t TheBackupLimit = BodyPipe::MaxCapacity;
-
 const SBuf Adaptation::Icap::ChunkExtensionValueParser::UseOriginalBodyName("use-original-body");
 
 Adaptation::Icap::ModXact::State::State()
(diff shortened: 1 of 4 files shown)


yadij (dismissed):

yadij:
Did you intend to apply this unrelated change?

kinkie:
> Did you intend to apply this unrelated change?

Obviously not. Too many open branches :( Reverting

src/adaptation/icap/ModXact.cc
```diff
@@ -42,8 +42,6 @@
 CBDATA_NAMESPACED_CLASS_INIT(Adaptation::Icap, ModXact);
 CBDATA_NAMESPACED_CLASS_INIT(Adaptation::Icap, ModXactLauncher);
 
-static const size_t TheBackupLimit = BodyPipe::MaxCapacity;
```
rousskov, line 45: I am against this removal because it obscures (currently self-documented) logic behind TheBackupLimit-using code. It may also complicate migration to smarter/flexible limits if we ever decide to change this logic. Can we do something like this instead?

```C++
constexpr auto TheBackupLimit = BodyPipe::MaxCapacity;
```

AFAIK, `static` is implied for namespace-scope constants, so we do not need it here, but, if you prefer, we can keep it:

```C++
static constexpr auto TheBackupLimit = BodyPipe::MaxCapacity;
```

rousskov (changes requested):

rousskov:
I removed the retelling of the changes from the PR description; I think they are self-documenting. The removal allows us to avoid arguing about the corresponding text details.

rousskov (approved):
I committed an AAA change (that was mentioned in my earlier suggestion but not implemented or discussed). If that change does not work, please undo, of course.

src/adaptation/icap/ModXact.cc
yadij: Hmm, IMO this would be better as a `using`/`typedef` no?

```
using TheBackupLimit = BodyPipe::MaxCapacity;
```

yadij (approved):

PR #1749, opened 2024-03-19
NoNewGlobals for MemBlob::Stats

kinkie:
Detected by Coverity. CID 1554656: Initialization or destruction

Also update MemBlobStats initialization.

### DIFF
--- a/src/sbuf/MemBlob.cc
+++ b/src/sbuf/MemBlob.cc
@@ -14,13 +14,9 @@
 
 #include <iostream>
 
-MemBlobStats MemBlob::Stats;
 InstanceIdDefinitions(MemBlob, "blob");
 
 /* MemBlobStats */
-
-MemBlobStats::MemBlobStats(): alloc(0), live(0), append(0), liveBytes(0)
-{}
 
 MemBlobStats&
 MemBlobStats::operator += (const MemBlobStats& s)
@@ -44,6 +40,19 @@ MemBlobStats::dump(std::ostream &os) const
        "\nlive MemBlob mean current allocation size: " <<
        (static_cast<double>(liveBytes)/(live?live:1)) << std::endl;
     return os;
+}
+
+static auto &
+WriteableStats()
+{
+    static const auto stats = new MemBlobStats();
+    return *stats;
+}
+
+const MemBlobStats &
+MemBlob::GetStats()
+{
+    return WriteableStats();
 }
 
 /* MemBlob */
@@ -72,8 +81,9 @@ MemBlob::~MemBlob()
 {
     if (mem || capacity)
         memFreeBuf(capacity, mem);
-    Stats.liveBytes -= capacity;
-    --Stats.live;
+    auto &stats = WriteableStats();
+    stats.liveBytes -= capacity;
+    --stats.live;
     SBufStats::RecordMemBlobSizeAtDestruct(capacity);
 
     debugs(MEMBLOB_DEBUGSECTION,9, "destructed, this="
@@ -99,9 +109,10 @@ MemBlob::memAlloc(const size_type minSize)
     debugs(MEMBLOB_DEBUGSECTION, 8,
            id << " memAlloc: requested=" << minSize <<
            ", received=" << capacity);
-    ++Stats.live;
-    ++Stats.alloc;
-    Stats.liveBytes += capacity;
+    auto &stats = WriteableStats();
+    ++stats.live;
+    ++stats.alloc;
+    stats.liveBytes += capacity;
 }
 
 void
@@ -109,7 +120,7 @@ MemBlob::appended(const size_type n)
 {
     Must(willFit(n));
     size += n;
-    ++Stats.append;
+    ++WriteableStats().append;
 }
 
 void
@@ -121,7 +132,7 @@ MemBlob::append(const char *source, const size_type n)
         memmove(mem + size, source, n);
         size += n;
     }
-    ++Stats.append;
+    ++WriteableStats().append;
 }
 
 void
@@ -143,12 +154,6 @@ MemBlob::consume(const size_type rawN)
         if (size)
             memmove(mem, mem + n, size);
     }
-}
-
-const MemBlobStats&
-MemBlob::GetStats()
-{
-    return Stats;
 }
 
 std::ostream&
--- a/src/sbuf/MemBlob.h
+++ b/src/sbuf/MemBlob.h
@@ -19,18 +19,16 @@
 class MemBlobStats
 {
 public:
-    MemBlobStats();
-
     /// dumps class-wide statistics
     std::ostream& dump(std::ostream& os) const;
 
     MemBlobStats& operator += (const MemBlobStats&);
 
 public:
-    uint64_t alloc;     ///< number of MemBlob instances created so far
-    uint64_t live;      ///< number of MemBlob instances currently alive
-    uint64_t append;    ///< number of MemBlob::append() calls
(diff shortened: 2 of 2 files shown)


src/sbuf/MemBlob.h
```diff
@@ -130,5 +128,12 @@ class MemBlob: public RefCountable
     MemBlob& operator =(const MemBlob &);
 };
 
+inline MemBlobStats&
+MemBlob::Stats()
+{
+    static auto stats = new MemBlobStats();
+    return *stats;
+}
+
```
yadij, line 137: Please move this to the .cc file instead of `inline`'ing.

yadij (changes requested):

src/sbuf/MemBlob.h
```diff
@@ -130,5 +128,6 @@ class MemBlob: public RefCountable
     MemBlob& operator =(const MemBlob &);
 };
 
+
```
yadij, line 131:
```suggestion
```

yadij (changes requested):

src/sbuf/MemBlob.cc
```diff
@@ -72,8 +75,9 @@ MemBlob::~MemBlob()
 {
     if (mem || capacity)
         memFreeBuf(capacity, mem);
-    Stats.liveBytes -= capacity;
-    --Stats.live;
+    auto stats = GetStatsInstance();
```
rousskov, line 78: This and similar PR code is buggy because it copies MemBlobStats -- among other "decorations", `auto` does not automatically relay references. We have to explicitly ask for a reference if we want a reference instead of a regular object (unlike, say, pointers, references are not C++ objects).

I am a bit surprised that the compiler cannot detect the following "pointless" `stats` update!

src/sbuf/MemBlob.cc
```diff
@@ -46,6 +42,13 @@ MemBlobStats::dump(std::ostream &os) const
     return os;
 }
 
+static auto &
+GetStatsInstance()
```
rousskov, line 46: I would call this function ModifiableStats() to better explain why we have two functions returning a reference to the same object. I do not insist on any particular spelling, but I do think we should rename.

src/sbuf/MemBlob.cc
```diff
@@ -148,7 +153,7 @@ MemBlob::consume(const size_type rawN)
 const MemBlobStats&
 MemBlob::GetStats()
 {
-    return Stats;
+    return GetStatsInstance();
 }
```
rousskov, line 157: Please either move GetStats() definition up or (forward-declare and) move GetStatsInstance() definition down so that the two "give me the stats" functions are defined _next to each other_. I recommend the former because moving GetStats() definition results in more compact code (at the expense of a few additional line modifications).

src/sbuf/MemBlob.cc
```diff
@@ -46,6 +42,13 @@ MemBlobStats::dump(std::ostream &os) const
     return os;
 }
 
+static auto &
+GetStatsInstance()
+{
+    static auto stats = new MemBlobStats();
```
rousskov, line 48: For consistency sake and to improve const-correctness, please add `const` that NoNewGlobals sketch contains:

```suggestion
    static const auto stats = new MemBlobStats();
```

rousskov (changes requested):

src/sbuf/MemBlob.cc
kinkie: How about RWStats?

src/sbuf/MemBlob.cc
kinkie: ok

src/sbuf/MemBlob.cc
kinkie: But this is not const, it's a read-write reference. GetStats returns the const version. Are you sure?

src/sbuf/MemBlob.cc
kinkie: Fixed. Doh. :(

src/sbuf/MemBlob.cc
rousskov: > How about RWStats?

I do not like the CAPS effect of that name and the abbreviations that some code readers may find difficult to decipher. I can live with WriteableStats() if you do not like ModifiableStats().

src/sbuf/MemBlob.cc
rousskov:
```C++
static const auto stats = new MemBlobStats();
```

> But this is not const, it's a read-write reference. Are you sure?

The `stats` object is constant. The value of that pointer (i.e. the address of the allocated MemBlobStats object) never changes. That is what NoNewGlobals pattern is declaring.

The `*stats` object (i.e., the allocated MemBlobStats object) is not constant. NoNewGlobals is not marking it as such.

If we wanted to mark the allocated object as constant (in addition to marking the pointer as constant), we would do something like this instead (but then the function would also need to return a const reference):

```C++
const auto &
Foo() {
    static const auto * const foo = new Type();
    return *foo;
}
```

I have designed NoNewGlobals pattern to correctly handle typical cases (like GetStatsInstance()). If you have questions about that pattern, I am happy to collaborate to find answers. Meanwhile, please use that exact pattern where the compiler allows you to use it (instead of _starting_ with other patterns and leaving them in if they happen to compile). Fixing the same mistakes over and over again is costly.

Moreover, please start with "const auto foo = ..." in all contexts that do not require a reference. In most cases, that is the right pattern. In most cases, the compiler will tell you when that pattern is not applicable. Adjust as needed to remove `const`, add reference, etc., but _start_ with that simplest (and usually correct) pattern.

rousskov:
@kinkie, this PR currently contains unrelated changes from other PRs/branches.

src/sbuf/MemBlob.cc
kinkie: WriteableStats it is then

kinkie:
> @kinkie, this PR currently contains unrelated changes from other PRs/branches.

Fixed, github got confused with some merge. I squashed commits

kinkie:
Run-tested; excerpt from http://localhost:3128/squid-internal-mgr/sbuf :

MemBlob created: 2115
MemBlob alive: 1297
MemBlob append calls: 3020
MemBlob currently allocated size: 62688

src/sbuf/MemBlob.cc
kinkie: Thanks

rousskov (approved):

kinkie:
@yadij any chance you could review this PR?

rousskov (commented):
> rousskov approved these changes on Mar 22, 2024

> kinkie merged branch 'master' into coverity-cid-1554656

> kinkie kinkie requested a review from rousskov on April 5, 2024 06:19

> rousskov removed their request for review on April 5, 2024 12:38

@kinkie, please do not request my re-review after uneventful/clean mergers of master branch[^1] unless there is something specific you want me to check or respond to. In the latter case, please mention what that is. Otherwise, I am facing a PR that, on the surface, should be the same as the one I have reviewed already, but with a new pending review request; I do not want to re-review everything from scratch unless there is a compelling reason to do that. This recommendation applies to the PRs I have approved (like this one), but it is not specific to such PRs.

[^1]: And merging master into feature branch should only be performed if necessary from PR code/review point of view. For example, these merges should not be abused for triggering or restarting CI tests. This footnote does not imply that an unwanted merger has happened in this PR.

yadij (approved):

PR #1750, opened 2024-03-19
NoNewGlobals for HttpHdrCc:ccLookupTable

kinkie:
Detected by Coverity. CID 1554655: Initialization or destruction

Also switched to compile-time checks for table initialization records.

### DIFF
--- a/src/HttpHdrCc.cc
+++ b/src/HttpHdrCc.cc
@@ -56,11 +56,12 @@ CcAttrs() {
     return attrsList;
 }
 
-static const auto&
-ccLookupTable() {
+static auto
+ccTypeByName(const SBuf &name) {
     const static auto table = new LookupTable<HttpHdrCcType>(HttpHdrCcType::CC_OTHER, CcAttrs());
-    return *table;
+    return table->lookup(name);
 }
+
 std::vector<HttpHeaderFieldStat> ccHeaderStats(HttpHdrCcType::CC_ENUM_END);
 
 /// used to walk a table of http_header_cc_type structs
@@ -119,7 +120,7 @@ HttpHdrCc::parse(const String & str)
         }
 
         /* find type */
-        const HttpHdrCcType type = ccLookupTable().lookup(SBuf(item,nlen));
+        const HttpHdrCcType type = ccTypeByName(SBuf(item, nlen));
 
         // ignore known duplicate directives
         if (isSet(type)) {
--- a/src/HttpHdrCc.cc
+++ b/src/HttpHdrCc.cc
@@ -9,6 +9,7 @@
 /* DEBUG: section 65    HTTP Cache Control Header */
 
 #include "squid.h"
+#include "base/EnumIterator.h"
 #include "base/LookupTable.h"
 #include "HttpHdrCc.h"
 #include "HttpHeader.h"
@@ -25,34 +26,35 @@
 #include <vector>
 #include <ostream>
 
-static const auto&
+constexpr LookupTable<HttpHdrCcType>::Record attrsList[] = {
+    {"public", HttpHdrCcType::CC_PUBLIC},
+    {"private", HttpHdrCcType::CC_PRIVATE},
+    {"no-cache", HttpHdrCcType::CC_NO_CACHE},
+    {"no-store", HttpHdrCcType::CC_NO_STORE},
+    {"no-transform", HttpHdrCcType::CC_NO_TRANSFORM},
+    {"must-revalidate", HttpHdrCcType::CC_MUST_REVALIDATE},
+    {"proxy-revalidate", HttpHdrCcType::CC_PROXY_REVALIDATE},
+    {"max-age", HttpHdrCcType::CC_MAX_AGE},
+    {"s-maxage", HttpHdrCcType::CC_S_MAXAGE},
+    {"max-stale", HttpHdrCcType::CC_MAX_STALE},
+    {"min-fresh", HttpHdrCcType::CC_MIN_FRESH},
+    {"only-if-cached", HttpHdrCcType::CC_ONLY_IF_CACHED},
+    {"stale-if-error", HttpHdrCcType::CC_STALE_IF_ERROR},
+    {"immutable", HttpHdrCcType::CC_IMMUTABLE},
+    {"Other,", HttpHdrCcType::CC_OTHER}, /* ',' will protect from matches */
+    {nullptr, HttpHdrCcType::CC_ENUM_END}
+};
+
+constexpr const auto &
 CcAttrs() {
-    // invariant: row[j].id == j
-    static LookupTable<HttpHdrCcType>::Record attrsList[] = {
-        {"public", HttpHdrCcType::CC_PUBLIC},
-        {"private", HttpHdrCcType::CC_PRIVATE},
-        {"no-cache", HttpHdrCcType::CC_NO_CACHE},
-        {"no-store", HttpHdrCcType::CC_NO_STORE},
-        {"no-transform", HttpHdrCcType::CC_NO_TRANSFORM},
-        {"must-revalidate", HttpHdrCcType::CC_MUST_REVALIDATE},
-        {"proxy-revalidate", HttpHdrCcType::CC_PROXY_REVALIDATE},
-        {"max-age", HttpHdrCcType::CC_MAX_AGE},
-        {"s-maxage", HttpHdrCcType::CC_S_MAXAGE},
-        {"max-stale", HttpHdrCcType::CC_MAX_STALE},
-        {"min-fresh", HttpHdrCcType::CC_MIN_FRESH},
-        {"only-if-cached", HttpHdrCcType::CC_ONLY_IF_CACHED},
-        {"stale-if-error", HttpHdrCcType::CC_STALE_IF_ERROR},
-        {"immutable", HttpHdrCcType::CC_IMMUTABLE},
-        {"Other,", HttpHdrCcType::CC_OTHER}, /* ',' will protect from matches */
-        {nullptr, HttpHdrCcType::CC_ENUM_END}
-    };
-    static auto invariantChecked = false;
-    if (!invariantChecked) {
-        for (unsigned int j = 0; attrsList[j].name != nullptr; ++j) {
-            assert(static_cast<decltype(j)>(attrsList[j].id) == j);
-        }
-        invariantChecked = true;
-    }
+    // TODO: Move these compile-time checks into LookupTable
+    ConstexprForEnum<HttpHdrCcType::CC_PUBLIC, HttpHdrCcType::CC_ENUM_END>([](const auto ev) {
+        const auto idx = static_cast<std::underlying_type<HttpHdrCcType>::type>(ev);
+        // invariant: each row has a name except the last one
+        static_assert(!attrsList[idx].name == (ev == HttpHdrCcType::CC_ENUM_END));
+        // invariant: row[ev].id == ev
+        static_assert(attrsList[idx].id == ev);
+    });
     return attrsList;
 }
 
--- a/src/HttpHdrCc.cc
+++ b/src/HttpHdrCc.cc
@@ -366,7 +366,7 @@ httpHdrCcStatDumper(StoreEntry * sentry, int, double val, double, int count)
 std::ostream &
 operator<< (std::ostream &s, HttpHdrCcType c)
 {
-    s << ccNameByType(c).value_or("*invalid hdrcc*") << '[' << static_cast<int>(c) << ']';
+    s << ccNameByType(c).value_or("INVALID") << '[' << static_cast<int>(c) << ']';
     return s;
 }
 
--- a/src/HttpHdrCc.cc
+++ b/src/HttpHdrCc.cc
@@ -62,6 +62,13 @@ ccTypeByName(const SBuf &name) {
     return table->lookup(name);
 }
 
+static auto
+ccNameByType(HttpHdrCcType type) {
+    if (type < HttpHdrCcType::CC_ENUM_END)
+        return CcAttrs()[static_cast<int>(type)].name;
+    return "*invalid hdrcc*";
+}
+
 std::vector<HttpHeaderFieldStat> ccHeaderStats(HttpHdrCcType::CC_ENUM_END);
 
 /// used to walk a table of http_header_cc_type structs
@@ -350,11 +357,7 @@ httpHdrCcStatDumper(StoreEntry * sentry, int, double val, double, int count)
 std::ostream &
 operator<< (std::ostream &s, HttpHdrCcType c)
 {
-    const unsigned char ic = static_cast<int>(c);
-    if (c < HttpHdrCcType::CC_ENUM_END)
-        s << CcAttrs()[ic].name << '[' << ic << ']' ;
-    else
-        s << "*invalid hdrcc* [" << ic << ']';
+    s << ccNameByType(c) << '[' << static_cast<int>(c) << ']';
     return s;
 }
 
--- a/src/HttpHdrCc.cc
+++ b/src/HttpHdrCc.cc
@@ -272,7 +272,7 @@ HttpHdrCc::packInto(Packable * p) const
         if (isSet(flag) && flag != HttpHdrCcType::CC_OTHER) {
 
             /* print option name for all options */
-            p->appendf((pcount ? ", %s": "%s"), CcAttrs()[flag].name);
+            p->appendf((pcount ? ", %s": "%s"), ccNameByType(flag));
 
             /* for all options having values, "=value" after the name */
             switch (flag) {
@@ -345,9 +345,9 @@ void
 httpHdrCcStatDumper(StoreEntry * sentry, int, double val, double, int count)
 {
     extern const HttpHeaderStat *dump_stat; /* argh! */
-    const int id = static_cast<int>(val);
-    const bool valid_id = id >= 0 && id < static_cast<int>(HttpHdrCcType::CC_ENUM_END);
-    const char *name = valid_id ? CcAttrs()[id].name : "INVALID";
+    const auto id = static_cast<HttpHdrCcType>(val);
+    const bool valid_id = id >= HttpHdrCcType::CC_PUBLIC && id < HttpHdrCcType::CC_ENUM_END;
+    auto name = valid_id ? ccNameByType(id) : "INVALID";
 
     if (count || valid_id)
         storeAppendPrintf(sentry, "%2d\t %-20s\t %5d\t %6.2f\n",
--- a/src/HttpHdrCc.cc
+++ b/src/HttpHdrCc.cc
@@ -17,6 +17,7 @@
 #include "HttpHeaderStat.h"
 #include "HttpHeaderTools.h"
 #include "sbuf/SBuf.h"
+#include "SquidMath.h"
 #include "StatHist.h"
 #include "Store.h"
 #include "StrList.h"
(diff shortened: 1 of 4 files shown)


kinkie:
As an implementation note: the CcAttrs table needs to be accessible,
as there are several lookup operations done from ID to SBuf.

There is an [abandoned branch](https://github.com/kinkie/squid/tree/coverity-cid-1554655-abandoned) where I tried to do a number of extra things, which might make sense to extract and do their own PR on:
1. add a const_iterator to LookupTable (done)
2. make LookupTable bidirectional (I've implemented a scan-based reverse_lookup function, but it would make sense to use a std::vector one instead)
3. make HttpHdrCc::packInto based on PackableStream

Do you think it's worth investing any time in these efforts?

rousskov (changes requested):
> Do you think it's worth investing any time in [the following] efforts?

> 1. add a const_iterator to LookupTable (done)

The table is not ordered by directive name or ID. Creating a useful iterator for such a table would not be trivial. Do it only if there are multiple uses for such an iterator.

FWIW, the current [attempt](https://github.com/squid-cache/squid/blob/4ee0668f90b24c43e0718612abac76b0e9f59b26/src/HttpHdrCc.cc#L352) to replace trivial `CcAttrs[id].name` with `ccLookupTable().reverse_lookup(static_cast<HttpHdrCcType>(id))` seems like a step in the wrong direction to me.

> 2. make LookupTable bidirectional

The current hash table is not "directional" at all, so I am not sure what you mean by bidirectional. What use cases are you targeting?

> 3. make HttpHdrCc::packInto based on PackableStream

Converting HttpHdrCc::packInto() to use std::ostream instead of Packable would be a useful step towards similar packInto() upgrades for Http::Message and friends.

src/HttpHdrCc.cc
```diff
@@ -345,7 +355,7 @@ operator<< (std::ostream &s, HttpHdrCcType c)
 {
     const unsigned char ic = static_cast<int>(c);
     if (c < HttpHdrCcType::CC_ENUM_END)
-        s << CcAttrs[ic].name << '[' << ic << ']' ;
+        s << CcAttrs()[ic].name << '[' << ic << ']' ;
```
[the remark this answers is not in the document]
kinkie: Good idea, implemented

kinkie:
It's hopefully ready for review now

src/HttpHdrCc.cc
```diff
@@ -9,24 +9,26 @@
...
 #include "util.h"
 
 #include <map>
 #include <vector>
+#include <optional>
 #include <ostream>
 
-// invariant: row[j].id == j
-static LookupTable<HttpHdrCcType>::Record CcAttrs[] = {
+constexpr LookupTable<HttpHdrCcType>::Record attrsList[] = {
```
rousskov, line 31: AFAICT, this compile-time constant is no longer a "global" in "No New Globals" sense. It cannot be localized inside CcAttrs() AFAICT because the compiler does not like us returning a reference to a local constexpr. Making this constexpr allows us to check the invariant at compile time, which is a significant improvement. Long-term, I suspect that this initialization-only code will be significantly refactored to avoid nil names and CC_ENUM_END hacks. However, if constexpr is not enough to make Coverity or others happy, we can go back to local statics and runtime assertions. I do not insist on trying constexpr here.

rousskov (approved):

src/HttpHdrCc.cc
```diff
@@ -60,9 +69,10 @@ operator++ (HttpHdrCcType &aHeader)
 void
 httpHdrCcInitModule(void)
 {
+    const auto attrs = CcAttrs();
     // check invariant on initialization table
-    for (unsigned int j = 0; CcAttrs[j].name != nullptr; ++j) {
-        assert(static_cast<decltype(j)>(CcAttrs[j].id) == j);
+    for (unsigned int j = 0; attrs[j].name != nullptr; ++j) {
+        assert(static_cast<decltype(j)>(attrs[j].id) == j);
```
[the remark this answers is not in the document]
kinkie: Above my skill level. Thanks for doing it

src/HttpHdrCc.cc
kinkie: I agree to try it. It might become a useful pattern, as you say.
I'm not sure we should strive to make Coverity happy; either we are solving real (if small) problems or we should just tell Coverity to be quiet

src/HttpHdrCc.cc
```diff
@@ -44,7 +46,40 @@ static LookupTable<HttpHdrCcType>::Record CcAttrs[] = {
...
+    return table->lookup(name);
+}
+
+/// Safely converts an integer into a Cache-Control directive name.
+/// \returns std::nullopt if the given integer is not a valid index of a named attrsList entry
+template <typename RawId>
+static std::optional<const char *>
+ccNameByType(const RawId rawId)
+{
+    // TODO: Store a by-ID index in (and move this functionality into) LookupTable.
```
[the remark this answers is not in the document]
kinkie: I agree.
This was exactly what I meant with my attempt for a reverse_lookup functionality

src/HttpHdrCc.cc
```diff
@@ -332,22 +357,16 @@ httpHdrCcStatDumper(StoreEntry * sentry, int, double val, double, int count)
...
 
 std::ostream &
 operator<< (std::ostream &s, HttpHdrCcType c)
 {
-    const unsigned char ic = static_cast<int>(c);
-    if (c < HttpHdrCcType::CC_ENUM_END)
-        s << CcAttrs[ic].name << '[' << ic << ']' ;
-    else
-        s << "*invalid hdrcc* [" << ic << ']';
+    s << ccNameByType(c).value_or("*invalid hdrcc*") << '[' << static_cast<int>(c) << ']';
```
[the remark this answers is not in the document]
kinkie: I like INVALID.

src/base/EnumIterator.h
```diff
@@ -224,5 +224,20 @@ class WholeEnum : public EnumRangeT<EnumType>
...
+constexpr void
+ConstexprForEnum(F &&f)
+{
+    using enumType = decltype(First);
+    using underlyingType = typename std::underlying_type<enumType>::type;
+    // std::integral_constant trick makes f(First) argument a constexpr
+    f(std::integral_constant<enumType, First>()); // including when First is Last
+    if constexpr (First < Last)
+        ConstexprForEnum<enumType(static_cast<underlyingType>(First) + 1), Last>(f);
+}
```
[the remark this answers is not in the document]
kinkie: Nono, this is brilliant and TIL

src/HttpHdrCc.cc
```diff
@@ -332,22 +357,16 @@ httpHdrCcStatDumper(StoreEntry * sentry, int, double val, double, int count)
...
     const int id = static_cast<int>(val);
-    const bool valid_id = id >= 0 && id < static_cast<int>(HttpHdrCcType::CC_ENUM_END);
-    const char *name = valid_id ? CcAttrs[id].name : "INVALID";
-
-    if (count || valid_id)
+    const auto name = ccNameByType(id);
+    if (count || name)
         storeAppendPrintf(sentry, "%2d\t %-20s\t %5d\t %6.2f\n",
-                          id, name, count, xdiv(count, dump_stat->ccParsedCount));
+                          id, name.value_or("INVALID"), count, xdiv(count, dump_stat->ccParsedCount));
```
kinkie, line 363: TIL about `value_or` as well

src/HttpHdrCc.cc
yadij: Returning a reference to a variable which exists only in compiler memory makes no sense. But leaving it a `static` member like all our other NoNewGlobals functions have should be fine to return and also more compatible with future conversion of the array to storing `SBuf`

src/HttpHdrCc.cc
yadij: ?

src/base/EnumIterator.h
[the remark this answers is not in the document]
yadij: Please be careful with compile-time recursive. They have a tendency towards compile failure/crash on resource limited small machines/VM/containers.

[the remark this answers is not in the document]
rousskov: > Please be careful with compile-time recursive.

I assume this advice is not meant to trigger some specific action in this PR.

> They have a tendency towards compile failure/crash on resource limited small machines/VM/containers.

FWIW, I am aware of the risks and do not insist on this addition (because of the risks and other complications). I hope that the current use case does not run into recursion limitations due to a small number of enum values (~16), but perhaps I am too optimistic. Still, I (obviously) think it is worth trying.

src/HttpHdrCc.cc
rousskov: AFAICT, Francesco is simply acknowledging/celebrating that he has [learned something](https://www.howtogeek.com/702929/what-does-til-mean-and-how-do-you-use-it/). The first usage of value_or() in Squid code goes back to 2021, but things like that are easy to overlook in PR noise. I am glad this code cut through that noise!

src/HttpHdrCc.cc
rousskov: TLDR: @yadij, if you think that the current PR code does not resolve these and other tradeoffs well and should go back to runtime checks, please say so explicitly, and I will adjust the code accordingly. I do not think we should merge this PR unless there is consensus regarding these compile-time checks. If you do not think these compile-time checks are worth their associated risks and limitations, we should not merge.

----

> Returning a reference to a variable which exists only in compiler memory makes no sense.

Why not?! IMHO, it may make sense to return a reference to a variable used by the compiler.

Moreover, [constexpr](https://en.cppreference.com/w/cpp/language/constexpr) does not imply that the variable exists only in compiler memory: The specifier means that the _value_ of that variable can be computed at compile time. The variable existence duration is governed by other considerations. For example, in the current PR code, attrsList exists at runtime in my tests:

```
$ nm src/squid | c++filt | grep attrsList
0000000000b241a0 d attrsList
```

```
(gdb) p attrsList
$1 = {{name = 0x555555e990a8 "public", id = CC_PUBLIC}, ...}
```

If we were to add `extern ... attrList[]` declaration, then attrList would become visible in other translation units as well (while still being constexpr).

> leaving it a `static` member ... should be fine to return 

Yes, but, as I tried to [explain](https://github.com/squid-cache/squid/pull/1750/files#r1532870287), doing so will kill compile-time checks because static variables are prohibited inside constexpr functions (until C++23):

```
HttpHdrCc.cc:34:53: error: ‘attrsList’ declared ‘static’ in ‘constexpr’ function
```

> and also more compatible with future conversion of the array to storing `SBuf`

I am not sure we would want to convert this particular array, but if we do start storing SBuf in it, then we would not be able to check its invariants at compile time.

src/HttpHdrCc.cc
```diff
@@ -44,7 +46,40 @@ static LookupTable<HttpHdrCcType>::Record CcAttrs[] = {
...
+
+constexpr const auto &
+CcAttrs() {
+    // TODO: Move these compile-time checks into LookupTable
+    ConstexprForEnum<HttpHdrCcType::CC_PUBLIC, HttpHdrCcType::CC_ENUM_END>([](const auto ev) {
+        const auto idx = static_cast<std::underlying_type<HttpHdrCcType>::type>(ev);
+        // invariant: each row has a name except the last one
+        static_assert(!attrsList[idx].name == (ev == HttpHdrCcType::CC_ENUM_END));
+        // invariant: row[idx].id == idx
+        static_assert(attrsList[idx].id == ev);
```
rousskov, line 58: I wanted to share how this static_assertion failure looks like (after injecting an incorrect attrsList entry):

```
HttpHdrCc.cc: In instantiation of ‘CcAttrs()::<lambda(auto:1)> [with auto:1 = std::integral_constant<HttpHdrCcType, CC_PUBLIC>]’:
HttpHdrCc.cc:62:39: error: static assertion failed
   62 |         static_assert(attrsList[j].id == j);
```

IMO, this is fairly readable and more informative than a runtime assertion would be (without using a debugger).

I wish I had tested the error message after injecting a bad value at the _end_ of the table!

kinkie: I did. Entries past HttpHdrCcType::CC_ENUM_END are ignored

rousskov: I meant a bad entry _before_ the end marker. I am worried about error message readability due to deep recursion...

We can add more static assertions to check for the presence of "extra" entries after the end marker as well, but, unlike duplicate or out-of-order entries before the end marker, such bugs would be pretty visible in the diff, so I am not particularly worried about them.

kinkie: On MacOS:
```patch
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:237:5: note: in instantiation of function template specialization 'CcAttrs()::(anonymous class)::operator()<std::integral_constant<HttpHdrCcType, CC_ENUM_END>>' requestediff --git a/src/HttpHdrCc.cc b/src/HttpHdrCc.cc
index 2bd9303208..179be7de24 100644
--- a/src/HttpHdrCc.cc
+++ b/src/HttpHdrCc.cc
@@ -44,8 +44,8 @@ constexpr LookupTable<HttpHdrCcType>::Record attrsList[] = {
     {"stale-if-error", HttpHdrCcType::CC_STALE_IF_ERROR},
     {"immutable", HttpHdrCcType::CC_IMMUTABLE},
     {"Other,", HttpHdrCcType::CC_OTHER}, /* ',' will protect from matches */
-    {nullptr, HttpHdrCcType::CC_ENUM_END}
-};
+    {nullptr, HttpHdrCcType::CC_OTHER},
+    {nullptr, HttpHdrCcType::CC_ENUM_END}};

 constexpr const auto &
 CcAttrs() {
```
results in
```
/Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:58:9: error: static assertion failed due to requirement 'attrsList[idx].id == ev'
        static_assert(attrsList[idx].id == ev);
        ^             ~~~~~~~~~~~~~~~~~~~~~~~
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:237:5: note: in instantiation of function template specialization 'CcAttrs()::(anonymous class)::operator()<std::integral_constant<HttpHdrCcType, CC_ENUM_END>>' requested here
    f(std::integral_constant<enumType, First>()); // including when First is Last
    ^
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:239:9: note: in instantiation of function template specialization 'ConstexprForEnum<CC_ENUM_END, CC_ENUM_END, (lambda at /Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:76) &>' requested here
        ConstexprForEnum<enumType(static_cast<underlyingType>(First) + 1), Last>(f);
        ^
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:239:9: note: in instantiation of function template specialization 'ConstexprForEnum<CC_OTHER, CC_ENUM_END, (lambda at /Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:76) &>' requested here
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:239:9: note: in instantiation of function template specialization 'ConstexprForEnum<CC_IMMUTABLE, CC_ENUM_END, (lambda at /Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:76) &>' requested here
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:239:9: note: in instantiation of function template specialization 'ConstexprForEnum<CC_STALE_IF_ERROR, CC_ENUM_END, (lambda at /Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:76) &>' requested here
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:239:9: note: (skipping 7 contexts in backtrace; use -ftemplate-backtrace-limit=0 to see all)
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:239:9: note: in instantiation of function template specialization 'ConstexprForEnum<CC_NO_TRANSFORM, CC_ENUM_END, (lambda at /Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:76) &>' requested here
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:239:9: note: in instantiation of function template specialization 'ConstexprForEnum<CC_NO_STORE, CC_ENUM_END, (lambda at /Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:76) &>' requested here
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:239:9: note: in instantiation of function template specialization 'ConstexprForEnum<CC_NO_CACHE, CC_ENUM_END, (lambda at /Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:76) &>' requested here
/Users/kinkie/src/squid/coverity-cid-1554655/src/base/EnumIterator.h:239:9: note: in instantiation of function template specialization 'ConstexprForEnum<CC_PRIVATE, CC_ENUM_END, (lambda at /Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:76) &>' requested here
/Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:5: note: in instantiation of function template specialization 'ConstexprForEnum<CC_PUBLIC, CC_ENUM_END, (lambda at /Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:53:76)>' requested here
    ConstexprForEnum<HttpHdrCcType::CC_PUBLIC, HttpHdrCcType::CC_ENUM_END>([](const auto ev) {
    ^
/Users/kinkie/src/squid/coverity-cid-1554655/src/HttpHdrCc.cc:58:41: note: expression evaluates to '14 == 15'
        static_assert(attrsList[idx].id == ev);
                      ~~~~~~~~~~~~~~~~~~^~~~~
1 error generated.
```

rousskov: That is exactly the test I was thinking about, thank you!

As I feared, it is not trivial to discover the enum value that violated the static assertion. The `14 == 15` error detail is probably enough to (eventually) figure out what is going. IMHO, this is not much worse than running a debugger after a runtime assertion (or even adding more code to report values that would lead to the assertion).

Overall, I still think that this static assertion experiment is worth merging, but I certainly do not insist on merging it.

src/base/EnumIterator.h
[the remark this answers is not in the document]
yadij: I just intended a warning to @kinkie not to get too attached to the usage.

src/HttpHdrCc.cc
yadij: I was wondering what "TIL" stood for, have not seen that one before.

src/HttpHdrCc.cc
yadij: Do you know whether ASAN correctly detects the limits of a loop scanning across the constexpr when only that virtual-reference is provided (ie before the expression is calculated?).

yadij (commented):
I objections from me, though I have enough doubts about the constexpr not to sign off explicitly.

kinkie:
> I objections from me, though I have enough doubts about the constexpr not to sign off explicitly.

Fair enough. This is a stepping stone towards improving LookupTable so that callers will be relieved of some of the invariant checks duty.

kinkie:
Merging on slow burner

src/HttpHdrCc.cc
rousskov: > Do you know whether ASAN correctly detects the limits of a loop scanning across the constexpr when only that virtual-reference is provided (ie before the expression is calculated?).

I assume that by "ASAN" you mean "address sanitizer" (more commonly abbreviated as ASan AFAICT).

I do not know much about address sanitizers, but I would expect their limitations to depend on the address sanitizer implementation. I can speculate that if a sanitizer is reusing initial processing stages from the "compiler", then it may not even have to deal with constexpr computations -- it would rely on their computed values received from those earlier processing stages.

I also cannot guess what "virtual-reference" means in this context.

PR #1752, opened 2024-03-22
HTTP: Protect just-parsed responses from accidental destruction

eduard-bagdasaryan:
The lack of HttpReply locking meant that we could not (safely) pass the
freshly created reply object to ACLFilledChecklist because the checklist
destructor would unwittingly destroy the object (by unlocking it). There
is at least one applicable ACL check in handle1xx(), but we got lucky
because that method adds the reply object to ALE, increasing its
reference counter. This change stops relying on such "external" locks.
    
This change also protects from HttpReply memory leaks when an exception
is thrown between HttpReply creation and a setVirginReply() call.
    
TODO: Fix all cases where newly created HttpReply objects are remembered
using raw pointers (but may be passed to locking/unlocking code).

### DIFF
--- a/src/http.cc
+++ b/src/http.cc
@@ -745,12 +745,10 @@ HttpStateData::processReplyHeader()
 
 /// ignore or start forwarding the 1xx response (a.k.a., control message)
 void
-HttpStateData::handle1xx(HttpReply *reply)
+HttpStateData::handle1xx(const HttpReply::Pointer &reply)
 {
     if (fwd->al)
         fwd->al->reply = reply;
-
-    HttpReply::Pointer msg(reply); // will destroy reply if unused
 
     // one 1xx at a time: we must not be called while waiting for previous 1xx
     Must(!flags.handling1xx);
@@ -774,7 +772,7 @@ HttpStateData::handle1xx(HttpReply *reply)
     if (Config.accessList.reply) {
         ACLFilledChecklist ch(Config.accessList.reply, originalRequest().getRaw());
         ch.al = fwd->al;
-        ch.reply = reply;
+        ch.reply = reply.getRaw();
         ch.syncAle(originalRequest().getRaw(), nullptr);
         HTTPMSGLOCK(ch.reply);
         if (!ch.fastCheck().allowed()) // TODO: support slow lookups?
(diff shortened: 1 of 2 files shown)


rousskov (dismissed):
Thank you. A lot more changes are needed to address similar cases (and also to remove unnecessary pointer use by non-storing code that requires object presence), but this is progress.

src/http.cc
```diff
@@ -720,7 +720,7 @@ HttpStateData::processReplyHeader()
     newrep->sources |= request->url.getScheme() == AnyP::PROTO_HTTPS ? Http::Message::srcHttps : Http::Message::srcHttp;
 
     if (newrep->sline.version.protocol == AnyP::PROTO_HTTP && Http::Is1xx(newrep->sline.status())) {
-        handle1xx(newrep);
+        handle1xx(newrep.getRaw());
```
yadij, line 723: Please include a change of the `handle1xx()` API. As you may see it already uses a `Pointer` internally and this is the only caller.

src/http.cc
```diff
@@ -733,7 +733,7 @@ HttpStateData::processReplyHeader()
     if (!peerSupportsConnectionPinning())
         request->flags.connectionAuthDisabled = true;
 
-    HttpReply *vrep = setVirginReply(newrep);
+    const auto vrep = setVirginReply(newrep.getRaw());
```
yadij, line 736: Please include updating the `setVirginReply()` API and its callers.  All suffer from this same "accidental destruction" flaw in regards to the `HttpReply` object they generate. The fix is the same, they all need to use a local `Pointer` instead of raw-pointer local variable.

yadij (changes requested):

src/http.cc
eduard-bagdasaryan: Done.

src/http.cc
eduard-bagdasaryan: This would require too much changes, I think...  The 'updating the setVirginReply() API' would require updating Client::theVirginReply to HttpReply::Pointer,  Client::virginReply() getter method and its callers. This, in turn, requires more changes, for otherwise we would need to 'getRaw()' in those contexts...
Note that currently there is no problem with setVirginReply(), because it locks theVirginReply before returning it. Callers may safely lock/unlock the raw pointer it returns.

eduard-bagdasaryan (commented):
Done (f7467c7).

src/http.cc
yadij: Hmm, the more the better. Though to make some progress I will not insist on this.

yadij (approved):

yadij:
Changed the title to clarify that this only affects the HTTP protocol. FTP still has this issue.

PR #1756, opened 2024-03-25
pinger: use std::chrono for timing

yadij:


src/icmp/Icmp4.cc
```diff
@@ -115,9 +116,10 @@ Icmp4::SendEcho(Ip::Address &to, int opcode, const char *payload, int len)
     // Construct ICMP packet data content
     echo = reinterpret_cast<icmpEchoData *>(reinterpret_cast<char *>(pkt) + sizeof(*icmp));
     echo->opcode = (unsigned char) opcode;
-    memcpy(&echo->tv, &current_time, sizeof(struct timeval));
+    const auto now = std::chrono::system_clock::now();
+    memcpy(&echo->tv, &now, sizeof(now));
```
kinkie, line 120: Are you sure? According to [cppreference](https://en.cppreference.com/w/cpp/chrono/system_clock/now), now() returns a `std::chrono::time_point`, I don't think it can be safely copied to a struct timeval like this

yadij: Oops, forgot to change `tv` type. Done now in 2dc4b66ed9

src/icmp/Icmp4.cc
```diff
@@ -115,9 +116,10 @@ Icmp4::SendEcho(Ip::Address &to, int opcode, const char *payload, int len)
...
 
-    icmp_pktsize += sizeof(struct timeval) + sizeof(char);
+    icmp_pktsize += sizeof(now) + sizeof(char);
```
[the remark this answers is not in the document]
yadij: Done in 2dc4b66ed9

src/icmp/Icmp4.cc
```diff
@@ -115,9 +114,9 @@ Icmp4::SendEcho(Ip::Address &to, int opcode, const char *payload, int len)
     // Construct ICMP packet data content
     echo = reinterpret_cast<icmpEchoData *>(reinterpret_cast<char *>(pkt) + sizeof(*icmp));
     echo->opcode = (unsigned char) opcode;
-    memcpy(&echo->tv, &current_time, sizeof(struct timeval));
+    echo->tv = std::chrono::system_clock::now();
 
-    icmp_pktsize += sizeof(struct timeval) + sizeof(char);
+    icmp_pktsize += sizeof(echo->tv) + sizeof(echo->opcode);
```
rousskov, line 119: Compared to `struct timeval`, the result of now() is more difficult to manipulate because we do not know the underlying type composition. Technically, the underlying non-trivial time_point type may not be trivially copyable and/or may create padding inside icmpEchoData.

I see two ways to address this concern:

1. If you insist on changing icmpEchoData::tv type, then this PR will need to assert that icmpEchoData is trivially copyable _and_ deal with possible _padding_ when computing related sizes. That new is_trivially_copyable_v assertion will probably [succeed](https://godbolt.org/z/qKb7q9jf5) on all platforms we care about (but that implementation-dependent outcome is not guaranteed by STL itself!).
2. Preserve icmpEchoData::tv type and convert to/from it as needed. I recommend this overall simpler option, especially if we want to minimize changes this PR has to make. The necessary conversion functions, if they do not exist already, will be eventually reused by other code.

rousskov (changes requested):

src/icmp/Icmp4.cc
yadij: AFAICT `time_point` are supposed to always be a numeric type, with some meta data about the clock it relates to. Added the static_assert in 6be60ef1f8

rousskov: > Added the static_assert in https://github.com/squid-cache/squid/commit/6be60ef1f8f2bb6b6dac6c4601a74b0e0783e8aa

It looks like you went for option 1. That option has "_and_ deal with possible padding when computing related sizes" part. There are also alignment concerns flagged in another change request. I doubt that, but if there are no possible padding and alignment issues in PR code, then we need to assert that explicitly.

FWIW, I still recommend option 2. Official code does not make it clear (and may have its own problems/violations), but we are dealing with serialization/de-serialization here. Doing that right is tricky. If we start changing underlying types, especially to STL-defined types with unknown internals, we become responsible for doing it right. Option 2 avoids all those problems.

src/icmp/Icmp.h
```diff
@@ -36,14 +38,15 @@ struct pingerReplyData {
...
+    int rtt = 0; // in milliseconds
     int hops = 0;
     int psize = 0;
     char payload[PINGER_PAYLOAD_SZ];
 };
 
 struct icmpEchoData {
-    struct timeval tv;
+    std::chrono::time_point<std::chrono::system_clock> tv;
+    static_assert(std::is_trivially_copyable_v<std::chrono::time_point<std::chrono::system_clock>>);
```
rousskov, line 49: Please move the assert to where the tested invariant is used.

If possible, I recommend testing the whole icmpEchoData structure instead of its individual data member for this assert. Doing so may also help with avoiding type duplication. If that duplication is still there, please name `std::chrono::time_point<std::chrono::system_clock>` type and reuse that name instead of spelling it out twice in this context.

src/icmp/Icmp4.cc
```diff
@@ -115,9 +114,9 @@ Icmp4::SendEcho(Ip::Address &to, int opcode, const char *payload, int len)
     // Construct ICMP packet data content
     echo = reinterpret_cast<icmpEchoData *>(reinterpret_cast<char *>(pkt) + sizeof(*icmp));
     echo->opcode = (unsigned char) opcode;
-    memcpy(&echo->tv, &current_time, sizeof(struct timeval));
+    echo->tv = std::chrono::system_clock::now();
```
rousskov, line 117: Official memcpy() code does not raise alignment concerns because it only uses the address of `echo->tv`. The new code does raise those concerns because it now treats memory in the middle of some I/O buffer as usable for time_point storage (an STL type with unknown (to us) alignment requirements).

rousskov (changes requested):

PR #1757, opened 2024-03-25
Portability: Drop unused GETTIMEOFDAY_NO_TZP

yadij:
Several helpers calling gettimeofday() have been widely used for years
and omitted this portability workaround without anyone noticing.

After commit 60bc6cf and 7072675 this define will no longer be used
by Squid at all.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2062,7 +2062,6 @@ SQUID_CHECK_SS_LEN_IN_SOCKADDR_STORAGE
 SQUID_CHECK_SIN_LEN_IN_SOCKADDR_IN
 
 dnl System-specific library modifications
-AH_TEMPLATE(GETTIMEOFDAY_NO_TZP,[Whether gettimeofday takes only one argument])
 AS_CASE(["$host"],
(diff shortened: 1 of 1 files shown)


kinkie (approved):
go for it

rousskov (changes requested):
AFAICT, initial GETTIMEOFDAY_NO_TZP support was added in 1996 commit 7149a49 to improve build on `*m88k*` and `*-*-solaris2.[0-4]` hosts. Hopefully, we do not want to support those and similar environments. The existence of conditionally-built helper and "make check" code that ignores GETTIMEOFDAY_NO_TZP is a good sign!

I support removal of this workaround. We can always add something like this later if needed.

However, this PR does not remove GETTIMEOFDAY_NO_TZP from the current master code. More work needs to be done in this PR, or we have to wait for the prerequisite PRs (at least) for this PR changes to become valid. I am OK with either strategy but assume that you do not want to add more changes to this PR; please ping me when those prerequisite PRs merge.

PR #1761, opened 2024-03-27
Update XMALLOC_STATISTICS timer to std::chrono

yadij:


src/stat.cc
```diff
@@ -764,8 +769,9 @@ void
 DumpMallocStatistics(StoreEntry* sentry)
 {
 #if XMALLOC_STATISTICS
-    xm_deltat = current_dtime - xm_time;
-    xm_time = current_dtime;
+    using namespace std::chrono;
+    xm_deltat = system_clock::now() - xm_time;
+    xm_time = system_clock::now();
```
rousskov, line 774: Squid is using "cached current time" by default (i.e. current_time and friends). This old code was using cached current time. Until a different approach is proposed and agreed on, please continue to use cached current time when upgrading to std::chrono API instead of switching to uncached now().

rousskov (changes requested):

src/stat.cc
```diff
@@ -421,8 +423,11 @@ info_get_mallstat(int size, int number, int oldnum, void *data)
     StoreEntry *sentry = (StoreEntry *)data;
 
 // format: "%12s %15s %6s %12s\n","Alloc Size","Count","Delta","Alloc/sec"
-    if (number > 0)
-        storeAppendPrintf(sentry, "%12d %15d %6d %.1f\n", size, number, number - oldnum, xdiv((number - oldnum), xm_deltat));
+    if (number > 0) {
+        const auto delta = number - oldnum;
+        const auto avg = xdiv(delta, (xm_deltat.count()/1000000.0))
```
rousskov, line 428: When upgrading to std::chrono, please avoid manual unit conversions like this. As you know from very recent experience, manual conversions are trivial to get wrong when writing and reviewing code. Use std::chrono APIs instead. For example, if you need seconds, convert to an std::chrono type representing seconds (including fractional seconds if needed).

In this particular case, it is may be best to _store_ the delta (i.e. xm_deltat) as fractional seconds, as the official code was doing. Is there some value in storing microseconds that will need to be converted to seconds on every use?

rousskov (changes requested):

PR #1764, opened 2024-03-30
Bug 5322: Do not leak HttpReply when checking http_reply_access

rousskov:
... as well as response_delay_pool and send_hit directives.

    auto check = clientAclChecklistCreate(...); // sets check->reply
    check->reply = reply; // self-assignment does nothing
    HTTPMSGLOCK(check->reply); // an unwanted/extra lock

When ACLFilledChecklist::reply is already set to X, resetting it to X
should not change HttpReply lock count, but some manual locking code did
not check that "already set" precondition and over-locked reply objects
set to ClientHttpRequest::al::reply by clientAclChecklistFill().

Current known leaks were probably introduced in 2021 commit e227da8 that
started supplying HttpReply to ACLChecklist in clientAclChecklistFill().
The added code locked the reply object correctly, but it was
incompatible with unconditional manual locks in three existing indirect
clientAclChecklistFill() callers (calling clientAclChecklistCreate()).

This change removes all known similar leaks and improves
ACLFilledChecklist API to prevent future similar leaks.

### DIFF
--- a/src/HttpHeaderTools.cc
+++ b/src/HttpHeaderTools.cc
@@ -298,12 +298,7 @@ httpHdrMangle(HttpHeaderEntry * e, HttpRequest * request, HeaderManglers *hms, c
     }
 
     ACLFilledChecklist checklist(hm->access_list, request, nullptr);
-
-    checklist.al = al;
-    if (al && al->reply) {
-        checklist.reply = al->reply.getRaw();
-        HTTPMSGLOCK(checklist.reply);
-    }
+    checklist.updateAle(al);
 
     // XXX: The two "It was denied" clauses below mishandle cases with no
     // matching rules, violating the "If no rules within the set have matching
@@ -498,12 +493,7 @@ void
 httpHdrAdd(HttpHeader *heads, HttpRequest *request, const AccessLogEntryPointer &al, HeaderWithAclList &headersAdd)
 {
     ACLFilledChecklist checklist(nullptr, request, nullptr);
-
-    checklist.al = al;
-    if (al && al->reply) {
-        checklist.reply = al->reply.getRaw();
-        HTTPMSGLOCK(checklist.reply);
-    }
+    checklist.updateAle(al);
 
     for (HeaderWithAclList::const_iterator hwa = headersAdd.begin(); hwa != headersAdd.end(); ++hwa) {
         if (!hwa->aclList || checklist.fastCheck(hwa->aclList).allowed()) {
--- a/src/acl/FilledChecklist.cc
+++ b/src/acl/FilledChecklist.cc
@@ -262,18 +262,19 @@ ACLFilledChecklist::setIdent(const char *ident)
 void
 ACLFilledChecklist::updateAle(const AccessLogEntry::Pointer &a)
 {
-    if (!al && a) {
-        al = a;
-        if (!request)
-            setRequest(a->request);
-        updateReply(a->reply);
-    }
+    if (!a)
+        return;
+
+    al = a; // could have been set already (to a different value)
+    if (!request)
+        setRequest(a->request);
+    updateReply(a->reply);
 }
 
 void
 ACLFilledChecklist::updateReply(const HttpReply::Pointer &r)
 {
     if (r)
-        reply_ = r; // may already be set to r
+        reply_ = r; // may already be set, including to r
 }
 
--- a/src/acl/FilledChecklist.cc
+++ b/src/acl/FilledChecklist.cc
@@ -26,7 +26,6 @@ CBDATA_CLASS_INIT(ACLFilledChecklist);
 
 ACLFilledChecklist::ACLFilledChecklist() :
     dst_rdns(nullptr),
-    reply (nullptr),
 #if USE_AUTH
     auth_user_request (nullptr),
 #endif
@@ -50,8 +49,6 @@ ACLFilledChecklist::~ACLFilledChecklist()
     assert (!asyncInProgress());
 
     safe_free(dst_rdns); // created by xstrdup().
-
-    HTTPMSGUNLOCK(reply);
 
     cbdataReferenceDone(conn_);
 
@@ -104,9 +101,9 @@ ACLFilledChecklist::verifyAle() const
         }
     }
 
-    if (reply && !al->reply) {
+    if (hasReply() && !al->reply) {
         showDebugWarning("HttpReply object");
-        al->reply = reply;
+        al->reply = reply_;
     }
 
 #if USE_IDENT
@@ -210,7 +207,6 @@ ACLFilledChecklist::markSourceDomainChecked()
  */
 ACLFilledChecklist::ACLFilledChecklist(const acl_access *A, HttpRequest *http_request, const char *ident):
     dst_rdns(nullptr),
-    reply(nullptr),
 #if USE_AUTH
     auth_user_request(nullptr),
 #endif
@@ -261,5 +257,23 @@ ACLFilledChecklist::setIdent(const char *ident)
 #else
     (void)ident;
 #endif
+}
+
+void
+ACLFilledChecklist::updateAle(const AccessLogEntry::Pointer &a)
+{
+    if (!al && a) {
+        al = a;
+        if (!request)
+            setRequest(a->request);
+        updateReply(a->reply);
+    }
+}
+
+void
+ACLFilledChecklist::updateReply(const HttpReply::Pointer &r)
+{
+    if (r)
+        reply_ = r; // may already be set to r
 }
 
--- a/src/acl/FilledChecklist.h
+++ b/src/acl/FilledChecklist.h
@@ -15,8 +15,8 @@
 #include "acl/forward.h"
 #include "base/CbcPointer.h"
 #include "error/forward.h"
-#include "HttpRequest.h"
 #include "HttpReply.h"
+#include "HttpRequest.h"
 #include "ip/Address.h"
 #if USE_AUTH
 #include "auth/UserRequest.h"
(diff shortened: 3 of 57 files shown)


src/HttpHeaderTools.cc
```diff
@@ -298,12 +298,7 @@ httpHdrMangle(HttpHeaderEntry * e, HttpRequest * request, HeaderManglers *hms, c
     }
 
     ACLFilledChecklist checklist(hm->access_list, request, nullptr);
-
-    checklist.al = al;
-    if (al && al->reply) {
-        checklist.reply = al->reply.getRaw();
-        HTTPMSGLOCK(checklist.reply);
-    }
+    checklist.updateAle(al);
```
rousskov, line 301: This PR uses new updateAle() in this and similar dangerous `HTTPMSGLOCK(checklist.reply)` places. After these changes, more than a dozen places in Squid code still set ALE::al directly instead of calling updateAle(). They all should eventually be upgraded to updateAle() or similar. To avoid mishaps, ALE::al member itself should made private, affecting dozens of readers. All those large/noisy changes are outside this bug-fixing PR scope.

src/http.cc
```diff
@@ -771,10 +771,9 @@ HttpStateData::handle1xx(const HttpReply::Pointer &reply)
     // check whether the 1xx response forwarding is allowed by squid.conf
     if (Config.accessList.reply) {
         ACLFilledChecklist ch(Config.accessList.reply, originalRequest().getRaw());
-        ch.al = fwd->al;
-        ch.reply = reply.getRaw();
+        ch.updateAle(fwd->al);
+        ch.updateReply(reply);
```
rousskov, line 775: In this and similar PR contexts, it is possible that ALE::reply exists and, hence, calling updateReply() is not necessary. However, investigating these hidden dependencies between ALE::reply and job reply fields requires significant effort (especially where a single job deals with multiple reply objects). I did not perform that investigation for this PR, preserving the original "update both ALE and reply" logic (except for trivial cases where the reply object was obviously taken from the ALE object in the same code snippet). I believe that investigation and related further improvements should be viewed as out of scope.

src/adaptation/icap/icap_log.cc
```diff
@@ -61,10 +61,7 @@ void icapLogLog(AccessLogEntry::Pointer &al)
 {
     if (IcapLogfileStatus == LOG_ENABLE) {
         ACLFilledChecklist checklist(nullptr, al->adapted_request, nullptr);
-        if (al->reply) {
-            checklist.reply = al->reply.getRaw();
-            HTTPMSGLOCK(checklist.reply);
-        }
+        checklist.updateAle(al);
```
rousskov, line 64: In places like this, it is possible that the new updateAle() will supply more information to the checklist than the original reply-setting code did. Hopefully, providing access to additional information is an improvement, but it may affect ACL-driven directives in some corner cases (e.g., a server_cert_fingerprint ACL might now match after getting access to ALE::hier.tcpServer that was not available to that ACL before this updateAle() call).

rousskov (commented):
The comments in this review just annotate the diff. They do not request any changes.

kinkie (commented):
Thanks for the extensive documentation.
How large is the leak? From the number and location of the removed locks, it seems rather large.

rousskov:
> How large is the leak? From the number and location of the removed locks, it seems rather large.

I do not know the exact answer to your question, but the leak ought to depend on the total response headers size (from hundreds of bytes to tens of kilobytes per transaction?). Squid is leaking HttpReply object (including its HttpHeader member containing header fields and various header parsing caches). In some environments, folks did notice significant (from their point of view) memory growth compared to v5 when evaluating v6, but I do not know specifics.

I speculate that this leak did not attract more attention because relatively few deployments use http_reply_access, send_hit, and response_delay_pool directives.

src/acl/FilledChecklist.h
```diff
@@ -55,14 +61,22 @@ class ACLFilledChecklist: public ACLChecklist
     /// set the client side FD
     void fd(int aDescriptor);
 
+    /// response added by updateReply()
+    /// \prec hasReply()
+    const HttpReply &reply() const { return *reply_; }
+
+    /// Remembers the given response (if it is not nil) or does nothing
+    /// (otherwise).
```
yadij, line 69: This call is implicitly part of reference counting and should clear anything already held when receiving a nil pointer. Otherwise we risk re-using checklists with wrong transaction pointers.

src/acl/HttpReqHeader.h
```diff
@@ -17,7 +17,7 @@ namespace Acl
 {
 
 /// a "req_header" ACL
-class HttpReqHeaderCheck: public ParameterizedNode< ACLData<HttpHeader*> >
+class HttpReqHeaderCheck: public ParameterizedNode< ACLData<const HttpHeader &> >
```
yadij, line 20: Please do not bundle `HttpHeader` type changes into a change documented as being for `HttpReply`. They are very different objects.

src/acl/FilledChecklist.h
```diff
@@ -43,6 +44,11 @@ class ACLFilledChecklist: public ACLChecklist
     /// configure rfc931 user identity for the first time
     void setIdent(const char *userIdentity);
 
+    /// Remembers the given ALE (if it is not nil) or does nothing (otherwise).
+    /// When (and only when) remembering ALE, populates other still-unset fields
+    /// with ALE-derived information, so that the caller does not have to.
+    void updateAle(const AccessLogEntry::Pointer &);
```
yadij, line 50: Please separate these `AccessLogEntry` changes from the `HttpReply` change.

src/client_side_reply.cc
```diff
@@ -1838,8 +1836,7 @@ clientReplyContext::processReplyAccess ()
     /** Process http_reply_access lists */
     ACLFilledChecklist *replyChecklist =
         clientAclChecklistCreate(Config.accessList.reply, http);
-    replyChecklist->reply = reply;
-    HTTPMSGLOCK(replyChecklist->reply);
+    replyChecklist->updateReply(reply);
```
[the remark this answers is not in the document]
yadij: Hmm, something seems wrong about that statement. These lines were correctly locking after setting the pointer.

Perhapse you meant that `clientAclChecklistCreate()` already set some value in reply that was not being correctly `HTTPMSGUNLOCK()`ed ?

src/acl/FilledChecklist.cc
```diff
@@ -263,3 +259,22 @@ ACLFilledChecklist::setIdent(const char *ident)
...
+    if (!request)
+        setRequest(a->request);
+    updateReply(a->reply);
+}
+
+void
+ACLFilledChecklist::updateReply(const HttpReply::Pointer &r)
+{
+    if (r)
+        reply_ = r; // may already be set, including to r
```
yadij, line 278: This breaks refcounting which already handles the nil check properly.
```suggestion
    reply_ = r; // refcounted
```

yadij (changes requested):

yadij:
> Thanks for the extensive documentation. How large is the leak? From the number and location of the removed locks, it seems rather large.

In addition to the thinks @rousskov already mentioned, the `HttpReply` also significantly stores a reference lock on the relevant `HttpRequest` and all its associated headers and caches etc. Which more than doubles the amount of memory used by just the reply message.

So I would expect the leak to be rather significant, if rarely encountered.

src/client_side_reply.cc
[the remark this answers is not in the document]
rousskov: > These lines were correctly locking after setting the pointer.

No, they were not. As I tried to explain in the PR description, this code _looks_ OK, but it is buggy:

```C++
foo = bar;
HTTPMSGLOCK(foo); // BUG: Overlocking when foo was bar before the assignment
```

> Perhapse you meant that clientAclChecklistCreate() already set some value in reply that was not being correctly HTTPMSGUNLOCK()ed ?

I did not mean to say that (and did mean to say what I said). However, if we were stuck with manual locks/unlocks, then one could fix this bug by carefully unlocking replyChecklist->reply while resetting it to the same value and locking it again. The end result would be the same -- the final value of reference counter will decrease by one in the problematic case of self-assignment (it will stay the same/unchanged instead of being increased by one as in the current official code):

```C++
// one possible implementation that does not assume anything about prior values:
LOCK(bar); // in case foo is bar
UNLOCK(foo);
foo = bar;
```

Fortunately smart pointers do all this work for us.

rousskov:
> the `HttpReply` also significantly stores a reference lock on the relevant `HttpRequest`

Where? I do not see HttpReply storing an HttpRequest reference;  mgr:mem snapshots I have seen for this leak did not show an increase in HttpRequest objects.

src/acl/HttpHeaderData.cc
```diff
@@ -36,20 +36,17 @@ ACLHTTPHeaderData::~ACLHTTPHeaderData()
 }
 
 bool
-ACLHTTPHeaderData::match(HttpHeader* hdr)
+ACLHTTPHeaderData::match(const HttpHeader &hdr)
```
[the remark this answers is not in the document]
rousskov: And those PR 1771 changes are now merged from master into this PR branch (commit 3f2e468c). I am removing S-waiting-for-PR label. This PR continues to wait for @yadij review.

rousskov:
@yadij, please unblock this PR. It has no (known to me) problems and has been waiting for your review for a month.

rousskov:
I have added "Bug 5322" prefix to the PR title. I am not sure this PR fixes _all_ leaks reported (by various people) in that bug report, but it does fix [some](https://bugs.squid-cache.org/show_bug.cgi?id=5322#c19). I do not insist on this prefix; please remove it if you dislike it.

kinkie (approved):
As it's been successfully run-tested, I'm approving this

yadij (approved):

PR #1766, opened 2024-03-31
Upgrade Acl::Node::name to SBuf; remove AclMatchedName global

rousskov:
AclMatchedName global has been localized into a regular Acl::Answer data
member (Acl::Answer maintains the result of ACLChecklist evaluations).
This long overdue change resolves an old TODO and XXXs, paving the way
for Acl::Node reference counting.

No significant functionality changes are expected, but it is possible
that some deny_info configurations will now be handled better in
reconfiguration corner cases (because Squid no longer forgets the name
of the last checked ACL when a slow ACL check crosses reconfiguration
barrier).

Most of these changes are performance-neutral or -positive because they
eliminate or reduce memory allocations and associated name copying (and
more reduction will become possible after upgrading squid.conf parsers
to use SBuf). This change adds SBuf object copies when Acl::Answer is
propagated to ACLCB callbacks, but those read-only copies are cheap.

Also renamed and polished aclGetDenyInfoPage() because we had to update
its parameter type (to supply the last evaluated ACL). All callers were
also supplying the same first argument (that is unlikely to change in
the foreseeable future); that argument is now gone. We  did not fix the
redirect_allowed name and debugs(): Fixing that behavior deserves a
dedicated change.

Also polished legacy aclIsProxyAuth() profile and description because we
have to change the parameter type (to supply the last evaluated ACL).

Also removed 63-character aclname parameter limit for acl directives.

### DIFF
--- a/src/acl/Acl.cc
+++ b/src/acl/Acl.cc
@@ -27,7 +27,7 @@
 #include <algorithm>
 #include <map>
 
-const char *AclMatchedName = nullptr;
+std::optional<SBuf> AclMatchedName;
 
 namespace Acl {
 
@@ -144,12 +144,12 @@ void Acl::Node::operator delete(void *)
 }
 
 Acl::Node *
-Acl::Node::FindByName(const char *name)
+Acl::Node::FindByName(const SBuf &name)
 {
     debugs(28, 9, "name=" << name);
 
     for (auto *a = Config.aclList; a; a = a->next)
-        if (!strcasecmp(a->name, name))
+        if (!a->name.caseCmp(name))
             return a;
 
     debugs(28, 9, "found no match");
@@ -161,9 +161,7 @@ Acl::Node::Node() :
     cfgline(nullptr),
     next(nullptr),
     registered(false)
-{
-    *name = 0;
-}
+{}
 
 bool
 Acl::Node::valid() const
@@ -179,7 +177,7 @@ Acl::Node::matches(ACLChecklist *checklist) const
     // XXX: AclMatchedName does not contain a matched ACL name when the acl
     // does not match. It contains the last (usually leaf) ACL name checked
     // (or is NULL if no ACLs were checked).
-    AclMatchedName = name;
+    AclMatchedName = std::make_optional(name);
 
     int result = 0;
     if (!checklist->hasAle() && requiresAle()) {
@@ -208,9 +206,9 @@ Acl::Node::matches(ACLChecklist *checklist) const
 void
 Acl::Node::context(const char *aName, const char *aCfgLine)
 {
-    name[0] = '\0';
+    name.clear();
     if (aName)
-        xstrncpy(name, aName, ACL_NAME_SZ-1);
+        name = SBuf(aName);
     safe_free(cfgline);
     if (aCfgLine)
         cfgline = xstrdup(aCfgLine);
@@ -239,13 +237,13 @@ Acl::Node::ParseAclLine(ConfigParser &parser, Node ** head)
 
     SBuf aclname(t);
     CallParser(ParsingContext::Pointer::Make(aclname), [&] {
-        ParseNamed(parser, head, aclname.c_str()); // TODO: Convert Node::name to SBuf
+        ParseNamed(parser, head, aclname);
     });
 }
 
 /// parses acl directive parts that follow aclname
 void
-Acl::Node::ParseNamed(ConfigParser &parser, Node ** const head, const char * const aclname)
+Acl::Node::ParseNamed(ConfigParser &parser, Node ** const head, const SBuf &aclname)
 {
     /* snarf the ACL type */
     const char *theType;
@@ -278,7 +276,7 @@ Acl::Node::ParseNamed(ConfigParser &parser, Node ** const head, const char * con
         }
         theType = "localport";
         debugs(28, DBG_IMPORTANT, "WARNING: UPGRADE: ACL 'myport' type has been renamed to 'localport' and matches the port the client connected to.");
-    } else if (strcmp(theType, "proto") == 0 && strcmp(aclname, "manager") == 0) {
+    } else if (strcmp(theType, "proto") == 0 && aclname.cmp("manager") == 0) {
         // ACL manager is now a built-in and has a different type.
         debugs(28, DBG_PARSE_NOTE(DBG_IMPORTANT), "WARNING: UPGRADE: ACL 'manager' is now a built-in ACL. Remove it from your config file.");
         return; // ignore the line
@@ -292,7 +290,7 @@ Acl::Node::ParseNamed(ConfigParser &parser, Node ** const head, const char * con
     if ((A = FindByName(aclname)) == nullptr) {
         debugs(28, 3, "aclParseAclLine: Creating ACL '" << aclname << "'");
         A = Acl::Make(theType);
-        A->context(aclname, config_input_line);
+        A->context(SBuf(aclname).c_str(), config_input_line);
         new_acl = 1;
     } else {
         if (strcmp (A->typeString(),theType) ) {
@@ -446,7 +444,7 @@ Acl::Node::~Node()
 {
     debugs(28, 3, "freeing ACL " << name);
     safe_free(cfgline);
-    AclMatchedName = nullptr; // in case it was pointing to our name
+    AclMatchedName.reset(); // in case it was pointing to our name
 }
 
 void
--- a/src/acl/Acl.cc
+++ b/src/acl/Acl.cc
@@ -27,8 +27,6 @@
 #include <algorithm>
 #include <map>
 
-std::optional<SBuf> AclMatchedName;
-
 namespace Acl {
 
 /// Acl::Node type name comparison functor
@@ -115,6 +113,13 @@ Acl::SetKey(SBuf &keyStorage, const char *keyParameterName, const char *newKey)
                         Here());
 }
 
+SBuf
+Acl::Answer::lastCheckDescription() const
+{
+    static const SBuf none("[no-ACL]");
+    return lastCheckedName.value_or(none);
+}
+
 /* Acl::ParsingContext */
 
 ScopedId
@@ -174,10 +179,7 @@ Acl::Node::matches(ACLChecklist *checklist) const
 {
     debugs(28, 5, "checking " << name);
 
-    // XXX: AclMatchedName does not contain a matched ACL name when the acl
-    // does not match. It contains the last (usually leaf) ACL name checked
-    // (or is NULL if no ACLs were checked).
-    AclMatchedName = std::make_optional(name);
+    checklist->setLastCheckedName(name);
 
     int result = 0;
     if (!checklist->hasAle() && requiresAle()) {
@@ -442,7 +444,6 @@ Acl::Node::~Node()
 {
     debugs(28, 3, "freeing ACL " << name);
     safe_free(cfgline);
-    AclMatchedName.reset(); // in case it was pointing to our name
 }
 
 void
--- a/src/acl/Acl.cc
+++ b/src/acl/Acl.cc
@@ -149,18 +149,32 @@ void Acl::Node::operator delete(void *)
     fatal ("unusable Acl::Node::delete");
 }
 
+/// implements both Acl::Node::FindByName() variations
+template <typename SBufOrCString>
+static Acl::Node *
+FindByNameT(const SBufOrCString name)
+{
+    for (auto a = Config.aclList; a; a = a->next) {
+        if (a->name.caseCmp(name) == 0) {
+            debugs(28, 8, "found " << a->name);
+            return a;
+        }
+    }
+
+    debugs(28, 8, "cannot find " << name);
+    return nullptr;
+}
+
 Acl::Node *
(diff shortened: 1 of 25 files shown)


src/acl/Node.h
```diff
@@ -30,14 +31,19 @@ class Node
 
     static void ParseAclLine(ConfigParser &parser, Acl::Node **head);
     static void Initialize();
+
+    /// A configured ACL with a given name or nil.
+    static Acl::Node *FindByName(const SBuf &);
+    /// \copydoc FindByName()
+    /// \deprecated Use to avoid performance regressions; remove with the last caller.
     static Acl::Node *FindByName(const char *name);
```
rousskov, line 39: This legacy function has one remaining caller: Acl::InnerNode::lineParse(). It will disappear as we upgrade squid.conf parsing code to use SBuf, but I think we should keep it for now because the remaining caller may be called frequently in some large configurations, and the corresponding to-SBuf conversion overheads (that would be added if this PR removes this function) may be noticed.

I adjusted FindByName() implementation so that the two FindByName() functions declared above do not duplicate each other code.

src/acl/AllOf.cc
```diff
@@ -56,13 +57,8 @@ Acl::AllOf::parse()
...
         // this acl saw a single line before; create a new OR inner node
 
-        MemBuf wholeCtx;
-        wholeCtx.init();
-        wholeCtx.appendf("(%s lines)", name);
-        wholeCtx.terminate();
-
         Acl::OrNode *newWhole = new Acl::OrNode;
-        newWhole->context(wholeCtx.content(), oldNode->cfgline);
+        newWhole->context(ToSBuf('(', name, " lines)"), oldNode->cfgline);
```
rousskov, line 61: This and similar name-generating lines are a bit difficult to read and write correctly. I am pretty sure we should add a couple of functions or a simple class that encapsulates this logic. However, IIRC, there are related XXXs/bugs that claim that wrong generated names are printed in mgr:config reporting or some other context. I think improving this generation code (and addressing those concerns or ruling them out of scope) deserves a dedicated PR.

For now, we are preserving the current official format for these generated/implicit ACL names and just converting them to SBuf (instead of using MemBuf to assemble them).

src/icp_v2.cc
```diff
@@ -443,14 +441,20 @@ icpDenyAccess(Ip::Address &from, char *url, int reqnum, int fd)
...
 
     ACLFilledChecklist checklist(Config.accessList.icp, icp_request, nullptr);
     checklist.src_addr = from;
     checklist.my_addr.setNoAddr();
-    return checklist.fastCheck().allowed();
+    const auto &answer = checklist.fastCheck();
+    if (answer.allowed())
+        return true;
+
+    debugs(12, 2, "Access Denied for " << from << " by " << answer.lastCheckDescription() << ".");
```
rousskov, line 456: This small icpAccessAllowed() refactoring preserves icpDenyAccess()-related behavior, including debugging. We refactored so that we do not have to pass `answer` to icpDenyAccess(). Ideally, the two functions should be merged instead of relying on icpAccessAllowed() caller to call icpDenyAccess() after icpAccessAllowed() returns false, but doing that right requires a lot more changes, including caller changes that we want to avoid in this PR.

rousskov (commented):
A portion of this PR is naturally similar (and, in parts, identical) to old and neglected #712, but this PR uses the need to change AclMatchedName type (and, hence, most of the code that uses that global) to completely remove that problematic global. This PR also takes advantage of std::optional that was not available to us in 2020 (#712 publication time) and addresses (or avoids) my [concerns](https://github.com/squid-cache/squid/pull/712#pullrequestreview-468455967) about some of that old PR changes.

src/acl/Checklist.h
```diff
@@ -152,9 +154,12 @@ class ACLChecklist
         return old;
     }
 
+    /// remember the name of the last ACL being evaluated
+    void setLastCheckedName(const SBuf &name) { lastCheckedName_ = name; }
+
 private:
     /// Calls non-blocking check callback with the answer and destroys self.
-    void checkCallback(Acl::Answer answer);
+    void checkCallback(const Acl::Answer &answer);
```
kinkie, line 162: This feels out of scope for the current PR

src/cache_cf.cc
```diff
@@ -4726,7 +4723,8 @@ static void parse_ftp_epsv(acl_access **ftp_epsv)
         *ftp_epsv = nullptr;
 
         if (ftpEpsvDeprecatedAction == Acl::Answer(ACCESS_DENIED)) {
-            if (auto *a = Acl::Node::FindByName("all"))
+            static const auto all = new SBuf("all");
```
kinkie, line 4726: What is the benefit of intentionally leaking a SBuf here versus
```
static const SBuf all("all");
```
?

This use case seems quite different from the NoNewGlobals one to me

kinkie:
Looks good to me, besides a small nit

src/acl/Checklist.h
rousskov: IMO, it is in scope because this PR makes Acl::Answer copies a bit more costly; passing Acl::Answer by value is no longer the best or preferred default way. This change avoids one such copy. There is more information about it in PR description:

> Most of these changes are performance-neutral or -positive because they eliminate or reduce memory allocations and associated name copying (and more reduction will become possible after upgrading squid.conf parsers to use SBuf). This change adds SBuf object copies when Acl::Answer is propagated to ACLCB callbacks, but those read-only copies are cheap.

If you insist, we will remove this change from this PR.

src/cache_cf.cc
rousskov: > What is the benefit of intentionally leaking a SBuf here versus

[ There is no real memory leak here, but I am ignoring that part of this question. ]

Same benefit as in NoNewGlobals: Unlike a [static local](https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables) SBuf, the value of this POD variable is guaranteed to exist during destruction at program exit.

> This use case seems quite different from the NoNewGlobals one to me

In what way? The concerns here are essentially the same AFAICT. You can even imagine this code as the result of future/hypothetical `Acl::All().name` inlining.

Needless to say, we can spend a lot of time (in aggregate) debating whether specific functions might be called during at-exit destruction. One of the NoNewGlobals policy goals is to _avoid_ such debates in all contexts (at the expense of a few extra pointer dereferences in some contexts).

If you insist, we can use a local static SBuf here, but that will create a dangerous precedent by going against NoNewGlobals intent/goals.

src/acl/Node.h
```diff
@@ -67,7 +73,12 @@ class Node
     /// printed parameters are collected from all same-name "acl" directives.
     void dumpWhole(const char *directiveName, std::ostream &);
 
-    char name[ACL_NAME_SZ];
```
yadij, line 70: Please check for other uses of `ACL_NAME_SZ` limitation and remove the macro as part of this.

src/acl/InnerNode.cc
```diff
@@ -82,7 +82,7 @@ Acl::InnerNode::dump() const
 {
     SBufList rv;
     for (Nodes::const_iterator i = nodes.begin(); i != nodes.end(); ++i)
-        rv.push_back(SBuf((*i)->name));
+        rv.push_back((*i)->name);
```
yadij, line 85: AFAIK this should be a ranged-for now, which simplifies this syntax quite a lot. Please consider upgrading this fully.

src/acl/Checklist.h
```diff
@@ -209,6 +214,9 @@ class ACLChecklist
     std::stack<Breadcrumb> matchPath;
     /// the list of actions which must ignored during acl checks
     std::vector<Acl::Answer> bannedActions_;
+
+    /// the name of the last evaluated ACL (if any ACLs were evaluated)
+    std::optional<SBuf> lastCheckedName_;
```
yadij, line 219: The `SBuf` object is essentially a smart pointer, so we should not use `std::optional` raw-pointers to them. Please try to replace this and other uses of `std::optional` with just `SBuf` variables (or `const SBuf &` in function parameters).

FWIW; Since all valid ACLs are named, the `SBuf::isEmpty()` method can be used to determine whether the name was set or not.

yadij (changes requested):

src/acl/Acl.cc
```diff
@@ -143,27 +149,39 @@ void Acl::Node::operator delete(void *)
...
-    debugs(28, 9, "found no match");
-
+    debugs(28, 8, "cannot find " << name);
     return nullptr;
 }
 
+Acl::Node *
+Acl::Node::FindByName(const SBuf &name)
+{
+    return FindByNameT<const SBuf &>(name);
```
kinkie, line 171: Why do we need to specify the template argument here? Can't we just rely on template argument deduction?
```suggestion
    return FindByNameT(name);
```

src/acl/Acl.cc
```diff
@@ -143,27 +149,39 @@ void Acl::Node::operator delete(void *)
...
+}
+
+Acl::Node *
+Acl::Node::FindByName(const char *name)
+{
+    return FindByNameT<const char *>(name);
```
kinkie, line 177: Just like above:
```suggestion
    return FindByNameT(name);
```

src/acl/Acl.cc
rousskov: > Why do we need to specify the template argument here?

We do not have to. I think being explicit adds clarity in this case. However, I removed it in hope to make progress with this [head-of-line](https://en.wikipedia.org/wiki/Head-of-line_blocking) PR.

> Can't we just rely on template argument deduction?

We can. Done in commit 01a25ae.

src/acl/Checklist.h
rousskov: > If you insist, we will remove this change from this PR.

@kinkie, do you insist on this removal?

src/acl/Checklist.h
rousskov: > The `SBuf` object is essentially a smart pointer

I strongly disagree. In this discussion context, SBuf semantics is a container of characters, not a pointer. SBuf internals (that happen to use a pointer) are irrelevant.

We have abused empty SBuf to mean "no SBuf" in some contexts, mostly before we had access to std::optional, but there is no good reason to introduce the associated risks here and now.

> so we should not use `std::optional` raw-pointers to them. 

std::optional is not a raw pointer. At some level of abstraction, it can be viewed as a smart pointer. Overall, std::optional is much better than "some containers have special meaning" hacks.

> Since all valid ACLs are named, the `SBuf::isEmpty()` method can be used to determine whether the name was set or not.

Yes, the code will "work" if we assign an empty SBuf a special meaning of "unknown name", but there is no good reason for such error-prone hacks here.

> Please try to replace this and other uses of `std::optional` with just `SBuf`

I will not. Please withdraw this change request.

src/acl/InnerNode.cc
rousskov: Done to avoid arguing (commit 216013e).

src/acl/Node.h
rousskov: Done to avoid arguing (commit cb4a4bb).

src/acl/Acl.cc
rousskov: Done in commit 01a25ae.

src/acl/Checklist.h
kinkie: I don't insist

kinkie (approved):
Good enough for me. The std::optional<SBuf> vs SBuf::isEmpty() conversation is best suited elsewhere

src/acl/Checklist.cc
```diff
@@ -60,6 +60,7 @@ ACLChecklist::markFinished(const Acl::Answer &finalAnswer, const char *reason)
     assert (!finished() && !asyncInProgress());
     finished_ = true;
     answer_ = finalAnswer;
+    answer_.lastCheckedName = lastCheckedName_;
     debugs(28, 3, this << " answer " << answer_ << " for " << reason);
```
rousskov, line 64: We thought of a new way to test PR changes. These new tests did not uncover PR bugs, but we have discovered that ACLChecklist is missing a couple of markFinished() calls. Adding those calls will improve reconfiguration support[^1], but the best solution is not to sprinkle ACLChecklist with more markFinished() calls. Instead, we should refactor a bit so that we will never forget to make those calls again. That non-trivial refactoring is outside this PR scope. Once this PR lands, we will follow up with a PR dedicated to that ACLChecklist refactoring.

[^1]: Reconfiguration support will improve by delivering the last evaluated ACL name to nonBlockingCheck() callbacks. Today, that name is missing in the official code because that code is missing markFinished() calls _and_ because of the various AclMatchedName-linked issues addressed in this PR. That name remains missing in PR code because those markFinished() calls remain missing in PR code. Again, we will address the underlying problem in a dedicated followup PR to keep this PR focused on Acl::Node::name/AclMatchedName upgrade/fixes.

yadij (approved):
Thank you for resolving most of my requests.

src/acl/Checklist.cc
eduard-bagdasaryan: Posted this follow-up [PR](https://github.com/squid-cache/squid/pull/1809).

PR #1767, opened 2024-04-01
Bug 5360: FwdState::noteDestinationsEnd() assertion "err"

rousskov:
FATAL: assertion failed: FwdState.cc:660: "err"

When FwdState decides to re-forward a request, it forgets the original
response[^1] but does not create an error object. Since commit e2bbd3b,
FwdState::noteDestinationsEnd() correctly assumed that we only idly wait
for additional destinations after encountering a problem, but
incorrectly asserted that past problems imply error object existence.

Now Squid generates an HTTP 502 (Bad Gateway) response while setting
%err_code/%err_detail to ERR_CANNOT_FORWARD/REFORWARD_TO_NONE.

TunnelStateData::noteDestinationsEnd() code is similar, but it probably
does not suffer from the same bug because an error object is created
before every retryOrBail() call, and there is no re-forwarding code that
forgets an HTTP error response without creating an error. Those
invariants are not well tracked, and this change mimics FwdState changes
in TunnelStateData just in case and to keep the two methods in sync. In
those hypothetical cases, ERR_CANNOT_FORWARD/RETRY_TO_NONE is logged.

[^1]: Long-term we probably want to preserve that original response so
that we do not have to replace it with a generated error, but doing so
requires significant refactoring and is outside this minimal fix scope.

### DIFF
--- a/src/tunnel.cc
+++ b/src/tunnel.cc
@@ -1364,7 +1364,15 @@ TunnelStateData::noteDestinationsEnd(ErrorState *selectionError)
     }
 
     // destinationsFound, but none of them worked, and we were waiting for more
-    assert(savedError);
+    debugs(17, 7, "no more destinations to try after " << n_tries << " failed attempts");
+    if (!savedError) {
+        // retryOrBail() must be preceded by saveError(), but in case we forgot:
+        const auto finalError = new ErrorState(ERR_CANNOT_FORWARD, Http::scBadGateway, request.getRaw(), al);
+        static const auto d = MakeNamedErrorDetail("RETRY_TO_NONE");
+        finalError->detailError(d);
+        saveError(finalError);
+    } // else use actual error from last forwarding attempt
+
     // XXX: Honor clientExpectsConnectResponse() before replying.
     sendError(savedError, "all found paths have failed");
 }
--- a/src/FwdState.cc
+++ b/src/FwdState.cc
@@ -647,7 +647,13 @@ FwdState::noteDestinationsEnd(ErrorState *selectionError)
     }
 
     // destinationsFound, but none of them worked, and we were waiting for more
-    assert(err);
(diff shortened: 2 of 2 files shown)


src/tunnel.cc
```diff
@@ -1364,7 +1364,15 @@ TunnelStateData::noteDestinationsEnd(ErrorState *selectionError)
     }
 
     // destinationsFound, but none of them worked, and we were waiting for more
-    assert(savedError);
+    debugs(17, 7, "no more destinations to try after " << n_tries << " failed attempts");
+    if (!savedError) {
+        // retryOrBail() must be preceded by saveError(), but in case we forgot:
```
kinkie, line 1369: This sounds like a non-fatal error condition. Might it be worthwhile to add a debugs() to aid traceability?

kinkie (approved):
User run-tested

src/tunnel.cc
rousskov: > Might it be worthwhile to add a debugs() to aid traceability?

Yes, but I would not do it _here_: We know we are in this section of the code due to debugs() immediately above this `if` statement. We know we are in this `if` statement because of fail() debugging triggered from inside of this `if` statement:

```
| 44,2| peer_select.cc(480) resolveSelected: PeerSelector1 found all 1 destinations for ...
| 17,7| FwdState.cc(650) noteDestinationsEnd: no more destinations to try after 1 failed attempts
| 17,3| FwdState.cc(461) fail: ERR_CANNOT_FORWARD "Bad Gateway"
| 17,3| FwdState.cc(185) stopAndDestroy: for all found paths have failed
```

We should add debugging to ErrorState constructor so that we do not have to wait for fail(). Done in #1768.

yadij (approved):

squidadm:
queued for backport to v6

PR #1768, opened 2024-04-01
Improve ErrorState debugging

rousskov:
In triage, it is often useful to know that ErrorState was created _when_
it was created rather than waiting for errorAppendEntry() or errorSend()
debugging, especially if those error handling stages are not reached.

Also report HTTP status code of the error response (when available).


src/FwdState.cc
```diff
@@ -458,7 +458,7 @@ FwdState::useDestinations()
 void
 FwdState::fail(ErrorState * errorState)
 {
-    debugs(17, 3, err_type_str[errorState->type] << " \"" << Http::StatusCodeString(errorState->httpStatus) << "\"\n\t" << entry->url());
+    debugs(17, 3, errorState << "; was: " << err);
```
rousskov, line 461: This change reduces code duplication, adds more debugging information for complex FwdState retries, synchronizes affected FwdState and TunnelStateData code, and also removes unnecessary (and excessive) multiline reporting for a level-3 debug message that complicates extracting kid-specific cache.log lines.

src/errorpage.cc
```diff
@@ -711,6 +723,7 @@ ErrorState::ErrorState(HttpRequest * req, HttpReply *errorReply, const AccessLog
     Must(errorReply);
     response_ = errorReply;
     httpStatus = errorReply->sline.status();
+    debugs(4, 3, "relaying " << *this);
```
rousskov, line 726: We could merge the two added debugs() lines above into the shared (private) ErrorState constructor instead, but I do not think we should because

* it is useful to distinguish the two ErrorState construction cases; and
* the shared constructor does not have access to HTTP status code information unless we add more code).
 
The unique (within ErrorState constructor debugging contexts) "relaying" prefix helps with the "distinguish" part (without having to map line numbers in cache.log messages to version-dependent Squid code).

rousskov (commented):
This PR was triggered by https://github.com/squid-cache/squid/pull/1767#discussion_r1545923891

Sample output:

```
| 4,3| errorpage.cc(712) ErrorState: ERR_CANNOT_FORWARD/http_status=502
| 17,3| FwdState.cc(461) fail: ERR_CANNOT_FORWARD/http_status=502; was: [nil]
| 4,4| errorpage.cc(739) errorAppendEntry: storing ERR_CANNOT_FORWARD/http_status=502 in e:=p2XIHWV/0x55c661abe320*4
| 4,2| errorpage.cc(1411) buildBody: No existing error page language negotiated for ERR_CANNOT_FORWARD/http_status=502. Using default error file.
```

The comments in this review annotate the diff. They do not request any code changes.

src/errorpage.cc
```diff
@@ -675,6 +676,16 @@ errorPageName(int pageId)
     return "ERR_UNKNOWN";   /* should not happen */
 }
 
+/// compactly prints top-level ErrorState information (for debugging)
+static std::ostream &
+operator <<(std::ostream &os, const ErrorState &err)
+{
+    os << errorPageName(err.type);
+    if (err.httpStatus != Http::scNone)
+        os << "/http_status=" << err.httpStatus;
```
yadij, line 685: Please use "comma SP" separator here, as is typical for lists of details about something.

```suggestion
        os << ", http_status=" << err.httpStatus;
```

src/errorpage.cc
```diff
@@ -698,6 +709,7 @@ ErrorState::ErrorState(err_type t, Http::StatusCode status, HttpRequest * req, c
         httpStatus = ErrorDynamicPages[page_id - ERR_MAX]->page_redirect;
     else
         httpStatus = status;
+    debugs(4, 3, *this);
```
yadij, line 712: Please use syntax required by the `find-alive.pl` script.
```suggestion
    debugs(4, 3, "ErrorState constructed, this=" << this, << ", " << *this);
```

This should also be placed at the end of the constructor to ease future output of the other stored details.

yadij (changes requested):
Just some minor polish.

src/errorpage.cc
rousskov: The formatting here follows `%err_code/%err_detail` formatting model that many admins use (and I recommend for human-read logs).

IMO, the information printed here is not really a _list_; the printed items do not belong to the same "level" where enumeration would be natural. Instead, it is a primary `ERR_FOO` entry with optional/secondary detailing. Rendering this information as a list and using whitespace (without surrounding the whole things with curly braces) will make some likely debugging use cases (that contain commas and spaces for higher-level reasons) more difficult to interpret.

Please keep this summary output looking as a single "field" or "token".

src/errorpage.cc
rousskov: `find-alive.pl` hints added to public ErrorState constructors and destructor in branch commit 5f65f2f.

src/errorpage.cc
yadij: Admin reading `cache.log` will be expecting to see the same syntax found in all the surrounding lines.

rousskov: > Admin reading `cache.log` will be expecting to see the same syntax found in all the surrounding lines. Which is key=value separated by comma.

In cache.log context, we do not have a standard format for this kind of information, and I do not think we should start defining one here. There are already many ways similar information is logged, including lists, structures, compact tokens, and even multiline records. Folks seeing these values are unlikely to misunderstand them because of this particular format choice (a format that fits this data better than a list would and that will look familiar to many admins that log `%err_code/%err_detail`).

PR #1771, opened 2024-04-02
Fix const-correctness of ACLHTTPHeaderData::match() parameter

rousskov:
ACLHTTPHeaderData::match() required a pointer to non-const HttpHeader
but does not (and should not) modify the supplied HttpHeader.

Also removed support for nil HttpHeader in that method. All callers
already require HttpHeader presence. If that changes, it is the _caller_
that should decide what HttpHeader absence means (match, mismatch,
exception/dunno, etc.); ACLHTTPHeaderData does not have enough
information to make the right choice and, hence, should not be asked to
choose.

Also polished related #includes to remove unnecessary ones.

### DIFF
--- a/src/acl/HttpHeaderData.cc
+++ b/src/acl/HttpHeaderData.cc
@@ -36,20 +36,17 @@ ACLHTTPHeaderData::~ACLHTTPHeaderData()
 }
 
 bool
-ACLHTTPHeaderData::match(HttpHeader* hdr)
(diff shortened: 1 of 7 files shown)


rousskov:
This prerequisite PR was created per Amos [request](https://github.com/squid-cache/squid/pull/1764#discussion_r1546914886) in #1764.

kinkie (approved):

